PR #1805, opened 2019-08-16
Change "Open cohort in Data Explorer" to new way of opening tools

melissachang:
FYI @seanhorgan @jeromechadel 

Before:

After:

Don't show "Data Explorer" for entities that aren't cohorts:

For cohorts, only show "Data Explorer" and don't show IGV or Workflow. 1 cohort selected:

2 cohorts selected:

(I disabled the button in this case but you can't tell from the looks of it.)

Also fix this harmless but confusing console error by wrapping `iframeResizer` in `onLoad()`.

```
Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('https://test-data-explorer.appspot.com') does not match the recipient window's origin ('http://localhost:3000').
```

PR #1881, opened 2019-10-04
Add "open cohort in notebook" tool [SATURN-1168]

melissachang:
@jeromechadel approved this flow. Lou said I could use this Jupyter icon. (I ran `npm run optimize-image-svgs`.)

FYI @seanhorgan @wnojopra @mbookman 

Notebook tool button:

Tooltip is "Create a Python 2 or 3 notebook with this notebook". I wanted to clarify that this works for Python and not R.

After you click the button, notebook creation dialog:

Then you're taken to notebook. The [notebook is here](https://app.terra.bio/#workspaces/fc-product-demo/melchang%20test/notebooks/launch/cohort-notebook.ipynb) if anyone wants to see it. It includes a visualization.

Tested:
- Notebook name dialog:
  - Error shown if you type in existing name
  - Clicking cancel and outside dialog work as expected
- Notebook
  - Works with Py2 and Py3
  - cohorts from 1000 Genomes, Framingham, Baseline, AMP PD
  - By default the [fapi library](https://github.com/broadinstitute/fiss) talks to prod (api.firecloud.org). I didn't see an easy way to change this, so I wasn't able to run the notebook in local Terra. I did test the notebook in prod Terra.
  - Workspace has BigQuery tables from multiple datasets. Correct BigQuery tables are used
- Normal way of creating notebook still works

Will merge after #1805


seanhorgan:
Thanks @melissachang, great to see this come together. I like the language note in the tooltip. Should we also clarify that this action isn't just an open but a create? I'd be interested in Jerome's thoughts here especially in the case where a user has already created a notebook from this cohort and wants to open it again. This could open a door to cohort provenance both in a workspace and beyond.

Can you share that test notebook with me (my @google address)?

seanhorgan:
One other question: is there a way to monitor usage? I'd like to be able to understand adoption, e.g. how many unique notebooks were created from a cohort in the last month, how many unique people created a notebook from a cohort in the last 3 months?

melissachang:
@seanhorgan - You should be able to see [notebook](https://app.terra.bio/#workspaces/fc-product-demo/melchang%20test/notebooks/launch/cohort-notebook.ipynb) now.

FYI due to [IA-1402](https://broadworkbench.atlassian.net/browse/IA-1402) you have to run a cell, restart kernel, then run all cells. Once [IA-1402](https://broadworkbench.atlassian.net/browse/IA-1402) is fixed, then you can simply run all cells.

That's a good point about `create` vs `open`. I changed tooltip to "Open cohort in a new Python 2 or 3 notebook". FYI Brett suggested only supporting `create` to begin with, so I don't have worry about supporting locking/syncing. I wanted to get something working; then it can be iterated upon.

Regarding monitoring -- I haven't been following Terra client-side logging. Once there is a client-side logging system in place, it won't be hard to add this use case.

melissachang:
This is ready to be reviewed. I rebased with #1805 and retested.

jeromechadel:
Open with a new notebook / open with an existing notebook.
There’s no need of a new pattern here.
Jerome

melissachang:
@jeromechadel -- Brett suggested to **initially** only support `create`, so I don't have to worry about locking/syncing. Can this PR support only `create`, then someone (not me) can add support for `open` if they want?

jeromechadel:
Yes, you are correct. I was just answering Sean. :-)
Jerome

seanhorgan:
Great. Sounds like we are aligned that this action should be referred to as
'create'.

melissachang:
Got it. Tooltip is now "Create a Python 2 or 3 notebook with this cohort".

src/components/notebook-utils.js
```diff
@@ -78,7 +79,7 @@ export const NotebookCreator = class NotebookCreator extends Component {
 
   render() {
     const { notebookName, notebookKernel, creating, nameTouched } = this.state
-    const { reloadList, onDismiss, namespace, bucketName, existingNames } = this.props
+    const { reloadList, onDismissNotebookCreated, onDismissNotebookNotCreated, namespace, bucketName, existingNames } = this.props
```
zarsky-broad: This seems like `onDismiss` vs `onSuccess`.

src/pages/workspaces/workspace/Data.js
```diff
@@ -360,14 +364,37 @@ const ReferenceDataContent = ({ workspace: { workspace: { namespace, attributes
   ])
 }
 
-const ToolDrawer = withModalDrawer()(({
-  workspace, workspace: { workspace: { workspaceId } }, onDismiss, onIgvSuccess, entityMetadata, entityKey, selectedEntities
+const ToolDrawer = withModalDrawer()(requesterPaysWrapper({
```
zarsky-broad: This should use `_.flow` for composition.

src/pages/workspaces/workspace/Data.js
```diff
@@ -360,14 +364,37 @@ const ReferenceDataContent = ({ workspace: { workspace: { namespace, attributes
...
+  onDismiss: ({ onDismiss }) => onDismiss()
+})(({
+  workspace, workspace: { workspace: { workspaceId } }, onDismiss, onIgvSuccess,
+  onRequesterPaysError, entityMetadata, entityKey, selectedEntities
 }) => {
+  const { workspace: { bucketName, name: wsName, namespace } } = workspace
```
zarsky-broad: Why not add these to the destructure above?

src/pages/workspaces/workspace/Data.js
```diff
@@ -360,14 +364,37 @@ const ReferenceDataContent = ({ workspace: { workspace: { namespace, attributes
...
+  const loadNotebookNames = _.flow(
+    withRequesterPaysHandler(onRequesterPaysError),
+    withErrorReporting('Error loading notebooks')
+  )(async () => {
+    const notebooks = await Buckets.listNotebooks(namespace, bucketName)
+    // slice removes 'notebooks/' and the .ipynb suffix
+    setNotebookNames(notebooks.map(notebook => notebook.name.slice(10, -6)))
+  })
+  Utils.useOnMount(() => {
+    loadNotebookNames()
```
zarsky-broad, line 407: If this is only used here, our convention is to move the definition inside of the `useOnMount` as well, for clarity.

src/pages/workspaces/workspace/Data.js
```diff
@@ -360,14 +364,37 @@ const ReferenceDataContent = ({ workspace: { workspace: { namespace, attributes
...
+  })
+
   const entitiesCount = _.size(selectedEntities)
   const isCohort = entityKey === 'cohort'
-  const dataExplorerButtonEnabled = isCohort && entitiesCount === 1 && _.values(selectedEntities)[0].attributes.data_explorer_url
+  const dataExplorerButtonEnabled = isCohort && entitiesCount === 1 && _.values(selectedEntities)[0].attributes.data_explorer_url !== undefined
   const dataExplorerUrl = dataExplorerButtonEnabled ? `${_.values(selectedEntities)[0].attributes.data_explorer_url}&wid=${workspaceId}` : undefined
+  const notebookButtonEnabled = isCohort && entitiesCount === 1
```
zarsky-broad, line 423: If you put this before `dataExplorerButtonEnabled`, it can build off of this.

src/pages/workspaces/workspace/Data.js
```diff
@@ -382,6 +409,24 @@ const ToolDrawer = withModalDrawer()(({
       title: 'YOUR WORKFLOWS',
       drawerContent: h(WorkflowSelector, { workspace, selectedEntities })
     })
+  ], [
+    'Notebook', () => ({
+      drawerContent: h(NotebookCreator, {
+        bucketName, namespace,
+        existingNames: notebookNames,
+        onDismissNotebookCreated: async notebookName => {
+          console.assert(isCohort && entitiesCount === 1, '1 cohort should be selected')
```
zarsky-broad: Is this necessary? We appear to be guarding for it above.

src/pages/workspaces/workspace/Data.js
```diff
@@ -382,6 +409,24 @@ const ToolDrawer = withModalDrawer()(({
...
+          const cohortName = _.values(selectedEntities)[0].name
+          const contents = cohortNotebookTemplate.replace('MY_COHORT', cohortName)
+          await Buckets.notebook(namespace, bucketName, notebookName).create(JSON.parse(contents))
+          Nav.goToPath('workspace-notebook-launch', { namespace, name: wsName, notebookName: notebookName + '.ipynb' })
```
```suggestion
          Nav.goToPath('workspace-notebook-launch', { namespace, name: wsName, notebookName: `${notebookName}.ipynb` })
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -382,6 +409,24 @@ const ToolDrawer = withModalDrawer()(({
...
+        },
+        onDismissNotebookNotCreated: () => {
+          setToolMode(undefined)
+        },
+        reloadList: () => {}
```
```suggestion
        reloadList: _.noop
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -382,6 +409,24 @@ const ToolDrawer = withModalDrawer()(({
...
+        bucketName, namespace,
+        existingNames: notebookNames,
+        onDismissNotebookCreated: async notebookName => {
+          console.assert(isCohort && entitiesCount === 1, '1 cohort should be selected')
+          const cohortName = _.values(selectedEntities)[0].name
+          const contents = cohortNotebookTemplate.replace('MY_COHORT', cohortName)
+          await Buckets.notebook(namespace, bucketName, notebookName).create(JSON.parse(contents))
+          Nav.goToPath('workspace-notebook-launch', { namespace, name: wsName, notebookName: notebookName + '.ipynb' })
+        },
+        onDismissNotebookNotCreated: () => {
```
```suggestion
        onDismissNotebookNotCreated: setToolMode
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -425,6 +470,23 @@ const ToolDrawer = withModalDrawer()(({
...
+            ]),
+            h(ModalToolButton, {
+              onClick: () => setToolMode('Notebook'),
+              disabled: !notebookButtonEnabled,
+              tooltip: Utils.cond(
+                [!entityMetadata.cohort, () => 'Talk to your dataset owner about setting up a Data Explorer. See the "Making custom cohorts with Data Explorer" help article.'],
+                [isCohort && entitiesCount > 1, () => 'Select exactly one cohort to open in notebook'],
+                [!isCohort, () => 'Only cohorts can be opened with notebooks'],
+                [notebookButtonEnabled, () => 'Create a Python 2 or 3 notebook with this cohort'],
+                () => undefined
```
zarsky-broad: Unneeded, undefined is the implicit default.

zarsky-broad (commented):
Instead of having `cohort-notebook` as an ipynb file, consider defining it as a function that takes the name and just returns JSON, and putting that in a js file. That would remove the need for the new loader, and also allow you to avoid doing string replacement where you're creating it.

src/components/notebook-utils.js
melissachang: Done

src/pages/workspaces/workspace/Data.js
melissachang: Done

src/pages/workspaces/workspace/Data.js
melissachang: Done

src/pages/workspaces/workspace/Data.js
melissachang: Done

src/pages/workspaces/workspace/Data.js
melissachang: I think it's cleaner and less bug-prone to be explicit

src/pages/workspaces/workspace/Data.js
melissachang: Not with this PR. I thought I would add it to prevent future bugs, since this code is complicated. Removed.

src/pages/workspaces/workspace/Data.js
melissachang: Done

src/pages/workspaces/workspace/Data.js
melissachang: Done

src/pages/workspaces/workspace/Data.js
melissachang: Done

src/pages/workspaces/workspace/Data.js
melissachang: Done

melissachang (commented):
Notebooks are typically represented as .ipynb files. I think it's cleaner as-is. Is that ok?

Typical workflow:
- Make changes to notebook in Terra. Then download `.ipynb`, copy to this repo

zarsky-broad (commented):
Adding in a new loader still feels like a lot to me, especially since it's to load a file with a custom extension that we're just loading the text out of. It also feels quite brittle to be leaning on that string substitution for the cohort name. I'd rather see one of the following two options:

```js
export const cohortNotebook = cohortName => ({
  'cells': [
   {
    'cell_type': 'code',
    'execution_count': null,
    'metadata': {},
    'outputs': [],
    'source': [
     `COHORT = '${cohortName}'`
    ]
   },
...
})
```
or
```js
export const cohortNotebook = cohortName => `
{
  "cells": [
   {
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
     "COHORT = '${cohortName}'"
    ]
   },
...
`
```

At the end, this gets stringified, so it may make the most sense just to treat it as a big string and sub in the one value; either way, I think string substitution on a file that we're adding with a custom extension so we can ignore the extension and parse it as text is an unnecessary complication. The benefit of the object/non-string version is that we can make sure we don't break the structure of the file.

melissachang:
Done, put notebook template in js

zarsky-broad (approved):
Clear to merge.

TOL: It occurs to me that the double-escaping of newlines is something we wouldn't have to worry about if we treated the template as JSON instead of a big string, but that's pretty minor.

melissachang:
Thanks. I'll merge after [WORKSPACE_NAME bug](https://broadinstitute.zendesk.com/agent/tickets/3776) is fixed (should be today or next couple days). Notebooks won't work with that bug.

PR #1882, opened 2019-10-08
add appcues script [SATURN-1104]

panentheos:
This provides an MVP implementation of the appcues JS, which should allow comms to start building out the user flows.

Note that ad-blockers may block the loading of appcues by default (mine did), so it might not be available in all cases. Hence the need to make the calls conditional.


ehanna4 (approved):
Discussed and tested (runs and can see script in network tab) and all looks & sounds good to me

PR #1885, opened 2019-10-08
no unnamed components in devtools

zarsky-broad:
1. ~Remove `pure` and thus recompose - no longer does anything in hooks-land.~ Replace `pure` with `memo` where helpful
2. Create `Utils.withDisplayName`
  i. Create `Utils.forwardRefWithName`
  ii. Create `Utils.memoWithName`


src/components/group-common.js
```diff
@@ -25,17 +24,17 @@ const styles = {
...
     onClick
   }, [
     div(['Add a User']),
     icon('plus-circle', { style: { marginTop: '0.5rem' }, size: 21 })
   ])
-})
+}
 
-export const MemberCard = pure(({ member: { email, roles }, adminCanEdit, onEdit, onDelete, adminLabel, userLabel }) => {
+export const MemberCard = ({ member: { email, roles }, adminCanEdit, onEdit, onDelete, adminLabel, userLabel }) => {
```
panentheos: FYI, `pure` was still having an effect, by adding a wrapper with a `shouldComponentUpdate` check. The preferred way to do that now is with the built-in `memo`. However, I suspect that these usages weren't providing much value and it's safe to remove them.

src/components/workspace-utils.js
```diff
@@ -28,17 +28,15 @@ export const useWorkspaces = () => {
 }
 
 export const withWorkspaces = WrappedComponent => {
-  const Wrapper = props => {
+  return Utils.withDisplayName('withWorkspaces()', props => {
```
panentheos: The recent devtools seems to strip off the parens when displaying these names. Since we're updating this code, it might be nice to remove the parens here to avoid confusion.

src/components/bucket-utils.js
```diff
@@ -20,7 +20,7 @@ export const withRequesterPaysHandler = _.curry((handler, fn) => async (...args)
 })
 
 export const requesterPaysWrapper = ({ onDismiss }) => WrappedComponent => {
-  const RequesterPaysWrapper = (props, ref) => {
+  return Utils.forwardRefWithName('RequesterPaysWrapper', (props, ref) => {
```
panentheos: This should probably be `'requesterPaysWrapper'` for consistency.

panentheos (approved):
Nice change! See a few small comments.

PR #1889, opened 2019-10-11
Add paste option for importing data to data-model table SATURN-1036

ehanna4:
Add ability for user to paste in tab-spaced-data for the data model table. Currently only supporting pasting in (not editable or type-able at all)


breilly2:
Can you add the JIRA ticket to the PR and make sure it has release notes? Thanks.

src/components/input.js
```diff
@@ -33,6 +33,14 @@ const styles = {
...
     cursor: 'text'
   },
+  textareaResizable: {
+    width: '100%', resize: 'vertical', display: 'flex',
+    justifyContent: 'flex-start',
+    border: `1px solid ${colors.light()}`, borderRadius: 4,
+    fontSize: 14, fontWeight: 400,
+    padding: '0.5rem 1rem',
+    cursor: 'text'
+  },
```
panentheos: Since this is almost identical to `textarea` with the exception of `resize`, it might be cleaner to just use `styles.textarea` below, and add one override for `resize`. (Note that the flex properties shouldn't be defined here)

src/components/input.js
```diff
@@ -320,3 +328,17 @@ export const TextArea = ({ onChange, ...props }) => {
...
+  Utils.useConsoleAssert(props.id || props['aria-label'], 'In order to be accessible, PasteOnlyInput needs a label')
+
+  return textarea(_.merge({
+    className: 'focus-style',
+    style: styles.textareaResizable,
+    onPaste: e => {
+      onPaste(e.clipboardData.getData('Text'))
+    },
+    onChange: undefined,
+    readOnly: readOnly ? readOnly : false
```
panentheos: nit: This logic could be expressed as a default value in the props destructuring above.

src/libs/data-utils.js
```diff
@@ -269,90 +268,130 @@ export const EntityUploader = class EntityUploader extends Component {
...
+          'The first column header must be:',
+          div({ style: { fontFamily: 'monospace', margin: '0.5rem', fontWeight: '600' } }, ['entity:[type]_id']),
+          'where ',
+          span({ style: { fontFamily: 'monospace', fontWeight: '600' } }, ['[type]']),
+          ` is the desired name of the data table in ${getAppName()}. For example, to create or update a `,
+          span({ style: { fontFamily: 'monospace', fontWeight: '600' } }, ['participant']),
+          ' table use ',
+          span({ style: { fontFamily: 'monospace', fontWeight: '600' } }, ['entity:participant_id']),
+          div({ style: { marginTop: '0.5rem', marginBottom: '0.5rem' } }, ['All of the values in the ID column must be unique.']),
+          div({ style: { height: '0.5rem' } }),
```
panentheos: nit: rather than having a spacer div, we could just extend the margin of the element above.

src/libs/data-utils.js
```diff
@@ -269,90 +268,130 @@ export const EntityUploader = class EntityUploader extends Component {
...
+          h(SimpleTabBar, {
+            style: { paddingTop: '0.5rem' },
```
panentheos: nit: This doesn't seem to be doing anything, and could be removed.

src/libs/data-utils.js
```diff
@@ -269,90 +268,130 @@ export const EntityUploader = class EntityUploader extends Component {
...
+              'aria-label': 'Paste text data here',
+              placeholder: 'entity:participant_id(tab)column1(tab)column2...',
+              onPaste: pastedText => {
+                this.setState({ file: new File([pastedText], 'upload.tsv'), fileContents: pastedText, isFileImportLastUsedMode: false })
+              },
+              readOnly: !!fileContents,
+              value: !isFileImportLastUsedMode ? fileContents : '',
+              wrap: 'off',
+              style: {
+                flex: 1, fontFamily: 'monospace', resize: 'vertical',
```
panentheos: nit: `flex` is unnecessary here since we're not in a flex container, and `resize` is already specified in the common component.

src/libs/data-utils.js
```diff
@@ -269,90 +268,130 @@ export const EntityUploader = class EntityUploader extends Component {
...
+                (currentFile && currentFile.name) ? currentFile.name : 'None')
+            ])
+          ]) : div([
+            h(PasteOnlyInput, {
+              'aria-label': 'Paste text data here',
+              placeholder: 'entity:participant_id(tab)column1(tab)column2...',
+              onPaste: pastedText => {
+                this.setState({ file: new File([pastedText], 'upload.tsv'), fileContents: pastedText, isFileImportLastUsedMode: false })
+              },
+              readOnly: !!fileContents,
```
panentheos, line 353: TOL: React currently prints a warning about the usage pattern of `onChange` and `readOnly` here. I understand the reasoning for doing it this way, but we should revisit this afterward and see if there's another way to accomplish that.

panentheos (approved):
Looks great, thanks for your patience on this. A few minor comments, then should be good to go.

PR #1891, opened 2019-10-11
Replace React Autosuggest with Downshift [SATURN-1169]

zarsky-broad:
Autosuggest is no longer updated.


src/components/input.js
```diff
@@ -199,114 +199,64 @@ export const ValidatedInput = props => {
...
+    inputId: id
+  }, [
+    ({ getInputProps, getMenuProps, getItemProps, isOpen, openMenu, toggleMenu, highlightedIndex }) => {
+      return div({ style: { width: '100%' } }, [
+        h(TextInput, getInputProps({
+          style,
+          type: 'search',
+          onFocus: openOnFocus ? openMenu : undefined,
+          onKeyDown: e => {
+            if (e.key === 'Escape') {
```
ehanna4, line 231: What is the intention of this logic with 'Escape'? It seems like it is trying to have 'Escape' escape the dropdown if that is open without escaping the whole model. However this is not working for me in testing (if my understanding is wrong then no issue there) However after I choose an option from the dropdown 'Escape' no longer does anything unless I re-click/focus on the model which I don't think is intended?

zarsky-broad: Hmm, this seems to be working for me.
The escape behavior should be this: 
1. Input unfocused: no change from normal, modal will close
2. Input focused:
  a. Menu open: menu will close
  b. Text entered: text is cleared
  c. Neither: menu will show; however normal behavior won't be prevented, so modal will close

ehanna4: Hmm testing today yielded the behavior you outlined above, I am not sure what I was running into last week, apologies, all looks good!

ehanna4 (approved):
Looks good to me! Behavior works as expected

src/components/group-common.js
```diff
@@ -145,19 +148,9 @@ export const NewUserModal = ajaxCaller(class NewUserModal extends Component {
...
-            },
-            onKeyDown: e => {
-              // 27 = Escape
-              if (e.which === 27 && !!userEmail) {
-                this.setState({ userEmail: '' })
-                e.stopPropagation()
-              }
-            },
-            suggestions,
+            suggestions: [...(!!userEmail && !suggestions.includes(userEmail) ? [userEmail] : []), ...suggestions],
```
panentheos, line 151: There's a slight behavior change on the groups page: typing and then clearing the box results in all the suggestions being shown, where before it closed the suggestions. That may be fine, but I wanted to mention it.

src/components/input.js
```diff
@@ -199,114 +199,64 @@ export const ValidatedInput = props => {
...
+                e.nativeEvent.preventDownshiftDefault = true
+                e.preventDefault()
+              }
+              toggleMenu()
+            }
+          },
+          nativeOnChange: true,
+          ...props
+        })),
+        h(AutocompleteSuggestions, {
```
panentheos: Rendering these unconditionally ends up putting a lot of hidden portals on the workflow configuration page, and noticeably impacts performance. We should switch back to conditionally rendering so there's only one at a time.

src/components/input.js
```diff
@@ -199,114 +199,64 @@ export const ValidatedInput = props => {
...
+          target: getInputProps().id,
+          containerProps: getMenuProps(),
+          isVisible: isOpen && !noSuggestions
```
panentheos: This renders a thin empty box if there are no matching suggestions. We should probably be looking at whether `suggestions` is empty, not `rawSuggestions`.

panentheos (approved):
Nice library! See a few comments.

src/components/input.js
zarsky-broad: Ah, I had gone with unconditional rendering because https://github.com/downshift-js/downshift#getmenuprops says
> Note that for accessibility reasons it's best if you always render this element whether or not downshift is in an `isOpen` state.

I'll switch back to conditional rendering, though, since none of these are for places where custom text isn't possible.

PR #1892, opened 2019-10-11
update deps & react

zarsky-broad:
tested and read changelogs; you may wish to test as well, since we're updating React here.


src/components/TopBar.js
```diff
@@ -98,7 +98,7 @@ const DropDownSection = ({ titleIcon, title, isOpened, onClick, children }) => {
       div({ style: { flexGrow: 1 } }),
       icon(isOpened ? 'angle-up' : 'angle-down', { size: 18, style: { flex: 'none' } })
     ]),
-    h(RCollapse, { isOpened, style: { flex: 'none' } }, [children])
```
panentheos, line 101: Without this, the collapse content won't assert its full height in an overflow situation, which means the menu can break on shorter pages.

panentheos (approved):
Looks good, see comment.

PR #1893, opened 2019-10-15
add fields specifier to workspaces list call

panentheos:
This effectively omits the non-description workspace attributes, which reduces payload size by quite a bit.


zarsky-broad (approved):
Should fields default to an empty array?

panentheos:
I think going forward the intention is to always specify explicitly, so no default is probably better.

PR #1895, opened 2019-10-17
display the current environment in Runtime Configuration [SATURN-1124]

cameron-ardell:
Previous issue: "I created a cluster with the Bioconductor image. When the cluster was ready I opened my Runtime environment and the drop-down said "Default". It would be better to display the runtime environment currently in use."

makes it so current runtime environment is shown as the default if one exists, otherwise uses default settings that previously existed

(also having some issues merging with dev so not updated fully yet)

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -20,6 +20,7 @@ import { clusterCost, currentCluster, machineConfigCost, normalizeMachineConfig,
 import colors from 'src/libs/colors'
 import { reportError, withErrorReporting } from 'src/libs/error'
 import * as Nav from 'src/libs/nav'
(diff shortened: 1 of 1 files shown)


src/components/ClusterManager.js
```diff
@@ -377,6 +379,13 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
       ])
     ])
   }
+
+  componentDidUpdate() {
+    StateHistory.update(_.pick(
```
panentheos, line 384: `StateHistory` only stores data associated with the current page, so this won't be a sufficient solution in this case. We'll need to make a separate call to the server to get the info we want.

PR #1896, opened 2019-10-17
minor visual tweaks to ws [SATURN-1170]

zarsky-broad:
1. Makes fields on dashboard move less as they populate
2. Brings disabled appearance of buttons in 'open with' drawer in line with Lou's mocks
  a. De-dupes code for buttons
  b. Opacity is applied to the image twice; this gives the desired appearance.


src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -267,7 +267,7 @@ export const WorkspaceDashboard = _.flow(
           h(InfoTile, { title: 'Submissions' }, [submissionsCount]),
           h(InfoTile, { title: 'Access level' }, [roleString[accessLevel]]),
           Utils.canWrite(accessLevel) && h(InfoTile, { title: 'Est. $/month' }, [
-            storageCostEstimate
+            storageCostEstimate || '$ ...'
```
ehanna4, line 270: I like the look of how this comes up on the page a lot! looks much more clean to me

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -304,7 +304,7 @@ export const WorkspaceDashboard = _.flow(
               }, [icon('times', { size: 14 })])
             ])
           }, tagsList),
-          busy && spinner({ style: { marginTop: '0.5rem' } })
+          (busy || tagsList === undefined) && spinner({ style: { marginTop: '0.5rem' } })
```
ehanna4: This seems to create more movement when a workspace has no tags (moving the "Google Bucket..." section down to fit the spinner and then back up when the page has finished rendering with no tags) Maybe could add a height to the container so that the space is consistently large enough to accommodate the spinner?

ehanna4 (approved):
Looks good, just see the one comment though

PR #1899, opened 2019-10-18
add filter for workflow io table [SATURN-865]

zarsky-broad:
Easier to review with whitespace changes hidden.


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -849,26 +853,34 @@ const WorkflowView = _.flow(
...
+          'aria-label': `Search ${key}`,
+          style: { marginLeft: '1rem', width: 200 },
+          placeholder: `SEARCH ${key.toUpperCase()}`,
+          value: filter,
+          onChange: filter => this.setState({ filter })
         })
-      ])
+      ]),
+      filteredData.length === 0 ?
+        `No matching ${key}.` :
```
panentheos: Elsewhere we use a pattern of rendering the table, but passing a `noContentMessage` to handle the case where everything is filtered out, which provides a nice visual differentiation vs just having no rows at all.

panentheos (approved):
Looks good, see one idea.

PR #1038, opened 2019-01-08
Validate special characters in notebook name

katevoss:
Fixes #1026 

- Notebook name cannot contain , : ; @ $ % = + / \
- Increased font size and weight to make validation errors slightly bolder

PR #1902, opened 2019-10-21
minor text changes for notebookNameValidator [SATURN-1187]

andy7i:
reorder regex to match the character order in the error message and remove extraneous '/r /r'

Old:

Screen Shot 2019-11-05 at 1 28 21 PM

based on discussions with Brian, the text is now on 2 lines:

New:
Screen Shot 2019-11-05 at 1 30 14 PM

### DIFF
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -42,8 +42,9 @@ export const findPotentialNotebookLockers = async ({ canShare, namespace, wsName
 export const notebookNameValidator = existing => ({
   presence: { allowEmpty: false },
   format: {
-    pattern: /^[^#[\]*?:;@$%+=\\,/]*$/,
-    message: 'can\'t contain these characters: \r \r  @ # $ % * + = ? , [ ] : ; / \\ '
+    pattern: /^[^@#$%*+=?,[\]:;/\\]*$/,
+    message: 'can\'t contain these characters: @ # $ % * + = ? , [ ] : ; / \\ '
+
   },
   exclusion: {
     within: existing,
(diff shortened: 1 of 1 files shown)


src/components/notebook-utils.js
```diff
@@ -42,8 +42,8 @@ export const findPotentialNotebookLockers = async ({ canShare, namespace, wsName
 export const notebookNameValidator = existing => ({
   presence: { allowEmpty: false },
   format: {
-    pattern: /^[^#[\]*?:;@$%+=\\,/]*$/,
-    message: 'can\'t contain these characters: \r \r  @ # $ % * + = ? , [ ] : ; / \\ '
+    pattern: /^[^@#$%*+=?,[\]:;/\\]*$/,
+    message: 'can\'t contain these characters: @ # $ % * + = ? , [ ] : ; / \\ '
```
breilly2: Those `\r`s are strange... I wonder if it was an attempt to make sure the list of characters didn't soft-wrap somewhere in the middle...

Confirmed by looking back at the previous PR: https://github.com/DataBiosphere/terra-ui/pull/1038#discussion_r246189122

It doesn't currently wrap on my screen, but any number of unrelated modifications could change that. Suggestion:

```
    message: h(Fragment, [
      div('Name can\'t contain these characters:'),
      div({ style: { margin: '0.5rem 1rem' } }, '@ # $ % * + = ? , [ ] : ; / \\ ')
    ])
```

breilly2 (approved):
👍 Great idea! One suggestion to satisfy the intent of the `\r`s. See comment.

breilly2:
I did some further testing to make sure the regex still works, checking each invalid character individually. All looks good. Brett fixed the underlying problem that was causing some react key warnings (#1940) as well as confirming the belief that using a react component for the `message` property is valid. I rebased the branch and confirmed that the warning is gone.

PR #1904, opened 2019-10-23
Show current  runtime [SATURN-1124]

cameron-ardell:
Previous issue: "I created a cluster with the Bioconductor image. When the cluster was ready I opened my Runtime environment and the drop-down said "Default". It would be better to display the runtime environment currently in use."

makes it so current runtime environment is shown as the default if one exists, otherwise uses default settings that previously existed. Also switches to showing the custom environment with their image name if they are using that instead.


panentheos (approved):
Looks good!

PR #1905, opened 2019-10-24
Initial functional test for CUJ2 [SATURN-1079]

ehanna4:
Note: No set up or tear down (will be taken care of in future PR/ticket) and the workspace name is parameterized as "WORKSPACE"


src/_tests/integration/cuj2.integration-test.js
```diff
@@ -0,0 +1,35 @@
+const { click, findText, waitForNoSpinners, signIntoTerra } = require(
+  './integration-utils')
```
zarsky-broad: nit: linebreak

src/_tests/integration/integration-utils.js
```diff
@@ -27,12 +27,23 @@ const waitForNoSpinners = page => {
   return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
 }
 
+const delay = ms => {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+const signIntoTerra = async page => {
+  await findText(page, 'requires a Google Account')
+  return (await page.evaluate(token => window.forceSignIn(token), process.env.TERRA_TOKEN))
```
zarsky-broad: Either await or return, don't need both.

src/_tests/integration/fire-cloud-utils.js
```diff
@@ -0,0 +1,33 @@
...
+    await waitForNoSpinners(page)
+    await findText(page, 'content you are looking for is currently only accessible')
+    await delay(500) // wait half a second for full load not accounted for by spinners
+    await page.evaluate(token => window.forceSignedIn(token), process.env.TERRA_TOKEN) // Note: function for Fire Cloud is forceSignedIn() while Terra is forceSignIn()
+  }
+
+  return { openFireCloudSelector, findFireCloudClickable, click, signIntoFireCloud, selectWorkSpace }
+}
+module.exports = {
+  fireCloud: fireCloud()
```
zarsky-broad: Instead of wrapping this all in a function that is called, why not just put each function here directly into the export?

src/_tests/integration/fire-cloud-utils.js
```diff
@@ -0,0 +1,33 @@
...
+    await openFireCloudSelector(page)
+    return (await page.waitForXPath(`//ul/li[contains(normalize-space(.),"${billingAccount}/${workSpace}")]`)).click()
+  }
+
+  const findFireCloudClickable = (page, text) => {
+    return page.waitForXPath(`(//a | //*[@role="button"])[contains(normalize-space(.),"${text}") or contains(@data-test-id,"${text}")]`)
+  }
+
+  const click = async (page, text) => {
+    return (await findFireCloudClickable(page, text)).click()
```
zarsky-broad: Either await or return, don't need both. This could be a synchronous function, like the one above it.

src/_tests/integration/fire-cloud-utils.js
```diff
@@ -0,0 +1,33 @@
+const { findText, waitForNoSpinners, delay } = require('./integration-utils')
+
+
+const fireCloud = () => {
+  const openFireCloudSelector = async page => {
+    return (await page.waitForXPath(`//*[@data-test-id="workspace-selector"]`)).click()
```
zarsky-broad: Either await or return, don't need both. This could be a synchronous function, like `findFirecloudClickable`.

zarsky-broad (commented):
Generally we refer to `firecloud` and `workspace` as one non-camelcase word, I'd change the names of functions and files here to reflect that.

src/_tests/integration/fire-cloud-utils.js
ehanna4: Since these are firecloud versions of the utils it seemed more clear to have them wrapped so that in the main test it's calling `firecloud.<function>` making it explicit that these utils are for firecloud not regular ones. I can certainly change it though if you think it would be better to directly export the functions

src/_tests/integration/fire-cloud-utils.js
ehanna4: My understanding was that this line has 2 nested async things happening, which is why there is both an `await` which is awaiting the promise returned from `findFirecloudClickable()` and then the `return` is returning the promise that comes back from the asynchronous function `click()` which is called at the end after the `findFirecloudclickable` has been resolved. So I believe it needs both of these?

src/_tests/integration/fire-cloud-utils.js
ehanna4: Similarly this one is awaiting the promise from `waitForXPath` and then the `return` is returning on the promise from the `click()`, but the `click()` has to await the `waitForXPath` being resolved which is why it's nested in the parens. `findFirecloudClickable` is only passing the promise it gets back from the `waitForXPath` function which is why there's no await/async on that one. So I believe it still needs both the `await` and `return` here?

src/_tests/integration/cuj2.integration-test.js
```diff
@@ -0,0 +1,34 @@
...
+  await click(page, 'code & workflows')
+  await click(page, workflowName)
+
+  await firecloud.signIntoFirecloud(page)
+  await waitForNoSpinners(page)
+  await findText(page, workflowName)
+  await firecloud.click(page, 'Export to Workspace...')
+  await waitForNoSpinners(page)
+  await findText(page, `gatk/${workflowName}-configured`)
+  await firecloud.click(page, `gatk-${workflowName}-configured-1-link`)
```
panentheos: If at all possible, it would be preferable to click these by their visible text, rather than by test-id. Also, if we do that, can we effectively combine this line with the previous one, since the `click` will effectively wait for the element to be there?

src/_tests/integration/cuj2.integration-test.js
```diff
@@ -0,0 +1,34 @@
...
+  await signIntoTerra(page)
+  await findText(page, 'GATK4 example workspaces')
+  await click(page, 'code & workflows')
+  await click(page, workflowName)
+
+  await firecloud.signIntoFirecloud(page)
+  await waitForNoSpinners(page)
+  await findText(page, workflowName)
+  await firecloud.click(page, 'Export to Workspace...')
+  await waitForNoSpinners(page)
```
panentheos: Can we get by without `waitForNoSpinners` in most of these cases? Ideally we'd only use this as a workaround for when the element we want is on-screen, but blocked by a spinner. Here, it looks like the element appears after the spinner disappears, so it might be possible to just wait on the interaction we want, directly.

src/_tests/integration/cuj2.integration-test.js
```diff
@@ -0,0 +1,34 @@
...
+const { firecloud } = require('./fire-cloud-utils.js')
+
+
+const workflowName = 'haplotypecaller-gvcf-gatk4'
+
+test.skip('integration', async () => {
+  await page.goto('localhost:3000')
+  await click(page, 'View Examples')
+  await signIntoTerra(page)
+  await findText(page, 'GATK4 example workspaces')
```
panentheos: This probably isn't strictly necessary, since the 'code & workflows' tab appears at the same time, so we could just go directly to the `click`, which will wait if needed.

src/_tests/integration/cuj2.integration-test.js
```diff
@@ -0,0 +1,34 @@
...
+  await firecloud.click(page, `gatk-${workflowName}-configured-1-link`)
+  await waitForNoSpinners(page)
+  await firecloud.click(page, 'use-selected-configuration-button')
+  await waitForNoSpinners(page)
+  await findText(page, 'Select a workspace')
+  await firecloud.selectWorkspace(page, 'general-dev-billing-account', process.env.WORKSPACE)
+  await firecloud.click(page, 'import-export-confirm-button')
+  await firecloud.click(page, 'Yes')
+
+  await signIntoTerra(page)
```
panentheos: We should be careful here, since at least in the local environment case, FC actually redirects back to the dev environment here (which might be why we're signed out). So we might need to either fix that, or hack around it by explicitly navigating back to the right server (for now).

src/_tests/integration/fire-cloud-utils.js
```diff
@@ -0,0 +1,33 @@
...
+    await waitForNoSpinners(page)
+    await findText(page, 'content you are looking for is currently only accessible')
+    await delay(500) // wait half a second for full load not accounted for by spinners
+    await page.evaluate(token => window.forceSignedIn(token), process.env.TERRA_TOKEN) // Note: function for Fire Cloud is forceSignedIn() while Terra is forceSignIn()
+  }
+
+  return { openFirecloudSelector, findFirecloudClickable, click, signIntoFirecloud, selectWorkspace }
+}
+module.exports = {
+  firecloud: firecloud()
```
panentheos: If these are going to live in their own file, then they're already effectively 'namespaced'. It would be more idiomatic to flatten out these exports, and instead import the module like `const firecloud = require('./fire-cloud-utils')`

src/_tests/integration/cuj2.integration-test.js
```diff
@@ -0,0 +1,34 @@
+const { click, findText, waitForNoSpinners, signIntoTerra } = require('./integration-utils')
+const { firecloud } = require('./fire-cloud-utils.js')
```
panentheos: nit: don't need to specify `.js` extension when importing.

src/_tests/integration/cuj2.integration-test.js
```diff
@@ -0,0 +1,34 @@
+const { click, findText, waitForNoSpinners, signIntoTerra } = require('./integration-utils')
```
panentheos: I think we should avoid references to 'CUJ' in the code, since that name may not have a long life span. How about naming the file e.g. `find-workflow.integration-test.js`?

src/_tests/integration/integration-utils.js
```diff
@@ -27,12 +27,23 @@ const waitForNoSpinners = page => {
   return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
 }
 
+const delay = ms => {
+  return new Promise(resolve => setTimeout(resolve, ms))
```
panentheos, line 67: TOL: There's an implementation of this in `utils.js`, but I think making a copy here is a good choice for now, since we're still figuring out all the ways this will get packaged up and operationalized. So this keeps things a bit more self-contained.

src/_tests/integration/cuj2.integration-test.js
ehanna4: Combined these two lines and instead used the visible text for selection. The only click now that does not use the visible text is the click of the export button in the modal since there are two buttons then (one in modal one outside) with the same visible text so used the test-id

src/_tests/integration/find-workflow.integration-test.js
```diff
@@ -0,0 +1,41 @@
...
+const firecloud = require('./firecloud-utils')
+
+
+const workflowName = 'haplotypecaller-gvcf-gatk4'
+const billingAccount = 'general-dev-billing-account'
+const workspace = process.env.WORKSPACE
+const url = 'http://localhost:3000'
+
+test.skip('integration', async () => {
+  await page.goto('http://localhost:3000')
```
panentheos: This should use `url` now that it's been split out.

panentheos (approved):
Looks good, see one comment.

PR #1906, opened 2019-10-24
Add test for CUJ1:  Find & Include data from library [SATURN-1078]

petesantos:
* Added additional helper functions 
* Added helper functions to find nodes specific to data explorer markup
* Added dev environment settings with appropriate warnings for puppeteer flakiness
* Added aria-labels where appropriate to find nodes with XPath
* Re-ordered module exports to be in alphabetical order
* This test is currently disabled (test.skip) until we add workspace setup


src/_tests/integration/cohort-data-cuj1.integration-test.js
```diff
@@ -0,0 +1,32 @@
...
+
+  const frame = await findIframe(page)
+  await dataExplorer.clickTextInAnyNS(frame, 'Has WGS Low Cov')
+  await click(frame, 'Save cohort')
+  await dataExplorer.fillIn(frame, 'name', cohortName)
+  await dataExplorer.click(frame, 'Save')
+
+  await findText(page, 'requires a Google Account')
+  await page.evaluate(token => window.forceSignIn(token), process.env.TERRA_TOKEN)
+  await waitForNoSpinners(page)
```
panentheos: Do we need to use `waitForNoSpinners`? Ideally we'd only use this as a workaround for cases where the element we want is onscreen, but obscured by a spinner. Otherwise it's preferable to just interact directly with the element, which will implicitly wait if needed.

src/_tests/integration/data-explorer-utils.js
```diff
@@ -0,0 +1,31 @@
...
+
+  const clickTextInAnyNS = async (page, text) => {
+    return (await findTextInAnyNS(page, text)).click()
+  }
+
+  return { findInput, fillIn, findClickable, click, findTextInAnyNS, clickTextInAnyNS }
+}
+
+module.exports = {
+  dataExplorer: dataExplorer()
```
panentheos: If these are going to live in their own file, then they're already effectively 'namespaced'. It would be more idiomatic to flatten out these exports, and instead import the module like const dataExplorer = require('./data-explorer-utils')

src/_tests/integration/integration-utils.js
```diff
@@ -1,5 +1,38 @@
+const waitForFrame = async url => {
+  const frameLoaded = new Promise(resolve => {
+    const navHandler = async frame => {
+      await frame.url().includes(url) && resolve({ frame, navHandler })
+    }
+    page.on('framenavigated', navHandler)
```
panentheos: To keep these utils pure, we should probably pass the `page` as a parameter instead of depending on the jest global.

panentheos: TOL: Is there any risk that the frame will already be navigated by the time we call this?

src/_tests/integration/integration-utils.js
```diff
@@ -1,5 +1,38 @@
...
+
+  return await waitForFrame(src)
+}
+
+const findInGrid = async (page, text) => {
+  return await page.waitForXPath(`//*[@role="grid"][contains(normalize-space(.),"${text}")]`)
+}
+
+const exactlyFindClickable = (page, text) => {
+  return page.waitForXPath(`(//a | //*[@role="button"] | //button)[text()="${text}"]`)
```
panentheos: We might want to use something like `normalize-space(.)="${text}"` here to correctly handle nested text.

src/pages/LandingPage.js
```diff
@@ -34,6 +34,7 @@ const makeDocLink = (href, title) => {
 }
 
 const makeCard = (link, title, body) => h(Clickable, {
+  'aria-label': title,
```
panentheos: Is this necessary? I think the contents of the element should be usable.

src/pages/library/Datasets.js
```diff
@@ -149,7 +149,8 @@ const thousandGenomesLowCoverage = () => h(Participant, {
 }, [
   h(ButtonPrimary, {
     href: Nav.getLink('data-explorer-public', { dataset: '1000 Genomes' }),
-    tooltip: browseTooltip
+    tooltip: browseTooltip,
+    'aria-label': 'Browse 1000 Genomes Low Coverage'
```
panentheos, line 153: TOL: I think this is a case where we could do a better job of annotating the page contents, and doing that would make it more accessible in addition to enabling the integration test. Conceptually, this page contains a list of things. If we marked that up with the `list`/`listitem` roles, we could write the selector as "within the listitem containing '1000 Genomes Low Coverage', click 'Browse data'". That's how a real user will ultimately navigate this page, so the more we can mirror that the better.

jest-puppeteer.config.js
```diff
@@ -1,5 +1,14 @@
...
+    devtools: (({ devtools, iframes }) => {
+      if (devtools) {
+        iframes && console.warn('You cannot enable devtools while testing a UI with iframes in headful mode')
+        return !iframes
+      }
+    })({ devtools: process.env.DEVTOOLS === 'true', iframes: process.env.IFRAMES === 'true' }),
+    headless: process.env.HEADLESS !== 'false',
+    // Workaround for issue when accessing cross domain iframes in puppeteer while not headless:
+    // https://github.com/GoogleChrome/puppeteer/issues/4960
+    args: process.env.HEADLESS === 'false' && process.env.IFRAMES === 'true' ? ['--disable-features=site-per-process'] : []
```
panentheos: Since we have tests with iframes, doesn't this always need to be true?

src/_tests/integration/integration-utils.js
petesantos: Yes, that is a concern of mine as well. I made a slight tweak to the URL in DataExplorerFrame to conform to what page.frames() returns. This is currently working. Let me know if that solution looks better - this uses the interval wrapped in a promise that we had discussed earlier

src/pages/library/Datasets.js
petesantos: I'd be happy to do this in another ticket - it feels a bit out of scope for this one. Do you have any issues with me creating a separate ticket for this?

jest-puppeteer.config.js
petesantos: This one is an annoyance of puppeteer. The flag is only needed when running this in headful mode (e.g. debugging). The tests run fine and can find frames in headless mode without the flag. Once in headful mode puppeteer will not see the frames.

I don't want to enable the flag all the time as I don't know what other side effects may occur when this is turned on - I'd rather run the tests with no flags enabled when running in an automated fashion.

I set it as a flag separate from headless because I want our parameters to be explicit. I wouldn't want someone's tests to fail while debugging because HEADLESS was set to false and they did not realize HEADLESS sets a second flag as well. My concern is that while this flag fixes iframe issues, it could potentially have some other unknown effect that could cause another issue.

panentheos: Does that mean if you run with just `HEADLESS=false`, the test will fail?

src/_tests/integration/import-cohort-data.integration-test.js
```diff
@@ -0,0 +1,30 @@
+const { findInGrid, exactClick, click, findIframe, findText, select } = require('./integration-utils')
+const dataExplorer = require('./data-explorer-utils')
+
+
+jest.setTimeout(10000)
```
panentheos: I think this might be unnecessary if we're specifying timeouts in individual tests.

src/_tests/integration/integration-utils.js
```diff
@@ -1,5 +1,40 @@
+const waitForFn = async ({ fn, interval = 2000, timeout = 10000 }) => {
+  const readyState = new Promise(resolve => {
+    const start = Date.now()
+    const intervalId = setInterval(() => {
```
panentheos, line 4: This will wait a minimum of 2 seconds before trying the function. Should we make one call up front to see if it's already successful?

src/_tests/integration/integration-utils.js
```diff
@@ -1,5 +1,40 @@
...
+      const success = fn()
+      success && resolve({ success, intervalId })
+      Date.now() - start > timeout && resolve({ intervalId })
```
panentheos, line 7: If this times out, we should abort the test anyway, so perhaps we should reject instead. (If we do that we'll have to modify the handling of `clearInterval`)

src/pages/library/Datasets.js
panentheos: Yep, separately is fine, I just wanted to bring up the idea.

jest-puppeteer.config.js
petesantos: Yes. I may be overly paranoid here. I can tie this flag into the HEADLESS option ( I did that initially)  - I just don't want that to cause test failures for someone in the future for other reasons. The headless / iframe issue was annoying for me, I would like to guard against it for future devs. 

I could also issue another log statement warning about the iframe issue.

breilly2 (approved):
👍 (LGTMob)

PR #1908, opened 2019-10-25
IA-1472 Allow users to delete Stopping and Starting clusters

akarukappadath:
https://broadworkbench.atlassian.net/browse/IA-1472


ehanna4 (approved):
Looks good to me!

PR #1909, opened 2019-10-25
Revert to previous version of react-focus-lock [SATURN-1191]

petesantos:
Version > 2.1.0 of react-focus lock seems to have introduced a backwards incompatibility. I am working on identifying and providing repro steps to the author

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,11 +34,11 @@
       }
     },
     "@babel/generator": {
-      "version": "7.6.2",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.6.2.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -18734,9 +18599,9 @@
       },
       "dependencies": {
         "@babel/runtime": {
-          "version": "7.6.2",
-          "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.6.2.tgz",
-          "integrity": "sha512-EXxN64agfUqqIGeEjI5dL5z0Sw0ZwWo1mLTi4mQowCZ42O59b7DRpZAnTC6OqdF28wMBMFKNb/4uFGrVaigSpg==",
+          "version": "7.6.3",
```
panentheos: This seems like it upgrades a lot of unrelated stuff. Can we double check that?

panentheos (approved):
Approach is good, but see comment.

package-lock.json
petesantos: I corrected this. I should have checked the lock diff before submitting the PR - thanks!

PR #1910, opened 2019-10-25
IA-1406 upgrade bio conductor image

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1406

The new version has a few bug fixes. Would like to get it out asap

### DIFF
--- a/src/data/leo-images.js
+++ b/src/data/leo-images.js
@@ -12,13 +12,13 @@ export const leoImages = [
   },
   {
     label: 'Bioconductor (R 3.6.1, Bioconductor 3.9, Tidyverse 1.2.1)',
-    version: '0.0.2',
-    updated: '2019-09-06',
+    version: '0.0.3',
+    updated: '2019-10-25',
     packages: {
-      python: 'https://storage.googleapis.com/terra-docker-image-documentation/terra-jupyter-bioconductor-0.0.2-python-packages.txt',
-      r: 'https://storage.googleapis.com/terra-docker-image-documentation/terra-jupyter-bioconductor-0.0.2-r-packages.txt'
+      python: 'https://storage.googleapis.com/terra-docker-image-documentation/terra-jupyter-bioconductor-0.0.3-python-packages.txt',
+      r: 'https://storage.googleapis.com/terra-docker-image-documentation/terra-jupyter-bioconductor-0.0.3-r-packages.txt'
     },
-    image: 'us.gcr.io/broad-dsp-gcr-public/terra-jupyter-bioconductor:0.0.2'
+    image: 'us.gcr.io/broad-dsp-gcr-public/terra-jupyter-bioconductor:0.0.3'
   }
 ]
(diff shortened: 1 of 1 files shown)


src/data/leo-images.js
```diff
@@ -12,13 +12,13 @@ export const leoImages = [
   },
   {
     label: 'Bioconductor (R 3.6.1, Bioconductor 3.9, Tidyverse 1.2.1)',
-    version: '0.0.2',
-    updated: '2019-09-06',
+    version: '0.0.3',
+    updated: '2019-10-25',
```
breilly2, line 16: Something I just noticed and Pete and I dug into understanding... when you don't specify a time, UTC is assumed. If you'd rather be more specific, you can add a time using ISO standard date/time format, e.g.: `2019-10-25T16:30:00-04:00` or `2019-10-25T20:30:00Z` (which represent the same instant). Be sure to include the `T` though because this ultimately gets passed to `new Date()` and Safari's implementation rejects it without the `T`.

Further info: Chrome, Firefox, and Safari assume UTC if you don't specify a time. Safari also assumes UTC if you specify a time with no time zone. Chrome and Firefox, however, when given a specific time assume _local time zone_!!! Weird, huh?

TL;DR: You don't have to specify a time if you're okay with midnight UTC. If you do specify a time, use correct ISO format with time zone.

breilly2 (approved):
👍 I tested by creating a cluster with the updated runtime. I did _not_ test (because I don't know how) that the runtime included updated versions of the packages as promised.

Qi77Qi:
@breilly2 the package list are generated from https://github.com/DataBiosphere/terra-docker/blob/master/build.sh#L57, so it should be accurate 

Thanks for reviewing!

src/data/leo-images.js
Qi77Qi: This timestamp is just informational, and exact time is not critical..So I'll leave it as is

breilly2:
Hmm... just thought of something. What about existing runtimes that have the old environment? The config screen isn't going to reflect that. Might need something a little more fancy, like keeping a record of old versions and displaying them (maybe disabled?) if they're active.

Should people be able to select older versions?

A quick fix would be up update the list to contain both. Then we can spend a little more time thinking about what we want the experience to be for old runtimes.

Qi77Qi:
> What about existing runtimes that have the old environment? The config screen isn't going to reflect that. 

You mean when people try to create a new runtime if they already have a bio conductor runtime?

> Then we can spend a little more time thinking about what we want the experience to be for old runtimes

I think this is planned work outside this PR

breilly2:
Thinking a little more, I guess an old runtime would appear to be running the newer versions, according to the config panel. That's misleading and confusing. They'd have to delete and create a new runtime.

Qi77Qi:
@akarukappadath told me there's ongoing work around that. @zarsky-broad should know?

PR #1911, opened 2019-10-28
Display current cluster's Start up Script in notebook runtime widget [SATURN-1119]

cameron-ardell:
[ticket here](https://broadworkbench.atlassian.net/browse/SATURN-1119)

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -182,6 +182,11 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
       } else {
         this.setState({ isCustomEnv: true, customEnvImage: dockerImage })
       }
+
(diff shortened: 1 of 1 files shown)


src/components/NewClusterModal.js
```diff
@@ -182,6 +182,11 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
       } else {
         this.setState({ isCustomEnv: true, customEnvImage: dockerImage })
       }
+
+      const { jupyterUserScriptUri } = await Ajax().Jupyter.cluster(currentCluster.googleProject, currentCluster.clusterName).details()
```
panentheos: Since we're already making this call above, we could just pull this additional information out above instead of making the call again.

panentheos (commented):
This approach looks good, see comment for simplification.

panentheos (approved):
Looks good, thanks!

PR #1912, opened 2019-10-29
Update PFB import notification [SATURN-1200]

breilly2:
This is a minor improvement until SATURN-1153 (in progress) can be completed (once the status endpoint is fully working).

Also makes it so, if the "Data" tab is active, clicking it again will refresh entity metadata thereby updating the table list.

### DIFF
--- a/src/pages/ImportData.js
+++ b/src/pages/ImportData.js
@@ -68,7 +68,10 @@ class Importer extends Component {
         ['PFB', () => wsAjax.importPFB(url)],
         [Utils.DEFAULT, () => wsAjax.importBagit(url)]
       )
-      notify('success', 'Data imported successfully.', { timeout: 3000 })
+      Utils.cond(
+        [format === 'PFB', () => notify('info', 'Data import in progress.', { message: 'Data will show up incrementally as the job progresses.' })],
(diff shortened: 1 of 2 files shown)


src/pages/ImportData.js
```diff
@@ -68,7 +68,10 @@ class Importer extends Component {
         ['PFB', () => wsAjax.importPFB(url)],
         [Utils.DEFAULT, () => wsAjax.importBagit(url)]
       )
-      notify('success', 'Data imported successfully.', { timeout: 3000 })
+      Utils.cond(
```
panentheos: nit: Since we're in imperative code, and there are only two branches, this might be slightly clearer as an `if`

src/pages/workspaces/workspace/Data.js
```diff
@@ -926,6 +926,7 @@ const WorkspaceData = _.flow(
 
   refresh() {
     this.setState(({ refreshKey }) => ({ refreshKey: refreshKey + 1 }))
+    this.loadMetadata()
```
panentheos, line 929: The `onSuccess` callback of the `EntityUploader` calls both `this.refresh` and `this.loadMetadata`, so we should remove the extra call from there.

panentheos (approved):
Looks good, see comments.

PR #1914, opened 2019-10-29
Updates to AMP PD section in datasets tab

wnojopra:
The upcoming AMP PD release now uses a 2-tiered access model, and there is an updated data explorer for each tier.

Screenshot from 2019-10-29 15-00-39

cc @mbookman @melissachang

### DIFF
--- a/src/data/datasets.js
+++ b/src/data/datasets.js
@@ -7,12 +7,12 @@ const datasets = [
   },
   {
     name: 'AMP PD - 2019_v1release_1015',
-    origin: 'https://amp-pd-data-explorers.appspot.com',
+    origin: 'https://data-explorer.amp-pd.org',
     authDomain: 'amp-pd-researchers'
   },
   {
     name: 'AMP PD Clinical - 2019_v1release_1015',
-    origin: 'https://amp-pd-clinical-data-explorers.appspot.com',
+    origin: 'https://clinical-data-explorer.amp-pd.org',
     authDomain: 'amp-pd-clinical-access'
   },
   {
--- a/src/data/datasets.js
+++ b/src/data/datasets.js
@@ -6,9 +6,14 @@ const datasets = [
     origin: 'https://test-data-explorer.appspot.com'
   },
   {
-    name: 'AMP PD - 2019_v1beta_0220',
-    origin: 'https://amp-pd-data-explorer.appspot.com',
+    name: 'AMP PD - 2019_v1release_1015',
+    origin: 'https://amp-pd-data-explorers.appspot.com',
     authDomain: 'amp-pd-researchers'
+  },
+  {
+    name: 'AMP PD Clinical - 2019_v1release_1015',
+    origin: 'https://amp-pd-clinical-data-explorers.appspot.com',
+    authDomain: 'amp-pd-clinical-access'
   },
   {
     name: 'Baseline Health Study',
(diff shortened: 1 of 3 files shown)


src/data/datasets.js
```diff
@@ -6,10 +6,15 @@ const datasets = [
     origin: 'https://test-data-explorer.appspot.com'
   },
   {
-    name: 'AMP PD - 2019_v1beta_0220',
-    origin: 'https://amp-pd-data-explorer.appspot.com',
+    name: 'AMP PD - 2019_v1release_1015',
+    origin: 'https://amp-pd-data-explorers.appspot.com',
```
mbookman: Please use the public URLs if you can:

- data-explorer.amp-pd.org
- clinical-data-explorer.amp-pd.org

Otherwise LGTM.

Thanks!

mbookman (changes requested):

mbookman (approved):
LGTM

src/data/datasets.js
```diff
@@ -6,10 +6,15 @@ const datasets = [
     origin: 'https://test-data-explorer.appspot.com'
   },
   {
-    name: 'AMP PD - 2019_v1beta_0220',
-    origin: 'https://amp-pd-data-explorer.appspot.com',
+    name: 'AMP PD - 2019_v1release_1015',
+    origin: 'https://data-explorer.amp-pd.org',
```
panentheos, line 10: There is code in `workspace/Data.js` that looks for the string 'appspot.com' to detect which data explorer it came from. If we're using different URLs here, we might need to change that logic as well.

melissachang: Good catch. I recommend changing back to appspot. Note that AMP PD users won't see the appspot, they'll see https://app.terra.bio/#library/datasets/public/AMP%20PD.../data-explorer

Willy, can you also manually test the following:
- [one time setup] In local Terra, create a group called amp-pd-researchers. (This is to pass auth domain check.)
- In local Terra, from data library, go to AMP PD DE and save a cohort
- From Data tab, open cohort in DE. Confirm cohort is opened in same tab (as opposed to in a new tab)

wnojopra: Thanks!

@melissachang , when creating the amp-pd-researchers or amp-pd-clinical-access group on local Terra, I get a 409 error, "A resource of this type and name already exists". Is there a way around this?

And what do you think about this suggested change to getDataset in `workspace/Data.js` - we try to find the dataset by origin first, then by name. Something like:

```
  dataset = _.find({ origin: new URL(dataExplorerUrl).origin }, datasets)
  if (!dataset) {
    const datasetName = unescape(dataExplorerUrl.split(/datasets\/(?:public\/)?([^/]+)\/data-explorer/)[1])
    return _.find({ name: datasetName }, datasets)
  }
```

melissachang: That error is normal (in dev). If you refresh, you should see the group.

That change looks good to me. (Please keep the comments I had.) Please test with cohort from public DE, in addition to AMP PD DE. Thanks.

wnojopra: Great, thanks. Updated the code, and emailed you a screencast of the manual test.

wnojopra:
@panentheos @melissachang I believe I've addressed all the comments. Does the rest of the code look good to you?

melissachang (approved):
Sorry, didn't realize I was a reviewer

mbookman:
Thanks! Can this now be merged?

melissachang:
@panentheos still needs to approve

panentheos (approved):
Looks good.

PR #1917, opened 2019-10-30
User will be able to delete workspaces with long names with text that wraps around [SATURN-1195]

cameron-ardell:


panentheos (approved):
Looks good.

PR #1918, opened 2019-11-01
Fix Copy link address for files in "Files" viewer in Data tab [SATURN-1142]

ehanna4:


src/pages/workspaces/workspace/Data.js
```diff
@@ -800,7 +800,7 @@ const BucketContent = _.flow(
     const prefixParts = _.dropRight(1, prefix.split('/'))
     const makeBucketLink = ({ label, target, onClick }) => h(Link, {
       style: { textDecoration: 'underline' },
-      href: `gs://${bucketName}/${target}`,
+      href: `${target}`,
```
panentheos: nit: We could get rid of the interpolation and just pass the value directly.

panentheos (approved):
Looks good, see minor comment.

PR #1921, opened 2019-11-04
Add dataJoined User Property for Appcues on-boarding [SATURN-1146]

ehanna4:
We want to use Appcues to show an on-boarding flow for new users. Appcues uses _User Properties_ to make decisions such as determining who the new users are. This change adds a _User Property_ called _dateJoined_, populated with the date that a user first logs into Terra (we are also already sending a unique user Id) so that Appcues can make this differentiation.
[ticket](https://broadworkbench.atlassian.net/browse/SATURN-1146?atlOrigin=eyJpIjoiZjQwNDA2YjBhZTY1NDg1YjhiYjdlZGFjMzFhOWI5MGIiLCJwIjoiaiJ9)

Note - This change also leverages _memoize_ so that the call to get the first-log-in-timestamp only occurs once and is cached (this value is also used in the NPS survey logic so are avoiding unnecessary network calls with this implementation)

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -131,6 +131,11 @@ const getServiceAccountToken = Utils.memoizeAsync(async (namespace, token) => {
 
 export const saToken = namespace => getServiceAccountToken(namespace, getUser().token)
 
+const getFirstTimeStamp = Utils.memoizeAsync(async token => {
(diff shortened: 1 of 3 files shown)


panentheos (approved):
Looks good!

PR #1923, opened 2019-11-05
Fix url for call_interval_list in Hg38 Reference Table

ehanna4:
ticket: https://broadworkbench.atlassian.net/browse/SUP-226


panentheos (approved):
Looks good.

It might be worth communicating that because this data gets copied to the workspace, anyone who previously added it would need to remove and re-add it in order to see the change.

zarsky-broad:
Do they even have to remove first? Or does adding overwrite?

ehanna4:
They do have to re-import it to see the updated link. How/where should I communicate this? In the Suggested Release Notes?

zarsky-broad:
That's the best idea I can think of, yeah

PR #1925, opened 2019-11-06
Prevent users from updating 'description' from workspace data [SATURN-1160]

ehanna4:
This PR solves the reported issue of having a 'description' entry in workspace data update the overall workspace description field.  [ticket](https://broadworkbench.atlassian.net/browse/SATURN-1160?atlOrigin=eyJpIjoiMGIwYzdiOWNiZGU4NGUyZWFhOTRjZTRhOTE1YWNiMmQiLCJwIjoiaiJ9)

It also stops users from being able to add an entry in workspace data with a key starting with 'referenceData_' (not reported but also found this to be a bug when testing)

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -148,11 +148,11 @@ const LocalVariablesContent = class LocalVariablesContent extends Component {
       ...filteredAttributes, ...(creatingNewVariable ? [['', '']] : [])
     ]
 
-    const inputErrors = editIndex !== undefined && editIndex >= 0 && [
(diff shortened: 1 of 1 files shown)


panentheos (approved):
Looks good.

Tested various cases, both with and without existing workspace data.

This is a good change, but I'm curious how it was possible to accidentally overwrite an existing description, since it looks like the uniqueness check would prevent saving in that case.

ehanna4:
It only allowed users to update a description on a workspace that did not have one before (so not overwrite them, only add one when there wasn't one before). I was able to test this by creating a new workspace, not adding a description and then going in to the workspace data and adding an entry with a key of 'description'

PR #1926, opened 2019-11-06
refactor integration test utils [SATURN-1211]

panentheos:
This decomposes the integration utils a bit to promote reuse and composition.
* Decouples actions (e.g. `click`) from the generation of xpath to find the target element (e.g. `clickable`).
* Parameterizes the xpath generation to cleanly enable different kinds of lookups. (e.g. `label` vs `placeholder`)

The resulting code is slightly longer than before, but with a smaller number of concepts, it hopefully  reads just as clearly. And this approach will enable composing xpath into more complex selectors when needed (e.g. "within the section containing text 'Foo', find the button with text 'Bar'")


integration-tests/tests/find-workflow.integration-test.js
```diff
@@ -1,36 +1,36 @@
...
   await findText(page, 'Select a workspace')
   await firecloud.selectWorkspace(page, billingProject, workspaceName)
-  await firecloud.click(page, 'import-export-confirm-button')
+  await click(page, clickable({ text: 'Export to Workspace' }))
 
   /* This else/if is necessary to "hack" going back to localhost:3000-Terra after going to Firecloud,
    without these lines it will redirect to dev-Terra even if started out at localhost:3000-Terra */
   if (testUrl === 'http://localhost:3000') {
-    await findClickable(page, 'Yes')
+    await waitForElement(page, clickable({ textContains: 'Yes' }))
```
breilly2: `await waitForElement` reads a little funny. Maybe `findElement` instead (matching `findText`, `findIframe`, `findInGrid`, etc.)?

zarsky-broad:
Would you consider `clickClickable()` as another util, since it's a common use case? To match the previous `click()` behavior.

integration-tests/utils/integration-utils.js
```diff
@@ -29,37 +29,39 @@ const findInGrid = async (page, text) => {
   return await page.waitForXPath(`//*[@role="grid"][contains(normalize-space(.),"${text}")]`)
 }
 
-const exactlyFindClickable = (page, text) => {
-  return page.waitForXPath(`(//a | //*[@role="button"] | //button)[normalize-space(.)="${text}"]`)
+const clickable = ({ text, textContains }) => {
```
breilly2, line 32: TOL: Currently, `textContains` is the more frequently used property. I think the more frequently used property should be the most concise one.

There are two ways to accomplish this:
1. Rename these to something like `{ text, exactText }`.
2. Change the current tests to use `text` as it's currently defined (updating the search text as needed).

This is an opportunity to be opinionated about which form is preferred.

Side note: I recognize that the existing naming is consistent with the `contains` xpath function. We could also choose to keep the current naming for consistency with xpath vocabulary. IMO (70%)*, these functions exist as an abstraction on top of xpath so consistency with xpath vocabulary is less important than clarity/fluency of these functions' APIs.

\* Trying out a notation to describe how strongly I hold this opinion.

integration-tests/utils/integration-utils.js
```diff
@@ -29,37 +29,39 @@ const findInGrid = async (page, text) => {
...
+  const base = '(//a | //*[@role="button"] | //button)'
+  if (text) {
```
breilly2, line 34: Did you intentionally choose to favor exact text over partial match in the case that both are specified? I think I agree with that choice.

TOL: An alternative would be something like `{ text, partialMatch = false}` to make self-contradictory parameters impossible.

integration-tests/utils/integration-utils.js
```diff
@@ -76,17 +78,25 @@ const signIntoTerra = async page => {
   return page.evaluate(token => window.forceSignIn(token), bearerToken)
 }
 
+const waitForElement = (page, xpath) => {
+  return page.waitForXPath(xpath)
+}
+
+const svgText = ({ textContains }) => {
```
breilly2, line 85: `textContains` should be consistent with whatever we choose to do with `clickable`.

integration-tests/utils/integration-utils.js
```diff
@@ -29,37 +29,39 @@ const findInGrid = async (page, text) => {
...
-  return page.waitForXPath(`(//a | //*[@role="button"] | //button)[contains(normalize-space(.),"${text}") or contains(@aria-label,"${text}")]`)
-}
-
-const click = async (page, text) => {
-  return (await findClickable(page, text)).click()
+const click = async (page, xpath) => {
+  return (await page.waitForXPath(xpath)).click()
 }
 
 const findText = (page, text) => {
```
breilly2: As `findText` is currently implemented,`text` should be `textContains` to be consistent with the current naming convention. Ultimately it should be consistent with `clickable` and `svgText` (and `labelContains` for `input` and `select`).

Change request note: This is the only inconsistency I noticed that we should definitely fix in this PR. I have not looked at more than the diff context that Github shows me so it's possible that there are others.

integration-tests/tests/import-cohort-data.integration-test.js
```diff
@@ -1,26 +1,25 @@
...
+  await click(page, clickable({ textContains: '1000 Genomes Low Coverage' }))
 
   const frame = await findIframe(page)
-  await dataExplorer.clickTextInAnyNS(frame, 'Has WGS Low')
-  await click(frame, 'Save cohort')
-  await dataExplorer.fillIn(frame, 'name', cohortName)
-  await dataExplorer.click(frame, 'Save')
+  await click(frame, svgText({ textContains: 'Has WGS Low Cov' }))
+  await click(frame, clickable({ textContains: 'Save cohort' }))
+  await fillIn(frame, input({ placeholder: 'cohort name' }), cohortName)
```
breilly2, line 17: TOL: Ouch! Scary that we have to rely on placeholder to find this input. 👍 to making that reliance explicit and clearly visible in the test code!

breilly2 (commented):
Great change overall! I really like the composability and I agree that it's a net win despite a slight increase in verbosity.

One requested change and quite a bit of TOL. Curious to hear your thoughts.

integration-tests/tests/find-workflow.integration-test.js
panentheos: Yep, good call.

integration-tests/utils/integration-utils.js
panentheos: I'm on the fence here. The main reason `textContains` is more used at the moment is because that's what I happened to implement initially. But there are cases where one or the other would be preferable. And both formulations here seem decent. 🤷‍♂

integration-tests/utils/integration-utils.js
panentheos: They're intended to be mutually exclusive, so perhaps we should assert that. Also, there may very well be other non-exclusive options here, e.g. `disabled: false`

panentheos (commented):
Regarding `clickClickable`, it's only a two character difference between the two:
```
click(page, clickable({ textContains: 'x' }))
clickClickable(page, { textContains: 'x' })
```
And I think calling out exact vs contains is helpful, since I don't expect either one to be an overwhelming majority (see comments below)

petesantos (commented):
TOL: It seems like there are three main categories of functionality: query building, existence checking and taking an action.

Any thoughts on naming that may may serve as an indicator for what each function returns? 
For instance if it returning and xpath/building a query would something like withinSvgText or withinInput work?

Seeing that type of naming convention may make it more obvious that some functions are returning an XPath rather than a node or taking an action.

This may also work well in the future for something like:
click(page, withinModal(...), withinClickable(...))
click(page, withinGrid(...), withinCell(...), withinClickable(...))
Where the "withins" could be flowed or joined together

panentheos:
@petesantos So far, the overall convention I'm suggesting is roughly _nouns_ vs _verbs_, where the nouns are the xpath helpers, and the verbs actually take some action (waiting, clicking, etc). The 'within' notion is about composition, which there aren't any examples of yet, but in the current scheme I'm imagining something like:
```
click(page, within(section({ textContains: 'Genomes' }), clickable({ text: 'Browse' })))
```
In other words, express nesting (or other relationships) by combining the basic building blocks.

integration-tests/utils/integration-utils.js
panentheos: I thought about this a little more, and given that substring matching is a more complex operation than regular equality, I like that it has the slightly more interesting name.

zarsky-broad (approved):
I think we've gone back and forth on this enough, let's live with it for a bit and see how we feel.

PR #1927, opened 2019-11-06
Resolved clicking stop runtime on Terminal screen shows starting runtime msg [SATURN-1101]

cameron-ardell:
ticket [here](https://broadworkbench.atlassian.net/browse/SATURN-1101)


cameron-ardell:
I did notice a bug separately when trying to test this. When deleting a notebook runtime and then starting one, we get this error `Cannot read property 'clusterUrl' of null`. Which appears already was happening in the code before my work as far I have examined. I think this is worth another ticket.

zarsky-broad:
This would be easier using utils.cond.

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
```diff
@@ -101,10 +101,18 @@ const TerminalLauncher = _.flow(
       })
     } else {
       return div({ style: { padding: '2rem' } }, [
-        spinner({ style: { color: colors.primary(), marginRight: '0.5rem' } }),
-        (clusterStatus === 'Creating' || !cluster) ?
-          'Creating runtime environment. You can navigate away and return in 5-10 minutes.' :
-          'Starting runtime environment, this may take up to 2 minutes.'
+        (clusterStatus !== 'Stopped') && spinner({ style: { color: colors.primary(), marginRight: '0.5rem' } }),
+        (() => {
+          if (clusterStatus === 'Creating' || !cluster) {
```
ehanna4: nit/tol: for consistency would it make sense to have all the status reports call the space the same name, either 'notebook runtime environment' or 'runtime environment'?

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
```diff
@@ -101,10 +101,18 @@ const TerminalLauncher = _.flow(
       })
     } else {
       return div({ style: { padding: '2rem' } }, [
-        spinner({ style: { color: colors.primary(), marginRight: '0.5rem' } }),
-        (clusterStatus === 'Creating' || !cluster) ?
-          'Creating runtime environment. You can navigate away and return in 5-10 minutes.' :
-          'Starting runtime environment, this may take up to 2 minutes.'
+        (clusterStatus !== 'Stopped') && spinner({ style: { color: colors.primary(), marginRight: '0.5rem' } }),
+        (() => {
```
breilly2: I think this would read nicely as a `Utils.cond`. Doing so would eliminate the need to create a function and immediately invoke it. Let's pair on this. Tag me in anytime.

breilly2:
I got the same `Cannot read property 'clusterUrl' of null` error when trying to open a terminal when I didn't yet have a runtime environment.

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
breilly2: Ah, I thought this felt familiar. Isaac made this comment already as a general comment which is why I didn't notice it when I made my comment. Anyway, still happy to pair on this.

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
```diff
@@ -101,10 +101,18 @@ const TerminalLauncher = _.flow(
...
+          if (clusterStatus === 'Creating' || !cluster) {
+            return 'Creating runtime environment. You can navigate away and return in 5-10 minutes.'
+          } else if (clusterStatus === 'Stopping') {
+            return 'Notebook runtime environment is stopping. You can restart it after it finishes.'
+          } else if (clusterStatus === 'Stopped') {
+            return 'Notebook runtime environment is stopped. Start it to edit your notebook.'
```
breilly2: "Start it to edit your notebook or use the terminal."

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
```diff
@@ -101,10 +101,18 @@ const TerminalLauncher = _.flow(
...
+          } else {
```
breilly2: I think we should explicitly check for `'Starting'`, even if just for documentation/readability. Also, that will allow us to use the `else` case to handle unexpected status.

breilly2: Follow-up: here are the status values defined in the leonardo source (https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/model/google/GoogleModel.scala#L55):
```
  case object Unknown extends ClusterStatus
  case object Creating extends ClusterStatus
  case object Running extends ClusterStatus
  case object Updating extends ClusterStatus
  case object Error extends ClusterStatus
  case object Deleting extends ClusterStatus

  // note: the below are Leo-specific statuses, not Dataproc statuses
  case object Deleted extends ClusterStatus
  case object Stopping extends ClusterStatus
  case object Stopped extends ClusterStatus
  case object Starting extends ClusterStatus
```
We should probably be ready for all of these, or at least specifically handle `'Starting'`.

ehanna4:
- Added more cluster status options per @breilly2's comment and observed the correct messages for:
            - Creating
            - Stopping
            - Stopped
            - Starting
            - Updating
            - Error
            - Unknown
            - null, "", and incorrect statuses

- Getting the null error (that Cameron is working on in the other ticket) when trying to test for:
            - Deleting
            - Deleted
(so not able to fully see if those are working or not but can confirm once that null bug is fixed.)

-̶ ̶N̶o̶t̶ ̶a̶b̶l̶e̶ ̶t̶o̶ ̶g̶e̶t̶ ̶a̶ ̶c̶l̶u̶s̶t̶e̶r̶ ̶s̶t̶a̶t̶u̶s̶ ̶o̶f̶ ̶U̶p̶d̶a̶t̶i̶n̶g̶ ̶(̶n̶e̶e̶d̶ ̶t̶o̶ ̶l̶o̶o̶k̶ ̶i̶n̶t̶o̶ ̶L̶e̶o̶ ̶c̶o̶d̶e̶ ̶m̶o̶r̶e̶ ̶f̶o̶r̶ ̶w̶h̶y̶)̶
-̶ ̶-̶D̶i̶d̶ ̶n̶o̶t̶ ̶t̶e̶s̶t̶ ̶E̶r̶r̶o̶r̶ ̶o̶r̶ ̶U̶n̶k̶n̶o̶w̶n̶,̶ ̶a̶s̶s̶u̶m̶i̶n̶g̶ ̶t̶h̶e̶s̶e̶ ̶a̶r̶e̶ ̶g̶e̶n̶e̶r̶a̶l̶l̶y̶ ̶g̶o̶i̶n̶g̶ ̶t̶o̶ ̶b̶e̶ ̶h̶a̶n̶d̶l̶e̶d̶ ̶b̶y̶ ̶o̶t̶h̶e̶r̶ ̶m̶e̶c̶h̶a̶n̶i̶s̶m̶s̶ ̶b̶u̶t̶ ̶a̶d̶d̶e̶d̶ ̶m̶e̶s̶s̶a̶g̶e̶s̶ ̶i̶f̶ ̶a̶l̶l̶ ̶e̶l̶s̶e̶ ̶f̶a̶l̶l̶s̶ ̶t̶h̶r̶o̶u̶g̶h̶-̶

EDIT:
 thanks to @breilly2 able to test all the statuses except Deleting and Deleted (known bug)

ehanna4:
Clusterstatus messages have been confirmed with @sharmatime, all is good to go, ready for review

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
```diff
@@ -101,10 +101,20 @@ const TerminalLauncher = _.flow(
       })
     } else {
       return div({ style: { padding: '2rem' } }, [
+        'Creating Stopping Starting Updating Deleting'.includes(clusterStatus) &&
         spinner({ style: { color: colors.primary(), marginRight: '0.5rem' } }),
-        (clusterStatus === 'Creating' || !cluster) ?
-          'Creating runtime environment. You can navigate away and return in 5-10 minutes.' :
-          'Starting runtime environment, this may take up to 2 minutes.'
+        Utils.cond(
```
petesantos: nit: this may be cleaner with a switchCase since all conditions are being checked against the same variable.

petesantos (approved):
Looks good.
Tested by starting and stopping and updating notebook runtimes.
Did notice the updating state shows the creating message, but after talking with Emily it looks like leo is sending the wrong status.

ehanna4:
I am following up with Leo team to see if there is an issue about the clusterStatus of 'Updating' coming through. 

At line 295 in this code there is a comment that changing an existing cluster's size should result in a status of Updating: 
https://github.com/DataBiosphere/leonardo/blob/3e0879e2586e2ddc303a3491557d31a54a91b40a/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/service/LeonardoService.scala#L295
However this is not happening (as far as I can see)

I was able to test that the 'Updating' conditional message added here does show by setting it with the ajx.overrides, but neither Pete nor I were able to see an 'Updating' status after updating the cluster size in regular testing circumstances.

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
zarsky-broad: +1

ehanna4:
Clarification from @zarsky-broad is that currently we are doing deletes even when updating (except in super rare instances) so it is currently expected that we cannot get an 'Updating' status and instead are seeing the 'Creating' one.

petesantos (approved):
Change to switchCase looks good

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
```diff
@@ -101,10 +101,20 @@ const TerminalLauncher = _.flow(
...
+          ['Creating', () => 'Creating notebook runtime environment. You can navigate away and return in 3-5 minutes.'],
+          ['Stopping', () => 'Notebook runtime environment is stopping, which takes ~4 minutes. You can restart it after it finishes.'],
+          ['Starting', () => 'Starting notebook runtime environment, this may take up to 2 minutes.'],
+          ['Updating', () => 'Updating notebook runtime environment. You can navigate away and return in 3-5 minutes.'],
+          ['Deleting', () => 'Deleting notebook runtime environment, you can create a new one after it finishes.'],
+          ['Stopped', () => 'Notebook runtime environment is stopped. Start it to edit your notebook or use the terminal.'],
+          ['Deleted', () => 'No existing notebook runtime environment, you can create a new one to edit your notebook or use the terminal.'],
+          ['Error', () => 'Error with the notebook runtime environment, please try again.'],
+          ['Unknown', () => 'Error with the notebook runtime environment, please try again.'],
+          [Utils.DEFAULT, () => 'Invalid notebook runtime environment state, please try again.']
```
rtitle, line 116: Hi, I was looking at a user-liaison question and noticed this PR. The user was saying they keep getting this `DEFAULT` error message when launching a terminal and have to manually refresh the page. When I try to launch a terminal I briefly see this message then the page seems to refresh by itself. Any idea what the issue might be for the user?

breilly2: Looks like this default case is catching the condition where we don't yet know the cluster status (the ajax call hasn't returned yet) or there is no cluster. We should clean that up. That said, I expect the behavior you experienced (brief flash of this message), not requiring a manual refresh. Is there a difference between the cases of having a running cluster, a paused cluster, and no cluster?

rtitle: Hm, I think I see the expected behavior in all cases: a brief flash of this message followed by an action (e.g. cluster creation) and the appropriate message. I'm not sure why the user was seeing it get stuck in the default message.

Thanks for replying!

PR #1928, opened 2019-11-06
fix integration test screenshots [SATURN-1212]

panentheos:
* Adds necessary `png` extension to filename.
* Makes screenshot capture opt-in via the environment variable.
* Explicitly creates the directory on Circle.

Tested locally and on Circle by temporarily pushing a breaking change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,7 +159,7 @@ jobs:
             JEST_JUNIT_OUTPUT_DIR: "/tmp/test-results/integration-test-results.xml"
             SCREENSHOT_DIR: "/tmp/failure-screenshots"
           command: |
-            mkdir -p /tmp/failure-screenshots
+            mkdir -p ${SCREENSHOT_DIR}
             TERRA_TOKEN=$(node getBearerToken.js "${ALPHA_TEST_UA}" "${ALPHA_USER_SA_KEY_JSON}") \
             npx npm@6.12 run test -- --ci --maxWorkers=2 --reporters=default --reporters=jest-junit
       - store_test_results:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -159,6 +159,7 @@ jobs:
             JEST_JUNIT_OUTPUT_DIR: "/tmp/test-results/integration-test-results.xml"
             SCREENSHOT_DIR: "/tmp/failure-screenshots"
           command: |
+            mkdir -p /tmp/failure-screenshots
```
breilly2: Assuming we have access to the env above, this could be `mkdir -p ${SCREENSHOT_DIR}`.

integration-tests/utils/integration-config.js
```diff
@@ -3,7 +3,7 @@ const {
   TERRA_TOKEN: bearerToken,
   BILLING_PROJECT: billingProject = 'general-dev-billing-account',
   WORKFLOW_NAME: workflowName = 'haplotypecaller-gvcf-gatk4',
-  SCREENSHOT_DIR: screenshotDir = 'failure-screenshots'
```
breilly2, line 6: TOL: I'm ever so slightly in favor of capturing screenshots by default. If we assume that the tests are pretty stable, then we'll most often see failures in edge cases that we might not be able to readily reproduce after opting-in.

breilly2 (approved):
👍 In this case, I did not test myself. I trust the testing that you explicitly said you did.

integration-tests/utils/integration-config.js
panentheos: Let's revisit this in a followup PR. It's a fair point, I just want to make sure we get the management of the files right.

breilly2 (approved):
:+1: Thanks for fixing this!

PR #1929, opened 2019-11-07
Create environment setup script [SATURN-1198]

zarsky-broad:
After discussion offline, this doesn't register a user because in the near future test users will be parameterized.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,8 +159,8 @@ jobs:
             JEST_JUNIT_OUTPUT_DIR: "/tmp/test-results/integration-test-results.xml"
             SCREENSHOT_DIR: "/tmp/failure-screenshots"
           command: |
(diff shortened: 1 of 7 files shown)


breilly2:
How do I run the script? I see a new `.js` file that's probably it, though no new npm script config or anything.

breilly2:
I think I answered my own question:
```
node integration-tests/scripts/initializeEnvironment.js
```
Request: add `integration-tests/README.md` with something like:

> Run `node integration-tests/scripts/initializeEnvironment.js` and follow the prompts.

Could also add some other stuff about the how to run the tests, what they depend on, etc. Just some basic info... nothing crazy.

breilly2 (changes requested):
Looking really good! I like the prompts! :)

One bug I found: it fails if there isn't already a `featured-methods.json` object in the alerts bucket.

breilly2 (approved):
👍 Very nice and robust! (Would still like a readme.) Great work!

PR #1931, opened 2019-11-08
Fix react warning about text import [SATURN-1194]

ehanna4:
- update the handling of onChange and readOnly to fix react warning about not properly handling these 
- add an error message for users trying to type in the paste only region
- update a few UI aspects of modal, including using the textarea styling used elsewhere in Terra and making things less wordy


src/components/input.js
```diff
@@ -275,7 +275,7 @@ export const PasteOnlyInput = ({ onPaste, readOnly = false, ...props }) => {
     onPaste: e => {
       onPaste(e.clipboardData.getData('Text'))
     },
-    onChange: undefined,
-    readOnly
+    readOnly,
+    onChange
```
panentheos: Since these are just passed through, we can just remove them entirely and they'll get rolled up into `...props`. (note that `readOnly` is already false by default)

src/libs/data-utils.js
```diff
@@ -293,20 +294,22 @@ export const EntityUploader = class EntityUploader extends Component {
...
-          'where ',
-          span({ style: { fontFamily: 'monospace', fontWeight: '600' } }, ['[type]']),
-          ` is the desired name of the data table in ${getAppName()}. For example, to create or update a `,
-          span({ style: { fontFamily: 'monospace', fontWeight: '600' } }, ['participant']),
-          ' table use ',
-          span({ style: { fontFamily: 'monospace', fontWeight: '600' } }, ['entity:participant_id']),
-          div({ style: { margin: '0.5rem 0 1rem' } }, ['All of the values in the ID column must be unique.']),
+          div({
+            style: { padding: '0 0 1rem' }
+          }, [
```
zarsky-broad: I'd condense this into one line

src/libs/data-utils.js
```diff
@@ -293,20 +294,22 @@ export const EntityUploader = class EntityUploader extends Component {
...
+            'Choose the data import option below. For more info on the table',
+            h(Link, {
+              'aria-label': 'Click for table info',
+              href: 'https://support.terra.bio/hc/en-us/articles/360025758392'
+            }, [
+              ' click here.'
+            ])
```
zarsky-broad: I'd condense this into one line, maybe make the whole sentence a link? Like, 'Click here for more info on the table.`

src/libs/data-utils.js
```diff
@@ -320,7 +323,7 @@ export const EntityUploader = class EntityUploader extends Component {
               h(TooltipTrigger, { content: 'Tab Separated Values', side: 'bottom' },
                 [span({ style: { textDecoration: 'underline dashed' } }, 'TSV')]),
               ' file containing your data: '
-            ]) : div(['Paste the data directly below:']),
+            ]) : div(['Copy and paste tab separated data directly below:']),
```
zarsky-broad:
```suggestion
            ]) : div(['Copy and paste tab separated data here:']),
```

src/libs/data-utils.js
```diff
@@ -380,11 +385,9 @@ export const EntityUploader = class EntityUploader extends Component {
               ...Utils.newTabLinkProps
             }, ['Learn more ', icon('pop-out', { size: 12 })])
           ]),
-          isInvalid && div({
-            style: { color: colors.danger(), fontWeight: 'bold', fontSize: 12, marginTop: '0.5rem' }
-          },
-          ['Invalid format: Data does not start with entity or membership definition.']
-          )
+          div({ style: errorTextStyle },
+            [(isInvalid && 'Invalid format: Data does not start with entity or membership definition.') ||
```
zarsky-broad: Wouldn't we potentially want both? What if I've pasted something incorrect but then I try to edit it in place?

src/libs/data-utils.js
```diff
@@ -293,20 +294,22 @@ export const EntityUploader = class EntityUploader extends Component {
...
-          span({ style: { fontFamily: 'monospace', fontWeight: '600' } }, ['participant']),
-          ' table use ',
-          span({ style: { fontFamily: 'monospace', fontWeight: '600' } }, ['entity:participant_id']),
-          div({ style: { margin: '0.5rem 0 1rem' } }, ['All of the values in the ID column must be unique.']),
+          div({
+            style: { padding: '0 0 1rem' }
+          }, [
+            'Choose the data import option below. For more info on the table',
+            h(Link, {
+              'aria-label': 'Click for table info',
```
panentheos: Since this is already a text link, it would be better for overall accessibility to tweak the language so the link could be more self-descriptive, instead of adding a separate aria label. Something like: 'Learn more about data tables.'

src/libs/data-utils.js
```diff
@@ -293,20 +294,22 @@ export const EntityUploader = class EntityUploader extends Component {
...
+              href: 'https://support.terra.bio/hc/en-us/articles/360025758392'
+            }, [
```
panentheos: This link should use `Utils.newTabLinkProps` so it opens in a new tab.

src/libs/data-utils.js
```diff
@@ -380,11 +385,9 @@ export const EntityUploader = class EntityUploader extends Component {
...
             }, ['Learn more ', icon('pop-out', { size: 12 })])
           ]),
-          isInvalid && div({
-            style: { color: colors.danger(), fontWeight: 'bold', fontSize: 12, marginTop: '0.5rem' }
-          },
-          ['Invalid format: Data does not start with entity or membership definition.']
-          )
+          div({ style: errorTextStyle },
+            [(isInvalid && 'Invalid format: Data does not start with entity or membership definition.') ||
+            (showInvalidEntryMethodWarning && 'Invalid Data Entry Method: Copy and paste only')])
```
panentheos: If these two cases need to be mutually exclusive, it might be cleaner to use a `Utils.cond` here

src/libs/data-utils.js
ehanna4: Both as in pasting and text editing? The decided on approach is to only allow copy&pasting right now in an effort to add the minimal requested feature and also avoiding user error particularly with the tab separation. If in the future there is a need/desire to add editing that is fine but will require figuring out how to deal with the tab separation and other user editing pitfalls here appropriately.

panentheos (approved):
Looks good!

PR #1932, opened 2019-11-08
Update deps [SATURN-1218]

zarsky-broad:
Read changelogs and tested, there are a bunch of updated packages so some more testing couldn't hurt, but I think I was pretty thorough.


package.json
```diff
@@ -11,32 +11,32 @@
...
+    "qs": "^6.9.1",
+    "react": "~16.11.0",
     "react-collapse": "^5.0.0",
-    "react-dom": "npm:@hot-loader/react-dom@~16.10.2",
-    "react-draggable": "^4.0.3",
+    "react-dom": "npm:@hot-loader/react-dom@~16.11.0",
+    "react-draggable": "^4.1.0",
     "react-dropzone": "^10.1.10",
-    "react-focus-lock": "^2.1.0",
+    "react-focus-lock": "~2.1.1",
```
petesantos: This update causes the issue we have been seeing. I am looking into this now

src/pages/Profile.js
```diff
@@ -238,7 +238,7 @@ const FenceLink = ({ provider, displayName }) => {
    * Render
    */
   const isBusy = isLoadingStatus || isLoadingAuthUrl || isLinking
-  const expireTime = addDays(30, parseISO(issuedAt)) // can't use parseJSON because dates don't specify time offset
+  const expireTime = addDays(30, parseJSON(issuedAt))
```
petesantos, line 241: Curious  - did the version update cause a breakage?

panentheos (approved):
Tested by clicking around, looks good.

src/pages/Profile.js
zarsky-broad: No, `parseJSON` is preferred because it's lighter; the version update gave it the ability to parse dates without offsets.

petesantos (approved):

PR #1933, opened 2019-11-14
Fix dockstore link to reflect  the current version of workflow [SATURN-1141]

ehanna4:
Previously the link for a dockstore sourced workflows did not update when changing versions, with these changes the version will show and also the redirect link will go to the version specific workflow.

To test this [workspace](https://bvdp-saturn-dev.appspot.com/#workspaces/this-is-a-testing-project-name/Test%20for%20Dockstore%20versioned%20link) has a dockstore sourced workflow with multiple versions and a terra sourced workflow with multiple snapshots

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -329,7 +329,7 @@ export const methodLink = config => {
   const { methodRepoMethod: { sourceRepo, methodVersion, methodNamespace, methodName, methodPath } } = config
   return sourceRepo === 'agora' ?
     `${getConfig().firecloudUrlRoot}/?return=${returnParam()}#methods/${methodNamespace}/${methodName}/${methodVersion}` :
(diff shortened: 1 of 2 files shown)


petesantos (commented):
Will the methodVersion field always be populated in the config?

ehanna4:
@petesantos I don't believe it's a guarantee that the methodVersion field will be populated in the config, however even the workflows that only have a single version/snapshot all appear to have a 'version 1', so it seems to me if the version is not in the config there's likely an issue with the workflow outside Terra?

petesantos (approved):

zarsky-broad (approved):
TOL: I feel like we should somehow combine these functionalities, getting the link and building the name, since they both use the same separators with the same things in the same order. That's probably for another time, and super minor.

PR #1934, opened 2019-11-14
Add prefer-template eslint rule

petesantos:


.eslintrc.js
```diff
@@ -1,4 +1,4 @@
-module.exports = {
+  module.exports = {
```
zarsky-broad:
```suggestion
module.exports = {
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -640,7 +639,7 @@ class EntitiesContent extends Component {
 
     const header = _.join('\t', [`entity:${entityKey}_id`, ...attributeNames])
 
-    const entityTsv = _.join('\n', [header, ..._.map(entityToRow, sortedEntities)]) + '\n'
+    const entityTsv = `${_.join('\n', [header, ..._.map(entityToRow, sortedEntities)])}\n`
```
zarsky-broad, line 642: What about this for these?
```suggestion
    const entityTsv = _.join('\n', [header, ..._.map(entityToRow, sortedEntities), ''])
```

panentheos: FWIW that suggestion seems less clear to me.

panentheos (approved):
Click tested, looks good, see whitespace comment.

src/pages/workspaces/workspace/Data.js
petesantos: kept this as-is, if there is a strong objection to not changing it let me know

PR #1937, opened 2019-11-14
update amppd access instructions [SATURN-1228]

zarsky-broad:


src/data/datasets.js
```diff
@@ -8,12 +8,14 @@ const datasets = [
   {
     name: 'AMP PD - 2019_v1release_1015',
     origin: 'https://data-explorer.amp-pd.org',
-    authDomain: 'amp-pd-researchers'
+    authDomain: 'amp-pd-researchers',
+    isAmpPd: true
```
panentheos: nit: Since these booleans are mutually exclusive, we could encode this as a single string field, e.g. `partner: 'amp-pd'`

panentheos (approved):
Click tested, looks good.

PR #1938, opened 2019-11-15
Correct issue with react-focus-lock [SATURN-1226]

petesantos:
The code change responsible for the issue is [here](https://github.com/theKashey/react-focus-lock/compare/v2.1.0...v2.1.1#diff-757809db6137e7d28cf62c382a00aead) - Lock.js lines 64-71 where the focus setting gets deferred in version 2.1.1 whereas it is not in v2.1.0

What happens is:

1. The focus trap is set to return focus when the trapped component unmounts
  a. This action is deferred ( I believe with a 0 timeout)
2. The trapped component unmounts
3. Our modal, which is not part of focus lock, mounts and gains focus
4. The focus trap executes the deferred focus code

This change uses the API's in the modal to gain back focus and properly return focus. I am essentially using the technique to do this in react-focus-lock, but am not thrilled with the solution (adding additional refs, etc...)

I tried a bunch of combinations of the modal and focus lock API, but could not get the desired behavior. This solution does work and gives us the same behavior that currently exists.

Tested by clicking through the application where modals are used and ensuring that setting focus is behaving correctly

To test the before / after behavior:
1. Run npm install on this branch
2. Switch to another branch or dev and open a modal on the workspaces page 
    a. You will see that the modal does not get focus
3. Switch back to this branch and the problem should be resolved


zarsky-broad:
So, the big question, then: is the complication that this introduces worth being able to update the package?

petesantos:
> So, the big question, then: is the complication that this introduces worth being able to update the package?

Overall I think it is. The focus locking code handles a tricky aspect of our app and has improved the accessibility and usability since we added it.  It is being maintained and the developer on it is heavily involved in many projects including the hot-loader.

The modal and focus lock do handle a lot of overlapping concerns when it comes to handling focus. It may make sense to look at creating our own modal at some point as we are starting to "turn off" a lot of the functionality that is baked in to the modal.

This solution works in both the 2.1.0 version of the focus trap and in the current version.
My biggest hesitation is that we don't typically manually handle these issues in our code base. At the same time, dealing with focus issues can be tricky and overall this solution doesn't seem too difficult to follow and is in line with the focus lock recommendation.

src/components/Modal.js
```diff
@@ -27,11 +28,25 @@ const styles = {
...
+  const previouslyFocusedNode = useRef()
   // react-modal applies aria-hidden to the app root *and* takes care of limiting what can be tab-focused - see appLoader.js
   return h(RModal, {
+    contentRef: node => modalNode.current = node,
     parentSelector: () => document.getElementById('modal-root'),
     isOpen: true,
-    onRequestClose: onDismiss,
+    onRequestClose: () => {
+      onDismiss()
+      previouslyFocusedNode.current.focus()
```
panentheos: It's possible to close the modal programmatically without coming through this code path, so we should probably use the mount/unmount cycle to handle this instead.

src/components/Modal.js
```diff
@@ -27,11 +28,25 @@ const styles = {
 
 const Modal = ({ onDismiss, title, titleExtras, children, width = 450, showCancel = true, cancelText = 'Cancel', showX, showButtons = true, okButton, ...props }) => {
   const titleId = Utils.useUniqueId()
+  const modalNode = useRef()
+  const previouslyFocusedNode = useRef()
   // react-modal applies aria-hidden to the app root *and* takes care of limiting what can be tab-focused - see appLoader.js
   return h(RModal, {
```
panentheos, line 40: Since we're handling focus ourselves now, can we turn off the Modal's default focus handling to simplify things?

panentheos (commented):
Approach looks good, see a few comments.

ehanna4 (approved):
Looks good to me.

- Went over flow of the code with Pete
- Tested changes by switching to the dev branch and upgrading the react-focus-lock library then clicked around and observed the incorrect focusing behavior when opening modals. I then switched to this branch and the focusing began behaving as I expected.

src/components/Modal.js
```diff
@@ -27,11 +28,30 @@ const styles = {
 
 const Modal = ({ onDismiss, title, titleExtras, children, width = 450, showCancel = true, cancelText = 'Cancel', showX, showButtons = true, okButton, ...props }) => {
   const titleId = Utils.useUniqueId()
+  const modalNode = useRef()
+  const previouslyFocusedNode = useRef()
+
+  Utils.useOnMount(() => {
+    previouslyFocusedNode.current = document.activeElement
+    return () => previouslyFocusedNode && previouslyFocusedNode.current && previouslyFocusedNode.current.focus()
```
panentheos: nit: The ref object itself should always be present, so I think this can be:
```suggestion
    return () => previouslyFocusedNode.current && previouslyFocusedNode.current.focus()
```

panentheos (approved):
Code looks good, didn't test.

PR #1939, opened 2019-11-18
[IA- 779] [SATURN-1240] Periodic Cookie Setter 

akarukappadath:
This feature is in preparation for RStudio.
Currently, we refresh the browser cookie with auth details through a custom notebooks extension. Since there is no equivalent for RStudio, this PR makes it so that Terra refreshes the cookie every fifteen minutes. This component will need to be moved to wherever RStudio is being rendered.


src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -452,6 +452,23 @@ const JupyterFrameManager = ({ onClose, frameRef }) => {
   return null
 }
 
+const PeriodicCookieSetter = ({ namespace, clusterName }) => {
+  const signal = useCancellation()
+
+  const periodicallySetCookie = withErrorReporting('Error setting cookie', async () => {
```
petesantos: We recently created a hook for the 15 minute signout that may simplify this code. Using the hook, this component could look like:

```
usePolling(300000)
withErrorReporting('Error setting cookie', () => Ajax().Jupyter.notebooks(namespace, clusterName).setCookie())
return null
```
This hook hasn't made it into dev yet, but it could be part of this PR

petesantos: I spoke with Brett - this approach would need some more work with an Ajax call

panentheos: With the error handler here, a single failed call would cause the loop to exit. Perhaps we should catch the error but continue looping, in case it was just a network blip. Also, we should think carefully about whether we actually want to report or ignore the error, since a string of failures might generate a lot of notifications.

panentheos (commented):
Nice self-contained component! See comment.

akarukappadath:
Note: In the event that the cookieSetter fails and the auth token doesn't get updated within an hour, the user is pushed back on to Preview mode for the notebook they were in.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -452,6 +452,23 @@ const JupyterFrameManager = ({ onClose, frameRef }) => {
   return null
 }
 
+const PeriodicCookieSetter = ({ namespace, clusterName }) => {
+  const signal = useCancellation()
+
+  const periodicallySetCookie = async () => {
+    while (!signal.aborted) {
+      withErrorIgnoring(() => Ajax().Jupyter.notebooks(namespace, clusterName).setCookie())
+      await Utils.delay(900000)
```
breilly2: Nit: Factor for clarity, i.e. `15 * 60 * 1000`.
Also, the PR description says 5 minutes but 900000 is 15 minutes. Which should it be?

akarukappadath: Ah good catch - I updated the code but forgot to update the PR description. We've decided 15 minutes would be better since the auth token expires in an hour

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -452,6 +452,23 @@ const JupyterFrameManager = ({ onClose, frameRef }) => {
   return null
 }
 
+const PeriodicCookieSetter = ({ namespace, clusterName }) => {
+  const signal = useCancellation()
+
+  const periodicallySetCookie = async () => {
+    while (!signal.aborted) {
+      withErrorIgnoring(() => Ajax().Jupyter.notebooks(namespace, clusterName).setCookie())
```
breilly2: `withErrorIgnoring` returns a function which is currently never called.
```suggestion
      withErrorIgnoring(() => Ajax().Jupyter.notebooks(namespace, clusterName).setCookie())()
```

config/dev.json
```diff
@@ -1,20 +1,19 @@
 {
-  "agoraUrlRoot": "https://agora.dsde-dev.broadinstitute.org",
+  "agoraUrlRoot": "https://agora-fiab.dsde-dev.broadinstitute.org:20443",
```
breilly2: Accidental commit of config?

akarukappadath: Agh sorry, just reverted it

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -452,6 +452,23 @@ const JupyterFrameManager = ({ onClose, frameRef }) => {
   return null
 }
 
+const PeriodicCookieSetter = ({ namespace, clusterName }) => {
+  const signal = useCancellation()
+
+  const periodicallySetCookie = async () => {
+    while (!signal.aborted) {
+      withErrorIgnoring(() => Ajax().Jupyter.notebooks(namespace, clusterName).setCookie())()
```
breilly2: One last thought. Not critical, but this can be cancellable.
```suggestion
      withErrorIgnoring(() => Ajax(signal).Jupyter.notebooks(namespace, clusterName).setCookie())()
```

breilly2 (approved):
👍 Looks good! One more minor thing I noticed and still the nit of factoring the time interval to `15 * 60 * 1000` but I won't block approval on that.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
akarukappadath: Since `setCookie` is a post, I thought it might be better to keep it from being cancellable

PR #1940, opened 2019-11-18
fix extraneous react key warnings [SATURN-1232]

panentheos:
This fixes a key react key warning when passing a `Fragment` as an error message. The deep merge was modifying the underlying react element and changing its behavior. In general we should probably avoid blindly deep merging, since it can cause subtle problems like this.

Tested by clicking through the UI.


petesantos (approved):
Looks good. Tested by copying a notebook and trying different valid and invalid names.

PR #1941, opened 2019-11-19
fix workflow import and launch [SATURN-1222]

zarsky-broad:
Some pre-CUJ3 (SATURN-1080) fixes.
* Imported workflows are named after their configs, if available, so use that in building the link.
* AJAX calls that create things shouldn't be cancelled.

(tested in the UI)


breilly2 (approved):
Witnessed the problem prior to these changes. Witnessed these changes fixing the proglem. LGTMob. 👍

breilly2:
Please add link to CUJ3 JIRA ticket to this PR.
Please fill in CLIA Impact for SATURN-1222 (probably N/A in this case).

PR #1943, opened 2019-11-19
Wait for nav when importing workflow [SATURN-1237]

zarsky-broad:
Makes the find workflow integration test run more reliably, especially in slower environments.


panentheos (approved):
Looks good.

PR #1944, opened 2019-11-19
Enable prevention of encoding the URI in routes (fixes import WDL) [PROD-300] [SATURN-1242]

petesantos:
I test this by pasting a the Terra import URI from Dockstore into my browser and modifying the URI to hit my localhost.

The URI I used is:
http://localhost:3000/#import-tool/dockstore/github.com/DataBiosphere/topmed-workflows/UM_variant_caller_wdl:1.32.0

This fails in with the current code, passes with this change

### DIFF
--- a/src/libs/routes.js
+++ b/src/libs/routes.js
@@ -60,13 +60,13 @@ const routes = _.flatten([
   NotFound.navPaths // must be last
 ])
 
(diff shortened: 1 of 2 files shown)


zarsky-broad (approved):
Verified manually: broken before, fixed now. Code looks good.

PR #1945, opened 2019-11-19
Run workflow integration test [SATURN-1080]

zarsky-broad:
Working reliably! Takes around 3.5-4 minutes locally.

### DIFF
--- a/integration-tests/jest-puppeteer.config.js
+++ b/integration-tests/jest-puppeteer.config.js
@@ -15,6 +15,7 @@ module.exports = {
   launch: {
     devtools: process.env.DEVTOOLS === 'true',
     headless: process.env.HEADLESS !== 'false',
+    slowMo: process.env.SLOW === 'true' ? 500 : undefined,
     // Workaround for issue when accessing cross domain iframes in puppeteer while not headless:
     // https://github.com/GoogleChrome/puppeteer/issues/4960
     args: (({ headfull, devtools }) => {
--- a/integration-tests/jest-puppeteer.config.js
+++ b/integration-tests/jest-puppeteer.config.js
@@ -15,7 +15,8 @@ module.exports = {
   launch: {
     devtools: process.env.DEVTOOLS === 'true',
     headless: process.env.HEADLESS !== 'false',
-    slowMo: process.env.SLOW === 'true' ? 500 : undefined,
+    slowMo: process.env.SLOW === 'true' ? 250 : undefined,
+    defaultViewport: { width: 1200, height: 800 },
     // Workaround for issue when accessing cross domain iframes in puppeteer while not headless:
     // https://github.com/GoogleChrome/puppeteer/issues/4960
     args: (({ headfull, devtools }) => {
--- a/integration-tests/scripts/initializeEnvironment.js
+++ b/integration-tests/scripts/initializeEnvironment.js
@@ -26,7 +26,7 @@ const wdl = `task echo_to_file {
   }
 
   runtime {
-    docker: "bashell/alpine-bash"
+    docker: "photon"
   }
 }
 
--- a/integration-tests/scripts/initializeEnvironment.js
+++ b/integration-tests/scripts/initializeEnvironment.js
@@ -26,7 +26,7 @@ const wdl = `task echo_to_file {
   }
 
   runtime {
-    docker: "alpine"
+    docker: "bashell/alpine-bash"
   }
 }
 
@@ -89,10 +89,10 @@ const setupMethod = async () => {
       namespace, name: `${name}-configured`,
       payload: JSON.stringify({
         ...configTemplate,
-        inputs: { 'echo_strings.echo_to_file.input1': 'foo' },
-        outputs: { 'echo_strings.echo_to_file.out': 'this.out' },
+        inputs: { 'echo_strings.echo_to_file.input1': 'this.input' },
+        outputs: { 'echo_strings.echo_to_file.out': 'this.output' },
         namespace, name: `${name}-configured`,
-        rootEntityType: 'participant'
+        rootEntityType: 'test_entity'
       }),
       entityType: 'Configuration'
     }
@@ -126,7 +126,7 @@ const setupMethod = async () => {
 
     const featuredMethodsUrl = `https://www.googleapis.com/storage/v1/b/firecloud-alerts-${env}/o/featured-methods.json`
 
-    const featuredMethods = await fetch(`${featuredMethodsUrl}?alt=media`, { headers: googleHeaders }).then(res => res.ok && res.json())
+    const featuredMethods = await fetch(`${featuredMethodsUrl}?alt=media`).then(res => res.ok && res.json())
 
     if (!featuredMethods || !featuredMethods.some(method => method.name === name && method.namespace === namespace)) {
       await fetchOk(
--- a/integration-tests/tests/run-workflow.integration-test.js
+++ b/integration-tests/tests/run-workflow.integration-test.js
@@ -1,7 +1,7 @@
 const pRetry = require('p-retry')
 const { testUrl, workflowName, billingProject } = require('../utils/integration-config')
 const { withWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, findElement, input, signIntoTerra, waitForNoSpinners, findInGrid, workspaceTab } = require('../utils/integration-utils')
+const { click, clickable, input, signIntoTerra, waitForNoSpinners, findInGrid, workspaceTab, findInDataTable } = require('../utils/integration-utils')
 
 
 const testEntity = { name: 'test_entity_1', entityType: 'test_entity', attributes: { input: 'foo' } }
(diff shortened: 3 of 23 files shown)


integration-tests/scripts/initializeEnvironment.js
```diff
@@ -26,7 +26,7 @@ const wdl = `task echo_to_file {
   }
 
   runtime {
-    docker: "alpine"
+    docker: "bashell/alpine-bash"
```
breilly2: I don't love the lack of transparency with this image. It's based on https://quay.io/repository/bashell/alpine which I can not find a Dockerfile for. I have to wonder why bashell is maintaining their own copy of an alpine base image.

How about photon? https://hub.docker.com/_/photon

Other than that, I don't see a great alpine alternative. As much as I don't want to have to maintain a separate artifact, I'm not fully confident in bashell.

integration-tests/scripts/initializeEnvironment.js
```diff
@@ -126,7 +126,7 @@ const setupMethod = async () => {
 
     const featuredMethodsUrl = `https://www.googleapis.com/storage/v1/b/firecloud-alerts-${env}/o/featured-methods.json`
 
-    const featuredMethods = await fetch(`${featuredMethodsUrl}?alt=media`, { headers: googleHeaders }).then(res => res.ok && res.json())
+    const featuredMethods = await fetch(`${featuredMethodsUrl}?alt=media`).then(res => res.ok && res.json())
```
breilly2, line 129: Did not test this script, but I did test accessing the featured methods JSON directly without auth which worked.

integration-tests/utils/integration-utils.js
```diff
@@ -25,8 +25,8 @@ const findIframe = async page => {
   return hasFrame() || await waitForFn({ fn: hasFrame })
 }
 
-const findInGrid = async (page, textContains) => {
-  return await page.waitForXPath(`//*[@role="grid"][contains(normalize-space(.),"${textContains}")]`)
+const findInGrid = (page, textContains) => {
```
breilly2, line 28: Interesting. I think I get that this doesn't need `async` because the function body doesn't `await`. However, having `async` is a good indicator that it returns a promise. How can we communicate that in the absence of `async`?

integration-tests/tests/run-workflow.integration-test.js
```diff
@@ -0,0 +1,56 @@
...
+  await click(page, clickable({ textContains: workflowName }))
+  await waitForNoSpinners(page)
+  await click(page, clickable({ textContains: 'Add to Workspace' }))
+  // note that this automatically brings in the highest numbered config, which isn't what happens when going through the method repo in FC
+
+  await waitForNoSpinners(page)
+  await click(page, clickable({ textContains: 'Select Data' }))
+  await click(page, input({ labelContains: 'Choose specific rows to process' }))
+  await click(page, `//*[@role="checkbox" and contains(@aria-label, "${testEntity.name}")]`)
+  await click(page, clickable({ textContains: 'OK' }))
```
petesantos: nit: given this is such a short string, would it be better to use text rather than contains?

integration-tests/utils/integration-utils.js
```diff
@@ -86,6 +86,10 @@ const svgText = ({ textContains }) => {
   return `//*[name()="text" and contains(normalize-space(.),"${textContains}")]`
 }
 
+const workspaceTab = tabName => {
```
breilly2: This is the first function in this file that is specific to a particular view. What would you think of moving this to a separate file to start a pattern of page-specific utilities so this file doesn't get out of control?

Alternatively, we could rename it to something more general since it really is just finding a named navigation item.

integration-tests/jest-puppeteer.config.js
```diff
@@ -15,6 +15,8 @@ module.exports = {
   launch: {
     devtools: process.env.DEVTOOLS === 'true',
     headless: process.env.HEADLESS !== 'false',
+    slowMo: process.env.SLOW === 'true' ? 250 : undefined,
```
petesantos, line 18: Nice option - thanks!

integration-tests/tests/run-workflow.integration-test.js
```diff
@@ -0,0 +1,56 @@
...
+  }, {
+    onFailedAttempt: async () => {
+      await page.reload()
+      await signIntoTerra(page)
+    }
+  })
+
+  await click(page, workspaceTab('data'))
+  await click(page, clickable({ textContains: 'test_entity' }))
+  await findElement(page, `//*[@role="grid"]//*[contains(.,"${testEntity.name}")]/following-sibling::*[contains(.,"result: ${testEntity.attributes.input}")]`)
```
petesantos: 2 questions:
1. This looks like the first instance of a findElement in a test that directly specifies an Xpath. Would it make sense to create a helper for this? 

2. Is this intended to find a second element on the same row, or anywhere in the grid?

breilly2 (commented):
Round 1. Basically the stuff outside of the test itself. Moving on to the test now...

integration-tests/tests/run-workflow.integration-test.js
zarsky-broad: Same row.

integration-tests/scripts/initializeEnvironment.js
zarsky-broad: Photon looks perfect

integration-tests/tests/run-workflow.integration-test.js
zarsky-broad: Also made a helper

integration-tests/utils/integration-utils.js
zarsky-broad: All of the verb things in this file are awaited, is that sufficient? See `findText` below.

integration-tests/tests/run-workflow.integration-test.js
petesantos: TOL: Thanks. While it isn't something we should address now - would this match any item in the table, even on a different row, given the way the markup is currently setup in the DOM?

I just wanted to put this out there as a thought as I am wondering if/how we should address this in the future.

petesantos (approved):
I tested by running the test to completion.
Overall looks good! I am good with this once Brian's points are resolved.
I added one TOL thought to the table sibling match

integration-tests/utils/integration-utils.js
breilly2: Deja vu; I've had this conversation before (with Brett, I think) and this was the conclusion then as well.

TOL: I'm still getting comfortable with not having typed function signatures. I think a little documentation could go a long way, especially in cases like this where we've decided on a convention that we're unlikely to change.

integration-tests/tests/run-workflow.integration-test.js
```diff
@@ -0,0 +1,56 @@
...
+
+const testEntity = { name: 'test_entity_1', entityType: 'test_entity', attributes: { input: 'foo' } }
+
+test('run workflow', withWorkspace(async ({ workspaceName }) => {
+  await page.goto(testUrl)
+  await signIntoTerra(page)
+
+  await page.evaluate((name, billingProject, testEntity) => {
+    return window.Ajax().Workspaces.workspace(billingProject, name).createEntity(testEntity)
+  }, workspaceName, billingProject, testEntity)
```
breilly2: Suggestion: move this to integration-helpers.js along with the other test data management functions.

integration-tests/tests/run-workflow.integration-test.js
```diff
@@ -0,0 +1,56 @@
...
+
+  await click(page, clickable({ textContains: 'View Workspaces' }))
+  await click(page, clickable({ textContains: workspaceName }))
+
+  await click(page, navChild('workflows'))
+  await waitForNoSpinners(page)
```
breilly2, line 19: Can we avoid having to `waitForNoSpinners()` or does `click` try to click "Find a Workflow" even though it's not clickable (because of the spinner overlay)?

TOL: If we do have to wait, I've discovered that I like it being inline with the test. I had a thought to suggest introducing page classes or something where we could encapsulate the code for navigating and waiting for the page to be ready. However, having it all here makes it very clear what the user experience is.

integration-tests/tests/run-workflow.integration-test.js
```diff
@@ -0,0 +1,56 @@
...
+  await click(page, clickable({ text: 'Select Data' }))
+  await click(page, input({ labelContains: 'Choose specific rows to process' }))
+  await click(page, `//*[@role="checkbox" and contains(@aria-label, "${testEntity.name}")]`)
+  await click(page, clickable({ text: 'OK' }))
+  await click(page, clickable({ text: 'Run analysis' }))
+
+  await Promise.all([
+    page.waitForNavigation(),
+    click(page, clickable({ text: 'Launch' }))
+  ])
```
breilly2, line 36: What is the intent here? My read is that we want to wait for the browser URL to change and for "Launch" to have been clicked, but we don't care (or even know?) what order these will happen in.

breilly2 (approved):
One suggestion and a couple of questions. Otherwise good-to-go. Nice work! 👍

integration-tests/tests/run-workflow.integration-test.js
zarsky-broad: We do indeed need to wait.

integration-tests/tests/run-workflow.integration-test.js
zarsky-broad: Yeah, this is [the recommended way to do it](https://pptr.dev/#?product=Puppeteer&version=v2.0.0&show=api-pagewaitfornavigationoptions).

PR #1947, opened 2019-11-20
Correct issue with data explorer open with [SATURN-1247]

petesantos:
Open with was not working in the app.
Corrected one of the calls in the data explorer chain, adding a guard.

This was an issue resulting from adding the eslint template rule

Tested by checking the flow by clicking on open with.
This fails without the fix, works with the fix.


breilly2 (approved):
👍 I tested after reproducing the bug in production. Looks good.

PR #1948, opened 2019-11-21
fix dockstore browse link

zarsky-broad:
Broken without this fix, working with it.


ehanna4 (approved):
Tested by observing the current behavior in dev where the link takes me to the dockstore page where no filters are applied and no workflows are showing in the list. Tried out this branch and the link now takes me to the dockstore page with a populated list of workflow wdls. (tested both instances of the DockstoreTile component) Much nicer user experience.

PR #1950, opened 2019-11-22
update select dropdown style [SATURN-1208]

zarsky-broad:
Worked with Joy; verified in UI.

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -3,7 +3,7 @@ import * as qs from 'qs'
 import { Fragment, useState } from 'react'
 import FocusLock from 'react-focus-lock'
 import { div, h, img, input, label, span } from 'react-hyperscript-helpers'
-import RSelect from 'react-select'
+import RSelect, { components as RSelectComponents } from 'react-select'
 import RAsyncCreatableSelect from 'react-select/async-creatable'
 import RSwitch from 'react-switch'
 import { centeredSpinner, icon } from 'src/components/icons'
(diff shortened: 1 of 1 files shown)


src/components/common.js
```diff
@@ -276,15 +276,26 @@ const commonSelectProps = {
     singleValue: base => ({ ...base, color: colors.dark() }),
     option: (base, { isSelected, isFocused, isDisabled }) => _.merge(base, {
       overflowWrap: 'break-word',
-      backgroundColor: isSelected ? colors.light(0.4) : isFocused ? colors.dark(0.15) : undefined,
-      color: isSelected ? colors.accent() : isDisabled ? undefined : colors.dark(),
+      fontWeight: isSelected ? 600 : undefined,
+      backgroundColor: isFocused ? colors.dark(0.15) : 'white',
```
ehanna4, line 280: nit: I haven't seen 'white' as a color used much in the app, should this be 'colors.light' instead?

petesantos: +1 The hard coded color may make it more difficult to change the theme in the future. Especially if adding a high contrast theme for a11y - the hardcoded color may make this more difficult

petesantos (approved):
Looks good - much more intuitive.
Brought up PR locally tested changes.

src/components/common.js
zarsky-broad: I see 66 uses of `white` vs just 27 uses of `light` in the code, I'm going to leave this as-is for now and we can consider a wider refactor later.

zarsky-broad: I see 66 uses of `white` vs just 27 uses of `light` in the code, I'm going to leave this as-is for now and we can consider a wider refactor later.

PR #1952, opened 2019-11-22
More test parallelization on Circle [SATURN-1254]

zarsky-broad:
This allows circle to split the integration tests up between different instances. With this config, we have 2 instances, each of which can run 2 tests in parallel. Circle decides how to split things based on historic data on test duration.

See [here](https://circleci.com/docs/2.0/parallelism-faster-jobs/#splitting-by-timing-data) and [here](https://dev.to/drazisil/test-splitting-with-jest-on-circleci-1i0c) to learn more.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,8 +162,8 @@ jobs:
             SCREENSHOT_DIR: "/tmp/failure-screenshots"
           command: |
             mkdir -p ${SCREENSHOT_DIR}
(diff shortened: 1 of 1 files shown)


petesantos (approved):
My only concern is that looking at the jobs on circle, it looks like these jobs took longer than the others. Though this may have been a fluke. I wouldn't mind trying it for a few days to see if it is faster or slower as we can always revert back.

zarsky-broad:
This is fairly consistently way slower and I can't figure out why, so I'm closing this for now.

PR #1953, opened 2019-11-25
Moved cancellation hook

petesantos:
The merge to dev failed due to an uncaught merge conflict with the periodic cookie setter change.
Merged the code and tested by throttling the network and checking cancellations in the network panel


petesantos:
Created new branch for the fix

PR #1955, opened 2019-11-25
Add rename workflow option to 'Find a Workflow' import modal [SATURN-1253]

ehanna4:
Add ability for users to rename a workflow before importing into a workspace from the 'Find a Workflow' modal accessible from the Workflows page view. (There is another place users can import workflows which will be addressed in another PR)

Tested by importing 10 different workflows into a workspace at least twice, once with default name and again with new name (including: switching order, using a name already in workspace, really long names, etc).

W̶h̶e̶n̶ ̶t̶e̶s̶t̶i̶n̶g̶ ̶p̶l̶e̶a̶s̶e̶ ̶n̶o̶t̶e̶ ̶t̶h̶a̶t̶ ̶t̶h̶e̶ ̶h̶a̶p̶l̶o̶t̶y̶p̶e̶c̶a̶l̶l̶e̶r̶-̶g̶v̶c̶f̶-̶g̶a̶t̶k̶4̶ ̶w̶o̶r̶k̶f̶l̶o̶w̶ ̶a̶l̶r̶e̶a̶d̶y̶ ̶h̶a̶d̶ ̶a̶ ̶s̶l̶i̶g̶h̶t̶l̶y̶ ̶d̶i̶f̶f̶e̶r̶e̶n̶t̶ ̶b̶e̶h̶a̶v̶i̶o̶r̶ ̶w̶h̶e̶n̶ ̶i̶m̶p̶o̶r̶t̶i̶n̶g̶ ̶(̶d̶e̶f̶a̶u̶l̶t̶ ̶i̶m̶p̶o̶r̶t̶ ̶n̶a̶m̶e̶ ̶g̶e̶t̶s̶ ̶'̶-̶c̶o̶n̶f̶i̶g̶u̶r̶e̶d̶'̶ ̶a̶d̶d̶e̶d̶ ̶t̶o̶ ̶e̶n̶d̶)̶.̶

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -906,13 +906,13 @@ const Methods = signal => ({
         return res.json()
       },
 
-      toWorkspace: async (workspace, config = {}) => {
+      toWorkspace: async (workspace, config = {}, workflowRename) => {
         const res = await fetchRawls(`workspaces/${workspace.namespace}/${workspace.name}/methodconfigs`,
           _.mergeAll([authOpts(), jsonBody(_.merge({
             methodRepoMethod: {
               methodUri: `agora://${namespace}/${name}/${snapshotId}`
             },
-            name,
+            name: workflowRename ? workflowRename : name,
             namespace,
             rootEntityType: '',
             prerequisites: {},
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -911,7 +911,7 @@ const Methods = signal => ({
             methodRepoMethod: {
               methodUri: `agora://${namespace}/${name}/${snapshotId}`
             },
-            name: workflowRename ? workflowRename : name,
+            name: workflowRename || name,
             namespace,
             rootEntityType: '',
             prerequisites: {},
--- a/src/pages/workspaces/workspace/Workflows.js
+++ b/src/pages/workspaces/workspace/Workflows.js
@@ -171,8 +171,10 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
       methods: undefined,
       selectedWorkflowDetails: undefined,
       exporting: undefined,
-      workflowRename: '',
-      isEditingName: false
+      workflowRename: undefined,
+      isEditingName: false,
+      config: undefined,
+      methodAjax: undefined
     }
   }
 
@@ -212,12 +214,12 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
       div({ style: { display: 'flex' } }, [
         div({ style: { flexGrow: 1, marginTop: '1rem' } }, [
           !isEditingName ?
-            div([span({ style: styles.sectionTitle }, ['Workflow Name: ']), `${workflowRename}`,
+            div([span({ style: styles.sectionTitle }, ['Workflow Name: ']), workflowRename ? `${workflowRename}` : '',
               h(Link, {
                 style: { marginRight: '0.5rem' },
                 onClick: () => this.setState({ isEditingName: !isEditingName }),
                 'aria-label': 'Edit workflow name'
-              }, [icon('edit', { style: { marginLeft: '0.5rem' }, size: 13 })])])
+              }, [workflowRename ? icon('edit', { style: { marginLeft: '0.5rem' }, size: 13 }) : undefined])])
             :
             h(Fragment, [
               h(IdContainer, [id => h(Fragment, [
@@ -293,10 +295,18 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
     const { ajax: { Methods } } = this.props
     const { namespace, name, snapshotId } = selectedWorkflow
 
-    this.setState({ selectedWorkflow, workflowRename: selectedWorkflow.name })
+    this.setState({ selectedWorkflow })
+
     try {
       const selectedWorkflowDetails = await Methods.method(namespace, name, snapshotId).get()
-      this.setState({ selectedWorkflowDetails })
+
+      const methodAjax = Ajax().Methods.method(selectedWorkflow.namespace, selectedWorkflow.name, selectedWorkflow.snapshotId)
+
+      const config = _.maxBy('snapshotId', await methodAjax.configs())
+
+      this.setState({ selectedWorkflowDetails, methodAjax })
+
+      config ? this.setState({ workflowRename: config.name, config }) : this.setState({ workflowRename: selectedWorkflow.name })
     } catch (error) {
       reportError('Error loading workflow', error)
       this.setState({ selectedWorkflow: undefined })
@@ -305,25 +315,20 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
 
   async exportMethod() {
     const { namespace, name } = this.props
-    const { selectedWorkflow, workflowRename } = this.state
+    const { selectedWorkflow, workflowRename, methodAjax, config } = this.state
 
     this.setState({ exporting: true })
 
     try {
-      const methodAjax = Ajax().Methods.method(selectedWorkflow.namespace, selectedWorkflow.name, selectedWorkflow.snapshotId)
+      const newConfig = (config && workflowRename !== selectedWorkflow.name) ?
+        _.merge({ name: workflowRename, payloadObject: { name: workflowRename } }, config) : undefined
 
-      const config = _.maxBy('snapshotId', await methodAjax.configs())
+      await methodAjax.toWorkspace({ namespace, name }, newConfig,
+        workflowRename !== selectedWorkflow.name ? workflowRename : undefined)
 
-      if (config && workflowRename !== selectedWorkflow.name) {
-        config.name = workflowRename
-        config.payloadObject.name = workflowRename
-      }
+      const { namespace: workflowNamespace } = newConfig || selectedWorkflow
 
-      await methodAjax.toWorkspace({ namespace, name }, config, workflowRename !== selectedWorkflow.name ? workflowRename : undefined)
-
-      const { namespace: workflowNamespace, name: workflowName } = config || selectedWorkflow
-
-      Nav.goToPath('workflow', { namespace, name, workflowNamespace, workflowName: config ? workflowName : workflowRename })
+      Nav.goToPath('workflow', { namespace, name, workflowNamespace, workflowName: workflowRename })
     } catch (error) {
       reportError('Error importing workflow', error)
       this.setState({ exporting: false })
--- a/src/pages/workspaces/workspace/Workflows.js
+++ b/src/pages/workspaces/workspace/Workflows.js
@@ -4,10 +4,10 @@ import { a, div, h, label, span } from 'react-hyperscript-helpers'
 import * as breadcrumbs from 'src/components/breadcrumbs'
 import { ViewToggleButtons, withViewToggle } from 'src/components/CardsListToggle'
 import {
-  ButtonOutline, ButtonPrimary, Clickable, IdContainer, Link, makeMenuIcon, MenuButton, methodLink, PageBox, Select, spinnerOverlay
+  ButtonOutline, ButtonPrimary, ButtonSecondary, Clickable, IdContainer, Link, makeMenuIcon, MenuButton, methodLink, PageBox, Select, spinnerOverlay
 } from 'src/components/common'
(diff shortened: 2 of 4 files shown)


src/pages/workspaces/workspace/Workflows.js
```diff
@@ -189,20 +198,68 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
 
     const { synopsis, managers, documentation } = selectedWorkflowDetails || {}
 
+    const errors = validate({ workflowRename }, {
+      workflowRename: {
```
ehanna4: Validation constraints matched to those of naming a workflow when duplicating or copying.

src/libs/ajax.js
```diff
@@ -906,13 +906,13 @@ const Methods = signal => ({
...
 
-      toWorkspace: async (workspace, config = {}) => {
+      toWorkspace: async (workspace, config = {}, workflowRename) => {
         const res = await fetchRawls(`workspaces/${workspace.namespace}/${workspace.name}/methodconfigs`,
           _.mergeAll([authOpts(), jsonBody(_.merge({
             methodRepoMethod: {
               methodUri: `agora://${namespace}/${name}/${snapshotId}`
             },
-            name,
+            name: workflowRename ? workflowRename : name,
```
breilly2: Nit: Can simplify to `workflowRename || name`

To other reviewers: What is the canonical way to express this?

src/pages/workspaces/workspace/Workflows.js
breilly2: It would be good to extract this validator.

breilly2:
TOL: Does edit mode need to be toggled? I could imagine having the text input always there, pre-populated with the workflow name. That would remove the complexity around toggling edit mode, including disabling the "Add to Workspace" button. Also, we could always pass in the entered name even if it happens to be the same as what the default would be.

Possible complexity: Where does the "-configured" part come from when importing "haplotypecaller-gvcf-gatk4"? I didn't see it in UI code so maybe it's happening in rawls?

zarsky-broad:
@breilly2 it's the name of the configuration, which is the name we actually use.

breilly2:
Can we read the config name in order to display and pre-populate the input with that name?

ehanna4:
@breilly2 good point, if the workflow is going to be named 'foo-configured' it would be much better for the user to see that at the import phase so that they aren't looking for a workflow just called 'foo' later. However if the '-configured' addition is shown at import time they would be able to delete that off since the name is now editable, is that an issue?

breilly2:
I don't think that would be an issue. We're letting them edit the name so they can already change it to whatever they want.

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -248,20 +308,21 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
...
-      const methodAjax = Ajax().Methods.method(selectedWorkflow.namespace, selectedWorkflow.name, selectedWorkflow.snapshotId)
-
-      const config = _.maxBy('snapshotId', await methodAjax.configs())
+      const newConfig = (config && workflowRename !== selectedWorkflow.name) ?
+        _.merge(config, { name: workflowRename, payloadObject: { name: workflowRename } }) : undefined
+      this.setState({ config: newConfig })
 
-      await methodAjax.toWorkspace({ namespace, name }, config)
+      await methodAjax.toWorkspace({ namespace, name }, newConfig,
+        workflowRename !== selectedWorkflow.name ? workflowRename : undefined)
```
panentheos, line 321: If we're modifying the `config` object already, then we shouldn't need to pass the new name in separately. I think we can remove the third argument to this.

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -248,20 +308,21 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
...
 
     this.setState({ exporting: true })
 
     try {
-      const methodAjax = Ajax().Methods.method(selectedWorkflow.namespace, selectedWorkflow.name, selectedWorkflow.snapshotId)
-
-      const config = _.maxBy('snapshotId', await methodAjax.configs())
+      const newConfig = (config && workflowRename !== selectedWorkflow.name) ?
+        _.merge(config, { name: workflowRename, payloadObject: { name: workflowRename } }) : undefined
+      this.setState({ config: newConfig })
```
panentheos, line 318: There's no need to set the `config` state here. In fact, in case the ajax call fails, this might lead to a slightly incorrect state, since it would incorrectly persist the edited name.

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -248,20 +308,21 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
...
+    const { selectedWorkflow, workflowRename, methodAjax, config } = this.state
 
     this.setState({ exporting: true })
 
     try {
-      const methodAjax = Ajax().Methods.method(selectedWorkflow.namespace, selectedWorkflow.name, selectedWorkflow.snapshotId)
-
-      const config = _.maxBy('snapshotId', await methodAjax.configs())
+      const newConfig = (config && workflowRename !== selectedWorkflow.name) ?
+        _.merge(config, { name: workflowRename, payloadObject: { name: workflowRename } }) : undefined
```
panentheos, line 317: This looks like it will result in `undefined` if the name is edited, which would mean other information from the config would be lost.

Also, since we're already maintaining `workflowRename` as a valid value throughout, we could probably remove the conditional logic here and just always write the new value (which might be a noop if it wasn't edited)

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -237,9 +289,17 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
...
+
     try {
       const selectedWorkflowDetails = await Methods.method(namespace, name, snapshotId).get()
-      this.setState({ selectedWorkflowDetails })
+
+      const methodAjax = Ajax().Methods.method(selectedWorkflow.namespace, selectedWorkflow.name, selectedWorkflow.snapshotId)
+
+      const config = _.maxBy('snapshotId', await methodAjax.configs())
+
+      this.setState({ selectedWorkflowDetails, methodAjax })
```
panentheos, line 300: `methodAjax` is a temporary object from the ajax code, and shouldn't be stored in state.

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -237,9 +289,17 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
     const { namespace, name, snapshotId } = selectedWorkflow
 
     this.setState({ selectedWorkflow })
+
     try {
       const selectedWorkflowDetails = await Methods.method(namespace, name, snapshotId).get()
-      this.setState({ selectedWorkflowDetails })
+
+      const methodAjax = Ajax().Methods.method(selectedWorkflow.namespace, selectedWorkflow.name, selectedWorkflow.snapshotId)
```
panentheos, line 296: nit: `selectedWorkflow` was already destructured above, so we can use the extracted variables directly here.

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -237,9 +289,17 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
...
+
+      const config = _.maxBy('snapshotId', await methodAjax.configs())
+
+      this.setState({ selectedWorkflowDetails, methodAjax })
+
+      config ? this.setState({ workflowRename: config.name, editedWorkflowName: config.name, config }) : this.setState({ workflowRename: selectedWorkflow.name, editedWorkflowName: selectedWorkflow.name })
```
panentheos, line 302: For clarity, we could combine all these `setState` calls into one, and parameterize the data being set instead.

src/pages/workspaces/workspace/Workflows.js
panentheos, line 296: This call looks like it could be made in parallel with the other one via `Promise.all`

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -189,20 +201,60 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
...
           div([synopsis || (selectedWorkflowDetails && 'None')]),
-          div({ style: { fontSize: 18, fontWeight: 600, margin: '1rem 0 0.5rem' } }, ['Method Owner']),
+          div({ style: styles.sectionTitle }, ['Method Owner']),
           div([_.join(',', managers)])
         ]),
         div({ style: { margin: '0 1rem', display: 'flex', flexDirection: 'column' } }, [
-          h(ButtonPrimary, { style: { marginBottom: '0.5rem' }, onClick: () => this.exportMethod() }, ['Add to Workspace']),
-          h(ButtonOutline, { onClick: () => this.setState({ selectedWorkflow: undefined, selectedWorkflowDetails: undefined }) }, ['Return to List'])
+          h(ButtonPrimary, {
+            disabled: !!errors || !!isEditingName,
```
panentheos, line 247: nit: Since `isEditingName` is already a boolean, there's no need to use `!!` on it.

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -189,20 +201,60 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
...
-        div({ style: { flexGrow: 1 } }, [
-          div({ style: { fontSize: 18, fontWeight: 600, margin: '1rem 0 0.5rem' } }, ['Synopsis']),
+        div({ style: { flexGrow: 1, marginTop: '1rem' } }, [
+          !isEditingName ?
+            div([span({ style: styles.sectionTitle }, ['Workflow Name: ']), workflowRename ? `${workflowRename}` : '',
+              h(Link, {
+                style: { marginRight: '0.5rem' },
+                onClick: () => this.setState({ isEditingName: !isEditingName }),
+                'aria-label': 'Edit workflow name'
+              }, [workflowRename ? icon('edit', { style: { marginLeft: '0.5rem' }, size: 13 }) : undefined])])
```
panentheos, line 215: nit: Rather than rendering the link with no contents, we could just make the whole link conditional via `&&`

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -189,20 +201,60 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
...
 
+    const errors = validateWorkflowName(editedWorkflowName, this.props.name)
+
     const renderDetails = () => [
       div({ style: { display: 'flex' } }, [
-        div({ style: { flexGrow: 1 } }, [
-          div({ style: { fontSize: 18, fontWeight: 600, margin: '1rem 0 0.5rem' } }, ['Synopsis']),
+        div({ style: { flexGrow: 1, marginTop: '1rem' } }, [
+          !isEditingName ?
+            div([span({ style: styles.sectionTitle }, ['Workflow Name: ']), workflowRename ? `${workflowRename}` : '',
```
panentheos, line 210: This starts on one line, and then jumps to a new line on edit. Could we move this below the label to keep everything consistent?

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -189,20 +201,60 @@ const FindWorkflowModal = ajaxCaller(class FindWorkflowModal extends Component {
...
+                onClick: () => this.setState({ isEditingName: !isEditingName }),
+                'aria-label': 'Edit workflow name'
+              }, [workflowRename ? icon('edit', { style: { marginLeft: '0.5rem' }, size: 13 }) : undefined])])
+            :
+            h(Fragment, [
+              h(IdContainer, [id => h(Fragment, [
+                h(FormLabel, { htmlFor: id, style: { margin: '0 0 0.25rem' } }, ['Workflow Name:']),
+                h(ValidatedInput, {
+                  error: Utils.summarizeErrors(errors),
+                  inputProps: {
```
panentheos, line 222: This should have `autoFocus: true` so the text box gets focus immediately.

ehanna4:
Got some very good feedback and want to spend appropriate effort fixing things and making the correct changes so closing this request for now. Will apply these feedbacks to a future pull request

PR #1957, opened 2019-11-25
Run integration tests nightly against alpha and staging [SATURN-1258]

zarsky-broad:


.circleci/config.yml
```diff
@@ -222,3 +265,13 @@ workflows:
...
+  nightly-integration-tests:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
+          filters:
+            branches:
+              only: dev
+    jobs:
+      - integration-tests-alpha
+      - integration-tests-staging
```
breilly2, line 277: I had a thought... what if we run this on every merge to dev? I believe the goal is to test the combination of bleeding-edge UI and stable server prior to deploying UI changes. Performing this check on every dev merge would cover us for on-demand UI releases as well as daily scheduled releases.

.circleci/config.yml
```diff
@@ -222,3 +265,13 @@ workflows:
...
               only: dev
+  nightly-integration-tests:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
+          filters:
+            branches:
+              only: dev
+    jobs:
+      - integration-tests-alpha
```
breilly2, line 276: TOL: A specific time is a good first pass. I think the alpha build is somewhat variable and sometimes run on-demand. I wonder if it would be possible to trigger CircleCI from successful completion of the alpha build in Jenkins... maybe something like Google Pub/Sub or Cloud Tasks.

zarsky-broad: All good thoughts on timing; for the moment these are based on Gary's requests when I talked to him yesterday.

ehanna4 (approved):
Looks good for first iteration. Can see passing tests here: https://circleci.com/workflow-run/f7723eca-a21f-4eb6-b8db-cb78e2ccd4ca. Expectation that will chat more later about more iterations as team sees fit.

PR #1958, opened 2019-11-27
Show the number of analyses in the launch modal [SATURN-1263]

zarsky-broad:
<!--
Hello, friend!
Thanks!
--!>
Verified functionality with all different launch types.

### DIFF
--- a/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
+++ b/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
@@ -39,25 +39,31 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
   }
 
   async componentDidMount() {
-    const { workspaceId: { namespace, name }, entitySelectionModel: { type, selectedEntities }, config: { rootEntityType }, ajax: { Workspaces } } = this.props
+    const { workspaceId: { namespace, name }, entitySelectionModel: { type, selectedEntities }, processSingle, config: { rootEntityType }, ajax: { Workspaces } } = this.props
 
-    const entities = await Promise.resolve(Utils.switchCase(type,
-      [EntitySelectionType.processAll, async () => _.map('name', await Workspaces.workspace(namespace, name).entitiesOfType(rootEntityType))],
-      [EntitySelectionType.chooseRows, () => _.keys(selectedEntities)],
-      [EntitySelectionType.processMergedSet, () => _.flow(
-        _.flatMap(`attributes.${rootEntityType}s.items`),
-        _.map('entityName')
-      )(selectedEntities)],
-      [Utils.DEFAULT, () => selectedEntities]
-    ))
+    if (!processSingle) {
+      const entities = await Promise.resolve(Utils.switchCase(type,
+        [EntitySelectionType.processAll, async () => _.map('name', await Workspaces.workspace(namespace, name).entitiesOfType(rootEntityType))],
+        [EntitySelectionType.chooseRows, () => _.keys(selectedEntities)],
+        [EntitySelectionType.processMergedSet, () => _.flow(
+          _.flatMap(`attributes.${rootEntityType}s.items`),
+          _.map('entityName')
+        )(selectedEntities)],
+        [Utils.DEFAULT, () => selectedEntities]
+      ))
 
-    this.setState({ entities })
+      this.setState({ entities })
+    }
   }
 
   render() {
     const { onDismiss, entitySelectionModel: { type }, processSingle } = this.props
     const { entities, launching, message, multiLaunchCompletions, launchError, multiLaunchErrors } = this.state
-    const entityCount = _.isArray(entities) ? _.uniq(entities).length : _.size(entities)
+    const entityCount = Utils.cond(
+      [processSingle, () => 1],
+      [_.isArray(entities), () => _.uniq(entities).length],
+      () => _.size(entities)
+    )
 
     return h(Modal, {
       title: !launching ? 'Confirm launch' : 'Launching Analysis',
@@ -78,7 +84,9 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
         'This will launch ', b([entityCount]), ` analys${entityCount === 1 ? 'is' : 'es'}`,
         type === EntitySelectionType.chooseSets && entityCount > 1 && ' simultaneously',
         '.',
-        type !== EntitySelectionType.chooseSets && entityCount !== entities.length && div({ style: { fontStyle: 'italic', marginTop: '0.5rem' } }, [
+        !processSingle && type !== EntitySelectionType.chooseSets && entityCount !== entities.length && div({
+          style: { fontStyle: 'italic', marginTop: '0.5rem' }
+        }, [
           '(Duplicate entities are only processed once.)'
         ])
       ]),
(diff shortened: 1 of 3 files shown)


ehanna4 (approved):
Looks good to me. Tested by launching an analysis of many different varieties and observed the different messages. Found a small thing to fix which was addressed with the last commit.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -37,10 +38,32 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
     }
   }
 
+  async componentDidMount() {
+    const { workspaceId: { namespace, name }, entitySelectionModel: { type, selectedEntities }, processSingle, config: { rootEntityType }, ajax: { Workspaces } } = this.props
+
+    if (!processSingle) {
+      const entities = await Promise.resolve(Utils.switchCase(type,
```
panentheos: It's ok to `await` on a non-promise, it just passes the value through. So this doesn't need to be wrapped in `Promise.resolve`

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -99,31 +128,23 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
...
-        this.launch(rootEntityType, _.head(entities))
+      if (entities.length === 1) {
+        this.launch(rootEntityType, entities[0])
       } else {
         this.createSetAndLaunch(entities)
       }
     } else if (type === EntitySelectionType.processMergedSet) {
-      if (_.size(selectedEntities) === 1) {
-        const { entityType, name } = _.values(selectedEntities)[0]
+      if (entities.length === 1) {
```
panentheos: This is looking at the entity count, rather than the set count. If you select a set of 1 and a set of 0, it should take the second branch, but it currently follows this one.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -99,31 +128,23 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
...
+        const { entityType, name } = entities[0]
```
panentheos: In this case the `entities` state is an array of strings, so these come back undefined.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -99,31 +128,23 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
...
-          _.map('entityName')
-        )(selectedEntities)
-
         this.createSetAndLaunch(entities)
       }
     } else if (type === EntitySelectionType.chooseSets) {
-      if (_.size(selectedEntities) === 1) {
-        this.launch(rootEntityType, _.values(selectedEntities)[0].name)
+      if (_.size(entities) === 1) {
+        this.launch(rootEntityType, _.values(entities)[0].name)
```
panentheos, line 138: TOL: It's confusing to me that `entities` has a different data type depending on what mode we're in.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -37,10 +38,32 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
...
+        [EntitySelectionType.processAll, async () => _.map('name', await Workspaces.workspace(namespace, name).entitiesOfType(rootEntityType))],
+        [EntitySelectionType.chooseRows, () => _.keys(selectedEntities)],
+        [EntitySelectionType.processMergedSet, () => _.flow(
+          _.flatMap(`attributes.${rootEntityType}s.items`),
+          _.map('entityName')
+        )(selectedEntities)],
+        [Utils.DEFAULT, () => selectedEntities]
+      ))
+
+      this.setState({ entities })
```
panentheos: TOL: We already have the `entityMetadata` loaded on the page, which includes counts. If we passed that in, we might be able to generate the info we need without any additional network calls, and leave the launch code paths as-is.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
zarsky-broad: I think we would want that case to take the first branch; this is running against the entities of the set, not the set itself, the alternative would be running against the members of a set with 1 member.

panentheos: We could, but it would be a change in behavior, and also doesn't match the message we give, e.g. 'Selected rows will have their membership combined...'

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -37,10 +38,32 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
...
+      const entities = Utils.switchCase(type,
+        [EntitySelectionType.chooseRows, () => _.keys(selectedEntities)],
+        [EntitySelectionType.processMergedSet, () => _.flow(
+          _.flatMap(`attributes.${rootEntityType}s.items`),
+          _.map('entityName')
+        )(selectedEntities)],
+        [Utils.DEFAULT, () => selectedEntities]
+      )
+
+      this.setState({ entities })
```
panentheos: Now that this calculation is synchronous, can we do it in `render` and eliminate this piece of state?

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
zarsky-broad: Hmm, that's true. Fair enough.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
zarsky-broad: Moved to a `this.` method, since it's also used in the launch method.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
zarsky-broad: This is true. I think it's a candidate for further simplification later.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -37,10 +38,31 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
...
   render() {
-    const { onDismiss, entitySelectionModel: { type, selectedEntities } } = this.props
+    const { onDismiss, entitySelectionModel: { type }, entityMetadata, config: { rootEntityType }, processSingle } = this.props
     const { launching, message, multiLaunchCompletions, launchError, multiLaunchErrors } = this.state
-    const entityCount = _.size(selectedEntities)
+    const entities = this.getEntities()
+    const entityCount = Utils.cond(
+      [processSingle, () => 1],
+      [type === EntitySelectionType.processAll, () => entityMetadata[rootEntityType].count],
+      [_.isArray(entities), () => _.uniq(entities).length],
```
breilly2, line 63: In which case is `entities` an array? I'm wondering if this is always doing the right thing. If the root entity type is a singular type and the selected entity is a set, then the effective number of entities (and therefore the number of workflows) will be the number of entities in the set.

zarsky-broad: Talked about this with faces last week. Gonna merge for now, this could use a refactor eventually.

PR #1959, opened 2019-12-02
Fix null bug in notebook terminal [SATURN-1215]

ehanna4:
Previously if you had a notebook runtime environment running and were in the terminal view then clicked the trashcan and deleted the runtime environment you would get an error "Unhandled Rejection (Type Error): Cannot read property 'clusterurl' of null. 

Changed the way the cluster object is de-structured in the terminalLauncher.js (found this guide helpful: https://medium.com/@crunchtech/object-destructuring-best-practice-in-javascript-9c8794699a0d). 

Also added a message status for a null cluster. "Deleting" and "Deleted" appear to be possible cluster statuses in Leo (so I did not remove the logic around these status messages in case they come through) however once a cluster is deleted on our end it seems to just be a null, so needed a case to show the correct message for that.

Testing:
Tested deleting a cluster that was in a running, stopping, and starting state while in the terminal view. Reproduced the error without these changes, then made these changes and did not see the error following the same steps.

### DIFF
--- a/src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
+++ b/src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
@@ -65,7 +65,7 @@ const TerminalLauncher = _.flow(
 
   async componentDidUpdate() {
     const { cluster } = this.props || {}
-    const { clusterUrl } = cluster || {}
+    const { clusterUrl } = cluster
     const { url } = this.state
 
     if (clusterUrl && !url) {
--- a/src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
+++ b/src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
@@ -64,7 +64,8 @@ const TerminalLauncher = _.flow(
   }
 
   async componentDidUpdate() {
-    const { cluster: { clusterUrl } = {} } = this.props
+    const { cluster } = this.props || {}
+    const { clusterUrl } = cluster || {}
     const { url } = this.state
 
     if (clusterUrl && !url) {
@@ -99,6 +100,8 @@ const TerminalLauncher = _.flow(
         },
         title: 'Interactive terminal iframe'
       })
+    } else if (clusterStatus === null) {
+      return div({ style: { padding: '2rem' } }, ['No existing notebook runtime environment, you can create a new one to edit your notebook or use the terminal.'])
     } else {
       return div({ style: { padding: '2rem' } }, [
         'Creating Stopping Starting Updating Deleting'.includes(clusterStatus) &&


src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
```diff
@@ -64,7 +64,8 @@ const TerminalLauncher = _.flow(
   }
 
   async componentDidUpdate() {
-    const { cluster: { clusterUrl } = {} } = this.props
+    const { cluster } = this.props || {}
+    const { clusterUrl } = cluster || {}
```
breilly2: `clusterUrl` is expected to be a string, but if `cluster` is falsy then it'll be an empty object. The way it was before, if `cluster` was falsy, `clusterUrl` would be `undefined`. This changes the truthiness relationship between `cluster` and `clusterUrl`. Is that intended?

ehanna4: Thanks for pointing that out, I misinterpreted how the de-structure worked in the original version. Only cluster should be set to `{}` if it doesn't exist, I will remove the `|| {}` from line 68.

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
```diff
@@ -64,7 +64,8 @@ const TerminalLauncher = _.flow(
   }
 
   async componentDidUpdate() {
-    const { cluster: { clusterUrl } = {} } = this.props
+    const { cluster } = this.props || {}
+    const { clusterUrl } = cluster
```
panentheos, line 68: As of the latest change, this still exhibits the bug. I think the previous comment was based on a misreading of the code; providing a fallback value on the right would make sure the destructring always had an object to unpack, and `clusterUrl` would be undefined in the null case, which is good.

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
```diff
@@ -64,7 +64,8 @@ const TerminalLauncher = _.flow(
   }
 
   async componentDidUpdate() {
-    const { cluster: { clusterUrl } = {} } = this.props
+    const { cluster } = this.props || {}
```
panentheos, line 67: `this.props` is always an object, so there's no need to provide a fallback value here.

src/pages/workspaces/workspace/notebooks/TerminalLauncher.js
breilly2: Good catch. This wasn't the change I was thinking of, but I also didn't recognize that it would reintroduce the problem.

Emily and I chatted in person yesterday. My worry is that:
```
const { cluster } = this.props || {}
const { clusterUrl } = cluster || {}
```
doesn't express that we expect that `null` is a valid value for `cluster`... especially problematic because this code is distant from the origin of the `null` (`wrapWorkspace`). In other words, I might incorrectly simplify these two lines, effectively reverting the change. I would almost certainly miss that nuance if I was reading it in someone else's PR.

My suggestion in person (repeated here for broader communication) was to be explicit about checking for `null`.

Side note: This pattern appears in two places in this file... both here and when checking cluster status below. I thought, maybe we could have a utility function for this? Maybe lodash already has something useful? Answer: We could do something like:
```
const clusterUrl = _.get("cluster.clusterUrl", this.props)
```
but that also hides the fact that `null` is valid.

@panentheos, do you have any other thoughts/suggestions?

panentheos: In this case, I think `cluster || {}` communicates fairly well that `cluster` may be falsy, and so that should be accounted for in any refactors. This code path doesn't actually care about the `null` case, so I would hesitate to introduce any specific checks for that. Maybe a comment, if we think that would be valuable.

PR #1960, opened 2019-12-03
15 minute timeout [SATURN-1065]

petesantos:
Adds 15 minute idle timeout to Terra. This will not be fully enabled until an email domain has been provided. This code will sync timeout and expiration modals across tabs. The timeout will continue working after the window is closed.

There is a storybook story associated with these modals. The storybook is primarily designed to test the UX of the timeouts

Tested by manually changing the default timeouts and email domain on the modals trying different timeout scenarios including multiple tabs, closing the window and waiting for the timeout.

Also tested using storybook. There are knobs added to allow testing of different timeout and email domain scenarios.


src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,93 @@
...
+    _.forEach(event => document.addEventListener(event, updateLastActive), targetEvents)
+
+    return () => {
+      _.forEach(event => document.removeEventListener(event, updateLastActive), targetEvents)
+    }
+  })
+
+  return Utils.cond([
+    expired || logoutRequested, () => {
+      removeLocalPref('terra-timeout')
```
petesantos: Push this side effect to an effect

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,93 @@
...
+      return iframe({ style: { display: 'none' }, src: 'https://www.google.com/accounts/Logout' })
+    }
+  ], [
+    timedOut, () => {
+      setExpired(true)
```
petesantos: Move to an effect

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,93 @@
...
+
+
+const displayRemainingTime = remainingSeconds => {
+  return _.join(':', [
+    `${Math.floor(remainingSeconds / 60).toString().padStart(2, '0')}`,
+    `${Math.ceil(remainingSeconds % 60).toString().padStart(2, '0')}`
+  ])
+}
+
+const IdleTimeoutModal = ({ timeout = 15 * 1000 * 60, countdownStart = 2 * 1000 * 60, emailDomain }) => {
```
petesantos: nit: See if there is a minutes to ms function

src/libs/utils.js
```diff
@@ -406,6 +407,51 @@ export const useCancellation = () => {
...
+  return [currentTime, delay => { delayRef.current = delay }]
+}
+
+/*
+ * Syncs state with a local storage key and will cause a render in components in
+ * other tabs that are using the same storage key with this hook
+ */
+export const useLocalStorageState = (keyToSync, initialValue) => {
+  const [value, setValue] = useState(() => {
+    initialValue !== undefined && setLocalPref(keyToSync, initialValue)
```
petesantos: local prefs can get timed out - consider not using dynamic storage to ensure this doesn't get removed

src/libs/utils.js
```diff
@@ -406,6 +407,51 @@ export const useCancellation = () => {
   return controller.current.signal
 }
 
+
+// Will cause a render to occur at each delay interval. The delay can be changed in the render.
+export const usePolling = (initialDelay = 250) => {
```
petesantos: nit: rename to useCurrentTime

src/libs/utils.js
```diff
@@ -406,6 +407,51 @@ export const useCancellation = () => {
...
+  })
+
+  return [currentTime, delay => { delayRef.current = delay }]
+}
+
+/*
+ * Syncs state with a local storage key and will cause a render in components in
+ * other tabs that are using the same storage key with this hook
+ */
+export const useLocalStorageState = (keyToSync, initialValue) => {
```
petesantos: look at syncAtomToSessionStorage to see if we can use that logic to sync local storage and add cross tab communication

src/libs/utils.js
petesantos: I added a function for this in utils. If you want me to remove it, let me know.

src/libs/utils.js
```diff
@@ -368,6 +368,8 @@ export const mergeQueryParams = (params, urlString) => {
   return url.href
 }
 
+export const hhmmssToMs = ({ hh = 0, mm = 0, ss = 0 }) => ((hh * 60 * 60) + (mm * 60) + ss) * 1000
```
breilly2: Nit suggestion: `durationToMillis = ({ hours = 0, minutes = 0, seconds = 0 })`

src/_stories/IdleTimeout.stories.js
```diff
@@ -0,0 +1,98 @@
...
+      logout && authTest.set({ user: { id: `${userEmail}-123` }, profile: {} })
+    })
+
+    observer.observe(document.body, { attributes: false, childList: true, subtree: true })
+    return () => { observer.disconnect() }
+  })
+
+  const agreement = h(Fragment, [
+    span({ style: { width: 300 } }, [
+      'Running this story will log you out of all of your google accounts in this browser.',
```
breilly2, line 57: Is this true? I didn't see myself being signed out. Maybe this message is out of date.

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+import { div, h, iframe } from 'react-hyperscript-helpers'
+import ButtonBar from 'src/components/ButtonBar'
+import Modal from 'src/components/Modal'
+import { getUser } from 'src/libs/auth'
+import colors from 'src/libs/colors'
+import { authStore, lastActiveTimeStore } from 'src/libs/state'
+import * as Utils from 'src/libs/utils'
+
+
+const displayRemainingTime = remainingSeconds => {
```
breilly2: I tested this in Storybook with a `timeout` of 125 seconds and `countdownStart` of 123 seconds. When it should say "2:00" it instead says "1:60".

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+}
+
+const setLastActive = lastActive => lastActiveTimeStore.update(existing => ({ ...existing, [getUser().id]: lastActive }))
+
+const IdleTimeoutModal = ({
+  timeout = Utils.hhmmssToMs({ mm: 15 }),
+  countdownStart = Utils.hhmmssToMs({ mm: 3 }), emailDomain = ''
+}) => {
+  const { isSignedIn, profile: { email }, user: { id } } = Utils.useStore(authStore)
+  const isClinicalDomain = email && emailDomain ? email.includes(`@${emailDomain}`) : false
```
breilly2: `endswith` instead of `includes`. For example, we wouldn't want to apply this to @blah.com if they were signed in to @blah.co.

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+  const activeTimeStore = Utils.useStore(lastActiveTimeStore)
```
breilly2: Is it possible to destructure `activeTimeStore[id]`? That's the only piece of data used from the object.

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+
+  const currentTime = Date.now()
+  const lastActive = activeTimeStore[id] === 'expired' ? null : activeTimeStore[id]
+  const lastActiveTime = lastActive ? parseInt(lastActive, 10) : currentTime
+  const timedOut = currentTime > lastActiveTime + timeout
```
breilly2: TOL: I think the only output of this is `timedOut`. Also, if I understand this logic correctly, I think `lastActive` can be a number (session is active) or `null` (session is expired) or `undefined` (not signed in?). Maybe extracting a function would make this more readable.

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+
+  useEffect(() => { timedOut && !isSignedIn && setLastActive('expired') }, [isSignedIn, timedOut])
```
breilly2: TOL: This might be a little too much responsibility for something that is named `IdleTimeoutModal`. What I finally understood about this line is that this is what's actually doing the session expiration. Maybe extract this into a separate component? (This idea is partially inspired by watching Isaac make use of `ClusterStatusMonitor` yesterday.)

breilly2 (commented):
Comments so far. Still need to review `InactivityTimer` and the Storybook code.

src/_stories/IdleTimeout.stories.js
petesantos: I was on the fence about this story. 
You _may_ get logged out. This story checks for the existence of the logout iframe and mocks the authstore to log the user out which will immediately unmount the iframe. Depending on how long this takes this may log you out.

In the real world case the iframe will not unmount until the user has actually been signed out as it uses the authstore state to determine if the user has signed out.

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+  ])
+}
+
+const setLastActive = lastActive => lastActiveTimeStore.update(existing => ({ ...existing, [getUser().id]: lastActive }))
+
+const IdleTimeoutModal = ({
+  timeout = Utils.hhmmssToMs({ mm: 15 }),
+  countdownStart = Utils.hhmmssToMs({ mm: 3 }), emailDomain = ''
+}) => {
+  const { isSignedIn, profile: { email }, user: { id } } = Utils.useStore(authStore)
```
panentheos: We should probably be reading `email` from `user` instead of `profile`, since `user` sticks around after logout, while `profile` doesn't. It might not matter at the moment, but it could be a source of future confusion or bugs.

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+  const showCountdown = currentTime > timeoutTime - countdownStart
+  const countdown = Math.max(0, timeoutTime - currentTime)
+
+  setDelay(showCountdown ? 1000 : Math.max(250, countdown - countdownStart))
+
+  Utils.useOnMount(() => {
+    const targetEvents = ['click', 'keydown']
+    const updateLastActive = () => setLastActive(Date.now().toString())
+
+    if (!activeTimeStore[id] || activeTimeStore[id] === 'expired') {
```
breilly2: Would it ever be wrong to update the last active time when mounting `InactivityTimer`?

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+      'You can extend your session to continue working',
+      h(ButtonBar, {
+        style: { marginTop: '1rem', display: 'flex', alignItem: 'baseline', justifyContent: 'flex-end' },
+        okText: 'Extend Session',
+        cancelText: 'Log Out',
+        onCancel: () => setLogoutRequested(true)
+      })
+    ])
+  ],
+  false)
```
breilly2: Not needed?

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+    null)
+}
+
+const InactivityTimer = ({ id, timeout, countdownStart }) => {
+  const activeTimeStore = Utils.useStore(lastActiveTimeStore)
+  const [logoutRequested, setLogoutRequested] = useState()
+  const [currentTime, setDelay] = Utils.useCurrentTime()
+
+  const lastActive = activeTimeStore[id] === 'expired' ? Date.now() : activeTimeStore[id]
+  const lastActiveTime = lastActive ? parseInt(lastActive, 10) : Date.now()
```
breilly2: Should `Date.now()` be `currentTime`?

src/components/IdleTimeoutModal.js
petesantos: Changed to endsWith. On this I am not too concerned about handling the domain at the moment as this information will most likely be coming in as part of the user's profile:
e.g.
`clinicalUser:  true`

I also added a comment reflecting that this is placeholder code for testing

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+  const timeoutTime = lastActiveTime + timeout
+  const timedOut = currentTime > timeoutTime
```
breilly2: These are ever so slightly different than the calculations in `IdleTimeoutModal`. Is that intentional/necessary? This stuff is always a pain to think through. It would be great to have a function that you pass a few facts and get back a set of derived properties.

breilly2 (commented):
A few more comments. Still haven't looked at the storybook code but I did use it to test and liked what I saw!

I tested the timeout in Terra locally by modifying the `IdleTimeoutModal` properties. I used three incognito windows: two in Terra and one on google.com. The timeout notification was synchronized across both Terra windows. After being automatically signed out, I refreshed google.com and was no longer signed in there either. 👍

src/components/IdleTimeoutModal.js
```diff
@@ -0,0 +1,105 @@
...
+
+
+const displayRemainingTime = remainingSeconds => {
+  return _.join(':', [
+    `${Math.floor(remainingSeconds / 60).toString().padStart(2, '0')}`,
+    `${Math.ceil(remainingSeconds % 60).toString().padStart(2, '0')}`
+  ])
+}
+
+const setLastActive = lastActive => lastActiveTimeStore.update(existing => ({ ...existing, [getUser().id]: lastActive }))
```
panentheos: nit: this could use `_.set`
```suggestion
const setLastActive = lastActive => lastActiveTimeStore.update(_.set(getUser().id, lastActive))
```

src/components/IdleTimeoutModal.js
panentheos: nit: Since this is computing a boolean, it could just use `&&` instead of a ternary.

panentheos (commented):
See a few small comments (to add to Brian's). The overall functionality looks solid, so after a bit of tidying I think this will be good to go.

src/components/IdleTimeoutModal.js
panentheos: `cond` returns undefined otherwise, which isn't valid to render

src/components/IdleTimeoutModal.js
petesantos: If there is already a last active time set we would not want to override that. If the user were to close the window and bring the application up again (say 30 minutes later) and we reset the timer we would lose the information of the last active time and not be able to log them out.

petesantos: Yes, if the user closed the window before expiring and reopened it 20 minutes later resetting the timer on mount would effectively bypass the timeout.

src/pages/Main.js
```diff
@@ -1,5 +1,6 @@
 import 'src/libs/routes'
 
+import IdleStatusMonitor from 'components/IdleStatusMonitor'
```
breilly2: Needs to be
```
import IdleStatusMonitor from 'src/components/IdleStatusMonitor'
```
(and also reordered)

src/CONTRIBUTING.md
```diff
@@ -0,0 +1,139 @@
+## Table of Contents
```
breilly2: Did you mean to include `CONTRIBUTING.md` in this PR?

breilly2 (approved):
:+1: Looks good! I tested some pieces in Storybook and some in the app. I had a bunch of questions about the state flow that I discussed with Pete in person. I have high confidence that, other than some unavoidable corner cases, this works the way we want.

src/CONTRIBUTING.md
petesantos: No - i must have accidentally added it when switching between branches - thanks!

PR #1961, opened 2019-12-04
Fixes before registration test

zarsky-broad:
Standardizes on `google-auth-library` for tests, handles some errors in registration flow. (For the appcues request, Rex was returning a 503 since the user wasn't set up in Sam yet; the alternative would be to make Rex requests use `fetch` instead of `fetchOk`)

### DIFF
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -159,7 +159,7 @@ authStore.subscribe(withErrorReporting('Error checking TOS', async (state, oldSt
  * doing this.
  */
 authStore.subscribe(async (state, oldState) => {
-  if (!oldState.isSignedIn && state.isSignedIn) {
+  if (!oldState.registrationStatus && state.registrationStatus) {
     window.newrelic.setCustomAttribute('userGoogleId', state.user.id)
     window.Appcues && window.Appcues.identify(state.user.id, {
       dateJoined: state.registrationStatus === 'registered' ? await getDateJoined() : Date.now()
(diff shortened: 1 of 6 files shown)


src/pages/Register.js
```diff
@@ -44,7 +44,6 @@ export default class Register extends Component {
       await refreshTerraProfile()
     } catch (error) {
       reportError('Error registering', error)
-    } finally {
```
panentheos, line 47: What's the intent of this change? It looks like this will leave a spinner active if the call fails.

zarsky-broad: The intent is not to try to dismiss the spinner after the component unmounts.

src/libs/auth.js
```diff
@@ -162,7 +162,7 @@ authStore.subscribe(async (state, oldState) => {
   if (!oldState.isSignedIn && state.isSignedIn) {
     window.newrelic.setCustomAttribute('userGoogleId', state.user.id)
     window.Appcues && window.Appcues.identify(state.user.id, {
-      dateJoined: await getDateJoined()
+      dateJoined: state.registrationStatus === 'registered' ? await getDateJoined() : Date.now()
```
ehanna4, line 165: TOL: Would it make more sense to omit sending appcues any data for a test user at all? It feels risky to me to be adding a logic specifically for a test user in the main code as it could have unintended consequences/make it less clear what should be happening. (not sure if that's feasible either but just wanted to ask) Or could we more specifically check if it's a test user before changing what is sent? Essentially just checking if the user is registered seems potentially risky/loose?

panentheos, line 165: As written, this will always use `Date.now()` because the registration check doesn't happen until after the sign-in transition. We might want to just delay this action until we see the registration transition.

panentheos (approved):
Looks good, but see comment.

src/libs/auth.js
zarsky-broad: This logic isn't specifically for test users, it's for any unregistered user.

zarsky-broad: Done.

PR #1962, opened 2019-12-04
User registration test [SATURN-1081]

zarsky-broad:
<!--
Hello, friend!
Thanks!
--!>

The non-test parts of this have been rolled into #1961 so they can be merged separately.


integration-tests/utils/integration-helpers.js
```diff
@@ -59,7 +60,30 @@ const createEntityInWorkspace = (page, billingProject, workspaceName, testEntity
...
 }
 
+const makeUser = async () => {
+  const { email } = await fetchLyle('create')
+  const { accessToken: token } = await fetchLyle('token', email)
+
+  return { email, token }
+}
+
+const withUser = test => async () => {
```
petesantos, line 70: Should the withUser wrapper return the test untouched, but decorate it with the additional functionality we are looking for - signing in? Right now wrapping a test in withUser executes the test, not allowing this wrapper to be composable with other wrappers we may write in the future. 

Also, this could potentially be factored out into two different wrappers: withErrorScreenshot and withUserSignin

If all of our tests require a sign in using Lyle, the sign in aspect may be better off directly implemented in the signIntoTerra function. If there are tests that do not require a sign with Lyle / multiple sign in paths keeping it as a wrapper makes sense.

petesantos (commented):
nit: It looks like some of these changes are already in the pre-cujx pr

zarsky-broad:
@petesantos indeed, as per the description that is intentional so they can be merged separately.

integration-tests/utils/integration-helpers.js
zarsky-broad: As is, these can be flowed together if they're made to pass their params through, but that hasn't been necessary as yet, so I haven't added it. The purpose of this wrapper isn't to sign a user in, just to create a temp user and pass through their email and token, so they can be used to sign in at the next level.

Indeed, when we need a temp user and a workspace, we'll probably break out screenshot from the other wrappers. I haven't done it here because that feels like premature optimization to me.

integration-tests/utils/integration-config.js
```diff
@@ -1,15 +1,19 @@
...
-  TERRA_TOKEN: bearerToken,
   BILLING_PROJECT: billingProject = 'general-dev-billing-account',
-  WORKFLOW_NAME: workflowName = 'echo_to_file',
-  SCREENSHOT_DIR: screenshotDir
+  LYLE_SA_KEY: lyleToken,
+  LYLE_URL: lyleUrl = 'https://terra-lyle.appspot.com',
+  SCREENSHOT_DIR: screenshotDir,
+  TERRA_TOKEN: bearerToken,
+  TEST_URL: testUrl = 'http://localhost:3000',
+  WORKFLOW_NAME: workflowName = 'echo_to_file'
```
breilly2: 👍 for alphabetization.

integration-tests/utils/integration-helpers.js
```diff
@@ -59,7 +60,30 @@ const createEntityInWorkspace = (page, billingProject, workspaceName, testEntity
...
+  } finally {
+    await fetchLyle('delete', email)
+  }
+}
+
 module.exports = {
   withWorkspace,
-  createEntityInWorkspace
+  createEntityInWorkspace,
+  withUser
```
breilly2, line 88: Nit: Maybe alphabetize these while you're at it?

integration-tests/utils/integration-helpers.js
breilly2: I'm going to agree on both points. Yes, we can refactor after this PR. Yes, it probably is a good idea to refactor. The piece that I see duplicated here is handling error screenshots.

petesantos: When I originally reviewed this I missed that a function **is** being returned - my apologies. The param flow could be modified later - I am good with that.

The only thing I still want to push a bit on is the screenshot handling. Two things I was thinking about when looking at this some more (consider this TOL):

1. With the screenshot being directly added to this we are using the email as part of the file name for the failure, coupling this with the makeUser code. If this were a separate "with" handler it would better separate the concerns and allow us to add a filename to the error similar to how we add a title to our other error reporting.

2. withWorkspace is also using screenshot handling. Would it make sense at this point to extract the error handling/screenshot as it seems to be generally useful?

breilly2 (approved):
👍 I ran this test (once Isaac showed me how to retrieve the Lyle key) and it worked.

Nit: Would be good to include instructions for running in the readme. Here's the command I used:
```
LYLE_SA_KEY=$(vault read --format=json secret/dsde/terra/envs/common/lyle-user-service-account-key |jq .data) npm run test -- register-user
```

PR #1963, opened 2019-12-05
Add polling effect hook [SATURN-1277]

petesantos:
Added a usePollingEffect hook. After noticing a common pattern in the code to setup polling on mount to run an async (usually AJAX) operation I wanted to take a stab at factoring this out into a hook. If this doesn't seem right we don't have to do this.

Tested by running the app and going to each call site and checking the behavior. In some cases I needed to temporarily modify some other code to mock out the intended effect/behavior.

Also added a story that demonstrates the effect is triggered, but a render cycle is not as a result of using this hook (a render could be triggered if state is set within the passed in function).

### DIFF
--- a/src/components/ServiceAlerts.js
+++ b/src/components/ServiceAlerts.js
@@ -4,6 +4,7 @@ import { div, h } from 'react-hyperscript-helpers'
 import { Link } from 'src/components/common'
 import { clearNotification, notify } from 'src/components/Notifications'
 import { Ajax } from 'src/libs/ajax'
+import { withErrorIgnoring } from 'src/libs/error'
 import * as Utils from 'src/libs/utils'
 
 
@@ -11,20 +12,9 @@ export const ServiceAlerts = () => {
   const [alerts, setAlerts] = useState([])
   const prevAlerts = Utils.usePrevious(alerts)
 
-  Utils.useOnMount(() => {
-    const checkAlerts = async () => {
-      while (true) {
-        try {
-          setAlerts(_.uniqWith(_.isEqual, await Ajax().Buckets.getServiceAlerts()))
-        } catch {
-          // swallowing error, yum!
-        }
-        await Utils.delay(60000)
-      }
-    }
-
-    checkAlerts()
-  })
+  Utils.usePollingEffect(withErrorIgnoring(
+    async () => setAlerts(_.uniqWith(_.isEqual, await Ajax().Buckets.getServiceAlerts()))
+  ), 60000)
 
   useEffect(() => {
     if (prevAlerts) {
--- a/src/components/ServiceAlerts.js
+++ b/src/components/ServiceAlerts.js
@@ -14,7 +14,7 @@ export const ServiceAlerts = () => {
 
   Utils.usePollingEffect(withErrorIgnoring(
     async () => setAlerts(_.uniqWith(_.isEqual, await Ajax().Buckets.getServiceAlerts()))
-  ), 60000)
+  ), { ms: 60000, leading: true })
 
   useEffect(() => {
     if (prevAlerts) {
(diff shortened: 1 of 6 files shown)


src/components/ServiceAlerts.js
```diff
@@ -4,27 +4,17 @@ import { div, h } from 'react-hyperscript-helpers'
...
-        }
-        await Utils.delay(60000)
-      }
-    }
-
-    checkAlerts()
-  })
+  Utils.usePollingEffect(withErrorIgnoring(
+    async () => setAlerts(_.uniqWith(_.isEqual, await Ajax().Buckets.getServiceAlerts()))
+  ), 60000)
```
panentheos: In some cases, like this one, we want to make the first call immediately, before waiting. We could make that a configurable option, e.g. `initial: true`

src/libs/utils.js
```diff
@@ -406,6 +406,20 @@ export const useCancellation = () => {
...
 }
 
+export const usePollingEffect = (effectFn, ms = 250) => {
+  const signal = useCancellation()
+
+  useOnMount(() => {
+    const poll = async () => {
+      while (!signal.aborted) {
+        await delay(ms)
+        !signal.aborted && effectFn()
```
panentheos: We should `await effectFn()` in case it takes some time, to guard against pile-up.

src/libs/utils.js
```diff
@@ -406,6 +406,20 @@ export const useCancellation = () => {
   return controller.current.signal
 }
 
+export const usePollingEffect = (effectFn, ms = 250) => {
```
panentheos: I don't think we can confidently specify a reasonable default delay, so we should probably make this non-optional. Also, if we're adding more options like `initial`, we should consider changing the second argument to an `options` object, rather than adding positional args.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -511,9 +494,11 @@ const NotebookEditorFrame = ({ mode, notebookName, workspace: { workspace: { nam
     setUpNotebook()
   })
 
+  const signal = Utils.useCancellation()
+  Utils.usePollingEffect(withErrorIgnoring(() => Ajax(signal).Jupyter.notebooks(namespace, clusterName).setCookie()), 15 * 60 * 1000)
+
   return h(Fragment, [
     notebookSetupComplete && h(Fragment, [
-      h(PeriodicCookieSetter, { namespace, clusterName }),
```
panentheos: This is relying on conditional rendering via `notebookSetupComplete` to execute at the right time, so we should keep it as its own component (but it can still use the new hook)

panentheos (commented):
This is a nice abstraction! See comments.

src/components/ServiceAlerts.js
petesantos: I went with "leading" as the option name to align with similar options in other time related functions (e.g. throttle in lodash). If initial seems clearer I will change it to that.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -456,18 +456,9 @@ const JupyterFrameManager = ({ onClose, frameRef }) => {
...
-    }
-  }
-
-  Utils.useOnMount(() => {
-    periodicallySetCookie()
-  })
-
+  Utils.usePollingEffect(
+    withErrorIgnoring(() => Ajax(signal).Jupyter.notebooks(namespace, clusterName).setCookie()),
+    { ms: 15 * 60 * 1000 })
```
panentheos, line 461: Note: This changes the previous behavior of making a call on the leading edge, but since there is a separate inline `setCookie` call in the setup code, the initial call was extraneous.

panentheos (approved):
Looks good, tested by examining the network tab.

PR #1964, opened 2019-12-06
Enable Important Prefix [SATURN-1280]

petesantos:
Allow the addition of "important" items in our dynamic local storage mechanism. Important items will not be deleted when running out of space.

This change will be utilized by upcoming 15 minute signout changes.

I tested the existing behavior by manipulating the table columns with the important flag turned on and off.


petesantos:
This will be handled in the other changes with the atom/storage syncing

PR #1965, opened 2019-12-06
Widen workflow ID column on Job History screen [no JIRA]

aednichols:
I was working with a user who needed to copy the workflow IDs for 30 workflows.

Because the workflow ID is truncated (see below) the best way she could find to accomplish this task was to open 30 browser tabs and copy the ID from the address bar.

(A suggested workaround of right clicking the workflow ID and selecting "copy" did not work; she only had the "copy link address" option.)

Before:

Screen Shot 2019-12-06 at 2 47 02 PM (1)

After:

Screen Shot 2019-12-06 at 3 53 46 PM

---

Testing done: ran my branch locally to take the "after" screenshot.

### DIFF
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -249,7 +249,7 @@ const SubmissionDetails = _.flow(
                 return h(TooltipCell, [h(Fragment, _.map(div, filteredWorkflows[rowIndex].messages))])
               }
             }, {
-              size: { basis: 150, grow: 0 },
+              size: { basis: 375, grow: 0 },
               headerRenderer: () => h(Sortable, { sort, field: 'workflowId', onSort: setSort }, ['Workflow ID']),
               cellRenderer: ({ rowIndex }) => {
                 const { workflowId, inputResolutions: [{ inputName } = {}] } = filteredWorkflows[rowIndex]


ehanna4:
Hi, is it possible for a submission id to be even longer than the ones this user reported? - if that occurs would this fix still work? A potential alternative/addition to the fix would be to enable the workaround functionality suggested (right clicking would have the copy option which would copy the submission id text not the link address) which could be done by adding a div() around the submission id like `div([submissionId])` (screenshot below showing that right clicking will have copy option with the addition of the div)
image

aednichols:
Hello and thanks for your review. UUIDs are defined by RFC to be 36 characters. Because this is not a monospaced font, slight variations in rendered length are possible, which is why I added a little extra space.

I feel that the right click option is non-discoverable; the user in question is highly technical and IIRC she thought to try it far from immediately.

petesantos (approved):
Thanks for identifying and addressing this!

While I am ok with this change as a quick workaround to address the issue, our team should dive deeper into this to identify the root of what the user was trying to do and if this is something other users may want to do. I created [SATURN-1282](https://broadworkbench.atlassian.net/browse/SATURN-1282) to do this.

Adam - if you have any other information associated with this it would me much appreciated if you could add it to the above ticket.

Brett had mentioned that a different column was considered the most important prior to this. You may want to await his input before merging.

aednichols:
@petesantos added some details to the ticket. No problem waiting, I will stand by for instrucitons.

aednichols:
Also great job making your project easy to clone & get up and running with 👏 👏 👏

aednichols:
@panentheos are you cool with this? Your name came up [earlier](https://github.com/DataBiosphere/terra-ui/pull/1965#pullrequestreview-329123723) as a person with opinions.

petesantos:
Hey Adam - I discussed this with the team at standup yesterday - feel free to merge

aednichols:
Thanks!

PR #1967, opened 2019-12-11
refactor browser storage to allow component connection [SATURN-1286]

panentheos:
Evolves `syncAtomToSessionStorage` into a more general 'static' storage mechanism.

It can purge dynamic storage entries as needed to make space, but is not itself a candidate for automatic deletion. As such, it can serve as a more resilient storage area for 'important' data.

It also listens to the `storage` event, so will automatically sync across browser tabs in real-time.

It no longer maintains state in an atom, but instead provides an atom-compatible interface that can be used interchangably via `connectAtom`.

Tested existing usages by clicking through. Tested cross-tab communication with some temporary example code.

### DIFF
--- a/src/libs/browser-storage.js
+++ b/src/libs/browser-storage.js
@@ -1,20 +1,17 @@
 import _ from 'lodash/fp'
-import { getUser } from 'src/libs/auth'
-import { maybeParseJSON } from 'src/libs/utils'
+import { maybeParseJSON, subscribable } from 'src/libs/utils'
 
 
-export const getDynamic = (storage, key) => {
-  const storageKey = `dynamic-storage/${key}`
-  const data = maybeParseJSON(storage.getItem(storageKey))
-  return data && data.value
-}
+/*
+ * This library provides a higher level interface on top of localStorage and sessionStorage.
+ * Values must be JSON-serializable. The 'dynamic' version is preferred if possible, but dynamic
+ * values might be deleted in case of space overflow. For critical data, use the 'static' version.
+ */
 
-export const setDynamic = (storage, key, value) => {
-  const storageKey = `dynamic-storage/${key}`
-  const storageValue = JSON.stringify({ timestamp: Date.now(), value })
+const forceSetItem = (storage, key, value) => {
   while (true) {
     try {
-      storage.setItem(storageKey, storageValue)
+      storage.setItem(key, value)
       return
     } catch (error) {
       const candidates = _.filter(([k]) => _.startsWith('dynamic-storage/', k), _.toPairs(storage))
@@ -31,12 +28,74 @@ export const setDynamic = (storage, key, value) => {
   }
 }
 
-export const removeDynamic = (storage, key) => {
-  const storageKey = `dynamic-storage/${key}`
-  storage.removeItem(storageKey)
+export const getStatic = (storage, key) => {
+  return maybeParseJSON(storage.getItem(key))
 }
 
-const withUserPrefix = key => `${getUser().id}/${key}`
+export const setStatic = (storage, key, value) => {
+  if (value === undefined) {
+    storage.removeItem(key)
+  } else {
+    forceSetItem(storage, key, JSON.stringify(value))
+  }
+}
 
-export const getLocalPref = key => getDynamic(localStorage, withUserPrefix(key))
-export const setLocalPref = (key, value) => setDynamic(localStorage, withUserPrefix(key), value)
+export const listenStatic = (storage, key, fn) => {
+  window.addEventListener('storage', e => {
+    if (e.storageArea === storage && e.key === key) {
+      fn(maybeParseJSON(e.newValue))
+    }
+  })
+}
+
+export const getDynamic = (storage, key) => {
+  const storageKey = `dynamic-storage/${key}`
+  const data = maybeParseJSON(storage.getItem(storageKey))
+  return data && data.value
+}
+
+export const setDynamic = (storage, key, value) => {
+  const storageKey = `dynamic-storage/${key}`
+  if (value === undefined) {
+    storage.removeItem(storageKey)
+  } else {
+    forceSetItem(storage, storageKey, JSON.stringify({ timestamp: Date.now(), value }))
+  }
+}
+
+export const listenDynamic = (storage, key, fn) => {
+  const storageKey = `dynamic-storage/${key}`
+  window.addEventListener('storage', e => {
+    if (e.storageArea === storage && e.key === storageKey) {
+      const data = maybeParseJSON(e.newValue)
+      fn(data && data.value)
+    }
+  })
+}
+
+/**
+ * Returns a stateful object that manages the given storage location.
+ * Implements an atom-compaible interface, and can be passed to useAtom.
+ */
+
+export const staticStorageSlot = (storage, key) => {
+  const { subscribe, next } = subscribable()
+  const get = () => getStatic(storage, key)
+  const set = newValue => {
+    setStatic(storage, key, newValue)
+    next(newValue)
+  }
+  listenStatic(storage, key, next)
+  return { subscribe, get, set, update: fn => set(fn(get())) }
+}
+
+export const dynamicStorageSlot = (storage, key) => {
+  const { subscribe, next } = subscribable()
+  const get = () => getDynamic(storage, key)
+  const set = newValue => {
+    setDynamic(storage, key, newValue)
+    next(newValue)
+  }
+  listenDynamic(storage, key, next)
+  return { subscribe, get, set, update: fn => set(fn(get())) }
+}
--- a/src/libs/browser-storage.js
+++ b/src/libs/browser-storage.js
@@ -88,14 +88,3 @@ export const staticStorageSlot = (storage, key) => {
   listenStatic(storage, key, next)
   return { subscribe, get, set, update: fn => set(fn(get())) }
 }
-
-export const dynamicStorageSlot = (storage, key) => {
-  const { subscribe, next } = subscribable()
-  const get = () => getDynamic(storage, key)
-  const set = newValue => {
-    setDynamic(storage, key, newValue)
-    next(newValue)
-  }
-  listenDynamic(storage, key, next)
-  return { subscribe, get, set, update: fn => set(fn(get())) }
-}
--- a/src/libs/state.js
+++ b/src/libs/state.js
@@ -1,3 +1,4 @@
+import { staticStorageSlot } from 'src/libs/browser-storage'
 import * as Utils from 'src/libs/utils'
 
 
@@ -11,7 +12,8 @@ export const authStore = Utils.atom({
   profile: {}
 })
 
-export const toggleStateAtom = Utils.atom({ notebooksTab: true })
+export const toggleStateAtom = staticStorageSlot(sessionStorage, 'toggleState')
+toggleStateAtom.update(v => v || { notebooksTab: true })
 
 export const freeCreditsActive = Utils.atom(false)
 
@@ -52,5 +54,5 @@ window.ajaxOverridesStore = ajaxOverridesStore
  * Modifies config settings for testing purposes.
  * Can be set to an object which will be merged with the loaded config object.
  */
-export const configOverridesStore = Utils.atom()
+export const configOverridesStore = staticStorageSlot(sessionStorage, 'config-overrides')
 window.configOverridesStore = configOverridesStore
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -7,31 +7,37 @@ import { div, h } from 'react-hyperscript-helpers'
 import uuid from 'uuid/v4'
 
 
+export const subscribable = () => {
+  let subscribers = []
+  return {
+    subscribe: fn => {
+      subscribers = append(fn, subscribers)
+      return {
+        unsubscribe: () => {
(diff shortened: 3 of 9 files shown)


src/libs/utils.js
```diff
@@ -7,31 +7,37 @@ import { div, h } from 'react-hyperscript-helpers'
...
+}
+
 /**
  * A simple state container inspired by clojure atoms. Method names were chosen based on similarity
  * to lodash and Immutable. (deref => get, reset! => set, swap! => update, reset to go back to initial value)
  */
 export const atom = initialValue => {
   let value = initialValue
-  let subscribers = []
+  const { subscribe, next } = subscribable()
```
zarsky-broad, line 33: `makeSubscribable`? `useSubscription`? Something verb-y feels right.

src/libs/utils.js
```diff
@@ -7,31 +7,37 @@ import { div, h } from 'react-hyperscript-helpers'
...
+  return {
+    subscribe: fn => {
+      subscribers = append(fn, subscribers)
+      return {
+        unsubscribe: () => {
+          subscribers = _.without([fn], subscribers)
+        }
+      }
+    },
+    next: (...args) => {
```
zarsky-broad, line 21: `notify`?

src/libs/utils.js
```diff
@@ -40,9 +46,7 @@ export const atom = initialValue => {
 export const useAtom = theAtom => {
   const [value, setValue] = useState(theAtom.get())
   useEffect(() => {
-    const handleChange = v => setValue(v)
-    theAtom.subscribe(handleChange)
-    return () => theAtom.unsubscribe(handleChange)
+    return theAtom.subscribe(v => setValue(v)).unsubscribe
```
zarsky-broad, line 49: Why not just `theAtom.subscribe(setValue).unsubscribe`?

src/libs/utils.js
```diff
@@ -7,31 +7,37 @@ import { div, h } from 'react-hyperscript-helpers'
 import uuid from 'uuid/v4'
 
 
+export const subscribable = () => {
+  let subscribers = []
+  return {
+    subscribe: fn => {
+      subscribers = append(fn, subscribers)
+      return {
```
zarsky-broad, line 15: Why return an object here? The pattern from the other libraries we use (and the browser built-ins) is to just return the unsubscribe function directly.

src/libs/state.js
```diff
@@ -52,5 +54,5 @@ window.ajaxOverridesStore = ajaxOverridesStore
  * Modifies config settings for testing purposes.
  * Can be set to an object which will be merged with the loaded config object.
  */
-export const configOverridesStore = Utils.atom()
+export const configOverridesStore = staticStorageSlot(sessionStorage, 'config-overrides')
```
zarsky-broad, line 57: This is a significant behavior change, is this what we want? I like being able to have versions with two different overrides in different tabs, and this would prevent that, unless I'm misunderstanding something.

src/libs/browser-storage.js
```diff
@@ -1,20 +1,17 @@
...
+ * This library provides a higher level interface on top of localStorage and sessionStorage.
+ * Values must be JSON-serializable. The 'dynamic' version is preferred if possible, but dynamic
+ * values might be deleted in case of space overflow. For critical data, use the 'static' version.
+ */
 
-export const setDynamic = (storage, key, value) => {
-  const storageKey = `dynamic-storage/${key}`
-  const storageValue = JSON.stringify({ timestamp: Date.now(), value })
+const forceSetItem = (storage, key, value) => {
   while (true) {
```
zarsky-broad, line 12: I know we use them a lot, but `while(true)` loops always make me feel icky. I want that escape logic encoded in the condition; could we make this function call itself rather than looping? Like, remove the `while`, and put `forceSetItem(storage, key, value)` at the end of the `catch` block?

src/libs/browser-storage.js
```diff
@@ -1,20 +1,17 @@
...
     try {
-      storage.setItem(storageKey, storageValue)
+      storage.setItem(key, value)
       return
```
zarsky-broad, line 15: We don't need this, correct?

src/libs/browser-storage.js
```diff
@@ -31,12 +28,74 @@ export const setDynamic = (storage, key, value) => {
...
+  })
 }
 
-const withUserPrefix = key => `${getUser().id}/${key}`
+/**
+ * Returns a stateful object that manages the given storage location.
+ * Implements an atom-compaible interface, and can be passed to useAtom.
+ */
+
+export const staticStorageSlot = (storage, key) => {
```
zarsky-broad, line 81: TOL: This pattern will set a new listener for every key. Why not just have one central storage listener, and have subscribers to that? Currently there's no provision to remove a listener, but that way we could stop listening more easily if we were done.

src/libs/state.js
```diff
@@ -11,7 +12,8 @@ export const authStore = Utils.atom({
   profile: {}
 })
 
-export const toggleStateAtom = Utils.atom({ notebooksTab: true })
+export const toggleStateAtom = staticStorageSlot(sessionStorage, 'toggleState')
+toggleStateAtom.update(v => v || { notebooksTab: true })
```
zarsky-broad, line 16: Something about having this here feels icky, but I don't have a better idea, other than making the `slot` allocators take an optional default, which would feel less gross to me now that I think about it.

zarsky-broad (commented):
Doesn't seem to sync across tabs: I opened two tabs on the workspaces list, and changed the view to list from cards. The other tab didn't change. Then I opened a new tab; it was also on card view.

We also have things like the free credits banner that write to storage directly, sidestepping this mechanism, although that may in fact be desired.

src/libs/browser-storage.js
panentheos: Since JS has no tail call optimization, each recursive call always creates a new stack frame, which could potentially result in an overflow. I agree that this `while` in particular is tricky, because it's actually used as a retry mechanism, not for polling. There's an abstraction that we could write to extract this, but I think it's beyond the scope of this PR.

src/libs/browser-storage.js
panentheos: This exits the loop after a successful write.

src/libs/browser-storage.js
panentheos: The multiplexing has to happen somewhere, and I don't think there's any reason to prefer doing it in our code as opposed to using the native listeners. Note that there's another layer of multiplexing happening here, since multiple components can connect (and disconnect) to a slot.

src/libs/state.js
panentheos: This was happening before in `syncAtomToSessionStorage`, so it's not new, just moved.

src/libs/state.js
panentheos: `sessionStorage` is per-tab and doesn't sync, so this should behave the same as before

src/libs/utils.js
panentheos: This mirrors the corresponding interfaces from RXJS (which we're gradually reinventing here). At some point we'll probably want to just bring in the library, so I'm trying to bring things closer together in the interim.

https://rxjs-dev.firebaseapp.com/api/index/interface/Unsubscribable

src/libs/utils.js
panentheos: https://rxjs-dev.firebaseapp.com/api/index/interface/Observer#next

src/libs/utils.js
panentheos: Since these are stateful, I figured a noun would be appropriate to communicate that.

src/libs/utils.js
panentheos: Atoms currently provide the previous value as a second argument, which would cause React to complain.

panentheos (commented):
NB: `sessionStorage` is tab-specific, and will not sync between tabs. We don't (yet) have any usages of real-time synced `localStorage`; The 15 min sign out will be the first one

src/libs/browser-storage.js
```diff
@@ -31,12 +28,74 @@ export const setDynamic = (storage, key, value) => {
   }
 }
 
-export const removeDynamic = (storage, key) => {
+export const getStatic = (storage, key) => {
+  return maybeParseJSON(storage.getItem(key))
+}
+
+export const setStatic = (storage, key, value) => {
+  if (value === undefined) {
```
petesantos, line 36: nit: would this be clearer / fit our coding style better as a ternary?

src/libs/browser-storage.js
```diff
@@ -31,12 +28,74 @@ export const setDynamic = (storage, key, value) => {
...
+    storage.removeItem(key)
+  } else {
+    forceSetItem(storage, key, JSON.stringify(value))
+  }
+}
+
+export const listenStatic = (storage, key, fn) => {
+  window.addEventListener('storage', e => {
+    if (e.storageArea === storage && e.key === key) {
```
petesantos, line 45: nit: would this fit our coding style better as
`e.storageArea === storage && e.key === key && fn(maybeParseJSON(e.newValue))
`

src/libs/browser-storage.js
```diff
@@ -31,12 +28,74 @@ export const setDynamic = (storage, key, value) => {
...
+
+export const getDynamic = (storage, key) => {
+  const storageKey = `dynamic-storage/${key}`
+  const data = maybeParseJSON(storage.getItem(storageKey))
+  return data && data.value
+}
+
+export const setDynamic = (storage, key, value) => {
+  const storageKey = `dynamic-storage/${key}`
+  if (value === undefined) {
```
petesantos, line 59: nit: same as above wrt ternary

src/libs/browser-storage.js
```diff
@@ -31,12 +28,74 @@ export const setDynamic = (storage, key, value) => {
...
+    setStatic(storage, key, newValue)
+    next(newValue)
+  }
+  listenStatic(storage, key, next)
+  return { subscribe, get, set, update: fn => set(fn(get())) }
+}
 
-export const getLocalPref = key => getDynamic(localStorage, withUserPrefix(key))
-export const setLocalPref = (key, value) => setDynamic(localStorage, withUserPrefix(key), value)
+export const dynamicStorageSlot = (storage, key) => {
```
petesantos: Since dynamic storage slots aren't currently used should we remove it for now, or do you envision this being used in the near future?

src/libs/browser-storage.js
petesantos, line 81: Should we allow the key to have a user prefix? This could be done with an added option of withUserPrefix.

petesantos (commented):
Tested by trying the local prefs, swapped the staticStorageSlot to localStorage to confirm cross tab syncing

src/libs/browser-storage.js
zarsky-broad: Oh, duh.

src/libs/state.js
zarsky-broad: Ah, totally brainfarted on that.

src/libs/utils.js
zarsky-broad: Ahh.

src/libs/browser-storage.js
panentheos: If we were in functional code and producing a return value, then definitely. In side-effectful imperative code like this, we tend to lean toward the more imperative flow control constructs. In addition to feeling a bit more natural, it creates a visual distinction between the types of code that can help readers.

src/libs/browser-storage.js
panentheos: With the current state of the code, that would require some additional restructuring. The user id can change over time, so the listener would need to take that into account.

One way to do that would be to move all this logic into a hook, which could use e.g. `useAtom` to monitor the user id. That would work fine, but it would mean this layer could only be used from within a component, potentially making it harder to reuse in other ways.

The other approach would be to use RxJS and define the whole thing as a 'stream' (which could be derived from the 'stream' of user ids). We're trending in that direction, but not quite there yet.

src/libs/browser-storage.js
panentheos: That's a fair point, I don't think we have an immediate use case so perhaps we should remove it.

src/libs/browser-storage.js
petesantos: What about exposing the withUserPrefix function? I am thinking about the case where we need to identify specific user has timed out - would there be a way to do this?

panentheos: I think the plan is to have a single entry which is an object that maps user ids to timeouts. That way components can connect to this one slot, as well as the `authStore`, and perform the lookup to find the timeout for the current user.

petesantos (approved):
Looks good. Tested earlier by trying out areas in the application affected by this change

breilly2:
TOL, and I've only just skimmed the code... you say, "It no longer maintains state in an atom, but instead provides an atom-compatible interface that can be used interchangably via `connectAtom`." This feels misleading to me. Is there a higher-level concept than `connectAtom` that we should introduce to avoid confusion about what is actually being passed to it?

panentheos:
@breilly2 Agreed, we could definitely do some renaming to better incorporate these ideas. I was thinking along the lines of 'Store' as the common interface name, e.g. `connectStore`. And `atom` could be `memoryStore`, to communicate its lifespan.

In any case, I think we can do that in a separate PR to keep things small.

PR #1968, opened 2019-12-12
Update deps [SATURN-1292]

zarsky-broad:
Tested fairly extensively, and read all changelogs. I think I tested everything relevant, and the tests are passing, but I'd appreciate someone else testing around, too.


petesantos (approved):
Tested by clicking around in the app where the deps changed.

PR #1970, opened 2019-12-15
run nightly integration after alpha deploy

zarsky-broad:
Alpha deploy runs around 1:30 E(S/D)T, which is 5:30 or 6:30 UTC, depending on DST. Setting this to 8:00 GMT to make sure it happens well after, and is done before morning.


panentheos (approved):
Looks good.

PR #1971, opened 2019-12-16
fix run workflow test

zarsky-broad:
<!--
Hello, friend!
Thanks!
--!>
Hopefully, this will fix intermittent failures of integration tests by ensuring that the loading spinner is always there immediately when loading the workflows tab. The refresh method is called right after mount, and it turns the spinner on, so this really only applies for one tick, but we think this may have been causing occasional failures.


petesantos (approved):
Looks good

PR #1972, opened 2019-12-16
Add ability to have groups in Selector

ehanna4:
Pulled out this small PR from the ticket: [SATURN-1261]

-> Add ability to have groups in selectors

Should not currently see any difference in existing selectors, tested with my upcoming changes to the Application Compute drawer and observed the groups there, and tested existing selectors on many pages and saw no change in functionality. 

Looking particularly for code style reviews please and thank you!

[SATURN-1261]: https://broadworkbench.atlassian.net/browse/SATURN-1261

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -303,7 +303,6 @@ const commonSelectProps = {
  * @param {Object} props - see {@link https://react-select.com/props#select-props}
  * @param props.value - a member of options
  * @param {Array} props.options - can be of any type; if objects, they should each contain a value and label, unless defining getOptionLabel
- * @param {boolean} [props.withGroups=false] - see {@link https://react-select.com/advanced#replacing-builtins} for info on formatting the options array
  */
 export const BaseSelect = ({ value, newOptions, id, findValue, ...props }) => {
   const newValue = props.isMulti ? _.map(findValue, value) : findValue(value)
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -306,7 +306,7 @@ const commonSelectProps = {
  * @param props.withGroups - boolean; see {@link https://react-select.com/advanced#replacing-builtins} for info on formatting the options array
  */
 export const Select = ({ value, options, id, withGroups = false, ...props }) => {
-  const newOptions = withGroups ? _.flatten(_.map('options', options)) : options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
+  const newOptions = withGroups ? _.flatMap('options', options) : options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
   const findValue = target => _.find({ value: target }, newOptions)
   const newValue = props.isMulti ? _.map(findValue, value) : findValue(value)
 
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -303,9 +303,10 @@ const commonSelectProps = {
  * @param {Object} props - see {@link https://react-select.com/props#select-props}
  * @param props.value - a member of options
  * @param {Array} props.options - can be of any type; if objects, they should each contain a value and label, unless defining getOptionLabel
+ * @param props.withGroups - boolean; see {@link https://react-select.com/advanced#replacing-builtins} for info on formatting the options array
  */
-export const Select = ({ value, options, id, ...props }) => {
-  const newOptions = options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
+export const Select = ({ value, options, id, withGroups = false, ...props }) => {
+  const newOptions = withGroups ? _.flatten(_.map('options', options)) : options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
   const findValue = target => _.find({ value: target }, newOptions)
   const newValue = props.isMulti ? _.map(findValue, value) : findValue(value)
 
@@ -314,7 +315,7 @@ export const Select = ({ value, options, id, ...props }) => {
     ...commonSelectProps,
     getOptionLabel: ({ value, label }) => label || value.toString(),
     value: newValue || null, // need null instead of undefined to clear the select
-    options: newOptions
+    options: withGroups ? options : newOptions
   }, props))
 }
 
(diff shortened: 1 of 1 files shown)


src/components/common.js
```diff
@@ -303,9 +303,10 @@ const commonSelectProps = {
  * @param {Object} props - see {@link https://react-select.com/props#select-props}
  * @param props.value - a member of options
  * @param {Array} props.options - can be of any type; if objects, they should each contain a value and label, unless defining getOptionLabel
+ * @param props.withGroups - boolean; see {@link https://react-select.com/advanced#replacing-builtins} for info on formatting the options array
```
zarsky-broad:
```suggestion
 * @param {boolean} [props.withGroups=false] - see {@link https://react-select.com/advanced#replacing-builtins} for info on formatting the options array
```

src/components/common.js
```diff
@@ -303,9 +303,10 @@ const commonSelectProps = {
...
  */
-export const Select = ({ value, options, id, ...props }) => {
-  const newOptions = options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
+export const Select = ({ value, options, id, withGroups = false, ...props }) => {
+  const newOptions = withGroups ? _.flatMap('options', options) : options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
```
zarsky-broad: Maybe this would be better as a `Utils.cond`?

src/components/common.js
```diff
@@ -314,7 +315,7 @@ export const Select = ({ value, options, id, ...props }) => {
     ...commonSelectProps,
     getOptionLabel: ({ value, label }) => label || value.toString(),
     value: newValue || null, // need null instead of undefined to clear the select
-    options: newOptions
+    options: withGroups ? options : newOptions
```
zarsky-broad: I don't think we want to change this line?

src/components/common.js
```diff
@@ -303,9 +303,10 @@ const commonSelectProps = {
  * @param {Object} props - see {@link https://react-select.com/props#select-props}
  * @param props.value - a member of options
  * @param {Array} props.options - can be of any type; if objects, they should each contain a value and label, unless defining getOptionLabel
+ * @param props.withGroups - boolean; see {@link https://react-select.com/advanced#replacing-builtins} for info on formatting the options array
  */
-export const Select = ({ value, options, id, ...props }) => {
-  const newOptions = options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
+export const Select = ({ value, options, id, withGroups = false, ...props }) => {
```
panentheos: Looking at the underlying `react-select` code, it detects groups by looking for the presence of an `options` attribute on the items, so we could probably do the same thing (instead of passing an explicit flag).

src/components/common.js
panentheos: The way it's written, we would need to change this line, because the `newOptions` line above flattens out the group structure. However, another approach would be to leave that line alone, and instead change `findValue` so that it understands how to find in groups. That would leave `newOptions` intact, and we could continue to use it here.

ehanna4: For the `withGroup` selector I'm passing in `options` exactly as the selector is expecting them to be formatted which is why I changed this line. The `newOptions` (when it's `withGroups` ) is a flattened out list of the available options which is needed for getting the correct `value` for the chosen `option`, which then means `newOptions` is not correctly formatted to be the `options` property of the selector. 

Would it be more straight forward to construct the correctly formatted `options` for a `withGroup` selector inside this function instead?

ehanna4: Okay I will look into changing the `findValue` approach rather than changing this line

src/components/common.js
ehanna4: Checking the presence of an `options` item in the object (as far as I could figure out) required all these checks to ensure the object existed `newOptions && newOptions[0] && newOptions[0].options` so to me it felt like passing an explicit flag would be simpler/clearer? However I might not be not following your suggestion correctly. Please let me know if you still think I should try for this approach rather than the flag option based on my latest changes. thanks!

src/components/common.js
ehanna4: Updated with latest commit

src/components/common.js
```diff
@@ -303,10 +303,12 @@ const commonSelectProps = {
  * @param {Object} props - see {@link https://react-select.com/props#select-props}
  * @param props.value - a member of options
  * @param {Array} props.options - can be of any type; if objects, they should each contain a value and label, unless defining getOptionLabel
+ * @param {boolean} [props.withGroups=false] - see {@link https://react-select.com/advanced#replacing-builtins} for info on formatting the options array
  */
-export const Select = ({ value, options, id, ...props }) => {
-  const newOptions = options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
-  const findValue = target => _.find({ value: target }, newOptions)
+export const Select = ({ value, options, id, withGroups = false, ...props }) => {
+  const newOptions = !withGroups && options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
```
petesantos: Is the withGroups flag needed here? I am assuming a grouped select will require all options to be an object

src/components/common.js
petesantos: It seems like the crux of the change is in findValue. When using a grouped select we need to flatten the options.
Another approach would be to make a property around how to treat options when finding an item, by default it would be a no-op. 

Or we could make an unexposed "BaseSelect" component that accepts a findFn and wrap it with other components  (Select and GroupedSelect) that pass in the find function as a prop. This would keep the component props the same for both select types, not exposing additional complexity when using the components and removing the conditional/branching logic for grouping within the component.

ehanna4: Thanks @petesantos for the review - Please see latest changes: I have updated the solution to utilize a "`baseSelect`" with "`GroupedSelect`" and "`Select`" wrappers to pass the appropriate `findValue` function.

panentheos (approved):
Looks good.

Tested existing behavior by clicking existing select boxes.

PR #1975, opened 2019-12-18
Move Custom Env from tab to grouped select in Application Compute Configuration 

ehanna4:
The majority of the changes associated with [SATURN-1261]
note the changes in common.js are being reviewed in another PR so no need to review that here I will merge that into here once review is complete

To do which will be seen in another PR:
 - move trash can out of widget and into drawer (and add the delete warning view)
 - move the back arrow to the right side of the warning views

Tested:
- tested by observing changes in the browser - all previous behavior is the same just the look of the drawer is changed: tab is gone and the select dropdown is grouped
- created notebooks (both custom and jupyter and observed that the drawer continues to open to the currently created cluster as is expected)
- tested out the regular cluster options, starting, stopping, deleting etc. and all behaved as expected
 

[SATURN-1261]: https://broadworkbench.atlassian.net/browse/SATURN-1261


src/components/NewClusterModal.js
```diff
@@ -111,7 +111,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     this.state = {
       profile: matchingProfile ? matchingProfile.name : 'custom',
       jupyterUserScriptUri: '',
-      isCustomEnv: false, customEnvImage: '', viewMode: undefined,
+      customEnvImage: '', viewMode: 'JupyterEnv', prevViewMode: 'JupyterEnv',
```
ehanna4: `prevViewMode` is more important when the `delete warn view` is added to the drawer since you will be able to delete from any of the environment views (Jupyter, Custom, and Other) having the `prevViewMode` ensures that clicking the `back arrow` or `back button` will return the user to the correct view.

zarsky-broad: Instead of this, why not build an array of where you are and have been, breadcrumb style?

ehanna4: I hadn't considered that, but it sounds like a more straightforward approach, I will look into that. Thanks!

ehanna4:
Ready for review. The final pieces of this ticket (changing the back arrow position, moving the delete into the drawer, and having a delete warning page) will be followed up in subsequent PRs.

src/components/NewClusterModal.js
```diff
@@ -216,209 +375,77 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-          div({
-            style: {
-              display: 'grid', gridTemplateColumns: '7rem 2fr 1fr', gridGap: '1rem', alignItems: 'center', margin: '1.5rem 0 1rem', minHeight: 100
-            }
-          }, [
-            isCustomEnv ?
-              h(Fragment, [
+          div({ style: { marginBottom: '1rem' } },
+            ['Create a compute instance to launch Jupyter Notebooks or a Project-Specific software application.']),
+          div({ style: { minHeight: 100 } },
```
panentheos: nit: This `minHeight` doesn't seem to be doing anything any more, so we can probably get rid of this div entirely.

src/components/NewClusterModal.js
```diff
@@ -216,209 +375,77 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+            [
+              div([
```
panentheos: nit: This div can be removed.

src/components/NewClusterModal.js
```diff
@@ -185,22 +185,181 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+                'Enter a valid docker image to use',
+              onClick: () => selectedLeoImage === CUSTOM_MODE ?
+                this.setState({ viewMode: 'Warning' }) :
+                this.createCluster()
+            }, !!currentCluster ? 'Replace' : 'Create')
+          ])
+        ])
+    ])
+
+    const machineConfig = () => h(Fragment, [
```
breilly2, line 230: I assume this is an unmodified extracted function. Let me know if there are any notable changes to review.

src/components/NewClusterModal.js
```diff
@@ -185,22 +185,181 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+        [
+          div({ style: { marginLeft: 'auto', marginRight: '2rem' } }, [
+            h(ButtonSecondary, {
+              onClick: onDismiss
+            }, 'Cancel')
+          ]),
+          div([
+            h(ButtonPrimary, {
+              disabled: selectedLeoImage === CUSTOM_MODE && isCustomImageInvalid,
+              tooltip: selectedLeoImage === CUSTOM_MODE && isCustomImageInvalid &&
```
panentheos: It might be nice to extract `selectedLeoImage === CUSTOM_MODE` to a local variable, since it's repeated several times throughout.

src/components/common.js
```diff
@@ -304,9 +304,7 @@ const commonSelectProps = {
  * @param props.value - a member of options
  * @param {Array} props.options - can be of any type; if objects, they should each contain a value and label, unless defining getOptionLabel
  */
-export const Select = ({ value, options, id, ...props }) => {
-  const newOptions = options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
-  const findValue = target => _.find({ value: target }, newOptions)
+const BaseSelect = ({ value, newOptions, id, findValue, ...props }) => {
```
breilly2, line 302: Please update docs for this function and/or provide docs for the exported `Select` and `GroupedSelect` functions.

src/components/NewClusterModal.js
```diff
@@ -216,209 +375,77 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-                      },
-                      isSearchable: false,
-                      isClearable: false,
-                      options: [
-                        ..._.map(({ name, label }) => ({ value: name, label: `${label} computer power` }), profiles),
-                        { value: 'custom', label: 'Custom' }
-                      ]
-                    })
+                    div({ style: { marginBottom: '0.5rem' } }, [label({ htmlFor: id, style: styles.label }, 'ENVIRONMENT')]),
+                    div({ style: { height: '45px' } }, [makeGroupedEnvSelect(id)])
```
panentheos: The validation error ends up a little cramped with this at 45px. We should bump it up to at least 52px or so.

panentheos (approved):
A couple minor markup suggestions, but otherwise looks good.

Tested by creating clusters from both standard and custom images.

breilly2 (approved):
One request about function docs then 👍. Looks good! I tested lots of interactions in the application compute widget, including replacing a pre-installed runtime with a custom runtime (however, I'm currently seeing a problem because the image list we're getting has `image` instead of `imageUrl` properties... but this appears to be a change on the leo side, not our side).

src/components/common.js
ehanna4: good catch!

src/components/NewClusterModal.js
ehanna4: correct - machineConfig did not get changed at all, just moved

breilly2:
Update on issue I was seeing: it was because I was using an RStudio custom image which we won't recognize until #1977 lands.

PR #1976, opened 2019-12-18
update public data google bucket references [DFE-6]

mmorgantaylor:
Edit to src/data/reference-data.js only - changed all references to `gs://broad-references` bucket to `gs://gcp-public-data--broad-references` to reflect where google is now hosting public data used here. Verified that the new bucket links do work.


zarsky-broad (approved):
Seems like these point to things that do indeed exist; beyond that, I trust they're correct.

PR #1977, opened 2019-12-19
Add RStudio launcher [SATURN-1259]

zarsky-broad:
Tested all the permutations (I think...!). This adds a button to launch RStudio if you have a cluster with an RStudio image. It also stops you from trying to edit notebooks. Not currently handled: navigating to the RStudio page with a Jupyter cluster or the terminal page with an RStudio cluster (Jupyter should just show its base file browser, RStudio shows an error with a back button).

Needed before merging: link to docs page with instructions on saving data.

Convenient image to try it with: `us.gcr.io/anvil-gcr-public/anvil-rstudio-base:0.0.2`


src/components/ClusterManager.js
```diff
@@ -58,10 +59,10 @@ export const ClusterErrorModal = ({ cluster, onDismiss }) => {
   const [loadingClusterDetails, setLoadingClusterDetails] = useState(false)
 
   const loadClusterError = _.flow(
-    withErrorReporting('Error loading notebook runtime details'),
+    withErrorReporting('Error loading application compute instance details'),
```
panentheos, line 62: TOL: We could have split the mechanical text changes into a separate PR to make this one smaller.

src/components/cluster-common.js
```diff
@@ -0,0 +1,72 @@
...
+  const [busy, setBusy] = useState()
+
+  const startClusterOnce = withErrorReporting('Error starting application compute instance', async () => {
+    while (!signal.aborted) {
+      const currentCluster = getCluster()
+      const { status, googleProject, clusterName } = currentCluster || {}
+      const currentStatus = currentCluster && status
+      if (currentStatus === 'Stopped') {
+        setBusy(true)
+        await Ajax().Jupyter.cluster(googleProject, clusterName).start()
```
panentheos: This should be `Clusters` instead of `Jupyter`

src/components/NewClusterModal.js
```diff
@@ -130,27 +130,27 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-      currentCluster ? Ajax().Jupyter.cluster(currentCluster.googleProject, currentCluster.clusterName).details() : null,
+      currentCluster ? Ajax().Clusters.cluster(currentCluster.googleProject, currentCluster.clusterName).details() : null,
       Ajax().Buckets.getObjectPreview('terra-docker-image-documentation', 'terra-docker-versions.json', namespace, true).then(res => res.json())
     ])
 
     this.setState({ leoImages: newLeoImages })
     if (currentClusterDetails) {
       const { clusterImages, jupyterUserScriptUri } = currentClusterDetails
-      const { imageUrl } = _.find({ imageType: 'Jupyter' }, clusterImages)
+      const { imageUrl } = _.find(({ imageType }) => ['Jupyter', 'RStudio'].includes(imageType), clusterImages)
```
panentheos: nit: We could use `_.includes`

src/components/cluster-common.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export const ApplicationHeader = ({ label, labelBgColor, bgColor, children }) => {
+  return div({
+    style: {
+      backgroundColor: bgColor, display: 'flex', alignItems: 'center', borderBottom: `2px solid ${colors.dark(0.2)}`, whiteSpace: 'pre'
+    }
+  }, [
+    b({ style: { backgroundColor: labelBgColor, padding: '0.75rem 2rem', alignSelf: 'stretch', display: 'flex', alignItems: 'center' } }, [label]),
+    h(Fragment, [children])
```
panentheos: nit: This `Fragment` looks unnecessary.

src/libs/ajax.js
```diff
@@ -1008,8 +1007,7 @@ const Jupyter = signal => ({
       },
 
       setCookie: () => {
-        return fetchLeo(`${oldRoot}/setCookie`,
-          _.merge(authOpts(), { signal, credentials: 'include' }))
+        return fetchLeo(`${root}/setCookie`, _.merge(authOpts(), { signal, credentials: 'include' }))
```
panentheos, line 1010: Is this an intentional change of endpoint?

src/libs/auth.js
```diff
@@ -25,7 +25,7 @@ export const signOut = () => {
 }
 
 export const reloadAuthToken = () => {
-  return getAuthInstance().currentUser.get().reloadAuthResponse().then(() => true, () => false)
+  return getAuthInstance().currentUser.get().reloadAuthResponse().catch(() => false)
```
panentheos, line 28: Are we sure this always returns a truthy value on resolve?

src/components/ClusterManager.js
```diff
@@ -316,27 +318,28 @@ export default ajaxCaller(class ClusterManager extends PureComponent {
...
-          () => 'Open terminal'
+          () => `Open ${appName}`
         ),
-        'aria-label': 'Open terminal',
+        'aria-label': `Open ${appName}`,
         disabled: !canCompute,
         style: { marginRight: '2rem', ...styles.verticalCenter },
-        ...Utils.newTabLinkProps
-      }, [icon('terminal', { size: 24 })]),
+        ...(!isRStudioImage ? Utils.newTabLinkProps : {})
```
breilly2: Nit: I think I'd prefer `isRStudioImage ? {} : Utils.newTabLinkProps` to better link the exception case (`isRStudioImage`) with the exception result (`{}`).

Also makes it congruent with the ternary for selecting the icon below.

src/components/ClusterManager.js
```diff
@@ -265,32 +266,33 @@ export default ajaxCaller(class ClusterManager extends PureComponent {
     }
     const currentCluster = this.getCurrentCluster()
     const currentStatus = currentCluster && currentCluster.status
+    const currentTool = currentCluster && currentCluster.labels.tool
```
breilly2: Nit: Move this closer to its only use (below `renderIcon`)? Could maybe be simplified/inlined at that point.

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+  Utils.useOnMount(() => {
+    const refreshCookie = async () => {
+      const { expires_in: expiresInSec } = await Auth.reloadAuthToken() // give us max time until expiry, should be 1 hour
+
+      cookieTimeout.current = setTimeout(refreshCookie, (expiresInSec - (15 * 60)) * 1000) // refresh 15 mins early
+
+      await Clusters.notebooks(namespace, getCluster().clusterName).setCookie()
+    }
+
+    const startCluster = async () => {
```
panentheos: `ClusterKicker` already handles this, can we use that here instead?

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+  const getCluster = Utils.useGetter(cluster)
+
+  const cookieTimeout = useRef()
+  const iframeRef = useRef()
+
+  Utils.useOnMount(() => {
+    const refreshCookie = async () => {
+      const { expires_in: expiresInSec } = await Auth.reloadAuthToken() // give us max time until expiry, should be 1 hour
+
+      cookieTimeout.current = setTimeout(refreshCookie, (expiresInSec - (15 * 60)) * 1000) // refresh 15 mins early
```
panentheos: If we happen to unmount during the reload call, then there won't be a timer to cancel and this will keep looping. It might be safer to use a `while` and explicitly check `signal.aborted`

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+  const [shownCreate, setShownCreate] = useState(false)
+  const [busy, setBusy] = useState(false)
+  const getCluster = Utils.useGetter(cluster)
+
+  const cookieTimeout = useRef()
+  const iframeRef = useRef()
+
+  Utils.useOnMount(() => {
+    const refreshCookie = async () => {
+      const { expires_in: expiresInSec } = await Auth.reloadAuthToken() // give us max time until expiry, should be 1 hour
```
panentheos: Does this actually generate a new token each time? If we're just looking for the expiration, we can call `gapi.auth2.getAuthInstance().currentUser.get().getAuthResponse()`.

Stepping back a level, can we simplify this logic by just examining our access token, and calling `setCookie` whenever it changes? The auth library already auto-refreshes at the right time, so we might be able to avoid doing that work ourselves.

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+      app === 'RStudio' && h(PlaygroundHeader, [
+        'This feature is in early development. Your R Code is saved on your compute instance but not to your workspace. We encourage you to frequently ',
+        h(Link, {
+          href: '', // TODO: put a link here
+          ...Utils.newTabLinkProps
+        }, ['back up your code manually']),
+        '.'
+      ]),
+      iframe({
+        src: `${cluster.clusterUrl}/${app === 'RStudio' ? '' : 'terminals/1'}`,
```
panentheos: This will expose the terminal on any url that doesn't end with `RStudio`, which probably isn't intended. We should probably whitelist the possible values for `app`.

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+      h(StatusMessage, { hideSpinner: ['Error', 'Unknown', null].includes(clusterStatus) }, [Utils.switchCase(clusterStatus,
+        ['Creating', () => 'Creating application compute instance. You can navigate away and return in 3-5 minutes.'],
+        ['Stopping', () => 'Application compute instance is stopping, which takes ~4 minutes.'],
+        ['Starting', () => 'Starting application compute instance. You can navigate away and return in ~2 minutes.'],
+        ['Updating', () => 'Updating application compute instance. You can navigate away and return in 3-5 minutes.'],
+        ['Deleting', () => 'Deleting application compute instance, you can create a new one after it finishes.'],
+        ['Error', () => 'Error with the application compute instance, please try again.'],
+        ['Unknown', () => 'Error with the application compute instance, please try again.'],
+        [null, () => 'Create an application compute instance to continue.'],
+        [Utils.DEFAULT, () => 'Getting ready...']
```
breilly2: I left a screen open to RStudio while eating lunch. When I came back, I found it showing "Getting ready..." continuously. It looks like the runtime auto-paused so its status is now "Stopped" so it's hitting this case. If on this page for a stopped runtime, we should probably just say that and offer an extra gesture for restarting the runtime.

Should also still handle "Deleted" since the runtime could be deleted in another tab.

panentheos: TOL: This message doesn't feel quite right to me. Right now it flashes briefly during the initial cluster load, but it would also show up indefinitely if we ever got a status we didn't expect.

Also:
* Is `'Unknown'` an expected status? What are we supposed to do with that?
* `Deleting` gets filtered out in the wrapper, so we shouldn't see it here.
* We need a message for `Stopped`

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+      )]),
+      h(NewClusterModal, {
+        isOpen: cluster === null && !shownCreate,
+        namespace, currentCluster: cluster,
+        onDismiss: () => setShownCreate(true),
+        onSuccess: withErrorReporting('Error creating cluster', async promise => {
+          setShownCreate(true)
+          setBusy(true)
```
panentheos: nit: This could use `withBusyState`

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+        ['Starting', () => 'Starting application compute instance. You can navigate away and return in ~2 minutes.'],
+        ['Updating', () => 'Updating application compute instance. You can navigate away and return in 3-5 minutes.'],
+        ['Deleting', () => 'Deleting application compute instance, you can create a new one after it finishes.'],
+        ['Error', () => 'Error with the application compute instance, please try again.'],
+        ['Unknown', () => 'Error with the application compute instance, please try again.'],
+        [null, () => 'Create an application compute instance to continue.'],
+        [Utils.DEFAULT, () => 'Getting ready...']
+      )]),
+      h(NewClusterModal, {
+        isOpen: cluster === null && !shownCreate,
```
panentheos: Currently if you go to the launcher page with a cluster running, and then delete it, it immediately pops up the create modal, which doesn't seem right. If we switch to using the `ClusterKicker` on this page, it has a mechanism for doing a one-time open on page load.

src/components/TopBar.js
```diff
@@ -174,7 +174,7 @@ const TopBar = Utils.connectAtom(authStore, 'authState')(class TopBar extends Co
               h(DropDownSubItem, {
                 href: Nav.getLink('clusters'),
                 onClick: () => this.hideNav()
-              }, ['Notebook Runtimes']),
+              }, ['Application Instances']),
```
breilly2: Should be "Application Compute" to match the text in the upper-right widget. Neither is a perfectly clear shorthand for "application compute instance" but I think "application compute" is slightly better and we should use it consistently when space available for the text is limited.

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+        try {
+          await refreshClusters()
+          const cluster = getCluster()
+          const status = cluster?.status
+
+          if (status === 'Running') {
+            await refreshCookie()
+            setCookieReady(true)
+            iframeRef.current.onload = function() { this.contentWindow.focus() }
+            return
```
panentheos: Since this currently only runs on mount at the page level, this flow will not be repeated if the cluster gets stopped and then restarted without leaving the page. We might want to do something like the `NotebookLauncher` page, where the iframe is pushed into a sub-component along with its setup code. That way, we can leverage the mount/unmount cycle to perform these steps each time.

breilly2 (commented):
Some initial comments.

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+  })
+
+  const clusterStatus = cluster && cluster.status
+
+  return (cookieReady && clusterStatus === 'Running') ?
+    h(Fragment, [
+      app === 'RStudio' && h(PlaygroundHeader, [
+        'This feature is in early development. Your R Code is saved on your compute instance but not to your workspace. We encourage you to frequently ',
+        h(Link, {
+          href: '', // TODO: put a link here
```
breilly2: Still need the link here.

src/pages/workspaces/workspace/notebooks/AppLauncher.js
```diff
@@ -0,0 +1,126 @@
...
+import * as Utils from 'src/libs/utils'
+import { wrapWorkspace } from 'src/pages/workspaces/workspace/WorkspaceContainer'
+
+
+const AppLauncher = _.flow(
+  Utils.forwardRefWithName('AppLauncher'),
+  wrapWorkspace({
+    breadcrumbs: props => breadcrumbs.commonPaths.workspaceDashboard(props),
+    title: _.get('app'),
+    activeTab: 'notebooks'
```
breilly2: Suggestion: remove `activeTab` from this use of `wrapWorkspace`. I realize that this also applies to terminal but I think it's misleading to put RStudio logically under "Notebooks".

src/components/NewClusterModal.js
zarsky-broad: Sure, but I don't see why?

src/libs/ajax.js
zarsky-broad: Yup, the underlying bug has been fixed.

src/libs/auth.js
zarsky-broad: Yes, the documentation says the auth response is always an object.

src/pages/workspaces/workspace/notebooks/AppLauncher.js
zarsky-broad: What if I put `terminal` as the only special case?

src/pages/workspaces/workspace/notebooks/AppLauncher.js
zarsky-broad: Delet[ing/ed] is filtered out by the wrapper.

src/components/ClusterManager.js
zarsky-broad: I think the idea is that ultimately most apps will open in the same tab, so I'm inclined to keep Jupyter as the exception. However, I'll definitely clean up this conditional.

src/components/cluster-common.js
```diff
@@ -0,0 +1,96 @@
...
+import { spinnerOverlay } from 'src/components/common'
+import { icon, spinner } from 'src/components/icons'
+import { Ajax } from 'src/libs/ajax'
+import colors from 'src/libs/colors'
+import { withErrorIgnoring, withErrorReporting } from 'src/libs/error'
+import * as Utils from 'src/libs/utils'
+
+
+export const StatusMessage = ({ hideSpinner, children }) => {
+  return div({ style: { padding: '1.5rem 2rem', display: 'flex' } }, [
```
panentheos: nit: This padding ends up changing the height of the `ApplicationHeader` when it's displayed, which is a minor visual change (probably unintended)

src/components/ClusterManager.js
```diff
@@ -74,7 +75,7 @@ export const ClusterErrorModal = ({ cluster, onDismiss }) => {
   Utils.useOnMount(() => { loadClusterError() })
 
   return h(Modal, {
-    title: userscriptError ? 'Notebook Runtime Creation Failed due to Userscript Error' : 'Notebook Runtime Creation Failed',
+    title: userscriptError ? 'Application Compute Instance Creation Failed due to Userscript Error' : 'Application instance Creation Failed',
     showCancel: false,
```
ehanna4, line 79: nit: Should these both say "Application Compute Instance" ?

src/components/NewClusterModal.js
```diff
@@ -236,7 +236,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
         }
       }, [
         div({ style: { fontSize: '0.875rem', fontWeight: 600, marginBottom: '0.5rem' } }, ['COMPUTE POWER']),
-        div({ style: { marginBottom: '1rem' } }, ['Select from one of the compute runtime profiles or define your own']),
+        div({ style: { marginBottom: '1rem' } }, ['Select from one of the compute instance profiles or define your own']),
```
ehanna4, line 239: I believe this text change is covered by another ticket but in the mock this text is: "Select from one of the default compute cluster profiles or define your own"

src/components/NewClusterModal.js
panentheos: Just that we generally use lodash utils over builtins.

panentheos (approved):
Looks good.

Tested locally by going through a wide variety of cluster and page transitions.

PR #1979, opened 2020-01-02
Remove queue status from UI [SATURN-1147]

cameron-ardell:
From the [ticket](https://broadworkbench.atlassian.net/browse/SATURN-1147):

> Queue status is hilariously wrong and has been forever. I believe it does more harm than good.
> Example:
> I submitted a workflow today at 5:09 PM. The queue status predicted a wait time of "about 1 hour". 
> The workflow succeeded at 5:10.


ehanna4:
I am still seeing `queue status` on the job history page that has the list of all the submissions, but it's not on the second page with the details about a specific submission anymore. I see that the screenshots in the ticket are only of the second page but it seems if the functionality is broken anyway we should take it out in all places?

cameron-ardell:
Oh yikes thanks! Didn't know there was a second one since it wasn't listed on the ticket. 
Not sure how to find it, but updated it so I think it's removed now. Can you check @ehanna4 ?

ehanna4:
Looks good, I don't see `queue status` anymore. The component I believe is `SubmissionQueueStatus` in a file called the same. After the latest changes it doesn't appear to be being used anywhere, so I think that's good. I'm not sure if maybe the whole file should just be removed now though since it's not used at all now? Or if it's good to keep for documentation and/or as a starting point if we want to add this back when it can be fixed later?

cameron-ardell:
I was thinking we should keep it since currently we're getting rid of it since estimation isn't great, but we could make it good later and then there's less work to do

breilly2:
In general, I'm in favor of removing dead code. Specifically as the person who added `SubmissionQueueStatus`, I have no problem with it going away. It was pretty straightforward to create and I might even structure it a little differently now.

cameron-ardell:
Can someone re-review this?

breilly2 (approved):
👍 Looks good! I confirmed that the Queue Status links are gone in both places.

PR #1981, opened 2020-01-03
rename useAtom to useStore [SATURN-1305]

panentheos:
Since a recent refactor separated out the interface of `atom` as a separate concept, names like `useAtom` are no longer completely accurate. This is a mechanical renaming of `useAtom` to `useStore`, and `connectAtom` to `connectStore`. It also improves a few comments to better describe things.

Tested by clicking around.


breilly2 (approved):
👍 Looks good. Tested by clicking around, specifically on changed pages, and making sure to hit something that would cause a notification.

PR #1982, opened 2020-01-06
Eh cluster trash to drawer SATURN-1261

ehanna4:
The final pieces to this ticket: [SATURN-1261](https://broadworkbench.atlassian.net/browse/SATURN-1261?atlOrigin=eyJpIjoiZDViNWRiMmNmMzBkNDg2Zjg4ZWQxNTMyM2E3YjNmNmIiLCJwIjoiaiJ9)

- Remove the trash icon from the cluster widget and add delete button inside drawer
- & add a warning delete view (previously a delete warning modal - which is where the wording was taken from as there is no mock for this view)

- Move 'back-arrow' the the right side of titlebar

Testing:
Was able to create and delete Jupyter and custom images. Clicked around and the back-arrows, etc. work as expected.

EDIT: Also checked the IGV drawer in the data table page as this also uses the title bar component. The confirmation from UI/UX is that all previous arrows across Terra should be migrated to the right corner rather than left

And checked out the deleteModal that is found in the clusters section and observed the text change looks okay there as well matching the deleteModalDrawer


ehanna4:
More changes coming:
- updating the look of the compute power section after discussion with Joy & Isaac

src/components/input.js
```diff
@@ -12,7 +12,7 @@ import * as Utils from 'src/libs/utils'
 const styles = {
   input: {
     height: '2.25rem',
-    border: `1px solid ${colors.light()}`, borderRadius: 4
+    border: `1px solid ${colors.dark(0.55)}`, borderRadius: 4
```
ehanna4, line 15: Got approval from Lou on this change - makes it consistent between selects and inputs now to both have dark borders

src/components/TitleBar.js
```diff
@@ -4,23 +4,24 @@ import { Link } from 'src/components/common'
 import { icon } from 'src/components/icons'
 
 
-const TitleBar = ({ onPrevious, title, onDismiss, titleExtras }) => {
+const TitleBar = ({ onPrevious, title, onDismiss, titleExtras, titleSize = '0.875rem', arrowSize = '17', dismissSize = '20' }) => {
```
ehanna4: Would appreciate input on this. Currently only the cluster config drawer and IGV drawer use this titlebar component. I added these extra props so could customize to match the mock for the cluster-config drawer and added defaults so that IGV drawer won't change. Wasn't sure if should have gone with a Wrapper approach instead? Or just had the IGV drawer also change?

src/components/ClusterManager.js
```diff
@@ -353,24 +344,16 @@ export default ajaxCaller(class ClusterManager extends PureComponent {
           disabled: isDisabled
         }, [
           div({ style: { marginLeft: '0.5rem', paddingRight: '0.5rem', color: colors.dark() } }, [
-            div({ style: { fontSize: 12, fontWeight: 'bold' } }, 'Notebook Runtime'),
+            div({ style: { fontSize: 12, fontWeight: 'bold' } }, 'Application Compute'),
```
breilly2: We're not changing this text until we launch RStudio support, right?

src/components/TitleBar.js
panentheos: This feels like a lot of props for a relatively small component, which suggests that maybe the abstraction isn't quite right. The fundamental question is whether this is intended to be a consistent pattern, or just a specific layout for this case. (UX may have some opinions on that). If it's a pattern, then we should reduce the amount of customization and let the other use cases inherit the changes as well. If it's more one-off, then perhaps we should dissolve this component (perhaps in favor of a few smaller ones that could be composed together).

src/components/NewClusterModal.js
```diff
@@ -47,46 +47,49 @@ const MachineSelector = ({ machineType, onChangeMachineType, diskSize, onChangeD
...
-          h(Select, {
-            isDisabled: readOnly,
-            id,
-            isSearchable: false,
-            value: currentCpu,
-            onChange: ({ value }) => onChangeMachineType(_.find({ cpu: value }, machineTypes).name),
-            options: _.uniq(_.map('cpu', machineTypes))
-          })
-        ])
+        readOnly ? div({ style: { border: `1px solid ${colors.dark(0.2)}`, borderRadius: '4px', padding: '0.5rem' } }, [currentCpu]) :
```
panentheos: Since this 'fake input' style is used in a couple places, we should factor it out into a local helper component to cut down on repetition.

src/components/NewClusterModal.js
```diff
@@ -47,46 +47,49 @@ const MachineSelector = ({ machineType, onChangeMachineType, diskSize, onChangeD
...
+          div([
+            h(Select, {
+              isDisabled: readOnly,
```
panentheos: nit: with the current logic, this will never appear disabled, so this could be removed (also below)

src/components/NewClusterModal.js
```diff
@@ -127,6 +130,13 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     }
   }
 
+  deleteCluster() {
+    const { currentCluster, onSuccess } = this.props
+    const { googleProject, clusterName } = currentCluster
+
+    onSuccess(Ajax().Jupyter.cluster(googleProject, clusterName).delete())
```
panentheos: Calling `onSuccess` here results in it getting called twice in the 'create' case, since `createCluster` calls it already. Also, nothing is being returned here, so the `Promise.all` isn't going to do the same thing as before. If we just returned the promise here, then we could call `onSuccess` in the click handler for the delete button, and everything should behave as expected.

src/components/TitleBar.js
ehanna4: Thanks for the guidance, I reached out to Lou and got confirmation that the style for the cluster drawer should be the new pattern to follow so I can simplify this/take out the customization

src/components/NewClusterModal.js
```diff
@@ -205,10 +215,16 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-      div({ style: { flexGrow: 1 } }),
       div(
-        { style: { display: 'flex', marginTop: '1rem' } },
+        { style: { display: 'flex', marginTop: '3rem', marginBottom: '1rem' } },
         [
+          currentCluster && div([
+            h(ButtonSecondary, {
+              onClick: () => this.setState({ isDeleteView: true }),
+              tooltip: 'Delete runtime configuration',
+              'aria-label': 'Delete runtime configuration'
```
panentheos: This tooltip and label seem redundant, since they're not really adding much information beyond the existing button text. Can we remove them?

src/components/NewClusterModal.js
```diff
@@ -335,8 +350,8 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
             ])
           ])
         ]),
-        div({ style: styles.row }, [
-          span({ style: { ...styles.label, marginRight: '0.25rem' } }, ['Cost:']),
+        div({ style: { backgroundColor: colors.dark(0.2), borderRadius: '100px', width: 'fit-content', padding: '0.75rem 1.25rem', ...styles.row } }, [
```
panentheos, line 351: I just learned about `fit-content`, thanks!

src/components/NewClusterModal.js
```diff
@@ -386,7 +400,29 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
           ])
         ])
       })],
-      [Utils.DEFAULT, () => ({
+      [isDeleteView === true, () => ({
```
panentheos: nit:
```suggestion
      [isDeleteView, () => ({
```

src/components/NewClusterModal.js
```diff
@@ -439,16 +475,24 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
     return h(Fragment, [
       h(TitleBar, {
-        title: viewMode === 'Packages' ? 'INSTALLED PACKAGES' : 'RUNTIME CONFIGURATION',
+        titleSize: '1rem',
+        arrowSize: '22',
+        dismissSize: '25',
+        title: Utils.cond(
+          [viewMode === 'Packages', () => 'INSTALLED PACKAGES'],
+          [viewMode === 'Warning', () => 'WARNING!'],
+          [isDeleteView === true, () => 'DELETE RUNTIME CONFIGURATION?'],
```
panentheos: nit:
```suggestion
          [isDeleteView, () => 'DELETE RUNTIME CONFIGURATION?'],
```

src/components/NewClusterModal.js
```diff
@@ -386,7 +400,29 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
           ])
         ])
       })],
-      [Utils.DEFAULT, () => ({
+      [isDeleteView === true, () => ({
+        onPrevious: () => this.setState({ isDeleteView: false }),
+        contents: h(Fragment, [
+          p({ style: { margin: '0px' } }, [
+            'Deleting your runtime configuration will stop all running notebooks and associated costs. '
```
panentheos: TOL: I know this language is mid-transition, but calling it 'configuration' seems like a miscommunication to me. What we're doing is deleting __the runtime__, which is a fairly heavy and destructive operation, and we should make sure we're representing that accurately. (I'm thinking here about more knowledgable users for whom 'configuration' might sound like something lighter, e.g. a yaml file, rather than a whole VM.)

src/components/ClusterManager.js
ehanna4: Correct, fixed thanks!

src/components/ClusterManager.js
```diff
@@ -97,13 +97,18 @@ export const DeleteClusterModal = ({ cluster: { googleProject, clusterName }, on
...
+      span({ style: { fontWeight: 600 } }, ['delete any files on the associated hard disk ']),
+      '(e.g. input data or analysis outputs) and installed packages. To permanently save these files, ',
+      h(Link, {
+        href: 'https://support.terra.bio/hc/en-us/articles/360026639112',
+        ...Utils.newTabLinkProps
+      }, ['move them to the workspace bucket.'])
+    ]),
+    p({ style: { margin: '14px 0px 0px' } },
+      ['Deleting your runtime will stop all running notebooks and associated costs. You can recreate your runtime later, ' +
+      'which will take several minutes.']),
```
panentheos: Since this text is the same in both places, we could extract it into a shared component to reduce the duplication.

src/components/common.js
```diff
@@ -290,14 +290,24 @@ const commonSelectProps = {
...
-        div({ style: { display: 'flex', alignItems: 'center' } }, [
+        div({ style: { display: 'flex', alignItems: 'center', height: 25 } }, [
           div({ style: { flexGrow: 1 } }, children),
           props.isSelected && icon('check', { size: 14, style: { flex: 'none', marginLeft: '0.5rem', color: colors.dark(0.5) } })
         ])
       ])
     }
   }
 }
+const formatGroupLabel = groupedLabels => (
```
panentheos: nit: `group` might be a better param name here

src/components/common.js
```diff
@@ -307,7 +317,8 @@ const BaseSelect = ({ value, newOptions, id, findValue, ...props }) => {
     ...commonSelectProps,
     getOptionLabel: ({ value, label }) => label || value.toString(),
     value: newValue || null, // need null instead of undefined to clear the select
-    options: newOptions
+    options: newOptions,
+    formatGroupLabel: groupedLabels => formatGroupLabel(groupedLabels)
```
panentheos: nit: the function already has the right signature and a matching name, so this could be:
```suggestion
    formatGroupLabel
```

src/pages/Clusters.js
```diff
@@ -56,7 +55,7 @@ const Clusters = () => {
   const clustersByProject = _.groupBy('googleProject', clusters)
 
   return h(Fragment, [
-    h(TopBar, { title: 'Notebook runtimes', href: Nav.getLink('clusters') }),
+    h(TopBar, { title: 'Notebook runtimes' }),
```
ehanna4, line 58: Removing this href allows the Terra logo to redirect to the home page from the notebook runtimes view

src/pages/billing/List.js
```diff
@@ -262,7 +262,7 @@ export const BillingList = _.flow(
     const breadcrumbs = `Billing > Billing Project`
 
     return h(Fragment, [
-      h(TopBar, { title: 'Billing', href: Nav.getLink('billing') }, [
+      h(TopBar, { title: 'Billing' }, [
```
ehanna4, line 265: same here

src/components/input.js
breilly2, line 15: Should also apply to `textarea`.

breilly2 (approved):
👍 Looks good! Just want to see textarea styling consistent with the change to text input styling.

PR #1983, opened 2020-01-06
fix minor style issue on notebook cards

panentheos:
The bottom box was extending too far and obscuring the shadow:
image
This fixes it:
image


breilly2 (approved):
👍 Looks good! I tested in both Chrome and Firefox.

PR #1985, opened 2020-01-06
extract TSV util

panentheos:
This factors out the relevant string manipulation, so the data manipulation can just use arrays.


breilly2 (approved):
👍 Looks good. I understand the refactoring and I tested by downloading a sample set (covered both entity and entity membership).

PR #1986, opened 2020-01-06
Use Yarn instead of NPM [SATURN-1309]

zarsky-broad:
I got tired of package-lock thrash, so this switches us to use yarn instead of npm.

To run with this:

```sh
brew install yarn
yarn install # yarn build, yarn test, yarn eslint src, yarn whatever
```


panentheos (approved):
Looks good, thanks for putting this together.

Tested locally, and everything works for me.

PR #1987, opened 2020-01-06
Wait for ws wf list page to load before waiting for spinners

zarsky-broad:
Makes sure we're really on the workspace's list page, by checking that the find workflow button is there, before waiting for the spinner covering it to go away, to make sure that we're waiting for the _right_ spinner before trying to click the button.

### DIFF
--- a/integration-tests/tests/run-workflow.integration-test.js
+++ b/integration-tests/tests/run-workflow.integration-test.js
@@ -1,10 +1,11 @@
 const pRetry = require('p-retry')
 const { testUrl, workflowName, billingProject } = require('../utils/integration-config')
 const { withWorkspace, createEntityInWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, input, signIntoTerra, waitForNoSpinners, findInGrid, navChild, findInDataTableRow } = require('../utils/integration-utils')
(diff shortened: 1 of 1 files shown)


integration-tests/tests/run-workflow.integration-test.js
```diff
@@ -1,10 +1,11 @@
 const pRetry = require('p-retry')
 const { testUrl, workflowName, billingProject } = require('../utils/integration-config')
 const { withWorkspace, createEntityInWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, input, signIntoTerra, waitForNoSpinners, findInGrid, navChild, findInDataTableRow } = require('../utils/integration-utils')
+const { click, clickable, findElement, input, signIntoTerra, waitForNoSpinners, findInGrid, navChild, findInDataTableRow } = require('../utils/integration-utils')
 
 
 const testEntity = { name: 'test_entity_1', entityType: 'test_entity', attributes: { input: 'foo' } }
+const findWorkflowButton = clickable({ textContains: 'Find a Workflow' })
```
breilly2, line 8: Nit: The `find` prefix feels out of place because this is a selector for a noun (`clickable`) rather than a verb (`findElement`). Maybe just name it `workflowButton` or come up with some other noun-like word.

breilly2 (approved):
👍 Logic looks good. Just one suggestion about naming.

PR #1988, opened 2020-01-10
update deps

zarsky-broad:
First yarn dep update PR! Turns out running `yarn upgrade` with `--latest` will cause it to update the package.json; it doesn't otherwise.

Read changelogs and tested.


package.json
```diff
@@ -11,25 +11,25 @@
     "array-move": "^2.2.1",
     "clipboard-polyfill": "^2.8.6",
     "color": "^3.1.2",
-    "date-fns": "^2.8.1",
+    "date-fns": "^2.9.0",
     "debounce-promise": "^3.1.2",
-    "downshift": "^3.4.8",
+    "downshift": "^4.0.4",
```
breilly2: This is already up to 4.0.5 as of ~3 hours ago.

panentheos (approved):
Looks good, tested locally by clicking through affected areas.

PR #1989, opened 2020-01-10
revert wording for now

zarsky-broad:
'instance' isn't yet ready for prime time, revert this to switch back


ehanna4 (approved):
Tested by looking through UI and code for no instances of "Application Compute"

PR #1990, opened 2020-01-13
Add ability to rename a workflow when importing from dockstore [SATURN-1145]

ehanna4:
Steps to test/reproduce:
From the homepage go to `browse data`. From there choose the `code & workflows` tab and from there choose to `import from dockstore`. Once you've chosen a wdl from dockstore select to `import to Terra` and once you are redirected be sure to change back from dev to localhost.

Tested by:
- importing a variety of dockstore workflows and renaming them
- attempting to rename workflows to names that were already in use
- attempting to rename workflows invalid names
- importing dockstore workflows and not renaming them
- importing dockstore workflows to brand new workspaces
- imported workflows on both Anvil and Datastage


breilly2:
First testing observation: The import button is not disabled if the workflow name is invalid because it contains a space. However, the import worked! If I try to duplicate it or copy it to another workspace, those buttons are disabled due to the space in the name.

I think it's entirely possible that we added this protection because of problems down the line (maybe when actually running the workflow). We should disable the import button if the name is invalid.

src/components/workspace-utils.js
```diff
@@ -59,7 +59,7 @@ export const WorkspaceSelector = ({ workspaces, value, onChange, ...props }) =>
 export const WorkspaceImporter = _.flow(
   Utils.withDisplayName('WorkspaceImporter'),
   withWorkspaces
-)(({ workspaces, refreshWorkspaces, onImport, authorizationDomain: ad, selectedWorkspaceId: initialWs }) => {
+)(({ workspaces, refreshWorkspaces, onImport, authorizationDomain: ad, selectedWorkspaceId: initialWs, isWorkflowInvalid = false }) => {
```
breilly2: Nit: I prefer boolean names based on the affirmative, i.e. `isWorkflowNameValid`. I've read this advice before but I don't remember where. I think it tends to avoid having to think through double negatives.

breilly2 (approved):
👍 Looks good! I tested locally by importing a workflow from Dockstore, checking various edge cases (similar to your very thorough testing).

src/components/workspace-utils.js
```diff
@@ -69,18 +69,23 @@ export const WorkspaceImporter = _.flow(
...
       value: selectedWorkspaceId,
       onChange: setSelectedWorkspaceId
     }),
     div({ style: { display: 'flex', alignItems: 'center', marginTop: '1rem' } }, [
       h(ButtonPrimary, {
-        disabled: !selectedWorkspace,
+        disabled: !selectedWorkspace || isWorkflowInvalid,
+        tooltip: Utils.cond([!selectedWorkspace, 'Select valid a workspace to import'],
+          [isWorkflowInvalid, 'Enter a valid a workflow name to import'],
+          'Import workflow to workspace'
```
panentheos, line 82: TOL: This component is also currently used on the import data page, so putting a workflow-specific tooltip here isn't technically correct. However, another PR is about to change that page to stop using the component, so it's probably not worth worrying about.

If we did want to preserve the generic nature of this component, we could change the prop to be something like 'additional errors', which if present would disable the button and drive the tooltip, without necessarily having to know what the content is.

src/libs/workflow-utils.js
```diff
@@ -0,0 +1,25 @@
...
+      pattern: /^[A-Za-z0-9_\-.]*$/,
+      message: 'can only contain letters, numbers, underscores, dashes, and periods'
+    }
+  })
+}
+
+export const validateWorkflowNameWithWorkSpace = (workflowName, selectedWorkspace) => {
+  return validate({ workflowName, selectedWorkspace }, {
+    selectedWorkspace: { presence: true },
+    workflowName: workflowNameValidation()
```
panentheos: TOL: In general, if we want to share validation logic in several places, it tends to be more natural to extract the configuration (either for a single field or several), but let the consumers make the actual call to `validate`. That way, consumers can compose the validators as they see fit using standard data manipulation.

src/pages/ImportWorkflow.js
```diff
@@ -72,8 +74,17 @@ const DockstoreImporter = ajaxCaller(class DockstoreImporter extends Component {
         wdl && h(WDLViewer, { wdl, style: { height: 500 } })
       ]),
       div({ style: { ...styles.card, margin: '0 2.5rem', maxWidth: 430 } }, [
-        div({ style: styles.title }, ['Destination Workspace']),
-        h(WorkspaceImporter, { onImport: ws => this.import_(ws) }),
+        div({ style: styles.title }, ['Workflow Name']),
+        h(ValidatedInput, {
+          inputProps: {
+            'aria-label': 'Workflow name',
```
panentheos: It's good that we're thinking about accessibility here. Instead of using an `aria-label`, we could turn the text above into an actual label and wire it up with an `IdContainer`

src/libs/workflow-utils.js
ehanna4: Please see latest commit and let me know if that's along the lines of what you were thinking

src/pages/ImportWorkflow.js
```diff
@@ -45,15 +48,22 @@ const DockstoreImporter = ajaxCaller(class DockstoreImporter extends Component {
...
   render() {
     const { path, version } = this.props
-    const { isImporting, wdl } = this.state
+    const { isImporting, wdl, workflowName } = this.state
+
+    const workflowNameValidationDetails = () => {
+      return (validate({ workflowName }, {
+        workflowName: workflowNameValidation()
+      }))
+    }
```
panentheos: nit: Since validation is inherently 'pure', we could remove the function wrapper and just directly call `validate` here, and then use the resulting value in both places below.

panentheos (approved):
Looks good, see one minor comment.

Tested by clicking through the flow locally.

panentheos (approved):
Latest changes look good, thanks.

PR #1991, opened 2020-01-13
refactor data import page

panentheos:
This is a preliminary refactor of the data import page in advance of the new functionality there.
Converts to a function component with hooks, and improves some names. There should be no visible changes.

Tested by importing data from a data explorer.


breilly2 (approved):
👍 Looks good. I also tested by importing from a data explorer. I would have liked to import a PFB, but there currently isn't any data available in Gen3 staging to exercise that flow. However, I did scrutinize the diff and this appears to be a pure refactor to me so I have high confidence.

PR #1992, opened 2020-01-13
Fix delete cluster call

ehanna4:
Need this change to be able to delete clusters.

Tested am able to successfully delete a cluster


breilly2 (approved):
👍 Looks good. I tested by deleting a notebook runtime from the notebook runtime drawer. I also glanced at the previous PR to double check that there aren't any other places affected by this refactor.

PR #1993, opened 2020-01-13
some fixes from LGTM

zarsky-broad:
see https://lgtm.com/projects/g/DataBiosphere/terra-ui/alerts/?mode=list

<!--
Hello, friend!
Thanks!
--!>


src/pages/workspaces/workspace/Data.js
```diff
@@ -326,10 +324,7 @@ const LocalVariablesContent = class LocalVariablesContent extends Component {
         onDismiss: () => this.setState({ deleteIndex: undefined }),
         title: 'Are you sure you wish to delete this variable?',
         okButton: h(ButtonPrimary, {
-          onClick: _.flow(
-            withErrorReporting('Error deleting workspace variable'),
-            Utils.withBusyState(v => this.setState({ saving: v }))
```
breilly2, line 331: Is it possible that this should have been setting `loading` instead of saving? Or that there was an intent to read `saving` that was dropped somewhere along the way? The other possible fix for this warning would be to add something that reads this state. I'd like to be convinced that we're applying the correct fix.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -872,11 +872,13 @@ const WorkflowView = _.flow(
           errors,
           onBrowse: name => this.setState({ variableSelected: name }),
           onChange: (name, v) => this.setState(_.set(['modifiedConfig', key, name], v)),
-          onSetDefaults: () => {
-            this.setState(_.set(['modifiedConfig', 'outputs'], _.fromPairs(_.map(({ name }) => {
-              return [name, `this.${_.last(name.split('.'))}`]
-            }, modifiedInputsOutputs.outputs))))
-          },
+          onSetDefaults: () => this.setState(oldState => {
```
breilly2, line 875: I think the old form is technically okay in this case, but I understand that lgtm has no way of knowing that.

src/pages/workspaces/workspace/Data.js
panentheos: Looks like I added the code originally. It never read the value, but I recognize the intended pattern, and it was almost certainly intended to drive the spinner, e.g. `(loading || saving) && spinner`. We might want to see if something like that still makes sense.

panentheos (dismissed):
Looks good, see comment.

src/pages/workspaces/workspace/workflows/WorkflowView.js
zarsky-broad: Its issue is that `modifiedInputsOutputs` is being read from old state, not the fresh callback state.

zarsky-broad:
Ok, there's now a spinner based on that state. This'll need a re-review; I _strongly_ recommend doing so with whitespace hidden.

src/pages/workspaces/workspace/workflows/WorkflowView.js
breilly2: Yes, and I have a loose grasp on why that can sometimes be a problem. What I don't understand is what could actually go wrong in this particular case. I'm curious to understand this better. Let's chat in person.

src/pages/workspaces/workspace/Data.js
```diff
@@ -129,216 +129,229 @@ const getReferenceData = _.flow(
...
+      onDismiss: () => setDeleteIndex(),
+      title: 'Are you sure you wish to delete this variable?',
+      okButton: h(ButtonPrimary, {
+        onClick: _.flow(
+          withErrorReporting('Error deleting workspace variable'),
+          Utils.withBusyState(setBusy)
+        )(async () => {
+          setDeleteIndex()
+          await Ajax().Workspaces.workspace(namespace, name).deleteAttributes([amendedAttributes[deleteIndex][0]])
+          loadAttributes()
```
panentheos: Since `loadAttributes` is writing to the same busy flag, this will cause the flag to get cleared early due to a race.
```suggestion
          await loadAttributes()
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -129,216 +129,229 @@ const getReferenceData = _.flow(
...
+        value: textFilter
+      })
+    ]),
+    div({ style: { flex: 1 } }, [
+      h(AutoSizer, [({ width, height }) => h(FlexTable, {
+        width, height, rowCount: amendedAttributes.length,
+        onScroll: y => saveScroll(0, y),
+        initialY,
+        hoverHighlight: true,
+        noContentMessage: _.isEmpty(amendedAttributes) ? 'No Workspace Data defined' : 'No matching data',
```
panentheos: TOL: There was previously a minor bug where 'No Workspace Data defined' shows when searching for a non-existent string. This refactor changes the display behavior, but preserves the bug. Since we're moving things around, perhaps we should fix it in the process. (It would involve separating the data filtering above back into two steps)

src/pages/workspaces/workspace/Data.js
```diff
@@ -129,216 +129,229 @@ const getReferenceData = _.flow(
...
-          })
-        },
-        'Delete Variable')
-      }, ['This will permanently delete the data from Workspace Data.'])
-    ])])
-  }
+      },
+      'Delete Variable')
+    }, ['This will permanently delete the data from Workspace Data.']),
+    busy && spinnerOverlay
```
panentheos, line 354: There are two spinners momentarily when loading this sub-page. One is due to refreshing the workspace, which we probably don't need to do now that this loads its own data.

src/pages/workspaces/workspace/Data.js
```diff
@@ -129,216 +129,229 @@ const getReferenceData = _.flow(
...
 
-    const download = withErrorReporting('Error downloading attributes', async () => {
-      const blob = await Ajax().Workspaces.workspace(namespace, name).exportAttributes()
-      FileSaver.saveAs(blob, `${name}-workspace-attributes.tsv`)
-    })
+  const download = _.flow(
+    withErrorReporting('Error downloading attributes'),
+    Utils.withBusyState(setBusy)
+  )(async () => {
+    const blob = await Ajax(signal).Workspaces.workspace(namespace, name).exportAttributes()
```
panentheos: Since this is a side-effectful call, it might be better not to make it cancellable.

src/pages/workspaces/workspace/Data.js
```diff
@@ -129,216 +129,229 @@ const getReferenceData = _.flow(
...
-    super(props)
-    this.state = {
-      editIndex: undefined,
-      deleteIndex: undefined,
-      editKey: undefined,
-      editValue: undefined,
-      editType: undefined,
-      textFilter: ''
-    }
+const LocalVariablesContent = ({ workspace, workspace: { workspace: { namespace, name } }, firstRender }) => {
```
panentheos: Clicking on 'Workspace Data' again while already on that sub-page no longer refreshes the workspace attributes. If we want to keep that behavior, we could pass `refreshKey` and trigger a reload when it changes. Also, we should stop passing `refreshWorkspace` since this is no longer used.

panentheos (commented):
TOL: The refactor on the data page ended up being much more complex than the original story, and could have been its own PR.

panentheos (approved):
Looks good, tested by clicking through the affected pages locally.

PR #1994, opened 2020-01-13
notify when an rstudio cluster is ready [SATURN-1262]

zarsky-broad:
Tested in browser.


breilly2 (commented):
This will only show the notification if the notebook runtime widget is being rendered, so not if you're somewhere other than a workspace (in the right billing project). This might be okay... should check with Adrian. Otherwise, the notification will need to be made more global, like the PFB import notification.

For the current scope, code looks good. I tested and it works, confirming that the notification shows up on any window where the notebook runtime widget is rendered (and confirming that the notification does not show up when viewing, e.g, the workspace list page). If that fits the requirements, let me know and I'll approve this PR.

zarsky-broad:
@breilly2 Adrian's ok with that.

breilly2 (approved):
:+1: Looks good.

PR #1995, opened 2020-01-13
prevent trying to select data of a non-existent type [SATURN-1224]

zarsky-broad:
Tested in browser. Crashes before, not after.


breilly2 (approved):
👍 Looks good. Not so much that it doesn't crash, but it prevents the click that causes the crash.

PR #1996, opened 2020-01-13
close menu when clicking notebook preview [SATURN-1227]

zarsky-broad:
Tested & working with this fix in browser.


breilly2 (approved):
👍 Nice solution! I tested this and it works. Also tested that it does not prevent links in the notebook preview from working.

Observation: This menu does not block clicks on clickable things. In a desktop app, I would expect a click that dismisses the menu to _not_ activate something clickable underneath. We currently do activate the thing underneath as well as dismissing the menu. The behavior seems consistent whether or not the click is in Terra or the notebook preview iframe.

ehanna4:
*TOL : I'm noticing when tabbing around the page I am unable to close the menu now unless I click somewhere with the mouse which is different than the behavior of 3-dot menus on some other pages on Terra like the workspace card list which allows closing of the menu using the keyboard. However other places on Terra have this same thing that you must use the mouse rather than keyboard to close out 3-dot menus so I'm not sure what is expected

*with further testing it looks like that is not a new behavior as of these changes. so TOL but should 3-dot menus across Terra all have same behavior and also allow closing from keyboard?

PR #1997, opened 2020-01-14
change data import flow to multi-step [SATURN-1316]

panentheos:
This introduces a wizard-style flow when importing data, in preparation for a third option to clone a template workspace. Adds the piece of the new left-hand text that's relevant to the existing choices.

Tested by clicking through the flows, and trying various edge cases.

### DIFF
--- a/src/pages/ImportData.js
+++ b/src/pages/ImportData.js
@@ -55,6 +55,7 @@ const ImportData = () => {
   const [isImporting, setIsImporting] = useState(false)
   const { query: { url, format, ad, wid } } = Nav.useRoute()
   const [mode, setMode] = useState(wid ? 'existing' : undefined)
+  const [isCreateOpen, setIsCreateOpen] = useState(false)
   const [selectedWorkspaceId, setSelectedWorkspaceId] = useState(wid)
 
   const selectedWorkspace = _.find({ workspace: { workspaceId: selectedWorkspaceId } }, workspaces)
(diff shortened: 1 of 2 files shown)


src/pages/ImportData.js
```diff
@@ -63,15 +90,73 @@ const ImportData = () => {
...
+                  disabled: !selectedWorkspace,
+                  onClick: () => onImport(selectedWorkspace.workspace)
+                }, ['Import']),
+                h(ButtonSecondary, { style: { marginLeft: '1rem' }, onClick: () => setMode() }, ['Back'])
+              ])
+            ])
+          }],
+          ['new', () => {
+            return h(Fragment, [
+              div({ style: styles.title }, ['Start with a new workspace']),
```
breilly2: Nit: I find it a little jarring/distracting for this pane to change at the same time as opening a modal. Part of it is that the modal partially obscures the new content so I feel like I missed something.

breilly2 (approved):
👍 Looks good! This turned out to be more simple than I expected (because of the existing componentization). I tested importing into both an existing workspace and a new workspace. I also accidentally tested the scenario where workspace creation succeeds but import fails. I got an error notification and the view changed to "Start with an existing workspace" with the newly created workspace selected. This was slightly jarring at first but makes perfect sense because of the state the system is left in.

src/pages/ImportData.js
```diff
@@ -63,15 +91,66 @@ const ImportData = () => {
...
+              h(ChoiceButton, {
+                style: { marginTop: '1rem' },
+                onClick: () => setMode('existing'),
+                iconName: 'folder-open',
+                title: 'Start with an existing workspace',
+                detail: 'Select one of your workspaces'
+              }),
+              h(ChoiceButton, {
+                style: { marginTop: '1rem' },
+                onClick: () => setIsCreateOpen(true),
```
breilly2, line 135: I think you could get away with folding this state into `mode`, more like it was previously. However, that would also require changing the `switchCase` to `cond`, and I can see that having an explicit `isCreateOpen` flag makes the state more closely mirror the visible UI state.

breilly2 (approved):
Still 👍. Thanks for the update!

I did notice one more small nit. After creating a new workspace, between the modal closing and being redirected to the new workspace, probably during the actual import, the view shows the "choose existing" interface. Not critical to fix.

PR #2000, opened 2020-01-14
[SATURN-1260] Adding Project-Specific view to cluster config drawer

ehanna4:
DO NOT MERGE UNTIL APPROVED BY PO

This ticket holds all the renaming of "Notebook Runtime" to "Application Compute" and will expose the RStudio Project-specific feature to users. Need to coordinate with PO before merging. Waiting on link to zendesk page to be made public.

Tested by:
- creating, deleting and replacing a variety of clusters from the Jupyter, Custom and Project-specific environments. 
- observed that you are able to paste the same image urls in either Custom or Project-specific environment (and validation works the same) but when you reopen the drawer it will open to the environment you have a current cluster on.
- observed the rstudio logo appears in the widget when using an appropriate image


src/components/NewClusterModal.js
```diff
@@ -115,7 +118,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     )
     this.state = {
       profile: matchingProfile ? matchingProfile.name : 'custom',
-      jupyterUserScriptUri: '', customEnvImage: '', viewMode: undefined, isDeleteView: false,
+      jupyterUserScriptUri: '', customEnvImage: '', projSpecImage: '', viewMode: undefined, isDeleteView: false,
```
panentheos: TOL: Do we want a separate piece of state for the project specific image?

It's functionally identical to the custom env image state, in terms of how it's ultimately used. Having it separate means that you can switch between custom and project-specific, and retain a separate temporary value for each. If that isn't a particularly important part of the UX, we could just use the same state value for both, which would eliminate some of the conditional logic around which value to use.

src/components/NewClusterModal.js
```diff
@@ -230,10 +244,15 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-              tooltip: isSelectedImageCustom && isCustomImageInvalid &&
-                'Enter a valid docker image to use',
-              onClick: () => isSelectedImageCustom ?
+              disabled: Utils.switchCase(selectedLeoImage,
+                [CUSTOM_MODE, () => { return this.isInputtedImageInvalid(customEnvImage) }],
+                [PROJECT_SPECIFIC_MODE, () => { return this.isInputtedImageInvalid(projSpecImage) }]
+              ),
+              tooltip: Utils.switchCase(selectedLeoImage,
+                [CUSTOM_MODE, () => { if (this.isInputtedImageInvalid(customEnvImage)) return 'Enter a valid docker image to use' }],
+                [PROJECT_SPECIFIC_MODE, () => { if (this.isInputtedImageInvalid(projSpecImage)) return 'Enter a valid rstudio image to use' }]
```
panentheos: This doesn't have to be an RStudio image, right? Should this just have the same message as the other one?

src/components/NewClusterModal.js
```diff
@@ -442,16 +465,40 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+              }],
+              [PROJECT_SPECIFIC_MODE, () => {
+                return h(Fragment, [
+                  div({ style: { lineHeight: '1.5' } }, [
+                    'Some consortium projects, such as ',
+                    h(Link, { href: rstudioBaseImages, ...Utils.newTabLinkProps }, ['AnVIL']),
+                    ', have created environments that are specific to their project. If you want to use one of these: ',
+                    div({ style: { marginTop: '0.5rem' } }, ['1. Find the environment image (',
+                      h(Link, { href: zendeskImagePage, ...Utils.newTabLinkProps }, ['view image list']),
+                      ') ']),
```
panentheos: nit: this code spacing is confusing when coupled with this particular text. I think expanding the children array fully would help, e.g.:
```
div([
  X,
  X,
  X
])
```

src/components/NewClusterModal.js
```diff
@@ -442,16 +465,40 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
                     ' be based off one of the ',
-                    h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra base images.'])
+                    h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images']),
+                    ' or a ',
+                    h(Link, { href: zendeskImagePage, ...Utils.newTabLinkProps }, ['Project-Specific image'])
+                  ])
+                ])
+              }],
+              [PROJECT_SPECIFIC_MODE, () => {
+                return h(Fragment, [
```
panentheos: nit: this `Fragment` is unnecessary since it only contains a single child.

src/components/NewClusterModal.js
ehanna4: I struggled with this decision and went back and forth about just using the customEnvImage. I ultimately went with two pieces of state so that the custom and project-specific environments would be more decoupled as it seemed to be emphasized in the requirement that this BYORstudio thing should be clearly separated from existing notebook stuff, but I am certainly open to changing it

src/components/NewClusterModal.js
ehanna4: Good catch yes I will fix that

src/components/NewClusterModal.js
ehanna4: Upon further reflection and discussion with Isaac I will remove the extra state and just use customEnvImage. Thanks for the call out!

ehanna4:
UPDATE: will be reverting to "Runtime" again while we wait on final-final approval of title change

src/components/common.js
```diff
@@ -309,11 +309,12 @@ const formatGroupLabel = group => (
     }
   }, [group.label]))
 
-const BaseSelect = ({ value, newOptions, id, findValue, ...props }) => {
+const BaseSelect = ({ value, newOptions, id, findValue, maxHeight, ...props }) => {
   const newValue = props.isMulti ? _.map(findValue, value) : findValue(value)
 
   return h(RSelect, _.merge({
     inputId: id,
+    maxMenuHeight: maxHeight || undefined,
```
panentheos: Since we already pass all additional props through to the underlying library, we could remove this special handling and just have the consumer pass `maxMenuHeight`, which would have the same effect.

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -68,7 +68,7 @@ const AppLauncher = _.flow(
             ['Stopping', () => 'Notebook runtime is stopping. This takes ~4 minutes.'],
             ['Stopped', () => 'Notebook runtime is stopped. Start it or reload the page to continue.'],
             ['Error', () => 'Error with the notebook runtime, please create a new one to continue.'],
-            [null, () => 'Create a notebook runtime to continue.'],
+            [null, () => 'Create an notebook runtime to continue.'],
```
panentheos: nit: grammar

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -85,7 +85,7 @@ const NotebookCard = ({ namespace, name, updated, metadata, listView, wsName, on
     content: h(Fragment, [
       h(MenuButton, {
         href: notebookLink,
-        tooltip: canWrite && 'Open without runtime',
+        tooltip: canWrite && 'Open without compute instance',
```
panentheos: Is this change intended?

src/components/NewClusterModal.js
```diff
@@ -216,12 +225,12 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
     const bottomButtons = () => h(Fragment, [
       div(
-        { style: { display: 'flex', marginTop: '3rem', marginBottom: '1rem' } },
+        { style: { display: 'flex', marginTop: '2.5rem', marginBottom: '1rem' } },
         [
           currentCluster && div([
             h(ButtonSecondary, {
               onClick: () => this.setState({ isDeleteView: true })
-            }, 'Delete Runtime')
+            }, 'Delete Compute')
```
panentheos: Is this change intended?

src/components/NewClusterModal.js
```diff
@@ -197,16 +199,23 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+    const isCustomImageInvalid = customEnvImage && !imageValidationRegexp.test(customEnvImage)
+    const isSelectedImageInputted = selectedLeoImage === CUSTOM_MODE || selectedLeoImage === PROJECT_SPECIFIC_MODE
+
     const makeGroupedEnvSelect = id => h(GroupedSelect, {
       id,
+      maxHeight: '25rem',
       value: selectedLeoImage,
       onChange: ({ value }) => {
-        this.setState({ selectedLeoImage: value })
+        this.setState({ selectedLeoImage: value, customEnvImage: '' })
```
panentheos, line 210: This means that if you switch to a prebuilt image and then back to custom, it loses any url you entered. Is that what we want?

panentheos (approved):
Looks good overall, see a few questions.

src/components/NewClusterModal.js
ehanna4: This was intended. While the functionality of custom env and pre-built are super similar there should appear to be a difference between them to users. This way the state is reduced but its still made clear there is a difference between these two options.

PR #2002, opened 2020-01-15
Add clinical user field to contact form [SATURN-1266]

cameron-ardell:
General premise:
There is no way to understand the type of research a user does when they contact us. This creates an issue for responses to clinical users, which will need a more expedited response.

Requirements:
When a user engages with the ZenDesk contact form there should be a field allowing them to input that they are a clinical user. This field should be presented to frontline support in Zendesk so they know that they have received an issue from a clinical user

[Link to ticket](https://broadworkbench.atlassian.net/browse/SATURN-1266?atlOrigin=eyJpIjoiNzA2MzlmZmU5NDJkNGRkN2FhNzlkZGRmOGU4NGRlMDgiLCJwIjoiaiJ9)


src/components/SupportRequest.js
```diff
@@ -208,6 +211,19 @@ const SupportRequest = _.flow(
             onChange: v => this.setState({ email: v })
           })
         ])]),
+        h(IdContainer, [id => h(Fragment, [
```
panentheos: For an `IdContainer` to work, the `id` needs to be attached to a label and also a target element. Radio buttons have their own built in labels though, so I don't think we need another one in this case.

src/components/SupportRequest.js
```diff
@@ -208,6 +211,19 @@ const SupportRequest = _.flow(
             onChange: v => this.setState({ email: v })
           })
         ])]),
+        h(Fragment, [
```
panentheos: nit: This `Fragment` is unnecessary since we're not doing any computation (e.g. conditionals) on this chunk of content. We could just inline the children directly.

panentheos (approved):
Looks good.

Tested by clicking through the form locally (but did not actually submit a ticket)

PR #2003, opened 2020-01-15
Use built-in react-scripts/storybook features [SATURN-1333]

zarsky-broad:
It's been a while since we brought in react-scripts and storybook, and since then both have simplified how they accept tweaks to their configs, and have built in a number of features that we had previously hacked on ourselves. This uses those built-in features and removes the custom additions.

See:
  * We're still using the rewire to add the custom eslintrc because the (still experimental) [built-in option](https://create-react-app.dev/docs/setting-up-your-editor#experimental-extending-the-eslint-config) wouldn't work for me
  * https://create-react-app.dev/docs/analyzing-the-bundle-size
  * https://create-react-app.dev/docs/adding-css-reset
  * https://create-react-app.dev/docs/adding-images-fonts-and-files#adding-svgs
  * https://create-react-app.dev/docs/importing-a-component/#absolute-imports
    * https://code.visualstudio.com/docs/languages/jsconfig
  * https://create-react-app.dev/docs/adding-custom-environment-variables
    * We're not [using a .env file](https://create-react-app.dev/docs/adding-custom-environment-variables#adding-development-environment-variables-in-env) because it doesn't evaluate code, at most it can just pass in external variables that are already defined
  * https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-52x-to-53x


src/components/FooterWrapper.js
```diff
@@ -11,7 +11,7 @@ const styles = {
   item: { marginLeft: '2rem' }
 }
 
-const buildTimestamp = new Date(SATURN_BUILD_TIMESTAMP)
+const buildTimestamp = new Date(parseInt(process.env.REACT_APP_BUILD_TIMESTAMP))
```
panentheos: `parseInt` needs an explicit base parameter

src/components/TopBar.js
```diff
@@ -305,10 +305,10 @@ const TopBar = Utils.connectStore(authStore, 'authState')(class TopBar extends C
             div({ style: { color: colors.dark(0.3), fontSize: 10, fontWeight: 600, marginTop: '0.5rem' } }, [
               'Built on: ',
               h(Clickable, {
-                href: `https://github.com/DataBiosphere/terra-ui/commits/${SATURN_VERSION}`,
+                href: `https://github.com/DataBiosphere/terra-ui/commits/${process.env.REACT_APP_VERSION}`,
                 ...Utils.newTabLinkProps,
                 style: { textDecoration: 'underline', marginLeft: '0.25rem' }
-              }, [new Date(SATURN_BUILD_TIMESTAMP).toLocaleString()])
+              }, [new Date(parseInt(process.env.REACT_APP_BUILD_TIMESTAMP)).toLocaleString()])
```
panentheos: `parseInt` here too

panentheos (approved):
Looks good, see comments.

Tested by running the build and storybook locally, and clicking around.

PR #2004, opened 2020-01-15
Some tweaks to rstudio experience [SATURN-1332]

zarsky-broad:
Made with Adrian, verified in-browser.

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -148,16 +148,20 @@ export default class ClusterManager extends PureComponent {
     const welderCutOff = new Date('2019-08-01')
     const createdDate = new Date(cluster.createdDate)
     const dateNotified = getDynamic(sessionStorage, `notifiedOutdatedCluster${cluster.id}`) || {}
+    const rStudioLaunchLink = Nav.getLink('workspace-app-launch', { namespace, name, app: 'RStudio' })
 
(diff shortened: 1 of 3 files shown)


src/components/ClusterManager.js
```diff
@@ -148,16 +148,20 @@ export default class ClusterManager extends PureComponent {
...
     if (cluster.status === 'Error' && prevCluster.status !== 'Error' && !_.includes(cluster.id, errorNotifiedClusters.get())) {
       notify('error', 'Error Creating Notebook Runtime', {
         message: h(ClusterErrorNotification, { cluster })
       })
       errorNotifiedClusters.update(Utils.append(cluster.id))
-    } else if (cluster.status === 'Running' && prevCluster.status && prevCluster.status !== 'Running' && cluster.labels.tool === 'RStudio') {
-      const rStudioNotificationId = notify('info', 'Your compute instance is ready.', {
+    } else if (
+      cluster.status === 'Running' && prevCluster.status && prevCluster.status !== 'Running' &&
+      cluster.labels.tool === 'RStudio' && window.location.hash !== rStudioLaunchLink
```
panentheos, line 160: TOL: At some point we should probably come up with a mechanism for matching URL paths based on data structures, rather than strings, since this could theoretically get tripped up by e.g. a trailing question mark.

panentheos (approved):
Looks good, tested by clicking locally.

PR #2005, opened 2020-01-17
Add group to handle 15 minute sign out [SATURN-1328]

petesantos:
Updated IdleStatusMonitor to check to see if the user belongs to the session_timeout group to determine if they are a clinical user.

The group name may change.


src/components/IdleStatusMonitor.js
```diff
@@ -33,19 +34,25 @@ const getIdleData = ({ currentTime, lastRecordedActivity, timeout, countdownStar
...
-  const isClinicalDomain = email && emailDomain && email.endsWith(`@${emailDomain}`)
+  const signal = Utils.useCancellation()
+  const [isClinicalUser, setIsClinicalUser] = useState()
+  const { isSignedIn, user: { id } } = Utils.useStore(authStore)
   const { [id]: lastRecordedActivity } = Utils.useStore(lastActiveTimeStore)
   const { timedOut } = getIdleData({ currentTime: Date.now(), lastRecordedActivity, timeout, countdownStart })
 
+  useEffect(() => {
+    const setClinicalStatus = async () => {
+      setIsClinicalUser(!!_.find({ groupName: 'session_timeout' }, await Ajax(signal).Groups.list()))
```
zarsky-broad: `_.some`?

src/components/IdleStatusMonitor.js
```diff
@@ -33,19 +35,25 @@ const getIdleData = ({ currentTime, lastRecordedActivity, timeout, countdownStar
...
+  const [isClinicalUser, setIsClinicalUser] = useState()
+  const { isSignedIn, registrationStatus, user: { id } } = Utils.useStore(authStore)
   const { [id]: lastRecordedActivity } = Utils.useStore(lastActiveTimeStore)
   const { timedOut } = getIdleData({ currentTime: Date.now(), lastRecordedActivity, timeout, countdownStart })
 
+  useEffect(() => {
+    const setClinicalStatus = withErrorReporting('Error loading group list', async () => {
+      setIsClinicalUser(_.some({ groupName: 'session_timeout' }, await Ajax(signal).Groups.list()))
+    })
+    isSignedIn && registrationStatus === 'registered' ? setClinicalStatus() : setIsClinicalUser()
```
panentheos: TOL: I think this logic will work, but we should get in the habit of being rigorous about what we're checking for in effects, because it's easy to get tripped up. For example, suppose that `signal` changed value (it won't, but just to illustrate), the effect would re-run, this check would succeed and we'd re-request the groups even though the person hadn't logged out.

What we actually want to detect is when `registrationStatus` transitions from `false` to `true`. To do that explicitly, we'd need to examine both the previous and current values. See the cascade of calls that happen in `auth.js` for some examples of that. Actually, thinking about it, I wonder if it would make sense for this call to happen in that file.

breilly2: +1 to moving the check to `auth.js`. It means we have to make `clinicalStatus` global, but that could be useful in other places like the Contact Us widget. We'd also have to remember to clear the global state on sign-out, which is also centralized in `auth.js`.

petesantos: I moved this to auth.js - it does look & feel much cleaner - thanks!

src/libs/auth.js
```diff
@@ -153,6 +154,14 @@ authStore.subscribe(withErrorReporting('Error checking TOS', async (state, oldSt
   }
 }))
 
+authStore.subscribe(withErrorReporting('Error checking groups for timeout status', async (state, oldState) => {
+  if (oldState.registrationStatus !== state.registrationStatus) {
```
panentheos: In `initializeAuth`, there's a pattern for resetting dependent data during signout, which reduces the number of state transitions (and so reduces the possibility of edge cases). If we handled the reset there, then this code path could just check for the exact transition we care about, e.g. `oldState.registrationStatus !== 'registered' && state.registrationStatus === 'registered'`

panentheos (approved):
Looks good! Tested locally by logging in and out with different users.

PR #2006, opened 2020-01-17
Better cluster replacement behavior [SATURN-1296]

zarsky-broad:
Verified all working in browser.

(Much better viewed with whitespace hidden)


src/components/NewClusterModal.js
```diff
@@ -348,136 +344,132 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
         ]),
-        div({ style: { backgroundColor: colors.dark(0.2), borderRadius: '100px', width: 'fit-content', padding: '0.75rem 1.25rem', ...styles.row } }, [
+        div({
+          style: { backgroundColor: colors.dark(0.2), borderRadius: '100px', width: 'fit-content', padding: '0.75rem 1.25rem', ...styles.row }
+        }, [
           span({ style: { ...styles.label, marginRight: '0.25rem' } }, ['COST:']),
           `${Utils.formatUSD(machineConfigCost(this.getMachineConfig()))} per hour`
         ])
-      ]),
-      !!currentCluster && div({ style: styles.warningBox }, [
```
panentheos, line 365: nit: we can remove `styles.warningBox` above since it's now unused

src/components/NewClusterModal.js
```diff
@@ -348,136 +344,132 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-              onClick: () => onSuccess(this.deleteCluster())
-            }, ['DELETE'])
+            h(ButtonSecondary, { style: { marginRight: '2rem' }, onClick: () => this.setState({ viewMode: undefined }) }, ['CANCEL']),
+            h(ButtonPrimary, { onClick: () => onSuccess(this.deleteCluster()) }, ['DELETE'])
           ])
         ])
       })],
-      () => ({
+      ['replace', () => ({
+        onPrevious: () => this.setState({ viewMode: isSelectedImageCustom ? 'warning' : undefined }),
```
panentheos: Clicking through this in the double-confirm case, it feels a little surprising to have to click 'back' twice to cancel out completely. Presumably if someone changes their mind, it's because they want to go back and change something on the original form.

panentheos (approved):
Looks good, see minor comments.

Tested locally by clicking through various cases.

src/components/NewClusterModal.js
breilly2: I thought the behavior was consistent with the notion of "back", which was also consistent with the behavior of the ⬅️. The current behavior feels more like "cancel" to me. However, if we changed the button to "cancel", I'm not sure what to do with the ⬅️.

I don't feel strongly enough to hold up the PR, but strongly enough to share my thoughts while it's being discussed.

breilly2 (approved):
👍 Looks good. I tested several scenarios, including creating custom runtimes, replacing runtimes, and deleting runtimes.

breilly2:
Please fill in Suggested Release Notes in the ticket so that Customer Success is aware of the change in behavior. Additionally, it would be good to give them a heads-up before deploying this change so they're not surprised by it.

PR #2007, opened 2020-01-17
disable unusable cluster configs [SATURN-1337]

zarsky-broad:
Verified that unusable configs are not selectable in browser


panentheos (approved):
Looks good, tested locally.

PR #2008, opened 2020-01-21
Check illegal names when selecting data [SATURN-1209]

cameron-ardell:
In the "Select Data" box, also titled "Select samples to process" located on the method page it allows you to change the name of the set you create in, "Selected rows will be saved as a new set named". If you give it an illegal name, it does not throw an error.  The error comes up later if the user submits the job. We want to fix this so the user sees the error, is able to appropriately correct, and cannot submit until they give a legal name.

Ticket [here](https://broadworkbench.atlassian.net/browse/SATURN-1209?atlOrigin=eyJpIjoiNjNiNzI4NjA1OWY3NDEzNWE3M2Q0OTZjYjA5ZjNiYzYiLCJwIjoiaiJ9)

Tested with @breilly2 by running through the text errors, along with making sure the new restrictions work for the error notification icon regardless of style parameters passed in (or lack thereof)


src/pages/workspaces/workspace/workflows/DataStepContent.js
```diff
@@ -133,17 +145,16 @@ export default class DataStepContent extends Component {
...
-            id,
-            style: { width: 500, marginLeft: '0.25rem' },
-            value: newSetName,
-            onChange: v => this.setNewSelectionModel({ newSetName: v })
-          })
-        ])])
+          ((isChooseRows || isProcessMergedSet) && _.size(selectedEntities) > 1)) && h(IdContainer,
+          [id => div({ style: { marginTop: '1rem' } }, [
+            h(FormLabel, { htmlFor: id }, [`Selected rows will ${isProcessMergedSet ? 'have their membership combined into' : 'be saved as'} a new set named:`]),
+            h(ValidatedInput, {
```
panentheos, line 152: If we're using a `ValidatedInput` we should pass the error so the user gets some feedback about what's wrong.

src/components/input.js
```diff
@@ -174,12 +174,10 @@ export const NumberInput = ({ onChange, onBlur, min = -Infinity, max = Infinity,
...
  * @param {object} [props.error] - error message content
  */
-export const ValidatedInput = props => {
-  const { inputProps, error } = props
-
+export const ValidatedInput = ({ inputProps, error }) => {
   return h(Fragment, [
     div({
-      style: { position: 'relative', display: 'flex', alignItems: 'center' }
+      style: { position: 'relative', display: 'flex', alignItems: 'center', width: inputProps?.style?.width }
```
panentheos: TOL: This works when setting a `width`, but wouldn't work when using other styles that affect the size, e.g. `maxWidth`. Another way to handle this would be to have the consumer create a width-constrained wrapper around the input. That should work because our inputs take up 100% of the available space by default.

breilly2: That occurred to me, but I also felt that `ValidatedInput` should render consistently and correctly regardless of container. Another thought: instead of reusing `inputProps.style.width`, we could add a `width` prop to `ValidatedInput` that gets applied to the wrapping container already exists in `ValidatedInput`. Thoughts?

panentheos: That seems a little better and more explicit than hooking onto the style, yes.

cameron-ardell: I think I added this feedback correctly, would one of you mind checking?

breilly2 (approved):
👍 Looks good! I tested with several invalid set names, including empty.

PR #2010, opened 2020-01-24
fix filtering workspace by tags [SATURN-1348]

panentheos:
This adds the `tag:tags` attribute to the list of workspace fields we request from the server, since it's needed for tag filtering.

Also simplifies the filtering code slightly.

Tested by using the filter locally.


zarsky-broad (approved):
Tested in browser.

PR #2011, opened 2020-01-24
break long option text in dropdowns [SATURN-1349]

panentheos:
See ticket.

Tested locally by interacting with the tag selector, and others.


zarsky-broad (approved):
Verified in browser

PR #2012, opened 2020-01-24
Tweak login for BDC [SATURN-1303]

zarsky-broad:
Verified in-browser, and with product and UX.


panentheos (approved):
Looks good!

Tested login flow locally on Terra, DataStage, and FireCloud versions.

PR #2013, opened 2020-01-27
Template Workspace when importing data [SATURN-1149]

ehanna4:
Add ability for user to import data directly to template workspace.

To Test:
Follow the flow of importing data and once you get to the screen with 'Destination Workspace' add query params to the url and refresh. (To see the json file the template workspace list is coming from the firecloud-alerts bucket in dev. There are currently two lists in the file with the keys 'Terra' and 'bdC'.)

i.e. add: `&template=Terra&template=bdC` to the url

EDIT: Any workspace that shows in the list now has been validated and should be able to be imported! 
*̶N̶O̶T̶E̶*̶ ̶O̶n̶l̶y̶ ̶t̶h̶e̶ ̶`̶j̶i̶m̶m̶y̶ ̶r̶e̶a̶d̶s̶`̶ ̶t̶e̶m̶p̶l̶a̶t̶e̶ ̶w̶o̶r̶k̶s̶p̶a̶c̶e̶ ̶w̶i̶l̶l̶ ̶w̶o̶r̶k̶ ̶c̶u̶r̶r̶e̶n̶t̶l̶y̶ ̶a̶s̶ ̶t̶h̶e̶ ̶o̶t̶h̶e̶r̶ ̶w̶o̶r̶k̶s̶p̶a̶c̶e̶s̶ ̶d̶i̶d̶ ̶n̶o̶t̶ ̶h̶a̶v̶e̶ ̶t̶h̶e̶ ̶r̶i̶g̶h̶t̶ ̶p̶e̶r̶m̶i̶s̶s̶i̶o̶n̶s̶.̶

Tested:
- With query params of 'Terra', 'bdC' and 'fakeNotReal' in different combinations
- Tried all the flows successfully: making a new workspace, starting with existing and starting with a template
- Successfully started with a template on Terra and DataStage


src/components/NewWorkspaceModal.js
```diff
@@ -191,6 +202,7 @@ export default _.flow(
             options: _.difference(_.uniq(_.map('groupName', allGroups)), existingGroups).sort()
           })
         ])]),
+        div({ style: { marginTop: '1rem', lineHeight: '1.5rem' } }, [customMessage]),
```
panentheos: nit: We should make this conditional on the content so we only add the margin if it's there.

src/components/NewWorkspaceModal.js
```diff
@@ -121,13 +124,21 @@ export default _.flow(
     return Utils.cond(
       [loading, spinnerOverlay],
       [hasBillingProjects, () => h(Modal, {
-        title: cloneWorkspace ? 'Clone a Workspace' : 'Create a New Workspace',
+        title: Utils.cond(
+          [cloneWorkspace && customCloneTitle, customCloneTitle],
```
panentheos: TOL: Granted there's only the one use case for these overrides right now, but since we already know we're in clone mode, I wonder if it would be simpler to just make them unconditional, e.g. just `title`, which takes precedence if it's there.

src/pages/ImportData.js
```diff
@@ -30,36 +33,84 @@ const styles = {
   }
 }
 
-const ChoiceButton = ({ iconName, title, detail, style, ...props }) => {
+const ChoiceButton = ({ onClick, iconName, title, detail, tooltip }) => {
```
panentheos: The `...props` will forward all additional props implicitly, so if we keep that, we can still pass `onClick` and `tooltip` without having to mention them. Also, nit: even though we're not currently using any style overrides, it might be nice to leave the option there since it's something we support on almost all of our shared components.

src/pages/ImportData.js
```diff
@@ -30,36 +33,84 @@ const styles = {
...
 
   const selectedWorkspace = _.find({ workspace: { workspaceId: selectedWorkspaceId } }, workspaces)
 
+  const getTemplates = fileContents => {
+    let temporaryList = []
+
+    !(_.isArray(template)) ? temporaryList = fileContents[template] :
+      _.forEach(currentTemplate => {
+        fileContents[currentTemplate] &&
+        temporaryList.push(fileContents[currentTemplate])
```
panentheos: There's a way to do this computation using lodash helpers and eliminate the mutation. Let's pair on it.

src/pages/ImportData.js
```diff
@@ -30,36 +33,84 @@ const styles = {
...
+      }, template)
+
+    return !temporaryList ? null : _.flatten(temporaryList)
+  }
+
+  Utils.useOnMount(() => {
+    const loadTemplateWorkspaces = async () => {
+      try {
+        const allTemplates = await fetch(`${getConfig().firecloudBucketRoot}/template-workspaces.json`).then(res => res.json())
+        setTemplateWorkspaces(getTemplates(allTemplates))
```
panentheos: This will work as written, but since `getTemplates` is such a straightforward computation, we could also choose to delay the work until render, and instead store the unmodified template list. That way, we avoid storing derived values in state, which reduces the risk of edge cases popping up.

src/pages/ImportData.js
```diff
@@ -30,36 +33,84 @@ const styles = {
...
+      _.forEach(currentTemplate => {
+        fileContents[currentTemplate] &&
+        temporaryList.push(fileContents[currentTemplate])
+      }, template)
+
+    return !temporaryList ? null : _.flatten(temporaryList)
+  }
+
+  Utils.useOnMount(() => {
+    const loadTemplateWorkspaces = async () => {
```
panentheos: This needs an error handler wrapper in case the fetch fails.

src/pages/ImportData.js
```diff
@@ -30,36 +33,84 @@ const styles = {
...
+        setTemplateWorkspaces(null)
+      }
+    }
+
+    loadTemplateWorkspaces()
+  })
+
+  const setSelectionInList = ws => {
+    setRadioButtonSelected(ws)
+    setSelectedTemplateWorkspace(ws)
```
panentheos: We can simplify the state handling here in a couple ways. This component already has the full list of workspaces from the `useWorkspaces` call above. If we use that, we can intersect that with the json list, so we only show workspaces that actually exist. We could also just store the selected id, and use that to look up the object, which minimizes our state. Also we probably don't need to make a separate details call, since we already have the necessary info.

src/pages/ImportData.js
```diff
@@ -151,7 +259,21 @@ const ImportData = () => {
...
         ),
-        isImporting && spinnerOverlay
+        isCloneOpen && h(NewWorkspaceModal, {
+          cloneWorkspace: selectedTemplateWorkspace,
+          customCloneTitle: `Clone ${selectedTemplateWorkspace.workspace.name} and Import Data`,
+          customCloneButtonText: 'Clone and Import',
+          customMessage: noteMessage,
+          onDismiss: () => setIsCloneOpen(false),
+          onSuccess: w => {
+            setMode('template')
```
panentheos: This should set the mode to 'existing', because the clone has succeeded at this point. The idea is that if the import step fails due to a network blip, we'll leave the UI in a consistent state, and they could just click 'Import' to try again with the workspace they just created.

src/pages/ImportData.js
```diff
@@ -34,32 +37,53 @@ const ChoiceButton = ({ iconName, title, detail, style, ...props }) => {
...
+
+  const noteMessage = 'Note that the import process may take some time after you are redirected into your destination workspace.'
 
   const selectedWorkspace = _.find({ workspace: { workspaceId: selectedWorkspaceId } }, workspaces)
 
+  const filteredTemplates =
+    _.flow(
+      _.flatMap(id => (allTemplates && allTemplates[id]) || []),
+      _.filter(({ name, namespace }) => _.some({ workspace: { namespace, name } }, workspaces))
+    )(_.isArray(template) ? template : Array.of(template))
```
panentheos: FYI: `_.castArray` is the shorthand we were looking for here!

src/pages/ImportData.js
```diff
@@ -34,32 +37,53 @@ const ChoiceButton = ({ iconName, title, detail, style, ...props }) => {
...
+
+  Utils.useOnMount(() => {
+    const loadTemplateWorkspaces = _.flow(
+      Utils.withBusyState(setIsImporting),
+      withErrorReporting('Import Error')
```
panentheos: This error message could be more descriptive, something like 'Error loading templates'

src/pages/ImportData.js
```diff
@@ -109,36 +134,92 @@ const ImportData = () => {
...
+                      text: h(Collapse, {
+                        buttonStyle: { color: colors.dark(), fontWeight: 600 },
+                        style: { fontSize: 14, marginLeft: '0.5rem' },
+                        title: h(Fragment, [
+                          name,
+                          hasNotebooks && img({ src: jupyterLogo, style: { height: 23, width: 23, marginLeft: '0.5rem' } }),
+                          hasWorkflows &&
+                            wdlIcon({ style: { height: 23, width: 23, marginLeft: '0.5rem', borderRadius: 3, padding: '8px 4px 7px 4px' } })
+                        ])
+                      }, [p({ style: { fontSize: 14, lineHeight: '1.5', marginRight: '1rem' } }, description)])
```
panentheos: nit: It shouldn't be a problem here, but we should bracket children for consistency.
```suggestion
                      }, [p({ style: { fontSize: 14, lineHeight: '1.5', marginRight: '1rem' } }, [description])])
```

src/pages/ImportData.js
```diff
@@ -109,36 +134,92 @@ const ImportData = () => {
...
+                    })
+                  ]
+                  )
+                }, Utils.toIndexPairs(filteredTemplates))
+              ]),
+              div({ style: { display: 'flex', alignItems: 'center', marginTop: '1rem' } }, [
+                h(ButtonSecondary, { style: { marginLeft: 'auto' }, onClick: () => setMode() }, ['Back']),
+                h(ButtonPrimary, {
+                  style: { marginLeft: '3rem' },
```
panentheos: nit: This margin is a bit wide, should probably be 2rem to match the other one

panentheos (approved):
Looks good, see a few small comments.

Tested locally with various combinations of templates.

PR #2014, opened 2020-01-27
fix back arrow on 'open with' modal [SATURN-1311]

panentheos:
See ticket. Tested by clicking locally.


ehanna4 (approved):
Looks good, tested in browser

PR #2016, opened 2020-01-28
Fix order of icon dict

ehanna4:
I noticed that the last 15ish entries to the icon-dict were not alphabetized like all the others, so this is just moving them into order with the rest of the entries.

TOL: Would be interested in others thoughts on updating the names in the dict to all follow the same format either snake-casing or camelCasing since currently its a mixture.


panentheos (approved):
Looks good, thanks!

PR #2017, opened 2020-01-28
Refactor integration tests [SATURN-1340]

breilly2:
There are two choices for how to organize the tests:

1. small individual shell files for each test, something like:
```
const { fn, name, timeout } = require('./register-user.integration-test')
test(name, fn({ context, page )), timeout)
```
2. one file with all tests

The consequence of #1 is obviously double the number of files. The consequences of #2 are:

- Need to use jest's `test.concurrent` to get them to run at the same time
- Selecting individual tests via CLI requires using `-t`
- CLI output appears to be running all tests even if a subset are selected
- CLI does lumps console output of all tests together
- CLI reports test execution time per-file so we wouldn't know execution time of individual tests

Given all of these consequences, I'm leaning toward #1 despite the extra files. Other thoughts?


integration-tests/tests/register-user.integration-test.js
```diff
@@ -3,7 +3,8 @@ const { withUser } = require('../utils/integration-helpers')
 const { findText, click, clickable, fillIn, input, signIntoTerra, waitForNoSpinners } = require('../utils/integration-utils')
 
 
-test('register user', withUser(async ({ token }) => {
+const testRegisterUserFn = withUser(async ({ page, token }) => {
+  page.setDefaultTimeout(7000)
```
breilly2: Unintentionally committed change. I will revert this line.

integration-tests/jest/jest-utils.js
```diff
@@ -0,0 +1,21 @@
+const { screenshotDir } = require('../utils/integration-config')
+
+
+const defaultTimeout = 5 * 60 * 1000
+
+const withGlobalJestPuppeteerContext = test => async () => await test({ context, page: await context.newPage() })
```
panentheos: nit: To avoid confusion with the global `test`, perhaps the argument should be named `fn`.

panentheos: This `async` and `await` look unnecessary. The test function will already return a promise.

panentheos: I think `jest-puppeteer` creates a default page and exposes a `page` global. Can we pass that along instead of creating another one?

integration-tests/jest/jest-utils.js
```diff
@@ -0,0 +1,21 @@
+const { screenshotDir } = require('../utils/integration-config')
+
+
+const defaultTimeout = 5 * 60 * 1000
```
panentheos: TOL: This timeout will need to apply in the web service as well, so it's not strictly a 'jest' thing. Not a big deal, but just wanted to call out a potential name confusion.

integration-tests/utils/integration-helpers.js
```diff
@@ -39,18 +39,13 @@ const deleteWorkspace = async workspaceName => {
   await ajaxPage.close()
 }
 
-const withWorkspace = test => async () => {
-  const workspaceName = await makeWorkspace()
+const withWorkspace = test => async ({ context, page }) => {
```
panentheos: We should capture and forward any additional options here, to make sure this remains fully composable.

integration-tests/jest/jest-utils.js
breilly2: Yes, good catch. This was left over from putting all of the tests in a single file and running them concurrently. Can also switch back from `test.concurrent` to just `test` below.

integration-tests/jest/jest-utils.js
breilly2: True. The mechanism for implementing the timeout (giving it to jest vs `Promise.race`) may be different but the meaning is the same.

I don't love the options currently available. I'm thinking `integration-config.js`, or maybe `integration-helpers.js`. Any preference?

panentheos: `integration-helpers.js` seems like a decent place to me

panentheos (approved):
Looks good.

PR #2018, opened 2020-01-28
Fix overflow of titles in modals [SATURN-1312]

ehanna4:
Stopped titles from overrunning edges of modals.

Observed this problem when editing in data table and copying a notebook. Tested in browser and no longer see these problems. Tested many other instances of modals to ensure they look fine still


petesantos (approved):
Looks good.
Tested by bringing up an edit modal for the bigquery table on dev and reproduced the issue. Did not see the issue with the new code in place

PR #2020, opened 2020-01-29
Make workspace selector case-insensitive [SATURN-1353]

ehanna4:
Update the workspace selector so that the alphabetization is case-insensitive. 

I.e. with this list: ['A', 'b', 'B', 'a']
previously it would have sorted as: 'A', 'B', 'a', 'b'
now it will sort as: 'A', 'a', 'b', 'B'

Tested by viewing in browser


src/components/workspace-utils.js
```diff
@@ -49,7 +49,7 @@ export const WorkspaceSelector = ({ workspaces, value, onChange, ...props }) =>
     value,
     onChange: ({ value }) => onChange(value),
     options: _.flow(
-      _.sortBy('workspace.name'),
+      _.sortBy([ws => { return (ws.workspace.name.toLowerCase()) }]),
```
panentheos: nit: with short, single expression functions, it's ok to use the brace-less shorthand. Also, the array isn't needed unless there are multiple sorts.
```suggestion
      _.sortBy(ws => ws.workspace.name.toLowerCase()),
```

panentheos (approved):
Looks good, see nit.

PR #2022, opened 2020-01-31
[IA-1508] Clusters can now be updated in some circumstances

jdcanas:
Heavy manual verification done.

Implements this ticket: https://broadworkbench.atlassian.net/browse/IA-1508, which:
- allows users to 'Update' their cluster via the leo patch endpoint
- Leverages leo functionality to stop/restart a cluster for updates that require the cluster to be stopped
- The Update status is used only in specific updates, but the user is allowed to continue editting their notebook during this time

There is one bug, where the stop transition does not actually work when pointing to dev, pending: https://github.com/DataBiosphere/leonardo/pull/1249

I still need to lint and there'e some logic (indicated in this PR) I want to noodle on a bit more, but I want to start getting feedback on the approach

**This cannot be merged until this PR is in prod: https://github.com/DataBiosphere/leonardo/pull/1271/files**

edit: that pr is in prod as of 2/10

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -140,7 +140,7 @@ export default class ClusterManager extends PureComponent {
   }
 
   componentDidUpdate(prevProps) {
-    const { namespace, name } = this.props
+    const { namespace, name, refreshClusters } = this.props
     const prevCluster = _.last(_.sortBy('createdDate', _.remove({ status: 'Deleting' }, prevProps.clusters))) || {}
     const cluster = this.getCurrentCluster() || {}
     const twoMonthsAgo = _.tap(d => d.setMonth(d.getMonth() - 2), new Date())
@@ -181,8 +181,10 @@ export default class ClusterManager extends PureComponent {
       notify('warn', 'Outdated Notebook Runtime', {
         message: 'Your notebook runtime is over two months old. Please consider deleting and recreating your runtime in order to access the latest features and security updates.'
       })
+    } else if (cluster.status === 'Running' && prevCluster.status === 'Updating') {
+      notify('success', 'Your runtime update has completed successfully.', { timeout: 3000 })
     }
-  }
+   }
 
   getActiveClustersOldestFirst() {
     const { clusters } = this.props
@@ -259,6 +261,7 @@ export default class ClusterManager extends PureComponent {
           })
         case 'Starting':
         case 'Stopping':
+        case 'Updating':
         case 'Creating':
           return h(ClusterIcon, {
             shape: 'sync',
@@ -288,7 +291,8 @@ export default class ClusterManager extends PureComponent {
     const totalCost = _.sum(_.map(clusterCost, clusters))
     const activeClusters = this.getActiveClustersOldestFirst()
     const creating = _.some({ status: 'Creating' }, activeClusters)
-    const isDisabled = !canCompute || creating || busy
+    const updating = _.some({ status: 'Updating' }, activeClusters)
+    const isDisabled = !canCompute || creating || busy || updating
 
     const isRStudioImage = currentCluster?.labels.tool === 'RStudio'
     const appName = isRStudioImage ? 'RStudio' : 'terminal'
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -140,7 +140,7 @@ export default class ClusterManager extends PureComponent {
   }
 
   componentDidUpdate(prevProps) {
-    const { namespace, name, refreshClusters } = this.props
+    const { namespace, name } = this.props
     const prevCluster = _.last(_.sortBy('createdDate', _.remove({ status: 'Deleting' }, prevProps.clusters))) || {}
     const cluster = this.getCurrentCluster() || {}
     const twoMonthsAgo = _.tap(d => d.setMonth(d.getMonth() - 2), new Date())
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -182,7 +182,7 @@ export default class ClusterManager extends PureComponent {
         message: 'Your notebook runtime is over two months old. Please consider deleting and recreating your runtime in order to access the latest features and security updates.'
       })
     } else if (cluster.status === 'Running' && prevCluster.status === 'Updating') {
-      notify('success', 'Your runtime update has completed successfully.', { timeout: 3000 })
+      notify('success', 'Your runtime update has completed successfully.', { timeout: 5000 })
     }
   }
 
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -292,8 +292,7 @@ export default class ClusterManager extends PureComponent {
     }
     const totalCost = _.sum(_.map(clusterCost, clusters))
     const activeClusters = this.getActiveClustersOldestFirst()
-    const creating = _.some({ status: 'Creating' }, activeClusters)
-    const updating = _.some({ status: 'Updating' }, activeClusters)
+    const { Creating: creating, Updating: updating } = _.countBy('status', activeClusters)
     const isDisabled = !canCompute || creating || busy || updating
 
     const isRStudioImage = currentCluster?.labels.tool === 'RStudio'
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -184,7 +184,7 @@ export default class ClusterManager extends PureComponent {
     } else if (cluster.status === 'Running' && prevCluster.status === 'Updating') {
       notify('success', 'Your runtime update has completed successfully.', { timeout: 3000 })
     }
-   }
+  }
 
   getActiveClustersOldestFirst() {
     const { clusters } = this.props
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -197,18 +197,17 @@ export default class ClusterManager extends PureComponent {
   }
 
   async executeAndRefresh(promise, waitBeforeRefreshMillis = 0) {
-    await setTimeout(async () => {
-      try {
-        const { refreshClusters } = this.props
-        this.setState({ busy: true })
-        await promise
-        await refreshClusters()
-      } catch (error) {
-        reportError('Notebook Runtime Error', error)
-      } finally {
-        this.setState({ busy: false })
-      }
-    }, waitBeforeRefreshMillis)
+    try {
+      const { refreshClusters } = this.props
+      this.setState({ busy: true })
+      await promise
+      waitBeforeRefreshMillis && await Utils.delay(waitBeforeRefreshMillis)
+      await refreshClusters()
+    } catch (error) {
+      reportError('Notebook Runtime Error', error)
+    } finally {
+      this.setState({ busy: false })
+    }
   }
 
   createDefaultCluster() {
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -196,17 +196,19 @@ export default class ClusterManager extends PureComponent {
     return currentCluster(clusters)
   }
 
-  async executeAndRefresh(promise) {
-    try {
-      const { refreshClusters } = this.props
-      this.setState({ busy: true })
-      await promise
-      await refreshClusters()
-    } catch (error) {
-      reportError('Notebook Runtime Error', error)
-    } finally {
-      this.setState({ busy: false })
-    }
+  async executeAndRefresh(promise, waitBeforeRefreshMillis = 0) {
+    await setTimeout(async () => {
+      try {
+        const { refreshClusters } = this.props
+        this.setState({ busy: true })
+        await promise
+        await refreshClusters()
+      } catch (error) {
+        reportError('Notebook Runtime Error', error)
+      } finally {
+        this.setState({ busy: false })
+      }
+    }, waitBeforeRefreshMillis)
   }
 
   createDefaultCluster() {
@@ -341,9 +343,9 @@ export default class ClusterManager extends PureComponent {
         namespace,
         currentCluster,
         onDismiss: () => this.setState({ createModalDrawerOpen: false }),
-        onSuccess: promise => {
+        onSuccess: (promise, waitBeforeRefreshMillis = 0) => {
           this.setState({ createModalDrawerOpen: false })
-          this.executeAndRefresh(promise)
+          this.executeAndRefresh(promise, waitBeforeRefreshMillis)
         }
       }),
       errorModalOpen && h(ClusterErrorModal, {
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -1,7 +1,7 @@
 import _ from 'lodash/fp'
 import PropTypes from 'prop-types'
 import { Component, Fragment } from 'react'
-import { b, div, h, image, label, p, span } from 'react-hyperscript-helpers'
+import { b, div, h, label, p, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, ButtonSecondary, GroupedSelect, IdContainer, LabeledCheckbox, Link, Select } from 'src/components/common'
 import { ImageDepViewer } from 'src/components/ImageDepViewer'
 import { NumberInput, TextInput, ValidatedInput } from 'src/components/input'
@@ -155,7 +155,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     const { googleProject, clusterName } = currentCluster
 
     return Ajax().Clusters.cluster(googleProject, clusterName).update({
-        machineConfig: this.getMachineConfig()
+      machineConfig: this.getMachineConfig()
     })
   }
 
@@ -167,7 +167,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
       Ajax().Buckets.getObjectPreview('terra-docker-image-documentation', 'terra-docker-versions.json', namespace, true).then(res => res.json())
     ])
 
-    console.log("Current cluster details: ", currentClusterDetails)
+    console.log('Current cluster details: ', currentClusterDetails)
 
     this.setState({ leoImages: newLeoImages })
     if (currentClusterDetails) {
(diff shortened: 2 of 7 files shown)


src/components/NewClusterModal.js
```diff
@@ -230,15 +239,54 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
           }
-        }, !!currentCluster ? 'Replace' : 'Create')
+        }, !!currentCluster ? getUpdateOrReplace() : 'Create')
       ])
     ])
 
+    //determines whether the changes are applicable for a call to the leo patch endpoint
+    //see this for a diagram of the conditional this implements https://drive.google.com/file/d/1mtFFecpQTkGYWSgPlaHksYaIudWHa0dY/view
+    //this function returns true for cases 2 & 3 in this diagram
+    const canUpdate = () => {
```
jdcanas: this is a bit nasty, I'm sure it can be simplified a bit more, but I doubt it will ever be both concise and readable, so I chose readable.

src/components/cluster-common.js
```diff
@@ -77,9 +77,10 @@ export const ClusterStatusMonitor = ({ cluster, onClusterStoppedRunning = _.noop
   const prevStatus = Utils.usePrevious(currentStatus)
 
   useEffect(() => {
-    if (prevStatus === 'Running' && currentStatus !== 'Running') {
+    const runningStatuses = ['Running', 'Updating']
```
jdcanas: is there a better place to put this?

panentheos: Since it seems like there are a couple places that share this logic, we could make a little shared function in `cluster-utils.js` that extracts this notion of 'ok to use'

src/components/NewClusterModal.js
panentheos: Whatever the final logic, expressing it like a simple functional computation like this is a good approach, and will enable us to read and/or change it later if needed.

panentheos (commented):
Thanks for letting us take an early look at this. It's looking good so far.

src/components/NewClusterModal.js
jdcanas: It's also worth mentioning that this can go away once [this](https://broadworkbench.atlassian.net/secure/RapidBoard.jspa?rapidView=35&projectKey=IA&view=planning&selectedIssue=IA-1507&issueLimit=100&search=allowDelete) ticket is implemented, but it hasn't been prioritized yet

src/components/ClusterManager.js
```diff
@@ -194,17 +196,19 @@ export default class ClusterManager extends PureComponent {
...
+        const { refreshClusters } = this.props
+        this.setState({ busy: true })
+        await promise
+        await refreshClusters()
+      } catch (error) {
+        reportError('Notebook Runtime Error', error)
+      } finally {
+        this.setState({ busy: false })
+      }
+    }, waitBeforeRefreshMillis)
```
jdcanas: this is needed because sometimes leo needs to do some background tasks before updating the clusters status.

src/components/NewClusterModal.js
```diff
@@ -149,6 +150,66 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+
+    if (isStopRequired) {
+      notify('success', 'To be updated, your runtime will now stop, and then start.', { timeout: 3000 })
+    }
+
+    return onSuccess(
+      Ajax().Clusters.cluster(googleProject, clusterName).update({
+        machineConfig: this.getMachineConfig()
+      }),
+      5000)
```
jdcanas: here, I wait 5 seconds before refreshing if we are going to stop. We cannot refresh immediately, because leonardo needs to do a bit of work before the cluster status is actually changed in this specific case. In my testing I found 5 seconds a good balance. The only consequence to missing the window, is, if the user isn't actually editting their notebook yet, the cluster can transition from 'Running' to 'Starting', and this case is covered by the notification on line 158.

src/components/ClusterManager.js
```diff
@@ -194,17 +196,19 @@ export default class ClusterManager extends PureComponent {
...
-      this.setState({ busy: true })
-      await promise
-      await refreshClusters()
-    } catch (error) {
-      reportError('Notebook Runtime Error', error)
-    } finally {
-      this.setState({ busy: false })
-    }
+  async executeAndRefresh(promise, waitBeforeRefreshMillis = 0) {
+    await setTimeout(async () => {
```
jdcanas: did I do this timeout + await with nested async await logic correctly? Everything checked out in my testing.

akarukappadath (approved):
lgtm!

src/components/NewClusterModal.js
```diff
@@ -149,6 +150,66 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+
+    return onSuccess(
+      Ajax().Clusters.cluster(googleProject, clusterName).update({
+        machineConfig: this.getMachineConfig()
+      }),
+      isStopRequired ? 5000 : 0)
+  }
+
+  //determines whether the changes are applicable for a call to the leo patch endpoint
+  //see this for a diagram of the conditional this implements https://drive.google.com/file/d/1mtFFecpQTkGYWSgPlaHksYaIudWHa0dY/view
```
akarukappadath, line 181: Should we have documentation accessible to users to show which updates require a full restart of their cluster and which don't? Might be worth talking to comms

src/components/ClusterManager.js
```diff
@@ -288,7 +293,8 @@ export default class ClusterManager extends PureComponent {
     const totalCost = _.sum(_.map(clusterCost, clusters))
     const activeClusters = this.getActiveClustersOldestFirst()
     const creating = _.some({ status: 'Creating' }, activeClusters)
-    const isDisabled = !canCompute || creating || busy
+    const updating = _.some({ status: 'Updating' }, activeClusters)
```
petesantos:
```js
const { Creating: creating, Updating: updating } = _.countBy('status', activeClusters)
```

src/components/cluster-common.js
```diff
@@ -77,9 +78,9 @@ export const ClusterStatusMonitor = ({ cluster, onClusterStoppedRunning = _.noop
   const prevStatus = Utils.usePrevious(currentStatus)
 
   useEffect(() => {
-    if (prevStatus === 'Running' && currentStatus !== 'Running') {
+    if (prevStatus === 'Running' && !_.includes(currentStatus, usableStatuses)) {
       onClusterStoppedRunning()
-    } else if (prevStatus !== 'Running' && currentStatus === 'Running') {
+    } else if (prevStatus !== 'Running' && !_.includes(currentStatus, usableStatuses)) {
```
petesantos:
```suggestion
    } else if (prevStatus !== 'Running' && _.includes(currentStatus, usableStatuses)) {
```

src/libs/ajax.js
```diff
@@ -978,6 +978,13 @@ const Clusters = signal => ({
         return fetchLeo(`api/cluster/v2/${project}/${name}`, _.mergeAll([authOpts(), jsonBody(body), { signal, method: 'PUT' }, appIdentifier]))
       },
 
+      update: clusterOptions => {
+        const body = _.merge(clusterOptions, {
+          allowStop: true
+        })
```
petesantos:
```js
const body = { ...clusterOptions, allowStop: true }
```

petesantos (commented):
We wen't over some of the code in mobbing, have some more to do

src/components/NewClusterModal.js
jdcanas: cc @sharmatime what do you think? I can take these documents to the appropriate person  and answer any questions they may have if you think we should and point me in the right direction.

src/components/NewClusterModal.js
```diff
@@ -227,18 +290,20 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
           disabled: isSelectedImageInputted && isCustomImageInvalid,
           tooltip: isSelectedImageInputted && isCustomImageInvalid && 'Enter a valid docker image to use',
           onClick: () => {
-            if (isSelectedImageInputted) {
+            if (isSelectedImageInputted && !this.canUpdate()) {
```
jdcanas, line 302: if the user has already seen the warning, and already has a cluster with this image, we don't want to warn them again and proceed with the replace logic

src/components/ClusterManager.js
```diff
@@ -181,6 +181,8 @@ export default class ClusterManager extends PureComponent {
       notify('warn', 'Outdated Notebook Runtime', {
         message: 'Your notebook runtime is over two months old. Please consider deleting and recreating your runtime in order to access the latest features and security updates.'
       })
+    } else if (cluster.status === 'Running' && prevCluster.status === 'Updating') {
+      notify('success', 'Your runtime update has completed successfully.', { timeout: 3000 })
```
breilly2: TOL: 3 seconds feels a little short for this. What would you think of 5 seconds?

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -385,7 +386,7 @@ const copyingNotebookMessage = div({ style: { paddingTop: '2rem' } }, [
 ])
 
 const NotebookEditorFrame = ({ mode, notebookName, workspace: { workspace: { namespace, name, bucketName } }, cluster: { clusterName, clusterUrl, status, labels } }) => {
-  console.assert(status === 'Running', 'Expected notebook runtime to be running')
+  console.assert(_.includes(status, usableStatuses), 'Expected notebook runtime to be running or updating')
```
breilly2: Nit:
```suggestion
  console.assert(_.includes(status, usableStatuses), `Expected notebook runtime to be one of: ${usableStatuses}`)
```
I think this will do what I expect, but please test to make sure.

src/components/NewClusterModal.js
```diff
@@ -162,17 +223,20 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+
       if (jupyterUserScriptUri) {
-        this.setState({ jupyterUserScriptUri, profile: 'custom' })
+        this.setState({ jupyterUserScriptUri, profile: 'custom', originalJupyterUserScriptUri: jupyterUserScriptUri })
+      } else {
+        this.setState({ originalJupyterUserScriptUri: '' })
       }
     } else {
-      this.setState({ selectedLeoImage: _.find({ id: 'terra-jupyter-gatk' }, newLeoImages).image })
+      this.setState({ selectedLeoImage: _.find({ id: 'terra-jupyter-gatk' }, newLeoImages).image, originalImageUrl: '', originalJupyterUserScriptUri: '' })
```
breilly2: Instead of worrying about interpreting `originalImageUrl` here, which only matters in one place in `canUpdate`. It would be simpler to put all of `currentClusterDetails` in state and  check for its existence in `canUpdate`.

Let me know if you want to chat through this. I'm happy to pair on the refactor.

src/components/NewClusterModal.js
```diff
@@ -227,18 +290,20 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
           disabled: isSelectedImageInputted && isCustomImageInvalid,
           tooltip: isSelectedImageInputted && isCustomImageInvalid && 'Enter a valid docker image to use',
           onClick: () => {
-            if (isSelectedImageInputted) {
+            if (isSelectedImageInputted && !this.canUpdate()) {
               this.setState({ viewMode: 'warning' })
             } else if (!!currentCluster) {
-              this.setState({ viewMode: 'replace' })
+              this.setState({ viewMode: _.lowerCase(getUpdateOrReplace()) })
```
breilly2: Nit: This would be more consistent with line 294 if `getUpdateOrReplace` returns lowercase symbols, followed by using `_.startCase` to render the button text.

src/components/NewClusterModal.js
```diff
@@ -149,6 +150,66 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+    const { currentCluster } = this.props
+    const { jupyterUserScriptUri, originalJupyterUserScriptUri, originalImageUrl, selectedLeoImage, customEnvImage } = this.state
+    const currentClusterConfig = currentCluster.machineConfig
+    const userSelectedConfig = this.getMachineConfig()
+
+    const hasStartupScriptChanged = jupyterUserScriptUri !== originalJupyterUserScriptUri
+
+    const hasImageChanged = !_.includes(originalImageUrl, [selectedLeoImage, customEnvImage])
+
+    const workersCantUpdate = currentClusterConfig.numberOfWorkers !== userSelectedConfig.numberOfWorkers &&
```
breilly2: Nit: `cantUpdateWorkers`. Better yet, `canUpdateWorkers` (see feedback on line 196).

src/components/NewClusterModal.js
```diff
@@ -149,6 +150,66 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+      currentClusterConfig.numberOfWorkerLocalSSDs !== userSelectedConfig.numberOfWorkerLocalSSDs
+
+    const hasWorkers = currentClusterConfig.numberOfWorkers >= 2 || currentClusterConfig.numberOfPreemptibleWorkers >= 2
+
+    const hasWorkersResourceChanged = hasWorkers && hasUnUpdateableResourceChanged
+
+    const hasDiskSizeDecreased = currentClusterConfig.masterDiskSize > userSelectedConfig.masterDiskSize
+
+    const cantUpdate = workersCantUpdate || hasWorkersResourceChanged || hasDiskSizeDecreased || hasImageChanged || hasStartupScriptChanged || !currentCluster
+    return !cantUpdate
```
breilly2, line 202: Nit: It's a bit awkward to calculate a negative and then negate it. Also, this feels like it would more obviously implement the diagram if written as if/else statements. If that's not the case and you want to keep it a set of boolean expressions, it might be easier to digest if the intermediate tests are extracted into separate functions. `hasStartupScriptChanged`, `hasImageChanged`, `workersCantUpdate`, `hasUnUpdateableResourceChanged`, `hasWorkers`, and `hasDiskSizeDecreased` are completely independent of each other. Still, I have a feeling that some of those are only needed some of the time which would be easier to recognize using if/else.

src/components/NewClusterModal.js
```diff
@@ -149,6 +150,66 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     ]))
   }
 
+  updateCluster(isStopRequired = false) {
+    const { currentCluster, onSuccess } = this.props
+    const { googleProject, clusterName } = currentCluster
+
+    if (isStopRequired) {
+      notify('success', 'To be updated, your runtime will now stop, and then start.', { timeout: 3000 })
```
breilly2: TOL: 3 seconds feels a little short for this. What would you think of 5 seconds? Or even not auto-dismissing in this case?

src/components/NewClusterModal.js
```diff
@@ -405,6 +470,22 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
         ])
       ])],
+      ['update', () => h(Fragment, [
+        this.isStopRequired()
+          ? p(['Changing the machine type (increasing or decreasing the # of CPUs or Mem) results in an update that requires a ',
+            b(['restart']),
+            ' of your runtime. This may take a few minutes.  Would you like to proceed? ',
+            b(['(You will not lose any files.)'])])
+          : p(['Increasing the disk size or changing the number of workers (when the number of workers is >2) results in a real-time update to your runtime. ',
+            'During this update, you can continue to work']),
```
breilly2: Nit: I know Isaac mentioned something about the formatting of this `?:` but I'm not 100% sure what the more conventional way is. It might be something like:
```
        this.isStopRequired() ? p([
          'Changing the machine type (increasing or decreasing the # of CPUs or Mem) results in an update that requires a ',
          b(['restart']),
          ' of your runtime. This may take a few minutes.  Would you like to proceed? ',
          b(['(You will not lose any files.)'])
        ]) : p([
          'Increasing the disk size or changing the number of workers (when the number of workers is >2) results in a real-time update to your runtime. ',
          'During this update, you can continue to work'
        ]),
```
but we should check with Isaac to make sure.

breilly2 (commented):
Some feedback, mostly small.

As a reminder of our PR lingo:
* "TOL": "thinking out loud" discussion points that may or may not lead to changes
* "Nit": non-critical suggestions that would be nice to make for improved readability/consistency/etc.

Most comments are negotiable and could be put off to a follow-on PR.

src/components/ClusterManager.js
jdcanas: Sure. I just saw all other places where the type is `success` having the timeout as 3000, was just trying to be consistent.

src/components/NewClusterModal.js
jdcanas: your suggestion seems to be consistent with other snippets I found:

```
   isFileImportCurrMode ? div([
              'Select the ',
              h(TooltipTrigger, { content: 'Tab Separated Values', side: 'bottom' },
                [span({ style: { textDecoration: 'underline dashed' } }, 'TSV')]),
              ' file containing your data: '
            ]) : div(['Copy and paste tab separated data here:']),
```
cc @zarsky-broad  ?

zarsky-broad: When both cases are multiline, I prefer this, because I find it a little easier to see where each block starts and ends:
```js
condition ?
  div([
    ...
  ]) :
  div([
    ...
  ])
```

breilly2: And Isaac has improved our eslint rules to help encourage consistency: https://github.com/DataBiosphere/terra-ui/pull/2046

src/components/NewClusterModal.js
jdcanas: I feel like this is the simplest way to do it. It's easier to reason about since each situation is completely isolated in its own boolean, as opposed to needing to reason about a nested structure similar to that in `componentDidMount`. 

I split the state-dependent lines into their own function with the piece of state they need, but I kept the other pieces, all of which rely on the below code in some way. I didn't like how those pieces looked in their own function because of all the duplication in variable initialization steps.

I agree that its atypical to calculate something, and then return its negation, but I found it was the easiest way to make each piece easy to reason about as it resulted in the least amount of negations overall. 

```    
const { currentCluster } = this.props
const currentClusterConfig = currentCluster.machineConfig
const userSelectedConfig = this.getMachineConfig()
```

breilly2 (approved):
👍 Looks good. I did a bunch of testing locally (PR site was not able to show jupyter iframe). Multi-window testing showed some corner cases that we could handle better, but that's not directly related to these changes.

PR #2023, opened 2020-01-31
Change tooltip language [SATURN-1221]

ehanna4:
Updated the tooltip for opening notebooks from the IGV drawer to use the existing more concise text option to be more clear/simple.

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -520,8 +520,9 @@ const ToolDrawer = _.flow(
               onClick: () => setToolMode('Notebook'),
               disabled: !notebookButtonEnabled,
               tooltip: Utils.cond(
+                [!entityMetadata.cohort, () => 'Unable to open with notebooks. See the "Making custom cohorts with Data Explorer" help article for more details.'],
                 [isCohort && entitiesCount > 1, () => 'Select exactly one cohort to open in notebook'],
-                [!isCohort || !entityMetadata.cohort, () => 'Only cohorts can be opened with notebooks'],
+                [!isCohort, () => 'Only cohorts can be opened with notebooks'],
                 [notebookButtonEnabled, () => 'Create a Python 2 or 3 notebook with this cohort']
               ),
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -520,9 +520,8 @@ const ToolDrawer = _.flow(
               onClick: () => setToolMode('Notebook'),
               disabled: !notebookButtonEnabled,
               tooltip: Utils.cond(
-                [!entityMetadata.cohort, () => 'Talk to your dataset owner about setting up a Data Explorer. See the "Making custom cohorts with Data Explorer" help article.'],
                 [isCohort && entitiesCount > 1, () => 'Select exactly one cohort to open in notebook'],
-                [!isCohort, () => 'Only cohorts can be opened with notebooks'],
+                [!isCohort || !entityMetadata.cohort, () => 'Only cohorts can be opened with notebooks'],
```
panentheos: nit: I don't think the change to this line does anything, because having a cohort selected implies that cohorts exist as a type in the metadata.

panentheos (approved):
Looks good, see comment. Tested locally in workspaces with and without cohorts.

TOL: Do we understand this change fully? I'm pretty sure the original message was put there intentionally to explain about Data Explorers, in case you hadn't used one. Without that message, someone might not know what a 'cohort' is or how to get one.

ehanna4:
Hmm, I see your point. When I made this change it was based on two things 1. It seemed to me the confusion about this tooltip was that it brings up Data Explorers when the user is trying to open a notebook, so I thought removing those words would make it more focused. 2. I read the article it suggests reading for help but I found it unhelpful because it focuses on how to open cohorts that can be opened with Notebooks, not on opening data in a notebook that doesn't fit that style which, is the sort of data this tooltip would be for. However I hadn't considered that it might be helpful just for providing more general understanding about the data and what a cohort is. I will consider alternatives. Thanks for the feedback!

panentheos (approved):
Looks good, thanks for the updated language.

PR #2024, opened 2020-02-03
fix loading indicies for igv [SATURN-1363]

zarsky-broad:
Verified fixed in-browser.

Previously, we were allowing IGV to assume the names of index files (i.e. foo.bam -> foo.bam.bai), but people sometimes have the file named foo.bai, without the .bam. This makes sure that an index can be found under one of those two naming conventions (when necessary), and explicitly passes that in to IGV.


ehanna4:
Good catch! - surprised it hasn't been an issue more earlier since it's explicitly mentioned in this Broad BAM documentation (https://software.broadinstitute.org/software/igv/BAM) that it's allowed to be `file_name.bam.bai` or `file_name.bai`

src/components/IGVBrowser.js
```diff
@@ -8,39 +8,26 @@ import colors from 'src/libs/colors'
...
-    const getTrack = filePath => {
-      const fileTypeToTrack = {
-        bam: 'alignment',
-        cram: 'alignment',
-        bed: 'annotation',
-        vcf: 'variant'
-      }
-
-      return {
-        type: fileTypeToTrack[_.last(filePath.split('.'))],
```
panentheos, line 26: Is `type` not needed any more?

src/components/IGVFileSelector.js
```diff
@@ -1,131 +1,105 @@
...
+        ['vcf', () => _.find(v => _.includes(v, [`${base}.idx`, `${base}.vcf.idx`, `${base}.tbi`, `${base}.vcf.tbi`]), allFiles)],
+        ['bed', () => undefined],
+        [Utils.DEFAULT, () => false]
+      )
 
-  render() {
-    const { onSuccess } = this.props
-    const { selectedFiles, refGenome } = this.state
-    const trackFiles = this.getIGVFileList()
+      return matchingIndex !== false ? _.set([filePath, 'indexFile'], matchingIndex, acc) : acc
```
panentheos: nit: we could use `flatMap` and `fromPairs` if we wanted to avoid the `reduce`

src/components/IGVFileSelector.js
```diff
@@ -1,131 +1,105 @@
...
+
+  return div({ style: Style.modalDrawer.content }, [
+    h(IdContainer, [id => div({ style: { fontWeight: 500 } }, [
+      label({ htmlFor: id }, ['Reference genome: ']),
+      div({ style: { display: 'inline-block', marginLeft: '0.25rem', marginBottom: '1rem', minWidth: 125 } }, [
+        h(Select, {
+          id,
+          options: ['hg38', 'hg19', 'hg18', 'mm10', 'panTro4', 'panPan2', 'susScr11', 'bosTau8', 'canFam3', 'rn6', 'danRer10', 'dm6', 'sacCer3'],
+          value: refGenome,
+          onChange: setRefGenome
```
panentheos: This needs to destructure `value`, it's not currently working.

src/components/IGVFileSelector.js
```diff
@@ -1,131 +1,105 @@
...
-                      checked: selectedFiles[name],
-                      onChange: () => this.toggleVisibility(name)
-                    })
-                  ]),
-                  h(Clickable, {
-                    style: styles.columnName,
-                    title: name,
-                    onClick: () => this.toggleVisibility(name)
-                  }, [_.last(name.split('/'))])
+  const toggleSelected = filePath => setSelections(_.set([filePath, 'isSelected'], !selections[filePath].isSelected))
```
panentheos: nit: Since the new value is a function of the old, we could use `_.update` instead of `_.set` to simplify slightly

src/components/IGVFileSelector.js
```diff
@@ -1,131 +1,105 @@
...
-  buttonIsDisabled() {
-    const selectedFilesList = this.getSelectedFilesList()
-    return !(selectedFilesList.length > 0 && selectedFilesList.length <= MAX_CONCURRENT_IGV_FILES)
-  }
+      if (!base) return acc
 
-  setAll(value) {
-    this.setState({ selectedFiles: _.fromPairs(_.map(v => [v, value], this.getIGVFileList())) })
-  }
+      const matchingIndex = Utils.switchCase(extension,
```
panentheos: TOL: Just making sure I understand the logic here. If we detect a known index file pattern, we'll explicitly pass that. If we don't, we'll pass `undefined`, which means IGV will infer based on the existing logic. So this will continue to work if e.g. the data didn't mention the index file explicitly, but it did exist at the inferred path.

panentheos (approved):
Looks good overall, but see comments.

Tested by opening up some bams locally.

src/components/IGVBrowser.js
zarsky-broad: Correct, it's implied from the extension by IGV, we'd just be duplicating their logic.

PR #2025, opened 2020-02-04
Give config to integration tests

breilly2:
Note: DRAFT. There are still a couple of references to global config that I missed.

Related to (but does not complete) https://broadworkbench.atlassian.net/browse/SATURN-1341. Will allow the service to dynamically point to different Terra deployments.

### DIFF
--- a/integration-tests/tests/import-dockstore-workflow.js
+++ b/integration-tests/tests/import-dockstore-workflow.js
@@ -1,14 +1,14 @@
 const fetch = require('node-fetch')
-const { testUrl } = require('../utils/integration-config')
 const { withWorkspace } = require('../utils/integration-helpers')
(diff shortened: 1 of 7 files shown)


integration-tests/tests/import-dockstore-workflow.js
```diff
@@ -19,7 +19,7 @@ const testImportDockstoreWorkflowFn = async ({ context, page }) => {
       await select(page, 'Select a workspace', workspaceName)
       await click(page, clickable({ text: 'Import' }))
       await findText(page, testWorkflowIdentifier)
-    })({ context })
+    })(options)
```
breilly2, line 22: This felt funky before and it feels a little more funky now. Is there a better way to write this that doesn't put `options` all the way down here?

breilly2:
Closing for now. Will re-open or replace later.

PR #2027, opened 2020-02-05
Don't need to wait 60s for sam to know about workspaces anymore

zarsky-broad:
"postgres is out so opendj caching should be a thing of the past" – @MatthewBemis


panentheos (approved):
Looks good!

PR #2028, opened 2020-02-06
fix submission page auto-refreshing

zarsky-broad:
Verified working in-browser.


integration-tests/tests/run-workflow.js
```diff
@@ -39,16 +39,11 @@ const testRunWorkflowFn = withWorkspace(async ({ page, workspaceName }) => {
 
   await pRetry(async () => {
     try {
-      await findInGrid(page, 'Succeeded')
+      await findInGrid(page, 'Succeeded', { timeout: 65 * 1000 }) // long enough for the submission details to refresh
```
petesantos, line 42: TOL: Just wondering if there is something we can do to identify specific timeouts for refresh in the app. We encode this in the test, based on the app timeout, if this changes in the app we may not know that the test is linked to it.
This may not be a big deal, just something to think about.

petesantos (approved):
Looks good

PR #2029, opened 2020-02-06
Initial workflow repo skeleton

zarsky-broad:
Adding as a dark feature: there are no links to these pages, but you can access the workflow list at `/#workflows`.

This is just the first functionality of the repo: list workflows, view snapshot details, and list configs.


src/components/common.js
```diff
@@ -131,18 +131,18 @@ export const makeIconButton = (shape, { disabled, size, iconProps = {}, ...props
   ])
 }
 
-export const TabBar = ({ activeTab, tabNames, refresh = _.noop, getHref, children }) => {
+export const TabBar = ({ activeTab, tabNames, displayNames = {}, refresh = _.noop, getHref, getClick = _.noop, children }) => {
```
breilly2: Nit: `onClick`?

src/pages/workflows/List.js
```diff
@@ -0,0 +1,156 @@
...
+        ])
+      ])
+    ])
+  ])
+}
+
+export const navPaths = [
+  {
+    name: 'workflows',
+    path: '/workflows',
```
breilly2, line 153: We probably need to support redirects from FireCloud which uses "methods" for these paths. Both here and in `WorkflowDetails`.

src/pages/workflows/List.js
```diff
@@ -0,0 +1,156 @@
...
+        fetch(`${getConfig().firecloudBucketRoot}/featured-methods.json`, { signal }).then(res => res.json())
+      ])
+
+      setWorkflows({
+        mine: _.filter(isMine, allWorkflows),
+        featured: _.flow(
+          _.map(featuredWf => _.find(featuredWf, allWorkflows)),
+          _.compact
+        )(featuredList),
+        public: _.filter('public', allWorkflows)
```
breilly2, line 55: TOL: Would this be a good candidate for `_.groupBy`? You could at least get `mine` and `public` out of that.

src/pages/workflows/List.js
```diff
@@ -0,0 +1,156 @@
...
+
+  const tabName = tab || 'mine'
+
+  Utils.useOnMount(() => {
+    const isMine = ({ public: isPublic, managers }) => !isPublic || _.includes(getUser().email, managers)
+
+    const loadWorkflows = async () => {
+      const [allWorkflows, featuredList] = await Promise.all([
+        Ajax(signal).Methods.definitions(),
+        fetch(`${getConfig().firecloudBucketRoot}/featured-methods.json`, { signal }).then(res => res.json())
```
breilly2, line 46: TOL/Nit: Don't need to wait for featured methods yet unless we're viewing the featured methods tab. May not be worth the extra complexity, though.

src/pages/workflows/List.js
```diff
@@ -0,0 +1,156 @@
...
+                size: { basis: 108, grow: 0, shrink: 0 }
+              },
+              {
+                headerRenderer: () => h(Sortable, { sort, field: 'numConfigurations', onSort: setSort }, [h(HeaderCell, ['Configurations'])]),
+                cellRenderer: ({ rowIndex }) => {
+                  const { numConfigurations } = sortedWorkflows[rowIndex]
+
+                  return div({ style: { textAlign: 'end', flex: 1 } }, [numConfigurations])
+                },
+                size: { basis: 140, grow: 0, shrink: 0 }
```
breilly2: The header text is getting snipped on my screen:
Screen Shot 2020-02-10 at 11 35 46 AM

src/pages/workflows/workflow/WorkflowDetails.js
```diff
@@ -0,0 +1,304 @@
...
+import colors from 'src/libs/colors'
+import { getConfig } from 'src/libs/config'
+import { withErrorReporting } from 'src/libs/error'
+import * as Nav from 'src/libs/nav'
+import { snapshotsListStore, snapshotStore } from 'src/libs/state'
+import * as Style from 'src/libs/style'
+import * as Utils from 'src/libs/utils'
+
+
+const styles = {
```
breilly2: Nit: Should extract refactor these styles that are shared with the workspace dashboard view.

src/pages/workflows/List.js
```diff
@@ -0,0 +1,156 @@
...
+  const [workflows, setWorkflows] = useState()
+
+  const getTabQueryName = newTab => newTab === 'mine' ? undefined : newTab
+
+  const getUpdatedQuery = ({ newTab = tab, newFilter = filter }) => {
+    // Note: setting undefined so that falsy values don't show up at all
+    return qs.stringify({ ...query, tab: getTabQueryName(newTab), filter: newFilter || undefined }, { addQueryPrefix: true })
+  }
+
+  const updateQuery = newParams => {
```
petesantos, line 30: Should this be wrapped in a useEffect with newTab and newFilter being set as state?
I view this as a side effect that is occurring as the result of a user action, but we are directly manipulating the nav location rather than capturing and calling it out in an effect

src/pages/workflows/List.js
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+
+  const tabName = tab || 'mine'
+
+  Utils.useOnMount(() => {
+    const isMine = ({ public: isPublic, managers }) => !isPublic || _.includes(getUser().email, managers)
+
+    const loadWorkflows = async () => {
+      const [allWorkflows, featuredList] = await Promise.all([
```
petesantos, line 44: Should this be wrapped in withErrorReporting?

src/pages/workflows/workflow/WorkflowDetails.js
```diff
@@ -0,0 +1,304 @@
...
+    snapshot ?
+      children :
+      centeredSpinner()
+  ])
+}
+
+const WorkflowSummary = () => {
+  const { namespace, name, snapshotId, createDate, managers, synopsis, documentation, public: isPublic } = Utils.useStore(snapshotStore)
+  const [importUrlCopied, setImportUrlCopied] = useState()
+  const importUrl = `${getConfig().orchestrationUrlRoot}/ga4gh/v1/tools/${namespace}:${name}/versions/${snapshotId}/plain-WDL/descriptor`
```
breilly2, line 117: Is this a new feature? I couldn't find this in FC UI. I'm guessing this is for composing WDLs?

src/pages/workflows/workflow/WorkflowDetails.js
```diff
@@ -0,0 +1,304 @@
...
+  const cachedSnapshotsList = Utils.useStore(snapshotsListStore)
+
+
+  const snapshotsList = cachedSnapshotsList && _.isEqual({ namespace, name }, _.pick(['namespace', 'name'], cachedSnapshotsList[0])) ?
+    cachedSnapshotsList :
+    undefined
+
+  Utils.useOnMount(() => {
+    const loadSnapshots = async () => {
+      snapshotsListStore.set(snapshotsList || await Ajax(signal).Methods.list({ namespace, name }))
```
petesantos, line 44: Should we add error reporting here?

src/pages/workflows/workflow/WorkflowDetails.js
```diff
@@ -0,0 +1,304 @@
...
+  const selectedSnapshot = (snapshotId * 1) || _.last(cachedSnapshotsList).snapshotId
+
+  const snapshot = cachedSnapshot &&
+  _.isEqual({ namespace, name, snapshotId: selectedSnapshot }, _.pick(['namespace', 'name', 'snapshotId'], cachedSnapshot)) ?
+    cachedSnapshot :
+    undefined
+
+  Utils.useOnMount(() => {
+    const loadSnapshot = async () => {
+      snapshotStore.set(await Ajax(signal).Methods.method(namespace, name, selectedSnapshot).get())
```
petesantos, line 79: Do we need error reporting?

breilly2 (approved):
👍 Looking great! I clicked through a bunch of things, comparing it to FireCloud UI, including incompatible configs. I didn't carefully re-review all of the detail around the navigation because we went over that during mobbing or office hours recently.

petesantos (commented):
Looking good! A few minor comments

src/pages/workflows/List.js
zarsky-broad: I don't think so, only because the groups aren't mutually exclusive.

src/pages/workflows/List.js
zarsky-broad: We may be removing featured methods altogether, since they all just say to find the latest version in Dockstore anyway

src/pages/workflows/workflow/WorkflowDetails.js
zarsky-broad: It's in the wdl tab in fc

zarsky-broad:
The questions about error reporting and setting a pattern for `useEffect` manipulating url query are really good; since this is just intended to be a skeleton, I'm not going to address them right now, but I'll leave comments in the code.

PR #2030, opened 2020-02-06
update deps

zarsky-broad:
Read changelogs and tested, no major changes found.

Perhaps of interest, Puppeteer is starting to support firefox with this version. We could think about running tests against chrome and ff in parallel at some point in the future, if we wanted...


petesantos (approved):
looks good, clicked around a bit

PR #2033, opened 2020-02-07
Update text in data library [SATURN-1361]

zarsky-broad:
Changed per ticket.

### DIFF
--- a/src/pages/library/Datasets.js
+++ b/src/pages/library/Datasets.js
@@ -116,13 +116,6 @@ class Participant extends Component {
 const browseTooltip = 'Look for the Export to Terra icon to export data from this provider.'
 
 
-const NIHCommonsButtons = () => h(ButtonPrimary, {
-  style: { margin: '0.25rem 0' },
-  href: 'https://gen3.datastage.io/explorer',
-  ...Utils.newTabLinkProps,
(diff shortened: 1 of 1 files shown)


src/pages/library/Datasets.js
```diff
@@ -116,13 +116,6 @@ class Participant extends Component {
 const browseTooltip = 'Look for the Export to Terra icon to export data from this provider.'
 
 
-const NIHCommonsButtons = () => h(ButtonPrimary, {
-  style: { margin: '0.25rem 0' },
-  href: 'https://gen3.datastage.io/explorer',
-  ...Utils.newTabLinkProps,
-  tooltip: browseTooltip
-}, ['Browse STAGE Repository'])
```
breilly2, line 124: This wording was specifically requested: https://broadworkbench.atlassian.net/browse/SATURN-397. I don't know the history and it would be a lot of text for a button, but we should check whether or not we have to say something like, "Browse BioData Catalyst Repository".

breilly2 (approved):
👍 Looks good. See comment.

src/pages/library/Datasets.js
zarsky-broad: This wording was specified in the ticket, so I'm gonna say we're good.

PR #2034, opened 2020-02-07
Try to make integration tests less flaky

ehanna4:
Hopefully fixes the notification issue that blocks on screen clickable things
- decided to dismiss all notifications immediately after login as these will be non-error type notifications at that time

Also adjusted the run-workflow time to be longer to test how circleCi will handle that for the tests tonight. Will go back and investigate if that's best solution

### DIFF
--- a/integration-tests/tests/find-workflow.js
+++ b/integration-tests/tests/find-workflow.js
@@ -1,12 +1,13 @@
 const firecloud = require('../utils/firecloud-utils')
 const { billingProject, testUrl, workflowName } = require('../utils/integration-config')
 const { withWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, findElement, findText, signIntoTerra } = require('../utils/integration-utils')
+const { click, clickable, dismissNotifications, findElement, findText, signIntoTerra } = require('../utils/integration-utils')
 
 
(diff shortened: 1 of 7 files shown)


zarsky-broad (approved):
do it, seen working locally

integration-tests/tests/find-workflow.js
```diff
@@ -1,12 +1,13 @@
...
 const { billingProject, testUrl, workflowName } = require('../utils/integration-config')
 const { withWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, findElement, findText, signIntoTerra } = require('../utils/integration-utils')
+const { click, clickable, dismissNotifications, findElement, findText, signIntoTerra } = require('../utils/integration-utils')
 
 
 const testFindWorkflowFn = withWorkspace(async ({ page, workspaceName }) => {
   await page.goto(testUrl)
   await signIntoTerra(page)
+  await dismissNotifications(page)
```
breilly2, line 10: Why not call `dismissNotifications` from `signIntoTerra`?

zarsky-broad: We made this decision because there are lots of sign ins just for Ajax, like when creating or deleting a workspce

breilly2: Good reason. 👍

PR #2035, opened 2020-02-07
updated ToS [SATURN-1351]

zarsky-broad:
On hold until the Mixpanel ToS updates are available.

TODO:
- [x] Tweak "as of" date before deploy
- [x] See if it's ok that I replaced the "cancel" button with "sign out" (since that's what it does)
- [x] Get the new version into the databases for alpha, perf, and staging (I've done the other envs already)
- [x] Accept new ToS as the test user on all envs so the integration tests pass OR teach the tests to handle this (done on dev already)

### DIFF
--- a/src/pages/TermsOfService.js
+++ b/src/pages/TermsOfService.js
@@ -295,7 +295,7 @@ including attorneys' fees and costs, arising out of or relating to your
 uploading, storing, or transferring of PHI without having fully
 implemented the Clinical Features.
 
-Terms as of February 11, 2020.
+Terms as of February 12, 2020.
 `
 
 const TermsOfServicePage = () => {
--- a/src/pages/TermsOfService.js
+++ b/src/pages/TermsOfService.js
@@ -1,7 +1,7 @@
-import { Component } from 'react'
+import { useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { backgroundLogo, ButtonPrimary, ButtonSecondary } from 'src/components/common'
-import { MarkdownViewer } from 'src/components/markdown'
+import { MarkdownViewer, newWindowLinkRenderer } from 'src/components/markdown'
 import { Ajax } from 'src/libs/ajax'
 import { signOut } from 'src/libs/auth'
 import colors from 'src/libs/colors'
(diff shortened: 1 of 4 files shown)


src/pages/TermsOfService.js
```diff
@@ -1,281 +1,341 @@
...
-  constructor(props) {
-    super(props)
-    this.state = { busy: false }
-  }
+const TermsOfServicePage = () => {
+  const [busy, setBusy] = useState()
+  const { isSignedIn, acceptedTos } = authStore.get() // can't change while viewing this without causing it to unmount, so doesn't need to subscribe
 
-  async accept() {
+  const accept = async () => {
```
breilly2, line 306: Can/should this use `Utils.withBusyState`?

src/pages/TermsOfService.js
```diff
@@ -1,281 +1,341 @@
...
+      div({ style: { maxHeight: 400, overflowY: 'auto', lineHeight: 1.5, marginTop: '1rem', paddingRight: '1rem' } }, [
+        h(MarkdownViewer, {
+          renderers: {
+            link: newWindowLinkRenderer,
+            heading: (text, level) => `<h${level} style="color: ${colors.dark()}; margin-bottom: 0">${text}</h${level}>`
+          }
+        }, [termsOfService])
+      ]),
+      isSignedIn && !acceptedTos && div({ style: { display: 'flex', justifyContent: 'flex-end', marginTop: '2rem' } }, [
+        h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: signOut }, 'Sign Out'),
```
breilly2: How about "Cancel and Sign Out"?

src/pages/TermsOfService.js
```diff
@@ -1,281 +1,341 @@
...
-Terra uses cookies to enable sign on and other essential features when signed in, and to provide
-statistics to our development team regarding how the site is used. For more information, see our
-privacy policy. Privacy policy, including privacy policy of uploaded data, is [here](/#privacy).
-You agree to read and consider before uploading data.
+
+Terra uses cookies to enable sign on and other essential features when
+signed in, and to provide statistics to our development team regarding
+how the site is used. For more information, see our privacy policy.
+Privacy policy, including privacy policy of uploaded data, is
+[here](/#privacy). You agree to read and consider before uploading data.
```
breilly2, line 42: This probably isn't a new problem with this PR, but I just noticed that I can't view the privacy policy. It show briefly and then redisplays the Terms of Service.

src/pages/TermsOfService.js
```diff
@@ -1,281 +1,341 @@
...
-      div({ style: styles.box }, [
-        termsTitle,
-        TOSMarkdown,
-        div({ style: { display: 'flex', justifyContent: 'flex-end', marginTop: '2rem' } }, [
-          h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: signOut }, 'Cancel'),
-          h(ButtonPrimary, { onClick: () => this.accept(), disabled: busy }, ['Accept'])
-        ])
+  return div({ style: { padding: '1rem', minHeight: '100%', display: 'flex', justifyContent: 'center', alignItems: 'center' } }, [
+    backgroundLogo,
+    div({ style: { backgroundColor: 'rgba(255,255,255,0.85)', borderRadius: 5, width: 800, padding: '2rem', boxShadow: Style.standardShadow } }, [
```
breilly2: We removed the opacity on the data import page at Joy's request. In the name of a11y, I suggest we do the same here.

src/pages/TermsOfService.js
```diff
@@ -1,281 +1,341 @@
...
+      div({ style: { color: colors.dark(), fontWeight: 600 } }, [
+        span({ style: { fontSize: 36 } }, ['TERRA ']), span({ style: { fontSize: 24 } }, ['Terms of Service'])
```
breilly2: TOL: Strange that "TERRA" is all caps. 🤔

breilly2 (approved):
👍 Looks good overall. See comments.

I'm in favor of manually accepting the new TOS for our test account in each env instead of complicating the test code with conditional TOS acceptance... code which will be run so infrequently that it could even go stale and stop working by the next time we need it.

src/pages/TermsOfService.js
zarsky-broad: That would cause a state change after unmount in the success case.

breilly2: Cool. At some point, could you teach me more about this? It's not quite clicking for me.

PR #2038, opened 2020-02-10
BDC sign-in needs to go through warning

zarsky-broad:
Discussed with Adrian.


breilly2 (approved):
👍 Looks good. Tested on PR site for Terra, BioData Catalyst, Datastage, and Anvil to make sure both (and only) BioData Catalyst and Datastage have the alternative sign-in sequence.

PR #2040, opened 2020-02-12
CUJ 4 & 5 Run Notebook [SATURN-1250]

ehanna4:
Integration test for running a notebook and interacting.
Create a notebook, run it, and run a line that just prints some numbers.

Tested running locally with the rest of the integration suite against dev and alpha

Had to update some selects to be 'searchable' and added aria-labels so they can be found by xpath - otherwise they weren't being found. Chatted with Brett briefly and he mentioned this is a known react-select bug, so this work around seemed reasonable for now? Would appreciate any other thoughts though.

Also made this test start with a 1 so that it runs first with circleCi instead of run-workflow because it has to wait for notebook creation which is at least 5 minutes.


ehanna4:
Because this test is doing runtime configuration in a billing account that other people have access to, cases needed to be added for handling a runtime configuration in the different statuses (Stopping, Running, Stopped, Creating, None). However I am not sure how this will work running a bunch of users on the same billing account at the same time for perf testing, since they will all be running into each others runtime configurations ?

^^ misunderstood how the runtime configurations work need to revert changes

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,43 @@
+const { withRegisteredUser, withBilling, withWorkspace } = require('../utils/integration-helpers')
+const { click, clickable, signIntoTerra, findElement, waitForNoSpinners, select, delay, fillIn, input, findIframe, findText, dismissNotifications } = require(
+  '../utils/integration-utils')
+
+
+const notebookName = 'TestNotebook'
+
+const testRunNotebookFn = withRegisteredUser(async ({ page, context, email, token }) => {
+  await withBilling(withWorkspace(async ({ workspaceName }) => {
+    await page.reload()
```
ehanna4: This is a left over from registeruser being already on workspaces, so instead make regiseterUser open a new window instead of relying on taking us to the workspaces page

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,43 @@
...
+    await click(page, clickable({ textContains: notebookName }))
+    await click(page, clickable({ text: 'Edit' }))
+    await select(page, 'Select Environment', 'Hail')
+    await click(page, clickable({ text: 'Create' }))
+    await findElement(page, clickable({ textContains: 'Creating' }))
+    await delay(300000) //creation takes about 5 minutes
+    await findElement(page, clickable({ textContains: 'Running' }))
+    await delay(5000) //wait for copy over or change of status
+
+    const frame = await findIframe(page)
```
ehanna4, line 29: consider adding check that 27 isn't here yet

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,43 @@
...
+    await click(page, clickable({ textContains: 'Create a' }))
+    await fillIn(page, input({ placeholder: 'Enter a name' }), notebookName)
+    await select(page, 'Select a language', 'Python 2')
+    await click(page, clickable({ text: 'Create Notebook' }))
+    await click(page, clickable({ textContains: notebookName }))
+    await click(page, clickable({ text: 'Edit' }))
+    await select(page, 'Select Environment', 'Hail')
+    await click(page, clickable({ text: 'Create' }))
+    await findElement(page, clickable({ textContains: 'Creating' }))
+    await delay(300000) //creation takes about 5 minutes
```
ehanna4: move this timeout to the findElement util with an options

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,43 @@
+const { withRegisteredUser, withBilling, withWorkspace } = require('../utils/integration-helpers')
+const { click, clickable, signIntoTerra, findElement, waitForNoSpinners, select, delay, fillIn, input, findIframe, findText, dismissNotifications } = require(
+  '../utils/integration-utils')
+
+
+const notebookName = 'TestNotebook'
+
+const testRunNotebookFn = withRegisteredUser(async ({ page, context, email, token }) => {
```
ehanna4, line 8: consider flow for this mulitple wrapper handling

src/components/notebook-utils.js
```diff
@@ -153,8 +153,8 @@ export const NotebookCreator = class NotebookCreator extends Component {
       h(IdContainer, [id => h(Fragment, [
         h(FormLabel, { htmlFor: id, required: true }, ['Language']),
         h(Select, {
-          id, isSearchable: false,
-          placeholder: 'Select a language',
+          id, isSearchable: true,
+          placeholder: 'Select a language', 'aria-label': 'Select a language',
```
ehanna4: might not need the aria-label and could likely use the placehold or formlabel instead

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,52 @@
...
+    await click(page, clickable({ textContains: notebookName }))
+    await click(page, clickable({ text: 'Edit' }))
+    await select(page, 'Select Environment', 'Hail')
+    await click(page, clickable({ text: 'Create' }))
+    await findElement(page, clickable({ textContains: 'Creating' }))
+    await findElement(page, clickable({ textContains: 'Running' }), { timeout: 3000000 })
+    await delay(5000) //wait for copy over or change of status
+
+    const frame = await findIframe(page)
+    await fillIn(frame, '//textarea', 'print(123456789099876543210990+9876543219)')
```
ehanna4, line 30: I chose this obnoxiously long number so that is longer than both the test user id and the test workspace id so there won't be a false positive by either of those numbers

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,52 @@
...
+    await fillIn(page, input({ placeholder: 'Enter a name' }), notebookName)
+    await select(page, 'Language', 'Python 2')
+    await click(page, clickable({ text: 'Create Notebook' }))
+    await click(page, clickable({ textContains: notebookName }))
+    await click(page, clickable({ text: 'Edit' }))
+    await select(page, 'Select Environment', 'Hail')
+    await click(page, clickable({ text: 'Create' }))
+    await findElement(page, clickable({ textContains: 'Creating' }))
+    await findElement(page, clickable({ textContains: 'Running' }), { timeout: 3000000 })
+    await delay(5000) //wait for copy over or change of status
```
breilly2: Is this needed if we're waiting for the iframe below?

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,52 @@
...
+
+    const frame = await findIframe(page)
+    await fillIn(frame, '//textarea', 'print(123456789099876543210990+9876543219)')
+    await click(frame, clickable({ text: 'Run' }))
+    await findText(frame, '123456789099886419754209')
```
breilly2, line 32: Wow!

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,52 @@
...
+    await click(page, clickable({ textContains: 'Create a' }))
+    await fillIn(page, input({ placeholder: 'Enter a name' }), notebookName)
+    await select(page, 'Language', 'Python 2')
+    await click(page, clickable({ text: 'Create Notebook' }))
+    await click(page, clickable({ textContains: notebookName }))
+    await click(page, clickable({ text: 'Edit' }))
+    await select(page, 'Select Environment', 'Hail')
+    await click(page, clickable({ text: 'Create' }))
+    await findElement(page, clickable({ textContains: 'Creating' }))
+    await findElement(page, clickable({ textContains: 'Running' }), { timeout: 3000000 })
```
breilly2: I'm assuming you meant `300000` (5 minutes) rather than `3000000` (50 minutes)?
Nit: use `5 * 60 * 1000` to make the intended time interval easier to read.

integration-tests/utils/integration-helpers.js
```diff
@@ -72,9 +72,60 @@ const withUser = test => async args => {
...
+  await addUserToBilling({ email })
+
+  try {
+    await test({ ...args, email })
+  } finally {
+    await removeUserFromBilling({ email })
+  }
+}
+
+const withRegisteredUser = test => withUser(async ({ page, token, ...args }) => {
```
breilly2, line 146: For later: At mobbing, we'd talked about having `with*` functions make API calls instead of driving the UI. Then these steps would move back into the register-user test.

src/components/NewClusterModal.js
```diff
@@ -187,9 +187,10 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
 
     const makeEnvSelect = id => h(Select, {
       id,
+      'aria-label': 'Select Environment',
       value: selectedLeoImage,
       onChange: ({ value }) => this.setState({ selectedLeoImage: value }),
-      isSearchable: false,
+      isSearchable: true,
```
breilly2, line 196: TOL: If this improves a11y, then I guess this is fine. Otherwise, I don't understand why we need to make this change in order to be able to drive the UI with puppeteer. Inspecting the select options is difficult in the browser so I don't have alternative suggestions without spending more time with it.

integration-tests/utils/integration-utils.js
```diff
@@ -86,11 +86,11 @@ const dismissNotifications = async page => {
 const signIntoTerra = async (page, token = bearerToken) => {
   await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true })
   await waitForNoSpinners(page)
-  await page.evaluate(token => window.forceSignIn(token), token)
+  return page.evaluate(token => window.forceSignIn(token), token)
```
breilly2, line 86: TOL: Good catch. I think it worked before, but maybe it wasn't 100% "correct"? I'd like to learn more about this.

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,52 @@
...
+    await delay(5000) //wait for copy over or change of status
+
+    const frame = await findIframe(page)
+    await fillIn(frame, '//textarea', 'print(123456789099876543210990+9876543219)')
+    await click(frame, clickable({ text: 'Run' }))
+    await findText(frame, '123456789099886419754209')
+    await click(page, clickable({ textContains: 'Notebook Runtime' }))
+
+    //Clean up TODO: Add failsafe sweep up script
+    await click(page, clickable({ textContains: 'Delete Runtime' }))
```
breilly2: Nit: This clean-up should be attempted even if the test fails.

integration-tests/tests/run-notebook.js
```diff
@@ -0,0 +1,52 @@
+const { withRegisteredUser, withBilling, withWorkspace } = require('../utils/integration-helpers')
+const { testUrl } = require('../utils/integration-config')
+const { click, clickable, signIntoTerra, findElement, waitForNoSpinners, select, delay, fillIn, input, findIframe, findText, dismissNotifications } = require(
+  '../utils/integration-utils')
```
breilly2: Nit: Odd place for a newline...

breilly2 (dismissed):
👍 Generally looks good. I have a few comments but, other than the runtime startup timeout, everything else is optional. Nice work!

integration-tests/tests/run-notebook.js
ehanna4: Oh yikes, good catch thanks!

integration-tests/utils/integration-utils.js
ehanna4: My understanding it that it still worked but was actually taking a bit of unnecessary time. Awaiting the `page.evaluation` inside the function meant that the `signIntoTerra` function waited until that `page.evaluation` had a result and then since it's an async function returned the overall result wrapped in a promise so that in the integration tests we still had to `await signIntoTerra`, so essentially we end up double awaiting on the result of `page.evaluation`. Returning the `page.evaluation` instead directly passes that promise resolution back to the integration test which will await just once on the result there. (maybe @panentheos can weigh in if that is an accurate description of async/await though)

integration-tests/utils/integration-helpers.js
ehanna4: I will add this to a second PR of clean up and optimizing

integration-tests/utils/integration-utils.js
panentheos: Yes, that's right! In practice, any actual time overhead would be too tiny to matter, but it's still unnecessary so it seems like a good change. Note that this does change the signature of `signIntoTerra`, since its returned promise now resolves to something instead of resolving to `undefined`, but we weren't doing anything with the return value anyway so it should be ok.

integration-tests/utils/integration-helpers.js
```diff
@@ -72,9 +75,94 @@ const withUser = test => async args => {
...
+
+  await ajaxPage.goto(testUrl)
+  await click(ajaxPage, clickable({ textContains: 'View Workspaces' }))
+  await signIntoTerra(ajaxPage, token)
+  await dismissNotifications(ajaxPage)
+  await fillIn(ajaxPage, input({ labelContains: 'First Name' }), 'Integration')
+  await fillIn(ajaxPage, input({ labelContains: 'Last Name' }), 'Test')
+  await click(ajaxPage, clickable({ textContains: 'Register' }))
+  await click(ajaxPage, clickable({ textContains: 'Accept' }))
+  await ajaxPage.close()
```
panentheos, line 158: It's a little surprising that we click 'Accept' both here and also in the test cases. I suspect what might be happening is that the click happens, but the page is closed before the network request completes, so the TOS isn't actually accepted here. If so, we should an additional wait for some piece of page content to ensure that the TOS acceptance is recorded before we tear down the page.

integration-tests/utils/integration-helpers.js
```diff
@@ -72,9 +75,94 @@ const withUser = test => async args => {
...
+}
+
+const trimClustersOldestFirst = _.flow(
+  _.remove({ status: 'Deleting' }),
+  _.sortBy('createdDate')
+)
+
+const currentCluster = _.flow(trimClustersOldestFirst, _.last)
+
+const getCurrentCluster = withUserToken(async ({ email, token }) => {
```
panentheos, line 120: Does this need to create its own page to make the call, or could we just inline it into `deleteCluster`?

src/components/NewClusterModal.js
```diff
@@ -212,12 +213,13 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
 
     const makeGroupedEnvSelect = id => h(GroupedSelect, {
       id,
+      'aria-label': 'Select Environment',
```
panentheos: This select looks like it already has a properly wired up label, so we should be able to use that instead of adding an `aria-label`

integration-tests/utils/integration-helpers.js
ehanna4: Thank you for pointing that out - I should have called that out as it seemed wrong but I could not determine how to fix it. I will try making that adjustment and see if it resolves the need to double accept.

integration-tests/utils/integration-helpers.js
ehanna4: Let me double check, it was not working when I inlined it initially but I also made a few other changes which may have been the actual issue.

panentheos (approved):
Looks good, thanks for all the iteration on this.

PR #2042, opened 2020-02-13
Correct issue with sporadic registration test failures [SATURN-1395]

petesantos:
When evaluated in the running browser, the forceSignIn call was not being await-ed.
From puppeteer's perspective the function call was complete and it would continue along with the test. On occasion, the force sign in would not have completed before the next page load causing no email address to get pre-populated in the form.

Added async/await to the evaluate call to correct this.
Also added additional debugging information to the readme to allow setting up breakpoints in node/jest (helpful to prevent the window from closing and general debugging of tests)


integration-tests/utils/integration-utils.js
```diff
@@ -86,7 +86,7 @@ const dismissNotifications = async page => {
 const signIntoTerra = async (page, token = bearerToken) => {
   await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true })
   await waitForNoSpinners(page)
-  await page.evaluate(token => window.forceSignIn(token), token)
+  await page.evaluate(async token => await window.forceSignIn(token), token)
```
breilly2: This seemed like a good theory, but I'm not sure anymore. I attempted to test it by inserting delays during `forceSignIn` and could not reproduce the failure. Looking more closely at `forceSignIn`:
```js
window.forceSignIn = async token => {
  const res = await fetch(
    'https://www.googleapis.com/oauth2/v3/userinfo',
    { headers: { Authorization: `Bearer ${token}` } }
  )
  const data = await res.json()
  authStore.update(state => {
    return {
      ...state,
      isSignedIn: true,
      registrationStatus: undefined,
      isTimeoutEnabled: undefined,
      profile: {},
      user: {
        token,
        id: data.sub,
        email: data.email,
        name: data.name,
        givenName: data.given_name,
        familyName: data.family_name,
        imageUrl: data.picture
      }
    }
  })
}
```
The registration form isn't going to render until the sign-in state changes. Both the sign-in state and the email address that is pre-populated into the form are updated at the same time. I'm starting to wonder if we're not getting a success response from `userinfo`. In fact, the way that I found to reproduce the failure behavior was to give `userinfo` a bad token so I would definitely get an error response.

petesantos: Thanks. I spoke with Isaac to clear up some confusion I had regarding page.evaluate as well.
I opted to add error reporting to this function so we can get better visibility into the issue when it happens again

panentheos (dismissed):
Looks good, hopefully this will help catch errors on this path.

src/components/Notifications.js
```diff
@@ -99,7 +99,7 @@ const NotificationDisplay = Utils.connectStore(notificationStore, 'notificationS
           }, ['Details'])
         ]),
         h(Clickable, {
-          'aria-label': 'Dismiss notification',
+          'aria-label': type ? `Dismiss ${type} notification` : 'Dismiss notification',
```
panentheos, line 102: TOL: This is fine as a short-term fix, but going forward we should look into improving the accessibility of notifications, which will also let us write a more natural selector in the tests.

panentheos (approved):

PR #2044, opened 2020-02-13
sort entity list in data tab

zarsky-broad:
Verified in browser


breilly2 (commented):
👍 Looks good. I tested on the PR site.
I ran the known run-workflow test failure locally with the following diff to verify that the test succeeds.
```
diff --git a/src/libs/ajax.js b/src/libs/ajax.js
index dd546e95..98c42118 100644
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -558,7 +558,8 @@ const Workspaces = signal => ({
               jsonBody({
                 ...payload,
                 methodConfigurationNamespace: configNamespace,
-                methodConfigurationName: configName
+                methodConfigurationName: configName,
+                deleteIntermediateOutputFiles: false
               }),
               { signal, method: 'POST' }
             ]))
```

breilly2 (approved):
Hmm... looks like I accidentally commented instead of approved. Here's the formal approval.

PR #2045, opened 2020-02-13
anonymous id and more info for mixpanel [SATURN-1366]

zarsky-broad:
Verified by logging out data in browser that would be sent.

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -4,6 +4,7 @@ import { h } from 'react-hyperscript-helpers'
 import { version } from 'src/data/clusters'
 import { getUser } from 'src/libs/auth'
 import { getConfig } from 'src/libs/config'
+import { withErrorIgnoring } from 'src/libs/error'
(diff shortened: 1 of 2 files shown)


src/libs/ajax.js
```diff
@@ -1086,8 +1087,19 @@ const Duos = signal => ({
...
-  captureEvent: (event, data) => metricsEnabled && fetchMetrics('api/event', _.mergeAll([authOpts(), jsonBody({ event, data }), { signal, method: 'POST' }]))
+  captureEvent: (event, details) => {
+    const body = {
+      event,
+      data: {
+        ...details,
+        userId: authStore.get().profile.anonymousGroup,
+        appId: window.location.hostname,
+        appPath: Nav.getCurrentRoute().name
+      }
```
petesantos, line 1100: Could you add current time?

petesantos (approved):
Looks good, see comment. Tested by sharing a workspace and checking the bard log for the data

PR #2048, opened 2020-02-14
allow json objects that aren't rawls lists [SATURN-1223]

zarsky-broad:
verified in-browser by simulating arbitrary object as attribute


breilly2 (approved):
👍 Looks good. I tested on the PR site by copying a record from the workspace mentioned in SATURN-1223 into a scratch workspace in dev (which I've since deleted) and verifying that it renders properly without crashing.

PR #2050, opened 2020-02-18
Refactor injection of user token into tests [SATURN-1341]

breilly2:
Refactor prior to standing up a test running service. Allows tests to be self-sufficient in getting an auth token for the pre-configured test user.

All tests still pass when running locally.


integration-tests/README.md
```diff
@@ -15,24 +15,27 @@ TERRA_TOKEN=[bearer token] LYLE_SA_KEY=[full key json] yarn test
 #### Configuration
 The following environment variables are parsed by the tests:
 
-1. `TERRA_TOKEN` **(required)**: bearer token used to log into Terra.
-2. `TEST_URL`: URL for the ui.
+1. `TERRA_SA_KEY` **(required\*)**: service account key for creating access tokens for test user.
+1. `TERRA_TOKEN` **(required\*)**: bearer token used to log into Terra.
```
breilly2: I was going to have to change all of the numbers anyway, so I'm making them all `1.` and letting markdown number them correctly so that future diffs can be smaller.

.circleci/config.yml
```diff
@@ -100,7 +100,6 @@ commands:
             BILLING_PROJECT: << parameters.billing_project >>
           command: |
             mkdir -p ${SCREENSHOT_DIR}
-            TERRA_TOKEN=$(node scripts/getBearerToken.js "${ALPHA_TEST_UA}" "${ALPHA_USER_SA_KEY_JSON}") \
```
breilly2, line 103: @panentheos I was able to remove this line, as you suggested, by adding the service account key to the circle environment with the right name.

integration-tests/scripts/getBearerToken.js
```diff
@@ -22,4 +14,4 @@ const getToken = async () => {
   }
 }
 
-getToken()
+main()
```
panentheos: Now that circle no longer uses this file, I think we can delete it.

integration-tests/utils/terra-sa-utils.js
```diff
@@ -0,0 +1,28 @@
...
+
+const withUserToken = testFn => async options => {
+  const token = bearerToken || await getToken(userEmail, JSON.parse(terraSaKeyJson))
+  return testFn({ ...options, token })
+}
+
+module.exports = {
+  getToken,
+  withUserToken
+}
```
panentheos, line 28: TOL: We're accumulating a lot of different 'utility' files, which could probably be consolidated at some point.

integration-tests/utils/terra-sa-utils.js
```diff
@@ -0,0 +1,28 @@
+const _ = require('lodash/fp')
+const { JWT } = require('google-auth-library')
+const { bearerToken, userEmail, terraSaKeyJson } = require('../utils/integration-config')
+
+
+const makeAuthClient = _.memoize((subject, { client_email, private_key }) => {
```
panentheos, line 6: `_.memoize` uses the first argument to the function as a cache key. In this case, since we always pass the same subject, it will only do this once. It might be clearer to use `_.once` to make that explicit.

panentheos (approved):
Looks good, see comments.

integration-tests/scripts/getBearerToken.js
breilly2: Interesting... you may be right. @zarsky-broad, do you have any opinions on this?

integration-tests/utils/terra-sa-utils.js
breilly2: That is true if we delete `getBearerToken.js`... and may remain true as this moves to a server context? I think `_.memoize` is the right thing as long as `getToken` has a `subject` parameter. If we find that we are always using a globally configured subject in all contexts, then I agree that this should be `_.once`. I'll keep an eye on that.

integration-tests/utils/terra-sa-utils.js
breilly2: These functions could probably move into `integration-helpers.js`. I was following `lyle-utils.js` as an example.

I like these being separate:
* `integration-utils.js` as higher-level functions on top of puppeteer
* `integration-helpers.js` as Terra-specific functions for use by tests (maybe it should be renamed to `terra-utils.js`)
* `firecloud-utils.js` as a separate thing that we can wholesale delete when we no longer have to drive the FireCloud UI
* `jest-utils.js` to keep jest dependencies separate and not deployed in the test service

integration-tests/README.md
zarsky-broad: I'd kind of like them all to be labeled 0, in that case, since it's clearer when reading the unrendered version, as that's not a number we'd use.

zarsky-broad: (Either that, or just make it an unordered list, since order doesn't matter)

integration-tests/scripts/getBearerToken.js
zarsky-broad: I think I agree.

integration-tests/README.md
breilly2: Oh, both good calls. I'm going to go with unordered list.

PR #2051, opened 2020-02-18
Update checkboxes to new design [SATURN-1257]

zarsky-broad:
Tested in browser.

Seen and okayed by product & UX.


src/components/common.js
```diff
@@ -186,20 +187,20 @@ export const MenuButton = ({ disabled, children, ...props }) => {
 }
 
 export const Checkbox = ({ checked, onChange, disabled, ...props }) => {
+  const getStateColor = (ifDisabled, ifChecked, ifDefault) => disabled ? ifDisabled : checked ? ifChecked : ifDefault
```
panentheos: nit: Since 2/3 cases below only use part of this logic, I suspect it might be clearer to spell things out inline with ternary operators or `cond`, even if the end result is slightly longer.

panentheos (commented):
TOL: I'm concerned that this approach is a step backwards from a usability and accessibility perspective:
* Without the dropdown arrow, the new control looks like a text box, but behaves like a select.
* The selection gesture doesn't actually change the select value, but instead resets it and adds an item to the list as a side-effect.
* If you already have a collaborator list that's longer than the scrolling content area, the addition of a new item isn't obvious, so it ends up looking like the select is broken.

petesantos:
> TOL: I'm concerned that this approach is a step backwards from a usability and accessibility perspective:
> 
> * Without the dropdown arrow, the new control looks like a text box, but behaves like a select.
> * The selection gesture doesn't actually change the select value, but instead resets it and adds an item to the list as a side-effect.
> * If you already have a collaborator list that's longer than the scrolling content area, the addition of a new item isn't obvious, so it ends up looking like the select is broken.

Isaac, Brian and I reviewed the UX of this PR in the pod on Friday. Overall, we had agreement that the UX "felt" good. 

> * Without the dropdown arrow, the new control looks like a text box, but behaves like a select.

This is similar to what other apps do (e.g. google calendar) where you click the text box to search for a person, and recent people will automatically show up as suggestions. Once you click a person, they get added to the list and the selection starts over. In terms of A11y, when a user is selected a screen reader reads out "<user> added to event"

> * If you already have a collaborator list that's longer than the scrolling content area, the addition of a new item isn't obvious, so it ends up looking like the select is broken.

On this one, Isaac added a scrolling feature to scroll the most recently added name into view

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -108,72 +108,69 @@ export default ajaxCaller(class ShareWorkspaceModal extends Component {
...
+        h(CreatableSelect, {
           id,
+          isLoading: !loaded,
           placeholder: 'Add people or groups',
-          value: searchValue,
-          onChange: v => this.setState({ searchValue: v }),
-          suggestions: _.difference(suggestions, _.map('email', acl)),
-          style: { fontSize: 16 }
+          'aria-label': 'Enter an email address of a person or group',
+          isValidNewOption: val => !validate.single(val, { email: true, exclusion: aclEmails }),
```
petesantos: There is a bug that crashes the app in some cases when entering/validating email addresses

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -108,72 +108,69 @@ export default ajaxCaller(class ShareWorkspaceModal extends Component {
     }, [
       h(IdContainer, [id => h(Fragment, [
         h(FormLabel, { htmlFor: id }, ['User email']),
-        h(AutocompleteTextInput, {
+        h(CreatableSelect, {
```
petesantos: What can we do to utilize a text input rather than a select? What a11y impacts does this have?
Would a combobox aria-label work for a select component?

petesantos (commented):
Review from mobbing

breilly2:
Nit: The focus ring padding is a little uneven.
Screen Shot 2020-03-20 at 2 27 13 PM

breilly2 (approved):
👍 Other than the focus ring being a little uneven, looks good! Once I understood it, I kinda liked the composition of elements making up the checkbox states.

PR #2052, opened 2020-02-18
[IA-1508] Fix canUpdate bug with a new cluster that has a custom environment

jdcanas:

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -493,15 +493,17 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
         ])
       ])],
       ['update', () => h(Fragment, [
-        this.isStopRequired() ? p([
-          'Changing the machine type (increasing or decreasing the # of CPUs or Mem) results in an update that requires a ',
-          b(['restart']),
-          ' of your runtime. This may take a few minutes.  Would you like to proceed? ',
-          b(['(You will not lose any files.)'])
-        ]) : p([
-          'Increasing the disk size or changing the number of workers (when the number of workers is >2) results in a real-time update to your runtime. ',
-          'During this update, you can continue to work'
-        ]),
+        this.isStopRequired() ?
+          p([
+            'Changing the machine type (increasing or decreasing the # of CPUs or Mem) results in an update that requires a ',
+            b(['restart']),
+            ' of your runtime. This may take a few minutes.  Would you like to proceed? ',
+            b(['(You will not lose any files.)'])
+          ]) :
+          p([
+            'Increasing the disk size or changing the number of workers (when the number of workers is >2) results in a real-time update to your runtime. ',
+            'During this update, you can continue to work'
+          ]),
         div({ style: { display: 'flex', justifyContent: 'flex-end', marginTop: '1rem' } }, [
           h(ButtonSecondary, {
             style: { marginRight: '2rem' },
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -153,7 +153,6 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
   }
 
   updateCluster() {
-    console.log('in update cluster')
     const { currentCluster } = this.props
     const { googleProject, clusterName } = currentCluster
 
@@ -169,8 +168,6 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
       currentCluster ? Ajax().Clusters.cluster(currentCluster.googleProject, currentCluster.clusterName).details() : null,
       Ajax().Buckets.getObjectPreview('terra-docker-image-documentation', 'terra-docker-versions.json', namespace, true).then(res => res.json())
     ])
-
-    console.log('Current cluster details: ', currentClusterDetails)
 
     this.setState({ leoImages: newLeoImages })
     if (currentClusterDetails) {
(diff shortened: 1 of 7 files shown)


jdcanas:
to make it easy to see what was actually changed with this fix post-approval of the last PR, i'd recommend `git diff dda5eec..fcb6e78`. It was two lines.

src/components/NewClusterModal.js
```diff
@@ -405,6 +482,26 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+      ['update', () => h(Fragment, [
+        this.isStopRequired() ?
+          p([
+            'Changing the machine type (increasing or decreasing the # of CPUs or Mem) results in an update that requires a ',
+            b(['restart']),
+            ' of your runtime. This may take a few minutes.  Would you like to proceed? ',
+            b(['(You will not lose any files.)'])
+          ]) :
+          p([
+            'Increasing the disk size or changing the number of workers (when the number of workers is >2) results in a real-time update to your runtime. ',
```
ehanna4: TOL: This message is shown when there are no changes to a cluster as well. This is confusing since the language only speaks to changing/increasing things and is even more confusing when you recreate a cluster that had <2 workers (since this message specifically calls out >2 workers). I think it would be beneficial to consider adding language to cover the recreate case in here or perhaps disabling the 'Update' button if there are no changes being made to the cluster

ehanna4: TOL: This message is shown when there are no changes to a cluster as well. This is confusing since the language only speaks to changing/increasing things and is even more confusing when you recreate a cluster that had <2 workers (since this message specifically calls out >2 workers). I think it would be beneficial to consider adding language to cover the recreate case in here or perhaps disabling the 'Update' button if there are no changes being made to the cluster

jdcanas: I added a `hasChanges` function that I've tested to disable the button if there are no changes as you suggest, but I would like to do it as a separate PR because otherwise identifying the changes after the previous PR approval would be unwieldy

ehanna4: that seems reasonable - I had a few other non-blocking comments but can hold those for a future PR as well. I would like to do a bit more testing but with the power out I don't have the current laptop power so will get back to that ASAP

breilly2 (approved):
I performed more thorough testing this time. Special thanks to @ehanna4 for help with testing.

- [x] Created a runtime using a custom docker image when starting with no cluster. (source of previous bug)
- [x] Updated between spark cluster and non-spark cluster, causing cluster replace
- [x] Live-updated a runtime with an increased disk size
- [x] Live-updated a spark cluster by changing number of workers
- [x] Updated CPU and/or memory of a running cluster, causing a stop/start cycle
- [x] Updated the environment/image of a running runtime, causing cluster replace
- [x] Updated the environment/image of a stopped runtime, causing cluster replace (and auto start... somewhat inconsistent with changing live-updatable parameters of a stopped runtime)
- [x] Updated CPU and/or memory of a stopped cluster, causing an immediate update (cluster remained stopped)

There are some oddities with wording. For example, when changing the machine type of a stopped cluster, it says that the change can be made without disrupting work. That's sort of true because there isn't any work to interrupt, but it's odd to me that it's the same message as when increasing the disk size of a running cluster. We should continue to refine the language to improve UX, but the functionality is consistent with expected behavior according to the flow diagram.

jdcanas:
>  Updated the environment/image of a stopped runtime, causing cluster replace (and auto start... somewhat inconsistent with changing live-updatable parameters of a stopped runtime)

After discussion with Adrian, this approach seems a bit risky cost-wise when it is avoidable. Dataproc clusters are `Running` by default after creation. Leo provides a flag for `stopAfterCreation` in the API, it may be worth leveraging this for `Replace`

src/components/NewClusterModal.js
```diff
@@ -149,6 +150,106 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+    const isMasterMachineTypeChanged = currentClusterConfig.masterMachineType !== userSelectedConfig.masterMachineType
+
+    const isClusterRunning = currentCluster.status === 'Running'
+
+    return this.canUpdate() && isMasterMachineTypeChanged && isClusterRunning
+  }
+
+  getRunningUpdateText() {
+    return this.isStopRequired() ?
+      p([
```
ehanna4, line 238: nit - This should have 1.5 line height

ehanna4 (approved):
Manually tested a lot of cases including:

-Changing CPU and Mem size to get a restart
 ^ also did this while working in a notebook and was appropriately switched to read-only mode
-Created a runtime using a custom docker image when starting with no cluster
-Changed number of workers while running and appropriately live updated
-Changed disk size while running and appropriately live updated
-Made no changes and saw improved tooltip and disabled button
-Updated a stopped runtime resulting in a replace

Overall looks good, thanks for making those wording changes! Minor comment on the line height for the warning messages.

src/components/NewClusterModal.js
```diff
@@ -152,6 +153,106 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+        'Changing the machine type (increasing or decreasing the # of CPUs or Mem) results in an update that requires a ',
+        b(['restart']),
+        ' of your runtime. This may take a 3-5 minutes. Would you like to proceed? ',
+        b(['(You will not lose any files.)'])
+      ]) :
+      p([
```
ehanna4, line 244: This should also get 1.5 line height

jdcanas: Hmm... I'm going to leave it because there are other p() tags that don't have it, and I'm unsure if the desire is for all of them to have it, or just the warning ones, which I feel is another discussion and would require a bit more refactoring.

ehanna4: It's not so much a p() thing as it is that when there is a block of text the line height should be 1.5 as 1.5 line height is the general accepted accessible line height and with the accessibility audit impending it's something we are trying to follow

PR #2053, opened 2020-02-20
remove table virtualization from workflow inputs

panentheos:
This converts the workflow IO table into a non-virtualized one. This will enable better handling of multi-line inputs, which will be coming up in a separate PR. Based on a sampling of existing workflows, we don't expect a large number of inputs, so this shouldn't be a performance concern.

Because the new table is not height-constrained, it will not scroll internally, which changes the UX of the page. However, users reported being confused by the internal scrolling, so this should be an improvement.

This PR also fixes a minor visual issue with the file selector button.

Tested locally by interacting with the table.

### DIFF
--- a/src/components/table.js
+++ b/src/components/table.js
@@ -263,10 +263,10 @@ export class FlexTable extends Component {
 
 /**
  * A basic table with a header and flexible column widths. Intended for small amounts of data,
- * since it does not provide scrolling.
+ * since it does not provide scrolling. See FlexTable for prop types.
  */
 export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, hoverHighlight }) => {
-  return div([
+  return h(Fragment, [
     div({ style: { height: 48, display: 'flex' } }, [
       _.map(([i, { size, headerRenderer }]) => {
         return div({
--- a/src/components/table.js
+++ b/src/components/table.js
@@ -262,6 +262,41 @@ export class FlexTable extends Component {
 }
 
 /**
+ * A basic table with a header and flexible column widths. Intended for small amounts of data,
+ * since it does not provide scrolling.
+ */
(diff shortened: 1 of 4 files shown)


zarsky-broad (commented):
Some observations from interacting:
1. When clicking on an input in the table, or tabbing from one to another, it takes a moment before my cursor moves, especially on a longer table, going from one cell showing an autocomplete to another.
2. When switching between inputs and outputs, I occasionally get an error in the console about updating state on an unmounted component, but I haven't yet tracked that down...

panentheos:
I think both of those issues stem from Downshift emitting `onChange` calls when the input loses focus, even if nothing has actually changed. I have a diff that's part of my other PR that changes that.

panentheos:
I tracked down and fixed the perf issues, this is ready for another look.

src/components/table.js
```diff
@@ -261,6 +261,41 @@ export class FlexTable extends Component {
   }
 }
 
+/**
+ * A basic table with a header and flexible column widths. Intended for small amounts of data,
+ * since it does not provide scrolling.
+ */
+export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, hoverHighlight }) => {
+  return div([
```
zarsky-broad:
```suggestion
  return h(Fragment, [
```

src/components/table.js
```diff
@@ -261,6 +261,41 @@ export class FlexTable extends Component {
   }
 }
 
+/**
+ * A basic table with a header and flexible column widths. Intended for small amounts of data,
+ * since it does not provide scrolling.
+ */
+export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, hoverHighlight }) => {
```
breilly2, line 268: PropTypes for columns (as discussed in mob)

breilly2 (approved):
👍 LGTMob!

PR #2054, opened 2020-02-20
Render workspace list page as table [SATURN-10]

zarsky-broad:
(Not merging until after India workshop ends, anyway)

- [x] Lou
- [x] Adrian
- [x] Comms notified

Data:
b.adm.firec has access to over 1400 workspaces on alpha.
Time to render on current alpha is about 27 seconds.
Time to render on this branch is about 13 seconds, which is less than a second longer than it took for the ajax request to complete!


ehanna4:
(Haven't reviewed just wanted to mention) The current super long time that it takes Alpha to load the workspaces list will cause intermittent failure in the tests, cleaning that up to load faster seems like a really good way to go since that could be something a user with tons of workspaces would experience.

zarsky-broad:
Tweaked and ready for re-review

panentheos (approved):
Looks good, thanks!

PR #2056, opened 2020-02-23
send event on every page view [SATURN-1327]

zarsky-broad:
Sends events only on route change, and only when signed in.


src/libs/events.js
```diff
@@ -1,3 +1,26 @@
...
+  workspaceShare: 'workspace:share',
+  pageView: 'page:view'
 }
+
+export const PageViewReporter = () => {
+  const { name } = useRoute()
+  const { isSignedIn } = Utils.useStore(authStore)
+
+  useEffect(() => {
+    if (isSignedIn) {
```
panentheos: This should be checking `registrationStatus === 'registered'` to be sure the user is in Sam.

panentheos (approved):
Looks good, see comment. Tested by clicking around locally.

PR #2057, opened 2020-02-23
allow workflow launch errors to wrap on periods [SATURN-1355]

zarsky-broad:
Verified in browser


src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -69,6 +69,7 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
       [_.isArray(entities), () => _.uniq(entities).length],
       () => _.size(entities)
     )
+    const wrappableOnPeriods = _.flatMap(char => char === '.' ? [wbr(), '.'] : char)
```
panentheos: TOL: This might potentially create a very large array, which I'd imagine creates extra work during rendering. Perhaps we could do something like use `split` to keep the non-period substrings intact to limit that.

panentheos (approved):
Looks good, see comment. Tested locally by mocking error messages.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
zarsky-broad: What do you think of this?

PR #1824, opened 2019-08-25
remove deprecated notification methods

zarsky-broad:
Also fixes an instance of NaN in the UI that I encountered while testing this.

PR #2058, opened 2020-02-23
update notifications to new styles

zarsky-broad:
In preparation for [an upcoming ticket](https://broadworkbench.atlassian.net/browse/SATURN-1276), update notifications to new colors and simplify code, including pulling out a bunch of stuff that isn't (was never? unclear.) needed.

Tested as extensively as I could in-browser.


src/libs/notifications.js
```diff
@@ -40,7 +32,7 @@ export const notify = (type, title, props) => {
   return notification.id
 }
 
-export const clearNotification = id => store.removeNotification(id)
+export const clearNotification = store.removeNotification
```
panentheos: Because this is implemented as a 'method' under the hood, this doesn't work properly (it breaks the `this` binding)

panentheos (commented):
Looks good, but there are couple of places where there are light colored links that are hard to read against the new backgrounds.
* 'Please Update Your Runtime' in ClusterManager
* ShibbolethLink in auth
* 'Cannot Edit Notebook' in NotebookLauncher
* 'Workspace access synchronizing' in WorkspaceContainer

src/libs/notifications.js
```diff
@@ -58,13 +50,14 @@ const NotificationDisplay = Utils.connectStore(notificationStore, 'notificationS
...
+      [type === 'success', () => colors.success],
+      [type === 'info', () => colors.accent],
+      [type === 'welcome', () => colors.accent],
+      [type === 'warn', () => colors.warning],
+      [type === 'error', () => colors.danger],
+      () => colors.accent
+    )
     const iconType = Utils.cond(
       [type === 'success', 'success-standard'],
       [type === 'warn' || type === 'error', 'warning-standard'],
```
breilly2: Mocks show error as an exclamation point in a circle while warning is an exclamation point in a triangle.

src/components/ServiceAlerts.js
```diff
@@ -2,9 +2,9 @@ import _ from 'lodash/fp'
 import { Fragment, useEffect, useState } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
 import { Link } from 'src/components/common'
-import { clearNotification, notify } from 'src/components/Notifications'
 import { Ajax } from 'src/libs/ajax'
 import { withErrorIgnoring } from 'src/libs/error'
+import { clearNotification, notify } from 'src/libs/notifications'
 import * as Utils from 'src/libs/utils'
```
breilly2, line 10: The "read more" link below (lines 34-35) is white which is now poor color contrast with the background.

breilly2 (approved):
👍 Overall looks good. I tested info notifications (NIH account linking notice) with different themes. I tested warning notifications by simulating system alerts (see comment... this should be addressed pre-merge). I tested error notifications by simulating an error fetching storage cost estimate (see comment).

Looks like we lost notification grouping a while back (https://github.com/DataBiosphere/terra-ui/pull/1824). Therefore, although the mocks show what this should look like, I was unable to test it. Did we mean to remove this feature?

zarsky-broad:
Grouping is still there, but only for multiples of the same message with the same id. This doesn't really happen often, since most id's are dynamic, but it's a capability we retain. I've addressed the other feedback, so I'm going to go ahead and merge.

breilly2:
It used to also group by type/title.

PR #2059, opened 2020-02-23
update deps

zarsky-broad:
1. Update deps. The only vaguely interesting change is that now source map explorer can display gzipped sizes.
2. While testing, I noticed that the labels on the paginator buttons kept shifting, so I fixed that.
3. Stop caching the `~/.cache/yarn` folder on CI. This folder never gets emptied out, it just grows with every new installed version, whereas the installed node_modules folder only contains the versions of packages used currently. As a result, this shrinks the cache that has to be restored for every step, and hopefully the amount of time it takes to restore it as well.

Tested in-browser.


ehanna4:
Hi just wanted to mention there will be a GATK workshop occurring next Monday (3/2/20) through Thursday (3/5/20) so per our agreement from the post mortem about dependency updates this should not be merged during those date. Thanks

panentheos (approved):
Looks good. Tested locally.

PR #2061, opened 2020-02-24
Add "delete intermediate output files" option to submissions [WA-110]

rsasch:
There is a link to a stub for documentation, which should probably be complete before release.

Closes https://broadworkbench.atlassian.net/browse/WA-110


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -724,7 +725,21 @@ const WorkflowView = _.flow(
               disabled: currentSnapRedacted || !!Utils.computeWorkspaceError(ws),
               checked: useCallCache,
               onChange: v => this.setState({ useCallCache: v })
-            }, [' Use call caching'])
+            }, [' Use call caching']),
+            span({ style: { marginLeft: '1rem', marginRight: '0.5rem' } }, [
```
ehanna4: nit: If there are multiple margin styles to apply I think the preferred convention is to use `margin: '0 0.5rem 0 1rem'` (which applies the margins clockwise with 0 for top, 0.5 for right, 0 for bottom, and 1 for left)

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -724,7 +725,21 @@ const WorkflowView = _.flow(
...
+                checked: deleteIntermediateOutputFiles,
+                onChange: v => this.setState({ deleteIntermediateOutputFiles: v }),
+                style: { marginLeft: '1rem' }
+              }, [' Delete intermediate outputs'])
+            ]),
+            h(InfoBox, [
+              'If the workflow succeeds, only the final output will be saved. Subsequently, call caching cannot be used as the intermediate steps will be not available. ',
+              h(Link, {
+                href: 'https://support.terra.bio/hc/en-us/articles/360039681632',
+                target: 'blank'
```
ehanna4: Rather than` target: 'blank' `there's a util that has been extracted out which can be applied with: `..Utils.newTabLinkProps,` which also specifies the 'rel' attribute when opening a link

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -724,7 +725,21 @@ const WorkflowView = _.flow(
...
+            }, [' Use call caching']),
+            span({ style: { marginLeft: '1rem', marginRight: '0.5rem' } }, [
+              h(LabeledCheckbox, {
+                checked: deleteIntermediateOutputFiles,
+                onChange: v => this.setState({ deleteIntermediateOutputFiles: v }),
+                style: { marginLeft: '1rem' }
+              }, [' Delete intermediate outputs'])
+            ]),
+            h(InfoBox, [
+              'If the workflow succeeds, only the final output will be saved. Subsequently, call caching cannot be used as the intermediate steps will be not available. ',
```
panentheos, line 741: If call caching is mutually exclusive with this option, we should enforce that in the UI somehow, maybe by clearing and disabling the call caching checkbox if you select this option.

rsasch: A workflow can take advantage of previous tasks (calls) whose outputs were cached and also delete it's own intermediate files, so the checkboxes are not mutually exclusive.

src/pages/workspaces/workspace/workflows/WorkflowView.js
rsasch: fixed!

src/pages/workspaces/workspace/workflows/WorkflowView.js
rsasch: changed — thanks!

panentheos (approved):
Looks good. Tested by submitting a workflow and verifying that the flag was passed.

PR #2062, opened 2020-02-26
fix doc link

zarsky-broad:
New link per Tiffany


breilly2 (approved):
👍 I didn't test the notification in Terra but I did test the link. The old one is a 404 and the new one is a useful and relevant document.

PR #2063, opened 2020-02-27
Fix link to gen3 [no JIRA]

breilly2:
Fixing this avoids warnings due to certificate mismatches (or something like that).

Related to, but does not complete, SATURN-1417 (https://broadworkbench.atlassian.net/browse/SATURN-1417)


petesantos (approved):
Could not test, will test when I get home.
Looks good.

PR #2060, opened 2020-02-24
Inject config into tests

breilly2:
*EDIT*: I may be reversing course on this. I attempted to keep refactoring separate from adding the service itself to reduce the complexity of the ultimate PR, but I have not done a great job making the right refactoring decisions ahead of time.

Another piece of refactoring for [SATURN-1341](https://broadworkbench.atlassian.net/browse/SATURN-1341)

PR #2064, opened 2020-02-28
Test runner service [SATURN-1341]

breilly2:
https://broadworkbench.atlassian.net/browse/SATURN-1341

First, this includes a refactor of config to extract the things that change (or could change) per environment. When running with jest, these come from environment variables. When running with the service, they are determined from a `targetEnv` query parameter. Either way, the values are passed to the test in the single object argument. The environments are defined in `terra-envs.js`.

That's just about every file except for `server.js` (and the usual `utils.js`). I haven't yet deployed this to GAE. However, I have run all of the existing tests through it, during which I discovered that express, by default, terminates connections after 2 minutes so I had to do a little extra work to extend that.

This PR partly undoes (and replaces) #2060.


zarsky-broad:
Note that the billing accounts have recently changed, see the updated circle config file.

breilly2:
I merged dev into this branch and updated billing projects accordingly.

integration-tests/src/server.js
```diff
@@ -0,0 +1,64 @@
...
+
+const getBrowser = _.once(() => puppeteer.launch())
+const getContext = async () => {
+  const browser = await getBrowser()
+  return browser.createIncognitoBrowserContext()
+}
+
+const testTimeout = async (context, timeout) => {
+  await delay(timeout)
+  context.close()
```
panentheos: We need to do this on both success and timeout, so perhaps this should be in a `finally` within the `try` below?

integration-tests/src/server.js
```diff
@@ -0,0 +1,64 @@
...
+const { testRegisterUser } = require('../tests/register-user')
+const { testRunWorkflow } = require('../tests/run-workflow')
+const { defaultTimeout } = require('../utils/integration-helpers')
+const { delay, withScreenshot } = require('../utils/integration-utils')
+const envs = require('../utils/terra-envs')
+
+
+const app = express()
+const server = http.createServer(app)
+server.setTimeout(defaultTimeout + 60 * 1000) // allow an extra minute for teardown
```
panentheos: Since some tests take longer than the default, this should probably be set to some larger upper bound.

integration-tests/src/server.js
```diff
@@ -0,0 +1,64 @@
...
+    const context = await getContext()
+    const page = await context.newPage()
+    try {
+      const result = await Promise.race([
+        withScreenshot(name)(fn)({ context, page, ...targetEnvParams }),
+        testTimeout(context, timeout)
+      ])
+      return new Response(200, result)
+    } catch (e) {
+      return new Response(200, JSON.stringify({
```
panentheos: nit: We don't need to stringify this manually, objects should get sent as JSON automatically.

integration-tests/utils/terra-envs.js
```diff
@@ -0,0 +1,27 @@
...
+    testUrl: 'https://bvdp-saturn-dev.appspot.com',
+    workflowName: 'echo_to_file'
+  },
+  alpha: {
+    billingProject: 'saturn-integration-test-alpha',
+    testUrl: 'https://bvdp-saturn-alpha.appspot.com',
+    workflowName: 'echo_to_file'
+  },
+  perf: {
+    billingProject: '???',
```
panentheos, line 18: Is this not set up yet?

integration-tests/src/server.js
breilly2: Oof... good point. I really wanted to set this for specific endpoints but I couldn't find any way to ask express to do that (at least not any way that worked). I had to do a bunch of searching to find this solution of going down a level and setting the timeout there.

integration-tests/src/server.js
breilly2: I'll experiment. There's some trickiness to making sure the context is closed at the right time. In truth, this isn't quite the right time because we don't currently get a screenshot when the test times out like this.

integration-tests/src/server.js
breilly2: I'll experiment. I didn't get much of anything useful when including `e` in the response. `JSON.stringify` might have been from an earlier attempt to do that which ultimately led to me constructing this object that manually extracts `name`, `message`, and `stack` from the exception.

integration-tests/src/server.js
breilly2: I finally remembered that this was more about extending the default 2-minute timeout.

integration-tests/utils/terra-envs.js
breilly2: Not that I saw in the CircleCI config.

integration-tests/utils/terra-envs.js
```diff
@@ -0,0 +1,27 @@
+module.exports = {
+  local: {
+    billingProject: 'saturn-integration-test-dev',
+    testUrl: 'http://localhost:3000',
+    workflowName: 'echo_to_file'
+  },
+  dev: {
+    billingProject: 'saturn-integration-test-dev',
+    testUrl: 'https://bvdp-saturn-dev.appspot.com',
+    workflowName: 'echo_to_file'
```
panentheos, line 10: TOL: Going forward we should think about which things should typically be allowed to vary between environments, and which things are expected to be standard.

panentheos (approved):
Looks good. Thanks for putting this together!

integration-tests/utils/terra-envs.js
breilly2: I agree when it comes to `workflowName`. I don't remember exactly why I put it here but not other things like `userEmail`. In both cases (and possibly others), they are things that need to be set up intentionally in each environment, however in practice the values are the same in all environments.

PR #2065, opened 2020-02-28
allow multi-line editing of workflow inputs [SATURN-1059]

panentheos:
This converts workflow inputs into resizable textareas, and adds a button to pretty-print existing JSON, if applicable.

### DIFF
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -142,6 +142,7 @@ const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange,
           const value = config[which][name] || ''
           const error = errors[which][name]
           const isFile = (inputType === 'File') || (inputType === 'File?')
+          const formattedValue = JSON.stringify(Utils.maybeParseJSON(value), null, 2)
           return div({ style: { display: 'flex', alignItems: 'center', width: '100%', paddingTop: '0.5rem', paddingBottom: '0.5rem' } }, [
             div({ style: { flex: 1, display: 'flex', position: 'relative', minWidth: 0 } }, [
               !readOnly ? h(DelayedAutocompleteTextArea, {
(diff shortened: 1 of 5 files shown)


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -158,6 +159,16 @@ const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange,
...
             ]),
+            !readOnly && h(Link, {
+              style: { marginLeft: '0.25rem' },
+              onClick: () => {
+                const json = Utils.maybeParseJSON(value)
+                if (json !== undefined) {
+                  onChange(name, JSON.stringify(json, null, 2))
+                }
+              },
+              tooltip: 'Format'
```
ehanna4: TOL: It was confusing for me to know what the format button would do and if it worked since it's always clickable even when the type is an int or a float or nothing is in there. A few thoughts: 
- If there could be some indication after clicking it that something did or did not happen as a result I feel like that would be nice for the user.
- Since this is only for JSON formatting I wonder if a more specific tool tip would make sense? Something like 'format JSON' ? 
- Could the button be disabled or maybe not even rendered if the input field is empty?

panentheos: For the wording, I wrestled with that a bit. While it is technically JSON, I don't know if we call that out specifically, and a non-programmer might be confused by it. I also considered 'Format struct', and while that would certainly make sense to a WDL author, I'm not sure if that would be clear to someone just running it.

Regarding disabling/feedback, it's a good point. I had originally skipped that due to worries about performance (since we'd need to pre-parse all the inputs to generate a good disabled flag), but now that we've tightened things up a bit, maybe we can revisit that.

ehanna4: Oh okay that's a good point about the wording, leaving it to just "format" sounds like a safe approach. 

Looking again I think what made the functionality confusing for me was that even when I clicked the format button the tool tip stayed visible and there is no visual/color change to the button so I couldn't be sure if I was really clicked or just hovering. 

No 'on click' visual change appears to be the norm across Terra but usually a click opens/closes something so there is some sort of visual clue something happened. Disabling does seem like it could cause a performance hit (especially when there are like 60 inputs) I wonder if maybe we could add a visual cue just to the on-click? (example: https://material.io/components/buttons/#)

zarsky-broad: In some places, specifically with copy to clipboard buttons, we briefly replace the button with a checkmark, maybe something like that?

panentheos:
Thanks for the comments. I added a disabled state, and modified the tooltip language to be clearer. This is ready for another look.

ehanna4 (approved):
LGTM!
- Tested the changes to the workflow page in the UI and I think the updated tool tips and disabled state for the format provide good feedback to the user - thanks for updating that!
- Tested instances of textarea to make sure they still work as normal
- Overall looked good in the UI and I believe the code looks good - since we mobbed on this and had an office hours session on wrappers I feel pretty confident in my code review of this PR as well.

PR #2066, opened 2020-02-28
Decrease notebook autosave frequency

akarukappadath:
Welder, the service that writes notebooks from disk to the workspace bucket, has been getting overwhelmed while trying to delocalize large notebooks through autosave because of frequent autosaving. This PR increases the autosave frequency while we find a more sustainable solution for Welder. 

<!--
Hello, friend!
Thanks!
--!>


ehanna4 (approved):
Looks good to me! Good intermediate solution to make the user experience better when dealing with large notebooks. Tested by working in a notebook and all worked as expected.

PR #2067, opened 2020-02-28
Fix reference to possible undefined [SATURN-1437]

breilly2:
https://broadworkbench.atlassian.net/browse/SATURN-1437

This affects cases where a workflow's root entity type is a set type and there are no entities of the base type in the workspace. I created an example of this and shared it with b.adm.firec:

https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/breilly-2020-02-28

There are 2 workflows in this workspace. The one ending in "_set" throws this error on dev. This PR fixes the error.

### DIFF
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -136,7 +136,7 @@ const deleteCluster = withUserToken(async ({ email, token }) => {
   await signIntoTerra(ajaxPage, token)
 
   const currentC = await getCurrentCluster()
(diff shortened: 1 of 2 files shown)


ehanna4 (approved):
Ran the failing test locally and it passed no problem. Because the rerun had multiple tests fail it seems to be a flaky thing occurring with the integration tests at this time. Fixing this bug is high priority. Will keep an eye on the new integration test to determine if it has any real issues.

PR #2068, opened 2020-03-02
Increase timeout for run-notebook test

ehanna4:
The run-notebook integration test failed four times over the weekend. At least two cases look to be due to the time out not being long enough waiting for a runtime cluster instance to be running.


panentheos (approved):
Looks good.

PR #2069, opened 2020-03-03
fix warning bg color for accessibility [SATURN-1320]

cameron-ardell:
Currently in Terra the warning messages have a dark orange background with white text. This is difficult to read and does not meet accessibility standards. UI/UX have put together a new color scheme for these messages.

Also, it was noted by myself and @zarsky-broad `TrialBanner.js` line 72, there's coding for a banner that either doesn't meet the accessibility requirements or it's in a color that isn't our utils. How should we handle this?

### DIFF
--- a/src/libs/data-utils.js
+++ b/src/libs/data-utils.js
@@ -26,7 +26,7 @@ import validate from 'validate.js'
 const warningBoxStyle = {
   backgroundColor: colors.warning(0.15),
   padding: '1rem 1.25rem',
-  color: colors.dark(), fontWeight: 'bold', fontSize: 12
+  color: colors.dark(1) , fontWeight: 'bold', fontSize: 12
 }
 
 const errorTextStyle = { color: colors.danger(), fontWeight: 'bold', fontSize: 12, marginTop: '0.5rem' }
--- a/src/libs/data-utils.js
+++ b/src/libs/data-utils.js
@@ -24,9 +24,9 @@ import validate from 'validate.js'
 
 
 const warningBoxStyle = {
-  backgroundColor: colors.warning(1),
+  backgroundColor: colors.warning(0.15),
   padding: '1rem 1.25rem',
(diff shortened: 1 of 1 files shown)


src/libs/data-utils.js
```diff
@@ -24,9 +24,9 @@ import validate from 'validate.js'
 
 
 const warningBoxStyle = {
-  backgroundColor: colors.warning(1),
+  backgroundColor: colors.warning(0.15),
   padding: '1rem 1.25rem',
-  color: colors.light(1), fontWeight: 'bold', fontSize: 12
+  color: colors.dark(), fontWeight: 'bold', fontSize: 12
```
ehanna4, line 29: This is being overridden by a style in line at 366, so it's still showing up as light

ehanna4: The icon also needs the dark color rather than the light

zarsky-broad:
When you have a chance, could you rename this PR and add the ticket number?

src/libs/data-utils.js
cameron-ardell: Is this new change (see below) what you meant should be done?

ehanna4: I'm testing based on the mocks in the ticket, this first photo is what I see when I deploy your branch:
image

and this photo is what the mocks look like:
image

so the text color seems correct now but the icon color still doesn't match.

cameron-ardell: ah okay! I resolved that now I believe.

breilly2 (approved):
👍 Looks good! I tested the data import warning locally. I also looked at the cluster replace warning, but that has moved to a separate sheet of the cluster management widget now and does not use the warning color.

It's a good point about `TrialBanner.js`. My take is that that's out of scope for this ticket and that we should involve UI/UX in a redesign of the free trial credit experience.

PR #2070, opened 2020-03-04
expose lodash/fp on console [SATURN-1238]

panentheos:
Previously, the `_` in the console pointed to the base `lodash`, rather than `lodash/fp`, which was confusing and unhelpful. This exposes `lodash/fp` instead, so we can use it for debugging.

### DIFF
--- a/src/appLoader.js
+++ b/src/appLoader.js
@@ -17,6 +17,7 @@ RModal.setAppElement(appRoot)
 window.SATURN_VERSION = process.env.REACT_APP_VERSION
 marked.setOptions({ sanitize: true, sanitizer: _.escape })
 
+window._ = _
 
 ReactDOM.render(h(Main), appRoot)
 initializeAuth()


petesantos (commented):
My only concern is that given lodash was exposed globally, do we know what was exposing it, and whether any of our dependencies are relying on this being exposed globally?

panentheos:
I didn't track down what was exposing it, but I clicked around and verified that nothing seemed broken after the change.

breilly2:
Pete brings up a good point, and I think it would be good to track down. If we miss this now and it does affect some corner of the app that we didn't detect, it's going to be difficult to track down the cause later. For example, maybe it's IGV.js or the markdown editor or (more likely) some other obscure corner of the app.

Would it be useful to expose it as something other than `window._`?

breilly2 (approved):
👍 I had attempted to break the site by doing `window._ = undefined` on the console but was unable to. More importantly, Isaac walked Pete and me through exploration using `yarn why` which proves that there are no non-prod dependencies to be worried about.

petesantos (approved):
+1 on Brian's comments

PR #2071, opened 2020-03-04
Make test failure more meaningful [SATURN-1438]

ehanna4:
- Updated the console.info in the integration tests to get more info for debugging when a test fails. 

- Also added a condition to only try to delete a cluster if it was successfully created. Since the cluster is created within the test it's possible that that fails, however the test then fails on deleting this nonexistent cluster as well and the real error stack trace can get covered up by that stack trace of not finding a cluster to delete.

- Also increased the timeout as I had a test that took 9 minutes to create a cluster and technically the max time in the UI warning is 10minutes, so I made it 10minutes now

- I couldn't finish testing locally because I got rate-limited by Sam so need to dig into that, but the integration tests passed here on dev so I think it should be okay

### DIFF
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -24,7 +24,7 @@ const testRunNotebookFn = withRegisteredUser(async ({ page, context, email, toke
     await select(page, 'ENVIRONMENT', 'Hail')
     await click(page, clickable({ text: 'Create' }))
     await findElement(page, clickable({ textContains: 'Creating' }))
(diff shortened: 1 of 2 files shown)


breilly2:
I think the 429 responses from sam are because we have too many pet service accounts in saturn-integration-test-dev, probably because of test failures that skip past clean-up. I'm going to try to clean this up manually.

breilly2 (approved):
👍 Looks good. Thanks for keeping attention on this!

PR #2072, opened 2020-03-10
Add bard configs [SATURN-1465]

petesantos:
1. Add bard configuration for all environments. Changed config name to "bardRoot" and "fetchBard"
2. Send "properties" to bard instead of "data" to pass validation


ehanna4 (approved):
LGTM! Tested running locally and enabling each environment individually (staging, perf, alpha, dev) and saw events in mixpanel when enabled correctly and ensured events did not come through when not enabled

PR #2073, opened 2020-03-10
Flip the flag to enable metrics tracking SATURN-1300

ehanna4:
Turn on Mixpanel - the privacy policy will come as a separate effort


zarsky-broad (approved):

src/libs/ajax.js
```diff
@@ -1103,12 +1101,13 @@ const Metrics = signal => ({
...
     }
+
     // Remove the metricsEnabled feature flag once TOS and all metrics projects are setup
-    return metricsEnabled && fetchBard('api/event', _.mergeAll([authOpts(), jsonBody(body), { signal, method: 'POST' }]))
+    fetchBard('api/event', _.mergeAll([authOpts(), jsonBody(body), { signal, method: 'POST' }]))
   }),
 
   syncProfile: withErrorIgnoring(() => {
-    return metricsEnabled && fetchBard('api/syncProfile', _.merge(authOpts(), { signal, method: 'POST' }))
+    fetchBard('api/syncProfile', _.merge(authOpts(), { signal, method: 'POST' }))
```
panentheos: We should `return` the promise here (and above) to keep everything linked up properly, even though we're currently ignoring errors.

panentheos (approved):
See comment, then looks good.

PR #2074, opened 2020-03-12
Make test service, bueller, work in GAE [SATURN-1450]

breilly2:
I think the docs are valid and useful. I'll give them a final self-review soon, but let me know if you run into any problems. I tested each test from CLI, running the server locally, and running in GAE.

### DIFF
--- a/integration-tests/Bueller.md
+++ b/integration-tests/Bueller.md
@@ -1,0 +1,30 @@
+# Bueller
+
+Test runner service
+
+## Overview
+
+This service provides a way to run an integration test via an HTTP request. The initial primary use case is to use a tool like [Gatling](https://gatling.io/) to initiate many simultaneous test runs to determine how well Terra performs under load.
+
+## Authentication
+
+Requires an OpenID Connect ID token for the service account `bueller-user@terra-bueller.iam.gserviceaccount.com`. A private key for that service account is stored in Vault at `secret/dsde/terra/envs/common/bueller-user-service-account-key`.
+
+To generate a token, create a JWT with the additional claim `target_audience: 'https://terra-bueller.appspot.com'`, sign it with the private key, and pass it to Google's OAuth2 `token` endpoint. Using one of Google's client libraries will make this easier.
+
+Pass the token in a header with every call: `Authentication: Bearer <token>`
+
+## Developing
+
+Install deps
+```bash
+yarn install
+```
+
+Start a dev server on port 8080 with auto-reload
+```bash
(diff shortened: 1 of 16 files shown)


integration-tests/Bueller.md
```diff
@@ -0,0 +1,30 @@
...
+
+This service provides a way to run an integration test via an HTTP request. The initial primary use case is to use a tool like [Gatling](https://gatling.io/) to initiate many simultaneous test runs to determine how well Terra performs under load.
+
+## Authentication
+
+Requires an OpenID Connect ID token for the service account `bueller-user@terra-bueller.iam.gserviceaccount.com`. A private key for that service account is stored in Vault at `secret/dsde/terra/envs/common/bueller-user-service-account-key`.
+
+To generate a token, create a JWT with the additional claim `target_audience: 'https://terra-bueller.appspot.com'`, sign it with the private key, and pass it to Google's OAuth2 `token` endpoint. Using one of Google's client libraries will make this easier.
+
+Pass the token in a header with every call: `Authentication: Bearer <token>`
```
panentheos:
```suggestion
Pass the token in a header with every call: `Authorization: Bearer <token>`
```

integration-tests/src/server.js
```diff
@@ -14,50 +13,43 @@ const { delay, withScreenshot } = require('../utils/integration-utils')
...
-  return browser.createIncognitoBrowserContext()
-}
+server.setTimeout(serverTimeout)
 
 const testTimeout = async timeout => {
   await delay(timeout)
   throw new Error(`Test timeout after ${timeout}ms`)
 }
 
+const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, timeout }) => {
```
panentheos, line 28: TOL: I'm not sure that all of these wrappers are adding clarity. Generally speaking, I think wrappers are appropriate if they enable 1. code reuse or 2. modularization of semi-independent code, usually via composition (e.g. flow). In this case, the code is only used once, and although things are modularized, the wrappers are being applied manually, which makes it harder to read.

I don't think this is something we need to address in this PR, but just something to think about as the code evolves.

panentheos (approved):
Looks good, see doc comment.

integration-tests/src/server.js
breilly2: I see what you're saying. My intent here was to separate concepts for readability, which I think aligns with your point #2. I really wanted to address this now to avoid the risk of not getting back around to it. However, at this point, I think I'll need to just merge and trust that my curiosity will bring me back around to this.

PR #2075, opened 2020-03-13
add custom branding for UK Biobank [SATURN-1473]

panentheos:
This applies custom branding for the UK Biobank demo site. To test on non-prod environments, open the console, enter `configOverridesStore.set({ isUKBiobank: true })`, and refresh.


ehanna4 (approved):
LGTM I compared the mocks to the UI and based on Adrian/Lou's comments about some things might be slightly out of date on the mocks I also cross-referenced with current Prod UI for the content. I also looked over the code and it looks good. 

TOL: The notification about access to NIH controlled workspaces is almost the same color as the top banner/header, I wonder if we should consider making the 'no title' notification color slightly different to make the contrast greater between them.

panentheos:
Good point about the background color. Seeing as this is just a demo for now, I'm inclined to leave that for a later design pass, assuming everything goes forward.

PR #2077, opened 2020-03-16
update deps & address vulnerabilities

zarsky-broad:
Read changelogs and tested. Nothing major; bump in version of React, but nothing that should affect us, just some better warnings.

Also updates some packages that were showing up as vulnerable.


panentheos (approved):
Looks good, tested locally.

PR #2078, opened 2020-03-16
Alternate implementation of share workspace modal to match mocks [SATURN-1477]

petesantos:
I created an implementation that uses the autocomplete text box and downshift to get the same effect as the implementation in [2051](https://github.com/DataBiosphere/terra-ui/pull/2051/files).

I believe this approach alleviates the issues Brett pointed out wrt the select component being used as a text input.

I added a new option to the Autocomplete component to allow for events to occur on selection and enter. I also added the option to display placeholder text if there are no suggestions.

If we choose to go with this, this PR can be merged independently of 2051 and some of the changes in 2051 can be removed.

### DIFF
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -250,7 +250,7 @@ const withAutocomplete = WrappedComponent => ({
             } else if (_.includes(e.key, ['ArrowUp', 'ArrowDown']) && !suggestions.length) {
               e.nativeEvent.preventDownshiftDefault = true
             } else if (e.key === 'Enter') {
-              onPick?.(value)
+              onPick && onPick(value)
             }
           },
           nativeOnChange: true,
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -213,7 +213,7 @@ const AutocompleteSuggestions = ({ target: targetId, containerProps, children })
 }
 
 const withAutocomplete = WrappedComponent => ({
-  instructions, value, onChange, onPick,
+  instructions, value, onChange, onPick, onClick,
   suggestions: rawSuggestions, style, id, renderSuggestion = _.identity, openOnFocus = true, placeholderText, ...props
 }) => {
   const suggestions = _.filter(Utils.textMatch(value), rawSuggestions)
@@ -236,6 +236,10 @@ const withAutocomplete = WrappedComponent => ({
         h(WrappedComponent, getInputProps({
           style,
           type: 'search',
+          onClick: evt => {
+            openOnFocus && openMenu()
+            onClick && onClick(evt)
+          },
           onKeyDown: e => {
             if (e.key === 'Escape') {
               (value || isOpen) && e.stopPropagation() // prevent e.g. closing a modal
@@ -253,24 +257,20 @@ const withAutocomplete = WrappedComponent => ({
           nativeOnChange: true,
           ...props
         })),
-        isOpen && Utils.cond(
-          [!suggestions.length && placeholderText, () => h(AutocompleteSuggestions, {
-            target: getInputProps().id,
-            containerProps: getMenuProps()
-          }, [
-            div({
-              style: { textAlign: 'center', paddingTop: '0.75rem', height: '2.5rem', color: colors.dark(0.8) }
-            }, [placeholderText])
-          ])],
-          [!!suggestions.length, () => h(AutocompleteSuggestions, {
-            target: getInputProps().id,
-            containerProps: getMenuProps()
-          }, _.map(([index, item]) => {
+        isOpen && h(AutocompleteSuggestions, {
+          target: getInputProps().id,
+          containerProps: getMenuProps()
+        },
+        Utils.cond(
+          [!suggestions.length && placeholderText, () => [div({
+            style: { textAlign: 'center', paddingTop: '0.75rem', height: '2.5rem', color: colors.dark(0.8) }
+          }, [placeholderText])]],
+          [!!suggestions.length, () => _.map(([index, item]) => {
             return div(getItemProps({
               item, key: item,
               style: styles.suggestion(highlightedIndex === index)
             }), [renderSuggestion(item)])
-          }, Utils.toIndexPairs(suggestions)))]
+          }, Utils.toIndexPairs(suggestions))])
         )
       ])
     }
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -261,10 +261,10 @@ const withAutocomplete = WrappedComponent => ({
           containerProps: getMenuProps()
         },
         Utils.cond(
-          [!suggestions.length && placeholderText, () => [div({
+          [_.isEmpty(suggestions) && placeholderText, () => [div({
             style: { textAlign: 'center', paddingTop: '0.75rem', height: '2.5rem', color: colors.dark(0.8) }
           }, [placeholderText])]],
-          [!!suggestions.length, () => _.map(([index, item]) => {
+          [!_.isEmpty(suggestions), () => _.map(([index, item]) => {
             return div(getItemProps({
               item, key: item,
               style: styles.suggestion(highlightedIndex === index)
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -212,11 +212,15 @@ const AutocompleteSuggestions = ({ target: targetId, containerProps, children })
   ])
 }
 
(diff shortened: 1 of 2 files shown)


src/components/input.js
```diff
@@ -225,11 +229,17 @@ const withAutocomplete = WrappedComponent => ({ value, onChange, suggestions: ra
...
-      return div({ style: { width: '100%', display: 'inline-flex' } }, [
+      return div({
+        onFocus: openOnFocus ? openMenu : undefined,
+        style: { width: '100%', display: 'inline-flex' }
+      }, [
         h(WrappedComponent, getInputProps({
           style,
           type: 'search',
-          onFocus: openOnFocus ? openMenu : undefined,
+          onClick: evt => {
```
petesantos: mob - Put focus on the newly created item, eliminating the need for this code

src/components/input.js
```diff
@@ -225,11 +229,17 @@ const withAutocomplete = WrappedComponent => ({ value, onChange, suggestions: ra
     inputId: id
   }, [
     ({ getInputProps, getMenuProps, getItemProps, isOpen, openMenu, toggleMenu, highlightedIndex }) => {
-      return div({ style: { width: '100%', display: 'inline-flex' } }, [
+      return div({
+        onFocus: openOnFocus ? openMenu : undefined,
+        style: { width: '100%', display: 'inline-flex' }
```
petesantos, line 237: mob - Add shadow to the drop down, when overlapping other components it looks too flat and blends in an odd way. 
Check the workflow config, group management screen, and a few other places to make sure it is behaving correctly.

petesantos: Added styles. Checked group config with the new code, l didn't see any issues. That is the only other spot in the code that I saw the autocomplete being used.

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -110,70 +115,75 @@ export default ajaxCaller(class ShareWorkspaceModal extends Component {
...
+    const isOld = _.find({ email }, originalAcl)
+    const numAdditions = _.filter(({ email }) => !_.some({ email }, originalAcl), acl).length
 
     return div({
-      style: { display: 'flex', padding: '0.5rem', borderTop: index && `1px solid ${colors.dark(0.4)}` }
+      style: {
+        display: 'flex', alignItems: 'center', borderRadius: 5,
+        padding: '0.5rem 0.75rem', marginBottom: 10,
+        border: `1px solid ${isOld ? colors.dark(0.25) : colors.success(0.5)}`,
+        backgroundColor: isOld ? colors.light(0.2) : colors.success(0.05)
```
ehanna4, line 170: I'm curious others thoughts but to me this would seem to violate the A11y rule D1: "Is information conveyed by methods other than color alone?" i.e. the information that this user in the list is a newly added one is only identified by this green color (which is not enough of a contrast to count as a lightness difference). But maybe that rule is only for necessary info? I've tried rereading the doc but am still unsure whether it would apply here or not, but just wanted to see what others thought.

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -110,70 +115,75 @@ export default ajaxCaller(class ShareWorkspaceModal extends Component {
         h(FormLabel, { htmlFor: id }, ['User email']),
         h(AutocompleteTextInput, {
           id,
+          openOnFocus: true,
+          placeholderText: _.includes(searchValue, aclEmails) ?
+            'This email has already been added to the list' :
```
ehanna4, line 120: Really appreciate this check - with the old version I could just keep pasting the same email and adding them over and over and only got an error message at the end when trying to save

src/pages/workspaces/workspace/ShareWorkspaceModal.js
petesantos: Thanks for keeping A11y in mind during the PR! I changed this to a dashed border for the new items. I will run it by Lou to get his thoughts

ehanna4: oh that's awesome, really like the look of that thanks!

petesantos: Checked with Lou, he liked the change and updated the mock

src/components/input.js
```diff
@@ -19,7 +19,10 @@ const styles = {
     position: 'fixed', top: 0, left: 0,
     maxHeight: 36 * 8 + 2, overflowY: 'auto',
     backgroundColor: 'white',
-    border: `1px solid ${colors.light()}`
+    border: `1px solid ${colors.light()}`,
+    margin: '8px 0',
+    borderRadius: '4px',
```
ehanna4: nit: Typically I've seen in the codebase using REMs for margins and padding favored over px (when possible) - is there a reason for choosing to use pixels here?

ehanna4 (approved):
Looks good to me, I viewed in the browser, comparing to the old style and found this to be an improvement. I also looked over the code and it looks good to me (perhaps another set of eyes on that though wouldn't hurt). Please just see the one comment. Thanks!

src/components/input.js
zarsky-broad: For border radii, px can be the way to go in order to match a mock, since small changes in number can have large visual impact.
```suggestion
    borderRadius: 4,
```

petesantos: +1 on Isaac's comment for border radii. On the margin I was matching the select styling. I changed this to (0.5) rem to be in-line with our margin settings elsewhere. Used suggestion as well.

src/components/input.js
```diff
@@ -240,20 +249,28 @@ const withAutocomplete = WrappedComponent => ({ value, onChange, suggestions: ra
               toggleMenu()
             } else if (_.includes(e.key, ['ArrowUp', 'ArrowDown']) && !suggestions.length) {
               e.nativeEvent.preventDownshiftDefault = true
+            } else if (e.key === 'Enter') {
+              onPick && onPick(value)
```
zarsky-broad, line 253:
```suggestion
              onPick?.(value)
```

src/components/input.js
```diff
@@ -240,20 +249,28 @@ const withAutocomplete = WrappedComponent => ({ value, onChange, suggestions: ra
...
           containerProps: getMenuProps()
-        }, _.map(([index, item]) => {
-          return div(getItemProps({
-            item, key: item,
-            style: styles.suggestion(highlightedIndex === index)
-          }), [renderSuggestion(item)])
-        }, Utils.toIndexPairs(suggestions)))
+        },
+        Utils.cond(
+          [!suggestions.length && placeholderText, () => [div({
```
zarsky-broad: Prefer `_.isEmpty`?

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -110,70 +115,75 @@ export default ajaxCaller(class ShareWorkspaceModal extends Component {
...
+            'Enter an email address',
+          onPick: value => {
+            !validate.single(value, { email: true, exclusion: aclEmails }) &&
+              this.setState(
+                _.flow(
+                  _.update('acl', Utils.append({ email: value, accessLevel: 'READER' })),
+                  _.update('searchValue', () => '')
```
zarsky-broad: `_.set`?

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -110,70 +115,75 @@ export default ajaxCaller(class ShareWorkspaceModal extends Component {
...
+              )
+          },
           placeholder: 'Add people or groups',
           value: searchValue,
           onChange: v => this.setState({ searchValue: v }),
-          suggestions: _.difference(suggestions, _.map('email', acl)),
+          suggestions: Utils.cond(
+            [searchValueValid && !_.includes(searchValue, aclEmails), () => [searchValue]],
+            [remainingSuggestions.length, () => remainingSuggestions],
+            []
```
zarsky-broad: `() => []` for clarity?

src/components/input.js
petesantos: I had to revert this change - eslint was complaining about it. It looks like a [known issue](https://github.com/facebook/create-react-app/issues/8107). I tried the two solutions with the babel plugin and typescript version of the rules but it let to a series of other (unwanted) warnings in our code.

zarsky-broad: Gross.

PR #2080, opened 2020-03-17
Add new and interesting workspaces [SATURN-1474]

zarsky-broad:
Verified in browser, featured workspaces json list in dev has one.


src/pages/workspaces/List.js
```diff
@@ -121,14 +121,22 @@ export const WorkspaceList = () => {
...
+
+    return {
+      myWorkspaces: _.filter(ws => !ws.public || Utils.canWrite(ws.accessLevel), workspaces),
+      public: _.filter('public', workspaces),
+      newAndInteresting: _.flow(
+        _.map(({ isNew, ...featuredWs }) => _.find({ workspace: featuredWs }, workspaces)),
+        _.compact
+      )(newWsList),
+      featured: _.flow(
+        _.map(featuredWs => _.find({ workspace: featuredWs }, workspaces)),
```
panentheos: We should probably explicitly `_.pick` out just the `namespace` and `name` here, so this behaves well even if there are other attributes present.

panentheos (approved):
Looks good, see comment.

PR #2082, opened 2020-03-19
New 'Update your Runtime' notification

akarukappadath:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -167,7 +167,7 @@ export default class ClusterManager extends PureComponent {
       setDynamic(sessionStorage, `notifiedOutdatedCluster${cluster.id}`, Date.now())
       notify('warn', 'Please Update Your Runtime', {
         message: h(Fragment, [
-          p(['On Sunday Oct 20th at 10am, we are introducing important updates to Terra, which are not compatible with the older notebook runtime in this workspace. After this date, you will no longer be able to save new changes to notebooks in one of these older runtimes.']),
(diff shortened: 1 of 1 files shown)


src/components/ClusterManager.js
```diff
@@ -167,7 +167,7 @@ export default class ClusterManager extends PureComponent {
       setDynamic(sessionStorage, `notifiedOutdatedCluster${cluster.id}`, Date.now())
       notify('warn', 'Please Update Your Runtime', {
         message: h(Fragment, [
-          p(['On Sunday Oct 20th at 10am, we are introducing important updates to Terra, which are not compatible with the older notebook runtime in this workspace. After this date, you will no longer be able to save new changes to notebooks in one of these older runtimes.']),
```
ehanna4, line 170: Should this date include the year and time zone? i.e. 'Sunday Oct 20th, 2020 at 10am EST' ? (Also if this time is based on the TerraUI deployment cycle we actually release at 11am now (due to daylights savings time switch and we didn't adjust it))

akarukappadath: I don't think it needs to be quite that specific, but I can update to "In October of 2019," if you think that would be helpful to users.

zarsky-broad: I think @ehanna4 was looking at the old removed text and not the replacement...?

ehanna4: 🤦‍♀️yup sorry, never mind on my last comment

ehanna4 (approved):
LGTM, tested in UI with the changes

PR #2085, opened 2020-03-23
Bug fix: viewport spinner [SATURN-1482]

petesantos:
Fixes an issue where if there is scrolling content on the page the spinner will be centered based on the content height. This pushes the spinner out of view, only showing the overlay. This can be seen on the billing page for the b.adm account and on the workflows page when there are a lot of inputs, outputs, etc... Enabling network throttling in the dev tools exacerbates the issue.

This change fixes this issue and it keeps the spinner centered when the window is sized below the min width.


src/components/icons.js
```diff
@@ -14,7 +14,16 @@ export const icon = (shape, { size = 16, ...props } = {}) => _.invokeArgs(shape,
 
 export const spinner = props => icon('loadingSpinner', _.merge({ size: 24, style: { color: colors.primary() } }, props))
 
-export const centeredSpinner = props => spinner(_.merge({ size: 48, style: { display: 'block', margin: 'auto' } }, props))
+export const centeredSpinner = ({ size = 48, ...props } = {}) => spinner(_.merge({
+  size, style: {
+    display: 'block',
+    position: 'sticky',
+    top: `calc(50% - ${size / 2}px)`,
```
ehanna4, line 21: This trick using css `calc` is neat, I looked it up to be sure I fully understood what it was doing and this certainly seems like an appropriate place for it, nice!

src/components/AuthContainer.js
```diff
@@ -13,15 +13,17 @@ import TermsOfService from 'src/pages/TermsOfService'
 const AuthContainer = ({ children }) => {
   const { name, public: isPublic } = useRoute()
   const { isSignedIn, registrationStatus, acceptedTos, profile } = Utils.useStore(authStore)
+  const spinner = () => h(centeredSpinner, { size: 48, style: { right: 'calc(50vw - 24px)', left: 'calc(50vw - 24px)' } })
```
ehanna4: The centeredSpinner has a default of `size: 48 `right? Is it necessary to have` size: 48` here as well?

petesantos: I want to explicitly set the size to 48 here since the calc is using half the size. If the default size were to change the calculation would be a bit off

ehanna4 (approved):
LGTM! tested the UI from the PR site and the spinner was now in view and the code changes look good to me too, albeit default 'maybe another eyes on the code wouldn't hurt' 👍

zarsky-broad (approved):

PR #2086, opened 2020-03-25
fire event when clicking new email [SATURN-1502]

zarsky-broad:
`onSelect` is a superset of `onChange`: it fires even when the selected value matches the value already in the input.


petesantos (approved):
Looks good, tested locally - thanks!

PR #2088, opened 2020-03-26
Check if onPick function is set before executing [SATURN-1509]

petesantos:


zarsky-broad (approved):
Verified on pr site: currently broken, fixed with this PR. I don't understand why it wasn't broken before, tbh...

PR #2089, opened 2020-03-27
Get more info on failure [SATURN-1475]

ehanna4:
This will enable getting a screenshot of one of the error notification details as well as the general page screenshot when a test fails. This will help us understand what is happening with the flaky notebooks test. 

Also I added a 1 second delay to the helper that adds the test user to the billing (similar to the 1 second delay that was added to the register user helper) to make sure the request has enough time to fully go through. When I ran the test locally I _think_ I saw an error of 403 which is a permissions issue (I only have ever been able to reproduce this error once on purpose), so I am thinking it is possible the issue is not allowing enough time for the user to get fully added to the project before trying to access the notebooks.

Tested: 
With config overrides got a similar error to appear and got screenshots of the page and notification details successfully. When no error notification on failure still gets the regular whole page screenshot.

**NOTE**
This is an imperfect solution. Brett and I discussed a much better long term solution [(ticket here)](https://broadworkbench.atlassian.net/browse/SATURN-1512) the general idea being to enable a button or keyboard shortcut to get a dump of the error info whenever an error notification comes about which could be used for both tests and users. For now this PR is pretty specifically for getting details on the notebooks failure and can easily be pulled out when we develop a better more universal approach.

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -169,5 +169,5 @@ module.exports = {
   navChild,
   findInDataTableRow,
   withScreenshot,
-  openNotification
+  openError
 }
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -107,7 +107,7 @@ const findInDataTableRow = (page, entityName, text) => {
   return findElement(page, `//*[@role="grid"]//*[contains(.,"${entityName}")]/following-sibling::*[contains(.,"${text}")]`)
 }
 
-const openNotification = async page => {
+const openError = async page => {
   //close out any non-error notifications first
   await dismissNotifications(page)
 
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -139,7 +139,7 @@ const withScreenshot = _.curry((testName, fn) => async options => {
         if (screenshotBucket) {
           const storage = new Storage()
           await storage.bucket(screenshotBucket).upload(path)
-          if (notificationsPresent) {
+          if (errorsPresent) {
             await storage.bucket(screenshotBucket).upload({ path: failureNotificationDetailsPath })
           }
         }
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -113,9 +113,7 @@ const openNotification = async page => {
 
   const errorDetails = await page.$x('(//a | //*[@role="button"] | //button)[contains(normalize-space(.),"Details")]')
 
-  await Promise.all(
-    errorDetails.map(handle => handle.click())
-  )
+  !!errorDetails[0] && await errorDetails[0].click()
 
   return errorDetails.length
 }
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -130,7 +130,7 @@ const withScreenshot = _.curry((testName, fn) => async options => {
 
         await page.screenshot({ path, fullPage: true })
 
-        const notificationsPresent = await openNotification(page)
+        const errorsPresent = await openError(page)
 
         if (!!notificationsPresent) {
           await page.screenshot({ path: failureNotificationDetailsPath, fullPage: true })
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -132,7 +132,7 @@ const withScreenshot = _.curry((testName, fn) => async options => {
 
         const errorsPresent = await openError(page)
 
-        if (!!notificationsPresent) {
+        if (errorsPresent) {
           await page.screenshot({ path: failureNotificationDetailsPath, fullPage: true })
         }
 
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -115,7 +115,7 @@ const openNotification = async page => {
 
   !!errorDetails[0] && await errorDetails[0].click()
 
-  return errorDetails.length
+  return !!errorDetails.length
 }
 
 const withScreenshot = _.curry((testName, fn) => async options => {
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -107,6 +107,19 @@ const findInDataTableRow = (page, entityName, text) => {
   return findElement(page, `//*[@role="grid"]//*[contains(.,"${entityName}")]/following-sibling::*[contains(.,"${text}")]`)
 }
 
+const openNotification = async page => {
+  //close out any non-error notifications first
+  await dismissNotifications(page)
+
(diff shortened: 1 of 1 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -107,6 +107,17 @@ const findInDataTableRow = (page, entityName, text) => {
   return findElement(page, `//*[@role="grid"]//*[contains(.,"${entityName}")]/following-sibling::*[contains(.,"${text}")]`)
 }
 
+const openNotification = async page => {
```
zarsky-broad:
```suggestion
const openError = async page => {
```

integration-tests/utils/integration-utils.js
```diff
@@ -115,10 +126,22 @@ const withScreenshot = _.curry((testName, fn) => async options => {
     if (screenshotDir) {
       try {
         const path = `${screenshotDir}/failure-${Date.now()}-${testName}.png`
+        const failureNotificationDetailsPath = `${screenshotDir}/failureDetails-${Date.now()}-${testName}.png`
+
         await page.screenshot({ path, fullPage: true })
+
+        const notificationsPresent = await openNotification(page)
```
zarsky-broad:
```suggestion
        const errorsPresent = await openError(page)
```

integration-tests/utils/integration-utils.js
```diff
@@ -115,10 +126,22 @@ const withScreenshot = _.curry((testName, fn) => async options => {
...
+
+        if (!!notificationsPresent) {
```
zarsky-broad:
```suggestion
        if (errorsPresent) {
```

integration-tests/utils/integration-utils.js
```diff
@@ -115,10 +126,22 @@ const withScreenshot = _.curry((testName, fn) => async options => {
...
+          await page.screenshot({ path: failureNotificationDetailsPath, fullPage: true })
+        }
+
         if (screenshotBucket) {
           const storage = new Storage()
           await storage.bucket(screenshotBucket).upload(path)
+          if (notificationsPresent) {
```
zarsky-broad:
```suggestion
          if (errorsPresent) {
```

integration-tests/utils/integration-utils.js
```diff
@@ -145,5 +168,6 @@ module.exports = {
   signIntoTerra,
   navChild,
   findInDataTableRow,
-  withScreenshot
+  withScreenshot,
+  openNotification
```
zarsky-broad:
```suggestion
  openError
```

integration-tests/utils/integration-utils.js
```diff
@@ -107,6 +107,17 @@ const findInDataTableRow = (page, entityName, text) => {
...
 
+const openNotification = async page => {
+  //close out any non-error notifications first
+  await dismissNotifications(page)
+
+  const errorDetails = await page.$x('(//a | //*[@role="button"] | //button)[contains(normalize-space(.),"Details")]')
+
+  !!errorDetails[0] && await errorDetails[0].click()
+
+  return errorDetails.length
```
zarsky-broad:
```suggestion
  return !!errorDetails.length
```

integration-tests/utils/integration-utils.js
```diff
@@ -107,6 +107,17 @@ const findInDataTableRow = (page, entityName, text) => {
   return findElement(page, `//*[@role="grid"]//*[contains(.,"${entityName}")]/following-sibling::*[contains(.,"${text}")]`)
 }
 
+const openNotification = async page => {
+  //close out any non-error notifications first
+  await dismissNotifications(page)
+
+  const errorDetails = await page.$x('(//a | //*[@role="button"] | //button)[contains(normalize-space(.),"Details")]')
```
zarsky-broad, line 114:
```suggestion
  const [errorDetails] = await page.$x('(//a | //*[@role="button"] | //button)[contains(normalize-space(.),"Details")]')
```

integration-tests/utils/integration-utils.js
```diff
@@ -107,6 +107,17 @@ const findInDataTableRow = (page, entityName, text) => {
...
+
+  !!errorDetails[0] && await errorDetails[0].click()
```
zarsky-broad, line 116:
```suggestion
  !!errorDetails && await errorDetails[0].click()
```

zarsky-broad (approved):
Makes sense to me. Didn't test extensively, some suggestions on clarification/simplification.

integration-tests/utils/integration-utils.js
ehanna4: I imagine this way is essentially saving the result of the await page.$x... function as an array. However how would this handle if that result was empty? Would it save errorDetails to be an empty array? If so isn't an empty array truthy?

zarsky-broad: `emptyArray[0]` is `undefined`

zarsky-broad: We're destructuring the array here, `errorDetails` is `array[0]`

integration-tests/utils/integration-utils.js
```diff
@@ -115,10 +126,22 @@ const withScreenshot = _.curry((testName, fn) => async options => {
...
+
+        if (errorsPresent) {
+          await page.screenshot({ path: failureNotificationDetailsPath, fullPage: true })
+        }
+
         if (screenshotBucket) {
           const storage = new Storage()
           await storage.bucket(screenshotBucket).upload(path)
+          if (errorsPresent) {
+            await storage.bucket(screenshotBucket).upload({ path: failureNotificationDetailsPath })
```
panentheos, line 143: This signature looks different from the other one, I think the path should be passed directly as the first arg.

integration-tests/utils/integration-utils.js
ehanna4: The `await page.$x...` function returns an empty array already when it doesn't find any elements that match. I tried using the change you suggested but then it returns `undefined` instead which doesn't seem preferable to me

zarsky-broad: Sure, if you prefer; we're doing `errorDetails[0]` below, so that's the same undefined. Not a huge difference, feel free to take it or leave it.

integration-tests/utils/integration-utils.js
```diff
@@ -107,6 +107,17 @@ const findInDataTableRow = (page, entityName, text) => {
   return findElement(page, `//*[@role="grid"]//*[contains(.,"${entityName}")]/following-sibling::*[contains(.,"${text}")]`)
 }
 
+const openError = async page => {
+  //close out any non-error notifications first
+  await dismissNotifications(page)
+
+  const errorDetails = await page.$x('(//a | //*[@role="button"] | //button)[contains(normalize-space(.),"Details")]')
```
panentheos, line 114: TOL: This looks like it could use the `clickable` helper, but doing that would create a circular dependency which I'm guessing would cause problems. Not something we need to unravel here, but it seems like we have a couple different 'utility' files that could be consolidated, which could enable some code re-use.

panentheos (approved):
Looks good, but see comment.

PR #2090, opened 2020-03-27
Update deps, deduplicate, handle vulnerability

zarsky-broad:
Read changelogs and tested! This removes vulnerable versions of `minimist` from our transitive deps by updating other transitive deps that depended on them directly or indirectly.


panentheos (approved):
Looks good, tested locally.

PR #2091, opened 2020-03-27
(Possibly Better) Fix for busy spinner with css max API

petesantos:
Here is a slightly simpler fix for the busy spinner. This one works better across the board as the spinner always sticks correctly.

However it does require 2 things:
1. A change to where the overflow is set on the root/body elements. I am not sure what the rationale is for having the overflows split between the elements, but I have not seen any issues. But I may be unaware of something else there.

2. The CSS "min" API. This is supported in all browsers except Firefox. It should be supported in the next version of Firefox which is projected to release on April 7th.

If we want to move forward with this we could theoretically merge around this time.


zarsky-broad:
We have people who use the Firefox ESR (currently 68), so I would be hesitant to do it this way.

zarsky-broad:
Closing for now, since we went with the alternate implementation.

PR #2093, opened 2020-03-30
Fix file path

ehanna4:
I missed Brett's PR feedback, so this is applying that. 

I tested and he was correct that the old file path was actually not working and with this new way it works correctly and I verified both screenshots made it into the google bucket.


zarsky-broad (approved):
lgtm

panentheos (approved):
Looks good, thanks!

PR #2094, opened 2020-03-31
Links to covid info [SATURN-1518]

zarsky-broad:
Good for review


breilly2 (commented):
In slack, I asked if we should only show this in Terra (not FireCloud, Anvil, BDC, etc.) and Adrian agreed with that. Other than that, this looks good.

src/components/TopBar.js
```diff
@@ -299,6 +299,14 @@ const TopBar = Utils.connectStore(authStore, 'authState')(class TopBar extends C
               ...Utils.newTabLinkProps
             }, ['Release Notes'])
           ]),
+          (isTerra() || isFirecloud()) && h(NavSection, {
```
breilly2: Terra or FireCloud? I thought we were doing just Terra (unless Adrian instructed otherwise).

breilly2 (approved):
👍 LGTM once it's limited to just Terra (not FC).

PR #2095, opened 2020-04-02
combine (non-integration) test & build

zarsky-broad:
Running test separately from build caused two sources to try to read the cache simultaneously, which slowed things down significantly.


cameron-ardell (approved):
LGTMob!

PR #2097, opened 2020-04-03
update metrics calls for new API [SATURN-1535]

panentheos:
This changes the metrics calls to conform to the new bard API, including making a call to the new `syncProfile` endpoint on load/login.


zarsky-broad (approved):
Verified in browser that syncProfile is called when it should be.

PR #2099, opened 2020-04-06
Fix overflowing text [SATURN-1542]

ehanna4:
Before:
Screen Shot 2020-04-06 at 4 38 36 PM

After:
Screen Shot 2020-04-06 at 4 58 48 PM


panentheos (approved):
LGTMob

PR #2101, opened 2020-04-09
Make Landing and Workspace list page better for screenreaders [SATURN-1506]

ehanna4:
**Part 1**
Remove `makeIconButton()` and replace its one usage with a more specific simplified version that is a11y compliant.

Looking for opinions:
TOL
The function `makeIconButton() `was in `common.js` however it was only being used one time in `LandingPage.js`. It also had a lot of options for dealing with props and merging props for the different pieces of the icon that were never used/needed. I took this all out to simplify my understanding while I made the change to make it an `image` rather than a `clickable` with the intention of putting that prop logic back in after. However now that I have finished it doesn't seem appropriate to add back as this is only used once and can be kept more simple and specific keeping it a function for this one icon in `LandingPage.js`. 

Is there a reason to keep the original makeIconButton that I am missing? Is it appropriate to take out the extra logic or should it be kept in so that in the future if there is need for it it's there?

Tested using ANDI to ensure the a11y errors were gone for the arrow buttons
Before:
Screen Shot 2020-04-09 at 1 14 53 PM

After:
Screen Shot 2020-04-09 at 1 14 41 PM

**Part 2**
Fix the workspace list page so that the 3-dot menus and the make new workspace button have a11y compliance

Before:
Screen Shot 2020-04-09 at 3 33 47 PM

After:
Screen Shot 2020-04-09 at 3 34 05 PM

### DIFF
--- a/src/pages/LandingPage.js
+++ b/src/pages/LandingPage.js
@@ -1,8 +1,9 @@
 import { Fragment } from 'react'
 import { div, h, h2 } from 'react-hyperscript-helpers'
-import { Clickable, HeroWrapper, Link, makeIconButton } from 'src/components/common'
+import { Clickable, HeroWrapper, Link } from 'src/components/common'
 import { icon } from 'src/components/icons'
 import covidHero from 'src/images/covid-hero.jpg'
+import hexButton from 'src/images/hex-button.svg'
 import colors from 'src/libs/colors'
 import { isFirecloud, isTerra } from 'src/libs/config'
 import * as Nav from 'src/libs/nav'
@@ -25,6 +26,18 @@ const makeDocLink = (href, title) => {
   ])
 }
 
+const makeRightArrowWithBackgroundIcon = () => div({
+  style: {
+    height: 30, width: 30,
+    display: 'flex', alignItems: 'center', justifyContent: 'center',
+    alignSelf: 'flex-end',
+    backgroundColor: colors.accent(),
+    ...(isTerra() ?
+      { mask: `url(${hexButton}) center no-repeat`, WebkitMask: `url(${hexButton}) center no-repeat` } :
+      { borderRadius: '1rem' })
+  }
+}, [icon('arrowRight', { color: 'white' })])
+
 const makeCard = (link, title, body) => h(Clickable, {
   href: Nav.getLink(link),
   style: { ...Style.elements.card.container, height: 245, width: 225, marginRight: '1rem', justifyContent: undefined },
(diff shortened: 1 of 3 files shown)


src/pages/LandingPage.js
```diff
@@ -25,6 +26,18 @@ const makeDocLink = (href, title) => {
   ])
 }
 
+const makeRightArrowWithBackgroundIcon = () => div({
```
breilly2, line 29: In response to your questions, I'm in favor of the simplification here. It's not clear to me that this is a pattern we want to reuse anyway.

TOL: Since this is specific to use in these cards, it could be inlined now instead of being a separate function. Might depend on whether or not we want to encourage reuse of this construct.

src/pages/workspaces/List.js
```diff
@@ -258,7 +258,7 @@ export const WorkspaceList = () => {
                   closeOnClick: true,
                   content: h(WorkspaceMenuContent, { namespace, name, onShare, onClone, onDelete })
                 }, [
-                  h(Link, { 'aria-label': 'Workspace menu', disabled: !canView }, [icon('cardMenuIcon', { size: 20 })])
+                  h(Link, { 'aria-label': 'Workspace menu', disabled: !canView, 'aria-hidden': false }, [icon('cardMenuIcon', { size: 20 })])
```
breilly2: TOL: It's very interesting to see these 2 changes in the same PR. Having to explicitly say `'aria-hidden': false`, especially at the point of use (rather than baked into a reusable component) feels less than ideal. I'm not sure what the right solution is, though.

Maybe `PopupTrigger` should have `role: 'button'` like `Interactive` (conditionally) does. Maybe `Link` (or `Clickable` or `Interactive`) should automatically add `'aria-hidden': false` when an `aria-label` is present.

Do others on the team have thoughts?

ehanna4: This piece I am not 100% sure on. In the video and in the acceptance criteria in the ticket these 3-dot menus are called out as lacking sufficient screen-reader transparency. However I was not finding the 3-dot menus to be problematic so I figured it might be a browser configuration thing and that setting aria-hidden to false rather than depending on it implicitly setting to false could alleviate that discrepancy.

Very much appreciate the feedback on potential other options, I will look into those

breilly2: There's also the question of whether or not the aria-label needs to be more specific, i.e. naming the workspace it applies to. The ticket references the old design in which it's clear that there is no other context, such as a table column header, to help disambiguate. In the new design, that is less clear (though from what I remember of the implementation, I think these aren't technically table columns).

zarsky-broad:
Fwiw, makeIconButton was going to be a thing when Terra was going to have hex buttons and other brands were going to have round ones, but that idea got dropped.

src/pages/workspaces/List.js
panentheos: The `'aria-hidden': false` almost certainly has no effect, so I think we should remove it. I agree with @breilly2 that for now we should probably just have the aria label include the workspace name, so it's clear what it applies to. In the future we can look at ways to annotate the table properly so that's less of a concern.

panentheos (approved):
Looks good, see comment.

PR #2102, opened 2020-04-09
Add support for GCE runtimes [SATURN-1411] [SATURN-1577]

zarsky-broad:
This can be tested against dev, but **can't merge until Leo changes are on prod.**

...probably review with whitespace hidden.


src/components/NewClusterModal.js
```diff
@@ -285,10 +288,10 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
     const { currentCluster, onDismiss, onSuccess } = this.props
     const {
-      profile, masterMachineType, masterDiskSize, workerMachineType, numberOfWorkers, numberOfPreemptibleWorkers, workerDiskSize,
+      profile, masterMachineType, masterDiskSize, sparkMode, workerMachineType, numberOfWorkers, numberOfPreemptibleWorkers, workerDiskSize,
       jupyterUserScriptUri, selectedLeoImage, customEnvImage, leoImages, viewMode
     } = this.state
-    const { version, updated, packages } = _.find({ image: selectedLeoImage }, leoImages) || {}
+    const { version, updated, packages, requiresSpark } = _.find({ image: selectedLeoImage }, leoImages) || {}
 
     const makeEnvSelect = id => h(Select, {
```
panentheos, line 301: Currently if you click "What's installed..." and choose a Hail runtime from there, it doesn't update spark mode. This `Select` needs to use the same logic as the other one.

src/components/NewClusterModal.js
```diff
@@ -380,7 +389,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
                   onChange: ({ value }) => {
                     this.setState({
                       profile: value,
-                      ...(value === 'custom' ? {} : normalizeMachineConfig(_.find({ name: value }, profiles).machineConfig))
+                      ...(value === 'custom' ? {} : normalizeRuntimeConfig(_.find({ name: value }, profiles).runtimeConfig))
```
panentheos: If you change to spark cluster mode and then select a non-custom profile, it zeroes out the worker count. This `Select` should now only control the master machine type and disk size.

src/components/ClusterManager.js
```diff
@@ -209,22 +209,22 @@ export default class ClusterManager extends PureComponent {
     const { namespace } = this.props
     this.executeAndRefresh(
       Ajax().Clusters.cluster(namespace, Utils.generateClusterName()).create({
-        machineConfig: normalizeMachineConfig({})
+        runtimeConfig: normalizeRuntimeConfig({})
```
panentheos: This tries to create a GCE machine, but using the spark version of the configuration, which is invalid.

src/components/NewClusterModal.js
```diff
@@ -110,41 +107,46 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
   constructor(props) {
     super(props)
     const { currentCluster } = props
-    const currentConfig = currentCluster ? currentCluster.machineConfig : profiles[0].machineConfig
-    const matchingProfile = _.find(
-      ({ machineConfig }) => machineConfigsEqual(machineConfig, currentConfig),
-      profiles
-    )
+    const { cloudService, ...currentConfig } = normalizeRuntimeConfig(currentCluster?.runtimeConfig || profiles[0].runtimeConfig)
+    const matchingProfile = _.find({ runtimeConfig: { masterMachineType: currentConfig.masterMachineType } }, profiles)
```
panentheos: Currently this will match even if the disk size is different, but since the `Select` controls access to that field, it should be included in the matching criteria here.

src/components/NewClusterModal.js
```diff
@@ -110,41 +107,46 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-      ...normalizeMachineConfig(currentConfig)
+      sparkMode: cloudService === 'GCE' ? false : currentConfig.numberOfWorkers === 0 ? 'master' : 'cluster',
+      ...currentConfig
     }
   }
 
-  getMachineConfig() {
-    const { numberOfWorkers, masterMachineType, masterDiskSize, workerMachineType, workerDiskSize, numberOfPreemptibleWorkers } = this.state
-    return {
+  getClusterConfig() {
```
panentheos: Was this rename intentional, or should it be `getRuntimeConfig`?

breilly2 (commented):
Testing notes:

Creating runtime with “Default (moderate)” profile and "Spark cluster" runtime type and then opening the runtime widget shows a “Custom” profile.

Same thing with "Spark master node" except, in addition, the "Update" button is enabled even though nothing has been changed.

Nit: I think a nice enhancement would be to indicate which, if any, fields have changed when selecting an environment. That would help call attention to the automatic change of runtime type.

Finally, once there are GCE runtimes in the wild, it will be difficult to roll back to the old `/cluster` API because that API throws errors when given GCE runtimes. Rollbacks are very infrequent so this shouldn't be a problem, but we should be aware of this.

src/components/ClusterManager.js
```diff
@@ -209,22 +212,22 @@ export default class ClusterManager extends PureComponent {
     const { namespace } = this.props
     this.executeAndRefresh(
       Ajax().Clusters.cluster(namespace, Utils.generateClusterName()).create({
-        machineConfig: normalizeMachineConfig({})
+        runtimeConfig: formatRuntimeConfig(normalizeRuntimeConfig({}))
```
panentheos, line 215: While this technically does the right thing, I think there might be a conceptual bug lurking. We essentially have three different types of data describing machines:
* L: the native Leo format, with `cloudService`, and the split sub-fields, e.g. `machineType` vs `masterMachineType`
* N: the normalized format, with `cloudService`, and the sub-fields merged into the dataproc names, e.g. `masterMachineType`
* S: the state format, with `cloudService` and `numWorkers` denormalized as `sparkMode`, and also merged sub-fields

`normalizeRuntimeConfig` is `L => N`, and formatRuntimeConfig is currently `S => L`, so they're technically not compatible. It happens to work, because `formatRuntimeConfig` looks for `sparkMode`, finds it `undefined`, and defaults `cloudService` to `'GCE'`. But I could see this leading to potential confusion or bugs later.

I think a good solution would be to change `formatRuntimeConfig` to `N => L` by having it look at `cloudService` instead, and write an inline `S => N` transform in `getRuntimeConfig`. That keeps all handling of `S` confined to `NewClusterModal`, which makes sense since that's where it's relevant.

src/components/NewClusterModal.js
```diff
@@ -110,41 +107,38 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
     super(props)
     const { currentCluster } = props
-    const currentConfig = currentCluster ? currentCluster.machineConfig : profiles[0].machineConfig
-    const matchingProfile = _.find(
-      ({ machineConfig }) => machineConfigsEqual(machineConfig, currentConfig),
-      profiles
-    )
+    const { cloudService, ...currentConfig } = normalizeRuntimeConfig(currentCluster?.runtimeConfig || profiles[0].runtimeConfig)
+    const { masterDiskSize, masterMachineType, numberOfWorkers } = currentConfig // want these to be put into state below, unlike cloudService
+    const matchingProfile = masterDiskSize === normalizeRuntimeConfig({}).masterDiskSize && _.find({ runtimeConfig: { masterMachineType } }, profiles)
```
panentheos: We probably shouldn't assume that the preset profiles always use the default disk size.
```suggestion
    const matchingProfile = _.find(({ runtimeConfig }) => _.isMatch({ masterMachineType, masterDiskSize }, normalizeRuntimeConfig(runtimeConfig)), profiles)
```

src/libs/cluster-utils.js
```diff
@@ -2,53 +2,69 @@ import _ from 'lodash/fp'
...
+  const { cloudService, masterMachineType, numberOfWorkers, numberOfPreemptibleWorkers, workerMachineType } = normalizeRuntimeConfig(config)
+  const { price: masterPrice, cpu: masterCpu } = findMachineType(masterMachineType)
+  const { price: workerPrice, preemptiblePrice, cpu: workerCpu } = findMachineType(workerMachineType)
   return _.sum([
     masterPrice,
     (numberOfWorkers - numberOfPreemptibleWorkers) * workerPrice,
     numberOfPreemptibleWorkers * preemptiblePrice,
-    machineStorageCost(config)
+    machineStorageCost(config),
+    cloudService === 'GCE' ? 0 : masterCpu + (workerCpu * numberOfWorkers) // dataproc costs $.01 per cpu per hour
```
panentheos: Currently this is adding $1 instead of $0.01
```suggestion
    cloudService === 'GCE' ? 0 : (masterCpu + (workerCpu * numberOfWorkers)) * 0.01 // dataproc costs $.01 per cpu per hour
```

src/components/NewClusterModal.js
```diff
@@ -285,16 +280,23 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+      profile, masterMachineType, masterDiskSize, sparkMode, workerMachineType, numberOfWorkers, numberOfPreemptibleWorkers, workerDiskSize,
       jupyterUserScriptUri, selectedLeoImage, customEnvImage, leoImages, viewMode
     } = this.state
-    const { version, updated, packages } = _.find({ image: selectedLeoImage }, leoImages) || {}
+    const { version, updated, packages, requiresSpark } = _.find({ image: selectedLeoImage }, leoImages) || {}
+
+    const onEnvChange = ({ value }) => {
+      this.setState({
+        selectedLeoImage: value, customEnvImage: '',
+        sparkMode: _.find({ image: value }, leoImages)?.requiresSpark ? (sparkMode || 'master') : false
```
panentheos: Choosing Hail, selecting spark cluster, switching to non-Hail, and then back to Hail ends up in a state where 'spark master' is selected, but there are also non-zero workers. If we change `sparkMode` here, we also need to update `numWorkers`

zarsky-broad:
@panentheos @breilly2 I _think_ this is ready for a re-review?

src/components/NewClusterModal.js
```diff
@@ -155,22 +151,23 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
       }),
       isStopRequired ? 5000 : 0)
   }
 
   hasStartUpScriptChanged() {
-    const { jupyterUserScriptUri, currentClusterDetails } = this.state
-    const originalJupyterUserScriptUri = currentClusterDetails?.jupyterUserScriptUri || ''
+    const { currentCluster } = this.props
+    const { jupyterUserScriptUri } = this.state
+    const originalJupyterUserScriptUri = currentCluster.jupyterUserScriptUri || ''
```
panentheos: The record from props doesn't have the `jupyterUserScriptUri`, so we need to look at the one from state.

panentheos (approved):
See one comment, then looks good. Tested locally.

PR #2103, opened 2020-04-09
Check workspace billing project access before using for requester pays [SATURN-1548]

breilly2:
Tested using this workspace: https://bvdp-saturn-dev.appspot.com/#workspaces/breilly-test-20200409/requester-pays-test%20copy which is owned by b.adm.firec who is also the sole owner of the workspace's billing project.

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -78,24 +78,18 @@ const checkRequesterPaysError = async response => {
  */
 const withRequesterPays = wrappedFetch => (url, ...args) => {
   const bucket = /\/b\/([^/?]+)[/?]/.exec(url)[1]
-  const workspace = workspaceStore.get()
-  const workspaceProject = workspace?.workspace?.namespace
-  const canCompute = workspace?.canCompute
+  const { workspace, canCompute } = workspaceStore.get() || {}
+  const workspaceProject = workspace?.namespace
 
-  const isProjectOwner = async () => {
-    const userProjects = await Billing().listProjects()
-    return _.some({ projectName: workspaceProject, role: 'Owner' }, userProjects)
+  const canUseWorkspaceProject = async () => {
+    return canCompute || _.some({ projectName: workspaceProject, role: 'Owner' }, await Ajax().Billing.listProjects())
   }
 
   const getUserProject = async () => {
-    if (requesterPaysProjectStore.get()) {
-      return requesterPaysProjectStore.get()
-    } else if (workspaceProject && (canCompute || (await isProjectOwner()))) {
+    if (!requesterPaysProjectStore.get() && workspaceProject && await canUseWorkspaceProject()) {
       requesterPaysProjectStore.set(workspaceProject)
-      return workspaceProject
-    } else {
-      return undefined
     }
+    return requesterPaysProjectStore.get()
   }
 
   const tryRequest = async () => {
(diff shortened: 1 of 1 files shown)


src/libs/ajax.js
```diff
@@ -76,10 +76,13 @@ const checkRequesterPaysError = async response => {
...
  */
-const withRequesterPays = wrappedFetch => (url, ...args) => {
+const withRequesterPays = wrappedFetch => async (url, ...args) => {
   const bucket = /\/b\/([^/?]+)[/?]/.exec(url)[1]
   const workspace = workspaceStore.get()
-  const userProject = workspace && Utils.canWrite(workspace.accessLevel) ? workspace.workspace.namespace : requesterPaysProjectStore.get()
+  const userProjects = await Billing().listProjects()
+  const userProject = workspace && _.some(['projectName', workspace.workspace.namespace], userProjects) ?
+    workspace.workspace.namespace :
+    requesterPaysProjectStore.get()
```
breilly2: I chose to not write this back to `requesterPaysProjectStore` at this time because that only holds a single project and is only cleared on sign-out. I think it would be better as a map from bucket to project. Otherwise, this bug could be encountered again by, in a single session, viewing two requester pays buckets in two different billing projects.

zarsky-broad (commented):
So, there are two side-effects of this change.
1. Non-writers who happen to be members of the billing project won't be prompted
2. Every bucket request will be preceded by a list billing projects request.

In light of #2, I think it would make more sense to only query for projects in the case that the bucket is known requester pays or fails. Does that make sense? Or am I confused?

src/libs/ajax.js
panentheos: This store is also cleared when navigating from one workspace to another (see `auth.js`). FWIW the originally intended behavior is that once the user selects a billing project, that project is used for all requester pays operations made from the context of that workspace, until they navigate away. Does that change your perspective on writing back to the store?

breilly2: Yes, it does. I saw `auth.js` and incorrectly assumed that it was being cleared on sign-out. I am okay with writing to `requesterPaysProjectStore` here.

That brings up another issue: That reference in `auth.js` is the only call to `reset()` for `requesterPaysProjectStore`. Are we not clearing this on sign-out? If so, that feels like a bug to me.

breilly2:
@zarsky-broad
#1 sounds like a bug that should be fixed in this PR.
#2 should become a non-issue by addressing Brett's comment about writing to `requesterPaysProjectStore`. We'll only ask for billing projects on the first requester-pays bucket request for a workspace.

src/libs/ajax.js
```diff
@@ -79,16 +79,34 @@ const checkRequesterPaysError = async response => {
 const withRequesterPays = wrappedFetch => (url, ...args) => {
   const bucket = /\/b\/([^/?]+)[/?]/.exec(url)[1]
   const workspace = workspaceStore.get()
-  const userProject = workspace && Utils.canWrite(workspace.accessLevel) ? workspace.workspace.namespace : requesterPaysProjectStore.get()
+  const workspaceProject = workspace?.workspace?.namespace
```
breilly2:
```suggestion
  const workspaceProject = workspace?.workspace.namespace
```

src/libs/ajax.js
```diff
@@ -79,16 +79,34 @@ const checkRequesterPaysError = async response => {
...
+  const canCompute = workspace?.canCompute
+
+  const isProjectOwner = async () => {
+    const userProjects = await Billing().listProjects()
```
breilly2:
```suggestion
    const userProjects = await Ajax().Billing.listProjects()
```

panentheos (approved):
LGTMob

PR #2104, opened 2020-04-13
Add back Terra's ToS and Privacy Policy [SATURN-1551]

cameron-ardell:
We removed Terra's ToS and privacy policy from the footer so we needed to add them back somewhere, so we decided to add them to the dropdown menu from the top bar.

I've tested these locally, they only show up for BDC and the links lead to a new tab with the correct information.

**I'm wondering if we want to add icons next to these titles as all the other menu items have icons.** Other than that, I believe this draft is complete.

Ticket is [here](https://broadworkbench.atlassian.net/browse/SATURN-1551?atlOrigin=eyJpIjoiZjQ0OTJmM2Y0ODE3NDk5YTgxMWFlNDQ5MzIyYWU2MmYiLCJwIjoiaiJ9)

### DIFF
--- a/src/components/TopBar.js
+++ b/src/components/TopBar.js
@@ -332,15 +332,13 @@ const TopBar = Utils.connectStore(authStore, 'authState')(class TopBar extends C
                   variant: 'light',
                   style: { display: 'block', textDecoration: 'underline', color: colors.accent(0.2) },
                   href: Nav.getLink('privacy'),
-                  onClick: () => this.hideNav(),
-                  ...Utils.newTabLinkProps
+                  onClick: () => this.hideNav()
                 }, ['Terra Privacy Policy']),
               h(Link, {
                 variant: 'light',
                 href: Nav.getLink('terms-of-service'),
                 style: { display: 'block', textDecoration: 'underline', color: colors.accent(0.2) },
-                onClick: () => this.hideNav(),
-                ...Utils.newTabLinkProps
+                onClick: () => this.hideNav()
               }, ['Terra Terms of Service'])
             ]),
             div({ style: { color: colors.dark(0.3), fontSize: 10, fontWeight: 600, marginTop: '0.5rem' } }, [
--- a/src/components/TopBar.js
+++ b/src/components/TopBar.js
@@ -327,19 +327,21 @@ const TopBar = Utils.connectStore(authStore, 'authState')(class TopBar extends C
           }, [
             h(CromwellVersionLink, { variant: 'light', style: { textDecoration: 'underline', color: colors.accent(0.2) } }),
             isBioDataCatalyst() && h(Fragment, [
-              div([h(Link,
+              h(Link,
                 {
-                  style: { textDecoration: 'underline', color: colors.accent(0.2) },
+                  variant: 'light',
+                  style: { display: 'block', textDecoration: 'underline', color: colors.accent(0.2) },
                   href: Nav.getLink('privacy'),
                   onClick: () => this.hideNav(),
                   ...Utils.newTabLinkProps
-                }, ['Terra Privacy Policy'])]),
-              div([h(Link, {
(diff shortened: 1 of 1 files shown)


src/components/TopBar.js
```diff
@@ -307,6 +307,16 @@ const TopBar = Utils.connectStore(authStore, 'authState')(class TopBar extends C
             icon('virus', { size: 24, style: styles.nav.icon }),
             'COVID-19 Data & Tools'
           ]),
+          isBioDataCatalyst() && h(NavSection, {
+            href: Nav.getLink('privacy'),
+            onClick: () => this.hideNav(),
+            ...Utils.newTabLinkProps
+          }, ['Terra Privacy Policy']),
+          isBioDataCatalyst() && h(NavSection, {
```
cameron-ardell: Is there a way to make it so I'm only checking `isBioDataCatalyst()` once and then create the two menu items?

cameron-ardell:
Update: I just pinged Lou regarding the icons

cameron-ardell:
Per a conversation with Lou, am moving the links down so they're minor links instead of major features of the menu (like how we have Cromwell version)

src/components/TopBar.js
zarsky-broad: `h(Fragment, [])`

cameron-ardell: thanks!

src/components/TopBar.js
```diff
@@ -324,6 +326,21 @@ const TopBar = Utils.connectStore(authStore, 'authState')(class TopBar extends C
             style: { flex: 'none', padding: 28, marginTop: 'auto' }
           }, [
             h(CromwellVersionLink, { variant: 'light', style: { textDecoration: 'underline', color: colors.accent(0.2) } }),
+            isBioDataCatalyst() && h(Fragment, [
+              div([h(Link,
+                {
+                  style: { textDecoration: 'underline', color: colors.accent(0.2) },
```
panentheos: The default hover color is not readable against the dark background. Adding a `variant: 'light'` prop will make it behave like the Cromwell link above, which is fine for now.

src/components/TopBar.js
```diff
@@ -324,6 +326,21 @@ const TopBar = Utils.connectStore(authStore, 'authState')(class TopBar extends C
...
+                  href: Nav.getLink('privacy'),
+                  onClick: () => this.hideNav(),
+                  ...Utils.newTabLinkProps
+                }, ['Terra Privacy Policy'])]),
+              div([h(Link, {
```
panentheos: nit: Keeping consistent spacing and line breaks helps with overall readability. Even though it's not enforced by the linter, generally speaking if an array or object is multiline, there should be a line break after the opener and before the closer, e.g.
```js
div([
  h(Link, {
    href: 'foo'
  }, ['This array is one-line'])
])
```

src/components/TopBar.js
```diff
@@ -324,6 +326,21 @@ const TopBar = Utils.connectStore(authStore, 'authState')(class TopBar extends C
...
+                href: Nav.getLink('terms-of-service'),
+                style: { textDecoration: 'underline', color: colors.accent(0.2) },
+                onClick: () => this.hideNav(),
+                ...Utils.newTabLinkProps
```
panentheos: If we want to keep the behavior from the original footer, these should not open in a new tab since they are local links.

src/components/TopBar.js
zarsky-broad: Also, adding `display: 'block'` will mean you don't need the wrapping divs.

cameron-ardell: I think we want to wrap the files so that way we only need to check for if it's biodatacatalyst once

cameron-ardell:
This is ready for a fast re-review

panentheos (approved):
Looks good, tested locally.

PR #2105, opened 2020-04-13
Fix color contrast of trial banner [SATURN-1505]

ehanna4:
Fix the color contrast of the Free Trial banner to meet a11y requirements. Also adjust the size of the title so that the NHLBI fits nicer.

Before:
Screen Shot 2020-04-13 at 12 15 24 PM

After:
Screen Shot 2020-04-13 at 12 22 24 PM

### DIFF
--- a/src/components/TrialBanner.js
+++ b/src/components/TrialBanner.js
@@ -76,8 +76,8 @@ export const TrialBanner = Utils.connectStore(authStore, 'authState')(class Tria
       [
         div({
           style: {
-            fontSize: '1.45rem', fontWeight: 500, textAlign: 'right', borderRight: '1px solid', paddingRight: '1rem', marginRight: '1rem',
-            maxWidth: 200, flexShrink: 0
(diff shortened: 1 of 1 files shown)


src/components/TrialBanner.js
```diff
@@ -69,14 +69,14 @@ export const TrialBanner = Utils.connectStore(authStore, 'authState')(class Tria
...
           display: 'flex', alignItems: 'center', padding: '1.5rem', height: 95,
-          backgroundColor: isWarning ? colors.warning() : '#359448',
+          backgroundColor: isWarning ? colors.warning() : '#28873b',
           justifyContent: 'center', color: 'white', width: '100%', fontSize: '1rem'
         }
       },
       [
         div({
           style: {
-            fontSize: '1.5rem', fontWeight: 500, textAlign: 'right', borderRight: '1px solid', paddingRight: '1rem', marginRight: '1rem',
```
zarsky-broad, line 79: Why not just increase the `maxWidth`?

zarsky-broad (approved):
One comment, otherwise looks solid.

src/components/TrialBanner.js
ehanna4: Oh yes that seems better 👍will make that change

PR #2107, opened 2020-04-14
update deps

zarsky-broad:
Tested and read changelogs.

We're switching to using the native react-scripts eslint extension method, no visible changes from that.

There are also two minor code changes to things that weren't problems before (although at least one should have been), we think this is because of sub-dependency updates from react-scripts that change how our code is being transpiled, probably mostly due to dropping support for older browser versions (this is governed by our browserslist config in the package.json).


panentheos (approved):
Looks good, tested locally.

PR #2111, opened 2020-04-22
Add security-related http response headers and report-only mode CSP 

adrazhi:
### Security Headers Added

- `Expect-CT`: More info in header [spec in Chrome.](https://www.chromestatus.com/feature/5677171733430272)
Needed for Certificate Transparency log validation. It's currently purposefully on **report-only** mode and likely will be for some time while we observe its behavior in the associating endpoint. 

- `Referrer-Policy`: The default as can be seen [here](https://www.chromestatus.com/feature/6251880185331712) is `strict-origin-when-cross-origin`. However, we set it to `no-referrer-when-downgrade` which would keep referrer data off HTTP connections. 

- `Content-Security-Policy-Report-Only:` The **report-only** directive does not block anything, but rather post a JSON blob with potential violations to the associating endpoint. This would allow us to observe assets that are loaded, forms posted, XHR requests made and everything else in-between,  over time and eventually craft a content security policy that would be as accurate as possible. 

### Testing Performed

There is no new functionality added to the application, just a change to the [app.yaml ](https://github.com/DataBiosphere/terra-ui/blob/dev/app.yaml)file. It's hard to test this since the acceptance criteria would be the data we get when a CSP "violation" occurs. That said, the above directives are in **report-only** mode so they do not impact application behavior.

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -15,3 +15,5 @@ handlers:
     Strict-Transport-Security: "max-age=31536000; includeSubdomains; preload"
     X-Content-Type-Options: "nosniff"
     X-XSS-Protection: "1; mode=block"
+    Expect-CT: max-age=0, report-uri="https://broadappsec.report-uri.com/r/d/ct/reportOnly"
+    Referrer-Policy: no-referrer-when-downgrade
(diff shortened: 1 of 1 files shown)


panentheos:
Closing this for now, pending an investigation of a different implementation.

PR #2112, opened 2020-04-24
add tutorials and videos to menu [SATURN-1545]

zarsky-broad:
Tested with adblocker active and disabled (appcues blocked and loaded). At PR time, the links on the appcues modal still need to be tweaked, and the fallback article hasn't been published yet, but comms is working on that, so this should be able to proceed.


breilly2 (approved):
👍 Looks good. I tested on the PR site. Thanks for taking the time to implement a fallback for when Appcues is blocked!

PR #2113, opened 2020-04-24
Update header to be more specific [SATURN-1557]

ehanna4:
Update header to be more specific for what the drop down actually is for, tested locally


zarsky-broad (approved):
LGTM, addresses the concern of duplicate text.

PR #2114, opened 2020-04-24
checkbox to allow requesting access to a group [SATURN-1604] [SATURN-1575]

zarsky-broad:
Adds a checkbox to the group creation modal and to the group details screen to control whether users can request access to a group if it's used as an AD.

Verified working in-browser.


cameron-ardell (approved):
LGTOfficeHours, pending info box

PR #2115, opened 2020-04-27
Make Free Credits Banner have appropriate language for screen reader [SATURN-1561]

cameron-ardell:
added aria-label to x button in the free credits banner

I'm wondering if there should be a different message if they're currently using their free credits since the message is about learning more, not getting more. I am currently talking to Adrian about this, but I figured functionality could be tested.

To test:

- be logged in as Barty Admiralson
- go to `TrialBanner.js`
- toggle `removeBanner` to false
- be on any page on Terra (I used Workspaces) and go to `TrialBanner` in React Components
- toggle `authState.profile.trialState` to `Enabled` / `Enrolled` / `Terminated`


panentheos (approved):
LGTMob

PR #2116, opened 2020-04-27
roll back to igv that works with promises for auth [SATURN-1607]

zarsky-broad:


panentheos (approved):
Looks good, tested locally.

PR #2119, opened 2020-04-28
Remove text reference to plus button [SATURN-1558]

breilly2:
I don't love the look of such long link text in such a big font, but this is what I remember us talking about.

Screen Shot 2020-04-28 at 3 10 56 PM

Let me know if you have suggestions for changes.

### DIFF
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -53,10 +53,10 @@ const billingProjectNameValidator = existing => ({
   }
 })
 
-const noBillingMessage = onClickCreateProject => div({ style: { fontSize: 20, margin: '2rem' } }, [
+const noBillingMessage = onClick => div({ style: { fontSize: 20, margin: '2rem' } }, [
   div([
     'To get started, ',
-    h(Link, { onClick: onClickCreateProject }, ['click here to ', span({ style: { fontWeight: 600 } }, ['create a Billing Project'])])
+    h(Link, { onClick }, ['click here to create a Billing Project'])
   ]),
   div({ style: { marginTop: '1rem', fontSize: 16 } }, [
     h(Link, {
@@ -250,7 +250,7 @@ export const BillingList = _.flow(
     !isOwner && _.map(project => _.includes('Owner', project.role) ? this.setState({ isOwner: true }) : null, billingProjects)
   }
 
-  onClickCreateProject = async () => {
+  showCreateProjectModal = async () => {
     if (Auth.hasBillingScope()) {
       this.setState({ creatingBillingProject: true })
     } else {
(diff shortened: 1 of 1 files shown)


src/pages/billing/List.js
```diff
@@ -53,9 +53,10 @@ const billingProjectNameValidator = existing => ({
   }
 })
 
-const noBillingMessage = div({ style: { fontSize: 20, margin: '2rem' } }, [
+const noBillingMessage = onClickCreateProject => div({ style: { fontSize: 20, margin: '2rem' } }, [
   div([
-    'To get started, click the plus button to ', span({ style: { fontWeight: 600 } }, ['create a Billing Project'])
+    'To get started, ',
+    h(Link, { onClick: onClickCreateProject }, ['click here to ', span({ style: { fontWeight: 600 } }, ['create a Billing Project'])])
```
breilly2: Alternative: make just the `click here` text a link and leave the rest of the styling alone. Thoughts?

```suggestion
    h(Link, { onClick: onClickCreateProject }, ['click here']),
    ' to ', span({ style: { fontWeight: 600 } }, ['create a Billing Project'])
```

panentheos: Generally I think it's better for accessibility for links to be self-describing if possible, so the current form seems better. Maybe we don't need the bold font anymore though?

src/pages/billing/List.js
```diff
@@ -249,6 +250,15 @@ export const BillingList = _.flow(
     !isOwner && _.map(project => _.includes('Owner', project.role) ? this.setState({ isOwner: true }) : null, billingProjects)
   }
 
+  onClickCreateProject = async () => {
```
panentheos: naming nit: I would probably name this `onCreateProject`, since that's the essence of what this function does. The fact that it's bound to a click is an implementation detail at the point of usage.

panentheos (approved):
Looks good, tested locally.

src/pages/billing/List.js
breilly2: I almost used `onCreateProject`, but it's not actually called when the project is created. It's called on the user gesture indicating that they want to create a project which has the effect of showing the modal for creating a project. Though now that I think about it, I could argue that `onClickCreateProject` could also sound like something called when clicking a "Create Project" button in the modal.

How about a different naming approach based on what the function actually does? Something like `openCreateProjectModal` or `showCreateProjectModal`? (Granted it does a little more than that, but that's the desired outcome.)

src/pages/billing/List.js
breilly2: Right. Thanks for the reminder of that accessibility rule. I agree that the bold can probably go.

src/pages/billing/List.js
panentheos: Yes, I think either of those would be good options.

PR #2120, opened 2020-04-29
fix a11y on pre-load template [SATURN-1560]

zarsky-broad:
Adds alt tags to images, and implements the same logic the app uses to keep the copyright year current in the pre-load.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -70,7 +70,7 @@
           <div style="font-weight: 600; font-size: 10px; color: white;" id="copyright"></div>
           <script>
             const buildTimestamp = new Date(parseInt('%REACT_APP_BUILD_TIMESTAMP%', 10))
-            window.document.getElementById('copyright').innerHTML = `Copyright © ${buildTimestamp.getFullYear()}`
+            window.document.getElementById('copyright').innerHTML = `Copyright ©${buildTimestamp.getFullYear()}`
           </script>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -59,15 +59,19 @@
...
             style="flex: 0 0 auto; display: flex; align-items: center; justify-content: space-between; height: 66px; padding: 0 1rem; background-color: rgb(109, 110, 112);">
-          <img role="img" src="%PUBLIC_URL%/logo-grey.svg"
+          <img role="img" src="%PUBLIC_URL%/logo-grey.svg" alt="Terra logo"
                style="height: 40px; margin-bottom: 4px;">
-          <div style="font-weight: 600; font-size: 10px; color: white;">Copyright ©2019</div>
+          <div style="font-weight: 600; font-size: 10px; color: white;" id="copyright"></div>
+          <script>
+            const buildTimestamp = new Date(parseInt('%REACT_APP_BUILD_TIMESTAMP%', 10))
+            window.document.getElementById('copyright').innerHTML = `Copyright ©${buildTimestamp.getFullYear()}`
+          </script>
```
panentheos: TOL: Having a special script for this seems like a lot of effort for something that's only on the screen briefly. A few thoughts:
* Is there a way to include logic in the HTML template, such that this is computed at compile time?
* Could we just remove the copyright here entirely?
* Should we just leave the number static and remember to bump it every year?

panentheos (approved):
Looks good, tested locally.

public/index.html
zarsky-broad: - For rendering it in the template, we'd have to add a new `REACT_APP_` variable, but we could do it.
- We could just remove it, totally.
- It had been static, and we hadn't updated for 5 months, so that feels like a non-starter.

zarsky-broad: I'm just gonna remove it, I don't think it's necessary.

PR #2121, opened 2020-04-30
Identify user to Appcues only after accepting ToS [SATURN-1615]

zarsky-broad:


src/libs/auth.js
```diff
@@ -155,6 +151,14 @@ authStore.subscribe(withErrorReporting('Error checking TOS', async (state, oldSt
   }
 }))
 
+authStore.subscribe(async (state, oldState) => {
+  if (!oldState.acceptedTos && state.acceptedTos) {
```
breilly2, line 155: Question/TOL: When a user who has already accepted the ToS signs in, do we first write `acceptedTos` as falsy and then truthy? If not, I worry that we might now miss that case. If so, that behavior isn't clear here... which honestly is more or less true of all of these listeners.

I think I just answered my own question by looking at the initialization of `authStore`:
```
export const authStore = Utils.atom({
  isSignedIn: undefined,
  registrationStatus: undefined,
  acceptedTos: undefined,
  user: {},
  profile: {}
})
```

I think I'm starting to feel uncomfortable with how these subscriptions work. I get that it's convenient to be able to watch for arbitrary changes to atoms. However, I worry about how fragile this is over time. I'm afraid it would be too easy to accidentally change code such that a particular state change never happens and not notice that a subscriber is effectively a no-op as a result.

Is there some way we can make this more clear?

src/libs/auth.js
```diff
@@ -155,6 +151,14 @@ authStore.subscribe(withErrorReporting('Error checking TOS', async (state, oldSt
...
+    window.Appcues && window.Appcues.identify(state.user.id, {
+      dateJoined: parseJSON((await Ajax().User.firstTimestamp()).timestamp).getTime()
```
breilly2, line 157: Hmm... I think there's a problem here. `User.firstTimestamp()` doesn't just return the data. It also is (almost) the only path to writing the data in the first place! I didn't notice that when this was previously made conditional on `window.Appcues`.

It looks like `NpsSurvey` also calls this on mount when a user is registered. Even then, it's buried in the conditional for determining whether or not to show the NPS prompt.

There may not be any problem with the changes in this PR, but I'm now even more uncomfortable with how much effort it took to analyze this and how coincidental all of this behavior feels.

src/libs/auth.js
panentheos: Yes, the cascade of events here is constructed in a very specific way, and there are some implicit dependencies on the behavior here. A couple thoughts:

* We could probably document some of the assumptions, or give a high level overview of how the auth flow works, as long as we're willing to keep that documentation up to date.
* One feature that I sometimes wonder about is using `undefined` as the 'unknown' value, e.g. we haven't loaded it yet. We could use an explicit unique symbol to represent that instead, though it would make some logic a bit wordier (this line specifically)

Either way, I think it's beyond the scope of this PR, so let's have a separate discussion about this.

src/libs/auth.js
panentheos: The `firstTimestamp` bit is weird to begin with, since it's not something that the UI should be responsible for in the first place. The logic there is such that we'll always call it at least once per user, and that was the original usage of this data anyhow. Again, it's written in a very specific way, but given the use case I don't see a way to simplify it much.

panentheos (approved):
LGTMob. Tested locally.

PR #2122, opened 2020-05-01
consistently represent very low prices [SATURN-1614]

zarsky-broad:
This applies the "< $0.01" terminology for prices less than 1 cent everywhere in the app.


breilly2 (approved):
👍 Looks good. I tested on the PR site, both downloading a small file from a bucket and looking at a stopped GCE runtime.

zarsky-broad:
Note that this also impacts display of prices for job submissions.

breilly2:
Good point. Also the runtime configuration widget (though that will likely never show < $0.01) and runtime list page (x2). I just tested all places that use `Utils.formatUSD` and everything looks good. 👍

PR #2124, opened 2020-05-04
Add alt text to rstudio img [SATURN-1573]

ehanna4:
Added an alt tag to the img of the Rstudio Logo in the cluster manager widget

Tested by running locally and checking that an alt tag showed when inspected and also ran ANDI to confirm it had no issues with the img.

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -306,11 +306,11 @@ export default class ClusterManager extends PureComponent {
         href: appLaunchLink,
         onClick: window.location.hash === appLaunchLink && currentStatus === 'Stopped' ? () => this.startCluster() : undefined,
         tooltip: canCompute ? `Open ${appName}` : noCompute,
-        'aria-label': !isRStudioImage ? `Open ${appName}`: undefined,
+        'aria-label': `Open ${appName}`,
         disabled: !canCompute,
         style: { marginRight: '2rem', ...styles.verticalCenter },
         ...(isRStudioImage ? {} : Utils.newTabLinkProps)
-      }, [isRStudioImage ? img({ src: rLogo, alt: 'Open RStudio', style: { maxWidth: 24, maxHeight: 24 } }) : icon('terminal', { size: 24 })]),
+      }, [isRStudioImage ? img({ src: rLogo, alt: '', style: { maxWidth: 24, maxHeight: 24 } }) : icon('terminal', { size: 24 })]),
       renderIcon(),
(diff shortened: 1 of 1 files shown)


src/components/ClusterManager.js
```diff
@@ -310,7 +310,7 @@ export default class ClusterManager extends PureComponent {
         disabled: !canCompute,
         style: { marginRight: '2rem', ...styles.verticalCenter },
         ...(isRStudioImage ? {} : Utils.newTabLinkProps)
-      }, [isRStudioImage ? img({ src: rLogo, style: { maxWidth: 24, maxHeight: 24 } }) : icon('terminal', { size: 24 })]),
+      }, [isRStudioImage ? img({ src: rLogo, alt: 'R-Studio Logo', style: { maxWidth: 24, maxHeight: 24 } }) : icon('terminal', { size: 24 })]),
```
breilly2: Nit: I think it should be "RStudio" (without the dash). That's what they seem to do consistently at https://rstudio.com/. However, now that I think about it, maybe "RStudio" isn't handled well by screen readers?

breilly2 (dismissed):
👍 Looks good.

TOL: I think we talked about this at refinement so I don't want to second guess that reasoning even though I don't remember the details. However, I feel like this now has 2 a11y accommodations: an `aria-label` of "Open RStudio" on the link and an `alt` tag of "R-Studio" on the image, which is the only child element of the link. Is this redundant or are we addressing different a11y needs by having both?

src/components/ClusterManager.js
```diff
@@ -306,11 +306,11 @@ export default class ClusterManager extends PureComponent {
         href: appLaunchLink,
         onClick: window.location.hash === appLaunchLink && currentStatus === 'Stopped' ? () => this.startCluster() : undefined,
         tooltip: canCompute ? `Open ${appName}` : noCompute,
-        'aria-label': `Open ${appName}`,
+        'aria-label': !isRStudioImage ? `Open ${appName}`: undefined,
```
panentheos: My vote would be to always include the label here, and give an empty alt attribute on the logo. Since this is primarily a control, it seems like we should describe the button directly. And the logo in this case doesn't add any additional information beyond that, so it seems fine to leave it empty.

ehanna4:
I made the change for an empty alt tag with the aria-label enabled per Brian and Brett's suggestion with the basis of the link rather than the image being the primary function of this piece of the UI

breilly2 (approved):
👍 Looks good. Thanks for the diligence!

PR #2126, opened 2020-05-04
Fix color contrast for info about entities being copied [SATURN-1572]

breilly2:
Pretty straightforward. Also had to update the style for warning/error banners in that modal. For example, take a workspace that has data, clone it, then try to copy some data from the workspace to the clone.

### DIFF
--- a/src/components/ExportDataModal.js
+++ b/src/components/ExportDataModal.js
@@ -17,9 +17,9 @@ import * as Utils from 'src/libs/utils'
 import validate from 'validate.js'
 
 
-const InfoTile = ({ infoStyle, content, iconName, iconColor }) => {
+const InfoTile = ({ infoStyle, content, icon: { name, color } }) => {
   return div({ style: { ...infoStyle, display: 'flex', alignItems: 'center' } }, [
-    icon(iconName, { size: 36, style: { color: iconColor, flex: 'none', marginRight: '0.5rem' } }),
+    icon(name, { size: 36, style: { color, flex: 'none', marginRight: '0.5rem' } }),
     content
   ])
 }
@@ -89,7 +89,7 @@ const ExportDataModal = withWorkspaces(class ExportDataModal extends Component {
       }, ['Copy'])
     }, [
       runningSubmissionsCount > 0 && InfoTile({
-        infoStyle: warningStyle, iconName: 'warning-standard', iconColor: colors.warning(),
+        infoStyle: warningStyle, icon: { name: 'warning-standard', color: colors.warning() },
         content: `WARNING: ${runningSubmissionsCount} workflows are currently running in this workspace. ` +
           'Copying the following data could cause failures if a workflow is using this data.'
       }),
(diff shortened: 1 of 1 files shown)


src/components/ExportDataModal.js
```diff
@@ -17,9 +17,9 @@ import * as Utils from 'src/libs/utils'
 import validate from 'validate.js'
 
 
-const InfoTile = ({ infoStyle, content, iconName }) => {
+const InfoTile = ({ infoStyle, content, iconName, iconColor }) => {
```
zarsky-broad: Since `infoStyle` and `iconColor` are connected, can we wrap them together into a single object that's just passed in as a prop once?

breilly2:
@zarsky-broad Updated. Is this what you meant? I'm happy to fix if not.

src/components/ExportDataModal.js
```diff
@@ -17,9 +17,24 @@ import * as Utils from 'src/libs/utils'
...
+  border: `1px solid ${colors.warning(0.8)}`, borderLeft: 'none', borderRight: 'none',
+  backgroundColor: colors.warning(0.15),
+  padding: '1rem 1.25rem', margin: '0 -1.25rem',
+  fontWeight: 'bold', fontSize: 12
+}
+const errorStyle = {
+  ...warningStyle,
+  border: `1px solid ${colors.danger(0.8)}`,
+  backgroundColor: colors.danger(0.15)
+}
```
breilly2, line 30: These were moved unchanged so they could be used by both `InfoTile` and `ExportDataModal`.

src/components/ExportDataModal.js
```diff
@@ -17,9 +17,24 @@ import * as Utils from 'src/libs/utils'
...
+
+const InfoTile = ({ type = 'warning', content }) => {
+  const [style, shape, color] = Utils.switchCase(type,
+    ['error', () => [errorStyle, 'error-standard', colors.danger()]],
+    [Utils.DEFAULT, () => [warningStyle, 'warning-standard', colors.warning()]])
```
zarsky-broad: Why not just:
```suggestion
const InfoTile = ({ isError = false, content }) => {
  const [style, shape, color] = isError ? 
    [errorStyle, 'error-standard', colors.danger()] :
    [warningStyle, 'warning-standard', colors.warning()]
```
and then make the corresponding prop change below?

breilly2: Yup. That's the logical next refactor. Thanks. The `switchCase` made sense to me right up until I added the default value for `type`. And you're absolutely right... this is a binary property right now. No need to add unused complexity.

src/components/ExportDataModal.js
```diff
@@ -17,9 +17,24 @@ import * as Utils from 'src/libs/utils'
...
+const errorStyle = {
+  ...warningStyle,
+  border: `1px solid ${colors.danger(0.8)}`,
+  backgroundColor: colors.danger(0.15)
+}
+
+const InfoTile = ({ isError = false, content }) => {
+  const [style, shape, color] = isError ?
+    [errorStyle, 'error-standard', colors.danger()] :
+    [warningStyle, 'warning-standard', colors.warning()]
```
zarsky-broad, line 35:
```suggestion
    [warningStyle, 'warning-standard', colors.warning()]

```

zarsky-broad (approved):
Just one last suggestion for readability.

src/components/ExportDataModal.js
breilly2: I'm all for readability, but I don't see a difference here. Is it indentation that github is somehow hiding or something like that?
Screen Shot 2020-05-07 at 6 54 29 PM

PR #2127, opened 2020-05-04
Update color of COVID19 text [SATURN-1631]

ehanna4:
Checked with Lou and Adrian about this change, and they said it was ok to pull into sprint

This wasn't caught during audit since it wasn't up then but I came across this while doing another remediation and it's a pretty visible thing.


panentheos (approved):
Looks good, tested locally.

PR #2129, opened 2020-05-06
A11y contrast fixes with Lou [SATURN-1565 SATURN-1566 SATURN-1568 SATURN-1569 SATURN-1571]

zarsky-broad:
Worked through all of these with Lou, verified now passing with ANDI in-browser.


panentheos (approved):
Looks good, tested locally.

PR #2130, opened 2020-05-07
use API to register test users [SATURN-1642]

panentheos:
See ticket for details.


breilly2 (approved):
👍 Looks good. I tested by running both the run-notebook (which uses `withRegisteredUser`) and register-user (which no longer uses `withRegisteredUser`) tests locally (non-headless so I could verify what they were doing) and both passed. As usual, I love how surgically precise and effective your PRs are!

PR #2131, opened 2020-05-08
improve handling of json error messages [SATURN-1641]

panentheos:
See ticket for details. Tested locally by instrumenting various calls to generate different types of errors.


breilly2 (approved):
👍 Looks good. I tested locally by simulating various types of error responses (json, HTML, string).

We should give the Customer Success team a heads-up so they're not surprised by the error details looking different, and so they can watch out for the unlikely case that this reduces the amount of information they get from users who encounter errors.

PR #2132, opened 2020-05-08
color contrast on left table nav [SATURN-1610]

cameron-ardell:
Mob work by BHW, BR, and CA
This is for billing and the data table.
We removed low contrast row selection and replaced it with a rectangle and bold text.

Tested locally using different themes.

### DIFF
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -85,10 +85,10 @@ export const navList = {
     display: 'flex', alignItems: 'center', padding: '0 1.5rem', height: 50, fontWeight: selected ? 700 : 500,
     backgroundColor: 'white',
     borderBottom: `1px solid ${colors.dark(0.2)}`,
-    boxShadow: selected ? `inset -10px 0px 0px 0px ${terraSpecial()}` : undefined
+    boxShadow: selected ? `inset -10px 0px ${terraSpecial()}` : undefined
   }),
   itemHover: selected => selected ? {} : {
-    boxShadow: `inset -10px 0px 0px 0px ${terraSpecial(0.5)}`
+    boxShadow: `inset -6px 0px ${terraSpecial(0.5)}`
   }
 }
 
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -85,10 +85,10 @@ export const navList = {
     display: 'flex', alignItems: 'center', padding: '0 1.5rem', height: 50, fontWeight: selected ? 700 : 500,
     backgroundColor: 'white',
     borderBottom: `1px solid ${colors.dark(0.2)}`,
-    borderRight: `${selected ? 10 : 0}px solid ${terraSpecial()}`
+    boxShadow: selected ? `inset -10px 0px 0px 0px ${terraSpecial()}` : undefined
   }),
   itemHover: selected => selected ? {} : {
-    borderRight: `5px solid ${terraSpecial(0.5)}`
+    boxShadow: `inset -10px 0px 0px 0px ${terraSpecial(0.5)}`
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -14,6 +14,7 @@
   --hover-boxShadow: var(--app-hover-boxShadow);
   --hover-opacity: var(--app-hover-opacity);
   --hover-textDecoration: var(--app-hover-textDecoration);
+  --hover-borderRight: var(--app-hover-borderRight);
```
breilly2: TOL: I wonder if there's a performance cost to adding more of these. In other words, should be be judicious about adding more?

breilly2 (approved):
👍 Though I was part of the mob, after being away from it and looking at it again, I think this still looks good. I had one TOL but I don't see it as something we need to be too worried about right now. I tested locally and tested all of the themes (caveat that I have a color vision deficiency and, with the exception of Terra, the accent colors all look very similar to me).

src/style.css
zarsky-broad: I doubt it; it's only adding one css rule to the stylesheet that's loaded, and only elements that change this particular style will have the variable change as it's applied to them, otherwise it's `initial`.

panentheos:
I found a small issue while testing this, hovering over a 'reference data' tab on the Data page would cause the 'remove' button to shift position, because the border would change the content size. This was already happening on selection before, but the hover effect made it much more disruptive.

Rather than trying to do math with the padding to compensate, I tried using an inset box shadow to generate the visual indicator instead. This has the advantage of using an already existing hover property, so no changes to Interactive are needed.

Thoughts on this approach?

src/libs/style.js
```diff
@@ -82,13 +82,14 @@ export const navList = {
...
   item: selected => ({
-    display: 'flex', alignItems: 'center', padding: '0 1.5rem', height: 50, fontWeight: 500,
-    backgroundColor: selected ? colors.accent(0.1) : 'white',
-    borderBottom: `1px solid ${colors.dark(0.2)}`, borderRightStyle: 'solid',
-    borderRightWidth: selected ? 10 : 0,
-    borderRightColor: selected ? terraSpecial() : colors.accent()
+    display: 'flex', alignItems: 'center', padding: '0 1.5rem', height: 50, fontWeight: selected ? 700 : 500,
+    backgroundColor: 'white',
+    borderBottom: `1px solid ${colors.dark(0.2)}`,
+    boxShadow: selected ? `inset -10px 0px 0px 0px ${terraSpecial()}` : undefined
```
breilly2: Nit: 2nd two numbers are 0 by default so aren't needed
```suggestion
    boxShadow: selected ? `inset -10px 0px ${terraSpecial()}` : undefined
```

src/libs/style.js
```diff
@@ -82,13 +82,14 @@ export const navList = {
...
   }),
-  itemHover: selected => selected ? {} : { backgroundColor: colors.accent(0.1) }
+  itemHover: selected => selected ? {} : {
+    boxShadow: `inset -10px 0px 0px 0px ${terraSpecial(0.5)}`
```
breilly2: Optional: I kinda liked the narrower hover effect we had going. I tried 5 and that felt a little too subtle, and 7 was maybe not quite different enough from 10. 6 felt about right to me.
```suggestion
    boxShadow: `inset -6px 0px ${terraSpecial(0.5)}`
```

breilly2 (approved):
👍 Looks good. It's a clever solution. Another idea I had was to always display the border and only change the color on hover. When not hovered, it would be the same as the cell background. That would avoid having to do math but would also not allow for the change in thickness that I liked.

PR #2109, opened 2020-04-15
Update the cohort notebook template.

deflaux:
This pull request contributes an updated copy of the generic "export a cohort to a notebook" template. It also adds a template written in R, but it does not enable it.

Compared to the current Python cohort notebook, this new version:
* Adds more narrative text to explain what each cell in the notebook is doing.
* Removes the obsolete `pip install` commands
* Uses `LIMIT 1000` so that the data download is quick the very first time someone runs it. [Previously all rows were downloaded, which is not desirable when exploring a large cohort for the very first time.]
* The cohort query is now a subquery within the data query. [Previously, data was downloaded for everyone, it was not limited to only the members of the cohort.]
* Adds more detail to the provenance section.
* Adds the open source footer.

Regarding testing please see these notebooks on Terra:
* [Python cohort notebook template.ipynb](https://ukbiobank.terra.bio/#workspaces/fc-product-demo/try-saturn/notebooks/launch/Python%20cohort%20notebook%20template.ipynb?mode=preview)
* [R cohort notebook template.ipynb](https://ukbiobank.terra.bio/#workspaces/fc-product-demo/try-saturn/notebooks/launch/R%20cohort%20notebook%20template.ipynb?mode=preview)

Note that I copied and pasted those notebooks into here and merely escaped the '`' and '\n' sequences.

PR #2133, opened 2020-05-08
Fix open with R notebook to actually create an R notebook instead of Python

wnojopra:
This is a continuation of #2109 that added a generic R version of "export a cohort to a notebook". This PR enables the selection of the R notebook. 

Previously, the notebook would always be a Python notebook, even if you selected R as the language.


zarsky-broad (approved):
Tested by creating an R notebook from a cohort, the resultant file is in R.

PR #2134, opened 2020-05-08
Update Deps

ehanna4:
Co-Authored-By: Isaac Zarsky <zarsky-broad@users.noreply.github.com>

Tested locally and read changelogs


breilly2 (approved):
👍 Looks good. I also reviewed some changelogs, clicked around locally (including dropping files in since dropzone was updated), and ran in integration test locally. Everything worked as expected.

PR #2136, opened 2020-05-12
Make modal title more descriptive [SATURN-1611]

breilly2:
From what I can tell, there were 3 issues related to `/groups`. This is one of them. The other two (B9 and J1) seem to indicate the same thing that seems to have been remediated in #2114. I could use a second opinion about whether or not there is anything else to be done for https://broadworkbench.atlassian.net/browse/SATURN-1611.


panentheos (approved):
Looks good, tested locally

PR #2137, opened 2020-05-12
Add notifications for framework service expirations [SATURN-1372]

ehanna4:
This will add notifications analogous to the NIH expiration notifications when the framework services are about to expire. 

Check out the profile page for verifying that the links are correct.

Giving 5 days of notice for the 'about to expire' status

Tested locally forcing the notifications to pop up


src/components/common.js
```diff
@@ -366,6 +366,13 @@ export const ShibbolethLink = ({ children, ...props }) => {
   ])
 }
 
+export const FrameworkServicesLink = (linkText, href) => {
```
panentheos: I think this would make sense as a component, the way `ShibbolethLink` is written. (If we decided not to do that, then the function should start with a lowercase letter by convention)

src/libs/auth.js
```diff
@@ -238,6 +239,79 @@ authStore.subscribe((state, oldState) => {
   }
 })
 
+authStore.subscribe(withErrorReporting('Error loading Framework Services account status', async (state, oldState) => {
+  const loadFenceStatus = async () => {
+    try {
+      const dcpStatus = await Ajax().User.getFenceStatus('fence')
+      const dcfStatus = await Ajax().User.getFenceStatus('dcf-fence')
```
panentheos: Having both of these in the same `try` block means that a failure in one will also prevent the other one from being returned. We could separate these calls into two different subscriptions to isolate them, which would also allow the calls to happen in parallel.

src/libs/auth.js
```diff
@@ -238,6 +239,79 @@ authStore.subscribe((state, oldState) => {
...
+    const notificationId = 'fence-dcp-link-warning'
+    const now = Date.now()
+    const expireTime = state.fenceDCPStatus && addDays(30, parseJSON(state.fenceDCPStatus.issued_at))
+    const expireStatus = Utils.cond(
+      [!expireTime, () => null],
+      [now >= expireTime, () => 'has expired'],
+      [now > expireTime - (1000 * 60 * 60 * 24 * 5), () => 'will expire soon']
+    )
+    const redirectUrl = `${window.location.origin}/${Nav.getLink('fence-callback')}`
+    const href = await Ajax().User.getFenceAuthUrl('fence', redirectUrl)
```
panentheos: Since this requires an ajax call, we could save some network traffic by moving it inside the `if` so it only happens if we need to show the link. Alternatively, since the profile page also needs this url, we could move this inside of `FrameworkServicesLink` itself (after converting it to a component). Wherever it ends up, we should make sure the call is wrapped in error reporting.

src/libs/auth.js
```diff
@@ -238,6 +239,79 @@ authStore.subscribe((state, oldState) => {
...
+    const expireTime = state.fenceDCFStatus && addDays(30, parseJSON(state.fenceDCFStatus.issued_at))
+    const expireStatus = Utils.cond(
+      [!expireTime, () => null],
+      [now >= expireTime, () => 'has expired'],
+      [now > expireTime - (1000 * 60 * 60 * 24 * 5), () => 'will expire soon']
+    )
+    const redirectUrl = `${window.location.origin}/${Nav.getLink('fence-callback')}`
+    const href = await Ajax().User.getFenceAuthUrl('dcf-fence', redirectUrl)
+    if (expireStatus) {
+      notify('info', div({}, [
```
panentheos: nit: If there are no props we can omit the object entirely.

src/libs/auth.js
```diff
@@ -238,6 +239,79 @@ authStore.subscribe((state, oldState) => {
...
+      ]), { id: notificationId })
+    } else {
+      clearNotification(notificationId)
+    }
+  }
+})
+
+//DCF
+authStore.subscribe(async (state, oldState) => {
+  if (state.fenceDCFStatus !== oldState.fenceDCFStatus) {
```
panentheos: There's a fair bit of similarity between these two subscriptions, so we could extract out a helper with the common code. There *might* be a case for trying to generalize over the NIH subscription as well, though that's less of a sure thing. It depends on how much we intend to unify those experiences.

panentheos (commented):
The overall approach here is good, but see comments.

src/libs/auth.js
breilly2: The structure of the code is also very similar to loading `nihStatus` above so there may be an opportunity to extract some common code.

src/libs/auth.js
```diff
@@ -1,12 +1,12 @@
-import { parseJSON } from 'date-fns/fp'
+import { addDays, parseJSON } from 'date-fns/fp'
 import _ from 'lodash/fp'
-import { Fragment } from 'react'
```
breilly2, line 3: We need to keep `Fragment`. It's still used on line 230.

breilly2 (commented):
Functionality seems good. I tested locally with the following ajax override:
```
ajaxOverridesStore.set([{ filter: 'https://broad-bond-dev.appspot.com/api/link/v1/(dcf-)?fence$', fn: () => () => {
    return new Response(JSON.stringify({ 'username': 'breilly', 'issued_at': new Date(Date.now() - (1000 * 60 * 60 * 24 * 25)) }))
}}])
```
Changing `25` to `31` to test the expired state.

src/libs/auth.js
ehanna4: Yeah not sure how/why that got removed haha

src/components/common.js
```diff
@@ -366,6 +367,31 @@ export const ShibbolethLink = ({ children, ...props }) => {
...
+  if (href) {
+    return h(Link, {
+      href,
+      style: { display: 'inline-flex', alignItems: 'center' }, ...Utils.newTabLinkProps
+    }, [
+      linkText,
+      icon('pop-out', { size: 12, style: { marginLeft: '0.2rem' } })
+    ])
+  } else {
+    return (h(Fragment, [linkText]))
```
zarsky-broad: don't need the outer parens around the `h(Fragment)`

src/components/common.js
```diff
@@ -366,6 +367,31 @@ export const ShibbolethLink = ({ children, ...props }) => {
...
+
+  Utils.useOnMount(() => {
+    const loadAuthUrl = withErrorReporting('Error getting Fence Link', async (provider, redirectUrl) => {
+      const result = await Ajax().User.getFenceAuthUrl(provider, redirectUrl)
+      setHref(result.url)
+    })
+    loadAuthUrl(provider, redirectUrl)
+  })
+
+  if (href) {
```
zarsky-broad: It would be more consistent with our usual style to write this as a ternary, I think, i.e. 
```js
return !!href ?
  ...Link... :
  ...Fragment...
```

src/components/common.js
```diff
@@ -366,6 +367,31 @@ export const ShibbolethLink = ({ children, ...props }) => {
...
+    return h(Link, {
+      href,
+      style: { display: 'inline-flex', alignItems: 'center' }, ...Utils.newTabLinkProps
```
zarsky-broad: nit: if href and style are on separate lines, I think the `newTabLinkProps` should probably be on its own

src/libs/auth.js
```diff
@@ -183,26 +184,55 @@ export const refreshTerraProfile = async () => {
   authStore.update(state => ({ ...state, profile }))
 }
 
+//helpers for external services statuses
+
+const loadStatus = async func => {
+  try {
+    return await func
+  } catch (error) {
+    if (error.status === 404) {
```
zarsky-broad, line 193: The result of this is that we'll get an empty object for a 404 error and undefined for any other error, is this still what we want to have happen? Just checking since it's more generic now.

src/components/common.js
```diff
@@ -366,6 +367,31 @@ export const ShibbolethLink = ({ children, ...props }) => {
   ])
 }
 
+
+export const FrameworkServicesLink = ({ linkText, provider, redirectUrl }) => {
```
zarsky-broad: Plural -> singular, `FrameworkServiceLink`

src/libs/auth.js
```diff
@@ -183,26 +184,55 @@ export const refreshTerraProfile = async () => {
   authStore.update(state => ({ ...state, profile }))
 }
 
+//helpers for external services statuses
+
+const loadStatus = async func => {
```
zarsky-broad: should the argument to this be called `promise` instead of `func`?

src/libs/auth.js
```diff
@@ -217,10 +247,11 @@ authStore.subscribe((state, oldState) => {
   if (state.nihStatus !== oldState.nihStatus) {
     const notificationId = 'nih-link-warning'
     const now = Date.now()
+
```
zarsky-broad: why this newline?

src/libs/auth.js
zarsky-broad: Also, `fetchServiceStatus`, maybe?

src/libs/auth.js
```diff
@@ -183,26 +184,55 @@ export const refreshTerraProfile = async () => {
...
+  } catch (error) {
+    if (error.status === 404) {
+      return {}
+    } else {
+      throw error
+    }
+  }
+}
+
+const fenceNotify = (status, oldStatus, notificationId, serviceName, provider) => {
```
zarsky-broad: So, since this is only used with state subscription, I wonder if it might make sense to refactor it so its signature was
```js
(stateKey, notificationId, serviceName, provider) => (state, oldState) => {
  const status = state[stateKey]
  const oldStatus = oldState[stateKey]

  ...
}
```
because then we could call it like
```js
authStore.subscribe(fenceNotify('fenceDCPStatus', 'fence-dcp-link-warning', 'DCP', 'fence'))
```

src/libs/auth.js
```diff
@@ -183,26 +184,55 @@ export const refreshTerraProfile = async () => {
...
+    const expireStatus = Utils.cond(
+      [!expireTime, () => null],
+      [now >= expireTime, () => 'has expired'],
+      [now > expireTime - (fiveDays), () => `will expire in ${getDay(expireTime - (fiveDays))} days`]
+    )
+    if (expireStatus) {
+      notify('info', div([
+        `Your access to ${serviceName} Framework Services ${expireStatus}. To `,
+        expireStatus === 'has expired' ? 'restore ' : 'renew ',
+        `access, log-in to Framework Services to `,
```
zarsky-broad: don't need a template string here

src/components/common.js
panentheos: FYI: `Link` can accept a `disabled` prop and will provide a reasonable UX (grey, not clickable), so we could use that here instead of the conditional if we want.

src/components/common.js
```diff
@@ -366,6 +367,31 @@ export const ShibbolethLink = ({ children, ...props }) => {
...
+  const [href, setHref] = useState()
+
+  Utils.useOnMount(() => {
+    const loadAuthUrl = withErrorReporting('Error getting Fence Link', async (provider, redirectUrl) => {
```
panentheos: nit: since `provider` and `redirectUrl` are already accessible in the outer scope, we could just reference them directly instead of passing them.

zarsky-broad (approved):
Tested in browser, all seems to work, so good there pending some comments.

src/pages/Profile.js
```diff
@@ -173,79 +175,34 @@ const FenceLink = ({ provider, displayName }) => {
...
-        throw error
-      }
-    }
-  })
-
-  const linkFenceAccount = _.flow(
-    withErrorReporting('Error linking NIH account'),
-    Utils.withBusyState(setIsLinking)
-  )(async () => {
+  const linkFenceAccount = Utils.withBusyState(setIsLinking)(async () => {
```
panentheos: Since the underlying link call can fail, I think we still want to keep the error reporting wrapper here.

src/libs/auth.js
```diff
@@ -183,26 +184,55 @@ export const refreshTerraProfile = async () => {
...
+  if (status !== oldStatus) {
+    const redirectUrl = `${window.location.origin}/${Nav.getLink('fence-callback')}`
+
+    const now = Date.now()
+    const expireTime = status && addDays(30, parseJSON(status.issued_at))
+    const fiveDays = 1000 * 60 * 60 * 24 * 5
+    const expireStatus = Utils.cond(
+      [!expireTime, () => null],
+      [now >= expireTime, () => 'has expired'],
+      [now > expireTime - (fiveDays), () => `will expire in ${getDay(expireTime - (fiveDays))} days`]
```
panentheos: The calculation of the message here is not showing what we want. `getDay` shows the day of the week, where we probably want to use `differenceInDays` between `now` and `expireTime`.

panentheos (approved):
Looks good overall, but see comments.

src/libs/auth.js
ehanna4: This is how the NIH status check had been handling it (check out line 193 of the old code) so I figured I'd leave that how it was

ehanna4:
Re-tested extensively with new changes

PR #2138, opened 2020-05-14
Allow creating sets of sets and launching in a single submission [SATURN-1321]

breilly2:
Treat the "chooseSets" case more-or-less the same as the "chooseRows" case. This relates to https://broadworkbench.atlassian.net/browse/SATURN-1321 and the discussions today in #saturn and #saturn-team slack channels.

I'm not entirely sure why we didn't do this in the first place. Maybe we thought that Rawls didn't support it? It turns out that Rawls does support sets of sets (though Orch's TSV import doesn't). If anyone remembers anything different, please comment.

I tested this with a simple "echo_string" workflow with sample_set as the root entity type:
* selecting a single existing set, which launches that one workflow in a single submission as before
* selecting multiple sets, which creates a sample_set_set containing the selected sets and launches multiple workflows in a single submission as desired

I addressed the core functionality bits that allowed me to test. There are still a few instances of `chooseSets` that need a closer look than I've given; I've put `???` comments near those. It occurs to me now that we'd also want to audit occurrences of `chooseRows` to see if those conditions should also apply to `chooseSets`.

### DIFF
--- a/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
+++ b/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
@@ -7,7 +7,6 @@ import Modal from 'src/components/Modal'
 import { Ajax, ajaxCaller } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { createEntitySet } from 'src/libs/data-utils'
-import { reportError } from 'src/libs/error'
 import * as Utils from 'src/libs/utils'
 import {
   chooseRows, chooseSetComponents, chooseSets, processAll, processAllAsSet, processMergedSet
@@ -18,8 +17,7 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
   constructor(props) {
     super(props)
     this.state = {
-      launching: undefined, message: undefined, launchError: undefined,
-      multiLaunchCompletions: undefined, multiLaunchErrors: undefined
+      launching: undefined, message: undefined, launchError: undefined
     }
   }
 
@@ -60,7 +58,7 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
 
   render() {
     const { onDismiss, entitySelectionModel: { type }, entityMetadata, config: { rootEntityType }, processSingle } = this.props
-    const { launching, message, multiLaunchCompletions, launchError, multiLaunchErrors } = this.state
+    const { launching, message, launchError } = this.state
     const entities = this.getEntities()
     const entityCount = Utils.cond(
       [processSingle, () => 1],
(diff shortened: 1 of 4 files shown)


breilly2:
One thing I thought of is that TSV download/upload won't work for sets of sets. While this process doesn't depend on that feature, it will create `*_set_set` entities in the workspace. We may want to detect this and disable download/upload so as not to imply that we support that type of round-trip operation with these the way we do for other entity types.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -215,34 +194,4 @@ export default ajaxCaller(class LaunchAnalysisModal extends Component {
       this.setState({ launchError: await error.text(), message: undefined })
     }
   }
-
```
breilly2, line 218: I'm sorry, @zarsky-broad . It pains me to remove code that you put good work into. I only do it because, with support for sets of sets, I don't think we'll be needing it anymore. :(

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -325,6 +325,8 @@ const WorkflowView = _.flow(
   resetSelectionModel(value, selectedEntities = {}, entityMetadata = this.state.entityMetadata) {
     const { workflowName } = this.props
 
+    // If the default for non-set types changes from `processAllAsSet` then the calculation of `noLaunchReason` in `renderSummary` needs to be updated accordingly.
+    // Currently, `renderSummary` assumes that it is not possible to have nothing selected for non-set types.
```
breilly2, line 329: `chooseSets` is now somewhat redundant with `chooseRows`, but I've left it in order to keep this PR relatively small and because it still reflects the current user experience. @panentheos and I discussed this and concluded that there are 2 (potentially separate) efforts we could undertake:

1. Refactor the code for simplicity
2. Rework the UI for simplicity and clarity

As part of understanding the changes required for this PR, I audited all occurrences of `chooseSet`. While doing so, I concluded that the conditions in this reset and the conditions the logic for validating the selection model below are linked. Therefore, I've added comments in both places to help future refactor/rework efforts.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
zarsky-broad: The only part it should pain you to remove is my beautiful sweet progress bar that almost nobody got to see.

panentheos (approved):
Looks good. Tested locally by launching against sets and sets of sets.

TOL: It's worth calling out that while this does allow creating sets of sets during launch, it doesn't provide a mechanism for selecting and re-using those set sets for future launches. This is because when the root entity type is a set, the option labelled 'Choose sets' actually means 'Choose rows from the root entity type', and there's no way to select from existing sets of those (e.g. set sets). We will almost certainly want to follow this with some UI improvements to introduce more consistency here, but this does solve the immediate problem.

breilly2:
That's a good point. We should definitely explore that as we consider remodeling this UI. For now, it may be inconvenient for these users... or, since they're thinking of sets more as their entities than as a unit of work for a submission, maybe they won't be inconvenienced.

PR #2143, opened 2020-05-15
handle rp with igv [SATURN-1613]

zarsky-broad:
Some interesting patterns here working within the limitations - I'm curious to hear folks' thoughts.

### DIFF
--- a/src/components/IGVBrowser.js
+++ b/src/components/IGVBrowser.js
@@ -1,45 +1,76 @@
 import _ from 'lodash/fp'
 import { Fragment, useRef, useState } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
+import { requesterPaysWrapper } from 'src/components/bucket-utils'
 import { Link } from 'src/components/common'
 import { centeredSpinner, icon } from 'src/components/icons'
-import { saToken } from 'src/libs/ajax'
+import { Ajax, saToken } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { getUserProjectForWorkspace, parseGsUri } from 'src/libs/data-utils'
 import { reportError } from 'src/libs/error'
-import { knownBucketRequesterPaysStatuses } from 'src/libs/state'
+import { knownBucketRequesterPaysStatuses, requesterPaysProjectStore } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
 // format for selectedFiles prop: [{ filePath, indexFilePath } }]
-const IGVBrowser = ({ selectedFiles, refGenome, workspace, onDismiss }) => {
+const IGVBrowser = _.flow(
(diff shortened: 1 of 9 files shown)


src/components/IGVBrowser.js
```diff
@@ -5,29 +5,36 @@ import { Link } from 'src/components/common'
...
     const options = {
       genome: refGenome,
-      tracks: _.map(({ filePath, indexFilePath }) => ({
-        name: `${_.last(filePath.split('/'))} (${filePath})`,
-        url: filePath,
-        indexURL: indexFilePath
-      }), selectedFiles)
+      tracks: _.map(({ filePath, indexFilePath }) => {
+        const [bucket] = parseGsUri(filePath)
+        const userProjectParam = { userProject: knownBucketRequesterPaysStatuses.get()[bucket] ? getUserProjectForWorkspace(workspace) : undefined }
```
s-rubenstein: Should we be prompting the user for the bucket if it is not in our list of known buckets?

src/components/bucket-utils.js
```diff
@@ -22,18 +22,23 @@ export const withRequesterPaysHandler = _.curry((handler, fn) => async (...args)
 export const requesterPaysWrapper = ({ onDismiss }) => WrappedComponent => {
   return Utils.forwardRefWithName('requesterPaysWrapper', (props, ref) => {
     const [showModal, setShowModal] = useState(false)
+    const [onSuccess, setOnSuccess] = useState(() => {})
```
s-rubenstein: Can we make this more specific than onSuccess? Like maybe `onRequesterPaysProjectSelected`?

Also, could we make this a prop passed into the wrapper? It feels kind of strange that it needs to be set on the error for `onSuccess` to be called.

src/libs/data-utils.js
```diff
@@ -31,6 +32,10 @@ const warningBoxStyle = {
 
 const errorTextStyle = { color: colors.danger(), fontWeight: 'bold', fontSize: 12, marginTop: '0.5rem' }
 
+export const parseGsUri = uri => _.drop(1, /gs:[/][/]([^/]+)[/](.+)/.exec(uri))
+
+export const getUserProjectForWorkspace = workspace => workspace && Utils.canWrite(workspace.accessLevel) ? workspace.workspace.namespace : requesterPaysProjectStore.get()
```
breilly2: Is `canWrite` correct here? I think this may need to be more like `ajax.js` > `withRequesterPays` > `canUseWorkspaceProject`.

breilly2 (commented):
Comments from mobbing w/ Cameron and Scout.

s-rubenstein (commented):
Code LGTMob, needs testing.

s-rubenstein (approved):
Tested! LGTMob

PR #2145, opened 2020-05-18
Add missing ShibbolethURLRoots

ehanna4:
Adding missing ShibbolethURLRoots to the configs where it was missing to help unblock a BDC user.


zarsky-broad (approved):
This is in line with the config for Firecloud UI.

PR #2146, opened 2020-05-19
 remediate terms of service [SATURN-1609]

cameron-ardell:
Part of this ticket was already done (adding alt text to an image), so I just fixed the other issue, that unhovered links were not discernable enough (weak color ratio). The guidelines for accessibility state that links should be underlined by default, which I ran by Lou, and stated that therefore it's not necessary to update the color ratio of link to body text.
This issue should now be fixed anywhere we parse markdown text in the site.


src/style.css
```diff
@@ -105,9 +105,11 @@ a {
 
 /* Styles for content rendered by marked (also used by react-simplemde-editor) */
 .markdown-body a, .simplemde-container .editor-preview a {
-  color: #4d72aa;
+  color: #4D72AA;
```
s-rubenstein: I noticed this color exists in `colors.JS`. Do you think it would be possible to plug that in here? (Recognizing that it is a CSS file, but it may be worth trying to figure out if we can get it in somehow.)

s-rubenstein (approved):
This seems reasonable to me. I would love some screenshots if possible? But not needed. I left one comment about trying to plug in `colors.js` but it definitely seems out of scope for this ticket.

src/style.css
cameron-ardell: it is interesting that we do all caps sometimes not otherwise, but they are the same shade.
I'm looking around at different use-cases in the code.

I feel like we should just stick to one but I'm not sure which one

cameron-ardell: Looking for any thoughts on this!

PR #2147, opened 2020-05-19
Add workflow source and snapshotId [SATURN-1621]

s-rubenstein:
This adds the source and snapshot ID to the bard reporting. I noticed that we also report here: https://github.com/DataBiosphere/terra-ui/blob/dev/src/pages/workspaces/workspace/workflows/FailureRerunner.js#L64 so I talked to Adrian. If this is an automated retry, we don't need all that info, because it isn't a user triggering it. If it is a manual process, we do. It looked automated to me, but I would love experienced eyes to confirm.

### DIFF
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -397,11 +397,17 @@ const WorkflowView = _.flow(
     // savedConfig: unmodified copy of config for checking for unsaved edits
     // modifiedConfig: active data, potentially unsaved
     const {
-      isFreshData, savedConfig, entityMetadata, launching, activeTab, useCallCache, deleteIntermediateOutputFiles,
+      currentSnapRedacted, isFreshData, savedConfig, entityMetadata, launching, activeTab, useCallCache, deleteIntermediateOutputFiles,
       entitySelectionModel, variableSelected, modifiedConfig, updatingConfig
     } = this.state
+    const { methodVersion, methodNamespace, methodName, methodPath, sourceRepo } = modifiedConfig.methodRepoMethod
     const { namespace, name, workspace } = this.props
     const workspaceId = { namespace, name }
+    const workflowMetadata = {
+      // Source follows the logic of the rendering on the page. It is not identical because the rendering requires elements to be included.
+      snapshotId: modifiedConfig.methodRepoMethod.methodVersion, source: sourceRepo === 'agora' || currentSnapRedacted ?
+        `${methodNamespace}/${methodName}/${methodVersion}` : `${methodPath}:${methodVersion}`
+    }
     return h(Fragment, [
       savedConfig && h(Fragment, [
         this.renderSummary(),
@@ -416,11 +422,11 @@ const WorkflowView = _.flow(
           processSingle: this.isSingle(), entitySelectionModel, useCallCache, deleteIntermediateOutputFiles,
           onDismiss: () => this.setState({ launching: false }),
           onSuccess: submissionId => {
-            Ajax().Metrics.captureEvent(Events.workflowLaunch, { multi: false })
+            Ajax().Metrics.captureEvent(Events.workflowLaunch, { ...workflowMetadata, multi: false })
             Nav.goToPath('workspace-submission-details', { submissionId, ...workspaceId })
           },
           onSuccessMulti: () => {
-            Ajax().Metrics.captureEvent(Events.workflowLaunch, { multi: true })
+            Ajax().Metrics.captureEvent(Events.workflowLaunch, { ...workflowMetadata, multi: true })
             Nav.goToPath('workspace-job-history', workspaceId)
           }
         }),
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -397,21 +397,11 @@ const WorkflowView = _.flow(
     // savedConfig: unmodified copy of config for checking for unsaved edits
     // modifiedConfig: active data, potentially unsaved
     const {
-      currentSnapRedacted, isFreshData, savedConfig, entityMetadata, launching, activeTab, useCallCache, deleteIntermediateOutputFiles,
+      isFreshData, savedConfig, entityMetadata, launching, activeTab, useCallCache, deleteIntermediateOutputFiles,
       entitySelectionModel, variableSelected, modifiedConfig, updatingConfig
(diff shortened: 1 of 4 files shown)


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -395,11 +395,17 @@ const WorkflowView = _.flow(
...
-      isFreshData, savedConfig, entityMetadata, launching, activeTab, useCallCache, deleteIntermediateOutputFiles,
+      currentSnapRedacted, isFreshData, savedConfig, entityMetadata, launching, activeTab, useCallCache, deleteIntermediateOutputFiles,
       entitySelectionModel, variableSelected, modifiedConfig, updatingConfig
     } = this.state
+    const { methodVersion, methodNamespace, methodName, methodPath, sourceRepo } = modifiedConfig.methodRepoMethod
     const { namespace, name, workspace } = this.props
     const workspaceId = { namespace, name }
+    const workflowMetadata = {
+      // Source follows the logic of the rendering on the page. It is not identical because the rendering requires elements to be included.
+      snapshotId: modifiedConfig.methodRepoMethod.methodVersion, source: sourceRepo === 'agora' || currentSnapRedacted ?
```
s-rubenstein: I also noticed I had to build the source by hand. It wasn't quite reusable/duplicated code, so I didn't duplicate it, but that may be my inexperience with hyperscript. I'd love some eyes on that too.

breilly2:
I'm pretty sure `rerunFailures` is for the user gesture on the Job History tab that is offered when some jobs in a submission have failed. Seems to me like we should probably capture the same events in both places.

A little bit of history... when event capturing was originally going to be done my a 3rd-party contractor, there was a detailed spec created for all of the events to be captured: https://docs.google.com/spreadsheets/d/1wNmVyYbWbggRegJw4_n0HBfY6j2Qsz8pJTdcmeisvJ4/edit#gid=1641519238
Specifically, look at the "Implementation Spec" tab. The relevant rows in that spreadsheet are 184 ("Workflow Run Started") and 242 ("Relaunch Job"). The properties in both cases are the same so, while workflow source and snapshot are not in this spec, it looks to me like the intent is to capture the same data in both cases.

s-rubenstein:
Got it, thank you @breilly2. Will update

s-rubenstein:
I changed our rerunner to submit its own event, so that we can track those separately (especially since they have considerably different information now)

I also added namespace and name to the rerun event, although I wasn't sure if I should submit them as one field or as two.

Per slack discussion, it is currently very difficult to get information on the source and snapshot for reruns, because they are stored in the configuration, and not the submission. On the list page, we only get submission information. Fixing this to submit that information on retry would probably be rawls changes. That aside, from a product perspective, it is perfectly alright to treat reruns separately, because there are no edits made, so we could consider it as part of the same run as before.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -414,11 +424,11 @@ const WorkflowView = _.flow(
           processSingle: this.isSingle(), entitySelectionModel, useCallCache, deleteIntermediateOutputFiles,
           onDismiss: () => this.setState({ launching: false }),
           onSuccess: submissionId => {
-            Ajax().Metrics.captureEvent(Events.workflowLaunch, { multi: false })
+            Ajax().Metrics.captureEvent(Events.workflowLaunch, { ...workflowMetadata, multi: false })
             Nav.goToPath('workspace-submission-details', { submissionId, ...workspaceId })
           },
           onSuccessMulti: () => {
```
panentheos: #2138 removed `onSuccessMulti` so we can actually delete this case entirely, and remove the `multi` property.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -395,11 +395,21 @@ const WorkflowView = _.flow(
...
     } = this.state
     const { namespace, name, workspace } = this.props
     const workspaceId = { namespace, name }
+    let workflowMetadata
+    if (savedConfig) {
+      const { methodRepoMethod: { methodVersion, methodNamespace, methodName, methodPath, sourceRepo } } = modifiedConfig
+
+      workflowMetadata = {
+        // Source follows the logic of the rendering on the page. It is not identical because the rendering requires elements to be included.
+        snapshotId: modifiedConfig.methodRepoMethod.methodVersion, source: sourceRepo === 'agora' || currentSnapRedacted ?
```
panentheos: I think there's actually a minor display bug in the render logic. There are currently two different method repos, dockstore and agora. Which fields are set depends on which repo the method came from. The render logic assumes that the redacted case only applies to agora, but versions can also be removed in dockstore, which results in trying to render the wrong fields.

I think the correct logic there would be to generate the display string based on the `sourceRepo` in all cases, and just have `currentSnapRedacted` control whether it's linkified or not. Then, if we want to use the same format here, we could extract out a function for computing the string from a config.

That being said, I wonder if we should be logging the fields separately, to make querying easier. E.g. if I want to see launches for any version of method X, that might be a bit harder if all I have is the string 'X/V'

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -395,11 +395,21 @@ const WorkflowView = _.flow(
     // savedConfig: unmodified copy of config for checking for unsaved edits
     // modifiedConfig: active data, potentially unsaved
     const {
-      isFreshData, savedConfig, entityMetadata, launching, activeTab, useCallCache, deleteIntermediateOutputFiles,
+      currentSnapRedacted, isFreshData, savedConfig, entityMetadata, launching, activeTab, useCallCache, deleteIntermediateOutputFiles,
       entitySelectionModel, variableSelected, modifiedConfig, updatingConfig
     } = this.state
     const { namespace, name, workspace } = this.props
     const workspaceId = { namespace, name }
+    let workflowMetadata
```
panentheos: I suspect this can go away if this is inlined after deleting the unused code below, but just as an FYI:

We avoid using `let` unnecessarily because it effectively creates mutable state, which can make code harder to reason about. Cases like this can be expressed like:

```js
const maybeValue = computeTheValue ? { x: 1 } : undefined
```

If the computation is more complex, it could also be split out as a locally-scoped helper function, which could then be called conditionally.

By being strict about this, we can treat the (necessary) `let` as a strong signal to readers that says: 'careful, there is state here'

panentheos (approved):
LGTMob. Tested by launching and relaunching workflows.

PR #2148, opened 2020-05-21
Add workspace details to metric events [SATURN-1624]

ehanna4:
Add the workspace name and namespace to all the events that are workspace related. 

Currently the event list is:
- applicationLaunch,
- notebookLaunch,
- pageView,
- workflowImport,
- workflowLaunch,
- workspaceDataImport,
- workspaceShare,
- Rerun workflow

So now workflowImport, workflowLaunch, workspaceDataImport, workspaceShare will have the workspace details (note notebookLaunch already got those details added in another PR).

Tested running locally then checking the MixPanel dashboard to ensure the event data went through.
BEFORE:
workflow import before
Screen Shot 2020-05-21 at 10 12 41 AM
Screen Shot 2020-05-21 at 10 12 55 AM
Screen Shot 2020-05-21 at 10 13 15 AM

AFTER:
workflow import after
workflow launch after
workspace share after
data import after


src/pages/ImportData.js
```diff
@@ -108,7 +108,7 @@ const ImportData = () => {
         notify('success', 'Data imported successfully.', { timeout: 3000 })
       }]
     )
-    Ajax().Metrics.captureEvent(Events.workspaceDataImport)
+    Ajax().Metrics.captureEvent(Events.workspaceDataImport, { 'Workspace Name': name, 'Workspace Namespace': namespace })
```
s-rubenstein: The strings 'Workspace Name' and 'Workspace Namespace' are repeated a bunch here. Could we either extract them as string constants, or create a helper function that takes a workspace returns that object (which we could destructure into the event metadata?)

I probably prefer the second one somewhat, because it would make it easier to add workspace metadata to all workspace metrics moving forwards?

panentheos: Seconded, I like the helper function idea.

Also, I think we should use camelcase identifiers for properties, e.g. `workspaceName`. I'm pretty sure there's a feature in Mixpanel UI to create friendly name mappings, so it would be better to keep the low level representation more 'data-like'.

s-rubenstein (approved):
LGTMob

PR #2149, opened 2020-05-21
Add encodeURIComponent any time an email is used in a URI [SATURN-1656]

breilly2:
Tested with group management by adding and removing `marc.test.terra%gmail.com@gtempaccount.com` from a group which succeeded. Also tested billing management and workspace sharing which failed, but for reasons that should be fixed in rawls/sam.

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -211,7 +211,7 @@ const User = signal => ({
   },
 
   getProxyGroup: async email => {
-    const res = await fetchOrchestration(`api/proxyGroup/${email}`, _.merge(authOpts(), { signal }))
+    const res = await fetchOrchestration(`api/proxyGroup/${encodeURIComponent(email)}`, _.merge(authOpts(), { signal }))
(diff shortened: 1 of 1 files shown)


src/libs/ajax.js
```diff
@@ -353,11 +353,12 @@ const Groups = signal => ({
     const resourceRoot = `api/resources/v1/managed-group/${groupName}`
 
     const addRole = (role, email) => {
-      return fetchSam(`${root}/${role}/${email}`, _.merge(authOpts(), { signal, method: 'PUT' }))
+      console.log(email, encodeURIComponent(email))
```
panentheos: This looks like leftover debugging.

panentheos (approved):
Looks good, see comment. Tested locally by adding/removing emails with special chars.

src/libs/ajax.js
breilly2: You're right. Thanks for catching it!

PR #2153, opened 2020-05-26
support for import service status values [no ticket; risk: low]

davidangb:
We (app services) made a mistake: when building the new import service, we changed the job status values the service could return. This PR makes the UI resilient to those.

We've just deployed the cutover for avro-import to import service into the dev env - given the planned release schedule, it won't hit prod for at least a week. During this period, we should support both sets of statuses, which this PR handles. After we're fully on import service and avro-import is gone, we can/should remove the old status values.

I tested this by running it locally and seeing a successful import message. Without this PR, a successful backend job shows up as "unexpected error" in the UI because the UI doesn't recognize the status.

### DIFF
--- a/src/components/ImportStatus.js
+++ b/src/components/ImportStatus.js
@@ -49,14 +49,17 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
     // TODO: only need to support both sets of statuses during the transition from avro-import to import service.
     // once import servie is fully adopted, we can/should remove the avro-import status values.
 
+    const successNotify = () => notify('success', 'Data imported successfully.',
+      { message: `Data import to workspace "${namespace} / ${name}" is complete. Please refresh the Data view.` })
+
+    const errorNotify = () => notify('error', 'Error importing PFB data.', message)
+
     if (!_.includes(status, ['PENDING', 'RUNNING', 'Pending', 'Translating', 'ReadyForUpsert', 'Upserting'])) {
       Utils.switchCase(status,
-        ['SUCCESS', () => notify('success', 'Data imported successfully.',
-          { message: `Data import to workspace "${namespace} / ${name}" is complete. Please refresh the Data view.` })],
-        ['Done', () => notify('success', 'Data imported successfully.',
-          { message: `Data import to workspace "${namespace} / ${name}" is complete. Please refresh the Data view.` })],
-        ['ERROR', () => notify('error', 'Error importing PFB data.', message)],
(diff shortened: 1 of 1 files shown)


src/components/ImportStatus.js
```diff
@@ -44,11 +44,19 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
...
+    // avro-import statuses: PENDING, RUNNING, SUCCESS, ERROR
+    // import service statuses: Pending, Translating, ReadyForUpsert, Upserting, Done, Error
+    // TODO: only need to support both sets of statuses during the transition from avro-import to import service.
+    // once import servie is fully adopted, we can/should remove the avro-import status values.
+
+    if (!_.includes(status, ['PENDING', 'RUNNING', 'Pending', 'Translating', 'ReadyForUpsert', 'Upserting'])) {
       Utils.switchCase(status,
         ['SUCCESS', () => notify('success', 'Data imported successfully.',
           { message: `Data import to workspace "${namespace} / ${name}" is complete. Please refresh the Data view.` })],
+        ['Done', () => notify('success', 'Data imported successfully.',
```
s-rubenstein: Could we merge these in with the above. ('Done' || 'SUCCESS')? If not, could we pull the lambda out into a variable so we can just reuse it? Same thing with error.

davidangb: does that syntax even work?
```
Utils.switchCase(status,
        [('SUCCESS' || 'Done'), () => ...
```
I tried it and got a JS error, but possibly did something wrong. I did think about pulling the lambda out into a var, but didn't think it was worth it, since this code should only exist for a couple weeks and then return to a non-duplicated case. If you think it's worth it, I'd do it though!

s-rubenstein: I don't think that syntax works. I wasn't sure what the right syntax is. Looks like the method doesn't support boolean operations like that.

I would say we should pull out the code into a variable, mainly because it isn't too difficult and I think string duplication is pretty rough to have in code (even if only for a brief period of time)

davidangb: @blrubenstein done!

breilly2: FYI, If you wanted a more flexible expression in the test, you can use `Utils.cond`, at the expense of having to repeat the thing you're checking:
```
Utils.cond(
  [status === 'SUCCESS' || status === 'Done', () => ...],
  [_.lowerCase(status) === 'error', () => ...],
  () => ... // default case
)
```
(something like that... I didn't test the code above)

s-rubenstein (approved):
This seems reasonable to me!

PR #2154, opened 2020-05-26
Add pagename to event name [SATURN-1619]

ehanna4:
Looking for thoughts on doing pageview:<page name> or some abbreviation on 'page view' to cut down on repetition.

Add the page path the event to make funnels and flows more obvious. (could previously be done by filtering by page view event and then filtering on appPath but this way you just filter by the event which is page:view:'pagename'

Tested locally.

Example funnel
Screen Shot 2020-05-26 at 2 51 53 PM

### DIFF
--- a/src/libs/events.js
+++ b/src/libs/events.js
@@ -23,7 +23,7 @@ export const PageViewReporter = () => {
 
   useEffect(() => {
     if (isSignedIn && registrationStatus === 'registered') {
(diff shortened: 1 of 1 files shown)


s-rubenstein (approved):
I personally think it is fine to have the pageView full text present. I am usually a fan of verboseness for the sake of clarity. (And have a dislike for acronyms for that reason)

PR #2155, opened 2020-05-27
Add tooltip for enabled create new workspace '+' button [SATURN-1606]

cameron-ardell:
Previously we did not have a tooltip for the button that allowed us to create a new workspace when it was enabled, so I added that.
I also noticed we were missing it for cloning a new workspace so I added that for that case as well.


cameron-ardell:
I am noticing some redundancy with the code and am wondering if there is a better way to format everything. I am suspicious though because each switch case is slightly different.

breilly2 (approved):
👍 Looks good. I checked it on the PR site.

PR #2157, opened 2020-05-29
Remove fence notifications on signout

ehanna4:
Fence notifications are not cleared on sign-out. This fixes that.


panentheos (approved):
Looks good, tested locally.

PR #2158, opened 2020-06-01
Try to conditionalize app id [SATURN-1267]

s-rubenstein:
I have no clue how to test that this works. If anyone else has ideas, I would love to know. Is there a dev data warehouse? How would I go about accessing it?

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -3,7 +3,7 @@ import * as qs from 'qs'
 import { h } from 'react-hyperscript-helpers'
 import { version } from 'src/data/machines'
 import { getUser } from 'src/libs/auth'
-import { getConfig } from 'src/libs/config'
+import { getConfig, isAnvil, isBioDataCatalyst, isDatastage, isFirecloud, isUKBiobank } from 'src/libs/config'
 import { withErrorIgnoring } from 'src/libs/error'
 import * as Nav from 'src/libs/nav'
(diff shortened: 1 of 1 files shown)


src/libs/ajax.js
```diff
@@ -24,7 +24,17 @@ window.ajaxOverrideUtils = {
 
 const authOpts = (token = getUser().token) => ({ headers: { Authorization: `Bearer ${token}` } })
 const jsonBody = body => ({ body: JSON.stringify(body), headers: { 'Content-Type': 'application/json' } })
-const appIdentifier = { headers: { 'X-App-ID': 'Saturn' } }
+const appIdentifier = {
```
panentheos, line 27: TOL: This will work and presumably won't require any backend changes, but I think we should check in with whoever is consuming this data and make sure it won't make their analysis more difficult. In particular, queries looking for stuff from any version of the UI would need to be rewritten to include all these new tokens, and would be at risk for missing data in the future if we added one.

Another way would be to store the variant info separately, e.g. by writing out the `Origin` header to the data warehouse. That way, consumers could filter the data at whichever granularity they want. That might require backend changes, though.

s-rubenstein:
Closing out this pull request because we want to try to maintain the consistent saturn header, in order to be able to aggregate information around 'all traffic from a terra-ui hosted site'. The new approach is to try to use `Origin`, which already is sent from the UI, but would need to be propagated through to kibana from the services.

PR #2159, opened 2020-06-02
Remove unrelated COVID pics and replace with relevant pics [SATURN-1601]

cameron-ardell:
in new and interesting, it was defaulted to make them all COVID virus images. However, this isn't related to every item in the list and a different, default image should be available. Now it only shows the COVID virus image if there is "COVID" in the title

### DIFF
--- a/src/pages/library/Showcase.js
+++ b/src/pages/library/Showcase.js
@@ -46,8 +46,8 @@ const makeCard = variant => ({ workspace: { namespace, name, attributes: { descr
         width: 87,
         ...Utils.switchCase(variant,
           ['new', () => Utils.switchCase(
-            [name.toLowerCase().includes('covid'), () =>({ backgroundImage: `url(${covidBg})` })],
-            [Utils.DEFAULT, () =>({ backgroundImage: `url(${featuredBg})`, opacity: 0.75 })]
+            [name.toLowerCase().includes('covid'), () => ({ backgroundImage: `url(${covidBg})` })],
+            [Utils.DEFAULT, () => ({ backgroundImage: `url(${featuredBg})`, opacity: 0.75 })]
           )],
           ['gatk', () => ({ backgroundColor: '#333', backgroundImage: `url(${gatkLogo})`, backgroundSize: undefined })],
           [Utils.DEFAULT, () => ({ backgroundImage: `url(${featuredBg})`, opacity: 0.75 })]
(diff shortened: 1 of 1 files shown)


src/pages/library/Showcase.js
```diff
@@ -45,7 +45,10 @@ const makeCard = variant => ({ workspace: { namespace, name, attributes: { descr
         backgroundRepeat: 'no-repeat', backgroundPosition: 'center', backgroundSize: 'auto 100%', borderRadius: '0 5px 5px 0',
         width: 87,
         ...Utils.switchCase(variant,
-          ['new', () => ({ backgroundImage: `url(${covidBg})` })],
+          ['new', () => Utils.switchCase(
```
panentheos: This won't work correctly, as `switchCase` uses the first parameter as the value to test against (it's trying to use the array in this case).

We should probably consolidate this logic a bit, and roll everything into a single `Utils.cond`, which will remove the duplication in the default case, and also do the right thing in case the covid workspaces are ever moved out of the 'new' column

src/pages/library/Showcase.js
```diff
@@ -44,10 +44,10 @@ const makeCard = variant => ({ workspace: { namespace, name, attributes: { descr
       style: {
         backgroundRepeat: 'no-repeat', backgroundPosition: 'center', backgroundSize: 'auto 100%', borderRadius: '0 5px 5px 0',
         width: 87,
-        ...Utils.switchCase(variant,
-          ['new', () => ({ backgroundImage: `url(${covidBg})` })],
-          ['gatk', () => ({ backgroundColor: '#333', backgroundImage: `url(${gatkLogo})`, backgroundSize: undefined })],
-          [Utils.DEFAULT, () => ({ backgroundImage: `url(${featuredBg})`, opacity: 0.75 })]
+        ...Utils.cond(
+          [name.toLowerCase().includes('covid'), { backgroundImage: `url(${covidBg})` }],
+          [variant === 'gatk', () => ({ backgroundColor: '#333', backgroundImage: `url(${gatkLogo})`, backgroundSize: undefined })],
```
breilly2, line 49: If a workspace card is both a `gatk` variant and contains `covid` in the name, which image should we use? As is, this will favor the COVID background. If that's not the desire, these lines will need to be swapped.

cameron-ardell: Ahh that's a good point. Is there a person I should talk to in order to clarify which to prioritize?

breilly2: I would start with Amelia or Adrian. If they can't answer the question, I bet they could tell you who to ask (probably someone in Customer Success).

PR #2160, opened 2020-06-03
Filter runtime list by creator to make sure we delete the correct runtimes [SATURN-1674]

breilly2:
~Also fixes an inconsistency between how test and non-test code filter runtimes when selecting the "current" one.~

Also delete all of the test user's runtimes.

### DIFF
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -110,42 +110,26 @@ const withBilling = test => async options => {
   try {
     await test({ ...options })
   } finally {
-    await deleteCluster(options)
+    await deleteRuntimes(options)
     await removeUserFromBilling(options)
   }
 }
 
-const trimClustersOldestFirst = _.flow(
-  _.remove({ status: 'Deleting' }),
-  _.sortBy('createdDate')
-)
-
-const currentCluster = _.flow(trimClustersOldestFirst, _.last)
-
-const getCurrentCluster = withUserToken(async ({ billingProject, context, email, testUrl, token }) => {
+const deleteRuntimes = withUserToken(async ({ billingProject, context, email, testUrl, token }) => {
   const ajaxPage = await context.newPage()
   await ajaxPage.goto(testUrl)
   await signIntoTerra(ajaxPage, token)
 
-  const clusters = await ajaxPage.evaluate((billingProject, email) => {
+  const runtimes = await ajaxPage.evaluate((billingProject, email) => {
     return window.Ajax().Clusters.list({ googleProject: billingProject, creator: email })
   }, billingProject, email)
 
-  await ajaxPage.close()
-  return currentCluster(clusters)
-})
-
-const deleteCluster = withUserToken(async ({ billingProject, context, email, testUrl, token }) => {
-  const ajaxPage = await context.newPage()
-  await ajaxPage.goto(testUrl)
-  await signIntoTerra(ajaxPage, token)
-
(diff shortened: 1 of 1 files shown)


integration-tests/utils/integration-helpers.js
```diff
@@ -117,34 +117,33 @@ const withBilling = test => async options => {
...
+  const clusters = await ajaxPage.evaluate((billingProject, email) => {
+    return window.Ajax().Clusters.list({ googleProject: billingProject, creator: email })
+  }, billingProject, email)
 
   await ajaxPage.close()
   return currentCluster(clusters)
 })
 
-const deleteCluster = withUserToken(async ({ billingProject, context, testUrl, token }) => {
+const deleteCluster = withUserToken(async ({ billingProject, context, email, testUrl, token }) => {
```
panentheos: Since this is using `withUserToken`, it's using Scarlet Flowerpicker's token to find and delete the cluster. Do we want to continue doing that? I can see arguments both ways (see other comment as well).

integration-tests/utils/integration-helpers.js
```diff
@@ -117,34 +117,33 @@ const withBilling = test => async options => {
...
   await ajaxPage.goto(testUrl)
   await signIntoTerra(ajaxPage, token)
 
-  const currentC = await getCurrentCluster({ billingProject, context, testUrl })
-  currentC && await ajaxPage.evaluate((currentC, billingProject) => {
-    return window.Ajax().Clusters.cluster(billingProject, currentC.runtimeName).delete()
-  }, currentC, billingProject)
+  const currentC = await getCurrentCluster({ billingProject, context, email, testUrl })
+  currentC && await ajaxPage.evaluate(currentC => {
+    return window.Ajax().Clusters.cluster(currentC.googleProject, currentC.runtimeName).delete()
```
panentheos: TOL: Since this is effectively a cleanup operation, I wonder if it would be better to find and delete ALL clusters, rather than just the current one. We only expect there to be one right now, but we could make it resilient to 0 or N to cover edge cases and/or future changes to the test code.

breilly2: We could do that, as long as we only delete runtimes for the current user.

A happy side effect would be eliminating the duplicate implementations of `trimClustersOldestFirst`. We probably still want to filter out runtimes that are already `Deleting`, but that makes logical sense and wouldn't be a duplication of intent with the UI code.

I'll sketch this out and either push another commit or post a follow-up comment here.

integration-tests/utils/integration-helpers.js
breilly2: I could go either way. I'm inclined to leave it as is for now, but I'm curious about your arguments. Either choice has the advantage of possibly exposing a bug in one of two code paths in leonardo. Using the user token would avoid the potential bug of deleting the wrong clusters. However, we should probably do some clean-up/refactoring to make it clear what tokens we're using where. I was just looking and found it really hard to trace... and I think we're even switching up which token is being passed through without calling it anything different than `token`. I'd rather we take that on in a separate PR and not muddy it further in this PR.

panentheos: I was initially surprised that the deletion was being done by a different user than the creation. But after thinking about this more as a general cleanup step, it doesn't seem that unreasonable, since S.F. is essentially an admin user.

I agree that there's some cleanup and clarifying we could do here, and I also agree that it doesn't need to go in this PR.

panentheos (approved):
LGTMob

PR #2161, opened 2020-06-05
refactor fence linking [SATURN-1686]

panentheos:
This unifies the code dealing with the existing fence link providers, since it looks like we're leaning into having a uniform experience across the various providers. This will make adding additional providers (e.g. Anvil) straightforward.

Also fixes a bug where current link information wasn't shown on the profile.

Testing note: I was not able to test the full flow locally, since the auth server rejects `localhost` as a return url. Still looking for a way to test end-to-end.

### DIFF
--- a/src/libs/providers.js
+++ b/src/libs/providers.js
@@ -1,0 +1,9 @@
+export const allProviders = ['fence', 'dcf-fence', 'anvil']
+
+export const providerName = provider => {
+  return {
+    fence: 'DCP',
+    'dcf-fence': 'DCF',
+    anvil: 'NHGRI AnVIL Data Commons'
+  }[provider]
+}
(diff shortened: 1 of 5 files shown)


src/libs/providers.js
```diff
@@ -0,0 +1,9 @@
+export const allProviders = ['fence', 'dcf-fence', 'anvil']
```
s-rubenstein: TOL: Ideally we could pull these out of the keys in `providerName`. Adding a note to address this in the renaming of our links. (Something like `_.keys(providerName...).alphabetize` Understanding that alphabetize may not be a function we get for free.)

s-rubenstein (approved):
LGTM

PR #2163, opened 2020-06-09
tidy up wrapper usage in integration tests

panentheos:
This simplifies and standardizes usage patterns around async wrappers in the integration tests. It should not change any behavior.

### DIFF
--- a/integration-tests/tests/find-workflow.js
+++ b/integration-tests/tests/find-workflow.js
@@ -1,10 +1,14 @@
+const _ = require('lodash/fp')
 const firecloud = require('../utils/firecloud-utils')
 const { withWorkspace } = require('../utils/integration-helpers')
 const { click, clickable, dismissNotifications, findElement, findText, signIntoTerra } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
-const testFindWorkflowFn = withUserToken(withWorkspace(async ({ billingProject, page, testUrl, token, workflowName, workspaceName }) => {
+const testFindWorkflowFn = _.flow(
+  withWorkspace,
+  withUserToken
+)(async ({ billingProject, page, testUrl, token, workflowName, workspaceName }) => {
   await page.goto(testUrl)
   await signIntoTerra(page, token)
   await dismissNotifications(page)
@@ -45,7 +49,7 @@ const testFindWorkflowFn = withUserToken(withWorkspace(async ({ billingProject, 
   await signIntoTerra(page, token)
   await findText(page, `${workflowName}-configured`)
   await findText(page, 'inputs')
-}))
+})
 
 const testFindWorkflow = {
   name: 'find-workflow',
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -1,35 +1,38 @@
+const _ = require('lodash/fp')
 const { withRegisteredUser, withBilling, withWorkspace } = require('../utils/integration-helpers')
 const { click, clickable, signIntoTerra, findElement, waitForNoSpinners, select, fillIn, input, findIframe, findText, dismissNotifications } = require('../utils/integration-utils')
 
 
 const notebookName = 'TestNotebook'
 
-const testRunNotebookFn = withRegisteredUser(async ({ billingProject, page, context, email, testUrl, token }) => {
-  await withBilling(withWorkspace(async ({ workspaceName }) => {
-    await page.goto(testUrl)
-    await click(page, clickable({ textContains: 'View Workspaces' }))
-    await signIntoTerra(page, token)
-    await dismissNotifications(page)
-    await findElement(page, clickable({ textContains: workspaceName }))
-    await waitForNoSpinners(page)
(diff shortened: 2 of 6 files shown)


integration-tests/tests/find-workflow.js
```diff
@@ -1,10 +1,14 @@
...
 const { withWorkspace } = require('../utils/integration-helpers')
 const { click, clickable, dismissNotifications, findElement, findText, signIntoTerra } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
-const testFindWorkflowFn = withUserToken(withWorkspace(async ({ billingProject, page, testUrl, token, workflowName, workspaceName }) => {
+const testFindWorkflowFn = _.flow(
+  withWorkspace,
+  withUserToken
+)(async ({ billingProject, page, testUrl, token, workflowName, workspaceName }) => {
```
breilly2, line 11: What would you think of using `_.flowRight`, or the `_.compose` alias in lodash/fp, to make the function order read more in-order? I think we talked about this before but I don't remember if there was a reason to avoid this.

panentheos: Normally I would use `compose` more when applying wrappers, since I think the order is clearer. But our lodash/fp eslint plugin in the main project enforces the `flowRight` alias, which I think is a terrible name. And even though the tests don't get linted (TOL: maybe they should?), I didn't want to use something here that's not allowed in the main code.

Actually looking at the plugin, it looks like we can override the preferred alias for that lint rule. So perhaps that would be a good follow-on.

breilly2: Sounds good. I agree that this is still more readable than before because it draws a clearer line between composing the functions and invoking the resulting function.

integration-tests/tests/run-notebook.js
```diff
@@ -1,35 +1,38 @@
+const _ = require('lodash/fp')
 const { withRegisteredUser, withBilling, withWorkspace } = require('../utils/integration-helpers')
 const { click, clickable, signIntoTerra, findElement, waitForNoSpinners, select, fillIn, input, findIframe, findText, dismissNotifications } = require('../utils/integration-utils')
 
 
 const notebookName = 'TestNotebook'
 
-const testRunNotebookFn = withRegisteredUser(async ({ billingProject, page, context, email, testUrl, token }) => {
-  await withBilling(withWorkspace(async ({ workspaceName }) => {
```
breilly2, line 8: Interesting. Nice observation that this could be collapsed. I wonder if it was like this because we used to try cleaning up the runtime here instead of in `withBilling`. That might also explain `billingProject`, `context`, and `email` being destructured here but ultimately going unused.

integration-tests/utils/integration-helpers.js
```diff
@@ -7,47 +7,45 @@ const { withUserToken } = require('../utils/terra-sa-utils')
 
 const defaultTimeout = 5 * 60 * 1000
 
-const makeWorkspace = async ({ billingProject, context, testUrl, token }) => {
-  const ajaxPage = await context.newPage()
-
-  await ajaxPage.goto(testUrl)
-  await signIntoTerra(ajaxPage, token)
+const withSignedInPage = fn => async options => {
```
breilly2, line 10: Nice!

integration-tests/utils/integration-helpers.js
```diff
@@ -7,47 +7,45 @@ const { withUserToken } = require('../utils/terra-sa-utils')
...
+})
 
 const withWorkspace = test => async options => {
   const workspaceName = await makeWorkspace(options)
 
   try {
     await test({ ...options, workspaceName })
   } finally {
-    await deleteWorkspace(workspaceName, options)
+    await deleteWorkspace({ ...options, workspaceName })
```
breilly2, line 48: Had to think about this for a second, but it makes sense to me now. Particularly important that it's this and not `{ workspaceName, ...options }`.

integration-tests/utils/integration-helpers.js
```diff
@@ -74,34 +72,20 @@ const withUser = test => async args => {
...
 }
 
-const addUserToBilling = withUserToken(async ({ billingProject, context, email, testUrl, token }) => {
-  const ajaxPage = await context.newPage()
-
-  await ajaxPage.goto(testUrl)
-  await signIntoTerra(ajaxPage, token)
-
-  await ajaxPage.evaluate((email, billingProject) => {
+const addUserToBilling = _.flow(withSignedInPage, withUserToken)(async ({ page, billingProject, email }) => {
```
breilly2, line 75: TOL question: Any reason why you chose to keep this `flow` on one line?

breilly2 (approved):
👍 Looks good. Nice improvement! I did not run the tests myself because CircleCI already ran them for this branch. I did check the output in CircleCI to make sure the tests all succeeded, especially making sure that `import-dockstore-workflow` actually ran instead of logging `"Skipping dockstore test, API appears to be down"`.

integration-tests/utils/integration-helpers.js
panentheos: No particular reason except that it seemed short and simple enough to be readable that way.

PR #2166, opened 2020-06-12
Fix display of fence status link [SATURN-1686]

breilly2:
We had some extra component-level state that wasn't needed anymore because the data is in `authStore`. Hooking that up fixes the display bug.

Note, this will need to be modified when merged with #2161.

Also, like #2161, I was not able to fully test locally. Linking does work on bvdp-saturn-dev.appspot.com, though, so I was able to establish a link there and verify that it is displayed on localhost.

### DIFF
--- a/src/pages/Profile.js
+++ b/src/pages/Profile.js
@@ -174,7 +174,11 @@ const FenceLink = ({ provider, displayName }) => {
   /*
    * Hooks
    */
-  const [{ username, issued_at: issuedAt }, setStatus] = useState({})
+  const auth = Utils.useStore(authStore)
+  const { username, issued_at: issuedAt } = Utils.switchCase(provider,
+    ['fence', () => auth.fenceDCPStatus || {}],
+    ['dcf-fence', () => auth.fenceDCFStatus || {}],
+    [Utils.DEFAULT, () => {}])
   const [isLinking, setIsLinking] = useState(false)
   const signal = Utils.useCancellation()
 
@@ -184,7 +188,8 @@ const FenceLink = ({ provider, displayName }) => {
     withErrorReporting('Error linking NIH account'),
     Utils.withBusyState(setIsLinking)
   )(async () => {
-    setStatus(await User.linkFenceAccount(provider, token, redirectUrl))
+    const resultStatus = await User.linkFenceAccount(provider, token, redirectUrl)
+    authStore.update(state => ({ ...state, resultStatus }))
   })
 
   Utils.useOnMount(() => {


src/pages/Profile.js
```diff
@@ -174,7 +174,11 @@ const FenceLink = ({ provider, displayName }) => {
   /*
    * Hooks
    */
-  const [{ username, issued_at: issuedAt }, setStatus] = useState({})
+  const auth = Utils.useStore(authStore)
+  const { username, issued_at: issuedAt } = Utils.switchCase(provider,
+    ['fence', () => auth.fenceDCPStatus || {}],
+    ['dcf-fence', () => auth.fenceDCFStatus || {}],
+    [Utils.DEFAULT, () => {}])
```
breilly2, line 181: This `Utils.switchCase` will need to be updated when merging with #2161. I suspect the result will also be simplified... possibly removing the `switchCase` altogether.

src/pages/Profile.js
```diff
@@ -184,7 +188,8 @@ const FenceLink = ({ provider, displayName }) => {
     withErrorReporting('Error linking NIH account'),
     Utils.withBusyState(setIsLinking)
   )(async () => {
-    setStatus(await User.linkFenceAccount(provider, token, redirectUrl))
+    const resultStatus = await User.linkFenceAccount(provider, token, redirectUrl)
+    authStore.update(state => ({ ...state, resultStatus }))
```
breilly2, line 192: This is the part that I couldn't test locally.

src/pages/Profile.js
s-rubenstein: TOL: It is interesting that we have an explicit DEFAULT option for switchCase, but not for cond (it just takes a function as its last arg). I wonder if we should bring them into alignment one way or the other? (I kinda like having the default there explicitly?)

s-rubenstein (approved):
LGTM

src/pages/Profile.js
breilly2: I've wondered about that, too.

It could just be how they came to be. I think `cond` is inspired by Clojure (?) where the default is an optional last arg while `switchCase` is inspired by the `switch` statement which explicitly declares the default.

Even so, I think I'm also in favor of standardizing, and I don't have a strong preference either way.

src/pages/Profile.js
panentheos, line 192: This create a key called `resultStatus`, which doesn't look right to me.

src/pages/Profile.js
panentheos: FWIW I've never loved the bare function as a default case for `cond`. A more consistent form would be an always-true clause like `[true, () => blah]`, which would work without any special handling. We could make that change as long as we visited every caller and updated them.

breilly2:
Fixed by #2161 instead.

PR #2168, opened 2020-06-16
[SATURN-1213] Update runtime Play button behavior

Shakespeared:
change runtime play button to open gear menu instead of starting default runtime

Verified in UI

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -12,7 +12,7 @@ import { dataSyncingDocUrl } from 'src/data/machines'
 import rLogo from 'src/images/r-logo.svg'
 import { Ajax } from 'src/libs/ajax'
 import { getDynamic, setDynamic } from 'src/libs/browser-storage'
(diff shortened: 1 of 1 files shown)


src/components/ClusterManager.js
```diff
@@ -280,7 +280,7 @@ export default class ClusterManager extends PureComponent {
         default:
           return h(ClusterIcon, {
             shape: 'play',
-            onClick: () => this.createDefaultCluster(),
+            onClick: () => this.setState({ createModalDrawerOpen: true }),
```
panentheos, line 274: This looks good to me. In addition, since this was the only usage of the `createDefaultCluster` method, we can remove that code entirely (which will also then necessitate removing a few imports -- check the linter output after you're done to make sure you got everything)

panentheos (commented):
Looks good so far, see the comment for an additional step we can take.

Also, a small suggestion: It's helpful to provide a bit of context in the PR title, so we can track and review the work more easily. It doesn't have to be wordy, even something like 'Update runtime Play button behavior' would give a good amount of context.

panentheos (approved):
Thanks for the changes, this looks good to me. I tested on my local machine.

PR #2169, opened 2020-06-16
[SATURN-1298] Update DCP&DCF naming in profile and notification

Shakespeared:
Update DCP & DCF names in user profile and notification

verified in UI

### DIFF
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -245,7 +245,7 @@ authStore.subscribe((state, oldState) => {
       )
       if (expireStatus) {
         notify('info', div([
-          `Your access to ${name} Framework Services ${expireStatus}. To `,
+          `Your access to ${name} ${expireStatus}. To `,
           expireStatus === 'has expired' ? 'restore ' : 'renew ',
           'access, log-in to Framework Services to ',
           h(FrameworkServiceLink, { linkText: 're-link', provider: key, redirectUrl }),
--- a/src/libs/providers.js
+++ b/src/libs/providers.js
@@ -1,4 +1,4 @@
 export const allProviders = [
-  { key: 'fence', name: 'DCP' },
-  { key: 'dcf-fence', name: 'DCF' }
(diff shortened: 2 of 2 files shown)


ehanna4:
Adding this from Jira ticket
@blrubenstein : Developer note: Currently our ordering is handled by duplicated strings from the object keys into an array. Ideally we would have this done via a _.keys method, and then following some ordering rule. We should take this opportunity to figure out what our ordering rule is. (I have a preference for alphabetical, but design probably has opinions)


Emily: ​thanks for pointing that out. I think the scope is probably bigger than this ticket for “Ordering Notifications” since these notifications are not the only ones that can appear at sign in time and so probably warrants a bigger discussion. What do you think

@blrubenstein: Possibly. This came out when @panentheos and I were talking on his PR about it, and he thought this might be a good place to tackle that duplication. If we do think there is a follow up/we don’t want to do it here, I think that is a piece of tech debt we should do soon or we will lose the context very quickly.

TOL: What does the team think about ordering of notifications - @Shakespeared and I had noted that DCP, DCF, NIH are not the only notifications that pop up at log in (service alerts etc. do as well) so it made this notification ordering techdebt seem larger than the scope of this ticket. But Scout/Brett bring up a good point that it is tech debt that should be addressed soon. What does the team think about including in scope with this ticket or creating a new ticket to cover that change?

Shakespeared:
Whether we leave it in this ticket or make a new one, I'm happy to keep working on this with @ehanna4 (or whomever) as a smallish ticket so I can have something to get practice on!

panentheos (approved):
Looks good. Tested the notifications and the profile locally.

Also, regarding the earlier comment from @blrubenstein , the data was restructured as a 'list of records', so the concern about duplication should be addressed already.

panentheos:
Regarding ordering, the profile page maintains the previous order, which I think is fine for now (we can always change if someone requests it). The notifications happen in whatever order the requests complete, since they're checked in parallel.

PR #2170, opened 2020-06-17
Mixpanel - track notebook renames on clone and edit [SATURN-1633]

cameron-ardell:
#### Acceptance Criteria

- [x] Mixpanel event triggered when a notebook is cloned called “notebook:clone”
- [x] Mixpanel event triggered when a notebook is renamed “notebook:rename”
- [x] Properties on each event for the name of the parent “oldName” and the name of the new notebook “newName”
- [x] Description of event added to MixPanel lexicon

### DIFF
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -8,6 +8,7 @@ import { ValidatedInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
 import { Ajax, ajaxCaller } from 'src/libs/ajax'
 import { reportError } from 'src/libs/error'
+import Events from 'src/libs/events'
 import { FormLabel } from 'src/libs/forms'
 import * as Nav from 'src/libs/nav'
 import * as Utils from 'src/libs/utils'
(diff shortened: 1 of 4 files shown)


ehanna4 (commented):
Checked out the data in Mixpanel and the event appears there (though I haven't yet tested trying to send an event myself yet). Only nit is the properties oldName and newName could probably use a description or display name that makes it clear it's the notebook's old&new names (since properties can be shared by events)

src/components/notebook-utils.js
```diff
@@ -215,20 +218,23 @@ export const NotebookDuplicator = ajaxCaller(class NotebookDuplicator extends Co
...
-              Nav.goToPath('workspace-notebook-launch', {
-                namespace, name: wsName, notebookName: `${newName}.ipynb`
-              })
-            }
+              Ajax().Buckets.notebook(namespace, bucketName, printName).copy(newName, bucketName, !destroyOld)
+            )
           } catch (error) {
             reportError(`Error ${destroyOld ? 'renaming' : 'copying'} notebook`, error)
           }
+          onSuccess()
```
panentheos: Moving this outside the `try` block means it will get called on errors too, which isn't the behavior we want. We should leave the general structure the same.

src/components/notebook-utils.js
```diff
@@ -169,6 +170,7 @@ export const NotebookCreator = class NotebookCreator extends Component {
   }
 }
 
+
```
panentheos: nit: These lines don't seem like they're improving readability, and they add noise to the diff. I find it helpful to glance over the whole diff right before creating a PR, and doing cleanup as necessary.

panentheos (commented):
See comments. Also, there's one more call to `copy` in the `ExportNotebookModal` that will need metrics. That case deals with copying from one workspace to another, so I'd imagine we probably want to record source and destination as part of the clone event.

src/components/notebook-utils.js
cameron-ardell: I shouldn't put the Mixpanel call in the block though right?

cameron-ardell: Never mind, I feel like I should. I was thinking about it in terms of the await.  But we don't want it reporting stuff to Mixpanel if there's an error.

cameron-ardell:
I just tested everything and I believe it's ready for a re-review

src/components/notebook-utils.js
```diff
@@ -215,17 +216,20 @@ export const NotebookDuplicator = ajaxCaller(class NotebookDuplicator extends Co
...
+              Ajax().Buckets.notebook(namespace, bucketName, printName).copy(newName, bucketName, !destroyOld)
+            )
             onSuccess()
             if (fromLauncher) {
               Nav.goToPath('workspace-notebook-launch', {
                 namespace, name: wsName, notebookName: `${newName}.ipynb`
               })
             }
+            const eventType = destroyOld ? Events.notebookRename : Events.notebookClone
+            Ajax().Metrics.captureEvent(eventType, { oldName: printName, newName })
```
panentheos: In order for these metrics to be valuable, we need to store workspace information, otherwise we won't be able to tell where these notebooks live. For rename, it's always within the same workspace, so there will just be one, but for copy, it can be between workspaces, so we'll want to store both source and destination.

panentheos (approved):
LGTMob, tested locally.

PR #2171, opened 2020-06-17
Log to firestore [SATURN-1370][SATURN-1691]

s-rubenstein:
Tested both the error case and the success case. Tested to make sure it was queryable using the following script:
```
const { Firestore } = require('@google-cloud/firestore')

const firestore = new Firestore()


const main = async () => {
  const { docs } = await firestore.collection('tests').where('testGroup', '=', '6/17/2020:10:51').get()
  console.log(docs)
}

main()
```
Which is runable with the following bash command:
`GCP_PROJECT=terra-bueller GOOGLE_APPLICATION_CREDENTIALS=/Users/brubenst/Desktop/code/keys/bueller-default-sa-key.json TERRA_SA_KEY=$(vault read --format=json secret/dsde/alpha/common/firecloud-account.pem | jq .data) LYLE_SA_KEY=$(vault read --format=json secret/dsde/terra/envs/common/lyle-user-service-account-key | jq .data) node scripts/getResultsByTestGroup.js`

I also tested the script above with `status = success` to ensure it didn't always pull in all the data. 

You can also see the many logs I made to firestore here: https://console.cloud.google.com/firestore/data/tests?project=terra-bueller

I went with UUIDs for the document names in firestore under the assumption that the use case involves querying by testGroup, rather than looking at individual runs.

### DIFF
--- a/integration-tests/src/server.js
+++ b/integration-tests/src/server.js
@@ -35,9 +35,12 @@ const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, 
   const testGroup = req.query.testGroup
   const startTime = new Date()
   const host = req.headers.host
+  if (testGroup === undefined) {
+    throw new Error('Must provide a test group to record test results')
+  }
   const reportTest = async error => {
     const endTime = new Date()
-    const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: error ? 'error' : 'success', testName: name, error, host, startTime }
+    const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: error ? 'error' : 'success', testName: name, error, host, startTime, endTime }
     await firestore.collection('tests').doc().create(testReport)
   }
   if (host === 'terra-bueller.appspot.com') {
--- a/integration-tests/src/server.js
+++ b/integration-tests/src/server.js
@@ -34,11 +34,10 @@ const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, 
   const targetEnvParams = { ...envs[req.query.targetEnv] }
   const testGroup = req.query.testGroup
   const startTime = new Date()
-  const testName = req.route.path.split('/')[2]
   const host = req.headers.host
   const reportTest = async error => {
     const endTime = new Date()
-    const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: error ? 'error' : 'success', testName, error, host }
+    const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: error ? 'error' : 'success', testName: name, error, host, startTime }
     await firestore.collection('tests').doc().create(testReport)
   }
   if (host === 'terra-bueller.appspot.com') {
--- a/integration-tests/src/server.js
+++ b/integration-tests/src/server.js
@@ -28,18 +28,19 @@ const testTimeout = async timeout => {
   throw new Error(`Test timeout after ${timeout}ms`)
 }
 
-const reportTest = (testGroup, startTime, host, testName, errorMessage) => {
-  const endTime = new Date()
-  const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: errorMessage ? 'error' : 'success', testName, errorMessage, host }
-  firestore.collection('tests').doc().create(testReport)
-}
+
 
 const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, timeout }) => {
   const targetEnvParams = { ...envs[req.query.targetEnv] }
   const testGroup = req.query.testGroup
   const startTime = new Date()
-  const testName = req.route.path
+  const testName = req.route.path.split('/')[2]
   const host = req.headers.host
+  const reportTest = async error => {
+    const endTime = new Date()
+    const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: error ? 'error' : 'success', testName, error, host }
+    await firestore.collection('tests').doc().create(testReport)
+  }
   if (host === 'terra-bueller.appspot.com') {
     throw new Error('Do not call without a version specified')
   }
@@ -49,7 +50,7 @@ const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, 
       test({ browser, context, page, ...targetEnvParams }),
       testTimeout(timeout)
     ])
-    reportTest(testGroup, startTime, host, testName, null)
+    await reportTest(null)
     return new Response(200, result)
   } catch (e) {
     const error = {
@@ -57,7 +58,7 @@ const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, 
       message: e.message,
       stack: e.stack
     }
-    reportTest(testGroup, startTime, host, testName, error)
+    await reportTest(error)
     return new Response(200, error)
   }
 })
--- a/integration-tests/src/server.js
+++ b/integration-tests/src/server.js
@@ -28,9 +28,9 @@ const testTimeout = async timeout => {
   throw new Error(`Test timeout after ${timeout}ms`)
 }
 
-const reportTest = (testGroup, startTime, host, errorMessage) => {
+const reportTest = (testGroup, startTime, host, testName, errorMessage) => {
   const endTime = new Date()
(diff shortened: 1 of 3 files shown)


integration-tests/src/server.js
```diff
@@ -20,26 +21,44 @@ const app = express()
...
 server.setTimeout(serverTimeout)
 
+const firestore = new Firestore()
+
 const testTimeout = async timeout => {
   await delay(timeout)
   throw new Error(`Test timeout after ${timeout}ms`)
 }
 
+const reportTest = (testGroup, startTime, host, testName, errorMessage) => {
```
panentheos: TOL: Rather than passing all these parameters, it might be cleaner to define this helper inside `runTest`, and just reference everything except `errorMessage` from the lexical scope.

integration-tests/src/server.js
```diff
@@ -20,26 +21,44 @@ const app = express()
...
+  const endTime = new Date()
+  const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: errorMessage ? 'error' : 'success', testName, errorMessage, host }
+  firestore.collection('tests').doc().create(testReport)
```
panentheos: This call is async, so we should make sure it gets properly `await`-ed all the way up to the original caller.

integration-tests/src/server.js
```diff
@@ -20,26 +21,44 @@ const app = express()
...
+}
+
 const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, timeout }) => {
   const targetEnvParams = { ...envs[req.query.targetEnv] }
+  const testGroup = req.query.testGroup
+  const startTime = new Date()
+  const testName = req.route.path
```
panentheos: This will give the full path, e.g. `'/test/foo'`, right? I think the fact that we're serving requests at `/test` is a current implementation detail, and not something that we should persist to the DB. What about just using `name`?

integration-tests/src/server.js
```diff
@@ -20,26 +21,44 @@ const app = express()
...
+const firestore = new Firestore()
+
 const testTimeout = async timeout => {
   await delay(timeout)
   throw new Error(`Test timeout after ${timeout}ms`)
 }
 
+const reportTest = (testGroup, startTime, host, testName, errorMessage) => {
+  const endTime = new Date()
+  const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: errorMessage ? 'error' : 'success', testName, errorMessage, host }
```
panentheos: Perhaps `errorMessage` would be better named `error`, since it contains `message` as well as other fields.

integration-tests/src/server.js
s-rubenstein: I was wondering about this one, because I could see a world where we didn't want to await on firestore (just fire and forget)

integration-tests/src/server.js
s-rubenstein: Yes, this will give the full path. I will change.

integration-tests/src/server.js
panentheos: In this case, the expectation is that the DB write will succeed, and it should be very fast, so waiting gives us a potential way to catch errors sooner in case they happen.

integration-tests/src/server.js
```diff
@@ -20,26 +21,45 @@ const app = express()
...
   throw new Error(`Test timeout after ${timeout}ms`)
 }
 
+
+
 const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, timeout }) => {
   const targetEnvParams = { ...envs[req.query.targetEnv] }
+  const testGroup = req.query.testGroup
+  const startTime = new Date()
+  const testName = req.route.path.split('/')[2]
```
panentheos: The `name` parameter from above should already have this value without having to compute it.

panentheos (commented):
See comment, then looks good to me. I was involved in pairing on this though, could someone else give the formal approval?

integration-tests/src/server.js
```diff
@@ -20,26 +21,44 @@ const app = express()
...
+  const testGroup = req.query.testGroup
+  const startTime = new Date()
+  const host = req.headers.host
+  const reportTest = async error => {
+    const endTime = new Date()
+    const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: error ? 'error' : 'success', testName: name, error, host, startTime }
+    await firestore.collection('tests').doc().create(testReport)
+  }
+  if (host === 'terra-bueller.appspot.com') {
+    throw new Error('Do not call without a version specified')
```
breilly2, line 47: E.g., `https://20200311t230511-dot-terra-bueller.appspot.com/`?
TOL: I guess it makes sense to make sure you know what version of the tests you're running, and that it's a stable version if you're doing a long-running performance test. We should probably consider enforcing (or at least allowing) specification of a stable target build as well.

integration-tests/src/server.js
```diff
@@ -20,26 +21,44 @@ const app = express()
...
+
+
 const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, timeout }) => {
   const targetEnvParams = { ...envs[req.query.targetEnv] }
+  const testGroup = req.query.testGroup
+  const startTime = new Date()
+  const host = req.headers.host
+  const reportTest = async error => {
+    const endTime = new Date()
+    const testReport = { testGroup, runtimeInMilliseconds: endTime - startTime, status: error ? 'error' : 'success', testName: name, error, host, startTime }
```
breilly2: Should we record `endTime` too? I know the 3rd number can be calculated from the other two but it might be useful to record the raw data.

integration-tests/src/server.js
```diff
@@ -20,26 +21,44 @@ const app = express()
...
+    await firestore.collection('tests').doc().create(testReport)
+  }
+  if (host === 'terra-bueller.appspot.com') {
+    throw new Error('Do not call without a version specified')
+  }
   try {
     const test = withScreenshot(name, fn)
     const result = await Promise.race([
       test({ browser, context, page, ...targetEnvParams }),
       testTimeout(timeout)
```
breilly2, line 53: TOL: We may want to consider recording test timeout events in a distinct way to make it easier to report on how often this happens. Could be particularly interesting for load tests.

integration-tests/src/server.js
```diff
@@ -20,26 +21,44 @@ const app = express()
...
 const testTimeout = async timeout => {
   await delay(timeout)
   throw new Error(`Test timeout after ${timeout}ms`)
 }
 
+
+
 const runTest = fn => withPuppeteer(async ({ browser, context, name, page, req, timeout }) => {
   const targetEnvParams = { ...envs[req.query.targetEnv] }
+  const testGroup = req.query.testGroup
```
breilly2, line 35: In testing, I got this error when I forgot to specify `testGroup`, and only after the test run completed:

> Error: Value for argument "data" is not a valid Firestore document. Cannot use "undefined" as a Firestore value (found in field "testGroup"). If you want to ignore undefined values, enable `ignoreUndefinedProperties`.

I think we should validate that we have this before running the test or provide a default value if it's missing, like `Date.now()`.

breilly2 (approved):
👍 Looks good. A few thoughts; see comments. Nothing absolutely critical for this type of utility. I tested various scenarios with some `--no-promote` GAE deploys, including test timeout and checking that a version was specified (which was a hack test and could use a quick check once merged and deployed).

integration-tests/src/server.js
s-rubenstein: I don't want to do that, because it could change mid-script. (If someone is deploying to the terra-bueller with --promote while a perf-test run script is in progress)

integration-tests/src/server.js
s-rubenstein: Sure! I can add endTime. TOL about use cases, I wanted to have a startTime so we could compare when runs started with other runs. I don't know how much value endTime will add, but its cheap and easy enough to add.

integration-tests/src/server.js
s-rubenstein: I also thought about this, but we won't catch timeouts within the test, only global timeouts on the entire run (on talking to Brett), so the inconsistency may hurt more than help.

integration-tests/src/server.js
breilly2: To clarify, I was talking about being able to lock down the Terra UI build. Right now, we can only specify an env such as `dev` but we might want to allow pointing at a specific GAE version such as `https://20200619t224522-dot-bvdp-saturn-dev.appspot.com/`.

PR #2172, opened 2020-06-17
[SATURN-1682] change datetime display format in job history

Shakespeared:
Change the display format of the submitted date in the "job history" page so it always displays MMM dd yyyy HH:mm instead of "today" 

checked in browser


cameron-ardell (approved):
Looks good to me!

PR #2173, opened 2020-06-22
Add xtra 50G disk for GCE runtimes [SATURN-1702]

ehanna4:
Leo has updated to use 2 dsisks for users, one will be the disk the user requests and the other will be for storing the installation. This is true for GCE runtimes.

See: https://broadworkbench.atlassian.net/browse/IA-1759 for details

tested creating a runtime in the UI and checking the cost

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -507,7 +507,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
         div({
           style: { backgroundColor: colors.dark(0.2), borderRadius: 100, width: 'fit-content', padding: '0.75rem 1.25rem', ...styles.row }
         }, [
-          span({ style: { ...styles.label, marginRight: '0.25rem' } }, ['COST:']),
+          span({ style: { ...styles.label, marginRight: '0.25rem', textTransform: 'uppercase' } }, ['cost:']),
           `${Utils.formatUSD(runtimeConfigCost(this.getRuntimeConfig(!currentCluster)))} per hour`
         ])
       ])
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -119,10 +119,10 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     }
   }
 
-  getRuntimeConfig() {
+  getRuntimeConfig(isNew = false) {
     return formatRuntimeConfig({
       cloudService: !!this.state.sparkMode ? 'DATAPROC' : 'GCE',
-      bootDiskSize: !!this.state.sparkMode ? null : 50,
+      isNew,
       ..._.pick(
         ['numberOfWorkers', 'masterMachineType', 'masterDiskSize', 'workerMachineType', 'workerDiskSize', 'numberOfPreemptibleWorkers'],
         this.state)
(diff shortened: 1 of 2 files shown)


s-rubenstein (dismissed):
This looks reasonable to me.

Qi77Qi:
not necessarily in this PR, but now that we're returning `bootDiskSize`, it'll be better to use this value instead of fixed `50G` in case we optimize boot disk size in the future...

breilly2:
Also need to adjust `runtimeConfigCost` just below these changes.

ehanna4:
@breilly2 `runtimeConfigCost ` calls `ongoingCost` which takes into account the bootdisk extra cost so it looked to me no change needed to be made to `runtimeconfigcost` ?

breilly2 (dismissed):
👍 Code looks good. I tested both the cost calculated from an actual runtime config from leo (in the upper-right runtime widget) and the cost displayed in the notebook runtime drawer (which is based on the config parameters we would send to leo for creating/updating a runtime).

src/components/NewClusterModal.js
```diff
@@ -122,6 +122,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
   getRuntimeConfig() {
     return formatRuntimeConfig({
       cloudService: !!this.state.sparkMode ? 'DATAPROC' : 'GCE',
+      bootDiskSize: !!this.state.sparkMode ? null : 50,
```
panentheos: TOL: This assumption will work for newly created runtimes after the Leo change, but may slightly over-estimate costs for updates to old runtimes. That may be fine, especially considering it's such a small amount and only affects this one particular case, (plus slightly over-estimating is better than under-estimating).

breilly2: Good point. To account for that, we'd need to:
* know that there is an existing runtime and whether or not it has a boot disk
* know whether or not the change being made is going to result in deleting and recreating the runtime
This is a lot of context that this function currently doesn't have.

Also, darn, I just noticed that this is used to generate the payload to leo where it's technically incorrect and possibly misleading to be providing a `bootDiskSize` property. Unless leo actually does accept this property... but I'm not sure we want to go down that road anyway.

In fact, this has implication on `NewClusterModal.hasChanges()`. This may benefit from just a little more thought.

breilly2: I thought about this a little more. `NewClusterModal.hasChanges()` uses `_.isEqual()` on 2 configs that are first run through `normalizeRuntimeConfig()`. Maybe `normalizeRuntimeConfig()` is the right place for this so that we have better control over what constitutes a difference between a config we build to use as a `POST` payload and a config we get back from leo. `runtimeConfigCost()` already calls `normalizeRuntimeConfig()` so that should work. I haven't yet studied other callers of `normalizeRuntimeConfig()` to look for unintended consequences.

ehanna4:
Updated logic and tested so only new clusters have the extra 50G of memory cost - see screenshots for example of creating a new one is $0.21 but for updating it stays $0.20.

Screen Shot 2020-06-27 at 4 27 06 PM
Screen Shot 2020-06-27 at 4 27 29 PM

src/components/NewClusterModal.js
```diff
@@ -507,7 +508,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
           style: { backgroundColor: colors.dark(0.2), borderRadius: 100, width: 'fit-content', padding: '0.75rem 1.25rem', ...styles.row }
         }, [
           span({ style: { ...styles.label, marginRight: '0.25rem' } }, ['COST:']),
```
s-rubenstein: Non-blocking question for the future: Why do we use 'COST' (all caps) here rather than 'Cost' with `textTransform: 'uppercase'`

s-rubenstein (approved):
This seems reasonable, and code looks reasonable. Tested locally by creating a cluster and verified new response had a `bootDiskSize` variable. Did not have a way to test old clusters.

src/components/NewClusterModal.js
ehanna4: Great question, I can't think of a reason and I don't see 'COST' anywhere else in the code so it's probably some tech debt? I will change it

PR #2174, opened 2020-06-22
Fix workspace namespace references for IGV [SATURN-1698]

breilly2:
This fixes an IGV bug with fetching a Google OAuth token and determining buckets that require requester-pays access. Testing is tricky because, as it was before, it was asking to use a Google project called `undefined` which _actually exists in non-prod environments_! If you look at the network tab, you'll see weird calls to sam with a path parameter of `undefined`, but because the project exists, these calls work. These same calls fail in prod. (I could not find a billing project called `undefined` in any of our usual non-prod environments so I have no idea who actually has permission to use it... but it seems like most/all of our usual dev accounts can use it.) Therefore, you have to look at the network tab before/after to see the difference to validate that the change is correct.

### DIFF
--- a/src/components/IGVBrowser.js
+++ b/src/components/IGVBrowser.js
@@ -33,7 +33,7 @@ const IGVBrowser = _.flow(
           return knownBucketStatus
         } else {
           try {
-            await Ajax(signal).Buckets.getObject(bucket, file, workspace.namespace, { fields: 'kind' })
+            await Ajax(signal).Buckets.getObject(bucket, file, workspace.workspace.namespace, { fields: 'kind' })
             return false
           } catch (e) {
             if (e.requesterPaysError) {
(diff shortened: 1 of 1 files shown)


s-rubenstein (approved):
This looks good to me. and seems reasonable. Tested locally but couldn't find the specific call, probably because I wasn't in a requester pays bucket?

breilly2:
@blrubenstein I looked into it further and the problem in prod does not involve a requester-pays bucket. The call to sam that we were seeing before is somehow avoided altogether now that we're getting the namespace property correctly. It bothers me that I can't definitely trace the code path that was being taken that caused the call to sam, but I'm confident that the code was wrong before.

It it helps increase confidence at all, I did do a sneaky test where I made my local UI use some of the prod backend so I could test this change in prod conditions and everything worked.

breilly2:
Note to self: Because of the differences in witnessed behavior between dev and prod, this definitely deserves a check in prod after release to make sure it's fixed.

PR #2175, opened 2020-06-23
use new Shibboleth service for NIH link [SATURN-1335]

panentheos:
See ticket for details. Tested locally using the fake login flow, and also tested a real login with some config overrides.

### DIFF
--- a/config/alpha.json
+++ b/config/alpha.json
@@ -13,10 +13,11 @@
   "leoUrlRoot": "https://leonardo.dsde-alpha.broadinstitute.org",
   "marthaUrlRoot": "https://us-central1-broad-dsde-alpha.cloudfunctions.net",
   "orchestrationUrlRoot": "https://firecloud-orchestration.dsde-alpha.broadinstitute.org",
+  "profileUrlRoot": "https://profile-dot-broad-shibboleth-prod.appspot.com/dev",
   "rawlsUrlRoot": "https://rawls.dsde-alpha.broadinstitute.org",
   "rexUrlRoot": "https://terra-rex-alpha.appspot.com",
   "samUrlRoot": "https://sam.dsde-alpha.broadinstitute.org",
-  "shibbolethUrlRoot": "http://mock-nih.dev.test.firecloud.org",
+  "shibbolethUrlRoot": "https://broad-shibboleth-prod.appspot.com/dev",
   "tCell": {
     "appId": "saturnnonprod-O9lUP",
     "apiKey": "AQEBBAEkx4iE2KxNyI7Wx08EwU1ycTM7E4FMSmaibbMUQxNU6uQvuAJt7fyABAtFYSYfgEE"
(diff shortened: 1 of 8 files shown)


config/alpha.json
```diff
@@ -13,10 +13,11 @@
   "leoUrlRoot": "https://leonardo.dsde-alpha.broadinstitute.org",
   "marthaUrlRoot": "https://us-central1-broad-dsde-alpha.cloudfunctions.net",
   "orchestrationUrlRoot": "https://firecloud-orchestration.dsde-alpha.broadinstitute.org",
+  "profileUrlRoot": "https://profile-dot-broad-shibboleth-prod.appspot.com/dev",
   "rawlsUrlRoot": "https://rawls.dsde-alpha.broadinstitute.org",
   "rexUrlRoot": "https://terra-rex-alpha.appspot.com",
   "samUrlRoot": "https://sam.dsde-alpha.broadinstitute.org",
-  "shibbolethUrlRoot": "http://mock-nih.dev.test.firecloud.org",
+  "shibbolethUrlRoot": "https://broad-shibboleth-prod.appspot.com/dev",
```
s-rubenstein, line 20: Unrelated to PR: The URL here seems super weird because of the prod in the URL host. If we take over shibboleth, I might push to see if we can change that.

config/alpha.json
```diff
@@ -13,10 +13,11 @@
   "leoUrlRoot": "https://leonardo.dsde-alpha.broadinstitute.org",
   "marthaUrlRoot": "https://us-central1-broad-dsde-alpha.cloudfunctions.net",
   "orchestrationUrlRoot": "https://firecloud-orchestration.dsde-alpha.broadinstitute.org",
+  "profileUrlRoot": "https://profile-dot-broad-shibboleth-prod.appspot.com/dev",
```
s-rubenstein, line 16: Nit: I might call this something like 'eraProfileUrlRoot'? TOL profileUrlRoot makes me think of SAM or a different service storing full profiles, and this is just hosting the eRA commons info.

s-rubenstein (approved):
A few nits, but not blocking. The only thing I'd probably consider near blocking is the `profileUrlRoot` comment, because I think that could cause problems with understanding down the line.

config/alpha.json
panentheos: Agree, the current approach is to have both dev and prod co-located, but I think it would be better to split them up.

config/alpha.json
panentheos: I thought about this as well. Although the new service currently only handles eRA Commons tokens, it's built and named ('terra-profile') in a way that suggests it could evolve into a general purpose profile service. It's unclear whether that will come to pass, but for now the name seems relatively unambiguous to me (Sam is limited to IAM, rather than profiles. Orch currently handles most of the profile stuff, but also does a lot of other things)

PR #2176, opened 2020-06-23
add anvil linkage [SATURN-1553]

panentheos:
This adds the the Anvil linkage to the UI. Because the auth process enforces that the redirect URL matches its whitelist, I had to temporarily hack the domain to be 'bvdp-saturn-dev.appspot.com' while testing locally, but was able to verify the flow.

Also fixes a react `key` warning.


breilly2 (approved):
👍 Looks good. I also tested by temporarily hacking the `return_uri` (good suggestion Brett).

breilly2:
One other thought: since this will use different deployments of other services (Bond, Gen3) than with our testing in dev, we should do a final test in prod after this change is released.

PR #2177, opened 2020-06-24
[SATURN-1701] fix runtimes sorting bug

Shakespeared:
in the 'your notebook runtimes' page, fix the ability to sort a user's existing runtimes by a) date created or b) date last accessed

checked in the UI


panentheos (approved):
Looks good. Tested locally.

PR #2178, opened 2020-06-24
Revert "use new Shibboleth service for NIH link (#2175)"

panentheos:
Reverting change pending changes to the whitelist.


breilly2 (approved):
👍 I tested in dev, which worked, but is of course a very different environment than prod. Will retest in prod when this is released.

PR #2179, opened 2020-06-24
reenable new Shibboleth service [SATURN-1335]

panentheos:
See ticket. This should now work properly in prod with the updated whitelist.


breilly2 (approved):
👍 Looks good. I tested several things:
1. Running locally, swapping in a bogus return URL results in a "Un-Trusted Return URL" result from shibboleth. Swapping in app.terra.bio successfully redirects to prod Terra.
2. Interrupting the redirect to app.terra.bio and sending the request back to localhost results in a successful link (assuming `/dev` is remove from both of the relevant config entries)

PR #2181, opened 2020-06-30
[SATURN-1654] branded experience for baseline

Shakespeared:
added a new color scheme and logo for Baseline users, with logic to select those according to subdomain once that's in place. Let me know if I missed anything with the branding. 

tested in browser using config overrides since the new subdomain isn't active yet.


src/libs/config.js
```diff
@@ -13,4 +13,6 @@ export const isDatastage = () => (window.location.hostname === 'datastage.terra.
 export const isAnvil = () => (window.location.hostname === 'anvil.terra.bio') || getConfig().isAnvil
 export const isBioDataCatalyst = () => (window.location.hostname.endsWith('.biodatacatalyst.nhlbi.nih.gov')) || getConfig().isBioDataCatalyst
 export const isUKBiobank = () => (window.location.hostname === 'ukbiobank.terra.bio') || getConfig().isUKBiobank
-export const isTerra = () => !isFirecloud() && !isDatastage() && !isAnvil() && !isBioDataCatalyst() && !isUKBiobank()
+export const isBaseline = () => (window.location.hostname === 'baseline.terra.bio') || getConfig().isBaseline
```
ehanna4, line 12: I̶ ̶t̶h̶i̶n̶k̶ ̶t̶h̶i̶s̶ ̶l̶i̶s̶t̶ ̶i̶s̶ ̶g̶e̶n̶e̶r̶a̶l̶l̶y̶ ̶a̶l̶p̶h̶a̶b̶e̶t̶i̶c̶a̶l̶ ̶s̶o̶ ̶w̶o̶u̶l̶d̶ ̶r̶e̶c̶o̶m̶m̶e̶n̶d̶ ̶m̶o̶v̶i̶n̶g̶ ̶`̶i̶s̶B̶a̶s̶e̶l̶i̶n̶e̶`̶ ̶a̶f̶t̶e̶r̶ ̶`̶i̶s̶A̶n̶v̶i̶l̶`̶

*Edit nevermind I only could see some of this list and that does not appear to be an accurate statement once the rest is expanded

src/images/brands/baseline/baseline-logo-color.svg
```diff
@@ -0,0 +1,127 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 24.2.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
panentheos: If you run `yarn run optimize-image-svgs`, it will strip out unnecessary stuff and make these files much smaller.

src/libs/colors.js
```diff
@@ -23,6 +23,7 @@ const colorPalette = Utils.cond(
   [isAnvil(), { ...baseColors, primary: '#e0dd10', accent: '#035c94', light: '#f6f7f4', dark: '#012840' }],
   [isBioDataCatalyst(), { ...baseColors, primary: '#c02f42', secondary: '#1a568c', accent: '#1a568c', light: '#f4f4f6', dark: '#12385a' }],
   [isUKBiobank(), { ...baseColors, primary: '#005f6f' }],
+  [isBaseline(), { ...baseColors, primary: '#c41061', secondary: '#31164c', accent: '#4d72aa', success: '#74ae43', warning: '#f7981c', danger: '#db3214', light: '#f6f7f4', dark: '#012840' }],
```
panentheos: For the colors that are the same as the `baseColors` above, we can just omit them here.

src/libs/config.js
Shakespeared: I can alphabetize it anyhow if that makes sense!

src/images/brands/baseline/baseline-logo-color.svg
Shakespeared: Thanks!

panentheos (approved):
See lint error, then looks good. Tested locally with a config override.

PR #2182, opened 2020-07-01
respect patchInProgress flag for runtime status [SATURN-1522]

panentheos:
See ticket for details.

Adds a `collapsedClusterStatus` function to transform the `patchInProgress` flag into a new pseudo-status for the purposes of driving display and behavior. Note that while we display this new state as 'Updating', it behaves differently from the preexisting `'Updating'` status (which we see when adding workers to a dataproc cluster).

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -12,7 +12,7 @@ import { dataSyncingDocUrl } from 'src/data/machines'
 import rLogo from 'src/images/r-logo.svg'
 import { Ajax } from 'src/libs/ajax'
 import { getDynamic, setDynamic } from 'src/libs/browser-storage'
-import { clusterCost, currentCluster, deleteText, trimClustersOldestFirst } from 'src/libs/cluster-utils'
+import { clusterCost, collapsedClusterStatus, currentCluster, deleteText, trimClustersOldestFirst } from 'src/libs/cluster-utils'
 import colors from 'src/libs/colors'
 import { reportError, withErrorReporting } from 'src/libs/error'
 import * as Nav from 'src/libs/nav'
@@ -194,12 +194,11 @@ export default class ClusterManager extends PureComponent {
     return currentCluster(clusters)
   }
 
-  async executeAndRefresh(promise, waitBeforeRefreshMillis = 0) {
+  async executeAndRefresh(promise) {
     try {
       const { refreshClusters } = this.props
       this.setState({ busy: true })
       await promise
-      waitBeforeRefreshMillis && await Utils.delay(waitBeforeRefreshMillis)
       await refreshClusters()
     } catch (error) {
       reportError('Notebook Runtime Error', error)
@@ -229,7 +228,7 @@ export default class ClusterManager extends PureComponent {
       return null
     }
     const currentCluster = this.getCurrentCluster()
-    const currentStatus = currentCluster?.status
+    const currentStatus = collapsedClusterStatus(currentCluster)
 
     const renderIcon = () => {
       switch (currentStatus) {
@@ -253,6 +252,7 @@ export default class ClusterManager extends PureComponent {
         case 'Stopping':
         case 'Updating':
         case 'Creating':
+        case 'Reconfiguring':
           return h(ClusterIcon, {
             shape: 'sync',
             disabled: true,
@@ -280,8 +280,8 @@ export default class ClusterManager extends PureComponent {
     }
     const totalCost = _.sum(_.map(clusterCost, clusters))
     const activeClusters = this.getActiveClustersOldestFirst()
-    const { Creating: creating, Updating: updating } = _.countBy('status', activeClusters)
-    const isDisabled = !canCompute || creating || busy || updating
+    const { Creating: creating, Updating: updating, Reconfiguring: reconfiguring } = _.countBy(collapsedClusterStatus, activeClusters)
+    const isDisabled = !canCompute || creating || busy || updating || reconfiguring
 
     const isRStudioImage = currentCluster?.labels.tool === 'RStudio'
     const appName = isRStudioImage ? 'RStudio' : 'terminal'
@@ -318,7 +318,7 @@ export default class ClusterManager extends PureComponent {
           div({ style: { marginLeft: '0.5rem', paddingRight: '0.5rem', color: colors.dark() } }, [
             div({ style: { fontSize: 12, fontWeight: 'bold' } }, 'Notebook Runtime'),
             div({ style: { fontSize: 10 } }, [
-              span({ style: { textTransform: 'uppercase', fontWeight: 500 } }, currentStatus || 'None'),
+              span({ style: { textTransform: 'uppercase', fontWeight: 500 } }, [currentStatus === 'Reconfiguring' ? 'Updating' : (currentStatus || 'None')]),
               currentStatus && ` (${Utils.formatUSD(totalCost)} hr)`
             ])
           ]),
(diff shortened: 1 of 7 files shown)


src/components/ClusterManager.js
```diff
@@ -318,7 +318,7 @@ export default class ClusterManager extends PureComponent {
           div({ style: { marginLeft: '0.5rem', paddingRight: '0.5rem', color: colors.dark() } }, [
             div({ style: { fontSize: 12, fontWeight: 'bold' } }, 'Notebook Runtime'),
             div({ style: { fontSize: 10 } }, [
-              span({ style: { textTransform: 'uppercase', fontWeight: 500 } }, currentStatus || 'None'),
+              span({ style: { textTransform: 'uppercase', fontWeight: 500 } }, [currentStatus === 'Reconfiguring' ? 'Updating' : (currentStatus || 'None')]),
```
s-rubenstein: This feels a little weird to be transforming the text in only one case, but I do agree that `Updating` is more clear than `Reconfiguring`. Is `Reconfiguring` how the API returns it?

src/libs/cluster-utils.js
```diff
@@ -78,6 +78,9 @@ export const trimClustersOldestFirst = _.flow(
 
 export const currentCluster = _.flow(trimClustersOldestFirst, _.last)
 
+export const collapsedClusterStatus = cluster => {
+  return cluster && (cluster.patchInProgress ? 'Reconfiguring' : cluster.status) // NOTE: preserves null vs undefined
```
s-rubenstein: Since we are setting this here, could we make this 'Updating' rather than 'Reconfiguring' so we don't have to do the text transforms?

panentheos: 'Updating' is already a real status that we get back when adjusting the number of workers in a spark cluster. We wanted this new state to display as 'Updating', but behave slightly differently (we need to kick you out of your notebook because of the machine stop, whereas the real 'Updating' status takes place while the machine is running)

breilly2: TOL: What if `Reconfiguring` someday becomes a real state returned from GCP/leo? I know it's probably unlikely, but I worry about this kind of thing and I think there are other ways to accomplish this that would avoid that potential problem... and possibly address Scout's comment at the same time.

(Side note: I won't be able to review this in one sitting, so I'm going to add single comments instead of queueing them up in a GitHub "review".)

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -69,6 +70,7 @@ const AppLauncher = _.flow(
             ['Running', () => 'Almost ready...'],
             ['Stopping', () => 'Notebook runtime environment is stopping, which takes ~4 minutes. You can restart it after it finishes.'],
             ['Stopped', () => 'Notebook runtime environment is stopped. Start it to edit your notebook or use the terminal.'],
+            ['Reconfiguring', () => 'Notebook runtime environment is updating, please wait.'],
```
breilly2: Tangent: You said `Updating` is a status that we can already get back. However, it looks like we're not handling that case well here. Should we add that case while we're at it?

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -253,6 +254,9 @@ const PreviewHeader = ({ queryParams, cluster, readOnlyAccess, onCreateCluster,
       [clusterStatus === 'Stopping', () => h(StatusMessage, [
         'Notebook runtime environment is stopping, which takes ~4 minutes. You can restart it after it finishes.'
       ])],
+      [clusterStatus === 'Reconfiguring', () => h(StatusMessage, [
```
breilly2: Same comment about `Updating` here as above.

src/libs/cluster-utils.js
panentheos: That's certainly a possibility. The two general options I see are 1. picking a value that's less likely to conflict, e.g. `'LeoReconfiguring'`, or 2. stop rolling the flag in and treat it as a separate value. I avoided 2 in the initial implementation because it would have involved much more rewriting of the existing logic. Thoughts?

src/pages/workspaces/workspace/applications/AppLauncher.js
panentheos: Actually, it looks like there's some incorrect behavior above WRT deciding when to show the iframe. Perhaps it's unlikely since it would involve a spark cluster and RStudio, but still something we should fix.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
panentheos: This one is fine, because `'Updating'` is in `usableStatuses`, so we should never get here with that status.

breilly2: Oh, that's interesting. I didn't expect that but I guess it makes sense. Maybe that's also the answer to the other case above? (Or at least evidence of the thinking behind why `Updating` isn't a case there.)

src/libs/cluster-utils.js
breilly2: I think it's worth doing option 1. Not only will it avoid the conflict but it will make it clear that this status value is special when seeing it elsewhere in the code, such as when mixed with other values in `Utils.switchCase` clauses and the case of calculating the display value that Scout commented on.

The consequence I see of option 2 is that the uses of `Utils.switchCase` on status would have to be transformed into `Utils.cond`. I agree that would generate a larger diff, but I'm inclined to think it would be the right thing to do in the absence of using option 1. (I would be open to discussion about that, and also whether or not that could be a follow-on refactor in a separate PR).

Side note: I had an even larger change in mind involving gathering the behavior around different cluster states and putting them in one place instead of having it spread throughout the code. I think I've talked myself out of that but wanted to mention that my gut told me to explore that option in case anyone else who ponders it might reach a different conclusion.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
breilly2: Also, TOL: The `usableStatuses` logic influencing this `Utils.cond` is a little too far away for my liking. I wish the evidence that this isn't a bug were at least within this component, thus potentially avoiding this conversation. Since we are already having the conversation, I want to take advantage of the moment and learn from it. (Thank you @dmohs for enabling this kind of thinking for me!)

 I bring it up so that we think about this in the future, thus getting the most possible value out of this conversation now that it's happened. :)

dmohs: I endorse this line of thinking :)

panentheos: Elsewhere in this file we use `console.assert` to encode these assumptions explicitly, perhaps we could do that here as well.

breilly2:
Testing observation: When I increase the disk size, the `PATCH` succeeds and the runtime stays available. However, when we immediately fetch the runtime list again, the result is `status: "Running"`, `patchInProgress: false`, and `runtimeConfig` with the old disk size. In this condition, we don't re-query the runtime list for another 2 minutes. During that time, the runtime configuration panel continues to show the old disk size even though the change is done in a few seconds.

panentheos:
@breilly2 Interesting. That seems like the same class of issue as what the flag was introduced for. There's an update in flight, but it hasn't registered any changes yet so we don't know about it.

breilly2:
Another testing observation: Changing number of workers in a Dataproc cluster also does not result in `patchInProgress: true` and, unlike what the screen in the UI says, does _not_ result in a real-time update. When forced a refresh of runtime status inside of the usual 2-minute polling interval, I found that the status was actually `Updating`.

breilly2:
Correction to my previous comment. Changing the number of workers (when > 2) _is_ a real-time update in the sense that you can still use the runtime. I was confused because we disable the runtime drawer in this case. I still wonder if leo should be telling us there is a `patchInProgress`, but we could also consider allowing users to open the drawer to see the configuration but prevent them from making modifications until the update is complete. This behavior is unrelated to this PR.

breilly2 (approved):
👍 Looks good. I tested various cases, including ones that caused leonardo to return `patchInProgress: true`. Specifically, this is when the runtime needs to be stopped and restarted.

I also tested some cases where I expected `patchInProgress: true` that do _not_ require a stop. However, leonardo seems to only return `patchInProgress: true` when the runtime needs to be stopped. I think this could be improved, but I'm confident that there isn't any regression here.

See my top-level PR comments for further details about testing observations.

PR #2183, opened 2020-07-02
Remove/disable free credit enrollment [SATURN-1752]

breilly2:
Entering the enrollment flow relies on setting global state `freeCreditsActive` in `state.js`. This removes just about every case where that happens, and also removes/changes handling of when the profile's `trialState` is `Enabled`.

### DIFF
--- a/src/components/TrialBanner.js
+++ b/src/components/TrialBanner.js
@@ -63,6 +63,7 @@ export const TrialBanner = Utils.connectStore(authStore, 'authState')(class Tria
     const { finalizeTrial } = this.state
     const { trialState } = profile
     const removeBanner = localStorage.getItem('removeBanner')
+    if (trialState === 'Enabled') return null // Disable new sign-ups
     if (!trialState || !isSignedIn || !acceptedTos || trialState === 'Finalized' || removeBanner === 'true') return null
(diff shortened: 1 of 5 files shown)


ehanna4:
Looked over code and think it all looks reasonable - I agree with the approach to leave the currently enabled users logic in and just not allowing anybody new. Haven't tested yet - will work on now

panentheos (approved):
Looks good. Verified locally with config overrides that Enabled now has no effect, while Enrolled still shows up.

src/components/TrialBanner.js
```diff
@@ -63,6 +63,7 @@ export const TrialBanner = Utils.connectStore(authStore, 'authState')(class Tria
     const { finalizeTrial } = this.state
     const { trialState } = profile
     const removeBanner = localStorage.getItem('removeBanner')
+    if (trialState === 'Enabled') return null // Disable new sign-ups
```
panentheos, line 66: TOL: This effectively removes the Enabled message, which is the only one with `isExternal: false`. That means that the last `freeCreditsActive.set(true)` is unreachable, so we could prune that state and the whole `FreeCreditsModal` if we want.

PR #2184, opened 2020-07-07
Update Deps (lightweight changes) [SATURN-1745]

ehanna4:
As this is my first solo "update deps" I am breaking down the dependency updates into a few PR's to manage the testing more easily. This PR is much of the very lightweight changes (i.e the changes that should be backwards compatible (minor and patch releases), see https://semver.org/#semantic-versioning-200)

Updated Dependencies:
storybook, animate, and font-awesome

Testing:
- For animate: Tested the notifications and their animations on and off page, the different hover states on various pages
- For font awesome: Checked various icons throughout the site to ensure they look the same and behave the same
- For storybook: Tested running storybook locally and ran through many of the stories to ensure they still work.

To run storybook you can run `yarn run storybook` from the terminal rather than `yarn start` and it will pop up a localhost that has all the stories we have created you can run through. For context Storybook is a tool we have used in the past to test small features without dealing with spinning up the entire application.


s-rubenstein (approved):
LGTM. Tested font-awesome by opening menu icons, animation by generating a toast, and ran through several storybook runs.

PR #2185, opened 2020-07-10
Update Deps Heftier Changes [SATURN-1746]

ehanna4:
Updated:
- array move
- iframe resizer
- uuid
- date-fns
- downshift
- lodash
- jszip

Tested by:
- checked the data explorer iframe
- checked adjusting columns order about in the data table
- the uuid in runtime name and notification id and validated
- Tested the date in notifications and profile
- Tested workflow page autocomplete
- tested downloading sample set


s-rubenstein:
@ehanna4 shouldn't FA, Storybook, and animate have gone in on a previous PR? I remember testing and approving those.

ehanna4:
@blrubenstein you are 100% correct, I was just copying commit messages into the description of the PR and did not recognize the repetition. I think when I made this branch originally I branched off my first update-deps so those commits messages are there but looking at the file diff those changes aren't in this PR. I will update the description. Thanks for pointing that out!

panentheos (approved):
Looks good. Tested by clicking around affected areas and running storybook.

PR #2188, opened 2020-07-21
Update integration test docs [No Jira]

breilly2:
Just some doc updates.

### DIFF
--- a/integration-tests/README.md
+++ b/integration-tests/README.md
@@ -45,13 +45,16 @@ The following environment variables are parsed by the tests:
 #### Convenient one-liner:
 
 ```sh
-[HEADLESS=false] \
-[ENVIRONMENT=local|dev|alpha|perf|staging] \
 TERRA_SA_KEY=$(vault read --format=json secret/dsde/alpha/common/firecloud-account.pem | jq .data) \
 LYLE_SA_KEY=$(vault read --format=json secret/dsde/terra/envs/common/lyle-user-service-account-key | jq .data) \
 yarn test
 ```
-`HEADLESS` and `ENVIRONMENT` are optional. By default, the tests will run headless against dev.
+Optionally, additional useful options are:
+```sh
+HEADLESS=false \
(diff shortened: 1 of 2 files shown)


integration-tests/README.md
```diff
@@ -45,10 +45,13 @@ The following environment variables are parsed by the tests:
 #### Convenient one-liner:
 
 ```sh
+[HEADLESS=false] \
+[ENVIRONMENT=local|dev|alpha|perf|staging] \
```
panentheos: Adding these in this format makes this command no longer paste-able. Perhaps we should make sure the options are documented sufficiently above and let this be a simple example.

integration-tests/README.md
```diff
@@ -45,10 +45,13 @@ The following environment variables are parsed by the tests:
...
 TERRA_SA_KEY=$(vault read --format=json secret/dsde/alpha/common/firecloud-account.pem | jq .data) \
 LYLE_SA_KEY=$(vault read --format=json secret/dsde/terra/envs/common/lyle-user-service-account-key | jq .data) \
 yarn test
 ```
+`HEADLESS` and `ENVIRONMENT` are optional. By default, the tests will run headless against dev.
```
panentheos: This appears to conflict with the default for ENVIRONMENT in the previous section.

panentheos (approved):
Looks good overall, see comments. Tested the code snippets and export steps.

integration-tests/README.md
breilly2: Thanks for catching. Will fix.

integration-tests/README.md
breilly2: Breaking copy/paste did cross my mind. I'm glad you weighed in. The full documentation of all of the options above is a bit heavy for someone looking for a convenient one-liner but could still benefit from a reminder of these two options in particular. I'll pull them out of the `sh` block and just mention them in this section.

PR #2189, opened 2020-07-21
[IA-2029] add community section to dropdown

jdcanas:
New field possible in the json that powers the image drop down to indicate community images.

To test, I uploaded an images file with some community images, and verified the dropdown has a new section with the community images

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -332,7 +332,8 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
       onChange: onEnvChange,
       isSearchable: true,
       isClearable: false,
-      options: [{ label: 'JUPYTER ENVIRONMENTS', options: _.map(({ label, image }) => ({ label, value: image }), leoImages) },
+      options: [{ label: 'JUPYTER ENVIRONMENTS', options: _.map(({ label, image }) => ({ label, value: image }), _.filter(({isCommunity}) => !isCommunity, leoImages)) },
+       { label: 'COMMUNITY-MAINTAINED JUPYTER ENVIRONMENTS (verified partners)', options: _.map(({ label, image }) => ({ label, value: image }), _.filter(({isCommunity}) => isCommunity, leoImages)) },
         {
           label: 'OTHER ENVIRONMENTS',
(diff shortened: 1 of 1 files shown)


breilly2:
@jdcanas What is your desired timeframe for getting this to production?

src/components/NewClusterModal.js
```diff
@@ -332,11 +332,17 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
       isSearchable: true,
       isClearable: false,
-      options: [{ label: 'JUPYTER ENVIRONMENTS', options: _.map(({ label, image }) => ({ label, value: image }), leoImages) },
+      options: [
         {
-          label: 'OTHER ENVIRONMENTS',
-          options: [{ label: 'Custom Environment', value: CUSTOM_MODE }, { label: 'Project-Specific Environment', value: PROJECT_SPECIFIC_MODE }]
-        }]
+          label: 'JUPYTER ENVIRONMENTS', options: _.map(({ label, image }) => ({ label, value: image }),
+            _.filter(({ isCommunity }) => !isCommunity, leoImages))
```
breilly2, line 338: Suggestion: I think you could use `_.groupBy` instead of two separate `_filter`s. That would make it more explicit that the two resulting sets are both disjoint and, together, still contain all elements.

jdcanas:
> @jdcanas What is your desired timeframe for getting this to production?

I'd like to get it in by the end of the week. There's a workshop on Monday and they'd like to have this functionality for it

src/components/NewClusterModal.js
jdcanas: hmm... i like the idea, but in the .json file `isCommunity` is optional, and only specified when true, which I like since most images are not community maintained. Without that the groups are ['true', 'undefined'], and there doesn't seem to be a clean way to map the keys without obfuscating the intent further

breilly2 (approved):
👍 Looks good. I tested with dev's current image list which did not have any `isCommunity` images, so there is no difference in that case. I also tested with the following ajax override:
```
ajaxOverridesStore.set([{ filter: 'https://storage.googleapis.com/storage/v1/b/terra-docker-image-documentation/o/terra-docker-versions.json', fn: ajaxOverrideUtils.mapJsonBody(items => _.map(i => i.image.match('gatk') ? { ...i, isCommunity: true } : i, items)) }])
```
which marks all of the GATK images with `isCommunity`. Worked as expected with the GATK images in a new section.

PR #2190, opened 2020-07-23
Update Deps Integration Tests [SATURN-1772]

ehanna4:
Updated the dependencies for the integration tests. Tested by running the suite of tests on alpha, local, dev, and perf.


panentheos (approved):
Looks good.

PR #2191, opened 2020-07-31
[IA-1904] Rstudio users experiencing 401s after sustained use

gpcarr:
This PR changes `PeriodCookieSetter` in `cluster-common.js` from setting every 15 minutes to setting every 5 minutes. We have been having Rstudio users report 401s after leaving Rstudio idle for some time and coming back to it. By setting our third party tool cookie more often we are hoping to mitigate this issue.

### DIFF
--- a/src/components/cluster-common.js
+++ b/src/components/cluster-common.js
@@ -93,6 +93,6 @@ export const PeriodicCookieSetter = ({ namespace, runtimeName, leading }) => {
   const signal = Utils.useCancellation()
   Utils.usePollingEffect(
     withErrorIgnoring(() => Ajax(signal).Clusters.notebooks(namespace, runtimeName).setCookie()),
-    { ms: 15 * 60 * 1000, leading })
+    { ms: 5 * 60 * 1000, leading })
   return null
 }


gpcarr:
@rtitle not sure who to add as a reviewer from the terra-ui team, lmk!

src/components/cluster-common.js
```diff
@@ -93,6 +93,6 @@ export const PeriodicCookieSetter = ({ namespace, runtimeName, leading }) => {
   const signal = Utils.useCancellation()
   Utils.usePollingEffect(
     withErrorIgnoring(() => Ajax(signal).Clusters.notebooks(namespace, runtimeName).setCookie()),
-    { ms: 15 * 60 * 1000, leading })
+    { ms: 5 * 60 * 1000, leading })
```
breilly2, line 96: According to https://broadworkbench.atlassian.net/browse/IA-779, the token should be good for an hour. I wonder if refreshing more frequently is going to help since refreshing every 15 minutes seems like more than enough. Do you have a sense of why 15 minutes might not be frequent enough?

breilly2 (approved):
👍 One comment/question, but we don't see any reason not to go ahead and try this.

src/components/cluster-common.js
gpcarr: We seem to only be seeing this issue with RStudio users and not Jupyter users. Specifically one screenshot we got from a user was the 401 during an R package installation, which can take a long time because they are compiled from source. Could have to do with how long RStudio users wait for things in their workflow but I don't have a definitive answer as we have seen this happen in different situations.

PR #2192, opened 2020-08-03
POC only: Galaxy UI

panentheos:
This is effectively just a mockup at this point, no need to review.


panentheos:
Closing now that the designs are done.

PR #2195, opened 2020-08-14
[Saturn-1722] Update text to say cloud environment over notebook runtime

s-rubenstein:
Tested by looking at a few places in the UI, did not exhaustively generate every error message, since the only change is the text. There are a few places I did not change it:

Still says runtime on NewClusterModal line 588 because that one is actually talking about specifically the runtime
WDLVIewer line 41 seems to have runtime in a stack of complicated | separated strings. I don't know what is going on there and didn't want to mess with it.
Clusters.js line 62 I did not change, because it seemed like it was actually still talking about notebook runtimes, but may potentially change in @panentheos 's work on SATURN-1747

List of changes was found by searching in the src directory for `runtime` (case insensitive) and then using intellij's breakdown to restrict me to looking at string constants only.

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -58,7 +58,7 @@ export const ClusterErrorModal = ({ cluster, onDismiss }) => {
   const [loadingClusterDetails, setLoadingClusterDetails] = useState(false)
 
   const loadClusterError = _.flow(
-    withErrorReporting('Error loading notebook runtime details'),
+    withErrorReporting('Error loading cloud environment details'),
     Utils.withBusyState(setLoadingClusterDetails)
   )(async () => {
     const { errors: clusterErrors } = await Ajax().Clusters.cluster(cluster.googleProject, cluster.runtimeName).details()
@@ -77,7 +77,7 @@ export const ClusterErrorModal = ({ cluster, onDismiss }) => {
   Utils.useOnMount(() => { loadClusterError() })
 
   return h(Modal, {
-    title: `Notebook Runtime Error${userscriptError ? ' due to Userscript Error' : ''}`,
+    title: `Cloud Environment Creation Failed${userscriptError ? ' due to Userscript Error' : ''}`,
     showCancel: false,
     onDismiss
   }, [
@@ -134,7 +134,7 @@ export default class ClusterManager extends PureComponent {
     const rStudioLaunchLink = Nav.getLink('workspace-app-launch', { namespace, name, app: 'RStudio' })
 
     if (cluster.status === 'Error' && prevCluster.status !== 'Error' && !_.includes(cluster.id, errorNotifiedClusters.get())) {
-      notify('error', 'Error Found In Notebook Runtime', {
+      notify('error', 'Error Creating Cloud Environment', {
         message: h(ClusterErrorNotification, { cluster })
       })
       errorNotifiedClusters.update(Utils.append(cluster.id))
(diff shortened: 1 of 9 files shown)


src/libs/disk-utils.js
```diff
@@ -1,21 +0,0 @@
-import { Fragment } from 'react'
```
s-rubenstein: Deleted this file because it is unused and we ended up not using the deletePDText in our delete work, so seemed not useful at this point.

src/components/ClusterManager.js
```diff
@@ -293,7 +293,7 @@ export default class ClusterManager extends PureComponent {
       (activeClusters.length > 1 || activeDisks.length > 1) && h(Link, {
         style: { marginRight: '1rem' },
         href: Nav.getLink('clusters'),
-        tooltip: 'Multiple runtimes and/or persistent disks found in this billing project. Click to select which to delete.'
+        tooltip: 'Multiple cloud environments found in this billing project. Click to select which to delete.'
```
s-rubenstein, line 277: This just says cloud environments instead of runtimes and/or persistent disks now.

src/components/NewClusterModal.js
```diff
@@ -612,7 +612,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
         div({ style: styles.whiteBoxContainer }, [
           div({ style: { fontSize: '0.875rem', fontWeight: 600, marginBottom: '0.5rem' } }, ['Cloud compute configuration']),
           // TODO PD: remove the profile select
-          div({ style: { marginBottom: '1rem' } }, ['Select from one of the default runtime profiles or define your own']),
+          div({ style: { marginBottom: '1rem' } }, ['Select from one of the default cloud compute profiles or define your own']),
```
s-rubenstein: I made this say `cloud compute` instead of `cloud environment` because it is talking specifically about the `Cloud compute configuration`

s-rubenstein:
Lint failures existing on mob-pd branch.

ehanna4:
I think I missed the conversations when the name changes were happening so just want to be sure I understand I have the right idea:
we have 'runtimes' which are the VMs
we have 'persistent disks' and 'disks' which are the storage pieces 
and 
we have 'cloud environments' which refer to the entire thing encapsulating both disk and VM?

and a 'cloud compute' is the same as a runtime?

ehanna4:
This has been Amelia and Lou approved:

For the Leonardo Powered Analysis:
1. Currently we have ‘runtimes’ (formally referred to as ‘clusters’ ) which are the VMs which will now be known as ‘cloud computes’
2. We currently have  ‘disks’ which are the storage pieces and going forward we will have these same ‘disks’ which refer to the volatile storage and in addition will have ‘persistent disks’ which will be the storage that can be saved
3. Currently we refer to the whole configuration of VM and Disk as “Runtime Configuration” but going forward when we want to refer to the thing that encapsulates both disk (persistent or not ) and VM we will say ‘Cloud Environment’

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -48,7 +48,7 @@ const AppLauncher = _.flow(
       h(Fragment, [
         h(PeriodicCookieSetter, { namespace, runtimeName }),
         app === 'RStudio' && h(PlaygroundHeader, [
-          'This feature is in early development. Your files are saved on your runtime but not to your workspace. We encourage you to frequently ',
+          'This feature is in early development. Your files are saved on your cloud environment but not to your workspace. We encourage you to frequently ',
```
ehanna4, line 50: Not sure how to re-phrase this,  is this still true even if you have a PD?

src/libs/disk-utils.js
ehanna4: 👍

src/pages/workspaces/workspace/applications/AppLauncher.js
s-rubenstein: I don't know if RStudio uses PD. How do we spin up RStudio?

ehanna4: @sharmatime does R Studio use PD too?

ehanna4: @s-rubenstein I think it's basically that an RStudio can be spun up with a custom docker image and if it's an image that relies on R then the code chooses to spin up an Rstudio environment rather than the Jupyter one. Since it gets decided in the same Cloud Environment Config i imagine it uses PD but just want to be sure

ehanna4 (approved):
These changes look good, thanks!

PR #2196, opened 2020-08-14
Make 'notebook runtimes' page disk-aware

panentheos:


panentheos:
Marking this reviewable. There's one TODO remaining that I think could use some group brainstorming, but it should be good otherwise.

src/pages/Clusters.js
```diff
@@ -52,88 +129,171 @@ const Clusters = () => {
...
+            cellRenderer: ({ rowIndex }) => {
+              return formatUSD(persistentDiskCostMonthly(filteredDisks[rowIndex]))
+            }
+          },
+          {
+            size: { basis: 50, grow: 0 },
+            headerRenderer: () => null,
+            cellRenderer: ({ rowIndex }) => {
+              const { id, status } = filteredDisks[rowIndex]
+              // TODO PD: there should be some way of identifying which disk is connected to which runtime
```
breilly2, line 282: Ah, interesting. One quick thought that doesn't quite address this is to add some icons to at least indicate whether or not a runtime has a disk attached and whether or not a disk is attached to a runtime. Currently the only way to see that it either in the runtime delete modal or whether or not the disk delete button is disabled.

breilly2 (approved):
👍  I checked it out in browser with both an attached disk and a floating disk and all looks good. I didn't test exhaustively... and only with the ajax overrides, not against a real leonardo.

src/pages/Clusters.js
panentheos: Yeah, one potential idea is to add icons, and let people click on them to highlight the linkages between rows. Not sure if that's a bit too weird though.

PR #2197, opened 2020-08-17
Add event for viewing 'about PD' [SATURN-1790]

ehanna4:
Add a Mixpanel event for when a user views the "About PD" screen. 

Properties:
Workspace Namespace
Workspace Name
currently has attached disk

tested with different overrides and clicking the two 'learn more' links

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -17,6 +17,7 @@ import {
 } from 'src/libs/cluster-utils'
 import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
+import Events, { extractWorkspaceDetails } from 'src/libs/events'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 import validate from 'validate.js'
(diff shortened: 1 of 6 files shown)


src/components/NewClusterModal.js
```diff
@@ -985,7 +1004,15 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
         // TODO PD: What do we do with this?
         !sparkMode && !isPersistentDisk && div([
           p(['Time to upgrade your compute runtime. Terra’s new persistent disk feature will safegard your work and data.']),
-          h(Link, { onClick: () => this.setState({ viewMode: 'aboutPersistentDisk' }) }, ['Learn more'])
+          h(Link, {
+            onClick: async () => {
```
panentheos: Since this `onClick` handler is the same as the other one, we could extract it into a helper method that handles the state transition and the event logging, and call it from both places.

src/components/NewClusterModal.js
```diff
@@ -985,7 +1004,15 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+              this.setState({ viewMode: 'aboutPersistentDisk' })
+              await Ajax().Metrics.captureEvent(Events.aboutPersistentDiskView, {
```
panentheos: TOL: There's no harm in the `await` here, but we normally just fire-and-forget metrics calls. Is there a need for it here?

src/components/NewClusterModal.js
ehanna4: good point!

src/components/NewClusterModal.js
ehanna4: ah nope I will remove that

Shakespeared (approved):
This makes sense to me! Tested by clicking 'about persistent disk' and viewing the event in mixpanel.

PR #2198, opened 2020-08-18
implement simplified form view

panentheos:


src/components/NewClusterModal.js
```diff
@@ -143,7 +143,8 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
       sparkMode: cloudService === cloudServices.GCE ? false : numberOfWorkers === 0 ? 'master' : 'cluster',
       ...currentConfig,
       masterDiskSize: currentCluster?.runtimeConfig?.masterDiskSize || currentCluster?.runtimeConfig?.diskSize || DEFAULT_DISK_SIZE,
-      deleteDiskSelected: false
+      deleteDiskSelected: false,
+      simplifiedForm: !currentCluster
```
breilly2, line 147: TOL: It feels to me like we should try to keep the user in the simplified view if that's how they created the runtime. That certainly has some challenges... such as detecting this case (maybe we could tag the runtime?) and also what happens when there's a newer version of the default image? Is that still "simple" or is it suddenly custom?

src/components/NewClusterModal.js
```diff
@@ -977,17 +963,60 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+                      'to keep your data even after you delete your compute'
+                    ])
+                  ])
+                ]),
+                renderActionButton()
+              ]),
+              renderCostBreakdown()
+            ]),
+            div({ style: styles.whiteBoxContainer }, [
+              div({ style: { display: 'flex', alignItems: 'center', justifyContent: 'space-between' } }, [
```
breilly2, line 990: TOL: Luck has it that `center` appears to also accomplish `baseline`. I could argue either one being the intent.

breilly2 (approved):
👍  Looks good. I tested creating a default runtime (actually calling leo) and everything behaved as I expected.

src/components/NewClusterModal.js
panentheos: I was under the impression that the simplified view was just intended as a quick-start flow, not something we intended to return to, but we can check on that and add it if needed. FWIW we currently have the "becoming custom" behavior with regard to the leo image, though that always feels a little weird .

src/components/NewClusterModal.js
panentheos: Going to leave it as center unless we see otherwise.

PR #2199, opened 2020-08-19
IA-2148 IA-2138 update patch criteria

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2148

* When a cluster is `Stopped`, we want to disallow users to update worker number (limitation on leo side).
* When a cluster is `Stopped`, we want to allow users to update disk size, machine type, and number of worker local SSDs

Tested in fiab
Screen Shot 2020-08-19 at 6 16 54 PM
Screen Shot 2020-08-19 at 6 16 36 PM

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -354,7 +354,6 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
 
     const bottomButtons = () => {
       const canUpdate = this.canUpdate()
-      console.log(`can update ${canUpdate}`)
       const updateOrReplace = canUpdate ? 'update' : 'replace'
 
       return h(Fragment, [
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -188,13 +188,14 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     const currentClusterConfig = normalizeRuntimeConfig(currentCluster.runtimeConfig)
     const userSelectedConfig = normalizeRuntimeConfig(this.getRuntimeConfig())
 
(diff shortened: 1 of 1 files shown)


src/components/NewClusterModal.js
```diff
@@ -352,6 +354,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
 
     const bottomButtons = () => {
       const canUpdate = this.canUpdate()
+      console.log(`can update ${canUpdate}`)
```
ehanna4: is this console.log() intentional?

Qi77Qi: oops..no...will delete

panentheos:
I thought the patch call was able to orchestrate status changes in order to update some properties. Is this a new restriction, or an existing issue?

Qi77Qi:
@panentheos it's an existing issue...If a user tries to update number of workers while the runtime is `Stopped` today, they'll be stuck in `Updating` forever (and we don't allow `delete` during `Updating`). I have an open PR to disallow this from leo more quickly...For other properties, I think it's okay for users to update even if runtme is `Stopped`

s-rubenstein:
@Qi77Qi I am going to suggest we wait on this for the persistent disk work to merge in. We are fairly close to that being ready, and completely overhauled our update logic, so would need to reimplement this on top of those changes, as well as handle what is likely to be a pretty gnarly rebase.

What is the urgency on this change?

Qi77Qi:
@s-rubenstein that's fine...I think it can wait a bit...or while you guys are in there, if you could update the patch logic, that'd be great :)

breilly2:
So if the cluster is running, we can allow changing # of workers which happens as a live update. But if the cluster is stopped and the number of workers changes, we have to delete and recreate the cluster?

ehanna4:
Closing this pull request since it will require extensive rework with the new PD stuff, so have added a ticket to the Saturn Backlog to address this work.

PR #2200, opened 2020-08-20
Mixpanel for open cloud env configure drawer [SATURN-1786]

ehanna4:
Add event to capture opening the cloud env config drawer

This should fire anytime the cloud environment configuration (aka newclustermodel) drawer is opened.


src/libs/events.js
```diff
@@ -8,6 +8,7 @@ import * as Utils from 'src/libs/utils'
 const eventsList = {
   aboutPersistentDiskView: 'about:persistentDisk:view',
   applicationLaunch: 'application:launch',
+  cloudEnvironmentConfigOpen: 'cloudEnvironment:Config:Open',
```
s-rubenstein: I think this should be `cloudEnvironment:config:open` rather than the capitals on Config and Open. If I get an agreement, I can make that change and merge this. Otherwise it LGTM.

src/components/NewClusterModal.js
```diff
@@ -134,13 +134,22 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+  }
+
   constructor(props) {
     super(props)
     const currentCluster = this.getCurrentCluster()
     const { cloudService, ...currentConfig } = normalizeRuntimeConfig(currentCluster?.runtimeConfig || { masterMachineType: 'n1-standard-4' })
     const { numberOfWorkers } = currentConfig // want these to be put into state below, unlike cloudService
     const currentPersistentDisk = this.getCurrentPersistentDisk()
 
+    Ajax().Metrics.captureEvent(Events.cloudEnvironmentConfigOpen, {
```
panentheos: This produces a side effect, so it should go in `componentDidMount` instead of the constructor.

src/libs/events.js
panentheos: Agreed.

PR #2202, opened 2020-08-20
[SATURN-1814] Filter deleting disk

s-rubenstein:
Considered filtering at the source, but decided that could be too destructive of a change for now. Added a comment about what I was thinking about if we want to consider it for the future. If it sounds like a good idea to the reviewer too, I can open a ticket for that refactor.


Shakespeared (approved):
This looks good to me! 

tested with ajax overrides by creating new runtime and making sure it doesn't attach the disk with status 'deleting'

PR #2203, opened 2020-08-20
Add Mixpanel Event for Create Cloud Environment

ehanna4:
This has large overlaps with the PR for adding an event for Update Cloud Environment. Please see that PR for details on the properties that should be extracted into this event and how to get those properties.

Also please note that makeWorkspaceObj() is being moved about and used by most events in the newClusterModel so watch out for that.


src/components/NewClusterModal.js
```diff
@@ -256,6 +261,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
       })
     }
     if (shouldCreateRuntime) {
+      Ajax().Metrics.captureEvent(Events.cloudEnvironmentCreate, { ...extractWorkspaceDetails(this.makeWorkspaceObj()), ...newRuntime, isDefaultConfig: !!this.state.simplifiedForm, pdSize: '', cost: '' })
```
panentheos, line 284: This will log a create event if a new runtime is being created implicitly because a previous one wasn't updatable, but I think we'd rather log that as an 'update' since that's how the user sees the action. It might be worth moving this into `renderApplyButton` and handling it alongside the 'update' and 'delete' cases, since that's where we figure out how to describe the action via the label.

PR #2204, opened 2020-08-21
Kf mixpanel delete environment event

Shakespeared:
send an event to mixpanel when either a PD, a runtime, or both are deleted. I'm pretty sure this should report correctly in all of the delete cases, but I'd appreciate a double check on that! 

the event should track these properties: 

Property: Delete Persistent Disk: [yes/no]
Property: Delete Cloud Compute: [yes/no]

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -210,7 +210,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     return { size: newPersistentDisk.size }
   }
 
-  getMixpanelDeleteMetrics(isDeleteRuntime, isDeletePersistentDisk) {
+  sendDeleteMetrics(isDeleteRuntime, isDeletePersistentDisk) {
     Ajax().Metrics.captureEvent(Events.cloudEnvironmentDelete, {
       ...extractWorkspaceDetails(this.makeWorkspaceObj()),
       isDeleteRuntime,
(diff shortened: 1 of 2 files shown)


src/components/NewClusterModal.js
```diff
@@ -205,6 +210,14 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     return { size: newPersistentDisk.size }
   }
 
+  getMixpanelDeleteMetrics(isDeleteRuntime, isDeletePersistentDisk) {
```
s-rubenstein: Nit: I'd remove get from this. Maybe `submitMixpanelDeleteMetrics`? `get` makes me think it is side-effect free.

src/components/NewClusterModal.js
```diff
@@ -247,6 +260,9 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
         })
       })
     }
+
+    this.getMixpanelDeleteMetrics(shouldDeleteRuntime, this.willDeletePersistentDisk())
```
s-rubenstein: We probably shouldn't submit this event if both are false (AKA update or create). Mind throwing this in an if? Or moving to both of the below deletion cases?

src/components/NewClusterModal.js
Shakespeared: Ah, yep good call. I'll do that now!

src/components/NewClusterModal.js
Shakespeared: oh, right, good catch I hadn't thought of that. Will update accordingly. I don't want to put it in both of the deletion cases, becuase then it will fire twice if you are deleting both (if that's what you meant)

s-rubenstein (approved):
LGTM

PR #2205, opened 2020-08-21
Kf mixpanel update environment options

Shakespeared:
Sending mixpanel metrics on the 'update' case for either a runtime or a PD. This one has a lot of suggested properties, I did my best to capture everything,  but if I missed something please let me know. Tested with multiple overrides, viewed results in mixpanel. 

See notes below on suggested properties:

Property: Default Config: Yes/No - NOT IMPLEMENTED (I couldn't find a way to determine the original defaults reliably without hardcoding them, Amelia okayed me to drop this)
Property: App Config: [name] - I believe this means the workspace info, but if anyone knows otherwise let me know
Property: CPUs: [number]
Property: Memory: [number]
Property: Disk: [number]
Property: Compute Type: [selection]
Property: Persistent Disk: [number]
Property: Cost [number]
Property: Delete Persistent Disk: [yes/no/NA] - N/A, this is handled under the 'delete' case
Property: Cloud Env: [GCE, dataproc]
Property: [placeholder for everything dataproc related]
Property: Reattach persistent disk [yes/no] - N/A, this is handled under the 'create' case

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -277,8 +277,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     }
     if (shouldUpdateRuntime) {
       await Ajax().Clusters.cluster(namespace, currentCluster.runtimeName).update({
-        runtimeConfig
-      })
+        runtimeConfig })
     }
     if (shouldCreateRuntime) {
       await Ajax().Clusters.cluster(namespace, Utils.generateClusterName()).create({
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -219,6 +219,22 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     })
   }
 
+  sendUpdateMetrics(newRuntime, newPersistentDisk) {
+    const machineType = newRuntime.cloudService === cloudServices.GCE ? newRuntime.machineType : newRuntime.masterMachineType
+    const { cpu, memory } = findMachineType(machineType)
+
+    Ajax().Metrics.captureEvent(Events.cloudEnvironmentUpdate, {
+      ...extractWorkspaceDetails(this.makeWorkspaceObj()),
+      ...newRuntime,
+      cpu,
+      memory,
+      persistentDiskSize: newPersistentDisk?.size,
+      runtimeCostPerHour: Utils.formatUSD(runtimeConfigCost(this.getPendingRuntimeConfig())),
+      runtimePausedCostPerHour: Utils.formatUSD(ongoingCost(this.getPendingRuntimeConfig())),
+      persistentDiskCostPerMonth: (newPersistentDisk && Utils.formatUSD(persistentDiskCostMonthly(newPersistentDisk)))
+    })
+  }
+
   applyChanges = _.flow(
     Utils.withBusyState(() => this.setState({ loading: true })),
     withErrorReporting('Error creating cloud environment')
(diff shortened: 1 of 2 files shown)


ehanna4:
You'll want to make this PR against the `mob-pd` branch rather than `dev`

panentheos:
FYI I changed the base branch to mob-pd

src/components/NewClusterModal.js
```diff
@@ -266,6 +282,10 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
       this.sendDeleteMetrics(shouldDeleteRuntime, this.willDeletePersistentDisk())
     }
 
+    if (shouldUpdateRuntime || shouldUpdatePersistentDisk) {
```
panentheos, line 285: This will catch PATCH calls, but not create/delete pairs, which we're also considering 'updates' from the user's perspective. We should consider moving this to `renderApplyButton` and handling it alongside the 'create' and 'delete' cases, to make sure the logic is consistent across them.

src/components/NewClusterModal.js
```diff
@@ -219,6 +219,22 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+
+    Ajax().Metrics.captureEvent(Events.cloudEnvironmentUpdate, {
+      ...extractWorkspaceDetails(this.makeWorkspaceObj()),
+      ...newRuntime,
+      cpu,
+      memory,
+      persistentDiskSize: newPersistentDisk?.size,
+      runtimeCostPerHour: Utils.formatUSD(runtimeConfigCost(this.getPendingRuntimeConfig())),
+      runtimePausedCostPerHour: Utils.formatUSD(ongoingCost(this.getPendingRuntimeConfig())),
+      persistentDiskCostPerMonth: (newPersistentDisk && Utils.formatUSD(persistentDiskCostMonthly(newPersistentDisk)))
```
panentheos, line 234: There's a `getPendingDisk` method that we should use to feed the cost calculation now. Passing `newPersistentDisk` technically works for the moment, but it's the wrong type. The cost functions expect leo-shaped objects, whereas the environment config stuff is an internal representation.

src/components/NewClusterModal.js
```diff
@@ -219,6 +219,22 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+    const machineType = newRuntime.cloudService === cloudServices.GCE ? newRuntime.machineType : newRuntime.masterMachineType
+    const { cpu, memory } = findMachineType(machineType)
+
+    Ajax().Metrics.captureEvent(Events.cloudEnvironmentUpdate, {
+      ...extractWorkspaceDetails(this.makeWorkspaceObj()),
+      ...newRuntime,
+      cpu,
+      memory,
+      persistentDiskSize: newPersistentDisk?.size,
+      runtimeCostPerHour: Utils.formatUSD(runtimeConfigCost(this.getPendingRuntimeConfig())),
```
panentheos, line 232: Do we want to format these? It might be easier to compute with on the analytics side if we leave it as a number.

Shakespeared:
> FYI I changed the base branch to mob-pd

Whoops, thanks for that! I shouldn't have done it last thing on Friday evening haha.

src/components/NewClusterModal.js
Shakespeared: Yeah I was thinking of 'create' and 'update' from the code perspective rather than from the users perspective, but if we just want to capture any update that comes from a user clicking 'update' (which does make sense) then I agree that the button is a better place to put this. My plan was to get this shipshape, merge it to mob-pd, and then build the 'create' case from the code I have in this PR as they're very similar.

PR #2206, opened 2020-08-25
include disk cost in cluster manager rollup

panentheos:
Looking for feedback on the change in cluster-utils.js specifically. Is it clear enough?

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -12,7 +12,7 @@ import { dataSyncingDocUrl } from 'src/data/machines'
 import rLogo from 'src/images/r-logo.svg'
 import { Ajax } from 'src/libs/ajax'
 import { getDynamic, setDynamic } from 'src/libs/browser-storage'
-import { clusterCost, collapsedClusterStatus, currentCluster, trimClustersOldestFirst } from 'src/libs/cluster-utils'
+import { clusterCost, collapsedClusterStatus, currentCluster, persistentDiskCost, trimClustersOldestFirst } from 'src/libs/cluster-utils'
 import colors from 'src/libs/colors'
 import { reportError, withErrorReporting } from 'src/libs/error'
 import * as Nav from 'src/libs/nav'
@@ -260,7 +260,7 @@ export default class ClusterManager extends PureComponent {
           })
       }
     }
-    const totalCost = _.sum(_.map(clusterCost, clusters))
+    const totalCost = _.sum(_.map(clusterCost, clusters)) + _.sum(_.map(persistentDiskCost, persistentDisks))
     const activeClusters = this.getActiveClustersOldestFirst()
     const activeDisks = _.remove({ status: 'Deleting' }, persistentDisks)
     const { Creating: creating, Updating: updating, LeoReconfiguring: reconfiguring } = _.countBy(collapsedClusterStatus, activeClusters)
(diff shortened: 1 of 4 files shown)


src/components/ClusterManager.js
```diff
@@ -260,7 +260,7 @@ export default class ClusterManager extends PureComponent {
           })
       }
     }
-    const totalCost = _.sum(_.map(clusterCost, clusters))
+    const totalCost = _.sum(_.map(clusterCost, clusters)) + _.sum(_.map(persistentDiskCost, persistentDisks))
```
s-rubenstein, line 263: Opt, TOL: Could we fold the sums together? Or is it worth wrapping in a sum call instead of a +?

src/libs/cluster-utils.js
```diff
@@ -51,13 +51,9 @@ export const runtimeConfigCost = config => {
...
 }
 
-export const persistentDiskCostMonthly = config => {
-  return config.size * monthlyStoragePrice
-}
-
-export const persistentDiskCost = config => {
-  return config.size * storagePrice
-}
+export const [persistentDiskCost, persistentDiskCostMonthly] = _.map(price => ({ size, status }) => {
```
s-rubenstein: This makes a lot of sense to me, and I like this change. It seems clear enough to me.

s-rubenstein (approved):
This LGTM. One optional thing, that I just thought of while reading it through, but it seems like probably a stylistic thing.

src/libs/cluster-utils.js
```diff
@@ -51,13 +51,9 @@ export const runtimeConfigCost = config => {
...
+  return _.includes(status, ['Deleting', 'Failed']) ? 0.0 : size * price
+}, [storagePrice, monthlyStoragePrice])
```
breilly2: It's clear to me in the context of this PR (since I can see the before and after of the refactor), however, it does take a bit of mental energy for me to read this, and I know I can't immediately forget my current knowledge of the intention to create 2 functions with the same implementation. Reading code that generates code is a difficult and infrequently used skill.

Maybe extracting the higher order function might help? Something like:

```
const generateDiskCostFunction = price => ({ size, status }) => {
  return _.includes(status, ['Deleting', 'Failed']) ? 0.0 : size * price
}
export const persistentDiskCost = generateDiskCostFunction(storagePrice)
export const persistentDiskCostMonthly = generateDiskCostFunction(monthlyStoragePrice)
```

Giving the function a name allows reuse without having to add the complexity of calling `_.map` and destructuring the result.

panentheos: Thanks, that was the other way I was considering, but I was having trouble coming up with a name.

s-rubenstein: I do think this is easier to read, seeing both ways. So +1 on Brian's comment.

src/components/ClusterManager.js
panentheos: I thought about it, this way felt slightly clearer to me.

src/libs/cluster-utils.js
breilly2: Yeah, naming is tough. I'm not committed to the name I suggested. I did choose the `generate` prefix because I don't think we use it extensively in any other context, and maybe we can have a convention of using it as a clue that we're in a code generation mindset.

PR #2207, opened 2020-08-25
Update runtime UI to enable persistent disks

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>


integration-tests/tests/run-notebook.js
```diff
@@ -24,7 +24,10 @@ const testRunNotebookFn = _.flow(
   await click(page, clickable({ text: 'Create Notebook' }))
   await click(page, clickable({ textContains: notebookName }))
   await click(page, clickable({ text: 'Edit' }))
-  await select(page, 'ENVIRONMENT', 'Hail')
+  // wait for drawer to slide in before clicking, otherwise we'll miss the button
+  await delay(1000)
```
ehanna4, line 40: I think our convention has typically been to use a findElement() that is on the page to ensure a thing is there rather than a delay, thoughts on finding something that's in the default drawer?

src/components/cluster-common.js
```diff
@@ -96,3 +96,16 @@ export const PeriodicCookieSetter = ({ namespace, runtimeName, leading }) => {
...
+export const SaveFilesHelp = () => {
+  return h(Fragment, [
+    p([
+      'If you want to save some files permanently, such as input data, analysis outputs, or installed packages, ',
+      h(Link, {
+        href: 'https://support.terra.bio/hc/en-us/articles/360026639112',
+        ...Utils.newTabLinkProps
+      }, ['move them to the workspace bucket.'])
+    ]),
+    p(['Note: Jupyter notebooks are autosaved to the workspace bucket, and deleting your disk will not delete your notebooks.'])
```
ehanna4, line 109: Why does this say Jupyter notebooks specifically?
1. If it is actually only about Jupyter notebooks I wonder if it is appropriate to have in this cluster-common file 
2. Or if it applies to all cloud environments then this wording should be updated

integration-tests/tests/run-notebook.js
breilly2: `click()` does await the thing we ask it to click. Unfortunately, it exists in the DOM and is clickable while it is animating in. I figured this out by capturing a screenshot of the element, which resulted in an image of the correct size but with the wrong content; it only showed right-hand side of the button. From this, I conclude that puppeteer is finding the element, calculating the point on the screen that it's going to click (or the rectangle it's going to capture for the screenshot) and then clicking (or capturing the pixels). Since the panel is still sliding in, the element has moved far enough left that the click misses the button.

We theoretically had this same problem with the `select()` for Hail below. However, since that is a wide select box, puppeteer's chosen click point (presumably closer to the right edge of the panel than we might expect) still activated the select whereas it is more unreliable for a narrower element like a button.

src/components/ClusterManager.js
```diff
@@ -273,13 +255,14 @@ export default class ClusterManager extends PureComponent {
             shape: 'play',
             onClick: () => this.setState({ createModalDrawerOpen: true }),
             disabled: busy || !canCompute,
-            tooltip: canCompute ? 'Create notebook runtime' : noCompute,
-            'aria-label': 'Create notebook runtime'
+            tooltip: canCompute ? 'Create cloud environment' : noCompute,
+            'aria-label': 'Create cloud environment'
```
s-rubenstein, line 259: TOL: Should this aria-label have the same ternary as above? (Existing issue if so, but may be worth thinking about?)

src/components/ClusterManager.js
```diff
@@ -310,16 +293,16 @@ export default class ClusterManager extends PureComponent {
...
           onClick: () => this.setState({ createModalDrawerOpen: true }),
           disabled: isDisabled
         }, [
           div({ style: { marginLeft: '0.5rem', paddingRight: '0.5rem', color: colors.dark() } }, [
-            div({ style: { fontSize: 12, fontWeight: 'bold' } }, 'Notebook Runtime'),
+            div({ style: { fontSize: 12, fontWeight: 'bold' } }, 'Cloud Environment'),
             div({ style: { fontSize: 10 } }, [
               span({ style: { textTransform: 'uppercase', fontWeight: 500 } }, [currentStatus === 'LeoReconfiguring' ? 'Updating' : (currentStatus || 'None')]),
-              currentStatus && ` (${Utils.formatUSD(totalCost)} hr)`
+              !!totalCost && ` (${Utils.formatUSD(totalCost)} hr)`
```
s-rubenstein, line 305: Why do we not still need currentStatus here? I would expect we might still want to know if there is a runtime or a disk before displaying this at all.

src/components/ClusterManager.js
ehanna4: this is also a thing in the above lines 241, 230, and 222. I haven't tested yet but i imagine the ternary should be used in the aria-label as you've suggested. Will look in the UI and see how it feels and report back

src/components/ClusterManager.js
ehanna4: currentStatus I think was used to determine if there was an existing cloud environment previously. Now totalCost (which is the sum of existing disk and cloud vm) will let us know if we have a current cloud enviornment since if not it would be 0 which i think is falsy

integration-tests/tests/run-notebook.js
ehanna4: ooh okay that makes sense, is the 1 second based on the time we selected for the animation to take to slide in?

src/components/NewClusterModal.js
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+        isClearable: false,
+        onlyInteger: true,
+        value,
+        onChange
+      })
+    ])
+  ])
+}
+
+const FancyRadio = ({ labelText, children, name, checked, onChange, style = {} }) => {
```
s-rubenstein: Nit: Fancy probably isn't the right prefix here. Maybe `BlockyWarningRadio`?

src/components/NewClusterModal.js
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+    children
+  ])
+}
+
 const CUSTOM_MODE = '__custom_mode__'
 const PROJECT_SPECIFIC_MODE = '__project_specific_mode__'
 
 export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterModal extends Component {
   static propTypes = {
     currentCluster: PropTypes.object,
```
s-rubenstein: Can we pull currentCluster out of propTypes now that we are passing the whole list?

src/components/NewClusterModal.js
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
-
-    if (isStopRequired) {
-      notify('info', 'To be updated, your runtime will now stop, and then start. This will take 3-5 minutes.')
-    }
-
-    return onSuccess(
-      Ajax().Clusters.cluster(googleProject, runtimeName).update({
-        runtimeConfig: this.getRuntimeConfig()
+  /**
+   * Transform the new environment config into the shape of runtime config
```
s-rubenstein: Nit: I would probably preface this with: 'this method transforms' or something, right now my instinctual reaction was that this is a todo.

src/components/NewClusterModal.js
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
-    return jupyterUserScriptUri !== originalJupyterUserScriptUri
+    }
   }
 
-  hasImageChanged() {
-    const { selectedLeoImage, customEnvImage, currentClusterDetails } = this.state
-    const { imageUrl } = currentClusterDetails ? this.getImageUrl(currentClusterDetails) : ''
-    return !_.includes(imageUrl, [selectedLeoImage, customEnvImage])
+  /**
+  * Transform new environment config into the shape of a disk returned from Leo.
```
s-rubenstein: Same as above, although for some reason this reads less like a todo to me.

src/pages/Clusters.js
```diff
@@ -51,88 +120,171 @@ const Clusters = () => {
...
+            cellRenderer: ({ rowIndex }) => {
+              return formatUSD(persistentDiskCostMonthly(filteredDisks[rowIndex]))
+            }
+          },
+          {
+            size: { basis: 50, grow: 0 },
+            headerRenderer: () => null,
+            cellRenderer: ({ rowIndex }) => {
+              const { id, status } = filteredDisks[rowIndex]
+              // TODO PD: there should be some way of identifying which disk is connected to which runtime
```
Shakespeared: TOL: if we wanted to identify which runtime a disk is attached to that's a little tricky since the runtimes don't have names or IDs visible in this table, but if we just wanted to show which disks were attached vs floating we could do that pretty easily, something like this: 

```
{
            size: { basis: 150, grow: 0 },
            headerRenderer: () => h(Sortable, { sort: diskSort, field: 'attached', onSort: setDiskSort }, ['Attached to Runtime']),
              const { id } = filteredDisks[rowIndex]
              return _.some({ runtimeConfig: { persistentDiskId: id } }, clusters) ? 'Attached' : 'Unattached'
            }
          }
````

which might be nice for users to be able to see whether they have floating disks or not. Conversely we could add a 'PD is attached' (or something) column under runtimes to communicate the same thing. 

(edit: the code formatting is being weird sorry!)

File: src/components/NewClusterModal.js [hunk 10]
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
 
-  //determines whether the changes are applicable for a call to the leo patch endpoint
-  //see this for a diagram of the conditional this implements https://drive.google.com/file/d/1mtFFecpQTkGYWSgPlaHksYaIudWHa0dY/view
-  //this function returns true for cases 2 & 3 in this diagram
-  canUpdate() {
-    const { currentCluster } = this.props
+  sendCloudEnvironmentMetrics() {
+    const { runtime: newRuntime, persistentDisk: newPersistentDisk } = this.getNewEnvironmentConfig()
+    const { runtime: oldRuntime, persistentDisk: oldPersistentDisk } = this.getOldEnvironmentConfig()
+    const newMachineType = newRuntime && (newRuntime.cloudService === cloudServices.GCE ? newRuntime.machineType : newRuntime.masterMachineType)
```
s-rubenstein, line 229: This feels like something we should have a helper for, considering the logic is repeated right below. (I was wondering if `findMachineType` would do it? Certainly seems like the right function.)

src/components/NewClusterModal.js
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+      [(this.state.viewMode === 'deleteEnvironmentOptions'), () => 'cloudEnvironmentDelete'],
+      [(!!oldRuntime), () => 'cloudEnvironmentUpdate'],
+      () => 'cloudEnvironmentCreate'
+    )
 
-    if (!currentCluster) return false
+    Ajax().Metrics.captureEvent(Events[metricsEvent], {
+      ...extractWorkspaceDetails(this.makeWorkspaceObj()),
+      ..._.mapKeys(key => `newRuntime_${key}`, newRuntime),
+      newRuntime_exists: !!newRuntime,
```
s-rubenstein, line 242: Optional: Maybe we can make a `getRuntimeMetadata` function or something? I feel like there's a lot of repeated keys here that we could maybe abstract?

src/components/NewClusterModal.js
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+              masterMachineType,
+              masterDiskSize,
+              numberOfWorkers: newNumberOfWorkers,
+              ...(newNumberOfWorkers && {
+                numberOfPreemptibleWorkers,
+                workerMachineType,
+                workerDiskSize
+              })
+            }),
+            toolDockerImage: _.includes(selectedLeoImage, [CUSTOM_MODE, PROJECT_SPECIFIC_MODE]) ? customEnvImage : selectedLeoImage,
```
s-rubenstein: Opt Nit: I might move these up to the top since they are for both cases, just think it may make that a little more clear?

src/components/NewClusterModal.js
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+          masterMachineType: runtimeConfig.masterMachineType || 'n1-standard-4',
+          masterDiskSize: runtimeConfig.masterDiskSize || 100,
+          numberOfWorkers,
+          ...(numberOfWorkers && {
+            numberOfPreemptibleWorkers: runtimeConfig.numberOfPreemptibleWorkers || 0,
+            workerMachineType: runtimeConfig.workerMachineType || 'n1-standard-4',
+            workerDiskSize: runtimeConfig.workerDiskSize || 100
+          })
+        }),
+        toolDockerImage: this.getImageUrl(currentClusterDetails),
```
s-rubenstein: Opt Nit: Same as above, I might move this up so its more clear it isn't part of the ternary

integration-tests/tests/run-notebook.js
breilly2: Unfortunately not. The animation time is set to 0.2s (I think). I tried delaying for 500ms in the test but was still seeing the same failures.

src/components/NewClusterModal.js
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
-        'Increasing the disk size or changing the number of workers (when the number of workers is >2) results in a real-time update to your runtime. ',
-        'Updating the number of workers can take around 2 minutes. ',
-        'During this update, you can continue to work.'
-      ])
+  // original diagram (without PD) for update runtime logic: https://drive.google.com/file/d/1mtFFecpQTkGYWSgPlaHksYaIudWHa0dY/view
+  isStopRequired() {
+    const { runtime: oldRuntime } = this.getOldEnvironmentConfig()
+    const { runtime: newRuntime } = this.getNewEnvironmentConfig()
+
+    // TODO PD: should we consider runtime status here? Consider what to do in the case of autopause
```
s-rubenstein: Still a todo here. I think the answer might be no?

src/components/NewClusterModal.js
```diff
@@ -42,217 +43,428 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
-  componentDidMount = withErrorReporting('Error loading cluster', async () => {
-    const { currentCluster, namespace } = this.props
+  componentDidMount = _.flow(
+    withErrorReporting('Error loading cluster'),
+    Utils.withBusyState(v => this.setState({ loading: v }))
+  )(async () => {
+    const { namespace } = this.props
+    const currentCluster = this.getCurrentCluster()
+
+    Ajax().Metrics.captureEvent(Events.cloudEnvironmentConfigOpen, {
```
s-rubenstein, line 471: Nit: I would probably call this event `openCloudEnvironmentConfig` because it starts with the action.

src/components/NewClusterModal.js
```diff
@@ -1,34 +1,35 @@
...
 import { withModalDrawer } from 'src/components/ModalDrawer'
 import { InfoBox } from 'src/components/PopupTrigger'
 import TitleBar from 'src/components/TitleBar'
-import { machineTypes, profiles } from 'src/data/machines'
+import { cloudServices, machineTypes } from 'src/data/machines'
 import { Ajax } from 'src/libs/ajax'
-import { deleteText, findMachineType, formatRuntimeConfig, normalizeRuntimeConfig, runtimeConfigCost } from 'src/libs/cluster-utils'
+import {
+  currentCluster,
+  DEFAULT_DISK_SIZE, findMachineType, normalizeRuntimeConfig, ongoingCost, persistentDiskCostMonthly, runtimeConfigCost
```
Shakespeared: Nit: maybe rename 'ongoingCost'? To me this name is unclear what 'ongoing' means and what the cost is for, and doesn't match our other cost names. It looks like in the UI we're displaying it as 'paused runtime cost per hour.' Maybe we could rename it to pausedRuntimeCost or baseRuntimeCost to indicate that it will be accrued whether the runtime is running or paused, as it's used as part of the overall runtime cost calculation.

src/components/NewClusterModal.js
Shakespeared: yeah a helper might be good for this. The way `machineType` and `masterMachineType` are used interchangeably in the environment configs was making it tricky to inline the call to `findMachineType` (see just below this line) which is why I pulled it out like that, but I agree it could be streamlined!

integration-tests/tests/run-notebook.js
ehanna4: okay this sounds reasonable to me, thanks for the explanation!

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
-          div({ style: { flex: 1 } }),
-          h(ButtonSecondary, { style: { marginRight: '2rem' }, onClick: onDismiss }, 'Cancel'),
-          h(ButtonPrimary, {
-            disabled: !this.hasChanges() || !!errors,
-            tooltip: Utils.summarizeErrors(errors),
+    const renderActionButton = () => {
+      const { runtime: oldRuntime } = this.getOldEnvironmentConfig()
+      const { runtime: newRuntime } = this.getNewEnvironmentConfig()
+      const commonButtonProps = { disabled: !this.hasChanges() || !!errors, tooltip: Utils.summarizeErrors(errors) }
+      const mayShowCustomImageWarning = viewMode === undefined
```
s-rubenstein: This is only used in one place, should we inline it or move the other logic into here?

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+      const mayShowEnvironmentWarning = _.includes(viewMode, [undefined, 'customImageWarning'])
```
s-rubenstein: This is only used in one place, should we inline it or move the other logic into here?

src/components/cluster-common.js
breilly2: At the moment, I believe this is specific to Jupyter. My understanding is that RStudio does not have "notebook" documents the same way Jupyter does. I do not know the file/data management model for Galaxy.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -148,32 +148,43 @@ const WorkspaceAccessError = () => {
   ])
 }
 
-const useClusterPolling = namespace => {
+const useCloudEnvironmentPolling = namespace => {
```
ehanna4, line 151: Did we decide to do an overall re-name of cluster to cloud environment? Seems to me like consistency would be best so either should change all or maybe keep this as cluster

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -148,32 +148,43 @@ const WorkspaceAccessError = () => {
...
     try {
-      const newClusters = await Ajax(signal).Clusters.list({ googleProject: namespace, creator: getUser().email })
+      const [newDisks, newClusters] = await Promise.all([
+        Ajax(signal).Disks.list({ googleProject: namespace, creator: getUser().email }),
+        Ajax(signal).Clusters.list({ googleProject: namespace, creator: getUser().email })
+      ])
+      // Consider whether we want to filter out deleting clusters and disks here, rather than lower down
+      // It is safe to do so, because detaching a disk from a cluster is a synchronous action, and the
+      // cluster will not be put in deleting state if the disk is still attached, so we don't need to worry
+      // about trying to attach disks that are attached to a deleting disk
```
ehanna4: This seems like a TODO

src/components/ClusterManager.js
breilly2: I wonder if there's an expectation that `aria-label` should be stable over time so people don't get lost in an ever-changing landscape.

src/components/ClusterManager.js
breilly2: Right, and `totalCost` could be non-zero without a runtime if there is a floating persistent disk.

src/components/NewClusterModal.js
```diff
@@ -502,179 +817,366 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-            'This will update your existing runtime. You will not lose any files. ',
-            'After the update is finished you will be able to start your runtime. ',
-            'Note that updating the number of workers requires your runtime to already be started.'
+      ])
+    }
+
+
+    const handleLearnMoreAboutPersistentDisk = () => {
+      this.setState({ viewMode: 'aboutPersistentDisk' })
+      Ajax().Metrics.captureEvent(Events.aboutPersistentDiskView, {
```
Shakespeared, line 1003: Nit: related to @s-rubenstein 's comment on the `cloudEnvironmentConfigOpen` mixpanel event, for consistency purposes we might also want to rename this to `viewAboutPersistentDisk`

src/components/NewClusterModal.js
```diff
@@ -502,179 +817,366 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+    const renderAboutPersistentDisk = () => {
+      return div({ style: styles.drawerContent }, [
+        h(TitleBar, {
+          style: styles.titleBar,
+          title: 'About persistent disk',
+          onDismiss,
+          onPrevious: () => this.setState({ viewMode: undefined })
+        }),
+        div({ style: { lineHeight: 1.5 } }, [
+          // TODO PD: this language could be cleaner, e.g. we don't abbreviate 'PD' anywhere else
```
Shakespeared: Just calling this out as an unresolved TODO, we should un-abbreviate PD here. Does anyone know if this was included in the full language pass? It seems like we're using 'compute' in some places where we might want to update to 'cloud environment' or 'cloud compute'

src/components/NewClusterModal.js
breilly2: Ah, thank you for giving me some guidance! I don't think I've ever settled on a personal preference for the phrasing of doc comments like this. I like the notion of making sure it sounds like a description rather than a command/request/TODO.

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+    children
+  ])
+}
+
 const CUSTOM_MODE = '__custom_mode__'
 const PROJECT_SPECIFIC_MODE = '__project_specific_mode__'
 
 export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterModal extends Component {
   static propTypes = {
     currentCluster: PropTypes.object,
```
breilly2:
```suggestion
```

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
-      notify('info', 'To be updated, your runtime will now stop, and then start. This will take 3-5 minutes.')
-    }
-
-    return onSuccess(
-      Ajax().Clusters.cluster(googleProject, runtimeName).update({
-        runtimeConfig: this.getRuntimeConfig()
+  /**
+   * Transform the new environment config into the shape of runtime config
+   * returned from leonardo. Specifically used as an input to cost calculation
+   * for potential new configurations.
```
breilly2:
```suggestion
   * Transforms the new environment config into the shape of runtime config
   * returned from leonardo. The cost calculation functions expect that shape,
   * so this is necessary to compute the cost for potential new configurations.
```

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
   }
 
-  hasImageChanged() {
-    const { selectedLeoImage, customEnvImage, currentClusterDetails } = this.state
-    const { imageUrl } = currentClusterDetails ? this.getImageUrl(currentClusterDetails) : ''
-    return !_.includes(imageUrl, [selectedLeoImage, customEnvImage])
+  /**
+  * Transform new environment config into the shape of a disk returned from Leo.
+  * Used as an input to cost calculation.
+  */
```
breilly2:
```suggestion
   * Transforms the new environment config into the shape of a disk returned
   * from leonardo. The cost calculation functions expect that shape, so this
   * is necessary to compute the cost for potential new disk configurations.
   */
```

src/components/NewClusterModal.js
Shakespeared: Upon consideration, I think leaving this as-is for now is the best option. We could pull out a function, but I don't think that the value of streamlining this block would outweigh the complication of having another function in an already long file - especially as it would only be useful for this particular instance, and it might be hard to name is so the purpose of the function is clear so I think it would impact the readability. So for those reasons I'm inclined to leave it alone for now, and potentially come back to refactor after this is merged

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+            } : {
+              masterMachineType,
+              masterDiskSize,
+              numberOfWorkers: newNumberOfWorkers,
+              ...(newNumberOfWorkers && {
+                numberOfPreemptibleWorkers,
+                workerMachineType,
+                workerDiskSize
+              })
+            }),
```
breilly2: Since there is a bunch of common stuff and a big block of stuff that is different based on the type of runtime, I think it might read well to extract the big block of conditional stuff. Then this object construction is a straightforward 4 lines with only inline conditionals.
```suggestion
    const runtimeDetails = {
      ...(cloudService === cloudServices.GCE ? {
        machineType: masterMachineType,
        ...(this.shouldUsePersistentDisk() ? {
          persistentDiskAttached: true
        } : {
          diskSize: masterDiskSize
        })
      } : {
        masterMachineType,
        masterDiskSize,
        numberOfWorkers: newNumberOfWorkers,
        ...(newNumberOfWorkers && {
          numberOfPreemptibleWorkers,
          workerMachineType,
          workerDiskSize
        })
      })
    }
    return {
      runtime: Utils.cond(
        [(viewMode !== 'deleteEnvironmentOptions'), () => {
          return {
            cloudService,
            ...runtimeDetails,
```
Does that work for you @s-rubenstein ?

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+        } : {
+          masterMachineType: runtimeConfig.masterMachineType || 'n1-standard-4',
+          masterDiskSize: runtimeConfig.masterDiskSize || 100,
+          numberOfWorkers,
+          ...(numberOfWorkers && {
+            numberOfPreemptibleWorkers: runtimeConfig.numberOfPreemptibleWorkers || 0,
+            workerMachineType: runtimeConfig.workerMachineType || 'n1-standard-4',
+            workerDiskSize: runtimeConfig.workerDiskSize || 100
+          })
+        }),
```
breilly2:
```suggestion
    const runtimeDetails = currentClusterDetails ? {
      ...(cloudService === cloudServices.GCE ? {
        machineType: runtimeConfig.machineType,
        ...(runtimeConfig.persistentDiskId ? {
          persistentDiskAttached: true
        } : {
          diskSize: runtimeConfig.diskSize
        })
      } : {
        masterMachineType: runtimeConfig.masterMachineType || 'n1-standard-4',
        masterDiskSize: runtimeConfig.masterDiskSize || 100,
        numberOfWorkers,
        ...(numberOfWorkers && {
          numberOfPreemptibleWorkers: runtimeConfig.numberOfPreemptibleWorkers || 0,
          workerMachineType: runtimeConfig.workerMachineType || 'n1-standard-4',
          workerDiskSize: runtimeConfig.workerDiskSize || 100
        })
      })
    } : {}
    return {
      runtime: currentClusterDetails ? {
        cloudService,
        ...runtimeDetails,
```

src/components/NewClusterModal.js
breilly2: I also prefer starting the name with the action. However, in the event list, starting with `cloudEnvironment` puts all of the related events together when sorted alphabetically:
```
const eventsList = {
  ...
  cloudEnvironmentConfigOpen: 'cloudEnvironment:config:open',
  ...
}
```

We could certainly do something like this instead:
```
const eventsList = {
  ...
  cloudEnvironment: {
    create: 'cloudEnvironment:create',
    delete: 'cloudEnvironment:delete',
    openConfig: 'cloudEnvironment:config:open',
    update: 'cloudEnvironment:update'
  },
  ...
}
```
but I think we can hold off on that refactor until another day.

src/components/NewClusterModal.js
breilly2: I think the naming of the idea could be valuable, though I'm a little confused at first read. Does "may" here mean "might" or "is allowed to"? I think it's the latter in this case, which is not how I typically use "may" in code.

Feels to me like these "may" conditions and the remainder of the logic are two different aspects of a "should". The "may" bits represent whether or not its the right time in the flow to show the warning. The remainder of the logic represents whether or not the warning is indicated based on the change being requested.

I'd prefer to keep them separate for now and consider some refactoring of names and/or logic for clarity at a later time.

integration-tests/tests/run-notebook.js
panentheos, line 40: I think the underlying need here is to wait for the initial ajax calls to finish, since the spinner blocks interaction. This might be better handled with `waitForNoSpinners`

src/libs/ajax.js
```diff
@@ -10,6 +10,165 @@ import { ajaxOverridesStore, knownBucketRequesterPaysStatuses, requesterPaysProj
...
+  dataproc: { runtimes: [dataprocRuntime], disks: [] },
+  dataprocCluster: { runtimes: [dataprocClusterRuntime], disks: [] },
+  disk: { runtimes: [], disks: [disk] },
+  diskAndDeletingDisk: { runtimes: [], disks: [disk, deletingDisk] },
+  gceAndDisk: { runtimes: [gceRuntime], disks: [disk] },
+  gceAndAttachedDisk: { runtimes: [gceRuntimeWithPd], disks: [disk] },
+  dataprocAndDisk: { runtimes: [dataprocRuntime], disks: [disk] },
+  attachedDiskEdgeCase: { runtimes: [gceRuntimeWithPd, dataprocRuntime], disks: [disk] }
+})
+
```
panentheos: We should remove these when we're done testing.

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
   }
 
   constructor(props) {
     super(props)
-    const { currentCluster } = props
-    const { cloudService, ...currentConfig } = normalizeRuntimeConfig(currentCluster?.runtimeConfig || profiles[0].runtimeConfig)
-    const { masterDiskSize, masterMachineType, numberOfWorkers } = currentConfig // want these to be put into state below, unlike cloudService
-    const matchingProfile = _.find(({ runtimeConfig }) => _.isMatch({ masterMachineType, masterDiskSize }, normalizeRuntimeConfig(runtimeConfig)), profiles)
+    const currentCluster = this.getCurrentCluster()
+    const { cloudService, numberOfWorkers, ...currentConfig } = normalizeRuntimeConfig(currentCluster?.runtimeConfig || { masterMachineType: 'n1-standard-4' })
```
panentheos: This initializes the form with the runtime info from the `list` call, but the `oldEnvironmentConfig` is built using the runtime info from the `details` call. If there's ever a discrepancy (due to timing, etc), then the form will show a false delta, which could result in some confusing behavior. If we're going to use the `details` call to populate the `oldEnvironmentConfig`, then we should also re-initialize the form with that same data to avoid that situation.

TOL: It's also worth pointing out that we only make the `details` call once, while we continue to poll `list` in the background. So this approach is slightly less resilient in a few edge cases involving multiple windows (e.g. if the runtime changes in the background while the modal is open). Probably not something that's worth worrying about, but just wanted to mention it.

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
-    ]))
+  getCurrentPersistentDisk() {
+    const currentCluster = this.getCurrentCluster()
+    const { clusters, persistentDisks } = this.props
+    const id = currentCluster?.runtimeConfig.persistentDiskId
+    const attachedIds = _.without([undefined], _.map(cluster => cluster.runtimeConfig.persistentDiskId, clusters))
+    return id ?
+      _.find({ id }, persistentDisks) :
+      _.last(_.sortBy('auditInfo.createdDate', _.filter(({ id }) => !_.includes(id, attachedIds),
+        _.filter(disk => disk.status !== 'Deleting', persistentDisks))))
```
panentheos: nit: these two `filter` calls could be combined into one

src/components/NewClusterModal.js
panentheos: If we want to do this, rather than guarding the expression with `currentClusterDetails`, we could make it an inline function that computes the value, to delay execution until we know we want it.

src/components/ClusterManager.js
s-rubenstein: canCompute is unlikely to change on folks though, right? Although I guess if I am using a screen reader and bouncing from one workspace to another, I might expect it to look the same.

src/components/NewClusterModal.js
s-rubenstein: SGTM

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
+      persistentDisk: currentPersistentDiskDetails ? { size: currentPersistentDiskDetails.size } : undefined
+    }
   }
 
-  hasChanges() {
-    const { currentCluster } = this.props
-    if (!currentCluster) return true
+  hasAttachedDisk() {
+    const { currentClusterDetails } = this.state
+    return currentClusterDetails?.runtimeConfig.persistentDiskId
```
panentheos: nit: This could use `oldEnvironmentConfig` instead

src/components/NewClusterModal.js
s-rubenstein: The fact that you are also confused now means that maybe we should rename it for clarity now, because we aren't going to have _more_ context later.

src/components/NewClusterModal.js
s-rubenstein: This looks good to me.

src/components/NewClusterModal.js
s-rubenstein: Looks good.

src/components/NewClusterModal.js
panentheos: Let's say no for now and remove this TODO. We might end up revisiting this when we incorporate the Qi's PR.

src/components/NewClusterModal.js
```diff
@@ -42,386 +43,717 @@ const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:
...
         this.setState({ selectedLeoImage: imageUrl })
       } else if (currentClusterDetails.labels.saturnIsProjectSpecific === 'true') {
         this.setState({ selectedLeoImage: PROJECT_SPECIFIC_MODE, customEnvImage: imageUrl })
       } else {
         this.setState({ selectedLeoImage: CUSTOM_MODE, customEnvImage: imageUrl })
       }
-
-      if (jupyterUserScriptUri) {
-        this.setState({ jupyterUserScriptUri, profile: 'custom' })
-      }
```
panentheos, line 276: We still need to set `jupyterUserScriptUri` here, since it's not included in the data from the `list` call. Note: This will probably interact with the re-initialization of the form state from the `details` call, as mentioned above.

src/components/NewClusterModal.js
```diff
@@ -502,179 +817,366 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-      ])],
-      ['delete', () => h(Fragment, [
-        h(deleteText),
-        div({ style: { display: 'flex', justifyContent: 'flex-end', marginTop: '1rem' } }, [
-          h(ButtonSecondary, { style: { marginRight: '2rem' }, onClick: () => this.setState({ viewMode: undefined }) }, ['CANCEL']),
-          h(ButtonPrimary, { onClick: () => onSuccess(this.deleteCluster()) }, ['DELETE'])
+      ])
+    }
+    const renderEnvironmentWarning = () => {
+      return this.willDetachPersistentDisk() ?
```
panentheos: TOL: This could be combined with the `cond` below

src/pages/workspaces/workspace/WorkspaceContainer.js
panentheos: In general we're leaving variables and methods named 'cluster', but since this now deals with both runtimes and disks, proactively changing the name here seems like a good idea.

src/components/ClusterManager.js
breilly2: That is exactly the thought process I went thorough as I was adding my comment. :)

integration-tests/tests/run-notebook.js
breilly2: I'll be happier if that works. I think that was my original thought but ultimately went with this. I'll try `waitForNoSpinners`. (This is a small enough task for me today, and I want to see it myself given the context I already have with debugging the issue.)

src/libs/ajax.js
breilly2: Yes. I recognize that these are likely to get out-of-date over time and could end up being a detriment instead of a benefit. However, I am sad that we'll have to recreate these any time we make changes to the panel in order to test it. As with the debug panels, there is a bit of learning built in to the existence of these cases, and we rely on that kind of learning to know what to manually test when making changes.

src/components/NewClusterModal.js
panentheos: Lou said that this was still the preferred language.

src/components/NewClusterModal.js
panentheos: I took Sky's suggestion here. I think keeping the data inline makes it easier to see the shape of this structure.

src/components/NewClusterModal.js
breilly2: Excellent points. I feel like, without a more holistic solution, we're going to end up having to trade off one edge case for another. The challenge is choosing which edge cases to prioritize.

src/components/ClusterManager.js
panentheos: I think this is a reasonable UX as-is. In fact, the accessible experience may be better than the visual one in this case. There's an `aria-label` that describes what the button does, and on focus/hover, we include an `aria-describedby` pointing to the tooltip, which describes the reason it's disabled (if applicable).

integration-tests/tests/run-notebook.js
breilly2: @panentheos Unfortunately it didn't work. This is really tricky to describe but I'll give it a try.

As a test case, I removed the slide-in and all delays between clicking "Edit" and "Customize". The test passes. I was curious to know what it thought it clicked. I instrumented capturing a screenshot of the element before clicking it. (This is not a perfect experiment because taking the screenshot takes >0 time.) Here's what it looks like:

button_no_slide_in_no_delay

My interpretation of this is that puppeteer locates the element, calculates the rectangle, then doesn't capture the screenshot until after something happens that causes the button's location on the screen to shift. Since I've disabled the slide-in, the only remaining cause I can think of is something being returned from an ajax call affecting the rendering of the panel. In this case, the test still passes because (I think) puppeteer chooses to click the middle of the rectangle, which luckily still hits the button.

When I add `waitForNoSpinners`, I've seen the test fail with a button that looks like this:

button_no_slide_wait_for_no_spinners

This looks to me like the button is covered by a spinner overlay. My hunch here is that `waitForNoSpinners` is returning before the spinner appears. Then, by coincidence, the spinner was rendered at the time that the screenshot was taken and the click attempted.

For reference, here's something like what I saw when I originally started debugging this:

button_slide_no_delay

Finally, when I added `waitForNoSpinners`, the test failed when not capturing a screenshot of the button. When I captured a screenshot of the button, the test _passed_, but the button looked like this:

button_slide_wait_for_not_spinners

I did not run multiple experiments with each conditions to determine how stable these behaviors are, but I think the evidence is clear enough that the ideal code does not reliably produce the desired behavior.

My conclusion is that the cause is elements moving due to a combination of drawer sliding and rendering changes after ajax calls return. `await delay(1000)` is the only reliable fix because of a timing mismatch between spinner rendering and test execution.

panentheos: Thanks for the detailed investigation! I wonder if we should actually have *both*: first a time delay to wait for the animation, and then a spinner check to make sure the ajax calls have finished (commented for intent). That would be more resilient to network delays. Thoughts?

breilly2: Sure. It's better than just the delay. I have worries and I started to express them here, but it's getting too complicated and I just need to move forward.

src/components/NewClusterModal.js
panentheos: Per that discussion, we'll keep these noun-first

panentheos (approved):
Looks good, thanks all!

PR #2209, opened 2020-08-25
Fix Dockstore extra inputs bug [WA-291]

aednichols:
Tested with local UI against methods created in dev.

https://github.com/DataBiosphere/terra-ui/pull/new/aen_wa_291

https://support.terra.bio/hc/en-us/community/posts/360056422292/comments/360009878072

### DIFF
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -951,6 +951,25 @@ const WorkflowView = _.flow(
         _.update('outputs', this.isSingle() ? () => ({}) : _.mapValues(_.trim))
       )
 
+      // Dockstore users who target floating tags can change their WDL via Github without explicitly selecting a new version in Terra.
+      // If the change removes or renames inputs or outputs, the `modifiedConfig` retrieved from the DB will include stale keys that
+      // are hidden from the UI and cause the workflow to fail at launch with "Validation errors: Extra inputs".
+      //
+      // The IO set `modifiedInputsOutputs` is always current because we regenerate it from the WDL each time; use it to filter out
+      // any extraneous IO keys before sending.
+      //
+      // https://broadworkbench.atlassian.net/browse/WA-291
+      const validInputs = _.map('name',  modifiedInputsOutputs['inputs'])
+      const validOutputs = _.map('name', modifiedInputsOutputs['outputs'])
+
+      modifiedConfig['inputs']  = Object.fromEntries(Object.entries(modifiedConfig['inputs']).filter(function(tuple) {
+        return validInputs.includes(tuple[0])
+      }))
+      modifiedConfig['outputs'] = Object.fromEntries(Object.entries(modifiedConfig['outputs']).filter(function(tuple) {
+        return validOutputs.includes(tuple[0])
+      }))
+
+      // POST to create new MC snapshot
       const validationResponse = await Ajax().Workspaces.workspace(namespace, name)
         .methodConfig(workflowNamespace, workflowName)
         .save(trimInputOutput(modifiedConfig))
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -962,12 +962,14 @@ const WorkflowView = _.flow(
       const validInputs = _.map('name', modifiedInputsOutputs['inputs'])
       const validOutputs = _.map('name', modifiedInputsOutputs['outputs'])
 
-      modifiedConfig['inputs'] = Object.fromEntries(Object.entries(modifiedConfig['inputs']).filter(function(tuple) {
-        return validInputs.includes(tuple[0])
-      }))
-      modifiedConfig['outputs'] = Object.fromEntries(Object.entries(modifiedConfig['outputs']).filter(function(tuple) {
-        return validOutputs.includes(tuple[0])
-      }))
+      const filterIO = (ioMap, allowedKeys) => {
+        return _.fromPairs(_.filter(pair => {
+          return _.includes(pair[0], allowedKeys);
+        }, _.toPairs(ioMap)));
+      };
+
+      modifiedConfig['inputs'] = filterIO(modifiedConfig['inputs'], validInputs);
+      modifiedConfig['outputs'] = filterIO(modifiedConfig['outputs'], validOutputs);
 
       // POST to create new MC snapshot
       const validationResponse = await Ajax().Workspaces.workspace(namespace, name)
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -951,6 +951,25 @@ const WorkflowView = _.flow(
...
+      // are hidden from the UI and cause the workflow to fail at launch with "Validation errors: Extra inputs".
+      //
+      // The IO set `modifiedInputsOutputs` is always current because we regenerate it from the WDL each time; use it to filter out
+      // any extraneous IO keys before saving.
+      //
+      // https://broadworkbench.atlassian.net/browse/WA-291
+      const validInputs = _.map('name', modifiedInputsOutputs['inputs'])
+      const validOutputs = _.map('name', modifiedInputsOutputs['outputs'])
+
+      modifiedConfig['inputs'] = Object.fromEntries(Object.entries(modifiedConfig['inputs']).filter(function(tuple) {
```
s-rubenstein: Nit: I think we prefer to use lodash `filter` (and probably `entries`). So this could be:
```
_.filter(tuple => _.includes(tuple[0], validInputs), _.fromPairs(_.toPairs(modifiedConfig['inputs'])))
```
(Or something similar. This hasn't been fully verified for syntactic correctness. See https://lodash.com/docs/#fromPairs for syntax. Although we use the functional version, so most of the time the args are switched.)

aednichols: It looks like
```
_.filter(_.toPairs(modifiedConfig['inputs']), pair => {return _.includes(desiredInputs, pair[0])})
```
works so it seems there is some truth to the arg switching statement (for both `includes` and `filter`)

aednichols:
I'm getting an error
```
 run-notebook

TimeoutError: waiting for XPath "(//a | //*[@role="button"] | //button)[contains(normalize-space(.),"Create a") or contains(@aria-label,"Create a")]" failed: timeout 30000ms exceeded
```
which doesn't seem related to my PR, so marking "ready for review"

s-rubenstein (approved):
LGTM. I would not worry about the timeout it seems unrelated

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -951,6 +951,27 @@ const WorkflowView = _.flow(
         _.update('outputs', this.isSingle() ? () => ({}) : _.mapValues(_.trim))
       )
 
+      // Dockstore users who target floating tags can change their WDL via Github without explicitly selecting a new version in Terra.
```
breilly2: Wow! To me, this breaks expectations of what a method config represents.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -951,6 +951,27 @@ const WorkflowView = _.flow(
...
+      const validInputs = _.map('name', modifiedInputsOutputs['inputs'])
+      const validOutputs = _.map('name', modifiedInputsOutputs['outputs'])
+
+      const filterIO = (ioMap, allowedKeys) => {
+        return _.fromPairs(_.filter(pair => {
+          return _.includes(pair[0], allowedKeys)
+        }, _.toPairs(ioMap)))
+      }
+
+      modifiedConfig['inputs'] = filterIO(modifiedConfig['inputs'], validInputs)
```
breilly2: I think `_.pick` (https://lodash.com/docs/4.17.15#pick) will work here:

```
modifiedConfig['inputs'] = _.pick(validInputs, modifiedConfig['inputs'])
```

The only thing to be careful about is that `_.pick` takes "paths", which are a convenience for picking properties out of nested structures. I don't remember if inputs/outputs might contain periods in their names. Even if so, I suspect the inputs/outputs objects are flat objects so it should just work. If there's any concern, though, you could use `_.pickBy`:

```
modifiedConfig['inputs'] = _.pickBy((value, key) => _.includes(key, validInputs), modifiedConfig['inputs'])
```

panentheos: Agreed, and to go one step further: If you take a look at `loadNewMethodConfig` (called when changing versions), we perform this operation as part of the pipeline of changes to the `modifiedConfig` state. Since we want to do the same thing in both places, we could extract that out as a helper, e.g. something like:
```js
const filterConfigIO = ({ inputs, outputs }) => {
  return _.flow(
    _.update('inputs', _.pick(_.map('name', inputs))),
    _.update('outputs', _.pick(_.map('name', outputs)))
  )
}
```

src/pages/workspaces/workspace/workflows/WorkflowView.js
panentheos: NB: Indeed, we explicitly warn people about this in `ImportWorkflow.js` when importing from dockstore.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -951,6 +951,27 @@ const WorkflowView = _.flow(
...
+      modifiedConfig['outputs'] = filterIO(modifiedConfig['outputs'], validOutputs)
+
+      // POST to create new MC snapshot
```
panentheos: TOL: Instead of doing this on save, we could do it on initial load, e.g. in `componentDidMount`. The difference would be that the Save button would be enabled from the start, giving a soft indication that something has changed (and also giving us the opportunity to explicitly call this out with a warning in the future, if we want)

panentheos: I just thought of a much stronger argument for this suggestion: Due to existing UI logic, filtering on load would force the user to save before being allowed to launch analysis, which should prevent the invalid fields from being sent to cromwell.

aednichols:
Thanks for the suggestions, they look great, I'll see what I can do

src/pages/workspaces/workspace/workflows/WorkflowView.js
aednichols: We intend that users who need reproducibility will be cutting Github releases so they have stable tags

Same principle as floating vs. fixed Docker images

s-rubenstein (approved):
This looks great, and is much more readable in my mind. Thanks for the changes! (And thank you to @breilly2 and @panentheos, I learned a lot from this PR)

aednichols:
@panentheos @breilly2 do I need a re-review from you folks?

Not impatient, just want to make sure I'm not inadvertently waiting for no reason

breilly2:
Yes, sorry. We noted yesterday morning that it was approved, but didn't recognize that the approval was from before reworking the code. In this case, I would like to do another round of review and testing. I will do that today.

aednichols:
Sounds good, thank you.

aednichols:
When the time comes to review this I am happy to demo/pair, I have a Dockstore demo workflow I've been testing with.

panentheos (approved):
Looks good, tested locally with ajax overrides.

PR #2211, opened 2020-08-26
Kf mixpanel metrics

Shakespeared:
Event to track mixpanel metrics for create, update, and delete cases. Sends a bunch of info to mixpanel so that metrics can be collected on what was deleted, updated, etc. 

Tested with different overrides and disk/runtime configurations

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -211,27 +211,39 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     return { size: newPersistentDisk.size }
   }
 
-  sendDeleteMetrics(isDeleteRuntime, isDeletePersistentDisk) {
-    Ajax().Metrics.captureEvent(Events.cloudEnvironmentDelete, {
-      ...extractWorkspaceDetails(this.makeWorkspaceObj()),
-      isDeleteRuntime,
-      isDeletePersistentDisk
-    })
-  }
+  sendCloudEnvironmentMetrics() {
+    const metricsEvent = Utils.cond(
+      [(this.state.viewMode === 'deleteEnvironmentOptions'), () => 'cloudEnvironmentDelete'],
+      [(!!this.getOldEnvironmentConfig().runtime), () => 'cloudEnvironmentUpdate'],
+      [() => 'cloudEnvironmentCreate']
+    )
+    const newRuntime = this.getNewEnvironmentConfig().runtime
+    const oldRuntime = this.getOldEnvironmentConfig().runtime
+    const newPersistentDisk = this.getNewEnvironmentConfig().persistentDisk
+    const oldPersistentDisk = this.getOldEnvironmentConfig().persistentDisk
+    const newMachineType = newRuntime && (newRuntime.cloudService === cloudServices.GCE ? newRuntime.machineType : newRuntime.masterMachineType)
+    const oldMachineType = oldRuntime && (oldRuntime?.cloudService === cloudServices.GCE ? oldRuntime.machineType : oldRuntime.masterMachineType)
+    const { cpu: newRuntimeCpus, memory: newRuntimeMemory } = findMachineType(newMachineType)
+    const { cpu: oldRuntimeCpus, memory: oldRuntimeMemory } = findMachineType(oldMachineType)
 
-  sendUpdateMetrics(newRuntime, newPersistentDisk) {
-    const machineType = newRuntime.cloudService === cloudServices.GCE ? newRuntime.machineType : newRuntime.masterMachineType
-    const { cpu, memory } = findMachineType(machineType)
-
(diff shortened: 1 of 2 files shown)


src/components/NewClusterModal.js
```diff
@@ -211,27 +211,35 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-    const machineType = newRuntime.cloudService === cloudServices.GCE ? newRuntime.machineType : newRuntime.masterMachineType
-    const { cpu, memory } = findMachineType(machineType)
+  sendCloudEnvironmentMetrics() {
+    const metricsEvent = Utils.cond(
+      [(this.state.viewMode === 'deleteEnvironmentOptions'), () => 'cloudEnvironmentDelete'],
+      [(!!this.getOldEnvironmentConfig().runtime), () => 'cloudEnvironmentUpdate'],
+      () => 'cloudEnvironmentCreate'
+    )
+    const { runtime: newRuntime, persistentDisk: newPersistentDisk } = this.getNewEnvironmentConfig()
+    const { runtime: oldRuntime, persistentDisk: oldPersistentDisk } = this.getOldEnvironmentConfig()
```
s-rubenstein, line 216: You can move this destructuring above the cond to use the destructuring in the cond.

src/components/NewClusterModal.js
```diff
@@ -211,27 +211,35 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
-      ...newRuntime,
-      cpu,
-      memory,
-      persistentDiskSize: newPersistentDisk?.size,
-      runtimeCostPerHour: Utils.formatUSD(runtimeConfigCost(this.getPendingRuntimeConfig())),
-      runtimePausedCostPerHour: Utils.formatUSD(ongoingCost(this.getPendingRuntimeConfig())),
-      persistentDiskCostPerMonth: (newPersistentDisk && Utils.formatUSD(persistentDiskCostMonthly(newPersistentDisk)))
+      ..._.mapKeys(key => `newRuntime_${key}`, newRuntime),
+      newRuntime_exists: !!newRuntime,
+      newRuntime_cpus: newRuntime && newRuntimeCpus,
```
s-rubenstein, line 231: Will these populate in mixpanel as undefined in delete events?

src/components/NewClusterModal.js
```diff
@@ -211,27 +211,35 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
+      newRuntime_memory: newRuntime && newRuntimeMemory,
+      newRuntime_costPerHour: newRuntime && runtimeConfigCost(this.getPendingRuntimeConfig()),
+      newRuntime_pausedCostPerHour: newRuntime && ongoingCost(this.getPendingRuntimeConfig()),
+      ..._.mapKeys(key => `oldRuntime_${key}`, oldRuntime),
+      oldRuntime_exists: !!oldRuntime,
+      oldRuntime_cpus: oldRuntime && oldRuntimeCpus,
```
s-rubenstein, line 237: Will these populate in mixpanel as undefined in create events?

src/components/NewClusterModal.js
```diff
@@ -594,7 +593,13 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
           return h(ButtonPrimary, { ...commonButtonProps, onClick: () => this.setState({ viewMode: 'environmentWarning' }) }, ['Next'])
         }],
         () => {
-          return h(ButtonPrimary, { ...commonButtonProps, onClick: () => this.applyChanges() }, [
+          return h(ButtonPrimary, {
+            ...commonButtonProps,
+            onClick: () => {
+              this.applyChanges()
+              this.sendCloudEnvironmentMetrics()
```
s-rubenstein: I might add this function to apply changes instead of here. Nothing is stopping us from calling applyChanges in other places, and I think it would be more correct to guard against it by putting this there.

Also, it guards against the chance of `applyChanges` erroring.

Shakespeared: Hmm yeah it was in applyChanges originally but broken out into the different cases. But I guess if it were just above the if blocks then that would work. A counter-argument is that most of the other mixpanel calls are in onClick, but I am not sure that's a good enough argument to keep it there.

src/components/NewClusterModal.js
Shakespeared: No, if it's undefined it just won't populate in mixpanel at all, so the whole field won't show up in the mixpanel event. I think that's an acceptable action, rather than forcing everything to be false or something if it's not populated.

src/components/NewClusterModal.js
Shakespeared: I think you're right, moved and pushed a new commit!

s-rubenstein (approved):
LGTM

PR #2212, opened 2020-08-27
Disable cloud environment debug panel by default

breilly2:
Also add a period to a sentence as talked about during a demo.

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -529,7 +529,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
         ]),
         p({ style: { marginBottom: 0 } }, [
           'You will continue to incur persistent disk cost at ',
-          span({ style: { fontWeight: 600 } }, [Utils.formatUSD(persistentDiskCostMonthly(this.getCurrentPersistentDisk())), ' per month'])
+          span({ style: { fontWeight: 600 } }, [Utils.formatUSD(persistentDiskCostMonthly(this.getCurrentPersistentDisk())), ' per month.'])
         ])
       ]),
       h(FancyRadio, {
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -24,6 +24,9 @@ import * as Utils from 'src/libs/utils'
 import validate from 'validate.js'
 
 
+// Change to true to enable a debugging panel (intended for dev mode only)
+const showDebugPanel = false
+
 const styles = {
   label: { fontWeight: 600, whiteSpace: 'pre' },
   titleBar: { marginBottom: '1rem' },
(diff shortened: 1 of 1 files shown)


panentheos (commented):
I think we should delete the code, rather than just leaving it inactive. We can always restore it from history if that particular implementation would be useful again.

breilly2:
I disagree. This feature has complex interactions that need to be considered when making any changes. Just the existence of this debugging aid contains a lot of embedded knowledge of the complexity of the problem and the kinds of things we have to consider when working with this code. I think the effort required to dig these up from git history will be sufficiently high such as to prevent it from happening and that will inevitably lead to less thorough testing and more time spent debugging. I might feel differently if we had some documentation in the form of executable checks or something in storybook. I do not want to lose all non-production artifacts of how this all works and how we got here (especially in the context of decentralized ownership of the code).

panentheos:
That's fair. Breaking this down, there are two potential things of value here: 1. The implementation of the debug panel itself, e.g. the convenience offered by that UX, and 2. The contents of this particular debug panel, which communicate the idea that 'Looking at the old and new environment configs is a good way to understand what the form is doing". We could think about the most effective way to capture each of those.

If we like the debug panel as a mechanism, maybe we should turn it into a generic tool that we could use in other situations too. (NB: There's also the built-in `useDebugValue` hook, which is only available in hook-based components, but might serve a similar purpose). As for communicating a high-level understanding of the form, we should definitely leave some sort of documentation of that, whether in the debug panel or in a block comment, or both.

src/components/NewClusterModal.js
```diff
@@ -141,6 +141,8 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
     const currentPersistentDisk = this.getCurrentPersistentDisk()
 
     this.state = {
+      // Uncomment this to enable the modal state debugger pane. Leave undefined for production.
+      // showDebugger: false,
```
s-rubenstein: TOL: I don't personally tend to like undefined as a trick for this, because it doesn't document itself as well in my opinion. If we want to start using this debug mode as a pattern, I might rather have a `debugModeEnabled` flag in the config (or something) so we can enable debug mode for local without enabling for prod. I think that also reduces the risk of the debug code accidentally hitting production by mistake.

That also has the advantage of being one switch for all our debugger tools, as opposed to potentially multiple switches.

breilly2: Yeah, I was definitely taking the shortcut of using `undefined` as a special value instead of adding another distinct piece of state. Seems like there's a bit more to explore here to decide how we want to organize our debugging utilities.

s-rubenstein: I am okay with doing this as a temporary step, but would want there to be a plan for moving to the new method of state. Do you have thoughts on how you would want to move forward on that after this PR/feature is released?

panentheos:
Changed the trigger to a simple code flag for now. We can expand on that later if desired. Also cleaned up the display a bit.

PR #2213, opened 2020-08-27
use disk details call on opening modal to display updated disk size faster

Shakespeared:
co-authored by Sky

refactored getCurrentCluster() and getCurrentPersistentDisk() to use currentClusterDetails and currentPersistentDiskDetails from state instead, added call to getDisks when you open the runtime drawer so we can display the most current info available on disk size.

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -142,6 +142,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
 
     this.state = {
       loading: false,
+      currentPersistentDiskDetails: undefined,
       selectedPersistentDiskSize: currentPersistentDisk ? currentPersistentDisk.size : DEFAULT_DISK_SIZE,
       jupyterUserScriptUri: '', customEnvImage: '', viewMode: undefined,
       sparkMode: cloudService === cloudServices.GCE ? false : numberOfWorkers === 0 ? 'master' : 'cluster',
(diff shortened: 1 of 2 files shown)


s-rubenstein:
This is the best way we could find to solve https://broadworkbench.atlassian.net/browse/SATURN-1833, because it uses the more complete version fetched on opening the modal for all cases, rather than just what it is needed for (which is a consistency win)

panentheos (commented):
Is the issue that updates might not be reflected immediately in the 'list' call? If so, then this might not be quite sufficient, because the form state still gets initialized in the constructor from the 'list' call, so it might be out of date.

We used to have an artificial 5 seconds of extra busy time after an update to account for cases like this. It's not super elegant, but it might be a more straightforward fix for now, and we could ask the Leo team about providing something on their end to smooth this over.

breilly2:
I think the intent here is to use the details rather than the results of the list call to fill the form. This feels clean to me. If that's not what this PR is doing then we probably missed something in the implementation.

I agree that it would be ideal for leonardo to give us some status info so we can detect that a change is in progress. I really don't want to go back to the hack of refreshing 5 seconds after posting the change.

Shakespeared:
Thanks for clarifying Brian - I'm not sure that I adequately explained the reasoning behind this change when I made the PR description, but as I understand it, that is correct.

Shakespeared:
Renamed the PR so it more accurately reflects the purpose of the change we're making here

s-rubenstein (approved):
LGTM. I am inclined to put this into the mob-pd branch so it can be part of the holistic review. I think that it drives us closer towards consistency, which is a good thing. (Rather than some environmentConfig info derived from the clusters call, and some from the details call)

PR #2216, opened 2020-08-28
[SATURN-1834] Add fixed positioning to menu

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -773,7 +773,6 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
                   id,
                   isSearchable: false,
                   value: sparkMode,
-                  menuPosition: 'fixed',
                   onChange: ({ value }) => this.setState({ sparkMode: value }),
                   options: [
                     { value: false, label: 'Standard VM', isDisabled: requiresSpark },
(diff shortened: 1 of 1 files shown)


Shakespeared (approved):
LGTM, tested in the UI

src/components/NewClusterModal.js
```diff
@@ -773,6 +773,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
                   id,
                   isSearchable: false,
                   value: sparkMode,
+                  menuPosition: 'fixed',
```
breilly2: This introduces a different problem. If the pane is scrollable and you open the select, the options don't scroll with the pane. Is there another way we can fix this?

s-rubenstein: The only other way I could think of to do it was to put was `menuPlacement: 'top'`, but that's a change in behavior.

panentheos: We could add `flex: 'auto'` to the scrollable div to have it expand to fill the space, which should fix it.

s-rubenstein: I will try that!

s-rubenstein: That solves it, pushing up shortly.

PR #2218, opened 2020-09-03
Generic Button style fixes

ehanna4:
Applied super basic style changes just to match the generic look of the other buttons

### DIFF
--- a/src/pages/workspaces/workspace/Notebooks.js
+++ b/src/pages/workspaces/workspace/Notebooks.js
@@ -319,15 +319,14 @@ const Notebooks = _.flow(
       div({
         style: {
           margin: '0 2.5rem 2.5rem 0', display: 'flex',
-          height: 350, width: 200, flexDirection: 'column',
+          width: 200, flexDirection: 'column',
           fontSize: 16, lineHeight: '22px'
         }
       }, [
         h(Clickable, {
           style: {
             ...Style.elements.card.container,
-            flex: 1,
-            color: colors.accent()
+            color: colors.accent(), height: 125
           },
           onClick: () => this.setState({ creating: true }),
           disabled: !canWrite,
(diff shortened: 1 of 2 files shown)


src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -319,7 +319,7 @@ const Notebooks = _.flow(
       div({
         style: {
           margin: '0 2.5rem 2.5rem 0', display: 'flex',
-          height: 250, width: 200, flexDirection: 'column',
+          height: 350, width: 200, flexDirection: 'column',
```
s-rubenstein: This seems like an unrelated change and doesn't look like what we want?

ehanna4: This is the height of the outer div - when it's only 250 the new button can't be very big. I don't know that this is 100% the way we want to go for the final choice (i.e. a hardcoded outer div height) but for now it allows the buttons to be big enough to see text.

ehanna4: ^ or that's how it seems to me and testing seems to fit this theory, but I could be misinterpreting what's controlling the button height style?

s-rubenstein: That will change existing buttons though, it seems to stretch them a lot taller.
Screen Shot 2020-09-03 at 1 13 41 PM

ehanna4: ah hmm I was trying to keep the pattern as similar to the existing as possible however now I'm wondering if this even should be using flex and maybe each button should just be setting their own height

ehanna4: or i could just add a max-height to the buttons for now?

s-rubenstein (approved):
LGTM

PR #2220, opened 2020-09-03
Initial working deleteApp code

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/libs/cluster-utils.js
+++ b/src/libs/cluster-utils.js
@@ -82,6 +82,17 @@ export const trimClustersOldestFirst = _.flow(
 
 export const currentCluster = _.flow(trimClustersOldestFirst, _.last)
 
+export const trimAppsOldestFirst = _.flow(
+  _.remove({ status: 'Deleting' }),
+  _.sortBy('createdDate')
+)
+
+export const currentApp = _.flow(trimAppsOldestFirst, _.last)
+
+export const appIsProvisioning = app => {
+  return app && app.status === 'PROVISIONING'
+}
+
 export const collapsedClusterStatus = cluster => {
   return cluster && (cluster.patchInProgress ? 'LeoReconfiguring' : cluster.status) // NOTE: preserves null vs undefined
 }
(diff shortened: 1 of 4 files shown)


src/libs/cluster-utils.js
```diff
@@ -82,6 +82,17 @@ export const trimClustersOldestFirst = _.flow(
 
 export const currentCluster = _.flow(trimClustersOldestFirst, _.last)
 
+export const trimAppsOldestFirst = _.flow(
```
s-rubenstein, line 85: I don't like the fact that this is duplicated from cluster logic, but don't want to rely on them being equivalent forever.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -176,6 +176,36 @@ const useClusterPolling = namespace => {
   return { clusters, refreshClusters }
 }
 
+const useAppPolling = namespace => {
```
s-rubenstein, line 179: Similarly don't like that this is duplicated from cluster code, but don't want to expect them to be identical forever.

src/libs/cluster-utils.js
panentheos: In fact the implementation of `trimClustersOldestFirst` has a bug, which is being fixed on the PD branch, so these will end up being different shortly. Even if that wasn't the case, this definitely looks like 'incidental equivalence' as opposed to 'intentional equivalence', so keeping them separate is a good way to communicate that.

src/pages/workspaces/workspace/WorkspaceContainer.js
panentheos: This one is interesting. There's potentially an abstraction here, along the lines of 'poll some resource, with variable interval, and return the data and a way to request a refresh'. It would have a fair number of parameters, though, so I'm not sure how nice it would look.

s-rubenstein: Yeah, this is the one that seems interesting. I know we changed useClusterPolling to include persistent disks in the PD work, so theoretically we could fold this in there as well, but while Clusters and PDs seem closely related, apps feel like their own resource and so probably shouldn't refresh or poll in the same loop?

panentheos: Yep, I agree that the polling loops should be independent since the resources are.

src/libs/cluster-utils.js
s-rubenstein: Ah yes! Indeed!

ehanna4 (approved):
LGTM

PR #2222, opened 2020-09-08
Super Bare Bones create/open drawer

ehanna4:
Adding the NewGalaxyModal.js which will be the panel that opens for the Galaxy configurations (aka create or delete).

this is WIP for create logic and is mostly for getting this file out there to add the NewGalaxyModal which will be used for the delete logic too.

### DIFF
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -1,0 +1,201 @@
+import _ from 'lodash/fp'
+import PropTypes from 'prop-types'
+import { Component, Fragment } from 'react'
+import { div, fieldset, h, label, legend, span } from 'react-hyperscript-helpers'
+import { ButtonPrimary, ButtonSecondary, IdContainer, Select } from 'src/components/common'
+import { NumberInput } from 'src/components/input'
+import { withModalDrawer } from 'src/components/ModalDrawer'
+import { InfoBox } from 'src/components/PopupTrigger'
+import TitleBar from 'src/components/TitleBar'
+import { profiles } from 'src/data/machines'
+import { formatRuntimeConfig, normalizeRuntimeConfig, runtimeConfigCost } from 'src/libs/cluster-utils'
+import colors from 'src/libs/colors'
+import * as Utils from 'src/libs/utils'
+
+
+const styles = {
+  row: {
+    display: 'flex',
+    alignItems: 'center',
+    marginTop: '1rem'
+  },
+  label: { fontWeight: 600, whiteSpace: 'pre' },
+  disabledInputs: {
+    border: `1px solid ${colors.dark(0.2)}`, borderRadius: 4, padding: '0.5rem'
+  }
+}
+
+export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewClusterModal extends Component {
+  static propTypes = {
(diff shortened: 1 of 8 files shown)


src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,119 @@
...
+  disabledInputs: {
+    border: `1px solid ${colors.dark(0.2)}`, borderRadius: 4, padding: '0.5rem'
+  },
+  whiteBoxContainer: { padding: '1rem', borderRadius: 3, backgroundColor: 'white' },
+  drawerContent: { display: 'flex', flexDirection: 'column', flex: 1, padding: '1.5rem' }
+}
+
+export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewClusterModal extends Component {
+  static propTypes = {
+    currentCluster: PropTypes.object,
```
ehanna4: a lot of this is copied from the NewClusterModal so I need to clean this section up and remove some unused stuff

ehanna4:
Clicking the create button here will open the side panel - the 'create' button there is disabled as I haven't finished adding the logic for actual creation

src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,91 @@
...
+          h(ButtonSecondary, { style: { margin: '0 2rem 0 1rem' }, onClick: onDismiss }, 'Cancel')
+        ])
+      ])
+    }
+
+    const getMachineDetails = () => {
+      return _.filter(({ name }) => name === 'n1-standard-8', machineTypes)[0]
+    }
+
+    return h(Fragment, [
```
s-rubenstein: I might go ahead and create the viewmode thing we did for NewClusterModal, just to follow the same patterns. (We can also do that when we add the second view mode for delete/confirmations)

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -294,12 +304,8 @@ const Notebooks = _.flow(
       Ajax().Apps.app(app.googleProject, app.appName).delete()
     }
 
-    const createGalaxyInstance = () => {
-
-    }
-
     const applyGalaxyChanges = () => {
-      return app ? deleteGalaxyInstance() : createGalaxyInstance()
+      return app ? deleteGalaxyInstance() : this.setState({ creatingGalaxy: true })
```
s-rubenstein, line 315: I wonder if we should make this state variable `modifyingGalaxy` or something? Since the panel will be used for more things?

s-rubenstein: Another option is we actually make separate components with some shared styling for different viewModes, which might be cleaner if we are doing it from the get-go.

s-rubenstein (commented):
This is looking good so far, a few preliminary comments.

src/components/NewGalaxyModal.js
ehanna4: I agree, I will make the update

src/pages/workspaces/workspace/Notebooks.js
s-rubenstein: Thinking more about it, I think we don't want separate components for the same reason we didn't do separate components for NewClusterModal.

PR #2223, opened 2020-09-09
Add martha_v3 endpoint support and integration test [WA-323]

rsasch:
- config option so that UI can use `martha_v2` or `martha_v3`; currently set to `martha_v2`
- logic to parse both shapes of responses
- integration test that will use an HCA URI for `martha_v2` and a TDR UIR for `martha_v3`

Closes https://broadworkbench.atlassian.net/browse/WA-323


config/alpha.json
```diff
@@ -12,6 +12,7 @@
   "jobManagerUrlRoot": "https://job-manager.dsde-alpha.broadinstitute.org/jobs",
   "leoUrlRoot": "https://leonardo.dsde-alpha.broadinstitute.org",
   "marthaUrlRoot": "https://us-central1-broad-dsde-alpha.cloudfunctions.net",
+  "metadataEndpoint": "martha_v2",
```
aednichols: Not loving this name collision

If we insist on having metadata in the name (perhaps because we plan to have non-metadata Martha endpoints in the future?) we can do
```suggestion
  "marthaMetadataEndpoint": "martha_v2",
```

aednichols (changes requested):

config/dev.json
```diff
@@ -11,6 +11,7 @@
   "jobManagerUrlRoot": "https://job-manager.dsde-dev.broadinstitute.org/jobs",
   "leoUrlRoot": "https://leonardo.dsde-dev.broadinstitute.org",
   "marthaUrlRoot": "https://us-central1-broad-dsde-dev.cloudfunctions.net",
+  "metadataEndpoint": "martha_v2",
```
panentheos: I'm curious why this should be a config option. Normally these are used for things that need to differ between environments, but these seem to all be the same. Is this being used like a feature flag, or for instrumentation? If that's the case, there are other mechanisms that might be a better fit for that.

rsasch: There are two different endpoints that provide metadata for DOS/DRS URIs, `martha_v2` and `martha_v3` (which supports the Terra Data Repo), and a separate endpoint that provides signed Urls (`getSignedUrlV1`).  We wanted to make it a configurable option so that we can roll back to `v2` if something is not working properly with `v3`, and the code needs to have a way for it to know which endpoint it's using because the responses come in different shapes and don't support all the same repos.  I would be happy to implement this in another way; this was the cleanest I could come up with.

config/alpha.json
rsasch: It has "metadata" in the name because Terra UI also calls the `getSignedUrlV1` Martha endpoint.

config/dev.json
aednichols: It's so that we can switch out `martha_v2` for `martha_v3` with just a config change and no code

A separation of concerns sort of thing

src/libs/ajax.js
```diff
@@ -1122,7 +1122,8 @@ const Dockstore = signal => ({
 
 const Martha = signal => ({
   getDataObjectMetadata: async url => {
-    const res = await fetchMartha('martha_v2', _.mergeAll([jsonBody({ url }), appIdentifier, { signal, method: 'POST' }]))
+    // TODO: switch to 'martha_v3' to point to updated endpoint
```
aednichols, line 1125: Is there a reason not to check in `martha_v3` today? It is in production in all environments.

aednichols (approved):

src/libs/ajax.js
rsasch: doing this to follow the model we've used for other codebases that are switching from `martha_v2` to `martha_v3`: first make sure code supporting both works, then do the switch-over

src/libs/ajax.js
```diff
@@ -1122,7 +1122,8 @@ const Dockstore = signal => ({
...
+    const res = await fetchMartha('martha_v2', _.mergeAll([jsonBody({ url }), authOpts(), appIdentifier, { signal, method: 'POST' }]))
```
panentheos, line 1126: This will send auth headers to martha, which will cause it to call bond. I don't think we want that (for v2 or v3), since we're not looking to use any bond SA from this call. (The SA gets looked up and used when generating a signed URL separately)

integration-tests/tests/preview-drs-uri.js
```diff
@@ -0,0 +1,46 @@
...
+
+  await click(page, clickable({ textContains: 'View Workspaces' }))
+  await fillIn(page, input({ placeholder: 'SEARCH WORKSPACES' }), workspaceName)
+  await click(page, clickable({ textContains: workspaceName }))
+
+  await click(page, navChild('data'))
+  await click(page, clickable({ textContains: testEntity.entityType }))
+  await click(page, elementInDataTableRow(testEntity.name, testEntity.attributes.file_uri))
+  await waitForNoSpinners(page)
+  await findText(page, 'View this file in the Google Cloud Storage Browser')
```
panentheos, line 38: TOL: This code looks fine, but a whole new integration test seems like a fairly high cost to pay for validating this one UI interaction. We should make sure that we're getting sufficient value in return.

src/libs/ajax.js
rsasch: `martha_v2` only calls Bond for certain providers (see https://github.com/broadinstitute/martha/blob/dev/common/bond.js#L44) and `martha_v3` behaves the same way; it is not based on whether auth is sent

integration-tests/tests/preview-drs-uri.js
rsasch: Also TOL: the test works (currently) on an HCA URI, which does not require Bond, so it coincidentally also verifies that `martha_v2` will not try to contact Bond (which would cause the request to fail) even if `auth` is sent.

src/libs/ajax.js
kshakir: Some of the new repos supported by `martha_v3` require `auth` to retrieve metadata. Instead of relying on clients like `terra-ui` to parse the URI and then figure out if the URI requires `auth` for metadata, `martha_v3` just always requires auth and then internally, based on the host, contacts bond.

There is an ongoing discussion regarding enabling "partial responses" by selecting only certain fields to include, such as `googleServiceAccount`. TOL: it would also be nice to perform the URL pre-signing in a single request, especially as Terra begins to support DOS/DRS servers that do not resolve to GCS. For example, the AnVIL returns `s3://` and `https://` but not `gs://`.

panentheos: FWIW from line https://github.com/broadinstitute/martha/blob/dev/common/bond.js#L45 the bond call is also conditional on whether the auth header is present. This may be a benign change, and in any case it's not intended to be  live for very long. (Thanks for the clarification on v3 needing auth, that makes sense)

Regarding 'partial responses', rather than trying to bundle multiple operations together and then provide additional mechanisms to separate them, we could instead provide smaller, single purpose building blocks and let downstream consumers compose them according to their needs. We ran into an issue previously with `fileSummaryV1` where we had to abandon it because it was trying to do too much, and didn't interact well with requester pays. Composing operations together at the point-of-use is much easier than taking a pre-bundled thing and trying to pull it apart.

panentheos (approved):
Looks good, tested locally.

PR #2225, opened 2020-09-10
reinitialize form state after making details calls

panentheos:
This addresses potential 'staleness' issues after loading runtime and disk details. The approach is as follows:
* First, when opening the modal, initialize the form with the data structures from props, e.g. the `list` call. This makes the form look mostly accurate while the spinner is up.
* Second, when the `details` calls return, re-initialize the form using the full details, which puts it into the fully accurate state. This should ensure that the form remains internally consistent, even if the `details` call returns different data than `list`.

Also restores the setting of `jupyterUserScriptUri` in `componentDidMount`, which fixes the issue with that field not populating.

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -12,10 +12,7 @@ import { InfoBox } from 'src/components/PopupTrigger'
 import TitleBar from 'src/components/TitleBar'
 import { cloudServices, machineTypes } from 'src/data/machines'
 import { Ajax } from 'src/libs/ajax'
-import {
-  currentCluster,
-  DEFAULT_DISK_SIZE, findMachineType, normalizeRuntimeConfig, ongoingCost, persistentDiskCostMonthly, runtimeConfigCost
-} from 'src/libs/cluster-utils'
+import { currentCluster, DEFAULT_DISK_SIZE, findMachineType, ongoingCost, persistentDiskCostMonthly, runtimeConfigCost } from 'src/libs/cluster-utils'
 import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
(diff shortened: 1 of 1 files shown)


src/components/NewClusterModal.js
```diff
@@ -139,24 +136,35 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
   constructor(props) {
     super(props)
     const currentCluster = this.getCurrentCluster()
```
breilly2, line 138: I notice that `this.getCurrentCluster()` and `this.getCurrentPersistentDisk()` are used for exactly 2 things:

1. Initializing the state backing the form elements (`constructor`)
2. Grabbing the keys used to fetch the details (`componentDidMount`)
(Yes, there is also a 3rd use of logging the event of opening the panel, but that is also done in `componentDidMount`.)

Is there a pattern we can use to make this more clear? Inlining the logic would make it more clear that the initialization is based on `props` while further operation of the form is based on fetched state, but would end up duplicating the code. Having a component-level function addresses the code duplication, but results in a function that reads like it could be used more broadly rather than just during initialization.

I just want to make sure that the dataflow is clear to future readers of this code.

Shakespeared (commented):
This seems reasonable to me, but I think someone else should review as well. I don't feel confident that I have the context (especially with the Leo images logic) to approve it.

breilly2 (approved):
👍  LGTM. I noted one place where I wonder if there could be an improvement to the representation of the dataflow in component initialization and operation. Not necessary to fix as part of this PR.

src/components/NewClusterModal.js
panentheos: Yeah, I think we should write a comment that describes the high-level data flow in the modal, to help people understand how the pieces fit together.

PR #2227, opened 2020-09-10
Add environment config, abstract render methods, fix flex

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -50,11 +50,16 @@ export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewGalaxyMod
     //todo work with leo team
   }
 
+  deleteGalaxy() {
+    const { app: oldApp } = this.getOldEnvironmentConfig()
+    return Ajax().Apps.app(oldApp.googleProject, oldApp.appName).delete()
+  }
+
   applyGalaxyChanges() {
     const { viewMode } = this.state
     const { app: oldApp } = this.getOldEnvironmentConfig()
     return Utils.switchCase(viewMode,
-      ['deleteWarn', () => Ajax().Apps.app(oldApp.googleProject, oldApp.appName).delete()],
+      ['deleteWarn', () => this.deleteGalaxy()],
       ['createWarn', () => this.createGalaxy()],
       [Utils.DEFAULT, () => !!oldApp ? this.setState({ viewMode: 'deleteWarn' }) : this.setState({ viewMode: 'createWarn' })]
     )
@@ -132,6 +137,7 @@ export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewGalaxyMod
               ]),
               h(Link, {
                 ...Utils.newTabLinkProps,
+                // TODO: Get the link from comms for this
                 href: ''
               }, ['Learn more about Galaxy interactive environments.'])
             ])
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -7,13 +7,14 @@ import { withModalDrawer } from 'src/components/ModalDrawer'
 import TitleBar from 'src/components/TitleBar'
 import { machineTypes } from 'src/data/machines'
 import { Ajax } from 'src/libs/ajax'
+import { currentApp } from 'src/libs/cluster-utils'
 import colors from 'src/libs/colors'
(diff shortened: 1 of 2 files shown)


src/components/NewGalaxyModal.js
```diff
@@ -7,29 +7,40 @@ import { withModalDrawer } from 'src/components/ModalDrawer'
...
 import { Ajax } from 'src/libs/ajax'
+import { currentApp } from 'src/libs/cluster-utils'
 import colors from 'src/libs/colors'
 import * as Utils from 'src/libs/utils'
 
 
 const styles = {
   whiteBoxContainer: { padding: '1rem', borderRadius: 3, backgroundColor: 'white' },
-  drawerContent: { display: 'flex', flexDirection: 'column', flex: 1, padding: '1.5rem' }
+  drawerContent: { display: 'flex', flexDirection: 'column', flex: '0 1 0%', padding: '1.5rem' }
```
s-rubenstein, line 17: Flex-grow 0 is to fix the buttons being all the way at the bottom of the page (per the mocks). If that's wrong, I can definitely revert this @ehanna4

src/components/NewGalaxyModal.js
```diff
@@ -7,29 +7,40 @@ import { withModalDrawer } from 'src/components/ModalDrawer'
...
 }
 
 export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewGalaxyModal extends Component {
   static propTypes = {
     onDismiss: PropTypes.func.isRequired,
     onSuccess: PropTypes.func.isRequired,
     namespace: PropTypes.string.isRequired,
-    app: PropTypes.object
+    apps: PropTypes.array
```
s-rubenstein, line 25: We started pulling in all the runtimes in the PD work for filtering, I think this does a bit of future proofing against that need here, and matches the new pattern

src/components/NewGalaxyModal.js
```diff
@@ -39,32 +50,29 @@ export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewGalaxyMod
     //todo work with leo team
   }
 
+  applyGalaxyChanges() {
+    const { viewMode } = this.state
+    const { app: oldApp } = this.getOldEnvironmentConfig()
+    return Utils.switchCase(viewMode,
+      ['deleteWarn', () => Ajax().Apps.app(oldApp.googleProject, oldApp.appName).delete()],
+      ['createWarn', () => this.createGalaxy()],
+      [Utils.DEFAULT, () => !!oldApp ? this.setState({ viewMode: 'deleteWarn' }) : this.setState({ viewMode: 'createWarn' })]
```
s-rubenstein, line 64: It may be worth pulling out the state changes into its own button component (similar to the new NewClusterModal), but this seems fine for now.

src/components/NewGalaxyModal.js
ehanna4: the change looks good in the UI to me

src/components/NewGalaxyModal.js
ehanna4: smart thinking!

src/components/NewGalaxyModal.js
ehanna4: I feel like we should be consistent with the create and delete and either have both ajax calls right in here or in their own functions. I think my preference is to put them in their own functions for quick readability and possibly error handling but could be swayed

src/components/NewGalaxyModal.js
```diff
@@ -74,61 +82,95 @@ export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewGalaxyMod
...
+                    [`${getMachineDetails().cpu} CPUS, ${getMachineDetails().memory}  GB of memory, 50 GB disk space`])
+                  //TODO: Define the disk space using DEFAULT_DISK_SIZE from the mob_pd branch
+                ]),
+                li({ style: { marginTop: '1rem' } }, [
+                  'Running cloud compute costs $0.19 per hr'
+                ])
+              ]),
+              h(Link, {
+                ...Utils.newTabLinkProps,
+                href: ''
```
ehanna4, line 141: just so it isn't forgotten could you throw a todo here to get the link for this? (i forgot to do that in my PR)

ehanna4 (approved):
Other than a few small questions/comments this looks good to me!

PR #2228, opened 2020-09-10
surface runtime and disk names on management page

panentheos:
Makes it possible to associate runtimes and disks on the 'cloud environments' page. It's not the most elegant solution, but this provides enough information to make it possible.


breilly2 (approved):
👍   This works. It's certainly not easy to recognize the connection, but any way that I can think of to improve it (without a major page overhaul) likely has problems with accessibility. I checked out the branch and tested locally.

PR #2231, opened 2020-09-11
Button always opens galaxy panel

ehanna4:
Enable create button and make it so the square button opens the panel/drawer/config for delete or create

### DIFF
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -30,7 +30,6 @@ export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewGalaxyMod
 
   constructor(props) {
     super(props)
-    const { apps } = props
 
     this.state = {
       viewMode: undefined,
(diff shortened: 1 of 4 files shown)


src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -477,12 +469,18 @@ const Notebooks = _.flow(
...
+          !!openGalaxyConfigDrawer && h(NewGalaxyModal, {
+            isOpen: openGalaxyConfigDrawer,
             namespace,
-            app: apps,
-            onDismiss: () => this.setState({ creatingGalaxy: false }),
-            onSuccess: () => { this.setState({ creatingGalaxy: false }) }
+            apps,
+            onDismiss: () => {
+              this.setState({ openGalaxyConfigDrawer: false })
+              this.refresh()
```
ehanna4: This should have been in the onSuccess() however even when there it doesn't appear to be updating correctly. Need to revisit this and how to update the button once the creation has begun

src/components/NewGalaxyModal.js
```diff
@@ -30,6 +30,8 @@ export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewGalaxyMod
 
   constructor(props) {
     super(props)
+    const { apps } = props
```
s-rubenstein: Do we need this here? Or is this more so we have a list of props here? (I'm worried it will fire unused variable errors)

src/libs/utils.js
```diff
@@ -185,7 +185,7 @@ export const abandonedPromise = () => {
 
 export const generateClusterName = () => `saturn-${uuid()}`
 
-export const generateKubernetesClusterName = () => `saturn-${uuid()}`
+export const generateKubernetesClusterName = () => `k8-${uuid()}`.slice(0, 30)
```
s-rubenstein: Since we are slicing it, should we use `saturn-k8-${uuid()}`

s-rubenstein (approved):
LGTM

PR #2234, opened 2020-09-14
make the switch from martha_v2 endpoint to martha_v3 [WA-349]

rsasch:
Once support for martha_v3  (#2223) has been merged and deployed and there are no issues, flip the switch to point to the `martha_v3` endpoint and change the integration test to use URI for Jade's Terra Data Repo.

Closes https://broadworkbench.atlassian.net/browse/WA-349


breilly2 (approved):
👍  Code looks good. As for testing, I'm relying on the integration test to verify the Jade Datarepo case. I also attempted to test though Gen3. Everything seems to be working on our side, including generating a signed URL to the file. However, I got an "Access Denied" error when trying to read the file. The behavior is the same on both the current dev build and this branch, so I believe this code is not introducing a new problem. Instead, it looks like possibly a permission problem on the Gen3 side, possibly only in staging. To be safe, we should perform a check in production after release to make sure everything works as expected.

breilly2:
I performed a pre-test from Gen3 in production. It worked to the point of giving me an error that I didn't specify a project for accessing a requester-pays bucket.

breilly2:
Post-release testing with Gen3 demonstrated the same behavior as pre-release, so I believe that is in good shape. I don't know how to test the Jade Datarepo integration so I have not done that.

PR #2235, opened 2020-09-14
Add Delete & Create for Galaxy [Saturn-1803] [SATURN-1805]

ehanna4:
This is the first Galaxy PR. Adding the button in the Notebooks tab which will be used for the 'Create' and 'Delete' flows. There is a new 'newGalaxyModal' file which closely follows the feel/look/style of the 'newClusterModal' for these flows. With this PR a user should be able to create a Galaxy App and Delete a Galaxy App when the feature flag is enabled.

Need to test! Don't merge yet
Also before merging ensure all feature flags are set to 'false'


src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,193 @@
...
+                  'Galaxy version xxx'
+                ]),
+                li({ style: { marginTop: '1rem' } }, [
+                  'Cloud Compute size of ', span({ style: { fontWeight: 600 } },
+                    [`${getMachineDetails().cpu} CPUS, ${getMachineDetails().memory}  GB of memory, 50 GB disk space`])
+                  //TODO: Define the disk space using DEFAULT_DISK_SIZE from the mob_pd branch
+                ]),
+                li({ style: { marginTop: '1rem' } }, [
+                  'Running cloud compute costs ',
+                  span({ style: { fontWeight: 600 } }, '$0.19 per hr')
```
ehanna4: this is the cost that was in the mocks - no calcualtion done

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -232,6 +250,15 @@ const Notebooks = _.flow(
     this.setState({ notebooks: _.reverse(_.sortBy('updated', notebooks)) })
   })
 
+  refreshApps = _.flow(
```
ehanna4, line 245: the above refresh should be called refreshNotebooks to match this name convention

src/components/NewGalaxyModal.js
s-rubenstein: I know we want things to be constant, but I don't know if we know for certain that this number is correct. However, we don't need that info in this PR. Ticket for cost info is here: https://broadworkbench.atlassian.net/browse/SATURN-1843

ehanna4: I've added a todo comment and changed this to '0.00' so that it is more obvious this will need to be changed.

config/dev.json
```diff
@@ -21,5 +21,6 @@
     "appId": "saturnnonprod-O9lUP",
     "apiKey": "AQEBBAEkx4iE2KxNyI7Wx08EwU1ycTM7E4FMSmaibbMUQxNU6uQvuAJt7fyABAtFYSYfgEE"
   },
-  "tosUrlRoot": "https://us-central1-broad-workbench-tos-dev.cloudfunctions.net/tos"
+  "tosUrlRoot": "https://us-central1-broad-workbench-tos-dev.cloudfunctions.net/tos",
+  "enableGalaxy": true
```
panentheos: TOL: Rather than adding a new entry here, we could just make it conditional on the existing `isProd` key.

src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,194 @@
...
+import * as Utils from 'src/libs/utils'
+
+
+const styles = {
+  whiteBoxContainer: { padding: '1rem', borderRadius: 3, backgroundColor: 'white' },
+  drawerContent: { display: 'flex', flexDirection: 'column', flex: '0 1 0%', padding: '0.5rem 1.5rem' },
+  headerText: { fontSize: 16, fontWeight: 600 }
+}
+
+export const NewGalaxyModal = withModalDrawer({ width: 675 })(class NewGalaxyModal extends Component {
```
panentheos: Generally speaking, new components should be written as function components rather than classes, to take advantage of hooks.

src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,194 @@
...
+    namespace: PropTypes.string.isRequired,
+    apps: PropTypes.array
+  }
+
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      viewMode: undefined,
+      isDeleting: false
```
panentheos: This state looks unused, we could delete it.

src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+
+  getOldEnvironmentConfig() {
```
panentheos: TOL: The environmentConfig concept was useful in the notebook runtime case, because we needed to manage multiple resources together, and the action logic was complex. Here, things are much more straightforward, so this might be an unnecessary indirection. This code will continue to change over time, and the simpler we can make this now, the easier it will be to evolve later.

src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,194 @@
...
+    // Currently, we only care about the existence of an app, and no other data.
+    const { apps } = this.props
+    const app = currentApp(apps)
+    return {
+      app: !!app ? app : undefined
+    }
+  }
+
+  async createGalaxy() {
```
panentheos: This will need an error handling wrapper, in case the ajax call fails.

src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,194 @@
...
+    const { namespace, onSuccess } = this.props
+    await Ajax().Apps.app(namespace, Utils.generateKubernetesClusterName()).create(Utils.generatePersistentDiskName())
+    return onSuccess()
+  }
+
+  async deleteGalaxy() {
```
panentheos: This will need an error handling wrapper, in case the ajax call fails.

src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,194 @@
...
+      div({ style: { padding: '0.5rem 1.5rem 1.5rem', flexGrow: 1, display: 'flex', flexDirection: 'column' } }, [contents, renderBottomButtons()])
+    ])
+  }
+
+  renderActionButton() {
+    const { viewMode } = this.state
+    const { app: oldApp } = this.getOldEnvironmentConfig()
+
+    return h(ButtonPrimary, {
+      onClick: () => this.applyGalaxyChanges()
```
panentheos: Since both the labels and the actions differ among these cases, it might be clearer to treat them as different buttons, rather than trying to unify them.

src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,194 @@
...
+    const renderDeleteWarning = () => {
+      return div({ style: { lineHeight: '22px' } }, [
+        div({ style: { marginTop: '0.5rem' } }, [
+          'Deleting your Cloud Environment will also ',
+          span({ style: { fontWeight: 600 } }, ['delete any files on the associated hard disk']),
+          ' (e.g. results files). Double check that your workflow results were written to ',
+          'the data tab in the workspace before clicking “Delete”'
+        ]),
+        div({ style: { marginTop: '1rem' } }, ['Deleting your Cloud Environment will stop your ',
+          'running Galaxy application and all associated costs. You can create a new Cloud Environment ',
```
panentheos: Based on how Kubernetes works, it isn't necessarily true that stopping your app will stop all costs (that's only the case if you're the last/only app running on the cluster). It will be tricky to communicate the cost model here, but I think we should try.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -42,6 +46,14 @@ const notebookCardCommonStyles = listView => _.merge({ display: 'flex' },
     }
 )
 
+const galaxyCardStyle = () => {
```
panentheos: nit: Since this is only used in one place and is pretty small, we could inline it

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -232,6 +250,14 @@ const Notebooks = _.flow(
     this.setState({ notebooks: _.reverse(_.sortBy('updated', notebooks)) })
   })
 
+  refreshApps = _.flow(
+    withErrorReporting('Error loading Apps'),
+    Utils.withBusyState(v => this.setState({ loading: v }))
```
panentheos, line 247: TOL: This adds another independent 'process' that writes to the `loading` state. It's worth considering what could happen if the two different refresh operations happened at the same time, or in close proximity (the end of one refresh might clear the spinner before the other refresh is done). In this case, it's unlikely, and the consequences seem fairly benign, so it's probably ok, but good to think about.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -315,10 +360,23 @@ const Notebooks = _.flow(
             icon('plus-circle', { style: { marginTop: '0.5rem' }, size: 21 })
           ])
         ]),
+        getConfig().enableGalaxy && h(Fragment, [
+          div({ style: { height: 15 } }),
```
panentheos: nit: Could this just be a margin?

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -416,7 +474,21 @@ const Notebooks = _.flow(
...
+            }
+          }),
+          //todo if status is deleting consider not opening the panel
+          !!openGalaxyConfigDrawer && h(NewGalaxyModal, {
+            isOpen: openGalaxyConfigDrawer,
+            namespace,
+            apps,
+            onDismiss: () => {
+              this.setState({ openGalaxyConfigDrawer: false })
+              this.refreshApps()
```
panentheos: If it was dismissed without an action, we shouldn't need to refresh.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -315,10 +360,23 @@ const Notebooks = _.flow(
             icon('plus-circle', { style: { marginTop: '0.5rem' }, size: 21 })
           ])
         ]),
+        getConfig().enableGalaxy && h(Fragment, [
```
panentheos: This should also be conditional on being in the Anvil site skin, for now.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -176,6 +176,34 @@ const useClusterPolling = namespace => {
...
+  const reschedule = ms => {
+    clearTimeout(timeout.current)
+    timeout.current = setTimeout(refreshAppsSilently, ms)
+  }
+  const loadApps = async () => {
+    try {
+      const newApps = await Ajax(signal).Apps.list(namespace, { creator: getUser().email })
+      setApps(newApps)
+      const app = currentApp(newApps)
+      reschedule(_.includes((app && (app.patchInProgress ? 'LeoReconfiguring' : app.status)), ['PROVISIONING', 'RUNNING']) ? 10000 : 120000)
```
panentheos: Is `patchInProgress` used on apps? Also, I don't think we should use the short timeout for the `RUNNING` status.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -416,7 +474,21 @@ const Notebooks = _.flow(
             onDismiss: () => this.setState({ deletingNotebookName: undefined }),
             onSuccess: () => {
               this.setState({ deletingNotebookName: undefined })
-              this.refresh()
+              this.refreshNotebooks()
+            }
+          }),
+          //todo if status is deleting consider not opening the panel
+          !!openGalaxyConfigDrawer && h(NewGalaxyModal, {
```
panentheos: In order to make the fade animation work for drawers, we need to always render them, and control their visibility with `isOpen`

src/libs/cluster-utils.js
```diff
@@ -82,6 +82,21 @@ export const trimClustersOldestFirst = _.flow(
 
 export const currentCluster = _.flow(trimClustersOldestFirst, _.last)
 
+export const trimAppsOldestFirst = _.flow(
+  _.remove({ status: 'Deleting' }),
+  _.sortBy('createdDate')
+)
+
+export const currentApp = _.flow(trimAppsOldestFirst, _.last)
```
panentheos, line 85: I noticed it's possible to get into a situation where you have multiple apps. Should we have some fallback UX for messaging and fixing that? Perhaps adding something to the 'Cloud Environments' page?

src/components/NewGalaxyModal.js
```diff
@@ -0,0 +1,194 @@
...
+          'for Galaxy later, which will take 8-10 minutes.'])
+      ])
+    }
+
+    const renderDefaultCase = () => {
+      return h(Fragment, [
+        div(['Environment consists of an application and cloud compute.']),
```
panentheos: UX TOL: Looking over the various screens, I think we might be muddying the idea of 'status' vs 'intention' in a couple places. We should make sure it's clear to the user when we're communicating the current state of the world (status), vs giving the opportunity to change to that state (intention).

For example, it looks like this default view is acting as 'intention' if you don't have an app, but a 'status' if you do. Even if the info isn't wrong (at this point), it's still confusing because it's presented the same.

I know this is just an MVP and will change substantially, but I wanted to bring that up to make sure we have the right frame as we go forward.

src/libs/cluster-utils.js
```diff
@@ -82,6 +82,21 @@ export const trimClustersOldestFirst = _.flow(
...
+
+export const appIsProvisioning = app => {
+  return app && app.status === 'PROVISIONING'
+}
+
+export const appIsDeleting = app => {
+  return app && app.status === 'DELETING'
```
panentheos, line 92: Is the status `'DELETING'` or `'Deleting'`, as above?

src/components/NewGalaxyModal.js
s-rubenstein: I think it is similar to clusters where the user is not aware of anyone else spending money in the project. For all intents and purposes, if I, as a user, shut down my app, I am no longer spending money in the project. (Similar to clusters, even if there is an underlying expensive K8 cluster that sticks around until the last app is gone.)

We are going ahead and changing the `all associated costs` to `your associated costs` since that is more accurate/correct though.

src/pages/workspaces/workspace/Notebooks.js
s-rubenstein: For ease of testing and development, we are posing that this happens as part of SATURN-1808

src/libs/cluster-utils.js
s-rubenstein: That sounds good, but sounds like a follow up piece of work probably. @ehanna4 is filing a ticket for it.

src/components/NewGalaxyModal.js
Shakespeared: TOL: From my conversation with Brian on slack, it sounds like whoever is paying for the billing project that the galaxy instances are created from will be the one charged. I don't have a lot of experience with the billing side of things, but would it be possible to identify the 'owner' of a billing project and continue to display the cost to them even if their specific instance is deleted? Disclaimer: I don't know how billing projects are handled so that might not make sense. 

Or maybe in the cost breakdown we could just say something (in a tooltip maybe?) indicating that K8 costs are grouped by billing project and won't be fully stopped until the entire K8 cluster is stopped. 

More TOL: Do we know what happens if someone creates a cluster, then someone else in the billing project is using it, but the first person wants to delete it and stop incurring costs?

s-rubenstein: I don't believe anyone can use an app that another person has created. If the app is deleted, it will only delete the K8 cluster if there are no other apps in the project (I believe)

Basically apps are 1-1 with users, and clusters are 1-1 with projects

src/components/NewGalaxyModal.js
s-rubenstein: We did a brief change to have it be a ternary on the word `consists` to either be `will consist` or `consists`.

src/pages/workspaces/workspace/Notebooks.js
s-rubenstein: Is worth thinking about for sure. I think probably relatively benign, so I am inclined to punt on it for now, if that is okay with you?

src/components/NewGalaxyModal.js
Shakespeared: That makes sense. I'm not sure this is a high-value conversation right now, I'm just curious because I don't fully understand the pay structure for the K8 clusters yet. My question is basically: who pays for the K8 cluster if the person who is currently using it is not the same as the person who created it? (E.g. person A creates the first galaxy instance and starts the K8 cluster, person B creates the second galaxy instance, person A deletes their galaxy instance - who pays for the cluster then?) Is it just always the person who owns the billing project? Or maybe there's something I'm missing and that's a moot point…

s-rubenstein: No worries! The answer is that currently I believe everything is always paid for by the billing project/billing project owner, regardless of who is using what.

Shakespeared: That makes sense, thanks! :)

src/libs/cluster-utils.js
ehanna4: 'DELETING' is what swagger says. I'm not finding where 'deleting' occurs in the code though, which line did you see that?

src/libs/logos.js
```diff
@@ -63,3 +64,13 @@ export const topBarLogo = () => isTerra() ?
...
+
+export const versionTag = (version, styles) => b({
+  style: {
+    fontSize: 8, lineHeight: '9px',
+    color: 'white', backgroundColor: colors.primary(1.5),
+    padding: '3px 5px', verticalAlign: 'middle',
+    borderRadius: 2, textTransform: 'uppercase',
+    ...styles
+  }
+}, version)
```
breilly2: I think `version` should be in an array since it's expected to be a single string. This could also be upgraded to a true component that takes children to render instead.

src/libs/utils.js
```diff
@@ -185,6 +185,8 @@ export const abandonedPromise = () => {
 
 export const generateClusterName = () => `saturn-${uuid()}`
 
+export const generateKubernetesClusterName = () => `saturn-k8-${uuid()}`.slice(0, 25)
```
breilly2: Is there a shorter form of UUID we can use instead of arbitrarily chopping off the end of this one? Maybe we can remove the `-`s? I know we're unlikely to see any collisions... probably ever... but if someone does, it will be practically impossible to reproduce and really hard to recognize what happened.

s-rubenstein: Ah, this should actually be fine instead of chopping off. Will update

src/libs/logos.js
```diff
@@ -63,3 +64,13 @@ export const topBarLogo = () => isTerra() ?
   brandLogoMaker(50, true)
 
 export const footerLogo = () => isTerra() ? terraLogoMaker(terraLogoWhite, { height: 40 }) : brandLogoMaker(40)
+
+export const versionTag = (children, styles) => b({
```
breilly2: As written, this isn't quite a component; it's a function that takes 2 arguments and returns a component. To be a component, it should take a `props` object (and should start with an uppercase letter):

```
export const VersionTag = ({styles, children}) => b({
  ...
```

And then `children` below should not be in `[]` because it should already be an array.

Alternatively, if you want to keep it as a function instead of a component, keep the name as is and change it back to `version`; `children` to me implies that it's a component.

s-rubenstein: Good point, lets leave it as a function.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -286,6 +307,23 @@ const Notebooks = _.flow(
       }))
     )(notebooks)
 
+    const getGalaxyText = () => {
+      return app ?
+        div({ style: { fontSize: 18, lineHeight: '22px', width: 160 } }, [
+          div(['Galaxy Interactive']),
```
breilly2, line 313: For the benefit of screen readers, we should try to merge these `div`s, inserting linebreaks if we want particular formatting. (same with "Create a Cloud Environment for Galaxy" below)

src/libs/ajax.js
```diff
@@ -1113,6 +1113,36 @@ const Clusters = signal => ({
...
+      _.mergeAll([authOpts(), appIdentifier, { signal }]))
+    return res.json()
+  },
+  app: (project, name) => {
+    const root = `api/google/v1/apps/${project}/${name}`
+    return {
+      delete: () => {
+        return fetchLeo(root, _.mergeAll([authOpts(), { signal, method: 'DELETE' }, appIdentifier]))
+      },
+      create: (diskName, appType, namespace, bucketName, workspaceName) => {
```
panentheos: TOL: With a large number of parameters, it's sometimes nice to instead take a single object and destructure it, rather than using positional arguments. That way, the  caller doesn't have to worry about getting the order right.

src/libs/cluster-utils.js
```diff
@@ -79,7 +79,20 @@ export const currentCluster = clusters => {
   return !clusters ? undefined : (_.flow(trimClustersOldestFirst, _.last)(clusters) || null)
 }
 
-export const currentApp = _.first // TODO galaxy: fill in the correct logic here
+export const trimAppsOldestFirst = _.flow(
+  _.remove({ status: 'DELETING' }),
+  _.sortBy('createdDate')
```
panentheos: I don't see `createdDate` on the apps returned from Leo, so this probably isn't doing what we want. That said, I don't see any other field that would be suitable for sorting, either.

src/components/NewAppModal.js
```diff
@@ -0,0 +1,161 @@
...
+      div({ style: { marginTop: '1rem' } }, ['Deleting your Cloud Environment will stop your ',
+        'running Galaxy application and your application costs. You can create a new Cloud Environment ',
+        'for Galaxy later, which will take 8-10 minutes.'])
+    ])
+  }
+
+  const renderDefaultCase = () => {
+    return h(Fragment, [
+      div([`Environment ${app ? 'consists' : 'will consist'} of an application and cloud compute.`]),
+      div({ style: { paddingTop: '0.5rem' } }, [
```
panentheos: nit: This wrapper could be removed in favor of some additional margin on the inner div.

src/components/NewAppModal.js
```diff
@@ -0,0 +1,161 @@
...
+    h(TitleBar, {
+      title: Utils.switchCase(viewMode,
+        ['deleteWarn', () => 'Delete Cloud Environment for Galaxy'],
+        [Utils.DEFAULT, () => 'Cloud environment']
+      ),
+      onDismiss,
+      onPrevious: !!viewMode ? () => setViewMode(undefined) : undefined,
+      style: { margin: '2rem 0 0 1.5rem' }
+    }),
+    div({ style: { padding: '0.5rem 1.5rem 1.5rem', flexGrow: 1, display: 'flex', flexDirection: 'column' } }, [contents, renderBottomButtons()])
```
panentheos: nit: This wrapper and the `TitleBar` are just adding extra padding inside of what's being provided by `styles.drawerContent`. We could simplify the layout by combining the padding.

src/components/NewAppModal.js
```diff
@@ -0,0 +1,161 @@
...
+      [Utils.DEFAULT, () => {
+        return !!app ?
+          h(ButtonPrimary, { onClick: () => setViewMode('deleteWarn') }, ['Delete']) :
+          h(ButtonPrimary, { onClick: () => setViewMode('createWarn') }, ['Next'])
+      }]
+    )
+  }
+
+  const renderBottomButtons = () => {
+    return h(Fragment, [
```
panentheos: nit: This `Fragment` is unnecessary since there's only one child.

src/components/NewAppModal.js
```diff
@@ -0,0 +1,161 @@
...
+// TODO: Once we have more appTypes, abstract text/component to support other apps
+export const NewAppModal = _.flow(
+  Utils.withDisplayName('NewAppModal'),
+  withModalDrawer({ width: 675 })
+)(({ onDismiss, onSuccess, namespace, apps, bucketName, workspaceName }) => {
+  const [viewMode, setViewMode] = useState(undefined)
+
+  const app = currentApp(apps)
+
+  const createApp = withErrorReporting('Error creating app', async () => {
```
panentheos: We should probably drive a busy spinner with these calls, in case they take some time.

src/components/NewAppModal.js
```diff
@@ -0,0 +1,161 @@
...
+          div({ style: { flex: '1', lineHeight: '1.5rem', minWidth: 0, display: 'flex' } }, [
+            span({ style: { marginRight: '0.5rem', marginTop: '0.5rem' } }, [icon('info-circle', { size: 25, color: colors.accent() })]),
+            div([
+              span({ style: styles.headerText }, ['Set up duration']),
+              div({ style: { lineHeight: '24px' } }, [
+                div(['Creating a cloud environment for Galaxy takes ', span({ style: { fontWeight: 600 } }, ['8-10 minutes.'])]),
+                div(['You can navigate away, and we will notify you when it\'s ready. '])
+              ]),
+              div({ style: { ...styles.headerText, marginTop: '0.5rem' } }, ['Continuation cost']),
+              div({ style: { lineHeight: '18px' } }, [
```
panentheos: nit: These line-height values seem arbitrary and slightly different from one another. Can we standardize a bit, and/or consider using a unitless value (e.g. `1.5`) so it's relative to the text size?

src/components/NewAppModal.js
```diff
@@ -0,0 +1,161 @@
...
+
+  const getMachineDetails = () => {
+    return _.filter(({ name }) => name === 'n1-standard-8', machineTypes)[0]
+  }
+
+  const renderCreateWarning = () => {
+    return h(Fragment, [
+      div({ style: { marginBottom: '1rem' } }, ['Environment will consist of an application and cloud compute.']),
+      div({ style: { ...styles.whiteBoxContainer, backgroundColor: colors.accent(.1), boxShadow: Style.standardShadow } }, [
+        div({ style: { display: 'flex', padding: `0.5rem 0 0`, minHeight: 120 } }, [
```
panentheos: nit: This wrapper could be removed, the `min-height` isn't having any effect and the padding looks extraneous to me.

src/components/NewAppModal.js
```diff
@@ -0,0 +1,161 @@
...
+          div({ style: { flex: '1', lineHeight: '1.5rem', minWidth: 0, display: 'flex' } }, [
+            span({ style: { marginRight: '0.5rem', marginTop: '0.5rem' } }, [icon('info-circle', { size: 25, color: colors.accent() })]),
+            div([
+              span({ style: styles.headerText }, ['Set up duration']),
+              div({ style: { lineHeight: '24px' } }, [
+                div(['Creating a cloud environment for Galaxy takes ', span({ style: { fontWeight: 600 } }, ['8-10 minutes.'])]),
+                div(['You can navigate away, and we will notify you when it\'s ready. '])
+              ]),
+              div({ style: { ...styles.headerText, marginTop: '0.5rem' } }, ['Continuation cost']),
```
panentheos: TOL: This language seems unclear to me. What about something like 'Cost warning' or 'Cost note'?

src/pages/workspaces/workspace/Notebooks.js
ehanna4: I'm not seeing this be an issue with ANDI or AXE, how did you discover the screenreader had trouble with this?

src/components/NewAppModal.js
ehanna4: I believe they were done in that manner to try to match the mocks. However, in the name of consistency and standardized I made them just 1.5

src/components/NewAppModal.js
ehanna4: This is what's in the mocks, but I would rather not hold this PR on this. I will add a note in the cost ticket to follow up with Joy on this language

src/pages/workspaces/workspace/Notebooks.js
breilly2: Ah, I didn't exactly test it. I'm assuming that a screenreader will think that the separate divs are separate pieces of content. It would be good to verify that. Time to relearn VoiceOver. :\

src/libs/cluster-utils.js
ehanna4: Sky is going to reach out to Leo to see about getting createdDate added to the apps they return. In the meantime going to remove this and will file a follow up ticket.

src/components/NewAppModal.js
```diff
@@ -0,0 +1,168 @@
...
+  withModalDrawer({ width: 675 })
+)(({ onDismiss, onSuccess, namespace, apps, bucketName, workspaceName }) => {
+  const [viewMode, setViewMode] = useState(undefined)
+  const [loading, setLoading] = useState(false)
+
+
+  const app = currentApp(apps)
+
+  const createApp = _.flow(
+    Utils.withBusyState(() => setLoading(true)),
```
panentheos: `withBusyState` passes the busy value as an argument. If we hardcode `true` here, this won't clear the spinner if the call fails. Instead, because `setLoading` is already a function of one argument that does what we want, we can do:
```suggestion
    Utils.withBusyState(setLoading),
```

src/components/NewAppModal.js
```diff
@@ -0,0 +1,168 @@
...
+    withErrorReporting('Error creating app')
+  )(async () => {
+    await Ajax().Apps.app(namespace, Utils.generateKubernetesClusterName()).create({
+      diskName: Utils.generatePersistentDiskName(), appType: 'GALAXY', namespace, bucketName, workspaceName
+    })
+    return onSuccess()
+  })
+
+  const deleteApp = _.flow(
+    Utils.withBusyState(() => setLoading(true)),
```
panentheos: Same as above:
```suggestion
    Utils.withBusyState(setLoading),
```

src/libs/ajax.js
```diff
@@ -1113,6 +1113,36 @@ const Clusters = signal => ({
   }
 })
 
+const Apps = signal => ({
+  list: async (project, labels = {}) => {
+    const res = await fetchLeo(`api/google/v1/apps/${project}/?${qs.stringify({ ...labels })}`,
```
panentheos: nit: Leo seems to accept this, but technically there should not be a `/` at the end here.
```suggestion
    const res = await fetchLeo(`api/google/v1/apps/${project}?${qs.stringify({ ...labels })}`,
```

panentheos (approved):
See a couple comments, then looks good. Tested create + delete locally.

breilly2 (commented):
I haven't done a thorough review, but I have left some single comments and done some testing.

One thing my testing revealed is that we should be deleting PDs when deleting apps because leonardo doesn't currently support re-attaching disks to apps (and is not planning to for alpha). Another thing I learned is that PDs are not sharable between Galaxy and Jupyter/RStudio so I think we should tag Galaxy disks and filter them out for non-Galaxy use.

ehanna4:
@breilly2 i will update the delete call to specify deleting the PD as well. Do you think the labels/filtering needs to be done in this PR or can that be a followup ?

breilly2 (commented):
Another bit of testing I want to call out: After creating a Galaxy app and deleting, I tested creating another app. I did this twice: once after waiting for the previous app deletion to complete and once while the delete was still happening in leonardo. Both cases worked.

On the second test, I noticed a state of `PREDELETING` which I think we should add to the condition for more frequent polling.

PR #2236, opened 2020-09-14
change PD copy [SATURN-1847]

panentheos:
See ticket for details

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -1,7 +1,7 @@
 import _ from 'lodash/fp'
 import PropTypes from 'prop-types'
 import { Component, Fragment } from 'react'
-import { b, div, fieldset, h, input, label, legend, li, p, span, ul } from 'react-hyperscript-helpers'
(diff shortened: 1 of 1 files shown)


src/components/NewClusterModal.js
```diff
@@ -1103,6 +1107,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
           onPrevious: () => this.setState({ viewMode: undefined })
         }),
         div({ style: { lineHeight: 1.5 } }, [
+          p(['Your persistent disk is mounted in the directory ', code({ style: { fontWeight: 600 } }, ['/home/jupyter-user/notebooks']), '. Please save your analysis data in this directory to ensure it’s stored on your disk.']),
```
ehanna4: in the mocks the `/home/jupyter-user/notebooks` got quotes and a new line, should that be the case here too?

panentheos: I added the line break. The quotes were copied directly from the Adrian's original proposal, but I think the monospace code font and font-weight do a better job of delineating this (and also making copy-pasting less error prone).

ehanna4 (approved):
LGTM

PR #2238, opened 2020-09-14
Fortify run-notebook integration test

breilly2:
Based on discussion in mob-pd PR, this fortifies how the run-notebook integration test interacts with the cloud environment modal drawer, along with some explanation about why we need to do it this way.


panentheos (approved):
Looks good, thanks for the writeup!

PR #2239, opened 2020-09-17
remove storybook [SATURN-1849]

panentheos:
See ticket for details


breilly2 (approved):
👍  LGTM. I didn't test because... well... there's really nothing to test, other than the build still working which the PR build already did.

PR #2240, opened 2020-09-17
implement galaxy status in cluster manager [SATURN-1804]

panentheos:
This adds galaxy information to the 'cluster manager' widget in the upper right (if galaxy is running). By itself, this PR doesn't have any visible changes, since the actual data isn't wired in yet. Tested locally by passing some stub data to the cluster manager.

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -277,7 +277,7 @@ export default class ClusterManager extends PureComponent {
       app && h(Clickable, {
         style: { display: 'flex', marginRight: '2rem' },
         onClick: () => {
-          // TODO galaxy: implement this action
+          // TODO galaxy SATURN-1855: implement this action
         }
       }, [
         img({ src: galaxyLogo, alt: '', style: { marginRight: '0.25rem' } }),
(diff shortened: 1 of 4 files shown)


src/components/ClusterManager.js
```diff
@@ -270,7 +271,21 @@ export default class ClusterManager extends PureComponent {
     const appName = isRStudioImage ? 'RStudio' : 'terminal'
     const appLaunchLink = Nav.getLink('workspace-app-launch', { namespace, name, app: appName })
 
+    const app = currentApp(apps)
+
     return div({ style: styles.container }, [
+      app && h(Clickable, {
+        style: { display: 'flex', marginRight: '2rem' },
+        onClick: () => {
+          // TODO galaxy: implement this action
```
s-rubenstein: Could we put a ticket number here? I think we want this to launch the galaxy cluster, so maybe TODO (SATURN-1855)? (And then add it as part of the AC to that ticket?)

s-rubenstein (approved):
This LGTM, although I don't know how comfortable I feel testing with stub data, given that we've had some inconsistency issues with Leo's swagger needing to change, but I feel confident we can fix in a follow-up.

PR #2242, opened 2020-09-21
Use GCE with PD instead of Dataproc for run-notebook test [SATURN-1859]

breilly2:
Small change to test as discussed over email with Brett, Amelia, Ambica, and Adrian.


panentheos:
Looks good. In addition, we should modify the cleanup code to delete any PDs, so they don't stick around indefinitely.

breilly2:
@panentheos Good call. In dev, at least, Interactive Analysis has a script that sweeps these things up (at least on the GCP side), but it's not perfect and may be missing from other envs. I changed the clean-up to request deletion of the PD if there is one.

panentheos (approved):
Looks good.

PR #2243, opened 2020-09-22
show notification when galaxy app is ready [SATURN-1806]

panentheos:
Note that this has no effect on its own, since the `apps` aren't wired up yet. Tested locally by passing stub data.

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -125,7 +125,7 @@ export default class ClusterManager extends PureComponent {
   }
 
   componentDidUpdate(prevProps) {
-    const { namespace, name } = this.props
+    const { namespace, name, apps } = this.props
     const prevCluster = _.last(_.sortBy('auditInfo.createdDate', _.remove({ status: 'Deleting' }, prevProps.clusters))) || {}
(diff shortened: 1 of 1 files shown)


src/components/ClusterManager.js
```diff
@@ -165,6 +167,10 @@ export default class ClusterManager extends PureComponent {
     } else if (cluster.status === 'Running' && prevCluster.status === 'Updating') {
       notify('success', 'Number of workers has updated successfully.')
     }
+    if (prevApp && prevApp.status !== 'RUNNING' && app && app.status === 'RUNNING') {
```
s-rubenstein, line 170: Do we want to check `not RUNNING` or `is PROVISIONING`? (I don't think there's a functional difference right now, but it is a difference in behavior)

s-rubenstein (approved):
LGTM. Is this getting wired up with the same cleanup task that is in progress?

src/components/ClusterManager.js
panentheos: I think this best captures the intended logic, that we want to catch the transition to `RUNNING` status.

panentheos:
I put a note in the 'launch' ticket to handle the wire up.

PR #2244, opened 2020-09-22
[SATURN-1861] Add delete disk parameter

s-rubenstein:
I looked at the network tab when deleting and saw the parameter attached:
Screen Shot 2020-09-22 at 2 19 03 PM

### DIFF
--- a/src/components/NewAppModal.js
+++ b/src/components/NewAppModal.js
@@ -45,7 +45,7 @@ export const NewAppModal = _.flow(
     Utils.withBusyState(setLoading),
     withErrorReporting('Error deleting app')
   )(async () => {
-    await Ajax().Apps.app(app.googleProject, app.appName).delete(true)
+    await Ajax().Apps.app(app.googleProject, app.appName).delete()
     return onSuccess()
   })
 
--- a/src/components/NewAppModal.js
+++ b/src/components/NewAppModal.js
@@ -45,7 +45,7 @@ export const NewAppModal = _.flow(
     Utils.withBusyState(setLoading),
     withErrorReporting('Error deleting app')
   )(async () => {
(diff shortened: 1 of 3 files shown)


src/libs/ajax.js
```diff
@@ -1122,8 +1122,9 @@ const Apps = signal => ({
   app: (project, name) => {
     const root = `api/google/v1/apps/${project}/${name}`
     return {
-      delete: () => {
-        return fetchLeo(root, _.mergeAll([authOpts(), { signal, method: 'DELETE' }, appIdentifier]))
+      delete: deleteDisk => {
+        return fetchLeo(`${root}${qs.stringify({ deleteDisk }, { addQueryPrefix: true })}`,
```
panentheos: I don't see this option in the swagger, is it actually available?

src/components/NewAppModal.js
```diff
@@ -45,7 +45,7 @@ export const NewAppModal = _.flow(
     Utils.withBusyState(setLoading),
     withErrorReporting('Error deleting app')
   )(async () => {
-    await Ajax().Apps.app(app.googleProject, app.appName).delete()
+    await Ajax().Apps.app(app.googleProject, app.appName).delete(true)
```
panentheos: Since we always intend to send `true` for now, perhaps we should hard code that in the ajax call.

src/libs/ajax.js
s-rubenstein: Yes, there's discussion on slack, they missed adding it to swagger.

src/components/NewAppModal.js
s-rubenstein: Sure. I can do that. Although, actually, maybe a default value would be better?

panentheos (approved):
Looks good, tested locally. Note the config change before merging.

PR #2245, opened 2020-09-22
Wire apps from wrapWorkspace [SATURN-1863]

breilly2:
Wires `apps`, through `WorkspaceContainer`, to `ClusterManager`.

Also, I noticed a state of `PREDELETING` that I felt should be included in the check list for quicker app polling.


panentheos (approved):
Looks good, tested locally.

PR #2251, opened 2020-09-24
[SATURN-1862] Update disk labels hide galaxy disks

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -280,10 +280,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
             name: currentPersistentDiskDetails.name
           } : {
             name: Utils.generatePersistentDiskName(),
-            size: newPersistentDisk.size,
-            labels: {
-              createdByTool: 'runtime'
-            }
+            size: newPersistentDisk.size
           }
         })
       } : {
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -280,7 +280,8 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
             name: currentPersistentDiskDetails.name
           } : {
             name: Utils.generatePersistentDiskName(),
-            size: newPersistentDisk.size
+            size: newPersistentDisk.size,
+            saturnApplication: 'runtime'
           }
         })
       } : {
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -280,7 +280,10 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
             name: currentPersistentDiskDetails.name
           } : {
             name: Utils.generatePersistentDiskName(),
-            size: newPersistentDisk.size
+            size: newPersistentDisk.size,
+            labels: {
+              createdByTool: 'runtime'
+            }
           }
         })
       } : {
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -280,8 +280,7 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
             name: currentPersistentDiskDetails.name
           } : {
             name: Utils.generatePersistentDiskName(),
-            size: newPersistentDisk.size,
-            saturnApplication: 'runtime'
+            size: newPersistentDisk.size
           }
         })
       } : {
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1131,7 +1131,7 @@ const Apps = signal => ({
           diskConfig: {
             name: diskName,
             labels: {
-              hidden: 'true'
+              saturnApplication: 'galaxy'
             }
           },
           customEnvironmentVariables: {
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1129,7 +1129,10 @@ const Apps = signal => ({
       create: ({ diskName, appType, namespace, bucketName, workspaceName }) => {
         const body = {
           diskConfig: {
-            name: diskName
+            name: diskName,
+            labels: {
+              createdByTool: 'app'
+            }
           },
           customEnvironmentVariables: {
             WORKSPACE_NAME: workspaceName,
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1131,7 +1131,7 @@ const Apps = signal => ({
           diskConfig: {
             name: diskName,
             labels: {
-              createdByTool: 'app'
+              hidden: 'true'
             }
           },
           customEnvironmentVariables: {
--- a/src/pages/workspaces/workspace/WorkspaceContainer.js
+++ b/src/pages/workspaces/workspace/WorkspaceContainer.js
@@ -161,12 +161,13 @@ const useCloudEnvironmentPolling = namespace => {
   }
   const load = async () => {
     try {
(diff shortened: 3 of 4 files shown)


s-rubenstein:
This is my proposal on how to hide galaxy disks. I thought this might be more extensible in case there are other cases where we want to hide disks. (And eventually we may want to create a display to show those hidden disks, for cost management, but this is probably good for now)

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -161,12 +161,14 @@ const useCloudEnvironmentPolling = namespace => {
   }
   const load = async () => {
     try {
-      const [newDisks, newClusters] = await Promise.all([
+      const [newDisks, newClusters, hiddenDisks] = await Promise.all([
         Ajax(signal).Disks.list({ googleProject: namespace, creator: getUser().email }),
-        Ajax(signal).Clusters.list({ googleProject: namespace, creator: getUser().email })
+        Ajax(signal).Clusters.list({ googleProject: namespace, creator: getUser().email }),
+        Ajax(signal).Disks.list({ googleProject: namespace, creator: getUser().email, hidden: 'true' })
```
panentheos: TOL: We need an extra call here because the list endpoint doesn't return labels, right? It might be good to highlight this to the Leo team, and see about including that data so we don't have to make multiple calls.

src/libs/ajax.js
```diff
@@ -1129,7 +1129,10 @@ const Apps = signal => ({
       create: ({ diskName, appType, namespace, bucketName, workspaceName }) => {
         const body = {
           diskConfig: {
-            name: diskName
+            name: diskName,
+            labels: {
+              hidden: 'true'
```
panentheos: 'Hidden' is a behavior specific to the UI, but that behavior might change, while these labels become permanent server state, so we should be careful. I think it would be better to store information about what they _are_, then let the code map that to some behavior. E.g., something like `saturnApplication: 'galaxy'` (the 'saturn' prefix is a courtesy since an API user might want to add their own labels)

s-rubenstein: Sounds good, I kept going back and forth on this, so good to know that you prefer this way! Will change it!

src/pages/workspaces/workspace/WorkspaceContainer.js
s-rubenstein: They actually are aware of this! We were talking about it and this is the solution I ended up brainstorming. I can flag it again though!

s-rubenstein: Yeah, they have pretty reasonable reasons for not having it return labels: https://broadinstitute.slack.com/archives/CA3NP1733/p1601302225045000

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -161,12 +161,14 @@ const useCloudEnvironmentPolling = namespace => {
...
+      const [newDisks, newClusters, galaxyDisks] = await Promise.all([
         Ajax(signal).Disks.list({ googleProject: namespace, creator: getUser().email }),
-        Ajax(signal).Clusters.list({ googleProject: namespace, creator: getUser().email })
+        Ajax(signal).Clusters.list({ googleProject: namespace, creator: getUser().email }),
+        Ajax(signal).Disks.list({ googleProject: namespace, creator: getUser().email, saturnApplication: 'galaxy' })
       ])
+      const galaxyDiskNames = _.map(disk => disk.name, galaxyDisks)
       setClusters(newClusters)
-      setPersistentDisks(newDisks)
+      setPersistentDisks(_.remove(disk => _.includes(disk.name, galaxyDiskNames), newDisks))
```
breilly2, line 171: I haven't tested yet, but I think you need `_.filter` instead of `_.remove`. `_.remove` mutates the array it's given and returns the removed elements.

s-rubenstein: You are right, and I am not sure why it was working… /shrug

breilly2: Well... apparently it was working because... I guess lodash/fp changes the behavior of `_.remove`??? I compared the REPL in the lodash docs and the console of Terra UI and they do indeed behave differently. lodash/fp's `_.remove` apparently returns the new array with the elements removed and does not modify the given array... and the FP Guide does not make any mention of this that I can tell! Not gonna lie, I'm kinda mad right now...

Anyway, I still prefer using `_.filter` because it behaves _predictably_ and _as documented_ in lodash/fp.

breilly2: Coming back to this, I think I'm changing my mind. Despite the fact that the lodash FP Guide says nothing about this other than "immutable auto-curried iteratee-first data-last methods", this does seem to be an intentional behavior change in lodash/fp. Furthermore, we are using `_.remove` in several other places already. It's also more clear than having to negate the condition given to `_.filter`, so I think it's better to keep it as `_.remove`.

I regret the time and energy taken by this thread. I hope that it ultimately pays off in learning (though I have a feeling I've been down this road before myself 😞 ).

s-rubenstein: Absolutely. And yeah, it would be very nice if they had documented the changes at all. /ugh 

I will revert it.

src/components/NewClusterModal.js
```diff
@@ -280,7 +280,8 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
             name: currentPersistentDiskDetails.name
           } : {
             name: Utils.generatePersistentDiskName(),
-            size: newPersistentDisk.size
+            size: newPersistentDisk.size,
+            saturnApplication: 'runtime'
```
panentheos: Should this be `labels: { saturnApplication: 'runtime' }`?

Also TOL: I think it's fine to add this, but since there have already been PDs created without it, it will be difficult to use this value by itself to make decisions, unless we back-fill the data on the server side.

panentheos (approved):
See comment, then looks good. Tested locally by creating a galaxy cluster and a standard runtime.

src/components/NewClusterModal.js
s-rubenstein: Yeah… I thought about adding this and kept going back and forth. Let's not have it, and that will let us act under the assumption that unlabeled ones are with runtime. (I hate that sort of un-clear assumption, but I don't know if there's another way)

breilly2 (approved):
👍  Looks good. I tested by having both a Cloud Environment and a Galaxy app in a single project. This did not lead to a warning in `ClusterManager` widget as it would if both disks were non-Galaxy. Additionally, `NewClusterModal` correctly ignores the Galaxy disk when choosing which disk to attach to a new runtime.

I also used the swagger UI to create an extra non-Galaxy disk and verified that `ClusterManager` shows a warning about there being an unexpected number of disks in the project.

On the topic of labeling the non-Galaxy disks or not... not labeling the disks means that we don't have to worry about a back-populating labels of existing disks, which is a definite plus. I believe it's fine to start labeling Galaxy disks now and not worry about back-populating existing Galaxy disks because this has not yet been released in production. We just have to watch out for the little bit of stale state that we currently have in non-prod environments.

breilly2:
One other note: An additional change we should probably make is to pay attention to this label on the Cloud Environments page (`clusters.js`). We could detect that a disk is for use with Galaxy, indicate that, and suppress the multi-disk warning we're currently showing.

PR #2252, opened 2020-09-25
Move cookie setting to top level

ehanna4:
Did some base testing but requires a bit more thorough click-through testing.

### DIFF
--- a/src/libs/state.js
+++ b/src/libs/state.js
@@ -13,6 +13,8 @@ export const authStore = Utils.atom({
   fenceStatus: {}
 })
 
+export const cookieReadyStore = Utils.atom(false)
+
 export const lastActiveTimeStore = staticStorageSlot(localStorage, 'idleTimeout')
 lastActiveTimeStore.update(v => v || {})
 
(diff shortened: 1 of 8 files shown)


src/libs/state.js
```diff
@@ -13,6 +13,8 @@ export const authStore = Utils.atom({
   fenceStatus: {}
 })
 
+export const cookieReadyStore = Utils.atom(false)
```
breilly2, line 16: For the sake of clarity/security/expression of intent, I think we should reset this to `false` on sign-out. This could be done when adding the call to clear the cookie, but I'd rather it be in this PR instead because this is where the global state is being added.

breilly2 (approved):
👍  Looks good. I tested locally by watching the browser's network inspector. First, I saw `setCookie` being called when I didn't have a runtime. Then I created a runtime and opened a notebook and saw it continue to be called. I also watched as the token "rolled over".

Interesting phenomenon: the token rolls over in between calls to `setCookie` resulting in requests containing the new token in the `Authorization` header and the old token in the `LeoToken` cookie. The next call to `setCookie` brings them back in sync.

PR #2253, opened 2020-09-28
[SATURN-1843] Add app cost hard coded

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/NewAppModal.js
+++ b/src/components/NewAppModal.js
@@ -131,8 +131,7 @@ export const NewAppModal = _.flow(
               'Running cloud compute costs ',
               span({ style: { fontWeight: 600 } }, `${Utils.formatUSD(
                 hourlyAppCost(
-                  app ? app.kubernetesRuntimeConfig : { machineType: 'n1-standard-8' },
-                  app ? app.status : 'RUNNING'
+                  app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8' } }
                 ) + persistentDiskCost({ size: 30, status: 'Running' })
               )} per hr`)
             ])
--- a/src/components/NewAppModal.js
+++ b/src/components/NewAppModal.js
@@ -130,7 +130,7 @@ export const NewAppModal = _.flow(
             li({ style: { marginTop: '1rem' } }, [
               'Running cloud compute costs ',
               // TODO: Actually calculate cost, rather than hardcode
-              span({ style: { fontWeight: 600 } }, `${Utils.formatUSD(appCost(app) + persistentDiskCost({ size: 30, status: 'Running' }))} per hr`)
+              span({ style: { fontWeight: 600 } }, `${Utils.formatUSD(appCost() + persistentDiskCost({ size: 30, status: 'Running' }))} per hr`)
             ])
           ]),
           h(Link, {
--- a/src/components/NewAppModal.js
+++ b/src/components/NewAppModal.js
@@ -7,7 +7,7 @@ import { withModalDrawer } from 'src/components/ModalDrawer'
 import TitleBar from 'src/components/TitleBar'
 import { machineTypes } from 'src/data/machines'
 import { Ajax } from 'src/libs/ajax'
-import { appCost, currentApp, persistentDiskCost } from 'src/libs/cluster-utils'
+import { currentApp, hourlyAppCost, persistentDiskCost } from 'src/libs/cluster-utils'
 import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
 import * as Style from 'src/libs/style'
@@ -129,8 +129,12 @@ export const NewAppModal = _.flow(
             ]),
             li({ style: { marginTop: '1rem' } }, [
               'Running cloud compute costs ',
-              // TODO: Actually calculate cost, rather than hardcode
-              span({ style: { fontWeight: 600 } }, `${Utils.formatUSD(appCost() + persistentDiskCost({ size: 30, status: 'Running' }))} per hr`)
+              span({ style: { fontWeight: 600 } }, `${Utils.formatUSD(
+                hourlyAppCost(
+                  app ? app.kubernetesRuntimeConfig : { machineType: 'n1-standard-8' },
+                  app ? app.status : 'RUNNING'
+                ) + persistentDiskCost({ size: 30, status: 'Running' })
+              )} per hr`)
             ])
           ]),
           h(Link, {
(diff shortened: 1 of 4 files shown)


src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -312,7 +312,8 @@ const Notebooks = _.flow(
         div({ style: { fontSize: 18, lineHeight: '22px', width: 160 } }, [
           div(['Galaxy Interactive']),
           div(['Environment']),
-          div({ style: { fontSize: 12, marginTop: 6 } }, [_.capitalize(app.status)]),
+          // TODO: Actually calculate cost, rather than hardcode in PD values
+          div({ style: { fontSize: 12, marginTop: 6 } }, [_.capitalize(app.status), ` ${Utils.formatUSD(appCost() + persistentDiskCost({size: 30, status: 'Running'}))}`]),
```
ehanna4:
```suggestion
          div({ style: { fontSize: 12, marginTop: 6 } }, [_.capitalize(app.status), ` ${Utils.formatUSD(appCost() + persistentDiskCost({ size: 30, status: 'Running' }))}`]),
```

ehanna4: (to fix the build/lint issue)

src/data/machines.js
```diff
@@ -41,6 +41,10 @@ export const monthlyStoragePrice = 0.04 // from https://cloud.google.com/compute
 export const storagePrice = monthlyStoragePrice / 730 // per GB hour using 730 hours per month from https://cloud.google.com/compute/pricing
 export const dataprocCpuPrice = 0.01 // dataproc costs $0.01 per cpu per hour
 
+// TODO calculate cost based on app, rather than constant
+// From https://cloud.google.com/compute/all-pricing#sustained_use monthly pricing
+export const appPrice = 194.18 // per GB hour using 730 hours per month from https://cloud.google.com/compute/pricing
```
panentheos: I don't think that we're expecting people to leave their clusters up for the whole month, so they probably won't get the full sustained use discount. It might be better to show the base hourly rate, since that's the worst case estimate.

Thoughts?

src/libs/cluster-utils.js
```diff
@@ -83,6 +83,9 @@ export const trimAppsOldestFirst = _.flow(
   _.remove({ status: 'DELETING' }),
   _.sortBy('auditInfo.createdDate'))
 
+// TODO: Actually calculate app cost and pass in app
+export const appCost = () => appPrice / 730
```
panentheos: It looks like the apps we get back from Leo already have a `kubernetesRuntimeConfig`, so we could actually compute the real cost if we have an app. If we did that, then we'd just need to fabricate a runtime config with the known defaults when we want to show the estimated cost (before you actually create one)

s-rubenstein: I was under the impression that we were punting on developing that for the time being to reduce scope. If we want to do the full calculation, that's totally fine, just want to make sure we are being intentional about effort

src/data/machines.js
s-rubenstein: It would worry me a bit to use a different calculation method than we are using for runtimes. Do runtimes also have a sustained use cost model? (I want to try to be consistent, cause my guess is the use case is consistent, but I am not sure. Maybe a product question?)

src/libs/cluster-utils.js
panentheos: That's fine too. I originally thought that we didn't have the data available, but when I was trying to reverse engineer where that number was coming from, I noticed we do get it.

In any case, if you buy my argument from the other comment about hourly rate, then this could be:
```js
_.find({ name: 'n1-standard-8' }, machineTypes).price
```
which is still hardcoded but documents where it's coming from.

src/data/machines.js
panentheos: Runtimes do too, but we use the base hourly cost. Note that this is currently doing the reverse of the calculation we do for runtimes. It's only for disks that we use the monthly and divide by 730.

(In case it wasn't clear, google's price sheet includes the discount when showing monthly price, but not for hourly)

src/libs/cluster-utils.js
s-rubenstein: Oh, sure, that's easy.

src/components/NewAppModal.js
```diff
@@ -130,8 +129,12 @@ export const NewAppModal = _.flow(
             ]),
             li({ style: { marginTop: '1rem' } }, [
               'Running cloud compute costs ',
-              span({ style: { fontWeight: 600 } }, '$0.00 per hr')
-              //TODO: Calculate cost
+              span({ style: { fontWeight: 600 } }, `${Utils.formatUSD(
+                hourlyAppCost(
+                  app ? app.kubernetesRuntimeConfig : { machineType: 'n1-standard-8' },
+                  app ? app.status : 'RUNNING'
+                ) + persistentDiskCost({ size: 30, status: 'Running' })
```
Shakespeared, line 135: I think in our other PR we changed the default disk size to 50

src/libs/cluster-utils.js
```diff
@@ -83,6 +83,11 @@ export const trimAppsOldestFirst = _.flow(
   _.remove({ status: 'DELETING' }),
   _.sortBy('auditInfo.createdDate'))
 
+// TODO: factor status into cost
+export const hourlyAppCost = (kubernetesRuntimeConfig, status) => {
```
panentheos: If we need multiple pieces of info, we could just pass the whole `app` object. That would simplify the code at the call sites too.

src/data/machines.js
panentheos: Now that we're computing the cost directly, this code could be removed.

src/components/NewAppModal.js
panentheos: I thought we observed that the default for Galaxy was 30 (coming from Leo)

s-rubenstein: It is, there was some confusion cause the text still says 50 in the launch PR, I think, so commented there instead.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -312,7 +312,11 @@ const Notebooks = _.flow(
         div({ style: { fontSize: 18, lineHeight: '22px', width: 160 } }, [
           div(['Galaxy Interactive']),
           div(['Environment']),
-          div({ style: { fontSize: 12, marginTop: 6 } }, [_.capitalize(app.status)]),
+          // TODO: Actually use status to calculate cost, and actually use disk rather than hardcoding
+          div({ style: { fontSize: 12, marginTop: 6 } }, [_.capitalize(app.status), ` ${Utils.formatUSD(
+            hourlyAppCost(app) +
+            persistentDiskCost({ size: 30, status: 'Running' })
+          )}`]),
```
panentheos: This should say 'per hr' as well.

panentheos (approved):
See one comment, then looks good. Tested locally.

PR #2255, opened 2020-10-01
deprecate passing non-function values to cond

panentheos:
`Utils.cond` currently treats values differently depending on whether they are functions or non-functions. This leads to some inconsistency in how it's used, and can open the door to some subtle or surprising bugs.

This deprecates the usage of non-function values. The implementation is unchanged for now, but passing a non-function value will result in a message on the console. This also changes all existing usages to conform to the stricter API. After some additional testing, the intention is to remove the option entirely.

### DIFF
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -139,7 +139,7 @@ export default _.flow(
         title: Utils.cond(
           [title, () => title],
           [cloneWorkspace, () => 'Clone a workspace'],
-          [true, () => 'Create a New Workspace']
+          () => 'Create a New Workspace'
         ),
         onDismiss,
         okButton: h(ButtonPrimary, {
@@ -149,7 +149,7 @@ export default _.flow(
         }, Utils.cond(
           [buttonText, () => buttonText],
           [cloneWorkspace, () => 'Clone Workspace'],
-          [true, () => 'Create Workspace']
+          () => 'Create Workspace'
         ))
       }, [
         h(IdContainer, [id => h(Fragment, [
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -134,12 +134,12 @@ export default _.flow(
     })
 
     return Utils.cond(
-      [loading, spinnerOverlay],
+      [loading, () => spinnerOverlay],
       [hasBillingProjects, () => h(Modal, {
         title: Utils.cond(
-          [title, title],
-          [cloneWorkspace, 'Clone a workspace'],
-          [Utils.DEFAULT, 'Create a New Workspace']
+          [title, () => title],
+          [cloneWorkspace, () => 'Clone a workspace'],
+          [true, () => 'Create a New Workspace']
         ),
         onDismiss,
         okButton: h(ButtonPrimary, {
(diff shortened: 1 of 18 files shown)


src/components/NewWorkspaceModal.js
```diff
@@ -134,22 +134,22 @@ export default _.flow(
...
-      [loading, spinnerOverlay],
+      [loading, () => spinnerOverlay],
       [hasBillingProjects, () => h(Modal, {
         title: Utils.cond(
-          [title, title],
-          [cloneWorkspace, 'Clone a workspace'],
-          [Utils.DEFAULT, 'Create a New Workspace']
+          [title, () => title],
+          [cloneWorkspace, () => 'Clone a workspace'],
+          [true, () => 'Create a New Workspace']
```
s-rubenstein: Why use true instead of `Utils.DEFAULT`? I personally think the default is clearer on meaning.

src/components/PrivateDataExplorer.js
```diff
@@ -103,7 +103,7 @@ export default _.flow(
           p([
             `Thank you for your interest in the Baseline Health Study data. Baseline data is currently only being shared
              with Baseline Health Study partner sites, Duke and Stanford. If you are a current researcher at one of those sites,
-             please reach out to your institutional contacts for information on how to obtain access. In the future, Baseline 
+             please reach out to your institutional contacts for information on how to obtain access. In the future, Baseline
```
s-rubenstein, line 106: This looks like an unrelated change. Is it for linting?

src/components/NewWorkspaceModal.js
s-rubenstein: I saw below that there are also some that just have a default value after without a conditional. Maybe these should move to that?

src/components/NewWorkspaceModal.js
```diff
@@ -134,22 +134,22 @@ export default _.flow(
...
           disabled: errors,
           tooltip: Utils.summarizeErrors(errors),
           onClick: () => this.create()
         }, Utils.cond(
-          [buttonText, buttonText],
-          [cloneWorkspace, 'Clone Workspace'],
-          [Utils.DEFAULT, 'Create Workspace']
+          [buttonText, () => buttonText],
+          [cloneWorkspace, () => 'Clone Workspace'],
+          [true, () => 'Create Workspace']
```
s-rubenstein: See above comment.

src/components/NewWorkspaceModal.js
panentheos: `Utils.DEFAULT` is a symbol with special meaning to `Utils.switchCase`. While it technically works here, I think it's misleading, because `cond` only looks for truthiness, so there's nothing special about `DEFAULT` as opposed to `true` or any other truthy value.

As for expressing the default, I've been pondering whether it would be good to also remove the special handling of the 'hanging function', and have everything use this form instead. Still thinking on that.

src/components/PrivateDataExplorer.js
panentheos: Sort of. My editor fixes trailing whitespace on save. The only reason it's not an actual lint error is because it's inside of a multiline string.

src/components/NewWorkspaceModal.js
s-rubenstein: I am happy with that change, but would prefer it moved to a new PR for simplicity and traceability's sake.

s-rubenstein (approved):
LGTM

PR #2257, opened 2020-10-01
work around stale PD data after an update [SATURN-1853]

panentheos:
This attempts to address the issue of runtime/disk status not being reflected immediately after an update. In cases where the status change may be delayed, we schedule a fast refresh for one polling cycle, in the hopes of picking up the new status.

Note that by itself, this won't solve the case of quickly re-opening the modal and pressing Create within 10 seconds after an update. If we want to address that, we'll need to go further, and add some additional state that disables the button during that time.

### DIFF
--- a/src/pages/workspaces/workspace/applications/AppLauncher.js
+++ b/src/pages/workspaces/workspace/applications/AppLauncher.js
@@ -81,10 +81,9 @@ const AppLauncher = _.flow(
           onSuccess: _.flow(
             withErrorReporting('Error creating cluster'),
             Utils.withBusyState(setBusy)
-          )(async promise => {
+          )(async () => {
             setShowCreate(false)
-            await promise
-            await refreshClusters()
+            await refreshClusters(true)
           })
         }),
         busy && spinnerOverlay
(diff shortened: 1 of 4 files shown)


src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -81,10 +81,9 @@ const AppLauncher = _.flow(
           onSuccess: _.flow(
             withErrorReporting('Error creating cluster'),
             Utils.withBusyState(setBusy)
-          )(async promise => {
+          )(async () => {
```
breilly2, line 84: Good clean-up. Not sure how `promise` went unused, but the code makes sense to me like this so I'm not going to worry.

breilly2 (approved):
👍 Looks good. Tested locally by simulating the browser connection network speed as "Slow 3G". This allowed me to see that, before this change, the old disk size is displayed when the modal opens but before the fetch triggered by opening the modal completes. With this change, as long as I wait 10 seconds before re-opening the modal, I can see the correct disk size even before the fetch triggered by opening the modal completes.

src/pages/workspaces/workspace/applications/AppLauncher.js
panentheos: Awaiting a non-promise value just returns the value immediately, so this had become a noop since we stopped passing an argument to `onSuccess`.

PR #2258, opened 2020-10-01
rename 'clusters' in Cloud Environments page

panentheos:
This is a subset of the overall rename effort.

### DIFF
--- a/src/pages/Clusters.js
+++ b/src/pages/Clusters.js
@@ -20,10 +20,10 @@ import * as Style from 'src/libs/style'
 import { cond, formatUSD, makeCompleteDate, useCancellation, useGetter, useOnMount, usePollingEffect, withBusyState } from 'src/libs/utils'
 
 
-const DeleteClusterModal = ({ cluster: { googleProject, runtimeName, runtimeConfig: { persistentDiskId } }, onDismiss, onSuccess }) => {
+const DeleteRuntimeModal = ({ runtime: { googleProject, runtimeName, runtimeConfig: { persistentDiskId } }, onDismiss, onSuccess }) => {
   const [deleteDisk, setDeleteDisk] = useState(false)
   const [deleting, setDeleting] = useState()
-  const deleteCluster = _.flow(
+  const deleteRuntime = _.flow(
     withBusyState(setDeleting),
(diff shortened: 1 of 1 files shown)


src/pages/Clusters.js
```diff
@@ -132,27 +132,27 @@ const Clusters = () => {
     size: 'size'
   }[diskSort.field]], [diskSort.direction], disks)
 
-  const totalCost = _.sum(_.map(clusterCost, clusters))
+  const totalCost = _.sum(_.map(clusterCost, runtimes))
```
s-rubenstein, line 135: Presumably clusterCost is going to be changed with `cluster-utils`?

src/pages/Clusters.js
```diff
@@ -132,27 +132,27 @@ const Clusters = () => {
...
         columns: [
           {
             headerRenderer: () => h(Sortable, { sort, field: 'project', onSort: setSort }, ['Billing project']),
             cellRenderer: ({ rowIndex }) => {
-              const cluster = filteredClusters[rowIndex]
-              const inactive = !_.includes(cluster.status, ['Deleting', 'Error']) &&
-                currentCluster(clustersByProject[cluster.googleProject]) !== cluster
+              const runtime = filteredRuntimes[rowIndex]
+              const inactive = !_.includes(runtime.status, ['Deleting', 'Error']) &&
+                currentCluster(runtimesByProject[runtime.googleProject]) !== runtime
```
s-rubenstein, line 153: Presumably currentCluster is going to change with `cluster-utils`?

s-rubenstein (commented):
Nit: We should probably also rename the file to Environments.js? And then could we put a note above line 345 saying that the path is to maintain backwards compatibility with old urls?

panentheos:
File renames will be done in a separate PR. If you make substantial changes to the contents of a file at the same time as renaming it, git sometimes fails to classify it as a rename, which can make archaeology harder.

src/pages/Clusters.js
panentheos: Yes, this PR is scoped to just local names within one file for easier reviewing.

s-rubenstein (approved):
LGTM

PR #2259, opened 2020-10-02
rename 'clusters' in NewClusterModal

panentheos:

### DIFF
--- a/src/components/NewClusterModal.js
+++ b/src/components/NewClusterModal.js
@@ -128,19 +128,19 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
 
   constructor(props) {
     super(props)
-    const currentCluster = this.getCurrentCluster()
+    const currentRuntime = this.getCurrentRuntime()
     const currentPersistentDisk = this.getCurrentPersistentDisk()
 
     this.state = {
       loading: false,
-      currentClusterDetails: currentCluster,
+      currentRuntimeDetails: currentRuntime,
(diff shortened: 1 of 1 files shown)


src/components/NewClusterModal.js
```diff
@@ -449,35 +449,35 @@ export const NewClusterModal = withModalDrawer({ width: 675 })(class NewClusterM
...
 
-  getImageUrl(clusterDetails) {
-    return _.find(({ imageType }) => _.includes(imageType, ['Jupyter', 'RStudio']), clusterDetails?.runtimeImages)?.imageUrl
+  getImageUrl(runtimeDetails) {
+    return _.find(({ imageType }) => _.includes(imageType, ['Jupyter', 'RStudio']), runtimeDetails?.runtimeImages)?.imageUrl
   }
 
   componentDidMount = _.flow(
-    withErrorReporting('Error loading cluster'),
+    withErrorReporting('Error loading environment'),
```
s-rubenstein: Nit: I would maybe say `cloud environment` since that's what we call it to the user.

s-rubenstein (commented):
https://github.com/DataBiosphere/terra-ui/pull/2259/files#diff-9fc1399eaeddca57bf342d5f239c71faR120 we should rename the component, probably. And also ideally rename the file.

src/components/NewClusterModal.js
panentheos: Done.

s-rubenstein (approved):
LGTM

PR #2260, opened 2020-10-02
Add a applicationLaunch event for galaxy launch [SATURN-1856]

ehanna4:
Adds an application launch event with the app name of 'Galaxy' for launching galaxy.
Tested by checking the mixpanel dashboard and can see the events came through

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -21,7 +21,6 @@ import {
 } from 'src/libs/cluster-utils'
 import colors from 'src/libs/colors'
 import { reportError, withErrorReporting } from 'src/libs/error'
-import Events from 'src/libs/events'
 import * as Nav from 'src/libs/nav'
 import { clearNotification, notify } from 'src/libs/notifications'
 import { errorNotifiedClusters } from 'src/libs/state'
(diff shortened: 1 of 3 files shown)


src/components/ClusterManager.js
```diff
@@ -181,7 +182,10 @@ export default class ClusterManager extends PureComponent {
           h(GalaxyWarning),
           h(GalaxyLaunchButton, {
             app,
-            onClick: () => clearNotification(galaxyId)
+            onClick: () => {
+              clearNotification(galaxyId)
+              Ajax().Metrics.captureEvent(Events.applicationLaunch, { app: 'Galaxy' })
```
panentheos: Since we want to do the same thing in both places, we could add this to `GalaxyLaunchButton` instead (as long as we make sure it respects the passed `onClick` as well)

ehanna4: oh I thought since the instances defined their own `onClick`s adding this to the definition would override that, I will look into moving it, thanks!

src/components/NewAppModal.js
```diff
@@ -57,7 +57,9 @@ export const NewAppModal = _.flow(
         return h(ButtonPrimary, { onClick: () => createApp() }, ['Create'])
       }],
       ['launchWarn', () => {
-        return h(GalaxyLaunchButton, { app, onClick: onDismiss })
+        return h(GalaxyLaunchButton, {
+          app, onClick: () => onDismiss()
+        })
```
panentheos: nit: This looks left-over from the previous implementation, it would be nice to remove this diff.

panentheos (approved):
Looks good, tested locally.

PR #2261, opened 2020-10-02
rename 'clusters' in ClusterManager

panentheos:


s-rubenstein (commented):
Also maybe should look at changing the filename and classname https://github.com/DataBiosphere/terra-ui/pull/2261/files#diff-10c62c0020a716096dd29166c6500c42L114 as well.

s-rubenstein (approved):
LGTM

PR #2262, opened 2020-10-02
send create and delete metrics for galaxy apps [SATURN-1807]

panentheos:
see ticket for details


src/libs/events.js
```diff
@@ -12,6 +12,8 @@ const eventsList = {
   cloudEnvironmentCreate: 'cloudEnvironment:create',
   cloudEnvironmentDelete: 'cloudEnvironment:delete',
   cloudEnvironmentUpdate: 'cloudEnvironment:update',
+  galaxyAppCreate: 'galaxyApp:create',
```
ehanna4: One day there might be more than just Galaxy apps, so I'm wondering if we should have Galaxy be a property and name the event just like `appCreate` instead?

panentheos: TOL: It's an interesting question, and one that's worth thinking extra hard about, because these events become permanent state that will affect the ease of future analytics. It seems like the relevant question is: Do 'apps' have common semantics outside of the specific software, or are they just an implementation detail? In other words, would we ever care about querying 'all app creations', as opposed to just 'all Galaxy spin-ups'? Then there's the complication that we're using 'application:launch' for Terminal and RStudio, even though those don't use Leo's 'apps' endpoints at all.

All that said, it seems like we're already establishing a pattern of going toward the more generic, so I'll follow that here.

ehanna4 (approved):
LGTM

PR #2263, opened 2020-10-02
add mixpanel tracking for user registration [SATURN-1775]

Shakespeared:
Add mixpanel tracking for registration to terra

### DIFF
--- a/src/pages/Register.js
+++ b/src/pages/Register.js
@@ -8,6 +8,7 @@ import { Ajax } from 'src/libs/ajax'
 import { getUser, refreshTerraProfile, signOut } from 'src/libs/auth'
 import colors from 'src/libs/colors'
 import { reportError } from 'src/libs/error'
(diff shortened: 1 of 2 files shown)


src/pages/Register.js
```diff
@@ -42,6 +43,7 @@ export default class Register extends Component {
       })
       authStore.update(state => ({ ...state, registrationStatus: 'registered' }))
       await refreshTerraProfile()
+      Ajax().Metrics.captureEvent(Events.userRegister)
```
s-rubenstein, line 46: Do we want any info about the user here? Or are we good without?

panentheos: Most of what's on the form is private info that we don't want in Mixpanel, and the anonymous contact email is automatically synced via the profile. So I think just the event is enough here.

s-rubenstein (approved):
LGTM

PR #2264, opened 2020-10-02
[SATURN-1870] First round of renames from galaxy to app

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/ClusterManager.js
+++ b/src/components/ClusterManager.js
@@ -128,7 +128,7 @@ export default class ClusterManager extends PureComponent {
     this.state = {
       createModalDrawerOpen: false,
       busy: false,
-      galaxyDrawerOpen: false
+      appDrawerOpen: false
     }
   }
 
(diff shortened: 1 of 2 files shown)


s-rubenstein:
One thing I wasn't sure whether we should change or not was the GalaxyLaunchButton. Right now it only opens galaxy specifically, but I could easily see it including all apps in the future.

panentheos (commented):
Is this the direction we want to go here? I know the stated intention is for Leo to use 'apps' for more things in the future, but it's not clear when that will actually happen, or what it will look like.

This change pre-supposes that there will be some sort of unified 'app management' UI, but personally I'm not sure that's what the future will look like. It seems equally likely that we'd build custom UI for each kind of software. In that case, this code would only ever deal with Galaxy, so having an overly generic name might feel unclear.

s-rubenstein:
I am worried because the APIs are all called `App`. I could see us building multiple shells over the api that hardcode type, but that hasn't been how we've gone there. We _could_ do that instead though. 

Since we don't know what we want to do, I am going to pose the question: "Should we put this back in the backlog for more refinement?" 

It doesn't quite feel ready. I took my best shot at what we could/should name things, but I wonder if there's misalignment/misunderstanding of how we should treat this code.

PR #2266, opened 2020-10-05
rename 'clusters' in NotebookLauncher

panentheos:


s-rubenstein (approved):
LGTM (Tested locally)

PR #2267, opened 2020-10-05
rename 'clusters' in AppLauncher

panentheos:

### DIFF
--- a/src/pages/workspaces/workspace/applications/AppLauncher.js
+++ b/src/pages/workspaces/workspace/applications/AppLauncher.js
@@ -79,7 +79,7 @@ const AppLauncher = _.flow(
           namespace, name, clusters, persistentDisks,
           onDismiss: () => setShowCreate(false),
           onSuccess: _.flow(
-            withErrorReporting('Error creating runtime'),
+            withErrorReporting('Error loading cloud environment'),
             Utils.withBusyState(setBusy)
           )(async () => {
             setShowCreate(false)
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -79,7 +79,7 @@ const AppLauncher = _.flow(
           namespace, name, clusters, persistentDisks,
           onDismiss: () => setShowCreate(false),
           onSuccess: _.flow(
-            withErrorReporting('Error creating cluster'),
+            withErrorReporting('Error creating runtime'),
```
breilly2: Nit: This should probably reference "cloud environment" instead of "runtime".
TOL: I think this error message exists elsewhere, too. It would be nice if somehow `NewClusterModal` could be responsible for this error message.

breilly2 (approved):
👍  Looks good. I tested locally by launching the terminal for a Jupyter runtime.

I considered whether or not this might be the time to locally rename the properties destructured from `AppLauncher`'s props, but those lines would need to change later anyway so I agree with this PR not changing those yet. That PR will be larger as a result, but will be more self-contained.

src/pages/workspaces/workspace/applications/AppLauncher.js
panentheos: Good call, and actually this should say 'Error loading cloud environment', since that's all it does now.

PR #2268, opened 2020-10-06
Update the baseline link [SATURN-1710]

ehanna4:
Update the link to request access so that for the baseline data set users will be able to fill out the google form rather than reaching out to Terra support (who would just have to send the same form to the users aka cut out the middle-man).


panentheos (approved):
Looks good, tested locally.

PR #2270, opened 2020-10-06
rename applauncher.js [SATURN-1877]

Shakespeared:
rename applauncher to galaxylauncher, tested galaxy launch flow in UI


s-rubenstein (commented):
This isn't actually used for galaxy. I believe it is used for RStudio and the Terminal? The confusion is why it needs a new name :D

Shakespeared:
misunderstood the point of the ticket, rolling back these changes sorry!

PR #2271, opened 2020-10-06
[SATURN-1878] Change names to galaxy on components

s-rubenstein:
I decided not to change methods like `currentApp` or `appIsSettingUp` or `trimAppsOldestFirst` because I thought that those are working with the leo representation, and I would want to do some filtering to make sure it really was only pulling galaxy information before changing those. (So some behavior changes would be needed)

Likewise did not change or do anything to the polling.


src/components/NewGalaxyModal.js
```diff
@@ -21,16 +21,15 @@ const styles = {
...
-export const NewAppModal = _.flow(
-  Utils.withDisplayName('NewAppModal'),
+export const NewGalaxyModal = _.flow(
+  Utils.withDisplayName('NewGalaxyModal'),
   withModalDrawer({ width: 675 })
 )(({ onDismiss, onSuccess, apps, workspace: { workspace: { namespace, bucketName, name: workspaceName } } }) => {
   const [viewMode, setViewMode] = useState(undefined)
   const [loading, setLoading] = useState(false)
 
   const app = currentApp(apps)
```
s-rubenstein, line 32: I wasn't sure if this should change. Given that it is still currentApp, but this does keep the abstraction of app here. May be okay cause its used in the scope of `NewGalaxyModal`?

panentheos (approved):
See lint issue, then looks good. Tested locally.

PR #2272, opened 2020-10-06
Add galaxy version

ehanna4:
Tiny piece of info that was missing from galaxy config drawer. 

Not currently passed in the API response. The galaxy image version is specified in the helm chart, and the helm chart version is specified in Leo config. So going to hardcode this in the UI for now and if it changes or starts to be an issue like that will update accordingly.


s-rubenstein (approved):
LGTM. Can you add a todo to read this from the config?

ehanna4:
@s-rubenstein I've added a ticket to the epic on post-workshop work to consider reading from config rather than hardcoding the version.

PR #2273, opened 2020-10-06
broken tests are annoying so I fix them

zarsky-broad:
The wider ranging fix would be to expose a mechanism to generate a new browser context, but that's much more involved.

### DIFF
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -43,6 +43,10 @@ const testRunNotebookFn = _.flow(
   await findElement(page, clickable({ textContains: 'Creating' }))
   await findElement(page, clickable({ textContains: 'Running' }), { timeout: 10 * 60 * 1000 })
 
+  await page.evaluate(async () => {
+    await window.Ajax().Runtimes.setCookie() // make sure the right access cookie is set
+  })
+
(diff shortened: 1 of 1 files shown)


s-rubenstein (approved):
LGTM

breilly2:
I know I'm late to the conversation, but I'm not a fan of this solution. The test script should stick to performing gestures that a user would perform.

Also, this didn't seem to work: https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/4310/workflows/b1af9a06-cb95-4233-8fbf-eb08d46ba43b/jobs/24530

I think it needs to be done sooner. Since we've already clicked "Edit", `NotebookLauncher` is just waiting for the runtime to start and will open the notebook as soon as it detects that. It's not going to wait for the test to also detect that the runtime is ready. Bumping it up a line or two, or even at the very beginning of the test, should work.

One more observation: This gives us one shot at setting the correct cookie. This call could fail. The shape of the error will be better than what we have now, but will still be an error. There's still an underlying problem which looks to me like we're opening the notebook even though we haven't successfully set the cookie since logging in... which doesn't make sense to me from reading the code. My hypothesis may be wrong, but I believe there's more investigation to do here.

PR #2274, opened 2020-10-06
update deps

zarsky-broad:
Read changelogs etc, tested locally and ran integration tests.

We can't upgrade to the new major version of `history` (it doesn't support hash navigation without leading slash), or `@google-cloud/secret-manager` (it has bugs); I didn't update IGV.


panentheos (approved):
Looks good, tested on the PR site.

PR #2278, opened 2020-10-07
rename 'clusters' in WorkspaceContainer and downstream props

panentheos:


breilly2 (approved):
👍 Looks good. I traced through the code to make sure all instances were handled. Good catch with `ClusterManager.componentDidUpdate`! I tested locally by managing a runtime, opening a notebook, and launching terminal (from a stopped state).

PR #2280, opened 2020-10-08
track users across registration, matching anon and signed in events in mixpanel [BW-409]

Shakespeared:


ehanna4:
In testing today I am noticing an identify event in mixpanel whenever I am already logged in then close Terra without signing out and re-open a Terra window.

src/libs/auth.js
```diff
@@ -197,6 +198,14 @@ authStore.subscribe(async (state, oldState) => {
   }
 })
 
+authStore.subscribe(async (state, oldState) => {
+  if (oldState.registrationStatus === 'unregistered' && state.registrationStatus === 'registered') {
```
panentheos: TOL: This does prevent extra `identify` calls by only triggering on registration, but there's still an underlying issue we might want to address at some point, specifically that on initial page load, if the path is public, the page view event will fire before the user's status is confirmed, so it will be 'orphaned' as an anonymous event.

A potential fix would be to modify the logic in `AuthContainer` to not render public pages as eagerly. The tradeoff in that case would be slightly longer perceived load time for registered users.

panentheos (dismissed):
Looks good, tested locally by running the integration test.

src/libs/auth.js
Shakespeared: What do you think about going ahead with that change vs merging as it is now, or maybe leaving a comment or TODO there to indicate a possible future improvement?

panentheos: I think a comment would be good, perhaps in `PageViewReporter` since that's what ultimately gets affected.

ehanna4: I just chatted with Amelia and the main question they are looking to get at is the conversion rate from people who browse anonymously and leave vs people who then register to use the site. So I think it might be worthwhile to try @panentheos 's suggestion in this PR so we don't muddy the data with orphaned anon events waiting to add it later

panentheos:
This is ready for a fresh look. The issue regarding page loads has been addressed by the addition of `ensureAuthSettled`, so we now avoid recording/identifying anonymous events unless the user is actually signed out. Tested locally.

salonishah11 (approved):
LGTM 👍

Shakespeared:
This update looks good to me, tested locally. Thanks for resolving the page view issue!

PR #2283, opened 2020-10-09
[SATURN-1798] Add delete workspace confirmation text

s-rubenstein:
Screen Shot 2020-10-09 at 3 21 23 PM

### DIFF
--- a/src/pages/workspaces/workspace/DeleteWorkspaceModal.js
+++ b/src/pages/workspaces/workspace/DeleteWorkspaceModal.js
@@ -1,18 +1,21 @@
 import { Component } from 'react'
-import { div, h, span } from 'react-hyperscript-helpers'
+import { div, h, label, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, Link, spinnerOverlay } from 'src/components/common'
+import { TextInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
 import { Ajax } from 'src/libs/ajax'
 import { bucketBrowserUrl } from 'src/libs/auth'
 import { reportError } from 'src/libs/error'
 import * as Utils from 'src/libs/utils'
+import * as _ from 'lodash/fp'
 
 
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,18 +1,21 @@
...
-import { div, h, span } from 'react-hyperscript-helpers'
+import { div, h, label, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, Link, spinnerOverlay } from 'src/components/common'
+import { TextInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
 import { Ajax } from 'src/libs/ajax'
 import { bucketBrowserUrl } from 'src/libs/auth'
 import { reportError } from 'src/libs/error'
 import * as Utils from 'src/libs/utils'
+import * as _ from 'lodash/fp'
```
panentheos: We usually import the default, e.g.
```suggestion
import _ from 'lodash/fp'
```
This may also work, but we should keep it consistent.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -55,6 +59,15 @@ export default class DeleteWorkspaceModal extends Component {
           marginTop: '1rem'
         }
       }, 'This cannot be undone.'),
+      div({ style: { marginTop: '1rem' } }, [
+        label({ htmlFor: 'understand' }, ['Please type \'I Understand\' to continue:']),
```
panentheos: I can imagine someone typing "I understand" without realizing what they're agreeing to. Perhaps we should have them type "delete workspace"?

s-rubenstein: Sounds good. Will change.

panentheos (approved):
Looks good, tested locally.

PR #2284, opened 2020-10-13
rename 'clusters' in exported identifiers

panentheos:


src/pages/Clusters.js
```diff
@@ -342,7 +342,7 @@ const Environments = () => {
 
 export const navPaths = [
   {
-    name: 'clusters',
+    name: 'environments',
     path: '/clusters',
```
s-rubenstein: Mind adding a comment here explaining why the path is different than the name now? (Just for future breadcrumbs)

s-rubenstein (approved):
LGTM with one comment

PR #2286, opened 2020-10-13
unlink fence accounts

Shakespeared:
Allow permanent unlinking of fence accounts from profile or notification


src/pages/Profile.js
```diff
@@ -117,24 +121,26 @@ const NihLink = ({ nihToken }) => {
...
-            div([h(ShibbolethLink, ['Log in to NIH to re-link your account'])])
-          ])
-        ]),
-        _.flow(
-          _.sortBy('name'),
-          _.map(renderDatasetAuthStatus)
-        )(datasetPermissions)
+    return div({ style: styles.identityLine }, [
+      h(Fragment, [
+        !linkedNihUsername && div({ style: styles.identityLine }, [ShibbolethLink, ['Log in to NIH to link your account']]),
```
panentheos: It looks like something got lost in this edit. The `ShibbolethLink` won't render properly here.

src/pages/Profile.js
```diff
@@ -117,24 +121,26 @@ const NihLink = ({ nihToken }) => {
...
-            div([Utils.makeCompleteDate(linkExpireTime * 1000)]),
-            div([h(ShibbolethLink, ['Log in to NIH to re-link your account'])])
-          ])
-        ]),
-        _.flow(
-          _.sortBy('name'),
-          _.map(renderDatasetAuthStatus)
-        )(datasetPermissions)
+    return div({ style: styles.identityLine }, [
+      h(Fragment, [
```
panentheos: This `Fragment` isn't necessary, since there's a `div` that can contain the children.

src/pages/Profile.js
```diff
@@ -207,17 +213,25 @@ const FenceLink = ({ provider: { key, name } }) => {
...
+            div({ style: { flex: 1 } }, ['Username:']),
+            div({ style: { flex: 2 } }, [username])
+          ]),
+          div({ style: styles.identityLine }, [
+            div({ style: { flex: 1 } }, ['Link Expiration:']),
+            div({ style: { flex: 2 } }, [Utils.makeCompleteDate(expireTime)])
+          ])
         ]),
-        h(FrameworkServiceLink, { linkText: 'Log-In to Framework Services to re-link your account', provider: key, redirectUrl })
+        div({ style: styles.identityLine }, [
```
panentheos, line 226: If this `div` was inside the flex column just above, then we wouldn't need the additional `div` wrapper around everything. In general, we should strive for as few wrappers as possible to achieve the desired layout.

Shakespeared: Removed a few extra divs here, which I think is what you mean. I can't see a way to remove the flex:1 and flex:2 divs without losing the formatting, but if there is a way (or if I messed up the formatting in a way I can't see) let me know!

src/pages/Profile.js
Shakespeared: good catch! Put this back the way it was.

panentheos (approved):
Looks good, tested locally.

PR #2287, opened 2020-10-14
replace ajaxCaller wrapper

panentheos:
`ajaxCaller` was an old component wrapper that injected a prop containing an `Ajax` object with cancellation pre-wired. The name was confusing, and it obscured the fact that it included cancellation.

This replaces `ajaxCaller` with a `withCancellation` wrapper, which injects the `signal` prop directly, which is then passed to the individual ajax calls. This makes cancellation more explicit, and paves the way for eventual conversion to the hook-based equivalent, `useCancellation`


zarsky-broad (approved):
There's a lot here, so I'm not able to really test it properly, but this looks like a really good change to me.

src/libs/utils.js
```diff
@@ -392,6 +392,13 @@ export const useCancellation = () => {
   return controller.current.signal
 }
 
+export const withCancellation = WrappedComponent => {
```
breilly2: I understand the evolution of the naming of `withCancellation` (which matches our `useCancellation` hook). However, if I were coming in fresh, I wouldn't immediately recognize `withCancellation` as a thing that introduces a prop called `signal`. What would you think about calling this `withCancellationSignal`?

Although the name is longer, usage of this should shrink over time as we have fewer class components. I find the use of `useCancellation` perfectly readable because the correlation between "cancellation" and "signal" is all together on one line:
```
```

panentheos: Good idea, I'll implement.

breilly2 (approved):
👍  Code looks good. I tested various pages locally, including making sure ajax cancellation still works by navigating to a workspace and quickly using the back button (keyboard shortcut).

I really like this change. I remember being initially confused about how to read the use of `ajaxCaller`. This feels much more straightforward.

PR #2290, opened 2020-10-19
Update cost calculation for galaxy, version number and cost text [IA-2288]

ehanna4:
- Updated the way the cost for a Galaxy app is calculated so that all the disks (250 + 10 + 100) are reflected in the cost
- Updated text so the disk size shown to the user is 250GB and the language reflects the changes Adrian wanted to see
- Update version of Galaxy

### DIFF
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -119,7 +119,7 @@ export const NewGalaxyModal = _.flow(
 
   const renderDefaultCase = () => {
     const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineTypes)
-    const cost = getGalaxyCost(app)
+    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8' } })
     return h(Fragment, [
       div([`Environment ${app ? 'consists' : 'will consist'} of an application and cloud compute.`]),
       div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -11,7 +11,7 @@ import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
 import Events, { extractWorkspaceDetails } from 'src/libs/events'
-import { currentApp, hourlyKubernetesAppCost, persistentDiskCost } from 'src/libs/runtime-utils'
+import { currentApp, getGalaxyCost } from 'src/libs/runtime-utils'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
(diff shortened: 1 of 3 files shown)


rtitle (approved):
:shipit: !

src/libs/runtime-utils.js
```diff
@@ -69,6 +69,11 @@ export const runtimeCost = ({ runtimeConfig, status }) => {
   }
 }
 
+export const getGalaxyCost = app => {
+  return hourlyKubernetesAppCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8' } }) +
```
jdcanas: its probably desirable to pull this machineType off the app itself, since the API GET response does specify it, and it can be passed in at creation time (though I don't think we support it in the UI currently, the above value being the default). Otherwise, if we change the default in the API conf this will be incorrect.

src/libs/runtime-utils.js
```diff
@@ -69,6 +69,11 @@ export const runtimeCost = ({ runtimeConfig, status }) => {
...
+    persistentDiskCost({ size: 250 + 10 + 100, status: 'Running' }) // 250Gb for the NFS disk, 10Gb for the postgres disk, and 100Gb for the boot disk
```
jdcanas: why do we pass in a hardcoded 'Running' here?

src/libs/runtime-utils.js
ehanna4: oh good call - I think when this logic was added the response did not include it? I will make the update now though

jdcanas: oh, i think it actually is doing that, its just appending that kubernetesRuntimeConfig if the app doesn't have it? Still probably good to update it, the app should always have it

src/components/NewGalaxyModal.js
```diff
@@ -119,23 +119,23 @@ export const NewGalaxyModal = _.flow(
 
   const renderDefaultCase = () => {
     const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineTypes)
-    const cost = hourlyKubernetesAppCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8' } }) + persistentDiskCost({ size: 30, status: 'Running' })
+    const cost = getGalaxyCost(app)
```
panentheos: I think the `app || ...` expression should remain here, since it's specific to this UI (we're either reporting on an existing app, or showing a preview of what the cost will be after creation). That way the helper function can always expect a value.

src/libs/runtime-utils.js
ehanna4: I believe the motivation is so that we are be able to show the user what the cost _will_ be when running, otherwise it could show as $0.00

panentheos: TOL: Are these numbers available via the persistent disk list that we get back? If so, it might be nice to compute the cost from there at some point (though it would mean wiring in the disks here).

src/libs/runtime-utils.js
ehanna4: oh I mistook where this was, this line is only if there is no app already then we are setting the default to be a 'n1-standard-8' machine. If there is an app then we get the machine type off the app

src/libs/runtime-utils.js
ehanna4: I'm not seeing these in the list, though this might be a needed piece in the future. I will make a note, thanks!

PR #2291, opened 2020-10-20
[IA-2255] add RStudio image to cloud environments dropdown

gpcarr:
Changes in this PR:
- Remove "Project Specific" option in dropdown (we are now including the RStudio Bioconductor image as an option in the dropdown so we no longer need the project specific option)
- Adds new section in cloud environments dropdown called `Community-Maintained RStudio Environments (verified partners)`
- Under the new section added, adds the Rstudio image with label `RStudio (R 4.0.0, Bioconductor 3.11.1, Python 2.7.17)`

Testing:
- Tested with the corresponding changes to this terra-docker PR: https://github.com/DataBiosphere/terra-docker/pull/164
- Verified that the new section + option are rendered as expected in the dropdown
- Verified that I can choose the new option in the dropdown and the rstudio cloud environment starts up as expected
- Verified that existing functionality still works as expected (if I have a runtime `Running` or `Stopped` that was built with the RStudio image that the image is selected in the dropdown and the R logo appears in the top right next to the pause button)


src/components/NewRuntimeModal.js
```diff
@@ -306,7 +305,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
       await Ajax().Runtimes.runtime(namespace, Utils.generateRuntimeName()).create({
         runtimeConfig,
         toolDockerImage: newRuntime.toolDockerImage,
-        labels: { saturnIsProjectSpecific: `${selectedLeoImage === PROJECT_SPECIFIC_MODE}` },
+        labels: {},
```
ehanna4: I don't think you need to keep `labels` here at all anymore since it's empty

src/components/NewRuntimeModal.js
```diff
@@ -328,7 +327,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
         [(viewMode !== 'deleteEnvironmentOptions'), () => {
           return {
             cloudService,
-            toolDockerImage: _.includes(selectedLeoImage, [CUSTOM_MODE, PROJECT_SPECIFIC_MODE]) ? customEnvImage : selectedLeoImage,
+            toolDockerImage: _.includes(selectedLeoImage, [CUSTOM_MODE]) ? customEnvImage : selectedLeoImage,
```
ehanna4: since it's only comparing the `selectedLeoImage` to `CUSTOM_MODE` now it might be simpler to use a `===` rather than `_.includes`

src/components/NewRuntimeModal.js
```diff
@@ -474,7 +473,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
     const foundImage = _.find({ image: imageUrl }, newLeoImages)
     this.setState({
       leoImages: newLeoImages, currentRuntimeDetails, currentPersistentDiskDetails,
-      selectedLeoImage: foundImage ? imageUrl : (currentRuntimeDetails?.labels.saturnIsProjectSpecific === 'true' ? PROJECT_SPECIFIC_MODE : CUSTOM_MODE),
+      selectedLeoImage: foundImage ? imageUrl : CUSTOM_MODE,
```
ehanna4, line 480: This is such a satisfying change - no more nested ternary! 🎉

panentheos:
I wasn't able to test without the new json file, but the code looks good (pending the comments and lint fixes). Glad to see this simplification!

rtitle:
Since we're adding RStudio to the drop-down and we now have PDs, should we also remove this yellow banner from the RStudio frame?

image

@sharmatime

rtitle:
Also to summarize a standup conversation: let's wait for this Leo bug to be merged/released before releasing this UI change, since it's a pretty significant impact to RStudio experience: https://broadworkbench.atlassian.net/browse/IA-2244

gpcarr:
> Since we're adding RStudio to the drop-down and we now have PDs, should we also remove this yellow banner from the RStudio frame?
> 
> image
> 
> @sharmatime

just reached out to Adrian about this. will update accordingly

gpcarr:
> I wasn't able to test without the new json file, but the code looks good (pending the comments and lint fixes). Glad to see this simplification!

If you replace `terra-docker-versions.json` in line https://github.com/DataBiosphere/terra-ui/blob/dev/src/components/NewRuntimeModal.js#L469 with `gabriela-test-rstudio.json` you should be able to test with the new json if you're curious! i pushed that test json file to the google bucket the UI is getting the original file from

src/components/NewRuntimeModal.js
gpcarr: is labels an optional parameter here?

panentheos: Yes, and here's how you can find out: If you take a look at the source of the `create` function, it passes this whole object to `_.merge`. That does a recursive deep merge, so it will combine these labels with the default ones specified there. Since we no longer have any additional labels to add, we can omit the key and it will leave the defaults unchanged. (You can try out `_.merge` in the browser console if you want to see the exact behavior)

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -42,14 +42,7 @@ const ApplicationLauncher = _.flow(
...
       h(Fragment, [
-        application === 'RStudio' && h(PlaygroundHeader, [
-          'This feature is in early development. Your files are saved on your cloud environment but not to your workspace. We encourage you to frequently ',
-          h(Link, {
-            href: 'https://support.terra.bio/hc/en-us/articles/360037269472#h_822db925-41fa-4797-b0da-0839580a74da',
-            ...Utils.newTabLinkProps
-          }, ['back up your files manually']),
-          '.'
-        ]),
+        application === 'RStudio',
```
panentheos: This expression was used to control the visibility of the `PlaygroundHeader` via the `&&`, it doesn't make sense on its own. So if we're removing the content entirely, we should remove this expression as well.

gpcarr:
@panentheos I pushed all the changes for the PR feedback I received and this PR is now ready for final review

ehanna4 (approved):
LGTM, the integration test failures do not appear to be relevant to this change so I don't believe should be a blocker to merge.

src/components/NewRuntimeModal.js
```diff
@@ -453,6 +449,22 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
     return _.find(({ imageType }) => _.includes(imageType, ['Jupyter', 'RStudio']), runtimeDetails?.runtimeImages)?.imageUrl
   }
 
+  getCurrentMountDirectory(currentRuntimeDetails) {
+    if (currentRuntimeDetails?.labels.tool) {
+      return currentRuntimeDetails?.labels.tool === 'RStudio' ? '/home/rstudio' : '/home/jupyter-user/notebooks'
+    } else {
+      return ''
```
ehanna4: In js I think the convention is using "null" or "undefined" in a case like this rather than ''

src/components/NewRuntimeModal.js
```diff
@@ -453,6 +449,22 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
...
+    }
+  }
+
+  getAboutPDWording(currentRuntimeDetails) {
+    if (this.getCurrentMountDirectory(currentRuntimeDetails) === '') {
```
ehanna4: Then rather than checking `this.getCurrentMountDirectory(currentRuntimeDetails) === ''` you can just do `if this.getCurrentMountDirectory(currentRuntimeDetails)`

src/components/NewRuntimeModal.js
```diff
@@ -504,15 +516,15 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
...
+    const { deleteDiskSelected, currentPersistentDiskDetails, currentRuntimeDetails } = this.state
     return h(Fragment, [
       h(RadioBlock, {
         name: 'delete-persistent-disk',
         labelText: 'Keep persistent disk, delete application configuration and compute profile',
         checked: !deleteDiskSelected,
         onChange: () => this.setState({ deleteDiskSelected: false })
       }, [
-        p(['Please save your analysis data in the directory ', code({ style: { fontWeight: 600 } }, ['/home/jupyter-user/notebooks']), ' to ensure it’s stored on your disk.']),
+        p(['Please save your analysis data in the directory ', code({ style: { fontWeight: 600 } }, [this.getCurrentMountDirectory(currentRuntimeDetails)]), ' to ensure it’s stored on your disk.']),
```
ehanna4, line 518: If there is no `currentMountDirectory` this sentence seems like it would read a little confusingly and also not be accurate enough to be helpful: "Please save your analysis data in the directory   to ensure it’s stored on your disk." I think maybe this whole paragraph should be conditional on `getCurrentMountDirectory` returning a value

src/components/NewRuntimeModal.js
jdcanas: This var is displayed in a string, so I think we want it to be empty in the case there isn't a mount Dir so we dont display null or undefined. However, we could do something like `'[No mount dir]'` to be explicit. I think this is fine though.

src/components/NewRuntimeModal.js
jdcanas: I think it would look better to assign the return of this function to a `const` than call the function twice

gpcarr: i like the `const` idea, will update!

src/components/NewRuntimeModal.js
gpcarr: hmm in order for this paragraph to display, there would need to be a PD otherwise we wouldn't display it at all correct? this is in the delete PD flow which we only display for GCE VMs not dataproc clusters since we don't support PDs in dataproc.

src/components/NewRuntimeModal.js
```diff
@@ -452,7 +452,8 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
   getCurrentMountDirectory(currentRuntimeDetails) {
     const rstudioMountPoint = '/home/rstudio'
     const jupyterMountPoint = '/home/jupyter-user/notebooks'
-    return currentRuntimeDetails?.labels.tool ? (currentRuntimeDetails?.labels.tool === 'RStudio' ? rstudioMountPoint : jupyterMountPoint) : jupyterMountPoint + ' for Jupyter environments and ' + rstudioMountPoint + ' for RStudio environments'
+    const noMountDirectory = ''.concat(jupyterMountPoint, ' for Jupyter environments and ', rstudioMountPoint, ' for RStudio environments')
```
ehanna4: `${jupyterMountPoint} for Jupyter environments and ${rstudioMountPoint}  for RStudio environments`

Typically string-interpolation is way you'll see strings get smushed together in this repo.

src/components/NewRuntimeModal.js
breilly2: I tend to agree with what Emily is seeing. While Gabriela is right about the interaction of those conditions, they feel a little distant in the code for it to read clearly.

TOL: I wonder if having the disk mount point included in the `currentPersistentDiskDetails` or `currentRuntimeDetails` would help.

TOL, part 2: It's a little unsettling that the UI has to know what the mount point is instead of being told by leonardo, especially since it's different depending on the app. It would be great if leonardo could provide this information to the UI.

PR #2292, opened 2020-10-20
Release the kracken! (show Galaxy on AnVIL site in prod) [IA-2285]

ehanna4:
Verified that the Galaxy create/delete button and also the widget in the top only shows on AnVIL

Do not release until Leo is released at 4pm on 10/21/2020.


rtitle (approved):
Thank you! It looks good to me.

🚀

src/components/RuntimeManager.js
```diff
@@ -292,7 +293,7 @@ export default class RuntimeManager extends PureComponent {
     const app = currentApp(apps)
 
     return h(Fragment, [
-      app && div({ style: { ...styles.container, borderRadius: 5, marginRight: '1.5rem' } }, [
+      app && isAnvil() && div({ style: { ...styles.container, borderRadius: 5, marginRight: '1.5rem' } }, [
```
panentheos: I think we should omit the `isAnvil` condition here, to remind people that they have an ongoing cost even if they go to another site skin.

panentheos (approved):
See comment, then looks good. Tested locally with an override.

src/components/RuntimeManager.js
ehanna4: Are other resources shared between site skins? Would we still want the users to be able to launch Galaxy from a non-AnVIL skin?

panentheos: I think the idea was to only show the launch button on the Anvil skin to limit the visibility of the feature, but once an app is running, we should show it regardless to make sure people know what they're spending money on.

PR #2293, opened 2020-10-21
undo new link

ehanna4:
Baseline folks do not have alignment on the form yet so this needs to be pulled


breilly2 (approved):
👍  I didn't test, but this looks like a straight revert to me.

PR #2294, opened 2020-10-22
AS-522: disable spellcheck for input and output fields in workflow configs [risk: no]

davidangb:
This disables spellcheck in the input/output textareas when configuring a workflow.

When spellcheck is enabled, it gives disconcerting warnings to end users if your input/output is namespaced:
image (2)

Tested by running locally, plus inspecting via dev tools.

### DIFF
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -156,6 +156,9 @@ const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange,
               !readOnly ? h(DelayedAutocompleteTextArea, {
                 autosize: true,
                 'aria-label': name,
+                // without quoting both the key and value for spellcheck, it won't make it to the browser
+                // eslint-disable-next-line quote-props
+                'spellcheck': 'false',
                 placeholder: optional ? 'Optional' : 'Required',
                 value,
                 style: isFile ? { paddingRight: '2rem' } : undefined,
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -156,6 +156,9 @@ const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange,
               !readOnly ? h(DelayedAutocompleteTextArea, {
                 autosize: true,
                 'aria-label': name,
+                // without quoting both the key and value for spellcheck, it won't make it to the browser
+                // eslint-disable-next-line quote-props
+                'spellcheck': 'false',
```
panentheos: Quoting `'false'` makes sense, since booleans get handled differently, but it shouldn't be necessary to quote the key. I just verified that it works without the quotes locally, could you please double check?

davidangb: @panentheos thank you! You're totally correct. I tried `spellcheck: false` which didn't work,  then `'spellcheck': false` which also didn't work, and finally `'spellcheck': 'false'` which did work, and never went back to try `spellcheck: 'false'`. Good catch.

Shakespeared (approved):
With Brett's comment, LGTM tested locally

zarsky-broad:
According to [the React docs](https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes), you actually want it in camelcase, i.e. `spellCheck: false`, that way there are no quotes at all.

davidangb:
ty @zarsky-broad ... wow, so may iterations on this PR for such a small change! now to make integration tests pass ...

zarsky-broad (approved):

PR #2295, opened 2020-10-22
Changing Galaxy startup time text to "5-7 minutes" from "8-10"

almahmoud:
Small text change reflecting faster startup for Galaxy.


ehanna4:
Hi, double checking with @sharmatime this change is good to go from Product side?
I will run the integration test manually but should be no issue there - just a text change

ehanna4:
Speed up has not been confirmed and is likely not there currently due to a slow down on Leo side

rtitle:
FWIW, I just launched Galaxy in a fresh project (no preexisting k8s cluster) and it took ~11 minutes.

PR #2296, opened 2020-10-23
[IA-2301] Update Galaxy cost

gpcarr:
We increased the default number of nodes for a Galaxy app from 1 to 2 so we need the cost in the UI to reflect this. Having an additional node pool also means having an additional boot disk.

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -71,9 +71,8 @@ export const runtimeCost = ({ runtimeConfig, status }) => {
 
 export const getGalaxyCost = app => {
   // numNodes * price per node + diskCost
-  ({ kubernetesRuntimeConfig }) => {
-    return kubernetesRuntimeConfig.numNodes * hourlyKubernetesAppCost(app) + persistentDiskCost({ size: 250 + 10 + 100 + 100, status: 'Running' })
-  }
+  return app.kubernetesRuntimeConfig.numNodes * hourlyKubernetesAppCost(app) + persistentDiskCost({ size: 250 + 10 + 100 + 100, status: 'Running' })
+
   // 250Gb for the NFS disk, 10Gb for the postgres disk, and 200Gb for boot disks (1 boot disk per nodepool)
   // to do: retrieve the disk sizes from the app not just hardcode them
 }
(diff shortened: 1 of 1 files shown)


src/libs/runtime-utils.js
```diff
@@ -70,8 +70,9 @@ export const runtimeCost = ({ runtimeConfig, status }) => {
 }
 
 export const getGalaxyCost = app => {
-  return hourlyKubernetesAppCost(app) + persistentDiskCost({ size: 250 + 10 + 100, status: 'Running' })
-  // 250Gb for the NFS disk, 10Gb for the postgres disk, and 100Gb for the boot disk
+  // 2 * hourlyKubernetesAppCost because a Galaxy app has 2 nodepools by default
```
panentheos: How does this affect already-existing galaxy apps? Are those being updated to 2 nodepools as well, or will this show an inaccurate cost for them?

Regardless, I think we should ultimately try to surface this information in the app object, so we could compute it directly, rather than having to infer.

src/libs/runtime-utils.js
```diff
@@ -70,8 +70,9 @@ export const runtimeCost = ({ runtimeConfig, status }) => {
 }
 
 export const getGalaxyCost = app => {
-  return hourlyKubernetesAppCost(app) + persistentDiskCost({ size: 250 + 10 + 100, status: 'Running' })
-  // 250Gb for the NFS disk, 10Gb for the postgres disk, and 100Gb for the boot disk
+  // numNodes * price per node + diskCost
+  return app.kubernetesRuntimeConfig.numNodes * hourlyKubernetesAppCost(app) + persistentDiskCost({ size: 250 + 10 + 100 + 100, status: 'Running' })
```
ehanna4, line 74: There is only one old Galaxy app that exists and was made by a Galaxy team person so I think it's safe to make this change
*meaning that any user making an App now will get the 2 nodepools so this will be accurate for all but that 1 old style one

rtitle (approved):
Looks good to me, thanks.

Yeah there is only 1 app in the system now (made by a developer) so this should be safe, but in the future we can derive the cost from the returned app json.

PR #2297, opened 2020-10-28
convert a few class components to functions

zarsky-broad:
Co-Authored-By: Kai <19541449+Shakespeared@users.noreply.github.com>

Tested locally!
_Reeeeeeally_ recommend reviewing with whitespace hidden.

### DIFF
--- a/src/components/TrialBanner.js
+++ b/src/components/TrialBanner.js
@@ -1,4 +1,3 @@
-import _ from 'lodash/fp'
 import { useState } from 'react'
 import { a, div, h, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, Clickable } from 'src/components/common'
@@ -50,10 +49,8 @@ const messages = {
 }
 
 
-export const TrialBanner = _.flow(
-  Utils.withDisplayName('TrialBanner'),
-  Utils.connectStore(authStore, 'authState')
-)(({ authState: { isSignedIn, profile: { trialState }, acceptedTos } }) => {
+export const TrialBanner = () => {
+  const { isSignedIn, profile: { trialState }, acceptedTos } = Utils.useStore(authStore)
   const [finalizeTrial, setFinalizeTrial] = useState(false)
   const removeBanner = localStorage.getItem('removeBanner')
 
@@ -131,4 +128,4 @@ export const TrialBanner = _.flow(
       }, ['Confirm'])
     }, ['Click confirm to remove banner forever.'])
   ])
-})
+}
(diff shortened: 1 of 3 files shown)


src/components/TrialBanner.js
```diff
@@ -50,90 +50,85 @@ const messages = {
...
-export const TrialBanner = Utils.connectStore(authStore, 'authState')(class TrialBanner extends Component {
-  constructor(props) {
-    super(props)
-    this.state = {
-      finalizeTrial: false
-    }
-  }
+export const TrialBanner = _.flow(
+  Utils.withDisplayName('TrialBanner'),
+  Utils.connectStore(authStore, 'authState')
```
panentheos: Now that this is a function component, we can use the `useStore` hook instead of this wrapper. (And removing this should remove the need for `withDisplayName`)

src/libs/nav.js
```diff
@@ -43,15 +43,10 @@ export const goToPath = (...args) => {
   history.push({ pathname: getPath(...args) })
 }
 
-export class Redirector extends Component {
-  componentDidMount() {
-    const { pathname, search } = this.props
-    history.replace({ pathname, search })
-  }
+export const Redirector = ({ pathname, search }) => {
+  useOnMount(() => history.replace({ pathname, search }))
```
panentheos: Writing this as a single-expression arrow function means that the result of `replace` is returned. Since this happens to be `undefined`, it works as expected, but if it returned something else React would try to register it as a callback on unmount. Since that's not what we intend here, I think it would be clearer to write it like this to be clear we're not returning anything:
```suggestion
  useOnMount(() => {
    history.replace({ pathname, search })
  })
```

src/libs/notifications.js
```diff
@@ -34,107 +34,102 @@ export const notify = (type, title, props) => {
...
-    this.state = { modal: false, notificationNumber: 0 }
-  }
-
-
-  render() {
-    const { notificationState, id } = this.props
-    const { modal, notificationNumber } = this.state
+const NotificationDisplay = _.flow(
+  Utils.withDisplayName('NotificationDisplay'),
+  Utils.connectStore(notificationStore, 'notificationState')
```
panentheos: Same comment as above regarding `useStore`

src/components/TrialBanner.js
zarsky-broad: Wow, you can really tell I'm out of practice, heh.

panentheos (approved):
Looks good, tested locally with some instrumentation.

PR #2299, opened 2020-10-29
Taking a whack at another few components to allow hooks

Shakespeared:
I didn't have too much else going on today so for practice I refactored a few more components to be able to use hooks. I didn't give them withDisplayName because they seemed fairly ephemeral, but could add it back. 

Tested locally

### DIFF
--- a/src/pages/workspaces/workspace/workflows/FailureRerunner.js
+++ b/src/pages/workspaces/workspace/workflows/FailureRerunner.js
@@ -11,14 +11,15 @@ import { rerunFailuresStatus } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
 
-const ToastMessageComponent = Utils.connectStore(rerunFailuresStatus, 'status')(({ status: { done, text } }) => {
+const ToastMessageComponent = () => {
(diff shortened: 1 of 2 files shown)


src/pages/workspaces/workspace/workflows/FailureRerunner.js
```diff
@@ -12,17 +11,13 @@ import { rerunFailuresStatus } from 'src/libs/state'
...
-    const { status: { done, text } } = this.props
-
-    return div({ style: { padding: '0.5rem 0', display: 'flex', alignItems: 'center', fontSize: 14 } }, [
-      done ?
-        icon('success-standard', { size: 24, style: { color: colors.success(), marginRight: '1rem' } }) :
-        spinner({ style: { marginRight: '1rem' } }),
-      text
-    ])
-  }
+const ToastMessageComponent = Utils.connectStore(rerunFailuresStatus, 'status')(({ status: { done, text } }) => {
```
panentheos: Now that this is a function component, we can use the `useStore` hook instead of the wrapper.

panentheos (approved):
Looks good, tested locally.

PR #2300, opened 2020-11-02
wait for definitive auth status before sending events [BW-409]

panentheos:
This is a precursor to capturing and linking anonymous metrics events. It defers event logging until the auth status has 'settled', e.g. we know the whether the user is signed in and registered. We're not currently sending any events that would be affected by this, so there's no immediate effect, but once we open up the logic around page view events, this will ensure that the initial page view event gets correctly attributed if the user is already logged in.


zarsky-broad (approved):
Code looks good, didn't test, I'll defer to others who understand the use case better on that.

cjllanwarne (approved):

PR #2301, opened 2020-11-03
convert Register page to function component

panentheos:
Extracted from #2280


Shakespeared (approved):
LGTM but did not test registration flow

PR #2302, opened 2020-11-03
[IA-2293] fix cost

jdcanas:
- makes the number of preemptible workers separate from the number of workers
- updates the cost estimation to make these workers additive


panentheos:
I tried testing this by creating a cluster with 2 workers and 3 preemptibles, and it ended up in an Error state. Not sure if that's an unrelated issue, but I wanted to mention it.

jdcanas:
> I tried testing this by creating a cluster with 2 workers and 3 preemptibles, and it ended up in an Error state. Not sure if that's an unrelated issue, but I wanted to mention it.

Hmm.. i tested it initially with different numbers but the same concept and it worked, and just now with those numbers and it worked. Not too sure what happened with your's, but Worked for me (tm). I suspect something unrelated.

panentheos (approved):
Looks good, tested locally.

PR #2304, opened 2020-11-04
[IA-2282] make galaxy apps per workspace per project

jdcanas:
This PR makes it so that galaxy apps are per user per workspace per project, as opposed to per user per project. 

I tested this by modifying `Notebooks.js` line 355 such that the galaxy app button always shows up. I then created an app in two different workspaces and observed that the UI treated them as two separate apps (separate status transitions, deleting one doesn't delete the other, etc). 

Had to do some heavy lifting for this.


rtitle (approved):
Nice that was easy! It looks good to me.

PR #2305, opened 2020-11-05
[AS-527] add baseline form to data explorer

Shakespeared:
Adds link to baseline google form to request access to their data instead of a link to terra support, and changes the associated description text.

### DIFF
--- a/src/components/PrivateDataExplorer.js
+++ b/src/components/PrivateDataExplorer.js
@@ -103,10 +103,10 @@ export default _.flow(
           p([
             `Thank you for your interest in the Baseline Health Study data. Baseline data is currently only being shared with our partner organizations at Duke and Stanford.
             If you are a researcher at one of our partner organizations, please reach out to your institutional contacts for information on how to obtain access.
-            If you are a researcher at our partner site and do not have an institutional contact, please complete the Terra Study Data Request form.`]),
-          p([`In the future, Baseline is planning to make this data available to qualified researchers outside of our partners. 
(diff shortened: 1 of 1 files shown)


src/components/PrivateDataExplorer.js
```diff
@@ -101,12 +101,22 @@ export default _.flow(
...
+            If you're interested in finding out when the data will be available to researchers outside of our partners, please complete the Terra Study Data Request Form.`
           ]),
-          p(['Please reach out to ', h(Link, { href: 'mailto:support@terra.bio' }, ['support@terra.bio']), ' if you have any additional questions.'])
+          p([
+            'Please fill out the ',
+            h(Link, {
+              href: 'https://forms.gle/d4vnyNvpagptcasf9',
+              ...Utils.newTabLinkProps
+            }, [
+              'Baseline Health Study Data Attestation form'
```
panentheos: TOL: Is this the same as the 'Terra Study Data Request Form' mentioned above? If so, we should probably use the same language to avoid confusion.

Shakespeared: TBH I'm not sure - this language is copied from the ticket, but now that you mention it the form name is from the previous PR to add the link, so I'm going to say you're probably right and it should be called the data request form.

Shakespeared: That does make the language somewhat redundant though

panentheos (approved):
Looks good, tested locally.

PR #2306, opened 2020-11-06
Move scheduled integration test run to 9am

breilly2:
We didn't adjust this for DST so it started running at 6am EST. Gary has requested that the test run happen at 9am EST.


panentheos (approved):
Looks good.

PR #2308, opened 2020-11-06
[CA-964] Update billing validation to include uppercase letters and underscores

s-rubenstein:
Tested by submitting the string `dm-AbcD4_-` which passes our validation and is URL safe, but does not pass googles. This should be considered part of our feature branch for the work to separate out google-project from billing-project and not merge until that code goes to production

### DIFF
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -44,8 +44,8 @@ const billingProjectNameValidator = existing => ({
   presence: { allowEmpty: false },
   length: { minimum: 6, maximum: 30 },
   format: {
-    pattern: /^[a-z]([a-z0-9-])*$/,
-    message: 'must start with a letter and can only contain lowercase letters, numbers, and hyphens.'
+    pattern: /^[A-z0-9-_]*$/,
+    message: 'can only contain letters, numbers, underscores and hyphens.'
   },
   exclusion: {
     within: existing,
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -44,7 +44,7 @@ const billingProjectNameValidator = existing => ({
   presence: { allowEmpty: false },
   length: { minimum: 6, maximum: 30 },
   format: {
-    pattern: /^[A-z0-9-_]*$/,
+    pattern: /(\w|-)+/,
     message: 'can only contain letters, numbers, underscores and hyphens.'
   },
   exclusion: {


s-rubenstein:
This matches the regex here: https://github.com/broadinstitute/rawls/blob/develop/model/src/main/scala/org/broadinstitute/dsde/rawls/StringValidationUtils.scala#L10

src/pages/billing/List.js
```diff
@@ -44,8 +44,8 @@ const billingProjectNameValidator = existing => ({
   presence: { allowEmpty: false },
   length: { minimum: 6, maximum: 30 },
```
dvoet, line 45: this size limitation should also be removed

s-rubenstein: I think we probably want to retain some size limitation for readability. I can check with Janki and UX to see if there's a product/design preference on this?

s-rubenstein: I've reached out to Janki. I realized if we remove the size limitation, we also need to start dealing with text overflow, which means there's almost certainly design considerations.

src/pages/billing/List.js
```diff
@@ -44,8 +44,8 @@ const billingProjectNameValidator = existing => ({
...
   format: {
-    pattern: /^[a-z]([a-z0-9-])*$/,
```
andy7i, line 47: did we decide no spaces to match rawls's regex?

andy7i (approved):
LGTM 👍  Interested in the answers to the questions, but not blocking. Updating the length could be a separate PR.

src/pages/billing/List.js
s-rubenstein: Yes, and also SAM requires that spaces are encoded, so we would need Rawl's regex to allow for `%` as well.

zarsky-broad:
Can we close this?

marctalbott:
> Can we close this?

@zarsky-broad We will still need this once our (very long lived) Rawls feature branch for one Google project per workspace merges

marctalbott (approved):

dmohs (approved):
The modified regular expression appears to be more permissive (and thus backwards-compatible) than the exiting one, so I'm inclined to suggest be merged now in anticipation of the PPW work.

src/pages/billing/List.js
zarsky-broad: This is exactly what `\w` is (EDIT: plus hyphen), should this just be `/(\w|-)+/`?

s-rubenstein: Done

andy7i: drive by comment: I like the brevity of it, but I think the readability suffers (especially if you don't know what `\w` means in regex)

s-rubenstein: I think it is probably okay with the \w. My thinking is that \w while not necessarily as clear at a glance, is something that is easily look-up-able if someone needs to futz with the regex for any reason. So the readability for folks is probably fine for when they need to make changes?

s-rubenstein:
@zarsky-broad is this set on your end? Do you want to give it another review?

zarsky-broad:
I'm good.

PR #2309, opened 2020-11-06
convert 'Resizable' to function component

Shakespeared:
In table.js convert the 'Resizable' component to a function component to allow hooks. 

tested in UI

### DIFF
--- a/src/components/ErrorWrapper.js
+++ b/src/components/ErrorWrapper.js
@@ -3,6 +3,7 @@ import { Component } from 'react'
 import { reportError } from 'src/libs/error'
 
 
+// No hook equivalent of componentDidCatch exists yet
 export default class ErrorWrapper extends Component {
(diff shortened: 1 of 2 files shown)


src/components/ErrorWrapper.js
```diff
@@ -13,9 +14,12 @@ export default class ErrorWrapper extends Component {
     this.state = { hasError: false }
   }
 
+  static getDerivedStateFromError(error) {
```
panentheos, line 17: I hadn't seen this lifecycle method before!

panentheos (approved):
Looks good, tested locally.

src/components/ErrorWrapper.js
Shakespeared: Isaac and I found it when we were looking up componentDidCatch!

PR #2310, opened 2020-11-12
add formatting for column namespaces [AS-503]

Shakespeared:
Added formatting for column namespaces when present

tested in UI 

(working on getting the wording for toast notification as specified in the ticket)

### DIFF
--- a/src/components/DataTable.js
+++ b/src/components/DataTable.js
@@ -191,6 +191,7 @@ export default ajaxCaller(class DataTable extends Component {
                   },
                   ..._.map(({ name }) => {
                     const thisWidth = theseColumnWidths[name] || 300
+                    const [, columnNamespace, columnName] = /(.+:)?(.+)/.exec(name)
                     return {
                       width: thisWidth,
                       headerRenderer: () => h(Resizable, {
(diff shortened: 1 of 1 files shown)


davidangb (approved):
lgtm!  tested by running locally.

src/components/DataTable.js
```diff
@@ -200,7 +201,10 @@ export default ajaxCaller(class DataTable extends Component {
                         }
                       }, [
                         h(Sortable, { sort, field: name, onSort: v => this.setState({ sort: v }) }, [
-                          h(HeaderCell, [name])
+                          !!columnNamespace && div({
+                            style: { fontStyle: 'italic', color: colors.dark(0.75), paddingRight: '0.2rem' }
+                          }, [columnNamespace]),
+                          h(HeaderCell, [columnName])
```
panentheos, line 207: If you have a long namespace, the text can overflow and bump into the resize handle. The `HeaderCell` handles overflow properly, so we should put both pieces of text inside it.

Shakespeared: Makes sense! Is it important to maintain the short circuit from ` !!columnNamespace &&` inside the headerCell? I wasn't sure how to make that work

panentheos: Yep, the `HeaderCell` can have multiple children, with one of them being conditional. (We'll probably want a `span` instead of a `div`, so the text flows properly)

Shakespeared: Ah, didn't realize `HeaderCell` could take children, that solves it, thanks!

PR #2312, opened 2020-11-13
display dockstore WDLs directly from github [BW-421]

panentheos:
For dockstore methods, the submission behavior was changed at some point to read the WDL code directly from github instead of using dockstore's duplicate copy. This can result in a mismatch in the UI if the dockstore copy is out of sync with the github reference.

This changes the UI to always load directly from github, to mirror what the backend does.


rsasch (approved):
LGTM 👍

aednichols (approved):
Looks good to me 👍 

A particular thank you from me because this is my bug in a way – I forgot to update the UI when I changed the backend in early 2020

I recommend including a release note in the JIRA ticket; took a stab:
> Fixed a bug that could cause a workflow imported from Dockstore to display an outdated version in the UI if a floating reference is used (e.g. "dev" or "master"). Workflows stored in Terra were not affected.

PR #2313, opened 2020-11-13
make profile pic infobox not super tiny

Shakespeared:
I noticed the infobox icon for changing your profile pic was eensy-weensy so I fixed it. 

Tested locally. Right now on local my profile pic itself is broken but I'm fairly sure that's unrelated.


panentheos (approved):
Looks good, tested locally. Thanks!

PR #2314, opened 2020-11-19
toast notification for new attribute namespace formatting [AS-503]

Shakespeared:
Updated the language in the import toast notification for PFB imports. Also addressed PR comments, which I thought had merged with my last PR. Tested locally.

### DIFF
--- a/src/components/ImportStatus.js
+++ b/src/components/ImportStatus.js
@@ -51,9 +51,9 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
 
     const successNotify = () => notify('success', 'Data imported successfully.',
       {
-        message: `Data import to workspace "${namespace} / ${name}" is complete. Please refresh the Data view.
-      Because this data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names. 
-      (I.e. "this.pfb:consent_codes" instead of "this.consent_codes.)`
+        message: `Data import to workspace "${namespace} / ${name}" is complete, please refresh the Data view.
+      Because this data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names, 
+      e.g. "this.pfb:consent_codes" instead of "this.consent_codes."`
       })
 
     const errorNotify = () => notify('error', 'Error importing PFB data.', message)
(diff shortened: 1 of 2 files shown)


src/components/ImportStatus.js
```diff
@@ -50,7 +50,11 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
     // once import servie is fully adopted, we can/should remove the avro-import status values.
 
     const successNotify = () => notify('success', 'Data imported successfully.',
-      { message: `Data import to workspace "${namespace} / ${name}" is complete. Please refresh the Data view.` })
+      {
+        message: `Data import to workspace "${namespace} / ${name}" is complete. Please refresh the Data view.
+      Because this data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names. 
+      (I.e. "this.pfb:consent_codes" instead of "this.consent_codes.)`
```
davidangb: grammar nits: no period after `attribute names`; don't capitalize `i.e.`, add closing double-quote after `this.consent_codes`, move final period outside the parenthesis:
```
Because this data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names 
(i.e. "this.pfb:consent_codes" instead of "this.consent_codes").`
```
I might even remove the parenthesis altogether:
```
Because this data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names,
i.e. "this.pfb:consent_codes" instead of "this.consent_codes."`
```
boy, now I'm questioning my grammar!

davidangb (changes requested):

src/components/ImportStatus.js
Shakespeared: Oop, good catch on that! How about this? I thing e.g. is actually more correct in this instance than i.e.

``Data import to workspace "${namespace} / ${name}" is complete, please refresh the Data view.
      Because this data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names, 
      e.g. "this.pfb:consent_codes" instead of "this.consent_codes".``

davidangb: good call on e.g. vs i.e., you're totally right there! This works for me. I thiiiiink the final period should be inside the quotation mark but that may be pedantic.

Shakespeared: Haha nope, you're totally right about that! My grammar is usually good so idk what happened here haha. I definitely appreciate the attention to detail.

davidangb (approved):
lgtm!

PR #2315, opened 2020-11-20
make tables/datatable function components

zarsky-broad:
Verified working locally.

All datatables now have selection type multiple, so removing the logic for selection type being single or absent.

_Definitely_ review with whitespace hidden.

### DIFF
--- a/src/components/DataTable.js
+++ b/src/components/DataTable.js
@@ -1,8 +1,8 @@
 import _ from 'lodash/fp'
-import { Component, createRef, Fragment } from 'react'
+import { Fragment, useEffect, useRef, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
-import { Checkbox, Clickable, Link, MenuButton, RadioButton, spinnerOverlay } from 'src/components/common'
+import { Checkbox, Clickable, Link, MenuButton, spinnerOverlay } from 'src/components/common'
 import { icon } from 'src/components/icons'
 import { ConfirmedSearchInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
@@ -11,14 +11,12 @@ import { ColumnSelector, GridTable, HeaderCell, paginator, Resizable, Sortable }
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { EditDataLink, EntityEditor, EntityRenamer, renderDataCell } from 'src/libs/data-utils'
-import { reportError } from 'src/libs/error'
+import { withErrorReporting } from 'src/libs/error'
 import { getLocalPref, setLocalPref } from 'src/libs/prefs'
 import * as StateHistory from 'src/libs/state-history'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
-
-const filterState = state => _.pick(['pageNumber', 'itemsPerPage', 'sort', 'activeTextFilter'], state)
 
 const entityMap = entities => {
   return _.fromPairs(_.map(e => [e.name, e], entities))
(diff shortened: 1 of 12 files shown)


src/components/DataTable.js
```diff
@@ -52,317 +89,256 @@ export default Utils.withCancellationSignal(class DataTable extends Component {
...
-    const nameWidth = theseColumnWidths['name'] || 150
+  const pageSelected = () => {
+    const entityKeys = _.map('name', entities)
+    const selectedKeys = _.keys(selected)
+    return entities.length && _.every(k => _.includes(k, selectedKeys), entityKeys)
+  }
 
-    const resetScroll = () => this.table.current.scrollToTop()
+  // Lifecycle
+  Utils.useOnMount(() => { loadData() })
```
panentheos: The effect below will already fire once on mount, so we can get rid of this one.

src/components/DataTable.js
```diff
@@ -52,317 +89,256 @@ export default Utils.withCancellationSignal(class DataTable extends Component {
...
-                        })
-                      }
+  useEffect(() => {
+    if (persist) {
+      setLocalPref(persistenceId, { columnWidths, columnState })
+    }
+  }, [columnWidths, columnState]) // eslint-disable-line react-hooks/exhaustive-deps
+
+  useEffect(() => {
+    table.current?.recomputeColumnSizes() // eslint-disable-line no-unused-expressions
```
panentheos, line 156: It seems unfortunate that this requires an eslint exception. Would adding the `allowShortCircuit` option to this rule suppress the message here?

panentheos, line 156: TOL: Since the underlying `GridTable` already measures cells on mount, I thought this effect would cause it to unnecessarily measure twice. It doesn't, but for a somewhat surprising reason: On first render, the `table.current` is undefined because the `entities` aren't loaded so the `GridTable` doesn't render, and therefore the effect does nothing. When the data loads and the `GridTable` does render, this effect is _not_ invoked again because we don't include `table.current` in the dependency array.

There are a couple confounding factors too: Why doesn't the exhaustive-deps rule throw a warning here? Also, if we did put the ref in the deps, then it definitely would measure twice, so how would we prevent that? Perhaps the `GridTable` should just delegate all measurement to its caller, so it could coordinate everything. Also, this component is currently the only user of `GridTable` anyway.

I'm not sure whether it's worth trying to 'fix' this as part of this PR, but perhaps a comment would be helpful to future readers.

panentheos (approved):

src/components/DataTable.js
zarsky-broad: It's a problem with this version of eslint not working with optional chaining. Once we update all the class components to hooks, I'll update us to the latest `react-scripts`, and that'll update eslint so we can get rid of these (there are a few)

zarsky-broad: Perhaps the thing to do is to make the `GridTable` recompute column sizes whenever the `columns` prop changes? That way we wouldn't have to expose it externally at all.

zarsky-broad: Also, I think if we convert hook in the `GridTable` to a layout effect, we wouldn't have to call the internal `ScrollSync` private method...?

panentheos: In order for `columns` to be the trigger, we'd need to memoize it to establish a stable reference, and its dependencies are a superset of what's required. Notably, it depends on the list of rows, which doesn't affect the column sizes. We could split the column width data out into a separate prop, but that would make the API less nice.

TBH I don't remember why we're calling `_onScroll`. It looks like it may be trying to initialize the `ScrollSync` internal state based on the saved position. But it doesn't look like the mechanism to save scroll positions is working correctly anyway.

zarsky-broad: I fixed the scroll restoration, take a look.

PR #2316, opened 2020-11-20
make some more components functional

zarsky-broad:
Tested locally.

The thing that makes this hard to review is that despite the fact that there aren't many major changes, in a lot of instances with the class components, helper or lifecycle methods are defined under the render method, and with function components the render is last, since that's what's returned.

We could get around this by making a named render function that's in the same place in the flow as it was before and then just returning that at the end, which would make the diffs easier to read, but that feels like an anti-pattern.


src/components/TopBar.js
```diff
@@ -95,53 +94,42 @@ const DropDownSection = ({ titleIcon, title, isOpened, onClick, children }) => {
...
     }
   }
 
-  hideNav() {
-    this.setState({ navShown: false, openUserMenu: false, openLibraryMenu: false, openSupportMenu: false })
+  const hideNav = () => {
+    setNavShown(false)
+    setOpenUserMenu(false)
+    setOpenLibraryMenu(false)
+    setOpenLibraryMenu(false)
```
panentheos: One of these should be `setOpenSupportMenu(false)`

src/libs/data-utils.js
```diff
@@ -67,128 +68,84 @@ export const EditDataLink = props => h(Link, {
...
-  }
-
-  async doDelete() {
-    const { onDismiss, onSuccess, namespace, name, selectedEntities, selectedDataType } = this.props
-    const { additionalDeletions } = this.state
+  const doDelete = async () => {
     const entitiesToDelete = _.concat(_.map(entityName => ({ entityName, entityType: selectedDataType }), selectedEntities), additionalDeletions)
 
-    this.setState({ deleting: true })
+    setDeleting(false)
```
panentheos: Should be `setDeleting(true)`

src/pages/billing/List.js
```diff
@@ -282,6 +274,7 @@ export const BillingList = _.flow(
         ]),
         creatingBillingProject && h(NewBillingProjectModal, {
           billingAccounts,
+          loadAccounts: () => this.loadAccounts(),
```
panentheos, line 277: Good catch!

panentheos (approved):
See comments then looks good. Tested locally.

PR #2318, opened 2020-11-24
Distinguish aborted vs failed workflows [BW-448]

cjllanwarne:
- Separates the `Aborted` from the general `Failed` status because these are significantly different outcomes. While I was there / additionally:

#### Job Status Page

- Updates the Status page to consider `Aborted` statuses.
- Updates the Status tooltip to _name_ the status icons (since Aborted/Failed and Queued/Submitted were otherwise confusingly similar icons)
- Updates the Status tooltip to only list actual statuses in that submission
- Updates the Status name to name Failure and Aborted statuses rather than just calling everything "Done"
- Updates the Status icon to highlight the most important status (eg a mix of Succeeded and Failed workflows implies a Failed submission)

#### Before:

image

#### After:

image

### DIFF
--- a/src/libs/icon-dict.js
+++ b/src/libs/icon-dict.js
@@ -36,6 +36,7 @@ const custom = _.curry((shape, { size, ...props }) => h(shape,
 const rotate = _.curry((rotation, shape, props) => shape(_.merge({ style: { transform: `rotate(${rotation}deg)` } }, props)))
 
 const iconDict = {
+  abort: fa(faBan),
   'angle-down': rotate(180, custom(angleUp)),
   'angle-left': rotate(-90, custom(angleUp)),
   'angle-right': rotate(90, custom(angleUp)),
@@ -101,8 +102,7 @@ const iconDict = {
   'view-cards': fa(faGripHorizontal),
   'view-list': custom(list),
   virus: fa(faVirus),
-  'warning-standard': fa(faExclamationTriangle),
-  'aborting-aborted': fa(faBan)
+  'warning-standard': fa(faExclamationTriangle)
 }
 
 export default iconDict
--- a/src/libs/icon-dict.js
+++ b/src/libs/icon-dict.js
@@ -101,7 +101,8 @@ const iconDict = {
   'view-cards': fa(faGripHorizontal),
   'view-list': custom(list),
   virus: fa(faVirus),
-  'warning-standard': fa(faExclamationTriangle)
+  'warning-standard': fa(faExclamationTriangle),
+  'aborting-aborted': fa(faBan)
 }
 
 export default iconDict
--- a/src/pages/workspaces/workspace/JobHistory.js
+++ b/src/pages/workspaces/workspace/JobHistory.js
@@ -76,7 +76,7 @@ const statusCell = workflowStatuses => {
           td(['Aborted']),
           td(styles.statusDetailCell, [abort || 0])
         ]): undefined
-      ].filter( element => element !== undefined ))
+      ].filter(element => element !== undefined))
     ])
   }, [
     div([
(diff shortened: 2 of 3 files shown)


src/libs/icon-dict.js
```diff
@@ -101,7 +101,8 @@ const iconDict = {
   'view-cards': fa(faGripHorizontal),
   'view-list': custom(list),
   virus: fa(faVirus),
-  'warning-standard': fa(faExclamationTriangle)
+  'warning-standard': fa(faExclamationTriangle),
+  'aborting-aborted': fa(faBan)
```
zarsky-broad: Perhaps just `abort`? Also, this list should be alphabetized.

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -44,31 +44,45 @@ const collapsedStatuses = _.flow(
...
-  const { succeeded, failed, running, submitted } = collapsedStatuses(workflowStatuses)
+  const { succeeded, failed, running, submitted, abort } = collapsedStatuses(workflowStatuses)
 
   return h(TooltipTrigger, {
     side: 'bottom',
     type: 'light',
     content: table({ style: { margin: '0.5rem' } }, [
       tbody({}, [
-        tr({}, [
+        submitted ? tr({}, [
```
zarsky-broad: This is where we'd use `&&` instead of the ternary. You could say `!!submitted` as well, for clarity.

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -44,31 +44,45 @@ const collapsedStatuses = _.flow(
...
-      ])
+          td(['Failed']),
+          td(styles.statusDetailCell, [failed])
+        ]): undefined,
+        abort ? tr({}, [
+          td(styles.statusDetailCell, [abortIcon()]),
+          td(['Aborted']),
+          td(styles.statusDetailCell, [abort])
+        ]): undefined
+      ].filter(element => element !== undefined))
```
zarsky-broad: If you're using `&&`, you won't have to filter.

Also, we'd use lodash for this if we were doing it, just to make sure we were returning rather than mutating.

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -221,8 +235,13 @@ const JobHistory = _.flow(
                 headerRenderer: () => h(HeaderCell, ['Status']),
                 cellRenderer: ({ rowIndex }) => {
                   const { workflowStatuses, status } = filteredSubmissions[rowIndex]
+                  const collapsedStatusList= collapsedStatuses(workflowStatuses)
```
zarsky-broad:
```suggestion
                  const collapsedStatusList = collapsedStatuses(workflowStatuses)
```

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -221,8 +235,13 @@ const JobHistory = _.flow(
...
+                  const statusString =
+                        'failed' in collapsedStatusList ? `Failed` :
+                          'abort' in collapsedStatusList ? 'Aborted' :
+                            status
```
zarsky-broad, line 242: @panentheos what do you think is the clearest way to do this?

zarsky-broad (changes requested):

cjllanwarne:
Pausing work on this one until #2325 merges since that PR adds some status elements and it'd be nice to get them all at the same time

src/pages/workspaces/workspace/JobHistory.js
dmohs: This is clear. We can change it later if we come up with some pattern that is obviously better, but this is quite nice.

dmohs (approved):

dmohs:
@cjllanwarne This seems like a nice change and the blocking PR has been merged. Shall we try to get this in?

kyuksel:
@cjllanwarne Closing the PR due to inactivity. Please re-open if needed.

PR #2319, opened 2020-11-24
Muscles's UI updates for the day

nmalfroy:
If you go through commit by commit, it's a bunch of small updates/fixes

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1188,7 +1188,22 @@ const driver = neo4j.driver(
 const Neo4j = signal => ({
   searchOntology: async (searchTerm) => {
     const session = driver.session();
-    const res = await session.run(`Match (n) where n.label[0] contains ("${searchTerm}") return n limit 10`);
+    const searchTermSanitized = searchTerm.replaceAll("'", "").toLowerCase();
+    const query = `
+      match (n) where (
+        toLower(n.label[0]) contains('${searchTermSanitized}') OR
+        n.definition contains('${searchTermSanitized}')
+      ) 
+      match (s) where (s)-[:SCO]-(n) 
+      return n.label as Name,
+             n.name as Code, 
+             n.definition as Definition,
+             n.uri as URI, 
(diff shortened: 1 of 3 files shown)


src/libs/ajax.js
```diff
@@ -1186,12 +1186,26 @@ const driver = neo4j.driver(
...
+             n.definition as Definition,
+             n.uri as URI, 
+             collect(n.label) as Synonyms,
+             collect(s.label) as Subclasses
+      limit 10
+    `
+    const res = await session.run(query)
     await session.close()
-    console.log(res);
+    console.log(res)
```
s-rubenstein, line 1208: We should probably clear out the console.logs

src/pages/ontology/OntologyModal.js
```diff
@@ -71,18 +71,18 @@ const AboutOntologySearch = () => {
 
 const OntologySearch = ({ searchTerm }) => {
   const [data, setData] = useState(undefined)
+  const [submittedSearchTerm, setSubmittedSearchTerm] = useState(undefined)
```
s-rubenstein, line 74: Rather than a new piece of state, I think we can use `prevState` here?

s-rubenstein: Talked offline

src/pages/ontology/OntologyContainer.js
```diff
@@ -13,7 +13,7 @@ const OntologyContainer = ({ title, setModalOpen, children }) => {
   return h(FooterWrapper, [
     h(TopBar, { title, href: Nav.getLink('root') }, [
       div({ style: { flexGrow: 1 } }),
-      div({ onClick: () => setModalOpen(true) }, ['hello'])
+      div({ onClick: () => setModalOpen(true), style: { marginRight: 10 } }, ['hello'])
       // TODO ontology search
```
Shakespeared: I think we can take out this TODO as well

Shakespeared (approved):
other than removing console logs etc, LGTM!

PR #2325, opened 2020-11-27
Hackathon: Cromwell dashboard [BW-447]

cjllanwarne:
Changes made during the cromwell dashboard hackathon project. 

Tested locally pointed at dev and alpha by opening submissions and checking how they looked.

#### Sample of the new submission details page:

image

#### Sample of the new workflow dashboard (alpha) page:

image


src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -1,12 +1,21 @@
...
-import { collapseStatus, failedIcon, runningIcon, statusIcon, submittedIcon, successIcon } from 'src/components/job-common'
+import {
+  collapseStatus,
+  failedIcon, makeSection,
+  makeStatusLine,
+  runningIcon,
+  statusIcon,
+  submittedIcon,
+  successIcon
+} from 'src/components/job-common'
```
zarsky-broad:
```suggestion
import {
  collapseStatus, failedIcon, makeSection, makeStatusLine, runningIcon, statusIcon, submittedIcon, successIcon
} from 'src/components/job-common'
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -20,6 +29,21 @@ import * as Utils from 'src/libs/utils'
 import { wrapWorkspace } from 'src/pages/workspaces/workspace/WorkspaceContainer'
 
 
+const CopyLink = ({ workflowId }) => {
```
zarsky-broad: I think I might prefer seeing this as a generic component in the general `components/common.js`, you'd just have to generalize the language and make it just take generic text. In that case, you'd probably want to make it take something like `{ text, style }` and then stick a `...style` on the end of the existing `style` object.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -127,10 +135,14 @@ const SubmissionDetails = _.flow(
...
-    h(Link, {
-      href: Nav.getLink('workspace-job-history', { namespace, name }),
-      style: { alignSelf: 'flex-start', display: 'flex', alignItems: 'center', padding: '0.5rem 0' }
-    }, [icon('arrowLeft', { style: { marginRight: '0.5rem' } }), 'Back to list']),
+    div({ style: { marginBottom: '1rem', display: 'flex', alignItems: 'center' } }, [
+      h(Link, {
+        href: Nav.getLink('workspace-job-history', { namespace, name }),
+        style: { alignSelf: 'flex-start', display: 'inline-flex', alignItems: 'center', padding: '0.5rem 0' }
+      }, [icon('arrowLeft', { style: { marginRight: '0.5rem' } }), 'Job History']),
+      span({ style: { whiteSpace: 'pre' } }, [' > ']),
```
zarsky-broad: Few things:
- How'd we decide to have double breadcrumbs? (We'll generally want to run all of this by UX)
- Inside a `display: 'flex'`, `span` and `div` behave the same, so you might as well use `div` since it's our generic go-to element.
- Instead of `whiteSpace: 'pre'`, try `margin: '0 0.5rem'` (and then you can remove the spaces)

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -234,10 +239,11 @@ const SubmissionDetails = _.flow(
               size: { basis: 125, grow: 0 },
               headerRenderer: () => h(Sortable, { sort, field: 'cost', onSort: setSort }, ['Run Cost']),
               cellRenderer: ({ rowIndex }) => {
-                return cost ? h(TextCell, [Utils.formatUSD(filteredWorkflows[rowIndex].cost)]) : 'N/A'
+                // handle undefined workflow cost as $0
+                return cost ? h(TextCell, [Utils.formatUSD(filteredWorkflows[rowIndex].cost || 0)]) : 'N/A'
```
zarsky-broad, line 216: Do we still want the ternary here?

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -234,10 +239,11 @@ const SubmissionDetails = _.flow(
...
               }
             }, {
-              size: { basis: 200 },
+              size: _.some(w => !_.isEmpty(w.messages), filteredWorkflows) ? { basis: 200 } : { basis: 100, grow: 0 },
```
zarsky-broad:
```suggestion
              size: _.some(({ messages }) => !_.isEmpty(messages), filteredWorkflows) ? { basis: 200 } : { basis: 100, grow: 0 },
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -246,16 +252,47 @@ const SubmissionDetails = _.flow(
...
             }, {
-              size: { basis: 375, grow: 0 },
+              size: { basis: 150 },
               headerRenderer: () => h(Sortable, { sort, field: 'workflowId', onSort: setSort }, ['Workflow ID']),
+              cellRenderer: ({ rowIndex }) => {
+                const { workflowId } = filteredWorkflows[rowIndex]
+                return workflowId ? [
+                  h(TooltipCell, { tooltip: workflowId }, [workflowId]),
+                  h(CopyLink, { workflowId })
+                ] : []
```
zarsky-broad:
```suggestion
                return workflowId && h(Fragment, [
                  h(TooltipCell, { tooltip: workflowId }, [workflowId]),
                  h(CopyLink, { workflowId })
                ])
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -246,16 +252,47 @@ const SubmissionDetails = _.flow(
...
+              }
+            }, {
+              size: { basis: 150, grow: 0 },
+              headerRenderer: () => div({ style: { display: 'flex' } }, ['Links']),
               cellRenderer: ({ rowIndex }) => {
                 const { workflowId, inputResolutions: [{ inputName } = {}] } = filteredWorkflows[rowIndex]
-                return h(TooltipCell, { tooltip: workflowId }, [
-                  inputName ? h(Link, {
+                return workflowId && [
```
zarsky-broad:
```suggestion
                return workflowId && h(Fragment, [
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -246,16 +252,47 @@ const SubmissionDetails = _.flow(
...
+                    h(Link, {
+                      ...Utils.newTabLinkProps,
+                      href: bucketBrowserUrl(`${bucketName}/${submissionId}/${inputName.split('.')[0]}/${workflowId}`),
+                      style: { padding: '.6rem' },
+                      tooltip: 'Execution directory'
+                    }, [div({ style: { display: 'flex', alignItems: 'center' } }, [
+                      icon('folder-open', { size: 18 })
+                    ])])
+                  ] : [])
+                ]
```
zarsky-broad:
```suggestion
                  inputName && h(Link, {
                    ...Utils.newTabLinkProps,
                    href: bucketBrowserUrl(`${bucketName}/${submissionId}/${inputName.split('.')[0]}/${workflowId}`),
                    style: { padding: '.6rem' },
                    tooltip: 'Execution directory'
                  }, [icon('folder-open', { size: 18 })])
                ])
              ])
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -265,6 +302,7 @@ const SubmissionDetails = _.flow(
   ])
 })
 
+
```
zarsky-broad:
```suggestion
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+import { centeredSpinner, icon } from 'src/components/icons'
+import {
+  collapseCromwellExecutionStatus,
+  failedIcon,
+  makeSection,
+  makeStatusLine,
+  runningIcon,
+  statusIcon, submittedIcon,
+  successIcon, unknownIcon
+} from 'src/components/job-common'
```
zarsky-broad:
```suggestion
import {
  collapseCromwellExecutionStatus, failedIcon, makeSection, makeStatusLine, runningIcon, statusIcon,
  submittedIcon, successIcon, unknownIcon
} from 'src/components/job-common'
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+const groupCallStatuses = callsObject => {
+  const statusCounts = {}
+  for (const callname in callsObject) {
+    for (const attempt in callsObject[callname]) {
+      const executionStatus = collapseCromwellExecutionStatus(callsObject[callname][attempt].executionStatus)
+      statusCounts[executionStatus] = (statusCounts[executionStatus] || 0) + 1
+    }
+  }
+  return statusCounts
+}
```
zarsky-broad: This should probably be rewritten using some combination of `_.reduce` and `_.map`. I'd be happy to talk through it with you, @panentheos can probably give you a more reasonable idea of how to start faster than I can, tho.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+
+const statusCell = workflowObject => {
+  const { succeeded, failed, running, submitted, ...others } = groupCallStatuses(workflowObject.calls)
+  console.log(others)
```
zarsky-broad:
```suggestion
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+    for (const attempt in callsObject[callname]) {
+      const executionStatus = collapseCromwellExecutionStatus(callsObject[callname][attempt].executionStatus)
+      statusCounts[executionStatus] = (statusCounts[executionStatus] || 0) + 1
+    }
+  }
+  return statusCounts
+}
+
+const statusCell = workflowObject => {
+  const { succeeded, failed, running, submitted, ...others } = groupCallStatuses(workflowObject.calls)
```
zarsky-broad:
```suggestion
const statusCell = ({ calls }) => {
  const { succeeded, failed, running, submitted, ...others } = groupCallStatuses(calls)
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+        ]): undefined,
+        _.map(other => tr({}, [
+          td(styles.statusDetailCell, [unknownIcon()]),
+          td([other]),
+          td(styles.statusDetailCell, [others[other]])
+        ]), Object.keys(others))
+      ].filter(element => element !== undefined))
+    ])
+
+  ])
```
zarsky-broad: There are a few things to talk about here:
- using the native js `.filter` and `Object.keys` rather than lodash
- unnecessary wrapping div
- ternaries rather than `&&`
- unneeded empty props for `tr`'s
- returning an array as a child, which makes React sad (see the console on the page)

Probably one to talk through in person, again ❤️

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+            'status',
+            'stderr',
+            'stdout',
+            'submission',
+            'submittedFiles:workflow',
+            'subworkflowId',
+            'workflowLog',
+            'workflowName',
+            'workflowRoot'
+          ]
```
zarsky-broad: This should probably be collapsed down to just a few lines, rather than a line per string.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+   */
+  const {
+    end,
+    failures,
+    start,
+    status,
+    workflowLog,
+    workflowName,
+    submittedFiles: { workflow: wdl } = {}
+  } = workflow
```
zarsky-broad: This can probably be collapsed down to one line.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+
+  const restructureFailures = _.map(({ message, causedBy }) => ({
+    message,
+    ...(!_.isEmpty(causedBy) ? { causedBy: simplifyDidNotStartFailures(restructureFailures(causedBy)) } : {})
+  }))
+
+  const simplifyDidNotStartFailures = failuresArray => {
+    const filtered = _.filter(({ message }) => !_.isEmpty(message) && !message.startsWith('Will not start job'), failuresArray)
+    const sizeDiff = !_.isEmpty(failuresArray) ? failuresArray.length - filtered.length : 0
```
zarsky-broad: This guard is unnecessary, since `filtered` is filtering `failuresArray`, so it'll always be the same length or shorter.
```suggestion
    const sizeDiff = failuresArray.length - filtered.length
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+    return [...filtered, ...newMessage]
+  }
+
+  return div({ style: { padding: '1rem 2rem 2rem', flex: 1, display: 'flex', flexDirection: 'column' } }, [
+    div({ style: { marginBottom: '1rem', display: 'flex', alignItems: 'center' } }, [
+      h(Link, {
+        href: Nav.getLink('workspace-job-history', { namespace, name }),
+        style: { alignSelf: 'flex-start', display: 'inline-flex', alignItems: 'center', padding: '0.5rem 0' }
+      }, [icon('arrowLeft', { style: { marginRight: '0.5rem' } }), 'Job History']),
+      span({ style: { whiteSpace: 'pre' } }, [' > ']),
```
zarsky-broad: Same comments as above

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+            tr([td({ style: styles.sectionTableLabel }, ['End:']), end ? td([Utils.makeCompleteDate(end)]) : 'N/A'])
+          ])
+        ])
+      ]),
+      makeSection('Links', [
+        div({ style: { display: 'flex' } }, [
+          h(Link, {
+            ...Utils.newTabLinkProps,
+            href: `${getConfig().jobManagerUrlRoot}/${workflowId}`,
+            style: { padding: '.6rem' },
```
zarsky-broad: Consider the following:
```suggestion
            style: { marginRight: '1.2rem' },
```
This will make the margin area non-clickable (padding is inside the element, so it adds to clickable area), as well as avoiding offsetting the first link from the section title.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+            tooltip: 'Execution directory'
+          }, [
+            div({ style: { display: 'flex', alignItems: 'center' } }, [
+              icon('folder-open', { size: 18 }),
+              span([' Execution Directory'])
+            ])
+          ]),
+          h(Link, { onClick: () => setShowLog(true), style: { padding: '.6rem' } }, [
+            icon('fileAlt', { size: 18 }),
+            ' View execution log'
```
zarsky-broad:
```suggestion
            icon('fileAlt', { size: 18, style: { marginRight: '0.5rem' } }),
            'View execution log'
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+          }, [
+            div({ style: { display: 'flex', alignItems: 'center' } }, [
+              icon('tasks', { size: 18 }),
+              span([' Job Manager'])
+            ])
+          ]),
+          h(Link, {
+            ...Utils.newTabLinkProps,
+            href: bucketBrowserUrl(`${bucketName}/${submissionId}/${workflowName}/${workflowId}`),
+            style: { padding: '.6rem' },
```
zarsky-broad: In general, when using a decimal, we tend to put the `0` in front of it.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,269 @@
...
+            ])
+          ]),
+          h(Link, { onClick: () => setShowLog(true), style: { padding: '.6rem' } }, [
+            icon('fileAlt', { size: 18 }),
+            ' View execution log'
+          ])
+        ])
+      ]),
+      makeSection('Call Statuses', [
+        div({ style: { display: 'flex' } }, [statusCell(workflow)])
```
zarsky-broad: I suspect this doesn't need the wrapper div? You could collapse the whole section into one line then. Also, we should handle when there are no call statuses, which I can see with some failed workflows.

zarsky-broad (changes requested):
Some thoughts! Happy to talk over them. Overall, great work, **especially** for a first go at UI work.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cjllanwarne: I think so. If `cost` hasn't loaded for the workflow yet, I think it's better to show nothing than state that it cost nothing?

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cjllanwarne: * IIRC, because 
  * Logically this set of breadcrumbs is *within* the "Job History" tab.
  * There's lots more space to show it fully on the main page
  * It has plenty of space to grow if (eg) we add "why isn't my stuff running", or "call x / subworkflow y / call y / inputs" sub-pages
  * It works as a nice title for the page contents 
  * It replaces the weirdly out-of-place `Back to list` link.
* I'll try these two suggestions. I'm not great at comparing layouts for aesthetic comparisons but if it looks reasonably similar with your changes, I'll make them

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: Mostly resolved

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: And reduced since we only need a handful of these entries

cjllanwarne: TOL one day it would be nice to bring the "constant time" requests and the "might be 10k or 100k values" requests separately to provide at least some of the information faster.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: This guard is to catch the case where `failuresArray` is undefined and thus calling `.length` on it would be an error.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: Switching to padding caused some layout issues, so I've reverted:

image

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
zarsky-broad: Ah, right, I was confusing myself.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
zarsky-broad: What if we defaulted it to an empty array where we declare the arg, then?

zarsky-broad:
You should update this from dev, I added a few autofixable lint rules that should help with cleanup here

src/libs/icon-dict.js
```diff
@@ -59,6 +59,7 @@ const iconDict = {
   copySolid: custom(copySolid),
   'copy-to-clipboard': fa(faClipboard),
   creditCard: fa(faCreditCard),
+  dashboard: fa(faTachometerAlt),
```
panentheos: Generally we should name these keys according to the underlying FA name, to improve traceability.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -200,22 +208,11 @@ const SubmissionDetails = _.flow(
...
-                }, ['View'])
-              }
-            }, {
-              size: { basis: 225, grow: 0 },
+              size: { basis: 225 },
               headerRenderer: () => h(Sortable, { sort, field: 'workflowEntity', onSort: setSort }, ['Data Entity']),
               cellRenderer: ({ rowIndex }) => {
                 const { workflowEntity: { entityName, entityType } = {} } = filteredWorkflows[rowIndex]
-                return h(TooltipCell, [entityName && `${entityName} (${entityType})`])
+                return h(TooltipCell, [entityName ? `${entityName} (${entityType})` : '--'])
```
panentheos: It might be good to make the placeholder value visually distinct from the actual values. E.g. make the `--` gray, or have it say `N/A` with italics.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -234,10 +231,11 @@ const SubmissionDetails = _.flow(
               size: { basis: 125, grow: 0 },
               headerRenderer: () => h(Sortable, { sort, field: 'cost', onSort: setSort }, ['Run Cost']),
               cellRenderer: ({ rowIndex }) => {
-                return cost ? h(TextCell, [Utils.formatUSD(filteredWorkflows[rowIndex].cost)]) : 'N/A'
+                // handle undefined workflow cost as $0
+                return cost ? h(TextCell, [Utils.formatUSD(filteredWorkflows[rowIndex].cost || 0)]) : 'N/A'
               }
             }, {
-              size: { basis: 200 },
+              size: _.some(({ messages }) => !_.isEmpty(messages), filteredWorkflows) ? { basis: 200 } : { basis: 100, grow: 0 },
```
panentheos, line 219: TOL: I see the intent, but making this dynamic is a little unusual, and could result in the layout shifting around after an auto-update. Perhaps there's a nicer way of handling these messages?

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -246,15 +244,44 @@ const SubmissionDetails = _.flow(
...
+              cellRenderer: ({ rowIndex }) => {
+                const { workflowId } = filteredWorkflows[rowIndex]
+                return workflowId && [
+                  h(TooltipCell, { tooltip: workflowId }, [workflowId]),
+                  h(CopyLink, { workflowId })
+                ]
+              }
+            }, {
+              size: { basis: 150, grow: 0 },
+              headerRenderer: () => div({ style: { display: 'flex' } }, ['Links']),
```
panentheos: nit: This div is unnecessary

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -246,15 +244,44 @@ const SubmissionDetails = _.flow(
...
                 const { workflowId, inputResolutions: [{ inputName } = {}] } = filteredWorkflows[rowIndex]
-                return h(TooltipCell, { tooltip: workflowId }, [
-                  inputName ? h(Link, {
+                return workflowId && h(Fragment, [
+                  h(Link, {
                     ...Utils.newTabLinkProps,
-                    href: bucketBrowserUrl(`${bucketName}/${submissionId}/${inputName.split('.')[0]}/${workflowId}`)
-                  }, [workflowId]) : workflowId
+                    href: `${getConfig().jobManagerUrlRoot}/${workflowId}`,
+                    style: { padding: '.6rem' },
```
panentheos: nit: If we add `display: 'flex'` here, we can remove the inner div wrapper (also below)

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -246,15 +244,44 @@ const SubmissionDetails = _.flow(
...
+                    tooltip: 'Job Manager'
```
panentheos: Since these buttons are just icons, we need to add an `'aria-label'` prop with the same text as the tooltip so screen readers can provide a proper description. (Someday we might automatically derive this from the tooltip if present, but that doesn't happen yet)

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
panentheos: I think this could be:
```suggestion
const groupCallStatuses = _.flow(
  _.values,
  _.flattenDepth(1),
  _.countBy(a => collapseCromwellExecutionStatus(a.executionStatus))
)
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,236 @@
...
+          const includeKey = [
+            'end', 'executionStatus', 'failures', 'start', 'status', 'submittedFiles:workflow', 'workflowLog', 'workflowName'
+          ]
+          const wf = await Ajax(signal).Workspaces.workspace(namespace, name).submission(submissionId).getWorkflow(workflowId, includeKey)
+          setWorkflow(wf)
+        }
+      })
+
+    initialize()
+  }, [workflow]) // eslint-disable-line react-hooks/exhaustive-deps
```
panentheos: I'm not a fan of using deps to implicitly trigger re-entry into this effect. I know we're currently doing it on the other page, but it feels fragile. I think it would be better to explicitly reschedule after loading, if applicable (see `useCloudEnvironmentPolling` for an example)

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
panentheos: It looks like this function and the one above are always called together. Perhaps we could combine them for clarity?

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,236 @@
...
+        href: Nav.getLink('workspace-job-history', { namespace, name }),
+        style: { alignSelf: 'flex-start', display: 'inline-flex', alignItems: 'center', padding: '0.5rem 0' }
+      }, [icon('arrowLeft', { style: { marginRight: '0.5rem' } }), 'Job History']),
+      div({ style: { margin: '0 0.5rem' } }, ['>']),
+      h(Link, {
+        href: Nav.getLink('workspace-submission-details', { namespace, name, submissionId }),
+        style: { alignSelf: 'flex-start', display: 'flex', alignItems: 'center', padding: '0.5rem 0' }
+      }, [`Submission ${submissionId}`]),
+      div({ style: { margin: '0 0.5rem' } }, ['>']),
+      div({ style: Style.elements.sectionHeader }, [`Workflow ${workflowId}`])
```
panentheos: Assuming we're ok with this new breadcrumb approach, it might be nice to extract out some of the shared bits between the pages.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,236 @@
...
+    ]),
+    _.isEmpty(workflow) ? centeredSpinner() : div({ style: { display: 'flex', flexWrap: 'wrap' } }, [
+      makeSection('Workflow Status', [
+        div({ style: { lineHeight: '24px' } }, [makeStatusLine(style => statusIcon(status, style), status)])
+      ]),
+      makeSection('Workflow Timing', [
+        table({ style: { marginTop: '0.3rem', lineHeight: '20px' } }, [
+          tbody([
+            tr([td({ style: styles.sectionTableLabel }, ['Start:']), start ? td([Utils.makeCompleteDate(start)]) : 'N/A']),
```
panentheos: The `N/A` should be inside the `td` to maintain proper table structure (next line also)

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,236 @@
...
+            tr([td({ style: styles.sectionTableLabel }, ['End:']), end ? td([Utils.makeCompleteDate(end)]) : 'N/A'])
+          ])
+        ])
+      ]),
+      makeSection('Links', [
+        div({ style: { display: 'flex' } }, [
+          h(Link, {
+            ...Utils.newTabLinkProps,
+            href: `${getConfig().jobManagerUrlRoot}/${workflowId}`,
+            style: { padding: '0.6rem' },
```
panentheos: nit: If we put `display: 'flex'` here, we can remove the inner div wrapper (below also)

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,236 @@
...
+            ])
+          ]),
+          h(Link, { onClick: () => setShowLog(true), style: { padding: '.6rem' } }, [
+            icon('fileAlt', { size: 18, style: { marginRight: '0.5rem' } }),
+            ' View execution log'
+          ])
+        ])
+      ]),
+      makeSection('Call Statuses', [
+        workflow.calls && false ? statusCell(workflow) : div({ style: { padding: '0.6rem' } }, ['No calls have been started by this workflow.'])
```
panentheos: We should either enable or remove this.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,236 @@
...
+          onClick: withErrorReporting('Error copying to clipboard', async e => {
+            e.stopPropagation() // this stops the collapse when copying
+            await clipboard.writeText(JSON.stringify(failures, null, 2))
+            setFailureCopied(true)
+            await Utils.delay(1500)
+            setFailureCopied(undefined)
+          })
+        }, [icon(failureCopied ? 'check' : 'copy-to-clipboard')])
+      ])
+    }, [h(ReactJson, {
```
panentheos, line 160: TOL: Since we're rearranging the structure of the JSON and adding pseudo-messages, I wonder if it might be confusing to present this as a JSON structure. We don't have another easy way to display nested data right now, but something to think about going forward.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cjllanwarne: Leaving for now. Let's reconsider if we can find a good example of an actual message in this field.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: Oops, this was me testing that it looked alright if the workflow had no calls. Wasn't intended to be committed 🤦

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: I kind of agree. We played around with what element to use for this for a while. Ultimately we couldn't find a nice "nested tree structure" element which really matched what we needed and displaying as an almost-"straight-from-Cromwell" json viewer for now felt like a reasonable compromise. Worth noting that (IIRC) the "copy to clipboard" link still gives you the original-from-Cromwell text

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
panentheos: Done

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
panentheos: Cleaned up.

panentheos (approved):
Looks good. Tested locally and during pairing.

src/components/common.js
```diff
@@ -525,3 +526,19 @@ export const WarningTitle = ({ children }) => {
     children
   ])
 }
+
+export const ClipboardButton = ({ text, onClick, ...props }) => {
```
zarsky-broad, line 530: TOL: do we want to put this in the other places that use it, like UriViewer etc? Or save that for another PR?

src/components/job-common.js
```diff
@@ -36,3 +60,46 @@ export const statusIcon = (status, style) => {
...
+}
+
+export const submissionDetailsBreadcrumbSubtitle = (namespace, workspaceName, submissionId) => {
+  return div({ style: { marginBottom: '1rem', display: 'flex', alignItems: 'center' } }, [
+    jobHistoryBreadcrumbPrefix(namespace, workspaceName),
+    div({ style: Style.elements.sectionHeader }, [`Submission ${submissionId}`])
+  ])
+}
+
+export const workflowDetailsBreadcrumbSubtitle = (namespace, workspaceName, submissionId, workflowId) => {
```
zarsky-broad, line 96: TOL: this and the submission details one above are only used in one place, do we want to move these definitions to those files?

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -200,22 +186,13 @@ const SubmissionDetails = _.flow(
...
-            }, {
-              size: { basis: 225, grow: 0 },
+              size: { basis: 225 },
               headerRenderer: () => h(Sortable, { sort, field: 'workflowEntity', onSort: setSort }, ['Data Entity']),
               cellRenderer: ({ rowIndex }) => {
                 const { workflowEntity: { entityName, entityType } = {} } = filteredWorkflows[rowIndex]
-                return h(TooltipCell, [entityName && `${entityName} (${entityType})`])
+                return h(TooltipCell, [
+                  entityName ? `${entityName} (${entityType})` : span({ style: { color: colors.dark(0.7) } }, ['--'])
+                ])
```
zarsky-broad: I think we don't need the tooltip when there's no entity:

```suggestion
                return !!entityName ?
                  h(TooltipCell, [`${entityName} (${entityType})`]) :
                  div({ style: { color: colors.dark(0.7) } }, ['--'])
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,181 @@
...
+
+  /*
+   * Page render
+   */
+  const { end, failures, start, status, workflowLog, workflowName, submittedFiles: { workflow: wdl } = {} } = workflow
+
+  const restructureFailures = failuresArray => {
+    const filtered = _.filter(({ message }) => !_.isEmpty(message) && !message.startsWith('Will not start job'), failuresArray)
+    const sizeDiff = failuresArray.length - filtered.length
+    const newMessage = sizeDiff > 0 ? [{ message: `${sizeDiff} jobs were queued in Cromwell but never sent to the cloud backend due to failures elsewhere in the workflow` }] : []
```
zarsky-broad: Just for line length:
```suggestion
    const newMessage = sizeDiff > 0 ? [{
      message: `${sizeDiff} jobs were queued in Cromwell but never sent to the cloud backend due to failures elsewhere in the workflow`
    }] : []
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -1,12 +1,12 @@
...
-import { div, h } from 'react-hyperscript-helpers'
+import { div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import * as breadcrumbs from 'src/components/breadcrumbs'
-import { Link, Select } from 'src/components/common'
+import { ClipboardButton, Link, Select } from 'src/components/common'
 import { centeredSpinner, icon } from 'src/components/icons'
 import { DelayedSearchInput } from 'src/components/input'
-import { collapseStatus, failedIcon, runningIcon, statusIcon, submittedIcon, successIcon } from 'src/components/job-common'
+import { collapseStatus, failedIcon, makeSection, makeStatusLine, runningIcon, statusIcon, submissionDetailsBreadcrumbSubtitle, submittedIcon, successIcon } from 'src/components/job-common'
```
zarsky-broad: I'd break this into multiple lines at this point:
```suggestion
import {
  collapseStatus, failedIcon, makeSection, makeStatusLine, runningIcon, statusIcon, submissionDetailsBreadcrumbSubtitle, submittedIcon, successIcon
} from 'src/components/job-common'
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,181 @@
...
+  _.countBy(a => collapseCromwellExecutionStatus(a.executionStatus))
+)
+
+const statusCell = ({ calls }) => {
+  const { succeeded, failed, running, submitted, ...others } = groupCallStatuses(calls)
+
+  const makeRow = (count, icon, text) => {
+    return !!count && div({ style: { display: 'flex', alignItems: 'center', marginTop: '0.25rem' } }, [
+      icon,
+      span([` ${count} ${text}`])
```
zarsky-broad: This doesn't need a wrapper:
```suggestion
      ` ${count} ${text}`
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -141,24 +122,29 @@ const SubmissionDetails = _.flow(
...
+                return h(Link, {
+                  href: Nav.getLink('workflow', { namespace, name, workflowNamespace, workflowName }),
+                  style: Style.noWrapEllipsis
+                }, [`${workflowNamespace}/${workflowName}`])
+              }],
+              [methodAccessible === false, () => {
+                return div({ style: { display: 'flex', alignItems: 'center' } }, [
+                  h(TooltipTrigger, {
+                    content: `${workflowNamespace}/${workflowName}` // TODO fix this width or wrap better
+                  }, [div({ style: Style.noWrapEllipsis }, [`${workflowNamespace}/${workflowName}`])]),
```
zarsky-broad: I think you might just want TooltipCell here?
```suggestion
                                    h(TooltipCell, [`${workflowNamespace}/${workflowName}`]), // TODO fix this width or wrap better
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,181 @@
...
+    breadcrumbs: props => breadcrumbs.commonPaths.workspaceDashboard(props),
+    title: 'Job History', activeTab: 'job history'
+  })
+)((props, ref) => {
+  const { namespace, name, submissionId, workflowId, workspace: { workspace: { bucketName } } } = props
+
+  /*
+   * State setup
+   */
+  const [workflow, setWorkflow] = useState({})
```
zarsky-broad: To allow us to use the standard pattern of undefined as not yet loaded:
```suggestion
  const [workflow, setWorkflow] = useState()
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,181 @@
...
+    loadWorkflow()
+    return () => {
+      clearTimeout(stateRefreshTimer.current)
+    }
+  })
+
+  /*
+   * Page render
+   */
+  const { end, failures, start, status, workflowLog, workflowName, submittedFiles: { workflow: wdl } = {} } = workflow
```
zarsky-broad: To allow us to use the standard pattern of undefined as not yet loaded:
```suggestion
  const { end, failures, start, status, workflowLog, workflowName, submittedFiles: { workflow: wdl } = {} } = workflow || {}
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,181 @@
...
+
+    return _.map(({ message, causedBy }) => ({
+      message,
+      ...(!_.isEmpty(causedBy) ? { causedBy: restructureFailures(causedBy) } : {})
+    }), simplifiedFailures)
+  }
+
+  return div({ style: { padding: '1rem 2rem 2rem', flex: 1, display: 'flex', flexDirection: 'column' } }, [
+    workflowDetailsBreadcrumbSubtitle(namespace, name, submissionId, workflowId),
+    _.isEmpty(workflow) ? centeredSpinner() : div({ style: { display: 'flex', flexWrap: 'wrap' } }, [
```
zarsky-broad: To allow us to use the standard pattern of undefined as not yet loaded:
```suggestion
    workflow === undefined ? centeredSpinner() : div({ style: { display: 'flex', flexWrap: 'wrap' } }, [
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,181 @@
...
+        ])
+      ]),
+      makeSection('Links', [
+        div({ style: { display: 'flex', marginTop: '0.5rem' } }, [
+          h(Link, {
+            ...Utils.newTabLinkProps,
+            href: `${getConfig().jobManagerUrlRoot}/${workflowId}`,
+            style: { display: 'flex' },
+            tooltip: 'Job Manager'
+          }, [icon('tasks', { size: 18 }), span([' Job Manager'])]),
```
zarsky-broad:
```suggestion
          }, [icon('tasks', { size: 18 }), ' Job Manager']),
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,181 @@
...
+          h(Link, {
+            ...Utils.newTabLinkProps,
+            href: bucketBrowserUrl(`${bucketName}/${submissionId}/${workflowName}/${workflowId}`),
+            style: { display: 'flex', marginLeft: '1rem' },
+            tooltip: 'Execution directory'
+          }, [icon('folder-open', { size: 18 }), span([' Execution Directory'])]),
```
zarsky-broad:
```suggestion
          }, [icon('folder-open', { size: 18 }), ' Execution Directory']),
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,181 @@
...
+          h(Link, {
+            onClick: () => setShowLog(true),
+            style: { display: 'flex', marginLeft: '1rem' }
+          }, [icon('fileAlt', { size: 18 }), span([' View execution log'])])
```
zarsky-broad:
```suggestion
          }, [icon('fileAlt', { size: 18 }), ' View execution log'])
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -0,0 +1,181 @@
+import _ from 'lodash/fp'
+import { Fragment, useRef, useState } from 'react'
+import { div, h, span } from 'react-hyperscript-helpers'
```
zarsky-broad: If you accept all span removals on this page, for your batch commit convenience:
```suggestion
import { div, h } from 'react-hyperscript-helpers'
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -1,12 +1,12 @@
 import _ from 'lodash/fp'
 import { Fragment, useEffect, useState } from 'react'
-import { div, h } from 'react-hyperscript-helpers'
+import { div, h, span } from 'react-hyperscript-helpers'
```
zarsky-broad: If you accept all span removals on this page, for your batch commit convenience:
```suggestion
import { div, h } from 'react-hyperscript-helpers'
```

zarsky-broad (approved):
Some (non-functional) comments that I'd like to see changed, but generally looks good!

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
zarsky-broad: (Generally with these I'd tell you to add marginRight to the icon rather than putting a space in the string, since spaces can often be collapsed when rendered, but it's not a huge deal.)

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cjllanwarne: Happy to make the change. OOI, what's the difference between `TooltipCell` vs `TooltipTrigger`?

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: Doing this caused problems later on due to destructuring on the `undefined`:
```
const { end, failures, start, status, workflowLog, workflowName, submittedFiles: { workflow: wdl } = {} } 
    = workflow
```

I could get around that with a ternary guard on the destructuring, but this now feels a little... obtuse? Interested in your thoughts:
```
const { end, failures, start, status, workflowLog, workflowName, submittedFiles: { workflow: wdl } = {} } 
    = workflow ? workflow : {}
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: Oops, I should probably have read this before making my comment above

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: This was answered in your next comment 🤦‍♂️

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: I thought `isEmpty(workflow)` was a strictly more safe variant of `workflow === undefined` because it ALSO covered an empty object as well as the various javascript values for "doesn't exist yet"?

src/components/job-common.js
cjllanwarne: IMO the benefits of declaring them both in the same place here are:
* If one of them changes, it's more clear that the others should change too
* If we copy/paste one to generate another subtitle link, it might be easier to refactor out some commonality

I don't see a huge downside to switching over to this file to see the method declarations so I'm inclined to leave them here.

src/components/common.js
```diff
@@ -527,11 +527,15 @@ export const WarningTitle = ({ children }) => {
...
 }
 
-export const ClipboardButton = ({ text, onClick, ...props }) => {
+export const ClipboardButton = ({ text, onClick, style, ...props }) => {
   const [copied, setCopied] = useState(false)
   return h(Link, {
     ...props,
     tooltip: 'Copy to clipboard',
+    style: {
+      color: colors.accent(),
```
cjllanwarne: @zarsky-broad only potentially controversial change since your last review: when I updated the URI viewer to re-use ClipboardButton I noticed that it was applying this color style. It felt like a reasonable idea to make this the default for all ClipboardButtons. Is this (plus the `...style` below) an appropriate way to do this?

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
zarsky-broad: Tooltip cell is basically a tooltip trigger wrapping around a div with no wrap ellipsis on it, with the same text in the tooltip and the div.

So it's exactly what you already were doing, just pre-made.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
zarsky-broad: Sure, the `=== undefined` is our standard pattern for "if this hasn't been loaded", though. The fact that `isEmpty` handles undefined is a convenience, it's not something we should consider semantic.

src/components/common.js
zarsky-broad: If you look at the definition of `Link`, it should be `colors.accent` by default, so you shouldn't even have to put this here, and then you can avoid destructuring style from props at all.

src/components/job-common.js
zarsky-broad: fair enough!

zarsky-broad (approved):
Left a couple comments, but looking good! Thanks for sticking with this!

PR #2326, opened 2020-11-30
Show bucket region during submit confirmation [BW-449]

cjllanwarne:
Tested running locally pointed at the `alpha` environment.

Example submit confirmation dialog:

~removed~ (see updates in comment below)

NB: styling suggestions welcome!

### DIFF
--- a/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
+++ b/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
@@ -120,9 +120,9 @@ const LaunchAnalysisModal = ({
       location ? h(Fragment, [span({ style: { marginRight: '0.5rem' } }, [flag]),
         span({ style: { marginRight: '0.5rem' } }, [regionDescription]),
         h(InfoBox, [
-          p(['Cromwell will instruct its job runners to delocalize call outputs to your workspace bucket in this region.']),
+          p(['Output (and intermediate) files will be written to your workspace bucket in this region.']),
           p(['Tasks within your workflow might additionally move or copy the data elsewhere. You should carefully vet the workflows you run if region requirements are a concern.']),
-          p(['Note that metadata about this run will also be stored in the US.'])
+          p(['Note that metadata about this run will be stored in the US.'])
         ])]) : 'Loading...'
     ])
   ])
--- a/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
+++ b/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
@@ -1,12 +1,15 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
-import { b, div, h, wbr } from 'react-hyperscript-helpers'
+import { b, div, h, p, span, wbr } from 'react-hyperscript-helpers'
 import { ButtonPrimary, CromwellVersionLink } from 'src/components/common'
 import { spinner } from 'src/components/icons'
 import Modal from 'src/components/Modal'
+import { InfoBox } from 'src/components/PopupTrigger'
+import { regionInfo } from 'src/components/region-common'
 import { Ajax } from 'src/libs/ajax'
 import { launch } from 'src/libs/analysis'
 import colors from 'src/libs/colors'
+import { withErrorReporting } from 'src/libs/error'
 import * as Utils from 'src/libs/utils'
 import {
   chooseRows, chooseSetComponents, chooseSets, processAll, processAllAsSet, processMergedSet
@@ -15,15 +18,24 @@ import {
 
 const LaunchAnalysisModal = ({
   onDismiss, entityMetadata,
-  workspace, workspace: { workspace: { namespace, name } }, processSingle,
+  workspace, workspace: { workspace: { namespace, name: workspaceName, bucketName } }, processSingle,
   entitySelectionModel: { type, selectedEntities, newSetName },
   config, config: { rootEntityType }, useCallCache, deleteIntermediateOutputFiles, onSuccess
 }) => {
   const [launching, setLaunching] = useState(undefined)
   const [message, setMessage] = useState(undefined)
   const [launchError, setLaunchError] = useState(undefined)
-
+  const [bucketLocation, setBucketLocation] = useState({})
   const signal = Utils.useCancellation()
+
+  Utils.useOnMount(() => {
+    const loadBucketLocation = withErrorReporting('Error loading bucket location', async () => {
+      const { location, locationType } = await Ajax(signal).Workspaces.workspace(namespace, workspaceName).checkBucketLocation(bucketName)
+      setBucketLocation({ location, locationType })
+    })
+
+    loadBucketLocation()
+  })
 
   const doLaunch = async () => {
     try {
@@ -32,7 +44,7 @@ const LaunchAnalysisModal = ({
         [processSingle, () => ({})],
         [type === processAll, async () => {
           setMessage('Fetching data...')
-          const selectedEntityNames = _.map('name', await Ajax(signal).Workspaces.workspace(namespace, name).entitiesOfType(rootEntityType))
+          const selectedEntityNames = _.map('name', await Ajax(signal).Workspaces.workspace(namespace, workspaceName).entitiesOfType(rootEntityType))
           return { selectedEntityType: rootEntityType, selectedEntityNames }
         }],
         [type === chooseRows || type === chooseSets, () => ({ selectedEntityType: rootEntityType, selectedEntityNames: _.keys(selectedEntities) })],
@@ -44,7 +56,7 @@ const LaunchAnalysisModal = ({
         [type === chooseSetComponents, () => ({ selectedEntityType: baseEntityType, selectedEntityNames: _.keys(selectedEntities) })],
         [type === processAllAsSet, async () => {
           setMessage('Fetching data...')
-          const selectedEntityNames = _.map('name', await Ajax(signal).Workspaces.workspace(namespace, name).entitiesOfType(baseEntityType))
+          const selectedEntityNames = _.map('name', await Ajax(signal).Workspaces.workspace(namespace, workspaceName).entitiesOfType(baseEntityType))
           return { selectedEntityType: baseEntityType, selectedEntityNames }
         }]
       )
@@ -71,6 +83,8 @@ const LaunchAnalysisModal = ({
     [type === processMergedSet, () => _.flow(mergeSets, _.uniqBy('entityName'))(selectedEntities).length]
   )
   const wrappableOnPeriods = _.flow(str => str?.split(/(\.)/), _.flatMap(sub => sub === '.' ? [wbr(), '.'] : sub))
+  const { location, locationType } = bucketLocation
+  const { flag, regionDescription } = regionInfo(location, locationType)
 
   return h(Modal, {
     title: !launching ? 'Confirm launch' : 'Launching Analysis',
@@ -100,6 +114,16 @@ const LaunchAnalysisModal = ({
     ]),
     div({ style: { color: colors.danger(), overflowWrap: 'break-word' } }, [
       h(Fragment, wrappableOnPeriods(launchError))
+    ]),
+    div(['Output files will be saved as workspace data in:']),
+    div({ style: { margin: '1rem' } }, [
+      location ? h(Fragment, [span({ style: { marginRight: '0.5rem' } }, [flag]),
+        span({ style: { marginRight: '0.5rem' } }, [regionDescription]),
+        h(InfoBox, [
+          p(['Cromwell will instruct its job runners to delocalize call outputs to your workspace bucket in this region.']),
+          p(['Tasks within your workflow might additionally move or copy the data elsewhere. You should carefully vet the workflows you run if region requirements are a concern.']),
+          p(['Note that metadata about this run will also be stored in the US.'])
+        ])]) : 'Loading...'
     ])
   ])
 }
--- a/src/components/region-common.js
+++ b/src/components/region-common.js
@@ -1,0 +1,27 @@
+// Get a { flag: ..., countryName: ... } object representing a google locationType/location input.
+// 'flag' will always be defined (even if it's a question mark.
+// 'regionDescription' is the same as location when locationType is 'multi-region', or a country name when locationType is 'region'.
+export const regionInfo = (location, locationType) => {
+  switch (locationType) {
+    case 'multi-region':
+      switch (location) {
+        case 'US':
+          return { flag: '🇺🇸', regionDescription: `${locationType}: ${location}` }
+        case 'EU':
+          return { flag: '🇪🇺', regionDescription: `${locationType}: ${location}` }
+        case 'ASIA':
+          return { flag: '🌏', regionDescription: `${locationType}: ${location}` }
+        default:
+          return { flag: '❓', regionDescription: `${locationType}: ${location}` }
+      }
+    case 'region':
+      switch (location) {
+        case 'EUROPE-NORTH1':
+          return { flag: '🇫🇮', regionDescription: `${locationType}: ${location} (Finland)` }
+        default:
+          return { flag: '❓', regionDescription: `${locationType}: ${location}` }
+      }
+    default:
+      return { flag: '❓', regionDescription: `${locationType}: ${location}` }
+  }
+}
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -544,6 +544,13 @@ const Workspaces = signal => ({
         return res.json()
       },
 
+      checkBucketLocation: async bucket => {
+        const res = await fetchBuckets(`storage/v1/b/${bucket}?fields=location%2ClocationType`,
+          _.merge(authOpts(await saToken(namespace)), { signal }))
+
+        return res.json()
+      },
+
       details: async fields => {
         const res = await fetchRawls(`${root}?${qs.stringify({ fields }, { arrayFormat: 'comma' })}`, _.merge(authOpts(), { signal }))
         return res.json()


src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -63,6 +82,10 @@ export default Utils.withCancellationSignal(class LaunchAnalysisModal extends Co
       ]),
       div({ style: { color: colors.danger(), overflowWrap: 'break-word' } }, [
         h(Fragment, wrappableOnPeriods(launchError))
+      ]),
+      location && locationType && span({}, [
+        'Data produced by this run will be stored in the following workspace bucket: ',
```
aednichols: Users can specify custom output locations, right?

The following is more concise and possibly more accurate by not over-promising:
```suggestion
        'Workspace bucket:',
```

cjllanwarne: I didn't think that they could? At least not in Terra?

cjllanwarne: (at least, at least not via this submit dialog?)

aednichols: I could write a task that uploads a localized file to an arbitrary bucket, I'm pretty sure I've heard of people doing this or similar somewhere

It's not likely to happen in Terra, OTOH making statements about where data always goes is a commitment

src/components/RegionFlaggedContent.js
```diff
@@ -0,0 +1,43 @@
...
+import { withErrorReporting } from 'src/libs/error'
+import * as clipboard from 'clipboard-polyfill/text'
+
+
+const RegionFlaggedContent = ({ location, locationType, text: content, ...props }) => {
+
+
+  const [copied, setCopied] = useState()
+  const { flag, aka } = (locationType === 'multi-region' && location === 'US') ? { flag: '🇺🇸', aka: 'USA' } :
+    (location === 'EUROPRE-NORTH1') ? { flag: '🇫🇮', aka: 'Finland' } :
```
aednichols:
```suggestion
    (location === 'EUROPE-NORTH1') ? { flag: '🇫🇮', aka: 'Finland' } :
```

cjllanwarne:
From conversation with Joy:

* Show the region name, not the bucket name on this page.
* No need for the copy-to-clipboard link.
* [Optional] Show the region flag and bucket name on the "Outputs" tab instead

src/components/RegionFlaggedContent.js
```diff
@@ -0,0 +1,43 @@
...
+import * as clipboard from 'clipboard-polyfill/text'
+
+
+const RegionFlaggedContent = ({ location, locationType, text: content, ...props }) => {
+
+
+  const [copied, setCopied] = useState()
+  const { flag, aka } = (locationType === 'multi-region' && location === 'US') ? { flag: '🇺🇸', aka: 'USA' } :
+    (location === 'EUROPE-NORTH1') ? { flag: '🇫🇮', aka: 'Finland' } :
+      { flag: undefined, aka: undefined }
```
evanbernstein: TOL. I worry about the tech debt of this block of code. The flags are nice and cute, but also now maintenance that we are creating for ourselves, that we have to remember to fix everytime we add a new region (which isn't in/near this code to remind us). I'd like a cleaner "failure" state that at least displays the text of the region. It might not be pretty, but would still display something to the user rather than nothing.

aednichols: +1

IMO we should display only the region, with `monospaced font styling`

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -63,6 +86,12 @@ export default Utils.withCancellationSignal(class LaunchAnalysisModal extends Co
       ]),
       div({ style: { color: colors.danger(), overflowWrap: 'break-word' } }, [
         h(Fragment, wrappableOnPeriods(launchError))
+      ]),
+      location && locationType && span({}, [
+        'Data produced by this run will be output to the following workspace bucket',
```
aednichols: I still believe we should reword this to make the promise less specific and drop the disclaimer.

```suggestion
        'Data written to the workspace bucket resides in region `region-name`'
```

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -47,7 +66,11 @@ export default Utils.withCancellationSignal(class LaunchAnalysisModal extends Co
             this.doLaunch()
           }
         }, ['Launch']) :
-        h(ButtonPrimary, { onClick: onDismiss }, ['OK'])
+        h(ButtonPrimary, { onClick: onDismiss }, ['OK']),
+      suffixContent: [div({ style: { fontSize: 10 } }, [
+        sup(['*']),
+        'Note: External actions (or workflow tasks) might act to move or copy this data. It is not guaranteed to remain only in this bucket. Metadata produced by this run is stored in the US.'
+      ])]
```
aednichols: Based on my experience working with our end-users, people will fall into three categories:

80% people who don't know they can do this
15% people who are aware this is a possibility but never do it
5% people who actually write data somewhere custom

For the 80% who don't know what this means, it kind of sounds like Terra is failing to protect their data from mysterious "external actions" that could move files somewhere else at any time ("not guaranteed to remain only in this bucket").

```suggestion
```

aednichols (changes requested):
Sorry to re-gull (did I just coin that?) but I think we have to be very careful with language that's going to be read (and attempted to be understood) by the entire cross-section of the workflows userbase

cjllanwarne:
Updates following feedback (inc. from @aednichols, @evanbernstein and Joy):

#### Launch modal:

* We now list the region rather than the bucket, as per the original ticket
* Wording has been updated to be closer in wording (and spirit) to the contents of the "Outputs" tab

image


* Hovering over the help tooltip icon yields:

image

* Unrecognized regions will now still show up (eg if I "unrecognize" the US multi-region:

image

#### Outputs tab

* Added new content in the form of the bucket region, and made it collapsed by default to avoid taking up too much screen space. 
* Made `Files` into a link back to the Data pages (since I didn't realize initially that this was what it was referencing)

image

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
cjllanwarne: I'm reluctant to switch to this wording.

1. I think 'Data written to the workspace bucket resides in region X' ignores the most important part of the story w.r.t. submitting a workflow - namely that this workflow will do some writing of data to that bucket.
1. I think "your outputs will be saved to X region" is true, even if it's not the full story.
1. I prefer optimizing the message for the common case and providing disclaimers for power users rather than making the message less useful in order to cover edges cases which the 80% will never have to worry about.

I'll also make the weak-sauce "resort to prior art" argument that the "Outputs" tab already has a form of words which is essentially identical to the statement being made here.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
cjllanwarne: Have re-worded. Let me know if the new form rings truer.

src/pages/workspaces/workspace/Data.js
```diff
@@ -1114,5 +1114,11 @@ export const navPaths = [
     path: '/workspaces/:namespace/:name/data',
     component: WorkspaceData,
     title: ({ name }) => `${name} - Data`
+  },
+  {
+    name: 'workspace-data-bucket',
+    path: '/workspaces/:namespace/:name/data/:initialDataType',
```
cjllanwarne: Doing this felt hacky. @panentheos / @zarsky-broad any ideas on ways to do it better?

zarsky-broad: We have an example somewhere of loading making a url including a query with something specific to load first and then removing that from the shown url once we do, I just can't remember where.

aednichols (commented):
Latest iteration looks great to me, will hold off with a full review until Kyle's requested changes land

src/pages/workspaces/workspace/Data.js
cjllanwarne: or is it possible to pass props to the constructor directly when you call a reference link?

src/components/RegionFlaggedContent.js
```diff
@@ -0,0 +1,36 @@
...
+import { Link } from 'src/components/common'
+import { icon } from 'src/components/icons'
+import { regionInfo } from 'src/components/region-common'
+import { TooltipCell } from 'src/components/table'
+import { withErrorReporting } from 'src/libs/error'
+import * as Utils from 'src/libs/utils'
+
+
+// Put a google-region appropriate flag next to some content (or a region description by default, if content is undefined).
+const RegionFlaggedContent = ({ location, locationType, content, clipboardText, children, ...props }) => {
```
panentheos: I'm not sure about the value of this abstraction as-is. Even if we do end up with another caller in the other PR, they're different enough that they would probably be cleaner as separate code paths. We can always extract out smaller shared bits, as we discover them.

src/components/RegionFlaggedContent.js
```diff
@@ -0,0 +1,36 @@
...
+  const [copied, setCopied] = useState()
+  const { flag, regionDescription } = regionInfo(location, locationType)
+
+  const bucketLocationTooltip = `${locationType}: ${location}${regionDescription ? ` (${regionDescription})` : ''}`
+
+  return span({ style: { display: 'flex', alignItems: 'center' } }, [
+    h(TooltipCell, { tooltip: bucketLocationTooltip, style: { padding: '1rem' } }, [flag]),
+    h(TooltipCell, { tooltip: !!clipboardText && clipboardText }, [clipboardText || bucketLocationTooltip]),
+    children,
+    clipboardText && h(Link, {
```
panentheos: Since this copy button only applies to the other caller, I think we should keep it inline there and not try to extract it.

src/pages/workspaces/workspace/Data.js
panentheos: Instead of using a path param, we could add an (optional) query param that the page consumes, to initialize the selected data type. That way we could remove it after the page loads so it doesn't become part of the permanent URL scheme.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -15,24 +17,33 @@ import {
...
   const [launching, setLaunching] = useState(undefined)
   const [message, setMessage] = useState(undefined)
   const [launchError, setLaunchError] = useState(undefined)
-
+  const [bucketLocation, setBucketLocation] = useState({})
   const signal = Utils.useCancellation()
 
+  Utils.useOnMount(() => {
+    const loadBucketLocation = async () => {
+      const { location, locationType } = await Ajax(signal).Workspaces.workspace(namespace, workspaceName).checkBucketLocation(bucketName) || {}
```
panentheos: Why do we need the fallback `|| {}` here?

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -100,6 +112,16 @@ const LaunchAnalysisModal = ({
     ]),
     div({ style: { color: colors.danger(), overflowWrap: 'break-word' } }, [
       h(Fragment, wrappableOnPeriods(launchError))
+    ]),
+    location && locationType && div({}, [
+      'Output files will be saved as workspace data in:',
+      span({ style: { layout: 'flex' } }, [
+        h(RegionFlaggedContent, { location, locationType }, [
+          h(TooltipTrigger,
+            { content: div([p(['Cromwell will instruct its job runners to delocalize call outputs to your workspace bucket in this region.']), p(['Tasks within your workflow might additionally move or copy the data elsewhere.', 'You should carefully vet the workflows you run if region requirements are a concern.']), p(['Note that metadata about this run will also be stored in the US.'])]) },
```
panentheos: We should add some linebreaks and indentation here to make this more readable.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -100,6 +112,16 @@ const LaunchAnalysisModal = ({
     ]),
     div({ style: { color: colors.danger(), overflowWrap: 'break-word' } }, [
       h(Fragment, wrappableOnPeriods(launchError))
+    ]),
+    location && locationType && div({}, [
+      'Output files will be saved as workspace data in:',
+      span({ style: { layout: 'flex' } }, [
```
panentheos: This isn't valid, if we want flexbox it should be `display: 'flex'`

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -100,6 +112,16 @@ const LaunchAnalysisModal = ({
...
+        h(RegionFlaggedContent, { location, locationType }, [
+          h(TooltipTrigger,
+            { content: div([p(['Cromwell will instruct its job runners to delocalize call outputs to your workspace bucket in this region.']), p(['Tasks within your workflow might additionally move or copy the data elsewhere.', 'You should carefully vet the workflows you run if region requirements are a concern.']), p(['Note that metadata about this run will also be stored in the US.'])]) },
+            [icon('info-circle', { style: { marginLeft: '0.5rem', marginRight: '0.2rem', cursor: 'pointer', color: colors.accent() } })])
```
panentheos: Right now this isn't accessible to keyboard or screen reader users. As a general rule, use the `Clickable` component or one of its derivatives to get the proper interactivity. In this case, there's an `InfoBox` component that might be a better fit here.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -771,24 +777,48 @@ const WorkflowView = _.flow(
...
-              ])
-            ]),
-            !!rootEntityType && h(Fragment, [
-              div({ style: { margin: '0.5rem 0', borderBottom: `1px solid ${colors.dark(0.55)}` } }),
-              div({ style: styles.outputInfoLabel }, 'References to outputs will be written to'),
-              div({ style: { display: 'flex', alignItems: 'center' } }, [
-                div({ style: { flex: 'none', display: 'flex', width: '1.5rem' } }, [icon('listAlt')]),
-                `Tables / ${rootEntityType}`
+            h(Collapse, { title: 'Output storage' }, [
+              h(Fragment, [
```
panentheos: nit: This `Fragment` and the one below are unnecessary, we could just include the content directly.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -771,24 +777,48 @@ const WorkflowView = _.flow(
...
+                div({ style: styles.outputInfoLabel }, 'Output files will be saved to your workspace bucket:'),
+                div({ style: { display: 'flex', alignItems: 'center', margin: '0.5rem' } }, [
```
panentheos: The horizontal margin is pushing this out of alignment with the rest of the content, we should probably just have vertical margins.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -771,24 +777,48 @@ const WorkflowView = _.flow(
...
+                // TODO: why can't I use the 'RegionFlaggedContent' element here...?
+                  div({ style: { flex: 'none', display: 'flex', width: '1.5rem' } }, [
+                    h(TooltipTrigger, { content: div([`${locationType}: ${location} (${bucketRegionDescription})`]) }, [div(bucketRegionFlag)])
```
panentheos: If this tooltip contains important information, then we should include some accessible text so screen readers can see it as well.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -771,24 +777,48 @@ const WorkflowView = _.flow(
...
+                  ]),
+                  div({ style: { flex: 1 } }, [`gs://${bucketName}`])
+                ])
               ]),
-              `Fill in the attributes below to add or update columns in your data table`
+              h(Fragment, [
+                div({ style: { margin: '0.5rem 0', borderBottom: `1px solid ${colors.dark(0.55)}` } }),
+                div({ style: styles.outputInfoLabel }, 'Raw outputs will be viewable in the Data tab under:'),
+                div({ style: { display: 'flex', alignItems: 'center' } }, [
+                  div({ style: { flex: 'none', display: 'flex' } }, [
```
panentheos: This flex wrapper isn't doing much as-is. If the intention was to preserve the alignment between the icon and the 'Files' text, then these styles should go on the `Link` instead, and this wrapper would be unnecessary.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -771,24 +777,48 @@ const WorkflowView = _.flow(
...
+                  ])
+                ])
+              ]),
+              !!rootEntityType && h(Fragment, [
+                div({ style: { margin: '0.5rem 0', borderBottom: `1px solid ${colors.dark(0.55)}` } }),
+                div({ style: styles.outputInfoLabel }, 'References to outputs can be written to'),
+                div({ style: { display: 'flex', alignItems: 'center' } }, [
+                  h(Link, {
+                    href: Nav.getLink('workspace-data-bucket', { namespace, name: workspaceName, initialDataType: 'localVariables' }),
+                    style: { marginRight: '0.2rem' }
```
panentheos: nit: We should probably keep this margin consistent with the 'Files' one above. Also, this should be a `flex` so the icon and the text are aligned.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -771,24 +777,48 @@ const WorkflowView = _.flow(
...
+                  }, [icon('listAlt', { style: { marginRight: '0.5rem' } }), 'Tables']),
+                  ` / ${rootEntityType}`
+                ]),
+                'Fill in the attributes below to add or update columns in your data table'
+              ])
```
panentheos: TOL: This seems like a nice change overall, but given the complexity with regard to the Data page, perhaps this could have been split out as a separate PR.

src/components/RegionFlaggedContent.js
cjllanwarne: I think it possibly made more sense before I extracted out the region functions (because there was one case where I couldn't re-use this component anyway). Separate code paths seems logical to me.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -100,6 +114,16 @@ const LaunchAnalysisModal = ({
     ]),
     div({ style: { color: colors.danger(), overflowWrap: 'break-word' } }, [
       h(Fragment, wrappableOnPeriods(launchError))
+    ]),
+    div(['Output files will be saved as workspace data in:']),
+    div({ style: { margin: '1rem' } }, [
+      location ? h(Fragment, [span({ style: { marginRight: '0.5rem' } }, [flag]),
+        span({ style: { marginRight: '0.5rem' } }, [regionDescription]),
+        h(InfoBox, [
+          p(['Cromwell will instruct its job runners to delocalize call outputs to your workspace bucket in this region.']),
```
panentheos: nit: Is 'delocalize' a commonly understood word, or could we say something more straightforward like 'write'?

panentheos (approved):
Looks good, thanks for splitting this out. Tested locally.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -100,6 +114,16 @@ const LaunchAnalysisModal = ({
...
+    ]),
+    div(['Output files will be saved as workspace data in:']),
+    div({ style: { margin: '1rem' } }, [
+      location ? h(Fragment, [span({ style: { marginRight: '0.5rem' } }, [flag]),
+        span({ style: { marginRight: '0.5rem' } }, [regionDescription]),
+        h(InfoBox, [
+          p(['Output (and intermediate) files will be written to your workspace bucket in this region.']),
+          p(['Tasks within your workflow might additionally move or copy the data elsewhere. You should carefully vet the workflows you run if region requirements are a concern.']),
+          p(['Note that metadata about this run will be stored in the US.'])
+        ])]) : 'Loading...'
```
salonishah11, line 126: What is the purpose of `Loading...` here? Is it like if the region hasn't been instantiated it displays `Loading` till then?

salonishah11: Never mind. Please ignore the question. I tested it locally and found out the purpose of `Loading`. It is exactly what I said ^

salonishah11 (approved):
LGTM 👍  I have tested it locally and it works as expected

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
cjllanwarne: It's so that if the API call to google hasn't replied yet we can still display the rest of the content around it and leave a placeholder for the region information.

cjllanwarne:
Ignoring flaky integration test failure from known issues.

PR #2327, opened 2020-11-30
AS-514: Add snapshot details to Workspace Data tab

ansingh7115:
<!--
Hello, friend!
Thanks!
--!>


config/dev.json
```diff
@@ -3,6 +3,7 @@
   "bardRoot": "https://terra-bard-dev.appspot.com",
   "bondUrlRoot": "https://broad-bond-dev.appspot.com",
   "calhounUrlRoot": "https://terra-calhoun-dev.appspot.com",
+  "dataRepoUrlRoot": "https://cors-anywhere.herokuapp.com/https://jade.datarepo-dev.broadinstitute.org",
```
ansingh7115: This was a hack to get around the CORS issue - it will be removed once Data Repo fixes it

zarsky-broad: Cool, I'm assuming this will be pre-merge

src/components/workspace-utils.js
```diff
@@ -56,6 +58,35 @@ export const WorkspaceSelector = ({ workspaces, value, onChange, ...props }) =>
...
+      const snapshotInfo = await Ajax(signal).DataRepoSnapshots.snapshots().getSnapshotDetails(snapshotId)
+      setSelectedSnapshotInfo(snapshotInfo)
+    }
+
+    loadSnapshotInfo()
+  })
+
+  const { name, description, createdDate } = snapshotInfo
+
+  console.log('snapshotInfo', snapshotInfo)
```
zarsky-broad:
```suggestion
```

src/components/workspace-utils.js
```diff
@@ -56,6 +58,35 @@ export const WorkspaceSelector = ({ workspaces, value, onChange, ...props }) =>
   })
 }
 
+export const SnapshotInfo = ({ snapshotId }) => {
+  const [snapshotInfo, setSelectedSnapshotInfo] = useState()
```
zarsky-broad, line 61:
```suggestion
  const [snapshotInfo, setSelectedSnapshotInfo] = useState({})
```

src/components/workspace-utils.js
```diff
@@ -56,6 +58,35 @@ export const WorkspaceSelector = ({ workspaces, value, onChange, ...props }) =>
...
+      setSelectedSnapshotInfo(snapshotInfo)
+    }
+
+    loadSnapshotInfo()
+  })
+
+  const { name, description, createdDate } = snapshotInfo
+
+  console.log('snapshotInfo', snapshotInfo)
+  return div({
```
zarsky-broad: Maybe render a spinner until the snapshot info is loaded?

src/components/workspace-utils.js
```diff
@@ -56,6 +58,35 @@ export const WorkspaceSelector = ({ workspaces, value, onChange, ...props }) =>
...
+    style: { ...Style.elements.card.container, color: colors.accent(), height: 125 }
+  }, [
+    div({ style: Style.elements.sectionHeader }, [name]),
+    div({ style: { display: 'flex', justifyContent: 'center', alignItems: 'center' } }, [
+      div([description]),
+      div({ style: { flexGrow: 1 } }),
+      div([`Creation date: ${createdDate}`])
+    ]),
+    div({ style: { margin: '1.5rem 0 1rem', borderBottom: `1px solid ${colors.dark(0.35)}` } })
```
zarsky-broad: Do we need this? There's already a shadow on the container.

src/libs/ajax.js
```diff
@@ -632,6 +633,12 @@ const Workspaces = signal => ({
         return res.json()
       },
 
+      listSnapshot: async (limit, offset) => {
+        const path = `${root}/snapshots?offset=${offset}&limit=${limit}`
+        const res = await fetchRawls(path, _.merge(authOpts(), { signal }))
```
```suggestion
        const res = await fetchRawls(`${root}/snapshots?offset=${offset}&limit=${limit}`, _.merge(authOpts(), { signal }))
```

src/libs/ajax.js
```diff
@@ -792,6 +804,19 @@ const Workspaces = signal => ({
...
+    const root = `repository/v1/snapshots`
+
+    return {
+      getSnapshotDetails: async snapshotId => {
+        const res = await fetchDataRepo(`${root}/${snapshotId}`, _.merge(authOpts(), { signal }))
+        return res.json()
+      }
+    }
+  }
+})
```
zarsky-broad: I might restructure this like so, to make it more generic for future data repo uses (you'd have to change it in other places too, of course):
```js
const DataRepo = signal => ({
  snapshots: () => {
    const root = `repository/v1/snapshots`

    return {
      snapshot: snapshotId => {

        return {
          details: async () => {
            const res = await fetchDataRepo(`${root}/${snapshotId}`, _.merge(authOpts(), { signal }))
            return res.json()
          }
        }
      }
    }
  }
})
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -107,6 +108,22 @@ const DataTypeButton = ({ selected, children, iconName = 'listAlt', iconSize = 1
   ])
 }
 
+const SnapshotCollapse = ({ title, buttonStyle, children, ...props }) => {
```
zarsky-broad: We already have a `Collapse` component, can you use that instead?

src/pages/workspaces/workspace/Data.js
```diff
@@ -702,59 +719,61 @@ class EntitiesContent extends Component {
   render() {
     const {
       workspace, workspace: { workspace: { namespace, name, attributes: { 'workspace-column-defaults': columnDefaults } }, workspaceSubmissionStats: { runningSubmissionsCount } },
-      entityKey, entityMetadata, loadMetadata, firstRender
+      entityKey, entityMetadata, snapshotDetails, loadMetadata, firstRender
     } = this.props
     const { selectedEntities, deletingEntities, copyingEntities, refreshKey, showToolSelector, igvData: { selectedFiles, refGenome } } = this.state
-
```
zarsky-broad: nit: add newline back in

src/pages/workspaces/workspace/Data.js
```diff
@@ -702,59 +719,61 @@ class EntitiesContent extends Component {
...
-          isOpen: showToolSelector,
-          onDismiss: () => this.setState({ showToolSelector: false }),
-          onIgvSuccess: newIgvData => this.setState({ showToolSelector: false, igvData: newIgvData }),
-          entityMetadata,
-          entityKey,
-          selectedEntities
-        })
-      ])
+    return Utils.cond(
+      [_.isArray(entityKey), ((snapshotDetails !== undefined) ? h(SnapshotInfo, { snapshotId: snapshotDetails[entityKey[0]].id } ) : spinnerOverlay())],
```
```suggestion
      [_.isArray(entityKey), () => snapshotDetails === undefined ? spinnerOverlay() : h(SnapshotInfo, { snapshotId: snapshotDetails[entityKey[0]].id })],
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -702,59 +719,61 @@ class EntitiesContent extends Component {
...
+      [selectedFiles ?
```
zarsky-broad: Why not make `selectedFiles` it's own cond branch, rather than having this ternary here? Also, remember that the outcomes for a cond should be functions.

src/pages/workspaces/workspace/Data.js
```diff
@@ -944,13 +963,33 @@ const WorkspaceData = _.flow(
...
     const { namespace, name, signal } = this.props
     const { selectedDataType } = this.state
-    const entityMetadata = await Ajax(signal).Workspaces.workspace(namespace, name).entityMetadata()
+
+    const [entityMetadata, { resources: snapshotMetadata }] = await Promise.all([
+      Ajax(signal).Workspaces.workspace(namespace, name).entityMetadata(),
+      Ajax(signal).Workspaces.workspace(namespace, name).listSnapshot(1000, 0)
+    ])
+
+
```
```suggestion
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -944,13 +963,33 @@ const WorkspaceData = _.flow(
...
     this.setState({
       selectedDataType: this.selectionType() === 'entities' && !entityMetadata[selectedDataType] ? undefined : selectedDataType,
-      entityMetadata
+      entityMetadata,
+      snapshotDetails
     })
   })
 
+
```
```suggestion
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -974,11 +1013,17 @@ const WorkspaceData = _.flow(
     )
   }
 
+
```
```suggestion
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -974,11 +1013,17 @@ const WorkspaceData = _.flow(
...
     const { namespace, name, workspace, workspace: { workspace: { attributes } }, refreshWorkspace, pfbImportJobs } = this.props
-    const { selectedDataType, entityMetadata, importingReference, deletingReference, firstRender, refreshKey, uploadingFile } = this.state
+    const { selectedDataType, entityMetadata, snapshotDetails, importingReference, deletingReference, firstRender, refreshKey, uploadingFile } = this.state
     const referenceData = getReferenceData(attributes)
     const sortedEntityPairs = _.flow(_.toPairs, _.sortBy(_.first))(entityMetadata)
+    const sortedSnapshotPairs = _.flow(_.toPairs, _.sortBy(_.first))(snapshotDetails)
+    console.log('SORTED ENTITY PAIRS')
+    console.log(sortedEntityPairs)
+    console.log('SORTED SNAPSHOT PAIRS')
+    console.log(sortedSnapshotPairs)
```
```suggestion
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -1002,6 +1047,26 @@ const WorkspaceData = _.flow(
               }
             }, [`${type} (${typeDetails.count})`])
           }, sortedEntityPairs),
+          div({ style: Style.navList.heading }, ['Snapshots']),
+          _.map(([snapshotName, { id: snapshotId, entityMetadata: snapshotTables }]) => {
```
zarsky-broad, line 1030: Make sure you're looking at the errors in the browser console when things are being rendered. Here, React is complaining that it's being given an array to render as a child (that is, within the children array of this div) without keys on its members. Whenever we're giving React a list of things to render within children, each should be given a unique `key` prop, so React can keep track of them. [See here](https://reactjs.org/docs/lists-and-keys.html) for more.

src/pages/workspaces/workspace/Data.js
```diff
@@ -1002,6 +1047,26 @@ const WorkspaceData = _.flow(
...
+            const snapshotTablePairs = _.flow(_.toPairs, _.sortBy(_.first))(snapshotTables)
+            return h(SnapshotCollapse, {
+              buttonStyle: { color: colors.dark(), fontWeight: 600 },
+              style: { fontSize: 14, marginRight: '0.5rem', lineHeight: '50px', padding: '0 1.5rem', borderBottom: `1px solid ${colors.dark(0.2)}` },
+              title: snapshotName
+            }, [
+              p({ style: { fontSize: 14, lineHeight: '1.5' } },
```
zarsky-broad: It might be nicer here to use a div with explicit padding rather than a p. P elements have semantic meaning (here is a paragraph of text) that will confuse screen readers if they're used purely visually.

src/pages/workspaces/workspace/Data.js
```diff
@@ -1002,6 +1047,26 @@ const WorkspaceData = _.flow(
...
+                _.map(([tableName, { count }]) => {
+                  return [h(DataTypeButton, {
```
zarsky-broad: This shouldn't return a one-member array, it should just be the `DataTypeButton`. Note that these buttons have keys on them, so that they can be members of the array from this `map` without anything becoming unhappy.

Stylistically, are we ok with the selection indicator being indented from the right margin? It looks a little weird to my eye; I think I'd prefer to see it indented from the left but not the right.

src/pages/workspaces/workspace/Data.js
```diff
@@ -1002,6 +1047,26 @@ const WorkspaceData = _.flow(
...
+          div({ style: Style.navList.heading }, ['Snapshots']),
+          _.map(([snapshotName, { id: snapshotId, entityMetadata: snapshotTables }]) => {
+            const snapshotTablePairs = _.flow(_.toPairs, _.sortBy(_.first))(snapshotTables)
+            return h(SnapshotCollapse, {
+              buttonStyle: { color: colors.dark(), fontWeight: 600 },
+              style: { fontSize: 14, marginRight: '0.5rem', lineHeight: '50px', padding: '0 1.5rem', borderBottom: `1px solid ${colors.dark(0.2)}` },
+              title: snapshotName
+            }, [
+              p({ style: { fontSize: 14, lineHeight: '1.5' } },
+                _.map(([tableName, { count }]) => {
```
zarsky-broad, line 1040: @panentheos What are your thoughts on having a map as the children array, rather than within it? I think we try to avoid that, right? Either keying it or putting it as the contents of a Fragment?

zarsky-broad (changes requested):

src/pages/workspaces/workspace/Data.js
ansingh7115: the selection indicator has been fixed!

src/pages/workspaces/workspace/Data.js
```diff
@@ -702,17 +704,18 @@ class EntitiesContent extends Component {
...
 
     const { initialX, initialY } = firstRender ? StateHistory.get() : {}
     const selectedKeys = _.keys(selectedEntities)
     const selectedLength = selectedKeys.length
 
-    return selectedFiles ?
-      h(IGVBrowser, { selectedFiles, refGenome, workspace, onDismiss: () => this.setState(_.set(['igvData', 'selectedFiles'], undefined)) }) :
-      h(Fragment, [
+    return Utils.cond(
+      [_.isArray(entityKey), () => snapshotDetails === undefined ? spinnerOverlay : h(SnapshotInfo, { snapshotId: snapshotDetails[entityKey[0]].id, snapshotName: entityKey[0] })],
```
panentheos: Since this code path is a completely separate case from the existing rawls entities, it might be cleaner to branch off one level up, e.g. have `selectionType` return a different token to represent snapshots, and then add a case to the `switchCase` in `render` that could go straight to the right content.

src/components/workspace-utils.js
```diff
@@ -56,6 +58,37 @@ export const WorkspaceSelector = ({ workspaces, value, onChange, ...props }) =>
...
+      setSelectedSnapshotInfo(snapshotInfo)
+    }
+
+    loadSnapshotInfo()
+  })
+
+  const { name, description, createdDate } = snapshotInfo
+
+  return _.isEqual(snapshotInfo, {}) ? spinnerOverlay : div({
+    style: { ...Style.elements.card.container }
```
panentheos: nit: The spread is unnecessary here, we could just reference the existing object directly.

src/components/workspace-utils.js
```diff
@@ -56,6 +58,37 @@ export const WorkspaceSelector = ({ workspaces, value, onChange, ...props }) =>
...
+  }, [
+    div({ style: Style.elements.sectionHeader }, [snapshotName]),
+    div({ style: { display: 'flex', justifyContent: 'center', marginTop: '10px' } }, [
+      div({ style: { flexGrow: 1, borderRight: `1px solid ${colors.dark(0.35)}`, paddingRight: '10px', marginRight: '10px' } }, [description]),
+
+      div([
+        div({ style: { marginBottom: '0.5rem', fontWeight: 'bold' } }, 'Underlying Snapshot'),
```
panentheos: nit: Children should have `[]` around them.

src/libs/ajax.js
```diff
@@ -792,6 +803,25 @@ const Workspaces = signal => ({
 })
 
 
+const DataRepo = signal => ({
+  snapshots: () => {
```
panentheos: This level of indirection isn't needed, because it's not capturing any variables the way `snapshot` below is. I think we should collapse this level to make it more concise.

src/pages/workspaces/workspace/Data.js
```diff
@@ -944,10 +948,28 @@ const WorkspaceData = _.flow(
...
+    const [entityMetadata, { resources: snapshotMetadata }] = await Promise.all([
+      Ajax(signal).Workspaces.workspace(namespace, name).entityMetadata(),
+      Ajax(signal).Workspaces.workspace(namespace, name).listSnapshot(1000, 0)
+    ])
+
+    const snapshotEntities = await Promise.all(_.map(({ name: snapshotName }) => {
+      return Ajax(signal).Workspaces.workspace(namespace, name).snapshotEntityMetadata(namespace, snapshotName)
+    }, snapshotMetadata))
+
+    const snapshotDetails = _.reduce(
```
panentheos: Instead of `_.reduce` and dealing with indices, we could use `_.zip` to associate the two arrays, and then `_.map` and `_.fromPairs` to create the object. I think it would increase readability a little by using more specific (less powerful) tools.

src/pages/workspaces/workspace/Data.js
panentheos: It's not the end of the world, but we should avoid dynamically generating the children array because it prevents React from optimizing rendering in some cases. Here, we should add an explicit array to the `div` above, such that this `_.map` expression is contained within it. And because we add a `key` to each `DataTypeButton`, it should work without warnings.

src/pages/workspaces/workspace/Data.js
```diff
@@ -1002,6 +1025,27 @@ const WorkspaceData = _.flow(
               }
             }, [`${type} (${typeDetails.count})`])
           }, sortedEntityPairs),
+          div({ style: Style.navList.heading }, ['Snapshots']),
+          _.map(([snapshotName, { id: snapshotId, entityMetadata: snapshotTables }]) => {
+            const snapshotTablePairs = _.flow(_.toPairs, _.sortBy(_.first))(snapshotTables)
+            return h(Collapse, {
+              key: snapshotName,
+              buttonStyle: { color: colors.dark(), fontWeight: 600 },
```
panentheos: The `Collapse` component has a built-in bottom margin, which is making this look a little off-center. This is ultimately a problem with `Collapse` itself, since it's not good practice for shared components to have external margins or padding (for this reason). For now, we might want to add `marginBottom: 0` here to cancel it out.

config/prod.json
```diff
@@ -3,6 +3,7 @@
   "bardRoot": "https://terra-bard-prod.appspot.com",
   "bondUrlRoot": "https://broad-bond-prod.appspot.com",
   "calhounUrlRoot": "https://terra-calhoun-prod.appspot.com",
+  "dataRepoUrlRoot": "https://jade.datarepo-dev.broadinstitute.org",
```
panentheos: Should this be a prod url?

src/pages/workspaces/workspace/Data.js
```diff
@@ -1002,6 +1025,28 @@ const WorkspaceData = _.flow(
...
+          _.map(([snapshotName, { id: snapshotId, entityMetadata: snapshotTables }]) => {
+            const snapshotTablePairs = _.flow(_.toPairs, _.sortBy(_.first))(snapshotTables)
+            return h(Collapse, {
+              key: snapshotName,
+              buttonStyle: { color: colors.dark(), fontWeight: 600, marginBottom: 0 },
+              style: { fontSize: 14, lineHeight: '50px', paddingLeft: '1.5rem', borderBottom: `1px solid ${colors.dark(0.2)}` },
+              title: snapshotName
+            }, [
+              div({ style: { fontSize: 14, lineHeight: '1.5' } }, [
+                div(_.map(([tableName, { count }]) => {
```
panentheos: nit: This inner div isn't necessary, the `_.map` expression can just sit inside the children array of the outer div.

panentheos (approved):

src/components/Collapse.js
```diff
@@ -4,18 +4,17 @@ import { Link } from 'src/components/common'
 import { icon } from 'src/components/icons'
 
 
-const Collapse = ({ title, buttonStyle, children, ...props }) => {
+const Collapse = ({ title, buttonStyle, children, titleFirst, ...props }) => {
   const [isOpened, setIsOpened] = useState(false)
+  const angleIcon = icon(isOpened ? 'angle-down' : 'angle-right', { style: { marginRight: '0.25rem', flexShrink: 0 } })
+  const titleAndIcon = titleFirst ? [div({ display: 'flex', style: { flexGrow: 1 } }, [title]), angleIcon] : [angleIcon, title]
```
zarsky-broad: Remember that, while children **is** an array, it shouldn't **contain** an array, unless that array has keys.
If you inline this where it's used below (which I recommend you do for simplicity), you'll see that you're putting the array you create here _inside_ the children array of the `Link`.

src/components/workspace-utils.js
```diff
@@ -56,6 +58,31 @@ export const WorkspaceSelector = ({ workspaces, value, onChange, ...props }) =>
...
+      const snapshotInfo = await Ajax(signal).DataRepo.snapshot(snapshotId).details()
+      setSelectedSnapshotInfo(snapshotInfo)
+    }
+
+    loadSnapshotInfo()
+  })
+
+  const { name, description, createdDate } = snapshotInfo
+
+  return _.isEqual(snapshotInfo, {}) ? spinnerOverlay : div({
```
zarsky-broad: We prefer leaving something that hasn't yet been loaded undefined:
```suggestion
  const [snapshotInfo, setSelectedSnapshotInfo] = useState()
  const signal = Utils.useCancellation()

  Utils.useOnMount(() => {
    const loadSnapshotInfo = async () => {
      const snapshotInfo = await Ajax(signal).DataRepo.snapshot(snapshotId).details()
      setSelectedSnapshotInfo(snapshotInfo)
    }

    loadSnapshotInfo()
  })

  const { name, description, createdDate } = snapshotInfo || {}

  return snapshotInfo === undefined ? spinnerOverlay : div({
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -556,6 +558,10 @@ const ToolDrawer = _.flow(
   ])
 })
 
+const SnapshotContent = ({ snapshotDetails, snapshotKey }) => {
+  return snapshotDetails === undefined ? spinnerOverlay : h(SnapshotInfo, { snapshotId: snapshotDetails[snapshotKey[0]].id, snapshotName: snapshotKey[0] })
```
zarsky-broad: Why is this a separate component, rather than just having this ternary in the cond below? And why are we passing in the whole snapshotKey array if we're only ever using the first element, rather than just passing `selectedDataType[0]` below?

zarsky-broad (commented):
A couple comments, and make sure you handle the lint error that's present currently, since you won't be able to merge without it.

src/pages/workspaces/workspace/Data.js
ansingh7115: This is a separate component because we will also need to add the snapshot table here, along with the snapshot info. I can change it to just pass in the first thing in the array (the snapshot name), but the next ticket (adding tables) will require using the second item in the array (the table name). So you'll just have to change it back anyway

zarsky-broad: Ah, cool. In that case, try this:
```suggestion
const SnapshotContent = ({ snapshotDetails, snapshotKey: [snapshotName] }) => {
  return snapshotDetails === undefined ? spinnerOverlay : h(SnapshotInfo, { snapshotId: snapshotDetails[snapshotName].id, snapshotName })
```

I would still recommend moving the spinner down to the cond, since that's in the component in charge of loading, and this shouldn't be rendered until there's something for it to show.

ansingh7115: looks like it renders the spinner even when I take the spinnerOverlay stuff out from here!

zarsky-broad (approved):
👍 pending checks!

PR #2329, opened 2020-11-30
Show workspace bucket's region on the workspace dashboard [BW-444]

cjllanwarne:
Adds the region information for a workspace bucket to the workspace dashboard.

#### Before

image

#### After:

||
|-|
|image |

### DIFF
--- a/src/pages/workspaces/workspace/Dashboard.js
+++ b/src/pages/workspaces/workspace/Dashboard.js
@@ -1,13 +1,13 @@
 import _ from 'lodash/fp'
 import { Component, Fragment } from 'react'
-import { div, h, i, span } from 'react-hyperscript-helpers'
+import { div, h, i, p, span } from 'react-hyperscript-helpers'
 import * as breadcrumbs from 'src/components/breadcrumbs'
-import { ButtonPrimary, ButtonSecondary, Link, spinnerOverlay } from 'src/components/common'
+import { ButtonPrimary, ButtonSecondary, ClipboardButton, Link, spinnerOverlay } from 'src/components/common'
 import { icon, spinner } from 'src/components/icons'
 import { MarkdownEditor, MarkdownViewer } from 'src/components/markdown'
 import { InfoBox } from 'src/components/PopupTrigger'
-import RegionFlaggedContent from 'src/components/RegionFlaggedContent'
-import { SimpleTable } from 'src/components/table'
+import { regionInfo, unknownRegionFlag } from 'src/components/region-common'
+import { SimpleTable, TooltipCell } from 'src/components/table'
 import TooltipTrigger from 'src/components/TooltipTrigger'
 import { WorkspaceTagSelect } from 'src/components/workspace-utils'
 import { displayConsentCodes, displayLibraryAttributes } from 'src/data/workspace-attributes'
@@ -190,6 +190,7 @@ export const WorkspaceDashboard = _.flow(
       consentStatus, tagsList, busy, bucketLocation, bucketLocationType
     } = this.state
     const isEditing = _.isString(editDescription)
+    const { flag, regionDescription } = regionInfo(bucketLocation, bucketLocationType)
 
     return div({ style: { flex: 1, display: 'flex' } }, [
       div({ style: Style.dashboard.leftBox }, [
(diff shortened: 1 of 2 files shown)


src/components/RegionFlaggedContent.js
```diff
@@ -18,7 +18,7 @@ const RegionFlaggedContent = ({ location, locationType, content, clipboardText,
 
   return span({ style: { display: 'flex', alignItems: 'center' } }, [
     h(TooltipCell, { tooltip: bucketLocationTooltip, style: { padding: '1rem' } }, [flag]),
-    h(TooltipCell, { tooltip: !!clipboardText && clipboardText }, [clipboardText || bucketLocationTooltip]),
+    h(TooltipCell, { tooltip: !!clipboardText && clipboardText, style: { flex: '0 0 55%' } }, [content || bucketLocationTooltip]),
```
cjllanwarne, line 21: This is the horrible style hack.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -302,23 +313,17 @@ export const WorkspaceDashboard = _.flow(
           'Google Bucket'
         ]),
         div({ style: { display: 'flex' } }, [
-          div({ style: Style.noWrapEllipsis }, [bucketName]),
+          (bucketLocation ?
```
panentheos: Thinking about the presentation of information here: We've already established a section that's about the bucket, so we don't necessarily need to squeeze everything onto one line. What about leaving this one as-is, and adding another line beneath with the region info? That way we could also display the region text directly, without relying on tooltips.

cjllanwarne: Thanks - I've restored the original link text and split the content across two lines:

image

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -94,22 +94,29 @@ export const WorkspaceDashboard = _.flow(
...
     const { signal, namespace, name, workspace: { accessLevel } } = this.props
     if (Utils.canWrite(accessLevel)) {
       const { estimate } = await Ajax(signal).Workspaces.workspace(namespace, name).storageCostEstimate()
       this.setState({ storageCostEstimate: estimate })
     }
   })
 
+  loadBucketLocation = withErrorReporting('Error loading bucket location data', async () => {
+    const { signal, namespace, workspaceName, workspace: { workspace: { bucketName } } } = this.props
+    const { location, locationType } = !_.isEmpty(bucketName) ? await Ajax(signal).Workspaces.workspace(namespace, workspaceName).checkBucketLocation(bucketName) : {}
```
panentheos: I don't think `bucketName` can be empty, so we don't have to guard this.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -301,19 +312,15 @@ export const WorkspaceDashboard = _.flow(
...
-          }, [icon(bucketCopied ? 'check' : 'copy-to-clipboard')])
+        div({ style: { marginBottom: '0.5rem', display: 'flex' } }, [
+          div({ style: { marginRight: '0.5rem', fontWeight: 500 } }, ['Name:']),
+          h(TooltipCell, { tooltip: bucketName, style: { marginRight: '0.5rem', ...Style.noWrapEllipsis } }, [bucketName]),
+          h(ClipboardButton, { text: bucketName, style: { marginLeft: '0.25rem' } })
+        ]),
+        div({ style: { marginBottom: '0.5rem', display: 'flex' } }, [
+          div({ style: { marginRight: '0.5rem', fontWeight: 500 } }, ['Location:']),
+          div({ style: { marginRight: '0.5rem' } }, [flag || unknownRegionFlag]),
+          regionDescription || 'loading...'
```
panentheos: To me, the 'unknown' icon during load feels a bit distracting (and the text fallback isn't working as written). What about something like:
```suggestion
          bucketLocation ? h(Fragment, [
            div({ style: { marginRight: '0.5rem' } }, [flag]),
            regionDescription
          ]) : 'Loading...'
```

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -301,19 +312,15 @@ export const WorkspaceDashboard = _.flow(
...
-            onClick: withErrorReporting('Error copying to clipboard', async () => {
-              await clipboard.writeText(bucketName)
-              this.setState({ bucketCopied: true }, () => {
-                setTimeout(() => this.setState({ bucketCopied: undefined }), 1500)
-              })
-            })
-          }, [icon(bucketCopied ? 'check' : 'copy-to-clipboard')])
+        div({ style: { marginBottom: '0.5rem', display: 'flex' } }, [
+          div({ style: { marginRight: '0.5rem', fontWeight: 500 } }, ['Name:']),
+          h(TooltipCell, { tooltip: bucketName, style: { marginRight: '0.5rem', ...Style.noWrapEllipsis } }, [bucketName]),
```
panentheos: Since `TooltipCell` already applies `noWrapEllipsis`, we could omit it here. Also, it will default the tooltip to be identical to the `children`, so we could omit the `tooltip` prop as well.

TOL: `TooltipCell` was originally created to be used in tables, as a way to see the full text that would otherwise be clipped by the cell width. However, there's nothing about its implementation that prevents it from being used outside of that context, and in fact the abstraction fits perfectly well here.

src/pages/workspaces/workspace/Dashboard.js
cjllanwarne: Sounds good to me

panentheos (approved):
Looks good. Tested locally.

rsasch (approved):

PR #2330, opened 2020-12-01
[IA-2391] fix cost calc for galaxy apps

jdcanas:
Cost was originally NaN when no app existed because  there was no default value for num nodes.

I also added the node that belongs to the cluster to the cost calculation. This will be an overestimate for users with multiple apps in a cluster, but otherwise it is an underestimate for users with single apps in a cluster. I believe the former is better.


ehanna4 (approved):
Looks good to me. Tested by pulling down changes and running on my local. Observed the price for galaxy apps displayed in the UI. Also reviewed the code and the changes look good and the cleanup of the machineCost method is a plus.

PR #2333, opened 2020-12-03
Trying to make notebooks test less noisy

ehanna4:
Call get() on profile before trying TOS acceptance 
- seems to fix the random auth error we started getting in the notebooks test

as @breilly2 puts it "Maybe the profile.get() does something that causes some side effect in Google land that then causes the tokeninfo call to be more reliable. ¯\_(ツ)_/¯"

Also added a slight delay to the test before checking for the "create notebook" button as that timeout has been a long standing annoyance


rtitle (approved):
Looks good to me, your explanation makes sense :)

PR #2335, opened 2020-12-03
Further optimize tables

zarsky-broad:
Moves from react-virtualized to react-window, which is a newer and lighter library from the same developer.

Adds memoization for the data table, which seems to help a lot.

I wanted to use https://github.com/AhmadMHawwash/scroll-sync-react for the scroll syncing, but react-window has a custom onScroll event, and scroll-sync-react requires the scroll event.


zarsky-broad:
Ugh, I could just fix the tests but that raises questions about accessibility, so I'm gonna poke at that.

zarsky-broad:
Closing in favor of #2339

PR #2337, opened 2020-12-04
IA-2289: adding getDefaultProperties to Bard Payload

xuscrbw:
<!--
Hello, friend!
Thanks!
--!>

Add getDefault Properties to the Bard Payload. Tested this in local browser on was able to see the referral properties
Screen Shot 2020-12-04 at 5 27 57 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.1",
   "private": true,
   "dependencies": {
+    "@databiosphere/bard-client": "^0.1.0",
     "@fortawesome/fontawesome-svg-core": "^1.2.32",
     "@fortawesome/free-regular-svg-icons": "^5.15.1",
     "@fortawesome/free-solid-svg-icons": "^5.15.1",
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1,3 +1,4 @@
+import { getDefaultProperties } from '@databiosphere/bard-client'
 import _ from 'lodash/fp'
 import * as qs from 'qs'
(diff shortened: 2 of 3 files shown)


jdcanas:
i see the field when i test it locally. I'm trying to think of a scenario in which the referred would be something meaningful though, do you have any ideas?

xuscrbw:
The referrer will have the url of the site that they were previously on, so we should be able to use it to determine where our traffic comes from.e

jdcanas:
> The referrer will have the url of the site that they were previously on, so we should be able to use it to determine where our traffic comes from.e

I understand what the field signifies, I'm more interested in testing (i.e. seeing) the field populated properly when there is an actual referrer

xuscrbw:
Just wanted to update @jdcanas: I am looking into testing a "real" example through changing the markup on another site to lead to localhost:3000. I did find a discrepancy, so I am looking into it more.

xuscrbw:
Did some additional testing:
1. Changed markup on some pages to create links to localhost:3000. 
2. Clicked the links and checked events from the Network Tab of dev tools real referrers
Here are some examples of referrers with real links. Think this should be thorough enough to validate the expected behavior 
Screen Shot 2020-12-08 at 11 09 21 AM
Screen Shot 2020-12-08 at 11 09 28 AM
Screen Shot 2020-12-08 at 11 09 37 AM

panentheos (approved):
Looks good, tested locally.

jdcanas (approved):

PR #2339, opened 2020-12-08
memoize cell rendering in datatable

zarsky-broad:
This should make datatable cell rendering a lot less expensive. Since lodash-fp doesn't allow providing a custom key function to memoize, I've added a memoKey as the first argument to the memoized renderers for the headers and cells that we map over.

I thought I might have some trouble with the checkbox cells, but they seem to be behaving nicely. I'm not honestly sure why- without the checked state in the key, I wouldn't expect them to re-render, so I wouldn't expect them to re-evaluate what's selected, and thus I wouldn't expect them to toggle correctly.

### DIFF
--- a/src/components/DataTable.js
+++ b/src/components/DataTable.js
@@ -1,5 +1,5 @@
 import _ from 'lodash/fp'
-import { Fragment, useEffect, useRef, useState } from 'react'
+import { Fragment, memo, useEffect, useRef, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import { Checkbox, Clickable, Link, MenuButton, spinnerOverlay } from 'src/components/common'
@@ -161,7 +161,7 @@ const DataTable = props => {
 
 
   // Memoized components
-  const checkboxCellRenderer = _.memoize((entityName, entity) => {
+  const CheckboxCell = memo(({ entityName, entity }) => {
     const checked = _.has([entityName], selected)
     return h(Checkbox, {
       'aria-label': entityName,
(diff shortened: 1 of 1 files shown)


zarsky-broad:
Apparently I don't need any of the memoKeys? I'm confused as to why. Does lodash-fp create its key from all the args somehow?

panentheos:
This is an interesting idea, but I think `_.memoize` uses an unbounded cache, so over time this could grow and lead to memory problems. If we're seeing a performance improvement from memoizing the cells, then maybe we should try splitting them into proper Components and wrapping them with `React.memo`? Since that's managed by React on a per-instance basis, I don't think it would have the same memory concerns.

zarsky-broad:
So, initially I used React.memo, but I had issues with that; I'll try it again and see if I can make it work.

Alternatively, we could crib from https://github.com/lodash/lodash/blob/master/.internal/memoizeCapped.js

zarsky-broad:
Here it is with React.memo. I stuck some console.log's in, and I do get a lot of logs from inside the render when I wouldn't expect to, contrary to the _.memoize version, which is why I'd gone from that. I don't know why. I pasted in the internal react shallow compare method and added some logging inside that, and I didn't see a single instance where it returned false, so my best guess is that with the virtualization, when things scroll offscreen react considers them unmounted (which is technically correct...) and flushes its cache. So, using react's memo, we only get memoization for things that are in the rendered area while they remain visible, whereas with the lodash memoize renders only need to be computed once ever.

I'm inclined towards making a version of memoize with a bounded cache, or something similar to that, but what do you think?

panentheos:
Stepping back a sec, what's the problem we're trying to solve here? Are we seeing unacceptable performance when scrolling?

zarsky-broad:
That's a typically excellent question.

This started when I was making the component a function, and I remembered about react-window and decided to look into it.
Then, when that wasn't usable for accessibility, I figured I'd take the memoization part and bring it back here.

So I guess I'm not really solving a problem. This is probably an unneeded pre-optimization, huh.

Do you think I should just close this?

panentheos:
If there's not a clear problem, then yeah let's close it for now, especially since review bandwidth is limited. Thanks for the exploration though!

PR #2342, opened 2020-12-10
re-enable snapshots

Shakespeared:
added a flag to hide snapshots on prod. This won't get merged until corresponding changes in rawls config have propagated. Also fixed incorrect URLs for datarepo


davidangb (approved):
:+1: for the second commit;  I'm assuming the revert commit is fine

ansingh7115 (approved):

PR #2346, opened 2020-12-14
allow dismissing service alerts [CA-1053]

panentheos:
This adds a 'Dismiss' link to service alerts, which allows users to suppress the alert for 1 day. Details are as follows:
* This only affects service alerts (the notifications published by comms)
* Pressing 'Dismiss' will hide that specific alert for 1 day, even across page reloads or new tabs.
* The identity of an alert is based on its full contents, so any changes to an alert will be considered a new alert (bypassing the previous dismissal)

Since we're using LocalStorage to save the dismissals, some caveats apply:
* If the user opens an incognito window or a different browser, their dismissals will not carry over.
* In the event we run out of space in LocalStorage, previous dismissals might be removed, meaning alerts could show up again (this is unlikely, but possible)
* If the user clears their browser data (cookies, etc), it will clear previous dismissals, meaning alerts could show up again.

Tested locally by adding an alert via overrides and dismissing it. Also manually modified LocalStorage to check date handling and edge cases.


src/components/notebook-utils.js
```diff
@@ -14,15 +14,7 @@ import * as Utils from 'src/libs/utils'
...
-export const notebookLockHash = async (bucketName, email) => {
-  const msgUint8 = new TextEncoder().encode(`${bucketName}:${email}`)
-  const hashBuffer = await crypto.subtle.digest('SHA-256', msgUint8)
-  const hashArray = new Uint8Array(hashBuffer)
-  return _.flow(
-    _.map(v => v.toString(16).padStart(2, '0')),
-    _.join('')
-  )(hashArray)
-}
+export const notebookLockHash = (bucketName, email) => Utils.sha256(`${bucketName}:${email}`)
```
s-rubenstein, line 17: This seems much cleaner, but I noticed we are now seeing notebooks test failures. Usually I'd assume a flake, but since we've made a change here I would want to double check to make sure its not a real issue. I could see an async issue. It looks functionally equivalent though, so I would be surprised.

src/components/ServiceAlerts.js
```diff
@@ -16,32 +22,48 @@ export const ServiceAlerts = () => {
...
+          ])
+        ]),
+        { id: `alert-${hash}` }
+      )
+    }, newAlertsWithHashes)
+  }
+  const clearOutdatedAlerts = async () => {
+    const outdatedAlerts = _.differenceWith(_.isEqual, prevAlerts, alerts)
+    const outdatedHashes = await alertHashes(outdatedAlerts)
+    _.forEach(hash => clearNotification(`alert-${hash}`), outdatedHashes)
```
s-rubenstein, line 61: I know when we were initially talking about this design, we had thought about clearing the cookies as well, for outdated alerts. This seems like maybe the place to do it? Unless there's another place that makes more sense.

s-rubenstein:
Would you mind adding what testing you did as part of implementing the feature to the PR description?

src/components/ServiceAlerts.js
panentheos: This approach relies on the automatic cleanup provided by the 'dynamic' storage mechanism. If LocalStorage fills up, these entries will be deleted starting with the oldest. I guess we could proactively remove some here, but it wouldn't catch everything, so it might be simpler to just let everything be handled the same way.

src/components/notebook-utils.js
panentheos: Re-ran tests, they no longer show a notebooks issue

s-rubenstein (approved):
LGTM

marctalbott (approved):

PR #2347, opened 2020-12-18
Use node 12 on app engine

zarsky-broad:
This should functionally change nothing, since we only use app engine for serving static files.


panentheos (approved):
Looks good. Nothing to test.

PR #2348, opened 2020-12-21
Convert data tab to hooks [pre- AS-577]

zarsky-broad:
Leaving this comment... for the future! OOooooooOOOOoooh it's a different year!

As usual, hide whitespace and the whole thing magically gets more readable.

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -4,7 +4,7 @@ import filesize from 'filesize'
 import JSZip from 'jszip'
 import _ from 'lodash/fp'
 import * as qs from 'qs'
-import { Component, createRef, Fragment, useEffect, useState } from 'react'
+import { Component, createRef, Fragment, useEffect, useRef, useState } from 'react'
 import { div, form, h, img, input } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import * as breadcrumbs from 'src/components/breadcrumbs'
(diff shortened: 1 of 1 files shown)


zarsky-broad:
@davidangb this is the PR I was talking about, fyi

src/pages/workspaces/workspace/Data.js
```diff
@@ -966,203 +968,191 @@ const WorkspaceData = _.flow(
...
+    setSelectedDataType(getSelectionType() === 'entities' && !entityMetadata[selectedDataType] ? undefined : selectedDataType)
+    setEntityMetadata(entityMetadata)
+    setSnapshotDetails(snapshotDetails)
   })
 
-  componentDidMount() {
-    this.loadMetadata()
-    this.setState({ firstRender: false })
-  }
+  const toSortedPairs = _.memoize(_.flow(_.toPairs, _.sortBy(_.first)))
```
panentheos: This `memoize` won't provide any benefit, because it will create a new memoized function each render. If we want to memoize this, we'd need to use `useMemo`, and probably actually three of them, because there are three different data structures. But do we need to do this? Is there a performance problem here?

src/pages/workspaces/workspace/Data.js
```diff
@@ -966,203 +968,191 @@ const WorkspaceData = _.flow(
...
 
-  refresh() {
```
panentheos, line 981: The parent attempts to call `refresh()` when re-clicking on the Data tab in the nav, so this component will need to expose that with `forwardRef` and `useImperativeHandle`.

panentheos (approved):
Looks good, tested locally.

PR #2353, opened 2021-01-07
IA-2342: Adding Pause Resume features to Terra-UI

xuscrbw:
I have added a few new buttons and screens for the New Galaxy Modal. 

First, we have the pause button. This button will stop the app. While it is in the stopping status, there will be a new modal with only a delete button. This modal will tell the user that the cloud environment is stopping and to return once it has stopped

Secondly, we have a resume button. This button will show up on the new Stopped screen of the New Galaxy Modal. It will trigger a start. When the app status is starting, we will display a new Starting modal with only a delete button. This Modal will tell the user to check back once it has started.

I think we should be able to use the 'Starting' workflow for 'Prestarting' (I've seen this status a few times, but can't replicate it consistently). 

I think there were some more changes that Joy wanted, but wanted to show this first draft.

I tested this locally and went through the workflow.

Mocks: https://projects.invisionapp.com/d/main?origin=v7#/console/20834248/438404970/preview?scrollOffset=0

### DIFF
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -88,7 +88,7 @@ export const NewGalaxyModal = _.flow(
       [Utils.DEFAULT, () => {
         return !!app ?
           h(Fragment, [
-            h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+            (app.status === 'RUNNING') && h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
             (app.status === 'RUNNING') && h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { pauseGalaxy() } }, ['Pause']),
             (app.status === 'STOPPED') && h(ButtonPrimary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume']),
             (app.status === 'RUNNING') && h(ButtonPrimary, { onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -81,21 +81,35 @@ export const NewGalaxyModal = _.flow(
       }],
       ['paused', () => {
         return h(Fragment, [
-                           h(ButtonPrimary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
-                           h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy(); setViewMode(Utils.DEFAULT);} }, ['Resume'])
-                         ])
+          h(ButtonPrimary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+          h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume'])
+        ])
       }],
       [Utils.DEFAULT, () => {
         return !!app ?
           h(Fragment, [
             h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
-            h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { pauseGalaxy(); setViewMode('paused'); }}, ['Pause']),
-            h(ButtonPrimary, { onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])
+            (app.status === 'RUNNING') && h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { pauseGalaxy() } }, ['Pause']),
+            (app.status === 'STOPPED') && h(ButtonPrimary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume']),
+            (app.status === 'RUNNING') && h(ButtonPrimary, { onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])
           ]) :
           h(ButtonPrimary, { onClick: () => setViewMode('createWarn') }, ['Next'])
       }]
     )
   }
+
+  const renderMessaging = () => {
+    return Utils.cond(
+      [viewMode === 'createWarn', renderCreateWarning],
+      [viewMode === 'deleteWarn', renderDeleteWarning],
+      [viewMode === 'launchWarn', renderLaunchWarning],
+      [!!app && (app.status === 'STOPPED'), renderPaused],
+      [!!app && (app.status === 'STOPPING'), renderStoppingWarning],
+      [!!app && (app.status === 'STARTING'), renderStartingWarning],
+      [Utils.DEFAULT, renderDefaultCase]
+    )
+  }
+
 
   const renderCreateWarning = () => {
     return h(Fragment, [
@@ -141,7 +155,30 @@ export const NewGalaxyModal = _.flow(
     ])
   }
 
+  const renderStoppingWarning = () => {
+    return h(Fragment, [
+      div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
+        div([
+          ul({ style: { paddingLeft: '1rem', lineHeight: 1.5 } }, [
+            'The cloud environment is now pausing. This process will take up to a few minutes. Please check back when the environment has stopped.'
+          ])
+        ])
+      ])
+    ])
+  }
 
+
+  const renderStartingWarning = () => {
+    return h(Fragment, [
+      div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
+        div([
+          ul({ style: { paddingLeft: '1rem', lineHeight: 1.5 } }, [
+            'The cloud environment is now starting. This process will take up to a few minutes. Please check back when the environment is running.'
+          ])
+        ])
+      ])
+    ])
+  }
 
   const renderLaunchWarning = () => {
     return div({ style: { lineHeight: '22px' } }, [
@@ -153,31 +190,31 @@ export const NewGalaxyModal = _.flow(
     const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineTypes)
     const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
     return h(Fragment, [
-          div([`Cloud environment is now paused...`]),
-          div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
-            div([
-              div({ style: styles.headerText }, ['Environment Settings']),
-              ul({ style: { paddingLeft: '1rem', lineHeight: 1.5 } }, [
-                li({ style: { marginTop: '1rem' } }, [
-                  'Galaxy version 20.09'
-                ]),
-                li({ style: { marginTop: '1rem' } }, [
-                  'Cloud Compute size of ',
-                  // Temporarily hard-coded disk size, once it can be customized this should be revisited
-                  span({ style: { fontWeight: 600 } }, [`${cpu} CPUS, ${memory} GB of memory, 250 GB disk space`])
-                ]),
-                li({ style: { marginTop: '1rem' } }, [
-                  'Estimated cost of cloud compute: ',
-                  span({ style: { fontWeight: 600 } }, [Utils.formatUSD(cost), ' per hr'])
-                ])
-              ]),
-              h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360050566271', ...Utils.newTabLinkProps }, [
-                'Learn more about Galaxy interactive environments',
-                icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
-              ])
+      div([`Cloud environment is now paused...`]),
+      div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
+        div([
+          div({ style: styles.headerText }, ['Environment Settings']),
+          ul({ style: { paddingLeft: '1rem', lineHeight: 1.5 } }, [
+            li({ style: { marginTop: '1rem' } }, [
+              'Galaxy version 20.09'
+            ]),
+            li({ style: { marginTop: '1rem' } }, [
+              'Cloud Compute size of ',
+              // Temporarily hard-coded disk size, once it can be customized this should be revisited
+              span({ style: { fontWeight: 600 } }, [`${cpu} CPUS, ${memory} GB of memory, 250 GB disk space`])
+            ]),
+            li({ style: { marginTop: '1rem' } }, [
+              'Estimated cost of cloud compute: ',
+              span({ style: { fontWeight: 600 } }, [Utils.formatUSD(cost), ' per hr'])
             ])
+          ]),
+          h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360050566271', ...Utils.newTabLinkProps }, [
+            'Learn more about Galaxy interactive environments',
+            icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
           ])
         ])
+      ])
+    ])
   }
 
   const renderDefaultCase = () => {
@@ -213,23 +250,19 @@ export const NewGalaxyModal = _.flow(
 
   return div({ style: styles.drawerContent }, [
     h(TitleBar, {
-      title: Utils.switchCase(viewMode,
-        ['launchWarn', () => h(WarningTitle, ['Launch Galaxy'])],
-        ['deleteWarn', () => 'Delete Cloud Environment for Galaxy'],
-        ['paused', () => 'Cloud environment is now pausing'],
-        [Utils.DEFAULT, () => 'Cloud environment']
+      title: Utils.cond(
+        [viewMode === 'launchWarn', () => h(WarningTitle, ['Launch Galaxy'])],
+        [viewMode === 'deleteWarn', () => 'Delete Cloud Environment for Galaxy'],
+        [!!app && (app.status === 'STOPPED'), () => 'Cloud environment is now paused'],
+        [!!app && (app.status === 'STOPPING'), () => 'Cloud environment is now pausing'],
+        [!!app && (app.status === 'STARTING'), () => 'Cloud environment is now starting'],
+        () => 'Cloud environment'
       ),
       style: { marginBottom: '0.5rem' },
       onDismiss,
       onPrevious: !!viewMode ? () => setViewMode(undefined) : undefined
     }),
-    Utils.switchCase(viewMode,
-      ['createWarn', renderCreateWarning],
-      ['deleteWarn', renderDeleteWarning],
-      ['launchWarn', renderLaunchWarning],
-      ['paused', renderPaused],
-      [Utils.DEFAULT, renderDefaultCase]
-    ),
+    renderMessaging(),
     div({ style: { display: 'flex', marginTop: '2rem', justifyContent: 'flex-end' } }, [
       renderActionButton()
     ]),
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -88,12 +88,12 @@ export const NewGalaxyModal = _.flow(
       [Utils.DEFAULT, () => {
         return !!app ?
           h(Fragment, [
-            (app.status === 'RUNNING') && h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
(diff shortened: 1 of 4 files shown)


src/libs/events.js
```diff
@@ -8,6 +8,8 @@ const eventsList = {
   applicationLaunch: 'application:launch',
   applicationCreate: 'application:create',
   applicationDelete: 'application:delete',
+  applicationPause: 'application:pause',
```
ehanna4, line 11: Thanks for remembering to add Mixpanel Events! Checked in the mixpanel dashboard and these events look good!

src/components/NewGalaxyModal.js
```diff
@@ -88,6 +127,11 @@ export const NewGalaxyModal = _.flow(
             div({ style: { lineHeight: 1.5 } }, [
               div(['Please delete the cloud environment when finished; it will']),
               div(['continue to ', span({ style: { fontWeight: 600 } }, ['incur charges ']), 'if it keeps running.'])
+            ]),
+            div({ style: { ...styles.headerText, marginTop: '0.5rem' } }, ['Pause and auto-pause']),
+            div({ style: { lineHeight: 1.5 } }, [
+              div(['You can pause anything during the compute, but it will auto-paused when']),
```
ehanna4: What does anything mean here? I think it would be good to be specific about what the user can pause, possibly listed out.

ehanna4: oh also probably should be "will auto-pause when" for the correct tense, also "idle" doesn't need to be past tense I don't think.

src/components/NewGalaxyModal.js
```diff
@@ -111,12 +155,68 @@ export const NewGalaxyModal = _.flow(
...
+    ])
+  }
+
   const renderLaunchWarning = () => {
     return div({ style: { lineHeight: '22px' } }, [
       h(GalaxyWarning)
     ])
   }
 
+  const renderPaused = () => {
```
panentheos: Since this is very similar to `renderDefaultCase`, we should consider using the same code path and adding some conditional logic to handle the differences.

ehanna4:
Could you please link the mocks in this PR? I'm having trouble finding the right ones in Jira

src/components/NewGalaxyModal.js
```diff
@@ -150,21 +250,19 @@ export const NewGalaxyModal = _.flow(
...
-        ['launchWarn', () => h(WarningTitle, ['Launch Galaxy'])],
-        ['deleteWarn', () => 'Delete Cloud Environment for Galaxy'],
-        [Utils.DEFAULT, () => 'Cloud environment']
+      title: Utils.cond(
+        [viewMode === 'launchWarn', () => h(WarningTitle, ['Launch Galaxy'])],
+        [viewMode === 'deleteWarn', () => 'Delete Cloud Environment for Galaxy'],
+        [!!app && (app.status === 'STOPPED'), () => 'Cloud environment is now paused'],
+        [!!app && (app.status === 'STOPPING'), () => 'Cloud environment is now pausing'],
+        [!!app && (app.status === 'STARTING'), () => 'Cloud environment is now starting'],
+        () => 'Cloud environment'
```
panentheos: We have similarly-shaped conditional logic in the title, messaging, and action buttons code, which suggests to me that the various cases are fairly distinct, and we should consider moving the conditional logic higher up, and making the code paths more independent. We did something similar in `NewRuntimeModal`.

ehanna4:
It looks like leo doesn't allow deleting while Galaxy is stopping? I am getting this error when trying
Screen Shot 2021-01-08 at 9 59 11 AM

src/components/NewGalaxyModal.js
```diff
@@ -61,17 +79,38 @@ export const NewGalaxyModal = _.flow(
...
     )
   }
 
+  const renderMessaging = () => {
+    return Utils.cond(
+      [viewMode === 'createWarn', renderCreateWarning],
+      [viewMode === 'deleteWarn', renderDeleteWarning],
+      [viewMode === 'launchWarn', renderLaunchWarning],
+      [!!app && (app.status === 'STOPPED'), renderPaused],
+      [!!app && (app.status === 'STOPPING'), renderStoppingWarning],
```
ehanna4: Probably this should also be for the "pre-stopping" case since that's basically a part of stopping

src/components/NewGalaxyModal.js
```diff
@@ -61,17 +79,38 @@ export const NewGalaxyModal = _.flow(
...
+      [!!app && (app.status === 'STARTING'), renderStartingWarning],
+      [Utils.DEFAULT, renderDefaultCase]
```
panentheos, line 112: TOL: There isn't anything inherently wrong with this code, but a couple thoughts on the structure:

Currently we're using `viewMode` as a nice top-level organizing principle to represent the multiple distinct 'modes' we flip between based on user interaction. Rather than mixing in the status-based logic here, we could move it into the 'default' mode, creating a two-level logical hierarchy. Conceptually, that would say "This dialog has multiple 'modes' that we switch between. Additionally, the default mode (that we see when it first opens) might vary based on the app's current status."

What I like about that framing is that it paves the way for some UX questions. For example, what's the reason for hiding the machine details when starting/stopping? We could easily continue to provide that info, and also show the current status.

src/components/NewGalaxyModal.js
```diff
@@ -111,12 +155,68 @@ export const NewGalaxyModal = _.flow(
     ])
   }
 
+  const renderStoppingWarning = () => {
+    return h(Fragment, [
```
panentheos, line 152: nit: A `Fragment` is only needed if there are multiple children, but since there's a single `div` here it's unnecessary.

src/components/NewGalaxyModal.js
```diff
@@ -111,12 +155,68 @@ export const NewGalaxyModal = _.flow(
...
+      div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
+        div([
+          ul({ style: { paddingLeft: '1rem', lineHeight: 1.5 } }, [
```
panentheos: The `ul` element represents a 'list' and should contain `li` elements. Since we're not conveying a list here, this content should probably just go directly in the `div`

panentheos (commented):
Looking good, added a couple comments.

src/components/NewGalaxyModal.js
xuscrbw: This sounds good. So just to make sure I understand, would this be the same as the renderPaused comment you mentioned above? Where we would move the app.status logic into the renderDefaultCase block?

panentheos: Yep, they're overlapping suggestions. If we go with the re-organization discussed here, then the similar code can be naturally combined as part of that.

src/components/NewGalaxyModal.js
xuscrbw: Hi Brett, Sorry for the late reply I thought I had sent this at the beginning of the week. Would this be similar to how renderEnvironmentWarning works in NewRuntimeModal? 
Also for viewMode = Create/Delete conditions, would this mean that we should also add in the modal titles into the RenderCreateWarning ... etc blocks?

panentheos: Yep, that's the suggestion. Even though every mode has a `TitleBar`, if the contents and behavior differ along the same lines as the body content, then it could be cleaner to push them down into the individual modes. Doing that would consolidate and simplify the conditional logic, at the cost of a slight bit of code bulk in the multiple `TitleBar`s.

src/components/NewGalaxyModal.js
ehanna4:
```suggestion
      [!!app && (app.status === 'STOPPING' || app.status === ''), renderStoppingWarning],
```

> It looks like leo doesn't allow deleting while Galaxy is stopping? I am getting this error when trying
> Screen Shot 2021-01-08 at 9 59 11 AM

Are you planning to make the change for allowing deletion in stopped/stopping on leo side with this ticket?

ehanna4:
> It looks like leo doesn't allow deleting while Galaxy is stopping? I am getting this error when trying
> Screen Shot 2021-01-08 at 9 59 11 AM

Are you planning to make the leo side change for allowing deleting while stopped/stopping with this ticket as well?

xuscrbw:
> It looks like leo doesn't allow deleting while Galaxy is stopping? I am getting this error when trying
> Screen Shot 2021-01-08 at 9 59 11 AM

Yes. Planning on changing this. I'll push a change that removes the delete on stopped. But I think Justin mentioned that it was an oversight. Should we make that change on a different ticket since it requires a leo change first?

ehanna4:
The leo change should go out before this one since it causes an error on the UI, I'm not 100% how best to ticket it, but I would say that leo PR should be most priority since it should basically be considered a blocker for this

src/components/NewGalaxyModal.js
```diff
@@ -96,32 +143,70 @@ export const NewGalaxyModal = _.flow(
...
     ])
   }
 
   const renderDefaultCase = () => {
     const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineTypes)
     const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
     return h(Fragment, [
-      div([`Environment ${app ? 'consists' : 'will consist'} of an application and cloud compute.`]),
+      h(TitleBar, {
+        title: Utils.cond(
```
ehanna4: This cond and the cond below are almost identical in a lot of ways, I wonder if a helper function could be extracted to reduce the repeated code

src/components/NewGalaxyModal.js
```diff
@@ -61,19 +79,43 @@ export const NewGalaxyModal = _.flow(
...
+          h(ButtonPrimary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+          h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume'])
+        ])
+      }],
       [Utils.DEFAULT, () => {
         return !!app ?
           h(Fragment, [
-            h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
-            h(ButtonPrimary, { onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])
+            (app.status === 'RUNNING') && h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
```
ehanna4: This should show the button to delete when it's STOPPED now too right?

xuscrbw: correct, it should show when it is stopped and running

src/components/NewGalaxyModal.js
```diff
@@ -96,32 +144,67 @@ export const NewGalaxyModal = _.flow(
...
+        h(GalaxyWarning)
+      ])
     ])
   }
 
+  const getEnvMessageBasedOnStatus = isTitle => {
+    return Utils.cond(
+      [!!app && (app.status === 'STOPPED'), () => `Cloud environment is now paused ${!isTitle ? '...' : undefined}`],
+      [!!app && (app.status === 'PRESTOPPING'), () => 'Cloud environment is preparing to stop.'],
+      [!!app && (app.status === 'STOPPING'), () => `Cloud environment is pausing ${!isTitle ? 'This process will take up to a few minutes' : undefined}`],
```
ehanna4:
```suggestion
      [!!app && (app.status === 'STOPPING'), () => `Cloud environment is pausing. ${!isTitle ? 'This process will take up to a few minutes' : undefined}`],
```

src/components/NewGalaxyModal.js
```diff
@@ -96,32 +144,67 @@ export const NewGalaxyModal = _.flow(
...
+      [!!app && (app.status === 'PRESTARTING'), () => 'Cloud environment is preparing to start.'],
+      [!!app && (app.status === 'STARTING'), () => `Cloud environment is starting ${!isTitle ? 'This process will take up to a few minutes' : undefined}`],
```
ehanna4:
```suggestion
      [!!app && (app.status === 'STARTING'), () => `Cloud environment is starting. ${!isTitle ? 'This process will take up to a few minutes' : undefined}`],
```

src/components/NewGalaxyModal.js
```diff
@@ -61,19 +79,45 @@ export const NewGalaxyModal = _.flow(
...
       }],
+      ['paused', () => {
+        return h(Fragment, [
+          h(ButtonPrimary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+          h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume'])
+        ])
+      }],
       [Utils.DEFAULT, () => {
         return !!app ?
           h(Fragment, [
```
ehanna4:
```suggestion
          h(Fragment, [
          (app.status === 'RUNNING') && h(Fragment, [
              h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
              h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { pauseGalaxy() } }, ['Pause']),
              h(ButtonPrimary, { onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])
            ]),
            (app.status === 'STOPPED') && h(Fragment, [
              h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
              h(ButtonPrimary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume'])
            ])
         ]) :
```

The status/button logic could be grouped a little more such as the code above

ehanna4 (approved):
Looks good to me! I tested locally running and pausing/resuming/deleting etc. and thanks to the issues in dev recently also got to see the error messaging works too.

src/components/NewGalaxyModal.js
```diff
@@ -61,19 +79,45 @@ export const NewGalaxyModal = _.flow(
       ['launchWarn', () => {
         return h(GalaxyLaunchButton, { app, onClick: onDismiss })
       }],
+      ['paused', () => {
+        return h(Fragment, [
+          h(ButtonPrimary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+          h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume'])
```
zarsky-broad:
```suggestion
          h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: resumeGalaxy }, ['Resume'])
```

src/components/NewGalaxyModal.js
```diff
@@ -61,19 +79,45 @@ export const NewGalaxyModal = _.flow(
...
           h(Fragment, [
-            h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
-            h(ButtonPrimary, { onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])
+            (app.status === 'RUNNING') && h(Fragment, [h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+              h(ButtonSecondary, { style: { marginRight: '1rem' }, onClick: () => { pauseGalaxy() } }, ['Pause']),
+              h(ButtonPrimary, { onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])]),
+            (app.status === 'STOPPED') && h(Fragment, [h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+              h(ButtonPrimary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume'])]),
+            (app.status === 'ERROR') && h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete'])
           ]) :
```
zarsky-broad: This should probably be a `Utils.switchCase` instead of this series of conditionals wrapped in a `Fragment`.
Also, be mindful of consistency with line breaks in your arrays, for instance there should be a linebreak after the open bracket for the children on 91 since they span multiple lines.

src/components/NewGalaxyModal.js
```diff
@@ -61,19 +79,45 @@ export const NewGalaxyModal = _.flow(
...
       }]
     )
   }
 
+  const renderMessaging = () => {
+    return Utils.cond(
+      [viewMode === 'createWarn', renderCreateWarning],
+      [viewMode === 'deleteWarn', renderDeleteWarning],
+      [viewMode === 'launchWarn', renderLaunchWarning],
+      [Utils.DEFAULT, renderDefaultCase]
```
zarsky-broad:
```suggestion
    return Utils.switchCase(viewMode
      ['createWarn', renderCreateWarning],
      ['deleteWarn', renderDeleteWarning],
      ['launchWarn', renderLaunchWarning],
      [Utils.DEFAULT, renderDefaultCase]
```

src/components/NewGalaxyModal.js
```diff
@@ -88,6 +132,11 @@ export const NewGalaxyModal = _.flow(
             div({ style: { lineHeight: 1.5 } }, [
               div(['Please delete the cloud environment when finished; it will']),
               div(['continue to ', span({ style: { fontWeight: 600 } }, ['incur charges ']), 'if it keeps running.'])
+            ]),
+            div({ style: { ...styles.headerText, marginTop: '0.5rem' } }, ['Pause and auto-pause']),
+            div({ style: { lineHeight: 1.5 } }, [
+              div(['You can pause anything during the compute, but it will auto-pause when']),
```
zarsky-broad: "anything" -> "at any time"? This message might benefit from a comms pass for clarity/grammar.

src/components/NewGalaxyModal.js
```diff
@@ -96,32 +145,68 @@ export const NewGalaxyModal = _.flow(
...
 
+  const getEnvMessageBasedOnStatus = isTitle => {
+    return Utils.cond(
+      [!!app && (app.status === 'STOPPED'), () => `Cloud environment is now paused ${(!isTitle ? '...' : '')}`],
+      [!!app && (app.status === 'PRESTOPPING'), () => 'Cloud environment is preparing to stop.'],
+      [!!app && (app.status === 'STOPPING'), () => `Cloud environment is pausing. ${!isTitle ? 'This process will take up to a few minutes' : ''}`],
+      [!!app && (app.status === 'PRESTARTING'), () => 'Cloud environment is preparing to start.'],
+      [!!app && (app.status === 'STARTING'), () => `Cloud environment is starting. ${!isTitle ? 'This process will take up to a few minutes' : ''}`],
+      [!!app && (app.status === 'ERROR'), () => `An error has occurred on your Cloud Environment. ${!isTitle ? 'This process will take up to a few minutes' : ''}`],
+      () => isTitle ? 'Cloud environment' : `Environment ${app ? 'consists' : 'will consist'} of an application and cloud compute.`
```
zarsky-broad: Two comments:
1. instead of `!!app && app.foo`, you can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining): `app?.foo`
2. This should be a `Utils.switchCase`; you can wrap it in a ternary checking if `app` is truthy.

src/libs/runtime-utils.js
```diff
@@ -116,3 +116,7 @@ export const appIsSettingUp = app => {
 export const collapsedRuntimeStatus = runtime => {
   return runtime && (runtime.patchInProgress ? 'LeoReconfiguring' : runtime.status) // NOTE: preserves null vs undefined
 }
+
+export const getAppStatus = app => {
```
zarsky-broad: This doesn't appear to be used anywhere, should it be deleted?

zarsky-broad (changes requested):
Some fairly straightforward changes to address before I start on functional testing

src/components/NewGalaxyModal.js
zarsky-broad: Also, I notice that the Delete button always seems to be there, should it be rendered outside of any conditional so it's not repeated?

src/components/NewGalaxyModal.js
xuscrbw: A question here about syntax:

For 1)

Would this look like: app?.status === "PRESTOPPING"

For 2)
Would we want the switchCase to depend on the app.status in this case?

src/libs/runtime-utils.js
xuscrbw: Removed this from the code

src/components/NewGalaxyModal.js
xuscrbw: Updated to match

src/components/NewGalaxyModal.js
xuscrbw: I was able to get the optional chaining to work, but having some trouble changing the block into a switchCase based on app?.status

src/components/NewGalaxyModal.js
xuscrbw: I think there are some instances where the Delete Button shouldn't be there: Starting/Stopping states stand out to me.

src/components/NewGalaxyModal.js
xuscrbw: Changed

zarsky-broad (approved):

PR #2354, opened 2021-01-07
Break apart Data.js [pre- AS-577]

zarsky-broad:
Just breaking up the big file of doom to make things easier to work with.

No actual code changes, all purely structural.


panentheos (approved):
Looks good, tested locally.

PR #2355, opened 2021-01-11
Add more details around the user bearer token to the logs for integration test

ehanna4:
I can get the same error we are seeing in the nightly test runs here: `https://firecloud-orchestration.dsde-alpha.broadinstitute.org/#/Profile/getAll` if I try to execute the request without authorizing/logging in. Looking at the request header there is no authorization in the request header. However once I log in I do see the auth in the request. 

I am hoping to use these added logs to see if the authorization header is there and is the correct token in the test.

### DIFF
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -58,7 +58,7 @@ const createEntityInWorkspace = (page, billingProject, workspaceName, testEntity
 const makeUser = async () => {
   const { email } = await fetchLyle('create')
   const { accessToken: token } = await fetchLyle('token', email)
-
+  console.info(`created a user with token: ${token}`)
   return { email, token }
 }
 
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):
disclaimer: I didn't try to reproduce this locally...but the change looks pretty trivial...

breilly2 (commented):
I'm somewhat not in favor of logging tokens like this because it's a public project. It's probably not a huge deal because these are very temporary users and don't have a lot of permissions, but I hesitate to create precedent for this practice. (Since I have a separate personal GitHub account, I was able to verify that I was able to see the CircleCI build logs from that account.)

breilly2:
Maybe it's safe to log a substring of the token? Though I'm not sure which part(s) might be both unique and not revealing enough to be a security risk.

ehanna4:
@breilly2 hmm I see your point, I think a substring would be sufficient

Qi77Qi:
ohhh..didn't realize this is logging a token....definitely doesn't sound good :)

integration-tests/utils/integration-helpers.js
```diff
@@ -58,7 +58,7 @@ const createEntityInWorkspace = (page, billingProject, workspaceName, testEntity
 const makeUser = async () => {
   const { email } = await fetchLyle('create')
   const { accessToken: token } = await fetchLyle('token', email)
-
+  console.info(`created a user with token: ...${token.toString().substr(-10, 10)}`)
```
zarsky-broad: is it worth declaring a helper somewhere above?

```js
```

breilly2: I like that idea.
Also, nit: I think `substr(-10)` is sufficient.

breilly2 (approved):
👍  Looks good. Thanks for adding the code to partially obscure the token.

PR #2356, opened 2021-01-11
load snapshots and entities separately [AS-577]

zarsky-broad:
Co-Authored-By: Kai <19541449+Shakespeared@users.noreply.github.com>

Verified locally by simulating errors.

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -20,7 +20,7 @@ import UriViewer from 'src/components/UriViewer'
 import { SnapshotInfo } from 'src/components/workspace-utils'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
-import { withErrorReporting } from 'src/libs/error'
+import { reportError, withErrorReporting } from 'src/libs/error'
 import { pfbImportJobStore } from 'src/libs/state'
 import * as StateHistory from 'src/libs/state-history'
 import * as Style from 'src/libs/style'
@@ -288,6 +288,17 @@ const BucketContent = _.flow(
   ])])
 })
 
+const DataTypeSection = ({ title, titleExtras, error, retryFunction, children }) => h(Fragment, [
+  div({ style: Style.navList.heading }, [
+    title,
+    error ? h(Link, {
+      onClick: retryFunction,
+      tooltip: 'Error loading, click to retry.'
+    }, [icon('sync', { size: 18 })]) : titleExtras
+  ]),
+  children
+])
+
 const WorkspaceData = _.flow(
   Utils.forwardRefWithName('WorkspaceData'),
   wrapWorkspace({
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -322,25 +335,45 @@ const WorkspaceData = _.flow(
...
-    },
-    _.zip(snapshotEntities, snapshotMetadata)))
+  const loadSnapshotMetadata = async () => {
+    try {
+      setSnapshotMetadataError(false)
+      const { resources: snapshotMetadata } = await Ajax(signal).Workspaces.workspace(namespace, name).listSnapshot(1000, 0)
+
+      const snapshotEntities = await Promise.all(_.map(({ name: snapshotName }) => {
+        return Ajax(signal).Workspaces.workspace(namespace, name).snapshotEntityMetadata(namespace, snapshotName)
+      }, snapshotMetadata))
```
davidangb, line 360: for this PR or a future one: can these be loaded lazily? Do we really have to proactively query for every single snapshot reference when the the Data tab is first opened? For a user with multiple snapshots, this could quickly become onerous, and involve a lot of queries that never end up getting  used.

zarsky-broad: Definitely, we don't have to load the tables in a snapshot before someone expands it.

zarsky-broad: I'd want to save that for a separate PR, though.

ehanna4:
Testing comparing how dev vs local pointed at this PR handle a failure of entity data loading by blocking the entity url (method for testing demonstrated by Isaac). See sceenshots, new way handles failure much nicer 
Dev:
Screen Shot 2021-01-12 at 3 15 10 PM
PR:
Screen Shot 2021-01-12 at 3 15 04 PM

ehanna4 (approved):
LGTM - I tested by pulling down PR code and played around with the data table (see screenshots in comment section) and tested out how it handles failing entities. Also checked the code over and it looks reasonable as far as I can tell. Really like the refresh of this file to be hook-y and smaller

src/pages/workspaces/workspace/Data.js
```diff
@@ -322,25 +335,45 @@ const WorkspaceData = _.flow(
...
-    }, snapshotMetadata))
+  const loadEntityMetadata = async () => {
+    try {
+      setEntityMetadataError(false)
+      const entityMetadata = await Ajax(signal).Workspaces.workspace(namespace, name).entityMetadata()
+
+      setSelectedDataType(getSelectionType() === 'entities' && !entityMetadata[selectedDataType] ? undefined : selectedDataType)
+      setEntityMetadata(entityMetadata)
+    } catch (error) {
+      reportError('Error loading workspace entity data', error)
```
davidangb, line 346: this is user-facing language. Should it change to something like  "Error loading workspace tables" or "Error loading data tables" to be consistent with how the rest of the UI identifies it? "entity data" may not be clear to users.

davidangb (approved):
one question about verbiage inline, otherwise :+1:. I don't feel particularly expert on UI code yet so  take my approval in context!

Also please make sure we don't lose the follow-on work to lazily load snapshot details.

src/pages/workspaces/workspace/Data.js
zarsky-broad: This is the same text as the previous error, so I'm gonna leave it as-is for now. I think there's a larger conversation about how we keep terminology in sync potentially to be had, so I'm not gonna poke it now.

PR #2357, opened 2021-01-12
Remove unused scan

zarsky-broad:
Per @ssyms this isn't the place where this scan happens, which is good, since it never actually ran here due to malformed command.


MatthewBemis (approved):

Shakespeared (approved):
👍

PR #2359, opened 2021-01-13
AS-556: add data format to mixpanel workspace:data:import event [risk: no]

davidangb:
This is a small part of AS-556 but does not complete that ticket.

PR to gather feedback.  My goal is to 1) start understanding mixpanel, and 2) improve how mixpanel tracks data imports by including the format for the user's import, e.g. "snapshot" vs. "pfb".

Reviewers: is this all I'd have to do to add a property to this mixpanel event? I have not updated the lexicon in mixpanel itself, but I would do so before merging this PR, assuming this PR is safe to merge.

I have not tested this in the app, I only tested syntax via chrome console.

### DIFF
--- a/src/pages/ImportData.js
+++ b/src/pages/ImportData.js
@@ -115,7 +115,7 @@ const ImportData = () => {
         notify('success', 'Data imported successfully.', { timeout: 3000 })
       }]
     )
-    Ajax().Metrics.captureEvent(Events.workspaceDataImport, { ...extractWorkspaceDetails({ workspace }) })
+    Ajax().Metrics.captureEvent(Events.workspaceDataImport, { format: format, ...extractWorkspaceDetails({ workspace }) })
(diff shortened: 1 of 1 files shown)


Shakespeared:
That should do it to add fields to an existing event - you can also optionally assign it a name which is nice if the event has a lot of properties already or if the value doesn't make it clear what it is. I've done a moderate amount of work with mixpanel on tracking persistent disk metrics, so let me know if you have questions or want to chat about it!

Shakespeared (approved):
LGTM tested in UI

breilly2 (approved):
👍  Looks like it works. I see 2 events in Mixpanel dev... one Wednesday and one Thursday.

gdlex4015:
As per code review policy all work must have JIRA ticket associated with it. Since JIRA is a source of truth for compliance fields such as CLIA impact and Security Impact Description
With no tickets we are bypassing those checks

PR #2360, opened 2021-01-14
[IA-2317] Disable users from changing num workers when cluster is not Running

gpcarr:
When a dataproc cluster is not in Running we want to disable users from being able to update the number of workers (currently this causes the cluster to get stuck updating):

no cluster:
Screen Shot 2021-01-14 at 12 37 52 PM

`Running` cluster:
Screen Shot 2021-01-14 at 12 42 25 PM

`Stopping` cluster:
Screen Shot 2021-01-14 at 12 43 11 PM

`Stopped` cluster:
Screen Shot 2021-01-14 at 12 44 23 PM

`Starting` cluster:
Screen Shot 2021-01-14 at 12 45 03 PM

Did confirm that every other update to Dataproc cluster (master or worker nodes) does work


src/components/NewRuntimeModal.js
```diff
@@ -770,6 +775,8 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
                   isClearable: false,
                   onlyInteger: true,
                   value: numberOfWorkers,
+                  disabled: !this.canUpdateNumberOfWorkers(),
+                  tooltip: !this.canUpdateNumberOfWorkers() ? 'Runtime must be in Running status to change number of workers' : undefined,
```
ehanna4: I just had a thought - is 'Runtime' the user facing language we use or should it be either 'Cloud Environment' or 'Cloud Compute' maybe?

ehanna4:
There is a moment when updating the cluster (if you change the worker number) that the widget (the box in the top right with the play button and gear button) doesn't yet reflect the new "Updating" status even though the Runtime drawer seems to have the correct status of 'Updating' since the box gets grayed out. It makes a bit of a bad user experience to see a tool tip say "Must be in Running state" when the top widget says the cluster _is_ in Running. Not a bug introduced by this code but I wonder if since this will expose that bug more if it should be fixed in this PR. 

Essentially it seems the Cluster.status is not consistent between the ClusterManager component and the NewRuntimeModal component
Screen Shot 2021-01-14 at 2 58 25 PM

src/components/NewRuntimeModal.js
ehanna4:
```suggestion
                  tooltip: !this.canUpdateNumberOfWorkers() ? 'Cloud Compute must be in Running status to change number of workers' : undefined,
```

gpcarr: will update!

gpcarr:
> There is a moment when updating the cluster (if you change the worker number) that the widget (the box in the top right with the play button and gear button) doesn't yet reflect the new "Updating" status even though the Runtime drawer seems to have the correct status of 'Updating' since the box gets grayed out. It makes a bit of a bad user experience to see a tool tip say "Must be in Running state" when the top widget says the cluster _is_ in Running. Not a bug introduced by this code but I wonder if since this will expose that bug more if it should be fixed in this PR.
> 
> Essentially it seems the Cluster.status is not consistent between the ClusterManager component and the NewRuntimeModal component
> Screen Shot 2021-01-14 at 2 58 25 PM

hmmm this could be weird behavior.. i'll try to recreate this and decide if we should fix it with this ticket or create a new ticket for the bug

ehanna4:
yeah if you wanna make a new ticket for that bug I think that would be ok - no need to hold up this PR

src/components/NewRuntimeModal.js
```diff
@@ -785,6 +792,8 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
                   isClearable: false,
                   onlyInteger: true,
                   value: numberOfPreemptibleWorkers,
+                  disabled: !this.canUpdateNumberOfWorkers(),
+                  tooltip: !this.canUpdateNumberOfWorkers() ? 'Cloud Compute must be in Running status to change preemptibles' : undefined,
```
ehanna4:
```suggestion
                  tooltip: !this.canUpdateNumberOfWorkers() ? 'Cloud Compute must be in Running status to change number of preemptibles' : undefined,
```

Just to match the language of the other tool tip

ehanna4 (approved):
Other than the final small language thing - this looks good to me! I reviewed the code and also tested the changes locally pulling down the PR branch and looking at the drawer in many states of runtime statuses

jdcanas (approved):

jdcanas:
LGTM

gpcarr:
test failures are unrelated and currently flaky ... going to merge

PR #2361, opened 2021-01-15
AS-513: Run analyses on snapshot tables

MatthewBemis:
Tested these scenarios:

Switching back and forth between snapshot and table sources to make sure everything looked correct (did not see issues)
Reloading the page and ensuring that snapshot dropdowns are properly loaded (dropdowns re-populated correctly based on existing MC)
Saving inputs/outputs that are invalid for that snapshot (showed validation errors, as expected)
Running an analysis that generated 4 workflows (completed successfully)
Trying to save and run an analysis on a snapshot without a table selected (buttons grayed out)


src/libs/analysis.js
```diff
@@ -61,3 +61,21 @@ export const launch = async ({
     useCallCache, deleteIntermediateOutputFiles
   })
 }
+
+export const launchSnapshot = async ({
```
MatthewBemis: I swing back and forth on wanting to honor DRY and not wanting to have conditionals _literally_ all over the place. I find that in the WorkflowView component, some of the logic may have gotten to be a bit convoluted or complicated because of the conditionals.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -80,7 +88,8 @@ const LaunchAnalysisModal = ({
     [type === processAll, () => entityMetadata[rootEntityType].count],
     [type === processAllAsSet, () => 1],
     [type === chooseSetComponents, () => 1],
-    [type === processMergedSet, () => _.flow(mergeSets, _.uniqBy('entityName'))(selectedEntities).length]
+    [type === processMergedSet, () => _.flow(mergeSets, _.uniqBy('entityName'))(selectedEntities).length],
+    [type === processSnapshotTable, () => 1] //TODO: could be table rowCount but rowCount to workflow conversion is currently unclear
```
MatthewBemis: Using `rowCount` from the snapshot table details is probably the right move in the end, but these submissions were spawning 0 workflows (but still writing outputs). I've followed the examples in the Rawls unit tests but I think there may be something missing either here or there.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -482,10 +526,13 @@ const WorkflowView = _.flow(
         savedConfig: config, modifiedConfig,
         currentSnapRedacted: isRedacted, savedSnapRedacted: isRedacted,
         entityMetadata,
+        availableSnapshots: snapshots,
+        selectedEntitySource,
```
MatthewBemis: Would a more appropriate piece of state be to make this a boolean called `snapshotSourceSelected` rather than doing string comps everywhere?

zarsky-broad (commented):
One thing before I review completely:
When we're setting the selection to a snapshot, we should probably set `selectedSnapshotTableNames: undefined, selectedTableName: undefined`

davidangb:
per the PR template, please indicate what testing you've already done!

MatthewBemis:
@davidangb thanks, I just updated the description!

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -705,33 +711,74 @@ const WorkflowView = _.flow(
...
+                  isClearable: false,
+                  isSearchable: false,
+                  value: selectedTableName,
+                  onChange: ({ value }) => {
+                    this.setState({ selectedTableName: value })
+                    this.setState(_.set(['modifiedConfig', 'rootEntityType'], value))
+                    this.setState(_.unset(['modifiedConfig', 'entityName']))
+                  },
+                  styles: { container: old => ({ ...old, display: 'inline-block', width: 200, marginLeft: '0.5rem' }) },
+                  options: selectedSnapshotTableNames
```
davidangb: could you make this dropdown sorted alphabetically? Otherwise it's very difficult for a user to find the table they're  searching for. `options: _.sortBy(_.identity)(selectedSnapshotTableNames)` should do it, but UI experts might have a better way and/or you might want to sort the array earlier in the code flow.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -705,33 +711,74 @@ const WorkflowView = _.flow(
...
                   placeholder: 'Select data type...',
                   styles: { container: old => ({ ...old, display: 'inline-block', width: 200, marginLeft: '0.5rem' }) },
                   value: selectedEntityType,
-                  onChange: selection => {
-                    const value = this.updateEntityType(selection)
-                    this.setState({ entitySelectionModel: this.resetSelectionModel(value) })
+                  onChange: async ({ value, source }) => {
+                    if (source === 'snapshot') {
+                      const snapshotId = _.find({ name: value }, availableSnapshots).reference.snapshot
+                      const { tables } = await Ajax(signal).DataRepo.snapshot(snapshotId).details()
```
davidangb: 1. could you cache the result of this somewhere, so we don't query for the same data every time if a user swaps the root entity type selection?
2. consider calling Rawls' entity type metadata endpoint, specifying the snapshot reference name, instead of calling data repo directly. It's the same data, but Rawls is already parsing it out for you. And if in the future Rawls supports entity types that are not 1:1 with table names - for instance, if Rawls supports an entity type that is the result of a join - you'd automatically inherit that here by calling Rawls instead of DR.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -705,33 +711,74 @@ const WorkflowView = _.flow(
...
-                    tooltip: Utils.editWorkspaceError(ws),
-                    onClick: () => this.setState({ selectingData: true })
-                  }, ['Select Data']),
-                  label({ style: { marginLeft: '1rem' } }, [`${this.describeSelectionModel()}`])
+              (entitySelectionModel.type === processSnapshotTable) ? h(IdContainer, [id => div({ style: { margin: '2rem 0 0 2rem' } }, [
+                h(Select, {
+                  id,
+                  isDisabled: !!Utils.editWorkspaceError(ws),
+                  isClearable: false,
+                  isSearchable: false,
```
davidangb: why disable `isSearchable`? Would be nice to be able to type to find the entity type I'm looking for.

src/pages/workspaces/workspace/workflows/WorkflowView.js
davidangb: further - if you called Rawls' entity type metadata endpoint instead, you'd also get all the attribute names for each entity type, which would allow you to satisfy the typeahead&validation when editing the method config inputs/outputs

davidangb (commented):
initial comments inline!

MatthewBemis:
@davidangb @zarsky-broad this is ready for review

i tested that once we loaded snapshot metadata once, it was cached and not loaded again. i also tested that input suggestions work for snapshots

davidangb (approved):
works for me! does the last remaining "Potentially inconsistent state update" warning need addressing?

davidangb (approved):
still :+1:

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -478,14 +470,27 @@ const WorkflowView = _.flow(
...
+      const selectedSnapshotTableNames = _.keys(selectedSnapshotEntityMetadata)
+
       this.setState({
         savedConfig: config, modifiedConfig,
         currentSnapRedacted: isRedacted, savedSnapRedacted: isRedacted,
         entityMetadata,
+        availableSnapshots: snapshots,
+        selectedEntitySource: modifiedConfig.dataReferenceName ? 'snapshot' : 'table',
+        selectedTableName: modifiedConfig.dataReferenceName ? modifiedConfig.rootEntityType : undefined,
+        cachedSnapshotEntityMetadata: modifiedConfig.dataReferenceName ? [{ snapshotName: modifiedConfig.dataReferenceName, metadata: selectedSnapshotEntityMetadata }] : [{}],
```
panentheos: Initializing this to `[{}]` adds an empty object to the cache, which will never match but might be confusing when debugging. Just an empty array `[]` should be fine.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -478,14 +470,27 @@ const WorkflowView = _.flow(
...
+        selectedSnapshotTableNames,
+        selectedSnapshotEntityMetadata,
```
panentheos, line 480: Since these two pieces of information can be derived from state we already have, it would be better to compute them dynamically as needed, instead of persisting them to state, since adding extra state opens up more ways things might accidentally get out of sync, either now or in the future.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -478,14 +470,27 @@ const WorkflowView = _.flow(
...
+      const selectedSnapshotEntityMetadata = modifiedConfig.dataReferenceName ? await Ajax(signal).Workspaces.workspace(namespace, name).snapshotEntityMetadata(namespace, modifiedConfig.dataReferenceName) : {}
+      const selectedSnapshotTableNames = _.keys(selectedSnapshotEntityMetadata)
+
       this.setState({
         savedConfig: config, modifiedConfig,
         currentSnapRedacted: isRedacted, savedSnapRedacted: isRedacted,
         entityMetadata,
+        availableSnapshots: snapshots,
+        selectedEntitySource: modifiedConfig.dataReferenceName ? 'snapshot' : 'table',
+        selectedTableName: modifiedConfig.dataReferenceName ? modifiedConfig.rootEntityType : undefined,
```
panentheos: We can compute these two things dynamically from the `modifiedConfig`, so it would be better to avoid storing them as extra state.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -557,13 +562,14 @@ const WorkflowView = _.flow(
...
+        `No ${rootEntityType}s selected`],
+      [type === processSnapshotTable, () => `process entire snapshot table`]
     )
   }
 
   canSave() {
-    const { modifiedConfig: { rootEntityType } } = this.state
-    return this.isSingle() || !!rootEntityType
+    const { modifiedConfig: { rootEntityType, dataReferenceName } } = this.state
+    return (this.isSingle() || !!rootEntityType) || (!!dataReferenceName && !!rootEntityType)
```
panentheos: Is this logic right? It looks like `!!rootEntityType` will always cause the whole expression to be true, regardless of `!!dataReferenceName`

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -478,14 +470,27 @@ const WorkflowView = _.flow(
...
+        selectedTableName: modifiedConfig.dataReferenceName ? modifiedConfig.rootEntityType : undefined,
+        cachedSnapshotEntityMetadata: modifiedConfig.dataReferenceName ? [{ snapshotName: modifiedConfig.dataReferenceName, metadata: selectedSnapshotEntityMetadata }] : [{}],
+        selectedSnapshotTableNames,
+        selectedSnapshotEntityMetadata,
         savedInputsOutputs: inputsOutputs,
         modifiedInputsOutputs: inputsOutputs,
         errors: isRedacted ? { inputs: {}, outputs: {} } : augmentErrors(validationResponse),
-        entitySelectionModel: this.resetSelectionModel(modifiedConfig.rootEntityType, readSelection ? selection.entities : {}, entityMetadata),
+        entitySelectionModel: this.resetSelectionModel(
+          modifiedConfig.dataReferenceName ? modifiedConfig.dataReferenceName : modifiedConfig.rootEntityType,
```
panentheos: Is this right? Though I think this value ends up getting ignored in snapshot mode, I would expect this to be `rootEntityType` in that case. Presumably this will become relevant when we add the ability to select snapshot table entities.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -478,14 +470,27 @@ const WorkflowView = _.flow(
...
+          readSelection ? selection.entities : {},
+          entityMetadata, modifiedConfig.dataReferenceName ? 'snapshot' : ''
```
panentheos: If the last argument is a 'source', we should probably pass `'table'` if we're in table mode, for consistency.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -705,33 +713,76 @@ const WorkflowView = _.flow(
...
+
+                      const selectedSnapshotEntityMetadata = selectedSnapshotCachedMetadata ?
+                        selectedSnapshotCachedMetadata.metadata :
+                        await Ajax(signal).Workspaces.workspace(namespace, workspaceName).snapshotEntityMetadata(namespace, value)
+
+                      this.setState(_.set(['modifiedConfig', 'dataReferenceName'], value))
+                      this.setState({
+                        selectedSnapshotEntityMetadata, selectedEntityType: value, selectedTableName: undefined, selectedSnapshotTableNames: _.keys(selectedSnapshotEntityMetadata),
+                        entitySelectionModel: this.resetSelectionModel(value, {}, {}, source), selectedEntitySource: source,
+                        cachedSnapshotEntityMetadata: _.concat(cachedSnapshotEntityMetadata, { snapshotName: value, metadata: selectedSnapshotEntityMetadata })
```
panentheos: This will add unconditionally add a new cache entry every time the selection changes, causing it to grow indefinitely. This could be confusing to debug, and would cause memory leaks in certain usage patterns.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -705,33 +713,76 @@ const WorkflowView = _.flow(
...
                   isSearchable: true,
                   placeholder: 'Select data type...',
                   styles: { container: old => ({ ...old, display: 'inline-block', width: 200, marginLeft: '0.5rem' }) },
                   value: selectedEntityType,
-                  onChange: selection => {
-                    const value = this.updateEntityType(selection)
-                    this.setState({ entitySelectionModel: this.resetSelectionModel(value) })
+                  onChange: async ({ value, source }) => {
+                    if (source === 'snapshot') {
+                      const selectedSnapshotCachedMetadata = _.find({ snapshotName: value }, cachedSnapshotEntityMetadata)
```
panentheos: TOL: If we're just doing plain caching here, it might be simpler to use something like `memoizeAsync` instead of managing the cache by hand.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -705,33 +713,76 @@ const WorkflowView = _.flow(
...
-                div({ style: { height: '2rem', fontWeight: 'bold' } }, ['Step 2']),
-                div({ style: { display: 'flex', alignItems: 'center' } }, [
-                  h(ButtonPrimary, {
-                    disabled: currentSnapRedacted || this.isSingle() || !rootEntityType ||
-                      !_.includes(selectedEntityType, [...entityTypes, ...possibleSetTypes]) || !!Utils.editWorkspaceError(ws),
-                    tooltip: Utils.editWorkspaceError(ws),
-                    onClick: () => this.setState({ selectingData: true })
-                  }, ['Select Data']),
-                  label({ style: { marginLeft: '1rem' } }, [`${this.describeSelectionModel()}`])
+              (entitySelectionModel.type === processSnapshotTable) ? h(IdContainer, [id => div({ style: { margin: '2rem 0 0 2rem' } }, [
```
panentheos: `IdContainer` is used to generate a unique id to associate a label. Since there's no label here, the id isn't doing anything and could be removed.

Also TOL: It doesn't look like the new select has any descriptive text (label or aria-label), so we could probably do better from an accessibility perspective.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -705,33 +713,76 @@ const WorkflowView = _.flow(
...
+                  isDisabled: !!Utils.editWorkspaceError(ws),
+                  isClearable: false,
+                  value: selectedTableName,
+                  onChange: ({ value }) => {
+                    this.setState({ selectedTableName: value })
+                    this.setState(_.set(['modifiedConfig', 'rootEntityType'], value))
+                    this.setState(_.unset(['modifiedConfig', 'entityName']))
+                  },
+                  styles: { container: old => ({ ...old, display: 'inline-block', width: 200, marginLeft: '0.5rem' }) },
+                  options: _.sortBy(_.identity)(selectedSnapshotTableNames)
```
panentheos: nit: if you're passing multiple params at once to a curried lodash function, you can just make a single call, e.g. `_.sortBy(_.identity, selectedSnapshotTableNames)`

src/pages/workspaces/workspace/workflows/WorkflowView.js
MatthewBemis: I believe that it's right. If you're part way through configuring the analysis to run on a snapshot (i.e. you've selected a snapshot name but not a table name) the expression will be `false` because `!!dataReferenceName` = `true`, `!!rootEntityType` = `false`, and `this.isSingle()` will evaluate to `false`

There was a bug that I fixed in a recent commit where the overall expression was evaluating to `true` when it shouldn't have because `rootEntityType` was never unset when selecting a snapshot name, so if you were testing locally that may have been the cause

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -375,21 +376,10 @@ const WorkflowView = _.flow(
   updateSingleOrMultipleRadioState(config) {
     this.setState({
       processSingle: !config.rootEntityType,
-      selectedEntityType: config.rootEntityType
+      selectedEntityType: config.dataReferenceName ? config.dataReferenceName : config.rootEntityType
```
zarsky-broad: This could actually be simplified to:
```suggestion
      selectedEntityType: config.dataReferenceName || config.rootEntityType
```

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -478,14 +470,27 @@ const WorkflowView = _.flow(
...
+        selectedTableName: modifiedConfig.dataReferenceName ? modifiedConfig.rootEntityType : undefined,
+        cachedSnapshotEntityMetadata: modifiedConfig.dataReferenceName ? [{ snapshotName: modifiedConfig.dataReferenceName, metadata: selectedSnapshotEntityMetadata }] : [],
+        selectedSnapshotTableNames,
+        selectedSnapshotEntityMetadata,
         savedInputsOutputs: inputsOutputs,
         modifiedInputsOutputs: inputsOutputs,
         errors: isRedacted ? { inputs: {}, outputs: {} } : augmentErrors(validationResponse),
-        entitySelectionModel: this.resetSelectionModel(modifiedConfig.rootEntityType, readSelection ? selection.entities : {}, entityMetadata),
+        entitySelectionModel: this.resetSelectionModel(
+          modifiedConfig.dataReferenceName ? modifiedConfig.dataReferenceName : modifiedConfig.rootEntityType,
```
zarsky-broad:
```suggestion
          modifiedConfig.dataReferenceName || modifiedConfig.rootEntityType,
```

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -478,14 +470,27 @@ const WorkflowView = _.flow(
...
+      const selectedSnapshotTableNames = _.keys(selectedSnapshotEntityMetadata)
+
       this.setState({
         savedConfig: config, modifiedConfig,
         currentSnapRedacted: isRedacted, savedSnapRedacted: isRedacted,
         entityMetadata,
+        availableSnapshots: snapshots,
+        selectedEntitySource: modifiedConfig.dataReferenceName ? 'snapshot' : 'table',
+        selectedTableName: modifiedConfig.dataReferenceName ? modifiedConfig.rootEntityType : undefined,
+        cachedSnapshotEntityMetadata: modifiedConfig.dataReferenceName ? [{ snapshotName: modifiedConfig.dataReferenceName, metadata: selectedSnapshotEntityMetadata }] : [],
```
zarsky-broad: Why not simplify the structure of this state to this:
```suggestion
        cachedSnapshotEntityMetadata: modifiedConfig.dataReferenceName ? { [modifiedConfig.dataReferenceName]: selectedSnapshotEntityMetadata } : {},
```

Then, you won't have to `_.find` to check if something is there, just see if there's a value for the key.

zarsky-broad (changes requested):
Strong +1 to Brett's comments about avoiding computed state and replacing caching with `memoizeAsync`.

MatthewBemis:
@zarsky-broad please take another look. I believe I've addressed all comments left by you and Brett at this point.

It seems to be working correctly but please pay careful attention to my use of memoizeAsync for snapshotEntityMetadata :)

zarsky-broad (approved):

PR #2363, opened 2021-01-15
Reprioritize the "this will launch X analyses" counter [BW-487]

cjllanwarne:
Re-prioritizes the "this will launch X analyses" counter by putting it right before the `Launch` button again (instead of hidden amongst the other text).

In personal experience, I found myself regularly submitting too many analyses because this warning was not as visible as it used to be.

### Testing

Tested against the workflow launch modal for: http://localhost:3000/#workspaces/testing-job-manager/lots_of_workflows/workflows/dumbledore_methods/preemptibles_echo_two_strings

#### Before

image

#### After

image

### DIFF
--- a/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
+++ b/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
@@ -101,6 +101,16 @@ const LaunchAnalysisModal = ({
       h(ButtonPrimary, { onClick: onDismiss }, ['OK'])
   }, [
     div({ style: { margin: '1rem 0' } }, ['This analysis will be run by ', h(CromwellVersionLink), '.']),
+    div(['Output files will be saved as workspace data in:']),
+    div({ style: { margin: '1rem' } }, [
+      location ? h(Fragment, [span({ style: { marginRight: '0.5rem' } }, [flag]),
+        span({ style: { marginRight: '0.5rem' } }, [regionDescription]),
+        h(InfoBox, [
+          p(['Output (and intermediate) files will be written to your workspace bucket in this region.']),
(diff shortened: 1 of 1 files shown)


breilly2 (approved):
👍   Seems reasonable to me.
Another change to consider (which may be a bad idea) would be to make the button text say, "Launch 10000 Analyses".

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -101,6 +101,16 @@ const LaunchAnalysisModal = ({
...
+      location ? h(Fragment, [span({ style: { marginRight: '0.5rem' } }, [flag]),
+        span({ style: { marginRight: '0.5rem' } }, [regionDescription]),
+        h(InfoBox, [
+          p(['Output (and intermediate) files will be written to your workspace bucket in this region.']),
+          p(['Tasks within your workflow might additionally move or copy the data elsewhere. You should carefully vet the workflows you run if region requirements are a concern.']),
+          p(['Note that metadata about this run will be stored in the US.'])
+        ])]) : 'Loading...'
+    ]),
     div({ style: { margin: '1rem 0' } }, [
       'This will launch ', b([entityCount]), ` analys${entityCount === 1 ? 'is' : 'es'}`,
```
panentheos, line 115: If this is the most important information, why not move it to the top instead of the bottom?

panentheos (approved):
See comment, but looks good. Tested locally.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
cjllanwarne: I guess this is all subjective... but to my brain at least, the "next to the button" location is more visible and my eye is drawn to it more readily than the "right below the title" location.

PR #2364, opened 2021-01-15
lazily load snapshot contents [post- AS-577]

Shakespeared:
make snapshot contents load lazily, added ability to retry if a snapshot fails to load

### DIFF
--- a/src/components/Collapse.js
+++ b/src/components/Collapse.js
@@ -1,12 +1,20 @@
-import { useState } from 'react'
+import { useCallback, useEffect, useState } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
 import { Link } from 'src/components/common'
 import { icon } from 'src/components/icons'
 
 
-const Collapse = ({ title, buttonStyle, initialOpenState, children, titleFirst, ...props }) => {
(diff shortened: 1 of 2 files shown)


MatthewBemis (approved):
LGTM. Tested locally by forcing a 4xx error loading snapshots

src/components/Collapse.js
```diff
@@ -1,21 +1,31 @@
...
   const [isOpened, setIsOpened] = useState(initialOpenState)
   const angleIcon = icon(isOpened ? 'angle-down' : 'angle-right', { style: { marginRight: '0.25rem', flexShrink: 0 } })
 
+  const firstOpenCallback = useCallback(onFirstOpen, [refreshKey])
+
+  useEffect(() => {
+    if (isOpened) {
+      firstOpenCallback()
+    }
+  }, [firstOpenCallback, isOpened, refreshKey])
```
davidangb: `useEffect` seems to be triggered every time I click on a snapshot table in the left nav. This means that `firstOpenCallback()` is called on every click, and thus the API call to `.../entities?...` occurs on every click.

Found by opening dev tools, watching network requests,  and clicking around on the tables.

davidangb (changes requested):

davidangb (approved):
nice! Working well for me, tested locally.

PR #2365, opened 2021-01-15
Some more hook conversions

zarsky-broad:
As always, better with whitespace hidden...

### DIFF
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -1,6 +1,5 @@
 import _ from 'lodash/fp'
-import PropTypes from 'prop-types'
-import { Component, Fragment } from 'react'
+import { Fragment, useState } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
 import { ButtonPrimary, IdContainer, Link, Select, spinnerOverlay } from 'src/components/common'
 import { icon } from 'src/components/icons'
@@ -13,7 +12,6 @@ import { withErrorReporting } from 'src/libs/error'
 import Events from 'src/libs/events'
 import { FormLabel } from 'src/libs/forms'
 import * as Nav from 'src/libs/nav'
-import { authStore } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 import validate from 'validate.js'
 
@@ -32,48 +30,41 @@ const constraints = {
   }
 }
 
-export default _.flow(
-  Utils.withCancellationSignal,
-  Utils.connectStore(authStore, 'authState')
-)(class NewWorkspaceModal extends Component {
-  static propTypes = {
-    cloneWorkspace: PropTypes.object,
-    onDismiss: PropTypes.func.isRequired,
-    customMessage: PropTypes.string,
-    title: PropTypes.string,
-    buttonText: PropTypes.string
+const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
+  cloneWorkspace, onSuccess, onDismiss, customMessage, requiredAuthDomain, title, buttonText
+}) => {
+  // State
+  const [billingProjects, setBillingProjects] = useState()
+  const [allGroups, setAllGroups] = useState()
+  const [name, setName] = useState(cloneWorkspace ? `${cloneWorkspace.workspace.name} copy` : '')
+  const [namespace, setNamespace] = useState(cloneWorkspace ? cloneWorkspace.workspace.namespace : undefined)
+  const [description, setDescription] = useState(cloneWorkspace ? cloneWorkspace.workspace.attributes.description : '')
+  const [groups, setGroups] = useState([])
+  const [nameModified, setNameModified] = useState(false)
+  const [loading, setLoading] = useState(false)
+  const [creating, setCreating] = useState(false)
+  const [createError, setCreateError] = useState()
+
+  const signal = Utils.useCancellation()
+
+
+  // Helpers
+  const getRequiredGroups = () => {
+    return _.uniq([
+      ...(cloneWorkspace ? _.map('membersGroupName', cloneWorkspace.workspace.authorizationDomain) : []),
+      ...(requiredAuthDomain ? [requiredAuthDomain] : [])
+    ])
   }
 
-  constructor(props) {
-    super(props)
-    const { cloneWorkspace } = props
-    this.state = {
-      billingProjects: undefined,
-      allGroups: undefined,
-      name: cloneWorkspace ? `${cloneWorkspace.workspace.name} copy` : '',
-      namespace: cloneWorkspace ? cloneWorkspace.workspace.namespace : undefined,
-      description: (cloneWorkspace && cloneWorkspace.workspace.attributes.description) || '',
-      groups: [],
-      nameModified: false,
-      loading: false,
-      creating: false,
-      createError: undefined
-    }
-  }
-
-  componentDidMount() {
-    this.loadProjectsGroups()
-  }
-
-  async create() {
-    const { onSuccess, cloneWorkspace } = this.props
-    const { namespace, name, description, groups } = this.state
+  const create = async () => {
     try {
-      this.setState({ createError: undefined, creating: true })
+      setCreateError(undefined)
+      setCreating(true)
+
       const body = {
         namespace,
         name,
-        authorizationDomain: _.map(v => ({ membersGroupName: v }), [...this.getRequiredGroups(), ...groups]),
+        authorizationDomain: _.map(v => ({ membersGroupName: v }), [...getRequiredGroups(), ...groups]),
         attributes: { description },
         copyFilesWithPrefix: 'notebooks/'
       }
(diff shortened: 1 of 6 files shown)


src/components/NewWorkspaceModal.js
```diff
@@ -32,48 +30,41 @@ const constraints = {
...
-    cloneWorkspace: PropTypes.object,
-    onDismiss: PropTypes.func.isRequired,
-    customMessage: PropTypes.string,
-    title: PropTypes.string,
-    buttonText: PropTypes.string
-  }
+const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
+  cloneWorkspace, onSuccess, onDismiss, customMessage, requiredAuthDomain, title, buttonText
+}) => {
+  // State
```
ehanna4, line 36: is this comment needed?

ehanna4: oh I see these are throughout, pointing out the code area's function nvm I think that's fine

src/pages/billing/Project.js
```diff
@@ -14,187 +14,176 @@ import * as StateHistory from 'src/libs/state-history'
...
+        div({ style: { flexGrow: 1 } },
+          _.map(member => {
+            return h(MemberCard, {
+              adminLabel: 'Owner',
+              userLabel: 'User',
+              member, adminCanEdit,
+              onEdit: () => setEditingUser(member),
+              onDelete: () => setDeletingUser(member)
+            })
+          }, projectUsers)
```
ehanna4, line 152: Why don't we need to filter anymore?

zarsky-broad: There's actually no search box rendered on the page, I think the state was a holdover.

ehanna4: ah okay

ehanna4:
FWIW in the future smaller PR's for individual files would be less intimidating and more manageable to review

src/pages/billing/Project.js
```diff
@@ -14,187 +14,176 @@ import * as StateHistory from 'src/libs/state-history'
...
+  ])
+}
 
-    StateHistory.update(_.pick(
-      ['projectUsers', 'filter'],
-      this.state)
-    )
-  }
-})
+export default ProjectDetail
```
ehanna4, line 189: What utility does it provide to export the default at the bottom rather than inlining it?

src/pages/billing/Project.js
```diff
@@ -14,187 +14,176 @@ import * as StateHistory from 'src/libs/state-history'
...
-          await Ajax().Billing.project(projectName).removeUser(deletingUser.roles, deletingUser.email)
-          this.refresh()
-        })
-      }),
-      (refreshing || loadingBillingInfo || updatingAccount || updating) && spinnerOverlay
-    ])
-  }
-
-  componentDidUpdate(prevProps, prevState) {
-    const hasBillingScope = Auth.hasBillingScope()
```
ehanna4, line 189: I'm confused about this logic and why it was there before. The new code is simpler but I can't figure out this excatly

ehanna4:
@Shakespeared We tested the changes all the files locally and reviewed that code. Left a few questions but approve generally so far - nothing exploded on testing, just have some questions to be sure we understand everything that's happening.

src/pages/billing/List.js
```diff
@@ -171,157 +170,147 @@ const NewBillingProjectModal = ({ onSuccess, onDismiss, billingAccounts, loadAcc
...
-    this.state = {
-      billingProjects: null,
-      creatingBillingProject: false,
-      billingAccounts: null,
-      isOwner: false,
-      ...StateHistory.get()
-    }
-  }
+export const BillingList = ({ queryParams: { selectedName } }) => {
+  // Hooks
```
ehanna4: Should this be  `// State` for consistency?

src/pages/billing/List.js
```diff
@@ -171,157 +170,147 @@ const NewBillingProjectModal = ({ onSuccess, onDismiss, billingAccounts, loadAcc
...
+  const [billingProjects, setBillingProjects] = useState(StateHistory.get().billingProjects || null)
```
ehanna4, line 175: Why do we use stateHistory here but nothing else compared to the old code which looks like it relied on stateHistory for setting a bit of state?

src/pages/SignIn.js
```diff
@@ -34,101 +34,97 @@ export const CookiesModal = (({ onDismiss }) => {
...
-              href: 'https://www.federalregister.gov/documents/2008/11/20/E8-27771/amendments-to-executive-order-9397-relating-to-federal-agency-use-of-social-security-numbers'
-            }, ['13478']),
-            `. Completing the form is voluntary, however, declining to provide any or all of the requested information may result in denial of access to controlled data.
-             The principal purpose for which the information will be used is to authenticate users who request access to controlled access data. The information will be
-             used to contact you in response to requests you have specifically made on this Web site. Your personal information may also be used to audit your activity
-             on the system in order to ensure compliance with NIH policies. The information you provide will be included in a Privacy Act system of Records, and will be
-             used and may be disclosed for the purposes and routine uses described and published in the following System of Records Notice (SORN):
-             09-90-1401, Records About Restricted Dataset Requesters, HHS/OS/Other `,
+  return h(HeroWrapper, { showMenu: false }, [
```
ehanna4: Can you add the // Render and // State comments in this file too?

src/pages/Profile.js
```diff
@@ -236,189 +236,176 @@ const FenceLink = ({ provider: { key, name } }) => {
...
 const sectionTitle = text => div({ style: styles.sectionTitle }, [text])
 
-const Profile = _.flow(
-  Utils.withCancellationSignal,
-  Utils.connectStore(authStore, 'authState')
-)(class Profile extends Component {
-  constructor(props) {
-    super(props)
+const Profile = ({ queryParams = {} }) => {
+  // Hooks
```
ehanna4: We @Shakespeared think "State" is more clear

src/pages/Profile.js
```diff
@@ -236,189 +236,176 @@ const FenceLink = ({ provider: { key, name } }) => {
...
+  const assignValue = _.curry((key, value) => {
+    setProfileInfo(_.set(key, value))
+  })
 
-  save = _.flow(
-    Utils.withBusyState(v => this.setState({ saving: v })),
-    withErrorReporting('Error saving profile')
-  )(async () => {
-    const { profileInfo } = this.state
+  const line = children => div({ style: styles.form.line }, children)
```
ehanna4, line 253: Why do we not need that spread here anymore?

src/components/NewWorkspaceModal.js
```diff
@@ -32,48 +30,41 @@ const constraints = {
...
-)(class NewWorkspaceModal extends Component {
-  static propTypes = {
-    cloneWorkspace: PropTypes.object,
-    onDismiss: PropTypes.func.isRequired,
-    customMessage: PropTypes.string,
-    title: PropTypes.string,
-    buttonText: PropTypes.string
-  }
+const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
+  cloneWorkspace, onSuccess, onDismiss, customMessage, requiredAuthDomain, title, buttonText
```
ehanna4, line 34: @Shakespeared and I want to know if there is no propType equivalent in hook type components?

src/components/NewWorkspaceModal.js
```diff
@@ -96,143 +87,141 @@ export default _.flow(
...
-                ...Utils.newTabLinkProps
-              }, ['Read more about authorization domains'])
-            ])
-          ]),
-          !!existingGroups.length && div({ style: { marginBottom: '0.5rem', fontSize: 12 } }, [
-            div({ style: { marginBottom: '0.2rem' } }, ['Inherited groups:']),
-            ...existingGroups.join(', ')
-          ]),
-          h(Select, {
+  // Hooks
```
ehanna4: Is this the right comment?

src/components/NewWorkspaceModal.js
ehanna4: We (@Shakespeared ) found these actually very nice and also appreciated the consistent ordering of the code in your changes (i.e` //State ` `//Helpers` `//Lifecycle` `//Render`) made the readability much nicer thanks!

src/pages/billing/Project.js
zarsky-broad: We actually require it with an eslint rule.

Specifically for a default, you can't name it if you export it inline and it's not a class:

```js
export default () => {} // this works, but no name associated with it, e.g. in the React dev tools*
export default class Foo extends Component {} // this works, but it's a class
const Bar = () => {}
export default Bar // this works and has a name!

export default const Baz = () => {} // not a thing, unfortunately
````

\* any component can have a name associated with it that'll show up in the dev tools manually, I made `Utils.withDisplayName` to make this easier, but it's still an extra step

File: src/pages/Profile.js (hunk 9 above)
Comment (zarsky-broad) [reply src/pages/Profile.js:253]: For consistency, I changed this helper from taking an arbitrary number of arguments to taking a standard children array.

File: src/pages/billing/List.js (hunk 6 above)
Comment (zarsky-broad) [reply src/pages/billing/List.js:175]: This is exactly the kind of clarification I like from converting to hooks: we were merging the whole thing into state, but down below it turns out that the billing project list is the only thing actually being saved into state at all!

File: src/pages/billing/Project.js (hunk 4 above)
Comment (zarsky-broad) [reply src/pages/billing/Project.js:189]: Yeah, previously it was sporadically writing this into state, now I'm just consulting Auth every time as the single source of truth rather than trying to duplicate the information.

File: src/components/NewWorkspaceModal.js (hunk 10 above)
Comment (zarsky-broad) [reply src/components/NewWorkspaceModal.js:34]: So there actually is, we could put this down at the bottom before the export:

```js
NewWorkspaceModal.propTypes = {
  cloneWorkspace: PropTypes.object,
  ...
```

I judged that it wasn't worth it in this case, but I'm open to discussion.

src/pages/billing/Project.js
ehanna4: This is a great explanation - is there someplace we could save this for future reference?

src/components/NewWorkspaceModal.js
ehanna4: I'm wary of not requiring an `onDismiss` function- what would be the default behavior without one?

ehanna4 (approved):
Looks good to me! Tested locally and reviewed code

Shakespeared (approved):
pair-reviewed with Emily, LGTM

PR #2366, opened 2021-01-19
[IA-2334] Account for preemptible disk in running compute cost

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2334

Preemptible workers incur disk costs while they are running, but not while they are paused. This PR adds preemptible storage to the 'running' cost calculation function. Note Leo sets preemptible disk size equal to worker disk size.

Tested locally side by side with dev, verified I saw the increased cost with preemptibles in this branch.

<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -40,13 +40,14 @@ export const findMachineType = name => {
 }
 
 export const runtimeConfigCost = config => {
-  const { masterMachineType, numberOfWorkers, numberOfPreemptibleWorkers, workerMachineType } = normalizeRuntimeConfig(config)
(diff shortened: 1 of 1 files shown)


ehanna4 (approved):
Looks good to me. Tested by comparing local and dev creating clusters with the same amount of preemptible disks and local price is more for the running cost, while stopped cost is the same
Screen Shot 2021-01-19 at 2 55 29 PM

rtitle:
Thanks for testing!

breilly2 (approved):
👍   Code looks good. Relying on Emily's testing, which I trust. I studied the change until I was convinced that we're also doing the right thing for preemptible workers when paused, and for non-preemptible workers in both cases.

PR #2367, opened 2021-01-20
[IA-2478] Account for Dataproc pricing for pre-emptibles

kyuksel:
[JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-2478)

Account for Dataproc premium for pre-emptibles when the cluster is running per [Google's pricing example](https://cloud.google.com/dataproc/pricing#pricing_example). Note that the aforementioned premium is separate from and charged in addition to any GCE and storage costs (which are correctly calculated as is). Since we remove pre-emptibles prior to stopping a Dataproc cluster, the cost addition only applies to clusters that are in _running_ state.

Tested manually on a local UI.

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -6,7 +6,10 @@ export const DEFAULT_DISK_SIZE = 50
 
 export const usableStatuses = ['Updating', 'Running']
 
-export const normalizeRuntimeConfig = ({ cloudService, machineType, diskSize, masterMachineType, masterDiskSize, numberOfWorkers, numberOfPreemptibleWorkers, workerMachineType, workerDiskSize, bootDiskSize }) => {
+export const normalizeRuntimeConfig = ({
+  cloudService, machineType, diskSize, masterMachineType, masterDiskSize, numberOfWorkers,
+  numberOfPreemptibleWorkers, workerMachineType, workerDiskSize, bootDiskSize
+}) => {
   const isDataproc = cloudService === cloudServices.DATAPROC
 
   return {
@@ -23,7 +26,9 @@ export const normalizeRuntimeConfig = ({ cloudService, machineType, diskSize, ma
 }
 
 export const runtimeConfigBaseCost = config => {
-  const { cloudService, masterMachineType, masterDiskSize, numberOfWorkers, workerMachineType, workerDiskSize, bootDiskSize } = normalizeRuntimeConfig(
+  const {
+    cloudService, masterMachineType, masterDiskSize, numberOfWorkers, workerMachineType, workerDiskSize, bootDiskSize
+  } = normalizeRuntimeConfig(
     config)
   const { cpu: masterCpu } = findMachineType(masterMachineType)
   const { cpu: workerCpu } = findMachineType(workerMachineType)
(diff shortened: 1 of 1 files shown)


src/libs/runtime-utils.js
```diff
@@ -40,14 +40,15 @@ export const findMachineType = name => {
...
-  const { masterMachineType, numberOfWorkers, numberOfPreemptibleWorkers, workerMachineType, workerDiskSize } = normalizeRuntimeConfig(config)
+  const { cloudService, masterMachineType, numberOfWorkers, numberOfPreemptibleWorkers, workerMachineType, workerDiskSize } = normalizeRuntimeConfig(config)
   const { price: masterPrice } = findMachineType(masterMachineType)
   const { price: workerPrice, preemptiblePrice } = findMachineType(workerMachineType)
   return _.sum([
     masterPrice,
     numberOfWorkers * workerPrice,
     numberOfPreemptibleWorkers * preemptiblePrice,
     numberOfPreemptibleWorkers * workerDiskSize * storagePrice,
+    cloudService === cloudServices.DATAPROC && numberOfPreemptibleWorkers * dataprocCpuPrice,
```
rtitle: should this be
```
cloudService === cloudServices.DATAPROC && numberOfPreemptibleWorkers * workerCpu * dataprocCpuPrice,
```
?

kyuksel: ah yes - thank you!

src/libs/runtime-utils.js
```diff
@@ -40,14 +40,15 @@ export const findMachineType = name => {
 }
 
 export const runtimeConfigCost = config => {
-  const { masterMachineType, numberOfWorkers, numberOfPreemptibleWorkers, workerMachineType, workerDiskSize } = normalizeRuntimeConfig(config)
+  const { cloudService, masterMachineType, numberOfWorkers, numberOfPreemptibleWorkers, workerMachineType, workerDiskSize } = normalizeRuntimeConfig(config)
   const { price: masterPrice } = findMachineType(masterMachineType)
   const { price: workerPrice, preemptiblePrice } = findMachineType(workerMachineType)
```
breilly2:
```suggestion
  const { price: workerPrice, preemptiblePrice, workerCpu } = findMachineType(workerMachineType)
```

breilly2:
TOL: I think it's worth considering extracting a `dataprocCost` function that takes `machineType` and `numInstances`. It would be called 3 times: from `runtimeConfigBaseCost` with `(masterMachineType, 1)` and `(workerMachineType, numberOfWorkers)` and from `runtimeConfigCost` with `(workerMachineType, numberOfPreemptibleWorkers)`.

breilly2:
Also, I'm really bummed that we missed this before. Thank you for fixing it!

rtitle:
+1 thanks Kaan for pushing this through, I also thought this was correct previously. This stuff is tricky

src/libs/runtime-utils.js
kyuksel: 👍🏼  with `cpu: ` prefix to `workerCpu`

kyuksel:
> TOL: I think it's worth considering extracting a dataprocCost function that takes machineType and numInstances. It would be called 3 times: from runtimeConfigBaseCost with (masterMachineType, 1) and (workerMachineType, numberOfWorkers) and from runtimeConfigCost with (workerMachineType, numberOfPreemptibleWorkers).

Thanks for your review @breilly2. Since I'm new to React/Javascript and even the basicest changes may take some time, I'd like to get this fix in without that refactoring so we stop displaying inaccurate cost and add the refactoring change on a subsequent PR. Could you please 👍🏼 my PR if that's reasonable?

breilly2 (approved):
👍   I am fine merging this as-is and extracting a dataproc cost function later.

ehanna4 (approved):
Looks good! Thanks for catching this!

PR #2369, opened 2021-01-25
[CA-1054] Add a cookie warning to bottom of page

s-rubenstein:
Tested locally by accepting and rejecting the cookie popup both logged in and logged out, as well as accepting after having rejected the cookie policy. Also tested the privacy policy link.


s-rubenstein:
Screen Shot 2021-01-25 at 11 42 35 AM
Adds the bar at the bottom.

src/components/CookieWarning.js
```diff
@@ -0,0 +1,56 @@
...
+      backgroundColor: 'white', borderTop: `6px solid ${colors.primary()}`
+    }
+  }, [
+    div({ style: { padding: '2rem' } }, ['Terra uses cookies to enable the proper functioning and security of our website,',
+      ' and to improve your experience on our site. By clicking Agree or continuing to use our site, you consent to the use of these functional',
+      ' cookies. If you do not wish to allow use of these cookies, you may tell us that by clicking on Reject. As a result, you will be unable',
+      ' to use our site. To find out more, read our ',
+      h(Link, { style: { textDecoration: 'underline' }, href: Nav.getLink('privacy') }, ['privacy policy']), '.']),
+    div({ style: { padding: '2rem', display: 'flex' } }, [
+      h(ButtonPrimary, { onClick: () => acceptCookies(true) }, ['Agree']),
```
andy7i, line 54: does `acceptCookies(true) ` set `acceptedCookies` = true ?

src/components/CookieWarning.js
```diff
@@ -0,0 +1,56 @@
...
+      h(ButtonSecondary, { style: { marginLeft: '2rem' }, onClick: () => { rejectCookies() } }, ['Reject'])
```
andy7i, line 55: I see that `rejectCookies()` is async here and `acceptCookies(true)` on the line above is not but they are called the same way. Async functions can be called the same way that sync functions can? I'm assuming yes because js, but just wanted to ask. Put another way, why doesn't this call need an `await`?

src/components/CookieWarning.js
```diff
@@ -0,0 +1,56 @@
...
+    const cookies = document.cookie.split(';')
+    acceptCookies(false)
+    await Ajax(signal).Runtimes.invalidateCookie()
+    // Expire all cookies
+    _.forEach(cookie => {
+      // Find an equals sign and uses it to grab the substring of the cookie that is its name
+      const eqPos = cookie.indexOf('=')
+      const cookieName = eqPos > -1 ? cookie.substr(0, eqPos) : cookie
+      document.cookie = `${cookieName}=;expires=Thu, 01 Jan 1970 00:00:00 GMT`
+    }, cookies)
```
andy7i, line 33: do we do this because we can't delete the user's cookies, so we expire them instead?

src/components/CookieWarning.js
```diff
@@ -0,0 +1,56 @@
...
+import colors from 'src/libs/colors'
+import * as Nav from 'src/libs/nav'
+import { getLocalPref, setLocalPref } from 'src/libs/prefs'
+import { authStore } from 'src/libs/state'
+import * as Utils from 'src/libs/utils'
+
+
+export const cookiesAcceptedKey = 'cookiesAccepted'
+
+export const CookieWarning = () => {
```
andy7i: how can I test this? what happens if I never hit accept or reject and try to use the site?

s-rubenstein: It works as planned. To test this, spin up a local UI server on this branch (`yarn install`, and then `yarn start` in the terra-ui directory)

If they never hit accept or reject, it leaves the popup open, but we consider that tacit acceptance of cookies.

src/components/CookieWarning.js
s-rubenstein: Yes, this is the standard way to delete cookies, and then the browser will clean them up. There are libraries which abstract this, but we do very little cookie management, so its probably fine not to add a library and to do it by hand.

src/components/CookieWarning.js
s-rubenstein: Yes, in the function here https://github.com/DataBiosphere/terra-ui/pull/2369/files#diff-c72ed808f8308541b6fbe2ff82b375ea71914747219e630c24c71053e075551eR22

src/components/CookieWarning.js
s-rubenstein: Ah! Yes, so it would need an `await` if we cared about the result. But because we don't care about the result, the lambda bound to onClick for reject is actually synchronous. It just pushes the async function on the evaluation stack and continues.

src/components/CookieWarning.js
andy7i: cool! I did see that and was surprised by the funky syntax

src/components/runtime-common.js
```diff
@@ -98,7 +100,7 @@ export const AuthenticatedCookieSetter = () => {
 
 export const PeriodicCookieSetter = () => {
   const signal = Utils.useCancellation()
-  Utils.usePollingEffect(
+  getLocalPref(cookiesAcceptedKey) !== false && Utils.usePollingEffect(
```
zarsky-broad: An effect shouldn't happen conditionally. Instead, conditionalize the rendering of this component where it's used.

src/libs/ajax.js
```diff
@@ -1056,6 +1056,10 @@ const Runtimes = signal => ({
     return res.json()
   },
 
+  invalidateCookie: () => {
+    return fetchLeo(`proxy/invalidateToken`, _.merge(authOpts(), { signal }))
```
```suggestion
    return fetchLeo('proxy/invalidateToken', _.merge(authOpts(), { signal }))
```

src/pages/PrivacyPolicy.js
```diff
@@ -101,8 +101,12 @@ We follow the laws of all the countries where we operate. We follow the laws of
 `
 
 const PrivacyPolicy = () => {
-  return div({ role: 'main', style: { padding: '1rem' } }, [
+  return div({ role: 'main', style: { padding: '1rem', marginBottom: 90 } }, [
     h(MarkdownViewer, {
+      // We want to cap the width of large amounts of text to ensure that they are readable.
+      style: {
+        maxWidth: '45rem'
+      },
```
zarsky-broad:
```suggestion
      style: { maxWidth: '45rem' },
```

src/pages/SignIn.js
```diff
@@ -1,37 +1,16 @@
...
-  }, [
-    `${getAppName()} uses cookies to enable sign on and other essential features when signed in, and to provide statistics to our development team regarding how the site is used. For more information, see our `,
-    h(Link, {
-      ...Utils.newTabLinkProps,
-      href: Nav.getLink('privacy')
-    }, ['privacy policy.'])
-  ])
-})
-
 export default class SignIn extends Component {
```
zarsky-broad: Since this no longer has any state, you might take this opportunity to convert it to a function component.

src/components/CookieWarning.js
zarsky-broad, line 55: In a hook component, there's no `this` binding to preserve, so functions don't need to be wrapped:
```suggestion
      h(ButtonSecondary, { style: { marginLeft: '2rem' }, onClick: rejectCookies } }, ['Reject'])
```

src/components/CookieWarning.js
```diff
@@ -0,0 +1,56 @@
...
+      position: 'fixed', height: 100, bottom: 0, width: '100%',
+      display: 'flex', alignItems: 'center', justifyContent: 'space-between',
+      backgroundColor: 'white', borderTop: `6px solid ${colors.primary()}`
+    }
+  }, [
+    div({ style: { padding: '2rem' } }, ['Terra uses cookies to enable the proper functioning and security of our website,',
+      ' and to improve your experience on our site. By clicking Agree or continuing to use our site, you consent to the use of these functional',
+      ' cookies. If you do not wish to allow use of these cookies, you may tell us that by clicking on Reject. As a result, you will be unable',
+      ' to use our site. To find out more, read our ',
+      h(Link, { style: { textDecoration: 'underline' }, href: Nav.getLink('privacy') }, ['privacy policy']), '.']),
```
zarsky-broad:
```suggestion
      h(Link, { style: { textDecoration: 'underline' }, href: Nav.getLink('privacy') }, ['privacy policy']), '.'
    ]),
```

src/components/CookieWarning.js
```diff
@@ -0,0 +1,56 @@
...
+    signOut()
+  }
+  return !(acceptedCookies) && div({
+    style: {
+      position: 'fixed', height: 100, bottom: 0, width: '100%',
+      display: 'flex', alignItems: 'center', justifyContent: 'space-between',
+      backgroundColor: 'white', borderTop: `6px solid ${colors.primary()}`
+    }
+  }, [
+    div({ style: { padding: '2rem' } }, ['Terra uses cookies to enable the proper functioning and security of our website,',
```
zarsky-broad:
```suggestion
    div({ style: { padding: '2rem' } }, [
      'Terra uses cookies to enable the proper functioning and security of our website,',
```
Optionally, you could combine these strings into a template string, which allows linebreaks.

src/components/CookieWarning.js
```diff
@@ -0,0 +1,56 @@
...
+    // Expire all cookies
+    _.forEach(cookie => {
+      // Find an equals sign and uses it to grab the substring of the cookie that is its name
+      const eqPos = cookie.indexOf('=')
+      const cookieName = eqPos > -1 ? cookie.substr(0, eqPos) : cookie
+      document.cookie = `${cookieName}=;expires=Thu, 01 Jan 1970 00:00:00 GMT`
+    }, cookies)
+    signOut()
+  }
+  return !(acceptedCookies) && div({
```
zarsky-broad:
```suggestion
  return !acceptedCookies && div({
```

src/components/CookieWarning.js
```diff
@@ -0,0 +1,56 @@
...
+import { getLocalPref, setLocalPref } from 'src/libs/prefs'
+import { authStore } from 'src/libs/state'
+import * as Utils from 'src/libs/utils'
+
+
+export const cookiesAcceptedKey = 'cookiesAccepted'
+
+export const CookieWarning = () => {
+  const signal = Utils.useCancellation()
+  const [acceptedCookies, setAcceptedCookies] = useState(getLocalPref(cookiesAcceptedKey))
```
zarsky-broad: Why is this using state rather than `Utils.useStore(authStore)`? That way you'd just have the single source of truth.

src/components/CookieWarning.js
zarsky-broad: Usually we might have this as a default export, since the file and the component have the same name.

Also, we prefer blank lines between functions in the body, for readability. Basically, anything multiline at the top level of the component should be padded.

zarsky-broad (commented):
Some notes

aednichols (commented):
I understand the regulatory need for this, but I suspect almost no users actually want to see it.

Can we make it say 50% smaller so that it occludes less of the UI people are actually interested in?

src/components/CookieWarning.js
s-rubenstein: The linter disagreed. It actually removed the blank lines I put between functions in the body, otherwise I would have done that.

src/components/CookieWarning.js
s-rubenstein: It may have been a remnant of a previous implementation, before I was storing that information in authStore. I will try that and see if it works.

src/components/CookieWarning.js
s-rubenstein: Great! Will do that.

src/components/CookieWarning.js
s-rubenstein: Hmmm… from a technical perspective I agree with you. However, my preference stylistically is to wrap, because it makes it much simpler of a diff to change if we need to add things to the onClick function, and is functionally equivalent. It also maintains stylistic consistency across the repo to always use the safer pattern, and not risk missing a lambda where it is important.

src/libs/ajax.js
s-rubenstein: Going to push back against this one. The backticks align with the setCookie endpoint above. I am happy to change both of them at once, but wanted to maintain consistency within the single Ajax object.

src/pages/SignIn.js
s-rubenstein: Yeah, sure! Will do.

src/components/CookieWarning.js
s-rubenstein: Nevermind, my linter was unhappy about spaces, fixed that.

s-rubenstein:
@aednichols there are no regulatory requirements that require us to have it this large, however we do want to make sure the text is not cut off at our smallest supported width. With the current height, we are cutting off a bit of the padding, but no text. If we made it too much shorter, we would start having text overflow issues, which compliance would certainly take issue with. We could make it smaller text, but that starts to be an accessibility issue.

aednichols:
Fair enough. Feel free to ignore my drive-by comment; I just lament how intrusive these things have gotten

src/libs/auth.js
```diff
@@ -184,6 +190,16 @@ authStore.subscribe(async (state, oldState) => {
   }
 })
 
+authStore.subscribe(state => {
+  // We can't guarantee that someone reopening the window is the same user,
+  // so we should not persist cookie acceptance across sessions for people signed out
+  // Per compliance recommendation, we could probably persist through a session, but
+  // that would require a second store in session storage instead of local storage
+  if (state.isSignedIn && state.cookiesAccepted !== getLocalPref(cookiesAcceptedKey)) {
+    setLocalPref(cookiesAcceptedKey, state.cookiesAccepted)
```
gpolumbo-broad, line 200: For users who stay signed in, does there need to be a cookie acceptance timeout for a session?

gpolumbo-broad (approved):

src/libs/auth.js
s-rubenstein: There wasn't a requirement for that :D

src/components/CookieWarning.js
```diff
@@ -1,28 +1,24 @@
...
-  const [acceptedCookies, setAcceptedCookies] = useState(getLocalPref(cookiesAcceptedKey))
-  authStore.subscribe(state => {
-    setAcceptedCookies(getLocalPref(cookiesAcceptedKey))
-  })
+  const { cookiesAccepted } = Utils.useStore(authStore)
+
   const acceptCookies = acceptedCookies => {
-    setLocalPref(cookiesAcceptedKey, acceptedCookies)
-    setAcceptedCookies(getLocalPref(cookiesAcceptedKey))
+    authStore.update(_.set('cookiesAccepted', acceptedCookies))
```
andy7i, line 20: where would I be able to see this? I don't see it appearing in local storage at all (I checked each localhost:3000 option). Actually, even if I hit accept and refresh the page, I get the banner again. 
cookies tab after hitting accept

src/components/CookieWarning.js
```diff
@@ -36,21 +32,28 @@ export const CookieWarning = () => {
...
-  return !(acceptedCookies) && div({
+
+  return !cookiesAccepted && div({
     style: {
       position: 'fixed', height: 100, bottom: 0, width: '100%',
       display: 'flex', alignItems: 'center', justifyContent: 'space-between',
-      backgroundColor: 'white', borderTop: `6px solid ${colors.primary()}`
+      backgroundColor: colors.dark(0.15), borderTop: `6px solid ${colors.primary()}`,
+      // Minimum supported page width
+      minWidth: 1200
```
andy7i: is this the minimum supported width for Terra? I did notice that if a user has a width of 1000px then they cannot see or scroll to the accept button. (I could not scroll the banner to the side)
cant see the buttons scrolled all the way to the right 1000px window

We don't support mobile devices right? I won't go there :)

andy7i (changes requested):

src/components/CookieWarning.js
s-rubenstein: We do not support mobile devices. The inability to scroll it at a smaller width is a problem, I will fix that. 1200px is our min-width, as set on the body file.

src/components/CookieWarning.js
s-rubenstein: So, you could see it in the Local Storage section on that area you are in. You're currently looking at cookies specifically. However, we don't save it for users who are signed out, because we can't verify that a user who comes back is the same. (E.g., going to the site on a library computer)

andy7i: I do see it now! It also seems to work as expected for a second user who has not yet accepted cookies (userA signs in, accepts, signs out, userB signs in and should see the cookies banner). I also tried logging in and out for both users.

src/components/CookieWarning.js
```diff
@@ -0,0 +1,59 @@
...
+  return !cookiesAccepted && div({
+    style: {
+      position: 'fixed', height: 100, bottom: 0, width: '100%',
+      display: 'flex', alignItems: 'center', justifyContent: 'space-between',
+      backgroundColor: colors.dark(0.15), borderTop: `6px solid ${colors.primary()}`
+    }
+  }, [
+    div({ style: { padding: '0.9rem 2rem', height: '100%', display: 'flex', alignItems: 'center' } }, [
+      div({ style: { overflowY: 'auto', height: '100%' } }, [
+        `Terra uses cookies to enable the proper functioning and security of our website,
```
zarsky-broad: Should this use `getAppName()`, btw?

andy7i (commented):
I was able to verify the behavior locally, but there's a lot of code changed since the original PR and I won't be able to review it this afternoon (sorry!)

zarsky-broad:
@andy7i fwiw it gets a lot smaller if you hide whitespace changes in the diff

src/components/CookieWarning.js
s-rubenstein: Definitely.

src/components/CookieWarning.js
```diff
@@ -36,21 +33,28 @@ export const CookieWarning = () => {
     }, cookies)
     signOut()
   }
-  return !(acceptedCookies) && div({
+  return !cookiesAccepted && div({
     style: {
       position: 'fixed', height: 100, bottom: 0, width: '100%',
       display: 'flex', alignItems: 'center', justifyContent: 'space-between',
-      backgroundColor: 'white', borderTop: `6px solid ${colors.primary()}`
+      backgroundColor: colors.dark(0.15), borderTop: `6px solid ${colors.primary()}`
```
andy7i: (especially) because I know you care about A11Y, I think the contrast suffers a bit with the darker background color -- lighthouse flags this as an issue with the Reject button and the privacy policy link.

Screen Shot 2021-01-29 at 6 09 36 PM

src/libs/prefs.js
```diff
@@ -3,6 +3,9 @@ import { getDynamic, setDynamic } from 'src/libs/browser-storage'
 
 
 const withUserPrefix = key => `${getUser().id}/${key}`
+const withUserPrefixForSpecifiedUserId = (userId, key) => `${userId}/${key}`
 
 export const getLocalPref = key => getDynamic(localStorage, withUserPrefix(key))
+// Needed for loading from localStorage on signin or signout event
+export const getLocalPrefForUserId = (userId, key) => getDynamic(localStorage, withUserPrefixForSpecifiedUserId(userId, key))
```
andy7i, line 10: I'm pretty sure it's safe, but just to be 100% sure, it's ok to give out the user ID right? (a one word answer is totally fine)

src/components/CookieWarning.js
```diff
@@ -36,21 +33,28 @@ export const CookieWarning = () => {
...
+
+    ]),
     div({ style: { padding: '2rem', display: 'flex' } }, [
       h(ButtonPrimary, { onClick: () => acceptCookies(true) }, ['Agree']),
       h(ButtonSecondary, { style: { marginLeft: '2rem' }, onClick: () => { rejectCookies() } }, ['Reject'])
     ])
   ])
 }
+
+export default CookieWarning
```
andy7i, line 60: just out of curiosity, why is it better to `export default` here instead of inline? (also noticed it was changed for `SignIn`). Is it easier to find the exported value when we specify it at the end of the file?
I see that `export default` is used when there is just 1 value so I think I understand that part at least.

andy7i (approved):
verified behavior/text for Terra (not the other apps)

1 change for a11y. 👍  assuming that change is added.

src/components/CookieWarning.js
zarsky-broad: See https://github.com/DataBiosphere/terra-ui/pull/2365#discussion_r567131664

andy7i: interesting, thanks!

src/libs/prefs.js
s-rubenstein: It is no more unsafe than what we are already doing. (since other storage mechanisms are also using that user information) I believe it is anonymous though.

PR #2370, opened 2021-01-25
Disable notebooks test

ehanna4:
Disabling the integration test for notebooks. This has been failing for over a month in either Staging or Alpha every night. Scheduled a mobbing session with the Terra-UI Champions to get help debugging. Until this gets sorted looking to reduce the failure notification which has gotten monotonous risking us ignoring other test failures.


breilly2 (commented):
👍   Looks good.

zarsky-broad (approved):
We could do this with fewer changes, I'd prefer if you didn't modify `tests/run-notebook.js` for example, I don't _think_ that's necessary

breilly2 (approved):
My thumb still stands. 👍

PR #2371, opened 2021-01-27
[CA-1162] Update warning notice on sign in page for anvil

marctalbott:
Ticket: [CA-1162](https://broadworkbench.atlassian.net/browse/CA-1162)
I tested this by using `configOverridesStore.set({isAnvil: true})` in my browser for my local Terra UI and verifying that the sign in text changed appropriately.

I did the same for base Terra, Firecloud, and BioDataCatalyst

### DIFF
--- a/src/pages/SignIn.js
+++ b/src/pages/SignIn.js
@@ -4,7 +4,7 @@ import { Clickable, HeroWrapper, Link } from 'src/components/common'
 import Modal from 'src/components/Modal'
 import SignInButton from 'src/components/SignInButton'
 import colors from 'src/libs/colors'
-import { isBioDataCatalyst, isFirecloud } from 'src/libs/config'
+import { isAnvil, isBioDataCatalyst, isFirecloud } from 'src/libs/config'
 import { getAppName } from 'src/libs/logos'
 import * as Nav from 'src/libs/nav'
 import * as Style from 'src/libs/style'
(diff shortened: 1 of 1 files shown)


src/pages/SignIn.js
```diff
@@ -96,6 +96,25 @@ export default class SignIn extends Component {
             }, ['https://www.federalregister.gov/documents/2018/03/14/2018-05176/privacy-act-of-1974-system-of-records'])
           ])
         ]),
+        isAnvil() && div({ style: { lineHeight: 1.5, marginTop: '2rem', paddingTop: '1rem', borderTop: Style.standardLine } }, [
```
s-rubenstein: Can we pull the style out into a shared style object, since they are identical? 

https://github.com/DataBiosphere/terra-ui/blob/dev/src/pages/workspaces/workspace/Data.js#L34 is an example of how to do that.

src/pages/SignIn.js
```diff
@@ -96,6 +96,25 @@ export default class SignIn extends Component {
...
+          div({ style: { fontWeight: 500 } }, ['WARNING NOTICE']),
```
s-rubenstein: Nit: I would rather do normal casing here, e.g., `'Warning Notice'` and apply the `textTransform: 'uppercase'` to the styles. It future proofs against the text changing and makes the idea of this div explicit in the styling.

gpolumbo-broad (approved):

src/pages/SignIn.js
```diff
@@ -4,13 +4,23 @@ import { Clickable, HeroWrapper, Link } from 'src/components/common'
...
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
 
+const styles = {
+  warningNoticeContainer: {
+    lineHeight: 1.5, marginTop: '2rem',
+    paddingTop: '1rem', borderTop: Style.standardLine
+  },
+  warningNotice: {
```
marctalbott, line 19: I wanted to use this on l.77 too, but wasn't sure how to partially apply the style since only part of that warning notice is uppercase

s-rubenstein: The way to do that would be to separate it out into two children within the div, and apply the styles to that one div. I am fine with punting on that though. (You would need to futz with the parent div so that the children appeared inline, which you could do by making it a flex container)

s-rubenstein (approved):
LGTM

s-rubenstein:
TOL: I wonder why the linter allowed for different indentation across the paragraphs…

PR #2372, opened 2021-01-28
[IA-2410] Update Leo calls with workspace v2 parameters.

jdcanas:
Tested this locally by running through all the pages effected and checking the leo calls in the network tab.

This ticket details a few different changes which this PR accomplishes: https://broadworkbench.atlassian.net/browse/IA-2410 .

Note that the label and env variables already exist on apps. On app, the label is named something slightly different (`saturnWorkspaceName`, see [here](https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/ajax.js#L1167)) and it cannot be changed easily because the galaxy app depends on it as is. I'm open to changing it here for consistency, but I did what was in the acceptance criteria for now.

### DIFF
--- a/src/components/NewRuntimeModal.js
+++ b/src/components/NewRuntimeModal.js
@@ -118,7 +118,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
   static propTypes = {
     runtimes: PropTypes.array,
     persistentDisks: PropTypes.array,
-    namespace: PropTypes.string.isRequired,
+    workspace: PropTypes.object.isRequired,
     onDismiss: PropTypes.func.isRequired,
     onSuccess: PropTypes.func.isRequired
   }
@@ -155,9 +155,8 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
     }
   }
 
-  makeWorkspaceObj() {
-    const { namespace, name } = this.props
-    return { workspace: { namespace, name } }
+  getWorkspaceObj() {
+    return this.props.workspace.workspace
   }
 
   getCurrentRuntime() {
@@ -227,7 +226,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
     )
 
     Ajax().Metrics.captureEvent(Events[metricsEvent], {
-      ...extractWorkspaceDetails(this.makeWorkspaceObj()),
+      ...extractWorkspaceDetails(this.getWorkspaceObj()),
       ..._.mapKeys(key => `newRuntime_${key}`, newRuntime),
       newRuntime_exists: !!newRuntime,
       newRuntime_cpus: newRuntime && newRuntimeCpus,
@@ -249,7 +248,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
     Utils.withBusyState(() => this.setState({ loading: true })),
     withErrorReporting('Error creating cloud environment')
   )(async () => {
-    const { onSuccess, namespace } = this.props
+    const { onSuccess } = this.props
     const { currentRuntimeDetails, currentPersistentDiskDetails } = this.state
     const { runtime: oldRuntime, persistentDisk: oldPersistentDisk } = this.getOldEnvironmentConfig()
     const { runtime: newRuntime, persistentDisk: newPersistentDisk } = this.getNewEnvironmentConfig()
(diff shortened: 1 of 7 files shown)


src/libs/ajax.js
```diff
@@ -1168,7 +1168,8 @@ const Apps = signal => ({
           diskConfig: {
             name: diskName,
             labels: {
-              saturnApplication: 'galaxy'
+              saturnApplication: 'galaxy',
+              workspaceName
```
s-rubenstein: This seems like a separate change, to me? From my understanding, this is passing the workspaceName as a label to Leo? Is that related to the workspace v2 parameters?

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -42,7 +42,7 @@ const NotebookLauncher = _.flow(
     showTabBar: false
   })
 )(
-  ({ queryParams, notebookName, workspace, workspace: { workspace: { namespace, name }, accessLevel, canCompute }, runtimes, persistentDisks, refreshRuntimes },
+  ({ queryParams, notebookName, workspace, workspace: { workspace: { namespace, name, bucketName }, accessLevel, canCompute }, runtimes, persistentDisks, refreshRuntimes },
```
s-rubenstein: TOL: (Existing issue, so no need to fix here) Why are we passing both workspace, and then a subset of the workspace fields? Could we remove the subset bit?

src/libs/ajax.js
jdcanas: Yes, it is part of the ticket https://broadworkbench.atlassian.net/browse/IA-2410. Not sure the exact use case, but it gives leo/the apps an awareness of the workspace for things like bucket access

kyuksel (approved):
👍🏼 ing based on the mob PR review

gpcarr:
looks like yarn build failed but once that's fixed, 👍

gpcarr (approved):

ehanna4:
so this failing test: import-dockstore-workflow, appears to be happening recently [Nightly Run](https://circleci.com/gh/DataBiosphere/terra-ui/26753) so not related to your changes but also not a flake - something seems to be up with SAM but shouldn't prevent you from merging. The failing run-notebooks test is a bit different than the usual issue we see so I'm going to re-run it. Will review again after they finish

ehanna4 (approved):
Signing off that the failures in Integration tests don't appear relevant to these changes
Been seeing a SAM issue recently (todayish) causing trouble in the tests and the typical 403 error in the notebooks test.

src/components/NewRuntimeModal.js
```diff
@@ -285,24 +286,32 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
       })
     }
 
+    const customEnvVars = {
+      WORKSPACE_NAME: name,
+      WORKSPACE_BUCKET: `gs://${bucketName}`,
+      GOOGLE_PROJECT: googleProject
```
rtitle, line 292: I think the Leo docker-compose already sets `GOOGLE_PROJECT` but shouldn't hurt to have it here, I think. 
https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/init-resources/jupyter-docker-compose-gce.yaml#L19

src/components/NewRuntimeModal.js
```diff
@@ -285,24 +286,32 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
       })
     }
 
+    const customEnvVars = {
+      WORKSPACE_NAME: name,
```
rtitle, line 290: Thinking through if setting `WORKSPACE_NAME` and `WORKSPACE_BUCKET` EVs will cause any issues.

Currently runtimes can still be used across workspaces. Leo has logic to dynamically set `WORKSPACE_NAME` and `WORKSPACE_BUCKET` depending on the current welder storage link. That way your notebook runtime is always connected to the workspace it's being used from.

I think that dynamic-ev-setting _should_ overwrite what's specified in `customEnvVars`. So I _think_ it should be okay to set it here. Once all workspaces are migrated to PPW we can remove the welder logic to dynamically set EVs (and maybe other simplifications too).

Does all that sound right? Maybe we could test using a notebook runtime from 2 workspaces, and make sure the EVs are still set correctly depending on the workspace in use.

rtitle:
Overall LGTM, one comment above about EVs. Just want to make sure we don't break anything while we're in this intermediate state of using `googleProject` but still sharing runtimes across workspaces.

src/components/NewRuntimeModal.js
jdcanas: Good call, I make two workspaces, created a notebook in each workspace, and then swapped between the two and ensure `printenv` behaved properly within a notebook.

It all seemed to work fine, even when I have both notebooks open in tabs and switch between

src/components/NewRuntimeModal.js
jdcanas: Yeah, I was thinking it shouldn't matter, I mainly added it for consistency between runtime and app custom env vars set in the UI.

rtitle (approved):
LGTM!

PR #2373, opened 2021-01-28
fix and re-enable run notebooks ticket (mob) [QA-1342]

zarsky-broad:
Working with The Mob, this should make the notebooks test much more resilient, and so we're re-enabling it!


panentheos (approved):
Looks good, assuming tests pass!

ehanna4 (approved):

PR #2374, opened 2021-01-29
load snapshot data [AS-516]

Shakespeared:
Load snapshot data in the data tab. Clicking the info-circle button by the snapshot name will show data about the snapshot itself instead of a data table for the selected entity


src/components/data/DataTable.js
```diff
@@ -235,7 +240,7 @@ const DataTable = props => {
                     }
                   }, [
                     h(Sortable, { sort, field: 'name', onSort: setSort }, [
-                      h(HeaderCell, [`${entityType}_id`])
+                      h(HeaderCell, [entityMetadata[entityType].idName])
```
davidangb, line 243: good call!

davidangb (commented):
code looks good to me, with the caveat that I still feel not-very-expert on this codebase.

tested locally, and everything works as expected, though I have some UX concerns (which I realize might need UX team or PO input). If you have a large number of non-snapshot tables, especially coupled with a large number of tables inside a snapshot, behavior gets pretty funky:
* if scrolled down very far, when clicking on a snapshot table, the table loads above the fold and I don't see it. Should we scroll to top when clicking on a table?
* if scrolled down very far, when clicking on a snapshot table, the spinner displays vertically-centered, which may also be above the fold, making me think my click did nothing.
* should the list of non-snapshot tables be collapsible, too?

I can share a workspace with you if you want to see this in action yourself.

Shakespeared:
@davidangb yeah if you could share a workspace that would be great, I don't think I have any with enough tables to check that out. @zarsky-broad and I can take a look and grab someone to advise if necessary.

Shakespeared:
@davidangb: @zarsky-broad and I changed the position of the loading spinner so it will display wherever you're looking (e.g. scrolled way down in the list of snapshot tables). 

The rest of your comments (scroll to top when loading, and potentially adding a collapse for the non-snapshot tables) we thought would be better discussed and handled separately from this ticket. We were thinking that some kind of transient notification that says 'data is loaded!' and when clicked scrolls you to the top might be good, but I think that's outside the scope of this ticket and would want us to talk to UX and/or PO first. 

If that sounds good to you, feel free to re-review with the spinner fix and we can discuss further UX improvements separately.

davidangb (approved):
:+1:.  I like the spinner change from a UX perspective.

I'm fine merging this PR as-is and following up with UX enhancements for scrolling/collapsing/etc  ... *if* you make sure those changes happen before launch and/or PO signs off on it. This PR is terrific progress but I am worried we can't launch with it as-is.

PR #2362, opened 2021-01-15
[Personal Hackathon] Call table, call details, call cache debug wizard [BW-486]

cjllanwarne:
A bunch of call details additions to the workflow dashboard in order to support a "Call Cache Debugging Wizard":

### Updated workflows page

image

### New Call Details page

image

### New Call Cache Debugging Wizard

image

### Tests:

* Comparing the diff between two Dev workflows:
  * http://localhost:3000/#workspaces/testing-job-manager/lots_of_workflows/job_history/4210652d-f68c-4faf-9d26-6d7354b07b87/049bd35c-68a7-4d36-8340-76fd875c3df6
  * http://localhost:3000/#workspaces/testing-job-manager/lots_of_workflows/job_history/1e55deba-28bb-4581-b290-5fa3b0e1629e/49a8b6b8-4b3c-4e0a-aaa1-ac37c575fa37

* Comparing the diff between two Dev workflows (currently "demonstrates" error handling because of a Cromwell bug):
  * http://localhost:3000/#workspaces/general-dev-billing-account/scale-abort-test/job_history/c4752583-839c-4319-bb0d-5eae15fa1fdc/d471802e-e1cc-405c-a816-536a6a7a69db
  * http://localhost:3000/#workspaces/general-dev-billing-account/scale-abort-test/job_history/b6647bde-603d-43d4-a24f-3b68bf04cf0e/2520d817-83ee-4e4d-961f-e582d29b1b6c

PR #2375, opened 2021-02-02
Personal Hackathon Part 1: Call Tables [BW-506]

cjllanwarne:
This is a hopefully more digestable "part 1" subset of https://github.com/DataBiosphere/terra-ui/pull/2362

Displays call-by-call summaries on the workflow dashboard page. Since we're already loading almost all of this information into browser memory already when we load workflow metadata, the overhead is not high compared to showing users nothing at all.

Example:

image

Recommended review order:

* Start with the changes to WorkflowDashboard
* Then look at the CallTable component that the dashboard is now displaying
* I had to move some parts of the "status line" generation around to let them look nice in the tables.

Recommended test pages (both in **alpha**):

- A workflow with a lot of calls: http://localhost:3000/#workspaces/cromwell-tests-billing-project/cromwell-tests-workspace/job_history/e60d7eab-4faa-4868-986e-5d86fc2b949d/e21fe7ab-8922-4552-9a66-3917925e3340
- Test workflow from the original "Workflows Dashboard" hackathon PR: http://localhost:3000/#workspaces/cromwell-tests-billing-project/cromwell-tests-workspace/job_history/d873e5da-054a-401e-ae8b-4d8c3adf03ad/b95fc8f8-4410-4e6c-accf-fb14b1766984

Note:

- There is still plenty of room for improvement! This PR isn't trying to boil the ocean, but just give a framework for showing off call details that are already being loaded into memory


src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -0,0 +1,98 @@
...
+import * as Utils from 'src/libs/utils'
+import FailuresModal from 'src/pages/workspaces/workspace/jobHistory/FailuresModal'
+
+
+const CallTable = ({ namespace, name, submissionId, workflowId, callName, callObjects }) => {
+  const [failuresModalParams, setFailuresModalParams] = useState()
+
+  return div([
+    h(FlexTable, {
+      height: _.min([callObjects.length * 100, 600]),
```
panentheos: This is a good approach if we're going to have multiple scrolling tables on the same page, but we should tighten up the logic to make it match the actual size used under the hood, which is 48px for both the header and row height. What about writing a helper function in `table.js` that takes a current number of rows and a desired maximum, and returns the container height based on that?

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -0,0 +1,98 @@
...
+      width: _.max([window.screen.width - 200, 0]),
```
panentheos: This will sample the current screen height once when rendering, but it won't respond if the browser is resized later. To handle that, wrap the table in an `AutoSizer` component which will continuously update the dimensions as needed (search for examples, the invocation is a little unusual). Since it looks like we're calculating the height ourselves, you can pass `disableHeight: true` on the `AutoSizer` so it only manages width.

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -0,0 +1,98 @@
...
+      rowCount: callObjects.length,
+      noContentMessage: 'No matching workflows',
```
panentheos: Should this be 'No matching calls`?

src/libs/ajax.js
```diff
@@ -650,9 +650,19 @@ const Workspaces = signal => ({
...
 
-          getWorkflow: async (workflowId, includeKey) => {
-            const res = await fetchRawls(`${submissionPath}/workflows/${workflowId}?${qs.stringify({ includeKey }, { arrayFormat: 'repeat' })}`, _.merge(authOpts(), { signal }))
-            return res.json()
+          // NB: This could one day perhaps redirect to CromIAM's 'workflow' like:
+          // workflow: workflowId => Ajax(signal).CromIAM.workflow(workflowId)
+          // But: Because of the slowness of asking via CromIAM, that's probably a non-starter for right now.
+          workflow: workflowId => {
+            return {
+              metadata: async (includeKey, excludeKey) => {
```
panentheos: TOL: Although it's not universally applied, I think it's nice in cases like this to bundle optional params into an object, e.g. `({ includeKey, excludeKey }) => ...`

That requires the caller to specify by name rather than argument order, which tends to make the downstream code clearer. It also would allow us to specify default values for one or more of the options.

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -0,0 +1,98 @@
...
+          cellRenderer: ({ rowIndex }) => {
+            const { shardIndex } = callObjects[rowIndex]
+            return h(TooltipCell, [shardIndex >= 0 ? shardIndex : 'N/A'])
+          }
+        }, {
+          size: { basis: 100, grow: 0 },
+          headerRenderer: () => 'Attempt',
+          cellRenderer: ({ rowIndex }) => {
+            const { attempt } = callObjects[rowIndex]
+            return h(TooltipCell, [attempt])
```
panentheos: `TooltipCell` is a good tool for exposing potentially long data, but since it creates visual noise when moving the mouse, we should try to avoid it when we can be reasonably sure the value won't be clipped.

src/pages/workspaces/workspace/jobHistory/FailuresModal.js
```diff
@@ -0,0 +1,19 @@
...
+    showButtons: false,
+    showX: true,
+    width: '50%'
+  }, [
+    `Failures in ${callFqn} / index ${index} / attempt ${attempt}`,
+    h(FailuresViewer, { failures })
+  ])
+}
+
+export default FailuresModal
```
panentheos: nit: Since this component is so small and only used in one place, it could be co-located in that file.

src/pages/workspaces/workspace/jobHistory/FailuresModal.js
```diff
@@ -0,0 +1,19 @@
...
+import FailuresViewer from 'src/pages/workspaces/workspace/jobHistory/FailuresViewer'
+
+
+const FailuresModal = ({ callFqn, index, attempt, failures, onDismiss }) => {
+  return h(Modal, {
+    title: `Error Messages`,
+    onDismiss,
+    showButtons: false,
+    showX: true,
+    width: '50%'
```
panentheos: nit: Does this need to vary based on browser width? If we want it to be larger than the default because of the content, we could pick a sensible fixed width.

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -0,0 +1,98 @@
...
+        {
+          size: { basis: 200, grow: 2 },
+          headerRenderer: () => 'Links',
+          cellRenderer: ({ rowIndex }) => {
+            const { failures, shardIndex: index, attempt } = callObjects[rowIndex]
+            const failureCount = _.size(failures)
+            return [
+              failures && h(Link, {
+                style: { marginLeft: '0.5rem' },
+                onClick: () => setFailuresModalParams({ callFqn: callName, index, attempt, failures })
```
panentheos: In most cases I would recommend reducing the amount of state, and e.g. storing the index of the selected row, and looking up the data dynamically. In this case, there's a complicating factor because we have a background polling loop that might update the underlying data while the modal is open. There are two things to think about regarding that:

1. What should the behavior be if the data changes out from under us? Common options are: A. continue showing the data we originally showed, B. show updated data in real time, C. close the modal.
2. Is there any potential for the calls array to reshuffle such that existing data changes index? That could be a factor if we try to do something dynamic.

In any case, there's at least some simplification we can do here: The `callName` doesn't vary within this component, so we don't need to store that in state.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -110,11 +118,16 @@ const WorkflowDashboard = _.flow(
     }), simplifiedFailures)
   }
 
+  const callNames = _.keys(calls)
```
panentheos: Technically objects are unordered, so we might want to e.g. sort these alphabetically to ensure some stability between updates.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -110,11 +118,16 @@ const WorkflowDashboard = _.flow(
...
+
   return div({ style: { padding: '1rem 2rem 2rem', flex: 1, display: 'flex', flexDirection: 'column' } }, [
     workflowDetailsBreadcrumbSubtitle(namespace, name, submissionId, workflowId),
+    workflow === undefined ?
+      div({ style: { fontStyle: 'italic', marginBottom: '1rem' } }, [`Fetching workflow metadata...`]) :
+      div({ style: { fontStyle: 'italic', marginBottom: '1rem' } }, [`Workflow metadata fetched in ${fetchTime}ms`]),
```
panentheos: This info is definitely helpful for debugging, but does showing the time help users? (aside from validating their frustration that this can take a while)

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -141,12 +154,9 @@ const WorkflowDashboard = _.flow(
             style: { display: 'flex', marginLeft: '1rem' }
           }, [icon('fileAlt', { size: 18 }), ' View execution log'])
         ])
-      ]),
-      makeSection('Call Statuses', [
-        workflow.calls ? statusCell(workflow) : div({ style: { marginTop: '0.5rem' } }, ['No calls have been started by this workflow.'])
       ])
     ]),
-    failures && h(Collapse, {
+    workflow !== undefined && failures && h(Collapse, {
```
panentheos: Since we have several pieces of content that are conditional on the workflow being loaded, we could bundle all of them into a `Fragment` and just do this check once.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -166,10 +176,32 @@ const WorkflowDashboard = _.flow(
...
+    workflow !== undefined && h(Collapse, {
+      title: div({ style: Style.elements.sectionHeader }, ['Calls']),
+      initialOpenState: true
+    }, [
+      div({ style: { marginLeft: '1rem' } },
+        [makeSection('Total Call Status Counts', [
+          !_.isEmpty(calls) ? statusCell(workflow) : div({ style: { marginTop: '0.5rem' } }, ['No calls have been started by this workflow.'])
+        ]),
+        !_.isEmpty(calls) && makeSection('Call Lists', [
+          _.map(callName => {
```
panentheos: When programmatically generating and rendering an array, we need to provide a unique `key` for each element, to help react maintain correctness and performance. There's usually a naturally available unique value that can be used (`callName` in this case). React will warn in the console if this isn't being done correctly, so that's a good way to verify.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -166,10 +176,32 @@ const WorkflowDashboard = _.flow(
...
+            return h(Collapse, {
+              style: { marginLeft: '1rem', marginTop: '0.5rem' },
+              title: div({ style: { ...Style.codeFont, ...Style.elements.sectionHeader } }, [`${callName} (x ${calls[callName].length})`]),
```
panentheos: nit: The unicode 'multiplication sign' `×` would be preferable to the letter `x` here, especially for screen readers.

src/pages/workspaces/workspace/jobHistory/CallTable.js
cjllanwarne: I've moved callName out.

1. Presumably if the data on the page refreshes the entire call table would re-render - and the modal selection would be zero'd out. I think ideally I think I would prefer to preserve the "which modal is open" state from the previous page render in that case, but I don't know the best way to handle that?
2. Yes, new calls can be added to the call list depending on when and in which order the Cromwell engine starts running them. Eg if you have `scatter(range(2))` over `{ call A -> call B }`, it's possible for shard 2 of call B to start even while shard 1 is still running call A. The call list for B will change dynamically as and when the shards start processing

aednichols:
This is still on my list but planning to look after Brett's requested changes go in to avoid race conditions

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: "One day" ™️  it would be nice to sort by start time instead of alphabetically, but I agree that alphabetically is infinitely better than random for now.

zarsky-broad: Why not use [`_.sortBy`](https://lodash.com/docs/4.17.15#sortBy) when you save `calls` into state?

zarsky-broad: (that would mean you had to convert `calls` to an array, so that might be way out of scope...)

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: Personally I like having it:

* Easier than checking the network tab when debugging
* I find putting a spotlight on something you want to improve helps to encourage that improvement
* Gives the users an extra piece of information to feed back to us (and lets them and us quantify just how frustrating the UI experience is)
* Gives the users a nudge/clue if something changes in their workflow and suddenly the response is significantly slower - that maybe there's something strange going on behind the scenes for them to either ask about (and maybe we can advise) or work through themselves.

I also checked on Google and they still have their "x responses in y time" message at the top of search results which makes me feel a little better about keeping it... But - I can see your point that this is not strictly necessary for users to see, so I'd be happy to remove this if you and/or Adam feel strongly that it's a mistake to include it.

cjllanwarne:
@aednichols now with the first set of requested changes. Unlikely to change again now before Monday.

config/dev.json
```diff
@@ -1,27 +1,27 @@
...
-  "bondUrlRoot": "https://broad-bond-dev.appspot.com",
-  "calhounUrlRoot": "https://terra-calhoun-dev.appspot.com",
-  "dataRepoUrlRoot": "https://jade.datarepo-dev.broadinstitute.org",
-  "devUrlRoot": "https://bvdp-saturn-dev.appspot.com",
+  "agoraUrlRoot": "https://agora.dsde-alpha.broadinstitute.org",
+  "bardRoot": "https://terra-bard-alpha.appspot.com",
+  "bondUrlRoot": "https://broad-bond-alpha.appspot.com",
+  "calhounUrlRoot": "https://terra-calhoun-alpha.appspot.com",
+  "dataRepoUrlRoot": "https://data.alpha.envs-terra.bio",
+  "devUrlRoot": "https://bvdp-saturn-alpha.appspot.com",
```
aednichols: Pretty sure we don't want to check in any changes to this file

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -110,66 +118,102 @@ const WorkflowDashboard = _.flow(
...
+                !_.isEmpty(calls) ? statusCell(workflow) : div({ style: { marginTop: '0.5rem' } }, ['No calls have been started by this workflow.'])
+              ]),
+              !_.isEmpty(calls) && makeSection('Call Lists', [
+                _.map(callName => {
+                  return h(Collapse, {
+                    key: callName,
+                    style: { marginLeft: '1rem', marginTop: '0.5rem' },
+                    title: div({ style: { ...Style.codeFont, ...Style.elements.sectionHeader } }, [`${callName} (× ${calls[callName].length})`]),
+                    initialOpenState: !_.isEmpty(_.filter(c => { return c.executionStatus !== 'Done' })(calls[callName]))
+                    // initialOpenState: calls[callName].length < 10
```
aednichols: Delete if not being checked in
```suggestion
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -110,66 +118,102 @@ const WorkflowDashboard = _.flow(
...
+            div({ style: { marginLeft: '1rem' } },
+              [makeSection('Total Call Status Counts', [
+                !_.isEmpty(calls) ? statusCell(workflow) : div({ style: { marginTop: '0.5rem' } }, ['No calls have been started by this workflow.'])
+              ]),
+              !_.isEmpty(calls) && makeSection('Call Lists', [
+                _.map(callName => {
+                  return h(Collapse, {
+                    key: callName,
+                    style: { marginLeft: '1rem', marginTop: '0.5rem' },
+                    title: div({ style: { ...Style.codeFont, ...Style.elements.sectionHeader } }, [`${callName} (× ${calls[callName].length})`]),
```
aednichols: Personal style preference, if in doubt omit parentheses
```suggestion
                    title: div({ style: { ...Style.codeFont, ...Style.elements.sectionHeader } }, [`${callName} × ${calls[callName].length}`]),
```

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -0,0 +1,101 @@
...
+import { FailuresModal } from 'src/pages/workspaces/workspace/jobHistory/FailuresViewer'
+
+
+const CallTable = ({ namespace, name, submissionId, workflowId, callName, callObjects }) => {
+  const [failuresModalParams, setFailuresModalParams] = useState()
+
+  return div([
+    h(AutoSizer, { disableHeight: true }, [
+      ({ width }) => h(FlexTable, {
+        height: tableHeight({ actualRows: callObjects.length, maxRows: 10 }),
```
aednichols: On a Mac notebook it was not immediately obvious to me that this table is scrollable.

This is because scroll bars are hidden by default on devices with trackpads, and the cells are perfectly aligned to their container.

If the container displayed e.g. 10.5 cells it would be more obvious that the table continues below.

Screenshot_2021-02-08 cromwell-tests-workspace - Workflow Dashboard

aednichols (commented):
Fantastic improvement, a few gremlins that prevent an outright approval.

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -0,0 +1,101 @@
...
+              const { failures, shardIndex: index, attempt } = callObjects[rowIndex]
+              const failureCount = _.size(failures)
+              return [
+                failures && h(Link, {
+                  style: { marginLeft: '0.5rem' },
+                  onClick: () => setFailuresModalParams({ index, attempt, failures })
+                }, [
+                  div({ style: { display: 'flex', alignItems: 'center' } }, [
+                    icon('warning-standard', { size: 18, style: { color: colors.warning(), marginRight: '0.5rem' } }),
+                    `${failureCount} Error Message${failureCount > 1 ? 's' : ''}`
```
aednichols: Marketing hat: a message stating that e.g. an attempt was preempted is not an error or at all unexpected.
```suggestion
                    `${failureCount} Message${failureCount > 1 ? 's' : ''}`
```

src/pages/workspaces/workspace/jobHistory/FailuresViewer.js
```diff
@@ -0,0 +1,44 @@
...
+    enableClipboard: false,
+    displayDataTypes: false,
+    displayObjectSize: false,
+    src: restructureFailures(failures)
+  })
+}
+
+export const FailuresModal = ({ callFqn, index, attempt, failures, onDismiss }) => {
+  return h(Modal, {
+    title: `Error Messages`,
```
aednichols: More marketing hat
```suggestion
    title: `Messages`,
```

config/dev.json
cjllanwarne: Oops - definitely not!

src/pages/workspaces/workspace/jobHistory/CallTable.js
cjllanwarne: I wonder whether @panentheos has thoughts on this. I was also tempted to have an extra half-row teaser to clue people in that there was more content below... but it felt kind of hacky. Is this a standard thing to do?

src/pages/workspaces/workspace/jobHistory/CallTable.js
cjllanwarne: I think the UI is valid in carrying this forward as what the metadata is describing as an error. IMO we could perhaps look to instead move "you got preempted" out of the `failures` section of metadata and into a new `warnings` or `messages` section? (I can think of other possible non-error-level messages that we might want to be able to highlight to users if the UI had such a section).

cjllanwarne: Or maybe we could describe them as "errors" if the call has failed, but just "messages" if it succeeded?

aednichols: IMO the UI doesn't have to match the underlying data structure, it should serve the user as clearly as possible and may need to add/remove meaning from the data

aednichols (approved):

src/pages/workspaces/workspace/jobHistory/CallTable.js
cjllanwarne: My reluctance is that by just calling them "messages" here that real errors would also just be labeled as "Messages" which seems worse to me than calling preemption messages "errors". 

That said - I'd be happy to see a little logic in a follow-up PR around whether to call individual messages "errors" vs "messages"

cjllanwarne: Also happy to be overridden if you still feel strongly that you still want me to change it. I guess I'm reluctant but not totally fixed in my opinion

aednichols (approved):

src/pages/workspaces/workspace/jobHistory/CallTable.js
panentheos: That actually sounds like a pretty good solution. There isn't a universal pattern for hinting that a table is scrollable. Some apps use shading tricks to accomplish that, but since we don't do that, adding an extra half row might be a good way to handle it.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
panentheos: Sounds good to me!

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -110,66 +118,101 @@ const WorkflowDashboard = _.flow(
...
+              [makeSection('Total Call Status Counts', [
+                !_.isEmpty(calls) ? statusCell(workflow) : div({ style: { marginTop: '0.5rem' } }, ['No calls have been started by this workflow.'])
+              ]),
+              !_.isEmpty(calls) && makeSection('Call Lists', [
+                _.map(callName => {
+                  return h(Collapse, {
+                    key: callName,
+                    style: { marginLeft: '1rem', marginTop: '0.5rem' },
+                    title: div({ style: { ...Style.codeFont, ...Style.elements.sectionHeader } }, [`${callName} × ${calls[callName].length}`]),
+                    initialOpenState: !_.isEmpty(_.filter(c => { return c.executionStatus !== 'Done' })(calls[callName]))
```
panentheos: I think this could be simplified to:
```suggestion
                    initialOpenState: !_.every({ executionStatus: 'Done' }, calls[callName])
```
(with the bonus that it will perform better because `_.every` will halt as soon as it finds a false case)

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -110,66 +118,101 @@ const WorkflowDashboard = _.flow(
     }), simplifiedFailures)
   }
 
+  const callNames = _.sortedUniq(_.keys(calls))
```
panentheos: `sortedUniq` isn't what we want here, try:
```suggestion
  const callNames = _.sortBy(_.identity, _.keys(calls))
```

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -0,0 +1,101 @@
...
+
+
+const CallTable = ({ namespace, name, submissionId, workflowId, callName, callObjects }) => {
+  const [failuresModalParams, setFailuresModalParams] = useState()
+
+  return div([
+    h(AutoSizer, { disableHeight: true }, [
+      ({ width }) => h(FlexTable, {
+        height: tableHeight({ actualRows: callObjects.length, maxRows: 10.5 }), // The half-row here hints at there being extra rows if scrolled
+        width: _.max([width, 0]),
```
panentheos: Do we need to do math on this? Usually you can just pass the width directly.

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -0,0 +1,101 @@
...
+              }
+            }
+          },
+          {
+            size: { basis: 200, grow: 2 },
+            headerRenderer: () => 'Links',
+            cellRenderer: ({ rowIndex }) => {
+              const { failures, shardIndex: index, attempt } = callObjects[rowIndex]
+              const failureCount = _.size(failures)
+              return [
```
panentheos: This is returning an array, which is triggering a key warning. You can return `null` to render nothing, so we could just return a ternary expression with the `Link` or `null`

panentheos (approved):
See comments, then looks good. Thanks!

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
cjllanwarne: Much nicer! I also didn't realize that the destructuring syntax could be used as predicate functions. Very very nice!

src/pages/workspaces/workspace/jobHistory/CallTable.js
cjllanwarne: No - I think this is an unintended carry-over from the old way I was doing this (where I was subtracting from the total page width). Will switch this to just be the width we're passed by the autosizer

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
panentheos: It's not destructuring, we're actually passing an object there. It's a lodash-specific feature: anywhere you pass a predicate function, you can substitute an object, and lodash will interpret it as "for each key in the passed object, test that the item has a matching value at that key"

PR #2376, opened 2021-02-02
Personal Hackathon Part 2: Cache miss debug wizard link on CacheMiss'd calls [BW-508]

cjllanwarne:
This is a hopefully more digestible "part 2" subset of #2362 to follow on from #2375's "part 1".

Implements the Call Cache Miss debug wizard as a link from the "calls" table for calls which tried but missed a call cache.

#### Example of the call cache modal in action:

See updated image in comment https://github.com/DataBiosphere/terra-ui/pull/2376#issuecomment-786845476 

#### Good example workflows to test against:

- Against the same **ALPHA** workflows as in #2375 to make sure the UX has not significantly regressed:
  - A workflow with a lot of calls: http://localhost:3000/#workspaces/cromwell-tests-billing-project/cromwell-tests-workspace/job_history/e60d7eab-4faa-4868-986e-5d86fc2b949d/e21fe7ab-8922-4552-9a66-3917925e3340
  - Test workflow from the original "Workflows Dashboard" hackathon PR: http://localhost:3000/#workspaces/cromwell-tests-billing-project/cromwell-tests-workspace/job_history/d873e5da-054a-401e-ae8b-4d8c3adf03ad/b95fc8f8-4410-4e6c-accf-fb14b1766984
- Against the same **DEV** workflows as in #2362 demonstrating the debug wizard:
  - Comparing the diff between two Dev workflows:
    - http://localhost:3000/#workspaces/testing-job-manager/lots_of_workflows/job_history/4210652d-f68c-4faf-9d26-6d7354b07b87/049bd35c-68a7-4d36-8340-76fd875c3df6
    - http://localhost:3000/#workspaces/testing-job-manager/lots_of_workflows/job_history/1e55deba-28bb-4581-b290-5fa3b0e1629e/49a8b6b8-4b3c-4e0a-aaa1-ac37c575fa37
  - Comparing the diff between two Dev workflows (currently "demonstrates" error handling because of a Cromwell bug):
    - http://localhost:3000/#workspaces/general-dev-billing-account/scale-abort-test/job_history/c4752583-839c-4319-bb0d-5eae15fa1fdc/d471802e-e1cc-405c-a816-536a6a7a69db
    - http://localhost:3000/#workspaces/general-dev-billing-account/scale-abort-test/job_history/b6647bde-603d-43d4-a24f-3b68bf04cf0e/2520d817-83ee-4e4d-961f-e582d29b1b6c
- Against a **DEV** workflow which has both a cache miss AND failure message:
  - http://localhost:3000/#workspaces/general-dev-billing-account/scale-abort-test/job_history/b5c5554e-2021-4ad4-ab58-34d1fe5f12f8/d37be105-559f-4e4a-aaf2-543b37aa51fa
- Against a **DEV** workflow which has a non-trivial drop down options:
  - http://localhost:3000/#workspaces/general-dev-billing-account/cjl_test_workspace/job_history/d4e06b20-3aad-480a-942b-ab5b48237f01/57734cb5-1452-49ad-a9d4-a8bd125a3e49
  - Compare `helloInScatter` shard 5's cache miss against shard 5 from the same task in `82b31604-c576-4c8d-b1dd-8f04f1e930d5`
  - Compare `read` against the same workflow to see another example of the "Cromwell can't call cache diff arrays" bug
  - Compare against `f9fbf5a7-2a17-4bf6-a661-0180a5905dc2` for a workflow which has exactly one shard index

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -130,7 +130,6 @@ const fetchSam = _.flow(withUrlPrefix(`${getConfig().samUrlRoot}/`), withAppIden
 const fetchBuckets = _.flow(withRequesterPays, withUrlPrefix('https://storage.googleapis.com/'))(fetchOk)
 const fetchGoogleBilling = withUrlPrefix('https://cloudbilling.googleapis.com/v1/', fetchOk)
 const fetchRawls = _.flow(withUrlPrefix(`${getConfig().rawlsUrlRoot}/api/`), withAppIdentifier)(fetchOk)
-const fetchCromIAM = _.flow(withUrlPrefix(`${getConfig().orchestrationUrlRoot}/api/`), withAppIdentifier)(fetchOk)
 const fetchDataRepo = withUrlPrefix(`${getConfig().dataRepoUrlRoot}/api/`, fetchOk)
 const fetchLeo = withUrlPrefix(`${getConfig().leoUrlRoot}/`, fetchOk)
 const fetchDockstore = withUrlPrefix(`${getConfig().dockstoreUrlRoot}/api/`, fetchOk)
@@ -502,7 +501,10 @@ const attributesUpdateOps = _.flow(
 )
 
 const CromIAM = signal => ({
-  callCacheDiff: async (thisWorkflowId, thisCallFqn, thisIndex, thatWorkflowId, thatCallFqn, thatIndex) => {
+  callCacheDiff: async (thisWorkflow, thatWorkflow) => {
+    const { workflowId: thisWorkflowId, callFqn: thisCallFqn, index: thisIndex } = thisWorkflow
+    const { workflowId: thatWorkflowId, callFqn: thatCallFqn, index: thatIndex } = thatWorkflow
+
     const params = {
       workflowA: thisWorkflowId,
       callA: thisCallFqn,
@@ -511,12 +513,12 @@ const CromIAM = signal => ({
       callB: thatCallFqn,
       indexB: (thatIndex !== -1) ? thatIndex : undefined
     }
-    const res = await fetchCromIAM(`workflows/v1/callcaching/diff?${qs.stringify(params)}`, _.merge(authOpts(), { signal }))
+    const res = await fetchOrchestration(`api/workflows/v1/callcaching/diff?${qs.stringify(params)}`, _.merge(authOpts(), { signal }))
     return res.json()
   },
 
   workflowMetadata: async (workflowId, includeKey, excludeKey) => {
-    const res = await fetchCromIAM(`workflows/v1/${workflowId}/metadata?${qs.stringify({ includeKey, excludeKey }, { arrayFormat: 'repeat' })}`, _.merge(authOpts(), { signal }))
+    const res = await fetchOrchestration(`api/workflows/v1/${workflowId}/metadata?${qs.stringify({ includeKey, excludeKey }, { arrayFormat: 'repeat' })}`, _.merge(authOpts(), { signal }))
     return res.json()
   }
 })
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -130,6 +130,7 @@ const fetchSam = _.flow(withUrlPrefix(`${getConfig().samUrlRoot}/`), withAppIden
 const fetchBuckets = _.flow(withRequesterPays, withUrlPrefix('https://storage.googleapis.com/'))(fetchOk)
 const fetchGoogleBilling = withUrlPrefix('https://cloudbilling.googleapis.com/v1/', fetchOk)
 const fetchRawls = _.flow(withUrlPrefix(`${getConfig().rawlsUrlRoot}/api/`), withAppIdentifier)(fetchOk)
+const fetchCromIAM = _.flow(withUrlPrefix(`${getConfig().orchestrationUrlRoot}/api/`), withAppIdentifier)(fetchOk)
 const fetchDataRepo = withUrlPrefix(`${getConfig().dataRepoUrlRoot}/api/`, fetchOk)
 const fetchLeo = withUrlPrefix(`${getConfig().leoUrlRoot}/`, fetchOk)
 const fetchDockstore = withUrlPrefix(`${getConfig().dockstoreUrlRoot}/api/`, fetchOk)
@@ -499,6 +500,27 @@ const attributesUpdateOps = _.flow(
       [{ op: 'AddUpdateAttribute', attributeName: k, addUpdateAttribute: v }]
   })
 )
+
+const CromIAM = signal => ({
+  callCacheDiff: async (thisWorkflowId, thisCallFqn, thisIndex, thatWorkflowId, thatCallFqn, thatIndex) => {
+    const params = {
+      workflowA: thisWorkflowId,
+      callA: thisCallFqn,
+      indexA: (thisIndex !== -1) ? thisIndex : undefined,
+      workflowB: thatWorkflowId,
+      callB: thatCallFqn,
+      indexB: (thatIndex !== -1) ? thatIndex : undefined
+    }
+    const res = await fetchCromIAM(`workflows/v1/callcaching/diff?${qs.stringify(params)}`, _.merge(authOpts(), { signal }))
+    return res.json()
+  },
+
+  workflowMetadata: async (workflowId, includeKey, excludeKey) => {
+    const res = await fetchCromIAM(`workflows/v1/${workflowId}/metadata?${qs.stringify({ includeKey, excludeKey }, { arrayFormat: 'repeat' })}`, _.merge(authOpts(), { signal }))
+    return res.json()
+  }
+})
+
 
 const Workspaces = signal => ({
   list: async fields => {
@@ -1324,7 +1346,8 @@ export const Ajax = signal => {
     Martha: Martha(signal),
     Duos: Duos(signal),
     Metrics: Metrics(signal),
-    Disks: Disks(signal)
+    Disks: Disks(signal),
+    CromIAM: CromIAM(signal)
   }
 }
 
--- a/src/libs/icon-dict.js
+++ b/src/libs/icon-dict.js
@@ -2,7 +2,7 @@ import { faClipboard, faClock, faClone, faEye, faFileAlt, faFolder, faFolderOpen
 import {
   faArrowLeft, faArrowRight, faBan, faCaretDown, faChalkboard, faCheck, faCheckCircle, faCircle, faCloud, faCog, faCreditCard, faDownload,
   faEllipsisV, faExclamationCircle, faExclamationTriangle, faFileInvoiceDollar, faGripHorizontal, faInfoCircle, faLock, faLongArrowAltDown,
-  faLongArrowAltUp, faMagic, faMinusCircle, faPause, faPen, faPlay, faPlus, faPlusCircle, faQuestion, faQuestionCircle, faSearch, faShareAlt, faSquare as faSquareSolid,
+  faLongArrowAltUp, faMinusCircle, faPause, faPen, faPlay, faPlus, faPlusCircle, faQuestion, faQuestionCircle, faSearch, faShareAlt, faSquare as faSquareSolid,
   faTachometerAlt, faTasks, faTerminal, faTrashAlt, faVirus
 } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
@@ -81,7 +81,6 @@ const iconDict = {
   lock: fa(faLock),
   'long-arrow-alt-down': fa(faLongArrowAltDown),
   'long-arrow-alt-up': fa(faLongArrowAltUp),
-  magic: fa(faMagic),
   'minus-circle': fa(faMinusCircle),
   pause: fa(faPause),
   play: fa(faPlay),
--- a/src/libs/icon-dict.js
+++ b/src/libs/icon-dict.js
@@ -2,7 +2,7 @@ import { faClipboard, faClock, faClone, faEye, faFileAlt, faFolder, faFolderOpen
 import {
   faArrowLeft, faArrowRight, faBan, faCaretDown, faChalkboard, faCheck, faCheckCircle, faCircle, faCloud, faCog, faCreditCard, faDownload,
   faEllipsisV, faExclamationCircle, faExclamationTriangle, faFileInvoiceDollar, faGripHorizontal, faInfoCircle, faLock, faLongArrowAltDown,
-  faLongArrowAltUp, faMinusCircle, faPause, faPen, faPlay, faPlus, faPlusCircle, faQuestion, faQuestionCircle, faSearch, faShareAlt, faSquare as faSquareSolid,
+  faLongArrowAltUp, faMagic, faMinusCircle, faPause, faPen, faPlay, faPlus, faPlusCircle, faQuestion, faQuestionCircle, faSearch, faShareAlt, faSquare as faSquareSolid,
   faTachometerAlt, faTasks, faTerminal, faTrashAlt, faVirus
 } from '@fortawesome/free-solid-svg-icons'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
@@ -81,6 +81,7 @@ const iconDict = {
   lock: fa(faLock),
   'long-arrow-alt-down': fa(faLongArrowAltDown),
   'long-arrow-alt-up': fa(faLongArrowAltUp),
+  magic: fa(faMagic),
   'minus-circle': fa(faMinusCircle),
   pause: fa(faPause),
   play: fa(faPlay),
--- a/src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
+++ b/src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
@@ -1,0 +1,263 @@
+import _ from 'lodash/fp'
+import { Fragment, useState } from 'react'
+import { div, h, hr } from 'react-hyperscript-helpers'
+import ReactJson from 'react-json-view'
+import Select from 'react-select'
+import { ButtonPrimary, ButtonSecondary } from 'src/components/common'
+import ErrorView from 'src/components/ErrorView'
+import { TextInput } from 'src/components/input'
+import { breadcrumbHistoryCaret } from 'src/components/job-common'
+import Modal from 'src/components/Modal'
+import { Ajax } from 'src/libs/ajax'
+import * as Style from 'src/libs/style'
+import * as Utils from 'src/libs/utils'
+
+
+const CallCacheWizard = ({
+  onDismiss, workflowId, callFqn, attempt, index
+}) => {
+  /*
+   * State setup
+   */
+
+  const [otherWorkflowId, setOtherWorkflowId] = useState()
+  const [otherWorkflowMetadata, setOtherWorkflowMetadata] = useState()
+  const [otherCallFqn, setOtherCallFqn] = useState()
+  const [otherIndex, setOtherIndex] = useState()
+  const [otherAttemptSucceeded, setOtherAttemptSucceeded] = useState()
+  const [otherCallSelected, setOtherCallSelected] = useState(false)
+  const [diff, setDiff] = useState()
+  const [metadataFetchError, setMetadataFetchError] = useState()
+  const [diffError, setDiffError] = useState()
+
+  const signal = Utils.useCancellation()
+
+  /*
+   * Data Fetchers
+   */
+
+  const readCalls = async otherWf => {
+    try {
+      const includeKey = [
+        'end', 'start', 'executionStatus'
+      ]
+      const excludeKey = []
+      // const wf = await Ajax(signal).Workspaces.workspace(otherNs, otherWs).submission(otherSub).getWorkflow(otherWf, includeKey, excludeKey)
+      const wf = await Ajax(signal).CromIAM.workflowMetadata(otherWf, includeKey, excludeKey)
+      setOtherWorkflowMetadata(wf)
+    } catch (error) {
+      console.log(error)
+      if (error instanceof Response) setMetadataFetchError(await error.text())
+      else if (_.isObject(error)) setMetadataFetchError(JSON.stringify(error))
+      else setMetadataFetchError(error)
+    }
+  }
+
+  const fetchDiff = async (otherWf, otherCall, otherIx) => {
+    try {
+      const diff = await Ajax(signal).CromIAM.callCacheDiff(workflowId, callFqn, Number(index), otherWf, otherCall, Number(otherIx))
+      setDiff(diff)
+    } catch (error) {
+      console.log(error)
+      if (error instanceof Response) setDiffError(await error.text())
+      else if (_.isObject(error)) setDiffError(JSON.stringify(error))
+      else setDiffError(error)
+    }
+  }
+
+  const otherCallFqnSelectionOptions = _.flow(
+    _.keys,
+    _.map(name => { return { value: name, label: name } })
+  )
+
+  const resetDiffResult = () => {
+    setDiff(undefined)
+    setDiffError(undefined)
+  }
+
+  const resetCallSelection = () => {
+    resetDiffResult()
+    setOtherCallFqn(undefined)
+    setOtherIndex(undefined)
+    setOtherAttemptSucceeded(undefined)
+    setOtherCallSelected(false)
+  }
+
+  const resetWorkflowSelection = () => {
+    resetCallSelection()
+    setOtherWorkflowId(undefined)
+    setMetadataFetchError(undefined)
+  }
+
+  /*
+   * Page render
+   */
+
+  const divider = hr({ style: { width: '100%', border: '1px ridge lightgray' } })
+
+  const step1 = () => {
+    return h(Fragment, [
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, ...Style.noWrapEllipsis } }, ['Step 1: Select the workflow you expected to cache from']),
+      div({ style: { marginTop: '0.5rem', marginBottom: '1rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+        div({ style: { paddingRight: '0.5rem' } }, ['Workflow ID:']),
+        div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [h(TextInput, { style: Style.codeFont, id: 'otherWorkflowId' })]),
+        div([h(ButtonPrimary, {
+          style: { float: 'right' },
+          onClick: () => {
+            const otherWf = document.getElementById('otherWorkflowId').value
+            setOtherWorkflowId(otherWf)
+            readCalls(otherWf)
+          }
+        }, ['Continue >'])])
+      ]),
+      divider,
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, color: 'lightgray', ...Style.noWrapEllipsis } }, ['Step 2: Select the call you expected to cache from']),
+      divider,
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, color: 'lightgray', ...Style.noWrapEllipsis } }, ['Result: View cache diff'])
+    ])
+  }
+  const step2 = () => {
+    const otherCallIndexSelector = (metadata, fqn) => {
+      const shards = _.flow(
+        _.map(c => Number(c.shardIndex)),
+        _.uniq
+      )(metadata.calls[fqn])
+
+      if (shards.length === 1) {
+        if (otherIndex !== shards[0]) { setOtherIndex(shards[0]) }
+        if (shards[0] === -1) {
+          return 'N/A (this call was not scattered)'
+        } else {
+          return `${shards[0]} (exactly one shard in the scatter)`
+        }
+      } else {
+        const shardOptions = _.map(i => { return { value: i, label: i } }).apply(shards)
+        return h(Select, { id: 'otherCallIndex', options: shardOptions, onChange: i => setOtherIndex(i) })
+      }
+    }
+
+    const otherCallAttemptDecision = (metadata, fqn, index) => {
+      const successfulAttempt = _.flow(
+        _.filter(c => c.shardIndex === index && c.executionStatus === 'Done'),
+        _.map(c => c.attempt),
+        _.first
+      )(metadata.calls[fqn])
+
+      if (otherAttemptSucceeded !== successfulAttempt) setOtherAttemptSucceeded(successfulAttempt)
+
+      if (successfulAttempt) {
+        return `${successfulAttempt} (the first successful attempt of this call)`
+      } else {
+        return 'This index of the call did not succeed'
+      }
+    }
+
+    return h(Fragment, [
+      div({ style: { display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Step 1: Selected workflow B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, otherWorkflowId)
+        ]),
+        h(ButtonSecondary, { style: { paddingLeft: '1rem', height: '20px', color: 'darkred', justifyContent: 'right' }, onClick: () => resetWorkflowSelection() }, ['[X]'])
+      ]),
+      metadataFetchError && h(ErrorView, { error: metadataFetchError }),
+      divider,
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, ...Style.noWrapEllipsis } }, ['Step 2: Select which call in that workflow you expected to cache from']),
+      !metadataFetchError && (otherWorkflowMetadata ?
+        div([
+          div({ style: { marginTop: '1rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            div({ style: { paddingRight: '0.5rem' } }, ['Call FQN:']),
+            div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [
+              h(Select, { id: 'otherCallFqn', options: otherCallFqnSelectionOptions(otherWorkflowMetadata.calls), onChange: v => setOtherCallFqn(v.value) })
+            ])
+          ]),
+          div({ style: { marginTop: '0.25rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            div({ style: { paddingRight: '0.5rem' } }, ['Shard Index:']),
+            otherCallFqn ?
+              div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [otherCallIndexSelector(otherWorkflowMetadata, otherCallFqn)]) :
+              'Select a call FQN first'
+          ]),
+          div({ style: { marginTop: '0.25rem', marginBottom: '1rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            div({ style: { paddingRight: '0.5rem' } }, ['Attempt:']),
+            (!!otherCallFqn && otherIndex !== undefined) ?
+              div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [otherCallAttemptDecision(otherWorkflowMetadata, otherCallFqn, otherIndex)]) :
+              'Select a call FQN and an index first'
+          ]),
+          !!(otherCallFqn && otherIndex !== undefined && otherAttemptSucceeded) && h(ButtonPrimary, { style: { float: 'right' }, onClick: () => { fetchDiff(otherWorkflowId, otherCallFqn, otherIndex); setOtherCallSelected(true) } }, ['Compare Diff >'])
+        ]) :
+        'Loading other workflow calls...'),
+      divider,
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, color: 'lightgray', ...Style.noWrapEllipsis } }, ['Result: View cache diff'])
+    ])
+  }
+
+  const compareDiffs = () => {
+    return h(Fragment, [
+      div({ style: { display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Step 1: Selected workflow B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, otherWorkflowId)
+        ]),
+        h(ButtonSecondary, { style: { paddingLeft: '1rem', height: '20px', color: 'darkred', justifyContent: 'right' }, onClick: () => resetWorkflowSelection() }, ['[X]'])
+      ]),
+      divider,
+      div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Step 2: Selected call B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, otherCallFqn),
+          breadcrumbHistoryCaret,
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, `index ${otherIndex}`)
+        ]),
+        h(ButtonSecondary, { style: { justifyContent: 'right', paddingLeft: '1rem', height: '20px', color: 'darkred' }, onClick: () => resetCallSelection() }, ['[X]'])
+      ]),
+      divider,
+      div({ style: { display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, ['Result: View cache diff']),
+      diffError ?
+        h(ErrorView, { error: diffError }) :
+        (diff ?
+          [div({ style: { marginTop: '0.5rem', marginBottom: '0.5rem' } }, ['Note: the diff is expressed in terms of hashes of values rather than raw values because it is hashes that determine cache hits.']),
+            h(ReactJson, {
+              style: { whiteSpace: 'pre-wrap', border: 'ridge', padding: '0.5rem' },
+              name: false,
+              shouldCollapse: ({ name }) => { return name === 'callA' || name === 'callB' },
+              enableClipboard: false,
+              displayDataTypes: false,
+              displayObjectSize: false,
+              src: { hashDifferential: diff.hashDifferential, ...diff }
+            })] :
+          'Cache diff loading...')
+    ])
+  }
+
+  const chooseStep = () => {
+    if (!otherWorkflowId) {
+      return step1()
+    } else if (!otherCallSelected) {
+      return step2()
+    } else {
+      return compareDiffs()
+    }
+  }
+
+  return h(Modal, {
+    title: 'Call Cache Debugging Wizard',
+    onDismiss,
+    width: '50%',
+    showButtons: false,
+    showX: true
+  }, [
+    div({ style: { padding: '1rem 2rem 2rem', flex: 1, display: 'flex', flexDirection: 'column' } }, [
+      'Debugging cache miss for call A:',
+      div({ style: { marginTop: '0.5rem', flex: 1, display: 'flex', alignItems: 'center', flexDirection: 'row' } }, [
+        div({ style: { fontSize: 16, fontWeight: 500, ...Style.codeFont } }, [workflowId]), breadcrumbHistoryCaret,
+        div({ style: { fontSize: 16, fontWeight: 500, ...Style.codeFont } }, [callFqn]),
+        Number(index) >= 0 && h(Fragment, [breadcrumbHistoryCaret, 'index', div({ style: { marginLeft: '0.25rem', fontSize: 16, fontWeight: 500, ...Style.codeFont } }, [index])]),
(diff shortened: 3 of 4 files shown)


src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,263 @@
...
+   * Data Fetchers
+   */
+
+  const readCalls = async otherWf => {
+    try {
+      const includeKey = [
+        'end', 'start', 'executionStatus'
+      ]
+      const excludeKey = []
+      // const wf = await Ajax(signal).Workspaces.workspace(otherNs, otherWs).submission(otherSub).getWorkflow(otherWf, includeKey, excludeKey)
```
rsasch: nit: remove commented-out line?

cjllanwarne:
Thanks to Joy for a number of recommendations on the layout. This now looks more like:

image

image

src/libs/ajax.js
```diff
@@ -130,6 +130,7 @@ const fetchSam = _.flow(withUrlPrefix(`${getConfig().samUrlRoot}/`), withAppIden
 const fetchBuckets = _.flow(withRequesterPays, withUrlPrefix('https://storage.googleapis.com/'))(fetchOk)
 const fetchGoogleBilling = withUrlPrefix('https://cloudbilling.googleapis.com/v1/', fetchOk)
 const fetchRawls = _.flow(withUrlPrefix(`${getConfig().rawlsUrlRoot}/api/`), withAppIdentifier)(fetchOk)
+const fetchCromIAM = _.flow(withUrlPrefix(`${getConfig().orchestrationUrlRoot}/api/`), withAppIdentifier)(fetchOk)
```
panentheos: This name is a bit confusing since it's just calling orch. I think we should skip this and just use `fetchOrchestration` below.

src/libs/ajax.js
```diff
@@ -500,6 +501,27 @@ const attributesUpdateOps = _.flow(
   })
 )
 
+const CromIAM = signal => ({
+  callCacheDiff: async (thisWorkflowId, thisCallFqn, thisIndex, thatWorkflowId, thatCallFqn, thatIndex) => {
```
panentheos: This is a lot of parameters, I think it would be good to combine them into either a single options object, or two objects having `workflow`, `call`, and `index` fields, whichever seems clearer.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      const includeKey = [
+        'end', 'start', 'executionStatus'
+      ]
+      const excludeKey = []
+      const wf = await Ajax(signal).CromIAM.workflowMetadata(otherWf, includeKey, excludeKey)
+      setOtherWorkflowMetadata(wf)
+    } catch (error) {
+      console.log(error)
+      if (error instanceof Response) setMetadataFetchError(await error.text())
+      else if (_.isObject(error)) setMetadataFetchError(JSON.stringify(error))
```
panentheos: What does this case handle? The `ErrorView` should already have some logic for handling different kinds of values.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+  const readCalls = async otherWf => {
+    try {
+      const includeKey = [
+        'end', 'start', 'executionStatus'
+      ]
+      const excludeKey = []
+      const wf = await Ajax(signal).CromIAM.workflowMetadata(otherWf, includeKey, excludeKey)
+      setOtherWorkflowMetadata(wf)
+    } catch (error) {
+      console.log(error)
```
panentheos: We should remove this.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      else setMetadataFetchError(error)
+    }
+  }
+
+  const fetchDiff = async (otherWf, otherCall, otherIx) => {
+    try {
+      const diff = await Ajax(signal).CromIAM.callCacheDiff(workflowId, callFqn, Number(index), otherWf, otherCall, Number(otherIx))
+      setDiff(diff)
+    } catch (error) {
+      console.log(error)
```
panentheos: We should remove this.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      if (error instanceof Response) setDiffError(await error.text())
+      else if (_.isObject(error)) setDiffError(JSON.stringify(error))
+      else setDiffError(error)
```
panentheos, line 68: Are we doing custom error handling here because this call can time out the same way metadata retrieval can?

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      console.log(error)
+      if (error instanceof Response) setMetadataFetchError(await error.text())
+      else if (_.isObject(error)) setMetadataFetchError(JSON.stringify(error))
+      else setMetadataFetchError(error)
+    }
+  }
+
+  const fetchDiff = async (otherWf, otherCall, otherIx) => {
+    try {
+      const diff = await Ajax(signal).CromIAM.callCacheDiff(workflowId, callFqn, Number(index), otherWf, otherCall, Number(otherIx))
```
panentheos: These indexes look like they're already numbers, so we probably don't need to convert them here.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      console.log(error)
+      if (error instanceof Response) setDiffError(await error.text())
+      else if (_.isObject(error)) setDiffError(JSON.stringify(error))
+      else setDiffError(error)
+    }
+  }
+
+  const otherCallFqnSelectionOptions = _.flow(
+    _.keys,
+    _.map(name => { return { value: name, label: name } })
```
panentheos: nit: Conventionally, single-line functions should use the braceless expression form (which needs parens because of the object in this case), e.g.
```suggestion
    _.map(name => ({ value: name, label: name }))
```

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+   */
+
+  const divider = hr({ style: { width: '100%', marginTop: '1rem', marginBottom: '1rem', border: '1px ridge lightgray' } })
+
+  const step1 = () => {
+    return h(Fragment, [
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, ...Style.noWrapEllipsis } }, ['Step 1: Select the workflow you expected to cache from']),
+      div({ style: { marginTop: '0.5rem', marginBottom: '1rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+        div({ style: { paddingRight: '0.5rem' } }, ['Workflow ID:']),
+        div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [h(TextInput, { style: Style.codeFont, id: 'otherWorkflowId' })]),
```
panentheos: This creates an 'uncontrolled' input, which means it's not managed by React, which isn't typical. If we add
```
value: otherWorkflowId, onChange: setOtherWorkflowId
```
to the props, it makes it 'controlled', and the current value will be updated in state on every keystroke. That means we don't have to reach into the DOM to get its value below.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+        div([h(ButtonPrimary, {
+          style: { float: 'right' },
```
panentheos: We shouldn't be using 'float', generally. Also, the div wrapper here is unnecessary.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+          onClick: () => {
+            const otherWf = document.getElementById('otherWorkflowId').value
+            setOtherWorkflowId(otherWf)
+            readCalls(otherWf)
+          }
+        }, ['Continue >'])])
```
panentheos: I don't think the `>` is necessary here, the text already communicates the action, and a screen reader would probably read this as "Continue greater than button".

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      ]),
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, color: 'lightgray', ...Style.noWrapEllipsis } }, ['Step 2: Select the call you expected to cache from']),
```
panentheos: The contrast ratio on this text is too light to be legible. TOL: Do we need to show this here? I suspect that the inclination to gray this out is an indication that it's not actually relevant yet, in which case we should just not show anything.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
panentheos: Also, this input doesn't have an accessible label. We could convert the text above into a `label`, and then link them together with an `IdContainer`. You'll know it's wired up correctly if you can click on the text to focus the input.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+  }
+  const step2 = () => {
+    const otherCallIndexSelector = (metadata, fqn) => {
+      const shards = _.flow(
+        _.map(c => Number(c.shardIndex)),
+        _.uniq
+      )(metadata.calls[fqn])
+
+      if (shards.length === 1) {
+        if (otherIndex !== shards[0]) { setOtherIndex(shards[0]) }
```
panentheos: This is setting state in the render code path, which should almost always be avoided. When you have semi-dependent state like this, the general solution is to reconcile/reset the 'child' state when you set the 'parent', but in this case it looks like we might not even need to do that, because of how the form gets reset. It might be enough to just have the final value be a computation of `shards.length === 1 ? shards[0] : otherIndex`.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    const otherCallAttemptDecision = (metadata, fqn, index) => {
+      const successfulAttempt = _.flow(
+        _.filter(c => c.shardIndex === index && c.executionStatus === 'Done'),
+        _.map(c => c.attempt),
+        _.first
+      )(metadata.calls[fqn])
+
+      if (otherAttemptSucceeded !== successfulAttempt) setOtherAttemptSucceeded(successfulAttempt)
```
panentheos: This is setting state in the render code path, but it looks like this doesn't even need to be state. We could compute this value as needed from the other state.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+        return h(Select, { options: shardOptions, onChange: i => { setOtherIndex(i.value) } })
+      }
+    }
+
+    const otherCallAttemptDecision = (metadata, fqn, index) => {
+      const successfulAttempt = _.flow(
+        _.filter(c => c.shardIndex === index && c.executionStatus === 'Done'),
+        _.map(c => c.attempt),
+        _.first
+      )(metadata.calls[fqn])
```
panentheos: Since we just want one, we could use `_.find` to simplify and save some work.
```suggestion
      const successfulAttempt = _.find({ shardIndex: index, executionStatus: 'Done' }, metadata.calls[fqn])?.attempt
```

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+
+  const step1 = () => {
+    return h(Fragment, [
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, ...Style.noWrapEllipsis } }, ['Step 1: Select the workflow you expected to cache from']),
+      div({ style: { marginTop: '0.5rem', marginBottom: '1rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+        div({ style: { paddingRight: '0.5rem' } }, ['Workflow ID:']),
+        div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [h(TextInput, { style: Style.codeFont, id: 'otherWorkflowId' })]),
+        div([h(ButtonPrimary, {
+          style: { float: 'right' },
+          onClick: () => {
```
panentheos, line 121: We should disable the button if the text input is empty (see above comment able making the input controlled first)

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      }
+    }
+
+    return h(Fragment, [
+      div({ style: { display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Selected workflow B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, otherWorkflowId)
+        ]),
+        h(ButtonSecondary, { style: { textTransform: 'none', paddingLeft: '1rem', height: '20px', justifyContent: 'right' }, onClick: () => resetWorkflowSelection() }, ['Reset'])
```
panentheos: Instead of overriding styles here, we can use the `Link` component with an `onClick` prop to get something that looks like a link, even though it's actually a button.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      ]),
+      metadataFetchError && h(ErrorView, { error: metadataFetchError }),
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, ...Style.noWrapEllipsis } }, ['Step 2: Select which call in that workflow you expected to cache from']),
```
panentheos: We should avoid using `noWrapEllipsis` unless we specifically want that behavior. This is a static message, and even if it was longer, it would be fine for it to wrap to a second line.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      !metadataFetchError && (otherWorkflowMetadata ?
+        div([
+          div({ style: { marginTop: '1rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            div({ style: { paddingRight: '0.5rem' } }, ['Call FQN:']),
+            div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [
+              h(Select, { id: 'otherCallFqn', options: otherCallFqnSelectionOptions(otherWorkflowMetadata.calls), onChange: v => setOtherCallFqn(v.value) })
```
panentheos: This select should have an accessible label. We could wire up the text above with an `IdContainer`

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+        return `${successfulAttempt} (the first successful attempt of this call)`
+      } else {
+        return 'This index of the call did not succeed'
+      }
+    }
+
+    return h(Fragment, [
+      div({ style: { display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Selected workflow B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
```
panentheos: nit: This `100px` looks like an arbitrary value. `flex: 1` should be suitable here.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+        ]),
+        h(ButtonSecondary, { style: { textTransform: 'none', paddingLeft: '1rem', height: '20px', justifyContent: 'right' }, onClick: () => resetWorkflowSelection() }, ['Reset'])
+      ]),
+      metadataFetchError && h(ErrorView, { error: metadataFetchError }),
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, ...Style.noWrapEllipsis } }, ['Step 2: Select which call in that workflow you expected to cache from']),
+      !metadataFetchError && (otherWorkflowMetadata ?
+        div([
+          div({ style: { marginTop: '1rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            div({ style: { paddingRight: '0.5rem' } }, ['Call FQN:']),
+            div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [
```
panentheos: The `flex` values here and below might be overly complex. A useful pattern is to have one or more fixed size elements, and one with `flex: 1`, which takes up all remaining space.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+              div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [otherCallIndexSelector(otherWorkflowMetadata, otherCallFqn)]) :
+              'Select a call FQN first'
+          ]),
+          div({ style: { marginTop: '0.25rem', marginBottom: '1rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            div({ style: { paddingRight: '0.5rem' } }, ['Attempt:']),
+            (!!otherCallFqn && otherIndex !== undefined) ?
+              div({ style: { paddingRight: '0.5rem', flex: '2 1 auto' } }, [otherCallAttemptDecision(otherWorkflowMetadata, otherCallFqn, otherIndex)]) :
+              'Select a call FQN and an index first'
+          ]),
+          !!(otherCallFqn && otherIndex !== undefined && otherAttemptSucceeded) && h(ButtonPrimary, { style: { float: 'right' }, onClick: () => { fetchDiff(otherWorkflowId, otherCallFqn, otherIndex); setOtherCallSelected(true) } }, ['Compare Diff >'])
```
panentheos: nit: Some of the lines here are pretty long. It might be nice to make some of the props objects or children arrays multiline for readability.

panentheos: We should remove `float` here

panentheos: The `>` isn't necessary in the button here.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, `index ${otherIndex}`)
+        ]),
+        h(ButtonSecondary, { style: { textTransform: 'none', paddingLeft: '1rem', height: '20px', justifyContent: 'right' }, onClick: () => resetCallSelection() }, ['Reset'])
+      ]),
+      divider,
+      div({ style: { display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, ['Result: View cache diff']),
+      diffError ?
+        h(ErrorView, { error: diffError }) :
+        (diff ?
+          [div({ style: { marginTop: '0.5rem', marginBottom: '0.5rem' } }, ['Note: the diff is expressed in terms of hashes of values rather than raw values because it is hashes that determine cache hits.']),
```
panentheos: The array around the div is causing a key warning. We should be able to just remove it.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, otherWorkflowId)
+        ]),
+        h(ButtonSecondary, { style: { textTransform: 'none', paddingLeft: '1rem', height: '20px', justifyContent: 'right' }, onClick: () => resetWorkflowSelection() }, ['Reset'])
+      ]),
+      div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Selected call B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, otherCallFqn),
+          breadcrumbHistoryCaret,
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, `index ${otherIndex}`)
```
panentheos: These `noWrapEllipsis` usages don't work because the container is asserting the content width as the minimum size. Try adding `minWidth: 0` to the container.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      div({ style: { paddingTop: '0.5rem', fontSize: 16, fontWeight: 500, color: 'lightgray', ...Style.noWrapEllipsis } }, ['Result: View cache diff'])
+    ])
+  }
+
+  const compareDiffs = () => {
+    return h(Fragment, [
+      div({ style: { paddingBottom: '0.5rem', display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Selected workflow B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
+          div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, otherWorkflowId)
```
panentheos: TOL: `noWrapEllipsis` is a good tool when the layout is constrained, e.g. inside tables or other fixed-size rows. Here, we could consider just letting the content flow (perhaps with `overflow-wrap` if there might be long tokens),

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      divider,
+      div({ style: { display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, ['Result: View cache diff']),
+      diffError ?
+        h(ErrorView, { error: diffError }) :
+        (diff ?
+          [div({ style: { marginTop: '0.5rem', marginBottom: '0.5rem' } }, ['Note: the diff is expressed in terms of hashes of values rather than raw values because it is hashes that determine cache hits.']),
+            h(ReactJson, {
+              style: { whiteSpace: 'pre-wrap', border: 'ridge', padding: '0.5rem' },
+              name: false,
+              shouldCollapse: ({ name }) => { return name === 'callA' || name === 'callB' },
```
panentheos: nit: Braceless expression would be preferred here, e.g.
```suggestion
              shouldCollapse: ({ name }) => name === 'callA' || name === 'callB',
```

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+              enableClipboard: false,
+              displayDataTypes: false,
+              displayObjectSize: false,
+              src: { hashDifferential: diff.hashDifferential, ...diff }
```
panentheos: This looks like it could just be:
```suggestion
              src: diff
```

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+            })] :
+          'Cache diff loading...')
+    ])
+  }
+
+  const chooseStep = () => {
+    if (!otherWorkflowId) {
```
panentheos: nit: `Utils.cond` is the preferred way to write decision trees like this

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+    } else if (!otherCallSelected) {
+      return step2()
+    } else {
+      return compareDiffs()
+    }
+  }
+
+  return h(Modal, {
+    title: [
+      icon('search'),
```
panentheos: We don't usually put icons in modal titles. I don't think the icon adds clarity here, so we could remove it.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+      ' Call Cache Miss Debugging Wizard'
+    ],
+    onDismiss,
+    width: '50%',
```
panentheos: This will make the width dependent on screen size, which probably isn't what we want. A fixed value will be more predictable.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,270 @@
...
+    showButtons: false,
+    showX: true
+  }, [
+    div({ style: { padding: '1rem 2rem 2rem', flex: 1, display: 'flex', flexDirection: 'column' } }, [
```
panentheos: This wrapper seems unnecessary. The modal content already has flex column layout and the standard amount of padding.

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -62,9 +64,20 @@ const CallTable = ({ namespace, name, submissionId, workflowId, callName, callOb
             size: { basis: 200, grow: 2 },
             headerRenderer: () => 'Call Caching Result',
             cellRenderer: ({ rowIndex }) => {
-              const { callCaching: { effectiveCallCachingMode, result } = {} } = callObjects[rowIndex]
+              const { shardIndex: index, attempt, callCaching: { effectiveCallCachingMode, result } = {} } = callObjects[rowIndex]
               if (effectiveCallCachingMode === 'ReadAndWriteCache' || effectiveCallCachingMode === 'ReadCache') {
-                return result ? h(TooltipCell, [result]) : div({ style: { color: colors.dark(0.7) } }, ['No Information'])
+                return result ? [
```
panentheos: This array is generating React key warnings. If we change it to a `Fragment`, it should fix the issue and allow us to remove the `key` below.

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -62,9 +64,20 @@ const CallTable = ({ namespace, name, submissionId, workflowId, callName, callOb
...
+                  h(TooltipCell, [result]),
+                  result === 'Cache Miss' && h(Link, {
+                    key: 'cc',
+                    style: { marginLeft: '0.5rem' },
+                    tooltip: 'Call Cache Debug Wizard',
```
panentheos, line 74: For accessibility, this should have an `aria-label` with this text as well.

src/libs/icon-dict.js
```diff
@@ -81,6 +81,7 @@ const iconDict = {
   lock: fa(faLock),
   'long-arrow-alt-down': fa(faLongArrowAltDown),
   'long-arrow-alt-up': fa(faLongArrowAltUp),
+  magic: fa(faMagic),
```
salonishah11: Sorry if this is a naive question, but what does this do?

salonishah11 (commented):
While resizing the window, I noticed that the text overflows outside the modal onto the background page. I am not sure regarding the exact terminology, but maybe it could be text-wrapped or hidden upon such overflow? Or have horizontal scroll bar?

image

src/libs/icon-dict.js
cjllanwarne: This is importing the "magic wand" icon from fontawesome. This was the icon I used to be using for the cache diff wizard but since it's now using `search` instead, we no longer need to import this icon and I should just remove it.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: There are two calls made from the wizard at different stages - 

1. Fetching the "other" workflow's metadata so that we can list calls in the drop down
2. Fetching the call cache diff between two workflows

Either of them can fail, independently, for a variety of different reasons. The most common failure in (1) is invalid workflow ID or a workflow which the user doesn't have access to. A timeout is also possible. For (2) there are also a couple of edge cases which can cause the endpoint to give us an error instead of a correct response. Eg this example from the PR description:

* Comparing the diff between two Dev workflows (currently "demonstrates" error handling because of a Cromwell bug):
http://localhost:3000/#workspaces/general-dev-billing-account/scale-abort-test/job_history/c4752583-839c-4319-bb0d-5eae15fa1fdc/d471802e-e1cc-405c-a816-536a6a7a69db
http://localhost:3000/#workspaces/general-dev-billing-account/scale-abort-test/job_history/b6647bde-603d-43d4-a24f-3b68bf04cf0e/2520d817-83ee-4e4d-961f-e582d29b1b6c

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: I believe they were coming in as strings from the drop down selector

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: I like that idea. Yes, the idea was "show which step you're currently in (and how many are left afterwards) so that people can tell how far through the process they are... but since it's two-steps that probably shouldn't really be a concern and I like the tidiness of not showing it at all

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: This is a really good point. I remember stumbling over it when I was writing this and wasn't sure what I did wrong, but now it makes total sense. I'll try to refactor this entire section to not set state from inside the renderer (for both here and below).

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: 👍  and in fact I managed to simplify even further. I don't think users will care which attempt succeeded, only whether the shard eventually succeeded or not.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: This seems to make my text box jump over to the left again. Any alternatives to keep the layout looking good?

With `float`:

image


Without:

image

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: Fixed by removing altogether, as suggested below.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: This was to make `hashDifferential` appear first

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: I tried to refactor into `Utils.cond` style but it caused horrible problems (functions being returned instead of values, etc..). Maybe we could pair on this to see what I was doing wrong?

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: Removing it seemed to make the content stick more to the sides of the modal. Feels worse to me, but maybe the layout can be recreated in a better way? (or maybe it really is better without squashing the content into the middle?)

Currently:

image


Without the wrapper `div`:

image

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: I seem to be getting `Warning: A component is changing an uncontrolled input of type undefined to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components`, I'm not sure whether it's related to this change or not?

cjllanwarne: Removing the `id` seemed to remove this warning

cjllanwarne: Removing `id` created much scarier errors, but changing the default to `''` seems to have solved both problems!

src/pages/workspaces/workspace/jobHistory/CallTable.js
cjllanwarne: Would like to check that I did this right

salonishah11:
It looks like the CircleCI build failed because of the unused variable warning posted by `lgtm-com` above
```'ButtonSecondary' is defined but never used```

**Comment by salonishah11** [comment]
After selecting the shard index, it looks like the value is not entirely visible because of the width of drop-down input. Would it be possible to make it bigger?

image

**Comment by cjllanwarne** [comment]
@salonishah11 thanks! 

image

**Review by salonishah11 — APPROVED** [review]
LGTM 👍

File: src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js [hunk 36]
```diff
@@ -0,0 +1,284 @@
...
+      div(['Debugging workflow A: ']),
+      div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
+        div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, workflowId)
+      ])
+    ]),
+    div({ style: { paddingTop: '0.5rem', display: 'flex', flexDirection: 'row', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+      div(['Debugging call A: ']),
+      div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
+        div({ style: { marginLeft: '0.5rem', ...Style.noWrapEllipsis, ...Style.codeFont } }, callFqn),
+        index && index >= 0 && [breadcrumbHistoryCaret,
```
panentheos: When `index` is `0`, this conditional is short-circuiting and rendering the `0`. Do we need to guard against this being potentially undefined? If so, we should explicitly check `index !== undefined`

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,284 @@
+import _ from 'lodash/fp'
+import { Fragment, useState } from 'react'
+import { div, h, hr, label } from 'react-hyperscript-helpers'
+import ReactJson from 'react-json-view'
+import Select from 'react-select'
```
panentheos: We have our own wrapper around `Select` in `src/components/common` that includes things like default styling and wiring up the `id` correctly. If we use that, it should fix the label accessibility.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,284 @@
...
+        div([
+          div({ style: { marginTop: '1rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            h(IdContainer, [id => h(Fragment, [
+              label({ htmlFor: id, style: { paddingRight: '0.5rem' } }, ['Call name:']),
+              div({ style: { flex: '1' } }, [
+                h(Select, {
+                  id,
+                  isSearchable: false,
+                  options: otherCallFqnSelectionOptions(otherWorkflowMetadata.calls),
+                  onChange: v => {
```
panentheos, line 165: This `Select` is currently 'uncontrolled' because we're not passing its `value`. Once we switch to our wrapped version of the component (see above), we should add `value: otherCallFqnDropdownValue` as a prop.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,284 @@
...
+            ])])
+          ]),
+          otherCallFqnDropdownValue && div({ style: { marginTop: '0.25rem', marginBottom: '0.5rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            h(IdContainer, [id => h(Fragment, [
+              label({ htmlFor: id, style: { paddingRight: '0.5rem' } }, ['Shard Index:']),
+              div({ style: { flex: '1' } }, [Utils.cond(
+                [selectedCallIndex === -1, () => 'N/A (not scattered)'],
+                () => {
+                  const options = _.uniqBy('value', _.map(({ shardIndex: i }) => { return { value: i, label: i } }, otherWorkflowMetadata.calls[otherCallFqnDropdownValue]))
+                  return h(Select, { id, isSearchable: false, options, onChange: i => { setOtherIndexDropdownValue(i.value) } })
```
panentheos: This `Select` is currently uncontrolled, which means the clearing of `otherIndexDropdownValue` in the previous `Select` isn't affecting this one the way we intend. After we switch to the wrapped version of the component (see above), we should add `value: otherIndexDropdownValue` as a prop.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,284 @@
...
+                })
+              ])
+            ])])
+          ]),
+          otherCallFqnDropdownValue && div({ style: { marginTop: '0.25rem', marginBottom: '0.5rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            h(IdContainer, [id => h(Fragment, [
+              label({ htmlFor: id, style: { paddingRight: '0.5rem' } }, ['Shard Index:']),
+              div({ style: { flex: '1' } }, [Utils.cond(
+                [selectedCallIndex === -1, () => 'N/A (not scattered)'],
+                () => {
```
panentheos, line 179: This logic currently shows the `Select` if there is only one call, even though the calculation of `selectedCallIndex` above defaults to implicitly choosing that. This seems confusing, in that case I think we should either not show the `Select` or remove the implicit selection logic.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,284 @@
...
+        ]),
+        resetLink(() => resetWorkflowSelection(undefined))
+      ]),
+      div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Selected call B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1 1 100px' } }, [
+          div({ style: { marginLeft: '0.5rem', ...Style.codeFont } }, [
+            otherCallFqnDropdownValue,
+            otherIndexDropdownValue !== undefined && breadcrumbHistoryCaret,
+            otherIndexDropdownValue !== undefined && div({ style: { marginLeft: '0.5rem', ...Style.codeFont } }, `index ${otherIndexDropdownValue}`)
```
panentheos: nit: Instead of specifying the same condition twice, this could be:
```suggestion
            otherIndexDropdownValue !== undefined && h(Fragment, [
              breadcrumbHistoryCaret,
              div({ style: { marginLeft: '0.5rem', ...Style.codeFont } }, `index ${otherIndexDropdownValue}`)
            ])
```

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,284 @@
...
+                style: { whiteSpace: 'pre-wrap', border: 'ridge', padding: '0.5rem' },
+                name: false,
+                shouldCollapse: ({ name }) => name === 'callA' || name === 'callB',
+                enableClipboard: false,
+                displayDataTypes: false,
+                displayObjectSize: false,
+                // This line is re-ordering the diff object to put the hashDifferential first:
+                src: { hashDifferential: diff.hashDifferential, ...diff }
+              })
+            ]) : 'Cache diff loading...')
```
panentheos, line 251: TOL: There's a potential subtle edge case here, which may or may not be worth worrying about. Because this operation can take a while and doesn't block interaction, it's possible to kick off a network request and then click 'Reset'. The request would continue in the background and eventually write its result, which could then be inconsistent with the form. There are two approaches for dealing with that:
1. block the whole form with a spinner overlay, so the only option is to wait (or cancel the whole modal)
2. wire up the network request with a 'local' cancellation signal, so we could cancel it on reset of the form

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,284 @@
...
+      return step1()
+    } else if (!otherCallSelected) {
+      return step2()
+    } else {
+      return compareDiffs()
+    }
+  }
+
+  return h(Modal, {
+    title: [
```
panentheos: The array wrapper isn't needed here.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
panentheos: Also, these two elements should be wrapped in an `Fragment` rather than a bare array.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,284 @@
...
+    const otherCallSucceeded = otherWorkflowMetadata && otherCallFqnDropdownValue && selectedCallIndex !== undefined &&
+      _.some({ shardIndex: selectedCallIndex, executionStatus: 'Done' }, otherWorkflowMetadata.calls[otherCallFqnDropdownValue])
+
+    return h(Fragment, [
+      div({ style: { display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Selected workflow B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1' } }, [
+          div({ style: { marginLeft: '0.5rem', ...Style.codeFont } }, otherWorkflowId)
+        ]),
+        resetLink(() => resetWorkflowSelection(undefined))
```
panentheos: We should reset to `''` rather than `undefined`, so the text is consistently a string.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  const compareDiffs = () => {
+    return h(Fragment, [
+      div({ style: { paddingBottom: '0.5rem', display: 'flex', alignItems: 'center', fontSize: 16, fontWeight: 500 } }, [
+        div(['Selected workflow B: ']),
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', flex: '1' } }, [
+          div({ style: { marginLeft: '0.5rem', ...Style.codeFont } }, otherWorkflowId)
+        ]),
+        resetLink(() => resetWorkflowSelection(undefined))
```
panentheos: This should also reset to `''`, same as above.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: Good spot. I think this was a carry-over from when there was also a search icon in the title next to the text

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
cjllanwarne: My first gut reaction is that this sounds like a lot of fiddly coding to cover an unlikely edge case. Would simply not showing the `reset` link until the request completes be an alternative option? I've wired it up like that for now (see lines 213 and 226)

panentheos: Yep, that works.

src/pages/workspaces/workspace/jobHistory/CallCacheWizard.js
```diff
@@ -0,0 +1,293 @@
...
+                })
+              ])
+            ])])
+          ]),
+          otherCallFqnDropdownValue && div({ style: { marginTop: '0.25rem', marginBottom: '0.5rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            h(IdContainer, [id => h(Fragment, [
+              label({ htmlFor: id, style: { paddingRight: '0.5rem' } }, ['Shard Index:']),
+              div({ style: { flex: '1' } }, [Utils.cond(
+                [selectedCallIndex === -1, () => 'N/A (not scattered)'],
+                [_.size(otherWorkflowMetadata.calls[otherCallFqnDropdownValue]) === 1, `${otherWorkflowMetadata.calls[otherCallFqnDropdownValue][0].shardIndex} (exactly one shard in scatter)`],
```
panentheos: The 'value' items in each pair need to be functions. Technically `cond` will accept bare values, but it's deprecated and will be removed soon.
```suggestion
                [_.size(otherWorkflowMetadata.calls[otherCallFqnDropdownValue]) === 1, () => `${otherWorkflowMetadata.calls[otherCallFqnDropdownValue][0].shardIndex} (exactly one shard in scatter)`],
```

panentheos (approved):
See one comment, then looks good. Thanks for all the work that went into this!

Tested locally.

PR #2379, opened 2021-02-05
[IA-2491] Refactor dataproc cost calculation

kyuksel:
- [JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-2491)

- Addresses @breilly2's feedback below on a previous [PR](https://github.com/DataBiosphere/terra-ui/pull/2367)

> TOL: I think it's worth considering extracting a dataprocCost function that takes machineType and numInstances. It would be called 3 times: from runtimeConfigBaseCost with (masterMachineType, 1) and (workerMachineType, numberOfWorkers) and from runtimeConfigCost with (workerMachineType, numberOfPreemptibleWorkers).

- Manually tested by comparing cost display on `local` vs. `dev` while flipping through cluster config parameter value combinations on the create cluster modal

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -25,39 +25,73 @@ export const normalizeRuntimeConfig = ({
   }
 }
 
+export const findMachineType = name => {
+  return _.find({ name }, machineTypes) || { name, cpu: '?', memory: '?', price: NaN, preemptiblePrice: NaN }
+}
+
+export const dataprocCost = (machineType, numInstances) => {
+  const { cpu: cpuPrice } = findMachineType(machineType)
+
+  return cpuPrice * numInstances * dataprocCpuPrice
+}
+
 export const runtimeConfigBaseCost = config => {
   const {
     cloudService, masterMachineType, masterDiskSize, numberOfWorkers, workerMachineType, workerDiskSize, bootDiskSize
   } = normalizeRuntimeConfig(
     config)
-  const { cpu: masterCpu } = findMachineType(masterMachineType)
-  const { cpu: workerCpu } = findMachineType(workerMachineType)
 
   return _.sum([
     (masterDiskSize + numberOfWorkers * workerDiskSize) * storagePrice,
-    cloudService === cloudServices.DATAPROC && (masterCpu + workerCpu * numberOfWorkers) * dataprocCpuPrice,
+    cloudService === cloudServices.DATAPROC && (dataprocCost(masterMachineType, 1) + dataprocCost(workerMachineType, numberOfWorkers)),
     bootDiskSize * storagePrice
   ])
 }
 
-export const findMachineType = name => {
-  return _.find({ name }, machineTypes) || { name, cpu: '?', memory: '?', price: NaN, preemptiblePrice: NaN }
-}
+// export const runtimeConfigBaseCost = config => {
+//   const {
+//     cloudService, masterMachineType, masterDiskSize, numberOfWorkers, workerMachineType, workerDiskSize, bootDiskSize
+//   } = normalizeRuntimeConfig(
+//     config)
+//   const { cpu: masterCpu } = findMachineType(masterMachineType)
+//   const { cpu: workerCpu } = findMachineType(workerMachineType)
+//
+//   return _.sum([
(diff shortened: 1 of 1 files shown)


src/libs/runtime-utils.js
```diff
@@ -25,36 +25,40 @@ export const normalizeRuntimeConfig = ({
   }
 }
 
+export const findMachineType = name => {
+  return _.find({ name }, machineTypes) || { name, cpu: '?', memory: '?', price: NaN, preemptiblePrice: NaN }
+}
```
kyuksel, line 30: No change - just moved up

src/libs/runtime-utils.js
```diff
@@ -25,36 +25,40 @@ export const normalizeRuntimeConfig = ({
...
+
+export const dataprocCost = (machineType, numInstances) => {
+  const { cpu: cpuPrice } = findMachineType(machineType)
+
+  return cpuPrice * numInstances * dataprocCpuPrice
+}
```
kyuksel: the function that was factored out

zarsky-broad: is there a reason to export this?

zarsky-broad (approved):
One comment, looks good otherwise!

src/libs/runtime-utils.js
zarsky-broad: It also occurs to me that we could change the signature to default `numInstances` to 1, so that for the master we wouldn't need the second argument, which might be nicer? It would look like this:
```js
const dataprocCost = (machineType, numInstances = 1) => { ... }
```

kyuksel: > is there a reason to export this?

No - thanks for the catch!

> It also occurs to me that we could change the signature to default numInstances to 1, so that for the master we wouldn't need the second argument, which might be nicer?

I like the conciseness of it but FWIW, our best practice in interactive analysis is to usually avoid default values to make it more difficult for the caller of the function to indirectly pass in a value by mistake that they wouldn't intend to, part of it coming from past pain. If you have the norm to default parameters in the terra-ui world, though, I would also respect that. Let me know either way 😊

gpcarr (approved):
lgtm 👍

src/libs/runtime-utils.js
zarsky-broad: I'm good with it this way. I think you're good to merge. I just kicked off a re-run of the integration tests, but the failure you had is known flake, I believe.

kyuksel: Thank you for kicking off a re-run. I've rerun it a few more times as it kept failing. The failure signature is

```
TimeoutError: waiting for XPath `(//a | //*[@role="button"] | //button)[contains(normalize-space(.),"Create a") or contains(@aria-label,"Create a")]` failed: timeout 30000ms exceeded
    at new WaitTask (/home/circleci/project/integration-tests/node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:411:34)
    at DOMWorld.waitForXPath (/home/circleci/project/integration-tests/node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:336:26)
    at Frame.waitForXPath (/home/circleci/project/integration-tests/node_modules/puppeteer/lib/cjs/puppeteer/common/FrameManager.js:866:51)
    at Page.waitForXPath (/home/circleci/project/integration-tests/node_modules/puppeteer/lib/cjs/puppeteer/common/Page.js:1325:33)
    at click (/home/circleci/project/integration-tests/utils/integration-utils.js:44:22)
    at /home/circleci/project/integration-tests/tests/run-notebook.js:22:9
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at /home/circleci/project/integration-tests/utils/integration-helpers.js:48:5
    at /home/circleci/project/integration-tests/utils/integration-helpers.js:97:5
    at /home/circleci/project/integration-tests/utils/integration-helpers.js:144:3
```

Is this known to happen several (4 so far) times in a row?

zarsky-broad: Yep, and dev sam is going to get taken down as well... @ehanna4 repeated failures are standard, right?

ehanna4 (approved):
Current Integration Test failure is a known transient occurrence - please go ahead and merge

PR #2382, opened 2021-02-11
[IA-2419] Factor in external IP address cost for VMs created by Leonardo

kyuksel:
[JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-2419)

**Background**: 
- VMs that Leonardo creates get assigned ephemeral external IP addresses. Google started charging for these IP addresses. Since these external IP addresses are ephemeral, they impact cost only when underlying VMs are running.
- Galaxy apps run on Kubernetes clusters. These clusters have default nodepools in addition to nodepools created per app. Nodepools incur all of the usual resource costs regardless of app statuses since they run as long as the cluster is alive.

This PR
- factors status into cost calculations
- accounts for the fact that default nodepools are always running
- adds comments to and refactors Galaxy cost calculation in hopes of making it easier to follow 
- adds a [slash](https://github.com/DataBiosphere/terra-ui/blob/58153135eef312a976b64cee7dd047490c98075d/src/components/RuntimeManager.js#L346) that was missing 

Manually tested by comparing cost displays in various runtime and app statuses between `prod` and `local`.

image

### DIFF
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -204,7 +204,7 @@ export const NewGalaxyModal = _.flow(
 
   const renderDefaultCase = () => {
     const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineBases)
-    const cost = getGalaxyCost(app || { status: 'RUNNING', kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
+    const cost = getGalaxyCost(app || { status: 'UNKNOWN', kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
     return h(Fragment, [
       h(TitleBar, {
         title: getEnvMessageBasedOnStatus(true),
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -187,7 +187,7 @@ export const NewGalaxyModal = _.flow(
   }
 
   const getEnvMessageBasedOnStatus = isTitle => {
-    const waitMessage = isTitle ? '' : 'This process will take up to a few minutes'
+    const waitMessage = isTitle ? '' : 'This process will take up to a few minutes.'
 
     return !app ?
       isTitle ? 'Cloud environment' : 'Environment will consist of an application and cloud compute.' :
@@ -198,7 +198,7 @@ export const NewGalaxyModal = _.flow(
         ['PRESTARTING', () => 'Cloud environment is preparing to start.'],
         ['STARTING', () => `Cloud environment is starting. ${waitMessage}`],
         ['RUNNING', () => 'Environment consists of an application and cloud compute.'],
-        ['ERROR', () => `An error has occurred on your Cloud Environment. ${waitMessage}`]
+        ['ERROR', () => `An error has occurred on your cloud environment.`]
       )
   }
 
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -204,7 +204,7 @@ export const NewGalaxyModal = _.flow(
 
   const renderDefaultCase = () => {
     const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineTypes)
-    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
+    const cost = getGalaxyCost(app || { status: 'Unknown', kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
     return h(Fragment, [
       h(TitleBar, {
         title: getEnvMessageBasedOnStatus(true),
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -204,7 +204,7 @@ export const NewGalaxyModal = _.flow(
 
   const renderDefaultCase = () => {
     const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineBases)
-    const cost = getGalaxyCost(app || { status: 'UNKNOWN', kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
+    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
     return h(Fragment, [
       h(TitleBar, {
         title: getEnvMessageBasedOnStatus(true),
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -204,7 +204,7 @@ export const NewGalaxyModal = _.flow(
 
   const renderDefaultCase = () => {
     const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineBases)
-    const cost = getGalaxyCost(app || { status: 'Unknown', kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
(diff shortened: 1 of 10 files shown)


src/data/machines.js
```diff
@@ -40,6 +40,10 @@ export const machineTypes = _.map(({ price, preemptiblePrice, ...details }) => (
 export const monthlyStoragePrice = 0.04 // from https://cloud.google.com/compute/pricing
 export const storagePrice = monthlyStoragePrice / 730 // per GB hour using 730 hours per month from https://cloud.google.com/compute/pricing
 export const dataprocCpuPrice = 0.01 // dataproc costs $0.01 per cpu per hour
+export const ephemeralExternalIpAddressPrice = { // per hour in dollars for Iowa (us-central1) region per https://cloud.google.com/vpc/network-pricing#ipaddress
+  standard: 0.004,
+  preemptible: 0.002
```
breilly2, line 36: I think this is already accounted for in lines 35-36 above.

kyuksel: I had missed this somehow. Thanks for pointing it out! 

If I'm reading the pre-existing code correctly, it's not accounting for VM status. Ephemeral external IP address cost is incurred only when the VM is running. Trying to see if I can find a clever way to adjust that. If you have any thoughts, let me know.

kyuksel: Okay, I pushed a commit that adds IP cost to runtimes only when they're running. But Galaxy cost calculation flow is different and currently it's adding the IP cost whether a Galaxy app is running or stopped. I may need help with figuring out how to make app status available to `getGalaxyCost()` or finding another way to make the distinction.

kyuksel: Paired with Emily and got help with factoring status into cost calculations.

src/libs/runtime-utils.js
```diff
@@ -81,14 +88,27 @@ export const runtimeCost = ({ runtimeConfig, status }) => {
...
+  const defaultNodepoolComputeCost = machineCost('n1-standard-1')
+  const defaultNodepoolIpAddressCost = ephemeralExternalIpAddressCost(1, 0)
+
+  // TODO: Retrieve disk sizes from the app instead of hardcoding them
+  const diskCost = persistentDiskCost({ size: 250 + 10 + 100 + 100, status: 'Running' })
+
+  const staticCost = defaultNodepoolComputeCost + defaultNodepoolIpAddressCost + diskCost
+  const dynamicCost = app.kubernetesRuntimeConfig.numNodes * machineCost(app.kubernetesRuntimeConfig.machineType) + ephemeralExternalIpAddressCost(app.kubernetesRuntimeConfig.numNodes, 0)
+
+  return (['RUNNING', 'PROVISIONING', 'PRECREATING', 'STARTING'].includes(appStatus.toUpperCase())) ? (staticCost + dynamicCost) : staticCost
```
kyuksel: Should the `STOPPING` case also display the whole cost (`staticCost + dynamicCost`) to be conservative? 🤔

src/components/NewGalaxyModal.js
```diff
@@ -203,8 +203,8 @@ export const NewGalaxyModal = _.flow(
   }
 
   const renderDefaultCase = () => {
-    const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineTypes)
-    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
+    const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineBases)
+    const cost = getGalaxyCost(app || { status: 'RUNNING', kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
```
ehanna4: It feels weird to me to have a non-existant app have a pseudo  status of RUNNING. I know it's only for the cost calculation but am wondering if since it's only within this context would it be less "hacky" to use a fake status of "WILL_RUN" or something similar.. or is that more "hacky" ?

kyuksel: Yea, I had the same feeling when I added it. Using a fake status may indeed be less misleading. I think I'll go with that unless I can find a better solution.

kyuksel: Feels like `UNKNOWN` may represent the situation a bit better and would not require introducing a fake status. If any thoughts, let me know.

ehanna4: Actually looking at this more I think the most standard way to deal with this would be to utilize `undefined` which would require a check for and `app` and an `app.status` when setting `appStatus` and then applying the uppercase to that ie:
`const appStatus = app && app.status ? app.status.toUpperCase() : undefined` 

then below you could just use `switch (appStatus)` and you also wouldn't need to add a status to the 'fake' app in runtime-utils

src/libs/runtime-utils.js
```diff
@@ -69,6 +71,11 @@ const generateDiskCostFunction = price => ({ size, status }) => {
 export const persistentDiskCost = generateDiskCostFunction(storagePrice)
 export const persistentDiskCostMonthly = generateDiskCostFunction(monthlyStoragePrice)
 
+const ephemeralExternalIpAddressCost = (numStandardVms, numPreemptibleVms) => {
```
ehanna4: If there are multiple parameters/arguments for a function the typical convention in this repo is to put them into an object i.e. 
`{numStandardVms, numPreemptibleVms}` 
that way when calling the function you can be transparent about which piece you are sending an arg for and without worrying about order i.e 
`ephemeralExternalIpAddressCost({numStandardVms: 1}`

src/components/NewGalaxyModal.js
zarsky-broad: `app?.status?.toUpperCase()` will default to `undefined` if any part of it is, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/libs/runtime-utils.js
kyuksel: 👍🏼  Thanks for the explanation.

kyuksel (commented):
@zarsky-broad @ehanna4 Thank you for the last round of great feedback! Please let me know if I missed something. I couldn't test the changes yet but will once OpenDJ is back up.

src/data/machines.js
```diff
@@ -1,7 +1,4 @@
-import _ from 'lodash/fp'
-
-
-const machineBases = [
+export const machineBases = [
```
breilly2: Nit: rename to `machineTypes` which is the thing that was previously exported. `machineBases` was probably added as an intermediate at the same time as the `_.map` that you're removing.

Making this change will reduce the number of files in this PR and reduce the diffs of the remaining files.

src/libs/runtime-utils.js
breilly2: TOL: I would allow either in this case. The object is a little more verbose, but it is more explicit as Emily pointed out. This is also especially powerful if you want to specify default values for both parameters. If they're not in an object, you'd have to provide the first value if you wanted a non-default 2nd value. If they're in an object then since, as Emily also said, you can provide them in any order when calling, you can omit specifying anything that you want to accept the default value for.

In this case, it looks like `numPreemptibleVms: 0` might be a reasonable default.

breilly2 (approved):
👍  Looks good!

src/libs/runtime-utils.js
kyuksel: This came up during another PR review where my thought was…I like the caller site conciseness of functions with default input parameters but our best practice in the interactive analysis team is to generally avoid default values to make it more difficult for the caller of the function to indirectly pass in a value by mistake that they wouldn't intend to, part of it coming from past pain. Especially since the cost estimate calculations are so convoluted as they are, I think I'd rather err on the side of verbosity.

kyuksel:
Re-tested for GCE, Dataproc and Galaxy compute creation/stopping/deletion  ✅

ehanna4 (approved):
This looks good! Thanks for all your work getting this through and being so thorough on getting cost right!

PR #2383, opened 2021-02-12
Arrange call lists by start time [BW-527]

cjllanwarne:
Tested locally against a workflow with a couple of calls: http://localhost:3000/#workspaces/general-dev-billing-account/cjl_test_workspace/job_history/82e9c326-8fc0-4d3f-aab7-d674bcaafa9e/23f98406-5d7a-4eae-be7f-bf6f6392c4ec

The more natural order makes it easier to decipher the story of what was going on in the workflow that the previous "alphabetical" order.

### DIFF
--- a/src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
+++ b/src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
@@ -118,7 +118,7 @@ const WorkflowDashboard = _.flow(
     }), simplifiedFailures)
   }
 
-  const callNames = calls ? _.sortBy(callName => _.minBy(callObject => callObject.start, calls[callName]).start, _.keys(calls)) : []
+  const callNames = _.sortBy(callName => _.minBy('start', calls[callName]).start, _.keys(calls))
 
   return div({ style: { padding: '1rem 2rem 2rem', flex: 1, display: 'flex', flexDirection: 'column' } }, [
     workflowDetailsBreadcrumbSubtitle(namespace, name, submissionId, workflowId),
--- a/src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
+++ b/src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
@@ -118,7 +118,7 @@ const WorkflowDashboard = _.flow(
     }), simplifiedFailures)
   }
 
-  const callNames = calls ? _.sortBy(callName => _.minBy(_.start, calls[callName]).start, _.keys(calls)) : []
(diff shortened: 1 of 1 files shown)


evanbernstein (approved):

salonishah11:
It looks like the test failure is related to changes made. Holding off the review for now.

cjllanwarne:
@salonishah11 thanks - I think the test failures are common and unrelated to my changes (see the many merges in https://github.com/DataBiosphere/terra-ui/pulls?q=is%3Apr+is%3Aclosed despite failures in this flaky test suite)

salonishah11:
@cjllanwarne Ah I see. Sorry. My bad!

salonishah11 (approved):
LGTM 👍

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -118,7 +118,7 @@ const WorkflowDashboard = _.flow(
     }), simplifiedFailures)
   }
 
-  const callNames = _.sortBy(_.identity, _.keys(calls))
+  const callNames = calls ? _.sortBy(callName => _.minBy(_.start, calls[callName]).start, _.keys(calls)) : []
```
zarsky-broad: I can't find any documentation of `_.start`, did you mean `'start'`?

zarsky-broad: (also, `_.keys` and `_.sortBy` should already be guarded to work normally if calls is undefined)

cjllanwarne: That's because it's the magic "I don't want to specify a variable name" `_` rather than the lodash `_`

cjllanwarne: ... but since it's confusingly mixed in with all the lodashes, I guess I really ought to expand to give it a full function definition with a named input

cjllanwarne: Does the `calls ? ...` at the start cover the guarding you're asking for?

zarsky-broad: Right, I mean they should already be guarded, you shouldn't need the ternary at all. I think you should be able to use the string as a lodash shorthand, then you won't have to spin out a whole function.

cjllanwarne: Oh wow, even nicer!

PR #2384, opened 2021-02-14
Workflow-level timing diagrams [BW-528]

cjllanwarne:
A "10,000 foot view" style timing diagram to show call-by-call timings (ie not x-ray-ing into scatters or backend events). Intended to give the broad picture of what happened in a workflow and let future diagrams (perhaps on call-specific pages) expand the call-specific events. 

The choice to collapse all rows of scatters and attempts into a single bar which does not break out per-call-event is based on:

* The observation that timing diagrams with >10,000 lines are not particularly useful.
* The assumption that seeing the shape of a workflow upfront is valuable (including which calls succeeded and which failed)
* The assumption that getting something high level out to users now - and adding a way to drill into per-call-details later - is also a good way to organize the user experience
* The fact that was much easier to bring the diagrams in than I thought it might be!

##### Successful workflow

image

##### Failed workflow

image

#### Review Recommendations

I would start with `WorkflowTimingDiagram.js`. I think my way of finding the "worst" status for each call and assigning it a color could almost certainly be improved and I'm very open to iterating on that code path. Otherwise, this change seemed fairly (even surprisingly!) straightforward.

#### Good test workflows:

* Workflow with scattered subworkflow:
  * A **DEV** [workflow](http://localhost:3000/#workspaces/general-dev-billing-account/cjl_test_workspace/job_history/5527bea1-3b3a-4331-9810-1728b5c01538/921c9bae-5e8a-40c9-90bc-726b9873ceac) which has a scattered subworkflow which failed.
  * The same **DEV** [workflow](http://localhost:3000/#workspaces/general-dev-billing-account/cjl_test_workspace/job_history/82e9c326-8fc0-4d3f-aab7-d674bcaafa9e/23f98406-5d7a-4eae-be7f-bf6f6392c4ec) which has a scattered subworkflow which succeeded.
  * DIY: Run the same workflow as in the previous few examples yourself (with call caching disabled or some slightly different inputs). See how the Running state is reported.
* A workflow which was aborted:
  * A **DEV** [workflow](http://localhost:3000/#workspaces/general-dev-billing-account/cjl_test_workspace/job_history/5527bea1-3b3a-4331-9810-1728b5c01538/921c9bae-5e8a-40c9-90bc-726b9873ceac) which has a task which was aborted.
  * Compare with the same workflow without the timing diagram. Note that the timing diagram makes is clear that the workflow was aborted (with the orange status bar) much more obviously even that the "Aborted" status text did.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "react-dropzone": "^11.3.2",
     "react-favicon": "^0.0.22",
     "react-focus-lock": "^2.5.0",
+    "react-google-charts": "^3.0.15",
     "react-hyperscript-helpers": "^1.2.0",
     "react-json-view": "^1.21.3",
     "react-modal": "^3.13.1",
--- a/src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
+++ b/src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
@@ -1,5 +1,6 @@
 import _ from 'lodash/fp'
 import { Fragment, useRef, useState } from 'react'
+import { Chart } from 'react-google-charts'
 import { div, h } from 'react-hyperscript-helpers'
 import ReactJson from 'react-json-view'
 import * as breadcrumbs from 'src/components/breadcrumbs'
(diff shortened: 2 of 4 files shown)


kyuksel:
@cjllanwarne Is this PR still applicable?

kyuksel:
Closing due to inactivity. Please feel free to re-open it if applicable.

PR #2386, opened 2021-02-22
Remove wait message from error

ehanna4:
Previously the error'ed galaxy app told users to wait but their only option once it's error'ed is to delete or reach out to support so the wait message was confusing.

Before:
Screen Shot 2021-02-22 at 10 17 29 AM

After:
Screen Shot 2021-02-22 at 10 22 05 AM


xuscrbw (approved):
Looks good. Must have missed this on the PR I made. Thanks for fixing it!

breilly2 (approved):
👍

PR #2387, opened 2021-02-22
Convert support modal to hooks

zarsky-broad:
Verified in-browser; also verified issues were created in zendesk when signed out and signed in.

### DIFF
--- a/src/components/SupportRequest.js
+++ b/src/components/SupportRequest.js
@@ -1,5 +1,5 @@
 import _ from 'lodash/fp'
-import { Component, Fragment } from 'react'
+import { Fragment, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, ButtonSecondary, Clickable, IdContainer, Link, RadioButton, Select, spinnerOverlay } from 'src/components/common'
 import Dropzone from 'src/components/Dropzone'
@@ -7,7 +7,7 @@ import { icon } from 'src/components/icons'
 import { TextArea, TextInput } from 'src/components/input'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
-import { reportError } from 'src/libs/error'
+import { reportError, withErrorReporting } from 'src/libs/error'
 import { FormLabel } from 'src/libs/forms'
 import { notify } from 'src/libs/notifications'
 import { authStore, contactUsActive } from 'src/libs/state'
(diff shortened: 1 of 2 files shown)


src/components/SupportRequest.js
```diff
@@ -24,248 +24,217 @@ const constraints = {
...
-            `%0AError%20reported%20from%20Zendesk%3A%0A%0A${JSON.stringify(error)}`,
-          ...Utils.newTabLinkProps
-        }, 'Click here to email support'), hasAttachment && ' and make sure to add your attachment to the email.']
-      ))
-    } finally {
-      SupportRequest.dismiss()
-    }
-  })
-})
+  return isActive && h(SupportRequest)
```
panentheos, line 236: This is nice, using unmounting to naturally clear the form state.

src/components/SupportRequest.js
```diff
@@ -24,248 +24,217 @@ const constraints = {
...
+        h(TextInput, {
+          id,
+          value: email,
+          placeholder: 'Enter your email address',
+          onChange: setEmail
+        })
+      ])]),
+      h(FormLabel, { required: true }, ['Are you a clinical user?']),
+      h(RadioButton, {
+        text: 'Yes', name: 'is-clinical-user', checked: clinicalUser,
```
panentheos: Since `clinicalUser` starts out undefined, React is warning that the component is switching from uncontrolled to controlled. We either need to keep the explicit comparison or use `!!`

panentheos (approved):
See comment, then looks good. Tested locally.

src/components/SupportRequest.js
breilly2: Is this a change in behavior? I remember being very careful to not throw away state if someone accidentally dismissed the form. I've been bitten by that in other applications which has led to me not submitting the feedback that I carefully put together before the app carelessly erased.

src/components/SupportRequest.js
```diff
@@ -24,248 +24,217 @@ const constraints = {
...
+            border: `1px dashed ${colors.dark(0.7)}`
+          },
+          onClick: openUploader
+        }, [
+          div({ style: { fontSize: 14, lineHeight: '30px', paddingLeft: '1rem', display: 'flex', alignItems: 'center' } }, [
+            'Drag or', h(Link, { style: { margin: '0 0.25rem' } }, ['Click']), 'to attach a file',
+            icon('upload-cloud', { size: 25, style: { opacity: 0.4, marginLeft: '0.5rem' } })
+          ])
+        ]),
+      uploadingFile && spinnerOverlay,
```
Shakespeared: @ehanna4 and I were testing this and we noticed that the spinner is not justified correctly in the modal. I don't think it came from this PR but might be worth fixing because it doesn't look very nice.

src/components/SupportRequest.js
panentheos: It's not a change, we were clearing the state manually on open before. Regarding usability, I remember we did try to be extra careful here, and the form doesn't close on click outside, escape key, or even page transition. The only way to clear it is to click Cancel.

Shakespeared (approved):
LGTU (looks good to us) but see comment

PR #2391, opened 2021-02-24
editable snapshots [AS-578]

zarsky-broad:
Verified locally doing all kinds of crazy fun editing...

### DIFF
--- a/src/components/workspace-utils.js
+++ b/src/components/workspace-utils.js
@@ -4,16 +4,17 @@ import { Fragment, useState } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
 import { AsyncCreatableSelect, ButtonPrimary, ButtonSecondary, IdContainer, Link, Select, spinnerOverlay } from 'src/components/common'
 import { icon } from 'src/components/icons'
-import { TextInput } from 'src/components/input'
+import { ValidatedInput } from 'src/components/input'
 import { MarkdownEditor, MarkdownViewer } from 'src/components/markdown'
 import Modal from 'src/components/Modal'
 import NewWorkspaceModal from 'src/components/NewWorkspaceModal'
 import { Ajax } from 'src/libs/ajax'
-import { withErrorReporting } from 'src/libs/error'
+import { reportError, withErrorReporting } from 'src/libs/error'
 import { FormLabel } from 'src/libs/forms'
 import { workspacesStore } from 'src/libs/state'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
+import validate from 'validate.js'
 
 
 export const useWorkspaces = () => {
@@ -84,9 +85,14 @@ export const SnapshotInfo = ({
 
   // Helpers
   const save = async () => {
-    setSaving(true) // this will be unmounted in the reload, so no need to reset this
-    await Ajax().Workspaces.workspace(namespace, name).snapshot(referenceId).update({ name: newName, description: newDescription })
-    onUpdate(newName)
+    try {
+      setSaving(true) // this will be unmounted in the reload, so no need to reset this
+      await Ajax().Workspaces.workspace(namespace, name).snapshot(referenceId).update({ name: newName, description: newDescription })
+      onUpdate(newName)
+    } catch (e) {
+      setSaving(false)
+      reportError('Error updating snapshot', e)
+    }
   }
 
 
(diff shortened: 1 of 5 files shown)


src/components/workspace-utils.js
```diff
@@ -77,32 +100,104 @@ export const SnapshotInfo = ({ snapshotId, snapshotName }) => {
...
+          }, [icon('edit')])
+        ]),
+        div({ style: { ...Style.elements.sectionHeader, marginBottom: '0.2rem' } }, [
+          'Description:',
+          !editingDescription && h(Link, {
+            style: { marginLeft: '0.5rem' },
+            onClick: () => setNewDescription(description),
+            'aria-label': 'Edit description'
+          }, [icon('edit')])
+        ]),
```
davidangb, line 147: I see the aria-labels for both name and description  ... could these also have hovertext/tooltips that indicate what they will do?

src/components/workspace-utils.js
```diff
@@ -77,32 +100,104 @@ export const SnapshotInfo = ({ snapshotId, snapshotName }) => {
...
+          h(FormLabel, { htmlFor: id }, ['New snapshot name']),
+          h(TextInput, {
+            inputProps: {
+              id,
+              autoFocus: true,
+              placeholder: 'Enter a name',
+              value: newName,
+              onChange: setNewName
+            }
+          })
```
davidangb, line 214: Add an aria-label - and avoid the errors in console!

src/components/workspace-utils.js
```diff
@@ -77,32 +100,104 @@ export const SnapshotInfo = ({ snapshotId, snapshotName }) => {
...
+              renderingConfig: {
+                singleLineBreaks: false,
+                markedOptions: { sanitize: true, sanitizer: _.escape }
+              },
+              status: false
+            },
+            className: 'simplemde-container',
+            value: newDescription,
+            onChange: setNewDescription
+          }),
```
davidangb, line 162: is there any config to suppress the deprecation warnings thrown by marked?

davidangb (changes requested):
See comments inline; I am also unable to edit the name field, as the save button never enables.

src/components/workspace-utils.js
zarsky-broad: Unfortunately no - the newer versions of marked remove sanitizing and have you do it for yourself, which is why we haven't upgraded, as we never put in the time to find a sanitizer we liked...

davidangb: yup, I understand why the warning exists, I was wondering if there's any way to turn off the logging for it so it doesn't show up in users' consoles

src/components/workspace-utils.js
```diff
@@ -77,32 +100,104 @@ export const SnapshotInfo = ({ snapshotId, snapshotName }) => {
...
+      }, [
+        h(IdContainer, [id => h(Fragment, [
+          h(FormLabel, { htmlFor: id }, ['New snapshot name']),
+          h(TextInput, {
+            id,
+            autoFocus: true,
+            placeholder: 'Enter a name',
+            value: newName,
+            onChange: setNewName
+          })
```
davidangb, line 214: 1. can you have this validate the input and disable the save button on an invalid name? Else it's far too easy for an end user to supply an illegal name and get an error on save. ("Name must be 1 to 63 alphanumeric characters or underscores, and cannot start with an underscore.")
2. errors on save appear not to be displayed - for me, they result in an infinite spinner with no feedback to the user.

davidangb (commented):
code changes look good, but I have more functional requests, sorry ...

src/components/workspace-utils.js
zarsky-broad: ah, no, I don't believe so.

davidangb (approved):
:+1: looks good.

I found one corner case in manual testing - if you have two snapshots "foo" and "bar" in your workspace, the UI allows you to attempt to rename "foo" to "bar", which fails in the backend due to uniqueness. This allowed me to see that error-display was working well! It would be a nicety if the UI caught this, but I don't think it should block this PR.

zarsky-broad:
> fails in the backend due to uniqueness

Yeah, that would require making the component aware of all of the other snapshot names, which is certainly doable, but I'm inclined to punt on it for now...

ansingh7115 (approved):
Tested locally - looks good to me!

PR #2392, opened 2021-02-25
[IA-2545] Update Galaxy default conf to 1 n1-highmem-8 node

rtitle:
See rationale in JIRA ticket: https://broadworkbench.atlassian.net/browse/IA-2545

See corresponding Leo PR: https://github.com/DataBiosphere/leonardo/pull/1893

I tested locally. With the change:
image

Current dev:
image

Nice cost savings for the user too.

### DIFF
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -203,8 +203,8 @@ export const NewGalaxyModal = _.flow(
   }
 
   const renderDefaultCase = () => {
-    const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineTypes)
-    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
+    const { cpu, memory } = _.find({ name: 'n1-highmem-8' }, machineTypes)
+    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-highmem-8', numNodes: 1 } })
     return h(Fragment, [
       h(TitleBar, {
         title: getEnvMessageBasedOnStatus(true),
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -203,8 +203,9 @@ export const NewGalaxyModal = _.flow(
   }
 
   const renderDefaultCase = () => {
-    const { cpu, memory } = _.find({ name: 'n1-highmem-8' }, machineTypes)
-    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-highmem-8', numNodes: 1 } })
+    const defaultMachineType = 'n1-highmem-8'
+    const { cpu, memory } = _.find({ name: defaultMachineType }, machineTypes)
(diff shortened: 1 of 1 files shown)


src/components/NewGalaxyModal.js
```diff
@@ -203,8 +203,8 @@ export const NewGalaxyModal = _.flow(
   }
 
   const renderDefaultCase = () => {
-    const { cpu, memory } = _.find({ name: 'n1-standard-8' }, machineTypes)
```
Qi77Qi, line 206: once we allow users to specify machine type, this needs to be updated as well right...not necessarily in this PR, but might deserve a ticket

Qi77Qi (approved):

src/components/NewGalaxyModal.js
rtitle: Yeah - this currently just displays the default, but the UI will eventually need to allow users to choose their machine type and number of nodes.

Qi77Qi: I see..this is just the defaults

rtitle:
Since Terra UI deploys daily, this should be coordinated with the Leo deploy of https://github.com/DataBiosphere/leonardo/pull/1893

src/components/NewGalaxyModal.js
```diff
@@ -203,8 +203,8 @@ export const NewGalaxyModal = _.flow(
...
-    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-standard-8', numNodes: 2 } })
+    const { cpu, memory } = _.find({ name: 'n1-highmem-8' }, machineTypes)
+    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: 'n1-highmem-8', numNodes: 1 } })
```
kyuksel: Might be good to factor out the default machine type so there's one place to change going forward:

```javascript
const defaultMachineType = 'n1-highmem-8'
```

kyuksel:
@rtitle While reviewing your PR I noticed that we've also not been factoring `numNodes` into the `Cloud Compute size` display:

```javascript
span({ style: { fontWeight: 600 } }, [`${cpu} CPUS, ${memory} GB of memory, 250 GB disk space`])
```

The information displayed on the page will be correct after your PR is merged since the new default config will include a single node instead of two but it would be good to make that dynamic in case the default config has two nodes again in the future. Not sure if this should loop in the UX team as well.

What do you think?

rtitle:
@kyuksel yeah me and @almahmoud noticed that too: the UI doesn't actually display the number of nodes in the `NewGalaxyModal`. However the number of nodes _is_ taken into account in the cost calculation (https://github.com/DataBiosphere/terra-ui/blob/e0e46a981e6b9355456ab036a73197b84cec5202/src/libs/runtime-utils.js#L109).

Since this code be changing anyway (I imagine) when we add compute configurability, I'm inclined to leave this PR as is to just make the default case correct for now.

kyuksel:
> Since this code be changing anyway (I imagine) when we add compute configurability

Yea I can look into this as part of the compute configurability ticket.

kyuksel (approved):

PR #2396, opened 2021-03-01
[SATURN-1884] Uploader page

andrewmarcus:
This PR implements a new feature allowing state labs to easily upload their entity data and create new tables in their workspace. Much of this functionality should eventually be ported back into the Workspaces view.

Amongst the recent updates:

* The abort upload button now works. I see that it can sometimes get into a state where the upload has stopped but the modal isn't dismissable, which I think happens because when the actual fetch is aborted, an unresolved promise is returned. But I didn't want to change the plumbing of the deep-down AJAX code. In the general case the button is working.
* I've switched out the UI to more of an accordion style rather than the nav buttons on the left side. After putting it in place, this interface seems MUCH more fluid and intuitive. See what you think.

I have not yet managed to resolve the issue with the `AutoSizer`. Please have your UX team take a look at it and tell me what I did wrong.

### DIFF
--- a/src/components/Dropzone.js
+++ b/src/components/Dropzone.js
@@ -1,14 +1,13 @@
 import { useState } from 'react'
 import { useDropzone } from 'react-dropzone'
 import { div, input } from 'react-hyperscript-helpers'
-import colors from 'src/libs/colors'
 
 
-const Dropzone = ({ disabled = false, onDragOver, onDrop, onDragLeave, onDropAccepted, onDropRejected,
-                    style = {},
-                    activeStyle = {},
-                    rejectedStyle = {},
-                    children, ...props }) => {
+const Dropzone = (
+  {
+    disabled = false, onDragOver, onDrop, onDragLeave, onDropAccepted, onDropRejected,
+    style = {}, activeStyle = {}, rejectedStyle = {}, children, ...props
+  }) => {
   // dropzone's built-in dragging status doesn't seem to work if there's anything rendered over the root div
   const [dragging, setDragging] = useState(false)
   const [rejected, setRejected] = useState(false)
--- a/src/components/Dropzone.js
+++ b/src/components/Dropzone.js
@@ -3,16 +3,11 @@ import { useDropzone } from 'react-dropzone'
 import { div, input } from 'react-hyperscript-helpers'
 
 
-const Dropzone = (
-  {
-    disabled = false, onDragOver, onDrop, onDragLeave, onDropAccepted, onDropRejected,
-    style = {}, activeStyle = {}, rejectedStyle = {}, children, ...props
-  }) => {
+const Dropzone = ({ disabled = false, onDragOver, onDrop, onDragLeave, style = {}, activeStyle = {}, children, ...props }) => {
   // dropzone's built-in dragging status doesn't seem to work if there's anything rendered over the root div
   const [dragging, setDragging] = useState(false)
-  const [rejected, setRejected] = useState(false)
 
-  const { getRootProps, getInputProps, open: openUploader, fileRejections, ...dropProps } = useDropzone({
+  const { getRootProps, getInputProps, open: openUploader, ...dropProps } = useDropzone({
     noClick: true,
     // Due to some sloppy internal state management, the keyboard handlers cause
     // re-renders on every focus/blur, which causes performance problems on some pages.
@@ -30,18 +25,10 @@ const Dropzone = (
       setDragging(false)
       onDragLeave && onDragLeave(...args)
     },
-    onDropAccepted: (...args) => {
-      setRejected(false)
-      onDropAccepted && onDropAccepted(...args)
-    },
-    onDropRejected: (...args) => {
-      setRejected(true)
-      onDropRejected && onDropRejected(...args)
-    },
     ...props
   })
 
-  return div(getRootProps({ tabIndex: -1, style: { ...style, ...(dragging ? activeStyle : {}), ...(rejected ? rejectedStyle : {}) } }), [
+  return div(getRootProps({ tabIndex: -1, style: { ...style, ...(dragging ? activeStyle : {}) } }), [
     // the input is disabled by react-dropzone when noClick is true, which makes it drag only, so we couldn't even open it manually
     input({ ...getInputProps({ disabled, 'aria-hidden': true }) }),
     children({ dragging, openUploader, ...dropProps })
--- a/src/components/Dropzone.js
+++ b/src/components/Dropzone.js
@@ -1,13 +1,19 @@
 import { useState } from 'react'
 import { useDropzone } from 'react-dropzone'
 import { div, input } from 'react-hyperscript-helpers'
+import colors from 'src/libs/colors'
 
 
-const Dropzone = ({ disabled = false, onDragOver, onDrop, onDragLeave, style = {}, activeStyle = {}, children, ...props }) => {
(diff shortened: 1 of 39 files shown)


andrewmarcus:
I just got some accessibility improvements in to the Uploader feature, some of which improve the site generally.

* Allow spacebar to select buttons (generally)
* Added aria-modal='true' to modals (generally)
* Show hover states on focus (generally)
* Move focus to section header when it opens, to help the user track their place (Uploads)
* Read out the current filter state to screen readers (Uploads > workspace)

I also hid the button for creating new folders for new

src/components/Dropzone.js
```diff
@@ -25,10 +30,18 @@ const Dropzone = ({ disabled = false, onDragOver, onDrop, onDragLeave, style = {
...
+    },
+    onDropRejected: (...args) => {
+      setRejected(true)
+      onDropRejected && onDropRejected(...args)
+    },
     ...props
   })
 
-  return div(getRootProps({ tabIndex: -1, style: { ...style, ...(dragging ? activeStyle : {}) } }), [
+  return div(getRootProps({ tabIndex: -1, style: { ...style, ...(dragging ? activeStyle : {}), ...(rejected ? rejectedStyle : {}) } }), [
```
dmohs: Not seeing usage of `rejectedStyle` or `fileRejections` anywhere.

src/components/Interactive.js
```diff
@@ -46,7 +46,7 @@ const Interactive = Utils.forwardRefWithName('Interactive', ({
     className: `hover-style ${className}`,
     style: { ...style, ...cssVariables, cursor: computedCursor, outline },
     onKeyDown: onKeyDown || (e => {
-      if (e.key === 'Enter') {
+      if (e.key === 'Enter' || e.key === ' ') {
```
dmohs: I think this makes sense, but should be a separate PR unless essential.

src/components/Modal.js
```diff
@@ -53,7 +53,7 @@ const Modal = ({ onDismiss, title, titleExtras, children, width = 450, showCance
       nodeToFocus.focus()
     },
     style: { overlay: styles.overlay, content: { ...styles.modal, width } },
-    aria: { labelledby: titleId },
+    aria: { labelledby: titleId, modal: true },
```
dmohs: This is a great change, but also a separate PR unless essential to this functionality.

src/components/Modal.js
```diff
@@ -74,9 +74,9 @@ const Modal = ({ onDismiss, title, titleExtras, children, width = 450, showCance
         }, [cancelText]) :
         null,
       Utils.cond(
-        [okButton === undefined, () => h(ButtonPrimary, { onClick: onDismiss }, 'OK')],
-        [_.isString(okButton), () => h(ButtonPrimary, { onClick: onDismiss }, okButton)],
-        [_.isFunction(okButton), () => h(ButtonPrimary, { onClick: okButton }, 'OK')],
+        [okButton === undefined, () => h(ButtonPrimary, { onClick: onDismiss, danger }, 'OK')],
+        [_.isString(okButton), () => h(ButtonPrimary, { onClick: onDismiss, danger }, okButton)],
+        [_.isFunction(okButton), () => h(ButtonPrimary, { onClick: okButton, danger }, 'OK')],
```
dmohs, line 79: I like this, but consider whether it can be omitted for the sake of preserving existing files as much as possible (when introducing new functionality). This would be a great addition as a follow-up PR.

src/components/table.js
```diff
@@ -394,25 +394,35 @@ GridTable.propTypes = {
...
         style: { display: 'flex' }, className: 'table-row',
-        hover: { backgroundColor: colors.light(0.4) }
+        hover: { backgroundColor: colors.light(0.4) },
+        role: 'row'
       }, [
         _.map(({ key, size }) => {
           return div({
             key,
             className: 'table-cell',
+            role: 'gridcell',
```
dmohs: Love these role additions, but as a separate PR so we can contain the new functionality as much as possible.

src/style.css
```diff
@@ -7,7 +7,7 @@
 /* This adds some low-level css resets based on the browserslist in package.json, see https://create-react-app.dev/docs/adding-css-reset */
 @import-normalize;
 
-.hover-style:hover {
+.hover-style:hover, .hover-style:focus {
```
dmohs: Separate PR? I think we want these kinds of changes unconditionally, whereas the upload feature has some organizational controversy around it.

dmohs (commented):
This first round is focused on keeping the changed set contained as much as possible. In general, I'm impressed with how well you follow the existing style.

src/components/Dropzone.js
andrewmarcus: Thanks for the feedback. I'd be happy to pull out the more general accessibility changes into separate PRs. There are a couple of things in here, such as this change to Dropzone that I was going to use but then ended up doing something else. They might still be useful in the future but I'll pull them out for this pass.

src/components/Modal.js
andrewmarcus: This particular change allowed me to make the Abort Upload button red instead of blue, which more accurately captures its purpose (don't press that button unless you really mean it). However, I'd be happy to pull that out into a different PR for cleanliness.

src/components/Modal.js
andrewmarcus: Sure. I'll pull this and the other changes out into a separate PR handling the first pass at accessibility improvements. These particular roles and keyboard actions impact the accessiblity of the new uploads page, which is why I included them, but there are no particular dependencies that these be in there for the first pass.

src/style.css
andrewmarcus: Thanks for that, good to know. Happy to split it up

src/components/Modal.js
dmohs: Tough call, probably should stay in this PR.

src/components/FloatingActionButton.js
```diff
@@ -23,7 +23,8 @@ const FloatingActionButton = ({ label, iconShape, onClick, bottom = 30, right =
       onClick: () => {
         onClick()
         setHover(false)
-      }
+      },
+      ...props
```
dmohs: This isn't right, I think. The ref you're passing here is creating a warning. I believe you should be able to find the right pattern elsewhere in the codebase.

andrewmarcus: In the FloatingActionButton? I'm not sure I understand which ref you're referring to. The only thing I changed in this file was to pass through additional optional properties to the button. That said, I don't recall now whether I actually ended up needing the extra props here. I'll double-check and remove them if I can.

dmohs: It's used to pass a ref on line 683 of Upload.js. I'd say leave that out until we can get it right. We can do a usability pass later.

dmohs: (the ref is used to focus a button)

andrewmarcus:
I got the table AutoSizer working. The problem turned out to be that I just hadn't been passing down the `height: 100%` appropriately through all the flexbox layers.

dmohs:
This is excellent. Changes are well-contained to new files, additions to existing files, and very few changes to existing code, where the existing style is respected. I judge this is safe to merge and we can follow-up with improvements in later PRs. I will work on getting a second approval.

dmohs:
Tested manually by running through the demonstrated workflow. Only minor cosmetic and usability issues found, nothing that should block release.

dmohs (approved):

abaumann (approved):
Reviewed the functionality and intended flow by running this branch locally and it all looks correct to me according to the requirements

PR #2398, opened 2021-03-03
[IA-2566] Fix Galaxy 401

gpcarr:
We recently added a referrer check to all leo proxy routes (this feature is currently disabled in all environments until https://broadworkbench.atlassian.net/browse/IA-2566 is complete which is dependent on the AoU team). here is the PR for the referrer check: https://github.com/DataBiosphere/leonardo/pull/1854. we essentially check that every request to the leonardo proxy has a `Referer` header which is in our list of valid referrers to help prevent CSRF attacks

While doing some testing in dev with the referrer check enabled - realized that since Galaxy opens in a new tab (unlike jupyter and rstudio) that we were throwing a 401 since previously we were stripping the referrer in terra-ui when a new tab is opened:
Screen Shot 2021-03-03 at 3 12 18 PM

This PR allows the referrer to be present in a new tab only when a user launches Galaxy. I tested this in my Fiab by pointing my local terra UI at my fiab and manually enabling the referrer check and verifying that Galaxy no longer fails with a 401:
Screen Shot 2021-03-03 at 3 04 56 PM

### DIFF
--- a/src/components/runtime-common.js
+++ b/src/components/runtime-common.js
@@ -136,7 +136,7 @@ export const GalaxyLaunchButton = ({ app, onClick, ...props }) => {
       onClick()
       Ajax().Metrics.captureEvent(Events.applicationLaunch, { app: 'Galaxy' })
     },
-    ...Utils.newTabLinkPropsGalaxy,
+    ...Utils.newTabLinkPropsWithReferrer, // Galaxy needs the referrer to be present so we can validate it, otherwise we fail with 401
     ...props
   }, ['Launch Galaxy'])
 }
--- a/src/components/runtime-common.js
+++ b/src/components/runtime-common.js
@@ -136,7 +136,7 @@ export const GalaxyLaunchButton = ({ app, onClick, ...props }) => {
       onClick()
       Ajax().Metrics.captureEvent(Events.applicationLaunch, { app: 'Galaxy' })
     },
-    ...Utils.newTabLinkProps,
+    ...Utils.newTabLinkPropsGalaxy,
     ...props
   }, ['Launch Galaxy'])
 }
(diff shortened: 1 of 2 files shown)


jdcanas:
Tested locally, looks good. For context, to anyone reading this who is curious, there is a lot of hidden stuff going on here. First of all, part of the snippet omitted by github specifies these attributes are on a button that has an `href`  attribute  (which indicates the button will open a link). Second of all, there are various attributes that can be specified within a `rel` attribute for elements that have a link. We are modifying one such attribute, which is detailed here: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noreferrer 

There's a lot of magic with links. For example, the `target` attribute specifies where the link will open, in this case a new tab as we use `_blank`. https://www.w3schools.com/tags/att_a_target.asp

jdcanas (approved):

ehanna4:
Since `newTabLinkPropsGalaxy` is only used in the galaxyLaunchButton it seems like adding it to the general utils file would make more sense if it was named to something not Galaxy specific ie `newTabLinkWithReferer` with the assumption we might want to include a referer for other new tabs in the future, otherwise another option would be to use the existing `newTabLinkProps` but overwrite the `rel` field when using it in the galaxyLaunchButton and then you wouldn't have to add a new thing to utils at all

i.e
`    },
    ...Utils.newTabLinkPropsGalaxy,
    rel: 'noopener',`

gpcarr:
> Since `newTabLinkPropsGalaxy` is only used in the galaxyLaunchButton it seems like adding it to the general utils file would make more sense if it was named to something not Galaxy specific ie `newTabLinkWithReferer` with the assumption we might want to include a referer for other new tabs in the future, otherwise another option would be to use the existing `newTabLinkProps` but overwrite the `rel` field when using it in the galaxyLaunchButton and then you wouldn't have to add a new thing to utils at all
> 
> i.e
> ` }, ...Utils.newTabLinkPropsGalaxy, rel: 'noopener',`

I like the idea of renaming the method to something more general so that if anyone wants to do the same in the future they can just use this method :)

ehanna4 (approved):
LGTM

PR #2399, opened 2021-03-05
Add button to open workspace in Seqr [AS-590]

Shakespeared:
Add link out to Seqr, tested in UI although not entirely by me. 

To test the actual Seqr side you have to have a seqr account, but I had someone who has an account test it on the PR site and they said it opened successfully. To see the link go to the 'files' section of the data tab.

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -238,9 +238,8 @@ const BucketContent = _.flow(
         ])
       ]),
       div({ style: { justifyContent: 'flex-end' } }, [
-        h(Link, {
-          onClick: () => ''
-        }, ['hello']
+        h(Link, { href: `https://seqr.broadinstitute.org/workspace/${namespace}/${workspace}` },
+          ['Analyze in Seqr ', icon('pop-out', { size: 14 })]
         )
       ])
 
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -237,11 +237,9 @@ const BucketContent = _.flow(
           }, _.range(0, prefixParts.length))
         ])
       ]),
-      [
-        h(Link, { href: `https://seqr.broadinstitute.org/workspace/${namespace}/${workspace.workspace.name}` },
-          ['Analyze in Seqr ', icon('pop-out', { size: 14 })]
-        )
-      ]
+      h(Link, { href: `https://seqr.broadinstitute.org/workspace/${namespace}/${workspace.workspace.name}` },
+        ['Analyze in Seqr ', icon('pop-out', { size: 14 })]
+      )
     ]),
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -223,18 +223,27 @@ const BucketContent = _.flow(
...
+          }, _.range(0, prefixParts.length))
         ])
-      }, [
-        { label: 'Files', target: '' },
-        ..._.map(n => {
-          return { label: prefixParts[n], target: _.map(s => `${s}/`, _.take(n + 1, prefixParts)).join('') }
-        }, _.range(0, prefixParts.length))
+      ]),
+
+      div({ style: { justifyContent: 'flex-end' } }, [
```
zarsky-broad: You should be able to remove `justifyContent`, since this doesn't have `display: 'flex'`.

src/pages/workspaces/workspace/Data.js
```diff
@@ -223,18 +223,27 @@ const BucketContent = _.flow(
...
+        h(Link, { href: `https://seqr.broadinstitute.org/workspace/${namespace}/${workspace.workspace.name}` },
+          ['Analyze in Seqr ', icon('pop-out', { size: 14 })]
+        )
       ])
+
```
```suggestion
```

src/pages/workspaces/workspace/Data.js
zarsky-broad: (In fact, you should be able to just remove this wrapper div altogether...)

src/pages/workspaces/workspace/Data.js
```diff
@@ -223,18 +223,27 @@ const BucketContent = _.flow(
...
-        ..._.map(n => {
-          return { label: prefixParts[n], target: _.map(s => `${s}/`, _.take(n + 1, prefixParts)).join('') }
-        }, _.range(0, prefixParts.length))
+      ]),
+
+      div({ style: { justifyContent: 'flex-end' } }, [
+        h(Link, { href: `https://seqr.broadinstitute.org/workspace/${namespace}/${workspace.workspace.name}` },
+          ['Analyze in Seqr ', icon('pop-out', { size: 14 })]
+        )
       ])
```
davidangb: could it cause any problems that all Terra envs point to the same production Seqr instance? For instance, if we have a workspace with the same namespace+name in both alpha and prod, would that cause any problems in Seqr?

Shakespeared: Hmm that's a good question. I don't know much about how Seqr works so I'll ask Janki about that.

davidangb (commented):
:+1: for code syntax.

A few more functional questions:
* if the user has zero files in the workspace, should the "Analyze in Seqr" link still show up? Is it useful to link to Seqr with no files?
* Could you (or your contact at seqr) test this on a workspace that includes spaces in the name? If there are problems with spaces I realize it could be an issue on seqr's side with their redirects, but we'd want to identify it prior to launch.

davidangb (approved):

zarsky-broad (approved):
Code looks good.

PR #2400, opened 2021-03-06
[CA-1304] Site-wide accessibility improvements

andrewmarcus:
This PR supports the new Uploader page as well as improvements that benefit other pages such as the file browser.

* Buttons now respond to the spacebar key in addition to the Enter key
* Modal dialogs now have the `aria-modal="true"` attribute to indicate the rest of the content is non-interactable
* `SimpleTable` now has correct grid roles
* Hover effects (button color change, icon visibility) are now active on focus as well as on hover. In particular, the Delete icon in the File Browser is now visible when it is focused


src/components/Interactive.js
```diff
@@ -46,7 +46,7 @@ const Interactive = Utils.forwardRefWithName('Interactive', ({
     className: `hover-style ${className}`,
     style: { ...style, ...cssVariables, cursor: computedCursor, outline },
     onKeyDown: onKeyDown || (e => {
-      if (e.key === 'Enter') {
+      if (e.key === 'Enter' || e.key === ' ') {
```
ehanna4, line 49: This is a nice change. Looks like from what I could find on standard keyboard shortcuts this is better/more standard. Also previously space did a weird half scroll behavior in the workspaces list and this now gives it a nice defined behavior.

andrewmarcus: Yep! In some widgets space and Enter are supposed to work differently (select boxes is an example) but for buttons they're supposed to do the same thing.

This change only applies when the user's focus in on an `Interactive` widget such as a `PrimaryButton`. So technically the space bar will still cause the page to scroll down a bit if the focus is on something else. But that is the standard expected browser behavior even if it seems a bit weird. :)

src/components/Modal.js
```diff
@@ -53,7 +53,7 @@ const Modal = ({ onDismiss, title, titleExtras, children, width = 450, showCance
       nodeToFocus.focus()
     },
     style: { overlay: styles.overlay, content: { ...styles.modal, width } },
-    aria: { labelledby: titleId },
+    aria: { labelledby: titleId, modal: true },
```
andrewmarcus, line 56: Adds the `aria-modal` attribute, indicating to screen readers that they should stay within the dialog and not read the rest of the page while the modal is active. This is still emergent, but has reasonable screen reader support at this point. https://a11ysupport.io/tech/aria/aria-modal_attribute

src/components/table.js
```diff
@@ -394,25 +394,35 @@ GridTable.propTypes = {
...
 }
 
-export const SimpleTable = ({ columns, rows }) => {
+export const SimpleTable = ({ columns, rows, ...props }) => {
   const cellStyles = { paddingTop: '0.25rem', paddingBottom: '0.25rem' }
   return h(Fragment, [
-    div({ style: { display: 'flex' } }, [
+    div({
+      style: { display: 'flex' },
+      role: 'grid',
```
andrewmarcus: Adds `grid`, `row`, `columnheader` and `gridcell` roles to indicate to screen readers that this is a table even though it's marked up with `<div>` tags.

`grid` elements are also supposed to support navigating through each cell using the arrow keys, but that's for later...

src/style.css
```diff
@@ -7,7 +7,7 @@
 /* This adds some low-level css resets based on the browserslist in package.json, see https://create-react-app.dev/docs/adding-css-reset */
 @import-normalize;
 
-.hover-style:hover {
+.hover-style:hover, .hover-style:focus {
```
andrewmarcus, line 10: These two changes ensure that the hover styles work when the user moves their focus to the element using the tab key, not just the mouse. Since the focus outlines are not always easy to see in Terra UI, it's rather jarring if the hover states don't appear.

src/style.css
```diff
@@ -78,7 +78,7 @@ a {
   overflow-y: scroll;
 }
 
-.table-row:not(:hover) .hover-only, .table-cell:not(:hover) .cell-hover-only {
+.table-row:not(:hover, :focus-within) .hover-only, .table-cell:not(:hover, :focus-within) .cell-hover-only {
```
andrewmarcus, line 81: This particular change allows the Delete button hidden in file browser tables to show up when the user tabs into it. Otherwise the focus ring and tooltip appear around an empty space.

ehanna4: Nice!

ehanna4 (approved):
LGTM - tested in browser with ANDI tool and the chrome extension screenreader. 

Really nice improvements! Thanks so much for adding this PR, these are a great fixes to make the site more accessible!

s-rubenstein (approved):
This looks great! Nice changes.

andrewmarcus:
Is this ready to merge? This PR added some ARIA roles to the SimpleTable, but I've got much more substantial changes to the rest of the tables in another PR I have just about ready. If we can get this one merged, I can resolve the merge conflicts with that one. Or I could pull the changes to `table.js` out of here...

andrewmarcus:
Update: I removed the `SimpleTable` table roles from this PR to avoid merge conflicts with https://github.com/DataBiosphere/terra-ui/pull/2448 which adds them to the other tables too, and uses `role="table"` instead of `role="grid"` for now

s-rubenstein:
This is good to merge since you have two thumbs!

PR #2401, opened 2021-03-08
CDC Uploader Page

dmohs:
Original description from https://github.com/DataBiosphere/terra-ui/pull/2396:
> This PR implements a new feature allowing state labs to easily upload their entity data and create new tables in their workspace. Much of this functionality should eventually be ported back into the Workspaces view.
> 
> Amongst the recent updates:
> 
> The abort upload button now works. I see that it can sometimes get into a state where the upload has stopped but the modal isn't dismissable, which I think happens because when the actual fetch is aborted, an unresolved promise is returned. But I didn't want to change the plumbing of the deep-down AJAX code. In the general case the button is working.
> I've switched out the UI to more of an accordion style rather than the nav buttons on the left side. After putting it in place, this interface seems MUCH more fluid and intuitive. See what you think.
> I have not yet managed to resolve the issue with the AutoSizer. Please have your UX team take a look at it and tell me what I did wrong.

Tested manually by running through the demonstrated workflow. Only minor cosmetic and usability issues found, nothing that should block release.

Moved to this branch to attempt fix for CI tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "marked": "^0.7.0",
     "outdated-browser-rework": "^2.10.0",
     "path-to-regexp": "^5.0.0",
-    "prismjs": "^1.21.0",
+    "prismjs": "^1.23.0",
     "prop-types": "^15.7.2",
     "qs": "^6.9.4",
     "react": "~16.13.0",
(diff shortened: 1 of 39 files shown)


dmohs:
CI tests passed on original PR, so merged that one and closing this one.

PR #2402, opened 2021-03-09
[SATURN-1885] Bugfix: resolve issue when creating new workspaces from the uploader

andrewmarcus:
This PR resolves issues which occur after users of the Uploader tool create a new workspace. It appears that I wasn't handling the state properly in the success callback, and was using the workspace's name rather that its ID.

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -902,9 +902,9 @@ const Buckets = signal => ({
     return res.json()
   },
 
-  listAll: async (namespace, bucket, pageToken = null) => {
+  listAll: async (namespace, bucket, prefix = null, pageToken = null) => {
     const res = await fetchBuckets(
-      `storage/v1/b/${bucket}/o?${qs.stringify({ pageToken })}`,
+      `storage/v1/b/${bucket}/o?${qs.stringify({ prefix, pageToken })}`,
       _.merge(authOpts(await saToken(namespace)), { signal })
     )
     const body = await res.json()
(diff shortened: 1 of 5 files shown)


andrewmarcus:
I added another bugfix to resolve what I believe may have been a race condition which caused the number of files in the collection prefix to not always return the correct count, thereby leaving the "Next" button disabled when it shouldn't have been.

Now, it will only re-count the files in the prefix when a file has been deleted or when an upload has completed, not while an upload is going on. It also lets Google handle all the prefix filtering rather than doing it in the browser, which should be a significant performance improvement for large workspaces with lots of folders.

src/libs/ajax.js
```diff
@@ -883,17 +883,17 @@ const Buckets = signal => ({
     return res.json()
   },
 
-  listAll: async (namespace, bucket, pageToken = null) => {
+  listAll: async (namespace, bucket, prefix = null, pageToken = null) => {
```
andrewmarcus, line 905: Since I introduced this method, the uploader should be the only code using it so this is safe.

andrewmarcus:
Just added validation to the modal for renaming the metadata table to ensure the name is valid and does not contain characters such as spaces. It is following the same rules that the server says it's following in its error message.

dmohs (approved):
Looks good. All changes contained to new uploader code, so minimal risk.

abaumann (approved):
Checked out this code and ran this locally to confirm these bugs were fixed

dmohs:
Recent runs of tests have passed, so re-running these to see if we can get a green.

dmohs:
All but one notebook test failed when I pushed the branch myself. Confident the failure is unrelated, so merging.

PR #2403, opened 2021-03-09
QA-1383: Adding logging when adding user to billing account

xuscrbw:
<!--
Hello, friend!
Thanks!
--!>
This PR doesn't fix the underlying issue for the run-notebooks test, but it does add some additional console logging to help us debug in the future if we run into this issue again. Looking at the screenshot accompanying this error, it looks like the user doesn't have "Edit" permissions on the notebook, so we wanted to track to make sure the user gets added to the billing project correctly at the time of the test.

If this issue still persists even if the user is added properly, I think the next step would be to look into Rawls/Sam to see if there's a race condition of some sort. 

Tested this locally with @ehanna4 to make sure that the information was printed correctly in the console.

### DIFF
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -85,7 +85,13 @@ const addUserToBilling = _.flow(withSignedInPage, withUserToken)(async ({ page, 
     return window.Ajax().Billing.project(billingProject).listUsers()
   }, billingProject)
 
-  console.info(`users in billing project: ${userList}`)
+  function getval(email) {
+
+    let obj = userList.filter(item => item.email === email)
+    return obj[0].role
+  }
+
+  console.info(`test user was added to the billing project with the role: ${getval(email)}`)
 })
 
 const removeUserFromBilling = _.flow(withSignedInPage, withUserToken)(async ({ page, billingProject, email }) => {
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -85,13 +85,12 @@ const addUserToBilling = _.flow(withSignedInPage, withUserToken)(async ({ page, 
     return window.Ajax().Billing.project(billingProject).listUsers()
   }, billingProject)
 
-  function getval(email) {
-
-    let obj = userList.filter(item => item.email === email)
-    return obj[0].role
+  const getVal = email => {
+    const userKey = _.find({email}, userList)
+    return userKey.map(user=> user.role)
   }
 
-  console.info(`test user was added to the billing project with the role: ${getval(email)}`)
+  console.info(`test user was added to the billing project with the role: ${getVal(email)}`)
 })
 
 const removeUserFromBilling = _.flow(withSignedInPage, withUserToken)(async ({ page, billingProject, email }) => {
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -80,6 +80,12 @@ const addUserToBilling = _.flow(withSignedInPage, withUserToken)(async ({ page, 
   }, email, billingProject)
 
   console.info(`added user to: ${billingProject}`)
+
+  const userList = await page.evaluate((billingProject) => {
+    return window.Ajax().Billing.project(billingProject).listUsers()
+  }, billingProject)
+
+  console.info(`users in billing project: ${userList}`)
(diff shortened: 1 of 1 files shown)


integration-tests/utils/integration-helpers.js
```diff
@@ -80,6 +80,18 @@ const addUserToBilling = _.flow(withSignedInPage, withUserToken)(async ({ page,
...
 
   console.info(`added user to: ${billingProject}`)
+
+  const userList = await page.evaluate((billingProject) => {
+    return window.Ajax().Billing.project(billingProject).listUsers()
+  }, billingProject)
+
+  function getval(email) {
+
+    let obj = userList.filter(item => item.email === email)
```
jdcanas: Two things,
1. we generally avoid `let` in favor of `const` to avoid mutability. 
2. we typically prefer the lodash variant of filter (i.e. `_.filter(({email}) => email == param, items)`)

integration-tests/utils/integration-helpers.js
```diff
@@ -80,6 +80,18 @@ const addUserToBilling = _.flow(withSignedInPage, withUserToken)(async ({ page,
   }, email, billingProject)
 
   console.info(`added user to: ${billingProject}`)
+
+  const userList = await page.evaluate((billingProject) => {
+    return window.Ajax().Billing.project(billingProject).listUsers()
+  }, billingProject)
+
+  function getval(email) {
```
jdcanas: Functions should be camel case, and typically defined via an arrow function, i.e. `const getVal = (email) => { ... }`

integration-tests/utils/integration-helpers.js
```diff
@@ -80,6 +80,18 @@ const addUserToBilling = _.flow(withSignedInPage, withUserToken)(async ({ page,
...
+
+    let obj = userList.filter(item => item.email === email)
+    return obj[0].role
```
jdcanas: We shouldn't have to hard index into the filter result. This is very risky as it could cause the entire web application to crash (very easy to imagine a scenario where the resultant list here is empty). It's better to map the filtered result and display that.

integration-tests/utils/integration-helpers.js
zarsky-broad: Also, we generally don't write the parentheses around a single argument.

integration-tests/utils/integration-helpers.js
zarsky-broad: In this case lodash can make life simple:
```js
const user = _.find({ email }, userList)
```

xuscrbw: Added this

integration-tests/utils/integration-helpers.js
xuscrbw: Fixed this

integration-tests/utils/integration-helpers.js
xuscrbw: I didn't end up needing to map this. Should I be accounting for needing to map?

integration-tests/utils/integration-helpers.js
```diff
@@ -80,6 +80,17 @@ const addUserToBilling = _.flow(withSignedInPage, withUserToken)(async ({ page,
...
+  const userList = await page.evaluate((billingProject) => {
+    return window.Ajax().Billing.project(billingProject).listUsers()
+  }, billingProject)
+
+  const getVal = email => {
+    const userKey = _.find({email}, userList)
+    return userKey.role
+  }
+
+  console.info(`test user was added to the billing project with the role: ${getVal(email)}`)
```
zarsky-broad: Why not just simplify this?
```suggestion
  const billingUser = _.find({ email }, userList)

  console.info(`test user was added to the billing project with the role: ${!!billingUser && billingUser.role}`)
```

zarsky-broad (approved):
Haven't been following along with the underlying troubleshooting, but if this is helpful, nifty.

integration-tests/utils/integration-helpers.js
jdcanas: No, the find covers it 👍

jdcanas (approved):
:shipit:

integration-tests/utils/integration-helpers.js
xuscrbw: made this change. very elegant suggestion!

PR #2405, opened 2021-03-11
Add more flag icons for regional buckets

wnojopra:
This pull requests gets all the regions from https://cloud.google.com/storage/docs/locations and adds an appropriate flag icon for it. 

Example screenshot from local testing:
Screenshot 2021-03-10 4 15 06 PM

### DIFF
--- a/src/components/region-common.js
+++ b/src/components/region-common.js
@@ -17,8 +17,48 @@ export const regionInfo = (location, locationType) => {
       }
     case 'region':
       switch (location) {
+        case 'ASIA-EAST1':
+          return { flag: '🇹🇼', regionDescription: `${locationType}: ${location} (Taiwan)` }
+        case 'ASIA-EAST2':
+          return { flag: '🇭🇰', regionDescription: `${locationType}: ${location} (Hong Kong)` }
+        case 'ASIA-NORTHEAST1':
+          return { flag: '🇯🇵', regionDescription: `${locationType}: ${location} (Tokyo)` }
+        case 'ASIA-NORTHEAST2':
+          return { flag: '🇯🇵', regionDescription: `${locationType}: ${location} (Osaka)` }
+        case 'ASIA-NORTHEAST3':
+          return { flag: '🇰🇷', regionDescription: `${locationType}: ${location} (Seoul)` }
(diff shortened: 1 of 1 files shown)


zarsky-broad (approved):
Looks fine to me. We're working on finding a second person to review.

src/components/region-common.js
```diff
@@ -17,8 +17,48 @@ export const regionInfo = (location, locationType) => {
...
+        case 'SOUTHAMERICA-EAST1':
+          return { flag: '🇧🇷', regionDescription: `${locationType}: ${location} (Sao Paulo)` }
+        case 'US-CENTRAL1':
+        case 'US-EAST1':
+        case 'US-EAST4':
+        case 'US-WEST1':
+        case 'US-WEST2':
+        case 'US-WEST3':
+        case 'US-WEST4':
+          return { flag: '🇺🇸', regionDescription: `${locationType}: ${location}` }
```
breilly2: Should this show a city/country/whatever like the others do?

breilly2 (approved):
👍   Looks good to me. I didn't test (not exactly sure how, and there are a lot of possible cases), but I don't see any problems, and risk of seriously breaking anything is very low.

src/components/region-common.js
wnojopra: Looked at this a little bit more, and I think this is fine as is. For non-US regions its sometimes important to know the city/country for regulatory reasons. Its less important for US regions, and showing the region name (e.g. US-WEST4) should suffice.

wnojopra: Actually now that I've worked on https://github.com/DataBiosphere/terra-ui/pull/2420/ I think we need the US cities too. I've added that.

breilly2:
Still looks good to me. Are you ready to merge now or do you want to see if anything else falls out of #2420 ?

wnojopra:
> Still looks good to me. Are you ready to merge now or do you want to see if anything else falls out of #2420 ?

This is good to merge. Would you mind merging this PR for me? I don't have write permission on this repo.

breilly2:
Since this PR is not able to run the integration tests automatically in CircleCI, I ran them locally against this branch. They passed.

PR #2410, opened 2021-03-18
convert notebooks tab and export modal to hooks

zarsky-broad:
Should be pretty straightforward, but you'll save your brain if you review with whitespace hidden...

### DIFF
--- a/src/pages/workspaces/workspace/Notebooks.js
+++ b/src/pages/workspaces/workspace/Notebooks.js
@@ -1,7 +1,7 @@
 import * as clipboard from 'clipboard-polyfill/text'
 import _ from 'lodash/fp'
 import * as qs from 'qs'
-import { Component, Fragment } from 'react'
+import { Fragment, useEffect, useState } from 'react'
 import { a, div, h, label, span } from 'react-hyperscript-helpers'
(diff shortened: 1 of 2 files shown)


rtitle (approved):
LGTM - the logic seems the same as before, as far as I can tell. I tested notebooks and Galaxy using this branch and all seems fine.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -148,7 +148,7 @@ const NotebookCard = ({ namespace, name, updated, metadata, listView, wsName, on
     style: _.merge({
       ...Style.elements.card.title, whiteSpace: 'normal', overflowY: 'auto'
     }, listView ? {
-      marginLeft: '1rem'
+      marginLeft: '1rem', flexGrow: 1
```
Shakespeared, line 151: What's the benefit of having flexGrow here? @ehanna4

Shakespeared (approved):
LGTU! 👍

src/pages/workspaces/workspace/Notebooks.js
zarsky-broad: It makes a difference in Firefox, otherwise the titles don't default to full width consistently like they do in Chrome.

PR #2411, opened 2021-03-18
POC: implement nested error messages in modals

panentheos:
This introduces a bit of infrastructure to allow notifications to be nested inside modals and drawers, which could be used to improve error handling within these parts of the UI.

There are three pieces that work together:
1. A new `useNotificationState` hook, which attaches a new local notification list to a component, and provides `notify` and `reportError` functions for updating it, which are compatible with their global counterparts.
2. A change to `withErrorReporting` to optionally accept a function, to allow routing error messages to the local list instead of the global one.
3. A new `LocalNotifications` component which can be used to render the local notifications within any 'positioned' container, e.g. modals, drawers.

See the `Environments.js` change for how this would look in practice. I tested by trying to delete one of the 'stuck' runtimes in dev, which generates an error.

This seems to work pretty well, but there are some caveats we might consider before moving forward:
* Currently we're rendering these similarly to the existing ones, as overlays in the upper right. Because modals are more compact, this might end up being overly disruptive. That said, the immediate use case is for errors, which are disruptive by definition.
* The local notifications don't currently use animation. Due to the positioning, we probably couldn't use the same 'slide' animation anyway, so we'd need to come up with something new if we wanted that.

### DIFF
--- a/src/libs/error.js
+++ b/src/libs/error.js
@@ -3,21 +3,26 @@ import { reloadAuthToken, signOut } from 'src/libs/auth'
 import { notify, sessionTimeoutProps } from 'src/libs/notifications'
 
 
-export const reportError = async (title, obj) => {
+export const reportErrorWith = _.curry(async (notifyFn, title, obj) => {
   if (obj instanceof Response && obj.status === 401 && !(await reloadAuthToken())) {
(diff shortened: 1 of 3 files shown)


zarsky-broad:
This was a PoC, so closing it out.

PR #2412, opened 2021-03-18
track mixpanel events for snapshots [AS-556]

Shakespeared:
This is to add some snapshot specific tracking to our mixpanel events. 

- (creating a snapshot reference was already taken care of by the data import tracking)
- added tracking for viewing snapshot entity contents 
- added a field to the launch workflow event to track whether it came from a snapshot or not

For the third point, I coerced it into a boolean, but I could also just track the import format. That would capture more info overall, but also require more work in mixpanel to be able to check how many workflows were launched from snapshots vs other data sources. If you have opinions on that please leave a comment! 

I did not create an event for deleting snapshot references because that function isn't in the UI yet

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -460,7 +460,7 @@ const WorkspaceData = _.flow(
           error: snapshotMetadataError,
           retryFunction: loadSnapshotMetadata
         }, [
-          _.map(([snapshotName, { resource: { referenceId }, entityMetadata: snapshotTables, error: snapshotTablesError }]) => {
+          _.map(([snapshotName, { resource: { referenceId, reference: { snapshot } }, entityMetadata: snapshotTables, error: snapshotTablesError }]) => {
             const snapshotTablePairs = toSortedPairs(snapshotTables)
             return h(Collapse, {
               key: snapshotName,
@@ -504,7 +504,8 @@ const WorkspaceData = _.flow(
                       setSelectedDataType([snapshotName, tableName])
                       Ajax().Metrics.captureEvent(Events.workspaceSnapshotContentsView, {
                         ...extractWorkspaceDetails({ workspace }),
-                        snapshotId: referenceId,
+                        referenceId,
+                        snapshotId: snapshot,
                         entityType: tableName
                       })
                       forceRefresh()
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -21,6 +21,7 @@ import { SnapshotInfo } from 'src/components/workspace-utils'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { reportError, withErrorReporting } from 'src/libs/error'
+import Events, { extractWorkspaceDetails } from 'src/libs/events'
 import { pfbImportJobStore } from 'src/libs/state'
 import * as StateHistory from 'src/libs/state-history'
 import * as Style from 'src/libs/style'
@@ -501,6 +502,11 @@ const WorkspaceData = _.flow(
                     selected: _.isEqual(selectedDataType, [snapshotName, tableName]),
                     onClick: () => {
                       setSelectedDataType([snapshotName, tableName])
+                      Ajax().Metrics.captureEvent(Events.workspaceSnapshotContentsView, {
+                        ...extractWorkspaceDetails({ workspace }),
+                        snapshotName: snapshotName,
+                        entityType: tableName
+                      })
                       forceRefresh()
                     }
                   }, [`${tableName} (${count})`])
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -460,7 +460,7 @@ const WorkspaceData = _.flow(
           error: snapshotMetadataError,
           retryFunction: loadSnapshotMetadata
         }, [
-          _.map(([snapshotName, { entityMetadata: snapshotTables, error: snapshotTablesError }]) => {
+          _.map(([snapshotName, { resource: { referenceId }, entityMetadata: snapshotTables, error: snapshotTablesError }]) => {
             const snapshotTablePairs = toSortedPairs(snapshotTables)
             return h(Collapse, {
               key: snapshotName,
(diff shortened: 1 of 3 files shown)


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -407,7 +407,7 @@ const WorkflowView = _.flow(
           onDismiss: () => this.setState({ launching: false }),
           onSuccess: submissionId => {
             const { methodRepoMethod: { methodVersion, methodNamespace, methodName, methodPath, sourceRepo } } = modifiedConfig
-            Ajax().Metrics.captureEvent(Events.workflowLaunch, { ...extractWorkspaceDetails(workspace), methodVersion, sourceRepo, methodPath: sourceRepo === 'agora' ? `${methodNamespace}/${methodName}` : methodPath })
+            Ajax().Metrics.captureEvent(Events.workflowLaunch, { ...extractWorkspaceDetails(workspace), fromSnapshot: entitySelectionModel.type.description === 'process snapshot table', methodVersion, sourceRepo, methodPath: sourceRepo === 'agora' ? `${methodNamespace}/${methodName}` : methodPath })
```
davidangb: does `entitySelectionModel.type === processSnapshotTable` work here, instead of looking at the description? Just in case the description changes in the future.

src/pages/workspaces/workspace/Data.js
```diff
@@ -501,6 +502,11 @@ const WorkspaceData = _.flow(
                     selected: _.isEqual(selectedDataType, [snapshotName, tableName]),
                     onClick: () => {
                       setSelectedDataType([snapshotName, tableName])
+                      Ajax().Metrics.captureEvent(Events.workspaceSnapshotContentsView, {
+                        ...extractWorkspaceDetails({ workspace }),
+                        snapshotName,
+                        entityType: tableName
+                      })
```
davidangb, line 510: could you include the snapshotId here? That is a more stable identifier than the snapshot name, since name is user-editable

davidangb (commented):
two comments inline. I tested by running locally and seeing the events fire off via ajax when loading snapshot tables, and seeing the new `fromSnapshot` property in the event when launching a workflow.

src/pages/workspaces/workspace/workflows/WorkflowView.js
Shakespeared: It's a symbol so reading directly from `entitySelectionModel.type` doesn't work. I agree reading the description is a bit clunky but I couldn't get anything else to read correctly, so if you have a better idea please let me know! I considered extracting an 'isSnapshot' property to state, but it's only used twice so that seemed like it might be overkill.

ansingh7115 (approved):
Looks good aside from David's comments. If we add deleting the snapshot reference, tracking the mixpanel event for that can be part of a different PR.

src/pages/workspaces/workspace/workflows/WorkflowView.js
davidangb: I see `=== processSnapshotTable` elsewhere in this file - line 751 for example - does that not work in this context? I haven't tried it myself and am unsure, so this is an honest question.

Shakespeared: Yeah I also thought that would work! That's what I did first but it seemed like it wasn't getting updated correctly. It would work fine the first time I ran a workflow, but then it would never change afterwards, so it seemed like it wasn't updating when you changed the root entity type from the dropdown.

Shakespeared: I'm honestly not 100% sure why reading the description works but just evaluating it doesn't…

Shakespeared: Ok! Got Isaac to explain what symbols are and now it's fixed, I was comparing it to a string instead of to the symbol itself, which was why it didn't work before but does now!

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -407,7 +407,7 @@ const WorkflowView = _.flow(
           onDismiss: () => this.setState({ launching: false }),
           onSuccess: submissionId => {
             const { methodRepoMethod: { methodVersion, methodNamespace, methodName, methodPath, sourceRepo } } = modifiedConfig
-            Ajax().Metrics.captureEvent(Events.workflowLaunch, { ...extractWorkspaceDetails(workspace), methodVersion, sourceRepo, methodPath: sourceRepo === 'agora' ? `${methodNamespace}/${methodName}` : methodPath })
+            Ajax().Metrics.captureEvent(Events.workflowLaunch, { ...extractWorkspaceDetails(workspace), fromSnapshot: entitySelectionModel.type === processSnapshotTable, methodVersion, sourceRepo, methodPath: sourceRepo === 'agora' ? `${methodNamespace}/${methodName}` : methodPath })
```
davidangb: could you also include the data repo snapshot id and/or the reference id in this event (see previous comment)? I don't know how easily available those are at this point; both would be great, getting at least one of those would be just fine.

src/pages/workspaces/workspace/Data.js
```diff
@@ -501,6 +502,11 @@ const WorkspaceData = _.flow(
                     selected: _.isEqual(selectedDataType, [snapshotName, tableName]),
                     onClick: () => {
                       setSelectedDataType([snapshotName, tableName])
+                      Ajax().Metrics.captureEvent(Events.workspaceSnapshotContentsView, {
+                        ...extractWorkspaceDetails({ workspace }),
+                        snapshotId: referenceId,
+                        entityType: tableName
+                      })
```
davidangb, line 510: nit: could you rename the `snapshotId` property to `referenceId`, since this is actually the reference's uuid? If possible, could you also include the underlying data repo snapshot's id?

Given this payload:
```
{
  "cloningInstructions": "COPY_NOTHING",
  "description": "",
  "name": "snappy",
  "reference": {
    "instanceName": "terra",
    "snapshot": "cf0e417d-1456-401d-bbf0-a3fa2bd3afa3"
  },
  "referenceId": "34542b10-e69a-4f02-84c2-6b4b8c6fc852",
  "referenceType": "DATA_REPO_SNAPSHOT",
  "workspaceId": "60bd54fe-5ca2-4763-b9f5-03a0a5083f89"
}
```

the ideal would be to include
```
referenceId: "34542b10-e69a-4f02-84c2-6b4b8c6fc852",
snapshotId: "cf0e417d-1456-401d-bbf0-a3fa2bd3afa3"
```

in the event.

davidangb (commented):
the code changes here look great, thank you for pushing through that! I have more scope-creep requests inline - sorry, it wasn't until I really inspected the payloads that I realized what we should include. Please push back if the requests aren't realistic!

davidangb (approved):
looks good, thank you!

PR #2413, opened 2021-03-19
convert share workspace modal to hooks

zarsky-broad:
Some rearranging here, so the changes look more significant than they actually are, even with whitespace hidden, but the logic is largely the same


src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -174,73 +133,111 @@ export default Utils.withCancellationSignal(class ShareWorkspaceModal extends Co
         email,
         pending && div({ style: styles.pending }, ['Pending']),
         h(AclInput, {
-          autoFocus: !!numAdditions,
+          autoFocus: email === lastAddedEmail,
```
ehanna4, line 136: nice simplification - much easier to understand

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -174,73 +133,111 @@ export default Utils.withCancellationSignal(class ShareWorkspaceModal extends Co
...
-    const eventData = { numAdditions, ...extractWorkspaceDetails(this.props.workspace) }
-
-    const aclUpdates = [
-      ..._.flow(
-        _.remove({ accessLevel: 'PROJECT_OWNER' }),
-        _.map(_.pick(['email', 'accessLevel', 'canShare', 'canCompute']))
-      )(acl),
-      ..._.map(({ email }) => ({ email, accessLevel: 'NO ACCESS' }), needsDelete)
-    ]
+    load()
```
ehanna4, line 177: Do we need an await here since load() is async? Also how come the old version didn't need the async?

@Shakespeared

zarsky-broad: So, we can't await, since this is inside a `useEffect` (that's what `useOnMount` actually is); we can only kick it off, it has to await internally. The old version was async, look at the old line 190.

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -74,93 +74,52 @@ const AclInput = ({ value, onChange, disabled, maxAccessLevel, autoFocus }) => {
...
+  const [searchValue, setSearchValue] = useState('')
+  const [acl, setAcl] = useState([])
+  const [loaded, setLoaded] = useState(false)
+  const [working, setWorking] = useState(false)
+  const [updateError, setUpdateError] = useState(undefined)
+  const [lastAddedEmail, setLastAddedEmail] = useState(undefined)
 
-    this.list = createRef()
-  }
+  const list = useRef()
```
ehanna4, line 89: Is this proper form to name a variable the same thing as a data-type?

@Shakespeared

zarsky-broad: I could have called it, like, `listDomNode` or something, I just kept the name the same as it was, tho.

ehanna4 (approved):
LGTU (pending 2 small questions)
@Shakespeared

Shakespeared (approved):
thumbs up, we tested in UI

src/pages/workspaces/workspace/ShareWorkspaceModal.js
ehanna4: oh yeah sorry should have clarified this wasn't a comment about a change due to this PR but was just wondering if in making this conversion if it would be appropriate to rename this variable (assuming `list` isn't a proper/standard variable name since it is the name of a data-type)

PR #2414, opened 2021-03-19
convert workflow tabs & components to hooks

zarsky-broad:
Much tidier to review with whitespace hidden. Tested locally!

### DIFF
--- a/src/pages/workspaces/workspace/JobHistory.js
+++ b/src/pages/workspaces/workspace/JobHistory.js
@@ -1,5 +1,5 @@
 import _ from 'lodash/fp'
-import { Fragment, useRef, useState } from 'react'
+import { Fragment, useImperativeHandle, useRef, useState } from 'react'
 import { div, h, span, table, tbody, td, tr } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import * as breadcrumbs from 'src/components/breadcrumbs'
@@ -147,6 +147,8 @@ const JobHistory = _.flow(
       }
     }
   })
+
+  useImperativeHandle(ref, () => ({ refresh }))
 
 
   // Render
--- a/src/pages/workspaces/workspace/JobHistory.js
+++ b/src/pages/workspaces/workspace/JobHistory.js
@@ -1,5 +1,5 @@
 import _ from 'lodash/fp'
-import { Component, Fragment } from 'react'
+import { Fragment, useRef, useState } from 'react'
 import { div, h, span, table, tbody, td, tr } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import * as breadcrumbs from 'src/components/breadcrumbs'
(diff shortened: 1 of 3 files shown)


src/pages/workspaces/workspace/Workflows.js
```diff
@@ -156,102 +156,48 @@ const WorkflowCard = Utils.memoWithName('WorkflowCard', ({ listView, name, names
...
+      setSelectedWorkflow(undefined)
     }
   }
 
-  async exportMethod() {
-    const { namespace, name, ws } = this.props
-    const { selectedWorkflow } = this.state
-
-    this.setState({ exporting: true })
+  const exportMethod = async () => {
```
rsasch, line 199: nit: I see that the additional line returns within this call existed before this PR, but maybe they can be removed (tried to select lines 200 - 212 but github threw an error)?

rsasch (approved):
small nit, otherwise 👍🏻

cjllanwarne (approved):
I didn't go through line-by-line but the pages are still functional when I browse to them, and I don't see anything obviously wrong here.

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -128,183 +125,188 @@ const JobHistory = _.flow(
...
+                    onClick: () => setAbortingId(submissionId)
+                  }, ['Abort workflows']),
+                  isTerminal(status) && (workflowStatuses['Failed'] || workflowStatuses['Aborted']) &&
+                  submissionEntity && !methodConfigurationDeleted && h(ButtonPrimary, {
+                    onClick: () => rerunFailures({
+                      workspace,
+                      submissionId,
+                      configNamespace: methodConfigurationNamespace,
+                      configName: methodConfigurationName,
+                      onDone: refresh
```
panentheos, line 258: TOL: It looks like there may be some pre-existing issues here, since this will kick off a separate refresh loop without stopping the first one. Not sure if it's worth fixing now, but we should probably do something like the `reschedule` pattern in `useCloudEnvironmentPolling`.

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -287,137 +287,134 @@ const noWorkflowsMessage = div({ style: { fontSize: 20, margin: '1rem' } }, [
...
-  constructor(props) {
-    super(props)
-    this.state = {
-      loading: true,
-      sortOrder: defaultSort.value,
-      filter: '',
-      ...StateHistory.get()
-    }
-  }
+  withViewToggle('workflowsTab')
```
panentheos: nit: There's a `useViewToggle` hook we could switch this to now.

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -91,28 +91,25 @@ const noJobsMessage = div({ style: { fontSize: 20, margin: '1rem' } }, [
...
   wrapWorkspace({
     breadcrumbs: props => breadcrumbs.commonPaths.workspaceDashboard(props),
     title: 'Job History', activeTab: 'job history'
-  }),
-  Utils.withCancellationSignal
-)(class JobHistory extends Component {
-  constructor(props) {
-    super(props)
+  })
+)(({ namespace, name, workspace, workspace: { workspace: { bucketName } } }, ref) => {
```
panentheos, line 99: We need a `useImperativeHandle` on this `ref` to wire up the `refresh` function to the tab bar.

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -287,137 +287,134 @@ const noWorkflowsMessage = div({ style: { fontSize: 20, margin: '1rem' } }, [
...
+)(({ namespace, name, listView, setListView, workspace: ws, workspace: { workspace } }, ref) => {
```
panentheos: This also needs a `useImperativeHandle` to wire up `refresh`.

panentheos (approved):

src/pages/workspaces/workspace/JobHistory.js
zarsky-broad: What if I just add this to the top of the refresh function itself?
```js
    if (scheduledRefresh.current) {
      clearTimeout(scheduledRefresh.current)
    }
```

PR #2415, opened 2021-03-22
Add a delete button to snapshots [AS-699]

zarsky-broad:
Verified locally! Waiting for final 👍 from UX.
Screen Shot 2021-03-22 at 16 01 39
Screen Shot 2021-03-22 at 16 01 43

### DIFF
--- a/src/components/workspace-utils.js
+++ b/src/components/workspace-utils.js
@@ -229,7 +229,7 @@ export const SnapshotInfo = ({
             Ajax().Metrics.captureEvent(Events.workspaceSnapshotDelete, {
               ...extractWorkspaceDetails({ workspace }),
               referenceId,
-              snapshotId,
+              snapshotId
             })
             onDelete()
           } catch (e) {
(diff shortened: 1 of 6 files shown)


Shakespeared (approved):
LGTM, tested locally in UI

we should remember to make a ticket to add deletion tracking to mixpanel once this is merged

davidangb:
1) Instead of a separate ticket, could we scope-creep this one to include the mixpanel event for deleting the snapshot?

2) If the call to delete the snapshot throws an error, the user is left with an infinite spinner. I simulated an error by forcing the snapshotId to be an invalid uuid in ajax.js:
```
const snapshotPath = `${root}/snapshots/${snapshotId}XXX`
```

src/components/workspace-utils.js
```diff
@@ -213,6 +219,37 @@ export const SnapshotInfo = ({
...
       ]),
+      deleting && h(Modal, {
+        onDismiss: () => setDeleting(false),
+        okButton: async () => {
+          try {
+            setSaving(true) // this will be unmounted in the reload, so no need to reset this
+            setDeleting(false)
+            await Ajax().Workspaces.workspace(namespace, name).snapshot(referenceId).delete()
+            Ajax().Metrics.captureEvent(Events.workspaceSnapshotDelete, {
+              ...extractWorkspaceDetails({ workspace }),
```
Shakespeared: Nit: I don't think you need the curly braces around workspace here

davidangb (approved):

PR #2416, opened 2021-03-23
[SATURN-1887] Ensure that the Upload In Progress modal cannot be dismissed

andrewmarcus:
This PR ensures that the Upload in Progress modal on the `/#upload` page can't be dismissed accidentally, except by explicitly aborting the upload with the big red "Abort Upload" button. Currently, pressing the Escape key or clicking outside the modal aborts the upload but it's not clear that that's what happened.

This disables all of the other means of dismissing the modal besides the button. The user can't do anything else in their project while the upload is still going on, but they won't accidentally cancel the upload without realizing it.

In general it would be better to allow uploads in the background in order to not block the user when uploading large files. However, the Metadata Table functionality, the next step, expects all of the files to already be present in the bucket, so there's not really anything else we'd want users to be able to do while waiting anyway.

On the other hand, it's not necessary that all of the bytes actually be in Google so long as we can match filenames in the bucket. Perhaps a better future approach could be to use Google's streaming APIs instead and start all of the uploads synchronously, but then stream them in the background. The Streaming API requires significantly more local state management, though...

### DIFF
--- a/src/components/ProgressBar.js
+++ b/src/components/ProgressBar.js
@@ -1,4 +1,5 @@
 import { dd, div, dl, dt, h, p, strong } from 'react-hyperscript-helpers'
+import { ButtonPrimary } from 'src/components/common'
 import Modal from 'src/components/Modal'
 import colors from 'src/libs/colors'
 import { friendlyFileSize } from 'src/libs/uploads'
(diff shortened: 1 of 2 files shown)


breilly2:
Could someone get themselves in trouble if they start interacting with the data in another window that is not blocked by a non-dismissible modal?

andrewmarcus:
That's a good question. I will argue that it isn't a problem; here's why.

For someone using the new `/#upload` interface, the only step remaining after they upload all their files is to upload their metadata file. That process looks for filenames in the Google bucket and replaces them in the upload TSV with full bucket paths.

If the files aren't there yet, it won't do the substitution, so they could end up with Data tables which don't reference actual files in the Google bucket. However, this can be corrected by importing the same metadata file a second time once all the files are in place.

dmohs (approved):
All changes are to Andrew's own files, so risk to rest of product is basically none.

abaumann (approved):
Reviewed the functionality and it worked as expected - was not able to close that dialog or abort without using the big red button

dmohs:
Tests fail because this is a fork. Merging with failing tests since risk is low.

PR #2417, opened 2021-03-23
[IA-2389][IA-2596] Custom hardware config for Galaxy apps and other small fixes

kyuksel:
[JIRA ticket
](https://broadworkbench.atlassian.net/browse/IA-2389)

### Highlights
- Galaxy users can now specify how many VMs to run their workflows on and how many/much vCPUs/memory each VM will have. 
- They can also select a size for their persistent data disk (metadata disk will remain fixed size of `10 GB`, which isn't directly exposed to users).
- Allowing users to update their hardware configuration is out of scope. To make this more clear, added a note to the info page right before app creation. Furthermore per UX request, revamped that page's visuals with Font Awesome icons for each bullet.
- Tried to make the app config customization window's layout as consistent as possible with that of the runtime modal's.
- Fixed a bug where hourly cost was being displayed for Galaxy disks although the text said `per month` 🤦🏼 
- Added a tooltip for the `Delete` button when a Galaxy app is in states like `pausing`, `paused` and `resuming`, informing user that they resume first. Also disabled the `Pause` and `Delete` buttons where applicable.
- Filtered out some of the low cpu/memory combinations based on feedback from the Galaxy team regarding usage patterns.
- Stylistic changes to fix typos and standardize user-facing messaging across Runtime and Galaxy modals.

### Notes
- Added plumbing to route the list of Galaxy data (NFS) disks from `WorkspaceContainer` through `Notebooks` and `RuntimeManager` to `NewGalaxyModal`.
- Put `MachineSelector` outside `NewGalaxyModal` so it can be called as a component, consistently with `NewRuntimeModal`'s. The other way, it was causing a very strange behaviour on the `Nodes` `NumberInput` component where incrementing and inserting double-digit numbers weren't working properly.
- Changed wording of messages displayed to users regarding what the cloud environment is doing at a given stage, in hopes of making them more grammatically consistent and also using the terms `pause/resume` consistently instead of a mix of them and `stop/start`. There is another ticket in progress for doing that throughout the codebase.
- Fixed typos.
- Punted on the `TODO`s around refactoring since the UI redesign slated for Q2 will probably make them irrelevant but kept them in in case that's not the case.

### Verification
- Manually clicked through lifecycles of a Galaxy app.
- [galaxy_screenshots_210326.pdf](https://github.com/DataBiosphere/terra-ui/files/6212342/galaxy_screenshots_210326.pdf) shows a lifecycle for a Galaxy app showing some of the changes in this PR. Note that the final versions are slightly different upon PR and UX feedback.


src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -307,15 +308,14 @@ const Notebooks = _.flow(
       }))
     )(notebooks)
 
-    const getGalaxyText = () => {
-      return app ?
+    const getGalaxyText = (app, galaxyDataDisks) => {
+      const dataDisk = currentDataDisk(app, galaxyDataDisks)
+      return (app && dataDisk) ?
```
kyuksel: Added `&& dataDisk` to prevent the error that occasionally happened (probably due to race condition at the first render of the component)

src/components/NewGalaxyModal.js
```diff
@@ -250,3 +322,59 @@ export const NewGalaxyModal = _.flow(
...
+        label({ htmlFor: id, style: styles.label }, ['Nodes']),
+        div([
+          h(NumberInput, {
+            id,
+            min: 1,
+            max: maxNodepoolSize,
+            isClearable: false,
+            onlyInteger: true,
+            value: value.numNodes,
+            onChange: n => onChange(prevState => { return { ...prevState, numNodes: n } })
```
kyuksel, line 370: Found out that updating JS objects with multiple (nested or not) fields works differently for function components than class components. For the latter, if only one field of such a JS object is updated, it gets merged as desired. For the former, one has to update the entire object, which is why I used the object spread syntax with `…` as suggested by [this blog](https://blog.logrocket.com/a-guide-to-usestate-in-react-ecb9952e406c).

src/components/NewGalaxyModal.js
```diff
@@ -186,59 +198,118 @@ export const NewGalaxyModal = _.flow(
...
   }
 
-  const getEnvMessageBasedOnStatus = isTitle => {
-    const waitMessage = isTitle ? '' : 'This process will take up to a few minutes.'
+  const getEnvMessageBasedOnStatus = app => {
+    const waitMessage = 'This process will take up to a few minutes.'
 
     return !app ?
-      isTitle ? 'Cloud environment' : 'Environment will consist of an application and cloud compute.' :
+      'consists of application configuration, cloud compute and persistent disk(s).' :
```
ehanna4: Was this change recommended by UX? Mostly curious because in the cloud env version it says `"Cloud environments consist of an application configuration, cloud compute and a persistent disk" `so this seems like a change that makes it less similar rather than more similar

src/components/NewGalaxyModal.js
```diff
@@ -186,59 +198,118 @@ export const NewGalaxyModal = _.flow(
...
+      ])
+    ])
+  }
+
+  const renderPersistentDiskSection = () => {
+    const gridStyle = { display: 'grid', gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', gridGap: '1rem', alignItems: 'center' }
+    return div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
+      div({ style: styles.headerText }, ['Persistent disk']),
+      div({ style: { marginTop: '0.5rem' } }, [
+        'Stores your analysis data '
```
ehanna4:
```suggestion
        'Stores your analysis data.'
```

ehanna4: Wondering if having the documentation about persistent disks linked here (like it is in the cloud env version) would be beneficial? ie:
`Stores your analysis data. Learn more about Persistent disks and where your disk is mounted`

edit: oops just saw your TODO on this haha

xuscrbw:
Screen Shot 2021-03-25 at 2 51 56 PM
Noticing this:

1. I don't know if the cost here makes sense. Am using the default App for this. Also, The text seems to fall out of the box. Probably has to do with the cost per hr being longer on top of provisioning being a long word. Not sure of a clean fix for this. @ehanna4 Have you run into this before?

EDIT: Was able to confirm here: I think the math being done here is wrong. We have a per Month cost from the disk being added to a per hour cost for everything else. This will probably need to be changed.

xuscrbw (commented):
Screen Shot 2021-03-25 at 3 12 58 PM
Question about this view: right now I have an app running and it looks like the numbers here can still be changed. Is this how we are intending for this view to look? Not sure if changing these will be applied automatically, or if something else needs to be happening.

kyuksel:
@xuscrbw Great catch on the wrong cost display! There used to be a bug in persistent disk reporting, which I fixed but that introduced the one you found. 

> I don't know if the cost here makes sense.
I don't think we need to display the cost there but let me see if I can get some UX input as I may be missing some historical context.

src/components/NewGalaxyModal.js
xuscrbw: Going off of this, I will say that preference wise the two separate headings across the same sentence feels a bit off. I don't think there's actually any functional difference here, but was more of stylistic preference for me

kyuksel: It wasn't particularly recommended by UX. I noticed that on the current runtime version below, `Cloud environment` is duplicated, and I wanted to remove duplication. I also wanted to be able to use text consistently across runtimes and apps, and since Galaxy apps have multiple disks, I turned `disk` into `disk(s)`. Note that both runtime and Galaxy modals now use it so it'll be the same. 

No functional difference indeed.

Do these address your questions?

image

kyuksel:
> Question about this view: right now I have an app running and it looks like the numbers here can still be changed. Is this how we are intending for this view to look?

@xuscrbw I wondered the same myself and then realized that the runtime modal has it the same way and decided to keep Galaxy's consistent with the runtime one. One difference is that Galaxy apps don't yet support updates for running instances so a user can play with configs (perhaps to see how that can change cost estimates) but nothing will change resource-wise.

src/components/NewGalaxyModal.js
```diff
@@ -68,9 +82,11 @@ export const NewGalaxyModal = _.flow(
     return onSuccess()
   })
 
-  const deleteButton = h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete'])
-
   const renderActionButton = () => {
```
kyuksel, line 85: The changes in this function below were upon the UX request to add disabled `Delete` and `Pause` buttons to transitionary status modals.

src/components/NewGalaxyModal.js
```diff
@@ -83,23 +99,29 @@ export const NewGalaxyModal = _.flow(
...
-            deleteButton,
-            h(ButtonPrimary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume'])
+            h(ButtonSecondary, { disabled: true, style: { marginRight: 'auto' }, tooltip: 'Cloud Compute must be resumed first.', onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+            resumeButton()
           ])],
-          ['ERROR', () => deleteButton]
+          ['ERROR', () => deleteButton],
+          [Utils.DEFAULT, () => {
+            return h(Fragment, [
+              deleteButton(true),
```
ehanna4: I think this should be `false`. Otherwise it is not allowing the user to delete the galaxy if it is errorred.

src/components/NewGalaxyModal.js
```diff
@@ -83,23 +99,29 @@ export const NewGalaxyModal = _.flow(
...
+            h(ButtonPrimary, { disabled: false, onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])
           ])],
           ['STOPPED', () => h(Fragment, [
-            deleteButton,
-            h(ButtonPrimary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume'])
+            h(ButtonSecondary, { disabled: true, style: { marginRight: 'auto' }, tooltip: 'Cloud Compute must be resumed first.', onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+            resumeButton()
           ])],
-          ['ERROR', () => deleteButton]
+          ['ERROR', () => deleteButton],
```
rtitle, line 118: Should this be `deleteButton()` (do parens matter?)

src/components/NewGalaxyModal.js
rtitle: Seems the `ERROR` case is handled above on L118?

kyuksel: Totally! Thanks for catching this 🙏🏼

kyuksel: Oh actually, the `ERROR` case is above and does have a `deleteButton`. The one you pointed out is for undelete-able cases.

src/components/NewGalaxyModal.js
kyuksel: Good catch! Parentheses matter in this case because the `deleteButton()` function has a parameter with a default value, which doesn't seem to get set if no `()`.

src/components/NewGalaxyModal.js
rtitle: No strong opinions but perhaps `Consists` should be capitalized?

src/components/NewGalaxyModal.js
```diff
@@ -186,59 +207,118 @@ export const NewGalaxyModal = _.flow(
...
-      isTitle ? 'Cloud environment' : 'Environment will consist of an application and cloud compute.' :
+      'consists of application configuration, cloud compute and persistent disk(s).' :
       Utils.switchCase(app.status,
-        ['STOPPED', () => `Cloud environment is now paused ${!isTitle ? '...' : ''}`],
-        ['PRESTOPPING', () => 'Cloud environment is preparing to stop.'],
-        ['STOPPING', () => `Cloud environment is pausing. ${waitMessage}`],
-        ['PRESTARTING', () => 'Cloud environment is preparing to start.'],
-        ['STARTING', () => `Cloud environment is starting. ${waitMessage}`],
-        ['RUNNING', () => 'Environment consists of an application and cloud compute.'],
+        ['STOPPED', () => `is paused.`],
```
rtitle: I see `pause`, `stop`, `start`, and `resume` here. Can we use consistent language?

src/components/NewGalaxyModal.js
```diff
@@ -83,23 +99,29 @@ export const NewGalaxyModal = _.flow(
...
-            deleteButton,
-            h(ButtonPrimary, { style: { marginRight: '1rem' }, onClick: () => { resumeGalaxy() } }, ['Resume'])
+            h(ButtonSecondary, { disabled: true, style: { marginRight: 'auto' }, tooltip: 'Cloud Compute must be resumed first.', onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+            resumeButton()
           ])],
-          ['ERROR', () => deleteButton]
+          ['ERROR', () => deleteButton],
+          [Utils.DEFAULT, () => {
+            return h(Fragment, [
+              deleteButton(),
```
rtitle: I think it should be `deleteButton(true)` here -- otherwise it makes `STOPPING`, `STARTING` deletable which is not what we want. (Error is handled above).

src/components/NewGalaxyModal.js
```diff
@@ -186,59 +207,118 @@ export const NewGalaxyModal = _.flow(
...
+      ])
+    ])
+  }
+
+  const renderPersistentDiskSection = () => {
+    const gridStyle = { display: 'grid', gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', gridGap: '1rem', alignItems: 'center' }
+    return div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
+      div({ style: styles.headerText }, ['Persistent disk']),
+      div({ style: { marginTop: '0.5rem' } }, [
+        'stores your analysis data. '
```
rtitle: Just IMO, I feel like these text snippets should be capitalized and be complete sentences.

src/components/NewGalaxyModal.js
```diff
@@ -186,59 +207,118 @@ export const NewGalaxyModal = _.flow(
...
+  }
+
   const renderDefaultCase = () => {
-    const defaultMachineType = 'n1-highmem-8'
-    const { cpu, memory } = _.find({ name: defaultMachineType }, machineTypes)
-    const cost = getGalaxyCost(app || { kubernetesRuntimeConfig: { machineType: defaultMachineType, numNodes: 1 } })
     return h(Fragment, [
       h(TitleBar, {
-        title: getEnvMessageBasedOnStatus(true),
+        title: 'Cloud environment',
```
rtitle: Cloud Environment

src/components/NewRuntimeModal.js
```diff
@@ -821,7 +821,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
           id => h(div, { style: { display: 'flex', flexDirection: 'column' } }, [
             label({ htmlFor: id, style: styles.label }, ['Persistent disk size (GB)']),
             div({ style: { marginTop: '0.5rem' } }, [
-              'Stores your analysis data. ',
+              'stores your analysis data. ',
```
rtitle: Same as above re capitalization

src/components/RuntimeManager.js
```diff
@@ -341,7 +342,7 @@ export default class RuntimeManager extends PureComponent {
             div({ style: { marginLeft: '0.5rem', paddingRight: '0.5rem', color: colors.dark() } }, [
               div({ style: { fontSize: 12, fontWeight: 'bold' } }, 'Cloud Environment'),
               div({ style: { fontSize: 10 } }, [
-                span({ style: { textTransform: 'uppercase', fontWeight: 500 } },
+                span({ style: { textTransform: 'capitalize', fontWeight: 500 } },
```
rtitle, line 345: Things look inconsistent for me now:
image

I think `uppercase` is fine..?

src/components/NewGalaxyModal.js
kyuksel: (For some reason Rob's comment above hadn't shown up on my screen when I wrote the above replies  🤷🏼‍♂️ )

src/components/NewGalaxyModal.js
rtitle: Example from local testing. The app now is `Starting` and the `DELETE` button is clickable, which is not correct:
image

rtitle:
Tested locally and it seems to be functional. Some comments above around presentation. In general I think the text snippets displayed in the various views should be capitalized and complete sentences. Curious what UX's opinion is though.

This PR also needs a rebase, probably conflicts with Isaac's recent PR I'm guessing.

src/components/NewGalaxyModal.js
kyuksel: Totally agreed - especially important for the user-facing ones. I've updated some things in my PR as I touched them but we have a [ticket in progress](https://broadworkbench.atlassian.net/browse/IA-2539) to go through the codebase for it.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -363,7 +361,7 @@ const Notebooks = _.flow(
             tooltip: appIsSettingUp(app) && 'Your Galaxy app is being created',
             onClick: () => setOpenGalaxyConfigDrawer(true)
           }, [
-            getGalaxyText()
+            getGalaxyText(app, galaxyDataDisks)
```
ehanna4, line 364: This tool tip is using `appIsSettingUp` so displays while app is creating but this button continues to say `create a galaxy app`. Probably should use that `appIsSettingUp` for this Galaxy text as well

ehanna4:
> > Question about this view: right now I have an app running and it looks like the numbers here can still be changed. Is this how we are intending for this view to look?
> 
> @xuscrbw I wondered the same myself and then realized that the runtime modal has it the same way and decided to keep Galaxy's consistent with the runtime one. One difference is that Galaxy apps don't yet support updates for running instances so a user can play with configs (perhaps to see how that can change cost estimates) but nothing will change resource-wise.

On this note - it feels weird to be able to play around with configs then hit the `launch` button and have nothing change. I wonder if we could make it clear that the user is just relaunching their existing app and the changes won't apply.

Maybe even on this view `Important: Please keep this tab open and logged in to Terra while using Galaxy.
Galaxy will open in a new tab.` could say like `Also please note to apply changes to your configuration please delete your existing app and create a new one with your changes blah blah` <- probably would need a UX input on that wording.

OR maybe near the `launch button` have an `i` or `*` with a note that it's the same app

ehanna4:
I don't see any update to mention PD in the delete view, when a user clicks delete is that killing everything or does the PD still exist? If it still exists how do we want users to be able to re-access it?

src/components/NewGalaxyModal.js
ehanna4: I can buy your rationale haha, though I agree with Rob I think capitalization would feel better. The bolded "Cloud Environment" is a title so it's feels funny to have it be like part of a fragmented sentence.

src/components/NewGalaxyModal.js
```diff
@@ -68,9 +82,11 @@ export const NewGalaxyModal = _.flow(
     return onSuccess()
   })
 
-  const deleteButton = h(ButtonSecondary, { style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete'])
-
   const renderActionButton = () => {
+    const deleteButton = (isDisabled = false) => h(ButtonSecondary, { disabled: isDisabled, style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete'])
```
ehanna4: nit: could be nice to include tooltips on these buttons when disabled

src/components/NewGalaxyModal.js
xuscrbw: For clarity, did you just mean the "is ..." text itself? Or do you mean, the statuses themselves?

kyuksel: We can skip the statuses like `STOPPED` and `STARTING` above since they're not user-facing but there are some that _are_ user-facing like [this one in RuntimeManager](https://github.com/DataBiosphere/terra-ui/blob/dd5cd6d782bf878d3c33ad6ed0df05c673dd61f6/src/components/RuntimeManager.js#L310) which would be good to rename. What do you think?

src/components/NewGalaxyModal.js
kyuksel: Changed it to

-------------------------

### Cloud Environment
A cloud environment consists of application configuration, cloud compute and persistent disk(s).

-------------------------

as having a full sentence looked better to me. Let me know if any objections.

src/components/RuntimeManager.js
kyuksel: +1 on the inconsistency. Fixed as in below:

image

Uppercasing signifies strong emphasis to me, at least subconsciously, which I doubt is what we want in this case.

kyuksel:
> I don't see any update to mention PD in the delete view, when a user clicks delete is that killing everything or does the PD still exist? If it still exists how do we want users to be able to re-access it?

@ehanna4 This PR doesn't change anything around persistence for Galaxy apps. Until we have the Galaxy Persistence ticket's changes in, we'll still be creating/deleting disks along with apps.

src/pages/workspaces/workspace/Notebooks.js
kyuksel: Thanks for this catch! I changed the logic around and it seems to work as intended ™

kyuksel:
@ehanna4 

>On this note - it feels weird to be able to play around with configs then hit the launch button and have nothing change. I wonder if we could make it clear that the user is just relaunching their existing app and the changes won't apply. Maybe even on this view Important: Please keep this tab open and logged in to Terra while using Galaxy. Galaxy will open in a new tab. could say like Also please note to apply changes to your configuration please delete your existing app and create a new one with your changes blah blah <- probably would need a UX input on that wording. OR maybe near the launch button have an i or * with a note that it's the same app

Yea, good thinking! I'm thinking the heads-up page that shows up right before creating an app may also be a good place (e.g. the last section I just added on the screenshot below) as it has more space to accommodate a wordy message. I'll try to get UX input on that though.

image

kyuksel:
@ehanna4 @rtitle @xuscrbw I believe I've addressed all of your comments as well as Joy's via [this diff](https://github.com/DataBiosphere/terra-ui/pull/2417/files/dd5cd6d782bf878d3c33ad6ed0df05c673dd61f6..48eb2e6ace25e9c96a0b468a65a0c228e8494790), and run it all by Joy. Could you please take another look? 

One section (`NewGalaxyModal.renderCreateWarning()`) that I revamped was per Emily's and Joy's feedback. It used to look like 

image

and now it looks like

image

src/components/RuntimeManager.js
```diff
@@ -297,16 +298,16 @@ export default class RuntimeManager extends PureComponent {
...
+            'View cloud environment',
           onClick: () => {
             this.setState({ galaxyDrawerOpen: true })
           }
         }, [
           img({ src: galaxyLogo, alt: '', style: { marginRight: '0.25rem' } }),
           div([
             div({ style: { fontSize: 12, fontWeight: 'bold' } }, ['Galaxy']),
-            div({ style: { fontSize: 10, textTransform: 'uppercase' } }, [app.status])
+            div({ style: { fontSize: 10, textTransform: 'capitalize' } }, [_.capitalize(app.status)])
```
jdcanas: I believe the use of the `textTransform` property in this div is duplicative of calling `_.capitalize` on the contents.

src/components/NewRuntimeModal.js
rtitle: Can this be capital `S`?

src/components/RuntimeManager.js
rtitle: Thanks.. am ok with `capitalize` instead of `uppercase`, as long as they're consistent. :)

src/components/NewGalaxyModal.js
```diff
@@ -186,59 +228,119 @@ export const NewGalaxyModal = _.flow(
...
+      div({ style: { ...gridStyle, marginTop: '0.75rem' } }, [
+        h(MachineSelector, { value: kubernetesRuntimeConfig, onChange: v => setKubernetesRuntimeConfig(v) })
+      ])
+    ])
+  }
+
+  const renderPersistentDiskSection = () => {
+    const gridStyle = { display: 'grid', gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', gridGap: '1rem', alignItems: 'center' }
+    return div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
+      div({ style: styles.headerText }, ['Persistent Disk']),
```
rtitle: Perhaps this should be `Persistent disk` to match the capitalization scheme of the rest of the page.

image

(I don't know why capitalization sticks out to me so much.. sorry if this is being too nit-picky 😁 )

rtitle (approved):
I tested quite a few cases locally (creating apps with non-standard configs, pause, resume, delete, etc). Everything looks good (modulo the bug https://github.com/DataBiosphere/leonardo/pull/1996 fixes). The code diff also looks good to me (capitalization nits aside ;)).

One general comment: as discussed above, it also felt weird to me to be able to change the compute settings for a `Running` app, with no way to actually update it besides delete and recreate. Perhaps the boxes could be made uneditable for `Running` apps (though they should still be displayed). I think the warning at app creation time covers this though, so I don't feel that strongly to make that change in this PR.

Nice work!

src/components/NewGalaxyModal.js
xuscrbw: I think this suggestion is good. Will look at the best way to accomplish this. Feels like we should have a helper function to make this work, might tackle this in its own PR.

xuscrbw (approved):
I think things look good. 

One difference I did notice was when the app is paused we have hover text for the "Delete" button in the Galaxy Modal, but we don't have similar hover text when the app is pausing or resuming. Is that intentional?

src/components/RuntimeManager.js
kyuksel: Hmm...interestingly, `textTransform: 'capitalize'` doesn't seem to have any effect regardless of `_.capitalize(...)` so I'm taking that out.

src/components/RuntimeManager.js
kyuksel: I'm all for consistency whenever possible 😉

src/components/NewGalaxyModal.js
kyuksel: Oh yea totally - for some reason I saw `Persistent Disk` at the same level as `Cloud Environment` 🤦🏼 

It's refreshing to see I'm not the only one with stylistic nits 😊

src/components/NewRuntimeModal.js
kyuksel: Made it a full sentence to be consistent with Galaxy modal's wording.

Messages Image(696994228)

kyuksel:
> One general comment: as discussed above, it also felt weird to me to be able to change the compute settings for a Running app, with no way to actually update it besides delete and recreate. Perhaps the boxes could be made uneditable for Running apps (though they should still be displayed). I think the warning at app creation time covers this though, so I don't feel that strongly to make that change in this PR.

@rtitle I agree that it's misleading. Will leave implementing a better solution to the UI revamp time if the situation still applies then.

kyuksel:
> One difference I did notice was when the app is paused we have hover text for the "Delete" button in the Galaxy Modal, but we don't have similar hover text when the app is pausing or resuming. Is that intentional?

@xuscrbw Good catch! I've added a slightly modified version of the hover text to the default case so it will be displayed for the missing cases.

PR #2418, opened 2021-03-24
[CA-1254] Ci billing project left menu

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -35,8 +35,6 @@ const styles = {
   }
 }
 
-
-// TODO: Make conform with new mocks from Jerome
 const ProjectListItem = ({ project: { projectName, role, creationStatus, message }, isActive }) => {
   const projectReady = creationStatus === 'Ready'
   const statusIcon = creationStatus === 'Creating' ?
@@ -277,6 +275,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
   const hasBillingProjects = !_.isEmpty(billingProjects)
   const breadcrumbs = `Billing > Billing Project`
 
+  const billingProjectListWidth = 330
   return h(FooterWrapper, [
     h(TopBar, { title: 'Billing' }, [
       !!selectedName && div({
@@ -289,7 +288,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
       ])
     ]),
     div({ role: 'main', style: { display: 'flex', flex: 1 } }, [
-      div({ style: { minWidth: 330, maxWidth: 330, boxShadow: '0 2px 5px 0 rgba(0,0,0,0.25)', height: `calc(100vh - ${topBarHeight})`, overflowY: 'scroll' /* TODO: Make scrollable */ } }, [
+      div({ style: { minWidth: billingProjectListWidth, maxWidth: billingProjectListWidth, boxShadow: '0 2px 5px 0 rgba(0,0,0,0.25)', height: `calc(100vh - ${topBarHeight}px)`, position: 'fixed', overflowY: 'scroll' } }, [
         div({
           style: {
             fontSize: 16, fontWeight: 600, padding: '2rem',
@@ -329,7 +328,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
           loadProjects()
         }
       }),
-      div({style: {minHeight: `calc(100vh - ${topBarHeight})`, flexGrow: 1 }}, [Utils.cond(
+      div({ style: { marginLeft: billingProjectListWidth, flexGrow: 1 } }, [Utils.cond(
         [selectedName && hasBillingProjects && !_.some({ projectName: selectedName }, billingProjects),
           () => div({ style: { margin: '1rem auto 0 auto' } }, [
             h2(['Error loading selected billing project.']),
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -26,7 +26,7 @@ import validate from 'validate.js'
 const ownerRole = 'Owner'
 
 const styles = {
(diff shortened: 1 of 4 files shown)


src/pages/billing/List.js
```diff
@@ -267,21 +287,37 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
         div({ style: Style.breadcrumb.textUnderBreadcrumb }, [selectedName])
       ])
     ]),
-    div({ role: 'main', style: { display: 'flex', flex: 1, position: 'relative' } }, [
-      div({ style: { width: 330, boxShadow: '0 2px 5px 0 rgba(0,0,0,0.25)' } }, [
-        div({ style: Style.navList.heading }, [
+    div({ role: 'main', style: { display: 'flex', flex: 1 } }, [
+      div({ style: { minWidth: billingProjectListWidth, maxWidth: billingProjectListWidth, boxShadow: '0 2px 5px 0 rgba(0,0,0,0.25)', height: `calc(100vh - ${topBarHeight}px)`, position: 'fixed', overflowY: 'scroll' } }, [
```
s-rubenstein: @panentheos this is the thing we were hoping to talk to you about. It seems like there should be a better way to make this 'fill up the total height of the screen, except the top bar, and scroll separately from the page' than this, but I couldn't figure it out.

gpolumbo-broad (approved):

src/pages/billing/List.js
```diff
@@ -267,21 +287,37 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
...
           'Billing Projects',
-          h(Clickable, {
+          h(ButtonPrimary, {
             'aria-label': 'Create new billing project',
-            onClick: showCreateProjectModal
+            onClick: showCreateProjectModal,
+            style: { marginTop: '1rem' }
           }, [
-            icon('plus-circle', { size: 21, style: { color: colors.accent() } })
+            'New Billing Project'
```
marctalbott: I can't see the mockup anymore (we hit some quota), but I'm guessing this is part of it? I prefer the plus circle button, but I leave it up to the experts

marctalbott (commented):
I'm not exactly sure where this is in the code otherwise I'd put it there, but the gap between the 'New Billing Project' button and the 'Owned by You' section seems too tall to me. I included screenshots of this and the existing to make it easier to compare
Screen Shot 2021-03-25 at 10 53 02 AM
Screen Shot 2021-03-25 at 10 53 11 AM

src/pages/billing/List.js
andy7i: yeah, exactly:

mock on left, local on right:

Screen Shot 2021-03-25 at 11 08 28 AM

marctalbott: 👍

Should 'Owned by You' and 'Shared with You' be bold like in the mock?

src/pages/billing/List.js
```diff
@@ -81,7 +81,7 @@ const BillingProjectSubheader = ({ title, children }) => h(Collapse, {
   title,
   initialOpenState: true,
   titleFirst: true,
-  buttonStyle: { padding: '1rem 1rem 1rem 2rem', color: colors.dark() }
+  buttonStyle: { padding: '1rem 1rem 1rem 2rem', color: colors.dark(), fontWeight: 'bold' }
```
andy7i, line 84: with fontWeight change:
Screen Shot 2021-03-25 at 11 15 11 AM

andy7i (approved):
added a fontWeight to match mock. not sure about what testing we need to do, but LGTM otherwise

src/pages/billing/List.js
andy7i: haha yeah I just added that

marctalbott (approved):

src/pages/billing/List.js
panentheos: Hmm, this is a tricky one. There are some assumptions in the top level wrappers that make this difficult. Not comprehensive, but a couple suggestions:
* The `position: fixed` is probably not necessary
* The righthand container also needs to be independently scrollable in this kind of layout
* Be sure to test in the BioDataCatalyst site skin, since that has a persistent footer that will interact with this.

s-rubenstein: Good calls, all around! `position: fixed` ended up not necessary, I think the righthand container being scrollable looks and feels better (otherwise it scrolled with the page, leaving the side list behind. The biodatacatalyst skin does have the issue of adding its own scrollbar because of the height issues. I am going to work around this by making a `pageHeightWithoutFooterOrHeader` function that takes that into account.

src/pages/billing/List.js
```diff
@@ -328,7 +332,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
           loadProjects()
         }
       }),
-      div({ style: { marginLeft: billingProjectListWidth, flexGrow: 1 } }, [Utils.cond(
+      div({ style: { flexGrow: 1, height: totalViewHeightMinusHeaderAndFooter, overflowY: 'scroll' } }, [Utils.cond(
```
gpolumbo-broad: I am not sure which change might have broken it, but the sidebar scrolling is now totally not working for me.  The whole page is scrolling off of the header.

gpolumbo-broad (changes requested):
In addition to this no longer being functional for me, there is one other thing that sort of fell off of our radar.

We had talked about possibly keeping the "New Billing Project" button and left nav heading "locked" at the top.  So in the left nav, when scrolling, you're only scrolling through the "Owned by you" and "Shared with you" lists.  People seemed on board with that idea before, just curious where people stand now.

gpolumbo-broad (approved):

PR #2419, opened 2021-03-25
[CA-1303] A11y: Landing Page, Top Menu and Sign-In Button

andrewmarcus:
This PR addresses accessibility challenges with the unauthenticated landing page that users first encounter.

## Moved GDPR Cookie Banner to the top

The way the banner overlapped the content at the bottom of the page is consistent with how many sites do it. However, it results in a sub-optimal user experience for mouse users, keyboard users, and screen reader users, each for different reasons. Since users must agree to accept cookies before they can use the site, I will argue that it really should be presented first.

## Fixed Google Sign-In button

Incomprehensibly, the Sign In with Google button, rendered by Google's SDK, is not accessible at all. It is a collection of `<div>` elements without any semantics indicating it's a button, which is not focusable and doesn't respond to the enter or spacebar keys. Fortunately, I found a way to wrap it into a semantic button element and pass down the keypress event to trigger the click.

## Added list roles to the main menu

Menu bars are often presented to screenreader users as bulleted lists. This PR adds some roles to make this work.

## Cleaned up Terra logo alt text for concision

The alt text for the Terra or brand logo at the top and bottom of the page included the word "logo". However, this is extraneous and unhelpful to a screen reader user. It is sufficient to just display the title itself.

## Added h1 and h2 to the landing page

The Terra UI has very few hierarchical headers, and any that it has are inconsistent. The landing page previously had on `<h2>` tag but no `<h1>`. This PR adds a primary header and some additional secondary headers in appropriate places t to the landing and sign-In pages (which use the same Hero component) without changing the styling or anything else. A header hierarchy should be added to other pages as well, but that will be handled in subsequent PRs.

## Defined fallback background colors for landing page regions

The aXe automated tool could not determine the background colors for the regions on the homepage because background images are used. This is a false-positive: in fact the color contrast is fine. 

To attempt to resolve this I added equivalent background colors to the two sections that were identified: the `main` section (hexagon background), and the Covid callout (virus illustration). This is good practice anyway as it constitutes a graceful fallback before the background images have finished loading, or if the user turns background images off. Unfortunately it did not resolve the aXe false positive. However, if you go into the Dev Console and disable the `background-image` rule in the 2 sections, then rerun aXe, it now will see the colors and stop complaining.


andrewmarcus:
The decision to move the GDPR cookie banner to the top of the page should probably get a review from the UX team. I'm happy to explain my reasoning for moving it to the top, but would also be happy to brainstorm other alternatives, if that would be preferred. The main thing is that whatever placement/markup is chosen needs to work for visual, keyboard-only and screen-reader users, which the previous placement did not.

src/components/CookieWarning.js
```diff
@@ -34,8 +34,9 @@ const CookieWarning = () => {
     signOut()
   }
   return !cookiesAccepted && div({
+    role: 'banner',
     style: {
-      position: 'fixed', height: 100, bottom: 0, width: '100%',
+      flex: 0, height: 100, width: '100%',
```
s-rubenstein, line 39: I ran this by UX, this is fine, but our UX team would want some other changes as well. Namely, the thought is:
Instead of `borderTop` on line 41, instead that should be `borderBottom`
For the background color, can we make it `colors.dark(0.15)`, just so it pops a bit more as being not part of the main site body?

src/components/SignInButton.js
```diff
@@ -14,7 +14,20 @@ const SignInButton = () => {
     })
   })
 
-  return div({ id: 'signInButton' })
+  // For some reason, Google's rendered Sign-In button is not at all keyboard accessible.
+  // To fix this, we wrap it as a button, and propagate the keyboard-accessible click event down to
+  // the inner DOM node inside the button, then let it bubble up to whatever it is that catches it.
```
s-rubenstein, line 20: Nice comment, thank you!

src/components/TopBar.js
```diff
@@ -55,6 +55,10 @@ const styles = {
     }),
     icon: {
       marginRight: 12, flex: 'none'
+    },
+    navSection: {
+      flex: 'none', height: 70, padding: '0 28px', fontWeight: 600,
+      borderTop: `1px solid ${colors.dark(0.55)}`, color: 'white'
```
s-rubenstein, line 61: So, I actually found an accessibility issue with `navSection` as well. If you go into the console, and type `configOverridesStore.set({isBioDataCatalyst: true})`, and then refresh the page (side note, you can use this to test out the different skins of the site), then you can see the background colors/rendering is different. I think for color here we need a ternary that detects which skin of the site we are using, similar to what you did for backgroundColor here: https://github.com/DataBiosphere/terra-ui/pull/2421/files#diff-c4ec3f6a5be9c7a24d3354c6aebf3544af26b150a4ce323fea36aa668cfd883cR323

src/pages/LandingPage.js
```diff
@@ -71,6 +71,7 @@ const LandingPage = () => {
     ]),
     isTerra() && div({
       style: {
+        backgroundColor: '#191a1c',
```
s-rubenstein: Can we make this use the color variables? (No worries if not, just wanted to check)

s-rubenstein (commented):
Hi! Found a few notes! Also, when running this locally, I found a couple lint errors, IDK if you have the linter set up, but if not, there are some instructions in the readme that should hopefully make it pretty easy to use.

src/components/CookieWarning.js
andrewmarcus: Sure! Both great suggestions

src/components/TopBar.js
andrewmarcus: Yep, makes sense! Thanks for the testing tip, I wasn't sure how to test those other code paths.

src/pages/LandingPage.js
andrewmarcus: The background is currently that virus image with a dark background on a gradient from left to right. I was trying to match the color of the left-most area so that the progressive enhancement would look reasonable for users who see the color before the image.  As far as I know there isn't a color name defined for that background color since this seems to be the only place using it.

Same thing with the other hexagon background image: I was trying to match the color of the image exactly so you don't see a line at the transition.

s-rubenstein: Fair enough! I am comfortable with that!

andrewmarcus: Tell you what: I'll add a comment explaining where that "magic" color code comes from. :)

src/components/TopBar.js
andrewmarcus: Ok, so I can get it into the BioCatalyst skin, but as far as I can tell the nav sections look fine to me. The background color of the menu is different, but it's still dark and the menu items in white text pop out with sufficient contrast.

I'm also seeing the same background image and the same not-quite-white background color still seems to match.

What are you seeing that needs to be fixed?

s-rubenstein: I attached two screenshots here of the top bar. Note the text that I highlight.:
Screen Shot 2021-04-08 at 2 05 02 PM
Screen Shot 2021-04-08 at 2 05 10 PM

andrewmarcus: What are the other skins I can try out?

andrewmarcus: Aha! Thanks

andrewmarcus: Ok, it looks like this will have to be fixed on each page that uses it. In the Workspaces section it's already styling the breadcrumbs properly with a ternary. I only see the problem on the Billing page. Anywhere else I should be looking?

s-rubenstein: If it is working in other pages (I'd try like 2 or 3), then we should be good to just fix this one. My guess is somewhere this one got missed. 

You can see all the skins that exist here: https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/colors.js#L21

However, the highest priority ones in my opinion are the default, and BioDataCatalyst. We ideally would get all of them, but BDC is the one that has the federal requirement.

andrewmarcus: Ok, I just pushed an update to the Billing page, which brings the way it renders breadcrumbs in line with the way it's done elsewhere, such as on the Workspaces pages, and ensures that the font color contrasts with the background. It increases the scope of this PR a little bit, but it's only a few lines of code

s-rubenstein (approved):
Okay, this looks good to me!

gpolumbo-broad (approved):

gpolumbo-broad:
The tests cannot run in @andrewmarcus's fork.  In order to get the tests to run before merging to `dev`, I am creating a new branch in in this repo that we can merge @andrewmarcus's code into and then we can merge from that branch into `dev`.

PR #2420, opened 2021-03-25
[PF-693] Add dropdown to select bucket location in Create Workspace dialog

wnojopra:
This adds a dropdown to the "Create a New Workspace" dialog for Bucket location. It defaults to the current default setting, US multi-regional.
image

I've tested that creating a workspace with the default location works, and a non-default location:

image
image

For full details on the motivation and discussion for this change, see [Terra Data Residency Components](https://docs.google.com/document/d/1--VH3sEGMVTzVvLu7suIaEvkdMgclS4kPLCpY2lgemE/edit?resourcekey=0-BdGkf63bItFtNRTeMixNJw#) doc.


aednichols (commented):
Just a drive-by observation, the red question mark in the second screenshot looks like an error.

If the system is working as intended, I don't think we should be alarming users with UI that looks like that.

wnojopra:
> Just a drive-by observation, the red question mark in the second screenshot looks like an error.
> 
> If the system is working as intended, I don't think we should be alarming users with UI that looks like that.

You can actually create a workspace with a Europe bucket today (using the API), and that is the what the UI shows. It is fixed by this PR: https://github.com/DataBiosphere/terra-ui/pull/2405/files (I'll actually need someone to merge that for me - I don't have permission to write to this repo).

aednichols (commented):
Gotcha, did not see that other PR.

The API today is undocumented so there is currently no supported path a user might take to make this happen; and I'm glad to see that as you add a supported path, you are making sure everything else continues to look good!

src/components/NewWorkspaceModal.js
```diff
@@ -30,6 +30,35 @@ const constraints = {
   }
 }
 
+const allRegions = [
```
cjllanwarne: This list probably belongs in https://github.com/DataBiosphere/terra-ui/blob/dev/src/components/region-common.js too

cjllanwarne:
You may want to add a warning that this sets the bucket location, but doesn't _necessarily_ change where compute runs.

Eg I think workflow submissions follow the workspace bucket location when setting default zones, but if a workflow has a `zones` runtime value hardcoded in then it won't be overridden. I'm also not sure Leonardo checks bucket location when it creates notebook clusters (cc @rtitle ?)

wnojopra:
> You may want to add a warning that this sets the bucket location, but doesn't _necessarily_ change where compute runs.
> 
> Eg I think workflow submissions follow the workspace bucket location when setting default zones, but if a workflow has a `zones` runtime value hardcoded in then it won't be overridden. I'm also not sure Leonardo checks bucket location when it creates notebook clusters (cc @rtitle ?)

I think that is fair. There are egress charges to think about if your workflow is run in a different region from the bucket. I'll figure out a good place to put warning/information about this.

On Leonardo, I am planning to submit two changes in the next few days:
1) terra-ui picks up the workspace bucket region and uses it for the createRuntime call
2) New Cluster Dialog allows researcher to set the location
But before these 2 happen Leonardo needs to support zone/region being passed in as a parameter. I am working with rtitle on this in https://github.com/DataBiosphere/leonardo/pull/1933

breilly2:
Notebook runtimes are currently per-project, not per-workspace. Until we move to project-per-workspace, one could still end up in a situation where their runtime is operating in a different region than the workspace bucket, which might come as a surprise to them.

src/components/region-common.js
```diff
@@ -26,3 +26,31 @@ export const regionInfo = (location, locationType) => {
       return { flag: unknownRegionFlag, regionDescription: `${locationType}: ${location}` }
   }
 }
+export const allRegions = [
```
mbookman, line 29: We have made a product decision that for now, we want to avoid creating within-country silos of life sciences community data (for example, data in `us-east` cannot be cross-analyzed with data in `us-central` without incurring egress charges). This means that we are going to be opinionated and pick specific regions in countries that have multiple regions. This comes to do US and Japan.

Thus we should remove from this list (and add a comment here) that `us-east*`, `us-west*`, and `asia-northeast2` are unavailable. We will put into documentation that if people want these regions, they should reach out to Terra support so that we can discuss the use cases.

petesantos:
Closing because this PR appears to be stale. If you need to finish this work, feel free to reopen.

PR #2421, opened 2021-03-26
[CA-1305] A11y: #/profile, #/clusters, #/test-login

andrewmarcus:
## Header hierarchy
* Added header hierarchy, placing the `<h1>` into the TopBar and setting each section header as an `<h2>`
* As I proceed through the various pages, I will continue to use the pattern of placing the `<h1>` into the TopBar if appropriate. There can only be one per page, so this seems like a sensible place to put it.
* Added `<h2>` to section headers in the `/clusters` page

## InfoBox accessibility
* Added aria-haspopup to InfoBox tooltips, to indicate to screenreader users what will happen when they click the button.

## Same link text, different behavior

This is the only issue identified by aXe on the `/profile` page. The problem, basically, is that for a user tabbing to the next link, the three links in a row all sound the same. The context is only given by the preceding line, and there was no semantic relationship between the two. To fix this, I made the provider names into `<h3>` tags and added differentiating text representing each provider to each link.

* Added a new short name for each provider. For now it includes just the org name (NCI, NHLBI, NHGRI) but could be something else if that would make more sense.
* Changed link text from "Log in to link your account" to "Log in to link your ${short} account" using the provider's short name
* Added an `aria-label` to the "Renew" and "Unlink" links providing the same short name. This adds context for screen readers but does not change the visual presentation of these two links side-by-side.

## Background color
* aXe reports that the top banner contrast is insufficient because it can't determine what color the background is due to the presence of background images. _This is a false positive._ I added a redundant fallback `background-color` property to try to help, but it doesn't seem to to do much.

## Notifications
* When error notifications pop, screenreaders are unaware of them. I added `role="alert"` to each notification to force screen readers to interrupt the user and read the notification text out loud. It will start with the image's aria-label: "error notification" and then proceed to read the title, text, details link and close button. This isn't perfect but at least it's better than not knowing the alerts exist. By default, `role="alert"` uses `aria-live="assertive"` which will interrupt whatever the user is doing.


src/libs/providers.js
```diff
@@ -1,5 +1,5 @@
 export const allProviders = [
-  { key: 'fence', name: 'NHLBI BioData Catalyst Framework Services' },
-  { key: 'dcf-fence', name: 'NCI CRDC Framework Services' },
-  { key: 'anvil', name: 'NHGRI AnVIL Data Commons Framework Services' }
+  { key: 'fence', name: 'NHLBI BioData Catalyst Framework Services', short: 'NHLBI' },
```
andrewmarcus: These could be changed to something else if it makes more sense...

src/components/TopBar.js
```diff
@@ -320,6 +320,7 @@ const TopBar = ({ showMenu = true, title, href, children }) => {
       role: 'banner',
       style: {
         ...styles.topBar,
+        backgroundColor: isTerra() ? colors.primary() : colors.ligth(), // Fallback color for a11y and if background images don't show
```
s-rubenstein: This is a typo `ligth` should be `light`?

src/libs/providers.js
s-rubenstein: I think this makes sense to me. We can change them down the road if it needs to change.

src/pages/Environments.js
```diff
@@ -141,7 +141,7 @@ const Environments = () => {
   return h(FooterWrapper, [
     h(TopBar, { title: 'Cloud Environments' }),
     div({ role: 'main', style: { padding: '1rem', flexGrow: 1 } }, [
-      div({ style: { ...Style.elements.sectionHeader, textTransform: 'uppercase', marginBottom: '1rem' } }, ['Your cloud environments']),
+      h2({ style: { ...Style.elements.sectionHeader, textTransform: 'uppercase', marginBottom: '1rem' } }, ['Your cloud environments']),
```
s-rubenstein: This change seems to have added some margin to the top, was that intentional? (I think it may just need an explicit margin set for marginTop to avoid inheriting from the header.)

s-rubenstein (commented):
Found a few discrepancies, but otherwise looks good!

src/components/TopBar.js
andrewmarcus: 🤦

src/pages/Environments.js
andrewmarcus: Right, true, I see that the heading style adds a bit of top and bottom padding which the original styling did not have.

If preserving the original spacing on this page is important, I'd be happy to 0 it out or go back to using a `div` and add `role="header"`. I would argue, however, that headings serve a visual purpose, and it's generally appropriate to offset them from the content a little bit.

andrewmarcus: Looks like the margins were being set explicitly before anyway. Just pushed an update to zero out any padding, as well at the top margin for the first h2 (the second h2 still has margins both top and bottom)

s-rubenstein: Thank you! I am not super worried about it (and even prefer a little more padding), but if we don't need it for accessibility, I would rather not make display changes without confirming with the UX team.

s-rubenstein (approved):
LGTM! Thanks for doing this, these are some great improvements!

src/pages/Environments.js
andrewmarcus: Sure, makes senes to me :)

src/components/TopBar.js
```diff
@@ -349,7 +350,9 @@ const TopBar = ({ showMenu = true, title, href, children }) => {
           div({
             style: title ? { fontSize: '0.8rem', lineHeight: '19px' } : { fontSize: '1rem', fontWeight: 600 }
           }, [versionTag('Beta')]),
-          title
+          title && h1({
+            style: { fontSize: '1em', fontWeight: 500, padding: 0, margin: 0 }
```
andy7i, line 378: just curious, why did you have to add in the style here?

src/libs/providers.js
```diff
@@ -1,5 +1,5 @@
 export const allProviders = [
-  { key: 'fence', name: 'NHLBI BioData Catalyst Framework Services', expiresAfter: 30 },
-  { key: 'dcf-fence', name: 'NCI CRDC Framework Services', expiresAfter: 15 },
-  { key: 'anvil', name: 'NHGRI AnVIL Data Commons Framework Services', expiresAfter: 30 }
+  { key: 'fence', name: 'NHLBI BioData Catalyst Framework Services', expiresAfter: 30, short: 'NHLBI' },
+  { key: 'dcf-fence', name: 'NCI CRDC Framework Services', expiresAfter: 15, short: 'NCI' },
+  { key: 'anvil', name: 'NHGRI AnVIL Data Commons Framework Services', expiresAfter: 30, short: 'NHGRI' }
```
andy7i, line 4: a new provider was added (or will be added) -- just a heads up for when this is rebased

src/pages/Profile.js
```diff
@@ -354,11 +354,15 @@ const Profile = ({ queryParams = {} }) => {
 
             sectionTitle('Program Info'),
 
-            div({ style: styles.form.title }, ['Non-Profit Status']),
-            div({ style: { margin: '1rem' } }, [
-              radioButton('nonProfitStatus', 'Profit'),
-              radioButton('nonProfitStatus', 'Non-Profit')
-            ]),
+            h(IdContainer, [id => div({
+              style: styles.form.title, role: 'radiogroup', 'aria-labelledby': id
```
andy7i: nitpick: I think this is applying`styles.form.title` on the radio button text as well
Screen Shot 2021-04-29 at 9 35 19 PM

src/components/TopBar.js
andy7i: oh, this is why
Screen Shot 2021-04-29 at 9 31 06 PM

andy7i (commented):
1 nitpick and then I think it should be good to go! Do you mind rebasing the branch?

src/components/TopBar.js
andrewmarcus: Yep! I specifically tried to preserve the formatting that was there before :)

andrewmarcus:
@andy7i If I Rebase and then Force Push, will Github preserve this PR or will I have to open a new one?

src/pages/Profile.js
andrewmarcus: Fixed!

andy7i (approved):
LGTM!

PR #2423, opened 2021-03-26
Warning about duplicate analyses of constant inputs [BW-599]

cjllanwarne:
Updates the launch modal to warn users if they're about to launch multiple workflows over the same constant set of input data.

This is certainly not a panacea for avoiding work-duplicating mistakes, but this little triangle would have saved at least one user in the region of $9000 recently due to an accidental duplicate analysis over 180 hard-coded input sets.

Example on dev: [link](http://localhost:3000/#workspaces/general-dev-billing-account/cjl_test_workspace/workflows/cjl/echo_input_string). Try switching the input type between inputs from the data table and "defined by file paths", and switching the input attribute from a constant string value to a participant ID to see how the modal changes in each case.

Example screenshot of the new warning message [updated 3/31]:

image

### DIFF
--- a/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
+++ b/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
@@ -2,6 +2,7 @@ import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
 import { b, div, h, p, span, wbr } from 'react-hyperscript-helpers'
 import { ButtonPrimary, CromwellVersionLink } from 'src/components/common'
+import { warningBoxStyle } from 'src/components/data/data-utils'
 import { icon, spinner } from 'src/components/icons'
 import Modal from 'src/components/Modal'
 import { InfoBox } from 'src/components/PopupTrigger'
@@ -11,8 +12,6 @@ import { launch } from 'src/libs/analysis'
 import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
 import * as Utils from 'src/libs/utils'
-import {warningBoxStyle} from 'src/components/data/data-utils'
-
 import {
   chooseRows, chooseSetComponents, chooseSets, processAll, processAllAsSet, processMergedSet, processSnapshotTable
 } from 'src/pages/workspaces/workspace/workflows/EntitySelectionType'
--- a/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
+++ b/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
@@ -93,13 +93,6 @@ const LaunchAnalysisModal = ({
   const onlyConstantInputs = _.every(i => !i || Utils.maybeParseJSON(i) !== undefined, config.inputs)
   const warnDuplicateAnalyses = onlyConstantInputs && entityCount > 1
 
-  const fullWidthWarningStyle = {
-    ...warningBoxStyle,
-    borderLeft: 'none', borderRight: 'none',
-    margin: '0 -1.25rem',
-    fontSize: 14
-  }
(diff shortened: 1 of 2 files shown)


ldgauthier:
Awesome, I love it!  This could be a huge help for new users.  Minor quibble -- can we make the word duplicate bold or underlined or something?  It looks very similar to the message I've seen when launching a big bunch of different workflows.

cjllanwarne:
Thanks @ldgauthier, I've updated the screenshot with a more emphatic "duplicate"

ldgauthier:
Perfect.

salonishah11 (approved):
💯

cjllanwarne:
Tested locally using: http://localhost:3000/#workspaces/general-dev-billing-account/cjl_test_workspace/workflows/cjl/possibly_parameterized_inputs

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -114,8 +125,14 @@ const LaunchAnalysisModal = ({
...
-    div({ style: { margin: '1rem 0' } }, [
-      'This will launch ', b([entityCount]), ` analys${entityCount === 1 ? 'is' : 'es'}`,
+    div({ style: { margin: '1rem 0', ...maybeWarningStyle } }, [
+      warnDuplicateAnalyses && icon('warning-standard', { size: 16, style: { color: colors.warning(), marginRight: '0.5rem' } }),
+      warnDuplicateAnalyses && 'Warning! ',
+      'This will launch ',
+      b([entityCount]),
+      warnDuplicateAnalyses && span({ style: { fontWeight: 'bold', textDecoration: 'underline' } }, [' duplicate']),
+      entityCount === 1 ? ' analysis' : ' analyses',
+      warnDuplicateAnalyses && ' of the same set of constant inputs',
```
panentheos: nit: Rather than doing word-by-word conditions, It might be cleaner to just have two separate branches, even if there's a little duplication in the content.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -114,13 +125,21 @@ const LaunchAnalysisModal = ({
...
-      type === processMergedSet && entityCount !== mergeSets(selectedEntities).length && div({
-        style: { fontStyle: 'italic', marginTop: '0.5rem' }
-      }, ['(Duplicate entities are only processed once.)'])
+    warnDuplicateAnalyses ? div({ style: { margin: '1rem 0', ...fullWidthWarningStyle } }, [
+      icon('warning-standard', { size: 16, style: { color: colors.warning(), marginRight: '0.5rem' } }),
+      'Warning! This will launch ',
+      b([entityCount]),
+      span({ style: { fontWeight: 'bold', textDecoration: 'underline' } }, [' duplicate']),
+      entityCount === 1 ? ' analysis' : ' analyses',
+      warnDuplicateAnalyses && ' of the same set of constant inputs.'
```
panentheos: Now that we have two code paths, these conditionals aren't needed, so this could be:
```suggestion
      ' analyses of the same set of constant inputs.'
```

panentheos (approved):
See comment, then looks good. Tested locally.

TOL: We might consider taking the opportunity to provide a more verbose explanation, with the aim of teaching about the system a bit. Something like "This will launch N analyses, but all the inputs are constant, so this will likely result in duplicated work." UX might have some ideas here.

cjllanwarne:
@panentheos and @salonishah11 heads up that I refactored this a little bit following feedback from Joy, please see the updated example image and you may want to glance at the most recent code changes again.

salonishah11 (approved):
LGTM 👍

PR #2424, opened 2021-03-30
standardize and improve UX around workflow entity selection [BW-492] [SATURN-1322]

panentheos:
This change addresses two long-standing issues:
1. The default behavior when launching a workflow using 'base' entity types (e.g. `sample`) is to run against ALL entities. This is almost never what users want, and it's easy to accidentally spend lots of money.
2. Due to historical reasons, the UX for launching with 'set' types (e.g. `sample_set`) behaves differently, even though there is no longer any need for the difference.

This change standardizes the UX and language across the various entity types, and defaults to an empty selection in all cases. That means users will need to explicitly choose what to run against, which should eliminate the accidental 'run against all' case. Note that it's still possible to run against all entities by explicitly selecting them (via the popup menu in the upper left of the selection table), in case someone actually wants to do that.

Note that the language can be a bit tricky, since we now support arbitrary levels of 'set' nesting, so it's important that we communicate that clearly. Language feedback is definitely welcome.


src/pages/workspaces/workspace/workflows/DataStepContent.js
```diff
@@ -109,66 +90,36 @@ export default class DataStepContent extends Component {
...
-                checked: isProcessMergedSet,
-                onChange: () => this.setNewSelectionModel({ type: processMergedSet, selectedEntities: {} }),
-                labelStyle: { marginLeft: '0.75rem' }
-              })
-            ])
-          ])]),
-        !isProcessAll && !isProcessAllAsSet && div({
+        div({ role: 'radiogroup', 'aria-label': 'Select entities' }, [
+          isSet && div([
+            h(RadioButton, {
```
rsasch, line 95: Since these are radio buttons, is there a scenario where the user will be choosing from two or more options (in testing, I was not able to come up with one)?

integration-tests/tests/run-workflow.js
```diff
@@ -33,7 +33,7 @@ const testRunWorkflowFn = _.flow(
 
   await waitForNoSpinners(page)
   await click(page, clickable({ text: 'Select Data' }))
-  await click(page, input({ labelContains: 'Choose specific rows to process' }))
+  await click(page, input({ labelContains: 'Choose specific test_entitys to process' }))
```
cjllanwarne, line 36: I think I see where you were going with this, but it scanned kind of weirdly as I was reading it:

```suggestion
  await click(page, input({ labelContains: 'Choose specific test_entities to process' }))
```

src/pages/workspaces/workspace/workflows/DataStepContent.js
```diff
@@ -44,16 +42,13 @@ export default class DataStepContent extends Component {
   }
 
   isValidSelectionModel() {
-    const { newSelectionModel: { newSetName, selectedEntities, type } } = this.state
-    const selectionSize = _.size(selectedEntities)
+    const { newSelectionModel: { newSetName, selectedEntities } } = this.state
+    return _.size(selectedEntities) > 0 && (!this.willCreateSet() || newSetName)
```
cjllanwarne: Nit, but I found this logic easier to parse this way round:

```suggestion
    return _.size(selectedEntities) > 0 && (newSetName || !this.willCreateSet())
```

cjllanwarne: Also - is this going to stop me if I choose to name my new set `"0"`? (perhaps the start of a sequence of sets...?)

src/pages/workspaces/workspace/workflows/DataStepContent.js
```diff
@@ -109,66 +90,36 @@ export default class DataStepContent extends Component {
...
-                onChange: () => this.setNewSelectionModel({ type: processMergedSet, selectedEntities: {} }),
-                labelStyle: { marginLeft: '0.75rem' }
-              })
-            ])
-          ])]),
-        !isProcessAll && !isProcessAllAsSet && div({
+        div({ role: 'radiogroup', 'aria-label': 'Select entities' }, [
+          isSet && div([
+            h(RadioButton, {
+              text: `Create a new ${rootEntityType} from selected ${baseEntityType}s`,
```
cjllanwarne, line 96: Thinking out loud, I'm getting a little lost in the distinction between `rootEntityType` and `baseEntityType` here. Eg - it's not totally obvious to me which is lower, the root or the base.

My intuition would be that "root" would be below base", but that _seems_ like the opposite of what's happening here?

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -80,11 +68,9 @@ const LaunchAnalysisModal = ({
...
     [processSingle, () => 1],
     [type === processSnapshotTable, () => entityMetadata[rootEntityType].count],
-    [type === chooseRows || type === chooseSets, () => _.size(selectedEntities)],
-    [type === processAll, () => entityMetadata[rootEntityType].count],
-    [type === processAllAsSet, () => 1],
-    [type === chooseSetComponents, () => 1],
-    [type === processMergedSet, () => _.flow(mergeSets, _.uniqBy('entityName'))(selectedEntities).length]
+    [type === chooseRootType, () => _.size(selectedEntities)],
+    [type === chooseBaseType, () => 1],
+    [type === chooseSetType, () => _.flow(mergeSets, _.uniqBy('entityName'))(selectedEntities).length]
```
cjllanwarne, line 73: does this `cond` need a base case?

cjllanwarne:
Note: I need to try this out for myself. Any advice (or RTFM links!) on how to set up a suitable data model to see these various hierarchy layers would be really appreciated.

integration-tests/tests/run-workflow.js
panentheos: Unfortunately, since we pluralize by adding an 's', this is in fact what shows up when the entity name is 'test_entity'. There are more sophisticated pluralization strategies out there, but it's not clear that they would always be better. E.g., should 'sample_from_goose' pluralize to 'sample_from_geese'? We could also say 'Choose specific test_entity rows to process', but that might be a bit clunky.

src/pages/workspaces/workspace/workflows/DataStepContent.js
panentheos: No, the string `'0'` is truthy. If we're dealing with numbers, that's when we need to worry about `0`, which is falsy.

src/pages/workspaces/workspace/workflows/DataStepContent.js
panentheos: Yes, you can get up to three options at once (e.g. `sample`, `sample_set`, `sample_set_set`), but only if those types exist in your data model.

src/pages/workspaces/workspace/workflows/DataStepContent.js
panentheos: Yeah, I wrestled with that a bit. The name `rootEntityType` is somewhat fixed, because that's what the method config calls it, and I think it's important to keep that the same. I considered 'subEntityType' but that seemed too obtuse.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
panentheos: A default? It shouldn't, since we're covering all available options.

rsasch (approved):

src/pages/workspaces/workspace/workflows/DataStepContent.js
```diff
@@ -177,31 +128,30 @@ export default class DataStepContent extends Component {
...
             childrenBefore: () => div({ style: Style.elements.sectionHeader }, [
-              Utils.cond(
-                [isProcessMergedSet, () => 'Select one or more sets to combine and process'],
-                [isChooseRows, () => `Select ${rootEntityType}s to process`],
-                [isChooseSetComponents, () => `Select ${baseEntityType}s to create a new set to process`],
-                [isChooseSets, () => `Select ${rootEntityType}s to process`]
+              Utils.switchCase(type,
+                [chooseSetType, () => `Select one or more ${setType}s to combine and process`],
+                [chooseRootType, () => `Select ${rootEntityType}s to process`],
+                [chooseBaseType, () => `Select ${baseEntityType}s to create a new set to process`]
```
cjllanwarne: Can we say what _type_ of set here?

src/pages/workspaces/workspace/workflows/DataStepContent.js
```diff
@@ -177,31 +128,30 @@ export default class DataStepContent extends Component {
...
           [id => div({ style: { marginTop: '1rem' } }, [
             h(FormLabel, { htmlFor: id }, [
-              Utils.cond(
-                [isProcessAll, () => `All ${rootEntityTypeCount} ${rootEntityType}s will be saved as a new set named:`],
-                [isProcessAllAsSet, () => `All ${baseEntityTypeCount} ${baseEntityType}s will be saved as a new set named:`],
-                [isProcessMergedSet, () => `Selected ${rootEntityType}s will have their membership combined into a new set named:`],
-                [isChooseSetComponents, () => `Selected ${baseEntityType}s will be saved as a new set named:`],
-                [isChooseSets, () => `Selected ${rootEntityType}s will be saved as a new set named:`],
-                [isChooseRows, () => `Selected ${rootEntityType}s will be saved as a new set named:`]
+              Utils.switchCase(type,
```
cjllanwarne, line 151: Can we say what _type_ of set here?

src/pages/workspaces/workspace/workflows/DataStepContent.js
```diff
@@ -177,31 +128,30 @@ export default class DataStepContent extends Component {
...
+                [chooseSetType, () => `Selected ${setType}s will have their membership combined into a new set named:`],
```
cjllanwarne: Can we say what _type_ of set here?

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -554,21 +548,17 @@ const WorkflowView = _.flow(
   }
 
   describeSelectionModel() {
-    const { modifiedConfig: { rootEntityType }, entityMetadata, entitySelectionModel: { newSetName, selectedEntities, type } } = this.state
+    const { modifiedConfig: { rootEntityType }, entitySelectionModel: { newSetName, selectedEntities, type } } = this.state
     const count = _.size(selectedEntities)
-    const newSetMessage = (type === processAll || type === processAllAsSet ||
-      (type === chooseSetComponents && count > 0) || count > 1) ? `(will create a new set named "${newSetName}")` : ''
+    const newSetMessage = type === chooseBaseType || count > 1 ? `(will create a new set named "${newSetName}")` : ''
```
cjllanwarne: Can we say what _type_ of set here?

cjllanwarne (approved):
I have a few "can we give this set a name" requests after manual testing by clicking around, otherwise LGTM 👍  (I may have been overzealous in my string matching for "new set" messages, so feel free to use your judgement here)

cjllanwarne (approved):

PR #2425, opened 2021-03-30
add mixpanel event for workspace:open, fix event for share workspace [AS-517]

Shakespeared:
- Adds a mixpanel event for opening a workspace, and making sure the workspace name and workspace namespace are tracked on other workspace events. 

- I think I caught all the events that need those properties, if you think they ought to be on anything else let me know. Most of them were already handled by the time I picked up the ticket. 

- Discovered that the share workspace event wasn't tracking those properly so I fixed that too. (Left it in the same PR because it's itty bitty)

Tested in the UI and verified seeing the properties in mixpanel


src/pages/workspaces/List.js
```diff
@@ -178,13 +179,17 @@ export const WorkspaceList = () => {
...
 
             return div({ style: styles.tableCellContainer }, [
               div({ style: styles.tableCellContent }, [
                 h(Link, {
                   style: { color: canView ? undefined : colors.dark(0.7), fontWeight: 600, fontSize: 16, ...Style.noWrapEllipsis },
                   href: canView ? Nav.getLink('workspace-dashboard', { namespace, name }) : undefined,
-                  onClick: !canView ? () => setRequestingAccessWorkspaceId(workspaceId) : undefined,
+                  onClick: () => {
+                    canAccessWorkspace()
+                    !!canView && Ajax().Metrics.captureEvent(Events.workspaceOpen, { workspaceName: name, workspaceNamespace: namespace })
```
breilly2: We already have page view events on workspace dashboard. Do we need this as well?

davidangb (commented):
I see the namespace/name in the share event now, that is good!

Brian's question about `page:view:workspace-dashboard` vs. `workspace:open` is a good one. If we decide to leave `workspace:open` in place, another issue: how else can users open a workspace, beyond the workspace list? For instance, coming from the showcase, or importing a workflow into a workspace? The `workspace:open` event would miss those.

src/pages/workspaces/List.js
Shakespeared: @breilly2 My thought with adding workspace:open was that it might be a better proxy for tracking unique opens, since the dashboard event will trigger if you're switching tabs after you've already opened the workspace. But I could also change tack and just throw the name and namespace on that event instead if that seems better, that's also an option. 

@davidangb Agreed that if we stick with workspace:open I'll need to track down other ways a user can open a workspace, that's a good callout.

Shakespeared: @breilly2 @davidangb I talked to Amelia and she asked that I add name and namespace tracking to the page:view events on workspaces, which I've done with @zarsky-broad 's help. She said that we may as well leave the other event to track opens from the list, as it might come in handy some time and it's already there. I think that should cover all the bases!

davidangb (approved):
lgtm, thank you!

src/pages/workspaces/List.js
breilly2: I really thought workspace namespace and name were already there. 🤦 🤷 
I'm fine with keeping the new event as well, as long as its name reflects what it is to avoid it from being misinterpreted. Adding details to the Mixpanel lexicon may help with that, and is probably good practice to do anyway.

Shakespeared: Yeah me too originally… 😬 But it's there now! Yep yep I renamed the new event to workspace:open:from:list (unless you can think of a more concise way to say that) to avoid confusion.

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -95,7 +95,7 @@ const ShareWorkspaceModal = ({ onDismiss, workspace, workspace: { workspace: { n
     const aclEmails = _.map('email', acl)
     const needsDelete = _.remove(entry => aclEmails.includes(entry.email), originalAcl)
     const numAdditions = _.filter(({ email }) => !_.some({ email }, originalAcl), acl).length
-    const eventData = { numAdditions, ...extractWorkspaceDetails(workspace) }
+    const eventData = { numAdditions, workspaceName: name, workspaceNamespace: namespace }
```
breilly2, line 98: Oh, yikes. `workspace.workspace` strikes again.

src/libs/events.js
```diff
@@ -45,11 +46,16 @@ export const extractCrossWorkspaceDetails = (fromWorkspace, toWorkspace) => {
 }
 
 export const PageViewReporter = () => {
-  const { name } = useRoute()
+  const { name, params } = useRoute()
 
   useEffect(() => {
-    Ajax().Metrics.captureEvent(`${eventsList.pageView}:${name}`)
-  }, [name])
+    const isWorkspace = /^#workspaces\/.+\/.+/.test(window.location.hash)
```
breilly2, line 55: TOL: This duplicates (and decentralizes) the `path` properties of numerous `navPaths`. I guess this coincidence already existed, but I don't think we're assuming that anywhere. We've been allowing each view to be in complete control of its URL path without affecting any other site functionality. 🤔

src/libs/events.js
```diff
@@ -45,11 +46,16 @@ export const extractCrossWorkspaceDetails = (fromWorkspace, toWorkspace) => {
...
+
+    Ajax().Metrics.captureEvent(
+      `${eventsList.pageView}:${name}`,
+      isWorkspace ? extractWorkspaceDetails(params) : undefined
```
breilly2, line 59: TOL: `params` isn't really a workspace object. It does happen to have `namespace` and `name` properties, though, so this technically works.

I wonder if we should just always record the URL params that the router pulled out, without any transformation. I think this could be done by adding something like:
```
  appPathParams: Nav.getCurrentRoute().params
```
to the properties here: http://github.com/DataBiosphere/saturn-ui/blob/42443921ba63ce5244a3556e54072ea50946bc14/src/libs/ajax.js#L1347-L1347 (with `Nav.getCurrentRoute()` extracted as a local `const`).

One downside of this is that it won't match how namespace/name are recorded in Mixpanel for events that are using `extractWorkspaceDetails`, but the long-term win of capturing the path parameters might be worth it.

Also, I was a little surprised to find that the `appPath` property we're recording is just the handler's name, not something more like the URL path. Maybe we were worried about privacy? Anyway, I suspect this is where we thought we were recording namespace/name in Mixpanel... but apparently not.

src/libs/events.js
```diff
@@ -45,11 +46,16 @@ export const extractCrossWorkspaceDetails = (fromWorkspace, toWorkspace) => {
...
+    )
+  }, [name]) // eslint-disable-line react-hooks/exhaustive-deps
```
breilly2: I think `params` should be included here. Since `name` is just something like `workspace-dashboard`, not including `params` would mean that we might miss events (which may actually be a bug today!).

breilly2 (commented):
One bug (I think) and some comments about things worth thinking about.

MatthewBemis (approved):
I don't seem to have a Mixpanel account so I can't do any testing of my own, but the code LGTM

PR #2427, opened 2021-03-31
[SATURN-1892] Uploader bugfixes and performance improvements

andrewmarcus:
## Apr 2 update

* I switched the reload buffering in the FileBrowser to 5 seconds (rather than 10), to be consistent with the comment and also because it provides the user with slight more real-time feedback. This still only matters when uploading lots of small files; if a big file takes longer than 5 seconds, it will still only refresh once the file is fully uploaded.
* I switched the FileBrowser component to use a `FlexTable` rather than a `GridTable`, which is better suited for this use case.
* Both `FlexTable` and `GridTable` now accept optional parameters to change the row heights and override the header and cell styles. This allowed me to condense the rows and remove borders, to get the file browser to more closely match the way it looks in the Data tab in Workspaces. FWIW, that one should probably be updated to use a FlexTable too, or probably just reuse the new FileBrowser component, but I figured it would be better not to make that change in this PR.
* I updated the way the Metadata preview table is rendered to show more clearly which columns will be updated versus added, when updating an existing table. This provides a better user experience while dedicating more screen real-estate to the table, better supporting smaller displays.

## Original

This PR switches the table of uploaded files from a `SimpleTable` to a `GridTable`, so that only the visible rows are rendered into the DOM. This appears to resolve the performance issues Joel encountered when labs uploaded hundreds of thousands of files.

In order to make this look good, I added some new styling options to GridTable. You can now set the desired height of each header row and data row, as well as apply additional styling options to the header. This makes the file table in the uploader look much more like the one in the Data tab within Workspaces.  I was unable to get the delete button to hide except on hover, but this does not seem to be important enough to spend time on.

To further improve performance, I reduced the frequency with which the file table is update while files are uploading. Previously, every time a file finished the table would refresh so that users can see the new file. However, this is rather meaningless when uploading hundreds of files at a time since the table only renders a few of them, and results in unnecessary work. I've debounced it so the table will refresh no more often than once every 5 seconds.

In addition, I fixed the bug which was causing only the first thousand filenames in a metadata file to be translated to their bucket URLs. We will now grab as many pages of 1000 URLs as we can before attempting to do the matching.

On the other, the one caveat of the file table is that it only displays the first 1000 files in the bucket. If it's important, I can figure out how to add a pager, or do automatic paging like I do to count the URLs. But I'm not sure whether it's necessary for most users.

### Metadata Uploader improvements

To improve the user experience, I reduced the vertical size of the Metadata Upload panel so that more of the metadata table is visible.  Further improvements could be made here, such as listing the columns to be updated in a row rather than a bulleted list. But this is better than it was.

I also fixed a bug preventing table columns from being sorted properly. This is working now.

### DIFF
--- a/src/components/ProgressBar.js
+++ b/src/components/ProgressBar.js
@@ -1,8 +1,8 @@
+import filesize from 'filesize'
 import { dd, div, dl, dt, h, p, strong } from 'react-hyperscript-helpers'
 import { ButtonPrimary } from 'src/components/common'
 import Modal from 'src/components/Modal'
 import colors from 'src/libs/colors'
-import { friendlyFileSize } from 'src/libs/uploads'
 
 
 export const ProgressBar = ({ max, now }) => {
@@ -77,7 +77,7 @@ export const UploadProgressModal = ({ status: { totalFiles, totalBytes, uploaded
       dt(['Size:']),
       dd({
         style: { margin: '0.4rem 0 1rem 0', fontWeight: 600 }
-      }, [friendlyFileSize(currentFile.size)])
+      }, [filesize(currentFile.size, { round: 1 })])
     ]),
     h(ProgressBar, {
       max: totalBytes,
@@ -85,9 +85,9 @@ export const UploadProgressModal = ({ status: { totalFiles, totalBytes, uploaded
     }),
     p({}, [
       'Transferred ',
-      strong([friendlyFileSize(uploadedBytes)]),
+      strong([filesize(uploadedBytes, { round: 1 })]),
       ' of ',
-      strong([friendlyFileSize(totalBytes)]),
+      strong([filesize(totalBytes, { round: 1 })]),
       ' ',
       strong(['(', (uploadedBytes / totalBytes * 100).toFixed(0), '%)'])
     ])
--- a/src/components/data/FileBrowser.js
+++ b/src/components/data/FileBrowser.js
@@ -1,15 +1,17 @@
 import filesize from 'filesize'
 import _ from 'lodash/fp'
-import { Fragment, useEffect, useState } from 'react'
+import { Fragment, useEffect, useRef, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
+import { AutoSizer } from 'react-virtualized'
 import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
 import { Link, topSpinnerOverlay } from 'src/components/common'
+import { saveScroll } from 'src/components/data/data-utils'
 import Dropzone from 'src/components/Dropzone'
 import FloatingActionButton from 'src/components/FloatingActionButton'
 import { icon } from 'src/components/icons'
 import { NameModal } from 'src/components/NameModal'
 import { UploadProgressModal } from 'src/components/ProgressBar'
-import { HeaderCell, SimpleTable, TextCell } from 'src/components/table'
+import { GridTable, HeaderCell, TextCell } from 'src/components/table'
 import UriViewer from 'src/components/UriViewer'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
@@ -23,10 +25,10 @@ import { DeleteObjectModal } from 'src/pages/workspaces/workspace/Data'
 export const FileBrowserPanel = _.flow(
   Utils.withDisplayName('DataUploadPanel'),
   requesterPaysWrapper({ onDismiss: ({ onClose }) => onClose() })
-)(({ workspace, workspace: { workspace: { namespace, bucketName } }, onRequesterPaysError, basePrefix, setNumFiles, collection, allowNewFolders = true, children }) => {
+)(({ workspace, workspace: { workspace: { namespace, bucketName } }, onRequesterPaysError, basePrefix, setNumFiles, collection, allowNewFolders = true, style, children }) => {
   const [prefix, setPrefix] = useState('')
-  const [prefixes, setPrefixes] = useState(undefined)
-  const [objects, setObjects] = useState(undefined)
+  const [prefixes, setPrefixes] = useState([])
+  const [objects, setObjects] = useState([])
   const [loading, setLoading] = useState(false)
   const [deletingName, setDeletingName] = useState(undefined)
   const [viewingName, setViewingName] = useState(undefined)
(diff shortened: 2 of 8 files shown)


src/components/ProgressBar.js
```diff
@@ -1,8 +1,8 @@
+import filesize from 'filesize'
 import { dd, div, dl, dt, h, p, strong } from 'react-hyperscript-helpers'
 import { ButtonPrimary } from 'src/components/common'
 import Modal from 'src/components/Modal'
 import colors from 'src/libs/colors'
-import { friendlyFileSize } from 'src/libs/uploads'
```
andrewmarcus, line 5: Turns out I wrote an unnecessary function, so I'm removing it and falling back on the one which already exists and is used everywhere else

src/components/data/FileBrowser.js
```diff
@@ -164,57 +179,109 @@ export const FileBrowserPanel = _.flow(
...
+                      icon('trash', {
+                        size: 16,
+                        className: 'hover-only'
+                      })
+                    ])
+                  }
+                }
+              },
+              {
+                width: width - 400, // Fill the remaining space
```
panentheos: Just a preliminary note: If this table isn't intended to scroll horizontally, the `FlexTable` component might be a better fit, since it will give you virtualized rows and you can specify flexbox-like constraints for the column sizes.

andrewmarcus: Ah, thanks!  Yeah, I saw that one but hadn't looked much into it.

Happy to switch to that one before this PR gets merged, or come back and fix it later. Whichever you'd prefer...

src/libs/ajax.js
```diff
@@ -939,8 +939,9 @@ const Buckets = signal => ({
     const items = body.items || []
 
     // Get the next page recursively if there is one
-    if (res.nextPageToken) {
-      return _.concat(items, await Buckets(signal).listAll(namespace, bucket, prefix, res.nextPageToken))
+    if (body.nextPageToken) {
+      const next = await Buckets(signal).listAll(namespace, bucket, prefix, body.nextPageToken)
+      return items.concat(next)
```
breilly2: Nit: May as well keep using `_.concat` here for consistency.

src/components/data/UploadPreviewTable.js
```diff
@@ -98,7 +98,7 @@ const UploadDataTable = props => {
   }, [])
 
   const sortedRows = useMemo(() => {
-    const i = _.findIndex(columns, sort.field)
+    const i = columns ? columns.indexOf(sort.field) : -1
```
breilly2, line 113: Nit: `_.findIndex` should work fine here, again for consistency.

src/components/data/FileBrowser.js
```diff
@@ -73,19 +79,26 @@ export const FileBrowserPanel = _.flow(
...
-  }, [prefix, uploadStatus]) // eslint-disable-line react-hooks/exhaustive-deps
+  }, [prefix, lastRefresh]) // eslint-disable-line react-hooks/exhaustive-deps
 
   useEffect(() => {
-    // Count all the files in the bucket any time the uploader completes
     if (!uploadStatus.active) {
+      // If the uploader has completed, count all the files in the bucket
       count()
+    } else {
+      // While the uploader is working, refresh the table no more often than every 5 seconds
```
breilly2, line 91: In the PR description (and the code below), you used 10 seconds. IMO, 5 is probably enough of a rate limit, and provides more immediate feedback, but as long as the code and this comment are consistent, it doesn't really matter to me.

src/components/data/FileBrowser.js
```diff
@@ -73,19 +79,26 @@ export const FileBrowserPanel = _.flow(
...
+      const now = Date.now()
+      if (lastRefresh < now - 10e3) {
```
breilly2: TIL about using exponents in numeric literals!

src/components/data/FileBrowser.js
breilly2: Andrew, I'm going to leave it up to you. As long as the current code meets your need, I think it's fine to go ahead with it and refactor later.

breilly2 (approved):
👍  Looks good. A couple of comments, but nothing blocking.

I mostly studied the shared code. I looked over the uploader code, but I'm not familiar with the feature and didn't try using it. Andrew, I'll trust whatever testing you did yourself. Let me know if you want more attention to testing in this review.

I did run the integration tests locally and they passed.

src/components/data/FileBrowser.js
andrewmarcus: I just pushed an update to switch to a `FlexTable`. You're right, the conversion was trivial and it's definitely a cleaner implementation for this use case.

I added `rowHeight` and `headerHeight` parameters, as well as `styleCell()` and `styleHeader()` functions, in order to attempt to match the styling of File Browsers elsewhere on the site. Since these are all optional, their presence should be a no-op for anything else using these tables.

src/components/data/UploadPreviewTable.js
andrewmarcus: Wasn't working for some reason. Might have just been that I got the lodash-fp argument order wrong, but switching to .indexOf fixed it

zarsky-broad: Yup, we're using lodash's functional variant, so the array comes last.

src/components/data/FileBrowser.js
andrewmarcus: Yeah, it's a nice trick to make time durations easier to read when you don't care about milliseconds. :)

src/components/data/FileBrowser.js
andrewmarcus: Yep, I tried both and left some inconsistency in there. Sorry. Switched back to 5, since you're right, it provides the user with more immediate feedback.

andrewmarcus:
Ok, posted some updates to resolve the things brought up in this PR. It also changes some more code in the Uploader codepath, so probably needs another glance. Sorry...

abaumann (approved):
Reviewed from a functional standpoint, and worked like a charm.  The UX for new vs updated columns is excellent - the color coding is clear and elegant

ehanna4:
BTW looks like the integration tests won't work since this is a fork, so I ran them locally and they all passed 👍 Screen Shot 2021-04-06 at 10 26 26 AM

PR #2429, opened 2021-04-01
[AS-518] Mixpanel instrumentation for data downloads

sagehen03:
I've verified that the events are showing in mix panel and it seems like I didn't break any functionality in the process :grimacing:


src/libs/events.js
```diff
@@ -14,6 +14,7 @@ const eventsList = {
   cloudEnvironmentCreate: 'cloudEnvironment:create',
   cloudEnvironmentDelete: 'cloudEnvironment:delete',
   cloudEnvironmentUpdate: 'cloudEnvironment:update',
+  download: 'workspace:data:download',
```
zarsky-broad: maybe rename this to `workspaceDataDownload` and alphabetize?

src/components/UriViewer.js
```diff
@@ -120,6 +121,7 @@ const DownloadButton = ({ uri, metadata: { bucket, name, fileName, size } }) =>
       div({ style: { display: 'flex', justifyContent: 'center' } }, [
         h(ButtonPrimary, {
           disabled: !url,
+          onClick: () => Ajax().Metrics.captureEvent(Events.download, { downloadFrom: 'bam file' }),
           href: url,
```
zarsky-broad: @panentheos There's no weirdness with having both an `onClick` and an `href`, right? I don't remember any offhand.

zarsky-broad (approved):
Just one small nit!

src/components/UriViewer.js
zarsky-broad: It occurs to me that this will fire on every download, do we want to scope it down a bit more? Maybe make it conditional on the extension of the file?

src/libs/events.js
sagehen03: fair enough

src/components/UriViewer.js
sagehen03: Good catch.  I'll play around it with it a bit and see if I can restrict to .bam extensions.

zarsky-broad: Maybe just ask Janki if we want an event for every download, and pull the extension dynamically?

src/components/UriViewer.js
```diff
@@ -120,6 +121,7 @@ const DownloadButton = ({ uri, metadata: { bucket, name, fileName, size } }) =>
       div({ style: { display: 'flex', justifyContent: 'center' } }, [
         h(ButtonPrimary, {
           disabled: !url,
+          onClick: () => Ajax().Metrics.captureEvent(Events.workspaceDataDownload, { downloadFrom: 'bam file' }),
```
davidangb: nit: this is for downloading any file, not just a bam

Edit: ah, I see Isaac already commented to this effect! Didn't mean to harp on it.

davidangb (commented):
one minor comment inline, else looks good. Nice job getting up to speed on this!

Shakespeared (approved):
Code looks good! 

This might be an Amelia question, but I think we should consider adding properties to track the workspace name and namespace on this. (There's a helper for this called extractWorkspaceDetails.) That would allow the ability to see if people are routinely downloading data from the same workspaces as well as whether a particular user has continued to use Terra or not.

sagehen03:
> Code looks good!
> 
> This might be an Amelia question, but I think we should consider adding properties to track the workspace name and namespace on this. (There's a helper for this called extractWorkspaceDetails.) That would allow the ability to see if people are routinely downloading data from the same workspaces as well as whether a particular user has continued to use Terra or not.

I've reached out to Amelia about including additional metadata, and she would like it.  My latest revision should reflect that.

src/components/UriViewer.js
```diff
@@ -120,6 +121,11 @@ const DownloadButton = ({ uri, metadata: { bucket, name, fileName, size } }) =>
       div({ style: { display: 'flex', justifyContent: 'center' } }, [
         h(ButtonPrimary, {
           disabled: !url,
+          onClick: () => {
+            if (uri.toLowerCase().endsWith('.bam')) {
+              Ajax().Metrics.captureEvent(Events.workspaceDataDownload, { downloadFrom: 'bam file' })
+            }
+          },
```
```suggestion
          onClick: uri.toLowerCase().endsWith('.bam') ?
            () => Ajax().Metrics.captureEvent(Events.workspaceDataDownload, { downloadFrom: 'bam file' }) :
            undefined,
```

sagehen03: For my own edification, what is the advantage of this change?

src/components/UriViewer.js
sagehen03: I will ask!

src/components/UriViewer.js
zarsky-broad: This way, if it's not a bam file, the onClick will be undefined rather than a no-op, so it doesn't even have to call it to find out.

sagehen03: Does that mean at render time we're calling endsWith() on all the url strings in the table?  If so, that seems like we might be trading off a slower render time for faster click handling.  I'm an infrequent front end developer, so if your instincts tell you this is right way to do it, I'll be quiet :)

zarsky-broad: Ah, so because this is in the modal, and that's not rendered until it's mounted, it'll only run once per render within that, and it's cheap enough that I'd tell you to do it this way for consistency. A good thought, though!

sagehen03: I see! I was conflating the other cases and forgot this was in the modal.

sagehen03: I spoke with Janki and she would like us to capture downloads for the other file types so this conditional is obsolete now I think.

zarsky-broad: Maybe something like this to capture the extension?
```js
_.tail(/\.\w+$/.exec(uri))
```

PR #2430, opened 2021-04-01
AS-704: Add integration test for snapshot features

ansingh7115:
<!--
Hello, friend!
Thanks!
--!>


integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,83 @@
...
+
+
+const snapshotName = 'testsnapshot'
+const findWorkflowButton = clickable({ textContains: 'Find a Workflow' })
+
+const testRunWorkflowOnSnapshotFn = _.flow(
+  withWorkspace,
+  withUserToken
+)(async ({ dataRepoUrl, environment, page, testUrl, snapshotColumnName, snapshotId, snapshotTableName, token, workflowName, workspaceName }) => {
+  if ( environment == "perf") {
```
zarsky-broad: Rather than wrapping this in a whole big if statement, what if we just return early? Like this:

```js
  if (environment === 'perf') {
    return
  }
```

Also, please make sure to use `===` instead of `==`. While we don't currently have a way to enforce lint on integration tests, you can run `npx eslint integration-tests/tests --fix` from the root of the repo to clean things up.

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,83 @@
...
+    // do nothing
+  } else {
+    // IMPORT SNAPSHOT
+    await page.goto(testUrl + "/#import-data?url=" + dataRepoUrl + "&snapshotId=" + snapshotId + "&snapshotName=" + snapshotName + "&format=snapshot")
```
zarsky-broad: Instead of concatenating strings, please convert this to a template string. (Here and for other string concatenation instances in this PR)

If you hit option+return on this string, IntelliJ should offer to do it for you.

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,83 @@
...
+const { withUserToken } = require('../utils/terra-sa-utils')
+
+
+const snapshotName = 'testsnapshot'
+const findWorkflowButton = clickable({ textContains: 'Find a Workflow' })
+
+const testRunWorkflowOnSnapshotFn = _.flow(
+  withWorkspace,
+  withUserToken
+)(async ({ dataRepoUrl, environment, page, testUrl, snapshotColumnName, snapshotId, snapshotTableName, token, workflowName, workspaceName }) => {
```
zarsky-broad: Instead of providing the data repo url, we should pull it from the config, like we do in `import-dockstore-workflow`:

```js
  const { dataRepoUrlRoot } = await fetch(`${testUrl}/config.json`).then(res => res.json())
```

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,83 @@
...
+    // START WORKFLOW ON SNAPSHOT
+    // note that this automatically brings in the highest numbered config, which isn't what happens when going through the method repo in FC
+    await waitForNoSpinners(page)
+    await select(page, 'Entity type selector', snapshotName)
+    await select(page, 'Snapshot table selector', snapshotTableName)
+
+    await click(page, clickable({ textContains: 'Inputs' }))
+    await (await findElement(page, input({ labelContains: 'echo_strings.echo_to_file.input1' }))).click({ clickCount: 3 })
+    await fillIn(page, input({ labelContains: 'echo_strings.echo_to_file.input1' }), 'this.' + snapshotColumnName)
+    await delay(500)
```
zarsky-broad: Could you add explanatory comments to the delays and the triple-click?

integration-tests/utils/terra-envs.js
```diff
@@ -1,16 +1,28 @@
...
+    snapshotColumnName: 'name',
+    snapshotId: 'f90f5d7f-c507-4e56-abfc-b965a66023fb',
+    snapshotTableName: 'tableA',
     testUrl: 'https://bvdp-saturn-dev.appspot.com',
     workflowName: 'echo_to_file'
   },
   alpha: {
     billingProject: 'saturn-integration-test-alpha',
+    dataRepoUrl: 'https://data.alpha.envs-terra.bio',
+    snapshotColumnName: 'VCF_File_Name',
```
zarsky-broad, line 20: I understand why the snapshot id would be different, but I think we should try to have the same contents of the snapshot in all environments.

zarsky-broad (changes requested):

zarsky-broad:
Any idea why it's currently failing to import the snapshot?

integration-tests/utils/terra-envs.js
ansingh7115: Agreed! But since QA wants us to manually test the UI until we have a UI integration test in place for these features, I think it would be best to add the tests now and then, once we have your script in place, add the test fixtures we want and change these tests.

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,94 @@
...
+const pRetry = require('p-retry')
+const { withWorkspace } = require('../utils/integration-helpers')
+const { click, clickable, delay, dismissNotifications, findElement, findText, fillIn, input, select, signIntoTerra, waitForNoSpinners, findInGrid, navChild } = require('../utils/integration-utils')
+const { withUserToken } = require('../utils/terra-sa-utils')
+
+
+const snapshotName = 'testsnapshot'
+const findWorkflowButton = clickable({ textContains: 'Find a Workflow' })
+
+const withDataRepoCheck = test => async options => {
```
panentheos, line 12: I know we do this with external services because we don't have control over their uptime, but perhaps we should be stricter with ourselves?

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+const testRunWorkflowOnSnapshotFn = _.flow(
+  withWorkspace,
+  withUserToken,
+  withDataRepoCheck
+)(async ({ dataRepoUrl, environment, page, testUrl, snapshotColumnName, snapshotId, snapshotTableName, token, workflowName, workspaceName }) => {
+  if (environment === 'perf') {
+    return
```
panentheos, line 29: If we're not guaranteed to have the environment pre-configured, what about explicitly poking in the state that we need, similar to what we do with `createEntityInWorkspace`?

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,94 @@
...
+  // START WORKFLOW ON SNAPSHOT
+  // note that this automatically brings in the highest numbered config, which isn't what happens when going through the method repo in FC
+  await waitForNoSpinners(page)
+  await select(page, 'Entity type selector', snapshotName)
+  await select(page, 'Snapshot table selector', snapshotTableName)
+
+  await click(page, clickable({ textContains: 'Inputs' }))
+  await (await findElement(page, input({ labelContains: 'echo_strings.echo_to_file.input1' }))).click({ clickCount: 3 })
+  await fillIn(page, input({ labelContains: 'echo_strings.echo_to_file.input1' }), `this.${snapshotColumnName}`)
+  await delay(500) // Without this delay, the input field sometimes reverts back to its default value
```
panentheos: TOL: Is this something the user could run into, if they were quick enough? If so, we might consider closing that gap in the operation of the page.

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,94 @@
...
+  await click(page, clickable({ text: 'Add to Workspace' }))
+
+  // START WORKFLOW ON SNAPSHOT
+  // note that this automatically brings in the highest numbered config, which isn't what happens when going through the method repo in FC
+  await waitForNoSpinners(page)
+  await select(page, 'Entity type selector', snapshotName)
+  await select(page, 'Snapshot table selector', snapshotTableName)
+
+  await click(page, clickable({ textContains: 'Inputs' }))
+  await (await findElement(page, input({ labelContains: 'echo_strings.echo_to_file.input1' }))).click({ clickCount: 3 })
```
panentheos: What is this doing?

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,94 @@
...
+  await (await findElement(page, input({ labelContains: 'echo_strings.echo_to_file.out' }))).click({ clickCount: 3 })
+  await fillIn(page, input({ labelContains: 'echo_strings.echo_to_file.out' }), 'workspace.result')
+  await delay(1000) // Without this delay, the input field sometimes reverts back to its default value
+  await click(page, clickable({ text: 'Save' }))
+
+  await delay(1000) // The Run Analysis button requires time to become enabled after hitting the save button
+  await click(page, clickable({ textContains: 'Run analysis' }))
+
+  await Promise.all([
+    page.waitForNavigation(),
```
panentheos: I thought that `waitForNavigation` was for sites that did full page transitions. Is this necessary here?

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,94 @@
...
+  ])
+
+  // CHECK WORKFLOW SUCCEEDED AND RESULT IS WRITTEN
+  await pRetry(async () => {
+    try {
+      await findInGrid(page, 'Succeeded', { timeout: 65 * 1000 }) // long enough for the submission details to refresh
+    } catch (e) {
+      throw new Error(e)
+    }
+  }, { retries: 10, factor: 1 })
```
panentheos, line 76: I see this pattern was copied from the other test, but I was curious if this needed to be a `pRetry` as opposed to just having an extra long `timeout` in the operation.

integration-tests/tests/run-workflow-on-snapshot.js
ansingh7115: Do you mean we shouldn't check this?

integration-tests/tests/run-workflow-on-snapshot.js
ansingh7115: I think adding the snapshot when we need to test is a good idea (Data Repo's perf env is volatile) - Isaac is working on a script that will let us create snapshots (currently we have to put in a request with the Data Repo team to make snapshots). Once we have that, we could let the tests create and delete their own artifacts.

integration-tests/tests/run-workflow-on-snapshot.js
ansingh7115: It's possible - I saw this with even this 1/2 second delay sometimes so I upped it to a full second. I haven't seen users complain about it though

integration-tests/tests/run-workflow-on-snapshot.js
ansingh7115: Added a comment - it's a triple click. The input box already contains a default string. Without the triple click, we simply append. Triple clicking highlights the string so that we can replace it

integration-tests/tests/run-workflow-on-snapshot.js
ansingh7115: I tried it without that line and it seemed to work!

integration-tests/tests/run-workflow-on-snapshot.js
ansingh7115: I think this is probably a better question for @zarsky-broad - I know less about best practice here :)

zarsky-broad: IIRC, this was the best way we could find to retry a certain number of times, I don't recall why just jacking the timeout didn't feel like a good option. I'm sure we must have thought of it, though, right? ...Right? Ugh. @breilly2 @ehanna4 @petesantos anybody remember anything here?

breilly2: Maybe because of how the background status refresh works, we have to make sure we re-run `findInGrid` for it to detect the update?

integration-tests/tests/run-workflow-on-snapshot.js
panentheos: I think we should consider that, yes. Right now this encodes the idea that "It's ok if data repo is down". Regardless of whether that's the right tactical decision at this point, I think it's good to remember that decisions like this can have a subtle influence on our expectations for our systems.

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,93 @@
...
+  await click(page, clickable({ text: 'Add to Workspace' }))
+
+  // START WORKFLOW ON SNAPSHOT
+  // note that this automatically brings in the highest numbered config, which isn't what happens when going through the method repo in FC
+  await waitForNoSpinners(page)
+  await select(page, 'Entity type selector', snapshotName)
+  await select(page, 'Snapshot table selector', snapshotTableName)
+
+  await click(page, clickable({ textContains: 'Inputs' }))
+  await (await findElement(page, input({ labelContains: 'echo_strings.echo_to_file.input1' }))).click({ clickCount: 3 }) // triple-click to replace the default text
```
panentheos: Ah, makes sense. This seems like a solid abstraction we could pull out into a utility function, which would be self-documenting and allow reuse.

integration-tests/tests/run-workflow-on-snapshot.js
panentheos: Consider this a TOL, feel free to make the right tactical decision on this.

integration-tests/tests/run-workflow-on-snapshot.js
panentheos: Good to know this is coming. In the meantime, I think this is a decent approach.

ansingh7115: Created https://broadworkbench.atlassian.net/browse/AS-715 for this work

integration-tests/tests/run-workflow-on-snapshot.js
ansingh7115: I'm going to leave this as is for now. I think the question of "what should the UI tests should do if urls we depend on don't work" is a broader question that I will leave to grown-up UI folks.

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -0,0 +1,89 @@
...
+
+  await click(page, clickable({ textContains: 'Outputs' }))
+  await fillInReplace(page, input({ labelContains: 'echo_strings.echo_to_file.out' }), 'workspace.result')
+  await click(page, clickable({ text: 'Save' }))
+
+  await delay(1000) // The Run Analysis button requires time to become enabled after hitting the save button
+  await click(page, clickable({ textContains: 'Run analysis' }))
+
+  await Promise.all([
+    click(page, clickable({ text: 'Launch' }))
```
panentheos: Since there's only one promise now, this doesn't need `Promise.all`

integration-tests/utils/integration-utils.js
```diff
@@ -61,6 +61,13 @@ const fillIn = async (page, xpath, text) => {
   return (await page.waitForXPath(xpath)).type(text, { delay: 20 })
 }
 
+// Replace pre-existing value
+const fillInReplace = async (page, xpath, text) => {
+  await (await findElement(page, xpath)).click({ clickCount: 3 }) // triple-click to replace the default text
+  await fillIn(page, xpath, text)
+  return await delay(1000) // Without this delay, the input field sometimes reverts back to its default value
```
panentheos: This `delay` is specific to the workflow page, right? If so, it shouldn't be baked into the utility. Otherwise, this util looks good.

panentheos (approved):
Looks good, thanks!

zarsky-broad (approved):

PR #2431, opened 2021-04-02
add better tooltips to table names [AS-677]

Shakespeared:
Fixes the problem of long table/entity names being cut off without a good tooltip, also makes it so the number of entities is never cut off. Added the nice tooltips with a slight delay so it's not annoying if you temporarily mouse over things. 

Tested in the UI

### DIFF
--- a/src/components/TooltipTrigger.js
+++ b/src/components/TooltipTrigger.js
@@ -33,11 +33,22 @@ const styles = {
   }
 }
 
-const Tooltip = ({ side = 'bottom', type, target: targetId, children, id }) => {
+const Tooltip = ({ side = 'bottom', type, target: targetId, children, id, delay }) => {
+  const [shouldRender, setShouldRender] = useState(!delay)
+  const renderTimeout = useRef()
   const elementRef = useRef()
   const [target, element, viewport] = useDynamicPosition([{ id: targetId }, { ref: elementRef }, { viewport: true }])
+
+  Utils.useOnMount(() => {
+    if (!!delay) {
+      renderTimeout.current = setTimeout(() => setShouldRender(true), delay)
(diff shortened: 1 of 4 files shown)


MatthewBemis (approved):
Code LGTM.

Tested in the UI as well and looked good. I like the delay.

src/components/TooltipTrigger.js
```diff
@@ -70,31 +70,46 @@ const Tooltip = ({ side = 'bottom', type, target: targetId, children, id }) => {
...
-const TooltipTrigger = ({ children, content, ...props }) => {
+const TooltipTrigger = ({ children, content, delay, ...props }) => {
   const [open, setOpen] = useState(false)
+  const openTimeout = useRef()
   const id = Utils.useUniqueId()
   const tooltipId = Utils.useUniqueId()
   const child = Children.only(children)
   const childId = child.props.id || id
+
+  const doOpen = !delay ?
```
panentheos: This logic looks good, but I had to think about it carefully.

As a general tip, I sometimes find it helpful to tie this kind of logic to the component lifecycle (if possible). For example, if we pushed the delay down into the `Tooltip`, then this component could continue to do the simple on/off toggle, and the `Tooltip` itself could just set a one-shot timer on mount (and clear it on unmount), and render nothing during the delay.

I haven't typed it out so I don't know if it would end up better, but maybe try it out if it sounds beneficial.

Shakespeared: @panentheos we moved the delay inside the tooltip itself, but we ended up having to use display: none on it to prevent the calculated dimensions from exploding… What do you think? Better or worse?

davidangb (approved):
lgtm!

PR #2432, opened 2021-04-07
Add extra scopes to fence link for signed URLs [BT-213]

breilly2:
https://broadworkbench.atlassian.net/browse/BT-213

@mikebaumann says that these additional scopes are needed in order for Gen3 to be able to generate and provide Terra with signed URLs for data access.

Please also review the CLIA and security fields on the Jira ticket as part of this review.

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -324,7 +324,7 @@ const User = signal => ({
 
   getFenceAuthUrl: async (provider, redirectUri) => {
     const queryParams = {
-      scopes: ['openid', 'google_credentials'],
+      scopes: ['openid', 'google_credentials', 'data', 'user'],
       redirect_uri: redirectUri,
       state: btoa(JSON.stringify({ provider }))
     }


kshakir (approved):
Those characters look like valid JavaScript to me. 👍

I'm assuming this was manually tested and did what it was supposed to do?

breilly2:
I manually tested this code for BDC (which was tricky because `localhost` isn't a valid redirect URI for Fence).

@mikebaumann presumably tested this change for the other projects when he put together the notebook about linking (https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/DRS%20and%20Signed%20URL%20Development%20-%20Dev/notebooks/launch/1_Linking_Terra_user_to_Fence_user_with_data_scope.ipynb). I verified that the AnVIL link in that notebook works, which means the Terra UI code changes should work too. I still have to test CRDC, though, which has different login requirements than the others so is taking a little extra effort. Stay tuned...

breilly2:
Successfully tested CRDC link with additional scopes.

mcovarr (approved):

PR #2435, opened 2021-04-07
AS-720: Fix display for gs uris that have a slash at the end

ansingh7115:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/UriViewer.js
+++ b/src/components/UriViewer.js
@@ -268,7 +268,7 @@ export const UriViewerLink = ({ uri, googleProject }) => {
         e.preventDefault()
         setModalOpen(true)
       }
-    }, [isGs(uri) ? _.last(uri.replace(/\/$/, '').split('/')) : uri]),
+    }, [isGs(uri) ? _.last(uri.split(/\/\b/)) : uri]),
     modalOpen && h(UriViewer, {
       onDismiss: () => setModalOpen(false),
       uri, googleProject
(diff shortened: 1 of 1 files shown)


zarsky-broad:
This is a significant change in behavior, previously we just showed the file name because the path before it could be very long; this will fix the issue if an address is for a location and not a specific file, but are we ok with the change to what's displayed for files?

ansingh7115:
That's true, I don't want to change the intent of this code - I updated it to continue showing only the file name while removing trailing slashes

src/components/UriViewer.js
```diff
@@ -268,7 +268,7 @@ export const UriViewerLink = ({ uri, googleProject }) => {
         e.preventDefault()
         setModalOpen(true)
       }
-    }, [isGs(uri) ? _.last(uri.split('/')) : uri]),
+    }, [isGs(uri) ? _.last(uri.replace(/\/$/, '').split('/')) : uri]),
```
```suggestion
    }, [isGs(uri) ? _.last(uri.split(/\/\b/)) : uri]),
```

zarsky-broad: What about this?

zarsky-broad: As a bonus, this will retain a trailing slash.

ansingh7115: That seems to work - it just leaves the extra slash at the end (which I think is probs fine)

ansingh7115: Updated

zarsky-broad (approved):

MatthewBemis (approved):
LGTM. Did not test myself!

PR #2438, opened 2021-04-08
[BW-577] Implement CSP

panentheos:
This adds a plugin that injects Content Security Policy information into a `meta` tag in the site HTML.

Two important caveats:
* Because the CSP interferes with the development environment, the plugin is necessarily disabled in (local) dev mode. It's active in any 'built' environment, which includes prod, the various appspot environments (e.g. dev, alpha), and the PR sites.
* The CSP restricts which external scripts are allowed to run, and how. This tends to affect third-party tools, like AppCues.

Since this introduces a configuration difference between dev and prod modes, it will **no longer be sufficient to use local testing for changes that introduce or modify third-party integrations**. Essentially, any change that introduces new script tags needs to be verified on a PR site, to make sure the CSP doesn't block the resource.


config-overrides.js
```diff
@@ -1,18 +1,37 @@
...
+          css: true
+        }
+      ]),
+      addWebpackPlugin(new CspHtmlWebpackPlugin({
+        // Per our security policy, 'script-src' cannot contain any of the options containing 'unsafe', e.g. 'unsafe-inline'
+        'script-src': [
+          '\'self\'',
+          'https://fast.appcues.com', // secondary script loaded by appcues
+          'https://us.jsagent.tcell.insight.rapid7.com', // secondary script loaded by tcell
+          'https://cdnjs.cloudflare.com' // libraries loaded by notebook preview
```
dmohs, line 24: Does this mean, "Any script can run if I can get it served from cdnjs?"

dmohs (approved):
I don't really understand what this means, but I trust your judgement. I would recommend an addition to `.github/pull_request_template.md` regarding the need for testing scripts.

zarsky-broad:
Did we consider https://github.com/YIZHUANG/react-csp? I'd love to avoid a new rewire, as I'd like to remove rewiring at some point.

panentheos:
@zarsky-broad Not specifically, but this plugin does a lot more than that one, including auto-generating nonces so existing scripts work without further steps.

config-overrides.js
panentheos: Yes. However, it's important to note that in order to inject an arbitrary link to cdnjs, you'd need to already have an XSS vector, at which point it's probably moot. Ultimately we're relying on React's built-in XSS protection like we always have.

zarsky-broad:
If you could run `npx yarn-deduplicate; yarn` I think it might tidy up the lock file a bit

kyuksel:
Closing due to inactivity. Please re-open if applicable.

dmohs:
We need to get this in, and Brett was the only one with context. I believe this repo's code owners are the best people to see this through.

rtitle:
Just want to call out that this is blocked by https://broadworkbench.atlassian.net/browse/IA-2687, which is to make calhoun work with the more restrictive CSP. IA should own making that change.

dmohs:
Verified hot reloading locally. Small changes reloaded fine without issue. Ran `yarn build` to create a production build and tested by clicking around a bit. Not sure if further verification is needed.

dmohs:
Not sure why the "run-notebook" test is failing. I re-ran without luck. Screenshot looks correct to me.

kyuksel:
> Not sure why the "run-notebook" test is failing. I re-ran without luck. Screenshot looks correct to me.

@dmohs The test seems to be timing out while waiting for the Jupyter kernel to start, which matches the failure signature we've been seeing as part of the flaky test monitoring. It has gotten better recently although we're not sure why. I just kicked off a re-run.

config-overrides.js
```diff
@@ -1,18 +1,37 @@
...
+          languages: ['bash', 'python'],
+          plugins: ['line-numbers'],
+          theme: 'default',
+          css: true
+        }
+      ]),
+      addWebpackPlugin(new CspHtmlWebpackPlugin({
+        // Per our security policy, 'script-src' cannot contain any of the options containing 'unsafe', e.g. 'unsafe-inline'
+        'script-src': [
+          '\'self\'',
```
zarsky-broad, line 21: @dmohs Since we can't use `unsafe-inline`, our one inline script (it loads the outdated browser warning) will have to be moved to a separate file. You'll find it near the top of our `index.html`.

Also, do we need `apis.google.com`?

kyuksel:
Converted to draft since there's an [external blocking dependency](https://broadworkbench.atlassian.net/browse/IA-2919).

jdcanas:
Calhoun issues should be resolved, the changes here have been deployed to calhoun prod https://github.com/DataBiosphere/calhoun/pull/22. cc @dmohs

dmohs:
Moved out of Draft since this is no longer blocked, but I need to debug the test failure.

config-overrides.js
dmohs: I don't see any difference when moving this to an external script. Advice on how to test?

dmohs:
Currently blocked on this because I can't get notebooks to work in dev:
Screen Shot 2021-11-24 at 14 55 53

config-overrides.js
zarsky-broad: Try to open the app in an unsupported browser, or change the list so yours is unsupported, and see if it warns you.

zarsky-broad:
Looks like a header on the response from Leo isn't being set. I believe this is the reason Safari hasn't worked, it expects a header that's missing as well.

integration-tests/utils/integration-utils.js
```diff
@@ -166,6 +166,8 @@ const withScreenshot = _.curry((testName, fn) => async options => {
             await storage.bucket(screenshotBucket).upload(failureNotificationDetailsPath)
           }
         }
+        console.error(e)
+        throw e
```
dmohs: I have another branch to clean this up, so I'll take this out before merging.

config-overrides.js
dmohs: This is it, right?
Screen Shot 2021-12-16 at 10 22 27

config-overrides.js
```diff
@@ -1,18 +1,37 @@
...
+    const newConfig = _.flow(
+      addBabelPlugin([
+        'prismjs', {
+          languages: ['bash', 'python'],
+          plugins: ['line-numbers'],
+          theme: 'default',
+          css: true
+        }
+      ]),
+      addWebpackPlugin(new CspHtmlWebpackPlugin({
```
petesantos, line 18: +1 on @zarsky-broad's question on `api.google.com`

I am also seeing this script getting loaded:
`https://us.jsagent.tcell.insight.rapid7.com/tcellagent.min.js`

from the [tcell file](https://github.com/DataBiosphere/terra-ui/blob/f1f60584e7c2d09e4d49191564df3bdc5343bcb8/src/libs/tcell.js#L7)

It looks like a tool we are or were using, but I am not sure of the context around this one.

config-overrides.js
dmohs: > Also, do we need `apis.google.com`?

How would we be able to determine this?

petesantos: I would think auth would fail. But I am surprised it isn't failing.
Have we been able to determine definitively whether the CSP is working?
Perhaps trying something like:
1. Pulling in some JS from a trusted site - adding it to our HTML file (do **_not_** add it to the CSP)
2. Commit the code so it is on the PR site where we can test it

We should see an error in the console saying that code violates the CSP - something like:
image

dmohs: apis.google.com appears to have a nonce, so I think that one is okay.

zarsky-broad: This specifically takes care of inline scripts for us, too: https://github.com/slackhq/csp-html-webpack-plugin#recommended-configuration

zarsky-broad: Or more accurately, it takes care of any script tags that are present when the page passes through webpack, whether they're inline or links, I think.

public/index.html
```diff
@@ -5,34 +5,7 @@
...
-              appStore: "Please update iOS from the Settings App"
-            },
-            url: "https://www.google.com/chrome/",
-            callToAction: (!!window.chrome ? "Update" : "Download") + " Chrome now",
-            close: "Close"
-          }
-        }
-      });
-    </script>
+    <script src="%PUBLIC_URL%/outdated-browser-message.js"></script>
```
dmohs, line 8: Tested by manually changing the acceptable version of Chrome in the production build and seeing the warning.

dmohs:
@zarsky-broad @petesantos ready for re-review.

zarsky-broad (approved):

PR #2439, opened 2021-04-08
[IA-2539] Changing user facing text 

xuscrbw:
<!--
Hello, friend!
Thanks!
--!>
This is a pretty straightforward change. I have added a helper function that helps to transform the Starting/Stopping/Stopped language to the Resuming/Pausing/Paused language from Kaan's ticket: https://broadworkbench.atlassian.net/browse/IA-2389

I tested this locally to make sure that the text matches. I also moved the capitalize functionality to within the helper for easy changes if needed

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -166,3 +166,16 @@ export const appIsSettingUp = app => {
 export const collapsedRuntimeStatus = runtime => {
   return runtime && (runtime.patchInProgress ? 'LeoReconfiguring' : runtime.status) // NOTE: preserves null vs undefined
 }
+
+export const convertedAppStatus = appStatus => {
+  switch (appStatus) {
+    case 'STOPPED':
+      return _.capitalize('PAUSED')
+    case 'STOPPING':
+      return _.capitalize('PAUSING')
+    case 'STARTING':
+      return _.capitalize('RESUMING')
+    default:
+      return _.capitalize(appStatus)
+  }
+}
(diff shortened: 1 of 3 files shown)


src/libs/runtime-utils.js
```diff
@@ -166,3 +166,16 @@ export const appIsSettingUp = app => {
 export const collapsedRuntimeStatus = runtime => {
   return runtime && (runtime.patchInProgress ? 'LeoReconfiguring' : runtime.status) // NOTE: preserves null vs undefined
 }
+
+export const convertedAppStatus = appStatus => {
```
jdcanas, line 171: try `Utils.switchCase` for style consistency, 
ex: https://github.com/DataBiosphere/terra-ui/blob/dev/src/components/NewRuntimeModal.js#L1130

xuscrbw: Addressed and fixed

jdcanas (approved):
tested locally

src/libs/runtime-utils.js
```diff
@@ -166,3 +167,12 @@ export const appIsSettingUp = app => {
...
+  return Utils.switchCase(appStatus,
```
```suggestion
  return Utils.switchCase(_.upperCase(appStatus),
```
might be good to ensure the conversion would still work in case `appStatus` is not all uppercase.

kyuksel (approved):
Looks great 👍🏼 
Tested locally. 
One suggestion inline that may make the code a bit more robust.

src/libs/runtime-utils.js
xuscrbw: addressed!

PR #2440, opened 2021-04-08
Add error notification when Galaxy App transitions to Error [IA-2655]

ehanna4:
Working on getting Galaxy apps to more parity with Notebooks
- previuosly a transition to "Error" status would not trigger a  popup, now it will
Screen Shot 2021-04-08 at 4 39 13 PM

Possibly some code duplication in the error handling though

### DIFF
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -22,7 +22,7 @@ import { clearNotification, notify } from 'src/libs/notifications'
 import {
   appIsSettingUp, collapsedRuntimeStatus, currentApp, currentRuntime, persistentDiskCost, runtimeCost, trimRuntimesOldestFirst
 } from 'src/libs/runtime-utils'
-import { errorNotifiedRuntimes, errorNotifiedApps } from 'src/libs/state'
+import { errorNotifiedApps, errorNotifiedRuntimes } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
 
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -236,7 +236,7 @@ export default class RuntimeManager extends PureComponent {
           })
         ])
       })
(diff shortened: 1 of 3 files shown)


Qi77Qi (approved):
🎉

src/components/RuntimeManager.js
```diff
@@ -108,6 +110,50 @@ const RuntimeErrorNotification = ({ runtime }) => {
...
+export const AppErrorModal = ({ app, onDismiss }) => {
+  const [error, setError] = useState()
+  const [loadingAppDetails, setLoadingAppDetails] = useState(false)
+
+  const loadAppError = _.flow(
+    withErrorReporting('Error loading app details'),
+    Utils.withBusyState(setLoadingAppDetails)
+  )(async () => {
+    const { errors: appErrors } = await Ajax().Apps.app(app.googleProject, app.appName).details()
+    setError(appErrors[0].errorMessage)
```
jdcanas: should this be `appErrors[0]?.errorMessage` to guard against when there are none (one of those 'shouldn't happen but why not be safe' things)

src/components/RuntimeManager.js
```diff
@@ -183,6 +229,11 @@ export default class RuntimeManager extends PureComponent {
           })
         ])
       })
+    } else if (prevApp && prevApp.status !== 'ERROR' && app && app.status === 'ERROR' && !_.includes(app.appName, errorNotifiedApps.get())) {
```
jdcanas: I think this should be `else if (app?.status === 'ERROR' && !_.includes(app.appName, errorNotifiedApps.get()))`. I think it's important to show the error on page load (aka if they navigate away when the error occurs and come back, or just open the page to an errored app).

jdcanas (approved):
Good change. Very annoying not to be able to see the error details. Few small changes I think are worth talking about.

src/components/RuntimeManager.js
ehanna4: ah yeah i like that idea, with the runtime it saves it to the little error icon so they can reopen it but galaxy doesnt have that

PR #2441, opened 2021-04-09
[PF-692] Pass workspace bucket zone/region parameter to Leonardo createRuntime…

wnojopra:
In https://github.com/DataBiosphere/leonardo/pull/1933, zone and region parameters were added to createRuntime so Leo can now create instances outside of us-central1. This PR will use the location of the workspace bucket to determine the region to create the instance.

In a future PR I plan to have a UI dropdown that will allow a user to specifically choose the location of their cloud environment (defaulting to their current workspace's bucket location). 

I have tested:
1) Creating a GCE instance with the default us-multiregional bucket.
2) Creating a cluster with the default us-multiregional bucket.
3) Creating a GCE instance in a non default location (like australia-southeast1)
4) Creating a cluster in a non default location (like australia-southeast1)

In all cases I verified that the instance was created in the specified zone:
image


src/components/region-common.js
```diff
@@ -1,74 +1,77 @@
...
 // 'flag' will always be defined (even if it's a question mark.
 // 'regionDescription' is the same as location when locationType is 'multi-region', or a country name when locationType is 'region'.
+// computeZone is generally the 'a' zone for each region, except for those regions where it is not available.
+// The choice to use the 'a' zone is arbitrary, choosing 'b' zone would also work.
+// The region choice for multi-region locations is arbitrary as well.
 export const unknownRegionFlag = '❓'
 export const regionInfo = (location, locationType) => {
   switch (locationType) {
     case 'multi-region':
       switch (location) {
```
jdcanas: we have a utility function `Utils.switchcase` (in `utils.js`) to clean up the syntax of switches a bit. I know it existed before your changes, but could be nice to clean up. Up to you

src/components/NewRuntimeModal.js
```diff
@@ -258,10 +261,12 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
     const shouldDeleteRuntime = oldRuntime && !this.canUpdateRuntime()
     const shouldCreateRuntime = !this.canUpdateRuntime() && newRuntime
     const { name, bucketName, googleProject } = this.getWorkspaceObj()
+    const { computeZone, clusterRegion } = regionInfo(bucketLocation, bucketLocationType)
```
jdcanas: can we call `clusterRegion` `computeRegion` for consistency? (comment applies to `region-common.js` too)

src/components/NewRuntimeModal.js
```diff
@@ -487,13 +499,17 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
       currentPersistentDisk ? Ajax().Disks.disk(currentPersistentDisk.googleProject, currentPersistentDisk.name).details() : null
     ])
 
+    const { location, locationType } = await Ajax().Workspaces.workspace(namespace, workspaceName).checkBucketLocation(bucketName)
```
jdcanas, line 508: we may want to provide a default here in case the remote call fails (google down-time, etc), otherwise there will be some errors that cascade (thinking about the failure mode of this line in particular `const { computeZone, clusterRegion } = regionInfo(bucketLocation, bucketLocationType)`. Unlikely that the app would actually work in the proposed scenario, but I like guarding against javascript crash scenarios wherever possible...

jdcanas (approved):
Played around with it locally and everything passed to leo and in the code looks good. Couple small nits for consistency/redundancy.
Ran the integration tests locally and they were all green, they will not pass here as this PR is from a fork. 
The conflict its complaining about will look daunting when you try to rebase, but its just a lot of renaming, which can be referenced here https://github.com/DataBiosphere/terra-ui/pull/2447

kyuksel (approved):
Looks good to me pending Justin's comments being addressed 👍🏼

src/components/region-common.js
```diff
@@ -1,74 +1,47 @@
...
-          return { flag: '🇺🇸', regionDescription: `${locationType}: ${location} (Salt Lake City)` }
-        case 'US-WEST4':
-          return { flag: '🇺🇸', regionDescription: `${locationType}: ${location} (Las Vegas)` }
-        default:
-          return { flag: unknownRegionFlag, regionDescription: `${locationType}: ${location}` }
-      }
-    default:
-      return { flag: unknownRegionFlag, regionDescription: `${locationType}: ${location}` }
-  }
+  return Utils.switchCase(locationType,
```
zarsky-broad, line 12: If you declare a 
```js
const regionDescription = `${locationType}: ${location}`
```
above this, you can simplify a lot of the lines below.

kyuksel:
@wnojopra Curious if we should expect any updates on this PR soon?

wnojopra:
Thanks for asking @kyuksel . We'd like to merge this PR together with #2420 as they both introduce rationality to Terra. By introducing regional buckets, users could end up accruing egress charges when data is moved between regions. This could happen between a regional bucket and a notebook VM or a workflow VM. To address this, I'm working on two PRs in Cromwell ([cromwell/pull/6332](https://github.com/broadinstitute/cromwell/pull/6332) and [cromwell/pull/6324](https://github.com/broadinstitute/cromwell/pull/6324))that adds options to minimize egress charges. I'd like to get those changes in before we make regional buckets/VMs in Terra more accessible. 

TL;DR - pending Cromwell PRs https://github.com/broadinstitute/cromwell/pull/6324 and https://github.com/broadinstitute/cromwell/pull/6332

petesantos:
Closing because this PR appears to be stale. If you need to finish this work, feel free to reopen.

PR #2445, opened 2021-04-16
IA-2634 IA-2668 use bootDisk returned from leo

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2634
https://broadworkbench.atlassian.net/browse/IA-2668

As part of gpu epic, we'll likely to change bootDiskSize. Hence, update cost estimate to rely on `bootDiskSize` returned from leo instead of hardcoded value.

Tested in Fiab. Notice `Running cloud compute cost` changed from 0.05 to 0.06 (I have terra-ui running against my fiab with `bootDisk` updated to 60G)
* Open the create runtime panel when there's no runtime existent
Screen Shot 2021-04-16 at 4 46 07 PM

* Open the create runtime panel when there is a runtime existent
Screen Shot 2021-04-16 at 4 51 22 PM

### DIFF
--- a/src/components/NewRuntimeModal.js
+++ b/src/components/NewRuntimeModal.js
@@ -340,7 +340,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
             ...(jupyterUserScriptUri && { jupyterUserScriptUri }),
             ...(cloudService === cloudServices.GCE ? {
               machineType: masterMachineType || defaultGceMachineType,
-              bootDiskSize: oldRuntime?.bootDiskSize,
+              bootDiskSize: existingRuntime?.bootDiskSize,
               ...(this.shouldUsePersistentDisk() ? {
                 persistentDiskAttached: true
               } : {
--- a/src/components/NewRuntimeModal.js
+++ b/src/components/NewRuntimeModal.js
@@ -341,7 +341,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
             ...(cloudService === cloudServices.GCE ? {
               machineType: masterMachineType || defaultGceMachineType,
               // bootDiskSize can be null if it's an really old runtime
-              bootDiskSize: existingRuntime?.bootDiskSize || 0,
+              bootDiskSize: existingRuntime?.bootDiskSize || 50,
               ...(this.shouldUsePersistentDisk() ? {
                 persistentDiskAttached: true
               } : {
--- a/src/components/NewRuntimeModal.js
+++ b/src/components/NewRuntimeModal.js
@@ -340,7 +340,8 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
             ...(jupyterUserScriptUri && { jupyterUserScriptUri }),
             ...(cloudService === cloudServices.GCE ? {
               machineType: masterMachineType || defaultGceMachineType,
-              bootDiskSize: existingRuntime?.bootDiskSize,
+              // bootDiskSize can be null if it's an really old runtime
+              bootDiskSize: existingRuntime?.bootDiskSize || 0,
               ...(this.shouldUsePersistentDisk() ? {
                 persistentDiskAttached: true
               } : {
--- a/src/components/NewRuntimeModal.js
+++ b/src/components/NewRuntimeModal.js
@@ -183,11 +183,12 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
    */
   getPendingRuntimeConfig() {
     const { runtime: desiredRuntime } = this.getDesiredEnvironmentConfig()
+
     return {
       cloudService: desiredRuntime.cloudService,
       ...(desiredRuntime.cloudService === cloudServices.GCE ? {
-        bootDiskSize: 50,
         machineType: desiredRuntime.machineType || defaultGceMachineType,
+        bootDiskSize: desiredRuntime.bootDiskSize,
         ...(desiredRuntime.diskSize ? {
           diskSize: desiredRuntime.diskSize
         } : {})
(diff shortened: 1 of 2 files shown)


src/components/NewRuntimeModal.js
```diff
@@ -184,7 +184,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
     return {
       cloudService: newRuntime.cloudService,
       ...(newRuntime.cloudService === cloudServices.GCE ? {
-        bootDiskSize: 50,
+        bootDiskSize: newRuntime.bootDiskSize,
```
jdcanas: Is this already returned by leonardo? When i try to print this out locally it isn't defined in `newRuntime`, which has these fields:

Screen Shot 2021-04-16 at 2 22 09 PM

I think a default value should be present (e.g. `bootDiskSize: oldRuntime.bootDiskSize ? oldRuntime.bootDiskSize : someDefaultValue`)  in case a runtime isn't present to reference. However, I think you want to look  into `this.getOldEnvironmentConfig()`, since you seem to be trying to source the bootdisk size off an existing runtime the user has when creating a new environment

Qi77Qi: so I realize we display the cost via different code (I think) when there's no runtime existent...see the screenshots in description

src/components/NewRuntimeModal.js
```diff
@@ -379,6 +381,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
         ...(currentRuntimeDetails?.jupyterUserScriptUri && { jupyterUserScriptUri: currentRuntimeDetails?.jupyterUserScriptUri }),
         ...(cloudService === cloudServices.GCE ? {
           machineType: runtimeConfig.machineType || defaultGceMachineType,
+          bootDiskSize: runtimeConfig.bootDiskSize,
           ...(runtimeConfig.persistentDiskId ? {
             persistentDiskAttached: true
           } : {
```
kyuksel, line 387: Any reason why we restrict getting `bootDiskSize` from Leo to the `cloudService === cloudServices.GCE` case?

kyuksel (commented):
LGTM with one question. Tested against `dev` Leo so couldn't verify the `bootDiskSize` increase but that would be coming from Leo so I'm not concerned.

src/components/NewRuntimeModal.js
Qi77Qi: there's no bootDisk vs user disk for dataproc clusters

kyuksel: Ah right…`masterDiskSize` and `workerDiskSize` are both boot disk sizes for Dataproc  👍🏼

kyuksel (approved):

kyuksel (changes requested):
Forgot to mention previously, [Emily's PR](https://github.com/DataBiosphere/terra-ui/pull/2447) renamed `old/new` as `existing/desired` so it'd be good to make sure the ones you've touched also use the new convention after resolving conflicts with the `dev` branch 😊

src/components/NewRuntimeModal.js
```diff
@@ -339,6 +340,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
             ...(jupyterUserScriptUri && { jupyterUserScriptUri }),
             ...(cloudService === cloudServices.GCE ? {
               machineType: masterMachineType || defaultGceMachineType,
+              bootDiskSize: existingRuntime?.bootDiskSize,
```
jdcanas, line 343: should we specify an optional value here to act as a stub in cost calculation if an existing runtime doesn't exist? I think we discussed wanting something like `existingRuntime?.bootDiskSize || 50`, like the above machineType has a default. You could specify it a level above  in the `Pending` function too

kyuksel (approved):

src/components/NewRuntimeModal.js
kyuksel: That'd be safer 👍🏼 

Let's not use a magic number like `50` though. Could use `DEFAULT_DISK_SIZE` which equals `50` or define a `DEFAULT_GCE_BOOT_DISK_SIZE` or something if we want to default to something other than `50` after increasing the GCE boot disk size.

Qi77Qi: @jdcanas , when we were talking about it, we thought we need to have a default value becuz we were thinking cost is displayed via this code if there's no existing runtime, but I don't think that's the case (I tested UI without any runtime, and things look correct)....`existingRuntime?` is already checking if `existingRuntime` is null I think, and if there's an existing runtime, `bootDiskSize` will always exist for gce runtimes.

jdcanas: its safe in terms of not accessing the value on an `undefined` variable. So we are safely handling the case where existingRuntime doesn't exist. However, in this case `existingRuntime.bootDiskSize` will be undefined. I think we do want to include a default value to be used in the cost calculation. See attached screenshot (I added these prints at beginning of `getPendingRuntimeConfig`)
Screen Shot 2021-04-21 at 9 11 02 AM

Qi77Qi: `bootDiskSize` is a required field in leo DB for GCE. Did you mock out leo's response or something?

jdcanas (approved):

src/components/NewRuntimeModal.js
Qi77Qi: talked with @jdcanas on zoom...there's default added in src/libs/runtime-utils.js now.

Qi77Qi:
@kyuksel there's a new `DEFAULT_BOOT_DISK_SIZE` added

kyuksel (approved):
re- 👍🏼 ing

ehanna4:
If you try merging dev into your branch hopefully it will build. Sky merged a [PR](https://github.com/DataBiosphere/terra-ui/pull/2452) that hopefully fixed the issue preventing builds from succeeding

PR #2446, opened 2021-04-16
[IA-2645] Delete apps when a workspace is deleted

jdcanas:
Currently, users are having issues because if a workspace is deleted without their apps being deleted the apps are orphaned, accruing cost, and not accessible by the user in any way.

We decided to detect this at workspace deletion time, and not allow workspace deletion if apps would be orphaned. We can auto-delete these apps if they are in 'RUNNING' or 'ERROR' status, but otherwise user intervention is required.

All of these changes are contained in the DeleteWorkspaceModal. Included below are screenshots of the various conditions

Delete with no apps existing (current use case):
deleteWithNoApps

Delete with non deletable apps detected (user action required before deletion). Note we remove all subsequent pieces that ask for user input when requesting this specific action from them:
deleteWithNonDeletableApps

Delete with only auto-deletable apps (no user action required, submitting will cause these apps to be deleted for them):
deleteWithDeletableApps

Delete with only auto-deletable apps with confirmation typed:
deleteWithDeletableAppsConfirmation

The network console when this modal is submitted with 5 auto-deletable apps detected (the requests seem to appear twice, but one of them is the pre-flight `options` method submitted by fetch. The `delete` action is submitted only once per app)
deleteCalled5Times

### DIFF
--- a/src/components/runtime-common.js
+++ b/src/components/runtime-common.js
@@ -124,10 +124,10 @@ export const SaveFilesHelp = () => {
 }
 
 export const GalaxyWarning = () => {
-  return div({ style: { fontWeight: 600 } },
-    [p('Important: Please keep this tab open and logged in to Terra while using Galaxy.'),
-      p('Galaxy will open in a new tab. ')]
-  )
+  return h(Fragment, [
+    p({ style: { fontWeight: 600 } }, 'Important: Please keep this tab open and logged in to Terra while using Galaxy.'),
+    p('Galaxy will open in a new tab. ')
+  ])
 }
 
 export const GalaxyLaunchButton = ({ app, onClick, ...props }) => {
--- a/src/components/runtime-common.js
+++ b/src/components/runtime-common.js
@@ -124,7 +124,10 @@ export const SaveFilesHelp = () => {
 }
 
 export const GalaxyWarning = () => {
-  return p([div({ style: { fontWeight: 600 } }, ['Important: Please keep this tab open and logged in to Terra while using Galaxy.']), ' Galaxy will open in a new tab. '])
+  return div({ style: { fontWeight: 600 } },
+    [p('Important: Please keep this tab open and logged in to Terra while using Galaxy.'),
+      p('Galaxy will open in a new tab. ')]
+  )
 }
 
 export const GalaxyLaunchButton = ({ app, onClick, ...props }) => {
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1233,7 +1233,7 @@ const Apps = signal => ({
     return res.json()
   },
   listWithoutProject: async labels => {
-    const res = await fetchLeo(`api/google/v1/apps?${qs.stringify({ ...labels })}`,
+    const res = await fetchLeo(`api/google/v1/apps?${qs.stringify(labels)}`,
       _.mergeAll([authOpts(), appIdentifier, { signal }]))
     return res.json()
   },
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1232,6 +1232,11 @@ const Apps = signal => ({
       _.mergeAll([authOpts(), appIdentifier, { signal }]))
     return res.json()
   },
+  listWithoutProject: async labels => {
+    const res = await fetchLeo(`api/google/v1/apps?${qs.stringify({ ...labels })}`,
+      _.mergeAll([authOpts(), appIdentifier, { signal }]))
+    return res.json()
+  },
   app: (project, name) => {
     const root = `api/google/v1/apps/${project}/${name}`
     return {
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -176,7 +176,7 @@ export const collapsedRuntimeStatus = runtime => {
   return runtime && (runtime.patchInProgress ? 'LeoReconfiguring' : runtime.status) // NOTE: preserves null vs undefined
 }
 
-export const isAppDeletable = app => app.status === 'RUNNING' || app.status === 'ERROR'
+export const isAppDeletable = app => app && app.status === 'RUNNING' || app.status === 'ERROR'
 
 export const convertedAppStatus = appStatus => {
   return Utils.switchCase(_.upperCase(appStatus),
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -176,6 +176,8 @@ export const collapsedRuntimeStatus = runtime => {
   return runtime && (runtime.patchInProgress ? 'LeoReconfiguring' : runtime.status) // NOTE: preserves null vs undefined
 }
 
+export const isAppDeletable = app => app.status === 'RUNNING' || app.status === 'ERROR'
+
 export const convertedAppStatus = appStatus => {
   return Utils.switchCase(_.upperCase(appStatus),
     ['STOPPED', () => _.capitalize('PAUSED')],
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -176,7 +176,7 @@ export const collapsedRuntimeStatus = runtime => {
   return runtime && (runtime.patchInProgress ? 'LeoReconfiguring' : runtime.status) // NOTE: preserves null vs undefined
 }
 
-export const isAppDeletable = app => app && app.status === 'RUNNING' || app.status === 'ERROR'
+export const isAppDeletable = app => app && (app.status === 'RUNNING' || app.status === 'ERROR')
 
 export const convertedAppStatus = appStatus => {
   return Utils.switchCase(_.upperCase(appStatus),
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -176,7 +176,7 @@ export const collapsedRuntimeStatus = runtime => {
   return runtime && (runtime.patchInProgress ? 'LeoReconfiguring' : runtime.status) // NOTE: preserves null vs undefined
 }
 
-export const isAppDeletable = app => app && (app.status === 'RUNNING' || app.status === 'ERROR')
+export const isAppDeletable = app => _.includes(app?.status, ['RUNNING', 'ERROR'])
 
 export const convertedAppStatus = appStatus => {
   return Utils.switchCase(_.upperCase(appStatus),
--- a/src/pages/workspaces/workspace/DeleteWorkspaceModal.js
+++ b/src/pages/workspaces/workspace/DeleteWorkspaceModal.js
@@ -14,30 +14,33 @@ import { isAppDeletable } from 'src/libs/runtime-utils'
 import * as Utils from 'src/libs/utils'
 
 
-const LoadApps = workspaceName => {
-  const signal = Utils.useCancellation()
-  const [deletableApps, setDeletableApps] = useState()
-  const [nonDeletableApps, setNonDeletableApps] = useState()
-
-  const load = async () => {
-    const [currentWorkspaceAppList] = await Promise.all([
-      Ajax(signal).Apps.listWithoutProject({ creator: getUser().email, saturnWorkspaceName: workspaceName })
-    ])
-    const appPartition = _.partition(isAppDeletable, currentWorkspaceAppList)
-    setDeletableApps(appPartition[0])
-    setNonDeletableApps(appPartition[1])
-  }
-
-  Utils.useOnMount(() => {
-    load()
-  })
-
-  return { deletableApps, nonDeletableApps }
-}
-
 const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucketName } }, onDismiss, onSuccess }) => {
   const [deleting, setDeleting] = useState(false)
   const [deleteConfirmation, setDeleteConfirmation] = useState('')
+  const [loadingApps, setLoadingApps] = useState(true)
+
+  const LoadApps = workspaceName => {
+    const signal = Utils.useCancellation()
+    const [deletableApps, setDeletableApps] = useState()
+    const [nonDeletableApps, setNonDeletableApps] = useState()
+
+    const load = async () => {
+      const [currentWorkspaceAppList] = await Promise.all([
+        Ajax(signal).Apps.listWithoutProject({ creator: getUser().email, saturnWorkspaceName: workspaceName })
+      ])
+      const appPartition = _.partition(isAppDeletable, currentWorkspaceAppList)
+      setDeletableApps(appPartition[0])
+      setNonDeletableApps(appPartition[1])
+      setLoadingApps(false)
+    }
+
+    Utils.useOnMount(() => {
+      load()
+    })
+
+    return { deletableApps, nonDeletableApps }
+  }
+
   const { deletableApps, nonDeletableApps } = LoadApps(name)
 
   const getAppDeletionMessage = (deletableApps, nonDeletableApps) => {
@@ -116,7 +119,7 @@ const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucke
         onChange: setDeleteConfirmation
       })
     ]),
-    deleting && spinnerOverlay
+    (deleting || loadingApps) && spinnerOverlay
   ])
 }
 
--- a/src/pages/workspaces/workspace/DeleteWorkspaceModal.js
+++ b/src/pages/workspaces/workspace/DeleteWorkspaceModal.js
@@ -1,42 +1,60 @@
 import _ from 'lodash/fp'
 import { useState } from 'react'
-import { div, h, label, span, p } from 'react-hyperscript-helpers'
+import { div, h, label, p, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, Link, spinnerOverlay } from 'src/components/common'
+import { warningBoxStyle } from 'src/components/data/data-utils'
+import { icon } from 'src/components/icons'
 import { TextInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
 import { Ajax } from 'src/libs/ajax'
 import { bucketBrowserUrl, getUser } from 'src/libs/auth'
+import colors from 'src/libs/colors'
 import { reportError } from 'src/libs/error'
 import * as Utils from 'src/libs/utils'
+import { isAppDeletable } from 'src/libs/runtime-utils'
 
 
 const LoadApps = workspaceName => {
   const signal = Utils.useCancellation()
-  const [apps, setApps] = useState()
+  const [deletableApps, setDeletableApps] = useState()
+  const [nonDeletableApps, setNonDeletableApps] = useState()
 
   const load = async () => {
     const [currentWorkspaceAppList] = await Promise.all([
       Ajax(signal).Apps.listWithoutProject({ creator: getUser().email, saturnWorkspaceName: workspaceName })
     ])
-    setApps(currentWorkspaceAppList)
+    const appPartition = _.partition(isAppDeletable, currentWorkspaceAppList)
+    setDeletableApps(appPartition[0])
+    setNonDeletableApps(appPartition[1])
   }
 
   Utils.useOnMount(() => {
     load()
   })
 
-  return { apps }
+  return { deletableApps, nonDeletableApps }
 }
 
 const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucketName } }, onDismiss, onSuccess }) => {
   const [deleting, setDeleting] = useState(false)
   const [deleteConfirmation, setDeleteConfirmation] = useState('')
-  const { apps } = LoadApps(name)
+  const { deletableApps, nonDeletableApps } = LoadApps(name)
 
-  const getDeletableApps = apps => _.filter(app => app.status === 'RUNNING' || app.status === 'ERROR', apps)
-  const getAppCountMessage = apps => {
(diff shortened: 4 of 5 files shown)


src/components/runtime-common.js
```diff
@@ -124,7 +124,10 @@ export const SaveFilesHelp = () => {
 }
 
 export const GalaxyWarning = () => {
-  return p([div({ style: { fontWeight: 600 } }, ['Important: Please keep this tab open and logged in to Terra while using Galaxy.']), ' Galaxy will open in a new tab. '])
+  return div({ style: { fontWeight: 600 } },
```
jdcanas: This is invalid html I found when digging around. `<p>` tags can have no children but text.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,70 @@
...
 
 
+const LoadApps = workspaceName => {
+  const signal = Utils.useCancellation()
+  const [deletableApps, setDeletableApps] = useState()
+  const [nonDeletableApps, setNonDeletableApps] = useState()
+
+  const load = async () => {
+    const [currentWorkspaceAppList] = await Promise.all([
+      Ajax(signal).Apps.listWithoutProject({ creator: getUser().email, saturnWorkspaceName: workspaceName })
```
jdcanas: we automatically label all apps with their workspace during creation UI-side, so we can list via this label and omit the project

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,70 @@
...
+      (deletableApps.length > 0 ||
+      nonDeletableApps.length > 0)
+  }
 
   const deleteWorkspace = async () => {
     try {
       setDeleting(true)
       await Ajax().Workspaces.workspace(namespace, name).delete()
+      await Promise.all(
+        _.map(async app => await Ajax().Apps.app(app.googleProject, app.appName).delete(), deletableApps)
```
jdcanas, line 61: this maps each deletable app to an async function which `await`s deletion before exitting. This can then be nested in Promise.all(), which must itself be `await`ed to ensure all async functions exit before continueing. Tested this extensively in console with multiple apps, prints after each delete call, and prints after the Promise.all itself.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -26,12 +73,17 @@ const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucke
...
   return h(Modal, {
     title: 'Delete workspace',
     onDismiss,
     okButton: h(ButtonPrimary, {
-      disabled: _.toLower(deleteConfirmation) !== 'delete workspace',
-      onClick: () => deleteWorkspace()
+      disabled: _.toLower(deleteConfirmation) !== 'delete workspace' || isDeleteDisabledWithApps(deletableApps, nonDeletableApps),
+      onClick: () => deleteWorkspace(),
+      tooltip: _.toLower(deleteConfirmation) !== 'delete workspace' ?
+        isDeleteDisabledWithApps(deletableApps, nonDeletableApps) ? 'You must ensure all apps in this workspace are deletable' : 'You must type the confirmation message' :
```
jdcanas: this nested conditional gives us the 3 different contextual tooltips shown in screenshots. Open to ideas for cleaner syntax...

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -156,6 +156,11 @@ const useCloudEnvironmentPolling = namespace => {
   const [persistentDisks, setPersistentDisks] = useState()
   const [galaxyDataDisks, setGalaxyDataDisks] = useState()
 
+  //clean up timeout on unmount.
+  useEffect(() => {
```
jdcanas: the application had a memory leak before we specify cleanup here. I'm not super familiar with useEffect, but it seems that  the timeout logic should eventually conform to use `useEffect` in some way. Currently the way I'm using `useEffect` is semantically equivalent to `Utils.useOnUnmount`. The effect method of cleanup is apparently preferred, since in its [description](https://reactjs.org/docs/hooks-effect.html#example-using-classes-1) it discourages the use of onMount/onUnmount cc @DataBiosphere/terra-devs

jdcanas: Hmm.. this doesn't seem to be quite correct... need to do some further digging.. probably will revert this and do it in a separate PR. I think this should still be refactored to use `useEffect` for proper cleanup...

jdcanas: Ahh... it seems useOnMount is an alias for useEffect.... may not actually be an issue... though I wonder why its still reporting a memory leak from around here...

jdcanas: Hmm.. it seems we are passing the function to set up this timeout around to a lot of components, only ensuring we clean up the timer when the wrapper component is unmounted... This wouldn't be a problem if the components using the timeout didn't try to re-render after they are unmounted. definitely a problem for another PR

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,73 @@
...
+import colors from 'src/libs/colors'
 import { reportError } from 'src/libs/error'
+import { isAppDeletable } from 'src/libs/runtime-utils'
 import * as Utils from 'src/libs/utils'
 
 
 const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucketName } }, onDismiss, onSuccess }) => {
   const [deleting, setDeleting] = useState(false)
   const [deleteConfirmation, setDeleteConfirmation] = useState('')
+  const [loadingApps, setLoadingApps] = useState(true)
```
jdcanas, line 20: this is used to ensure the spinner stays present until the apps are loaded, so the user can't 'quickly delete' it before we detect apps that cant be deleted. It's not super relevant in the happy path where the call returns almost instantly, but if you modify your network speed to 3G or the like in the console and remove this its quite jarring

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,73 @@
...
+  const LoadApps = workspaceName => {
+    const signal = Utils.useCancellation()
+    const [deletableApps, setDeletableApps] = useState()
+    const [nonDeletableApps, setNonDeletableApps] = useState()
+
+    const load = async () => {
+      const [currentWorkspaceAppList] = await Promise.all([
+        Ajax(signal).Apps.listWithoutProject({ creator: getUser().email, saturnWorkspaceName: workspaceName })
+      ])
+      const appPartition = _.partition(isAppDeletable, currentWorkspaceAppList)
```
ehanna4: This is neat I haven't come across` _.partition` before I don't think, seems like a great use case for it.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,73 @@
...
+import { isAppDeletable } from 'src/libs/runtime-utils'
 import * as Utils from 'src/libs/utils'
 
 
 const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucketName } }, onDismiss, onSuccess }) => {
   const [deleting, setDeleting] = useState(false)
   const [deleteConfirmation, setDeleteConfirmation] = useState('')
+  const [loadingApps, setLoadingApps] = useState(true)
+
+  const LoadApps = workspaceName => {
```
ehanna4: Is `LoadApps` a component within the `DeleteWorkspaceModal` component?

I'm curious about nesting a component, rather than having that state at the top level and just having a function for the loadApps work? This seems to work well when testing on the UI I'm just not sure I've seen that pattern before and am curious if it has some benefit

jdcanas: Hmm.. I played around this a lot initially and I was having trouble setting the state up properly (and getting more familiar with hooks), so I modeled this after how we load apps/runtimes in the WorkspaceCoontainer https://github.com/DataBiosphere/terra-ui/blob/dev/src/pages/workspaces/workspace/WorkspaceContainer.js#L153. 

I can see the benefit in that you can split state logic into smaller components, and then pass that around (for example, multiple components care about refreshingRuntimes). WorkspaceContainer is a bit different in that the state its tracking is more complex and multiple components need to trigger the state update though. 

I took another stab at it, and It seems to be working in a plain function.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
jdcanas: Yeah... one of those things where I knew lodash had to have something to do what I wanted but I had no idea the name of it and dug through all array functions in the docs to find it 😂

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,67 @@
...
+          icon('warning-standard', { size: 19, style: { color: colors.warning(), flex: 'none', marginRight: '0.5rem' } }),
+          'Undeletable Workspace Warning'
+        ]),
+        p({ style: { fontWeight: 'normal' } }, `You cannot delete this workspace because there are ${nonDeletableApps.length} application(s) you must delete first. Only applications in ('ERROR', 'RUNNING') status can be automatically deleted.`)
+      ]) :
+      p({ style: { marginLeft: '1rem', fontWeight: 'bold' } }, `Detected ${deletableApps.length} automatically deletable application(s).`)
+  }
+
+  const hasApps = (deletableApps, nonDeletableApps) => {
+    return deletableApps !== undefined && nonDeletableApps !== undefined &&
```
jdcanas, line 51: for some reason this wasn't working without the explicit undefined check (even with the `?.` syntax).

xuscrbw (approved):
Tested this out locally. Looks good! Emily had some good feedback.

I might need to test this again since I see you are making changes

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,67 @@
...
+
+  const signal = Utils.useCancellation()
+
+  const loadApps = async workspaceName => {
+    const [currentWorkspaceAppList] = await Promise.all([
+      Ajax(signal).Apps.listWithoutProject({ creator: getUser().email, saturnWorkspaceName: workspaceName })
+    ])
+    const appPartition = _.partition(isAppDeletable, currentWorkspaceAppList)
+    setDeletableApps(appPartition[0])
+    setNonDeletableApps(appPartition[1])
```
panentheos: TOL: Instead of pre-partitioning the apps, it might work out better to store the whole list in state, and partition/count in the render path. That would enable some simplification further down, e.g. the message 'getters' could just be computed as inline values as well.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
xuscrbw: How would you have used the ?. in this case. Would it just be:  deletableApps?.length > 0 || nonDeleteableApps?.length >0

ehanna4:
Also it looks like the build is failing due to something this recently merged PR was trying to fix: https://github.com/DataBiosphere/terra-ui/pull/2452 if you merge in dev now _**hopefully**_ the build will work

src/components/runtime-common.js
zarsky-broad: A `p` tag should be able to have any children: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p

Also note that a paragraph element has margins applied to it automatically, but a div does not. You may wish to apply the font weight style to the p itself and remove the div.
(That is, with the current two nested paragraphs, the padding between the lines will be doubled.)

src/components/runtime-common.js
```diff
@@ -124,7 +124,10 @@ export const SaveFilesHelp = () => {
...
+    [p('Important: Please keep this tab open and logged in to Terra while using Galaxy.'),
+      p('Galaxy will open in a new tab. ')]
+  )
```
zarsky-broad: I'd prefer we kept brackets matched on their own lines, here you can see the difference in indentation between lines 128 and 129, even though the elements are siblings. Also, while a string is technically valid, we try to always provide children as an array, for consistency.
All of this is so that developers who are more junior or are new to the codebase will have consistent examples.

src/libs/runtime-utils.js
```diff
@@ -168,6 +168,8 @@ export const collapsedRuntimeStatus = runtime => {
   return runtime && (runtime.patchInProgress ? 'LeoReconfiguring' : runtime.status) // NOTE: preserves null vs undefined
 }
 
+export const isAppDeletable = app => app && (app.status === 'RUNNING' || app.status === 'ERROR')
```
zarsky-broad:
```suggestion
export const isAppDeletable = app => _.includes(app?.status, ['RUNNING', 'ERROR'])
```

src/components/runtime-common.js
zarsky-broad: (this note applies a few more times throughout this PR)

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,67 @@
...
+    setLoadingApps(false)
+  }
+
+  Utils.useOnMount(() => {
+    loadApps()
+  })
+
+  const getAppDeletionMessage = (deletableApps, nonDeletableApps) => {
+    return nonDeletableApps.length > 0 ?
```
zarsky-broad: Generally, when you're checking if it's empty, it may be simpler to use `_.isEmpty`, for example here:
```suggestion
    return !_.isEmpty(nonDeletableApps) ?
```

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
zarsky-broad: +1

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
zarsky-broad: Is there a reason this takes `deletableApps` and `nonDeletableApps` as params, since it has access to them from state?

zarsky-broad:
@jdcanas I'd be happy to pair with you on this, if you'd like, just to work on code style and simplifying logic.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
jdcanas: Ah, they were variables that were state elsewhere first, changed @zarsky-broad .  Yeah Wilson, I tried that and `deleteableApps && ....`

jdcanas:
> @jdcanas I'd be happy to pair with you on this, if you'd like, just to work on code style and simplifying logic.

I understood and implemented the changes you recommend, let me know if they look good. We can chat if we end up having some back and forth

src/components/runtime-common.js
zarsky-broad: What I meant here was that this was valid before, although a bit much to have on one line. You should be able to revert these lines, and maybe just add some line breaks:
```suggestion
  return p([
    div({ style: { fontWeight: 600 } }, ['Important: Please keep this tab open and logged in to Terra while using Galaxy.']),
    'Galaxy will open in a new tab.'
  ])
```

src/libs/ajax.js
```diff
@@ -1232,6 +1232,11 @@ const Apps = signal => ({
       _.mergeAll([authOpts(), appIdentifier, { signal }]))
     return res.json()
   },
+  listWithoutProject: async labels => {
+    const res = await fetchLeo(`api/google/v1/apps?${qs.stringify({ ...labels })}`,
```
```suggestion
    const res = await fetchLeo(`api/google/v1/apps?${qs.stringify(labels)}`,
```

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
zarsky-broad: This seems unnecessary; why not just check if `apps === undefined` to see if they've been loaded yet?

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,72 @@
...
+  const loadApps = async workspaceName => {
+    const [currentWorkspaceAppList] = await Promise.all([
+      Ajax(signal).Apps.listWithoutProject({ creator: getUser().email, saturnWorkspaceName: workspaceName })
+    ])
+    setApps(currentWorkspaceAppList)
+    setLoadingApps(false)
+  }
+
+  Utils.useOnMount(() => {
+    loadApps(name)
```
zarsky-broad, line 33: Since this is only called here, you should move the declaration of `loadApps` into the `useOnMount` function body. This makes the scope where the function is used clearer to other developers.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -26,12 +75,17 @@ const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucke
...
     title: 'Delete workspace',
     onDismiss,
     okButton: h(ButtonPrimary, {
-      disabled: _.toLower(deleteConfirmation) !== 'delete workspace',
-      onClick: () => deleteWorkspace()
+      disabled: _.toLower(deleteConfirmation) !== 'delete workspace' || isDeleteDisabledFromApps(),
+      onClick: () => deleteWorkspace(),
+      tooltip: _.toLower(deleteConfirmation) !== 'delete workspace' ?
+        isDeleteDisabledFromApps() ? 'You must ensure all apps in this workspace are deletable' : 'You must type the confirmation message' :
+        'Delete Workspace'
```
zarsky-broad: Nested ternaries can be very hard to parse. Instead, consider using a `Utils.cond`:
```suggestion
      tooltip: Utils.cond(
        [isDeleteDisabledFromApps(), () => 'You must ensure all apps in this workspace are deletable'],
        [_.toLower(deleteConfirmation) !== 'delete workspace', () => 'You must type the confirmation message'],
        () => 'Permanently delete workspace'
      )
```

Note that I'm not sure the default tooltip is needed at all.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -26,12 +75,17 @@ const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucke
     }
   }
 
+  const isDeleteDisabledFromApps = () => hasApps() && !_.isEmpty(getNonDeletableApps())
```
zarsky-broad: Since this is called repeatedly every render, you should make it a computed value rather than a function and change the places that reference it.
```suggestion
  const isDeleteDisabledFromApps = hasApps() && !_.isEmpty(getNonDeletableApps())
```

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -1,23 +1,72 @@
...
+    setLoadingApps(false)
+  }
+
+  Utils.useOnMount(() => {
+    loadApps(name)
+  })
+
+  const getDeletableApps = () => _.filter(isAppDeletable, apps)
+
+  const getNonDeletableApps = () => _.filter(app => !isAppDeletable(app), apps)
```
zarsky-broad: This iterates through `apps` multiple times. Consider:
```suggestion
  const [deletableApps, nonDeletableApps] = _.partition(isAppDeletable, apps)
```

src/components/runtime-common.js
jdcanas: react complains when you nest tags in a `<p>`, I'd prefer to clean up the console

zarsky-broad: Ah, my bad, I didn't realize React was complaining about it. In that case, I'd still like to clean up the spacing here:
```suggestion
  return h(Fragment, [
    p({ style: { fontWeight: 600 } }, 'Important: Please keep this tab open and logged in to Terra while using Galaxy.'),
    p('Galaxy will open in a new tab.')
  ])
```

Just note that this will still change the spacing between the text from what it was before this PR. I've moved the style so that only the section that's currently bold will be bold.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -26,12 +68,18 @@ const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucke
...
   return h(Modal, {
     title: 'Delete workspace',
     onDismiss,
     okButton: h(ButtonPrimary, {
-      disabled: _.toLower(deleteConfirmation) !== 'delete workspace',
-      onClick: () => deleteWorkspace()
+      disabled: _.toLower(deleteConfirmation) !== 'delete workspace' || isDeleteDisabledFromApps,
+      onClick: () => deleteWorkspace(),
+      tooltip: Utils.cond(
+        [_.toLower(deleteConfirmation) !== 'delete workspace', () => 'You must type the confirmation message'],
```
ehanna4, line 80: I'm seeing the message "You must type the confirmation message" when I have an App in `precreating`, I think you want to flip these cond checks so the App one is first

ehanna4 (approved):
Overall looks good! Tested locally and reviewed code - one small change needed for the tooltip message but otherwise looks good to go to me

PR #2447, opened 2021-04-16
Rename confusing variables in the config in NewClusterModal 

ehanna4:
No Functional Change

Things previously prefixed with "new" are now prefixed by "desired"
Things previously prefixed with "old" now are prefixed with "existing"

Open to different name prefixes but thought these made it more clear what status the pieces of the config are.


jdcanas (approved):
definitely better than new/old naming

Qi77Qi (approved):
👍  thanks @ehanna4 !

kyuksel (approved):
I'm very glad you did the renaming. It'd taken me a while to realize what `old`/`new` meant 😊 

TOL: `requested*` could be an alternative prefix to `desired*`.

PR #2448, opened 2021-04-17
[CA-1312] A11y: Table roles

andrewmarcus:
This PR adds proper table rows to every type of table, to improve the accessibility of many pages within the app. This applies to the following table types:

* SimpleTable
* SimpleFlexTable
* FlexTable
* GridTable

### Table Names

* Each of the 4 table types now supports an optional `tableName` property, which is converted to an `aria-label` on the root table element. If not specified, it will be "data table" by default.

### Table Roles

* Each table has `role="table"` at the top level. This role does not imply that the table supports keyboard navigation. If we go back later and add keyboard navigation in order to make it easier to scroll through the virtualized DOM, then at that point these should be switched to `role="grid"` to imply that [keyboard navigation is possible](https://www.digitala11y.com/grid-role/).
* The header row has `role="row"` and each header column has `role="columnheader"`.
* Each data row has `role="row"` and each data cell has `role="cell"`. If the top-level is changed to `role="grid"` in the future to support arrow-key navigation, then each cell should be changed to `role="gridcell"`.

### Virtualized Tables

In FlexTable and GridTable, not all cells will be included in the DOM. To support this:
* Each cell now includes `aria-rowindex` and `aria-colindex` attributes to indicate where in the table these cells are supposed to be.
* Each of the 3 main table types now also includes consistent `rowIndex` and `columnIndex` properties when calling its respective `cellRenderer` functions. I didn't change existing behavior, but added new optional properties that could be used by table definitions if they wish. These two properties should now be presented consistently regardless of which table type you're using.

### Sorting

Table definitions currently declare `Sortable` headers within the output of the `headerRenderer()` function for each column. However, to properly use the `aria-sort` ARIA attribute, the attribute must be placed directly on the element with `role="columnheader"`, which is several levels higher in the DOM. To support this:
* The `Sortable` and `MiniSortable` renderers now produce columns which are keyboard interactive, and provide an ARIA description indicating that the column will resort itself when clicking on it.
* Each sortable table must now be passed the `sort` state object.
* Each column which supports sorting must now specify the sorting field name, the same way it is defined in the `Sortable` renderer.

_Note:_ I don't like the duplication. It's problematic for a developer to have to add each field name in two places -- in the column definition and again in the `Sortable` definition -- and to keep them in sync. I'd prefer to try to recursively find a `Sortable` object in the `children` array and grab the 'sort' and `field` properties from there. But so far I've been trying to avoid that under the impression that it's bad practice to try to grab attributes from inside hyperscript children arrays. I certainly haven't seen this done anywhere else.

Is there a better way to do this without adding duplicate work for developers?

### Page-Specific updates

I've added duplicate field names to support `aria-sort` to:

* UploadPreviewTable.js (/#upload)
* Environment.js (/#clusters)
* List.js (/#workspaces/list)
* DataTable.js (/#workspaces/data)
* SubmissionDetails.js (/#workspaces/:workspace/jobHistory
* WorkflowView.js (/#workspaces/:workspace/workflows
* List.js (/#workflows/list)

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -198,6 +198,7 @@ const DataTable = props => {
               onScroll,
               initialX,
               initialY,
+              sort,
               columns: [
                 {
                   width: 70,
@@ -206,7 +207,8 @@ const DataTable = props => {
                       h(Checkbox, {
                         checked: pageSelected(),
                         disabled: !entities.length,
-                        onChange: () => pageSelected() ? deselectPage() : selectPage()
+                        onChange: () => pageSelected() ? deselectPage() : selectPage(),
+                        'aria-label': 'Select all'
                       }),
                       h(PopupTrigger, {
                         closeOnClick: true,
@@ -233,6 +235,7 @@ const DataTable = props => {
                   }
                 },
                 {
+                  field: 'name',
                   width: nameWidth,
                   headerRenderer: () => h(Resizable, {
                     width: nameWidth, onWidthChange: delta => {
@@ -259,6 +262,7 @@ const DataTable = props => {
                   const thisWidth = columnWidths[name] || 300
                   const [, columnNamespace, columnName] = /(.+:)?(.+)/.exec(name)
                   return {
+                    field: name,
                     width: thisWidth,
                     headerRenderer: () => h(Resizable, {
                       width: thisWidth, onWidthChange: delta => setColumnWidths(_.set(name, thisWidth + delta))
--- a/src/components/table.js
+++ b/src/components/table.js
@@ -267,12 +267,18 @@ FlexTable.propTypes = {
  * since it does not provide scrolling. See FlexTable for prop types.
  */
 export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, hoverHighlight }) => {
-  return h(Fragment, [
-    div({ style: { height: 48, display: 'flex' } }, [
+  return div({
+    role: 'grid',
+  }, [
+    div({
+      style: { height: 48, display: 'flex' },
+      role: 'row'
+    }, [
       _.map(([i, { size, headerRenderer }]) => {
         return div({
           key: i,
-          style: { ...styles.flexCell(size), ...styles.header(i * 1, columns.length) }
+          style: { ...styles.flexCell(size), ...styles.header(i * 1, columns.length) },
+          role: 'columnheader'
         }, [headerRenderer()])
       }, Utils.toIndexPairs(columns))
     ]),
@@ -281,6 +287,7 @@ export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, hoverHigh
         key: rowIndex,
         as: 'div',
         className: 'table-row',
+        role: 'row',
         style: { backgroundColor: 'white', display: 'flex', minHeight: 48 },
         hover: hoverHighlight ? { backgroundColor: colors.light(0.4) } : undefined
       }, [
@@ -288,6 +295,7 @@ export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, hoverHigh
           return div({
             key: i,
             className: 'table-cell',
+            role: 'gridcell',
             style: { ...styles.flexCell(size), ...styles.cell(i * 1, columns.length) }
           }, [cellRenderer({ rowIndex })])
         }, Utils.toIndexPairs(columns))
@@ -335,7 +343,11 @@ export const GridTable = Utils.forwardRefWithName('GridTable', ({
     ref: scrollSync
   }, [
     ({ onScroll, scrollLeft }) => {
-      return div([
+      return div({
+        role: 'grid',
+        'aria-rowcount': rowCount + 1, // count the header row too
+        'aria-colcount': columns.length
+      }, [
         h(RVGrid, {
           ref: header,
           width: width - scrollbarSize,
@@ -344,6 +356,9 @@ export const GridTable = Utils.forwardRefWithName('GridTable', ({
           rowHeight: headerHeight,
           rowCount: 1,
           columnCount: columns.length,
+          role: 'rowgroup',
+          containerRole: 'row',
+          'aria-label': '',
           cellRenderer: data => {
             return div({
               key: data.key,
@@ -352,7 +367,10 @@ export const GridTable = Utils.forwardRefWithName('GridTable', ({
                 ...data.style,
                 ...styles.header(data.columnIndex, columns.length),
                 ...styleHeader(data)
-              }
+              },
+              role: 'columnheader',
+              'aria-colindex': data.columnIndex,
+              'aria-rowindex': 1, // the rowindex property must start at 1
             }, [
               columns[data.columnIndex].headerRenderer(data)
             ])
@@ -369,6 +387,9 @@ export const GridTable = Utils.forwardRefWithName('GridTable', ({
           rowHeight,
           rowCount,
           columnCount: columns.length,
+          role: 'rowgroup',
+          containerRole: 'presentation',
+          'aria-label': '',
           noContentRenderer: () => div({ style: { marginTop: '1rem', textAlign: 'center', fontStyle: 'italic' } }, [noContentMessage]),
           onScrollbarPresenceChange: ({ vertical, size }) => {
             setScrollbarSize(vertical ? size : 0)
@@ -377,6 +398,9 @@ export const GridTable = Utils.forwardRefWithName('GridTable', ({
             return div({
               key: data.key,
               className: 'table-cell',
+              role: 'gridcell',
+              'aria-colindex': data.columnIndex,
+              'aria-rowindex': data.rowIndex + 2, // The header row is 1
               style: {
                 ...data.style,
                 ...styles.cell(data.columnIndex, columns.length),
--- a/src/components/table.js
+++ b/src/components/table.js
@@ -499,10 +499,11 @@ GridTable.propTypes = {
   })
 }
 
-export const SimpleTable = ({ columns, rows }) => {
+export const SimpleTable = ({ columns, rows, tableName = 'data table' }) => {
   const cellStyles = { paddingTop: '0.25rem', paddingBottom: '0.25rem' }
   return h(div, {
-    role: 'table'
+    role: 'table',
+    'aria-label': tableName
   }, [
     div({ style: { display: 'flex' } }, [
       _.map(({ key, header, size }) => {
--- a/src/components/table.js
+++ b/src/components/table.js
@@ -6,7 +6,7 @@ import Draggable from 'react-draggable'
 import { button, div, h, label, option, select } from 'react-hyperscript-helpers'
 import Pagination from 'react-paginating'
 import { SortableContainer, SortableElement, SortableHandle } from 'react-sortable-hoc'
-import { AutoSizer, Grid as RVGrid, List, ScrollSync as RVScrollSync } from 'react-virtualized'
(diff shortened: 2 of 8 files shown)


src/components/data/DataTable.js
```diff
@@ -233,6 +235,7 @@ const DataTable = props => {
                   }
                 },
                 {
+                  field: 'name',
```
marctalbott, line 238: Are the single quotes around `name` here intentional? If so, what's the difference between this field and the other fields where name isn't enclosed in single quotes?

src/components/table.js
```diff
@@ -259,37 +277,52 @@ FlexTable.propTypes = {
...
+      role: 'row'
+    }, [
+      _.map(([i, { size, headerRenderer, field }]) => {
         return div({
           key: i,
-          style: { ...styles.flexCell(size), ...styles.header(i * 1, columns.length) }
-        }, [headerRenderer()])
+          style: { ...styles.flexCell(size), ...styles.header(i * 1, columns.length) },
+          role: 'columnheader',
+          'aria-sort': !sort || !field ? null : sort.field === field ? sort.direction === 'asc' ? 'ascending' : 'descending' : 'none'
```
marctalbott: Is there a way to extract this out into a variable? I see it is also used in other tables

src/components/table.js
```diff
@@ -177,25 +178,33 @@ export const FlexTable = ({
     body.current.scrollToPosition({ scrollTop: initialY })
   })
 
-  return div([
+  return div({
+    role: 'table',
+    'aria-rowcount': rowCount,
```
marctalbott: Not sure if this is how it's supposed to be, but when I was testing this out locally, I found that this `rowCount` doesn't include the column headers, but the row indices do. Should those match up? It seems inconsistent to have a table report that there are 2 rows, but then have the index of the last row be 3

src/components/data/DataTable.js
andrewmarcus: No difference. I agree they might as well be consistent. Is there a preferred quote type (single vs double) in the repo?

src/components/table.js
andrewmarcus: Hmm. I could calculate it each time within the renderer function and then just use the value here, if that would be preferred. I suppose I could also make this a shared function for all 3 tables, but it would basically be a one-liner.  Probably shouldn't memoize it, though, since the whole point of these virtual tables is to not have the entire thing in memory at a time...

What I would ideally prefer to do is extract it from the `Sortable` renderer nested inside the results from the `headerRenderer()` call. See comments above about avoiding duplication...  I'm just not sure whether it's good practice to introspect hyperscript objects. If you think that's ok, I can remove the `field` property here entirely.

src/components/table.js
andrewmarcus: Good catch!

src/components/data/DataTable.js
andrewmarcus: Ah. Here the name of the field is `'name'` to match up with the `Sortable` definition attached to the column. In the situation below, `name` is a variable attached to each column created by the iterator

src/components/table.js
s-rubenstein: I would probably extract it into a function, sort of like:
`(sort, field) => !sort || !field ? null : sort.field === field ? sort.direction === 'asc' ? 'ascending' : 'descending' : 'none'`, and then each place this is duplicated, we can just call that function. I think it could make things a little more readable.

src/components/data/DataTable.js
andrewmarcus: I agree this is confusing, but I just pulled up the column names that were already in the corresponding `Sortable` definitions.

src/components/table.js
```diff
@@ -161,14 +161,18 @@ const styles = {
 // Note: We always need 1 extra row's worth of height for the table header row:
 export const tableHeight = ({ actualRows, maxRows, heightPerRow = 48 }) => (_.min([actualRows, maxRows]) + 1) * heightPerRow
 
+const ariaSort = (sort, field) => !sort || !field ? null :
+  sort.field !== field ? 'none' : sort.direction === 'asc' ? 'ascending' : 'descending'
+
```
andy7i: I had some trouble parsing this. Would it be possible to refactor this to make it more readable? Maybe not really possible but here are a couple suggestions:

* I think the ordering of negative case first and positive case last puts the most important piece when it might be easier if it came first. Not exactly the same but related: https://softwareengineering.stackexchange.com/questions/223086/should-if-else-statements-be-arranged-by-rareness-of-cases-or-difficulty-of-deal/223093
  * I'd be ok with keeping the current order if you feel strongly about it

* There are a lot of `!`s to keep track of. There's a comment about that in the link above
> Personally I would concentrate on avoiding not conditions. They have been proven to induce more cognitive load to understand than "positive" conditions and are detrimental to readability / understandability of code. I tend to use them in guard statements only

* the ternary chain is hard to follow. I think ternary operators are best for simple one-liners. Maybe it's also the fact that it's all on 2 lines.

Taking all of that into consideration, what do you think about this? I _think_ it should be functionally equivalent but can guarantee nothing!

```suggestion
const ariaSort = (sort, field) => {
  if (!!sort || !!field)
    if (field == sort.field)
      if (sort.direction === 'asc')
        'ascending'
      else 'descending'
    else 'none'
  else null
}
```

src/components/table.js
```diff
@@ -212,20 +228,24 @@ export const FlexTable = ({
...
           style: { ...data.style, backgroundColor: 'white', display: 'flex' },
           hover: hoverHighlight ? { backgroundColor: colors.light(0.4) } : undefined
         }, [
           _.map(([i, { size, cellRenderer }]) => {
             return div({
               key: i,
               className: 'table-cell',
+              role: 'cell',
+              'aria-rowindex': data.rowIndex + 2, // The header row is 1
+              'aria-colindex': i + 1,
```
andy7i:
```suggestion
              'aria-colindex': i + 1, // aria-colindex must be ≥ 1 https://www.w3.org/TR/wai-aria-1.1/#aria-colindex
```

src/components/table.js
```diff
@@ -259,37 +280,52 @@ FlexTable.propTypes = {
...
 
 /**
  * A basic table with a header and flexible column widths. Intended for small amounts of data,
  * since it does not provide scrolling. See FlexTable for prop types.
  */
-export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, hoverHighlight }) => {
-  return h(Fragment, [
-    div({ style: { height: 48, display: 'flex' } }, [
-      _.map(([i, { size, headerRenderer }]) => {
+export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, hoverHighlight, tableName = 'data table', sort = null }) => {
```
andy7i, line 320: I'm worried that it might be easy to forget to put a descriptive name in `tableName` if we provide a default value here.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -108,9 +108,11 @@ const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange,
   return h(SimpleFlexTable, {
     rowCount: sortedData.length,
     noContentMessage: `No matching ${which}.`,
+    sort,
```
andy7i, line 111: is `data table` the right description for this table? (the answer might be yes, but I just wanted to double check)

src/components/table.js
```diff
@@ -259,37 +280,52 @@ FlexTable.propTypes = {
...
+  return div({
```
andy7i, line 321: Why is the div better than the `h(Fragment` here? -- just curious

src/components/table.js
andy7i: I think you have similar comments like this in other places. Maybe worth adding that in the other places where we offset the `colindex` and `rowindex` since folks aren't super familiar with aria rules.

src/components/table.js
andy7i: same goes for the other 3 tables

src/components/table.js
```diff
@@ -457,31 +554,35 @@ export const HeaderCell = props => {
 }
 
 export const Sortable = ({ sort, field, onSort, children }) => {
-  return div({
-    style: { flex: 1, display: 'flex', alignItems: 'center', cursor: 'pointer', width: '100%' },
-    onClick: () => onSort(Utils.nextSort(sort, field))
+  return h(IdContainer, [id => h(Clickable, {
```
andy7i, line 584: where does `id` come from?

andy7i (commented):
Hey Andrew! Thanks for helping us do this important work! I appreciated the refactoring and comments you added.
I had some questions to try to understand some of the changes better (I haven't done too much front end dev recently) as well as some nitpicks. Let me know if I can help clarify anything!

src/components/table.js
andrewmarcus: I am in favor of making this as easy to understand as possible! Glad to reformat. I'll add some comments too. :)

src/components/table.js
andrewmarcus: Hmm, that's a great point. I was trying to keep these changes as unobtrusive to developers as possible, and trying to avoid having to make changes to every table.

But on the other hand, I already changed all the other tables to add support for `aria-sort`, so that's probably not a big deal at this point.

And it would certainly be better to force devs to explicitly think about accessibility where they need to. Now that I think about it, this caption is sort of like adding `alt` text to an image: sure you could just use a default but that really defeats the purpose.

Ok, I'll make it a required field, and update all of the tables accordingly.

src/pages/workspaces/workspace/workflows/WorkflowView.js
andrewmarcus: I was trying to keep the PRs isolated to either core functionality or specific pages, but I'll go ahead and make the `tableName` required, which means I'll also add accessible names to every table including this one.

src/components/table.js
andrewmarcus: It seemed like `Fragment` doesn't actually render any markup on its own, it's just a container if you have more than one child. I changed it to a `div` since I needed to add ARIA properties and needed an actual tag to attach them to.

src/components/table.js
andrewmarcus: The `IdContainer` generates stateful unique ids and passes them down to its children. I just copied and pasted an existing design pattern I found in use throughout the site.

I'm not entirely sure how that `id` gets passed in there, and what would happen if I had more than one child in that array; would they each be called with the `id` parameter or would only the first one be? But it seems to do what I want when written like this.

src/components/table.js
zarsky-broad: Might a `Utils.cond` be simpler?

src/components/data/DataTable.js
marctalbott: Ah okay I think it's fine then, just wanted to double check that this wasn't meant to be a variable like it was down below

src/components/table.js
andrewmarcus: This thread is outdated, I already updated it to a ternary inside some ifs. That also let me put comments in describing the various cases, which seemed useful...

src/pages/workspaces/workspace/workflows/WorkflowView.js
andrewmarcus: I confirmed with @andy7i that I'm going to keep `tableName` optional in this PR. I will publish a new PR in which I make it required and add table names to the 15+ tables which need it, in order to keep all those changes out of this PR.

src/components/data/DataTable.js
andrewmarcus: Yeah, it's a good question. Thanks for double checking

marctalbott (approved):

andy7i (approved):
👍 👍  Looks great! Thank you

PR #2450, opened 2021-04-21
[CA-1306] A11y: Workflows pages

andrewmarcus:
This PR addresses accessibility issues on workflows pages.

## Summary of changes

* Marked up the TabBar widget as role=tablist
* Adjusted Prism code editor theme colors to increase contrast to 4.5:1
* Added a label to the snapshot selector input
* Added a "main" landmark surrounding the Workflows page contents
* Converted dashboard headers to actual h2 tags

## Notes and Questions

How do users actually navigate to this set of pages? I don't see it in the main menu, and the Workflows tab in the Workspaces section appears to not go here directly.

I've changed the Prism theme colors directly in the style.css file with `!important` overrides. However, it appears that Prism is based on a theme system and ships with a number of built-in themes, so perhaps it's also possible to define a custom theme? Would that be a better solution than what I did here?

How would I see an example of the code editor not being readonly? In every one of the handful of workflows I looked at on the dev server, nothing seemed to be editable...

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -127,3 +127,27 @@ a {
   animation: progress-bar-stripes 1s linear infinite;
 }
 
+/* Prism theme overrides to increase contrast for accessibility */
+.token.punctuation {
+  color: #767676 !important; /* #999; */
+}
+
+.token.selector,
+.token.attr-name,
+.token.string,
+.token.char,
+.token.builtin,
+.token.inserted {
+  color: #518400 !important; /* #690; */
+}
+
+.token.function,
+.token.class-name {
+  color: #d3405e !important; /* #DD4A68; */
+}
+
+.token.regex,
+.token.important,
+.token.variable {
+  color: #b56000 !important; /* #e90; */
+}
--- a/src/components/WDLViewer.js
+++ b/src/components/WDLViewer.js
@@ -75,6 +75,7 @@ const WDLViewer = ({ wdl, ...props }) => {
 
   return pre(_.merge(
     {
+      tabIndex: 0,
       className: 'line-numbers',
       style: { border: Style.standardLine, backgroundColor: 'white' }
     },
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -119,11 +119,15 @@ export const ButtonOutline = ({ disabled, children, ...props }) => {
 }
 
 export const TabBar = ({ activeTab, tabNames, displayNames = {}, refresh = _.noop, getHref, getOnClick = _.noop, children }) => {
-  const navTab = currentTab => {
+  const navTab = (i, currentTab) => {
     const selected = currentTab === activeTab
     const href = getHref(currentTab)
 
     return h(Clickable, {
+      role: 'tab',
+      'aria-posinset': i + 1,
+      'aria-setsize': tabNames.length,
+      'aria-selected': selected,
       style: { ...Style.tabBar.tab, ...(selected ? Style.tabBar.active : {}) },
       hover: selected ? {} : Style.tabBar.hover,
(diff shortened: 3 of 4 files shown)


s-rubenstein (approved):
This looks good to me. I believe you get to the workflow details page by running a workflow and then clicking into it. For the WDLViewer page, I believe it is when you go to import a workflow from the methods repository, but I am not certain.

src/style.css
```diff
@@ -127,3 +127,27 @@ a {
   animation: progress-bar-stripes 1s linear infinite;
 }
 
+/* Prism theme overrides to increase contrast for accessibility */
```
andy7i, line 130: >I've changed the Prism theme colors directly in the style.css file with !important overrides. However, it appears that Prism is based on a theme system and ships with a number of built-in themes, so perhaps it's also possible to define a custom theme? Would that be a better solution than what I did here?


It looks like there is a way to create a custom theme (http://k88hudson.github.io/syntax-highlighting-theme-generator/www/) and there are some a11y-compliant themes (https://github.com/ericwbailey/a11y-syntax-highlighting) but I don't have a strong preference with any of these 3 approaches. Any thoughts from you or @s-rubenstein ?

andy7i (approved):
LGTM, did not do any manual testing. One question but not blocking

to answer your questions:

>How do users actually navigate to this set of pages? I don't see it in the main menu, and the Workflows tab in the Workspaces section appears to not go here directly.

I asked around and it sounds like the `Workflows` list is not a completed feature yet, and so it is not linked anywhere in Terra. I think you're right in that it can only be accessed by manually pointing the browser to the URL https://bvdp-saturn-dev.appspot.com/#workflows . Similarly, `WorkflowDetails` seems to only be used by the `Workflows` list page. 

>I've changed the Prism theme colors directly in the style.css file with !important overrides. However, it appears that Prism is based on a theme system and ships with a number of built-in themes, so perhaps it's also possible to define a custom theme? Would that be a better solution than what I did here?

Answered this in a code comment so it's easier to reply to.

>How would I see an example of the code editor not being readonly? In every one of the handful of workflows I looked at on the dev server, nothing seemed to be editable...

I didn't see anything for editing wdl workflows either. I think users would go to the firecloud method repo to make changes to a wdl.

src/style.css
andrewmarcus: Hmm, I like that Eric Bailey one, and the fact that it has support for high-contrast mode. That's a feature we should be considering as well.

The main thing, though, if I switch to a different theme, is that the colors could change rather dramatically. I'm not sure anyone would notice or care, but that's definitely a thing to think about. In this PR I've just taken the existing colors and darkened them according to ANDI's recommendations to get them just barely to 4.5:1

s-rubenstein: Yeah, I think changing to a new theme would be a larger conversation with UX. There's a lot of consideration that has gone into our themes color design both from UX, but also likely from project partners. My guess is there are some theme requirements.

andrewmarcus: To be clear, we're just talking about the color scheme of the code editor, not the site theme overall.

Do you think the adjustments I made will be ok without a review from UX? The most obvious one is that the orange has turned into a sort of orangey-brown. The other color switches are pretty subtle.

zarsky-broad: Fwiw, I think we just... _picked one_.

zarsky-broad: (As such I think whatever you did is fine @andrewmarcus)

s-rubenstein: Ohhhhh… I didn't realize only the code editor. I think what you did is fine, and then we can re-evaluate if the workflows teams want to change it longer term.

andy7i: ok, that all sounds good, thanks for the input!

PR #2454, opened 2021-04-21
add sample tsv to data import modal

Shakespeared:
Adds a downloadable sample TSV file and a link to a terra support page from the 'import table data' modal on the data page. 

Also adds mixpanel tracking for downloading the example TSV and for uploading data through the modal. 

Tested in the UI and verified events in mixpanel.

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -17,6 +17,7 @@ import ReferenceData from 'src/data/reference-data'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { reportError } from 'src/libs/error'
+import Events from 'src/libs/events'
 import { FormLabel } from 'src/libs/forms'
 import { requesterPaysProjectStore } from 'src/libs/state'
 import * as StateHistory from 'src/libs/state-history'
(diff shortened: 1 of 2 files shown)


src/components/data/data-utils.js
```diff
@@ -212,6 +213,9 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
       const workspace = Ajax().Workspaces.workspace(namespace, name)
       await (useFireCloudDataModel ? workspace.importEntitiesFile : workspace.importFlexibleEntitiesFile)(file)
       onSuccess()
+      Ajax().Metrics.captureEvent(Events.workspaceDataUpload, {
```
ansingh7115, line 216: Intellij gives me a warning here: "Missing await for an async function call "

src/components/data/data-utils.js
```diff
@@ -212,6 +213,9 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
...
+        workspaceNamespace: namespace, workspaceName: name
+      })
```
davidangb, line 218: noting that we only log this event if the upload was successful. I assume that's your intention!

davidangb (approved):
lgtm! one minor comment inline, not a blocker

src/components/data/data-utils.js
Shakespeared: @davidangb yes, that was my intention! If you disagree with that implementation I can change it, but I figured we wouldn't want to duplicate events if someone tries to upload a bunch of times with something not formatted correctly, etc.

src/components/data/data-utils.js
Shakespeared: Yeah, that's on most of the mixpanel implementations, it's a known condition but isn't an issue. It happens when the mixpanel call is inside an async function even though the mixpanel event itself doesn't need to be awaited.

src/components/data/data-utils.js
davidangb: that works for me! If our business folks decide they want to track failures too, we can always change it later.

sagehen03 (approved):
:+1:

PR #2455, opened 2021-04-22
retry failed tests [QA-1437]

zarsky-broad:
This adds the ability to retry failed tests in jest, which uses a test runner called `jest-circus`. This will be the default test runner in the next version of Jest. See [here](https://github.com/facebook/jest/tree/master/packages/jest-circus) and [here](https://jestjs.io/docs/jest-object#jestretrytimes) for documentation.

Ran integration tests locally against alpha, which passed after retrying a failed test!

### DIFF
--- a/integration-tests/jest/1-run-notebook.integration-test.js
+++ b/integration-tests/jest/1-run-notebook.integration-test.js
@@ -2,4 +2,4 @@ const { registerTest } = require('./jest-utils')
 const { testRunNotebook } = require('../tests/run-notebook')
 
 
-registerTest(testRunNotebook)
+registerTest(testRunNotebook, 1)
--- a/integration-tests/jest/1-run-workflow-on-snapshot.integration-test.js
+++ b/integration-tests/jest/1-run-workflow-on-snapshot.integration-test.js
@@ -2,4 +2,4 @@ const { registerTest } = require('./jest-utils')
 const { testRunWorkflowOnSnapshot } = require('../tests/run-workflow-on-snapshot')
 
 
-registerTest(testRunWorkflowOnSnapshot)
+registerTest(testRunWorkflowOnSnapshot, 1)
--- a/integration-tests/jest/1-run-workflow.integration-test.js
+++ b/integration-tests/jest/1-run-workflow.integration-test.js
@@ -7,4 +7,4 @@ const { registerTest } = require('./jest-utils')
 const { testRunWorkflow } = require('../tests/run-workflow')
 
 
-registerTest(testRunWorkflow)
+registerTest(testRunWorkflow, 1)
--- a/integration-tests/package.json
+++ b/integration-tests/package.json
@@ -8,9 +8,10 @@
     "start-dev": "nodemon"
   },
   "devDependencies": {
-    "jest": "^26.5.2",
+    "jest": "^26.6.3",
+    "jest-circus": "^26.6.3",
     "jest-junit": "^12.0.0",
-    "jest-puppeteer": "^4.4.0",
+    "jest-puppeteer": "^5.0.2",
     "node-fetch": "^2.6.1",
     "prompts": "^2.3.2"
   },
@@ -29,6 +30,6 @@
     "lodash": "^4.17.20",
     "nodemon": "^2.0.4",
     "p-retry": "^4.2.0",
-    "puppeteer": "^5.3.1"
+    "puppeteer": "^9.0.0"
   }
 }
--- a/integration-tests/jest.config.js
+++ b/integration-tests/jest.config.js
@@ -1,4 +1,5 @@
 module.exports = {
   preset: 'jest-puppeteer',
-  testRegex: '\\.integration-test\\.js$'
+  testRegex: '\\.integration-test\\.js$',
+  testRunner: 'jest-circus/runner'
 }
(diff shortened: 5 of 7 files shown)


breilly2:
I don't like this for several reasons:

1. Can mask real unreliability that users could experience.
2. Increases test suite run time, especially if the number of test cases grows.
3. Allows new tests may be written to a lower standard of reliability, reducing the effectiveness of the retries.

For #2 in particular, Cromwell is currently feeling this pain and we're actively considering rolling retries (after having made reliability improvements to individual tests).

I want less noise as much as everyone else. At the same time, I've seen the negative consequences of this kind of "just for now" fix too many times.

(Also, FWIW, the `run-workflow-on-snapshot` test mentioned in https://broadworkbench.atlassian.net/browse/QA-1437 has been a bit less flaky than it used to be.)

zarsky-broad:
@gdlex4015 @ungwudik thoughts?

zarsky-broad:
@breilly2 on the point of reliability, the nightly integration tests have a [43% success rate over the last month](https://app.circleci.com/insights/github/DataBiosphere/terra-ui/workflows/nightly-integration-tests/overview?reporting-window=last-30-days&branch=dev). I could turn down the number of retries to 2, so that if a particular test fails twice in a row it would fail the workflow.

breilly2:
Neat! I didn't know CircleCI had that report!

That pass rate feels accurate to me. Most of the time, the failure is in `run-notebook`, and most of the time, it's the screenshot showing a 403 response (though there are also some other common ones).

As humans dealing with these, we can recognize those patterns and have confidence that it's the tests that are flaky and not the product. If we automate that work, I'd want to build in that kind of recognition. Otherwise, adding one retry could make us miss degrading product reliability until it drops below 50%. (Yes, that's an extreme case, and users would start complaining before then, but I don't think we want to do that to them.)

zarsky-broad:
What if I only put the retry on those tests that are known to be flaky, and only have it happen once?

ungwudik:
Do we know the root cause of the flakiness? Should that be explored first? For example, if the test is attempting to act on a UI element faster than that element loads, maybe we can add/increase wait time instead.
I see that we are already sleeping and waiting for 'no spinners' in the different places in the test, but I wonder if there's anything else that can be done around there.

That retry feature seems handy though, especially as more tests are added to this project. However, I appreciate Brian's concerns and would +1 Isaac's suggestion of 'retrying only for specific tests' as a compromise, and as a last resort.

integration-tests/package.json
```diff
@@ -29,6 +30,6 @@
     "lodash": "^4.17.20",
     "nodemon": "^2.0.4",
     "p-retry": "^4.2.0",
-    "puppeteer": "^5.3.1"
+    "puppeteer": "^9.0.0"
```
breilly2, line 33: This gives us aria selectors, which could be useful for some tests!

integration-tests/jest/1-run-notebook.integration-test.js
```diff
@@ -2,4 +2,4 @@ const { registerTest } = require('./jest-utils')
 const { testRunNotebook } = require('../tests/run-notebook')
 
 
-registerTest(testRunNotebook)
+registerTest(testRunNotebook, 1)
```
breilly2, line 5: This has several failure modes that I've seen frequently. The most common is the "403" response. The next most common is probably where the notebook is open but hasn't evaluated the expression for some reason. (It happened just this morning on alpha: https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/5492/workflows/b550addd-1e83-4483-8a08-797e588315da/jobs/28464). I really feel like we should be able to fix these up.

integration-tests/jest/1-run-workflow-on-snapshot.integration-test.js
```diff
@@ -2,4 +2,4 @@ const { registerTest } = require('./jest-utils')
 const { testRunWorkflowOnSnapshot } = require('../tests/run-workflow-on-snapshot')
 
 
-registerTest(testRunWorkflowOnSnapshot)
+registerTest(testRunWorkflowOnSnapshot, 1)
```
breilly2, line 5: I haven't looked at this enough to know why it fails, but I've always seen the screenshot be on the workflow configuration screen. Everything there should be in our control, so I feel like we really should be able to fix that. (It might occasionally fail because the workflow took too long, same as `run-workflow`).

integration-tests/jest/1-run-workflow.integration-test.js
```diff
@@ -7,4 +7,4 @@ const { registerTest } = require('./jest-utils')
 const { testRunWorkflow } = require('../tests/run-workflow')
 
 
-registerTest(testRunWorkflow)
+registerTest(testRunWorkflow, 1)
```
breilly2, line 10: This one really don't fail very often. Occasionally, Cromwell and/or Rawls are overloaded and the workflow gets stuck in "queued" or "submitted" or "running" and doesn't finish in time. A re-run might fix it, but might just apply more load.

I think I've also seen an occasional failure where it's trying to navigate to the workspace where the workspace list has been filtered, the workspace is visible, but for some reason doesn't get clicked. I bet we could fix this. (This could probably also affect most other tests, but for some reason, I feel like it has mostly affected this one.)

breilly2 (commented):
This is better. I'm still really on the fence. I also realized that this makes the failure modes transparent to us since we won't get stack traces (probably) and screenshots (definitely), so we'll start to lose track of which failure modes are most common... basically all of the characterizations I added in comments on the specific tests here.

panentheos (commented):
I agree with Brian's comments, I don't like codifying the idea that failures are acceptable. Granted, the remaining test problems are the hardest ones to track down and fix, but ultimately the purpose of any test code is to provide a _reliable_ signal about the quality of the underlying system. If they're not doing that, I think we should either commit resources to fix them (including getting help from other teams, if needed) or remove the test(s) and stop paying for something that's not serving its purpose.

zarsky-broad:
Ok. Based on the feedback here, I'm closing this out.

PR #2456, opened 2021-04-22
POC: add query state hook

panentheos:
This is a proof of concept for a hook that manages state that lives in the URL. It moves some logic around, but the goal is removing the imperative effect code that writes the query string.

While I think this idea is interesting and may be worth exploring, it's not a slam dunk. The two included usages are pretty clean, but there is a usage pattern in the (still unfinished) Workflows page that this wouldn't completely replace, specifically being able to generate an `href` value for a link, _ahead_ of actually navigating. Ideally there would be some way to define the state transformation logic once, and use it both ways, but this solution doesn't allow for that.

### DIFF
--- a/src/libs/nav.js
+++ b/src/libs/nav.js
@@ -113,3 +113,15 @@ export const updateSearch = (query, params) => {
     history.replace({ search: newSearch })
   }
 }
+
+export const useQueryState = field => {
+  const { query } = useRoute()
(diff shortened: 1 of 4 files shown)


petesantos:
This looks like a promising way to handle query params. Closing until we have bandwidth to handle this issue.

PR #2461, opened 2021-04-23
Big deps update [AS-736]

zarsky-broad:
One functional change, otherwise only a couple changes due to updating React, see [here](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#changes-to-event-delegation) and [here](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#effect-cleanup-timing) for explanation.

Edit: I'm breaking out the markdown changes into a separate PR.

~The functional change is to markdown rendering: previously, we escaped markdown contents, which had the effect of making html tags render literally, counter to the markdown specification.~

~Now, we purify html tags properly, which means that you can use them, but things like scripts will be removed. This is safer than the old way we were handling it.~
~According to some querying of the prod database, less than 1% of workspaces have html-like code in their descriptions, so I don't expect this to effect many people, although someone out there is going to be very happy that they can now embed images and write superscripts.~


integration-tests/utils/terra-sa-utils.js
```diff
@@ -3,12 +3,12 @@ const { JWT } = require('google-auth-library')
 const { getSecrets, userEmail } = require('../utils/integration-config')
 
 
-const makeAuthClient = _.memoize((subject, { client_email, private_key }) => {
+const makeAuthClient = _.memoize((subject, { client_email: email, private_key: key }) => {
```
breilly2, line 6: Curious... is this just a style preference? I feel like I might have chosen the style I used for a reason, but I don't remember what that reason might have been.

src/components/DataExplorerFrame.js
```diff
@@ -11,6 +11,8 @@ const DataExplorerFrame = ({ dataset }) => {
   const { origin } = _.find({ name: dataset }, datasets)
 
   useEffect(() => {
+    const instance = elem.current
```
breilly2, line 14: What is the purpose of this change? (I trust that it is very intentional... I'm just trying to learn.)

src/components/IdleStatusMonitor.js
```diff
@@ -88,10 +88,10 @@ const InactivityTimer = ({ id, timeout, countdownStart }) => {
       setLastActive(Date.now())
     }
 
-    _.forEach(event => document.addEventListener(event, updateLastActive), targetEvents)
+    _.forEach(event => document.addEventListener(event, updateLastActive, true), targetEvents)
```
breilly2, line 107: Can you explain this? Docs say this last parameter is `wantsUntrusted` and is specifically for Firefox. This isn't a change I would expect to see in a deps PR.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -215,7 +215,7 @@ const Notebooks = _.flow(
 )(({
   apps, galaxyDataDisks, name: wsName, namespace, workspace, workspace: { accessLevel, canShare, workspace: { bucketName } },
   refreshApps, onRequesterPaysError, listView, setListView
-}) => {
+}, ref) => {
```
breilly2, line 218: What is this for?

breilly2 (commented):
Initial comments on clean-up changes. Still need to go through package updates.

integration-tests/utils/terra-sa-utils.js
zarsky-broad: The linter doesn't like when we declare variables that aren't camelcase; we don't enforce linting on the test files, but I like to keep it happy anyway, which is the reason for these changes outside the `src` folder that we lint normally.

src/components/DataExplorerFrame.js
zarsky-broad: See here, it's from the react upgrade: https://reactjs.org/blog/2020/08/10/react-v17-rc.html#potential-issues

src/components/IdleStatusMonitor.js
zarsky-broad: See here, it's from the react upgrade: https://reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues

zarsky-broad: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener the third argument is capture, the fourth would be wantsUntrusted.

src/pages/workspaces/workspace/Notebooks.js
zarsky-broad: All pages in a workspace wrapper are surrounded by a `forwardRef`, so if they don't declare ref in their arguments (even if it's unused), react complains in the js console.

src/components/markdown.js
```diff
@@ -30,5 +33,15 @@ export const newWindowLinkRenderer = (href, title, text) => {
...
 export const MarkdownEditor = props => {
-  return h(Suspense, { fallback: centeredSpinner() }, [h(SimpleMDE, props)])
+  return h(Suspense, { fallback: centeredSpinner() }, [h(SimpleMDE, _.merge({
+    options: {
+      autofocus: true,
+      renderingConfig: {
+        singleLineBreaks: false
+      },
+      previewClass: ['editor-preview', 'markdown-body'],
+      previewRender: renderMarkdown,
```
breilly2, line 43: Nice! How did you discover this? I don't see it in the docs. :(

breilly2 (approved):
👍   LGTM. I clicked around and tested a bunch of things, including the workspace description markdown rendering and editing.

src/components/markdown.js
zarsky-broad: The docs say that [`options` is passed through](https://github.com/RIP21/react-simplemde-editor#options) to the underlying library, so I [looked through its docs](https://github.com/Ionaru/easy-markdown-editor#configuration)

breilly2: Oh, oops, of course. I was looking at the wrong docs. 🤦

package.json
```diff
@@ -4,59 +4,59 @@
...
-    "iframe-resizer": "^4.2.11",
+    "iframe-resizer": "^4.3.2",
     "igv": "2.3.5",
-    "jszip": "^3.5.0",
-    "lodash": "^4.17.20",
-    "marked": "^0.7.0",
-    "outdated-browser-rework": "^2.10.0",
+    "jszip": "^3.6.0",
+    "lodash": "^4.17.21",
+    "marked": "^2.0.3",
```
ehanna4, line 27: Is this supposed to be changed? I thought it looked like you said this was going to be in another PR

zarsky-broad: I'm upgrading the version, but I'm not overhauling the way we parse HTML in this PR.

package.json
```diff
@@ -4,59 +4,59 @@
...
+    "react-dropzone": "^11.3.2",
+    "react-favicon": "^0.0.21",
+    "react-focus-lock": "^2.5.0",
     "react-hyperscript-helpers": "^1.2.0",
-    "react-json-view": "^1.19.1",
-    "react-modal": "^3.11.2",
-    "react-notifications-component": "^2.4.1",
-    "react-onclickoutside": "^6.9.0",
+    "react-json-view": "^1.21.3",
+    "react-modal": "^3.13.1",
```
ehanna4, line 42: @Shakespeared and me made it this far in testing so far

package.json
```diff
@@ -4,59 +4,59 @@
...
+    "qs": "^6.10.1",
+    "react": "17.0.1",
+    "react-collapse": "^5.1.0",
+    "react-dom": "npm:@hot-loader/react-dom@17.0.1",
     "react-draggable": "^4.4.3",
-    "react-dropzone": "^11.2.0",
-    "react-favicon": "^0.0.18",
-    "react-focus-lock": "^2.4.1",
+    "react-dropzone": "^11.3.2",
+    "react-favicon": "^0.0.21",
```
ehanna4: @Shakespeared and i found that the BioDataCatalyst favicon is not showing up anymore (checked on dev branch locally and it works there)

src/components/RuntimeManager.js
```diff
@@ -226,7 +226,7 @@ export default class RuntimeManager extends PureComponent {
     } else if (runtime.status === 'Running' && prevRuntime.status === 'Updating') {
       notify('success', 'Number of workers has updated successfully.')
     }
-    if (prevApp && prevApp.status !== 'RUNNING' && app && app.status === 'RUNNING') {
+    if (prevApp?.status !== 'RUNNING' && app?.status === 'RUNNING') {
```
ehanna4: Something is not working with this logic. When switching between tabs if Galaxy is running a new "launch galaxy" notification pops up on every tab switch which it was not previously doing. (@Shakespeared )
Screen Shot 2021-05-10 at 3 13 09 PM

zarsky-broad: Good catch, I re-ordered it to be clearer.

package.json
zarsky-broad: try again with this latest push?

src/components/data/UploadPreviewTable.js
```diff
@@ -41,12 +41,14 @@ const UploadDataTable = props => {
...
-    _.flow( // eslint-disable-line lodash-fp/no-unused-result
+    const loadMetadata = _.flow(
       withErrorReporting('Error loading entity data'),
       Utils.withBusyState(setMetadataLoading)
     )(async () => {
       setEntityMetadata(await Ajax(signal).Workspaces.workspace(namespace, name).entityMetadata())
-    })()
+    })
+
+    loadMetadata()
```
ehanna4, line 51: Is this for readability or does it have a functional improvement?

zarsky-broad: It makes eslint happy, so we don't have to disable it above.

src/libs/browser-storage.js
```diff
@@ -51,7 +51,7 @@ export const listenStatic = (storage, key, fn) => {
 export const getDynamic = (storage, key) => {
   const storageKey = `dynamic-storage/${key}`
   const data = maybeParseJSON(storage.getItem(storageKey))
-  return data && data.value
+  return data?.value
```
ehanna4, line 54: We are skeptical if this will always produce the same results as the current model due to returning "undefined" rather than `data`s falsey value and then relies on anything using this function to know to handle this. Could we just leave this as is?
@Shakespeared

src/libs/utils.js
```diff
@@ -401,9 +401,10 @@ export const useCancelable = () => {
 
 export const useCancellation = () => {
   const controller = useRef()
-  useEffect(() => {
-    return () => controller.current.abort()
-  }, [])
+  useOnMount(() => {
+    const instance = controller.current
```
ehanna4, line 405: Why would breaking this into two lines be beneficial? @Shakespeared

src/pages/billing/List.js
```diff
@@ -113,9 +113,9 @@ const NewBillingProjectModal = ({ onSuccess, onDismiss, billingAccounts, loadAcc
     onDismiss,
     shouldCloseOnOverlayClick: false,
     title: 'Create Billing Project',
-    showCancel: !(billingAccounts && billingAccounts.length === 0),
+    showCancel: !(billingAccounts?.length === 0),
     showButtons: !!billingAccounts,
-    okButton: billingAccounts && billingAccounts.length !== 0 ?
+    okButton: billingAccounts?.length !== 0 ?
```
ehanna4: Can't do optional chaining for this like this - `billingAccounts?.length !== 0` because if billingAccounts is falsey this will be true overall which is not correct

@Shakespeared and i are now scholars when it comes to optional chaining vs short circuit evaluation

src/pages/billing/List.js
```diff
@@ -124,14 +124,14 @@ const NewBillingProjectModal = ({ onSuccess, onDismiss, billingAccounts, loadAcc
...
-    billingAccounts && billingAccounts.length === 0 && h(Fragment, [
+    billingAccounts?.length === 0 && h(Fragment, [
       `You don't have access to any billing accounts.  `,
       h(Link, {
         href: `https://support.terra.bio/hc/en-us/articles/360026182251`,
         ...Utils.newTabLinkProps
       }, ['Learn how to create a billing account.', icon('pop-out', { size: 12, style: { marginLeft: '0.5rem' } })])
     ]),
-    billingAccounts && billingAccounts.length !== 0 && h(Fragment, [
+    billingAccounts?.length !== 0 && h(Fragment, [
```
ehanna4: Same here

src/pages/workspaces/List.js
```diff
@@ -56,7 +56,7 @@ const WorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDelete }) =
   })
 
   const canRead = workspace && Utils.canRead(workspace.accessLevel)
-  const canShare = workspace && workspace.canShare
+  const canShare = workspace?.canShare
```
ehanna4, line 59: Could we add an `!!` here to be more fool-proof that `canShare` will always be a boolean ? @Shakespeared

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -302,7 +293,7 @@ const WorkspaceDashboard = _.flow(
             }, [icon('times', { size: 14 })])
           ])
         }, tagsList),
-        !!tagsList && tagsList.length === 0 && i(['No tags yet'])
+        tagsList?.length === 0 && i(['No tags yet'])
```
ehanna4: How come previously the `tagsList` was coerced into a boolean? @Shakespeared

src/libs/browser-storage.js
zarsky-broad: So, since `maybeParseJson` will default to `undefined`, this shouldn't be an issue; either the item is unset, in which case it's always returned `undefined`, or it's an object; there aren't any other falsy values possible here.

src/libs/utils.js
zarsky-broad: See https://reactjs.org/blog/2020/08/10/react-v17-rc.html#potential-issues

src/pages/billing/List.js
zarsky-broad: Ahh, I'll add some clarifying parens, the original should have been written as `billingAccounts && (billingAccounts.length !== 0)` to make that clearer

zarsky-broad: (actually I'll just change it to use `_.isEmpty`)

src/pages/workspaces/workspace/Dashboard.js
zarsky-broad: It was to make clear that we were checking for its existence, I'll change this to `_.isEmpty`

package.json
ehanna4: yep works now thanks!

src/pages/workspaces/List.js
zarsky-broad: Since this is only used in two places below, and both are negating it anyway, I don't think it's worth it, personally

ehanna4 (approved):
Looks good - did lots of code review and testing with @Shakespeared and the final commit appeared to address our concerns.👍👍👍

PR #2462, opened 2021-04-23
Add link to "workflow details archived" article [BW-566]

cjllanwarne:
Tested manually by hard-coding a `|| true` to the end of the `metadataArchiveStatus === 'ArchivedAndDeleted'` conditional for displaying this message.

Sample page when this message is displayed:

image

Note to reviewers: this change appears more dramatic than it really is due to whitespace changes in the render code. The only real changes are:
1. Converting the `return workflow === undefined ? ... : ...` into a `Utils.cond()`
2. Adding an additional special case renderer, covering `metadataArchiveStatus === 'ArchivedAndDeleted'` 

Blocked because: we need a support article to link to.


src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -122,97 +123,119 @@ const WorkflowDashboard = _.flow(
...
-        div({ style: { display: 'flex', flexWrap: 'wrap' } }, [
-          makeSection('Workflow Status', [
-            div({ style: { lineHeight: '24px', marginTop: '0.5rem' } }, [makeStatusLine(style => statusIcon(status, style), status)])
-          ]),
-          makeSection('Workflow Timing', [
-            div({ style: { marginTop: '0.5rem', display: 'grid', gridTemplateColumns: 'auto 1fr', gap: '0.5rem' } }, [
-              div({ style: styles.sectionTableLabel }, ['Start:']), div([start ? Utils.makeCompleteDate(start) : 'N/A']),
-              div({ style: styles.sectionTableLabel }, ['End:']), div([end ? Utils.makeCompleteDate(end) : 'N/A'])
-            ])
+        div({ style: { display: 'flex', flexDirection: 'column' } }, [
```
panentheos: nit: This wrapper isn't necessary since we're already in a column-oriented container.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -122,97 +123,119 @@ const WorkflowDashboard = _.flow(
...
+          div({ style: { lineHeight: '24px', marginTop: '0.5rem', display: 'flex', flexDirection: 'row', alignItems: 'center' } }, [
+            icon('info-circle', { size: 18, style: { color: colors.accent(), marginRight: '0.5rem' } }),
```
panentheos: Normally these blue info icons are used for popup information, and are clickable. Since all the relevant information is already spelled out inline, we could probably do without an icon here.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -122,97 +123,119 @@ const WorkflowDashboard = _.flow(
...
+            div({ style: Style.elements.sectionHeader, paddingRight: '0.5rem' }, ' Workflow Details Archived')
```
panentheos: This `paddingRight` needs to be passed in the `style` object to take effect (but we should probably just remove it).

panentheos (approved):
See comments, otherwise looks good pending the link.

Tested locally with an ajax override.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -122,97 +123,117 @@ const WorkflowDashboard = _.flow(
 
   return div({ style: { padding: '1rem 2rem 2rem', flex: 1, display: 'flex', flexDirection: 'column' } }, [
     workflowDetailsBreadcrumbSubtitle(namespace, name, submissionId, workflowId),
-    workflow === undefined ?
-      h(Fragment, [
+    Utils.cond([
```
zarsky-broad: Our normal `cond` spacing goes like this:

```js
Utils.cond(
  [predicate, () => result], // all on one line when possible
  [predicate2, () => { // open bracket, predicate, and start of function on one line when function is multiline
    return result
  }]
)
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -122,97 +123,117 @@ const WorkflowDashboard = _.flow(
...
+                }, [
+                  h(CallTable, { namespace, name, submissionId, workflowId, callName, callObjects: calls[callName] })
+                ])
+              }, callNames)
+            ])]
+          )
+        ]
+      ),
+      wdl && h(Collapse,
+        {
```
zarsky-broad: This should be one line, we keep the open brace for props on the same line as the component (here and elsewhere in this PR):
```js
      wdl && h(Collapse, {
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -122,97 +123,117 @@ const WorkflowDashboard = _.flow(
...
+          title: div({ style: Style.elements.sectionHeader }, ['Submitted workflow script'])
+        }, [h(WDLViewer, { wdl })]
+      ),
+      showLog && h(UriViewer, { googleProject: namespace, uri: workflowLog, onDismiss: () => setShowLog(false) })
+    ])
+    )
```
zarsky-broad: If close brackets of any kind have the same indentation, on two lines in a row, it means they were opened on the same line, so they should be closed on the same line for consistency.

zarsky-broad:
I've left some comments on lines I know aren't really yours, but if we could clean up while you're in here that would be awesome...

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -102,7 +103,7 @@ const WorkflowDashboard = _.flow(
   /*
    * Page render
    */
-  const { calls, end, failures, start, status, workflowLog, workflowName, submittedFiles: { workflow: wdl } = {} } = workflow || {}
+  const { metadataArchiveStatus, calls, end, failures, start, status, workflowLog, workflowName, submittedFiles: { workflow: wdl } = {} } = workflow || {}
```
salonishah11, line 105: Will this throw an error if field `metadataArchiveStatus` is not present in the metadata response?  Or the value for metadataArchiveStatus is set to `undefined`?

cjllanwarne: If it's not there it will be `undefined`

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -122,97 +123,114 @@ const WorkflowDashboard = _.flow(
...
+      () => h(Fragment, [
         div({ style: { fontStyle: 'italic', marginBottom: '1rem' } }, [`Fetching workflow metadata...`]),
         centeredSpinner()
-      ]) :
-      h(Fragment, [
+      ])
+    ], [
+      metadataArchiveStatus === 'ArchivedAndDeleted',
+      () => h(Fragment, [
         div({ style: { fontStyle: 'italic', marginBottom: '1rem' } }, [`Workflow metadata fetched in ${fetchTime}ms`]),
```
salonishah11: Should this be removed? I feel it's confusing to see `.. metadata fetched in x seconds` if the next line says 'Workflow Details Archived'. I do agree that technically UI did get some metadata back, but this might be confusing for users no?

salonishah11 (approved):
Just one comment above, otherwise LGTM 👍

PR #2463, opened 2021-04-26
Send Appcues events to Mixpanel [BT-222]

breilly2:
https://broadworkbench.atlassian.net/browse/BT-222

Tested with various Appcues flows, including http://localhost:3000/?appcue=-M2cUbHaOgXdP2P4kMJu and opening the "Support > Tutorials and Videos" menu.

Reviewed the resulting events in Mixpanel with Amelia. We experimented with building funnels and she is fairly confident that she will be able to get the data she is looking for. Following the release to production, I will update the lexicon in the Mixpanel prod environment and create several custom events for convenience.

Khalid & Miguel, I probably only need one of you along with Isaac. Let me know if you want a walk-through.

### DIFF
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -7,6 +7,7 @@ import { cookiesAcceptedKey } from 'src/components/CookieWarning'
 import { Ajax, fetchOk } from 'src/libs/ajax'
 import { getConfig } from 'src/libs/config'
 import { withErrorReporting } from 'src/libs/error'
+import Events from 'src/libs/events'
 import { getAppName } from 'src/libs/logos'
 import * as Nav from 'src/libs/nav'
 import { clearNotification, notify, sessionTimeoutProps } from 'src/libs/notifications'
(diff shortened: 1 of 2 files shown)


src/libs/auth.js
```diff
@@ -185,9 +186,12 @@ authStore.subscribe(withErrorReporting('Error checking TOS', async (state, oldSt
...
 authStore.subscribe(async (state, oldState) => {
   if (!oldState.acceptedTos && state.acceptedTos) {
-    window.Appcues && window.Appcues.identify(state.user.id, {
-      dateJoined: parseJSON((await Ajax().User.firstTimestamp()).timestamp).getTime()
-    })
+    if (window.Appcues) {
+      window.Appcues.identify(state.user.id, {
+        dateJoined: parseJSON((await Ajax().User.firstTimestamp()).timestamp).getTime()
+      })
+      window.Appcues.on('all', captureAppcuesEvent)
```
breilly2, line 193: I don't see anything in the Appcues docs (https://docs.appcues.com/article/302-working-with-client-side-appcues-events) that say the callback can return a `Promise` and they'll make sure it completes (if that's even a thing) but I also can't find any way I can resolve the promise myself that doesn't just result in a new promise... and it seems to work, so... I think this is fine? 🤷

src/libs/events.js
```diff
@@ -63,4 +65,48 @@ export const PageViewReporter = () => {
...
+    const eventProps = {
+      'appcues.flowId': event.flowId,
+      'appcues.flowName': event.flowName,
+      'appcues.flowType': event.flowType,
+      'appcues.flowVersion': event.flowVersion,
+      'appcues.id': event.id,
+      'appcues.interaction.category': event.interaction?.category,
+      'appcues.interaction.destination': event.interaction?.destination,
+      'appcues.interaction.element': event.interaction?.element,
+      'appcues.interaction.fields': event.interaction?.fields ? JSON.stringify(event.interaction.fields) : undefined,
```
kshakir: Still (re-)new to JS. The linter over in Martha has yelled at me for using the `undefined` literal before. Would this line be the same if written as:

```javascript
'appcues.interaction.fields': event.interaction?.fields && JSON.stringify(event.interaction.fields),
```

kshakir (approved):
LGTM with an additional review from a modern full-stack dev.

src/libs/events.js
```diff
@@ -63,4 +65,48 @@ export const PageViewReporter = () => {
...
+    'flow_aborted',
+    'step_started',
+    'step_completed',
+    'step_skipped',
+    'step_aborted',
+    'step_interacted',
+    'form_submitted',
+    'form_field_submitted'
+  ]
+  if (_.includes(eventName, publicEvents)) {
```
mcovarr, line 83: TOL curious why the lodash `_.includes` over the regular array `includes`?

mcovarr (approved):

src/libs/auth.js
zarsky-broad: It's not a thing to make sure it completes; basically, they don't care if the callback returns a promise, or indeed if it returns anything at all. They just call it.

```js
const aPromise = async () => {
  await doSomething()
}

aPromise()
```

They just dispatch and forget it, so we're good.

src/libs/events.js
breilly2: I considered that, but I was afraid of it evaluating to a boolean.

src/libs/events.js
breilly2: Habit, I guess. 🤷

zarsky-broad: The lodash version is guarded against the array being undefined or not an array somehow, so we prefer it as best practice.

src/libs/events.js
zarsky-broad: We're fine with undefined as a literal in this codebase, I'm not sure why Martha wouldn't be.

If you try it in the browser console, though, you'll find that, due to short-circuiting, `undefined && true` is `undefined`, rather than `false`, so you should be good with @kshakir's syntax.

src/libs/events.js
```diff
@@ -63,4 +65,48 @@ export const PageViewReporter = () => {
...
+    const eventProps = {
```
zarsky-broad: I think we can simplify this.

If we change the signature of this function to this:
```js
captureAppcuesEvent = (eventName, { interaction: { fields, ...interaction } = {}, ...event }) => {
```
We can build the result like this:

```js
const eventProps = _.flow(
  _.mapKeys(k => `appcues.${k}`),
  _.merge(_.mapKeys(k => `appcues.interaction.${k}`, interaction)),
  _.set(['appcues.interaction.fields'], fields && JSON.stringify(fields)), // note the [] here that let us set a key with a period in it
)(event)
```

That might need a bit of tweaking, but I think it should put you in the right direction.

src/libs/events.js
kshakir: Here's the linter's explanation: https://eslint.org/docs/rules/no-undefined

src/libs/events.js
breilly2: I considered trying something like this. I didn't want to take the time to write a recursive solution, and I'm not sure about the general transform with the special case for how to handle `interaction`. I also wanted to reduce any risk of this code breaking due to a change in the Appcues event structure, and for it to be easy to fix if it ever breaks. It is a bit verbose, but I think it's simple and easy to understand.

src/libs/events.js
```diff
@@ -63,4 +65,48 @@ export const PageViewReporter = () => {
...
+      'appcues.flowId': event.flowId,
+      'appcues.flowName': event.flowName,
+      'appcues.flowType': event.flowType,
+      'appcues.flowVersion': event.flowVersion,
+      'appcues.id': event.id,
+      'appcues.interaction.category': event.interaction?.category,
+      'appcues.interaction.destination': event.interaction?.destination,
+      'appcues.interaction.element': event.interaction?.element,
+      'appcues.interaction.fields': event.interaction?.fields && JSON.stringify(event.interaction.fields),
```
zarsky-broad: Actually, prodded a bit deeper, turns out that `JSON.stringify(undefined) === undefined`, so you can just
```suggestion
      'appcues.interaction.fields': JSON.stringify(event.interaction?.fields),
```

src/libs/events.js
zarsky-broad:
```suggestion
    const eventProps = { // Building the props manually to make sure we're resilient to any changes in Appcues
```

zarsky-broad (approved):
2 minor suggestions

PR #2460, opened 2021-04-22
[CA-1312] A11y: Table roles (#2448)

andy7i:
This branch and PR was created because @andrewmarcus's fork could not run tests in his PR: #2448 

So I created this branch and PR so that the tests can run before merging into dev.

Note that the original PR is already approved, so this one should not require more approvals.

PR #2464, opened 2021-04-27
[CA-1312] A11y: Enforce every table has a name

andrewmarcus:
This PR builds on top of #2448 / #2460 by making `tableName` a required field for all tables, as I discussed with @andy7i in that PR. This ensures that every table now and in the future has an accessible name.

In order to avoid PropType warnings, I went through every usage of all 4 table types and added names to every table. In many cases the name basically just matches the name of the function in which it's embedded. However, in situations where tabs or paging were being used, I tried to capture the name of the tab and/or the page number in order to provide users with as much context as reasonable.

*Note*: This branch builds on top of branch `a11y/awm-tables`. However, that branch has not been fully merged into `dev` so the diff here might be bigger than the one commit which it actually should include. Once #2460 is merged that will clean up this PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,14 +65,14 @@ commands:
           at: .
       - restore_cache:
           keys:
-            - test-deps5-{{ checksum "integration-tests/package.json" }}-{{ checksum "integration-tests/yarn.lock" }}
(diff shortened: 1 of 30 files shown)


s-rubenstein (approved):
LGTM

marctalbott (approved):

PR #2466, opened 2021-04-28
[CA-1312] Updated integration tests to properly identify tables

andrewmarcus:
Previously, RVGrid was incorrectly marking up table headers and bodies separately with `role="grid"`.  #2460 changes this to use `role="table"` instead at the root of each of the 4 table types.

This tiny PR updates the integration tests to match. This should resolve the Circle CI build issues.


andy7i:
cherry-picked the relevant commit to https://github.com/DataBiosphere/terra-ui/pull/2460. I think we can close this PR

PR #2470, opened 2021-04-29
[IA-2379] Galaxy persistence

gpcarr:
This PR adds two features: Galaxy persistence when creating a Galaxy app and adds Galaxy apps + disks to the "Cloud Environments" page

Galaxy persistence was tested by creating a galaxy app, deleting it, recreating an app and then checking that the data disk and postgres disk from the first app get re-attached to the new app. This PR also disables changing the PD size upon Galaxy creation if the user has an existing persistent disk that is currently detached. When a user goes to delete their Galaxy app they get the following warning screen and get to choose whether or not to keep their persistent disk:
Screen Shot 2021-04-29 at 2 55 46 PM
The majority of these code changes live in `NewGalaxyModal.js`

The second part of this PR focuses on the "Cloud Environments" page and was tested by validating that apps now show up in the "Your cloud environments" table and that galaxy disks show up in the "Your persistent disks" table. This PR also adds a column to the "Your cloud environments" table names "Type" which displays what type (GCE, Dataproc, or Galaxy) of cloud environment that row is representing. The changes look like this:
Screen Shot 2021-04-29 at 2 56 13 PM
The majority of these changes live in `Environments.js`

There are still 3 todos that I'm working on that are super minor:
~~1) There's a bug when clicking on the red warning triangle for error'd galaxy apps in the cloud environments page~~ This is fixed. see screenshot below:
Screen Shot 2021-04-29 at 5 32 18 PM

~~2) There's a bug in displaying the yellow warning triangle for having more than one persistent disk in the same billing project. need to get this logic right since a user can now have both a galaxy disk and a GCE VM disk in the same billing project~~ This is fixed, see screenshot below
Screen Shot 2021-05-04 at 12 02 23 PM

~~3) I would also like to add a tooltip to the persistent disk size selector in the `NewGalaxyModal` when it's disabled, explaining to users why it's disabled~~ This was implemented and the design was changed based on Joy's feedback. See screenshot below
Screen Shot 2021-05-04 at 12 04 04 PM

I will update this PR description as I work on these 3 todos.
I also need to fix all the indentation warnings

### DIFF
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -18,6 +18,7 @@ import {
   getGalaxyDiskCost, RadioBlock
 } from 'src/libs/runtime-utils'
 import * as Style from 'src/libs/style'
+import { newTabLinkProps } from 'src/libs/utils'
 import * as Utils from 'src/libs/utils'
 
 // TODO Factor out common pieces with NewRuntimeModal.styles into runtime-utils
@@ -289,8 +290,11 @@ export const NewGalaxyModal = _.flow(
     return div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
       div({ style: styles.headerText }, ['Persistent disk']),
       div({ style: { marginTop: '0.5rem' } }, [
-        'Persistent disks store analysis data.'
-        // TODO Add info on PDs for Galaxy (similarly to in NewRuntimeModal.js)
+        'Persistent disks store analysis data. ',
+        h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360050566271', ...Utils.newTabLinkProps }, [
+          'Learn more about about persistent disks',
+          icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
+        ])
       ]),
       renderPersistentDiskSizeSection()
     ])
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -49,7 +49,7 @@ export const NewGalaxyModal = _.flow(
   const [kubernetesRuntimeConfig, setKubernetesRuntimeConfig] = useState(app?.kubernetesRuntimeConfig || defaultKubernetesRuntimeConfig)
   const [viewMode, setViewMode] = useState(undefined)
   const [loading, setLoading] = useState(false)
-  const [deleteDiskSelected, setDeleteDiskSelected] = useState(false)
+  const [shouldDeleteDisk, setShouldDeleteDisk] = useState(false)
 
   const currentDataDisk = currentPersistentDisk(apps, galaxyDataDisks)
 
@@ -69,7 +69,7 @@ export const NewGalaxyModal = _.flow(
     Utils.withBusyState(setLoading),
     withErrorReporting('Error deleting galaxy instance')
   )(async () => {
-    await Ajax().Apps.app(app.googleProject, app.appName).delete(attachedDataDisk ? deleteDiskSelected : false)
+    await Ajax().Apps.app(app.googleProject, app.appName).delete(attachedDataDisk ? shouldDeleteDisk : false)
     Ajax().Metrics.captureEvent(Events.applicationDelete, { app: 'Galaxy', ...extractWorkspaceDetails(workspace) })
     return onSuccess()
   })
@@ -291,7 +291,7 @@ export const NewGalaxyModal = _.flow(
       div({ style: { marginTop: '0.5rem' } }, [
         'Persistent disks store analysis data. ',
         h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360050566271', ...Utils.newTabLinkProps }, [
-          'Learn more about about persistent disks',
+          'Learn more about persistent disks',
           icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
         ])
       ]),
@@ -344,7 +344,7 @@ export const NewGalaxyModal = _.flow(
         onDismiss,
         onPrevious: () => {
           setViewMode(undefined)
-          setDeleteDiskSelected(false)
+          setShouldDeleteDisk(false)
         }
       }),
       div({ style: { lineHeight: '1.5rem' } }, [
@@ -352,8 +352,8 @@ export const NewGalaxyModal = _.flow(
           h(RadioBlock, {
             name: 'keep-persistent-disk',
             labelText: 'Keep persistent disk, delete application configuration and compute profile',
-            checked: !deleteDiskSelected,
-            onChange: () => setDeleteDiskSelected(false),
+            checked: !shouldDeleteDisk,
+            onChange: () => setShouldDeleteDisk(false),
             style: { marginTop: '1rem' }
           }, [
             p([
@@ -368,8 +368,8 @@ export const NewGalaxyModal = _.flow(
           h(RadioBlock, {
             name: 'delete-persistent-disk',
             labelText: 'Delete everything, including persistent disk',
-            checked: deleteDiskSelected,
-            onChange: () => setDeleteDiskSelected(true),
+            checked: shouldDeleteDisk,
+            onChange: () => setShouldDeleteDisk(true),
             style: { marginTop: '1rem' }
           }, [
             p([
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -14,7 +14,8 @@ import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
 import Events, { extractWorkspaceDetails } from 'src/libs/events'
 import {
-  currentApp, currentAttachedDataDisk, findMachineType, getGalaxyComputeCost, getGalaxyDiskCost, RadioBlock
+  currentApp, currentAttachedDataDisk, currentPersistentDiskIncludingUnattached, findMachineType, getGalaxyComputeCost,
+  getGalaxyDiskCost, RadioBlock
 } from 'src/libs/runtime-utils'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
@@ -50,16 +51,7 @@ export const NewGalaxyModal = _.flow(
   const [loading, setLoading] = useState(false)
   const [deleteDiskSelected, setDeleteDiskSelected] = useState(false)
 
-  const getCurrentPersistentDiskName = (apps, galaxyDataDisks) => {
-    const currentGalaxyApp = currentApp(apps)
-    const currentDataDiskName = currentGalaxyApp?.diskName
-    const attachedDataDiskNames = _.without([undefined], _.map(app => app.diskName, apps))
-    return currentDataDiskName ?
-      _.find({ currentDataDiskName }, galaxyDataDisks) :
-      _.last(_.sortBy('auditInfo.createdDate', _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), galaxyDataDisks)))
-  }
-
-  const currentDataDisk = getCurrentPersistentDiskName(apps, galaxyDataDisks)
+  const currentDataDisk = currentPersistentDiskIncludingUnattached(apps, galaxyDataDisks)[0]
 
   const createGalaxy = _.flow(
     Utils.withBusyState(setLoading),
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -311,9 +311,9 @@ export const NewGalaxyModal = _.flow(
         return div({ marginTop: '0.75rem' }, [
           div({ style: { ...gridStyle, marginTop: '0.5rem' } }, [
             div({ style: styles.label }, ['Size (GB): ']), h(TooltipTrigger,
-          { content: ['Persistent disk of size ', currentDataDisk.size, ' GB already exists and will be attached upon Galaxy creation'] }, [
-              div({ marginTop: '0.5rem', width: '5rem' }, [currentDataDisk.size])
-            ])
+              { content: ['Persistent disk of size ', currentDataDisk.size, ' GB already exists and will be attached upon Galaxy creation'] }, [
+                div({ marginTop: '0.5rem', width: '5rem' }, [currentDataDisk.size])
+              ])
           ])
         ])
       }],
--- a/src/components/NewGalaxyModal.js
+++ b/src/components/NewGalaxyModal.js
@@ -1,27 +1,32 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
-import { div, h, label, span } from 'react-hyperscript-helpers'
+import { div, h, label, p, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, ButtonSecondary, IdContainer, Link, Select, spinnerOverlay, WarningTitle } from 'src/components/common'
 import { icon } from 'src/components/icons'
 import { NumberInput } from 'src/components/input'
 import { withModalDrawer } from 'src/components/ModalDrawer'
-import { GalaxyLaunchButton, GalaxyWarning } from 'src/components/runtime-common'
+import { GalaxyLaunchButton, GalaxyWarning, SaveFilesHelpGalaxy } from 'src/components/runtime-common'
 import TitleBar from 'src/components/TitleBar'
+import TooltipTrigger from 'src/components/TooltipTrigger'
 import { machineTypes } from 'src/data/machines'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
 import Events, { extractWorkspaceDetails } from 'src/libs/events'
-import { currentApp, currentDataDisk, findMachineType, getGalaxyComputeCost, getGalaxyDiskCost } from 'src/libs/runtime-utils'
+import {
+  currentApp, currentAttachedDataDisk, findMachineType, getGalaxyComputeCost, getGalaxyDiskCost, RadioBlock
+} from 'src/libs/runtime-utils'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
 // TODO Factor out common pieces with NewRuntimeModal.styles into runtime-utils
 const styles = {
   label: { fontWeight: 600, whiteSpace: 'pre' },
+  value: { fontWeight: 400, whiteSpace: 'pre' },
   whiteBoxContainer: { padding: '1rem', borderRadius: 3, backgroundColor: 'white' },
   drawerContent: { display: 'flex', flexDirection: 'column', flex: 1, padding: '1.5rem' },
-  headerText: { fontSize: 16, fontWeight: 600 }
+  headerText: { fontSize: 16, fontWeight: 600 },
+  warningView: { backgroundColor: colors.warning(0.1) }
 }
 
 const defaultDataDiskSize = 500 // GB
@@ -37,19 +42,32 @@ export const NewGalaxyModal = _.flow(
 )(({ onDismiss, onSuccess, apps, galaxyDataDisks, workspace, workspace: { workspace: { namespace, bucketName, name: workspaceName } } }) => {
   // Assumption: If there is an app defined, there must be a data disk corresponding to it.
   const app = currentApp(apps)
-  const dataDisk = currentDataDisk(app, galaxyDataDisks)
+  const attachedDataDisk = currentAttachedDataDisk(app, galaxyDataDisks)
 
-  const [dataDiskSize, setDataDiskSize] = useState(dataDisk?.size || defaultDataDiskSize)
+  const [dataDiskSize, setDataDiskSize] = useState(attachedDataDisk?.size || defaultDataDiskSize)
   const [kubernetesRuntimeConfig, setKubernetesRuntimeConfig] = useState(app?.kubernetesRuntimeConfig || defaultKubernetesRuntimeConfig)
   const [viewMode, setViewMode] = useState(undefined)
   const [loading, setLoading] = useState(false)
+  const [deleteDiskSelected, setDeleteDiskSelected] = useState(false)
+
+  const getCurrentPersistentDiskName = (apps, galaxyDataDisks) => {
+    const currentGalaxyApp = currentApp(apps)
+    const currentDataDiskName = currentGalaxyApp?.diskName
+    const attachedDataDiskNames = _.without([undefined], _.map(app => app.diskName, apps))
+    return currentDataDiskName ?
+      _.find({ currentDataDiskName }, galaxyDataDisks) :
+      _.last(_.sortBy('auditInfo.createdDate', _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), galaxyDataDisks)))
+  }
+
+  const currentDataDisk = getCurrentPersistentDiskName(apps, galaxyDataDisks)
 
   const createGalaxy = _.flow(
     Utils.withBusyState(setLoading),
     withErrorReporting('Error creating app')
   )(async () => {
-    await Ajax().Apps.app(namespace, Utils.generateKubernetesClusterName()).create({
-      kubernetesRuntimeConfig, diskName: Utils.generatePersistentDiskName(), diskSize: dataDiskSize, appType: 'GALAXY', namespace, bucketName, workspaceName
+    await Ajax().Apps.app(namespace, Utils.generateGalaxyAppName()).create({
+      kubernetesRuntimeConfig, diskName: currentDataDisk ? currentDataDisk.name : Utils.generatePersistentDiskName(), diskSize: dataDiskSize,
+      appType: 'GALAXY', namespace, bucketName, workspaceName
     })
     Ajax().Metrics.captureEvent(Events.applicationCreate, { app: 'Galaxy', ...extractWorkspaceDetails(workspace) })
     return onSuccess()
@@ -59,7 +77,7 @@ export const NewGalaxyModal = _.flow(
     Utils.withBusyState(setLoading),
     withErrorReporting('Error deleting galaxy instance')
   )(async () => {
-    await Ajax().Apps.app(app.googleProject, app.appName).delete()
+    await Ajax().Apps.app(app.googleProject, app.appName).delete(attachedDataDisk ? deleteDiskSelected : false)
     Ajax().Metrics.captureEvent(Events.applicationDelete, { app: 'Galaxy', ...extractWorkspaceDetails(workspace) })
     return onSuccess()
   })
@@ -83,7 +101,8 @@ export const NewGalaxyModal = _.flow(
   })
 
   const renderActionButton = () => {
-    const deleteButton = h(ButtonSecondary, { disabled: false, style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') }, ['Delete'])
+    const deleteButton = h(ButtonSecondary, { disabled: false, style: { marginRight: 'auto' }, onClick: () => setViewMode('deleteWarn') },
+      ['Delete Environment Options'])
     const pauseButton = h(ButtonSecondary, { disabled: false, style: { marginRight: '1rem' }, onClick: () => { pauseGalaxy() } }, ['Pause'])
     const resumeButton = h(ButtonSecondary, { disabled: false, style: { marginRight: '1rem' }, onClick: resumeGalaxy }, ['Resume'])
 
@@ -112,14 +131,21 @@ export const NewGalaxyModal = _.flow(
             h(ButtonPrimary, { disabled: false, onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])
           ])],
           ['STOPPED', () => h(Fragment, [
-            h(ButtonSecondary, { disabled: true, style: { marginRight: 'auto' }, tooltip: 'Cloud Compute must be resumed first.', onClick: () => setViewMode('deleteWarn') }, ['Delete']),
+            h(ButtonSecondary, {
+              disabled: true, style: { marginRight: 'auto' }, tooltip: 'Cloud Compute must be resumed first.',
+              onClick: () => setViewMode('deleteWarn')
+            }, ['Delete']),
             resumeButton
           ])],
           ['ERROR', () => deleteButton],
           [Utils.DEFAULT, () => {
             return h(Fragment, { tooltip: 'Cloud Compute must be resumed first.' }, [
-              h(ButtonSecondary, { disabled: true, style: { marginRight: 'auto' }, tooltip: 'Cloud Compute must be running.', onClick: () => setViewMode('deleteWarn') }, ['Delete']),
-              h(ButtonSecondary, { disabled: true, style: { marginRight: '1rem' }, tooltip: 'Cloud Compute must be running.', onClick: () => { pauseGalaxy() } }, ['Pause'])
+              h(ButtonSecondary, {
+                disabled: true, style: { marginRight: 'auto' }, tooltip: 'Cloud Compute must be running.', onClick: () => setViewMode('deleteWarn')
+              }, ['Delete']),
+              h(ButtonSecondary,
+                { disabled: true, style: { marginRight: '1rem' }, tooltip: 'Cloud Compute must be running.', onClick: () => { pauseGalaxy() } },
+                ['Pause'])
             ])
           }]
         )]
@@ -129,14 +155,14 @@ export const NewGalaxyModal = _.flow(
   const renderMessaging = () => {
     return Utils.switchCase(viewMode,
       ['createWarn', renderCreateWarning],
-      ['deleteWarn', renderDeleteWarning],
+      ['deleteWarn', renderDeleteDiskChoices],
       ['launchWarn', renderLaunchWarning],
       [Utils.DEFAULT, renderDefaultCase]
     )
   }
 
   const renderCreateWarning = () => {
-    return h(Fragment, [
+    return div({ style: styles.drawerContent }, [
       h(TitleBar, {
         title: 'Cloud Environment',
         style: { marginBottom: '0.5rem' },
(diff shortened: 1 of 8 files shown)


src/components/NewGalaxyModal.js
```diff
@@ -37,19 +38,24 @@ export const NewGalaxyModal = _.flow(
...
+
+  const currentDataDiskName = getCurrentPersistentDiskName(apps, galaxyDataDisks)
 
   const createGalaxy = _.flow(
     Utils.withBusyState(setLoading),
     withErrorReporting('Error creating app')
   )(async () => {
-    await Ajax().Apps.app(namespace, Utils.generateKubernetesClusterName()).create({
-      kubernetesRuntimeConfig, diskName: Utils.generatePersistentDiskName(), diskSize: dataDiskSize, appType: 'GALAXY', namespace, bucketName, workspaceName
+    console.log('hiiiii')
```
ehanna4: 😂 might wanna remove these `console.logs()`

src/libs/ajax.js
```diff
@@ -1126,7 +1126,10 @@ const Runtimes = signal => ({
   list: async (labels = {}) => {
     const res = await fetchLeo(`api/google/v1/runtimes?${qs.stringify({ saturnAutoCreated: true, ...labels })}`,
       _.mergeAll([authOpts(), appIdentifier, { signal }]))
-    return res.json()
+    console.log('TAKE 2')
+    const finalRes = res.json()
+    console.log(finalRes)
```
kyuksel: just a reminder to clean these up :)

src/components/NewGalaxyModal.js
```diff
@@ -286,13 +285,24 @@ export const NewGalaxyModal = _.flow(
...
       div({ style: styles.headerText }, ['Persistent disk']),
       div({ style: { marginTop: '0.5rem' } }, [
         'Persistent disks store analysis data.'
         // TODO Add info on PDs for Galaxy (similarly to in NewRuntimeModal.js)
       ]),
+      renderPersistentDiskSizeSection()
+    ])
+  }
+
+  const renderPersistentDiskSizeSection = () => {
```
gpcarr, line 302: Not sure if this logic is the best way to do this - could use some feedback

src/components/NewGalaxyModal.js
```diff
@@ -286,13 +285,24 @@ export const NewGalaxyModal = _.flow(
...
+    const gridStyle = { display: 'grid', gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', gridGap: '1rem', alignItems: 'center' }
+    return currentDataDisk ? div({ marginTop: '0.75rem' }, [div({ style: { ...gridStyle, marginTop: '0.5rem' } }, [
+      div({ style: styles.label }, ['Size (GB): ']), h(TooltipTrigger,
+        { content: ['Persistent disk of size ', currentDataDisk.size, ' GB already exists and will be attached upon Galaxy creation'] }, [
+          div({ marginTop: '0.5rem', width: '5rem' }, [currentDataDisk.size])
+        ])
+    ])]) :
```
jdcanas: i feel like line 305 indicates line 300 should be spaced out across a few lines. 

You could LAO use a `Utils` method to get out of using a ternary (like `Utils.cond`), which could help clean it up

src/pages/Environments.js
```diff
@@ -75,30 +75,71 @@ const DeleteDiskModal = ({ disk: { googleProject, name }, onDismiss, onSuccess }
...
   const getDeleteDiskId = useGetter(deleteDiskId)
+  const [errorAppId, setErrorAppId] = useState()
+  const getErrorAppId = useGetter(errorAppId)
+  const [deleteAppId, setDeleteAppId] = useState()
+  const getDeleteAppId = useGetter(deleteAppId)
   const [sort, setSort] = useState({ field: 'project', direction: 'asc' })
   const [diskSort, setDiskSort] = useState({ field: 'project', direction: 'asc' })
+  // TODO: Can we not use a state variable for appSort, and use a local variable instead?
+  // eslint-disable-next-line
+  const [appSort, setAppSort] = useState({ field: 'project', direction: 'asc' })
```
jdcanas: Yeah, why are we using a state variable here if the setter is never used?

src/pages/Environments.js
```diff
@@ -132,82 +179,173 @@ const Environments = () => {
...
+              return cloudEnvironment.appName ? renderBillingProjectApp(cloudEnvironment) : renderBillingProjectRuntime(cloudEnvironment)
+            }
+          },
+          {
+            size: { basis: 90, grow: 0 },
+            headerRenderer: () => h(Sortable, { sort, field: 'created', onSort: setSort }, ['Type']),
+            cellRenderer: ({ rowIndex }) => {
+              const cloudEnvironment = filteredCloudEnvironments[rowIndex]
+              // TODO: update return logic once we support more app types (will need a backend change to return appType in list apps endpoint as well)
+              return cloudEnvironment.appName ? 'Galaxy' : (cloudEnvironment.runtimeConfig.cloudService === 'DATAPROC' ? 'Dataproc' : cloudEnvironment.runtimeConfig.cloudService)
```
jdcanas, line 307: hmm... what exactly is this trying to do, check the appType? I see you have a TODO to update it once we return that and its a bit of scope creep (can't believe we aren't returning this!), but right now this is a nested ternary where the first condition is always false I'm fairly sure. App name is auto-generated in the UI, like GCE runtimes. I added a print with an app I created printing the value to confirm. 

image

jdcanas, line 322: edit: nvm misread

Consider using `Utils.cond` to un-nest this ternary (thanks isaac!)

src/components/RuntimeManager.js
```diff
@@ -125,6 +125,8 @@ export const AppErrorModal = ({ app, onDismiss }) => {
     withErrorReporting('Error loading app details'),
     Utils.withBusyState(setLoadingAppDetails)
   )(async () => {
+    console.log('helllloooo')
+    console.log(JSON.stringify(app))
```
kyuksel: reminder to remove before merging :)

src/libs/ajax.js
```diff
@@ -1232,11 +1235,17 @@ const Apps = signal => ({
       _.mergeAll([authOpts(), appIdentifier, { signal }]))
     return res.json()
   },
+  listWithoutProject: async (labels = {}) => {
+    const res = await fetchLeo(`api/google/v1/apps?${qs.stringify(labels)}`,
+      _.mergeAll([authOpts(), appIdentifier, { signal }]))
+    const resFinal = res.json()
+    return resFinal
```
kyuksel:
```suggestion
    return res.json()
```

src/components/NewGalaxyModal.js
```diff
@@ -306,17 +316,72 @@ export const NewGalaxyModal = _.flow(
...
+            onChange: () => setDeleteDiskSelected(false),
+            style: { marginTop: '1rem' }
+          }, [
+            p([
+              'Deletes your application configuration and cloud compute profile, but detaches your persistent disk and saves it for later. ',
+              'The disk will be automatically reattached the next time you create a Galaxy application.'
+            ]),
+            p({ style: { marginBottom: 0 } }, [
+              'You will continue to incur persistent disk cost at ',
+              span({ style: { fontWeight: 600 } }, [Utils.formatUSD(getGalaxyDiskCost(dataDiskSize)), ' per month.'])
```
kyuksel:
```suggestion
              span({ style: { fontWeight: 600 } }, [Utils.formatUSD(getGalaxyDiskCost(dataDiskSize)), ' per hour.'])
```
since `getGalaxyDiskCost()` returns hourly disk cost.

kyuksel (commented):
Looking good so far. A few comments inline. Will finish reviewing later tonight or tomorrow morning.

src/libs/runtime-utils.js
```diff
@@ -164,10 +169,24 @@ export const machineCost = machineType => {
 
 export const currentApp = _.flow(trimAppsOldestFirst, _.last)
 
-export const currentDataDisk = (app, galaxyDataDisks) => {
+export const currentAttachedDataDisk = (app, galaxyDataDisks) => {
   return _.find({ name: app?.diskName }, galaxyDataDisks)
 }
 
+export const getCurrentPersistentDiskName = (apps, galaxyDataDisks) => {
```
kyuksel: Since this method is used only by `NewGalaxyModal`, it may be good to move it to `NewGalaxyModal.js` file.

src/libs/runtime-utils.js
```diff
@@ -164,10 +169,24 @@ export const machineCost = machineType => {
...
+  return currentDataDiskName ?
+    _.find({ currentDataDiskName }, galaxyDataDisks) :
+    _.last(_.sortBy('auditInfo.createdDate', _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), galaxyDataDisks)))
+}
+
+export const trimDisksOldestFirst = _.flow(
+  _.remove({ status: 'Deleting' }),
+  _.sortBy('auditInfo.createdDate')
+)
+
```
kyuksel: Is this method used anywhere?

src/pages/Environments.js
```diff
@@ -132,82 +179,173 @@ const Environments = () => {
...
+  const appsByProject = _.groupBy('googleProject', apps)
+
+  const renderBillingProjectApp = app => {
+    const inactive = !_.includes(app.status, ['DELETING', 'ERROR']) &&
+      currentApp(appsByProject[app.googleProject]) !== app
+    return h(Fragment, [
+      app.googleProject,
+      inactive && h(TooltipTrigger, {
+        content: 'This billing project has multiple active cloud environments. Only the most recently created one will be used.'
+      }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
```
kyuksel, line 207: This tooltip showed up when I pressed the trash icon for an app on `Cloud Environments` page and it was briefly in `PREDELETING` status. Should we add that status to the list `inactive` statuses?

image

kyuksel (changes requested):

kyuksel:
image

is shown when a user clicks on the trash icon on Cloud Environments page for a Galaxy disk. I don't think the note on the bottom shouldn't be shown since it's a Galaxy app?

src/pages/Environments.js
kyuksel: Also thinking more generally, does this tooltip still apply the way it was conceived since users can now have multiple cloud environments routinely and it may not make sense to notify them if they have multiple cloud environments because one is for Galaxy and one is a runtime?

kyuksel:
I logged into local Terra with Hermione's test account, went to Cloud Environments page and clicked on one of the red triangles belonging to a Galaxy app and saw the following:

image

The error message says "`Cluster` not found...". `Cluster` here is a Kubernetes cluster that the app was associated with, correct?

src/components/NewGalaxyModal.js
```diff
@@ -286,13 +285,24 @@ export const NewGalaxyModal = _.flow(
   }
 
   const renderPersistentDiskSection = () => {
-    const gridStyle = { display: 'grid', gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', gridGap: '1rem', alignItems: 'center' }
     return div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
       div({ style: styles.headerText }, ['Persistent disk']),
       div({ style: { marginTop: '0.5rem' } }, [
         'Persistent disks store analysis data.'
         // TODO Add info on PDs for Galaxy (similarly to in NewRuntimeModal.js)
```
kyuksel: We're awaiting User Ed to be able to remove the TODO above, correct?

src/libs/runtime-utils.js
```diff
@@ -164,10 +169,24 @@ export const machineCost = machineType => {
...
   return _.find({ name: app?.diskName }, galaxyDataDisks)
 }
 
+export const getCurrentPersistentDiskName = (apps, galaxyDataDisks) => {
+  const currentGalaxyApp = currentApp(apps)
+  const currentDataDiskName = currentGalaxyApp?.diskName
+  const attachedDataDiskNames = _.without([undefined], _.map(app => app.diskName, apps))
+  return currentDataDiskName ?
+    _.find({ currentDataDiskName }, galaxyDataDisks) :
+    _.last(_.sortBy('auditInfo.createdDate', _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), galaxyDataDisks)))
```
kyuksel: I'm having to spend a lot of time to wrap my head around the logic here. Need to think more about whether we could make it easier to parse.

src/components/NewGalaxyModal.js
kyuksel: What logic exactly were you unsure about?

src/components/NewGalaxyModal.js
```diff
@@ -306,17 +316,72 @@ export const NewGalaxyModal = _.flow(
                 onlyInteger: true,
                 value: dataDiskSize,
                 style: { marginTop: '0.5rem', width: '5rem' },
+                disabled: !!currentDataDisk,
+                tooltip: !!currentDataDisk ?
+                  ['Persistent disk of size ', currentDataDisk.size, ' GB already exists and will be attached upon Galaxy creation'] :
+                  undefined,
```
kyuksel: Is this portion applicable? The `div` these lines are part of, will only be rendered if `currentDataDisk` is `falsy`, in which case `!!currentDataDisk` should always be `false`, and so should `disabled` and `tooltip`?

src/components/NewGalaxyModal.js
```diff
@@ -306,17 +316,72 @@ export const NewGalaxyModal = _.flow(
...
+        onDismiss,
+        onPrevious: () => {
+          setViewMode(undefined)
+          setDeleteDiskSelected(false)
+        }
+      }),
+      div({ style: { lineHeight: '1.5rem' } }, [
+        h(Fragment, [
+          h(RadioBlock, {
+            name: 'delete-persistent-disk',
```
kyuksel: Should `name` be `keep-persistent-disk` or something?

kyuksel (changes requested):
Done reviewing 😊  Lots of good work 🚀  

Tested through various app lifecycle transitions. Reported a couple of behaviour inline that may need attention.

Also inline are other comments and questions.

kyuksel:
Noticed the warning below

image

when I deleted an app from Cloud Environments page (while keeping the disk in case it matters...). 
Didn't get to look into it.

kyuksel:
We should look into the scenario described below and think about how we can improve the UX.

As part of the sprint review demo, I
1. created a new app with a new disk
2. deleted the app keeping the disk
3. opened the Galaxy modal to create a new app with the existing disk
4. noticed that the UI wasn't showing the existing disk and the disk size was enabled to be adjusted
5. tried 4. a few times and the same thing happened
6. went to Cloud Environments page and saw that the app was still being deleted
7. saw that the app disappeared from the page
8. went back to the Galaxy modal and this time saw that the existing disk was picked up to be re-used

I'm guessing it takes some time for the disk to be marked as `not attached to an app` and if a user tries to create a new app during that time, the UI doesn't get the memo and does allow the user to create a new disk instead of using the existing one.

Qi77Qi:
the galaxy version on [this](https://user-images.githubusercontent.com/23626109/117033924-d57f5900-acd0-11eb-8e99-082fdf485e2d.png) should come from leo's response eventually. It needs to be the version of previous galaxy that was launched with the PD. If it's too much to resolve for this ticket, can you create a new ticket to track?

jdcanas:
> Noticed the warning below
> 
> image
> 
> when I deleted an app from Cloud Environments page (while keeping the disk in case it matters...).
> Didn't get to look into it.

This is the memory leak I investigated, its because there's an async refresh callback that is triggered in the synchronous modal dismissal, such that the callback returns to a component that is detached, but the callback updates state of the container (i.e. a component still attached on modal dismissal) so is still relevant. Its a non-issue since its not something that's rescheduled, just a one-time callback

src/libs/ajax.js
gpcarr: oops meant to remove this, did it for testing purposes

src/components/NewGalaxyModal.js
gpcarr: yeah, i'll check in with them to see how that's going

src/libs/runtime-utils.js
gpcarr: yeah it took me a bit to understand as well - we do the same kind of logic for determining whether or not to delete a PD for a GCE VM. see https://github.com/DataBiosphere/terra-ui/blob/dev/src/components/NewRuntimeModal.js#L169

gpcarr:
> I logged into local Terra with Hermione's test account, went to Cloud Environments page and clicked on one of the red triangles belonging to a Galaxy app and saw the following:
> 
> image
> 
> The error message says "`Cluster` not found...". `Cluster` here is a Kubernetes cluster that the app was associated with, correct?

yes, this is referring to the kubernetes cluster and is an error that gets returned by leo

src/components/NewGalaxyModal.js
gpcarr: meant to remove this actually - will commit that with the rest of the changes from the feedback received so far. this was my original approach to not letting the user choose a disk size if they have an existing PD that is going to be attached to their app but after syncing with joy, we decided to go with a different approach seen here: 
Screen Shot 2021-05-12 at 10 42 48 PM

src/pages/Environments.js
gpcarr: the warning triangle will not show up if the user has a runtime and galaxy app in the same billing project. the logic that handles this is in `renderBillingProjectApp` and `renderBillingProjectRuntime` - these methods ensure that the warning triable only shows up if there is more than one galaxy app in a project or more than one runtime in a project but i tested to make sure that it doesn't show up if the user has 1 galaxy app and 1 runtime in a project

gpcarr:
> image
> 
> is shown when a user clicks on the trash icon on Cloud Environments page for a Galaxy disk. I don't think the note on the bottom shouldn't be shown since it's a Galaxy app?

good catch - pushing up the fix for this soon

src/libs/runtime-utils.js
gpcarr: meant to delete - thought it was needed but it is not. will remove

src/components/NewGalaxyModal.js
gpcarr: the ternary in `renderPersistentDiskSizeSection` - justin addressed this in a comment below and i will try out using a `Utils` method to clean this up a bit

src/components/NewGalaxyModal.js
gpcarr: will do - will update `NewRuntimeModal` as well since it also only uses `'delete-persistent-disk'`

gpcarr:
> We should look into the scenario described below and think about how we can improve the UX.
> 
> As part of the sprint review demo, I
> 
> 1. created a new app with a new disk
> 2. deleted the app keeping the disk
> 3. opened the Galaxy modal to create a new app with the existing disk
> 4. noticed that the UI wasn't showing the existing disk and the disk size was enabled to be adjusted
> 5. tried 4. a few times and the same thing happened
> 6. went to Cloud Environments page and saw that the app was still being deleted
> 7. saw that the app disappeared from the page
> 8. went back to the Galaxy modal and this time saw that the existing disk was picked up to be re-used
> 
> I'm guessing it takes some time for the disk to be marked as `not attached to an app` and if a user tries to create a new app during that time, the UI doesn't get the memo and does allow the user to create a new disk instead of using the existing one.

Fixed this by disabling the create button on the notebooks page like we mentioned in our slack conversations :)

gpcarr:
Added a hyperlink to the user ed article that will be updated once this PR is released to prod
Screen Shot 2021-05-18 at 5 32 59 PM

src/components/NewGalaxyModal.js
```diff
@@ -286,37 +286,110 @@ export const NewGalaxyModal = _.flow(
...
     return div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
       div({ style: styles.headerText }, ['Persistent disk']),
       div({ style: { marginTop: '0.5rem' } }, [
-        'Persistent disks store analysis data.'
-        // TODO Add info on PDs for Galaxy (similarly to in NewRuntimeModal.js)
+        'Persistent disks store analysis data. ',
+        h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360050566271', ...Utils.newTabLinkProps }, [
+          'Learn more about about persistent disks',
+          icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
+        ])
```
kyuksel: The article doesn't seem to say anything about disks. Will it be updated later?

gpcarr:
> this

must have missed this! will create a separate ticket to track this work since this PR is already large

kyuksel:
> must have missed this! will create a separate ticket to track this work since this PR is already large

What work were you referring to?

src/libs/runtime-utils.js
```diff
@@ -177,6 +179,19 @@ export const appIsSettingUp = app => {
...
+  const currentGalaxyApp = currentAppIncludingDeleting(apps)
+  const currentDataDiskName = currentGalaxyApp?.diskName
+  const attachedDataDiskNames = _.without([undefined], _.map(app => app.diskName, apps))
+  const currentDatDisk = currentDataDiskName ?
+    _.find({ name: currentDataDiskName }, galaxyDataDisks) :
+    _.last(_.sortBy('auditInfo.createdDate', _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), galaxyDataDisks)))
+  const isCurrentDiskDetaching = currentGalaxyApp && (currentGalaxyApp.status === 'DELETING' || currentGalaxyApp.status === 'PREDELETING')
+  const res = [currentDatDisk, isCurrentDiskDetaching]
+  console.dir(res)
+  return res
```
```suggestion
  return [currentDataDisk, isCurrentDiskDetaching]
```

src/libs/runtime-utils.js
```diff
@@ -177,6 +179,19 @@ export const appIsSettingUp = app => {
   return app && (app.status === 'PROVISIONING' || app.status === 'PRECREATING')
 }
 
+export const currentPersistentDiskIncludingUnattached = (apps, galaxyDataDisks) => {
+  const currentGalaxyApp = currentAppIncludingDeleting(apps)
+  const currentDataDiskName = currentGalaxyApp?.diskName
+  const attachedDataDiskNames = _.without([undefined], _.map(app => app.diskName, apps))
+  const currentDatDisk = currentDataDiskName ?
```
```suggestion
  const currentDataDisk = currentDataDiskName ?
```

src/pages/Environments.js
```diff
@@ -325,11 +322,6 @@ const Environments = ({ namespace }) => {
               const cloudEnvironment = filteredCloudEnvironments[rowIndex]
               console.log(JSON.stringify(cloudEnvironment))
```
kyuksel:
```suggestion
```

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -357,8 +357,8 @@ const Notebooks = _.flow(
             style: {
               ...Style.elements.card.container, height: 125, marginTop: 15
             },
-            disabled: appIsSettingUp(app),
-            tooltip: appIsSettingUp(app) && 'Galaxy app is being created',
+            disabled: appIsSettingUp(app) || _.last(currentPersistentDiskIncludingUnattached(apps, galaxyDataDisks)),
+            tooltip: (appIsSettingUp(app) && 'Galaxy app is being created') || (_.last(currentPersistentDiskIncludingUnattached(apps, galaxyDataDisks)) && 'Your persistent disk is still attached to your previous app; you can create a new app once your previous app finishes deleting, which will take a few minutes.'),
```
kyuksel: I wonder if it'd make the code easier to read if ```_.last(currentPersistentDiskIncludingUnattached(apps, galaxyDataDisks))``` were factored out so it reads something like

```javascript
disabled: appIsSettingUp(app) || isCurrentDiskDetaching(apps, galaxyDataDisks)
```

src/libs/runtime-utils.js
```diff
@@ -177,6 +179,19 @@ export const appIsSettingUp = app => {
...
+    _.find({ name: currentDataDiskName }, galaxyDataDisks) :
+    _.last(_.sortBy('auditInfo.createdDate', _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), galaxyDataDisks)))
```
kyuksel, line 191: To make sure I understand, would the line above get executed if a user deleted their Galaxy app but kept the disks?

kyuksel (changes requested):
Looks great overall. Mostly minor comments inline.

Testing:
- created a Galaxy app
- deleted the while keeping the disk
- tried to create another app right away and saw that the create button was disabled
- monitored it on Cloud Environments page
- attempted again after detaching
- saw the app was successfully created

src/components/NewGalaxyModal.js
gpcarr: yeah, anton has the changes in draft mode and will publish the changes once this PR is released

src/pages/workspaces/workspace/Notebooks.js
gpcarr: I can make a helper method that calls `currentPersistentDiskIncludingUnattached` and returns the second value in the tuple. i don't want to change the name of this method since i use it in a couple of different places within this component

src/libs/runtime-utils.js
gpcarr: yes, that's correct. it would only get executed in that case. the other case is that the user has a current app that's either active or in deleting status so the disk is either attached or detaching

gpcarr: i'll add a comment to make this a bit more clear

kyuksel: The comments are helpful. Thank you.

src/pages/workspaces/workspace/Notebooks.js
kyuksel: AFAICS, all use cases depend on either `currentDataDisk` or `isCurrentDiskDetaching` so how about breaking up `currentPersistentDiskIncludingUnattached` a bit such that it returns only `currentDataDisk` as the name implies, and the remaining detaching status boolean is returned by `isCurrentDiskDetaching`, like so:

```javascript
export const currentPersistentDiskIncludingUnattached = (apps, galaxyDataDisks) => {
  // a user's PD can either be attached to their current app, detaching from a deleting app or unattached
  const currentGalaxyApp = currentAppIncludingDeleting(apps)
  const currentDataDiskName = currentGalaxyApp?.diskName
  const attachedDataDiskNames = _.without([undefined], _.map(app => app.diskName, apps))
  // if the disk is attached to an app (or being detached from a deleting app), return that disk. otherwise,
  // return the newest galaxy disk that the user has unattached to an app
  const currentDataDisk = currentDataDiskName ?
    _.find({ name: currentDataDiskName }, galaxyDataDisks) :
    _.last(_.sortBy('auditInfo.createdDate', _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), galaxyDataDisks)))
  return currentDataDisk
}
```

```javascript
const isCurrentGalaxyDiskDetaching = apps => {
  const currentGalaxyApp = currentAppIncludingDeleting(apps)    
  return currentGalaxyApp && (currentGalaxyApp.status === 'DELETING' || currentGalaxyApp.status === 'PREDELETING')
}
```

src/pages/Environments.js
```diff
@@ -145,75 +305,67 @@ const Environments = () => {
...
           },
           {
             size: { basis: 120, grow: 0.2 },
             headerRenderer: () => 'Location',
             cellRenderer: ({ rowIndex }) => {
-              const { runtimeConfig: { zone, region } } = filteredRuntimes[rowIndex]
-              return zone || region
+              const cloudEnvironment = filteredCloudEnvironments[rowIndex]
+              // This logic works under the assumption that all Galaxy apps get created in zone 'us-central1-a'
+              const zone = cloudEnvironment.runtimeConfig ? cloudEnvironment.runtimeConfig.zone : 'us-central1-a'
```
dmohs: `us-central1-a` is a zone with in the `us-central1` region. This leads me to think that either the logic here or the variable names are not quite right.

src/pages/Environments.js
```diff
@@ -108,6 +150,12 @@ const Environments = () => {
     if (!_.some({ id: getDeleteDiskId() }, newDisks)) {
       setDeleteDiskId(undefined)
     }
+    if (!_.some({ appName: getErrorAppId() }, newApps)) {
```
dmohs: These getters surprise me. Can you explain the need for this (offline if needed) over using the raw form?

gpcarr:
> > must have missed this! will create a separate ticket to track this work since this PR is already large
> 
> What work were you referring to?

I was referring to Qi's comment. i captured the work in this ticket https://broadworkbench.atlassian.net/browse/IA-2778

src/libs/runtime-utils.js
```diff
@@ -164,14 +169,33 @@ export const machineCost = machineType => {
...
+  // if the disk is attached to an app (or being detached from a deleting app), return that disk. otherwise,
+  // return the newest galaxy disk that the user has unattached to an app
+  return currentDataDiskName ?
+    _.find({ name: currentDataDiskName }, galaxyDataDisks) :
+    _.last(_.sortBy('auditInfo.createdDate', _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), galaxyDataDisks)))
+}
+
+export const isCurrentGalaxyDiskDetaching = apps => {
+  const currentGalaxyApp = currentAppIncludingDeleting(apps)
+  return currentGalaxyApp && (currentGalaxyApp.status === 'DELETING' || currentGalaxyApp.status === 'PREDELETING')
```
dmohs: Alternatively:
```suggestion
  return currentGalaxyApp && _.includes(currentGalaxyApp.status, ['DELETING', 'PREDELETING'])
```

dmohs (approved):
Looks good. Very minor suggested tweaks which you are free to reject.

src/components/NewGalaxyModal.js
```diff
@@ -286,37 +286,110 @@ export const NewGalaxyModal = _.flow(
...
   const renderPersistentDiskSection = () => {
-    const gridStyle = { display: 'grid', gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', gridGap: '1rem', alignItems: 'center' }
     return div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
       div({ style: styles.headerText }, ['Persistent disk']),
       div({ style: { marginTop: '0.5rem' } }, [
-        'Persistent disks store analysis data.'
-        // TODO Add info on PDs for Galaxy (similarly to in NewRuntimeModal.js)
+        'Persistent disks store analysis data. ',
+        h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360050566271', ...Utils.newTabLinkProps }, [
+          'Learn more about about persistent disks',
```
kyuksel:
```suggestion
          'Learn more about persistent disks',
```

kyuksel: Can we also change `Learn more about about persistent disks in the Terra Support site` in `NewRuntimeModal.js` to read `Learn more about persistent disks`?

src/pages/Environments.js
kyuksel: This appears to be the pattern that's been used since the earlier days of Terra UI. I see [Brett's comment](https://github.com/DataBiosphere/terra-ui/blob/dbf333204e2c10aa36cdefce08765ca77c0890af/src/libs/utils.js#L335) noting a use case. If it isn't necessary, I think it'd be good to drop all cases for consistency, which should be beyond the scope of this PR.

kyuksel (approved):
Looks great 👍🏼  Requested a couple of text changes inline.

src/pages/Environments.js
dmohs: I agree that fixing all unnecessary usage is beyond the scope of this PR, but let's not add more. I've added just a few suggestions to help out.
```suggestion
    if (!_.some({ appName: errorAppId }, newApps)) {
```

src/pages/Environments.js
```diff
@@ -108,6 +150,12 @@ const Environments = () => {
...
+      setErrorAppId(undefined)
+    }
+    if (!_.some({ appName: getDeleteAppId() }, newApps)) {
```
dmohs:
```suggestion
    if (!_.some({ appName: deleteAppId }, newApps)) {
```

src/pages/Environments.js
```diff
@@ -70,35 +70,77 @@ const DeleteDiskModal = ({ disk: { googleProject, name }, onDismiss, onSuccess }
...
+  const [galaxyDisks, setGalaxyDisks] = useState()
   const [loading, setLoading] = useState(false)
   const [errorRuntimeId, setErrorRuntimeId] = useState()
   const getErrorRuntimeId = useGetter(errorRuntimeId)
   const [deleteRuntimeId, setDeleteRuntimeId] = useState()
   const getDeleteRuntimeId = useGetter(deleteRuntimeId)
   const [deleteDiskId, setDeleteDiskId] = useState()
   const getDeleteDiskId = useGetter(deleteDiskId)
+  const [errorAppId, setErrorAppId] = useState()
+  const getErrorAppId = useGetter(errorAppId)
```
dmohs:
```suggestion
```

src/pages/Environments.js
```diff
@@ -70,35 +70,77 @@ const DeleteDiskModal = ({ disk: { googleProject, name }, onDismiss, onSuccess }
...
+  const [deleteAppId, setDeleteAppId] = useState()
+  const getDeleteAppId = useGetter(deleteAppId)
```
dmohs:
```suggestion
```

src/components/NewGalaxyModal.js
```diff
@@ -37,19 +43,23 @@ export const NewGalaxyModal = _.flow(
...
   const app = currentApp(apps)
-  const dataDisk = currentDataDisk(app, galaxyDataDisks)
+  const attachedDataDisk = currentAttachedDataDisk(app, galaxyDataDisks)
 
-  const [dataDiskSize, setDataDiskSize] = useState(dataDisk?.size || defaultDataDiskSize)
+  const [dataDiskSize, setDataDiskSize] = useState(attachedDataDisk?.size || defaultDataDiskSize)
   const [kubernetesRuntimeConfig, setKubernetesRuntimeConfig] = useState(app?.kubernetesRuntimeConfig || defaultKubernetesRuntimeConfig)
   const [viewMode, setViewMode] = useState(undefined)
   const [loading, setLoading] = useState(false)
+  const [deleteDiskSelected, setDeleteDiskSelected] = useState(false)
```
dmohs: This name threw me off until I saw usage. No need to fix this, but you might consider whether something like `shouldDeleteDisk` or `shouldDeleteDiskSelected` would make the intent more obvious. Sorry for the extra noise, but this comment got deleted from my earlier review.

src/pages/Environments.js
gpcarr: the reason I did it this way was because for GCE runtimes we have access to `zone` whereas for for Dataproc we have access to `region` - i can try to think of a way to clean up this logic. we display that `zone` or `region` as the "Location" column

src/libs/runtime-utils.js
gpcarr: will update!

src/components/NewGalaxyModal.js
gpcarr: will do!

src/components/NewGalaxyModal.js
gpcarr: no worries! will update

src/pages/Environments.js
kyuksel: Right. I think what Mohs was pointing out is…the default below for `region` (`us-central1-a`) is actually a `zone` value. In other words it should be `us-central1` instead.

Also, optional chaining could simplify the two definitions, like so:

```javascript
const zone = cloudEnvironment?.runtimeConfig?.zone || 'us-central1-a'
const region = cloudEnvironment?.runtimeConfig?.region || 'us-central1'
```

and then we could just

```javascript
return zone || region
```

src/pages/Environments.js
gpcarr: made updates to not use getters anymore, thanks for pointing that out!

src/pages/Environments.js
gpcarr: ah i see, will update!

gpcarr: I did
```
const zone = cloudEnvironment?.runtimeConfig?.zone
const region = cloudEnvironment?.runtimeConfig?.region
return zone || region || 'us-central1-a'
```
instead because we don't want to return `'us-central1-a'` for Dataproc

gpcarr:
@kyuksel @dmohs i've addressed all feedback now - going to merge once tests pass if that sounds good!

PR #2475, opened 2021-05-03
[CA-1312] A11y: Enforce every table has a name (#2464)

andy7i:
This branch and PR was created because @andrewmarcus's fork could not run tests in his PR: #2464

So I created this branch and PR so that the tests can run before merging into dev.

See the original PR for additional PR review details.

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -194,6 +194,7 @@ const DataTable = props => {
               ref: table,
               width, height,
               rowCount: entities.length,
+              tableName: `${entityType} data table, page ${pageNumber} of ${Math.ceil(totalRowCount / itemsPerPage)}`,
               noContentMessage: `No ${entityType}s to display.`,
               onScroll,
               initialX,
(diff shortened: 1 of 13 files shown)


andy7i:
[test failure](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/5559/workflows/59d48d50-5537-47c1-b6b7-1d045a720646/jobs/28705/artifacts) seems to be a flaky test, which has come up in the past -- rerunning test

marctalbott (approved):

s-rubenstein (approved):
LGTM, we should notify the terra UI champions that tablename is now required, and why.

andy7i:
I think Andrew Marcus captured it really well in this comment:

> Hmm, that's a great point. I was trying to keep these changes as unobtrusive to developers as possible, and trying to avoid having to make changes to every table.
> 
> But on the other hand, I already changed all the other tables to add support for aria-sort, so that's probably not a big deal at this point.
> 
> And it would certainly be better to force devs to explicitly think about accessibility where they need to. Now that I think about it, this caption is sort of like adding alt text to an image: sure you could just use a default but that really defeats the purpose.
> 
> Ok, I'll make it a required field, and update all of the tables accordingly.

https://github.com/DataBiosphere/terra-ui/pull/2448#discussion_r618022434

PR #2476, opened 2021-05-03
[IA-2719] Spark master machine type in Terra UI is not reflected in GCP

rtitle:
See https://broadworkbench.atlassian.net/browse/IA-2719

The fix is to specify `masterMachineType` instead of `machineType` as the key for Dataproc runtimes as that is the key used in the Dataproc case when creating a runtime.

Tested locally by creating Dataproc + GCE runtimes with different settings, and validating the POST request payload.

### DIFF
--- a/src/components/NewRuntimeModal.js
+++ b/src/components/NewRuntimeModal.js
@@ -347,7 +347,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
                 diskSize: masterDiskSize
               })
             } : {
-              machineType: masterMachineType || defaultDataprocMachineType,
+              masterMachineType: masterMachineType || defaultDataprocMachineType,
               masterDiskSize,
               numberOfWorkers: desiredNumberOfWorkers,
               ...(desiredNumberOfWorkers && {


src/components/NewRuntimeModal.js
```diff
@@ -347,7 +347,7 @@ export const NewRuntimeModal = withModalDrawer({ width: 675 })(class NewRuntimeM
                 diskSize: masterDiskSize
               })
             } : {
-              machineType: masterMachineType || defaultDataprocMachineType,
+              masterMachineType: masterMachineType || defaultDataprocMachineType,
```
kyuksel, line 350: 🤦🏼

kyuksel (approved):
Thanks for fixing this! Verified it by creating a Dataproc runtime with the following config:

image

and then seeing the corresponding request payload on `Network Tab`

image

Qi77Qi (approved):
unfortunate bug...thanks for fixing!

gpcarr:
tested locally by creating a cluster with 8 CPUs and verifying `masterMachineType` in request payload and verifying that the correct master machine type gets created in google

also recreated the bug in dev and saw that even thought i selected 8 CPUs, the machine type was `n1-standard-2` in the request payload and in google

Thanks for fixing @rtitle!

gpcarr (approved):

rtitle:
Thanks all - I also tested `PATCH` for Dataproc and it seems to work as expected.

PR #2486, opened 2021-05-10
Fix idle timeout [PROD-537]

zarsky-broad:
Verified in-browser.

We now ensure that signing back in to your google account in another tab won't sign you back in to Terra.

### DIFF
--- a/src/components/IdleStatusMonitor.js
+++ b/src/components/IdleStatusMonitor.js
@@ -1,10 +1,12 @@
 import _ from 'lodash/fp'
+import * as qs from 'qs'
 import { useEffect, useState } from 'react'
 import { div, h, iframe } from 'react-hyperscript-helpers'
 import ButtonBar from 'src/components/ButtonBar'
 import Modal from 'src/components/Modal'
 import { getUser } from 'src/libs/auth'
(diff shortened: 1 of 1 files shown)


petesantos (approved):
Paired and reviewed

src/components/IdleStatusMonitor.js
```diff
@@ -35,30 +37,45 @@ const IdleStatusMonitor = ({
   timeout = Utils.durationToMillis({ minutes: 15 }),
   countdownStart = Utils.durationToMillis({ minutes: 3 })
 }) => {
+  // State
```
dmohs, line 40: I judge comments like this do not add enough value to justify them.

dmohs (approved):

src/components/IdleStatusMonitor.js
zarsky-broad: I've been adding these indicators in this order with this spacing to all hook-based components, and I've consistently received feedback that they're useful to people in lieu of the class-based methods that previously served as waypoints in the component's structure.

zarsky-broad:
Test failures are known flake, merging.

PR #2488, opened 2021-05-11
Replace JM link with infobox for archived workflows [BW-672]

cjllanwarne:
Tested locally by moving the "archive limit" up and down. Starting at 6 years so that this doesn't _actually_ show up yet.

See screenshots in ticket.

### DIFF
--- a/src/components/PopupTrigger.js
+++ b/src/components/PopupTrigger.js
@@ -72,17 +72,19 @@ const PopupTrigger = Utils.forwardRefWithName('PopupTrigger', ({ content, childr
 
 export default PopupTrigger
 
-export const InfoBox = ({ size, children, style, side }) => {
+export const InfoBox = ({ size, children, style, side, tooltip, iconOverride }) => {
   const [open, setOpen] = useState(false)
   return h(PopupTrigger, {
     side,
+    width: '100px',
     onChange: setOpen,
     content: div({ style: { padding: '0.5rem', width: 300 } }, [children])
   }, [
     h(Clickable, {
+      tooltip,
       as: 'span', 'aria-label': 'More info', 'aria-expanded': open, 'aria-haspopup': true
     }, [
-      icon('info-circle', { size, style: { cursor: 'pointer', color: colors.accent(), ...style } })
+      icon(iconOverride || 'info-circle', { size, style: { cursor: 'pointer', color: colors.accent(), ...style } })
     ])
   ])
 }
--- a/src/components/PopupTrigger.js
+++ b/src/components/PopupTrigger.js
@@ -76,7 +76,6 @@ export const InfoBox = ({ size, children, style, side, tooltip, iconOverride }) 
   const [open, setOpen] = useState(false)
   return h(PopupTrigger, {
     side,
-    width: '100px',
     onChange: setOpen,
     content: div({ style: { padding: '0.5rem', width: 300 } }, [children])
   }, [
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -1,15 +1,16 @@
 import _ from 'lodash/fp'
 import { Fragment, useEffect, useState } from 'react'
-import { div, h } from 'react-hyperscript-helpers'
+import { div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import * as breadcrumbs from 'src/components/breadcrumbs'
-import { ClipboardButton, Link, Select } from 'src/components/common'
+import { Clickable, ClipboardButton, Link, Select } from 'src/components/common'
 import { centeredSpinner, icon } from 'src/components/icons'
 import { DelayedSearchInput } from 'src/components/input'
 import {
   collapseStatus, failedIcon, makeSection, makeStatusLine, runningIcon, statusIcon,
   submissionDetailsBreadcrumbSubtitle, submittedIcon, successIcon
 } from 'src/components/job-common'
+import { InfoBox } from 'src/components/PopupTrigger'
 import { FlexTable, Sortable, TextCell, TooltipCell } from 'src/components/table'
 import TooltipTrigger from 'src/components/TooltipTrigger'
 import { Ajax } from 'src/libs/ajax'
@@ -110,6 +111,35 @@ const SubmissionDetails = _.flow(
   )(workflows)
 
   const { succeeded, failed, running, submitted } = _.groupBy(wf => collapseStatus(wf.status), workflows)
+
+  const archiveLimitYears = 6
+  const archiveLimitString = `${archiveLimitYears} year${archiveLimitYears > 1 ? 's' : ''}`
+
+  const archived = lastChangedDate => {
+    const dateYearsAgo = new Date(new Date().setFullYear(new Date().getFullYear() - archiveLimitYears))
+    return lastChangedDate && Date.parse(lastChangedDate) < dateYearsAgo
+  }
+
+  const archivedInfoIcon = ({ name, iconOverride }) => {
+    return h(InfoBox, {
+      style: { color: colors.light(), margin: '0.5rem' },
+      tooltip: `${name} (unavailable)`,
+      iconOverride
+    }, [
+      div({ style: Style.elements.sectionHeader }, 'Workflow Details Archived'),
+      div({ style: { paddingTop: '0.5rem', paddingBottom: '0.5rem' } }, [`This workflow ran more than ${archiveLimitString} ago and its details have been archived. ${name} is no longer available.`]),
+      div([
+        'Please refer to the ',
+        h(Link, {
+          href: 'https://support.terra.bio/hc/en-us/articles/360060601631',
+          ...Utils.newTabLinkProps
+        }, [icon('pop-out', { size: 18 }), ' Workflow Details Archived']),
+        ' support article for details on how to access the archive.'
+      ])
+    ])
+  }
+
+
   /*
    * Page render
    */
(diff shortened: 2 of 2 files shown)


src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +111,35 @@ const SubmissionDetails = _.flow(
   )(workflows)
 
   const { succeeded, failed, running, submitted } = _.groupBy(wf => collapseStatus(wf.status), workflows)
+
+  const archiveLimitYears = 6
```
evanbernstein: The AC for this ticket had this bullet:
"comment everywhere we have the deletion delay constant in code, a pointer to all the other locations this constant exists in code"
I'd love to see a comment here that points to the other repositories that this constant exists as well as a comment in that code that points here.

rsasch (commented):
I found the UX of this confusing — because the greyed out icons have a hover and are a lighter color, I wouldn't think to click on them, but clicking on them is the only way to get a more detailed message and a link to a support article.

src/components/PopupTrigger.js
```diff
@@ -72,17 +72,19 @@ const PopupTrigger = Utils.forwardRefWithName('PopupTrigger', ({ content, childr
 
 export default PopupTrigger
 
-export const InfoBox = ({ size, children, style, side }) => {
+export const InfoBox = ({ size, children, style, side, tooltip, iconOverride }) => {
   const [open, setOpen] = useState(false)
   return h(PopupTrigger, {
     side,
+    width: '100px',
```
zarsky-broad: I'm _extremely_ hesitant to make a change like this to the base structure of something that's used in multiple places, rather than make it an optional addition.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +111,35 @@ const SubmissionDetails = _.flow(
...
+
+  const archived = lastChangedDate => {
+    const dateYearsAgo = new Date(new Date().setFullYear(new Date().getFullYear() - archiveLimitYears))
+    return lastChangedDate && Date.parse(lastChangedDate) < dateYearsAgo
+  }
+
+  const archivedInfoIcon = ({ name, iconOverride }) => {
+    return h(InfoBox, {
+      style: { color: colors.secondary(), margin: '0.5rem' },
+      tooltip: `${name} (unavailable)`,
```
zarsky-broad: If there's going to be a tooltip, perhaps it should be something like "learn more". I'd recommend checking with UX, if you haven't already.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -248,23 +278,28 @@ const SubmissionDetails = _.flow(
...
-                  }, [icon('tasks', { size: 18 })]),
-                  h(Link, {
-                    href: Nav.getLink('workspace-workflow-dashboard', { namespace, name, submissionId, workflowId }),
-                    style: { padding: '.6rem', display: 'flex' },
-                    tooltip: 'Workflow Dashboard [alpha]',
-                    'aria-label': 'Workflow Dashboard [alpha]'
-                  }, [icon('tachometer', { size: 18 })]),
+                  archived(filteredWorkflows[rowIndex].statusLastChangedDate) ? [
+                    archivedInfoIcon({ name: 'Job Manager', iconOverride: 'tasks' }),
+                    archivedInfoIcon({ name: 'Workflow Dashboard', iconOverride: 'tachometer' })
```
zarsky-broad: I don't love overriding the icon. The info icon means click for more information. These mean other things. However, the infobox here just gives more information.

src/components/PopupTrigger.js
cjllanwarne: Oops, this wasn't meant to make it this far (and didn't work anyway), I'll remove it

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cjllanwarne: 👍  I've pinged Joy on whether this is a good idea

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +111,35 @@ const SubmissionDetails = _.flow(
   )(workflows)
 
   const { succeeded, failed, running, submitted } = _.groupBy(wf => collapseStatus(wf.status), workflows)
+
+  const archiveLimitYears = 6
+  const archiveLimitString = `${archiveLimitYears} year${archiveLimitYears > 1 ? 's' : ''}`
+
+  const archived = lastChangedDate => {
```
zarsky-broad: This is being used for a boolean, so it should be `isArchived`.

rsasch (commented):
I would defer to the UX team, but I think, instead of adding an additional visual variant (i.e. grey-with-hover-but-clickable icons), we keep things simple and use an existing one, e.g. the blue "i" that appears on the "Workflow Submission" page.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +111,35 @@ const SubmissionDetails = _.flow(
...
+    const dateYearsAgo = new Date(new Date().setFullYear(new Date().getFullYear() - archiveLimitYears))
+    return lastChangedDate && Date.parse(lastChangedDate) < dateYearsAgo
```
zarsky-broad: You should add parentheses here so it's whether you want it to evaluate

```js
(lastChangedDate && Date.parse(lastChangedDate)) < dateYearsAgo
```
vs
```js
lastChangedDate && (Date.parse(lastChangedDate) < dateYearsAgo)
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +111,35 @@ const SubmissionDetails = _.flow(
   )(workflows)
 
   const { succeeded, failed, running, submitted } = _.groupBy(wf => collapseStatus(wf.status), workflows)
+
+  const archiveLimitYears = 6
+  const archiveLimitString = `${archiveLimitYears} year${archiveLimitYears > 1 ? 's' : ''}`
+
+  const archived = lastChangedDate => {
+    const dateYearsAgo = new Date(new Date().setFullYear(new Date().getFullYear() - archiveLimitYears))
```
zarsky-broad: You should look in to https://date-fns.org/v2.21.3/docs/differenceInYears, we have this library already.

zarsky-broad: You can find an example in `ServiceAlerts.js` of using `differenceInDays`, note that we import it from the `fp` submodule of date-fns

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
zarsky-broad: I think she's transitioned off of Terra, I believe Jerome is the UX person now.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +111,35 @@ const SubmissionDetails = _.flow(
...
+  }
+
+  const archivedInfoIcon = ({ name, iconOverride }) => {
+    return h(InfoBox, {
+      style: { color: colors.secondary(), margin: '0.5rem' },
+      tooltip: `${name} (unavailable)`,
+      iconOverride
+    }, [
+      div({ style: Style.elements.sectionHeader }, 'Workflow Details Archived'),
+      div({ style: { paddingTop: '0.5rem', paddingBottom: '0.5rem' } }, [`This workflow's details have been archived (> ${archiveLimitString} old).`]),
```
zarsky-broad:
```suggestion
      div({ style: { padding: '0.5rem 0' } }, [`This workflow's details have been archived (> ${archiveLimitString} old).`]),
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +111,35 @@ const SubmissionDetails = _.flow(
...
+      div([
+        'Please refer to the ',
+        h(Link, {
+          href: 'https://support.terra.bio/hc/en-us/articles/360060601631',
+          ...Utils.newTabLinkProps
+        }, [icon('pop-out', { size: 18 }), ' Workflow Details Archived']),
```
zarsky-broad: I think the icon usually goes after the link text.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cjllanwarne: I dipped a toe into using the original Info icon here (https://github.com/DataBiosphere/terra-ui/pull/2489) but the UX feedback was to prefer this original method. Let me know if you had a different alternative in mind?

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cjllanwarne: No longer applies after switching to `differenceInDays` as suggested below

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cjllanwarne: 👍  I ended up using `differenceInDays` because the `differenceInYears` was in _whole_ years, which isn't fine-grained enough for this info

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +112,33 @@ const SubmissionDetails = _.flow(
   )(workflows)
 
   const { succeeded, failed, running, submitted } = _.groupBy(wf => collapseStatus(wf.status), workflows)
+
+  // Note: This 'archiveLimitYears' value should reflect the current 'archive-delay' value configured for PROD in firecloud-develop's
+  // 'cromwell.conf.ctmpl' file:
+  const archiveLimitYears = 6
```
salonishah11: Just a heads up, [this](https://github.com/broadinstitute/firecloud-develop/pull/2541) PR in firecloud-develop will set the archive-delay in prod to 1 year

salonishah11:
It looks like there is still some discussion happening about grey-with-hover-and-clickable icons v/s tooltip-with-blue-icon that needs to checked with UX. Holding off reviewing until then.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cjllanwarne: See also https://github.com/broadinstitute/firecloud-develop/pull/2540. We need to keep them in sync whenever one is changed the other must be changed too

cjllanwarne:
@salonishah11 @rsasch the feedback from Joy was to prefer this option over the info box

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +112,33 @@ const SubmissionDetails = _.flow(
...
+    }, [
+      div({ style: Style.elements.sectionHeader }, 'Workflow Details Archived'),
+      div({ style: { padding: '0.5rem 0' } }, [`This workflow's details have been archived (> ${archiveLimitString} old).`]),
+      div([
+        'Please refer to the ',
+        h(Link, {
+          href: 'https://support.terra.bio/hc/en-us/articles/360060601631',
+          ...Utils.newTabLinkProps
+        }, ['Workflow Details Archived ', icon('pop-out', { size: 18 })]),
+        ' support article for more details.'
```
rsasch: Nit: include the words "support article" in the linked text
```suggestion
        'Please refer to the ',
        h(Link, {
          href: 'https://support.terra.bio/hc/en-us/articles/360060601631',
          ...Utils.newTabLinkProps
        }, ['Workflow Details Archived  support article', icon('pop-out', { size: 18 })]),
        ' for more details.'
```

rsasch (approved):
One small nit, otherwise 👍🏻

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
salonishah11: Based on the tech talk discussion, I think this should be `deletionLimitInYears` since the metadata would still be available on archiving, but not after it is moved to GCS. Same applies for places below: 'archive' => 'deletion' limit

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +112,33 @@ const SubmissionDetails = _.flow(
...
+  const isDeleted = statusLastChangedDate => differenceInDays(Date.parse(statusLastChangedDate), Date.now()) > (deletionDelayYears * 365)
+
+  const deletedInfoIcon = ({ name, iconOverride }) => {
+    return h(InfoBox, {
+      style: { color: colors.secondary(), margin: '0.5rem' },
+      tooltip: `${name} unavailable. Click to learn more.`,
+      iconOverride
+    }, [
+      div({ style: Style.elements.sectionHeader }, 'Workflow Details Archived'),
+      div({ style: { padding: '0.5rem 0' } }, [`This workflow's details have been archived (> ${deletionDelayString} old).`]),
```
aednichols, line 129: TOL, I think it flows a bit smoother without the parenthetical and symbolic greater-than
```suggestion
      div({ style: { padding: '0.5rem 0' } }, [`This workflow ran more than ${deletionDelayString} ago and is archived.`]),
```
Screen Shot 2021-05-13 at 12 54 32 PM

aednichols (approved):
Very slick and user friendly... 👍 I couldn't help myself and threw in a copy-editing TOL

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -110,6 +112,33 @@ const SubmissionDetails = _.flow(
   )(workflows)
 
   const { succeeded, failed, running, submitted } = _.groupBy(wf => collapseStatus(wf.status), workflows)
+
+  // Note: This 'deleteLimitYears' value should reflect the current 'deletion-delay' value configured for PROD in firecloud-develop's
+  // 'cromwell.conf.ctmpl' file:
+  const deletionDelayYears = 6
+  const deletionDelayString = `${deletionDelayYears} year${deletionDelayYears > 1 ? 's' : ''}`
+  const isDeleted = statusLastChangedDate => differenceInDays(Date.parse(statusLastChangedDate), Date.now()) > (deletionDelayYears * 365)
```
zarsky-broad: Should this use `date-fns`'s `parseJson`?

zarsky-broad (approved):
Just one thought.

PR #2489, opened 2021-05-11
InfoBox alternative to "Replace JM link with infobox for archived workflows" [BW-672]

cjllanwarne:
Instead of greyed-out links which produce a pop-up when clicked, this option adds a new info-box _next to_ the options in the `Links` column:

image


cjllanwarne:
Closing. UX opinion was to prefer the original option (grey icons with pop-up info box) in #2488

PR #2495, opened 2021-05-12
Support html in markdown [AS-736]

zarsky-broad:
Supports sanitized html in markdown.


dinvlad (approved):

ansingh7115 (approved):
Had to learn a bit about xss sanitization, but it looks good to me!

PR #2498, opened 2021-05-13
Use ram disk for CI

zarsky-broad:
This switches jobs on circle to run in the ram disk that they attach to all docker executors. This makes disk-intensive steps complete in the same amount of time every run.


Shakespeared (approved):

dmohs (approved):
This sounds pretty interesting and may have positive implications for other projects.

PR #2499, opened 2021-05-17
Download via accessUrl if found BT-196

kshakir:
Checked functionality in my local Chrome browser using a combo of `yarn start`, terra-dev, and a hacked-up local copy of `getSignedUrlv1` and `martha_v3`. The latter cloud function was repeatedly modified to returned various `accessUrl` entries w/ and w/o [headers](https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.1.0/docs/#_accessurl).

<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/UriViewer.js
+++ b/src/components/UriViewer.js
@@ -120,7 +120,7 @@ const DownloadButton = ({ uri, metadata: { bucket, name, fileName, size }, acces
         const { url } = await Ajax(signal).Martha.getSignedUrl({
           bucket,
           object: name,
-          dataObjectUri: isDrs(uri) ? uri : null
+          dataObjectUri: isDrs(uri) ? uri : undefined
         })
         const workspace = workspaceStore.get()
         const userProject = await getUserProjectForWorkspace(workspace)
--- a/src/components/UriViewer.js
+++ b/src/components/UriViewer.js
@@ -1,3 +1,5 @@
+// noinspection RegExpUnnecessaryNonCapturingGroup
+
 import filesize from 'filesize'
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
@@ -5,7 +7,7 @@ import { div, h, img, input } from 'react-hyperscript-helpers'
 import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
 import Collapse from 'src/components/Collapse'
 import { ButtonPrimary, ClipboardButton, Link } from 'src/components/common'
-import { getUserProjectForWorkspace, parseGsUri } from 'src/components/data/data-utils'
+import { getDownloadCommand, getUserProjectForWorkspace, parseGsUri } from 'src/components/data/data-utils'
 import { spinner } from 'src/components/icons'
 import Modal from 'src/components/Modal'
 import DownloadPrices from 'src/data/download-prices'
@@ -54,6 +56,8 @@ const isFilePreviewable = ({ size, ...metadata }) => {
 
 const isGs = uri => _.startsWith('gs://', uri)
 
+const isDrs = uri => _.startsWith('dos://', uri) || _.startsWith('drs://', uri)
+
 const getMaxDownloadCostNA = bytes => {
   const nanos = DownloadPrices.pricingInfo[0].pricingExpression.tieredRates[1].unitPrice.nanos
   const downloadPrice = bytes * nanos / DownloadPrices.pricingInfo[0].pricingExpression.baseUnitConversionFactor / 10e8
@@ -99,17 +103,33 @@ const PreviewContent = ({ uri, metadata, metadata: { bucket, name }, googleProje
   ])
 }
 
-const DownloadButton = ({ uri, metadata: { bucket, name, fileName, size } }) => {
+const DownloadButton = ({ uri, metadata: { bucket, name, fileName, size }, accessUrl }) => {
   const signal = Utils.useCancellation()
   const [url, setUrl] = useState()
   const getUrl = async () => {
-    try {
-      const { url } = await Ajax(signal).Martha.getSignedUrl({ bucket, object: name, dataObjectUri: isGs(uri) ? undefined : uri })
-      const workspace = workspaceStore.get()
-      const userProject = await getUserProjectForWorkspace(workspace)
-      setUrl(knownBucketRequesterPaysStatuses.get()[bucket] ? Utils.mergeQueryParams({ userProject }, url) : url)
-    } catch (error) {
-      setUrl(null)
+    if (accessUrl && accessUrl.url) {
+      /*
+      NOTE: Not supporting downloading using `accessUrl.headers`:
+      - https://ga4gh.github.io/data-repository-service-schemas/preview/release/drs-1.1.0/docs/#_accessurl
+
+      If we want to support supplying `accessUrl.headers` here we'll probably need a bigger solution.
+      As of 2021-05-17 a google search turned up this c. 2018 result that mentioned something called `ServiceWorker`
+      - https://stackoverflow.com/questions/51721904/make-browser-submit-additional-http-header-if-click-on-hyperlink#answer-51784608
+       */
+      setUrl(_.isEmpty(accessUrl.headers) ? accessUrl.url : null)
+    } else {
+      try {
+        const { url } = await Ajax(signal).Martha.getSignedUrl({
+          bucket,
+          object: name,
+          dataObjectUri: isDrs(uri) && uri
+        })
+        const workspace = workspaceStore.get()
+        const userProject = await getUserProjectForWorkspace(workspace)
+        setUrl(knownBucketRequesterPaysStatuses.get()[bucket] ? Utils.mergeQueryParams({ userProject }, url) : url)
+      } catch (error) {
+        setUrl(null)
+      }
     }
   }
   Utils.useOnMount(() => {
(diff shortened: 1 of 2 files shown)


src/components/UriViewer.js
```diff
@@ -99,17 +103,33 @@ const PreviewContent = ({ uri, metadata, metadata: { bucket, name }, googleProje
...
+      As of 2021-05-17 a google search turned up this c. 2018 result that mentioned something called `ServiceWorker`
+      - https://stackoverflow.com/questions/51721904/make-browser-submit-additional-http-header-if-click-on-hyperlink#answer-51784608
+       */
+      setUrl(_.isEmpty(accessUrl.headers) ? accessUrl.url : null)
+    } else {
+      try {
+        const { url } = await Ajax(signal).Martha.getSignedUrl({
+          bucket,
+          object: name,
+          dataObjectUri: isDrs(uri) && uri
```
mcovarr: should be `uri && isDrs(uri)`?

mcovarr (approved):

src/components/UriViewer.js
kshakir: Nope. This is saying "if this is a DOS URI or a DRS URI, then feed it into the `getSignedUrlV1` field `dataObjectUri`, otherwise leave the `dataObjectUri` as undefined".

We can go over the `getSignedUrlV1` API in more detail and how it needs... help. For now, from a high level, `getSignedUrlV1` \*always\* expects to be passed a GCS bucket and GCS object. `getSignedUrlV1` _optionally_ accepts an additional DOS URI or DRS URI as `dataObjectUri`. If `dataObjectUri` is provided then [the credentials retrieved](https://github.com/broadinstitute/martha/blob/3.6.0/handlers/getSignedUrlV1.js#L9-L11) to generate the GCS signed URL are from one-of-the-Bonds instead of SAM.

zarsky-broad: I think in this case a ternary might be clearer than relying on the short-circuiting, since we actively want `undefined` in the falsy case: `dataObjectUri: isDrs(uri) ? uri : undefined`

src/components/UriViewer.js
```diff
@@ -1,11 +1,13 @@
+// noinspection RegExpUnnecessaryNonCapturingGroup
```
zarsky-broad: If this is for an intellij inspection, I'd rather not commit it into the code. Is it possible for you to just disable the inspection for yourself?

src/components/UriViewer.js
```diff
@@ -99,17 +103,33 @@ const PreviewContent = ({ uri, metadata, metadata: { bucket, name }, googleProje
...
   const [url, setUrl] = useState()
   const getUrl = async () => {
-    try {
-      const { url } = await Ajax(signal).Martha.getSignedUrl({ bucket, object: name, dataObjectUri: isGs(uri) ? undefined : uri })
-      const workspace = workspaceStore.get()
-      const userProject = await getUserProjectForWorkspace(workspace)
-      setUrl(knownBucketRequesterPaysStatuses.get()[bucket] ? Utils.mergeQueryParams({ userProject }, url) : url)
-    } catch (error) {
-      setUrl(null)
+    if (accessUrl && accessUrl.url) {
```
zarsky-broad: Can you use `accessUrl?.url`?

src/components/data/data-utils.js
```diff
@@ -36,6 +36,25 @@ const errorTextStyle = { color: colors.danger(), fontWeight: 'bold', fontSize: 1
 
 export const parseGsUri = uri => _.drop(1, /gs:[/][/]([^/]+)[/](.+)/.exec(uri))
 
+export const getDownloadCommand = (fileName, gsUri, accessUrl) => {
+  if (accessUrl && accessUrl.url) {
```
zarsky-broad: `accessUrl?.url`

src/components/data/data-utils.js
```diff
@@ -36,6 +36,25 @@ const errorTextStyle = { color: colors.danger(), fontWeight: 'bold', fontSize: 1
...
+    const headers = _.flow(
+      _.toPairs,
+      _.reduce(
+        (headers, [header, value]) => {
+          return `${headers}--header='${header}: ${value}' `
+        },
+        ''
+      )
+    )(accessUrl.headers || {})
```
zarsky-broad: `_.toPairs` is guarded against undefined and other falsy values, so you shouldn't have to provide an alternative here.

src/components/data/data-utils.js
```diff
@@ -36,6 +36,25 @@ const errorTextStyle = { color: colors.danger(), fontWeight: 'bold', fontSize: 1
 
 export const parseGsUri = uri => _.drop(1, /gs:[/][/]([^/]+)[/](.+)/.exec(uri))
 
+export const getDownloadCommand = (fileName, gsUri, accessUrl) => {
```
zarsky-broad, line 39: Since you seem to be accessing properties of `accessUrl` a lot, does it make sense to do some preemptive destructuring here and in `UriViewer`? Something like:

```js
const exampleFnDeclaration = (accessUrl: { url: accessUrl, headers: accessHeaders } = {}) => {}
```

(I added some renaming for clarity)

src/components/UriViewer.js
kshakir: Meybe. What version of ES does terra-ui use? And how can one tell?

src/components/data/data-utils.js
kshakir: ~Meybe? I feel like there was some comment about the number of variables to help the browser... I'll try to find the old comment that I may have misinterpreted.~

Edit: see below

src/components/UriViewer.js
kshakir: It is ultimately for any developer. It helps visibly linting. If the file is ✅ when you start, and not when you finish, you likely should fix your changes before committing.

We could also just fix the unused regex parts. BT is very ticket happy so I'll submit a Jira for follow up and then BT can refine, prioritize, etc.

src/components/data/data-utils.js
kshakir: Hunh. I saw `guarded` in the docs and had know idea what it meant. Thanks!

src/components/data/data-utils.js
kshakir: ~Found it- here's the previous comment about creating more variables: [derefed link](http://www.dereferer.org/?https%3A%2F%2Fgithub%2Ecom%2FDataBiosphere%2Fterra%2Dui%2Fpull%2F2233%23discussion%5Fr487900931).~

~So it had something to do with `state`, that may not be applicable here? TODO: At some point if I'm going to be ~tau~ tee or ~table~ pi shaped I should take some time and do a react course I guess...~

Edit: see below

kshakir: Ok, I think I see the difference. You suggestion is to add additional destructuring for only local function variables vs. additional destructuring into whatever React's `state` is... I think. I'm assuming `state` is lurking around the browser memory as the name implies.

src/components/UriViewer.js
zarsky-broad: Would it be crazy to just run the intention to fix the regex in this PR? I'd rather that than disabling an inspection, if we're going to be changing lines for it anyway.

src/components/UriViewer.js
zarsky-broad: https://create-react-app.dev/docs/supported-browsers-features
This was added in https://github.com/facebook/create-react-app/releases/tag/v3.3.0
Note, though, that it [should work in most browsers](https://caniuse.com/mdn-javascript_operators_optional_chaining), which is really what matters.

src/components/data/data-utils.js
zarsky-broad: Right; state sticks around from render to render, whereas this is just sugar quick access to contents of things, and can be handled by the optimization when code gets minified automagically if appropriate.

kshakir: Gave passing by name plus destructuring a shot. Had to figure out how to get rid of `null` parameter values so they would be defaulted. Found something on StackOverflow so let me know if there's something more modern?

src/components/UriViewer.js
```diff
@@ -122,7 +120,7 @@ const DownloadButton = ({ uri, metadata: { bucket, name, fileName, size }, acces
         const { url } = await Ajax(signal).Martha.getSignedUrl({
           bucket,
           object: name,
-          dataObjectUri: isDrs(uri) && uri
+          dataObjectUri: isDrs(uri) ? uri : null
```
zarsky-broad: I think we want undefined here, right? So that the field is just left off? Is there a reason it should be defined and null?

src/components/UriViewer.js
kshakir: Used `null` as an explicit value instead. I've been chastised by eslint for daring to explicitly use `undefined` before. Apparently, `undefined` is not a keyword and thus can be shadowed to do unexpected things: https://eslint.org/docs/rules/no-undefined

src/components/data/data-utils.js
```diff
@@ -36,6 +36,21 @@ const errorTextStyle = { color: colors.danger(), fontWeight: 'bold', fontSize: 1
...
+  const { url: httpUrl, headers: httpHeaders } = accessUrl || {}
+  if (httpUrl) {
+    const headers = _.flow(
+      _.toPairs,
+      _.reduce((acc, [header, value]) => `${acc}-H '${header}: ${value}' `, '')
+    )(httpHeaders)
+    const output = fileName ? `-o '${fileName}' ` : '-O '
+    return `curl ${headers}${output}'${httpUrl}'`
+  }
+  if (gsUri) {
```
zarsky-broad:
```suggestion
  }

  if (gsUri) {
```

src/components/UriViewer.js
```diff
@@ -212,24 +230,24 @@ const UriViewer = _.flow(
         ]),
         h(PreviewContent, { uri, metadata, googleProject }),
         els.cell([els.label('File size'), els.data(filesize(size))]),
-        els.cell([
+        !accessUrl && gsUri && els.cell([
```
```suggestion
        !accessUrl && !!gsUri && els.cell([
```

src/components/UriViewer.js
zarsky-broad: We use [this rule](https://eslint.org/docs/rules/no-shadow-restricted-names#disallow-shadowing-of-restricted-names-no-shadow-restricted-names) to make sure that you get `undefined` when you write `undefined`, so I'd prefer that you use `undefined` here, unless the server requires that the field be present but null.

zarsky-broad (approved):
Minor last notes

PR #2500, opened 2021-05-17
Sr v1 workspace integration tests

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -1,5 +1,5 @@
 const _ = require('lodash/fp')
-const { withRegisteredUser, withBilling, withWorkspace } = require('../utils/integration-helpers')
+const { withRegisteredUser, withBilling, withWorkspace, withV1Workspace } = require('../utils/integration-helpers')
 const { click, clickable, delay, signIntoTerra, findElement, waitForNoSpinners, select, fillIn, input, findIframe, findText, dismissNotifications } = require('../utils/integration-utils')
 
 
@@ -10,6 +10,18 @@ const testRunNotebookFn = _.flow(
   withBilling,
   withRegisteredUser
 )(async ({ workspaceName, page, testUrl, token }) => {
+  await testRunNotebookHelper(workspaceName, page, testUrl, token)
+})
+
+const testRunNotebookWithV1WorkspaceFn = _.flow(
+  withV1Workspace,
+  withBilling,
+  withRegisteredUser
+)( async ({ v1WorkspaceName, page, testUrl, token }) => {
+  await testRunNotebookHelper(v1WorkspaceName, page, testUrl, token)
+})
+
+const testRunNotebookHelper = async (workspaceName, page, testUrl, token) => {
   await page.goto(testUrl)
   await click(page, clickable({ textContains: 'View Workspaces' }))
   await signIntoTerra(page, token)
@@ -48,11 +60,18 @@ const testRunNotebookFn = _.flow(
   await fillIn(frame, '//textarea', 'print(123456789099876543210990+9876543219)')
   await click(frame, clickable({ text: 'Run' }))
   await findText(frame, '123456789099886419754209')
-})
+}
+
 const testRunNotebook = {
   name: 'run-notebook',
   fn: testRunNotebookFn,
   timeout: 20 * 60 * 1000
 }
 
-module.exports = { testRunNotebook }
+const testRunNotebookWithV1Workspace = {
+  name: 'run-notebook-v1-workspace',
+  fn: testRunNotebookWithV1WorkspaceFn,
+  timeout: 20 * 60 * 1000
+}
+
+module.exports = { testRunNotebook, testRunNotebookWithV1Workspace }
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -9,19 +9,19 @@ const testRunNotebookFn = _.flow(
   withWorkspace,
   withBilling,
   withRegisteredUser
-)(async ({ workspaceName, page, testUrl, token }) => {
-  await testRunNotebookHelper(workspaceName, page, testUrl, token)
+)(async ({ billingProject, workspaceName, page, testUrl, token }) => {
+  await testRunNotebookHelper(billingProject, workspaceName, page, testUrl, token)
 })
 
 const testRunNotebookWithV1WorkspaceFn = _.flow(
   withV1Workspace,
   withBilling,
   withRegisteredUser
-)( async ({ v1WorkspaceName, page, testUrl, token }) => {
-  await testRunNotebookHelper(v1WorkspaceName, page, testUrl, token)
+)( async ({ billingProject, v1WorkspaceName, page, testUrl, token }) => {
+  await testRunNotebookHelper(billingProject, v1WorkspaceName, page, testUrl, token)
 })
 
-const testRunNotebookHelper = async (workspaceName, page, testUrl, token) => {
+const testRunNotebookHelper = async (billingProject, workspaceName, page, testUrl, token) => {
   await page.goto(testUrl)
   await click(page, clickable({ textContains: 'View Workspaces' }))
   await signIntoTerra(page, token)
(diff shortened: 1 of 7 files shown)


integration-tests/utils/integration-helpers.js
```diff
@@ -74,14 +74,29 @@ const withUser = test => async args => {
   }
 }
 
+const addUserToV1Workspace = _.flow(withSignedInPage, withUserToken)(async ({ page, billingProject, v1WorkspaceName, email }) => {
+  await page.evaluate((email, billingProject, v1WorkspaceName) => {
+    return window.Ajax().Workspaces.workspace(billingProject, v1WorkspaceName).updateAcl([{ email, accessLevel: 'WRITER', canCompute: true }])
```
jdcanas, line 79: nit: It might be worth making a constants file (or use an existing one) for these access levels, see https://github.com/DataBiosphere/terra-ui/pull/2500/files#diff-ada5c6a1752c4db6c566600346028c476c874042c68d93f0d187b2e0a3a07e04R110 as well. I dislike scattered magic strings with external dependencies. We are trying to do the same for Leo strings in terra-ui too...

integration-tests/tests/run-notebook.js
```diff
@@ -48,11 +60,18 @@ const testRunNotebookFn = _.flow(
...
 }
 
-module.exports = { testRunNotebook }
+const testRunNotebookWithV1Workspace = {
+  name: 'run-notebook-v1-workspace',
+  fn: testRunNotebookWithV1WorkspaceFn,
+  timeout: 20 * 60 * 1000
+}
+
+module.exports = { testRunNotebook, testRunNotebookWithV1Workspace }
```
jdcanas, line 93: TOL: is it worth calling `testRunNotebook` `testRunNotebookWithV2Workspace` ?

jdcanas (approved):
Looks good overall when they pass, few nits. Worth mentioning I haven't had a chance to do much beyond mess with the tests a bit locally, so I'd take my approval with a grain of salt. 

As an aside, we are currently doing a large revamp of notebooks, so I expect all notebook related tests to change in terms of the UI flow. I don't see this logic changing at all though, it seems modular enough to be adapted easily.

integration-tests/tests/run-notebook.js
s-rubenstein: I am going to wait on that until CI takes a look! My thought is no, because it is whatever the default for creating a workspace is, not explicitly v2.

integration-tests/utils/integration-helpers.js
s-rubenstein: Agreed. I can add that.

integration-tests/tests/run-notebook.js
```diff
@@ -9,50 +9,80 @@ const testRunNotebookFn = _.flow(
...
+const testRunNotebookHelper = async (billingProject, workspaceName, page, testUrl, token) => {
   await page.goto(testUrl)
   await click(page, clickable({ textContains: 'View Workspaces' }))
   await signIntoTerra(page, token)
   await dismissNotifications(page)
   await findElement(page, clickable({ textContains: workspaceName }))
   await waitForNoSpinners(page)
   await click(page, clickable({ textContains: workspaceName }))
-  await delay(5000)
+  await delay(10000)
```
s-rubenstein, line 16: This was bumped up due to an issue with timing on asynchronous permissions updates (namely google granting bucket access)

integration-tests/utils/integration-helpers.js
s-rubenstein: I might add that as a separate tech debt ticket to avoid scope creep. We have a few of these scattered throughout the code-base as well, and I would want to hit those at the same time: https://broadworkbench.atlassian.net/browse/CA-1360

integration-tests/tests/run-workflow.js
```diff
@@ -28,38 +42,54 @@ const testRunWorkflowFn = _.flow(
...
+    }, { retries: 10, factor: 1 })
 
-  await click(page, navChild('data'))
-  await click(page, clickable({ textContains: 'test_entity' }))
-  await findInDataTableRow(page, testEntity.name, testEntity.attributes.input)
-})
+    await click(page, navChild('data'))
+    await click(page, clickable({ textContains: 'test_entity' }))
+    await findInDataTableRow(page, testEntity.name, testEntity.attributes.input)
+  } finally {
```
andy7i, line 57: (just curious) could we add a `if (!newEntity)` here? to skip some of this cleanup?

integration-tests/utils/integration-helpers.js
```diff
@@ -90,6 +105,14 @@ const addUserToBilling = _.flow(withSignedInPage, withUserToken)(async ({ page,
   console.info(`test user was added to the billing project with the role: ${!!billingUser && billingUser.role}`)
 })
 
+const removeUserFromV1Workspace = _.flow(withSignedInPage, withUserToken)(async ({ page, billingProject, v1WorkspaceName, email }) => {
+  await page.evaluate((email, billingProject, v1WorkspaceName) => {
+    return window.Ajax().Workspaces.workspace(billingProject, v1WorkspaceName).updateAcl([{ email, accessLevel: 'NO ACCESS' }])
+  }, email, billingProject, v1WorkspaceName)
+
+  console.info(`removed user from: ${v1WorkspaceName}`)
```
andy7i: is it worth logging the user's email here?

andy7i (approved):
👍 , couple questions but not blocking. Did not try running tests locally

integration-tests/tests/run-workflow.js
s-rubenstein: Yes, I can add that. It might speed up the tests in some places.

integration-tests/utils/integration-helpers.js
s-rubenstein: Sure! Why not!

zarsky-broad:
Can you give me a heads up when this is almost ready to merge so I can do a review pass? <3

s-rubenstein:
@zarsky-broad it is almost ready to merge. Just waiting for another review from @gpolumbo-broad

integration-tests/tests/run-workflow.js
```diff
@@ -1,6 +1,8 @@
+
+
```
zarsky-broad: Remove these?

integration-tests/tests/run-notebook.js
```diff
@@ -9,50 +9,82 @@ const testRunNotebookFn = _.flow(
...
+
+const testRunNotebookWithV1WorkspaceFn = _.flow(
+  withV1Workspace,
+  withBilling,
+  withRegisteredUser
+)(async ({ billingProject, v1WorkspaceName, page, testUrl, token }) => {
+  await testRunNotebookHelper(billingProject, v1WorkspaceName, page, testUrl, token, false)
+})
+
+const testRunNotebookHelper = async (billingProject, workspaceName, page, testUrl, token, newEntity = true) => {
```
zarsky-broad: I'd like to see this defined above its usages. Also, I'm not sure that `newEntity` is the right name for this, maybe like `isWorkspaceSingleUse`?

integration-tests/tests/run-workflow.js
```diff
@@ -12,11 +14,23 @@ const testRunWorkflowFn = _.flow(
...
+})
+
+const testRunWorkflowWithV1WorkspaceFn = _.flow(
+  withV1Workspace,
+  withRegisteredUser
+)(async ({ billingProject, page, testUrl, token, workflowName, v1WorkspaceName }) => {
+  await testRunWorkflowHelper(billingProject, page, testUrl, token, workflowName, v1WorkspaceName, false)
+})
+
+const testRunWorkflowHelper = async (billingProject, page, testUrl, token, workflowName, workspaceName, newEntity = true) => {
```
zarsky-broad: As in the other file, I'd prefer to see this defined above where it's invoked.

zarsky-broad (commented):
I don't know what this v1 workspace thing is about, so my comments aren't about that, just a few things. I think it might be worth making sure teams know that this ~doubles the amount of time it takes tests to run, I might think there was something wrong otherwise.

integration-tests/tests/run-notebook.js
s-rubenstein: Done

integration-tests/tests/run-workflow.js
s-rubenstein: Done

integration-tests/tests/run-workflow.js
s-rubenstein: Done

s-rubenstein:
@zarsky-broad I wouldn't expect time to double? Since don't we run tests in parallel? It might be a bit longer, but I would expect it to be small, not huge. (At least, `integration-tests/yarn test` runs them in parallel)

zarsky-broad:
> @zarsky-broad I wouldn't expect time to double? Since don't we run tests in parallel? It might be a bit longer, but I would expect it to be small, not huge. (At least, `integration-tests/yarn test` runs them in parallel)

The issue is that on circle we only have 2 cpu threads to work with, so our parallelism is very low; I'm working on another solution for that now. Comparing the time taken for the tests on this branch vs others, I'm seeing ~10 mins vs ~6:30

integration-tests/tests/run-notebook.js
gpolumbo-broad: I suggest leaving a comment to explain this.  I think the point here is that we expect `testRunNotebook` to run in a Workspace created via the latest mechanism (might be v1, v2 or some future version) and `testRunNotebookWithV1Workspace` is here while we transition to PPW, and after we've completed the PPW migration, we should no long need it.  I might also suggest creating a Jira issue in the PPW Cleanup epic to come back and clean-up/delete the tests no longer being used.

integration-tests/tests/run-workflow.js
```diff
@@ -28,38 +26,71 @@ const testRunWorkflowFn = _.flow(
...
 }
 
-module.exports = { testRunWorkflow }
+const testRunWorkflowWithV1Workspace = {
+  name: 'run-workflow-v1-workspace',
+  fn: testRunWorkflowWithV1WorkspaceFn,
+  timeout: 15 * 60 * 1000
+}
+
+module.exports = { testRunWorkflow, testRunWorkflowWithV1Workspace }
```
gpolumbo-broad, line 100: Same in this file, if one of these tests is intended to be around forever and one is only intended to be around until some external condition is met (i.e., PPW migration is done) then let's comment as such and leave a marker for how to find this and how to identify when it can be safely removed.

s-rubenstein: Filed https://broadworkbench.atlassian.net/browse/CA-1366

gpolumbo-broad (approved):

PR #2501, opened 2021-05-19
WIP: [CA-1359] A11y: Workspaces List Page

andrewmarcus:
This PR introduces a number of global improvements which specifically benefit the Workspaces List page and top-level page for an individual workspace.

### Skip Navigation Link

Most websites provide a link, first in the tab order on the page, allowing screen reader users to skip repetitive content at the top of the page. This is a standard design practice which users expect to be present. In this case there isn't a ton of content in the header, but it still seems useful to be able to skip it.

Because of the way React passes refs down to child components in order to be able to target elements for focusing, I decided it was important to encapsulate all of the skip functionality in as few places as possible, to avoid having to update every page to accept an addition forwarded ref. I was able to set it up entirely within the `TopBar`, so the skip link is first, and the skip target is last, and appears visually as focus on the entire page. The next tab stop is the first element below the header.

The one potential problem with this approach is that the header isn't entirely identical on every page. The `<h1>` is in it on most pages except the landing page, and it sometimes also contains contextual info such as breadcrumbs, a search bar or info about the billing project. Consequently a user could skip things that are relevant to the page without knowing it. However, these elements still appear visually and semantically within the header, so it seems reasonable. And it's a pattern users of assistive technologies expect.

### Tab Search Filters

The `SimpleTabBar` component is used on the Workspaces List page to render search filters for the workspaces table. Semantically, these are tabs: only one may be active at a time, and they cause only a portion of the page to redrawn when selected. I updated the semantic markup to make them function properly as tabs according to the ARIA specs.

In particular, I created a new reusable utility allowing a collection of `Interactive` elements to accept focus navigation using the arrow keys as well as Home/End, in either a vertical or horizontal orientation, by passing refs down into each child element. This seems to work like a charm! And it can be used in other places now too!

I also added `PropTypes` to the `SimpleTabBar` component to enforce that it has a `label` property for compliance with ARIA.

### Tab Navigation Bar

The `TabBar` component is used on each Workspace and Workflow page to render a navigation section at the top, pointing to different subpages within the workflow or workspace. Semantically, these are actually menu items: only one can be active at a time, and they cause the entire page to refresh with a different URL. It's essentially a submenu for the page, even though they visually appear to be tabs.

I previously added Tab semantic markup to them and updated the integration test. However, I concluded that they in fact encode menu items more accurately, so I essentially reverted them. They now support keyboard navigation just like `SimpleTabBar`, though.

I also added `PropTypes` to the `TabBar` component to enforce that it has a `label` property for compliance with ARIA. And I added labels to all other pages that use this component.

### Combobox components

The Comboboxes produced by the `react-select` module are somewhat accessible. They manage focus well, and include an `aria-live` section to announce changes the user makes. However, they do not identify themselves properly as [Combobox](https://www.digitala11y.com/combobox-role/) widgets, or implement the necessary [semantic elements](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html).

For almost all of the internal components I was able to add overrides which add proper ARIA roles and properties to the original renderers. However, I wasn't quite able to get it to work for widgets with `isSearchable == false` because they use a `DummyInput` component for which no overrides are possible. 

I've been thinking about reaching out to the maintainers of the `react-select` module to see if they can bake some of these improvements into their widget out of the box, but haven't done so yet.

### Tooltip Triggers

To be more accessible, I duplicated the tooltip text as an `aria-describedby` reference, so screen reader users will always be presented that text as an aside without needing to explicitly open the tooltip which isn't an obvious interaction.

### Tables without content

I found that, despite all of the ARIA cleanups I previously did with tables, they were still not semantically correct in the case where there was no content, because the message indicating this was just jammed in there without being wrapped properly into a row. I've fixed this and made it consistent across all of the different table components.


integration-tests/utils/integration-utils.js
```diff
@@ -106,7 +106,7 @@ const svgText = ({ textContains }) => {
 }
 
 const navChild = text => {
-  return `//*[@role="tablist"]/*[contains(normalize-space(.),"${text}")]`
```
andrewmarcus, line 109: I decided to make this test more robust by targeting the nav element by ID instead of by ARIA role. This allows the team more flexibility to refine the roles further in the future

src/pages/workspaces/List.js
```diff
@@ -197,8 +202,9 @@ export const WorkspaceList = () => {
             return div({ style: styles.tableCellContainer }, [
               div({ style: styles.tableCellContent }, [
                 h(Link, {
-                  style: { color: canView ? undefined : colors.dark(0.7), fontWeight: 600, fontSize: 16, ...Style.noWrapEllipsis },
+                  style: { color: canView ? undefined : colors.dark(0.8), fontWeight: 600, fontSize: 16, ...Style.noWrapEllipsis },
```
andrewmarcus, line 204: This change was to increase the color contrast to a sufficient degree

src/pages/workspaces/List.js
```diff
@@ -252,7 +258,7 @@ export const WorkspaceList = () => {
           },
           size: { basis: 120, grow: 1, shrink: 0 }
         }, {
-          headerRenderer: () => null,
+          headerRenderer: () => div({ className: 'sr-only' }, ['Actions']),
```
andrewmarcus, line 260: This addresses a specific issue identified by aXe in which a column header was missing text. In this case the text will only appear for screen reader users

src/pages/workspaces/List.js
```diff
@@ -262,14 +268,17 @@ export const WorkspaceList = () => {
             const lastRunStatus = workspaceSubmissionStatus(workspace)
 
 
-            return div({ style: { ...styles.tableCellContainer, paddingRight: 0 } }, [
+            return canView ? div({ style: { ...styles.tableCellContainer, paddingRight: 0 } }, [
```
andrewmarcus: I decided to remove the button to popup the menu unless the user can actually access it. There is much discussion online about how disabled buttons do not provide a good user experience since it isn't obvious how a user might go about enabling them. In this particular case, the menu button itself provides no meaning if the user can't interact with the workspace anyway

src/style.css
```diff
@@ -151,3 +151,27 @@ a {
 .token.variable {
   color: #b56000 !important; /* #e90; */
 }
+
+/* Restore focus outlines to the wrapper elements around RSelect boxes */
+[role="combobox"]:focus-within {
+  outline: -webkit-focus-ring-color auto 1px;
+}
```
andrewmarcus, line 158: For some inexplicable reason, `react-select` explicitly hides the focus ring from combo boxes. Keyboard users have no idea where they are... :(

src/components/SkipNavLink.js
```diff
@@ -0,0 +1,42 @@
...
+      border: '1px solid black',
+      borderColor: colors.accent(),
+      padding: '1em',
+      boxShadow: 'rgba(0, 0, 0, 0.5) 0px 0px 4px 0px',
+      zIndex: 9998
+    },
+    onClick: event => {
+      event.preventDefault()
+      event.stopPropagation()
+      ref.current && ref.current.focus()
```
zarsky-broad:
```suggestion
      ref.current?.focus()
```

src/components/TooltipTrigger.js
```diff
@@ -66,7 +66,8 @@ const Tooltip = ({ side = 'bottom', type, target: targetId, children, id, delay
 
   return h(PopupPortal, [
     div({
-      id, role: 'tooltip',
+      id,
+      role: 'tooltip',
```
zarsky-broad: Is there a reason for this line change? Or just sneaked in?

src/components/TopBar.js
```diff
@@ -328,16 +329,19 @@ const TopBar = ({ showMenu = true, title, href, children }) => {
       ])
     ])
   }
+  const mainRef = useRef()
 
-  return h(Fragment, [
+  return div({
+    role: 'banner'
+  }, [
```
zarsky-broad: I'd tell you to just collapse this to one line, for simplicity

src/components/common.js
```diff
@@ -118,56 +121,165 @@ export const ButtonOutline = ({ disabled, children, ...props }) => {
...
+ * @param getOnClick An optional click handler function, given the current tab
+ * @param label The ARIA label for the menu, which is required for accessibility
+ * @param tabProps Optionally, properties to add to each tab
+ * @param id Optionally, an ID to apply to the wrapper `nav` element so it can be easily found in the
+ *  integration tests. This defaults to 'tab-navigation', but you should change if it this ID would appear
+ *  more than once on a page
+ * @param children Children, which will be appended to teh end of the tab bar
+ * @param props Any additional properties to add to the container menu element
+ */
+export const TabBar = ({
```
zarsky-broad: I think it might be time to break these two out into something like `src/components/tabBars.js`, I'd say you might as well do it in this PR personally

src/components/common.js
```diff
@@ -118,56 +121,165 @@ export const ButtonOutline = ({ disabled, children, ...props }) => {
...
+      div({ style: { flexGrow: 1 } }),
+      children
+    ])
   ])
 }
+TabBar.propTypes = {
+  activeTab: PropTypes.string.isRequired,
+  tabNames: PropTypes.arrayOf(PropTypes.string).isRequired,
+  displayNames: PropTypes.arrayOf(PropTypes.string),
+  label: PropTypes.string.isRequired,
```
zarsky-broad: For requiring this sort of thing, you might want to look at our uses of `Utils.useConsoleAssert`, which we currently use specifically for ensuring that accessibility labels get added, since sometimes people don't develop with the console open, and it makes things explode

src/components/common.js
```diff
@@ -260,43 +372,80 @@ export const comingSoon = span({
...
-      boxShadow: 'none'
-    }),
-    singleValue: base => ({ ...base, color: colors.dark() }),
-    option: (base, { isSelected, isFocused, isDisabled }) => _.merge(base, {
-      fontWeight: isSelected ? 600 : undefined,
-      backgroundColor: isFocused ? colors.dark(0.15) : 'white',
-      color: isDisabled ? undefined : colors.dark(),
-      ':active': { backgroundColor: colors.accent(isSelected ? 0.55 : 0.4) }
+const commonSelectProps = (menuId, isOpen) => {
+  return {
```
zarsky-broad: I'd tell you to use the shorthand for returning a literal object here, `() => ({})`

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,93 @@
+import * as _ from 'lodash/fp'
```
zarsky-broad:
```suggestion
import _ from 'lodash/fp'
```

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,93 @@
...
+import { useRef } from 'react'
+import { h } from 'react-hyperscript-helpers'
+import { toIndexPairs } from 'src/libs/utils'
```
zarsky-broad: As a rule for utils.js, we import `* as Utils` and then call those as `Utils.foo`

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,93 @@
...
+
+
+export const withArrowKeyNavigation = ({ onKeyDown, children }) => {
+  const count = children.length
+  const refs = _.map(() => useRef(), _.range(0, count))
```
zarsky-broad: 1. Can you just do `_.map(useRef, _.range...)`?
2. Does this cause issues with React if the number of children changes between renders? That would change the number of hooks, right?

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,93 @@
...
+
+  const focusOn = index => {
+    // Wrap around the ends and ensure the number is positive
+    const i = index % count + (index < 0 ? count : 0)
+    refs[i].current && refs[i].current.focus()
```
zarsky-broad: `refs[i].current?.focus()`

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  // Pass a ref into each child, using an alternative property name
+  return _.map(([i, child]) => {
+    return h(child, { key: i, forwardedRef: refs[i], onKeyDown: onKeyDown(i, focusOn) })
```
zarsky-broad: This keying by index is likely to cause issues if the children are re-ordered or the number of children changes. Can we use something more durable?

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,93 @@
...
+ * @param children The children provided to this element.
+ *  These must be functions which accept a props object containing `forwardedRef` and `onKeyDown` properties.
+ * @returns An array of children for inclusion in higher-level elements
+ */
+export const withVerticalNavigation = children => {
+  return withArrowKeyNavigation({
+    children,
+    onKeyDown: (i, focusOn) => {
+      return event => {
+        switch (event.key) {
```
zarsky-broad, line 71: It might be a bit more consistent to use `Utils.switchCase`

src/components/table.js
```diff
@@ -275,7 +276,20 @@ export const FlexTable = ({
       },
       style: { outline: 'none' },
       onScroll: ({ scrollTop }) => onScroll(scrollTop),
-      noContentRenderer: () => div({ style: { marginTop: '1rem', textAlign: 'center', fontStyle: 'italic' } }, [noContentMessage]),
+      noContentRenderer: () => {
+        noContentMessage = noContentMessage || noContentRenderer() || 'Nothing to display'
```
zarsky-broad: I don't like reassigning this. Can you make a new `const` with a new name? Or inline this?

src/components/table.js
```diff
@@ -317,10 +332,15 @@ FlexTable.propTypes = {
  * A basic table with a header and flexible column widths. Intended for small amounts of data,
  * since it does not provide scrolling. See FlexTable for prop types.
  */
-export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, hoverHighlight, tableName, sort = null }) => {
+export const SimpleFlexTable = ({ columns, rowCount, noContentMessage, noContentRenderer = _.noop, hoverHighlight, tableName, sort = null }) => {
+  if (!rowCount && !noContentMessage) {
+    noContentMessage = noContentRenderer() || 'Nothing to display'
```
zarsky-broad: I don't like reassigning this. Can you make a new `const` with a new name?  Or default it to a function that returns this?

src/pages/workspaces/List.js
```diff
@@ -58,23 +58,28 @@ const WorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDelete }) =
   const canRead = workspace && Utils.canRead(workspace.accessLevel)
   const canShare = workspace?.canShare
   const isOwner = workspace && Utils.isOwner(workspace.accessLevel)
-  return h(Fragment, [
+  return div({
+    role: 'menu'
+  }, [
```
zarsky-broad: I'd tell you to collapse this into one line.

src/pages/workspaces/List.js
zarsky-broad: Instead of a ternary, you can just do `canView && div(...`

src/components/common.js
```diff
@@ -310,23 +459,31 @@ const formatGroupLabel = group => (
     }
   }, [group.label]))
 
-const BaseSelect = ({ value, newOptions, id, findValue, maxHeight, ...props }) => {
+const BaseSelect = ({ value, newOptions, id = null, findValue, maxHeight, ...props }) => {
   const newValue = props.isMulti ? _.map(findValue, value) : findValue(value)
+  const [isOpen, setOpen] = useState(false)
+  const menuId = Utils.useUniqueId()
+  const myId = Utils.useUniqueId()
+  const inputId = id || myId
```
s-rubenstein, line 312: Can you inline the definition of `myId` here? (`id || Utils.useUniqueId()`) Just curious, because that could avoid us generating the unique id if we already have an id to use, and avoids needing to name the variable.

src/components/keyboard-nav.js
andrewmarcus: Hmm, yes I suppose it would. FWIW, I haven't seen any situations in the Terra UI where that is the case. Even if we were to add vertical keyboard navigation to the little 3-item popup menus, the convention seems to be to disable items rather than removing them. So in theory, at least, there will always be the same number.

Is this something we should add support for? I agree it would be more robust if we do, but I'm not sure a good way around it while continuing to use refs.

src/components/keyboard-nav.js
andrewmarcus: Would it be legit to see if the child itself has a key property and if so use that, using the index as a fallback?

andrewmarcus: No, I guess that wouldn't work since these children are functions which haven't been rendered yet.

2 potential options:

1. Don't use `h` to create wrapper elements at all, and just call the chlidren directly as functions. You'd lose the convenient type coersion, but it's possible the key property wouldn't be necessary anymore.

2. Pass in a keyed object of children rather than an array. This would force the caller to create keys for all their elements, which may or may not make much sense in context. For tabs it would be fine since they already have keys, but for menu items it may not. It wouldn't solve the `useRef` issue if the number of children changes, however.

Are there any other options?

andrewmarcus: It appears that option 1 works and makes the key unnecessary, so I'm going to try that for now. Let me know if you'd prefer a different solution.

src/components/keyboard-nav.js
andrewmarcus: What's the advantage of `Utils.switchCase` over using native `switch` semantics?

src/components/SkipNavLink.js
andrewmarcus: ok

src/components/TooltipTrigger.js
andrewmarcus: will revert

src/components/TopBar.js
andrewmarcus: ok

src/components/keyboard-nav.js
andrewmarcus: Looks like with `Utils.switchCase` I would need to define the last case `'Enter'` or `' '` twice? Is that true?

src/components/common.js
andrewmarcus: Ha! This is exactly what I tried to do, but then React complained about the number of hooks changing. :)

andrewmarcus: `Utils.useUniqueId()` uses `useRef` under the hood

zarsky-broad:
FYI: if you update from dev, you should have more reliable and faster test runs going forward.

integration-tests/utils/integration-utils.js
```diff
@@ -106,7 +106,7 @@ const svgText = ({ textContains }) => {
...
+  return `//*[@id="tab-navigation"]//a[contains(normalize-space(.),"${text}")]`
```
zarsky-broad: I hate to poo-poo the idea, but we actually avoid using id's for the integration tests, on the basis that users don't target things by id, and we want to simulate user behavior. So, I'd rather target by aria role and then have to change it if we change the role in the future, rather than rely on something that's not exposed to the user.

andrewmarcus: Ok, that makes sense

gpolumbo-broad (approved):

src/components/TabBar.js
```diff
@@ -0,0 +1,186 @@
+import _ from 'lodash/fp'
```
zarsky-broad: We generally only name a file with a capital letter when it has a default export of a component with the same name; since there are two exports here, would you consider either merging them to have a prop that says which version someone wants, and then making it a default export, or changing the name to i.e. `tabBars.js`?

src/components/TabBar.js
```diff
@@ -0,0 +1,186 @@
...
+ * @param tabNames An array of keys for each tab
+ * @param displayNames An optional array of display names for each tab (otherwise the keys will be displayed)
+ * @param refresh If provided, a function to refresh the current tab
+ * @param getHref A function to get the href for a given tab
+ * @param getOnClick An optional click handler function, given the current tab
+ * @param label The ARIA label for the menu, which is required for accessibility
+ * @param tabProps Optionally, properties to add to each tab
+ * @param id Optionally, an ID to apply to the wrapper `nav` element so it can be easily found in the
+ *  integration tests. This defaults to 'tab-navigation', but you should change if it this ID would appear
+ *  more than once on a page
```
zarsky-broad: Since we're not using this, do we want to pull it out?

src/components/common.js
```diff
@@ -310,23 +303,31 @@ const formatGroupLabel = group => (
     }
   }, [group.label]))
 
-const BaseSelect = ({ value, newOptions, id, findValue, maxHeight, ...props }) => {
+const BaseSelect = ({ value, newOptions, id = null, findValue, maxHeight, ...props }) => {
```
zarsky-broad: I don't _think_ you should need to default `id` to `null`?

src/components/keyboard-nav.js
zarsky-broad: Legit.

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,92 @@
+import _ from 'lodash/fp'
+import { useRef } from 'react'
+import * as Utils from 'src/libs/utils'
+
+
+export const withArrowKeyNavigation = ({ onKeyDown, children }) => {
+  const count = children.length
+  const refs = _.map(useRef, _.range(0, count))
```
zarsky-broad, line 8: I think the Right Thing To Do here is to return a div with a ref attached to it, and use that ref to access its children, which can then be indexed into; we'd just have to support passing props to the wrapper div.

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const focusOn = index => {
+    // Wrap around the ends and ensure the number is positive
+    const i = index % count + (index < 0 ? count : 0)
+    refs[i].current?.focus()
+  }
+
+  // Pass a ref into each child, using an alternative property name
+  return _.map(([i, child]) => {
```
zarsky-broad, line 17: Then if we wrap this in a div, I think the right thing to do might be to wrap this in a Fragment, so we're not passing the result of a map directly to a dom element.

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,92 @@
...
+    return child({ forwardedRef: refs[i], onKeyDown: onKeyDown(i, focusOn) })
```
zarsky-broad, line 18: Maybe `cloneElement`?

src/components/TabBar.js
andrewmarcus: OOHH! 💡  I was wondering why there was a difference in capitalization!  Thanks for explaining. I think I'll just rename the file. :)

src/components/keyboard-nav.js
andrewmarcus: I think I'm following. I'd keep a single ref to the wrapper div, which would returns a raw 'HTMLElement', and then access the children just using the DOM's `children` property?

When I first tried just using `document.querySelector` to find the element and call `focus()` on it directly, I could find the element but the focus didn't seem to transfer. So I thought maybe React doesn't let you fully break out into the DOM to manage focus state, or at least you're not supposed to. But I don't see why it shouldn't work. I'll give it a try.

src/components/keyboard-nav.js
andrewmarcus: This might be moot if I end up rewriting the ref passing. But... why add a clone here? This should be the only place the element is used, right? How would returning a clone rather than the original affect the ref?

src/components/keyboard-nav.js
andrewmarcus: Hmm. I'm trying to attach the keydown event to the wrapper div rather than each individual child. But then I have to figure out which child triggered the event, and the actual Clickable element might be several levels down in the child. I could recursively scan the entire DOM tree below the wrapper element until I find the `event.target`, but that doesn't seem like the React way to do things...

andrewmarcus: Figured it out!  Addressed in #2512 (I'll close this PR and use those ones instead)

andrewmarcus:
I've replaced this PR with 3 new ones: #2511, #2512, #2513, #2515

PR #2502, opened 2021-05-19
Wait for workspace list to load when testing

zarsky-broad:
I noticed that one difference between these tests, which often fail because they don't actually open the workspace when they think they've clicked it in the list, and those that _don't_ have this problem, is that they actually wait to make sure there's no spinners...


s-rubenstein (approved):
LGTM

s-rubenstein:
I assume that the waitForNoSpinners is a noop if there are already none on the page? Is there any reason to not bake that into our click method? I assume we shouldn't click if there are spinners?

breilly2 (approved):
👍   Nice! Thanks for finding this!  🎉

zarsky-broad:
@s-rubenstein It is indeed a no-op. In theory we could bake it into everything, but the intent when writing the tests is to be aware of what the user experiences, so I us to put it in explicitly when it's needed.

PR #2503, opened 2021-05-20
split integration tests between multiple runners

zarsky-broad:
I've gone with 4 runners for now, somewhat arbitrarily, but I know there are more tests planning to be added and I want to leave a bit of headroom.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,7 @@ commands:
           command: |
             mkdir -p ${SCREENSHOT_DIR}
             TESTS_TO_RUN=$(circleci tests glob "jest/**.js" | circleci tests split --split-by=timings)
-            yarn test --ci --maxWorkers=2 --reporters=default --reporters=jest-junit $TESTS_TO_RUN
+            yarn test $TESTS_TO_RUN --ci --maxWorkers=2 --reporters=default --reporters=jest-junit
       - store_test_results:
           path: /tmp/test-results
       - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,7 +95,8 @@ commands:
           no_output_timeout: 16m
           command: |
             mkdir -p ${SCREENSHOT_DIR}
-            yarn test --ci --maxWorkers=2 --reporters=default --reporters=jest-junit
+            TESTS_TO_RUN=$(circleci tests glob "jest/**.js" | circleci tests split --split-by=timings)
+            yarn test --ci --maxWorkers=2 --reporters=default --reporters=jest-junit $TESTS_TO_RUN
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -95,7 +95,8 @@ commands:
           no_output_timeout: 16m
           command: |
             mkdir -p ${SCREENSHOT_DIR}
-            yarn test --ci --maxWorkers=2 --reporters=default --reporters=jest-junit
+            TESTS_TO_RUN=$(circleci tests glob "jest/**.integration-test.js" | circleci tests split --split-by=timings)
```
dmohs, line 98: What is this `circleci` command and what does it do?

dmohs (approved):

.circleci/config.yml
zarsky-broad: See: https://circleci.com/blog/a-guide-to-test-splitting/

zarsky-broad: (and further https://circleci.com/docs/2.0/parallelism-faster-jobs/)

zarsky-broad: Basically, we tell it all of the test files and have it break them up into the number of groups specified by the `parallelism` key below, using historical timing data from previous test results as the basis for the split.

Shakespeared (approved):

s-rubenstein (approved):
Looks great! Nice find!

integration-tests/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "jest-junit": {
     "suiteName": "Integration tests",
-    "addFileAttribute": true,
+    "addFileAttribute": "true",
```
breilly2, line 19: This doesn't seem right. Does `jest-junit` really want the string `"true"`?

.circleci/config.yml
```diff
@@ -95,7 +95,8 @@ commands:
...
+            yarn test $TESTS_TO_RUN --ci --maxWorkers=2 --reporters=default --reporters=jest-junit
```
breilly2, line 99: Do we still want `--maxWorkers=2`?

breilly2 (commented):
You have plenty of thumbs, so I'll just post comments.

integration-tests/package.json
zarsky-broad: Stunningly, yes. https://github.com/jest-community/jest-junit#configuration specifies that it must be a string for this particular config key. I think it's because it's being passed through to some underlying thing.

.circleci/config.yml
zarsky-broad: Yes. Each container has 2 cpu threads available, and will use them if the container is assigned more than one test.

PR #2504, opened 2021-05-21
Sr billing page changes

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -23,7 +23,10 @@ import ProjectDetail from 'src/pages/billing/Project'
 import validate from 'validate.js'
 
 
-export const ownerRole = 'Owner'
+export const billingRoles = {
+  owner: 'Owner',
+  user: 'User'
+}
 
 const styles = {
   projectListItem: selected => {
@@ -48,7 +51,7 @@ const ProjectListItem = ({ project: { projectName, role, creationStatus, message
     hover: Style.navList.itemHover(isActive)
   }, [projectName, !projectReady && statusIcon])
   const unselectableProject = div({ style: { ...styles.projectListItem(isActive), color: colors.dark() } }, [projectName, !projectReady && statusIcon])
-  return div([_.includes(ownerRole, role) && projectReady ? selectableProject : unselectableProject])
+  return div([_.includes(billingRoles.owner, role) && projectReady ? selectableProject : unselectableProject])
 }
 
 const billingProjectNameValidator = existing => ({
@@ -237,7 +240,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
   })
 
   const checkOwner = () => {
-    !isOwner && _.map(project => _.includes(ownerRole, project.role) ? setIsOwner(true) : null, billingProjects)
+    !isOwner && _.map(project => _.includes(billingRoles.owner, project.role) ? setIsOwner(true) : null, billingProjects)
   }
 
   const showCreateProjectModal = async () => {
@@ -313,14 +316,14 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
           _.map(project => h(ProjectListItem, {
             project, key: project.projectName,
             isActive: !!selectedName && project.projectName === selectedName
-          }), _.filter(project => _.includes(ownerRole, project.role), billingProjects))
+          }), _.filter(project => _.includes(billingRoles.owner, project.role), billingProjects))
         ]),
 
         h(BillingProjectSubheader, { title: 'Shared with You' }, [
           _.map(project => h(ProjectListItem, {
             project, key: project.projectName,
             isActive: !!selectedName && project.projectName === selectedName
-          }), _.filter(project => !_.includes(ownerRole, project.role), billingProjects))
+          }), _.filter(project => !_.includes(billingRoles.owner, project.role), billingProjects))
         ])
       ]),
       creatingBillingProject && h(NewBillingProjectModal, {
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -23,7 +23,7 @@ import ProjectDetail from 'src/pages/billing/Project'
 import validate from 'validate.js'
 
 
-const ownerRole = 'Owner'
+export const ownerRole = 'Owner'
 
 const styles = {
   projectListItem: selected => {
@@ -43,7 +43,7 @@ const ProjectListItem = ({ project: { projectName, role, creationStatus, message
       style: { color: colors.danger(), margin: '0 1rem 0 0.5rem' }, side: 'right'
     }, [div({ style: { wordWrap: 'break-word', whiteSpace: 'pre-wrap' } }, [message || 'Error during project creation.'])])
   const selectableProject = h(Clickable, {
-    style: { ...styles.projectListItem(isActive), color: colors.accent() },
+    style: { ...styles.projectListItem(isActive), color: isActive ? colors.dark() : colors.accent() },
     href: `${Nav.getLink('billing')}?${qs.stringify({ selectedName: projectName, type: 'project' })}`,
     hover: Style.navList.itemHover(isActive)
   }, [projectName, !projectReady && statusIcon])
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -14,7 +14,7 @@ import * as Nav from 'src/libs/nav'
 import * as StateHistory from 'src/libs/state-history'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
-import { ownerRole } from 'src/pages/billing/List'
+import { billingRoles } from 'src/pages/billing/List'
 
 
 const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace }) => div({
@@ -41,7 +41,7 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
   const [showBillingModal, setShowBillingModal] = useState(false)
   const [selectedBilling, setSelectedBilling] = useState()
   const [workspaces, setWorkspaces] = useState(() => StateHistory.get().projectUsers || null)
-  const [tab, setTab] = useState(query.tab || 'users')
+  const [tab, setTab] = useState(query.tab || 'workspaces')
 
   const signal = Utils.useCancellation()
 
@@ -56,8 +56,8 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
       div({ style: { flexGrow: 1 } },
         _.map(member => {
           return h(MemberCard, {
-            adminLabel: ownerRole,
-            userLabel: 'User',
+            adminLabel: billingRoles.owner,
+            userLabel: billingRoles.user,
             member, adminCanEdit,
             onEdit: () => setEditingUser(member),
             onDelete: () => setDeletingUser(member)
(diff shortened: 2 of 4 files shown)


src/pages/billing/List.js
```diff
@@ -23,7 +23,7 @@ import ProjectDetail from 'src/pages/billing/Project'
 import validate from 'validate.js'
 
 
-const ownerRole = 'Owner'
+export const ownerRole = 'Owner'
```
marctalbott: Could we do this with the `userRole` as well? Seems nicer to me than duplicating strings in Project.js

marctalbott (commented):
A few questions:

Would it be possible to add column headers and format the last modified like we do in the workspace list? Or is that out of scope for this ticket? I know there's another more workspace focused ticket.
Screen Shot 2021-05-21 at 11 23 07 AM

Not sure if this is because I've lost access to the billing account, but it doesn't show me what the current billing account on the project is, there's just a pencil below the name of the billing project 
Screen Shot 2021-05-21 at 11 26 00 AM

Should the selected tab persist when the page is reloaded? It reverts to the users tab for me when I reload

src/pages/billing/List.js
s-rubenstein: Absolutely, will do.

s-rubenstein (commented):
I was trying to do minimal possible styling that it was functional for release. https://broadworkbench.atlassian.net/browse/CA-1255 outlines the actual finalizing of that table. There's an argument that that work should just happen here, but since we have separate tickets, I figured it would be useful to keep the change as small as possible.

marctalbott (approved):
Would it be possible to have the selected tab persist instead of defaulting to the Workspaces tab? If I'm looking at the Users tab and I reload the page, it reverts back to the Workspaces tab, but it could be nice for it to just reopen the same tab. Same goes for when I'm clicking between different billing projects -- it might be nice to have the same tab open in the new project that I had open in the previous project.

Not a blocking comment

s-rubenstein:
@marctalbott done!

src/pages/billing/Project.js
```diff
@@ -58,7 +114,9 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
     setDeletingUser(false)
     setUpdating(false)
     setEditingUser(false)
-
+    const rawWorkspaces = await Ajax(signal).Workspaces.list()
```
gpolumbo-broad: This might take some time, possibly many seconds depending on the Billing Project.  Will this code be resilient to that?

s-rubenstein: I think it should be be independent of the billing project, because it grabs all workspaces. This is currently our approach on the workspaces list page, and should be as resilient as any other load.

That said, there's an argument here that we could pull this up a level higher, and do the filtering on project select, plumbing this in as a prop. (thus loading once, and effectively caching)

However, I wanted to: A) Keep it as close to the usage as possible) and B) Ensure that users can reload workspaces, if one is created while they have the page open. (Which they can do by just clicking the tab again). I thought that likely won out on the tradeoffs for me, until such a time as we hear complaints about the loading speed on this page, or the endpoint gets slow enough we start failing to load/erroring frequently.

src/pages/billing/Project.js
```diff
@@ -10,12 +11,25 @@ import * as Auth from 'src/libs/auth'
...
 import * as Utils from 'src/libs/utils'
+import { billingRoles } from 'src/pages/billing/List'
 
 
+const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace }) => div({
+  style: Style.cardList.longCard
+}, [
+  div({ style: { flex: 1 } }, [workspace.name]),
+  div({ style: { flex: '0 0 350px' } }, [workspace.createdBy]),
+  div({ style: { flex: 'none' } }, [workspace.lastModified])
```
dmohs: Is this preferred over not specifying at all?

src/pages/billing/List.js
```diff
@@ -23,7 +23,10 @@ import ProjectDetail from 'src/pages/billing/Project'
 import validate from 'validate.js'
 
 
-const ownerRole = 'Owner'
+export const billingRoles = {
+  owner: 'Owner',
+  user: 'User'
+}
```
dmohs, line 29: 👍

src/pages/billing/Project.js
```diff
@@ -58,7 +114,9 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
     setDeletingUser(false)
     setUpdating(false)
     setEditingUser(false)
-
+    const rawWorkspaces = await Ajax(signal).Workspaces.list()
+    console.log(rawWorkspaces)
```
dmohs:
```suggestion
```

src/pages/billing/Project.js
```diff
@@ -58,7 +114,9 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
+    setWorkspaces(_.map(workspaceResponse => workspaceResponse.workspace, _.filter(workspaceResponse => workspaceResponse.workspace.namespace === projectName, rawWorkspaces)))
```
dmohs: This is a personal preference, so please feel free to ignore, but I think the variable name `workspaceResponse` here is long-enough that it takes up more space than it deserves considering its function as just a way to access nested data. I would shorten it to `response` or `wr` or even `x`. I use `x` when (a) the context of its use is extremely narrow as it is in this case and (b) to communicate that the name is really, really unimportant. In this case, it may be quite clear from locality that `rawWorkspaces` is a workspace response object, so calling that out is less helpful to the reader than highlighting (via percentage of space taken) what is happening with it.

dmohs (approved):

src/pages/billing/Project.js
```diff
@@ -27,9 +41,51 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
+    workspaces: div({ style: { flexGrow: 1 } }, [
+      _.map(workspace => h(WorkspaceCard, { workspace }), workspaces)
+    ]),
+    users: h(Fragment, [
+      h(NewUserCard, {
+        onClick: () => setAddingUser(true)
+      }),
+      div({ style: { flexGrow: 1 } },
+        _.map(member => {
+          return h(MemberCard, {
```
petesantos, line 68: I believe you will need to provide a `key` here and at line 51 for the mapped over elements to prevent the warning:
`Warning: Each child in a list should have a unique "key" prop.`

src/pages/billing/Project.js
zarsky-broad: We do have a `useWorkspaces` hook.

src/pages/billing/Project.js
s-rubenstein: I had assumed so, given that is the existing pattern here: https://github.com/DataBiosphere/terra-ui/blob/dev/src/components/group-common.js#L59.

As an aside, there's an argument to abstracting this styling, but I could see a point where the workspace display diverged from the group display, so didn't want to pre-maturely optimize that.

src/pages/billing/Project.js
s-rubenstein: In general, CI has had the practice of verbose variable names for understanding, over conciseness. However, I think `response` is just as clear as `workspaceResponse`, and cuts down the verbosity considerably, so I think that is a clear win. Will do.

gpolumbo-broad (approved):

src/pages/billing/Project.js
```diff
@@ -1,21 +1,37 @@
...
+}, [
+  div({ style: { flex: 1 } }, [workspace.name]),
+  div({ style: { flex: '0 0 350px' } }, [workspace.createdBy]),
+  div({ style: { flex: 'none' } }, [workspace.lastModified])
+]))
+
 const ProjectDetail = ({ project, project: { projectName, creationStatus }, billingAccounts, authorizeAndLoadAccounts }) => {
   // State
+  const { query } = Nav.useRoute()
+  const { workspaces, refresh: refreshWorkspaces } = useWorkspaces()
```
s-rubenstein, line 41: Using this method removes the option to use `StateHistory` that I noticed we use for project owners. I didn't see any noticeable loading differences, so it is probably fine, but I wanted to call it out. Looks like we reload project owners on component mount anyway, so it is just a display cache from what I can tell.

src/pages/billing/Project.js
```diff
@@ -1,21 +1,37 @@
...
 import * as StateHistory from 'src/libs/state-history'
+import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
+import { billingRoles } from 'src/pages/billing/List'
 
 
+const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace }) => div({
+  style: Style.cardList.longCard
+}, [
+  div({ style: { flex: 1 } }, [workspace.name]),
```
s-rubenstein: I noticed that for long workspace names, this expands off the page (and, in fact, pushes the other text off of the card.) However, this seems to be an existing issue with the pattern I copied this from with `MemberCard`, so it seems okay to leave it for now. We should examine this general pattern for https://broadworkbench.atlassian.net/browse/CA-1256 and https://broadworkbench.atlassian.net/browse/CA-1255, because it seems like an issue, but because it is existing, and I am trying to maintain this as minimal set of table changes, that should probably be in the following tickets.

src/pages/billing/Project.js
```diff
@@ -27,9 +51,51 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
   const [billingAccountName, setBillingAccountName] = useState(null)
   const [showBillingModal, setShowBillingModal] = useState(false)
   const [selectedBilling, setSelectedBilling] = useState()
+  const [tab, setTab] = useState(query.tab || 'workspaces')
 
   const signal = Utils.useCancellation()
 
+  const tabToTable = {
+    workspaces: div({ style: { flexGrow: 1, width: '100%' } }, [
+      _.map(([i, workspace]) => h(WorkspaceCard, { workspace, key: i }), Utils.toIndexPairs(_.map(response => response.workspace, _.filter(response => response.workspace.namespace === projectName, workspaces))))
```
zarsky-broad, line 60: Two things, for the future:
1. We shouldn't use index as the key, it's unsafe because if something is inserted in the middle of a list, the index and the item won't be matched anymore.
2. When mapping to just get a property, there's a shorthand: `_.map('workspace', workspaces)`

src/pages/billing/Project.js
```diff
@@ -81,23 +149,26 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
 
 
   // Render
-  const adminCanEdit = _.filter(({ roles }) => _.includes('Owner', roles), projectUsers).length > 1
-  const { displayName = null } = _.find({ accountName: billingAccountName }, billingAccounts) || {}
+  const adminCanEdit = _.filter(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers).length > 1
+  const { displayName = null } = _.find({ accountName: billingAccountName }, billingAccounts) || { displayName: 'No Access' }
```
zarsky-broad, line 153: Why not just change the default?
```js
const { displayName = 'No Access' } = _.find({ accountName: billingAccountName }, billingAccounts) || {}
```

src/pages/billing/Project.js
```diff
@@ -27,9 +51,51 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
+      _.map(([i, workspace]) => h(WorkspaceCard, { workspace, key: i }), Utils.toIndexPairs(_.map(response => response.workspace, _.filter(response => response.workspace.namespace === projectName, workspaces))))
+    ]),
+    users: h(Fragment, [
+      h(NewUserCard, {
+        onClick: () => setAddingUser(true)
+      }),
+      div({ style: { flexGrow: 1 } },
+        _.map(([i, member]) => {
+          return h(MemberCard, {
+            key: i,
```
zarsky-broad, line 69: Key here shouldn't be the index again, I'd encourage you to use `'member'` if it's a string.

src/pages/billing/Project.js
s-rubenstein: To me, this felt semantically more correct, because it is the name being displayed? (Also, the logic gets more complex down below, because we want the `Billing Account` prefix to show up regardless, so this folds that logic in cleanly.

s-rubenstein (commented):
Hi @zarsky-broad, saw your comments! I went ahead and merged since you weren't a primary reviewer on the PR before you responded again. I will do the index changes as a follow up, although to be honest, I think worrying about something being inserted into the middle of a list is not a huge problem, because the rendering of the cards here should be redone when the list changes, so it would only be a problem if someone did state mutations without using setState (which is an unsafe practice)

That said, I don't expect us to need to remember original order (the benefits to using index) because I don't think that users will need to clear all sorting here, especially since default sorting is probably going to be alphabetical by name anyhow.

PR #2505, opened 2021-05-21
No multiple runners for overnight tests

zarsky-broad:
Multiple runners for the overnight tests were reporting multiple times to slack. Since it doesn't matter if the overnight tests take a bit longer, remove them.


ehanna4 (approved):
Makes sense to me

breilly2 (approved):

PR #2506, opened 2021-05-21
[CA-1365] A11y: Increased the banner header color contrast to get to a 4.5:1

andrewmarcus:
The aXe and ANDI tools both fail to measure the color contrast of the top banner of the Terra UI primary theme, because of the presence of a background image. However, if I disable the image but keep the background color, then they determine that it only has a color contrast of 3.3:1, not the 4.5:1 which is required by accessibility guidelines.

Unfortunately, since the text is already white and can’t be lightened any further, the only way to increase this ratio is to make the background of the banner significantly darker. Since this is present on every page, it does change the look of the page overall in a noticeable way. Darkening the banner also causes the overlaid honeycomb image to pop out significantly more.

Note that this only applies to the primary (green) theme. The Firecloud, Data Biosphere, etc. themes do not have background images in the banner, and also do not have color contrast issues because they use dark text on light backgrounds.

### What does this PR do?

It appears that React merges the `backgroundColor` and `background` properties on the same element automatically. Consequently it is difficult to disable the background image while retaining the color, for the purposes of evaluation. To make this easier, I split them into two separate elements: the parent element has a background color, and next element down has a background image overlay. Now it is trivial to disable the background image in Dev Tools before running the aXe or ANDI analysis.

This PR also make a few additional color contrast modifications: 
* Disabled workspaces on the Workspace List page are now a little darker, and I made them italic so they continue to stand out from active rows
* The Config Overrides block that shows up at the bottom when testing themes is now a bit darker. But normal users will probably never see this anyway

### Before

Screen Shot 2021-05-21 at 11 37 43 AM

### After

Screen Shot 2021-05-21 at 11 36 44 AM


zarsky-broad:
We should check with UX that they don't want to change the glow color for the logo background; also it looks like the left hex background image has shifted left, I think it's not supposed to be behind the menu icon.

andrewmarcus:
> We should check with UX that they don't want to change the glow color for the logo background; also it looks like the left hex background image has shifted left, I think it's not supposed to be behind the menu icon.

Yep! Agree on all counts.

You're right, I did shift the left hex background because it looked really weird when I didn't, now that the contrast is so much greater. But I can certainly put it back if the UX team would prefer that.

Perhaps the UX team can provide me with 2 new honeycomb images, to reduce the texture contrast back closer to what it was?

gpolumbo-broad (approved):

zarsky-broad:
When you sub in the new assets, remember to run the yarn optimize svgs script, please and thank you!

src/pages/workspaces/List.js
```diff
@@ -197,7 +197,11 @@ export const WorkspaceList = () => {
             return div({ style: styles.tableCellContainer }, [
               div({ style: styles.tableCellContent }, [
                 h(Link, {
-                  style: { color: canView ? undefined : colors.dark(0.7), fontWeight: 600, fontSize: 16, ...Style.noWrapEllipsis },
+                  style: {
+                    color: canView ? undefined : colors.dark(0.8),
+                    fontStyle: canView ? undefined : 'italic',
+                    fontWeight: 600, fontSize: 16, ...Style.noWrapEllipsis
+                  },
```
dmohs:
```suggestion
                  style: {
                    ...(canView ? {} : { color: colors.dark(0.8), fontStyle: 'italic' }),
                    fontWeight: 600, fontSize: 16, ...Style.noWrapEllipsis
                  },
```

dmohs (approved):
One tiny cleanup suggestion.

andrewmarcus:
Updated screenshot

Screen Shot 2021-06-14 at 9 37 05 AM

andrewmarcus:
> When you sub in the new assets, remember to run the yarn optimize svgs script, please and thank you!

@zarsky-broad Hmm, I just did ran the optimizer script, and it bumped the opacity value from `0.15` back down to `0.1`. Is there a way to preserve that value through the optimizer?

zarsky-broad:
Ugh. @andrewmarcus it's because the optimizer has `-p 1`, rounding decimals to a precision of one place. We could change that, but I'm trying to think of an alternative solution. Is opacity of `0.2` a no-go?

zarsky-broad:
I suppose one alternative would be to have them with no opacity set on them and then apply them with a floating element underneath the real one, and apply opacity there ourselves on the new element? Then we could just conditionally render the element at all.

s-rubenstein (approved):
LGTM

andrewmarcus:
> Ugh. @andrewmarcus it's because the optimizer has `-p 1`, rounding decimals to a precision of one place. We could change that, but I'm trying to think of an alternative solution. Is opacity of `0.2` a no-go?

@zarsky-broad In my opinion 0.1 was too subtle an effect. I'm going to try 0.2 and see how that looks.

andrewmarcus:
Here's 20% opacity (0.2)

Screen Shot 2021-06-14 at 10 22 53 PM

PR #2507, opened 2021-05-21
[IA-2696] Create analysis tab and perform some initial migration

jdcanas:
This PR begins to the IA analysis tab migration, which involves migrating everything from the `notebooks` tab into a more `analysis` focused model. To facilitate this migration, this PR creates a new tab that is not visible by default. It also duplicates a lot of existing components/service calls that are designed to be notebook-specific, and generalizes them for all analyses. The duplication is necessary to not affect existing functionality while we work on this. 

To see the wireframes this implements for the base analysis page, see: https://broadworkbench.atlassian.net/browse/IA-2696. You can see the linked epic here too for more info on the migration.

In terms of migration not explicitly related to the wireframes, there is seemingly a lot of lines in this PR, so it may be easier to go through it with the lens of what it actually accomplishes. In other words, when reviewing, comparing the 'new' components to their existing notebook counterparts and evaluating only the difference can simplify the task. Mobbing to review this PR may be a good team exercise to spread knowledge of the various components. 
- duplicates and migrates all notebook specific components in `notebook-utils.js`. Most of these are modals that appear from the `...` menu on each analysis 'row' in the AnalysisTab
- completely migrates the entire `Notebook.js` page into `Analysis.js` (which represents the contents of the tab itself), with all bucket calls and wording generalized
     - as a result, the analysis tab now provides all functionality it does for `.ipynb` to `.rmd` files (upload/export/rename, display in the analysis list, searchability, etc)
- begins to migrate `NotebookLauncher.js` to `AnalysisLauncher.js`. This launcher is used from `Analysis.js` (as `NotebookLauncher.js` is from `Notebook.js`), but is largely non-functional for `.rmds` There are a lot of TODOs slated for other tickets

I propose we delay all style review for the sake of iteration expediency, and not re-doing styles due to wireframe iteration, until we are closer to actually releasing these pages to users. (There will be a dedicated ticket for PO/UX review/style touch-up prior to release). There are also other tickets for testing the existing `run-notebook` flows in the analysis tab once we are further along.


src/components/notebook-utils.js
```diff
@@ -48,8 +49,16 @@ export const notebookNameValidator = existing => ({
     message: 'already exists'
   }
 })
+// removes leading dirs and a file ext suffix on paths
+export const getDisplayName = name => stripExtension(getFileName(name))
```
jdcanas: There's a lot of inconsistent usage of files (curse Javascript lack of typing...). Some places we pass around the file in the form `notebooks/fileName.ext`, some places `fileName.ext`, and some places `fileName`. This PR keeps all params in the exact same format, and adds some utilities to make it a bit clearer what exactly is being used where, but doesn't address any consistency to minimize changes/complexity overhead since these strings are used in conjunction with bucket/welder calls and displayed in various ways. We may want to clean this up a bit at some point.

src/components/notebook-utils.js
```diff
@@ -58,6 +67,9 @@ export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput,
   }
 })
 
+export const getAnalysisType = name => name.endsWith('.ipynb') ? 'python' : 'r'
```
jdcanas: I miss scala case classes here :(. Not sure if there's a cleaner way to map from `analysisType` to the various use cases (i.e. the file extension associated with an analysisType) other than a bunch of `Utils.switchCase`

package.json
```diff
@@ -24,9 +24,10 @@
     "iframe-resizer": "^4.3.2",
     "igv": "2.3.5",
     "jszip": "^3.6.0",
-    "lodash": "^4.17.21",
+    "lodash": "^4.17.20",
```
zarsky-broad: why are you downgrading this?

package.json
```diff
@@ -24,9 +24,10 @@
...
     "marked": "^2.0.3",
     "outdated-browser-rework": "^3.0.1",
+    "path": "^0.12.7",
```
zarsky-broad: I'd rather not add this in, since it's only being used for one minor thing.

src/components/notebook-utils.js
```diff
@@ -48,8 +49,16 @@ export const notebookNameValidator = existing => ({
...
 })
+// removes leading dirs and a file ext suffix on paths
+export const getDisplayName = name => stripExtension(getFileName(name))
 
-export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput, {
+// removes all paths up to and including the last slash
+export const getFileName = name => path.basename(name)
+
+// removes a file extension from a string
+export const stripExtension = name => name.replace(/\.[^/.]+$/, '')
```
zarsky-broad: Something like this might be simpler
```suggestion
export const getFileName = _.replace(/^.+\/.+\//, '')

// removes a file extension from a string
export const stripExtension = _.replace(/\.[^/.]+$/, '')
```

src/components/notebook-utils.js
zarsky-broad: Yeah, you'd do a switchcase on the last chunk of the name. I'd encourage you to use the same method of selection between this and `getAnalysisFileExtension`, whichever way you do it.

src/components/notebook-utils.js
```diff
@@ -58,6 +67,9 @@ export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput,
   }
 })
 
+export const getAnalysisType = name => name.endsWith('.ipynb') ? 'python' : 'r'
+
+export const getAnalysisFileExtension = analysisType => Utils.switchCase(analysisType, ['r', () => 'rmd'], ['python', () => 'ipynb'])
```
zarsky-broad: Another option would be to make an object for mapping:
```suggestion
const typeExtensionMapping = { r: 'rmd', python: 'ipynb' }
export const getAnalysisFileExtension = analysisType => typeExtensionMapping[analysisType]
```

src/images/rstudio-logo.svg
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 22.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
zarsky-broad: I strongly encourage you to run `yarn optimize-image-svgs` to clean this up.

src/libs/ajax.js
```diff
@@ -1012,6 +1033,73 @@ const Buckets = signal => ({
...
+  analysis: (namespace, bucket, name, analysisType) => {
+    const bucketUrl = `storage/v1/b/${bucket}/o`
+
+    const encodeFileName = n => Utils.switchCase(analysisType,
+      ['python', () => nbName(getDisplayName(n))], ['r', () => rName(getDisplayName(n))])
+
+    const calhounPath = () => Utils.switchCase(analysisType,
+      ['python', () => 'api/convert'], ['r', () => 'api/convert/rmd'])
+
+    const mimeType = () => Utils.switchCase(analysisType, ['python', () => 'application/x-ipynb+json'], ['r', () => 'text/plain'])
```
zarsky-broad: Is there a reason for these to be functions that are called rather than values that are just evaluated?

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -40,9 +40,13 @@ const WorkspaceTabs = ({ namespace, name, workspace, activeTab, refresh }) => {
     { name: 'dashboard', link: 'workspace-dashboard' },
     { name: 'data', link: 'workspace-data' },
     { name: 'notebooks', link: 'workspace-notebooks' },
+    // the spread operator results in no array entry if the config value is false
+    // we want this feature gated until it is ready for release
+    ...(isAnalysisTabVisible() ? [{ name: 'analysis', link: 'workspace-analysis' }] : []),
     { name: 'workflows', link: 'workspace-workflows' },
     { name: 'job history', link: 'workspace-job-history' }
   ]
+
```
zarsky-broad: Any reason for this change?

src/pages/workspaces/workspace/notebooks/ExportNotebookModal.js
```diff
@@ -14,9 +19,10 @@ import * as Nav from 'src/libs/nav'
 import * as Utils from 'src/libs/utils'
 import validate from 'validate.js'
 
-
+//TODO: deprecate once notebook tab is removed
```
zarsky-broad, line 23: You need another blank line here, we have 2 blank lines between imports and the rest of the file.

src/pages/workspaces/workspace/notebooks/ExportNotebookModal.js
```diff
@@ -29,7 +35,6 @@ const ExportNotebookModal = ({ fromLauncher, onDismiss, printName, workspace })
   const signal = Utils.useCancellation()
   const { workspaces } = useWorkspaces()
 
-
```
zarsky-broad: We generally leave 2 lines between named sections of a component, just to make it stand out.

zarsky-broad (changes requested):

zarsky-broad:
You may want to run the new image through https://tinypng.com/.

package.json
jdcanas: oops, must have messed up the merge

src/components/notebook-utils.js
jdcanas: Hmm... it would be elegant to do this, and then `_.invert` the object for extension to type mapping. I think I'll keep it as switch cases though, since it will allow us to be more flexible with handling new types (namely, we can use the default case as we're implementing new types if applicable)

src/libs/ajax.js
jdcanas: ah yeah, I think they originally were somewhere else, doesn't make sense with this refactor though 👍

src/components/notebook-utils.js
jdcanas: The get file name regex didn't work for me... point taken though, removed the import and found another solution
image

src/components/notebook-utils.js
```diff
@@ -48,8 +48,18 @@ export const notebookNameValidator = existing => ({
     message: 'already exists'
   }
 })
+// removes leading dirs and a file ext suffix on paths
+export const getDisplayName = name => stripExtension(getFileName(name))
```
zarsky-broad: Since `stripExtension` and `getFileName` aren't used anywhere else, you should remove them.
```suggestion
export const getDisplayName = str => /([^/]+?)(?:\.[^.]+)?$/.exec(str)[1]
```

Generally, we'd avoid a regex, since it's complicated to parse for other developers; I'm suggesting it here only because it requires trimming before and after the string.

src/components/notebook-utils.js
```diff
@@ -48,8 +48,18 @@ export const notebookNameValidator = existing => ({
...
 
-export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput, {
+// removes all paths up to and including the last slash
+export const getFileName = path => path.split('/').pop()
+
+// removes a file extension from a string
+export const stripExtension = _.replace(/\.[^/.]+$/, '')
+
+export const getExtension = name => name.split('.').pop()
```
zarsky-broad: To avoid mutating anything, and for consistency, consider this:
```suggestion
export const getExtension = _.flow(_.split('.'), _.last)
```

zarsky-broad (changes requested):
This PR is very big and hard to parse. I don't have time to go over all the things I'd like to see touched here, unfortunately. Please find a more experienced developer to help you work through this.

src/components/notebook-utils.js
jdcanas: `getFileName` is used in `Analysis.js` (probably missed a ctrl+f because it's collapsed by default by github). I like the file utilities being separate too to support a possible refactor I discuss here https://github.com/DataBiosphere/terra-ui/pull/2507/files#r637038725

zarsky-broad: In that case, here's my suggestion:
```suggestion
export const getDisplayName = _.flow(stripExtension, getFileName)
```

src/components/notebook-utils.js
```diff
@@ -48,8 +48,18 @@ export const notebookNameValidator = existing => ({
...
 })
+// removes leading dirs and a file ext suffix on paths
+export const getDisplayName = name => stripExtension(getFileName(name))
 
-export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput, {
+// removes all paths up to and including the last slash
+export const getFileName = _.flow(_.split('/'), _.last)
+
+// removes a file extension from a string
+export const stripExtension = _.replace(/\.[^/.]+$/, '')
```
zarsky-broad, line 57: If you define this after `getExtension`, it can be:

```js
const stripExtension = str => str.slice(0, -(getExtension(str).length + 1)) // + 1 to account for the period
```

src/components/notebook-utils.js
kyuksel: Punting on cleaning up this sort of thing for the time being while we get our functional ducks in order makes sense. Can we file tech debt tickets for them so we have references to go back to when the functional migration is done? I'd vote for adding the tickets to the UI revamp epic so it is easier to group them.

src/components/notebook-utils.js
```diff
@@ -58,6 +68,9 @@ export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput,
   }
 })
 
+export const getAnalysisType = fileName => Utils.switchCase(getExtension(fileName), ['ipynb', () => 'python'], ['rmd', () => 'r'])
```
kyuksel: Should `python` and `r` be `notebook` and `rstudio`, respectively? `ipynb` files can also correspond to R notebooks, and `rmd` files are specific to `RStudio`.

Similarly for [rName](https://github.com/DataBiosphere/terra-ui/pull/2507/files#diff-d23f5bf16a31795a4bfd83c17503ba4f7e6ddae66c73283f13f4eb95d1cfe410R145) in `ajax.js`.

src/libs/ajax.js
```diff
@@ -919,6 +921,24 @@ const Buckets = signal => ({
     return _.filter(({ name }) => name.endsWith('.ipynb'), items)
   },
 
+  listRmd: async (namespace, name) => {
```
kyuksel: `listRmds` for consistency?

kyuksel (commented):
To be continued... 

Only minor comments inline so far. Will pick it up tomorrow morning.

One high-level question: Have we defined what an _analysis_ is? From the code, it seems like we refer to notebook and RStudio _artifacts_ as _analyses_, which would throw me off if I were new. The new tab will also be named _Analyses_, which will entail more than just artifacts/files. I'm asking because now is probably the best time to get on the same page as far as terminology and so we can be consistent going forward 😊

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -40,6 +40,9 @@ const WorkspaceTabs = ({ namespace, name, workspace, activeTab, refresh }) => {
     { name: 'dashboard', link: 'workspace-dashboard' },
     { name: 'data', link: 'workspace-data' },
     { name: 'notebooks', link: 'workspace-notebooks' },
+    // the spread operator results in no array entry if the config value is false
+    // we want this feature gated until it is ready for release
+    ...(isAnalysisTabVisible() ? [{ name: 'analysis', link: 'workspace-analysis' }] : []),
```
kyuksel: good to know :)

src/pages/workspaces/workspace/notebooks/ExportNotebookModal.js
```diff
@@ -15,8 +20,10 @@ import * as Utils from 'src/libs/utils'
 import validate from 'validate.js'
 
 
+//TODO: deprecate once notebook tab is removed
 const cutName = name => name.slice(10, -6) // removes 'notebooks/' and the .ipynb suffix
 
+//TODO: deprecate once notebook tab is removed. See ExportAnalysisModal for replacement
 const ExportNotebookModal = ({ fromLauncher, onDismiss, printName, workspace }) => {
```
kyuksel, line 27: Why are we updating this function given it'll be deprecated?

src/pages/workspaces/workspace/notebooks/ExportNotebookModal.js
```diff
@@ -102,7 +109,107 @@ const ExportNotebookModal = ({ fromLauncher, onDismiss, printName, workspace })
         ])]),
         h(IdContainer, [id => h(Fragment, [
           h(FormLabel, { htmlFor: id, required: true }, ['Name']),
-          notebookNameInput({
+          analysisNameInput({
+            error: Utils.summarizeErrors(Utils.summarizeErrors(errors?.newName)),
```
kyuksel: Why is it necessary to call `summarizeErrors()` twice?

src/components/notebook-utils.js
```diff
@@ -48,8 +48,18 @@ export const notebookNameValidator = existing => ({
...
+export const getDisplayName = name => stripExtension(getFileName(name))
 
-export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput, {
+// removes all paths up to and including the last slash
+export const getFileName = _.flow(_.split('/'), _.last)
+
+// removes a file extension from a string
+export const stripExtension = _.replace(/\.[^/.]+$/, '')
+
+export const getExtension = _.flow(_.split('.'), _.last)
```
kyuksel: Would it be good to unit-test the string parsing functions above?

src/libs/routes.js
```diff
@@ -45,11 +47,13 @@ const routes = _.flatten([
   Dashboard.navPaths,
   Data.navPaths,
   Notebooks.navPaths,
+  Analysis.navPaths,
```
kyuksel, line 50: `Analyses` instead of `Analysis`, analogously to `Notebooks`?

kyuksel (commented):
fancy  🎉 

image

src/pages/workspaces/workspace/Analysis.js
```diff
@@ -0,0 +1,469 @@
...
+        div({ style: { flex: 2 } }),
+        !_.isEmpty(analyses) && h(DelayedSearchInput, {
+          'aria-label': 'Search analyses',
+          style: { marginRight: '0.75rem', width: 220 },
+          placeholder: 'Search analyses',
+          onChange: setFilter,
+          value: filter
+        }),
+        !_.isEmpty(analyses) && h(IdContainer, [id => h(Fragment, [
+          label({ htmlFor: id, style: { marginLeft: 'auto', marginRight: '0.75rem' } }, ['Sort By:']),
```
kyuksel: `Sort By` text seems to have gotten misaligned for some reason:

image

kyuksel: I know these are not important for the time being but we tend to forget them later…would be good to change the menu items like `Most Recently Updated` to `Most recently updated` so they have the same casing as every other text on the page.

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -0,0 +1,545 @@
...
+        isOpen: createOpen,
+        workspace,
+        runtimes,
+        persistentDisks,
+        onDismiss: () => {
+          chooseMode(undefined)
+          setCreateOpen(false)
+        },
+        onSuccess: _.flow(
+          withErrorReporting('Error creating runtime'),
```
kyuksel, line 80: Should we start removing the `runtime` wording, at least where it's user-facing? As far as I know, going forward, the main terms will be:
- app
- cloud environment
- cloud compute
- persistent disk

kyuksel: We could also create a ticket to remind us to change text and variable names in one fell swoop via a PR that contains no functional changes, which would make reviewing easier.

src/pages/workspaces/workspace/Analysis.js
```diff
@@ -0,0 +1,469 @@
...
+    busy && spinnerOverlay
+  ])])
+})
+
+export const navPaths = [
+  {
+    name: 'workspace-analysis',
+    path: '/workspaces/:namespace/:name/analysis',
+    component: Analysis,
+    title: ({ name }) => `${name} - Analysis`
```
kyuksel:
```suggestion
    path: '/workspaces/:namespace/:name/analyses',
    component: Analyses,
    title: ({ name }) => `${name} - Analyses`
```

kyuksel (commented):
LGTM overall. Tested by clicking around the page. Some inline comments/questions, mostly regarding naming. It'd be nice to set a good precedent by using forward-looking terminology consistently 😊

src/components/notebook-utils.js
jdcanas: Hmm... I could see it either way. I chose the analysis type to be the language the analysis is in. Don't feel strongly either way

src/pages/workspaces/workspace/notebooks/ExportNotebookModal.js
jdcanas: Ah, this is most likely something I messed up while merging in changes from dev. It was an annoying merge since I had to apply the changes to the duplicated components as well...

src/pages/workspaces/workspace/notebooks/ExportNotebookModal.js
jdcanas: So I leave ExportNotebookModal as is, and add `ExportAnalysisModal`.  Any updates to `ExportNotebookModal` are exclusively from renaming things used in both places (that don't functionally affect the existing modal) and/or merging in changes from dev

src/libs/routes.js
jdcanas: Ah, the tab name was incorrect as `Analysis` too, I updated all references to `Analysis` to `Analyses` where applicable

src/pages/workspaces/workspace/Analysis.js
jdcanas: Yeah, I didn't spend too much time here since eventually this bar will go away entirely in favor of table based sorting. See https://broadworkbench.atlassian.net/browse/IA-2700 . Also, the ticket notes that the wireframes themselves are inconsistent in how this should look, so I very much did not want to touch it this iteration (working on that in the background...) :) .

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
jdcanas: Yeah, I think this is a good idea to do after we've completed most of the migration. Do you want to make the ticket in the epic? I'm not sure if that list is exhaustive... (noteboook -> analysis and notebooks -> analyses come to mind as a good addition to the list)

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -28,495 +28,495 @@ import { wrapWorkspace } from 'src/pages/workspaces/workspace/WorkspaceContainer
...
-export const navPaths = [
-  {
+export const navPaths = [{
     name: 'workspace-notebook-launch',
     path: '/workspaces/:namespace/:name/notebooks/launch/:notebookName',
     component: NotebookLauncher,
     title: ({ name, notebookName }) => `${notebookName} - ${name}`
-  }
-]
+}]
```
kyuksel: missing new line 
```suggestion
}]

```

src/components/notebook-utils.js
kyuksel: Then I'd suggest we have a mapping from `file type` to `application` because
1. `ipynb` can map to both `python` and `r`
2. There may be apps in the near future where language is N/A (e.g. visualization tools such as Genome Browser)

kyuksel: Also a TOL: For the reason #2 above, I feel using the terms `application` and `artifact` instead of `analysis` may be more future-proof.

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
kyuksel: https://broadworkbench.atlassian.net/browse/IA-2799

kyuksel (changes requested):
1. Resolved comments that were addressed
2. Added a few new comments 
3. Linting needed
4. Something seems to be broken in nav paths

image

src/components/notebook-utils.js
jdcanas: hmm.. I think it would be, but there are only stub tests currently... I'll make a unit test ticket in the epic with this and runtime-utils as candidates for testing

src/components/notebook-utils.js
```diff
@@ -58,6 +69,9 @@ export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput,
   }
 })
 
+export const getTool = fileName => Utils.switchCase(getExtension(fileName), ['ipynb', () => 'Jupyter'], ['rmd', () => 'RStudio'])
```
jdcanas: make the tool a constant

src/libs/ajax.js
```diff
@@ -1012,6 +1033,75 @@ const Buckets = signal => ({
...
+  },
+
+  analysis: (namespace, bucket, name, analysisType) => {
+    const bucketUrl = `storage/v1/b/${bucket}/o`
+
+    const calhounPath = Utils.switchCase(analysisType,
+      ['python', () => 'api/convert'], ['r', () => 'api/convert/rmd'])
+
+    const mimeType = Utils.switchCase(analysisType,
+      ['python', () => 'application/x-ipynb+json'], ['r', () => 'text/plain'])
```
jdcanas: change keys to use constants from util function

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -0,0 +1,469 @@
...
+      width: 400,
+      flexDirection: 'column',
+      padding: 0
+    }
+)
+
+const noWrite = 'You do not have access to modify this workspace.'
+
+const sortTokens = {
+  lowerCaseName: notebook => notebook.name.toLowerCase()
```
jdcanas, line 73: rename

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -0,0 +1,469 @@
...
+    style: _.merge({
+      ...Style.elements.card.title, whiteSpace: 'normal', overflowY: 'auto'
+    }, listView ? {
+      marginLeft: '1rem', flexGrow: 1
+    } : { height: 60, padding: '1rem' })
+  }, getDisplayName(name))
+
+
+  const appIconSrc = Utils.switchCase(analysisType, ['python', () => jupyterLogo], ['r', () => rLogo])
+  const appLabel = Utils.switchCase(analysisType, ['python', () => 'Jupyter'], ['r', () => 'RStudio'])
```
jdcanas: use analysis type constant

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -0,0 +1,469 @@
...
+  //TODO: does this prevent users from making an .rmd with the same name as an .ipynb?
+  const existingNames = _.map(({ name }) => getDisplayName(name), analyses)
+
+  const refreshAnalyses = _.flow(
+    withRequesterPaysHandler(onRequesterPaysError),
+    withErrorReporting('Error loading analyses'),
+    Utils.withBusyState(setBusy)
+  )(async () => {
+    const notebooks = await Ajax(signal).Buckets.listNotebooks(namespace, bucketName)
+    const enhancedNotebooks = _.map(notebook => _.merge(notebook, { analysisType: 'python' }), notebooks)
```
jdcanas: use constant here for analysisType

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -0,0 +1,469 @@
...
+  //TODO: eventually load app artefacts
+  // const doAppRefresh = _.flow(
+  //   withErrorReporting('Error loading Apps'),
+  //   Utils.withBusyState(setBusy)
+  // )(refreshApps)
+
+  const uploadFiles = Utils.withBusyState(setBusy, async files => {
+    try {
+      await Promise.all(_.map(async file => {
+        const name = file.name.endsWith('.ipynb') ? file.name.slice(0, -6) : file.name.slice(0, -4)
```
jdcanas: use the get displayName switchcase

yarn.lock
```diff
@@ -8473,6 +8473,14 @@ path-type@^4.0.0:
   resolved "https://registry.npmjs.org/path-type/-/path-type-4.0.0.tgz"
   integrity sha512-gDKb8aZMDeD/tZWs9P6+q0J9Mwkdl6xMV8TjnGP3qJVJ06bdMgkbBlLU8IdfOsIsFz2BW1rNVT3XuNEl8zPAvw==
 
+path@^0.12.7:
```
jdcanas: should be no diff in this (uninstall properly)

src/pages/workspaces/workspace/Analyses.js
kyuksel: `Workflows.js` has an identical `sortTokens` functions. Does it make sense to move it to a common place so we have a single copy instead of three?

kyuksel (approved):
👍🏼

src/components/notebook-utils.js
```diff
@@ -58,6 +69,13 @@ export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput,
   }
 })
 
+export const tools = { rstudio: { label: 'RStudio', ext: 'rmd' }, jupyter: { label: 'Jupyter', ext: 'ipynb' } }
```
rtitle: Is the `ext` case insensitive? The RStudio extension is generally `.Rmd` (capital R).

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -0,0 +1,478 @@
...
+    load()
+  })
+
+  useEffect(() => {
+    StateHistory.update({ analyses, sortOrder, filter })
+  }, [analyses, sortOrder, filter])
+
+  const noAnalysisBanner = div([
+    div({ style: { fontSize: 48 } }, ['A place for all your analyses ']),
+    div({ style: { display: 'flex', flexDirection: 'row', justifyContent: 'center' } }, [
```
rtitle: It could be just me, but I feel like it's more vertically centered in the wireframes.

image

vs

image

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  // Render
+  return h(Dropzone, {
+    accept: `.${tools.jupyter.ext}, .${tools.rstudio.ext}`,
+    disabled: !Utils.canWrite(accessLevel),
+    style: { flexGrow: 1 },
+    activeStyle: { backgroundColor: colors.accent(0.2), cursor: 'copy' },
+    onDropRejected: () => reportError('Not a valid analysis file',
+      'The selected file is not a .ipynb notebook file or an .rmd rstudio file. Ensure your file has the proper extension.'),
```
rtitle: Should probably say `.Rmd` and `RStudio`

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -0,0 +1,545 @@
+import * as clipboard from 'clipboard-polyfill/text'
```
rtitle, line 1: Wondering if we should make a new `src/pages/workspaces/workspace/analysis` directory and start populating files there?

(Also thinking about where to put new files related to the context bar..)

src/components/notebook-utils.js
jdcanas: hmm... it is. I've just been using test files i made with `.rmd`, ill fix this

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
jdcanas: probably worth it.. theres a ticket I made already to re-sort the util files and the modals scattered across a few different places. I'll add some details there for this so we can do some re-structuring at a safer point when there's less going on in parallel (merge conflicts are really risky with copy/pasted duplicated components, when there are relevant changes to the base components...)

jdcanas: Updated the description here https://broadworkbench.atlassian.net/browse/IA-2807

src/pages/workspaces/workspace/Analyses.js
jdcanas: Accomplished this by setting the parent div to 100% height, and then setting the child div height to 80% (to vertically align and account for the fixed buttons at the top)

dmohs (approved):
Reviewed via meeting.

src/components/notebook-utils.js
```diff
@@ -49,7 +49,18 @@ export const notebookNameValidator = existing => ({
   }
 })
 
-export const notebookNameInput = ({ inputProps, ...props }) => h(ValidatedInput, {
+// removes all paths up to and including the last slash
+export const getFileName = _.flow(_.split('/'), _.last)
+
+export const getExtension = _.flow(_.split('.'), _.last)
+
+// removes a file extension from a string
```
dmohs: For this comment and the one above, I judge the code does a better (and more maintainable) job of saying the same thing. I doubt this adds enough value to pay for the space it consumes, so I'd suggest removal.

jdcanas: Yeah, was much more relevant when it was a regex

rtitle (approved):

jdcanas:
It seems that the same 4-6 integration tests are failing across PRs due to timeouts from other services. The previous commit of this passed, I just changed some style of a non-visible component in 1 line, so I'm confident this isn't introducing failures.

zarsky-broad (approved):
Not re-reviewing, but you've already gone through things with other folks! Thanks!

PR #2508, opened 2021-05-24
AS-757: Fix editing and deleting entities for entity types that are equivalent when ignoring case

ansingh7115:
Case sensitivity for entity types is interesting! 

In the rawls db, entity types are saved preserving their case. But entity names have to be unique across entity types ignoring case. Meaning that if I have an entity of type "sample" and name "entity1" already saved, I cannot create another entity of type "sAmple" and name "entity1" - it fails due to a db uniqueness violation. But I _can_ create an entity of type "sAmple" and name "entity2".

The UI fixes here are minimal - just enough to fix the errors the user is seeing without changing expected behavior. The fix involves sending in the entity's actual type (with the correct case) when deleting or editing the entity. 

In the future, we may want to reassess entity type cases - though the backend preserves cases for entity types, the UI does not. For example, an entity of type "sample" and type "sAmple" are shown as the same type in the UI.

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -327,7 +327,7 @@ const DataTable = props => {
       onDismiss: () => setViewData(undefined)
     }, [div({ style: { maxHeight: '80vh', overflowY: 'auto' } }, [displayData(viewData)])]),
     renamingEntity !== undefined && h(EntityRenamer, {
-      entityType: _.flow(_.filter(entity => entity.name == renamingEntity), _.first, _.get('entityType'))(entities),
+      entityType: _.flow(_.filter(entity => entity.name === renamingEntity), _.first, _.get('entityType'))(entities),
       entityName: renamingEntity,
       workspaceId,
       onSuccess: () => {
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -327,7 +327,7 @@ const DataTable = props => {
       onDismiss: () => setViewData(undefined)
     }, [div({ style: { maxHeight: '80vh', overflowY: 'auto' } }, [displayData(viewData)])]),
     renamingEntity !== undefined && h(EntityRenamer, {
-      entityType: _.flow(_.filter(entity => entity.name === renamingEntity), _.first, _.get('entityType'))(entities),
+      entityType: _.find(entity => entity.name === renamingEntity, entities).entityType,
       entityName: renamingEntity,
       workspaceId,
       onSuccess: () => {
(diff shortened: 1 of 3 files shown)


src/components/data/DataTable.js
```diff
@@ -327,7 +327,8 @@ const DataTable = props => {
       onDismiss: () => setViewData(undefined)
     }, [div({ style: { maxHeight: '80vh', overflowY: 'auto' } }, [displayData(viewData)])]),
     renamingEntity !== undefined && h(EntityRenamer, {
-      entityType, entityName: renamingEntity,
+      entityType: _.flow(_.filter(entity => entity.name === renamingEntity), _.first, _.get('entityType'))(entities),
```
zarsky-broad: I think you want `_.find` for these.

src/components/data/data-utils.js
```diff
@@ -152,8 +152,10 @@ export const EntityDeleter = ({ onDismiss, onSuccess, namespace, name, selectedE
   const [additionalDeletions, setAdditionalDeletions] = useState([])
   const [deleting, setDeleting] = useState(false)
 
+  const selectedKeys = _.keys(selectedEntities)
+
   const doDelete = async () => {
-    const entitiesToDelete = _.concat(_.map(entityName => ({ entityName, entityType: selectedDataType }), selectedEntities), additionalDeletions)
+    const entitiesToDelete = _.concat(_.map(entity => ({ entityName: entity['name'], entityType: entity['entityType'] }), selectedEntities), additionalDeletions)
```
zarsky-broad: For a literal string key, just do `entity.name`. The brackets are for when the key needs to be interpreted in some way.

ungwudik (approved):
LGTM.

I don't think we have a test for this scenario (or that we want to add one), but I assume you manually tested (or plan to test) it as part of the changes in this PR?

ansingh7115:
Yes, I have tested the deleting and editing functionality locally. we don't have a UI test for it because the scenario where it breaks is extremely rare, and our UI tests were written to be purposefully sparse

src/components/data/data-utils.js
```diff
@@ -152,8 +152,10 @@ export const EntityDeleter = ({ onDismiss, onSuccess, namespace, name, selectedE
   const [additionalDeletions, setAdditionalDeletions] = useState([])
   const [deleting, setDeleting] = useState(false)
 
+  const selectedKeys = _.keys(selectedEntities)
+
   const doDelete = async () => {
-    const entitiesToDelete = _.concat(_.map(entityName => ({ entityName, entityType: selectedDataType }), selectedEntities), additionalDeletions)
+    const entitiesToDelete = _.concat(_.map(entity => ({ entityName: entity.name, entityType: entity.entityType }), selectedEntities), additionalDeletions)
```
zarsky-broad: This might be a clearer way to do this, leading people through the operations in order and making the map on the entity objects a bit clearer: 

```suggestion
    const entitiesToDelete = _.flow(
      _.map(({ name: entityName, entityType }) => ({ entityName, entityType })),
      entities => _.concat(entities, additionalDeletions)
    )(selectedEntities)
```

That third line with the arrow function can just be `_.concat(additionalDeletions)` if the order of deletions doesn't matter.

zarsky-broad (approved):
One suggestion, then good to go.

PR #2511, opened 2021-05-25
[CA-1359] A11y: Skip Navigation Links

andrewmarcus:
This functionality was extracted from #2501, where it was already reviewed and approved. I'm splitting that PR up into smaller components to try to figure out what's causing the integration tests to fail.

### Skip Navigation Link

Most websites provide a link, first in the tab order on the page, allowing screen reader users to skip repetitive content at the top of the page. This is a standard design practice which users expect to be present. In this case there isn't a ton of content in the header, but it still seems useful to be able to skip it.

Because of the way React passes refs down to child components in order to be able to target elements for focusing, I decided it was important to encapsulate all of the skip functionality in as few places as possible, to avoid having to update every page to accept an addition forwarded ref. I was able to set it up entirely within the TopBar, so the skip link is first, and the skip target is last, and appears visually as focus on the entire page. The next tab stop is the first element below the header.

The one potential problem with this approach is that the header isn't entirely identical on every page. The `<h1>` is in it on most pages except the landing page, and it sometimes also contains contextual info such as breadcrumbs, a search bar or info about the billing project. Consequently a user could skip things that are relevant to the page without knowing it. However, these elements still appear visually and semantically within the header, so it seems reasonable. It's a pattern users of assistive technologies expect, and on some pages the header is quite extensive.


s-rubenstein (approved):
This seems reasonable to me.

gpolumbo-broad (approved):

src/components/skipNavLink.js
```diff
@@ -0,0 +1,42 @@
...
+
+export const SkipNavLink = Utils.forwardRefWithName('SkipNavLink', (props, ref) => {
+  return h(Link, {
+    as: 'a',
+    href: '#',
+    className: 'reveal-on-focus',
+    style: {
+      position: 'absolute',
+      top: 0,
+      left: 0,
```
dmohs: Nit: One line?

src/components/skipNavLink.js
```diff
@@ -0,0 +1,42 @@
...
+      background: 'white',
+      color: colors.accent(),
+      border: '1px solid black',
+      borderColor: colors.accent(),
```
dmohs: Just curious why you did this instead of:
```suggestion
      border: '1px solid',
      borderColor: colors.accent(),
```

src/components/skipNavLink.js
```diff
@@ -0,0 +1,42 @@
...
+      padding: '1em',
```
dmohs: Stick with `rem` here? We try to discourage use of `em` in the codebase since people don't always expect it to be different.

src/components/skipNavLink.js
```diff
@@ -0,0 +1,42 @@
...
+      boxShadow: 'rgba(0, 0, 0, 0.5) 0px 0px 4px 0px',
+      zIndex: 9998
+    },
+    onClick: event => {
+      event.preventDefault()
+      event.stopPropagation()
+      ref.current?.focus()
+    },
+    ...props
```
dmohs: `_.merge` here to allow merging of styles?

src/components/skipNavLink.js
```diff
@@ -0,0 +1,42 @@
...
+      position: 'absolute',
+      top: 0,
+      left: 0,
+      background: 'white',
+      color: colors.accent(),
+      border: '1px solid black',
+      borderColor: colors.accent(),
+      padding: '1em',
+      boxShadow: 'rgba(0, 0, 0, 0.5) 0px 0px 4px 0px',
+      zIndex: 9998
```
dmohs: We trust your judgment that this is the best way to do this, but would you also please add a line comment?

src/style.css
```diff
@@ -151,3 +151,22 @@ a {
...
+  opacity: 1;
+}
+
+.skip-navigation-content {
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  z-index: 9999;
```
dmohs: Reciprocal line comment here?

dmohs (approved):

src/components/skipNavLink.js
andrewmarcus: Habit, I think. ;-)  Thanks for pointing it out

src/components/skipNavLink.js
andrewmarcus: Sure, make sense. I mostly just need to ensure that the button and outline both appear on top of anything else defined after them in the DOM

src/components/skipNavLink.js
andrewmarcus: I added `_.merge` to the `SkipNavLink` component. For some reason it wasn't working properly in the `SkipNavTarget` component, probably because of the `ref`, so I left that one as-is.

PR #2512, opened 2021-05-25
[CA-1359] A11y: Accessible Tab Bars with keyboard navigation

andrewmarcus:

In this PR I've re-implemented the keyboard navigation using native DOM elements rather than React, in response to PR comments from @zarsky-broad. This seems to be working well, and it lets me convert the helper functions into actual React components without opaque children. Thanks @zarsky-broad !

From the original PR:

### Tab Search Filters

The SimpleTabBar component is used on the Workspaces List page to render search filters for the workspaces table. Semantically, these are tabs: only one may be active at a time, and they cause only a portion of the page to redrawn when selected. I updated the semantic markup to make them function properly as tabs according to the ARIA specs.

In particular, I created a new reusable utility allowing a collection of Interactive elements to accept focus navigation using the arrow keys as well as Home/End, in either a vertical or horizontal orientation, by passing refs down into each child element. This seems to work like a charm! And it can be used in other places now too!

I also added PropTypes to the SimpleTabBar component to enforce that it has a label property for compliance with ARIA.

### Tab Navigation Bar

The TabBar component is used on each Workspace and Workflow page to render a navigation section at the top, pointing to different subpages within the workflow or workspace. Semantically, these are actually menu items: only one can be active at a time, and they cause the entire page to refresh with a different URL. It's essentially a submenu for the page, even though they visually appear to be tabs.

I previously added Tab semantic markup to them and updated the integration test. However, I concluded that they in fact encode menu items more accurately, so I essentially reverted them. They now support keyboard navigation just like SimpleTabBar, though.

I also added PropTypes to the TabBar component to enforce that it has a label property for compliance with ARIA. And I added labels to all other pages that use this component.


src/components/keyboard-nav.js
```diff
@@ -0,0 +1,97 @@
...
+export const ArrowKeyNavigation = ({ keyDownHandler, children, ...props }) => {
+  const ref = useRef()
+
+  // Figure out which child elements are focusable/clickable
+  const focusableElements = () => {
+    return ref.current.querySelectorAll('a, button, input, select, textarea, [tabindex]')
+  }
+
+  // Figure out which focusable element sent the event
+  const findIndex = event => _.findIndex(f => f === event.target, focusableElements())
```
zarsky-broad, line 15: I _think_ you can just use the shorthand for matching:
```suggestion
  const findIndex = event => _.findIndex(event.target, focusableElements())
```

andrewmarcus: Unfortunately that doesn't seem to work for some reason.  I just tried it again and it always returns `0` (but not `-1`, odd...)

Maybe lodash does something weird because it's an object. All I need here is referential equality.

s-rubenstein (approved):
This all looks reasonable. I did not run locally. As discussed offline, ignore v1 workspace integration test failures, those are a work in progress atm. Other test failures speak of likely real issues.

andrewmarcus:
What's going on with integration tests on this branch? I'm getting a different error every time I run it :(

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,97 @@
+import _ from 'lodash/fp'
+import { useRef } from 'react'
+import { div, h } from 'react-hyperscript-helpers'
+
+
+export const ArrowKeyNavigation = ({ keyDownHandler, children, ...props }) => {
```
dmohs, line 6: I love this utility!

dmohs (approved):

andy7i (approved):
since this code was looked at by others (including in another pr) I mostly focused on running it and testing it manually on the workspace and on the new(!) billing pages. It worked really well (and includes the kebab menu icon which is great). 👍 

outside the scope of this PR question: Will the kebab menu also get keyboard navigation?
Screen Shot 2021-06-11 at 10 57 15 AM

andrewmarcus:
@andy7i Yes! #2529 adds the vertical keyboard nav component to the popup menus.

And of course it's available as a component now to add anywhere else you want it. :) It should work with any kind of focusable children, although I only tested it in these two specific configurations.

PR #2513, opened 2021-05-25
[CA-1359] A11y: Comboboxes

andrewmarcus:

From the original: 

### Combobox components

The Comboboxes produced by the react-select module are somewhat accessible. They manage focus well, and include an aria-live section to announce changes the user makes. However, they do not identify themselves properly as Combobox widgets, or implement the necessary semantic elements.

For almost all of the internal components I was able to add overrides which add proper ARIA roles and properties to the original renderers. However, I wasn't quite able to get it to work for widgets with isSearchable == false because they use a DummyInput component for which no overrides are possible.

I've been thinking about reaching out to the maintainers of the react-select module to see if they can bake some of these improvements into their widget out of the box, but haven't done so yet.


src/components/common.js
```diff
@@ -348,7 +393,14 @@ export const GroupedSelect = ({ value, options, id, ...props }) => {
 }
 
 export const AsyncCreatableSelect = props => {
-  return h(RAsyncCreatableSelect, _.merge(commonSelectProps, props))
+  const [isOpen, setOpen] = useState(false)
```
zarsky-broad: For consistency, can you make these `setIsOpen`?

src/components/common.js
```diff
@@ -290,15 +290,51 @@ const commonSelectProps = {
...
+          'aria-selected': props.isSelected
+        })
+      }, [
         div({ style: { display: 'flex', alignItems: 'center', minHeight: 25 } }, [
           div({ style: { flex: 1, minWidth: 0, overflowWrap: 'break-word' } }, [children]),
           props.isSelected && icon('check', { size: 14, style: { flex: 'none', marginLeft: '0.5rem', color: colors.dark(0.5) } })
         ])
       ])
+    },
+    SelectContainer: props => {
```
zarsky-broad: For these, it might be tidier to destructure the argument:
```js
({ innerProps, ...props }) => {
  return RSelectComponents.SelectContainer({
    ...props
    innerProps: {
      ...innerProps,
      role: 'combobox',
      'aria-haspopup': 'listbox',
      'aria-expanded': isOpen
    }
  })
}
```

Alternatively, we could just merge once, since `_.merge` is a deep merge:

```js
return RSelectComponents.SelectContainer(_.merge(props, {
  innerProps: {
    role: 'combobox',
    'aria-haspopup': 'listbox',
    'aria-expanded': isOpen
  }
}))
```

andrewmarcus: Thanks! `_.merge()` worked once I switched everything to using the `h` wrapper and pulling out the children property. But it's quite a bit cleaner now.

s-rubenstein (approved):
LGTM. Thank you for breaking it into separate PRs. As discussed offline, merging with failing v1 tests is fine, other tests could be real failures.

marctalbott (approved):

dmohs (approved):

PR #2515, opened 2021-05-25
[CA-1359] A11y: Tables without content

andrewmarcus:

From the original:

### Tables without content

I found that, despite all of the ARIA cleanups I previously did with tables, they were still not semantically correct in the case where there was no content, because the message indicating this was just jammed in there without being wrapped properly into a row. I've fixed this and made it consistent across all of the different table components.


src/components/table.js
```diff
@@ -187,13 +187,17 @@ const ariaSort = (sort, field) => {
   return null
 }
 
+const noContent = ({ noContentMessage, noContentRenderer = _.noop, defaultMessage = 'Nothing to display' }) => {
```
zarsky-broad: Since `defaultMessage` isn't overridden anywhere, can we make it not be an argument?

andrewmarcus: The idea was that different things that use this function could potentially have different defaults. But you're right, I'm not currently using that, and don't have any plans to. I'll strip it out.

src/pages/workspaces/List.js
```diff
@@ -176,7 +176,7 @@ export const WorkspaceList = () => {
       rowCount: sortedWorkspaces.length,
       tableName: currentTab?.tableName || 'workspaces',
       noContentRenderer: () => Utils.cond(
-        [loadingWorkspaces, () => null],
+        [loadingWorkspaces, () => 'Loading...'],
```
s-rubenstein, line 179: Nit: This seems like an unrelated change. Is that necessary for this PR? (I am fine with it going in, just want to make sure it is necessary)

src/components/table.js
```diff
@@ -450,7 +484,21 @@ export const GridTable = Utils.forwardRefWithName('GridTable', ({
           containerRole: 'presentation',
           'aria-readonly': null, // Clear out ARIA properties which have been moved one level up
           'aria-label': `${tableName} content`, // The whole table is a tab stop so it needs a label
-          noContentRenderer: () => div({ style: { marginTop: '1rem', textAlign: 'center', fontStyle: 'italic' } }, [noContentMessage]),
+          noContentRenderer: () => {
+            return div({
```
s-rubenstein: This exact div formatting and styling seems oft repeated. Can we abstract that out to a component? Just noting that the changes are repeated three times, and I could see that being easy to overlook in the future.

andrewmarcus: Yep, that's a great point! Actually, now that I look at it some other parts of the table structure could probably be extracted out too, but I'll at least do this one. :)

src/pages/workspaces/List.js
andrewmarcus: The table previously displayed nothing while it was loading. But now with the new functionality, the new default value "Nothing to display" will kick in if the renderer function returns something falsey. That's not quite right in this case, so I changed it to something more accurate.

s-rubenstein (approved):
LGTM

src/components/table.js
```diff
@@ -187,13 +187,27 @@ const ariaSort = (sort, field) => {
...
+])
+
 /**
  * A virtual table with a fixed header and flexible column widths. Intended to take up the full
  * available container width, without horizontal scrolling.
  */
 export const FlexTable = ({
   initialY = 0, width, height, rowCount, variant, columns = [], hoverHighlight = false,
-  onScroll = _.noop, noContentMessage = null, headerHeight = 48, rowHeight = 48,
+  onScroll = _.noop, noContentMessage, noContentRenderer = _.noop, headerHeight = 48, rowHeight = 48,
```
jacmrob, line 210: out of curiosity, what is the difference between `_.noop` and `null`?

jacmrob (approved):

src/components/table.js
andrewmarcus: `_.noop` can be called as a function without producing errors and without needing to check for existence

BAD:
```
let noContentRenderer = null
noContentRenderer() // throws an exception
```

GOOD:
```
let noContentRenderer = _.noop
noContentRenderer() // does nothing
```

dmohs (approved):

PR #2518, opened 2021-05-27
Attempting to fix the run-notebook integration test

andrewmarcus:
● run-notebook

    TimeoutError: waiting for XPath `(//a | //*[@role="button"] | //button)[contains(normalize-space(.),"Create a") or contains(@aria-label,"Create a")]` failed: timeout 30000ms exceeded

      42 |
      43 | const click = async (page, xpath, options) => {
    > 44 |   return (await page.waitForXPath(xpath, options)).click()
         |                      ^
      45 | }
      46 |
      47 | const findText = (page, textContains, options) => {

      at new WaitTask (node_modules/puppeteer/src/common/DOMWorld.ts:780:28)
      at DOMWorld.waitForXPath (node_modules/puppeteer/src/common/DOMWorld.ts:675:22)
      at Frame.waitForXPath (node_modules/puppeteer/src/common/FrameManager.ts:1193:47)
      at Page.waitForXPath (node_modules/puppeteer/src/common/Page.ts:2111:29)
      at click (utils/integration-utils.js:44:22)
      at tests/run-notebook.js:23:9
      at utils/integration-helpers.js:48:5
      at utils/integration-helpers.js:105:5
      at utils/integration-helpers.js:152:3

failure-1622081100171-run-notebook


breilly2:
Oh no, I'm sorry you spent time looking into this. This is a somewhat known issue. This test uses a unique "user" (really a service account) for each test run. Running a notebook causes the creation of a pet service account in the workspace's project, in this case saturn-integration-test-dev. Terra doesn't clean these up, so they can accumulate over time and eventually hit quota. The process for cleaning up is documented here: https://github.com/DataBiosphere/terra-ui/wiki/Integration-Tests'-Service-Accounts

I'll do this clean-up now. Hopefully everything will return to normal. However, the last time we hit quota, it was fine for a couple of days and then the quota system got out of sync and it didn't autocorrect for another week.

zarsky-broad (approved):
I'd still be happy to see this merge, if you don't mind removing the delay. The fix to use navChild is worthwhile to me.

breilly2 (approved):
Oh, good point. Yes, I agree with Isaac. Also, I'm not certain of the need for `waitForNoSpinners`, but I can imagine a case where it might matter so I think it's worth keeping.

andrewmarcus:
Ok, the delay has been removed. Ready to merge!

PR #2520, opened 2021-06-01
Set 'deletionDelayYears' to 3 years [BW-581]

salonishah11:
To test this, I updated the limit to 1 year and tested against alpha environment (this is because I don't know of workflows that ran 3 years ago in test envs that can be used for this)

user: dumbledore.admin@test.firecloud.org
workspace: cromwell-tests-workspace
submission id: ef5b6712-58a2-4bbe-80f2-9b7fb9cdf33d
status last updated: Mar 18, 2020, 1:37 PM

Screen Shot 2021-06-01 at 11 23 22 AM

Screen Shot 2021-06-01 at 11 23 40 AM

### DIFF
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -115,7 +115,7 @@ const SubmissionDetails = _.flow(
 
   // Note: This 'deleteLimitYears' value should reflect the current 'deletion-delay' value configured for PROD in firecloud-develop's
   // 'cromwell.conf.ctmpl' file:
-  const deletionDelayYears = 6
+  const deletionDelayYears = 3
   const deletionDelayString = `${deletionDelayYears} year${deletionDelayYears > 1 ? 's' : ''}`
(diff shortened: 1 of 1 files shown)


aednichols (approved):

zarsky-broad (approved):

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -115,7 +115,7 @@ const SubmissionDetails = _.flow(
 
   // Note: This 'deleteLimitYears' value should reflect the current 'deletion-delay' value configured for PROD in firecloud-develop's
```
breilly2: Nit: This comment appears to be out-of-date with respect to the const name. Not part of the change in this PR, but probably worth fixing.

breilly2 (approved):

breilly2:
The test failure is related to reaching our service account quota on saturn-integration-test-dev. Andrew Marcus ran into this recently (#2518) and I cleaned out the service accounts. However, we've again hit the problem we had a month ago where the quota system is out of sync with reality and is incorrectly throwing quota errors. I've submitted a quota increase request, though it may take a day or two to go through. In the mean time, I believe it is okay to merge this change despite the test failure.

salonishah11:
@breilly2 thank you for taking a look at the error. Since everything else is green, I will merge this PR.

PR #2522, opened 2021-06-01
Update deps

zarsky-broad:
Read changelogs and tested locally.
Also handles open dependabot pr's.


dmohs (approved):

petesantos (approved):
LGTM. Ran through the app where the deps were touched.

PR #2523, opened 2021-06-01
[CA-1256] Add sorting and tab to billing page

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -355,7 +355,7 @@ export const PageBoxVariants = {
   LIGHT: 'light'
 }
 
-export const PageBox = ({ children, variant = undefined, style = {}, ...props }) => {
+export const PageBox = ({ children, variant, style = {}, ...props }) => {
   return div(_.merge({
     style: {
       margin: '1.5rem', padding: '1.5rem 1.5rem 0', minHeight: 125, flex: 'none', zIndex: 0,
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -24,6 +24,7 @@ import * as Nav from 'src/libs/nav'
 import { notify } from 'src/libs/notifications'
 import { authStore } from 'src/libs/state'
 import * as Style from 'src/libs/style'
+import { switchCase } from 'src/libs/utils'
 import * as Utils from 'src/libs/utils'
 
 
@@ -351,10 +352,17 @@ export const AsyncCreatableSelect = props => {
   return h(RAsyncCreatableSelect, _.merge(commonSelectProps, props))
 }
 
-export const PageBox = ({ children, style = {}, ...props }) => {
+export const PageBoxVariants = {
+  LIGHT: 'light'
+}
+
+export const PageBox = ({ children, variant = undefined, style = {}, ...props }) => {
   return div(_.merge({
     style: {
-      margin: '1.5rem', padding: '1.5rem 1.5rem 0', minHeight: 125, flex: 'none', zIndex: 0, ...style
+      margin: '1.5rem', padding: '1.5rem 1.5rem 0', minHeight: 125, flex: 'none', zIndex: 0,
+      ...Utils.switchCase(variant,
+        [PageBoxVariants.LIGHT, () => { return { backgroundColor: colors.light(), margin: 0, padding: '3rem 3rem 1.5rem' } }],
+        [Utils.DEFAULT, () => {}]), ...style
     }
   }, props), [children])
 }
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -361,7 +361,7 @@ export const PageBox = ({ children, variant, style = {}, ...props }) => {
       margin: '1.5rem', padding: '1.5rem 1.5rem 0', minHeight: 125, flex: 'none', zIndex: 0,
       ...Utils.switchCase(variant,
         [PageBoxVariants.LIGHT, () => ({ backgroundColor: colors.light(), margin: 0, padding: '3rem 3rem 1.5rem' })],
-        [Utils.DEFAULT, () => {}]), ...style
+        [Utils.DEFAULT, () => ({})]), ...style
     }
   }, props), [children])
 }
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -24,7 +24,6 @@ import * as Nav from 'src/libs/nav'
 import { notify } from 'src/libs/notifications'
 import { authStore } from 'src/libs/state'
 import * as Style from 'src/libs/style'
-import { switchCase } from 'src/libs/utils'
 import * as Utils from 'src/libs/utils'
 
 
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -360,7 +360,7 @@ export const PageBox = ({ children, variant, style = {}, ...props }) => {
     style: {
       margin: '1.5rem', padding: '1.5rem 1.5rem 0', minHeight: 125, flex: 'none', zIndex: 0,
       ...Utils.switchCase(variant,
-        [PageBoxVariants.LIGHT, () => { return { backgroundColor: colors.light(), margin: 0, padding: '3rem 3rem 1.5rem' } }],
+        [PageBoxVariants.LIGHT, () => ({ backgroundColor: colors.light(), margin: 0, padding: '3rem 3rem 1.5rem' })],
         [Utils.DEFAULT, () => {}]), ...style
     }
   }, props), [children])
--- a/src/components/group-common.js
+++ b/src/components/group-common.js
@@ -59,6 +59,8 @@ const UserMenuContent = ({ onEdit, onDelete }) => {
   ])
 }
 
+const menuCardSize = 20
+
 export const MemberCardHeaders = Utils.memoWithName('MemberCardHeaders', ({ sort, onSort }) => {
   const makeHeaderRenderer = name => h(MiniSortable, { sort, field: name, onSort }, [
     div({ style: { fontWeight: 600 } }, [Utils.normalizeLabel(name)])
@@ -72,14 +74,13 @@ export const MemberCardHeaders = Utils.memoWithName('MemberCardHeaders', ({ sort
       makeHeaderRenderer('roles')
     ]),
     // Width is the same as the menu icon.
-    div({ style: { width: 20 } })
+    div({ style: { width: menuCardSize } })
   ])
 })
 
 export const MemberCard = Utils.memoWithName('MemberCard', ({ member: { email, roles }, adminCanEdit, onEdit, onDelete, adminLabel, userLabel }) => {
   const canEdit = adminCanEdit || !_.includes(adminLabel, roles)
   const tooltip = !canEdit && `This user is the only ${adminLabel}`
-  const menuSize = 20
 
   return div({
     style: Style.cardList.longCardShadowless
@@ -87,7 +88,7 @@ export const MemberCard = Utils.memoWithName('MemberCard', ({ member: { email, r
     div({ style: { flex: '1', whiteSpace: 'nowrap', textOverflow: 'ellipsis', overflow: 'hidden', height: '1rem' } }, [email]),
     div({ style: { flex: '1', textTransform: 'capitalize', height: '1rem' } }, [_.includes(adminLabel, roles) ? adminLabel : userLabel]),
     div({ style: { flex: 'none' } }, [
-      h(TooltipTrigger, { content: tooltip, style: { height: menuSize, width: menuSize } }, [
+      h(TooltipTrigger, { content: tooltip, style: { height: menuCardSize, width: menuCardSize } }, [
         // This div exists because popup trigger overrides the tooltip trigger if it is a direct descendant, which would make there be no tooltip
         div([
           h(PopupTrigger, {
@@ -95,7 +96,7 @@ export const MemberCard = Utils.memoWithName('MemberCard', ({ member: { email, r
             closeOnClick: true,
             content: h(UserMenuContent, { onEdit, onDelete })
           }, [
-            h(Link, { 'aria-label': `Menu for User: ${email}`, disabled: !canEdit }, [icon('cardMenuIcon', { size: menuSize })])
+            h(Link, { 'aria-label': `Menu for User: ${email}`, disabled: !canEdit }, [icon('cardMenuIcon', { size: menuCardSize })])
           ])
         ])
       ])
--- a/src/components/group-common.js
+++ b/src/components/group-common.js
@@ -1,11 +1,11 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
 import { b, div, h, label } from 'react-hyperscript-helpers'
-import { ButtonPrimary, Clickable, IdContainer, LabeledCheckbox, Link, spinnerOverlay } from 'src/components/common'
+import { ButtonPrimary, IdContainer, LabeledCheckbox, Link, makeMenuIcon, MenuButton, spinnerOverlay } from 'src/components/common'
 import { icon } from 'src/components/icons'
 import { AutocompleteTextInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
-import { InfoBox } from 'src/components/PopupTrigger'
+import PopupTrigger, { InfoBox } from 'src/components/PopupTrigger'
 import TooltipTrigger from 'src/components/TooltipTrigger'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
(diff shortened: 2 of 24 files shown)


src/components/group-common.js
```diff
@@ -38,37 +38,48 @@ export const AdminNotifierCheckbox = ({ checked, onChange }) => {
...
 }
 
 export const MemberCard = Utils.memoWithName('MemberCard', ({ member: { email, roles }, adminCanEdit, onEdit, onDelete, adminLabel, userLabel }) => {
   const canEdit = adminCanEdit || !_.includes(adminLabel, roles)
   const tooltip = !canEdit && `This user is the only ${adminLabel}`
+  const menuSize = 20
 
   return div({
-    style: Style.cardList.longCard
+    style: { ...Style.cardList.longCard, boxShadow: 'none' }
```
s-rubenstein: I wanted to explicitly have `boxShadow: 'none'` everywhere, instead of abstracting, because even if it is in every place that longCard is used, it breaks our general card patterns, and I didn't want someone to use `longCard` expecting standard card patterns and get otherwise.

src/pages/billing/Project.js
```diff
@@ -6,6 +6,7 @@ import { ButtonPrimary, IdContainer, Link, Select, SimpleTabBar, spinnerOverlay
 import { DeleteUserModal, EditUserModal, MemberCard, NewUserCard, NewUserModal } from 'src/components/group-common'
 import { icon, spinner } from 'src/components/icons'
 import Modal from 'src/components/Modal'
+import { MiniSortable } from 'src/components/table'
```
s-rubenstein: I'm almost inclined to pull this out into its own file, since nothing stops us from sorting non-table things with this, but wanted to open for discussion.

src/pages/groups/List.js
```diff
@@ -196,23 +195,23 @@ const GroupList = () => {
         value: filter
       })
     ]),
-    h(PageBox, { role: 'main', style: { flexGrow: 1 } }, [
+    h(PageBox, { role: 'main', style: { flexGrow: 1, backgroundColor: colors.light(), margin: 0, padding: '3rem 3rem 1.5rem' } }, [
```
s-rubenstein: This style may be abstract-able. Maybe add `darkMode` as a prop to `PageBox`?

s-rubenstein (commented):
Screen Shot 2021-06-01 at 4 41 58 PM
Screen Shot 2021-06-01 at 4 42 09 PM
Screen Shot 2021-06-01 at 4 42 14 PM

src/components/group-common.js
marctalbott: Could we have a different card name or something? Like `longCardShadowless`? It seems like in the interest of DRY, we should abstract it if we have it in every place that we use `longCard`, but I also agree with what you're saying about having `longCard` conform with our general card patterns

marctalbott (commented):
Could we add column headers to the group list and group member list? Each column is pretty self explanatory, but we do have column headers elsewhere so it could be nice for consistency.

The user email is getting partially cut off on the bottom in the group member list (see the bottom of the '@' and the 'g' in the screenshot)
Screen Shot 2021-06-01 at 5 35 29 PM

dmohs (approved):
All of these patterns look solid to me and I love the discussion.

src/components/group-common.js
s-rubenstein: Yeah, I can do that.

s-rubenstein:
@marctalbott I have added headers everywhere!
Screen Shot 2021-06-02 at 9 57 12 AM
Screen Shot 2021-06-02 at 9 57 20 AM

marctalbott (approved):

src/components/group-common.js
```diff
@@ -38,37 +39,65 @@ export const AdminNotifierCheckbox = ({ checked, onChange }) => {
...
+  const makeHeaderRenderer = name => h(MiniSortable, { sort, field: name, onSort }, [
+    div({ style: { fontWeight: 600 } }, [Utils.normalizeLabel(name)])
+  ])
+
+  return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem' } }, [
+    div({ style: { flex: 1 } }, [
+      makeHeaderRenderer('email')
+    ]),
+    div({ style: { flex: 1 } }, [
+      makeHeaderRenderer('roles')
```
andy7i, line 74: nit:
```suggestion
      makeHeaderRenderer('role')
```

src/pages/groups/List.js
```diff
@@ -94,10 +95,26 @@ const DeleteGroupModal = ({ groupName, onDismiss, onSubmit }) => {
...
+const GroupCardHeaders = Utils.memoWithName('GroupCardHeaders', ({ sort, onSort }) => {
+  const makeHeaderRenderer = name => h(MiniSortable, { sort, field: name, onSort }, [
+    div({ style: { fontWeight: 600 } }, [Utils.normalizeLabel(name)])
+  ])
+  return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem' } }, [
+    div({ style: { width: '30%', marginRight: '1rem' } }, [
+      makeHeaderRenderer('groupName')
+    ]),
+    div({ style: { flexGrow: 1 } }, [
+      makeHeaderRenderer('groupEmail')
```
andy7i, line 109: should we add a header for Role as well?

src/pages/groups/List.js
```diff
@@ -196,29 +213,32 @@ const GroupList = () => {
...
-            return div({ style: { flexGrow: 1 } }, [
-              _.map(group => {
-                return h(GroupCard, {
-                  group, key: `${group.groupName}`,
-                  onDelete: () => setDeletingGroup(group)
-                })
-              }, filteredGroups)
+            return h(Fragment, [
+              h(GroupCardHeaders, { sort, onSort: setSort }),
+              div({ style: { flexGrow: 1, marginTop: '1rem' } }, [
```
andy7i, line 242: Nit: The column with roles is pushed right against the trash icon. It looks a bit odd to me since we don't do that in the other screenshots. 
A11Y question: should we have a header for the column with a trash can? maybe we could ask AM?

src/components/group-common.js
```diff
@@ -38,37 +39,65 @@ export const AdminNotifierCheckbox = ({ checked, onChange }) => {
...
-    icon('plus-circle', { style: { marginTop: '0.5rem' }, size: 21 })
+    icon('plus', { size: 14 }),
+    div({ style: { marginLeft: '0.5rem' } }, ['Add User'])
   ])
 }
 
+const UserMenuContent = ({ onEdit, onDelete }) => {
+  return h(Fragment, [
+    h(MenuButton, {
+      onClick: () => onEdit()
```
zarsky-broad: These shouldn't be wrapped in functions:
```js
{ onClick: onEdit }
```

src/pages/groups/List.js
zarsky-broad: `Link` takes a `variant` prop that can be `'light'`, maybe that same pattern here? `variant: 'light'`?

src/pages/billing/Project.js
```diff
@@ -53,20 +54,29 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
       h(NewUserCard, {
         onClick: () => setAddingUser(true)
-      }),
-      div({ style: { flexGrow: 1 } },
+      }, [
+        icon('plus-circle', { size: 14 }),
+        div({ style: { marginLeft: '0.5rem' } }, ['Add User'])
+      ]),
+      h(MemberCardHeaders, { sort, onSort: setSort }),
+      div(
```
zarsky-broad: When reasonable, I don't like to have elements with things in them declared on their own line; can you bring the `_.map` up?

src/pages/billing/Project.js
```diff
@@ -53,20 +54,29 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
   const [selectedBilling, setSelectedBilling] = useState()
   const [tab, setTab] = useState(query.tab || 'workspaces')
 
+  const [sort, setSort] = useState({ field: 'email', direction: 'asc' })
+
+
```
zarsky-broad: Generally we only have two blank lines after deps, between labelled sections inside a Component, or between components in a long file; could you just glance through your whitespace changes?

src/components/group-common.js
s-rubenstein: We appear to have a lack of consistency in our codebase on this, I was following the pattern here: https://github.com/DataBiosphere/terra-ui/blob/dev/src/pages/workspaces/List.js#L66 which seems like the most equivalent thing.

Should I do a comprehensive pass of the codebase and change in all places?

src/pages/billing/Project.js
s-rubenstein: I personally find the indentation of this makes it more readable, is this a codebase pattern? If so, I can adapt it to that, but want to make sure its a codebase pattern first.

src/components/group-common.js
zarsky-broad: That would be nifty. Some context:

Back in the days of class components, it was necessary to wrap callbacks in arrow functions, in order to ensure that the value of `this` was preserved. In the arrow function world, we don't use `this` (indeed, there is no `this` in an arrow function).

src/pages/billing/Project.js
s-rubenstein: If we don't have a codebase pattern there, maybe it makes sense to just catch up real quick about that first before I make the change?

zarsky-broad: It is a pattern in the codebase, yes. For example, this `div` previously had its props object on the same line as the function call; now, with no props, I'd expect the children on the same line.

src/components/group-common.js
s-rubenstein: Ah, that makes sense. I was wondering how we got around the this binding

src/components/group-common.js
```diff
@@ -38,37 +39,65 @@ export const AdminNotifierCheckbox = ({ checked, onChange }) => {
...
+
+  return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem' } }, [
+    div({ style: { flex: 1 } }, [
+      makeHeaderRenderer('email')
+    ]),
+    div({ style: { flex: 1 } }, [
+      makeHeaderRenderer('roles')
+    ]),
+    // Width is the same as the menu icon.
+    div({ style: { width: 20 } })
```
andy7i: could we codify this by setting `menuSize` at a higher level? (and similarly creating another const in List.js?)

andy7i (commented):
a few comments and questions but otherwise looks good!

src/components/group-common.js
s-rubenstein: @zarsky-broad done, ptal. I did not do comprehensive testing, instead relying on tests and review here. If you think it would be worth going through and testing all of the refactors, I can do that.

FWIW, in the future, for this sort of refactor (especially given how long it took), it would be extremely helpful to be given context. Why did we decide to move towards direct bindings, rather than arrow functions? What benefits do we gain? What are the cons of this approach? I think that generally providing more of that information will help people understand the decisions behind the Terra UI codebase, and the code style recommendations.

src/pages/billing/Project.js
s-rubenstein: Done

src/components/group-common.js
s-rubenstein: This has to be `roles`. Due to the nature of it, the name passed in has to match the field name, and presently we pass back an array of roles, so the name has to be `roles`. We could change this if we wanted to, but IMO it is unneeded complexity.

src/pages/groups/List.js
s-rubenstein: Sure, will do!

src/pages/groups/List.js
s-rubenstein: @andrewmarcus any thoughts here?

s-rubenstein: Re: the column with the roles being pushed right against the trash can… I am inclined to punt on that. This is an aside to the ticket, and is the existing behavior on the group page. I was mainly doing the header and styling on this page to ensure it stays in alignment, future changes/extra changes should probably have their own ticket. Any objections?

s-rubenstein: Actually, the trashcan is in the same column as the role, so the A11Y question is actually handled now, so I am going to say that section of that is resolved.

andy7i (approved):
👍

src/pages/groups/List.js
andrewmarcus: Hey there, thanks for asking. You definitely need a column heading on every column, although you can use the `sr-only` class to make it only available for screen readers and not visible if that makes sense. I haven't looked through the code for this PR, but in the second pair of screenshots above, you'll probably need a header over the column that has the values "Admin" and "Member".

I'm not sure about the extra menu toggle and delete button at the end of the row. Probably fine to leave them where they are rather than breaking them into their own column. I guess the only reason to give them a separate column would be that a screen reader users navigating the table will hear something like "Admin Delete, cell, row 3 column 3"

s-rubenstein: That screen reader text there is compelling to me. Will break them out!

src/components/common.js
```diff
@@ -351,10 +351,17 @@ export const AsyncCreatableSelect = props => {
   return h(RAsyncCreatableSelect, _.merge(commonSelectProps, props))
 }
 
-export const PageBox = ({ children, style = {}, ...props }) => {
+export const PageBoxVariants = {
+  LIGHT: 'light'
+}
+
+export const PageBox = ({ children, variant = undefined, style = {}, ...props }) => {
```
zarsky-broad: `undefined` is the default value for anything, so defining it as the default here is a no-op.
```suggestion
export const PageBox = ({ children, variant, style = {}, ...props }) => {
```

src/components/common.js
```diff
@@ -351,10 +351,17 @@ export const AsyncCreatableSelect = props => {
...
   return div(_.merge({
     style: {
-      margin: '1.5rem', padding: '1.5rem 1.5rem 0', minHeight: 125, flex: 'none', zIndex: 0, ...style
+      margin: '1.5rem', padding: '1.5rem 1.5rem 0', minHeight: 125, flex: 'none', zIndex: 0,
+      ...Utils.switchCase(variant,
+        [PageBoxVariants.LIGHT, () => { return { backgroundColor: colors.light(), margin: 0, padding: '3rem 3rem 1.5rem' } }],
```
zarsky-broad:
```suggestion
        [PageBoxVariants.LIGHT, () => ({ backgroundColor: colors.light(), margin: 0, padding: '3rem 3rem 1.5rem' })],
```

src/components/common.js
```diff
@@ -351,10 +351,17 @@ export const AsyncCreatableSelect = props => {
...
+        [Utils.DEFAULT, () => {}]), ...style
```
zarsky-broad: This currently actually returns `undefined`, instead of a literal object.
```suggestion
        [Utils.DEFAULT, () => ({})]), ...style
```

src/components/group-common.js
```diff
@@ -38,37 +39,68 @@ export const AdminNotifierCheckbox = ({ checked, onChange }) => {
...
-        }, ['Remove'])
+      h(TooltipTrigger, { content: tooltip, style: { height: menuCardSize, width: menuCardSize } }, [
+        // This div exists because popup trigger overrides the tooltip trigger if it is a direct descendant, which would make there be no tooltip
+        div([
+          h(PopupTrigger, {
+            side: 'left',
+            closeOnClick: true,
+            content: h(UserMenuContent, { onEdit, onDelete })
+          }, [
+            h(Link, { 'aria-label': `Menu for User: ${email}`, disabled: !canEdit }, [icon('cardMenuIcon', { size: menuCardSize })])
```
zarsky-broad: I'm a bit confused why we don't just use `Link`'s native tooltip prop? (Also why we didn't before, but maybe that's the easiest fix now?)

src/pages/billing/Project.js
```diff
@@ -26,8 +26,9 @@ const styles = {
   }
 }
 
+
```
zarsky-broad:
```suggestion
```

src/pages/billing/Project.js
```diff
@@ -53,30 +54,37 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
   const [selectedBilling, setSelectedBilling] = useState()
   const [tab, setTab] = useState(query.tab || 'workspaces')
 
+  const [sort, setSort] = useState({ field: 'email', direction: 'asc' })
```
zarsky-broad, line 55: For consistency with other places in the codebase, I'd tell you to get rid of the newline above this, so it's grouped with the other `useState` hooks.

zarsky-broad (approved):
Didn't test, purely reviewing code; basically looks good, just a few comments, then 👍.

src/components/group-common.js
s-rubenstein: I have no clue why we didn't before! I can check if that would work!

PR #2525, opened 2021-06-02
[CA-1382] A11y: Tooltip Labels

andrewmarcus:
To support accessibility, all icons without any accompanying text need an `aria-label`. Every such icon I could find within the application already includes a `tooltip` (great!) or an `aria-label`, and a previous PR attaches the text of every tooltip to its element as an `aria-describedby`. However, in the case of icons these need to be `aria-labelledby` instead; otherwise there’s still no label even if there’s a description.

This PR introduces a new, optional, `useTooltipAsLabel` parameter, which if set to true forces the tooltip to be used as a label rather than as a description (the default). It also introduces a convenient mechanism to do this automatically on any `Clickable` or `TooltipTrigger` element if it detects that the element has only a single `icon` without an `aria-label` as a child.

This PR also adds support for setting an explicit `aria-label` property on an `icon` element, which, if set, removes the `aria-hidden="true"` which would normally be applied. I believe this property is already being used in a few places but I didn't do an exhaustive search to see where.

Finally, I've reviewed the codebase and added the `useTooltipAsLabel` property where necessary, in situations which have a tooltip already but where the auto-detection might not work and an `aria-label` is still necessary. Many of these cases already had an explicit `aria-label` which duplicated the `tooltip` and which I've been able to remove because of the auto-detection. In some case I replaced the `aria-label` with a `tooltip` instead, since it seemed relevant for all users.


integration-tests/utils/integration-utils.js
```diff
@@ -34,9 +34,9 @@ const findInGrid = (page, textContains, options) => {
 const clickable = ({ text, textContains }) => {
   const base = '(//a | //*[@role="button"] | //button)'
   if (text) {
-    return `${base}[normalize-space(.)="${text}" or @aria-label="${text}"]`
+    return `${base}[normalize-space(.)="${text}" or @aria-label="${text}" or @aria-labelledby=//*[normalize-space(.)="${text}"]/@id]`
   } else if (textContains) {
-    return `${base}[contains(normalize-space(.),"${textContains}") or contains(@aria-label,"${textContains}")]`
+    return `${base}[contains(normalize-space(.),"${textContains}") or contains(@aria-label,"${textContains}") or @aria-labelledby=//*[contains(normalize-space(.),"${textContains}")]/@id]`
```
andrewmarcus, line 39: This does not seem to be necessary to get the tests to pass currently. However, it allows the `tooltip` property to be used as a selection criteria for clickables in the future in cases where there isn't a duplicative `aria-label` since it's no longer necessary.

marctalbott (approved):

src/components/common.js
```diff
@@ -62,8 +62,15 @@ export const Clickable = ({ href, as = (!!href ? 'a' : 'div'), disabled, tooltip
     ...props
   }, [children])
 
+  // If there's only one child which is an icon, and no aria-label, and the user hasn't explicitly set useTooltipAsLabel to false,
+  // then use the tooltip as the label rather than the description.
```
gpolumbo-broad: I keep rereading this trying to get the logic straight in my head and wondering if there's a clearer way to express this (even if it might be wordier).  Maybe:

`If "aria-label" on the Clickable is unset and there is only one child which is an "icon", then the "aria-label" will be automatically set to be the same as the "icon"'s tooltip.  If the user explicitly sets "useTooltipAsLabel" to false, then the icon's description will be used as Clickable's "aria-label"`

As I read this, I'm wondering also if `useTooltipAsLabel` would perhaps be more descriptive/helpful if it were `useDescriptionAsAriaLabel` instead?

gpolumbo-broad (approved):

src/components/common.js
andrewmarcus: Technically, it's using the tooltip as either a label or a description, but not both. But it's not taking an existing description and turning it into a label...

You're absolutely right that this is confusing language! I'll clean it up.

andrewmarcus: The wording has been updated! Now includes a better description of why this check is even necessary, and frames the logic in a positive way, rather than all the negation I had before

src/components/common.js
```diff
@@ -62,8 +62,22 @@ export const Clickable = ({ href, as = (!!href ? 'a' : 'div'), disabled, tooltip
...
   }, [children])
 
+  // To support accessibility, every link must have a label or contain text or a labeled child.
+  // If an unlabeled link contains just a single unlabeled icon, then we should use the tooltip as the label,
+  // rather than as the description as we otherwise would.
+  //
+  // If the auto-detection can't make the proper determination, for example, because the icon is wrapped in other elements,
+  // you can explicitly pass in a boolean as `useTooltipAsLabel` to force the correct behavior.
+  //
+  // Note that TooltipTrigger does this same check with its own children, but since we'll be passing it an
```
dmohs, line 72: Appreciate that you called this out.

dmohs (approved):
I'm grateful to you for this boost in the quality of the codebase. Really nice duplication reduction.

PR #2526, opened 2021-06-03
[CA-1307] A11y: Workspace pages

andrewmarcus:
This PR attempts to address accessibility issues on all of the pages related to an individual workspace, enumerated here: https://broadworkbench.atlassian.net/browse/CA-1307

To get these pages fully compliant, this PR depends on the updates to Tooltips from #2525, which was split out from this branch and should be merged first.

Although this PR contains a large number of changes, the main theme is better semantics. Links which open popup modals now use `aria-haspopup="dialog"`, collapsible sections now properly link to the region which they control, and Autocomplete widgets now link properly to a label.

In addition, I made significant semantic improvements to the navigation section of the Data page on workspaces. The sidebar is now a nested `role="list"` within a `role="navigation`" region, with `aria-current` used to indicate which table is active, and it no longer contains buttons within buttons.


src/components/common.js
```diff
@@ -52,7 +52,7 @@ const styles = {
   }
 }
 
-export const Clickable = ({ href, as = (!!href ? 'a' : 'div'), disabled, tooltip, tooltipSide, tooltipDelay, onClick, children, ...props }) => {
+export const Clickable = ({ href, as = (!!href ? 'a' : 'div'), disabled, tooltip, tooltipSide, tooltipDelay, useTooltipAsLabel, onClick, children, ...props }) => {
```
andrewmarcus: I snuck this in here so I could start to use the attribute without waiting for #2525 to get merged and without causing DOM warnings. May possible cause a merge conflict once that branch is merged, but it'll be easy to resolve.

src/components/common.js
```diff
@@ -546,7 +546,7 @@ export const ClipboardButton = ({ text, onClick, ...props }) => {
   const [copied, setCopied] = useState(false)
   return h(Link, {
     ...props,
-    tooltip: 'Copy to clipboard',
+    tooltip: copied ? 'Copied to clipboard' : 'Copy to clipboard',
```
andrewmarcus: #2525 is necessary in order to resolve the original issue identified by aXe where this button has no `aria-label`.

src/components/data/EntitiesContent.js
```diff
@@ -345,9 +345,11 @@ const EntitiesContent = ({
           onClick: () => setDeletingEntities(true)
         }, ['Delete Data'])
       ])
-    }, [h(Link, { style: { marginRight: '1rem' } }, [
-      icon('ellipsis-v-circle', { size: 24 })
-    ])])
+    }, [
+      h(Link, { style: { marginRight: '1rem' } }, [
+        icon('ellipsis-v-circle', { size: 24, 'aria-label': 'selection menu' })
```
andrewmarcus: #2525 will set `aria-hidden` to false automatically if `aria-label` is present. No need to add that here and then remove it once branch is merged...

src/components/data/EntitiesContent.js
```diff
@@ -373,7 +375,11 @@ const EntitiesContent = ({
           !snapshotName && renderDownloadButton(columnSettings),
           !_.endsWith('_set', entityKey) && renderCopyButton(entities, columnSettings),
           div({ style: { margin: '0 1.5rem', height: '100%', borderLeft: Style.standardLine } }),
-          div({ style: { marginRight: '0.5rem' } }, [`${selectedLength} row${selectedLength === 1 ? '' : 's'} selected`]),
+          div({
+            style: { marginRight: '0.5rem' },
+            role: 'status',
+            'aria-atomic': true
+          }, [`${selectedLength} row${selectedLength === 1 ? '' : 's'} selected`]),
```
andrewmarcus, line 381: This announces the current selection status to screen readers whenever it changes. `role="status"` uses `aria-live="polite"` by default.

src/style.css
```diff
@@ -151,3 +151,10 @@ a {
 .token.variable {
   color: #b56000 !important; /* #e90; */
 }
+
+.token.comment,
+.token.prolog,
+.token.doctype,
+.token.cdata {
+  color: #687888 !important; /* slategray */
+}
```
andrewmarcus, line 165: I missed these colors in the first update to the Prism colors. This time I found a workflow that had tokens from this group, and ANDI flagged the contrast again.

src/components/data/LocalVariablesContent.js
```diff
@@ -170,6 +170,7 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { namespace,
               editIndex === rowIndex ?
                 h(Fragment, [
                   h(Select, {
+                    'aria-label': 'what type of value is this?',
```
marctalbott: Is this the right aria label? Completely fine if the answer is yes, but it seems like they're generally shorter and I'd expect it to be more along the lines of "Edit Type"

src/components/data/EntitiesContent.js
```diff
@@ -345,9 +345,11 @@ const EntitiesContent = ({
           onClick: () => setDeletingEntities(true)
         }, ['Delete Data'])
       ])
-    }, [h(Link, { style: { marginRight: '1rem' } }, [
-      icon('ellipsis-v-circle', { size: 24 })
-    ])])
+    }, [
+      h(Link, { style: { marginRight: '1rem' } }, [
```
s-rubenstein: Was the whitespace change here intentional? (newline)

src/components/table.js
```diff
@@ -683,9 +688,10 @@ export const ColumnSelector = ({ onSave, columnSettings }) => {
 
   return h(Fragment, [
     h(Clickable, {
-      'aria-label': 'Select columns',
```
s-rubenstein: Out of curiosity, how come we don't need the `aria-label`?

src/pages/workspaces/workspace/Data.js
```diff
@@ -48,22 +49,34 @@ const styles = {
...
   }, [
-    div({ style: { flex: 'none', display: 'flex', width: '1.5rem' } }, [
-      icon(iconName, { size: iconSize })
-    ]),
-    div({ style: { flex: 1, ...Style.noWrapEllipsis } }, [
-      entityName || children
+    h(Clickable, {
+      style: { flex: '1 1 auto', maxWidth: 232, ...Style.navList.item(selected), color: colors.accent(1.2), ...buttonStyle },
+      ...(isEntity ? {
+        tooltip: entityName ? `${entityName} (${entityCount} row${entityCount === 1 ? '' : 's'})` : undefined,
```
s-rubenstein, line 64: I'm surprised that undefined works here, given some of the other things about tooltip. Should this be `unnamed entity` or something? (That way we guarantee a label)

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -233,7 +258,7 @@ const JobHistory = _.flow(
               cellRenderer: ({ rowIndex }) => {
                 const { workflowStatuses, status } = filteredSubmissions[rowIndex]
                 return h(Fragment, [
-                  statusCell(workflowStatuses), _.keys(collapsedStatuses(workflowStatuses)).length === 1 && status
+                  statusCell(workflowStatuses, status)
```
s-rubenstein: It looks like there's a conditional removed here. What is the goal of that change?

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -151,22 +151,22 @@ const WorkflowDashboard = _.flow(
             ])
           ]),
           makeSection('Links', [
-            div({ style: { display: 'flex', flexFlow: 'row wrap', marginTop: '0.5rem' } }, [
+            div({ style: { display: 'flex', flexFlow: 'row wrap', marginTop: '0.5rem', lineHeight: '2rem' } }, [
               h(Link, {
                 ...Utils.newTabLinkProps,
                 href: `${getConfig().jobManagerUrlRoot}/${workflowId}`,
-                style: { display: 'flex' },
+                style: { display: 'flex', marginRight: '2em', alignItems: 'center' },
```
s-rubenstein: Nit: In general, I prefer `marginLeft` to `marginRight`, because it means that an element is more concerned with its own positioning, rather than the positioning of other elements around it. 

Also, if possible, I would recommend doing this with rem rather than em

s-rubenstein: Nit: Because there are more styles being applied to these than just flex now, I would say we should probably pull out to a variable.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -139,28 +140,34 @@ const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange,
...
+                })
+              ])]) : h(TextCell, { style: { flex: 1 } }, [value]),
               !readOnly && isFile && h(Clickable, {
                 style: { position: 'absolute', right: '0.5rem', top: 0, bottom: 0, display: 'flex', alignItems: 'center' },
                 onClick: () => onBrowse(name),
-                tooltip: 'Browse bucket files'
+                tooltip: 'Browse bucket files',
+                'aria-haspopup': 'dialog'
               }, [icon('folder-open', { size: 20 })])
+
```
s-rubenstein: Nit: RM newline

marctalbott (approved):

src/components/data/EntitiesContent.js
andrewmarcus: I had added other stuff, then removed it with the Tooltips enhancements. I'll revert the whitespace

src/components/data/LocalVariablesContent.js
andrewmarcus: Fair point. I was trying to describe that it applies a data type to other content, but maybe "data type" would be a reasonable enough label

src/pages/workspaces/workspace/Data.js
andrewmarcus: Technically, in this instance the tooltip is only necessary in situations where the `noWrapEllipsis` hides some of the content. Otherwise it's basically just duplicative of what the user already sees. 

As far as I have seen, the only cases where this widget displays its children rather than an entity name is for the Reference Data, and the words "Workspace Data" and "Files" at the end. Since these don't extend beyond 280px, a tooltip is technically unnecessary for them.

src/pages/workspaces/workspace/JobHistory.js
andrewmarcus: Ah. I moved the conditional logic into the `statusCell` function since I needed the content to be nested rather than parallel. I'll remove the `Fragment` since there's now only one child here

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
andrewmarcus: Hmm, the problem had to do with wrapping in this section on narrow windows. If one or two of the links bumped down to the next line, they were indented by 1rem compared to the first line. But looking at it now, I'm not sure these fixes solve that either.

Since it's not technically linked to accessibility, perhaps I'll just revert this part and let the Broad team figure out how you want to handle it.
Screen Shot 2021-06-04 at 11 37 01 PM

andrewmarcus: OBE: I reverted to closer to how they were

src/components/table.js
andrewmarcus: Oh. Right. I pulled this out because #2525 will make it redundant; the tooltip will be used as a label once that branch is merged. At one time those changes were in this branch...

I can restore that line for now, to keep the two PRs entirely independent. Or leave it like this assuming they'll be merged at roughly the same time.  Your choice. :)

andrewmarcus: Nevermind. The change is already in that PR. I'll revert it here

src/pages/workspaces/workspace/Workflows.js
```diff
@@ -89,6 +89,7 @@ const sortOptions = [
 
 const WorkflowCard = Utils.memoWithName('WorkflowCard', ({ listView, name, namespace, config, onExport, onCopy, onDelete, workspace }) => {
   const { namespace: workflowNamespace, name: workflowName, methodRepoMethod: { sourceRepo, methodVersion } } = config
+  const sourceRepoName = sourceRepo === 'agora' ? 'Terra' : sourceRepo
```
s-rubenstein: Nit (existing, so feel free to ignore): We should probably have consistent capitalization on this, either `Agora` instead of `agora` or `terra` instead of `Terra`

s-rubenstein (approved):
This looks reasonable to me! Thanks for the changes.

src/pages/workspaces/workspace/Workflows.js
andrewmarcus: This is just the human-readable name. Maybe use `Utils.normalizeLabel()`?

s-rubenstein: That sounds like exactly the right method to use!

src/components/data/EntitiesContent.js
```diff
@@ -374,7 +374,11 @@ const EntitiesContent = ({
           !snapshotName && renderDownloadButton(columnSettings),
           !_.endsWith('_set', entityKey) && renderCopyButton(entities, columnSettings),
           div({ style: { margin: '0 1.5rem', height: '100%', borderLeft: Style.standardLine } }),
-          div({ style: { marginRight: '0.5rem' } }, [`${selectedLength} row${selectedLength === 1 ? '' : 's'} selected`]),
+          div({
+            style: { marginRight: '0.5rem' },
+            role: 'status',
+            'aria-atomic': true
```
dmohs: I believe we are placing 'aria-...' attributes first in the list of properties for consistency.

dmohs (approved):

PR #2529, opened 2021-06-04
[CA-1385] A11y: Accessible Menu Buttons Popups

andrewmarcus:
This PR is based on the #2512 branch, which hasn't yet been merged, since I needed the keyboard navigation component.

In this PR, I create a new `MenuTrigger` component, which wraps `PopupTrigger` and properly sets up a menu in a dialog box with all the proper roles and relationships.

Some accessibility features:
* The `PopupTrigger` component now allows different roles to be used in the modal: 'dialog', 'menu', 'listbox', etc. The modal itself has the indicated role, and the `aria-haspopup` attached to the trigger button has the matching value.
* The label attached to the trigger button will now also be used as a label for the dialog box
* Users can use the up, down, home and end arrow keys to navigate the menu options

I moved `MenuButton` and associated functionality into the `PopupTrigger.js` file. In addition, I updated all applicable instances of `PopupTrigger` in the codebase to use `MenuTrigger` instead and get the new functionality.


src/components/PopupTrigger.js
```diff
@@ -51,21 +59,32 @@ const PopupTrigger = Utils.forwardRefWithName('PopupTrigger', ({ content, childr
...
+      id: menuId,
       target: childId,
       handleClickOutside: () => setOpen(false),
       outsideClickIgnoreClass: childId,
       onClick: closeOnClick ? () => setOpen(false) : undefined,
-      ...props
+      popupProps: {
+        role,
+        'aria-labelledby': labelledby,
+        ...props
```
s-rubenstein: Passing props here as opposed to a level up is technically a functional change. I would be worried about props not relating to the popup that should be passed in getting lost. However, it looks like we weren't using the general props there anyway, so it might be fine? Just worth double checking.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -21,7 +21,9 @@ import { reportError, withErrorReporting } from 'src/libs/error'
 import { versionTag } from 'src/libs/logos'
 import * as Nav from 'src/libs/nav'
 import { notify } from 'src/libs/notifications'
-import { appIsSettingUp, convertedAppStatus, currentApp, currentAttachedDataDisk, getGalaxyCost, isCurrentGalaxyDiskDetaching } from 'src/libs/runtime-utils'
+import {
```
s-rubenstein: Nit: My preference would be to either put these on the line with the curly braces, as they were before, or newline each of them/put a newline in the middle, so we are making better use of the improved horizontal space.

s-rubenstein (approved):
LGTM, a few changes but nothing blocking.

src/components/PopupTrigger.js
andrewmarcus: So the problem with just pulling in ...props in `Popup` was that it also grabs all of the internal `onClickOutside` properties for some reason. I couldn't just stick them straight into the `div`; I was getting console errors complaining about things like `stopPropagation` not being a valid DOM property. So I did it this way instead to only pass in the ones we actually care about going into the DOM, like ARIA properties.

But now that I think about it, you're right! Properties `Popup` needs, like `side`, need to be passed in one level further up since they're not explicitly pulled out in `PopupTrigger`.

I think I can either expose `popupProps` as a parameter in `PopupTrigger`, or pull out the `side`, `target` and `onClick` properties explicitly.

src/pages/workspaces/workspace/Notebooks.js
andrewmarcus: Sorry, I ran Intellij's "Organize Imports" script and it did this. I'll revert it.

src/components/PopupTrigger.js
andrewmarcus: It looks like the only unnamed prop which could possibly be passed through was `side`, so I just pulled that out here. I decided not to push them into `popupProps` inside the `PopupTrigger`, because it could change existing calls, and then it's not clear what should be in versus out. For example, I'm pulling `role` out explicitly because I'm using it elsewhere, however it also gets attached to the DOM node so it should by that logic be in popupProps instead. Better to just leave it all in props and do the right thing. :)

andrewmarcus: Nevermind! Users of the `PopupTrigger` can also pass in properties to be consumed by the `onClickOutside` functionality. I forgot about that. I couldn't find anywhere that's using that, but it should be possible.

I've moved the properties back into `popupProps` inside the `PopupTrigger`. At least now it's consistent and clear what those particular properties will be used for...

andy7i (commented):
I'll try to review this PR by today. In the meantime, could you look into this? I'm seeing some errors when I try to run this branch locally.

I ran

```
yarn install
yarn start
```
and saw:
```

Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db
```
before it errored with:

```
Failed to compile
./src/components/FooterWrapper.js
Error: [BABEL] /Users/ajang/Firecloud/terra-ui/src/components/FooterWrapper.js: Cannot find module '@babel/plugin-syntax-class-properties'
Require stack:
...
```

zarsky-broad:
@andy7i try to `rm -rf node_modules` and then install again.

src/components/keyboard-nav.js
```diff
@@ -85,9 +85,9 @@ export const VerticalNavigation = ({ children, ...props }) => h(ArrowKeyNavigati
         focusOn(i - 1); break
       case 'ArrowDown':
         focusOn(i + 1); break
-      case 'Home': case 'PageDown':
+      case 'Home': case 'PageUp':
         focusOn(0); break
-      case 'End': case 'PageUp':
+      case 'End': case 'PageDown':
```
andy7i, line 90: not sure where these buttons are on my mac keyboard -- haven't tested this

andy7i (approved):
looks great! verified new aria-haspopup on several types of roles and everything seemed good in manual testing as well

src/components/keyboard-nav.js
zarsky-broad: Fn + up/down arrow

src/components/PopupTrigger.js
```diff
@@ -87,3 +110,36 @@ export const InfoBox = ({ size, children, style, side, tooltip, iconOverride })
...
+  return div({ role: 'menuitem' }, [
+    h(Clickable, _.merge({
+      ref,
+      disabled,
+      style: {
+        display: 'flex', alignItems: 'center',
+        fontSize: 12, minWidth: 125, height: '2.25rem',
+        color: disabled ? colors.dark(0.7) : undefined,
+        padding: '0.875rem',
+        cursor: disabled ? 'not-allowed' : 'pointer'
```
dmohs:
```suggestion
        padding: '0.875rem',
        ...(disabled ? { color: colors.dark(0.7), cursor: 'not-allowed' } : { cursor: 'pointer' })
```

dmohs (approved):
With @petesantos and @zarsky-broad.

src/components/PopupTrigger.js
```diff
@@ -20,27 +22,35 @@ const styles = {
 
 // This is written as a "function" function rather than an arrow function because react-onclickoutside wants it to have a prototype
 // eslint-disable-next-line prefer-arrow-callback
-export const Popup = onClickOutside(function({ side = 'right', target: targetId, onClick, children }) {
+export const Popup = onClickOutside(function({ id, side = 'right', target: targetId, onClick, children, popupProps = {} }) {
+  // We're passing popupProps here rather than just props, because ...props also includes lots of internal onClickOutside properties which
+  // aren't valid to be dropped on a DOM element.
+  Utils.useConsoleAssert('aria-label' in popupProps || 'aria-labelledby' in popupProps, 'In order to be accessible, Popup needs a label')
```
zarsky-broad: We don't tend to use `in`, so to match other places:
```suggestion
  Utils.useConsoleAssert(popupProps['aria-label'] ||  popupProps['aria-labelledby'], 'In order to be accessible, Popup needs a label')
```

andrewmarcus: Ok, sounds good. I think I had been trying to account for someone intentionally setting the `aria-label` to `''`, but that should probably also throw a warning. I'll clean all of these useConsoleAssert statements in a new PR anyway.

PR #2530, opened 2021-06-06
[CA-1308] A11y: Library pages

andrewmarcus:
The library pages didn't have much to do to ensure accessibility. All I did was resolve the explicitly enumerated issues from https://broadworkbench.atlassian.net/browse/CA-1308 on the "Datasets" and "Code" pages.

"Code & Workflows" page:
* Increased color contrast of links in right sidebar
* Changed styles applied to cards to make the overflow regions work with aXe so it didn't detect overlapping regions (no visual impact)

"Datasets" page:
* Added `aria-label`s to every button, in order to ensure each button has a unique name. I kinda made these labels up based on the names of each dataset, but stripping out info I thought may be extraneous. But feel free to pick them apart and suggest better labels. :)

I did not make any changes on the "Showcase & Tutorials" page. aXe appears to have detected links with the same name, but since these are entirely dynamic this can't be resolved in code.

### DIFF
--- a/src/pages/library/Code.js
+++ b/src/pages/library/Code.js
@@ -36,9 +36,9 @@ export const MethodCard = ({ method: { name, synopsis }, ...props }) => {
       position: 'relative'
     }
   }, [
-    div({ style: { flex: 1, padding: '15px 20px' } }, [
+    div({ style: { flex: 'none', padding: '15px 20px', height: 140 } }, [
       div({ style: { color: colors.accent(), fontSize: 16, lineHeight: '20px', height: 40, marginBottom: 7 } }, [name]),
-      div({ style: { lineHeight: '20px', height: 100, whiteSpace: 'pre-wrap', overflow: 'hidden' } }, [synopsis])
+      div({ style: { lineHeight: '20px', ...Style.noWrapEllipsis, whiteSpace: 'pre-wrap', height: 60 } }, [synopsis])
     ]),
     wdlIcon({ style: { position: 'absolute', top: 0, right: 8 } })
   ])
@@ -61,7 +61,10 @@ export const DockstoreTile = () => {
   return div({ style: { display: 'flex' } }, [
     h(LogoTile, { logoFile: dockstoreLogo, style: { backgroundColor: 'white' } }),
     div([
-      h(Link, { href: `${getConfig().dockstoreUrlRoot}/search?_type=workflow&descriptorType=WDL&searchMode=files` }, 'Dockstore'),
+      h(Link, {
+        href: `${getConfig().dockstoreUrlRoot}/search?_type=workflow&descriptorType=WDL&searchMode=files`,
+        style: { color: colors.accent(1.1) } // For a11y, we need at least 4.5:1 contrast against the gray background
+      }, 'Dockstore'),
       div(['Browse WDL workflows in Dockstore, an open platform used by the GA4GH for sharing Docker-based workflows'])
     ])
   ])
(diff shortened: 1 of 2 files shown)


src/pages/library/Code.js
```diff
@@ -61,7 +61,10 @@ export const DockstoreTile = () => {
   return div({ style: { display: 'flex' } }, [
     h(LogoTile, { logoFile: dockstoreLogo, style: { backgroundColor: 'white' } }),
     div([
-      h(Link, { href: `${getConfig().dockstoreUrlRoot}/search?_type=workflow&descriptorType=WDL&searchMode=files` }, 'Dockstore'),
+      h(Link, {
+        href: `${getConfig().dockstoreUrlRoot}/search?_type=workflow&descriptorType=WDL&searchMode=files`,
+        style: { color: colors.accent(1.1) } // For a11y, we need at least 4.5:1 contrast against the gray background
```
s-rubenstein, line 66: How frequently is the contrast on links needing to get bumped up? If its frequent, maybe we should look at the base color.

src/pages/library/Datasets.js
```diff
@@ -248,7 +255,8 @@ const fcDataLib = () => h(Participant, {
   h(ButtonPrimary, {
     href: `${getConfig().firecloudUrlRoot}/?return=${returnParam()}#library`,
     ...Utils.newTabLinkProps,
-    tooltip: 'Search for dataset workspaces'
+    tooltip: 'Search for dataset workspaces',
+    'aria-label': 'Browse Broad Institute datasets'
```
s-rubenstein: It seems strange to me that we say "datasets" here (and below), and data elsewhere. My inclination would be to unify naming schemes to either dataset/datasets or data.

s-rubenstein (approved):
LGTM from an accessibility standpoint! One comment on naming, but it is more nitpick-y than I am inclined to block on.

src/pages/library/Datasets.js
andrewmarcus: Agree! I thought I was just preserving the names already here. But happy to change them if it makes sense. Or hand it over and let your team take care of that... :)

s-rubenstein: Let's leave it for now, and I will tag… @breilly2 is this workflows? (It seems like that is my best guess)

andy7i (approved):
👍

src/pages/library/Datasets.js
```diff
@@ -168,10 +169,12 @@ const amppd = () => h(Participant, {
 }, [
   h(ButtonPrimary, {
     style: { marginBottom: '1rem' },
-    href: Nav.getLink('data-explorer-private', { dataset: 'AMP PD Clinical - 2020_v2release_1218' })
+    href: Nav.getLink('data-explorer-private', { dataset: 'AMP PD Clinical - 2020_v2release_1218' }),
+    'aria-label': 'Browse AMP-PD Tier 1 data'
```
dmohs: To keep with existing patterns, could you please put new "aria" props above others.

dmohs (approved):
With @petesantos and @zarsky-broad.

src/pages/library/Code.js
andrewmarcus: I discussed this with @s-rubenstein in a different PR. In this particular case, this link is on top of a gray background rather than a white one, so the link color has to be darkened a bit to maintain the contrast ratio.

But I agree it would make sense to re-evaluate the site colors in the context of contrast ratios. If the colors were darker to begin with, they might work on a gray background without having to darken them further.

PR #2531, opened 2021-06-07
[CA-1309] A11y: Groups and Billing Projects pages

andrewmarcus:
## Jun 9 comment

This PR will largely be superceded by #2532. Hold off on review for now until @s-rubenstein and I determine what to do about this branch.

## Original

This PR is based on the #2529 branch, since I use a `MenuTrigger` feature from there and a `SimpleTabBar` feature from the #2512 branch from which it's derived. Each of those two PRs should be merged for this one.

The most significant change in this PR was that I added table semantics to the card lists across these sets of pages. They were already functionally tables, with elements within each card lined up in columns, and all but one of the tables already had sortable column headers.

I might suggest a future refactor to convert the card table-ish structures into true `FlexTable`s for consistency with the rest of the site. This would be particularly good if some groups or billing projects have enough rows to warrant virtualizing the tables to achieve better performance. It would probably even be straightforward to build a `FlexCardTable` which uses the existing card styling on each row rather than typical table styling, if desired. However, I didn't do that in this PR because it would require restructuring the elements to define the bililng and group data column-wise rather than row-wise as it currently is.

### Changes to the Groups pages

* Turned card lists into semantic tables using existing sorting headers
* Updated to the new `MenuTrigger` for launching the menu (this particular change didn't make it into the #2529 branch)
* Added `<h2>` headers

### Changes to the Billing page

* Turned card lists into semantic tables using existing sorting headers
* Added new sorting headers to the Workspaces table
* Updated `SimpleTabBar` to include a label and move the focus to its `tabpanel`
* Added `<h2>` and `<h3>` headers


src/pages/billing/Project.js
```diff
@@ -54,15 +76,18 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
 
   const adminCanEdit = _.filter(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers).length > 1
 
   const tabToTable = {
-    workspaces: div({ style: { flexGrow: 1, width: '100%' } }, [
+    workspaces: div({ role: 'table', 'aria-label': 'workspaces list', style: { flexGrow: 1, width: '100%' } }, [
+      h(WorkspaceCardHeaders, { sort: workspaceSort, onSort: setWorkspaceSort }),
       _.flow(
         _.filter({ workspace: { namespace: projectName } }),
+        _.orderBy([`workspace.${workspaceSort.field}`], [workspaceSort.direction]),
```
s-rubenstein: Hold off on this, I have a more complicated implementation coming in with https://github.com/DataBiosphere/terra-ui/pull/2532

src/pages/groups/Group.js
```diff
@@ -103,18 +103,20 @@ const GroupDetails = ({ groupName }) => {
...
-            return h(MemberCard, {
-              adminLabel: 'admin',
-              userLabel: 'member',
-              member, adminCanEdit,
-              onEdit: () => setEditingUser(member),
-              onDelete: () => setDeletingUser(member)
-            })
-          }, _.orderBy([sort.field], [sort.direction], _.filter(({ email }) => Utils.textMatch(filter, email), members)))
-        ),
+        div({ role: 'table', 'aria-label': 'member list' }, [
```
s-rubenstein: This looks very similar to https://github.com/DataBiosphere/terra-ui/pull/2531/files#diff-63f09855608a6a241dc6a5b78476307d1ea6ef749c662b0bf3323c2fc87a8f28R101, do you think we could extract to a component? (I was hesitant before, because there were different stylings, but if we are wrapping in identical divs, that feels componentable.)

s-rubenstein (approved):
Looks fairly reasonable, and no blocking changes, just a few comments to think about!

src/pages/groups/Group.js
andrewmarcus: Yep, I was thinking the same thing, and was curious why there were different stylings. As in, was that intentional, or should they just be the same?

At any rate, I won't make any further changes to this PR right now, since #2532 supercedes portions of it.

dmohs (approved):
Focused my review on code style and state and found no issues.

zarsky-broad:
Oh, my bad, you already had a thumb from @dmohs, didn't mean to dismiss...

src/components/group-common.js
```diff
@@ -61,15 +62,15 @@ const UserMenuContent = ({ onEdit, onDelete }) => {
 const menuCardSize = 20
 
 export const MemberCardHeaders = Utils.memoWithName('MemberCardHeaders', ({ sort, onSort }) => {
-  return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem' } }, [
-    div({ style: { flex: 1 } }, [
+  return div({ role: 'row', style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem' } }, [
```
s-rubenstein, line 65: Are we at all worried about this semantically being different from the aria perspective than the workspace table?

s-rubenstein (approved):
LGTM

andrewmarcus:
In theory yes, but in practice I doubt a user is going to really care that
much as long as they can find what they're looking for. My colleague agreed
when he suggested going with a list for this one.

src/components/group-common.js
andrewmarcus: In theory yes, but in practice I doubt a user is going to really care that
much as long as they can find what they're looking for. My colleague agreed
when he suggested going with a list for this one.

src/components/group-common.js
```diff
@@ -61,15 +62,15 @@ const UserMenuContent = ({ onEdit, onDelete }) => {
...
+    div({ role: 'columnheader', 'aria-sort': ariaSort(sort, 'email'), style: { flex: 1 } }, [
       h(HeaderRenderer, { sort, onSort, name: 'email' })
     ]),
-    div({ style: { flex: 1 } }, [
+    div({ role: 'columnheader', 'aria-sort': ariaSort(sort, 'roles'), style: { flex: 1 } }, [
       h(HeaderRenderer, { sort, onSort, name: 'roles' })
     ]),
     // Width is the same as the menu icon.
-    div({ style: { width: menuCardSize } }, [
+    div({ role: 'columnheader', style: { width: menuCardSize } }, [
```
andy7i, line 73: just curious - action/buttons in a table don't need a header (like "actions" or "delete")?

andy7i (approved):
changes look reasonable. Tested manually 👍 
1 question but not blocking merge

src/components/group-common.js
andrewmarcus: They do need a header, but it's acceptable for it to be screenreader-only if the column is obvious visually (i.e. action buttons which are all the same). A header won't fit in a really narrow column anyway with looking weird.

`.sr-only` (borrowed from Bootstrap but was already in the Terra codebase) makes the header visible to screen readers but hides it visually (unlike `display: none` which hides it from both)

PR #2532, opened 2021-06-08
[CA-1255] Add billing project workspace list styling

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -2,7 +2,7 @@ import _ from 'lodash/fp'
 import * as qs from 'qs'
 import { Fragment, useEffect, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
-import { ButtonPrimary, IdContainer, Link, Select, SimpleTabBar, spinnerOverlay } from 'src/components/common'
+import { ButtonPrimary, HeaderRenderer, IdContainer, Link, Select, SimpleTabBar, spinnerOverlay } from 'src/components/common'
 import { DeleteUserModal, EditUserModal, MemberCard, MemberCardHeaders, NewUserCard, NewUserModal } from 'src/components/group-common'
 import { icon, spinner } from 'src/components/icons'
 import Modal from 'src/components/Modal'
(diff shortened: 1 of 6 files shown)


src/pages/billing/Project.js
```diff
@@ -52,18 +110,32 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
   const [showBillingModal, setShowBillingModal] = useState(false)
   const [selectedBilling, setSelectedBilling] = useState()
   const [tab, setTab] = useState(query.tab || 'workspaces')
+  const [expandedWorkspace, setExpandedWorkspace] = useState()
```
dmohs: Based on the usage I see, I might call this `expandedWorkspaceName` instead.

src/pages/billing/Project.js
```diff
@@ -52,18 +110,32 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
+        _.flow(
+          // TODO (Post PPW): Remove billing account name here, and move back to just returning workspace. This is for a seamless transition to PPW, where `billingAccountName` should be a field on the workspace.
+          _.map(({ workspace }) => { return { billingAccountName, ...workspace } }),
+          _.filter({ namespace: projectName }),
+          _.orderBy([workspaceSort.field], [workspaceSort.direction]),
+          _.map(workspace => {
+            const isExpanded = expandedWorkspace === workspace.name
+            return h(WorkspaceCard, {
+              workspace, key: workspace.workspaceId, isExpanded,
+              onExpand: () => isExpanded ? setExpandedWorkspace(undefined) : setExpandedWorkspace(workspace.name)
```
dmohs: Alternatively:
```suggestion
              onExpand: () => setExpandedWorkspace(isExpanded ? undefined : workspace.name)
```

dmohs (approved):
Looks good. Tiny suggestions.

src/pages/billing/Project.js
```diff
@@ -19,20 +20,76 @@ import * as Utils from 'src/libs/utils'
...
 
-const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace }) => div({
-  style: Style.cardList.longCardShadowless
-}, [
-  div({ style: styles.workspaceCardField }, [workspace.name]),
-  div({ style: styles.workspaceCardField }, [workspace.createdBy]),
-  div({ style: { flex: `0 0 ${workspaceLastModifiedWidth}px` } }, [workspace.lastModified])
-]))
+  return div({ style: expandedInfoStyles.row, role: 'row' }, [
+    div({ role: 'cell', style: expandedInfoStyles.title }, [title]),
```
s-rubenstein: @andrewmarcus, this is the one I most don't know about. Semantically, should this be a cell or a row header?

src/pages/billing/Project.js
```diff
@@ -19,20 +20,76 @@ import * as Utils from 'src/libs/utils'
...
+  }
+
+  return div({ role: 'row', style: { ...Style.cardList.longCardShadowless, flexDirection: 'column' } }, [
+    div({ style: workspaceCardStyles.row }, [
+      div({ role: 'rowheader', style: { ...workspaceCardStyles.field, display: 'flex', alignItems: 'center' } }, [
+        h(Link, {
+          style: { fontWeight: 600, fontSize: 16, ...Style.noWrapEllipsis },
+          href: Nav.getLink('workspace-dashboard', { namespace, name })
+        }, [name]),
+        h(Link, {
```
s-rubenstein, line 74: Does this need anything else for accessibility @andrewmarcus? Or do our components handle all that for us?

src/pages/billing/Project.js
andrewmarcus: Could go either way. I made these `rowheader`s in my version because it seemed that the title pretty cleanly described the row, and was generally unique within the set.

If you do include a rowheader, then users of assistive technologies will hear both the row header and column header on each cell, which helps with orientation. I didn't do this in the case of the general tables though, since I wasn't sure the first column would always be the row header. I probably should go back and add that, though, now that I think about it.

src/pages/billing/Project.js
andrewmarcus: Well, you'll need the link to have either an `aria-label` or a `tooltip` which becomes a label once #2525 is merged. The `aria-expanded` is perfect though

src/pages/billing/Project.js
```diff
@@ -52,18 +110,32 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
-    workspaces: div({ style: { flexGrow: 1, width: '100%' } }, [
-      _.flow(
-        _.filter({ workspace: { namespace: projectName } }),
-        _.map(({ workspace }) => h(WorkspaceCard, { workspace, key: workspace.workspaceId }))
-      )(workspaces)
+    workspaces: h(Fragment, [
+      h(WorkspaceCardHeaders, { sort: workspaceSort, onSort: setWorkspaceSort }),
+      div({ style: { flexGrow: 1, width: '100%' } }, [
+        _.flow(
+          // TODO (Post PPW): Remove billing account name here, and move back to just returning workspace. This is for a seamless transition to PPW, where `billingAccountName` should be a field on the workspace.
```
marctalbott, line 126: Can you make a ticket for this and put it in the cleanup epic if it doesn't exist already?

src/pages/billing/Project.js
andrewmarcus: It appears you're trying to create an expandable section within a row. You can't just nest `row`s inside each other, and you can't add more `cell`s to a row than there are `columnheader`s. These roles essentially just replicate `<table>`, `<tr>`, `<th>` and `<td>`.

Because of the expanding interaction, it probably no longer makes sense to use table semantics. You could try `treegrid`, or grouping these cards into `rowgroup`s with extra columns for the inner tables. But all of those are much more complicated than this really needs.

I conferred with a colleague. We recommend it's probably best to just go back to normal list semantics, where each list item has a button to expand or collapse the card.

```
<div role="list">
  <div role="listitem">
    <button type="button" aria-expanded="true" aria-controls="id1">2019_ASGH</button>
    <div>cloreth@broadinstitute.org</div>
    <div>Jan 25, 2021</div>
    <div id="id1">
        <div>Google project</div>
        ...
    </div>
  </div>
</div>
```

The sortable column headers will be interesting. But I think they are already buttons and already have ARIA descriptions telling users to click them to sort. That's likely good enough.

andrewmarcus: I added similar semantics to the `Collapse` component in one of the PRs, I think it was #2526. You might be able to reuse that component here, or at least do the same thing it does

marctalbott (commented):
The bottom of emails, google project names, and billing accounts is being cut off a little -- see the bottom of the 'g' in the screenshot
Screen Shot 2021-06-09 at 5 10 35 PM

I also noticed that you can only have one workspace expanded at once, is that intended?

Did Janki end up weighing in on whether all workspace users should be able to see the name of the Google billing account or not?

s-rubenstein:
@marctalbott Janki has not weighed in yet. I will check in with her later today on that. 

Janki did say that she preferred one workspace expanded at a time for now, and we can change it if users complain.

jacmrob (commented):
From my limited knowledge, code LGTM.

One question – from clicking around I found that when you expand one set of workspace details, any open one for another workspace closes. effectively, you can only have one workspace details open at a time. is this intended behavior?

jacmrob:
Ah nevermind I just saw you answered that question. Great! 👍

jacmrob (approved):
Question answered 👍

s-rubenstein:
@marctalbott I have fixed the 'g's, I think!

marctalbott (approved):
Looks awesome!

PR #2533, opened 2021-06-09
[CA-1311] A11y: Terms of Service, HoF and Cookiebar

andrewmarcus:
This PR makes the following updates:

### Terms of Service page

* Added 'main' region
* Marked up the title as an `<h1>`
* Converted all headers in the markdown to `<h2>` instead of `<h4>`. This significantly changed their styling, but I'm not sure whether that matters.
* Fixed the Markup link text to include more than just "here".
* Increased the height of the scrollable region on larger screens for easier reading

These changes have a noticeable visual effect on the page. In addition, they involved changes to the Terms of Service markdown. The ToS changes are insubstantial and do not constitute a legal change to the terms people agreed to, in that they change the styling but not the text itself.

Are both of these effects OK? Or do I need to consult with the UX team or take a different approach?

### GDPR Cookie Bar

* Increased contrast of links and buttons
* Used `<header>` tag instead of `role="banner"` since the latter must be unique on the page

### Hall of Fame

* Added headers and list semantics


gpolumbo-broad:
FWIW, looks good to me from a code perspective, @dmohs or @zarsky-broad, as for actual visual changes, who approves those?  I'm inclined to run them by Jerome?  

I'll withhold my thumb until we can get that sorted.

gpolumbo-broad:
FWIW - the style changes seem totally reasonable/acceptable to me

src/components/CookieWarning.js
```diff
@@ -47,13 +46,13 @@ const CookieWarning = () => {
         and to improve your experience. By clicking Agree or continuing to use our site, you consent to the use of these functional
         cookies. If you do not wish to allow use of these cookies, you may tell us that by clicking on Reject. As a result, you will be unable
         to use our site. To find out more, read our `,
-        h(Link, { style: { textDecoration: 'underline' }, href: Nav.getLink('privacy') }, ['privacy policy']), '.'
+        h(Link, { style: { textDecoration: 'underline', color: colors.accent(1.1) }, href: Nav.getLink('privacy') }, ['privacy policy']), '.'
```
s-rubenstein, line 49: I am definitely a little concerned at how often we need to do the 1.1 colors.accent. I am going to make a note to flag that to our design team.

src/pages/TermsOfService.js
```diff
@@ -314,12 +315,15 @@ const TermsOfServicePage = () => {
...
     backgroundLogo,
-    div({ style: { backgroundColor: 'white', borderRadius: 5, width: 800, padding: '2rem', boxShadow: Style.standardShadow } }, [
-      div({ style: { color: colors.dark(), fontSize: 38 } }, ['Terra Terms of Service']),
+    div({ style: { backgroundColor: 'white', borderRadius: 5, width: 800, maxHeight: '100%', padding: '2rem', boxShadow: Style.standardShadow } }, [
+      h1({ style: { color: colors.dark(), fontSize: 38, fontWeight: 400 } }, ['Terra Terms of Service']),
       needToAccept && div({ style: { fontSize: 18, fontWeight: 600 } }, ['Please accept the Terms of Service to continue.']),
-      div({ style: { maxHeight: 400, overflowY: 'auto', lineHeight: 1.5, marginTop: '1rem', paddingRight: '1rem' } }, [
+      div({
+        ref: scrollRef,
+        style: { height: '50vh', overflowY: 'scroll', lineHeight: 1.5, marginTop: '1rem', paddingRight: '1rem' }
```
s-rubenstein: Where does the 50vh come from? And why does 'auto' not work for the overflow here? I would expect/hope to avoid use of 'scroll' because it means when the screen is large enough, you still have an empty scrollbar.

src/components/CookieWarning.js
andrewmarcus: True! It's definitely worth having a conversation with them about this.

Keep in mind, though, that context matters. For instance, `colors.accent(1.0)` has sufficient contrast against white, either when used as blue text on a white background (links) or white text on a blue background (primary buttons). The problem here was that the background is not white. The cookie warning uses a gray background to offset it from the main part of the page.

That said, the green color used in the header banner, `colors.primary(1.0)` does not have sufficient contrast with white, hence #2506. That one could definitely used to be darkened, although I didn't want to make that sort of global change unilaterally.

Similarly, the orange `colors.warning(1.0)` does not have nearly enough contrast against background. When I darkened it to 4.5:1 in #2538 it basically became red. So maybe that color should just go away completely and use `colors.danger()` everywhere instead.

src/pages/TermsOfService.js
andrewmarcus: Basically, I was just trying to get more text into the panel and make it clear that the user must scroll, while not changing the look and feel too much. Macs don't show the scrollbar until you start scrolling, so the first time I saw this page it wasn't obvious that there was more to read.

I played around with adding a button to automatically scroll down a page, (and to indicate that scrolling is necessary) but this was complicated and didn't really seem to help the situation much so I pulled that back out. Hitting the spacebar key does the same thing without needing to do any scroll state synchronization.

`auto` vs `scroll`: yeah, you're right, auto would be fine. On the other hand, the ToS text is baked into the source code and is WAY longer than one page, so I can't imagine there would ever be a scenario where the text fits on a single page without scrolling, even now that the panel can get bigger than 400px. Mostly I was trying to see if I could force MacOS to render the scrollbars. It didn't make a difference, so I could change it back if you prefer. Technically that `scrollRef` isn't necessary anymore either.

`50vh`: I tested a couple of different values relative to the viewport height, and this one seemed to work pretty well for both small and large browser windows. At moderate screen sizes it is fairly similar to how it looks pinned to `400px`, but it allows the panel to expand with taller screens, and it doesn't collapse too much in really short browser windows. At "50% of the viewport height", the window itself has to be something like 300px or less to get a double scrollbar on the text region. Not perfect, but since the page won't get narrower than 1200px anyway that seemed reasonable. It doesn't appear that any of the pages are intended to be mobile-responsive.

src/components/CookieWarning.js
s-rubenstein: Great. I will add a ticket for our team to re-evaluate our base colors, and bring that to the UX team's attention. (I don't think you should worry any more than this question on it though!)

s-rubenstein (approved):
Thank you for moving back to `auto`! LGTM

src/pages/TermsOfService.js
```diff
@@ -314,12 +314,12 @@ const TermsOfServicePage = () => {
...
-  return div({ style: { padding: '1rem', minHeight: '100%', display: 'flex', justifyContent: 'center', alignItems: 'center' } }, [
+  return div({ role: 'main', style: { padding: '1rem', minHeight: '100%', display: 'flex', justifyContent: 'center', alignItems: 'center' } }, [
     backgroundLogo,
-    div({ style: { backgroundColor: 'white', borderRadius: 5, width: 800, padding: '2rem', boxShadow: Style.standardShadow } }, [
-      div({ style: { color: colors.dark(), fontSize: 38 } }, ['Terra Terms of Service']),
+    div({ style: { backgroundColor: 'white', borderRadius: 5, width: 800, maxHeight: '100%', padding: '2rem', boxShadow: Style.standardShadow } }, [
+      h1({ style: { color: colors.dark(), fontSize: 38, fontWeight: 400 } }, ['Terra Terms of Service']),
       needToAccept && div({ style: { fontSize: 18, fontWeight: 600 } }, ['Please accept the Terms of Service to continue.']),
-      div({ style: { maxHeight: 400, overflowY: 'auto', lineHeight: 1.5, marginTop: '1rem', paddingRight: '1rem' } }, [
+      div({ style: { height: '50vh', overflowY: 'auto', lineHeight: 1.5, marginTop: '1rem', paddingRight: '1rem' } }, [
```
dmohs, line 322: I wonder if I'm the only person who is in love with the availability of the vx length specifiers.

dmohs (approved):

s-rubenstein:
@gpolumbo-broad usually Jerome is who I'd check in with, however he is out and this is time sensitive. I'd try Joy or Lou?

@andrewmarcus, would you mind adding screenshots of the style changes to PRs if its significant changes?

andrewmarcus:
## Tall screens
### Old
Screen Shot 2021-06-11 at 12 19 26 PM
### New
Screen Shot 2021-06-11 at 12 18 00 PM

## Short screens
### Old
Screen Shot 2021-06-11 at 12 19 36 PM
### New
Screen Shot 2021-06-11 at 12 20 09 PM

gpolumbo-broad (approved):

PR #2538, opened 2021-06-10
[CA-1310] A11y: Import Pages

andrewmarcus:
Addressed accessibility on the import-data and import-workflow pages.

I tested with the following: 
* http://localhost:3000/#import-data?url=https://jade.datarepo-dev.broadinstitute.org&snapshotId=f90f5d7f-c507-4e56-abfc-b965a66023fb&snapshotName=testsnapshot&format=snapshot
* http://localhost:3000/#import-workflow/dockstore/github.com/DataBiosphere/topmed-workflows/UM_variant_caller_wdl:1.32.0

### Changes in this PR

* Added h2 headers as appropriate
* Ensured Select elements are properly linked to labels
* Increased contrast of warning text
* Moved background image into the main section so it's not floating off by itself
* Surrounded radio buttons into a radiogroup

I also noticed that many instances of Select combobox components do not have `aria-label`s or linked `label`s, even after #2513 has been merged. I'll need to address that in another PR.

### DIFF
--- a/src/pages/ImportData.js
+++ b/src/pages/ImportData.js
@@ -1,8 +1,8 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
-import { div, h, img, p } from 'react-hyperscript-helpers'
+import { div, h, h2, img, p } from 'react-hyperscript-helpers'
 import Collapse from 'src/components/Collapse'
-import { backgroundLogo, ButtonPrimary, ButtonSecondary, Clickable, RadioButton, spinnerOverlay } from 'src/components/common'
+import { backgroundLogo, ButtonPrimary, ButtonSecondary, Clickable, IdContainer, RadioButton, spinnerOverlay } from 'src/components/common'
 import FooterWrapper from 'src/components/FooterWrapper'
 import { icon, wdlIcon } from 'src/components/icons'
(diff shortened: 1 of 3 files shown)


gpolumbo-broad (approved):

src/pages/ImportData.js
```diff
@@ -135,16 +136,19 @@ const ImportData = () => {
...
-                onChange: setSelectedWorkspaceId
-              }),
+              h2({ style: styles.title }, ['Start with an existing workspace']),
+              h(IdContainer, [id => h(Fragment, [
+                h(FormLabel, { htmlFor: id, style: { marginBottom: '0.25rem' } }, ['Select one of your workspaces']),
+                h(WorkspaceSelector, {
+                  id,
+                  workspaces: _.filter(ws => {
+                    return Utils.canWrite(ws.accessLevel) &&
+                      (!ad || _.some({ membersGroupName: ad }, ws.workspace.authorizationDomain))
```
dmohs, line 146: Not your code, but this was an interesting case. I generally want the ability for a variable to describe itself to vary with the physical size of the scope. `x` is preferred when contained within a few lines, since the context is readily visible, whereas something more descriptive is better when the scope is larger. In this case, `ad` may be fine even though the scope is a bit larger because it appears to be a common pattern.

dmohs (approved):

PR #2540, opened 2021-06-10
[CA-1255] UX Improvement ideas on workspace list

s-rubenstein:
This PR addresses some issues Andrew Marcus and I found with the workspace list on the billing page. Specifically, with the larger font-size for the workspace links, it was removing some of the focus rings around that and the expand button. It also looked cleaner to have the expand button move to the right side, since it is a different affordance from the link to the workspace, and reduces the chance of accidental wrong clicks.

Old: (Importantly, the highlight bars are cut off on the expand button and the link to workspace)
Screen Shot 2021-06-10 at 4 07 57 PM

New:
Screen Shot 2021-06-10 at 4 06 53 PM

### DIFF
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -34,7 +34,7 @@ const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ sort,
     div({ 'aria-sort': ariaSort(sort, 'lastModified'), style: { flex: `0 0 ${workspaceLastModifiedWidth}px` } }, [
       h(HeaderRenderer, { sort, onSort, name: 'lastModified' })
     ]),
-    div({ role: 'columnheader', style: { flex: `0 0 ${workspaceExpandIconSize}px` } }, [
+    div({ style: { flex: `0 0 ${workspaceExpandIconSize}px` } }, [
       div({ className: 'sr-only' }, ['Expand'])
     ])
   ])
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -21,10 +21,11 @@ import { billingRoles } from 'src/pages/billing/List'
 
 
 const workspaceLastModifiedWidth = 150
+const workspaceExpandIconSize = 20
 
 const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ sort, onSort }) => {
   return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem', marginBottom: '0.5rem' } }, [
-    div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1 } }, [
+    div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '1rem' } }, [
       h(HeaderRenderer, { sort, onSort, name: 'name' })
     ]),
     div({ 'aria-sort': ariaSort(sort, 'createdBy'), style: { flex: 1 } }, [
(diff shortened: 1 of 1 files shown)


jacmrob (approved):
Like having the expand icon all the way to the right!

dmohs (approved):
>reduces the chance of accidental wrong clicks

Absolutely beautiful.

andrewmarcus:
Love it! Thanks!

src/pages/billing/Project.js
```diff
@@ -21,17 +21,21 @@ import { billingRoles } from 'src/pages/billing/List'
...
+    div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '1rem' } }, [
       h(HeaderRenderer, { sort, onSort, name: 'name' })
     ]),
     div({ 'aria-sort': ariaSort(sort, 'createdBy'), style: { flex: 1 } }, [
       h(HeaderRenderer, { sort, onSort, name: 'createdBy' })
     ]),
     div({ 'aria-sort': ariaSort(sort, 'lastModified'), style: { flex: `0 0 ${workspaceLastModifiedWidth}px` } }, [
       h(HeaderRenderer, { sort, onSort, name: 'lastModified' })
+    ]),
+    div({ role: 'columnheader', style: { flex: `0 0 ${workspaceExpandIconSize}px` } }, [
```
andrewmarcus: Thank you for thinking to put the column header in! But actually, since this is no longer a table, this will probably be flagged by aXe as an invalid role: a `columnheader` floating around outside a `table` or `row` container. I'd say for now you don't need this, unless you change it back to being a table at a later point.

andrewmarcus: Incidentally, this is an issue the `axe-core` linter should be able to catch. It very easily catches too-much-ARIA, it's the not-enough-ARIA case that it sometimes struggles with. :)

src/pages/billing/Project.js
```diff
@@ -21,17 +21,21 @@ import { billingRoles } from 'src/pages/billing/List'
...
+      div({ className: 'sr-only' }, ['Expand'])
```
andy7i, line 38: "sr-only" className?

andy7i (approved):
looks great! tested locally. 👍 
1 question but not blocking

src/pages/billing/Project.js
andy7i: sky and I slacked about this:
"sr" stands for screenreader
https://getbootstrap.com/docs/4.0/utilities/screenreaders/

dmohs (approved):
With @petesantos.

src/pages/billing/Project.js
```diff
@@ -55,33 +59,34 @@ const ExpandedInfoRow = Utils.memoWithName('ExpandedInfoRow', ({ title, details,
...
     expandedInfoContainer: { display: 'flex', flexDirection: 'column', width: '100%' }
   }
 
   return div({ role: 'listitem', style: { ...Style.cardList.longCardShadowless, flexDirection: 'column' } }, [
     div({ style: workspaceCardStyles.row }, [
-      div({ style: { ...workspaceCardStyles.field, display: 'flex', alignItems: 'center' } }, [
+      div({ style: { ...workspaceCardStyles.field, display: 'flex', alignItems: 'center', paddingLeft: '1rem' } }, [
         h(Link, {
-          style: { fontWeight: 600, fontSize: 16, ...Style.noWrapEllipsis },
+          style: { ...Style.noWrapEllipsis },
```
zarsky-broad:
```suggestion
          style: Style.noWrapEllipsis,
```

PR #2541, opened 2021-06-10
[CA-1398] A11Y: Ensure that Select comboboxes have a label

andrewmarcus:
In this PR, I did a search through the code for all usages of `Select` and `GroupedSelect` to ensure they are labeled properly. Most were, but this PR fixes the few remaining ones that weren't. In addition, I added an assertion to alert developers if they're on a page with an improperly labeled `Select` component.

In addition, I disabled the logic in `react-modal` which marks the whole page as `aria-hidden` whenever a modal is open. Since the modal now properly sets `aria-modal="true"`, this is unnecessary since screen readers will ignore the rest of the page anyway. Instead it just results in false positives in aXe and other a11y testing tools, which incorrectly indicate the `main` role and `h1` are missing whenever a modal is open.

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -362,6 +362,8 @@ const BaseSelect = ({ value, newOptions, id, findValue, maxHeight, ...props }) =
  * @param props.id - The HTML ID to give the form element
  */
 export const Select = ({ value, options, id, ...props }) => {
+  Utils.useConsoleAssert(props.id || props['aria-label'] || props['aria-labelledby'], 'In order to be accessible, Select needs a label')
+
   const newOptions = options && !_.isObject(options[0]) ? _.map(value => ({ value }), options) : options
   const findValue = target => _.find({ value: target }, newOptions)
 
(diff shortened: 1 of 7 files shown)


src/components/common.js
```diff
@@ -362,6 +362,8 @@ const BaseSelect = ({ value, newOptions, id, findValue, maxHeight, ...props }) =
  * @param props.id - The HTML ID to give the form element
  */
 export const Select = ({ value, options, id, ...props }) => {
+  Utils.useConsoleAssert(props.id || props['aria-label'] || props['aria-labelledby'], 'In order to be accessible, Select needs a label')
```
s-rubenstein, line 365: Nice, this is very clean.

s-rubenstein (approved):
LGTM, nice and clean!

src/components/common.js
andrewmarcus: Yeah, actually, at some point this should be made consistent in all of the widgets like this. I'm not sure it is...

marctalbott (approved):

dmohs (approved):
With @petesantos.

PR #2546, opened 2021-06-15
[CA-1407] A11y: Ensure all components have accessible labels

andrewmarcus:
This PR standardizes the way labels are enforced across all components. It provides an actionable warning in the console if a component is missing a label that it should have.

It accepts labels in any of the following formats:
* `aria-label` containing the text
* `aria-labelledby` linking to the `id` of a component with the text (except for tables which need the text directly)
* `id` linked to by a label with a `for` property (for form elements)
* `tooltip` containing the text (for `Clickable` and `NumberInput` components)

This PR also renames the `tableName` property in tables and `label` property in TabBars to just use `aria-label` like everything else. This change affected everywhere where these components where used, which is why this PR is large.

### DIFF
--- a/src/components/PopupTrigger.js
+++ b/src/components/PopupTrigger.js
@@ -25,7 +25,7 @@ const styles = {
 export const Popup = onClickOutside(function({ id, side = 'right', target: targetId, onClick, children, popupProps = {} }) {
   // We're passing popupProps here rather than just props, because ...props also includes lots of internal onClickOutside properties which
   // aren't valid to be dropped on a DOM element.
-  Utils.useConsoleAssert(popupProps['aria-label'] || popupProps['aria-labelledby'], 'In order to be accessible, Popup needs a label')
+  Utils.useLabelAssert('Popup', { ...popupProps })
 
   const elementRef = useRef()
   const [target, element, viewport] = useDynamicPosition([{ id: targetId }, { ref: elementRef }, { viewport: true }])
--- a/src/components/PopupTrigger.js
+++ b/src/components/PopupTrigger.js
@@ -25,7 +25,7 @@ const styles = {
 export const Popup = onClickOutside(function({ id, side = 'right', target: targetId, onClick, children, popupProps = {} }) {
   // We're passing popupProps here rather than just props, because ...props also includes lots of internal onClickOutside properties which
   // aren't valid to be dropped on a DOM element.
-  Utils.useLabelAssert('Popup', { ...popupProps })
+  Utils.useLabelAssert('Popup', popupProps)
 
   const elementRef = useRef()
   const [target, element, viewport] = useDynamicPosition([{ id: targetId }, { ref: elementRef }, { viewport: true }])
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -192,9 +192,9 @@ const DataTable = props => {
           ({ width, height }) => {
             return h(GridTable, {
               ref: table,
+              'aria-label': `${entityType} data table, page ${pageNumber} of ${Math.ceil(totalRowCount / itemsPerPage)}`,
               width, height,
               rowCount: entities.length,
-              tableName: `${entityType} data table, page ${pageNumber} of ${Math.ceil(totalRowCount / itemsPerPage)}`,
               noContentMessage: `No ${entityType}s to display.`,
               onScroll,
               initialX,
--- a/src/components/data/FileBrowser.js
+++ b/src/components/data/FileBrowser.js
@@ -192,10 +192,10 @@ export const FileBrowserPanel = _.flow(
         h(AutoSizer, {}, [
           ({ width, height }) => h(FlexTable, {
             ref: table,
+            'aria-label': 'file browser',
             width,
             height,
             rowCount: numPrefixes + numObjects,
-            tableName: 'file browser',
             noContentMessage: 'No files have been uploaded yet',
             onScroll: saveScroll,
             initialY,
--- a/src/components/data/UploadPreviewTable.js
+++ b/src/components/data/UploadPreviewTable.js
@@ -209,9 +209,9 @@ const UploadDataTable = props => {
           ({ width, height }) => {
             return h(GridTable, {
               ref: table,
+              'aria-label': `metadata preview table`,
               width, height,
               rowCount: sortedRows.length,
-              tableName: `metadata preview table`,
               noContentMessage: `No ${entityType}s to display.`,
               onScroll: saveScroll, initialX, initialY, sort,
               columns: [
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -67,7 +67,7 @@ export const withDebouncedChange = WrappedComponent => {
 }
 
 export const TextInput = Utils.forwardRefWithName('TextInput', ({ onChange, nativeOnChange = false, ...props }, ref) => {
-  Utils.useConsoleAssert(props.id || props['aria-label'], 'In order to be accessible, TextInput needs a label')
+  Utils.useLabelAssert('TextInput', { ...props, allowId: true })
 
   return input({
     ..._.merge({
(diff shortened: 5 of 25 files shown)


src/components/PopupTrigger.js
```diff
@@ -25,7 +25,7 @@ const styles = {
 export const Popup = onClickOutside(function({ id, side = 'right', target: targetId, onClick, children, popupProps = {} }) {
   // We're passing popupProps here rather than just props, because ...props also includes lots of internal onClickOutside properties which
   // aren't valid to be dropped on a DOM element.
-  Utils.useConsoleAssert(popupProps['aria-label'] || popupProps['aria-labelledby'], 'In order to be accessible, Popup needs a label')
+  Utils.useLabelAssert('Popup', { ...popupProps })
```
zarsky-broad: Any reason we need this spread? Or can we just pass `popupProps`?

src/components/data/DataTable.js
```diff
@@ -192,9 +192,9 @@ const DataTable = props => {
           ({ width, height }) => {
             return h(GridTable, {
               ref: table,
+              'aria-label': `${entityType} data table, page ${pageNumber} of ${Math.ceil(totalRowCount / itemsPerPage)}`,
```
petesantos, line 195: could you move the `aria` props to the top?

src/components/PopupTrigger.js
andrewmarcus: Yep, I was doing this everywhere since some also have other properties. But you're right about this one.

src/libs/utils.js
```diff
@@ -382,6 +382,55 @@ export const useConsoleAssert = (condition, message) => {
   }
 }
 
+/**
+ * Asserts that a component has an accessible label, and alerts the developer how to fix it if it doesn't.
+ *
+ * @param componentName The name of the component, which will be printed to the console if there's an alert.
+ * @param allowLabelledBy If true (default), the component can have an aria-labelledby linked to another element. Set to false to only allow aria-label.
```
zarsky-broad: Would you mind marking the optional ones using the [jsdoc syntax](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values)?

src/libs/utils.js
```diff
@@ -382,6 +382,55 @@ export const useConsoleAssert = (condition, message) => {
...
+
+/**
+ * Returns the aria-label to use for a component if applicable.
+ *
+ * If aria-label was provided, it will be returned
+ * If aria-labelledby was provided, or an id was provided and allowId is true, then this will return nothing because the component is already labeled
+ * If a tooltip was provided without any of the other attributes, it will be returned to be used as the aria-label
+ */
+export const useAriaLabelOrTooltip = ({ allowId = false, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, id, tooltip }) => {
+  return ariaLabel || ariaLabelledBy || (allowId && id) ? ariaLabel : tooltip
```
zarsky-broad: Can you add some clarifying parens to this? If I'm reading it right, they'd go around `(allowId && id) ? ariaLabel : tooltip`

zarsky-broad (approved):
I love love love the way you did this. It's been a pleasure having you working on this codebase. I really like establishing the pattern of putting the label assert first in the component and putting a newline after it.

src/libs/utils.js
andrewmarcus: Oh, you're right. That is confusing.

src/libs/utils.js
```diff
@@ -382,6 +382,55 @@ export const useConsoleAssert = (condition, message) => {
...
+
+  if (!printed.current) {
+    // Ensure that the properties contain a label
+    if (!(ariaLabel ||
+      (allowLabelledBy && ariaLabelledBy) ||
+      (allowId && id) || (allowTooltip && tooltip)
+    )) {
+      printed.current = true
+
+      console.warn(`For accessibility, ${componentName} needs a label. Resolve this by doing any of the following: ${allowContent ? `
```
s-rubenstein, line 412: Love this message! Very clear.

s-rubenstein (approved):
LGTM

src/components/tabBars.js
```diff
@@ -75,10 +75,13 @@ export const TabBar = ({
     ])
   }
 
-  return div({ role: 'navigation', 'aria-label': label }, [
+  return div({
+    role: 'navigation',
```
petesantos, line 79: move `aria-label` to the top of the props list

src/components/input.js
```diff
@@ -146,11 +146,14 @@ export const SearchInput = ({ value, onChange, ...props }) => {
...
 export const NumberInput = ({ onChange, onBlur, min = -Infinity, max = Infinity, onlyInteger = false, isClearable = true, tooltip, value, ...props }) => {
-  Utils.useConsoleAssert(props.id || props['aria-label'], 'In order to be accessible, NumberInput needs a label')
+  // If the user provided a tooltip but no other label, use the tooltip as the label for the input
+  Utils.useLabelAssert('NumberInput', { tooltip, ...props, allowTooltip: true })
+
   const [internalValue, setInternalValue] = useState()
 
   const numberInputChild = div([input(_.merge({
     type: 'number',
+    'aria-label': Utils.useAriaLabelOrTooltip({ tooltip, ...props }),
```
petesantos: move `aria-label` to the top of the props list

src/components/data/UploadPreviewTable.js
```diff
@@ -209,9 +209,9 @@ const UploadDataTable = props => {
           ({ width, height }) => {
             return h(GridTable, {
               ref: table,
+              'aria-label': `metadata preview table`,
```
petesantos, line 212: move `aria-label` to the top of the props list

src/components/data/FileBrowser.js
```diff
@@ -192,10 +192,10 @@ export const FileBrowserPanel = _.flow(
         h(AutoSizer, {}, [
           ({ width, height }) => h(FlexTable, {
             ref: table,
+            'aria-label': 'file browser',
```
petesantos, line 195: move `aria-label` to the top of the props list

src/components/table.js
```diff
@@ -222,7 +224,7 @@ export const FlexTable = ({
     role: 'table',
     'aria-rowcount': rowCount + 1, // count the header row too
     'aria-colcount': columns.length,
-    'aria-label': tableName,
+    'aria-label': ariaLabel,
```
petesantos, line 228: Could you go through this file and move the aria props to the top?
If you would prefer to do a follow up PR for this to keep the PR noise down, I am fine with that as well.

src/libs/utils.js
```diff
@@ -382,6 +382,55 @@ export const useConsoleAssert = (condition, message) => {
...
+}
+
+/**
+ * Returns the aria-label to use for a component if applicable.
+ *
+ * If aria-label was provided, it will be returned
+ * If aria-labelledby was provided, or an id was provided and allowId is true, then this will return nothing because the component is already labeled
+ * If a tooltip was provided without any of the other attributes, it will be returned to be used as the aria-label
+ */
+export const useAriaLabelOrTooltip = ({ allowId = false, 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledBy, id, tooltip }) => {
```
petesantos: Could this be renamed to `getAriaLabelOrTooltip`.
The reason being this isn't a react hook and is being used deeper in the render function (rather than at the top) to provide values to the props of other components. 

The `use` in the name could confuse developers

petesantos (approved):
Looks good. 
Please see comments around nits regarding the order of aria props and one comment around a function rename.

src/libs/utils.js
MatthewBemis: +1

MatthewBemis (approved):
Not a UI expert but this looks good to me!

src/components/data/DataTable.js
andrewmarcus: Sure, I'd be happy to move the ARIA properties even farther to the top. However:

* The reason I listed `role` first, where applicable, is that `role` is an accessibility attribute. An HTML element's `role` determines which `aria-*` properties are available for that element. It seemed they should go together.
* The reason I listed `ref` and `key` first is that they have special meaning in React, which it seemed useful to call out as early as possible. This seems to be consistent with how these properties are used elsewhere in the codebase.

However, if you'd prefer I can push `aria-*` properties to the very top above these.

src/components/input.js
andrewmarcus: Like the others, `type` is pretty tightly bound to the form element and dictates how it will appear and its behavior from an accessibility perspective. But again, happy to bump the `aria-*` properties above it if you insist.

src/components/data/DataTable.js
petesantos: Thanks for the explanation - I like your rationale and am ok with the aria-* as-is - thanks!

PR #2547, opened 2021-06-16
[CA-1408] A11y: Resolved issues identified in 2nd audit

andrewmarcus:
Look at https://broadworkbench.atlassian.net/browse/CA-1408 for a complete list of resolutions included here.

By far the most complex one was the `ModalDrawer` updates. I'm not super happy with the way the ID of the title must be known in advance before the component is constructed, but that seemed like the easiest way to fit into the existing structure while allowing the title to change contextually. You may want to refactor this at some point.

### DIFF
--- a/src/components/ModalDrawer.js
+++ b/src/components/ModalDrawer.js
@@ -3,6 +3,7 @@ import { h } from 'react-hyperscript-helpers'
 import RModal from 'react-modal'
 import { Transition } from 'react-transition-group'
 import colors from 'src/libs/colors'
+import * as Utils from 'src/libs/utils'
 
 
 const drawer = {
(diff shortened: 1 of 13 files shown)


s-rubenstein (approved):
LGTM

dmohs (approved):
Looks great to me.

src/components/ModalDrawer.js
```diff
@@ -24,25 +25,28 @@ const drawer = {
...
     isOpen: true,
     onRequestClose: onDismiss,
     style: { overlay: drawer.overlay(transitionState), content: { ...drawer.container(transitionState), width } },
-    ariaHideApp: false,
     ...props
   }, [children])])
 }
 
-export const withModalDrawer = ({ width } = {}) => WrappedComponent => {
+export const withModalDrawer = ({ width, ...modalProps } = {}) => WrappedComponent => {
```
andy7i, line 47: do we need the spread operator here?

andy7i (approved):
1 question but not blocking. 👍 👍  otherwise

src/components/ModalDrawer.js
s-rubenstein: It is more accurate to the goal (which is as individual props), but not strictly speaking necessary. My inclination is to keep it with the spread operator, since that's more indicative of what it is trying to do.

PR #2548, opened 2021-06-16
[CA-1409] Update workflow config header

s-rubenstein:
I noticed an issue where `aria-label` isn't actually an allowed property on a span. This aligns this table with the way we do tables elsewhere.


src/pages/workflows/workflow/WorkflowDetails.js
```diff
@@ -215,9 +215,7 @@ const WorkflowConfigs = () => {
           rowCount: allConfigs.length,
           columns: [
             {
-              headerRenderer: () => span({
-                'aria-label': 'warnings'
-              }),
+              headerRenderer: () => div({ className: 'sr-only' }, ['Warnings']),
```
marctalbott, line 218: If it's a `div` now instead of a `span` should/can there be an `aria-label`? Or does the `sr-only` class name accomplish the same thing?

s-rubenstein: `sr-only` accomplishes the same thing. `aria-label` from my understanding is specifically for like… links/inputs/etc (interactive elements).

What `sr-only` does is basically hides the element, but because it is present still, screen readers will pick it up!

marctalbott: TIL! Thanks!

marctalbott (approved):

petesantos (approved):

andy7i (approved):
👍  LGTM, did not try testing manually.

PR #2551, opened 2021-06-16
handle null snapshot reference descriptions in markdown editor [AS-799]

davidangb:
Fixes a full-page crash when attempting to view the details of a snapshot reference inside a workspace. This only happened for snapshot references that had a null description; if you had previously given the reference a description, the bug did not happen. Unfortunately, descriptions are null for all newly-added snapshot references.

The error in question is in the markdown editor/viewer:
Screen Shot 2021-06-14 at 8 40 01 AM

This PR fixes the issue by sending the empty string to the markdown editor/viewer in cases where the underlying value is null.

I personally smoke-tested locally, but not comprehensively; @ungwudik did comprehensive testing as noted in comments below.

### DIFF
--- a/src/components/workspace-utils.js
+++ b/src/components/workspace-utils.js
@@ -142,7 +142,7 @@ export const SnapshotInfo = ({
           'Description:',
           !editingDescription && h(Link, {
             style: { marginLeft: '0.5rem' },
-            onClick: () => setNewDescription(description),
+            onClick: () => setNewDescription(description || ''), // description is null for newly-added snapshot references
             tooltip: 'Edit description'
           }, [icon('edit')])
         ]),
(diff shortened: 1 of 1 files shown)


ungwudik:
Checks passed. I'm setting up my local environment to poke around and also test.

ungwudik:
I tested locally and the change looks good. I'll also test in dev once this PR is merged.

Below are the test case steps I ran through:
1. add snapshot to an _existing_ workspace
2. rename snapshot
3. update snapshot description
4. repeat 2 and 3 with different input strings
5. delete snapshot from workspace
6. repeat 1 to 4
7. rename snapshot using text/strings like 'null', 'undefined'
8. update snapshot description using text/strings like 'null', 'undefined', empty string
9. Add snapshot to a _new_ workspace and repeat the above steps

ansingh7115 (approved):

src/components/workspace-utils.js
```diff
@@ -142,7 +142,7 @@ export const SnapshotInfo = ({
           'Description:',
           !editingDescription && h(Link, {
             style: { marginLeft: '0.5rem' },
-            onClick: () => setNewDescription(description),
+            onClick: () => setNewDescription(description || ''), // description is null for newly-added snapshot references
```
dmohs, line 145: Excellent comment.

dmohs (approved):

PR #2552, opened 2021-06-22
Add a small delay after typing text into an input field [QA-1484]

breilly2:
https://broadworkbench.atlassian.net/browse/QA-1484

See the code comment for the high-level what and why.

In this case, I found at least 2 causes of extra renders on the workspace list page as state was settling.

1. `WorkspaceList` is being rendered twice for each debounced keystroke. The first render results from calling `setFilter` to update component state. The second render is in response to `Nav.history.replace` to put the search filter in the query string.
2. When the workspace list goes from more than a page to all fitting on a page, the scrollbar is no longer necessary. This changes the `Grid`s calculated width which results in an extra call to `cellRenderer`.

#1 is something that we can fix (and I will have a PR for shortly... though I don't know if there are downsides so I haven't created a ticket for it yet). #2 is less in our control. Maybe we could change something to prevent things from shifting around depending on whether or not there is a scrollbar. However, I argue that these are not the kinds of errors that these tests are supposed to identify. Therefore, this PR proposes a more general solution to the problem of pages that automatically update when text is typed into an input field.

I've had CircleCI run the tests against this branch several times and have not see the QA-1484 error.

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -58,7 +58,20 @@ const input = ({ labelContains, placeholder }) => {
 }
 
 const fillIn = async (page, xpath, text) => {
-  return (await page.waitForXPath(xpath)).type(text, { delay: 20 })
+  const input = await page.waitForXPath(xpath)
+  await input.type(text, { delay: 20 })
+  // There are several places (e.g. workspace list search) where the page responds dynamically to
+  // typed input. That behavior could involve extra renders as component state settles. We strive to
+  // avoid the kinds of complex, multi-stage state transitions that can result in extra renders.
+  // But we aren't perfect.
+  //
+  // The impact on these tests is that elements found in the DOM immediately after typing text might
+  // get re-rendered (effectively going away) before the test can interact with them, leading to
+  // frustrating intermittent test failures. This test suite is _not_ intended to guard against
+  // unnecessary renders. It is to check that some specific critical paths through the application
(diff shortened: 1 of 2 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -58,7 +58,20 @@ const input = ({ labelContains, placeholder }) => {
...
+  // avoid the kinds of complex, multi-stage state transitions that can result in extra renders.
+  // But we aren't perfect.
+  //
+  // The impact on these tests is that elements found in the DOM immediately after typing text might
+  // get re-rendered (effectively going away) before the test can interact with them, leading to
+  // frustrating intermittent test failures. This test suite is _not_ intended to guard against
+  // unnecessary renders. It is to check that some specific critical paths through the application
+  // (Critical User Journeys) are not broken. Therefore, we'll delay briefly here instead of
+  // charging forward at a super-human pace.
+  return delay(250)
```
dmohs: This is an absolutely fantastic code comment.

dmohs (approved):

integration-tests/utils/integration-utils.js
zarsky-broad: So these delayed renders come from using `withDebouncedChange` under the hood, which has a debounce time of 250 hardcoded in it. I'd prefer we waited longer than that to be safe.

```suggestion
  return delay(300) // withDebouncedChange in input.js specifies 250ms, so waiting longer than that
```

zarsky-broad (approved):
One suggestion, then 👍

PR #2553, opened 2021-06-22
Remove duplicate state to avoid re-renders [SATURN-1903]

breilly2:
https://broadworkbench.atlassian.net/browse/SATURN-1903

Discovered during investigation for https://broadworkbench.atlassian.net/browse/QA-1484

I had a theory that the test code was locating an element to click on but that the element was being re-rendered before the click actually happened. My plan was to add a small sleep after entering the search text to give react a chance to settle down. However, I could not reliably reproduce the error locally, so I had no confidence that the fix was working. The closest I got was one case where puppeteer reported: "Node is detached from document" which seemed consistent with my theory.

I took a step back and analyzed `WorkspaceList` render patterns and found that it was rendering twice for every change in the search input: once from changing state with `setFilter` and once in reaction to `Nav.history.replace`. My conclusion is that the filter state was duplicated between the component state in `WorkspaceList` and the value in the query string.

The solution here removes the component state and relies only on the value in the query string.

This was not the only source of extra rendering, so it does not fix QA-1484, but I thought it might be worth doing anyway, though I don't know if there are any downsides to this. If it looks okay to folks, I'll create a Jira ticket for it.

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -58,9 +58,7 @@ const input = ({ labelContains, placeholder }) => {
 }
 
 const fillIn = async (page, xpath, text) => {
-  const input = await page.waitForXPath(xpath)
-  await input.type(text, { delay: 20 })
-  return delay(250)
+  return (await page.waitForXPath(xpath)).type(text, { delay: 20 })
 }
 
 // Replace pre-existing value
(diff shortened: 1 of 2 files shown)


dmohs (approved):
Thank you for not making us fix the dozen or so bugs this would have caused in the future.

petesantos (approved):
Nice - thanks!

zarsky-broad (approved):
👌

PR #2554, opened 2021-06-24
[IA-2679][IA-2814] UI support for creating GCE VMs with GPUs

kyuksel:
**Note:** This PR needs to be merged after other GPU support dependencies are in place.

## [IA-2814](https://broadworkbench.atlassian.net/browse/IA-2814)
image

Small amount of non-functional changes to fix wording on the window above around Galaxy pause, resume and auto-pause functionality. Please see the ticket for details. In hindsight, I should've PRed these changes separately but things have gone differently than I'd anticipated.

## [IA-2679](https://broadworkbench.atlassian.net/browse/IA-2679)

image

### Requirements
The requirements described in the ticket map to quite a few lower-level scenarios to handle differently for GPUs. I listed them [here](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2039906312/2021-06-24+GPU+UI+Requirements). These should also serve as helpful reference while manually testing the PR.

### Known Issue
Google "[projects have a global GPU quota](https://cloud.google.com/compute/docs/gpus#restrictions), which limits the total number of GPUs you can create in any supported zone." Therefore GPU creation requests can fail several times in a row. We are hoping to improve UX experience around this but this is beyond the scope of this PR.

### Remaining TODOs
✅   Don't display `Enable GPUs` when there is a GCE compute without GPUs
✅   Fix spacing such that Spark cluster disk size box fits into the section 
✅   Update `margin` and `padding` across `NewRuntimeModal` as per UX mock
🔲   Update the link to point to the GPU documentation once the User Ed team is done with it

### Beyond Scope & Future Work
- Add MixPanel analytics for GPU usage pattern to inform selections and resource quota issue severity
- Improving UX around failures due to the aforementioned quota issue
- Enable `GPU machine family (A2 series)` which is optimized for GPU use cases
- Improve UI messaging so people can find out more easily all available GPU types and numbers, even when some of them may not be allowed due to number of CPUs and/or memory restrictions
- Creating or patching non-GCE compute (i.e. Dataproc and K8s) with GPUs
- Attaching GPUs to an existing GCE compute
- Patching a GCE compute created with GPUs


src/components/NewGalaxyModal.js
```diff
@@ -168,7 +168,7 @@ export const NewGalaxyModalBase = Utils.withDisplayName('NewGalaxyModal')(
           div({ style: { flex: '1', lineHeight: '1.5rem', minWidth: 0, display: 'flex' } }, [
```
kyuksel, line 168: There is no functional change in this file.

src/components/NewRuntimeModal.js
```diff
@@ -46,7 +51,7 @@ const safeImageDocumentation = 'https://support.terra.bio/hc/en-us/articles/3600
 // distilled from https://github.com/docker/distribution/blob/95daa793b83a21656fe6c13e6d5cf1c3999108c7/reference/regexp.go
 const imageValidationRegexp = /^[A-Za-z0-9]+[\w./-]+(?::\w[\w.-]+)?(?:@[\w+.-]+:[A-Fa-f0-9]{32,})?$/
 
-const MachineSelector = ({ value, machineTypeOptions, onChange }) => {
+const WorkerSelector = ({ value, machineTypeOptions, onChange }) => {
```
kyuksel, line 55: `MachineSelector` used to be used in two different places to select GCE VMs' as well as Dataproc main machines' machine type. Now that GCE VMs can have GPU(s) attached, this component didn't factor well since we want available GPU type and number to dynamically and directly depend on CPU and Memory selections, which doesn't apply to Dataproc backends. Therefore, I renamed this component as (Dataproc) `WorkerSelector` which reflects its only purpose going forward. For the rest, I included the `Select` components flatly in `renderCloudComputeProfileSection()` (formerly `renderRuntimeSection()`).

src/components/NewRuntimeModal.js
```diff
@@ -131,15 +136,22 @@ export class NewRuntimeModalBase extends Component {
...
-      sparkMode: runtimeConfig?.cloudService === cloudServices.DATAPROC ? (runtimeConfig.numberOfWorkers === 0 ? 'master' : 'cluster') : false,
+      sparkMode,
       masterMachineType: runtimeConfig?.masterMachineType || runtimeConfig?.machineType,
       masterDiskSize: runtimeConfig?.masterDiskSize || runtimeConfig?.diskSize || DEFAULT_DISK_SIZE,
       numberOfWorkers: runtimeConfig?.numberOfWorkers || 2,
       numberOfPreemptibleWorkers: runtimeConfig?.numberOfPreemptibleWorkers || 0,
       workerMachineType: runtimeConfig?.workerMachineType || defaultDataprocMachineType,
-      workerDiskSize: runtimeConfig?.workerDiskSize || DEFAULT_DISK_SIZE
+      workerDiskSize: runtimeConfig?.workerDiskSize || DEFAULT_DISK_SIZE,
+      gpuEnabled: (!!gpuConfig && !sparkMode) || false,
```
kyuksel, line 152: We don't want `Enable GPUs` to be displayed for Dataproc compute and hence `!sparkMode`.

src/components/NewRuntimeModal.js
```diff
@@ -131,15 +136,22 @@ export class NewRuntimeModalBase extends Component {
...
+      hasGpu: !!gpuConfig,
```
kyuksel, line 153: We need to behave differently when there's an existing compute with GPUs (more on that later) vs. at compute creation time with no preexisting compute. That's why we have `hasGpu` in addition to `gpuEnabled`.

src/components/NewRuntimeModal.js
```diff
@@ -131,15 +136,22 @@ export class NewRuntimeModalBase extends Component {
...
+      gpuType: gpuConfig?.gpuType || DEFAULT_GPU_TYPE,
+      numGpus: gpuConfig?.numOfGpus || DEFAULT_NUM_GPUS
```
kyuksel, line 155: In general, we're defaulting to and ranking more highly in menus the least expensive configurations applicable.

src/components/NewRuntimeModal.js
```diff
@@ -751,51 +773,138 @@ export class NewRuntimeModalBase extends Component {
...
-      const gridStyle = { display: 'grid', gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', gridGap: '0.8rem', alignItems: 'center' }
+    const renderCloudComputeProfileSection = () => {
+      const { hasGpu, gpuEnabled, gpuType, numGpus } = this.state
+      const { cpu: currentNumCpus, memory: currentMemory } = findMachineType(mainMachineType)
+      const validGpuOptions = getValidGpuTypes(currentNumCpus, currentMemory)
+      const validGpuNames = _.flow(_.map('name'), _.uniq, _.sortBy('price'))(validGpuOptions)
+      const validGpuName = _.includes(displayNameForGpuType(gpuType), validGpuNames) ? displayNameForGpuType(gpuType) : _.head(validGpuNames)
+      const validNumGpusOptions = _.flow(_.filter({ name: validGpuName }), _.map('numGpus'))(validGpuOptions)
+      const validNumGpus = _.includes(numGpus, validNumGpusOptions) ? numGpus : _.head(validNumGpusOptions)
+      const gridStyle = { display: 'grid', gridGap: '1.5rem', alignItems: 'center', marginTop: '0.75rem' }
```
kyuksel: One-`gridStyle`-fit-all wasn't easy to make work since GPU selections have different widths than CPU selections, so I moved `gridTemplateColumns`down to lower-level `div`s for more granular control.

src/components/NewRuntimeModal.js
```diff
@@ -751,51 +773,138 @@ export class NewRuntimeModalBase extends Component {
...
-    const renderRuntimeSection = () => {
-      const gridStyle = { display: 'grid', gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', gridGap: '0.8rem', alignItems: 'center' }
+    const renderCloudComputeProfileSection = () => {
+      const { hasGpu, gpuEnabled, gpuType, numGpus } = this.state
+      const { cpu: currentNumCpus, memory: currentMemory } = findMachineType(mainMachineType)
+      const validGpuOptions = getValidGpuTypes(currentNumCpus, currentMemory)
+      const validGpuNames = _.flow(_.map('name'), _.uniq, _.sortBy('price'))(validGpuOptions)
+      const validGpuName = _.includes(displayNameForGpuType(gpuType), validGpuNames) ? displayNameForGpuType(gpuType) : _.head(validGpuNames)
+      const validNumGpusOptions = _.flow(_.filter({ name: validGpuName }), _.map('numGpus'))(validGpuOptions)
+      const validNumGpus = _.includes(numGpus, validNumGpusOptions) ? numGpus : _.head(validNumGpusOptions)
```
kyuksel, line 783: These capture the logic that controls dynamic updating of available GPU type and numbers based on CPU number and memory selections. GPU numbers also depend on GPU type. 

I thought about making helper functions for these but they'd be used only here and thought including them here directly may be more reader-friendly in this case.

src/data/machines.js
```diff
@@ -23,6 +23,32 @@ export const machineTypes = [
...
+  { name: 'NVIDIA Tesla K80', type: 'nvidia-tesla-k80', numGpus: 8, maxNumCpus: 64, maxMem: 208, price: 1.8000, preemptiblePrice: 1.0800 },
+  { name: 'NVIDIA Tesla P4', type: 'nvidia-tesla-p4', numGpus: 1, maxNumCpus: 24, maxMem: 156, price: 0.6000, preemptiblePrice: 0.2160 },
+  { name: 'NVIDIA Tesla P4', type: 'nvidia-tesla-p4', numGpus: 2, maxNumCpus: 48, maxMem: 312, price: 1.2000, preemptiblePrice: 0.4320 },
+  { name: 'NVIDIA Tesla P4', type: 'nvidia-tesla-p4', numGpus: 4, maxNumCpus: 96, maxMem: 624, price: 1.8000, preemptiblePrice: 0.8640 },
+  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 1, maxNumCpus: 12, maxMem: 78, price: 2.4800, preemptiblePrice: 0.7400 },
+  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 2, maxNumCpus: 24, maxMem: 156, price: 4.9600, preemptiblePrice: 1.4800 },
+  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 4, maxNumCpus: 48, maxMem: 312, price: 9.9200, preemptiblePrice: 2.9600 },
+  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 8, maxNumCpus: 96, maxMem: 624, price: 19.8400, preemptiblePrice: 5.9200 }
+]
```
kyuksel, line 50: `name`s are user-facing and `type`s are used internally and by the Leo APIs.

src/components/NewRuntimeModal.js
```diff
@@ -751,51 +770,138 @@ export class NewRuntimeModalBase extends Component {
...
+            !isPersistentDisk ?
+              h(DiskSelector, { value: masterDiskSize, onChange: v => this.setState({ masterDiskSize: v }) }) :
+              div({ style: { gridColumnEnd: 'span 2' } })
+          ]),
+          // GPU Enabling
+          !sparkMode && div({ style: { gridColumnEnd: 'span 6', marginTop: '1.25rem' } }, [
+            h(LabeledCheckbox, {
+              checked: gpuEnabled,
+              onChange: v => this.setState({ gpuEnabled: v || hasGpu })
+            }, [span({ style: { marginLeft: '0.5rem', ...styles.label, verticalAlign: 'top' } }, ['Enable GPUs ', versionTag('Beta', { color: colors.primary(1.5), backgroundColor: 'white', border: `1px solid ${colors.primary(1.5)}` })]),
```
rtitle: Just curious: is there a particular reason this is marked `Beta`? Is there criteria for moving it out of Beta? Just thinking that while there may be evolution of other ML functionality, GPU support specifically is pretty much done.

src/components/NewRuntimeModal.js
```diff
@@ -175,9 +187,8 @@ export class NewRuntimeModalBase extends Component {
       ...(desiredRuntime.cloudService === cloudServices.GCE ? {
         machineType: desiredRuntime.machineType || defaultGceMachineType,
         bootDiskSize: desiredRuntime.bootDiskSize,
-        ...(desiredRuntime.diskSize ? {
-          diskSize: desiredRuntime.diskSize
-        } : {})
+        ...(desiredRuntime.gpuConfig ? { gpuConfig: desiredRuntime.gpuConfig } : {}),
```
dmohs, line 191: Good call moving to single line here.

src/components/NewRuntimeModal.js
```diff
@@ -356,17 +370,20 @@ export class NewRuntimeModalBase extends Component {
...
+    const gpuConfig = runtimeConfig?.gpuConfig
     return {
+      hasGpu,
       runtime: currentRuntimeDetails ? {
         cloudService,
         toolDockerImage: this.getImageUrl(currentRuntimeDetails),
         ...(currentRuntimeDetails?.jupyterUserScriptUri && { jupyterUserScriptUri: currentRuntimeDetails?.jupyterUserScriptUri }),
         ...(cloudService === cloudServices.GCE ? {
           machineType: runtimeConfig.machineType || defaultGceMachineType,
+          ...(hasGpu ? { gpuConfig } : {}),
```
dmohs: Should this be `(hasGpu && gpuConfig ? { gpuConfig } : {})`?

src/components/NewRuntimeModal.js
kyuksel: Good question. I think I'd feel more comfortable with removing `Beta` once we
1. gain more confidence that our revamped container image stack is more proven
2. have a better idea of how much the Google GPU quota issue will affect user experience, and what we can do about it
3. get good feedback from beta testers that it's usable for them

The criteria are admittedly hand-wavy but I think that's okay. 

What do you think?

rtitle: That all sounds good to me 👍 , thanks

rtitle (approved):
I've been running this branch a lot to test https://github.com/DataBiosphere/terra-docker/pull/209. I've tested creating GPU, non-GPU, and Dataproc runtimes; everything seems to work as expected. It also seems like all requirements from [here](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2039906312/2021-06-24+GPU+UI+Requirements) are fulfilled. Nice work!

Note: this should not merge until the above terra-docker PR is merged and released to production.

src/components/NewRuntimeModal.js
```diff
@@ -751,51 +770,138 @@ export class NewRuntimeModalBase extends Component {
...
-              label({ htmlFor: id, style: styles.label }, ['Startup script']),
-              div({ style: { marginTop: '0.5rem' } }, [
-                h(TextInput, {
-                  id,
-                  placeholder: 'URI',
-                  value: jupyterUserScriptUri,
-                  onChange: v => this.setState({ jupyterUserScriptUri: v })
-                })
+        div([
+          div({ style: { ...gridStyle, gridTemplateColumns: `0.25fr 4.5rem 1fr 5.5rem 1fr 5.5rem` } }, [
```
dmohs:
```suggestion
          div({ style: { ...gridStyle, gridTemplateColumns: '0.25fr 4.5rem 1fr 5.5rem 1fr 5.5rem' } }, [
```
I actually never use backticks in my own code because
```js
`Hello ${name}!`
```
saves barely any room over
```js
'Hello '+name+'!'
```

src/components/NewRuntimeModal.js
```diff
@@ -751,51 +770,138 @@ export class NewRuntimeModalBase extends Component {
...
+              onChange: v => this.setState({ gpuEnabled: v || hasGpu })
+            }, [span({ style: { marginLeft: '0.5rem', ...styles.label, verticalAlign: 'top' } }, ['Enable GPUs ', versionTag('Beta', { color: colors.primary(1.5), backgroundColor: 'white', border: `1px solid ${colors.primary(1.5)}` })]),
+              // TODO Update the article link when it's ready
+              h(Link, { style: { marginLeft: '1rem', verticalAlign: 'top' }, href: 'https://support.terra.bio/hc/en-us/articles/', ...Utils.newTabLinkProps }, [
+                'Learn more about GPU cost and restrictions.',
+                icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
+              ])])
+          ]),
+          // GPU Selection
+          gpuEnabled && !sparkMode && div({ style: { ...gridStyle, gridTemplateColumns: `0.75fr 12rem 1fr 5.5rem 1fr 5.5rem`, marginTop: '0.75rem' } }, [
```
dmohs:
```suggestion
          gpuEnabled && !sparkMode && div({ style: { ...gridStyle, gridTemplateColumns: '0.75fr 12rem 1fr 5.5rem 1fr 5.5rem', marginTop: '0.75rem' } }, [
```

dmohs (approved):
Functional approach is really nice across the PR.

src/components/NewRuntimeModal.js
zarsky-broad: We do backticks over string concat as a standard in this codebase, especially since we put spaces around operators. That said, this doesn't have any interpolation, so it should indeed be single-quoted.

src/libs/runtime-utils.js
```diff
@@ -211,6 +226,16 @@ export const convertedAppStatus = appStatus => {
   )
 }
 
+export const displayNameForGpuType = type => {
+  return Utils.switchCase(type,
+    ['nvidia-tesla-t4', () => 'NVIDIA Tesla T4'],
```
zarsky-broad: Do we need this case if we have the default?

src/data/machines.js
zarsky-broad: Do we need user facing names here if we have `displayNameForGpuType`?

src/components/NewRuntimeModal.js
```diff
@@ -751,51 +770,138 @@ export class NewRuntimeModalBase extends Component {
...
+                ])
               ])
             ])
           ])
         ]),
         sparkMode === 'cluster' && fieldset({ style: { margin: '1.5rem 0 0', border: 'none', padding: 0 } }, [
           legend({ style: { padding: 0, ...styles.label } }, ['Worker config']),
           // grid styling in a div because of display issues in chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=375693
-          div({ style: { ...gridStyle, marginTop: '0.75rem' } }, [
+          div({ style: { ...gridStyle, gridTemplateColumns: `0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem`, marginTop: '0.75rem' } }, [
```
zarsky-broad:
```suggestion
          div({ style: { ...gridStyle, gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', marginTop: '0.75rem' } }, [
```

src/components/NewRuntimeModal.js
```diff
@@ -751,51 +770,138 @@ export class NewRuntimeModalBase extends Component {
...
-                  isSearchable: false,
-                  value: sparkMode,
-                  onChange: ({ value }) => this.setState({ sparkMode: value }),
-                  options: [
-                    { value: false, label: 'Standard VM', isDisabled: requiresSpark },
-                    { value: 'master', label: 'Spark master node' },
-                    { value: 'cluster', label: 'Spark cluster' }
-                  ]
-                })
+          div({ style: { ...gridStyle } }, [
```
zarsky-broad:
```suggestion
          div({ style: gridStyle }, [
```

src/components/NewRuntimeModal.js
zarsky-broad: This is an extremely long line, could you break it up? As a guideline, try to avoid lines much over ~120 characters.

src/data/machines.js
kyuksel: We don't need it per se but
- `renderCloudComputeProfileSection()` seems to be called multiple times upon state updates and I wanted to err on the side of optimizing for performance vs. memory assuming a dictionary look-up is a lot quicker than branching through conditionals.
- It also provides a bit more self-documentation I'd say 😊

kyuksel (commented):
@zarsky-broad @dmohs @rtitle Thank you very much for the feedback and the testing 🙏🏼 
I've pushed changes to address your comments (or replied to them). A non-white-space diff since your review is [here](https://github.com/DataBiosphere/terra-ui/pull/2554/files/69ede2fb447d47ac01ec5a5904b0f2ecd9ea8515..12dfc0cf7aa65adc0e44bf4bbcb4e68e006f9bfb?w=1) if it helps. The changes also include the following improvements:

- Disabling the `Enable GPUs` checkbox when there is a GCE compute without GPUs, and show a tooltip: This was favoured over not displaying the checkbox at all for UX purposes since the latter can be a bit confusing.
- Fixing a spacing issue where Spark cluster disk size box wasn't fitting into the section.
- Adjusting `margin` and `padding` per the specification received from the UX team below:

gpu2

src/data/machines.js
zarsky-broad: I'd say that's a serious problem that's separate from this, then: a render function should only be called once per update.

kyuksel: I don't know React well enough to comment on the rendering piece.

If you feel strongly about removing `name`, I can do that.

zarsky-broad: We can leave it, I just want to deduplicate the calls to the render method.

kyuksel: I may work on converting `NewRuntimeModal` to be hook-based as part of the jamboree, during which time I may be able to investigate it further.

rtitle (approved):
Retested basic GPU functionality, it looks good to me. 

Btw to summarize in-person conversations: we're targeting merge/releasing this PR along with Leo next week to release GPUs in Terra.

kyuksel:
[GPU Release Checklist](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2071789569/2021-07-01+GPU+Release+Checklist) for reference

PR #2559, opened 2021-07-08
Add UI control for retrying a task with more memory, BW-290.

cahrens:
This exposes the "retry with more memory" feature for Workflows. When a task has maxRetries > 1, this option allows the machine memory to be increased on each retry, if the task fails only for memory reasons.

I have tested this against dev, using this WDL: https://github.com/broadinstitute/cromwell/blob/develop/centaur/src/main/resources/standardTestCases/retry_with_more_memory/retry_with_more_memory.wdl

For details of how to confirm that the machine size was increased on each retry, see https://broadworkbench.atlassian.net/browse/BW-749.

### DIFF
--- a/src/pages/workspaces/workspace/workflows/FailureRerunner.js
+++ b/src/pages/workspaces/workspace/workflows/FailureRerunner.js
@@ -27,7 +27,7 @@ export const rerunFailures = async ({ workspace, workspace: { workspace: { names
   const eventData = { ...extractWorkspaceDetails(workspace), configNamespace, configName }
 
   try {
-    const [{ workflows, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks }, config] = await Promise.all([
+    const [{ workflows, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks, memoryRetryMultiplier }, config] = await Promise.all([
       Ajax().Workspaces.workspace(namespace, name).submission(submissionId).get(),
       Ajax().Workspaces.workspace(namespace, name).methodConfig(configNamespace, configName).get()
     ])
@@ -40,7 +40,7 @@ export const rerunFailures = async ({ workspace, workspace: { workspace: { names
         _.map('workflowEntity.entityName')
       )(workflows),
       newSetName: Utils.sanitizeEntityName(`${configName}-resubmission-${new Date().toISOString().slice(0, -5)}`),
-      useCallCache, deleteIntermediateOutputFiles, useReferenceDisks,
+      useCallCache, deleteIntermediateOutputFiles, useReferenceDisks, memoryRetryMultiplier,
       onProgress: stage => {
         rerunFailuresStatus.set({ text: { createSet: 'Creating set from failures...', launch: 'Launching new job...', checkBucketAccess: 'Checking bucket access...' }[stage] })
       }
--- a/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
+++ b/src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
@@ -19,7 +19,8 @@ const LaunchAnalysisModal = ({
   onDismiss, entityMetadata,
   workspace, workspace: { workspace: { namespace, name: workspaceName, bucketName } }, processSingle,
   entitySelectionModel: { type, selectedEntities, newSetName },
-  config, config: { rootEntityType }, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks, onSuccess
+  config, config: { rootEntityType }, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks,
+  retryWithMoreMemory, retryMemoryFactor, onSuccess
 }) => {
   const [launching, setLaunching] = useState(undefined)
   const [message, setMessage] = useState(undefined)
@@ -53,6 +54,7 @@ const LaunchAnalysisModal = ({
       const { submissionId } = await launch({
         isSnapshot: type === processSnapshotTable,
         workspace, config, selectedEntityType, selectedEntityNames, newSetName, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks,
+        memoryRetryMultiplier: retryWithMoreMemory ? retryMemoryFactor : undefined,
         onProgress: stage => {
           setMessage({ createSet: 'Creating set...', launch: 'Launching analysis...', checkBucketAccess: 'Checking bucket access...' }[stage])
         }
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -8,7 +8,7 @@ import {
 } from 'src/components/common'
 import Dropzone from 'src/components/Dropzone'
 import { centeredSpinner, icon } from 'src/components/icons'
-import { DelayedAutocompleteTextArea, DelayedSearchInput } from 'src/components/input'
+import { DelayedAutocompleteTextArea, DelayedSearchInput, NumberInput } from 'src/components/input'
 import { MarkdownViewer } from 'src/components/markdown'
 import Modal from 'src/components/Modal'
 import { InfoBox, makeMenuIcon, MenuButton, MenuTrigger } from 'src/components/PopupTrigger'
@@ -316,6 +316,8 @@ const WorkflowView = _.flow(
       useCallCache: true,
       deleteIntermediateOutputFiles: false,
       useReferenceDisks: false,
+      retryWithMoreMemory: false,
+      retryMemoryFactor: 1.2,
       includeOptionalInputs: true,
       filter: '',
       errors: { inputs: {}, outputs: {} },
@@ -350,6 +352,15 @@ const WorkflowView = _.flow(
     })
   }
 
+  /** Returns the href link for the specified article. */
+  getSupportLink(article) { return `https://support.terra.bio/hc/en-us/articles/${article}` }
+
+  /** Returns true if the increasing memory on retries is enabled */
+  isMemoryIncreasedOnRetries() { return this.state.retryWithMoreMemory }
+
+  /** Returns true if the memory retry factor is outside the expected range */
+  warnMemoryFactor() { return this.state.retryMemoryFactor > 2 }
+
   render() {
     // isFreshData: controls spinnerOverlay on initial load
     // variableSelected: field of focus for bucket file browser
@@ -357,7 +368,8 @@ const WorkflowView = _.flow(
     // modifiedConfig: active data, potentially unsaved
     const {
       isFreshData, savedConfig, launching, activeTab, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks,
-      entitySelectionModel, variableSelected, modifiedConfig, updatingConfig, selectedSnapshotEntityMetadata, availableSnapshots
+      retryWithMoreMemory, retryMemoryFactor, entitySelectionModel, variableSelected, modifiedConfig, updatingConfig,
+      selectedSnapshotEntityMetadata, availableSnapshots
     } = this.state
     const { namespace, name, workspace } = this.props
     const workspaceId = { namespace, name }
@@ -373,7 +385,8 @@ const WorkflowView = _.flow(
         launching && h(LaunchAnalysisModal, {
           workspace, config: savedConfig, entityMetadata: selectedSnapshotEntityMetadata,
           accessLevel: workspace.accessLevel, bucketName: workspace.workspace.bucketName,
-          processSingle: this.isSingle(), entitySelectionModel, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks,
+          processSingle: this.isSingle(), entitySelectionModel, useCallCache, deleteIntermediateOutputFiles,
+          useReferenceDisks, retryWithMoreMemory, retryMemoryFactor,
           onDismiss: () => this.setState({ launching: false }),
           onSuccess: submissionId => {
             const { methodRepoMethod: { methodVersion, methodNamespace, methodName, methodPath, sourceRepo } } = modifiedConfig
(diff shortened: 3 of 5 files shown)


src/pages/workspaces/workspace/workflows/FailureRerunner.js
```diff
@@ -27,7 +27,7 @@ export const rerunFailures = async ({ workspace, workspace: { workspace: { names
   const eventData = { ...extractWorkspaceDetails(workspace), configNamespace, configName }
 
   try {
-    const [{ workflows, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks }, config] = await Promise.all([
+    const [{ workflows, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks, memoryRetryMultiplier }, config] = await Promise.all([
```
cahrens, line 30: To test that this works, I had to hack the JS to show the "Relaunch Failures" Action on the Job History page. It is not enabled for the test WDL I am running.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -53,6 +54,7 @@ const LaunchAnalysisModal = ({
       const { submissionId } = await launch({
         isSnapshot: type === processSnapshotTable,
         workspace, config, selectedEntityType, selectedEntityNames, newSetName, useCallCache, deleteIntermediateOutputFiles, useReferenceDisks,
+        memoryRetryMultiplier: retryWithMoreMemory ? retryMemoryFactor : undefined,
```
cahrens, line 57: The back end code expects just memoryRetryMultiplier. The default in the back end code is 1.0, which means no increase in memory. In the front end, the minimum allowable value is 1.1, so as long as the checkbox was selected we should send the user-specified value.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -762,39 +778,86 @@ const WorkflowView = _.flow(
...
-            ]),
-            h(InfoBox, [
-              'If the workflow succeeds, only the final output will be saved. Subsequently, call caching cannot be used as the intermediate steps will be not available. ',
-              h(Link, {
-                href: 'https://support.terra.bio/hc/en-us/articles/360039681632',
-                ...Utils.newTabLinkProps
-              }, ['Click here to learn more.'])
-            ]),
-            span({ style: { margin: '0 0.5rem 0 1rem' } }, [
+          div([
```
cahrens: I recommend hiding whitespace changes for this part of the review.

cahrens:
@petesantos can you explain how the ci/circleci build check works? It has been in the same state since I created the merge request yesterday.

salonishah11:
@cahrens is it intentional that when the screen size is decreased, the `Memory retry factor` input box is wrapped onto new line while the `Retry with more memory` checkbox remains in the same line? 
 
Screen Shot 2021-07-09 at 3 43 27 PM

cahrens:
@salonishah11 it is not intentional, but I did not attempt to prevent it because it does not appear to me that the UI makes an effort to be responsive at smaller screen sizes. The layout works at what I was told to be our minimum screen size (1280 width). As you know, I initially put the new field below the existing ones to keep the line of controls shorter, but we decided it looked better to put them all on the same line.

Also, part of the challenge here is that in order to prevent the row of controls from resizing vertically when the numeric input is displayed, I had to put the numeric input and the associated checkbox in separate divs.

salonishah11:
@cahrens Ah I see. How hard would it be to have both the checkbox and the input field stay on the same line when we decrease the screen size? I am asking because I feel it would be cleaner to have them stay on the same line even if the screen size is decreased. That being said, if the UX is happy with the input field wrapping to the next line, then I don't mind.

cahrens:
@salonishah11 I am not sure how hard it would be. But I would like to get @petesantos's opinion on the importance before I spend the time on it, especially given that the existing components on the page are not responsive and simply get cut off for narrow screen sizes. Making the controls wrap nicely will require changing the existing layout.

cahrens:
@salonishah11 I experimented with changing the div/span nesting so that the checkbox and the input field move together to the next line. However, I am finding that is very jarring/unexpected for the user because the checkbox moves from one line to the next based on whether it is checked (when the browser is narrow enough that they don't both fit on the same line). I think that if we are really concerned about the length of that line of options, it would be better to always put the new checkbox/input on its own line so that things do not move around as much.

@petesantos can you offer your opinion when get a chance, and also let me know what the minimum browser size is that we design for?

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -596,6 +609,9 @@ const WorkflowView = _.flow(
     const inputsValid = _.isEmpty(errors.inputs)
     const outputsValid = _.isEmpty(errors.outputs)
     const sourceDisplay = sourceRepo === 'agora' ? `${methodNamespace}/${methodName}/${methodVersion}` : `${methodPath}:${methodVersion}`
+    const checkBoxSpanMargins = { margin: '0 0.5rem 0 1rem' }
+    const checkBoxLeftMargin = { marginLeft: '1rem' }
```
petesantos: nit: could you move these styles to the styles object on line 51?

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -762,39 +778,86 @@ const WorkflowView = _.flow(
...
+                  'If a task has a maxRetries value greater than zero and fails because it ran out of memory, retry it with more memory. ',
+                  h(Link, { href: this.getSupportLink('4403215299355'), ...Utils.newTabLinkProps },
+                    [clickToLearnMore])
+                ]),
+              this.isMemoryIncreasedOnRetries() && this.warnMemoryFactor() && h(InfoBox,
+                { style: { color: colors.warning() }, iconOverride: 'warning-standard' }, [
+                  'Retry factors above 2 are not recommended. The retry factor compounds and may substantially increase costs. ',
+                  h(Link, { href: this.getSupportLink('4403215299355'), ...Utils.newTabLinkProps },
+                    [clickToLearnMore])
+                ])
```
petesantos: How do you feel about a ternary as a minor simplification? prefer this as I find it a bit easier to read.

```suggestion
              (retryWithMoreMemory && retryMemoryFactor > 2 ?
                h(InfoBox,
                  { style: { color: colors.warning() }, iconOverride: 'warning-standard' }, [
                    'Retry factors above 2 are not recommended. The retry factor compounds and may substantially increase costs. ',
                    h(Link, { href: this.getSupportLink('4403215299355'), ...Utils.newTabLinkProps },
                      [clickToLearnMore])
                  ]) :
                h(InfoBox,
                  [
                    'If a task has a maxRetries value greater than zero and fails because it ran out of memory, retry it with more memory. ',
                    h(Link, { href: this.getSupportLink('4403215299355'), ...Utils.newTabLinkProps },
                      [clickToLearnMore])
                  ])
              )
```

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -350,14 +352,24 @@ const WorkflowView = _.flow(
...
   }
 
+  /** Returns the href link for the specified article. */
+  getSupportLink(article) { return `https://support.terra.bio/hc/en-us/articles/${article}` }
+
+  /** Returns true if the increasing memory on retries is enabled */
+  isMemoryIncreasedOnRetries() { return this.state.retryWithMoreMemory }
+
+  /** Returns true if the memory retry factor is outside the expected range */
+  warnMemoryFactor() { return this.state.retryMemoryFactor > 2 }
```
petesantos: nit: With these two functions I prefer directly using the state it is more straightforward to me when I read the code - I don't have to bounce back to the function to see what it is doing.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -350,14 +352,24 @@ const WorkflowView = _.flow(
     })
   }
 
+  /** Returns the href link for the specified article. */
+  getSupportLink(article) { return `https://support.terra.bio/hc/en-us/articles/${article}` }
```
petesantos, line 362: TOL: I like this function - looking through the code this could be generally useful as we use the article numbers a lot. It would be nice to create a mapping of numbers to titles and use a function like this - something like: 

```getSupportLink(article.OOM_RETRY)```

**Review by petesantos — APPROVED** [review]
Some notes on terms: 
"nit" - nitpick / minor issue
"TOL" - thinking out loud, no need to take any action, unless you want to follow up

Looks good! I tested locally and saw the memory increase. A few minor comments.

File: src/pages/workspaces/workspace/workflows/WorkflowView.js (hunk 4 above)
Comment (cahrens) [reply src/pages/workspaces/workspace/workflows/WorkflowView.js]: Thanks for pointing me to the styles object, done.

File: src/pages/workspaces/workspace/workflows/WorkflowView.js (hunk 5 above)
Comment (cahrens) [reply src/pages/workspaces/workspace/workflows/WorkflowView.js]: I agree it is simpler, done.

File: src/pages/workspaces/workspace/workflows/WorkflowView.js (hunk 6 above)
Comment (cahrens) [reply src/pages/workspaces/workspace/workflows/WorkflowView.js]: Done.

File: src/pages/workspaces/workspace/workflows/WorkflowView.js (hunk 7 above)
Comment (cahrens) [reply src/pages/workspaces/workspace/workflows/WorkflowView.js:362]: Would you suggest then that `getSupportLink` should be in a different file so that it can be reused across other pages? If so, can you point me to an appropriate file?

If `getSupportLink` moves, I'm not sure if it would make sense to put the mapping of titles to numbers in the same location, or keep mappings in the individual page files so they are more closely related to the code referencing them.

File: src/pages/workspaces/workspace/workflows/WorkflowView.js [hunk 8]
```diff
@@ -762,39 +776,87 @@ const WorkflowView = _.flow(
...
-                href: 'https://support.terra.bio/hc/en-us/articles/360039681632',
-                ...Utils.newTabLinkProps
-              }, ['Click here to learn more.'])
-            ]),
-            span({ style: { margin: '0 0.5rem 0 1rem' } }, [
+          div([
+            // Because the retries numeric input is not always visible, and will be taller than the other components
+            // on this line, wrap the components that are always visible with a div and add top/bottom margins to
+            // avoid vertical resizing when the numeric input is rendered.
+            div({ style: { display: 'inline-block', marginTop: '0.5rem', marginBottom: '0.5rem' } }, [
```
petesantos: To avoid wrapping and the height change I think we can do:
```suggestion
div({ style: { display: 'flex', alignItems: 'center', minHeight: '2.5rem', minWidth: 'max-content', margin: '0.5rem 0' } }, [
```

With this you could remove the `inline-block` styling and push the conditional render into this div block

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -762,39 +776,87 @@ const WorkflowView = _.flow(
...
+                  { style: { color: colors.warning() }, iconOverride: 'warning-standard' }, [
+                    'Retry factors above 2 are not recommended. The retry factor compounds and may substantially increase costs. ',
+                    h(Link, { href: this.getSupportLink('4403215299355'), ...Utils.newTabLinkProps },
+                      [clickToLearnMore])
+                  ]) :
+                h(InfoBox, [
+                  'If a task has a maxRetries value greater than zero and fails because it ran out of memory, retry it with more memory. ',
+                  h(Link, { href: this.getSupportLink('4403215299355'), ...Utils.newTabLinkProps },
+                    [clickToLearnMore])
+                ])
```
petesantos: If you use the above suggestion this block can be moved to the above flex div. You should be able to remove the inline-block styling as well.

petesantos:
@salonishah11 @cahrens 
re: The width of the options line.

I would prefer to see the line not wrap so it is consistent with the rest of the page. While having a responsive design would be a nice feature that our users could benefit from, it is not something we have designed this page for at the moment. I confirmed with Lou that the min width for the app is 1200 - I will add it to the repo docs. The options will fit within this width.

I added some suggestions to the PR to make the retry fields not wrap.

Thanks for bringing these things up - it is nice to get fresh eyes on the app!

src/pages/workspaces/workspace/workflows/WorkflowView.js
petesantos: I added a [ticket ](https://broadworkbench.atlassian.net/secure/RapidBoard.jspa?rapidView=110&projectKey=TUB&modal=detail&selectedIssue=TUB-4) for this improvement linking to this line / diff. 

You can leave it here for now.

src/pages/workspaces/workspace/workflows/WorkflowView.js
cahrens: @petesantos I can't figure out a way to use 'center' for alignItems and keep the checkbox text aligned properly vertically (and in comparison to the input label). I pushed up a change for you to consider that puts all the checkboxes within a span with top/bottom margin spacing so that I can use 'baseline' for alignItems.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -762,40 +776,86 @@ const WorkflowView = _.flow(
...
-                style: { marginLeft: '1rem' }
-              }, [' Delete intermediate outputs'])
-            ]),
-            h(InfoBox, [
-              'If the workflow succeeds, only the final output will be saved. Subsequently, call caching cannot be used as the intermediate steps will be not available. ',
-              h(Link, {
-                href: 'https://support.terra.bio/hc/en-us/articles/360039681632',
-                ...Utils.newTabLinkProps
-              }, ['Click here to learn more.'])
+          div({ style: { display: 'flex', alignItems: 'baseline', minWidth: 'max-content' } }, [
```
petesantos, line 779: The changes for the wrapping fix LGTM

petesantos (approved):
Checked wrapping changes, lgtm.

salonishah11 (approved):
LGTM 👍

PR #2560, opened 2021-07-13
Jamboree new showcase [BT-350]

tcjiang9:
Complete redesign of the Showcase page.
Goals:
* Make featured workspace list and full workspace descriptions viewable when not signed in
* Customer Success should continue to control which workspaces are featured and how they're categorized
* Improve the browsing/filtering/searching experience


breilly2:
@tcjiang9 & @jyang-broad, could you review the finished product? I've done a bit of work since jamboree week and I think this is ready to go now. We need one Terra UI approver (Isaac, Pete, or Mohs) and at least one other. Since you helped work on this and are already familiar with the project, it makes sense for you to be the other approvers. Thanks!

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,332 @@ import * as Style from 'src/libs/style'
...
     marginBottom: '1rem'
+  },
+  sidebarRow: {
+    display: 'flex', justifyContent: 'space-between', alignItems: 'baseline'
+  },
+  nav: {
+    background: {
+      position: 'absolute', left: 0, right: 0, top: 0, bottom: 0,
+      overflow: 'auto', cursor: 'pointer',
+      zIndex: 2
```
dmohs: I will usually try to avoid `zIndex` by moving things around in the DOM as needed, so it may be worth having a convention where we comment on usage. In this case, I can't see where this is used, so I can't infer the need for the `zIndex` here.

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
+        _.map(section => ({ ...section, tags: _.map(_.toLower, section.labels) })),
+        _.remove(section => _.isEmpty(section.labels))
+      ])(sidebarSections)
+
+      setFeaturedList(featuredWorkspaces)
+      setWorkspacesByTag(workspacesByTag)
+      setSections(activeSections)
+
+      // TODO: is this actually doing anything for us?
+      StateHistory.update({ featuredWorkspaces })
```
jyang-broad, line 255: are we leaving this TODO here for now?

jyang-broad (approved):
looks good! There's one trailing TODO not sure if that's a later'd task or just forgotten.

src/pages/library/Showcase.js
breilly2: Hmm... yeah, I'd like to understand this better than I do. I believe the intent of this is a performance optimization, but I don't think it's entirely working. In addition, since there are a few calculations that are happening from this data, we'd either need to include those in state history, change the `useOnMount` logic to continue doing the calculations based on the cached data from state history, or move the calculations into `_.once` or `_.memoize` functions... or some other solution I'm not thinking of.

For now, I'm out of time for working on this project, so unfortunately I'm going to have to leave this TODO alone.

tcjiang9:
> @tcjiang9 & @jyang-broad, could you review the finished product? I've done a bit of work since jamboree week and I think this is ready to go now. We need one Terra UI approver (Isaac, Pete, or Mohs) and at least one other. Since you helped work on this and are already familiar with the project, it makes sense for you to be the other approvers. Thanks!

Lgtm, but I can't approve since I created the PR

breilly2 (approved):
Using my approval for @tcjiang9 review:

> Lgtm, but I can't approve since I created the PR

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
+
+  const unionSectionWorkspaces = section => {
+    return _.uniq(_.flatMap(tag => workspacesByTag[tag], section.tags))
+  }
+
+  return div({ style: { display: 'flex', flexDirection: 'column' } }, [
+    _.map(section => {
+      return section.keepCollapsed ?
+        h(Clickable, {
+          key: section.name,
```
dmohs, line 173: Since the `key` property is usually reserved for when React is misbehaving, I wonder if we should have a convention around adding a comment about why it is needed when we use it?

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
-        div({ style: { display: 'flex' } }, [
-          !_.isEmpty(newAndInteresting) && div({ style: styles.column }, [
-            div({ style: styles.header }, 'New and interesting'),
-            ..._.map(makeCard('new'), newAndInteresting)
+    libraryTopMatter('featured workspaces'),
+    !featuredList ?
+      centeredSpinner() :
+      h(Fragment, [
+        div({ style: { display: 'flex', margin: '1rem 1rem 0', alignItems: 'baseline' } }, [
+          div({ style: { width: '19rem', flex: '0 0 auto' } }, [
```
dmohs: Since 'none' is a synonym for '0 0 auto', and the former is used above, I wonder if it makes sense to always use one or the other?

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
+              sections,
+              selectedSections,
+              selectedTags,
+              workspacesByTag: groupByFeaturedTags(filteredWorkspaces)
+            })
           ]),
-          div({ style: { ...styles.column, marginRight: undefined } }, [
-            div({ style: styles.header }, 'GATK4 example workspaces'),
-            ..._.map(makeCard('gatk'), bestPractices)
+          div({ style: { marginLeft: '1rem', minWidth: 0 } }, [
```
dmohs, line 340: Wouldn't a `minWidth` of 0 be implied?

src/pages/workspaces/workspace/DashboardPublic.js
```diff
@@ -0,0 +1,72 @@
...
+  marginTop: '2rem',
+  display: 'flex',
+  flexDirection: 'column',
+  borderRadius: 5,
+  textAlign: 'center',
+  alignItems: 'center',
+  justifyContent: 'space-between',
+  fontSize: 20
+}
+//render
```
dmohs:
```suggestion
```

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
 
 
+// Description of the structure of the sidebar. Case is preserved when rendering but all matching is case-insensitive.
+const sidebarSections = [
+  {
+    name: 'Getting Started',
+    labels: ['Workflow Tutorials', 'Notebook Tutorials', 'Data Tutorials', 'RStudio Tutorials', 'Galaxy Tutorials'],
+    keepCollapsed: true
+  },
+  {
```
dmohs: Would this increase concision without harming readability?
```suggestion
  }, {
```

dmohs (approved):
Looks great. Most of my comments are in the interest of learning.

src/pages/library/Showcase.js
zarsky-broad: I think we should pretty reliably expect it with any `_.map`.

zarsky-broad (changes requested):
Please add a Route using the normal system.

src/components/SignInButton.js
```diff
@@ -3,14 +3,14 @@ import { Clickable } from 'src/components/common'
 import * as Utils from 'src/libs/utils'
 
 
-const SignInButton = (props = {}) => {
+const SignInButton = (props = { theme: 'light' }) => {
```
zarsky-broad:
```suggestion
const SignInButton = ({ theme = 'light', ...props } = {}) => {
```

src/components/SignInButton.js
```diff
@@ -3,14 +3,14 @@ import { Clickable } from 'src/components/common'
...
   Utils.useOnMount(() => {
     window.gapi.signin2.render('signInButton', {
       scope: 'openid profile email',
       width: 250,
       height: 56,
       longtitle: true,
-      theme: 'light',
+      theme: props.theme,
```
zarsky-broad:
```suggestion
      theme,
```

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
+      display: 'flex', flexDirection: 'column'
+    }),
+    navSection: {
+      alignItems: 'center', flex: 'none', padding: '1.2rem 0', fontWeight: 700,
+      borderTop: `1px solid ${colors.dark(0.35)}`
+    }
+  },
+  pill: {
+    width: '4.5rem', padding: '0.25rem', fontWeight: 500, textAlign: 'center',
+    border: '1px solid', borderColor: colors.dark(0.25), borderRadius: '1rem',
```
zarsky-broad, line 82:
```suggestion
    border: `1px solid ${colors.dark(0.25)}`, borderRadius: '1rem',
```

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
+    backgroundColor: 'white'
   }
 }
+styles.hilightedPill = { ...styles.pill, color: 'white', backgroundColor: colors.primary(1), borderColor: colors.primary(1) }
```
zarsky-broad: To avoid mutating styles, could this just be `styles.pillHighlight`, and then just be conditionally merged in over the base pill style when needed?

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
+  ])
+}
+
+const groupByFeaturedTags = workspaces => _.flow([
+  _.map(tag => [tag, _.filter(w => _.includes(tag, w.tags.items), workspaces)]),
+  _.fromPairs
+])(uniqueSidebarTags)
+
+const Sidebar = props => {
+  const { onSectionFilter, onTagFilter, sections, selectedSections, selectedTags, workspacesByTag } = props
```
zarsky-broad:
```suggestion
const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSections, selectedTags, workspacesByTag }) => {
```

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
+          div({
+            style: _.includes(section, selectedSections) ? styles.hilightedPill : styles.pill
+          }, [_.size(unionSectionWorkspaces(section))])
+        ]) :
+        h(SidebarCollapser,
+          {
+            key: section.name,
+            title: section.name,
+            onClick: () => setOpenState(_.update(section.name, open => !open, openState)),
+            isOpened: openState[section.name]
```
zarsky-broad: If we do this condition as `!(openState[section.name] === false)`, we won't have to initialize the object above.

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
+  _.map(tag => [tag, _.filter(w => _.includes(tag, w.tags.items), workspaces)]),
+  _.fromPairs
+])(uniqueSidebarTags)
+
+const Sidebar = props => {
+  const { onSectionFilter, onTagFilter, sections, selectedSections, selectedTags, workspacesByTag } = props
+
+  // setup open-ness state for each sidebar section
+  const initialOpenState = _.fromPairs(_.map(section => [section.name, true], sidebarSections))
+  const [openState, setOpenState] = useState(initialOpenState)
```
zarsky-broad: If you make the change below, this can default to `{}`

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,337 @@ import * as Style from 'src/libs/style'
...
+  return div({ style: { display: 'flex', flexDirection: 'column' } }, [
+    _.map(section => {
+      return section.keepCollapsed ?
+        h(Clickable, {
+          key: section.name,
+          onClick: () => onSectionFilter(section),
+          style: { ...styles.sidebarRow, ...styles.nav.navSection }
+        }, [
+          div({ style: { flex: 1 } }, [section.name]),
+          div({ style: { flexGrow: 1 } }),
```
zarsky-broad: Rather than a spacer here, consider changing the alignment of the content of the name div.

zarsky-broad:
> Please add a Route using the normal system.

Specifically, I think the thing to do might be to make the dashboard page public and merging this logic in there.

src/pages/library/Showcase.js
breilly2: Good catch. I think this was copied over from `TopBar`. I doubt it's needed here so I will remove it if doing so doesn't break anything.

src/pages/library/Showcase.js
breilly2: React also complains very loudly on the console if you omit `key`.

src/pages/library/Showcase.js
breilly2: Without this, the workspace cards blow out past the right edge of the screen. I fiddled with this quite a bit to find something that worked, but I'm open to other solutions.

src/pages/library/Showcase.js
dmohs: Yes, I remember now. I actually had a utility to apply children (effectively extracting them from a list) just to avoid this warning. Please do disregard.

src/pages/library/Showcase.js
breilly2: Would you then put the first open and last close on the same lines as the opes and close square brackets as well? Feels unbalanced otherwise, but I'm not sure I like the end result:
```
const sidebarSections = [{
  name: 'Getting Started',
  labels: ['Workflow Tutorials', 'Notebook Tutorials', 'Data Tutorials', 'RStudio Tutorials', 'Galaxy Tutorials'],
  keepCollapsed: true
}, {
  ...
}, {
  name: 'Projects',
  labels: ['HCA', 'AnVIL', 'BRAIN Initiative', 'BioData Catalyst', 'NCI']
}]

File: src/components/SignInButton.js (hunk 8 above)
Comment (breilly2) [reply src/components/SignInButton.js]: This was following Jerome's mocks: https://www.sketch.com/s/a4f14c9b-e39f-48e3-9f2c-8d69c3cd8bbb/a/8yOvy2D
Screen Shot 2021-08-04 at 5 33 10 PM
My guess was that he thought it wasn't prominent enough with the light theme. I can ask him about it.

File: src/pages/library/Showcase.js (hunk 11 above)
Comment (breilly2) [reply src/pages/library/Showcase.js]: Ah, good call. I didn't love this. I just I didn't remember the technique you suggested.

File: src/pages/library/Showcase.js (hunk 10 above)
Comment (breilly2) [reply src/pages/library/Showcase.js:82]: Actually, this doesn't work along with the idea of merging in the highlight style because the highlight only changes the border color. I could repeat the `1px solid` part in the highlight style, but I'd rather keep the conditionally overridden border color separate instead.

File: src/pages/library/Showcase.js (hunk 13 above)
Comment (breilly2) [reply src/pages/library/Showcase.js]: I don't love the readability of `!(... === false)`, but I can get rid of the special initialization if I reverse it to keep track of collapsed sections instead of open sections. Also, I can turn it into an array and use `_.xor` to toggle the state, just like the filter selections. I prefer thinking of it like that instead of having a sparse object where keys we're looking up might be `undefined` in addition to `false` or `true`.

File: src/pages/library/Showcase.js (hunk 15 above)
Comment (breilly2) [reply src/pages/library/Showcase.js]: Looks like we don't even need the spacer anymore. That must have been left over from a previous layout attempt.

**Comment by breilly2** [comment]
> Specifically, I think the thing to do might be to make the dashboard page public and merging this logic in there.
@zarsky-broad Makes sense. As it turns out, there's also a problem with this currently... viewing a dashboard for a non-featured workspace when not signed in. The dashboard will still need auth-container-like functionality, but you're right that the dashboard page is a better home for it (especially now that it needs to be a little more complex).

File: src/components/SignInButton.js (hunk 8 above)
Comment (breilly2) [reply src/components/SignInButton.js]: Oh, I totally didn't read your suggestion closely enough. Sorry. 🤦

**Comment by breilly2** [comment]
Updated based on feedback. @zarsky-broad , @dmohs , please re-review.

To test the cases of viewing a workspace dashboard, I used http://localhost:3000/#workspaces/help-gatk/GATK4-RNA-Germline-VariantCalling as a featured workspace and http://localhost:3000/#workspaces/broad-dsde-alpha/2000_samples as a non-featured workspace.

When signed in:
* the non-featured workspace dashboard is rendered with all details
* the featured workspace dashboard _would_ be rendered but it doesn't exist in dev, so instead you get the "Could not display workspace" message

When _not_ signed in:
* the featured workspace shows the full description, even though it doesn't exist in dev, because it's included in the `showcase.json` file
* the non-featured workspace shows the sign-in page like it always has

There's a bit more changed than you might have expected due to I discovered a bug with that last case where it would try to render a non-featured workspace as if it were a featured workspace despite not having the description from `showcase.json`.

File: src/pages/library/Showcase.js [hunk 16]
```diff
@@ -16,104 +18,333 @@ import * as Style from 'src/libs/style'
...
+          })
+        ]) :
+        h(SidebarCollapser,
+          {
+            key: section.name,
+            title: section.name,
+            onClick: () => setCollapsedSections(_.xor([section], collapsedSections)),
+            isOpened: !_.includes(section, collapsedSections)
+          },
+          [
```
dmohs:
```suggestion
          }, [
```

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,333 @@ import * as Style from 'src/libs/style'
...
+          style: { ...styles.sidebarRow, ...styles.nav.navSection }
+        }, [
+          div({ style: { flex: 1 } }, [section.name]),
+          h(Pill, {
+            count: _.size(unionSectionWorkspaces(section)),
+            highlight: _.includes(section, selectedSections)
+          })
+        ]) :
+        h(SidebarCollapser,
+          {
```
dmohs:
```suggestion
        h(SidebarCollapser, {
```
(unless there is a reason for these choices that I am not seeing)

src/pages/library/Showcase.js
dmohs: I suspected there may have been more thought invested into this than is obvious from the number alone. It might be worth commenting as such.

src/pages/workspaces/workspace/DashboardPublic.js
```diff
@@ -0,0 +1,70 @@
...
+          `${namespace}/${name}`
+        ])
+      ])
+    ]),
+    div({ style: { flex: 1, display: 'flex' } }, [
+      div({ style: Style.dashboard.leftBox }, [
+        div({ style: Style.dashboard.header }, [
+          'About the workspace'
+        ]),
+        !!description && h(MarkdownViewer, [description])
```
dmohs: How about?
```suggestion
        workspace?.description && h(MarkdownViewer, [workspace.description])
```
(and remove from above)

src/pages/library/Showcase.js
dmohs: Yes. This is our style, for better or worse. I think it's pretty readable once you get used to it.

src/components/SignInButton.js
```diff
@@ -1,24 +1,34 @@
...
+  const isGoogleAuthInitialized = authSettled(auth)
+
+  useEffect(() => {
+    if (isGoogleAuthInitialized) {
+      window.gapi.signin2.render('signInButton', {
+        scope: 'openid profile email',
+        width: 250,
+        height: 56,
+        longtitle: true,
+        theme: props.theme,
```
dmohs: With the above, I think you can:
```suggestion
        theme,
```

src/libs/auth.js
```diff
@@ -54,10 +54,11 @@ export const ensureBillingScope = async () => {
   }
 }
 
+export const authSettled = ({ isSignedIn, registrationStatus }) => {
```
dmohs: I think perhaps `isAuthSettled` since this became a function.

dmohs (approved):

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,333 @@ import * as Style from 'src/libs/style'
...
+    'gnomAD', 'NCI', 'COVID-19']
+}, {
+  name: 'Utilities',
+  labels: ['Format Conversion', 'Developer Tools']
+}, {
+  name: 'Projects',
+  labels: ['HCA', 'AnVIL', 'BRAIN Initiative', 'BioData Catalyst', 'NCI']
+}]
+
+const uniqueSidebarTags = _.flow([
```
zarsky-broad: Inline this?

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,333 @@ import * as Style from 'src/libs/style'
...
+        ]) :
+        h(SidebarCollapser,
+          {
+            key: section.name,
+            title: section.name,
+            onClick: () => setCollapsedSections(_.xor([section], collapsedSections)),
+            isOpened: !_.includes(section, collapsedSections)
+          },
+          [
+            ..._.map(label => {
```
zarsky-broad: instead of spread merging this array, could you just do the map directly and use a key on the clickable?

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,333 @@ import * as Style from 'src/libs/style'
...
+
+  const [selectedSections, setSelectedSections] = useState([])
+  const [selectedTags, setSelectedTags] = useState([])
+  const [searchFilter, setSearchFilter] = useState()
+  const [sort, setSort] = useState('most recent')
+
+  const sortWorkspaces = Utils.cond(
+    [sort === 'most recent', () => _.orderBy(['created'], ['desc'])],
+    [sort === 'alphabetical', () => _.orderBy(w => _.toLower(_.trim(w.name)), ['asc'])],
+    () => _.identity
```
zarsky-broad: I don't see a need for this default case since there's no way for this to be something else; that said, I'd tell you to use a switchcase instead of a cond, but with only two options could you just use a ternary?

src/pages/library/Showcase.js
```diff
@@ -16,104 +18,333 @@ import * as Style from 'src/libs/style'
...
+  )
 
   Utils.useOnMount(() => {
     const loadData = async () => {
-      const featuredList = await Ajax().Buckets.getFeaturedWorkspaces()
+      const showcase = await Ajax().Buckets.getShowcaseWorkspaces()
+
+      // Immediately lowercase the workspace tags so we don't have to think about it again.
+      // Also pre-compute lower name and description.
+      const featuredWorkspaces = _.flow([
```
zarsky-broad: flow is variadic, don't pass it an array plz

src/components/SignInButton.js
breilly2: I just found that I somehow didn't push all of my changes. :( Update to SignInButton incoming...

src/pages/library/Showcase.js
breilly2: This is used by `groupByFeaturedTags` which operates on the current, possibly filtered, list of workspaces. I was worried that that computation was already a little heavier than I wanted. My intent here was to compute `uniqueSidebarTags` once ahead of time since that is only dependent on the static `sidebarSections`.

Do you still want me to inline it? Or is there a better way to express this? (I considered `_.once`, but that seemed unnecessary since there's no reason to defer the calculation.)

src/pages/library/Showcase.js
breilly2: Interesting. Recognizing that `_.map` would return an array, I removed the surrounding `[]` and... react didn't complain about not specifying `key`s. Sure enough, the same is true for the surrounding map over `sections`.

Should I remove the `[]`s and add/keep keys or remove all of it? From my reading, it looks like keeping them in an array and providing keys might be more performant.

zarsky-broad: yup, your target is:
```js
div([
  _.map(x => div({ key: x.uniqueId }), list)
])
```

src/pages/library/Showcase.js
zarsky-broad: I think there are a bunch of things that are being recomputed on every render now, but I think a bunch of them may just need to be recomputed when the featured workspace list is loaded; I bet we could take care of that with some memoization...

src/pages/library/Showcase.js
breilly2: Yeah, I definitely agree with the `switchCase`, and you're right that the default case isn't needed. I do want to stop short of the ternary; neither feels like a natural primary or secondary case to me.

src/pages/library/Showcase.js
breilly2: I wanted to do a better job with that, but unfortunately I just ran out of time.

I also couldn't make sense of the `StateHistory.update({ featuredWorkspaces })`, which looks like it wants to avoid repeated downloads, but there wasn't any condition guarding that fetch.

Anyway, I'm definitely open to adding some memoization. My first candidate is `groupByFeaturedTags`. Let me know if you have other suggestions.

src/components/SignInButton.js
```diff
@@ -1,24 +1,34 @@
...
+        prompt: 'select_account'
+      })
+    }
+  }, [isGoogleAuthInitialized, theme])
 
   // For some reason, Google's rendered Sign-In button is not at all keyboard accessible.
   // To fix this, we wrap it as a button, and propagate the keyboard-accessible click event down to
   // the inner DOM node inside the button, then let it bubble up to whatever it is that catches it.
-  return h(Clickable, {
+  return !isGoogleAuthInitialized ? h(spinner) : h(Clickable, {
```
zarsky-broad: While this is in fact valid, the pattern in the codebase is to call `spinner()` directly; we should probably think about making it `h(Spinner)` at some point, but for now would you mind changing this for consistency?
```suggestion
  return !isGoogleAuthInitialized ? spinner() : h(Clickable, {
```

src/pages/library/Showcase.js
zarsky-broad: FWIW, the intention of these state history saves isn't to prevent a fetch, it's to show old data while the new data is being fetched, so users can start to navigate without waiting.

zarsky-broad (approved):
Not doing a full re-review, happy if others are happy. One little nit, good to go once that's addressed!

src/pages/library/Showcase.js
dmohs: https://user-images.githubusercontent.com/1545444/129941034-ae6e6fbc-55b3-4a5c-aa10-9b1ad280a0f8.mov

Ignore the error notifications and focus on the interaction. When I load the page, you see a spinner. When I navigate back to it, the page loads instantaneously.  This is the type of attention to detail and user empathy that puts Terra UI in the top tier.

PR #2561, opened 2021-07-14
AS-879: Add support for tsv async imports

davidangb:
This changes TSV upload to an async process, similar to PFB import.

### DIFF
--- a/src/components/ImportStatus.js
+++ b/src/components/ImportStatus.js
@@ -4,16 +4,16 @@ import { h } from 'react-hyperscript-helpers'
 import { Ajax } from 'src/libs/ajax'
 import { withErrorReporting } from 'src/libs/error'
 import { clearNotification, notify } from 'src/libs/notifications'
-import { pfbImportJobStore } from 'src/libs/state'
+import { asyncImportJobStore } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
 
 const ImportStatus = () => {
-  const jobs = Utils.useStore(pfbImportJobStore)
+  const jobs = Utils.useStore(asyncImportJobStore)
   return h(Fragment, _.map(job => h(ImportStatusItem, {
     job,
     onDone: () => {
-      pfbImportJobStore.update(_.reject({ jobId: job.jobId }))
+      asyncImportJobStore.update(_.reject({ jobId: job.jobId }))
     }
   }), jobs))
 }
@@ -22,14 +22,14 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
   const signal = Utils.useCancellation()
 
   Utils.usePollingEffect(
-    withErrorReporting('Problem checking status of PFB data import', async () => {
+    withErrorReporting('Problem checking status of data import', async () => {
       await checkForCompletion(targetWorkspace, jobId)
     }), { ms: 5000 })
 
   const checkForCompletion = async ({ namespace, name }, jobId) => {
     const fetchImportStatus = async () => {
       try {
-        return await Ajax(signal).Workspaces.workspace(namespace, name).importPFBStatus(jobId)
+        return await Ajax(signal).Workspaces.workspace(namespace, name).getImportJobStatus(jobId)
       } catch (error) {
         // Ignore 404; We're probably asking for status before the status endpoint knows about the job
         if (error.status === 404) {
@@ -64,7 +64,7 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
         ['Done', successNotify],
         ['ERROR', errorNotify],
         ['Error', errorNotify],
-        [Utils.DEFAULT, () => notify('error', 'Unexpected error importing PFB data', response)]
+        [Utils.DEFAULT, () => notify('error', 'Unexpected error importing data', response)]
       )
       clearNotification(jobId)
       onDone()
--- a/src/components/ImportStatus.js
+++ b/src/components/ImportStatus.js
@@ -47,12 +47,12 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
     // avro-import statuses: PENDING, RUNNING, SUCCESS, ERROR
     // import service statuses: Pending, Translating, ReadyForUpsert, Upserting, Done, Error
     // TODO: only need to support both sets of statuses during the transition from avro-import to import service.
-    // once import servie is fully adopted, we can/should remove the avro-import status values.
+    // once import service is fully adopted, we can/should remove the avro-import status values.
 
     const successNotify = () => notify('success', 'Data imported successfully.',
       {
         message: `Data import to workspace "${namespace} / ${name}" is complete, please refresh the Data view.
-      Because this data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names, 
+      If your data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names, 
       e.g. "this.pfb:consent_codes" instead of "this.consent_codes."`
       })
 
--- a/src/components/ImportStatus.js
+++ b/src/components/ImportStatus.js
@@ -56,7 +56,7 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
       e.g. "this.pfb:consent_codes" instead of "this.consent_codes."`
       })
 
-    const errorNotify = () => notify('error', 'Error importing PFB data.', message)
+    const errorNotify = () => notify('error', 'Error importing data.', { message: message })
 
     if (!_.includes(status, ['PENDING', 'RUNNING', 'Pending', 'Translating', 'ReadyForUpsert', 'Upserting'])) {
       Utils.switchCase(status,
--- a/src/components/ImportStatus.js
+++ b/src/components/ImportStatus.js
@@ -56,7 +56,7 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
       e.g. "this.pfb:consent_codes" instead of "this.consent_codes."`
       })
 
-    const errorNotify = () => notify('error', 'Error importing data.', { message: message })
+    const errorNotify = () => notify('error', 'Error importing data.', { message })
 
     if (!_.includes(status, ['PENDING', 'RUNNING', 'Pending', 'Translating', 'ReadyForUpsert', 'Upserting'])) {
       Utils.switchCase(status,
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -20,7 +20,8 @@ import colors from 'src/libs/colors'
 import { reportError } from 'src/libs/error'
 import Events from 'src/libs/events'
 import { FormLabel } from 'src/libs/forms'
-import { requesterPaysProjectStore } from 'src/libs/state'
+import { notify } from 'src/libs/notifications'
+import { pfbImportJobStore, requesterPaysProjectStore } from 'src/libs/state'
 import * as StateHistory from 'src/libs/state-history'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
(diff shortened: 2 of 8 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -447,6 +447,18 @@ const WorkspaceData = _.flow(
...
+      const runningJobs = await Ajax(signal).Workspaces.workspace(namespace, name).listImportJobs(true)
+      _.map(job => {
+        const jobStatus = _.lowerCase(job.status)
+        if (!(jobStatus === "success" || jobStatus === "error" || jobStatus === "done"))
+          asyncImportJobStore.update(Utils.append({ targetWorkspace: { namespace, name }, jobId: job.jobId }))
+      }, runningJobs)
+    } catch (error) {
+      reportError('Error loading running import jobs in this workspace')
+    }
+  }
```
ansingh7115, line 462: This change means that the Data tab checks for running import jobs and polls any that exist. The user impact is that reloading the data page will indicate running jobs.

src/pages/Upload.js
```diff
@@ -596,7 +596,7 @@ const MetadataUploadPanel = _.flow(
       // Convert the table data structure back into a TSV, in case the user made changes
       const file = Utils.makeTSV([metadata.table.columns, ...metadata.table.rows])
       const workspace = Ajax().Workspaces.workspace(namespace, name)
-      await workspace.importFlexibleEntitiesFile(file)
+      await workspace.importFlexibleEntitiesFile(false, file)
```
ansingh7115: I preserved this as a non-async import, but am open to changing it to true

src/pages/workspaces/workspace/Data.js
```diff
@@ -81,7 +81,7 @@ const DataTypeButton = ({ selected, entityName, children, entityCount, iconName
 }
 
 const DataImportPlaceholder = () => {
-  return div({ style: { ...Style.navList.item(false), color: colors.dark(0.7) } }, [
+  return div({ style: { ...Style.navList.item(false), color: colors.dark(0.7), marginLeft: '0.5rem' } }, [
```
ansingh7115, line 84: The "Data import in Progress" icon and message were right up against the edge of the window, which looked bad.

src/libs/ajax.js
```diff
@@ -833,19 +833,29 @@ const Workspaces = signal => ({
...
-      importPFBStatus: async jobId => {
-        const res = await fetchOrchestration(`api/${root}/importPFB/${jobId}`, _.merge(authOpts(), { signal }))
+      getImportJobStatus: async jobId => {
+        const res = await fetchOrchestration(`api/${root}/importJob/${jobId}`, _.merge(authOpts(), { signal }))
+        return res.json()
+      },
+
+      listImportJobs: async isRunning => {
+        // ToDo: This endpoint should be deprecated in favor of more generic "importJob" endpoint
+        const res = await fetchOrchestration(`api/${root}/importPFB?running_only=${isRunning}`, _.merge(authOpts(), { signal }))
```
davidangb, line 862: ah, good call, that's work we need to do on the back end at some point

src/components/ImportStatus.js
```diff
@@ -47,24 +47,24 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
...
     const successNotify = () => notify('success', 'Data imported successfully.',
       {
         message: `Data import to workspace "${namespace} / ${name}" is complete, please refresh the Data view.
-      Because this data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names, 
+      If your data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names, 
       e.g. "this.pfb:consent_codes" instead of "this.consent_codes."`
       })
 
-    const errorNotify = () => notify('error', 'Error importing PFB data.', message)
+    const errorNotify = () => notify('error', 'Error importing data.', { message: message })
```
dmohs: Nit:
```suggestion
    const errorNotify = () => notify('error', 'Error importing data.', { message })
```

src/components/data/data-utils.js
```diff
@@ -221,6 +222,13 @@ export const EntityDeleter = ({ onDismiss, onSuccess, namespace, name, selectedE
 
 const supportsFireCloudDataModel = entityType => _.includes(entityType, ['pair', 'participant', 'sample'])
 
+export const notifyDataImportProgress = (jobId) => {
```
dmohs: Nit: I believe our convention is to leave of the parens when it is possible to do so:
```suggestion
export const notifyDataImportProgress = jobId => {
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -447,6 +447,18 @@ const WorkspaceData = _.flow(
 
   const toSortedPairs = _.flow(_.toPairs, _.sortBy(_.first))
 
+  const getRunningImportJobs = async () => {
+    try {
+      const runningJobs = await Ajax(signal).Workspaces.workspace(namespace, name).listImportJobs(true)
+      _.map(job => {
+        const jobStatus = _.lowerCase(job.status)
+        if (!(jobStatus === "success" || jobStatus === "error" || jobStatus === "done"))
```
dmohs: I think this is a place where a bit more concision might help readability:
```suggestion
        if (!_.includes(jobStatus, ['success', 'error', 'done']))
```

src/libs/ajax.js
```diff
@@ -833,19 +833,29 @@ const Workspaces = signal => ({
         return fetchOrchestration(`api/${root}/importEntities`, _.merge(authOpts(), { body: formData, signal, method: 'POST' }))
       },
 
-      importFlexibleEntitiesFile: file => {
+      importFlexibleEntitiesFile: async (isAsync, file) => {
```
dmohs: I don't love the name of the option `isAsync`. I judge this function is always async. I might call this option `shouldReadJsonBody` or `readBody` or somesuch.

dmohs (approved):

src/pages/Upload.js
davidangb: where is this in the UI?

src/pages/workspaces/workspace/Data.js
davidangb: should this pop open the "import in progress" modal as well?

I worry a little bit here about cases where a workspace has a failed import that's stuck in some running state … in that condition, we'll *always* show a running job. I realize the proper fix for this is in import service/Rawls, to make sure we never have stuck jobs!

src/components/ImportStatus.js
```diff
@@ -47,24 +47,24 @@ const ImportStatusItem = ({ job: { targetWorkspace, jobId }, onDone }) => {
...
     // import service statuses: Pending, Translating, ReadyForUpsert, Upserting, Done, Error
     // TODO: only need to support both sets of statuses during the transition from avro-import to import service.
-    // once import servie is fully adopted, we can/should remove the avro-import status values.
+    // once import service is fully adopted, we can/should remove the avro-import status values.
 
     const successNotify = () => notify('success', 'Data imported successfully.',
       {
         message: `Data import to workspace "${namespace} / ${name}" is complete, please refresh the Data view.
-      Because this data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names, 
+      If your data comes from a PFB, you must include the namespace "pfb:" when referring to attribute names,
```
davidangb, line 55: if we wanted this message to be tailored to TSV vs. PFB, what would be involved? We'd probably have to persist the filetype into the `asyncImportJobStore`, right?

davidangb (commented):
some questions/comments inline

src/components/ImportStatus.js
ansingh7115: Yeah I tried to figure out a way to do that here, but at this stage we don't know what the file type is. If we could preserve file type and return it when we get/list jobs, that would help here.

src/pages/Upload.js
ansingh7115: I couldn't figure that out either! 
Isaac says:
> That's a whole separate upload pathway for covid stuff (I think?) that was built with external help

Brian R says:
> Correct. That page isn’t linked from anywhere. Presumably, those who have been invited to use it were given links.

So, I think we should keep it false for now to avoid messing up that page and then having to hotfix or rollback. After the fire is over, we could contact the external person or whoever owns that page and see if they want to convert the import there to async. How does that sound?

src/pages/workspaces/workspace/Data.js
ansingh7115: We could - I was thinking, in the case that a job got stuck in progress, a modal popping up might be annoying for the user. 

Yeah, I was a little worried about stuck imports too. I think one way to fix it might be to pass up when the job started, and the UI could ignore jobs that have been running for more than, say, 24 hours. (obviously, that's more of a quick, in-case-things-go-wrong solution. Ideally, as you said, the back end should be able to handle stuck jobs)

src/libs/ajax.js
ansingh7115: Ah, I see! The function itself is async, while the endpoint it's calling has a query parameter called `async`. So that could be confusing. What about `isAsyncImport` or `isAsyncJob`?

src/pages/Upload.js
davidangb: I saw answers in Slack - agree to leave this exactly as-is for compatibility

src/pages/workspaces/workspace/Data.js
davidangb: all sounds good - I say leave this UI PR as-is, and if/when we realize stuck jobs are a problem in practice, we deal with it in the back end

src/components/ImportStatus.js
davidangb: Ah, yeah - especially after reloading the page, you'd need the filetype from the job list. Got it - another ticket to add to the queue!

src/libs/ajax.js
ansingh7115: The response from an async import looks like:
```
{   "url": "string",   "jobId": "string",   "workspace": {     "namespace": "string",     "name": "string"   } }
```

while the response from the synchronous import just returns the entity type as a string (so `participant` or `sample_set`)

zarsky-broad: I think this should be a separate entry in this file called `startAsyncFlexibleImport` or similar, then; generally, I want one definition here per endpoint.

ansingh7115: It is the same endpoint: https://firecloud-orchestration.dsde-dev.broadinstitute.org/#/Entities/flexibleImportEntities , whether or not it does the import asynchronously is decided by the `isAsync` query parameter

ansingh7115: or are you considering that to be different endpoints?

I have nothing against writing it as two separate functions - one for synchronous and one for async, especially since the responses are different. Just clarifying that, to me, this is a singular endpoint, and if we are following "one definition per endpoint" guideline, then it would be one definition

src/libs/auth.js
```diff
@@ -13,7 +13,7 @@ import * as Nav from 'src/libs/nav'
 import { clearNotification, notify, sessionTimeoutProps } from 'src/libs/notifications'
 import { getLocalPref, getLocalPrefForUserId, setLocalPref } from 'src/libs/prefs'
 import allProviders from 'src/libs/providers'
-import { authStore, cookieReadyStore, pfbImportJobStore, requesterPaysProjectStore, workspacesStore, workspaceStore } from 'src/libs/state'
+import { authStore, cookieReadyStore, asyncImportJobStore, requesterPaysProjectStore, workspacesStore, workspaceStore } from 'src/libs/state'
```
sagehen03: Not sure if it's worth worrying about, but I'm getting a linter warning about the imports not being alphabetized.

davidangb:
Github won't let me approve since I technically started this PR, so here's my thumb: 👍

src/libs/ajax.js
zarsky-broad: Ahh, I missed that it was the same endpoint, since the returns are so different. I would treat it like a separate one for that reason.

sagehen03 (approved):
lgtm modulo any linter errors.

PR #2562, opened 2021-07-14
Reduce motion of Terra banners [BW-755][hackathon]

aednichols:
Replace slide-in animation with a subtle fade. This is an [accessibility measure](https://a11y-101.com/development/reduced-motion) that helps users who have trouble concentrating in the face of distraction (like me!).

*Before:*

Jul-14-2021 15-37-05

*After:*

Jul-14-2021 15-37-12

You may have come across this similar setting in iOS:

IMG_5103

In this case I did not elect to make it a setting because I don't think the absence of sliding will impair anyone from noticing the banners, and it would be hard to make the setting discoverable.

### DIFF
--- a/src/libs/notifications.js
+++ b/src/libs/notifications.js
@@ -147,8 +147,9 @@ const showNotification = ({ id, timeout }) => {
     ]),
     container: 'top-right',
     dismiss: { duration: !!timeout ? timeout : 0, click: false, touch: false },
-    animationIn: ['animate__animated', 'animate__slideInRight'],
-    animationOut: ['animate__animated', 'animate__slideOutRight'],
+    animationIn: ['animate__animated', 'animate__fadeIn'],
+    animationOut: ['animate__animated', 'animate__fadeOut'],
+    insert: "bottom",
     width: 350
   })
 }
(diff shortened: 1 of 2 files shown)


s-rubenstein (approved):
I really like this change from a visual clutter perspective. 

Just to check, have you run this by UX?

src/libs/notifications.js
```diff
@@ -147,8 +147,9 @@ const showNotification = ({ id, timeout }) => {
     ]),
     container: 'top-right',
     dismiss: { duration: !!timeout ? timeout : 0, click: false, touch: false },
-    animationIn: ['animate__animated', 'animate__slideInRight'],
-    animationOut: ['animate__animated', 'animate__slideOutRight'],
+    animationIn: ['animate__animated', 'animate__fadeIn'],
+    animationOut: ['animate__animated', 'animate__fadeOut'],
```
s-rubenstein, line 151: I might leave the animationOut as a slide? Just since that one (I think) needs to be initiated by the user, a harder out animation might be nice for stronger immediate feedback, but it would disrupt the symmetry there. I would defer to UX.

salonishah11 (approved):
As long as UX is happy, it LGTM 👍

aednichols:
[Joy says 💯](https://broadinstitute.slack.com/archives/C01EHNUM73R/p1626795251152900?thread_ts=1626291845.151600&cid=C01EHNUM73R)

aednichols:
Adding @breilly2 as the most proximate Terra code owner

aednichols:
Apologies for putting @breilly2 on the spot, have learned that he is no longer a code owner & removed from review list (but if you have feedback anyway I always welcome it!)

dmohs (approved):
Thank you for caring about stuff like this, Adam.

PR #2563, opened 2021-07-14
A11y hackathon

marctalbott:
<!--
Hello, friend!
Thanks!
--!>


petesantos (changes requested):
As is, this only tests pages that don't require a login, and the login screen. To make this work we would need to run this with a system capable of signing in to the application similar to what the integration tests do.

kyuksel:
Closing because it seems stale. Please re-open if applicable.

PR #2564, opened 2021-07-16
[IA-2710] Cloud status page

jdcanas:
This PR does a lot of things with some fairly complex logic.... I'm considering setting up a bug bashing session for it. Below I'll enumerate the big things:

- Modifies the context bar to have a cloud status page that has a panel for each tool we support. Each tool has 3 buttons
    - Button 1: There is a cloud env button that allows users to create/delete/modify a cloud env for a given tool. This should be disabled when creation isn't allowed today for the tool 
        - We only allow envs for (jupyter **or** Rstudio) **and** galaxy (max of 2)
    - Button 2: This shows the status for the current runtime, similar to the widget in the top right currently, while allowing stop/starting.
    - Button 3: This allows 'launching' of the app. This takes you to the RStudio editor for R, the tree view for Jupyter, and the current launch functionality for galaxy. This should be disabled when launch doesn't make sense. 
    - tooltips for all states should make sense and be contextual
- Refactors a few functions for naming consistency

There are so many possible states its impossible to enumerate via screenshots... here are some examples. Note that this modal also houses the `NewRuntime` and `NewGalaxy` modals via the cloud env buttons.

image

image


src/components/ContextBar.js
```diff
@@ -19,50 +25,77 @@ const contextBarStyles = {
   }
 }
 
-export const ContextBarButtons = ({ setDeletingWorkspace, setCloningWorkspace, setSharingWorkspace, isOwner, canShare }) => {
```
jdcanas, line 23: not sure why git decided to not identify the diffs when all I really did was rename this component.... Ill call out the actual different stuff inline below

src/components/ContextBar.js
```diff
@@ -19,50 +25,77 @@ const contextBarStyles = {
...
+  const terminalDisabled = !(currentRuntimeTool === tools.Jupyter.label) || getIsRuntimeBusy(currentRuntime)
+  const terminalLaunchLink = Nav.getLink('workspace-application-launch', { namespace, name: workspaceName, application: 'terminal' })
+
+  const startCurrentRuntime = () => {
+    const { googleProject, runtimeName } = currentRuntime
+    Ajax().Runtimes.runtime(googleProject, runtimeName).start()
+  }
+
+  return h(Fragment, [
+    h(CloudEnvironmentModal, {
```
jdcanas, line 82: added this modal, which is the main addition of this PR

src/components/ContextBar.js
```diff
@@ -19,50 +25,77 @@ const contextBarStyles = {
...
-      tooltipDelay: 100,
-      'aria-label': 'Terminal'
-    }, [icon('terminal', { size: 24 })])
+export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setSharingWorkspace, isOwner, canShare, canCompute, runtimes, apps, galaxyDataDisks, refreshRuntimes, refreshApps, workspace, persistentDisks, workspace: { workspace: { namespace, bucketName, name: workspaceName } } }) => {
+  const [isCloudEnvOpen, setCloudEnvOpen] = useState(false)
+  const currentRuntime = getCurrentRuntime(runtimes)
+  const currentRuntimeTool = currentRuntime?.labels?.tool
+
+  const terminalDisabled = !(currentRuntimeTool === tools.Jupyter.label) || getIsRuntimeBusy(currentRuntime)
+  const terminalLaunchLink = Nav.getLink('workspace-application-launch', { namespace, name: workspaceName, application: 'terminal' })
```
jdcanas, line 45: added terminal functionality, with proper disabling

src/components/ContextBar.js
```diff
@@ -19,50 +25,77 @@ const contextBarStyles = {
...
+          onClick: () => setCloudEnvOpen(!isCloudEnvOpen),
+          ...{ tooltip: 'Compute Configuration', tooltipDelay: 100 },
+          'aria-label': 'Compute Configuration'
+        }, [icon('cloudBolt', { size: 24 })]),
+        h(Clickable, {
+          style: { ...contextBarStyles.contextBarButton, color: terminalDisabled ? colors.dark(0.7) : contextBarStyles.contextBarButton.color },
+          hover: { boxShadow: `inset -6px 0px ${terraSpecial(0.9)}` },
+          disabled: terminalDisabled,
+          href: terminalLaunchLink,
+          onClick: window.location.hash === terminalLaunchLink && currentRuntime?.status === 'Stopped' ? () => startCurrentRuntime() : undefined,
```
jdcanas, line 140: the terminal will start the runtime if its stopped. Users should not be able to click on this if the terminal is not in some other ready state.

src/components/NewAnalysisModal.js
```diff
@@ -24,33 +24,41 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
     const [busy, setBusy] = useState()
 
     const renderNewRuntimeModal = tool => h(NewRuntimeModalBase, {
-      isOpen: true, //always true, this is detached from dom for other viewmodes within this modal
+      isOpen: viewMode === NEW_JUPYTER_MODE || viewMode === NEW_RSTUDIO_MODE,
```
jdcanas, line 27: this was a bug. @rtitle, you actually identified it when you asked if it should re-open to the previous page. I thought it was just remembering choices, but in fact it was still attached because the viewMode was never reset on modal close, which is a bit annoying I found after some playing around with it. Same thing was happening with new modal. Most other changes in this file ensure the viewmode is properly reset upon modal close.

src/components/NewGalaxyModal.js
```diff
@@ -308,11 +308,11 @@ export const NewGalaxyModalBase = Utils.withDisplayName('NewGalaxyModal')(
       const gridStyle = { display: 'grid', gridTemplateColumns: '0.75fr 4.5rem 1fr 5.5rem 1fr 5.5rem', gridGap: '1rem', alignItems: 'center' }
       return Utils.cond(
         [currentDataDisk, () => {
-          return div({ marginTop: '0.75rem' }, [
```
jdcanas, line 305: the two changes involving margins in this file are correcting bugs that showed up in the console. Note that these margins are not actually in a style prop, and are floating top-level props (aka not doing anything)

src/components/notebook-utils.js
```diff
@@ -69,8 +69,8 @@ export const analysisNameInput = ({ inputProps, ...props }) => h(ValidatedInput,
 })
 
 export const tools = {
-  RStudio: { label: 'RStudio', ext: 'Rmd', imageIds: ['terra-jupyter-bioconductor', 'RStudio', 'terra-jupyter-bioconductor_legacy'], defaultImageId: 'terra-jupyter-bioconductor' },
-  Jupyter: { label: 'Jupyter', ext: 'ipynb', imageIds: ['terra-jupyter-hail', 'terra-jupyter-python', 'terra-jupyter-gatk', 'Pegasus', 'terra-jupyter-gatk_legacy'], defaultImageId: 'terra-jupyter-gatk' },
+  RStudio: { label: 'RStudio', ext: 'Rmd', imageIds: ['RStudio'], defaultImageId: 'RStudio' },
```
jdcanas, line 72: this was a bug, in that the default RStudio image was not actually an RStudio image

src/libs/runtime-utils.js
```diff
@@ -196,17 +196,25 @@ export const isCurrentGalaxyDiskDetaching = apps => {
   return currentGalaxyApp && _.includes(currentGalaxyApp.status, ['DELETING', 'PREDELETING'])
 }
 
-export const collapsedRuntimeStatus = runtime => {
```
jdcanas, line 217: I disliked this function name having `collapsed` as a non-description. Renamed to be consistent with the one we have for apps that essentially does the same thing.

src/libs/runtime-utils.js
```diff
@@ -167,7 +167,7 @@ export const machineCost = machineType => {
   return _.find(knownMachineType => knownMachineType.name === machineType, machineTypes).price
 }
 
-export const currentApp = _.flow(trimAppsOldestFirst, _.last)
+export const getCurrentApp = _.flow(trimAppsOldestFirst, _.last)
```
jdcanas, line 188: functions are verbs variables are nouns, otherwise the variable that holds the result has to be named weird.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -308,23 +313,20 @@ const Notebooks = _.flow(
...
-          div(['Galaxy ', versionTag('Beta', { color: colors.primary(1.5), backgroundColor: 'white', border: `1px solid ${colors.primary(1.5)}` })]),
-          icon('plus-circle', { style: { marginTop: '0.5rem' }, size: 21 })
-        ])
-    }
+    const getGalaxyText = (app, galaxyDataDisks) => app ?
+      div({ style: { fontSize: 18, lineHeight: '22px', width: 160 } }, [
+        div(['Galaxy Interactive']),
+        div(['Environment']),
+        div({ style: { fontSize: 12, marginTop: 6 } },
+          getGalaxyCostTextChildren(app, galaxyDataDisks)),
```
jdcanas, line 321: refactored this to be able to re-use the cost text

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -140,15 +142,16 @@ const WorkspaceContainer = ({ namespace, name, breadcrumbs, topBarContent, title
...
-          div({ style: Style.elements.contextBarContainer }, [
-            h(ContextBarButtons, {
-              setDeletingWorkspace,
-              setCloningWorkspace,
-              setSharingWorkspace,
-              isOwner,
-              canShare
-            })
-          ])
+          workspace && h(ContextBar, {
```
jdcanas, line 144: removed the styling from the workspace container and put it in the context bar itself

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -0,0 +1,401 @@
...
+            },
+            hover: { backgroundColor: colors.accent(0.2) },
+            tooltip: Utils.cond([isCloudEnvModalDisabled(toolLabel), () => 'Edit disabled, processing'],
+              [doesCloudEnvForToolExist, () => 'Edit existing environment'],
+              [!doesCloudEnvForToolExist, () => 'Create new environment (may overwrite existing)']),
+            disabled: isCloudEnvModalDisabled(toolLabel),
+            tooltipDelay: 100,
+            onClick: () => setViewMode(toolLabel)
+          }, [
+            icon('cloudBolt', { size: 20 }), //TODO: why doesn't this icon disable properly color-wise?
```
jdcanas: This doesn't look good like the rest of the icons when disabled (it doesn't inherit the color of the parent clickable).... I think its a fact that its an `svg` in the icons foldler and not a font-awesome icon. Any ideas on how to fix this? Its available as a font-awesome pro icon but I think that would require some sort of credentials to pull at yarn install time, which seems like a pain to set up...

kyuksel:
The text `UPLOAD` on the button below doesn't look vertically centered. Is that intentional?

image

kyuksel:
Linting errors need fixing.

src/components/ContextBar.js
```diff
@@ -20,54 +36,114 @@ const contextBarStyles = {
...
-      // TODO: add click handler
-      tooltip: 'Terminal',
-      tooltipSide: 'left',
-      tooltipDelay: 100,
-      'aria-label': 'Terminal'
-    }, [icon('terminal', { size: 24 })])
+          disabled: !isTerminalEnabled,
+          href: terminalLaunchLink,
+          onClick: window.location.hash === terminalLaunchLink && currentRuntime?.status === 'Stopped' ? () => startCurrentRuntime() : undefined,
+          tooltip: !isTerminalEnabled ? 'Terminal disabled due to current environment status' : 'Terminal',
```
kyuksel: It's not just _status_ that can cause terminal to be disabled but also the kind of app (`Jupyter` vs. `RStudio` vs. `Galaxy`). Therefore, I wonder if something like below would be more accurate:
```suggestion
          tooltip: !isTerminalEnabled ? 'Terminal can be launched only with Jupyter environments when in running status.' : 'Terminal',
```

kyuksel:
We should probably not display `None` for cost estimates of non-existent apps.

image

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -0,0 +1,404 @@
...
+      disabled: isDisabled,
+      style: {
+        ...toolButtonStyles,
+        color: isDisabled ? colors.dark(0.7) : colors.accent()
+      },
+      hover: { backgroundColor: colors.accent(0.2) },
+      tooltip: Utils.cond(
+        [doesCloudEnvForToolExist && !isDisabled, () => 'Launch'],
+        [doesCloudEnvForToolExist && isDisabled && toolLabel !== tools.Jupyter.label, () => 'Launch disabled'],
+        [doesCloudEnvForToolExist && isDisabled && toolLabel === tools.Jupyter.label, () => 'Generic launch disabled for Jupyter. Select a notebook in the analyses tab to launch Jupyter'],
```
kyuksel: I eventually understood what the first sentence in the message is trying to convey but the first time I saw it, I had to think what `generic launch` would be. Since it's not something established per se, I'd suggest removing it, and using the space to also mention creating a notebook in case users don't have one already, like so

```suggestion
        [doesCloudEnvForToolExist && isDisabled && toolLabel === tools.Jupyter.label, () => 'Select or create a notebook in the analyses tab to launch Jupyter'],
```

kyuksel:
When the new Terminal button is clicked, the new tab that is launched doesn't have the new Analyses tab. Will this automatically change when the new UI is enabled by default?

image

kyuksel:
Cost estimate isn't displayed when Galaxy is in `Precreating` status:

image

I also wonder if cost estimate should be clearly displayed, perhaps by having a dedicated column or something  🤔

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -0,0 +1,404 @@
...
+      'aria-label': `Launch ${toolLabel}`,
+      disabled: isDisabled,
+      style: {
+        ...toolButtonStyles,
+        color: isDisabled ? colors.dark(0.7) : colors.accent()
+      },
+      hover: { backgroundColor: colors.accent(0.2) },
+      tooltip: Utils.cond(
+        [doesCloudEnvForToolExist && !isDisabled, () => 'Launch'],
+        [doesCloudEnvForToolExist && isDisabled && toolLabel !== tools.Jupyter.label, () => 'Launch disabled'],
```
kyuksel: I wonder if this should say something like 

```suggestion
        [doesCloudEnvForToolExist && isDisabled && toolLabel !== tools.Jupyter.label, () => 'Please wait until Galaxy is running'],
```

src/components/ContextBar.js
jdcanas: I'll add the note about Jupyter environments, but its not true it can only be launched when the runtime is running

kyuksel:
It'd be good to capitalize `g` in `Launch galaxy` below to be consistent:

image

jdcanas:
> The text `UPLOAD` on the button below doesn't look vertically centered. Is that intentional?
> 
> image

This will button be removed soon, pending a ticket I currently have

> We should probably not display None for cost estimates of non-existent apps.

What are you thinking? This is what the wireframes had, its an intentonally set string, not like Scala None. 

> When the new Terminal button is clicked, the new tab that is launched doesn't have the new Analyses tab. Will this automatically change when the new UI is enabled by default?

Yeah, right now enabling the tab is set in a console variable, aka to state associated with current browser window, so when it's always visible this problem won't exist.

> Cost estimate isn't displayed when Galaxy is in Precreating status:

This is consistent with current behavior.

src/libs/runtime-utils.js
```diff
@@ -214,17 +214,25 @@ export const isCurrentGalaxyDiskDetaching = apps => {
...
+export const getConvertedAppStatus = appStatus => {
   return Utils.switchCase(_.upperCase(appStatus),
-    ['STOPPED', () => _.capitalize('PAUSED')],
-    ['STOPPING', () => _.capitalize('PAUSING')],
-    ['STARTING', () => _.capitalize('RESUMING')],
+    ['STOPPED', () => 'Paused'],
+    ['STOPPING', () => 'Pausing'],
+    ['STARTING', () => 'Resuming'],
+    ['PRESTARTING', () => 'Starting'],
+    ['PRESTOPPING', () => 'Stopping'],
```
```suggestion
    ['PRESTARTING', () => 'Preresuming'],
    ['PRESTOPPING', () => 'Prepausing'],
```

kyuksel:
Looks like there are a few more of `start/stop` wording left over:

image

image

kyuksel:
- created a Galaxy app
- `LAUNCH GALAXY` pop-up appeared when the app was ready
- clicked on `Pause Galaxy` on the `CloudEnvironmentModal`
- the app status changed to `Pausing`
- the pop-up remained
- when I clicked on the pop-up, a new tab was created displaying a JSON with a proxy error

This scenario doesn't seem applicable to the current UI.

kyuksel:
Good work! 

image
Displaying app logos underneath the cloud looks especially cool!

Made my way through [Cloud Page bug bash](https://docs.google.com/spreadsheets/d/1IHflzBfG1upQj42zE8iOnFjev0ZjFcQDTzXB9rXga54) spreadsheet and marked most of them as verified, and noted otherwise where I couldn't and added a PR comment where applicable.

Will review the code once the comments so far have been addressed.

src/libs/runtime-utils.js
jdcanas: Ah. I changed these to 'Resuming' and 'Pausing'. The pre- statuses aren't intended for display to users.

jdcanas:
> Looks like there are a few more of `start/stop` wording left over:
> 
> image
> 
> image

This has how we currently do it for Jupyter/RStudio. Good time to change it for consistency though.

jdcanas:
> * created a Galaxy app
> * `LAUNCH GALAXY` pop-up appeared when the app was ready
> * clicked on `Pause Galaxy` on the `CloudEnvironmentModal`
> * the app status changed to `Pausing`
> * the pop-up remained
> * when I clicked on the pop-up, a new tab was created displaying a JSON with a proxy error
> 
> This scenario doesn't seem applicable to the current UI.

Hmm.. I'm able to re-create this with the existing flow too. I tried just disabling the button by app status, but unfortunately its hooked up [here](https://github.com/DataBiosphere/terra-ui/blob/dev/src/components/RuntimeManager.js#L229) as a static popup and doesnt receive state refreshes... I'm not sure there's an easy fix based on how its currently hooked up. I'll create a ticket

src/components/ContextBar.js
```diff
@@ -20,54 +35,114 @@ const contextBarStyles = {
...
+    return h(Fragment, [
+      ...(currentRuntime ? [getIconForTool(currentRuntimeTool, currentRuntime.status)] : []),
+      ...(currentApp ? [getIconForTool(tools.galaxy.label, currentApp.status)] : [])
+    ])
+  }
+
+  return h(Fragment, [
+    h(CloudEnvironmentModal, {
+      isOpen: isCloudEnvOpen,
+      onSuccess: () => {}, //TODO: I don't know if we want to do anything?
```
dmohs, line 84: If you don't know, then it's not a TODO. Instead, commit to the decision (do nothing) and then change it later if and when you gain information.

src/components/ContextBar.js
```diff
@@ -20,54 +35,114 @@ const contextBarStyles = {
...
+    }),
+    div({ style: Style.elements.contextBarContainer }, [
+      div({ style: contextBarStyles.contextBarContainer }, [
+        h(MenuTrigger, {
+          closeOnClick: true,
+          content: h(Fragment, [
+            h(MenuButton, { onClick: () => setCloningWorkspace(true) }, [makeMenuIcon('copy'), 'Clone']),
+            h(MenuButton, {
+              disabled: !canShare,
+              tooltip: !canShare && 'You have not been granted permission to share this workspace',
```
dmohs, line 99: I've been looking for this pattern lately (when the thing is disabled, explain why). Thank you.

src/components/ContextBar.js
```diff
@@ -20,54 +35,114 @@ const contextBarStyles = {
...
           tooltipSide: 'left',
-          onClick: () => setSharingWorkspace(true)
-        }, [makeMenuIcon('share'), 'Share']),
-        h(MenuButton, { disabled: true }, [makeMenuIcon('export'), 'Publish', comingSoon]),
-        h(MenuButton, {
-          disabled: !isOwner,
-          tooltip: !isOwner && 'You must be an owner of this workspace or the underlying billing project',
+          onClick: () => setCloudEnvOpen(!isCloudEnvOpen),
+          ...{ tooltip: 'Environment Configuration', tooltipDelay: 100 },
+          'aria-label': 'Environment Configuration'
```
dmohs: I think you can get this with `useTooltipAsLabel: true`, which would eliminate the need for copying it here.

src/components/ContextBar.js
```diff
@@ -20,54 +35,114 @@ const contextBarStyles = {
...
-      tooltipSide: 'left',
-      tooltipDelay: 100,
-      'aria-label': 'Terminal'
-    }, [icon('terminal', { size: 24 })])
+          disabled: !isTerminalEnabled,
+          href: terminalLaunchLink,
+          onClick: window.location.hash === terminalLaunchLink && currentRuntime?.status === 'Stopped' ? () => startCurrentRuntime() : undefined,
+          tooltip: !isTerminalEnabled ? 'Terminal can only be launched for Jupyter environments' : 'Terminal',
+          tooltipDelay: 100,
+          'aria-label': 'Terminal',
```
dmohs: Same as above or ignore if I'm wrong.

src/libs/icon-dict.js
```diff
@@ -1,9 +1,45 @@
...
+  faQuestionCircle,
+  faRocket,
+  faSearch,
+  faShareAlt,
+  faSquare as faSquareSolid,
+  faTachometerAlt,
+  faTasks,
+  faTerminal,
+  faTrashAlt,
+  faVirus
```
dmohs, line 42: Good call on expanding this out.

src/libs/runtime-utils.js
dmohs: I agree.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -0,0 +1,404 @@
...
+  const currentRuntimeStatus = getConvertedRuntimeStatus(currentRuntime)
+  const currentRuntimeTool = currentRuntime?.labels?.tool
+
+  const currentApp = getCurrentApp(apps)
+
+  const RuntimeIcon = ({ shape, onClick, disabled, messageChildren, toolLabel, style, ...props }) => {
+    return h(Clickable, {
+      'aria-label': `${toolLabel} Status`,
+      hover: { backgroundColor: colors.accent(0.2) },
+      tooltipDelay: 100,
```
dmohs: This value is set in several places in this PR. Is there a chance we have the wrong default?

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -0,0 +1,404 @@
...
+  }], [Utils.DEFAULT, () => {
+    const { googleProject, runtimeName } = currentRuntime
+    executeAndRefresh(toolLabel,
+      Ajax().Runtimes.runtime(googleProject, runtimeName).start())
+  }])
+
+  const stopApp = toolLabel => Utils.cond([toolLabel === tools.galaxy.label, () => {
+    const { googleProject, appName } = currentApp
+    executeAndRefresh(toolLabel,
+      Ajax().Apps.app(googleProject, appName).pause())
```
dmohs: Nit: Possibly not getting enough value out of the increase vertical space here.

src/components/ContextBar.js
dmohs: Also, I like your `() => undefined` pattern for a no-op (used later) better.

dmohs (approved):

src/libs/runtime-utils.js
kyuksel: I like the idea of not displaying `pre-` statuses to users as that fits well the semantics of `app`s as designed them. However, their conversion to`start/stop`sounds confusing to me. For instance, with
```
['PRESTARTING', () => 'Starting'],
['PRESTOPPING', () => 'Stopping']
```
when a user wants to create a Galaxy app, the statuses go as: `Starting` => `Provisioning` => `Running`, which to me begs the questions:
- how can we be _starting_ an app that didn't exist to begin with?
- wouldn't something have to be provisioned _before_ starting it?

How about 

```
['PRESTARTING', () => 'Processing'],
['PRESTOPPING', () => 'Processing']
```

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -0,0 +1,404 @@
...
+      await promise
+      await toolLabel === tools.galaxy.label ? refreshApps() : refreshRuntimes()
+    } catch (error) {
+      reportError('Cloud Environment Error', error)
+    } finally {
+      setBusy(false)
+    }
+  }
+
+  // We assume here that button disabling is working properly, so the only thing to check is whether its galaxy or the current (assumed to be existing) runtime
```
kyuksel:
```suggestion
  // We assume here that button disabling is working properly, so the only thing to check is whether it's a Galaxy app or the current (assumed to be existing) runtime.
```

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -0,0 +1,404 @@
...
+    [NEW_JUPYTER_MODE, () => 675],
+    [NEW_RSTUDIO_MODE, () => 675],
+    [NEW_GALAXY_MODE, () => 675],
+    [Utils.DEFAULT, () => 430]
+  )
+
+  const modalBody = h(Fragment, [
+    h(TitleBar, {
+      id: titleId,
+      title: 'Cloud environment details',
```
kyuksel: Would 
```suggestion
      title: 'Cloud Environment Details',
```
be more consistent since we use title case for `Cloud Environment`

image

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -0,0 +1,404 @@
...
+    [NEW_RSTUDIO_MODE, () => renderNewRuntimeModal(NEW_RSTUDIO_MODE)],
+    [NEW_GALAXY_MODE, renderNewGalaxyModal],
+    [Utils.DEFAULT, renderDefaultPage]
+  )
+
+  const width = Utils.switchCase(viewMode,
+    [NEW_JUPYTER_MODE, () => 675],
+    [NEW_RSTUDIO_MODE, () => 675],
+    [NEW_GALAXY_MODE, () => 675],
+    [Utils.DEFAULT, () => 430]
```
kyuksel, line 366: Why is it narrower for the default case?

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
jdcanas: hmm... I'm actually not sure why it was included in the ContextBar to begin with. cc @rtitle. I'll remove it remove the places I put it in the modal

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
jdcanas: the default case has different content than the NewGalaxy/NewRuntime modals, so a different width is provided for both, with room for other modals of different width in the future

src/components/ContextBar.js
jdcanas: this is a pattern that has existed for a lot of analysis components... I'm guessing it corresponds to when you can't share a workspace, in which case they don't have permission to clone it either. It makes sense semantically, and the clone workspace functionality is gated by the same condition elsewhere

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -374,7 +369,7 @@ export const CloudEnvironmentModal = ({ isOpen, onDismiss, onSuccess, canCompute
   const modalBody = h(Fragment, [
     h(TitleBar, {
       id: titleId,
-      title: 'Cloud environment details',
+      title: 'Cloud Environment details',
```
kyuksel: Can we also capitalize `d` in `details` so the whole title is following title case?

kyuksel: Can we also capitalize `d` in `details` so the whole title is following title case?

src/components/ContextBar.js
```diff
@@ -124,8 +124,9 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
           hover: contextBarStyles.hover,
           tooltipSide: 'left',
           onClick: () => setCloudEnvOpen(!isCloudEnvOpen),
-          ...{ tooltip: 'Environment Configuration', tooltipDelay: 100 },
-          'aria-label': 'Environment Configuration'
+          tooltip: 'Environment Configuration',
+          tooltipDelay: 100,
```
kyuksel, line 128: Can this be removed as well?

kyuksel (approved):
👍🏼  with a minor comment inline and also

1. can the other `tooltipDelay`s (mostly in `ContextBar`) be removed as well?
2. bumping [this comment](https://github.com/DataBiosphere/terra-ui/pull/2564#discussion_r681212351) in case it got lost under the pile of comments  :)

src/components/runtime-common.js
```diff
@@ -145,10 +145,13 @@ export const GalaxyWarning = () => {
   ])
 }
 
-export const GalaxyLaunchButton = ({ app, onClick, ...props }) => {
+export const GalaxyLaunchButton = ({ app, key = app.status, onClick, ...props }) => {
   const cookieReady = Utils.useStore(cookieReadyStore)
+  console.log('in galaxy launch button, status')
+  console.log(app.status)
```
dmohs: Errant debugging?

src/components/ContextBar.js
jdcanas: i want to keep the ones in the contextbar since i didnt add them

PR #2566, opened 2021-07-19
show reference data when only one is added [AS-816]

davidangb:
[AS-816](https://broadworkbench.atlassian.net/browse/AS-816):

> When I start with a new Workspace and add Reference Data, nothing appears until I have selected 2 Reference Data items, at which point both appear.
> 
> Repro steps:
> 1) Create a new Workspace
> 2) Under Data, Reference Data, choose B37Human. It does not appear under Data (even if you refresh the page)
> 3) Add Reference Data hg38.
> 4) Now both Reference Data entries appear.
> 5) If you delete one, they both disappear again from view.

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -354,7 +354,7 @@ const DataTypeSection = ({ title, titleExtras, error, retryFunction, children })
       tooltip: 'Error loading, click to retry.'
     }, [icon('sync', { size: 18 })]) : titleExtras
   ]),
-  children?.length && div({
+  !!children?.length && div({
     style: { display: 'flex', flexDirection: 'column', width: '100%' },
     role: 'list'
   }, [children])
@@ -573,7 +573,7 @@ const WorkspaceData = _.flow(
               onClick: () => setImportingReference(true),
               'aria-haspopup': 'dialog'
             }, [icon('plus-circle', { size: 21 })])
-          }, _.map(type => h(DataTypeButton, {
+          }, [_.map(type => h(DataTypeButton, {
             key: type,
             selected: selectedDataType === type,
             onClick: () => {
@@ -590,7 +590,7 @@ const WorkspaceData = _.flow(
               }
             }, [icon('minus-circle', { size: 16 })])
           }, [type]), _.keys(referenceData)
-          )),
+          )]),
           importingReference && h(ReferenceDataImporter, {
             onDismiss: () => setImportingReference(false),
             onSuccess: () => {


src/pages/workspaces/workspace/Data.js
```diff
@@ -573,7 +573,7 @@ const WorkspaceData = _.flow(
               onClick: () => setImportingReference(true),
               'aria-haspopup': 'dialog'
             }, [icon('plus-circle', { size: 21 })])
-          }, _.map(type => h(DataTypeButton, {
+          }, [_.map(type => h(DataTypeButton, {
```
davidangb, line 576: every other instance of `DataTypeSection` puts their children inside a `[...]`, so I'm doing the same here. I'll also note that every other instance of `DataTypeSection` is mapping over an array of objects, not an array of strings.

src/pages/workspaces/workspace/Data.js
```diff
@@ -354,7 +354,7 @@ const DataTypeSection = ({ title, titleExtras, error, retryFunction, children })
       tooltip: 'Error loading, click to retry.'
     }, [icon('sync', { size: 18 })]) : titleExtras
   ]),
-  children?.length && div({
+  !!children?.length && div({
```
davidangb, line 357: without this, the Reference Data children displays as "0" (literally the string "0") … let's make sure this resolves to a boolean.

petesantos (approved):
LGTM. Tested locally - was able to reproduce on dev branch, saw the behavior of the code change fix the issue.

ansingh7115 (approved):

src/pages/workspaces/workspace/Data.js
cahrens: @davidangb can you explain how "!!" converts the string to a boolean? I haven't seen this syntax before (nor have I seen "children?.length"). Thanks!

davidangb: The `!!` is a terse (but common) way of saying "cast to Boolean". The first `!` casts truthy/falsy values to Boolean, and then returns the `NOT` of that value; the second `!` adds another `NOT`, so the two `NOT`s cancel each other out and the end result is the original value, but cast to Boolean. Here's a reference:
https://betterprogramming.pub/javascript-bang-bang-i-shot-you-down-use-of-double-bangs-in-javascript-7c9d94446054

The `.?` is a null-safe way of referencing values. Here's a reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

cahrens: @davidangb Thanks for the references! The optional chaining is really cool-- I've been stuck writing "old JavaScript" for various reasons, and it's exciting to actually start learning about these newer features.

PR #2567, opened 2021-07-21
[IA-2871] Increase default disk size to 60GB for both GCE and Dataproc

kyuksel:
Dataproc main and worker machines' disk sizes need to be 60GB at the minimum or they fail to be created. Please see [IA-2871](https://broadworkbench.atlassian.net/browse/IA-2871) for more details.

This PR
- sets the default disk size for Dataproc main and worker machines to 60GB (instead of 50GB)
- doesn't allow selection of less than 60GB for Dataproc main and worker machines
- doesn't change anything for GCE VMs
- fixes a bug where we were over-estimating the cost of Dataproc clusters (see [inline comment](https://github.com/DataBiosphere/terra-ui/pull/2567/files#r674416045))
- renames some of the constants and a component to clarify their usage

Tested by checking that the following are correct for a GCE VM, a Dataproc cluster with a single node and a Dataproc cluster with two workers:

1. the default persistent and non-persistent disk sizes are as expected when no cloud compute exists
2. upon creation of those computes, the UI displays the expected values on the modal
3. the Leo DB has the expected values

### DIFF
--- a/src/components/NewRuntimeModal.js
+++ b/src/components/NewRuntimeModal.js
@@ -21,7 +21,10 @@ import { withErrorReporting } from 'src/libs/error'
 import Events, { extractWorkspaceDetails } from 'src/libs/events'
 import { versionTag } from 'src/libs/logos'
 import {
-  currentRuntime, DEFAULT_DISK_SIZE, DEFAULT_GPU_TYPE, DEFAULT_NUM_GPUS, defaultDataprocMachineType, defaultGceMachineType, displayNameForGpuType,
+  currentRuntime, DEFAULT_DATAPROC_DISK_SIZE, DEFAULT_GCE_BOOT_DISK_SIZE, DEFAULT_GCE_PERSISTENT_DISK_SIZE, DEFAULT_GPU_TYPE, DEFAULT_NUM_GPUS,
+  defaultDataprocMachineType,
+  defaultGceMachineType,
+  displayNameForGpuType,
   findMachineType,
   getDefaultMachineType, getValidGpuTypes,
   persistentDiskCostMonthly,
@@ -86,13 +89,13 @@ const WorkerSelector = ({ value, machineTypeOptions, onChange }) => {
   ])
 }
 
-const DiskSelector = ({ value, onChange }) => {
+const DataprocDiskSelector = ({ value, onChange }) => {
   return h(IdContainer, [
     id => h(Fragment, [
       label({ htmlFor: id, style: styles.label }, ['Disk size (GB)']),
       h(NumberInput, {
         id,
-        min: 60, // sizes below this are too small for images
+        min: 60, // less than this size causes failures in cluster creation
         max: 64000,
         isClearable: false,
         onlyInteger: true,
@@ -139,16 +142,17 @@ export class NewRuntimeModalBase extends Component {
     const runtimeConfig = runtime?.runtimeConfig
     const gpuConfig = runtimeConfig?.gpuConfig
     const sparkMode = runtimeConfig?.cloudService === cloudServices.DATAPROC ? (runtimeConfig.numberOfWorkers === 0 ? 'master' : 'cluster') : false
+    const isDataproc = !sparkMode && !runtimeConfig?.diskSize
 
     return {
-      selectedPersistentDiskSize: disk?.size || DEFAULT_DISK_SIZE,
+      selectedPersistentDiskSize: disk?.size || DEFAULT_GCE_PERSISTENT_DISK_SIZE,
       sparkMode,
       masterMachineType: runtimeConfig?.masterMachineType || runtimeConfig?.machineType,
-      masterDiskSize: runtimeConfig?.masterDiskSize || runtimeConfig?.diskSize || DEFAULT_DISK_SIZE,
+      masterDiskSize: runtimeConfig?.masterDiskSize || runtimeConfig?.diskSize || (isDataproc ? DEFAULT_DATAPROC_DISK_SIZE : DEFAULT_GCE_BOOT_DISK_SIZE),
       numberOfWorkers: runtimeConfig?.numberOfWorkers || 2,
       numberOfPreemptibleWorkers: runtimeConfig?.numberOfPreemptibleWorkers || 0,
       workerMachineType: runtimeConfig?.workerMachineType || defaultDataprocMachineType,
-      workerDiskSize: runtimeConfig?.workerDiskSize || DEFAULT_DISK_SIZE,
+      workerDiskSize: runtimeConfig?.workerDiskSize || DEFAULT_DATAPROC_DISK_SIZE,
       gpuEnabled: (!!gpuConfig && !sparkMode) || false,
       hasGpu: !!gpuConfig,
       gpuType: gpuConfig?.gpuType || DEFAULT_GPU_TYPE,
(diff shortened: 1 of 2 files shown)


rtitle:
Can we increase it just for Dataproc? I think GCE works fine with 50G. The limitation is really just for the Dataproc image.

Qi77Qi:
yea..the 50G for gce is just the PD size..it's not the boot disk size...

src/libs/runtime-utils.js
```diff
@@ -10,7 +10,7 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
 
-export const DEFAULT_DISK_SIZE = 50
+export const DEFAULT_DISK_SIZE = 60
 export const DEFAULT_BOOT_DISK_SIZE = 50
```
rtitle: The 60G limitation may actually be with the boot disk, so I wonder if that const needs to change instead.

Actually I'm confused how `DEFAULT_DISK_SIZE` and `DEFAULT_BOOT_DISK_SIZE` are actually used in the UI. Only diskSize should actually be changeable by the user -- the boot disk should be constant. Maybe will look at UI code a bit more to understand better.

kyuksel: Oh yea I see - I'll fix it

src/libs/runtime-utils.js
```diff
@@ -63,13 +64,14 @@ const dataprocCost = (machineType, numInstances) => {
...
+  } = normalizeRuntimeConfig(config)
+
+  const isDataproc = cloudService === cloudServices.DATAPROC
 
   return _.sum([
     (masterDiskSize + numberOfWorkers * workerDiskSize) * storagePrice,
-    cloudService === cloudServices.DATAPROC && (dataprocCost(masterMachineType, 1) + dataprocCost(workerMachineType, numberOfWorkers)),
-    bootDiskSize * storagePrice
+    isDataproc && (dataprocCost(masterMachineType, 1) + dataprocCost(workerMachineType, numberOfWorkers)),
+    !isDataproc && (bootDiskSize * storagePrice)
```
kyuksel: I think this was a bug. I've added the `!isDataproc` condition because for `Dataproc`, the line above (`(masterDiskSize + numberOfWorkers * workerDiskSize) * storagePrice`) already accounts for all disk costs.

src/components/NewRuntimeModal.js
```diff
@@ -139,16 +142,17 @@ export class NewRuntimeModalBase extends Component {
...
     const sparkMode = runtimeConfig?.cloudService === cloudServices.DATAPROC ? (runtimeConfig.numberOfWorkers === 0 ? 'master' : 'cluster') : false
+    const isDataproc = !sparkMode && !runtimeConfig?.diskSize
 
     return {
-      selectedPersistentDiskSize: disk?.size || DEFAULT_DISK_SIZE,
+      selectedPersistentDiskSize: disk?.size || DEFAULT_GCE_PERSISTENT_DISK_SIZE,
       sparkMode,
       masterMachineType: runtimeConfig?.masterMachineType || runtimeConfig?.machineType,
-      masterDiskSize: runtimeConfig?.masterDiskSize || runtimeConfig?.diskSize || DEFAULT_DISK_SIZE,
+      masterDiskSize: runtimeConfig?.masterDiskSize || runtimeConfig?.diskSize || (isDataproc ? DEFAULT_DATAPROC_DISK_SIZE : DEFAULT_GCE_BOOT_DISK_SIZE),
```
Qi77Qi, line 146: hmm..this is a bit awkward. If it is gce, `masterDiskSize` means boot disk size? we don't intend to let user set bootDisk size, so we probably don't need to care for GCE (probably should just get rid of it in Leo)...

src/libs/runtime-utils.js
```diff
@@ -10,8 +10,9 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
 
-export const DEFAULT_DISK_SIZE = 50
-export const DEFAULT_BOOT_DISK_SIZE = 50
+export const DEFAULT_DATAPROC_DISK_SIZE = 60 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
+export const DEFAULT_GCE_BOOT_DISK_SIZE = 70 // GCE boot disk size is not customizable by users. We use this for cost estimate calculations only.
```
Qi77Qi, line 14: for cost estimation, we use the `bootDiskSize` returned from Leo...we shouldn't use the default defined in UI...or is this just a default if there's nothing returned form Leo?

kyuksel: Yea we display cost estimates when user has no compute too so we need that info. It's not ideal in that we need to remember to change it if the default changes in Leo. The only way around that would be to get that info from a Leo API.

src/components/NewRuntimeModal.js
kyuksel: Yea it is tough to follow too. `masterDiskSize` ends up being assigned the value of
-  `masterDiskSize` returned from Leo for Dataproc clusters
- `diskSize` returned from Leo for legacy GCE VMs without PDs

If we stop supporting GCE VMs without PDs, I think it can be simplified as

```javascript
masterDiskSize: runtimeConfig?.masterDiskSize || DEFAULT_DATAPROC_DISK_SIZE
```

rtitle (approved):
Logic was a little tricky at first but I think I get it. Yeah it would simplify things if we could stop supporting GCE VMs without PDs.

Also tested locally creating GCE and Dataproc runtimes 👍

src/libs/runtime-utils.js
zarsky-broad: Would this be clearer?
```js
isDataproc ?
  (dataprocCost(masterMachineType, 1) + dataprocCost(workerMachineType, numberOfWorkers)) :
  (bootDiskSize * storagePrice)
```

src/components/NewRuntimeModal.js
```diff
@@ -21,7 +21,10 @@ import { withErrorReporting } from 'src/libs/error'
 import Events, { extractWorkspaceDetails } from 'src/libs/events'
 import { versionTag } from 'src/libs/logos'
 import {
-  currentRuntime, DEFAULT_DISK_SIZE, DEFAULT_GPU_TYPE, DEFAULT_NUM_GPUS, defaultDataprocMachineType, defaultGceMachineType, displayNameForGpuType,
+  currentRuntime, DEFAULT_DATAPROC_DISK_SIZE, DEFAULT_GCE_BOOT_DISK_SIZE, DEFAULT_GCE_PERSISTENT_DISK_SIZE, DEFAULT_GPU_TYPE, DEFAULT_NUM_GPUS,
+  defaultDataprocMachineType,
+  defaultGceMachineType,
+  displayNameForGpuType,
```
zarsky-broad:
```suggestion
  defaultDataprocMachineType, defaultGceMachineType, displayNameForGpuType,
```

src/libs/runtime-utils.js
```diff
@@ -63,13 +64,14 @@ const dataprocCost = (machineType, numInstances) => {
 export const runtimeConfigBaseCost = config => {
   const {
     cloudService, masterMachineType, masterDiskSize, numberOfWorkers, workerMachineType, workerDiskSize, bootDiskSize
-  } = normalizeRuntimeConfig(
-    config)
+  } = normalizeRuntimeConfig(config)
+
+  const isDataproc = cloudService === cloudServices.DATAPROC
```
ehanna4, line 69: Here the logic for `isDataproc` utilizes `cloudservices` but in NewRuntimeModal the logic is `const isDataproc = !sparkMode && !runtimeConfig?.diskSize` are these supposed to be different?

zarsky-broad (approved):

src/libs/runtime-utils.js
kyuksel: Yea, they are intended to be `const`s local to the functions they are defined in with different meanings.

PR #2568, opened 2021-07-21
[QA-1528][IA-2872] adding fixes for kernel error and other python 2 update

xuscrbw:
<!--
Hello, friend!
Thanks!
--!>
Updated the Python 2 notebook use to Python 3 (Looks like we don't support the Python 2 Kernel anymore)

There was a subsequent bug where the println wouldn't execute if the kernel wasn't ready. I added a wait for the Kernel Idle button in the frame (top right). I had to add a 10 second more delay to allow for slow runtime creation. 

Would love for a second opinion on how to do this delay.

Also, removed the option to create Python 2 notebook in the UI. It looks like there's some Python 2 language in the repo, but going to leave it for now

### DIFF
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -47,11 +47,11 @@ const testRunNotebookFn = _.flow(
 
   const frame = await findIframe(page)
   await findElement(frame, '//*[@title="Kernel Idle"]')
-  //Delay here to allow runtime to finish creating.
-  await delay(10000)
   await fillIn(frame, '//textarea', 'print(123456789099876543210990+9876543219)')
   await click(frame, clickable({ text: 'Run' }))
   await findText(frame, '123456789099886419754209')
+  // Save notebook to avoid "unsaved changes" modal when test tear-down tries to close the window
+  await click(frame, clickable({ text: 'Save and Checkpoint' }))
 })
 const testRunNotebook = {
   name: 'run-notebook',
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -22,7 +22,7 @@ const testRunNotebookFn = _.flow(
   await waitForNoSpinners(page)
   await click(page, clickable({ textContains: 'Create a' }))
   await fillIn(page, input({ placeholder: 'Enter a name' }), notebookName)
-  await select(page, 'Language', 'Python 2')
+  await select(page, 'Language', 'Python 3')
   await click(page, clickable({ text: 'Create Notebook' }))
   await click(page, clickable({ textContains: notebookName }))
   await click(page, clickable({ text: 'Edit' }))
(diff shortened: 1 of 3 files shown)


Qi77Qi (commented):
lgtm

rtitle (approved):
Looks good.. btw I think we should also remove `Python 2` from the Language drop-drown in the UI since it's not supported in Leo. I made this ticket: https://broadworkbench.atlassian.net/browse/IA-2872. Could be a separate PR though, I think.

xuscrbw:
Removed the Python 2 code in the UI.

rtitle (approved):
:shipit: !

integration-tests/tests/run-notebook.js
```diff
@@ -46,6 +46,9 @@ const testRunNotebookFn = _.flow(
   await findElement(page, clickable({ textContains: 'Running' }), { timeout: 10 * 60 * 1000 })
 
   const frame = await findIframe(page)
+  await findElement(frame, '//*[@title="Kernel Idle"]')
+  //Delay here to allow runtime to finish creating.
+  await delay(10000)
```
petesantos: What are your thoughts on correcting this without adding a delay?

breilly2: I would love if we didn't have to add a delay. I know that's tough sometimes. I'll think on it and let you know if I come up with anything by tomorrow morning.

xuscrbw: I was trying to verify if there was a visual cue for when things are ready. But I don't think I was able to see one. I don't think it is reliant on autosave duration. Maybe it could be the Cloud Environment area in the top right.

breilly2: From what I can tell, waiting for `Kernel Idle` is taking care of making sure the runtime is ready, so that's good. However, when the test is shutting down and trying to close the window, it will display a modal if there are unsaved changes to the notebook. Without the 10 second delay, the modal blocks closing the window which causes the test to fail... and in my local runs, I don't even get a stack trace! With the 10 second delay, the test timing is shifted enough that an auto-save happens between running and closing the window during test tear-down. (I found that the 10 second delay was also effective even if moved to after running the notebook.)

I made a PR (#2569) into this branch with a change that replaces the 10 second delay with a click on the save button. (Since this PR isn't going to dev, it shouldn't have the 2-reviewer requirement. It'll ultimately be reviewed as part of this PR if you choose to merge it to this branch.)

xuscrbw:
Looks like there’s an integration test still failing. Will take a look at this later today to see if this is related to any of the changes we made.

integration-tests/tests/run-notebook.js
xuscrbw: Merged your PR. Thanks!

xuscrbw:
It looks like all tests passed! I think this should be good to merge once we get one more thumbs up!

petesantos (approved):
looks good - thanks!

PR #2575, opened 2021-08-02
[CA-1459] Rollback change to google project

s-rubenstein:
This change was actually due to a backwards incompatible change in rawls. We are going to roll it back here, deprecate the old field in rawls, and make the change after PPW release.

<!--
Hello, friend!
Thanks!
--!>


dmohs (approved):

andy7i (approved):
nit: (no need to change anything for this PR) I think it would be nice to use git revert next time since it is explicitly a rollback, and also guarantees that we don't miss anything when we want to roll back a change.

PR #2579, opened 2021-08-09
unstyled search bar

tcjiang9:
manual testing

### DIFF
--- a/src/pages/library/DataBrowser.js
+++ b/src/pages/library/DataBrowser.js
@@ -21,7 +21,7 @@ const DataBrowser = () => {
     libraryTopMatter('browse & explore'),
     h(DelayedSearchInput, {
       style: { flex: 1, marginLeft: '1rem' },
-      'aria-label': 'Search Featured Data Collections',
+      'aria-label': 'Search Data Collections',
       placeholder: 'Search Name or Description',
       value: searchFilter,
       onChange: setSearchFilter
--- a/src/pages/library/DataBrowser.js
+++ b/src/pages/library/DataBrowser.js
@@ -4,7 +4,7 @@ import { h } from 'react-hyperscript-helpers'
 import FooterWrapper from 'src/components/FooterWrapper'
 import { DelayedSearchInput } from 'src/components/input'
 import { libraryTopMatter } from 'src/components/library-common'
-import {textMatch} from "src/libs/utils";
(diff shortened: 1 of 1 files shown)


dmohs (commented):
Fix your lint and then I'll take a look.

src/pages/library/DataBrowser.js
```diff
@@ -1,11 +1,35 @@
...
 const DataBrowser = () => {
+  const items = ['test1', 'test2']
+
+  const [searchFilter, setSearchFilter] = useState()
+
+  const filterBySearch = items => {
+    const lowerSearch = _.toLower(searchFilter)
+    return _.isEmpty(lowerSearch) ?
+      items :
+      _.filter(item => _.includes(lowerSearch, item), items)
```
dmohs:
```suggestion
    return _.isEmpty(lowerSearch) ? items : _.filter(item => _.includes(lowerSearch, item), items)
```
Optional. We value concision, so separating out these lines might not be worth doing. This is a very straightforward concept, so it should probably take up as little space as possible.

src/pages/library/DataBrowser.js
```diff
@@ -1,11 +1,35 @@
...
+  }
+
+  const filteredItems = filterBySearch(items)
+
```
dmohs:
```suggestion
```

src/pages/library/DataBrowser.js
```diff
@@ -1,11 +1,35 @@
...
-    libraryTopMatter('browse & explore')
+    libraryTopMatter('browse & explore'),
+    h(DelayedSearchInput, {
+      style: { flex: 1, marginLeft: '1rem' },
+      'aria-label': 'Search Featured Workspaces',
+      placeholder: 'Search Name or Description',
+      value: searchFilter,
+      onChange: setSearchFilter
+    }),
+    h(Fragment, filteredItems)
```
dmohs:
```suggestion
    h(Fragment, filterBySearch(items))
```
No reason to force the reader to look up how this is defined, better to show them.

dmohs (approved):
Approved with suggestions.

src/pages/library/DataBrowser.js
dmohs: Actually, Utils.textMatch might be more appropriate here. Please take a look.

pshapiro4broad (approved):

PR #2580, opened 2021-08-10
[IA-2656] Create analysis flow additions

jdcanas:
This PR adds some state to NewAnalysisModal to guide users through creating/uploading an artifact for their analysis. It will also prompt the user to create an environment if none exists to run their analysis. If no environment for an analysis (i.e. Jupyter for a notebook) exists  at creation time and they arent able to create one (i.e. they have an Rstudio runtime in a non-deletable status) we will allow them to create the artifact, but warn them they dont have an applicable env to run. 

Initial page
image

Creating a notebook
image

Creating an rmd
image

Creating an rmd when I have a creating (non-deletable) Jupyter env
image

### DIFF
--- a/src/components/NewAnalysisModal.js
+++ b/src/components/NewAnalysisModal.js
@@ -1,37 +1,107 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
-import { div, h, hr, img } from 'react-hyperscript-helpers'
-import { spinnerOverlay } from 'src/components/common'
+import { div, h, h2, h3, hr, img } from 'react-hyperscript-helpers'
+import { ButtonPrimary, IdContainer, Select, spinnerOverlay } from 'src/components/common'
 import ModalDrawer from 'src/components/ModalDrawer'
 import { NewGalaxyModalBase } from 'src/components/NewGalaxyModal'
 import { NewRuntimeModalBase } from 'src/components/NewRuntimeModal'
-import { tools } from 'src/components/notebook-utils'
+import {
+  analysisNameInput,
+  analysisNameValidator,
+  getDisplayName,
+  NotebookCreator, notebookData,
+  tools
+} from 'src/components/notebook-utils'
 import TitleBar from 'src/components/TitleBar'
 import galaxyLogo from 'src/images/galaxy-logo.png'
 import jupyterLogoLong from 'src/images/jupyter-logo-long.png'
 import rstudioLogo from 'src/images/rstudio-logo.svg'
+import { Ajax } from 'src/libs/ajax'
+import { getUser } from 'src/libs/auth'
 import colors from 'src/libs/colors'
-import { withErrorReporting } from 'src/libs/error'
+import { reportError, withErrorReporting } from 'src/libs/error'
 import * as Utils from 'src/libs/utils'
+import { getConvertedRuntimeStatus, getCurrentApp, getCurrentRuntime } from 'src/libs/runtime-utils'
+import { FormLabel } from 'src/libs/forms'
+import validate from 'validate.js'
 
 //TODO: title id in titleBar and aria-labelled-by in withModalDrawer
 
 const titleId = 'new-analysis-modal-title'
 
 export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
-  ({ isOpen, onDismiss, onSuccess, runtimes, apps, galaxyDataDisks, refreshRuntimes, refreshApps, workspace, persistentDisks, workspace: { workspace: { namespace, bucketName, name: workspaceName } } }) => {
+  ({ isOpen, onDismiss, onSuccess, runtimes, apps, galaxyDataDisks, refreshRuntimes, refreshApps, refreshAnalyses, analyses, workspace, persistentDisks, workspace: { workspace: { namespace, bucketName, name: workspaceName } } }) => {
     const [viewMode, setViewMode] = useState(undefined)
     const [busy, setBusy] = useState()
 
-    const renderNewRuntimeModal = tool => h(NewRuntimeModalBase, {
-      isOpen: viewMode === NEW_JUPYTER_MODE || viewMode === NEW_RSTUDIO_MODE,
+
+    const [notebookKernel, setNotebookKernel] = useState('python3')
+    const [nameTouched, setNameTouched] = useState(false)
+    const [analysisName, setAnalysisName] = useState('')
+    const currentRuntime = getCurrentRuntime(runtimes)
+    const currentRuntimeTool = currentRuntime?.labels?.tool
+    const currentApp = getCurrentApp(apps)
+
+    const [currentTool, setCurrentTool] = useState(undefined)
+
+    const NEW_ANALYSIS_MODE = 'NEW_ARTIFACT'
+    const NEW_ENVIRONMENT_MODE = 'NEW_ENVIRONMENT'
+
+    const resetView = () => {
+      setViewMode(undefined)
+      setCurrentTool(undefined)
+    }
+
+    console.log('viewmode')
+    console.log(viewMode)
+
+    const setNextViewMode = () => {
+      const doesCloudEnvForToolExist = currentRuntimeTool === currentTool || (currentApp && currentTool === tools.galaxy.label)
+
+      console.log('in setNextViewMode, currentTool')
+      console.log(currentTool)
+      return Utils.switchCase(viewMode,
+        [NEW_ANALYSIS_MODE, () => {
+          if (doesCloudEnvForToolExist) {
+            resetView()
+            onSuccess()
+          } else {
+            //TODO, if (current runtime exists && isNotDeletable), if before is true dont take them to new env creation because they cannot create one at this time. Should this be handled in the modal itself?
+            setViewMode(NEW_ENVIRONMENT_MODE)
+          }
+        }],
+        [NEW_ENVIRONMENT_MODE, () => resetView()],
+        [Utils.DEFAULT, () => currentTool === tools.galaxy.label ? setViewMode(NEW_ENVIRONMENT_MODE) : setViewMode(NEW_ANALYSIS_MODE)]
+      )
+    }
+
+    const setPreviousViewMode = () => Utils.switchCase(viewMode,
+      [NEW_ANALYSIS_MODE, () => setViewMode(undefined)],
+      [NEW_ENVIRONMENT_MODE, () => setViewMode(undefined)],
+      [Utils.DEFAULT, () => undefined]
+    )
+
+
+    const getView = () => Utils.switchCase(viewMode,
+      [NEW_ANALYSIS_MODE, renderCreateAnalysis],
+      [NEW_ENVIRONMENT_MODE, getNewEnvironmentView],
+      [Utils.DEFAULT, renderSelectAnalysisBody])
+
+    const getNewEnvironmentView = () => Utils.switchCase(currentTool,
+      [tools.Jupyter.label, renderNewRuntimeModal],
+      [tools.RStudio.label, renderNewRuntimeModal],
+      [tools.galaxy.label, renderNewGalaxyModal]
+    )
+
+    const renderNewRuntimeModal = () => h(NewRuntimeModalBase, {
+      isOpen: currentTool === tools.Jupyter.label || currentTool === tools.RStudio.label,
       isAnalysisMode: true,
       workspace,
-      tool,
+      tool: currentTool,
       runtimes,
       persistentDisks,
       onDismiss: () => {
-        setViewMode(undefined)
+        resetView()
         onDismiss()
       },
       onSuccess: _.flow(
@@ -45,7 +115,7 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
     })
 
     const renderNewGalaxyModal = () => h(NewGalaxyModalBase, {
-      isOpen: viewMode === NEW_GALAXY_MODE,
+      isOpen: viewMode === tools.galaxy.label,
       isAnalysisMode: true,
       workspace,
       apps,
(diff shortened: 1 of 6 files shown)


kyuksel:
@jdcanas Could you please reformat the code following the steps listed on the [README](https://github.com/DataBiosphere/terra-ui/blob/54b6c278ae880d79c8cd8e1f8a4d5bc714c1d835/README.md)?


> In IntelliJ:
> When you open the project, right-click .eslintrc.js, click Apply Eslint Code Style Rules, and select Automatic ESLint Configuration; then go to Preferences -> Editor -> Code Style -> Javascript, click the gear next to Scheme -> Import Scheme -> Intellij IDEA code style XML, then select js-style.xml.

dmohs:
Taking a look. Love the screenshots for context.

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,112 @@
...
+    const currentRuntime = getCurrentRuntime(runtimes)
+    const currentRuntimeTool = currentRuntime?.labels?.tool
+    const currentApp = getCurrentApp(apps)
+    const [currentTool, setCurrentTool] = useState(undefined)
 
-    const renderNewRuntimeModal = tool => h(NewRuntimeModalBase, {
-      isOpen: viewMode === NEW_JUPYTER_MODE || viewMode === NEW_RSTUDIO_MODE,
+
+    const NEW_ANALYSIS_MODE = 'NEW_ARTIFACT'
+    const NEW_ENVIRONMENT_MODE = 'NEW_ENVIRONMENT'
```
petesantos: Since these are constants, could these be moved outside of the render function?

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,112 @@
...
+  ({ isOpen, onDismiss, onSuccess, uploadFiles, openUploader, runtimes, apps, galaxyDataDisks, refreshRuntimes, refreshApps, refreshAnalyses, analyses, workspace, persistentDisks, workspace: { workspace: { namespace, bucketName, name: workspaceName } } }) => {
     const [viewMode, setViewMode] = useState(undefined)
     const [busy, setBusy] = useState()
+    const [notebookKernel, setNotebookKernel] = useState('python3')
+    const [nameTouched, setNameTouched] = useState(false)
+    const [analysisName, setAnalysisName] = useState('')
+    const currentRuntime = getCurrentRuntime(runtimes)
+    const currentRuntimeTool = currentRuntime?.labels?.tool
+    const currentApp = getCurrentApp(apps)
+    const [currentTool, setCurrentTool] = useState(undefined)
```
petesantos, line 36: Could you move this up with the other states, keeping all of the state grouped together in a block and the derived state in the next section?

src/components/NewAnalysisModal.js
```diff
@@ -64,33 +139,120 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
...
-    const NEW_GALAXY_MODE = tools.galaxy.label
-    const getView = () => Utils.switchCase(viewMode,
-      [NEW_JUPYTER_MODE, () => renderNewRuntimeModal(NEW_JUPYTER_MODE)],
-      [NEW_RSTUDIO_MODE, () => renderNewRuntimeModal(NEW_RSTUDIO_MODE)],
-      [NEW_GALAXY_MODE, renderNewGalaxyModal],
-      [Utils.DEFAULT, renderCreateAnalysisBody]
-    )
-
     const toolCardStyles = { backgroundColor: 'white', borderRadius: 5, padding: '1rem', display: 'inline-block', verticalAlign: 'middle', marginBottom: '1rem', textAlign: 'center', width: '100%', height: 60 }
     const imageStyles = { verticalAlign: 'middle', height: '100%', width: '40%' }
```
petesantos: For the these styles, could you add a `styles` object at the top of the file with the different styles that will be used throughout? Something like:

```js
const styles = {
  toolCard: {
    backgroundColor: 'white', borderRadius: 5, padding: '1rem', display: 'inline-block', verticalAlign: 'middle', marginBottom: '1rem',
    textAlign: 'center', width: '100%', height: 60
  },
  image: { verticalAlign: 'middle', height: '100%', width: '40%' }
}

File: src/components/NewAnalysisModal.js [hunk 4]
```diff
@@ -64,33 +139,120 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
...
+        }, style: { marginTop: '1rem', fontSize: 16, lineHeight: '20px', ...Style.elements.card.container, alignItems: 'center', width: '100%', height: 150, backgroundColor: colors.dark(0.1), border: `1px dashed ${colors.dark(0.7)}`, boxShadow: 'none' }
+      }, [
+        div(['Or Click / Drag to upload an analysis file']),
+        icon('upload-cloud', { size: 25, style: { opacity: 0.4, marginTop: '0.5rem' } })
+      ])])
+    ])
+
+    const getArtifactLabel = toolLabel => Utils.switchCase(toolLabel, [tools.RStudio.label, () => 'R markdown file'],
+      [tools.Jupyter.label, () => 'notebook'],
+      [Utils.DEFAULT, () => console.error(`Should not be calling getArtifactLabel for ${toolLabel}, artifacts not implemented`)])
```
petesantos: Having the first case on its own line is a bit clearer to read:

```suggestion
  const getArtifactLabel = toolLabel => Utils.switchCase(toolLabel,
    [tools.RStudio.label, () => 'R markdown file'],
    [tools.Jupyter.label, () => 'notebook'],
    [Utils.DEFAULT, () => console.error(`Should not be calling getArtifactLabel for ${toolLabel}, artifacts not implemented`)]
  )
```

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,112 @@
...
+    const NEW_ENVIRONMENT_MODE = 'NEW_ENVIRONMENT'
+
+    const resetView = () => {
+      setViewMode(undefined)
+      setAnalysisName('')
+      setCurrentTool(undefined)
+      setNotebookKernel('python3')
+    }
+
+    const setNextViewMode = (currentTool, baseViewMode = viewMode) => {
```
petesantos: Since this is not a react state setter can we rename this?
```suggestion
    const goToNextView= (currentTool, baseViewMode = viewMode) => {
```

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,112 @@
...
+      const doesCloudEnvForToolExist = currentRuntimeTool === currentTool || (currentApp && currentTool === tools.galaxy.label)
```
petesantos, line 53: nit: parens
```suggestion
      const doesCloudEnvForToolExist = (currentRuntimeTool === currentTool) || (currentApp && currentTool === tools.galaxy.label)
```

src/components/NewAnalysisModal.js
```diff
@@ -64,33 +139,120 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
...
-      div({ style: toolCardStyles, onClick: () => setViewMode(NEW_JUPYTER_MODE) }, [img({ src: jupyterLogoLong, style: _.merge(imageStyles, { width: '30%' }) })]),
-      div({ style: toolCardStyles, onClick: () => setViewMode(NEW_RSTUDIO_MODE) }, [img({ src: rstudioLogo, style: imageStyles })]),
-      div({ style: toolCardStyles, onClick: () => setViewMode(NEW_GALAXY_MODE) }, [img({ src: galaxyLogo, style: _.merge(imageStyles, { width: '30%' }) })])
+      div({ style: toolCardStyles, onClick: () => { setCurrentTool(tools.Jupyter.label); setNextViewMode(tools.Jupyter.label) } }, [img({ src: jupyterLogoLong, style: _.merge(imageStyles, { width: '30%' }) })]),
+      div({ style: toolCardStyles, onClick: () => { setCurrentTool(tools.RStudio.label); setNextViewMode(tools.RStudio.label) } }, [img({ src: rstudioLogo, style: imageStyles })]),
+      div({ style: { opacity: currentApp ? '.5' : '1', ...toolCardStyles }, onClick: () => { setCurrentTool(tools.galaxy.label); setNextViewMode(tools.galaxy.label) }, disabled: !currentApp, title: currentApp ? 'You already have a galaxy environment' : '' }, [img({ src: galaxyLogo, style: _.merge(imageStyles, { width: '30%' }) })])
```
petesantos: The `'.5'` and `'1'` values don't need to be strings, they can be numeric. Also please put set `.5` to `0.5` for better clarity

src/components/NewAnalysisModal.js
```diff
@@ -64,33 +139,120 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
...
+        div(['Or Click / Drag to upload an analysis file']),
+        icon('upload-cloud', { size: 25, style: { opacity: 0.4, marginTop: '0.5rem' } })
+      ])])
+    ])
+
+    const getArtifactLabel = toolLabel => Utils.switchCase(toolLabel, [tools.RStudio.label, () => 'R markdown file'],
+      [tools.Jupyter.label, () => 'notebook'],
+      [Utils.DEFAULT, () => console.error(`Should not be calling getArtifactLabel for ${toolLabel}, artifacts not implemented`)])
+
+    const renderCreateAnalysis = () => div({ style: { display: 'flex', flexDirection: 'column', flex: 1, padding: '.5rem 1.5rem 1.5rem 1.5rem' } }, [
```
petesantos: Could you go through the file and set any `.5` values to `0.5`?

src/components/NewAnalysisModal.js
```diff
@@ -64,33 +139,120 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
...
+      h2({ style: { fontWeight: 600, marginBottom: 0 } }, [`Create a new ${getArtifactLabel(currentTool)}`]),
+      renderCreateAnalysisBody(currentTool)
+    ])
+
+    const existingNames = _.map(({ name }) => getDisplayName(name), analyses)
+
+    const errors = validate(
+      { analysisName, notebookKernel },
+      {
+        analysisName: analysisNameValidator(existingNames),
```
petesantos: Since this is used once, in-lining may be cleaner 
```suggestion
        analysisName: _.map(({ name }) => getDisplayName(name), analyses),
```

src/components/NewAnalysisModal.js
```diff
@@ -64,33 +139,120 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
...
+          inputProps: {
+            id, value: analysisName,
+            onChange: v => {
+              setAnalysisName(v)
+              setNameTouched(true)
+            }
+          }
+        })
+      ])]),
+      toolLabel === tools.Jupyter.label && h(IdContainer, [id => h(Fragment, [
```
petesantos: Since this is used multiple times, could you set this to a variable and reuse it?
```js
      toolLabel === tools.Jupyter.label
```

src/components/NewAnalysisModal.js
```diff
@@ -100,8 +262,11 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
         title: 'Select an application',
         titleStyles: _.merge(viewMode === undefined ? {} : { display: 'none' }, { margin: '1.5rem 0 0 1.5rem' }),
```
petesantos: Since merge does a deep merge, and styles are flat, we can avoid that usage.
```suggestion
      titleStyles: { margin: '1.5rem 0 0 1.5rem', display: !!viewMode ? undefined : 'none' },
```

src/components/NewAnalysisModal.js
```diff
@@ -64,33 +139,120 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
...
+          tooltip: Utils.summarizeErrors(errors),
+          onClick: async () => {
+            try {
+              const contents = toolLabel === tools.Jupyter.label ? notebookData[notebookKernel] : '# Starter Rmd file'
+              if (toolLabel === tools.Jupyter.label) {
+                // notebook creation requires special handling of json in request
+                await Ajax().Buckets.notebook(namespace, bucketName, analysisName).create(contents)
+              } else {
+                await Ajax().Buckets.analysis(namespace, bucketName, analysisName, toolLabel).create(contents)
+              }
```
petesantos: Could you make this a ternary?

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -8,8 +8,8 @@ import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/b
 import { withViewToggle } from 'src/components/CardsListToggle'
 import {
```
petesantos: Could you put these on one line. An import per line takes up a lot of vertical space in the file. Could you do the same on line 21?

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,112 @@
...
 import { NewGalaxyModalBase } from 'src/components/NewGalaxyModal'
 import { NewRuntimeModalBase } from 'src/components/NewRuntimeModal'
-import { tools } from 'src/components/notebook-utils'
+import {
+  analysisNameInput,
+  analysisNameValidator,
+  getDisplayName, getTool,
+  notebookData,
+  tools
+} from 'src/components/notebook-utils'
```
petesantos: Could you put these imports on one line, as this takes up a lot of vertical space in the file

src/components/NewAnalysisModal.js
jdcanas: I thought the syntax was a bit odd with awaiting in a ternary... i do think it looks cleaner as a ternary though

src/pages/workspaces/workspace/Analyses.js
jdcanas: ah my IDE likes doing this

src/components/NewAnalysisModal.js
jdcanas: I don't think extra the parens here are necessary. The linting rules are set up to enforce parens in equality expressions insofar as to ensure no ambiguity, but the above expression evaluates with a clear order with the current parens. If I were to remove the latter parens, the linter would complain.

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,106 @@
...
+import { getCurrentApp, getCurrentRuntime, isRuntimeDeletable } from 'src/libs/runtime-utils'
+import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
+import validate from 'validate.js'
 
-//TODO: title id in titleBar and aria-labelled-by in withModalDrawer
 
 const titleId = 'new-analysis-modal-title'
+const NEW_ANALYSIS_MODE = 'NEW_ARTIFACT'
+const NEW_ENVIRONMENT_MODE = 'NEW_ENVIRONMENT'
```
dmohs:
```suggestion
const newAnalysisMode = Symbol('new-artifact')
const newEnvironmentMode = Symbol('new-environment')
```
We owe you some notes about this convention in our contribution guide. Using `Symbol` isn't that important (but it's a nice technique). We are trying to avoid SCREAMING_SNAKE_CASE.

src/components/NewAnalysisModal.js
dmohs: I agree, and prefer to let economy (concision) be the deciding factor when possible.

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,106 @@
...
+          }]
+        )],
+        [NEW_ENVIRONMENT_MODE, () => {
+          resetView()
+          onSuccess()
+        }],
+        [Utils.DEFAULT, () => Utils.cond(
+          [currentTool === tools.RStudio.label || currentTool === tools.Jupyter.label, () => setViewMode(NEW_ANALYSIS_MODE)],
+          [currentTool === tools.galaxy.label && !currentApp, () => setViewMode(NEW_ENVIRONMENT_MODE)],
+          [currentTool === tools.galaxy.label && currentApp, () => resetView()] //This shouldn't be possible, as the button is disabled in this case
```
dmohs: If your comment is true then wouldn't an assertion be more appropriate? I judge we'd want to find out if the assumption proves false.

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,106 @@
...
+        )]
+      )
+    }
+
+    const setPreviousViewMode = () => Utils.switchCase(viewMode,
+      [NEW_ANALYSIS_MODE, () => resetView()],
+      [NEW_ENVIRONMENT_MODE, () => resetView()],
+      [Utils.DEFAULT, () => undefined]
+    )
```
dmohs: As far as I can tell, viewMode is one of those values or undefined, in which case this function can be simplified to:
```suggestion
    const setPreviousViewMode = () => !!viewMode && resetView()
```
If there is only one usage, it should be inlined for clarity.

src/components/NewAnalysisModal.js
```diff
@@ -64,44 +133,135 @@ export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
...
-      div({ style: toolCardStyles, onClick: () => setViewMode(NEW_JUPYTER_MODE) }, [img({ src: jupyterLogoLong, style: _.merge(imageStyles, { width: '30%' }) })]),
-      div({ style: toolCardStyles, onClick: () => setViewMode(NEW_RSTUDIO_MODE) }, [img({ src: rstudioLogo, style: imageStyles })]),
-      div({ style: toolCardStyles, onClick: () => setViewMode(NEW_GALAXY_MODE) }, [img({ src: galaxyLogo, style: _.merge(imageStyles, { width: '30%' }) })])
+      div({ style: styles.toolCard, onClick: () => { setCurrentTool(tools.Jupyter.label); enterNextViewMode(tools.Jupyter.label) } }, [img({ src: jupyterLogoLong, style: _.merge(styles.image, { width: '30%' }) })]),
+      div({ style: styles.toolCard, onClick: () => { setCurrentTool(tools.RStudio.label); enterNextViewMode(tools.RStudio.label) } }, [img({ src: rstudioLogo, style: styles.image })]),
+      div({ style: { opacity: currentApp ? '0.5' : '1', ...styles.toolCard }, onClick: () => { setCurrentTool(tools.galaxy.label); enterNextViewMode(tools.galaxy.label) }, disabled: !currentApp, title: currentApp ? 'You already have a galaxy environment' : '' }, [img({ src: galaxyLogo, style: _.merge(styles.image, { width: '30%' }) })])
```
dmohs, line 143: Nit: wrap. We're currently debating a line limit of 100 or 120 characters. For now, please just treat 120 as a soft limit.

dmohs (approved):
Approved with minor suggestions.

kyuksel:
Clicked through the cases in the PR description as well as some others. All but one looked good. Please see below for details on that case.

While having the default Jupyter app with the config:

image

I clicked on a previously created RStudio analysis as shown below:

image

which brought up the R markdown preview:

image

When I clicked on `EDIT` on the above page, my browser downloaded the `Rmd` file as shown below:

image

The same downloading happens when I click on `PLAYGROUND MODE` instead of `EDIT` as well.

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,106 @@
...
+const NEW_ANALYSIS_MODE = 'NEW_ARTIFACT'
+const NEW_ENVIRONMENT_MODE = 'NEW_ENVIRONMENT'
 
 export const NewAnalysisModal = Utils.withDisplayName('NewAnalysisModal')(
-  ({ isOpen, onDismiss, onSuccess, runtimes, apps, galaxyDataDisks, refreshRuntimes, refreshApps, workspace, persistentDisks, workspace: { workspace: { namespace, bucketName, name: workspaceName } } }) => {
+  ({ isOpen, onDismiss, onSuccess, uploadFiles, openUploader, runtimes, apps, galaxyDataDisks, refreshRuntimes, refreshApps, refreshAnalyses, analyses, workspace, persistentDisks, workspace: { workspace: { namespace, bucketName, name: workspaceName } } }) => {
     const [viewMode, setViewMode] = useState(undefined)
     const [busy, setBusy] = useState()
+    const [notebookKernel, setNotebookKernel] = useState('python3')
+    const [nameTouched, setNameTouched] = useState(false)
```
kyuksel: I wonder if there's a more self-explanatory name than `nameTouched`. I couldn't be sure what it represents.

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,106 @@
...
+    const currentApp = getCurrentApp(apps)
+
+    const resetView = () => {
+      setViewMode(undefined)
+      setAnalysisName('')
+      setCurrentTool(undefined)
+      setNotebookKernel('python3')
+    }
+
+    const enterNextViewMode = (currentTool, baseViewMode = viewMode) => {
```
kyuksel, line 52: I'm concerned that the function argument `baseViewMode` having a default argument can make it easy to call that method without carefully thinking what value should be passed into `baseViewMode` (the same reason why we try to avoid default function argument values in IA codebases).

src/components/NewAnalysisModal.js
dmohs: I meant to comment here as well. This piece of state feels a bit dirty to me (ha ha, it's a dirty flag, ha ha). I wonder if `usePrevious` with a comparison would make things a bit cleaner.

src/components/NewAnalysisModal.js
jdcanas: I'll put a console.error

jdcanas:
> Clicked through the cases in the PR description as well as some others. All but one looked good. Please see below for details on that case.
> 
> While having the default Jupyter app with the config:
> _..._
> I clicked on a previously created RStudio analysis as shown below:
> _..._
> which brought up the R markdown preview:
> _..._
> When I clicked on `EDIT` on the above page, my browser downloaded the `Rmd` file as shown below:
> _..._
> The same downloading happens when I click on `PLAYGROUND MODE` instead of `EDIT` as well.

The launch functionality for RStudio is not implemented yet despite the preview (calhoun working yet). There are 3 associated tickets, [one](https://broadworkbench.atlassian.net/browse/IA-2653https://broadworkbench.atlassian.net/browse/IA-2653) due to this delocalize ticket for RMDs) for getting delocalize working in welder, [two](https://broadworkbench.atlassian.net/browse/IA-2701) for leveraging that delocalize in the UI and a little bit of this [third](https://broadworkbench.atlassian.net/browse/IA-2702) one for completeness

src/components/NewAnalysisModal.js
jdcanas: Hmm. My thoughts were the default (and currently predominant) use case is to call it with just the tool and no view mode, and have it be smart enough to figure out what should be next. Only in specific cases, where want to go out of the normal use case order due to an edge case and skip around in the modals is purposely passing a view mode applicable. 

I'm of the opinion that, since there's only 1 case where we explicitly hotwire it to an off-order transition, the default should be present and we should use this simplified approach for now. To be honest, this is the culmination of something @petesantos and I discussed a while ago. Tl;dr, a modal wizard that uses an object built of numbered steps representing modals to allow more free transitions. This seems like a good candidate, but I also feel it's not necessarily complex enough to warrant it yet.

kyuksel: > Hmm. My thoughts were the default (and currently predominant) use case is to call it with just the tool and no view mode, and have it be smart enough to figure out what should be next. Only in specific cases, where want to go out of the normal use case order due to an edge case and skip around in the modals is purposely passing a view mode applicable.

Fair enough :)

kyuksel: > To be honest, this is the culmination of something @petesantos and I discussed a while ago. Tl;dr, a modal wizard that uses an object built of numbered steps representing modals to allow more free transitions. This seems like a good candidate, but I also feel it's not necessarily complex enough to warrant it yet.

I'd be curious to know more about this.

src/components/NewAnalysisModal.js
```diff
@@ -1,37 +1,105 @@
...
+        )],
+        [newEnvironmentMode, () => {
+          resetView()
+          onSuccess()
+        }],
+        [Utils.DEFAULT, () => Utils.cond(
+          [currentTool === tools.RStudio.label || currentTool === tools.Jupyter.label, () => setViewMode(newAnalysisMode)],
+          [currentTool === tools.galaxy.label && !currentApp, () => setViewMode(newEnvironmentMode)],
+          [currentTool === tools.galaxy.label && currentApp, () => {
+            console.error('This shouldn\'t be possible, as you aren\'t allowed to create a galaxy analysis when one exists; the button should be disabled')
```
kyuksel: Nit: missing `.`

```suggestion
            console.error('This shouldn\'t be possible, as you aren\'t allowed to create a galaxy analysis when one exists; the button should be disabled.')
```

kyuksel (approved):
👍🏼   with a nit

PR #2583, opened 2021-08-12
Help avoid missing config on deploy

dmohs:
I've had this bite me twice now, so I'm hoping this is strategically placed in a way that would tell me what I was missing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "setenv": "echo REACT_APP_VERSION=$(git rev-parse HEAD) REACT_APP_BUILD_TIMESTAMP=$(date -u \"+%s000\") EXTEND_ESLINT=true",
     "start": "env $(yarn -s setenv) react-app-rewired start",
     "build": "env $(yarn -s setenv) react-app-rewired build",
-    "postbuild": "rm build/config.json",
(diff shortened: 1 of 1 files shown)


zarsky-broad (approved):

package.json
```diff
@@ -62,7 +62,7 @@
     "setenv": "echo REACT_APP_VERSION=$(git rev-parse HEAD) REACT_APP_BUILD_TIMESTAMP=$(date -u \"+%s000\") EXTEND_ESLINT=true",
     "start": "env $(yarn -s setenv) react-app-rewired start",
     "build": "env $(yarn -s setenv) react-app-rewired build",
-    "postbuild": "rm build/config.json",
+    "postbuild": "rm build/config.json && echo && echo '*** Be sure to copy a config/xxx.json to build/config.json. ***'",
```
kyuksel, line 65: [Asking for my edification] How does  ^^^  work?

petesantos (approved):

package.json
dmohs: It's just a message that appears after you run `yarn build`. It'll be sitting pretty close to your prompt in the terminal and might be enough to prevent a mistake.

PR #2589, opened 2021-08-16
update node, yarn, deps

zarsky-broad:
This makes a few changes:
1. Node version is changed to 14, which has been the LTS for a while.
    - No major changes as a result of this. I'm personally using [nvm](https://github.com/nvm-sh/nvm) to manage node versions these days, but it's a bit involved, so you might be happier just using Homebrew and linking the version you want to use at any given moment, which is what the node version check will tell you to do if you're on the wrong version.
2. We change to the latest version of Yarn, which we've been holding off on for a while.
    - New Yarn handles scripts differently, so the `postbuild` script is now explicitly part of the `build` script.
    - New Yarn doesn't allow scripts to log to the console, so the node version check is now a Yarn plugin. The Yarn check script remains as before, as it's only designed to be run by npm.
    - As a bonus, I added a script, `yarn lint`, which will run `eslint src --fix` for you, so you don't have to remember the command.
    - New Yarn doesn't allow implicit dependencies, so we link the version of `eslint` that we already have (in our case specified by `react-scripts`) so that it appears to be a top-level dependency, without actually making it one.
    - For the time being, we're having new Yarn act like old Yarn, in that you have to `yarn install` and it creates a `node_modules` folder. New Yarn actually has a [zero-install mode](https://yarnpkg.com/features/zero-installs), where a compressed version of the yarn cache is actually checked in and you never have to install anything; as soon as you check a branch out you're good to go. We may switch to this in the future.
3. Dependency updates are all basically no-ops.
    - `array-move` changed the style of its export, so our use of it is tweaked to match.
    - While running `optimize-image-svgs` to make sure it was happy, one file was thrashing, so an unused `<g>` was removed, which made the file stable.
    - The new version of Husky, our git hook manager, changes how it's structured, so the pre-commit secret check moved.


app.yaml
```diff
@@ -1,4 +1,4 @@
-runtime: nodejs12
+runtime: nodejs14
```
zarsky-broad, line 1: The runtime doesn't matter, since it's just serving static files; however, we keep this in sync because we have to choose _something_, so it might as well be consistent.

dmohs (approved):

petesantos (approved):

PR #2590, opened 2021-08-18
[IA-2866] Convert NewRuntimeModal to be a function component

kyuksel:
**Note for reviewers:** The only functional (albeit non-user-facing) changes should be in `ComputeModal` (formerly `NewRuntimeModal`). The rest has mostly styling and renaming changes.

### Ticket
https://broadworkbench.atlassian.net/browse/IA-2866

### Locating the Modal
Click on the gear icon on the upper right of corner of the `Notebooks` page below

Screen Shot 2021-08-18 at 3 28 22 PM

which should open the modal drawer below

image

### Highlights
- Replaced `NewRuntimeModal` with its function counterpart named `ComputeModal`. There should be no user-facing difference in functionality. As for `compute` vs. `runtime, we have been favouring the term `compute` over `runtime` to encompass all of the backends including K8s used for apps (Galaxy for now), especially when it's user-facing.
- Removed the redundant state updates that used to be in `NewRuntimeModal`'s `constructor` (via `getInitialState()`).
- Styling improvements
  - Helpers, render functions, etc. are alphabetically listed in their respective sections
  - Consistent casing for defaults/constants
  - Ran `Code Reformatter` on files specific to interactive analysis functionality.
- Renamed `NewGalaxyModal` and `NewAnalysisModal` to remove `New`, which caused confusion for me for a while and feels redundant at best. In fact, since these modals are used for more than just creating new instances, the name had become misleading too.
- Updated `README` with a reference to how to create a macro to streamline code linting and reformatting.

### Out of Scope
- Refactoring of individual functions, other than when it was necessary for the component conversion and/or trivial
- Splitting the very long modal file into multiple

### Testing
- Locally deployed the changes and ran against dev backends
- Went through state transitions for interactive analysis resources, via the current and the revamped UI
- Went through state updates via the new modal


src/libs/runtime-utils.js
```diff
@@ -10,17 +10,21 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
```
kyuksel, line 12: Thought about renaming this file as `compute-utils` but it was going to increase the diff's even more. Could still do it if nudged 😉

src/components/ComputeModal.js
```diff
@@ -0,0 +1,1335 @@
...
+  // State -- end
+
+  const isPersistentDisk = shouldUsePersistentDisk(sparkMode, currentRuntimeDetails, upgradeDiskSelected)
+
+  const isCustomImage = selectedLeoImage === customMode
+  const { version, updated, packages, requiresSpark, label: packageLabel } = _.find({ image: selectedLeoImage }, leoImages) || {}
+
+  const minRequiredMemory = sparkMode ? 7.5 : 3.75
+  const validMachineTypes = _.filter(({ memory }) => memory >= minRequiredMemory, machineTypes)
+  const mainMachineType = _.find({ name: computeConfig?.masterMachineType }, validMachineTypes)?.name || getDefaultMachineType(sparkMode)
```
kyuksel: `computeConfig` state variable should be defined but opted to use the optional chaining operator (`?`) when referencing a property of its in case I was missing something.

src/components/ComputeModal.js
```diff
@@ -0,0 +1,1335 @@
...
+            if (!_.includes(foundImage, filteredNewLeoImages) && isAnalysisMode) {
+              return _.find({ id: tools[tool].defaultImageId }, newLeoImages).image
+            } else {
+              return imageUrl
+            }
+          } else {
+            return customMode
+          }
+        }
+
```
kyuksel, line 561: The states updates below used to be also made in `NewRuntimeModal`'s `constructor` (in addition to `componentDidMount` which is the class component equivalent of `useEffect` that `useOnMount` uses under the hood). This seemed redundant as the modal is mounted quickly. Not sure if it may be undesirable in case of major network latencies.

src/libs/runtime-utils.js
```diff
@@ -10,17 +10,21 @@ import * as Style from 'src/libs/style'
...
 
 
-export const DEFAULT_DATAPROC_DISK_SIZE = 60 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
-export const DEFAULT_GCE_BOOT_DISK_SIZE = 70 // GCE boot disk size is not customizable by users. We use this for cost estimate calculations only.
-export const DEFAULT_GCE_PERSISTENT_DISK_SIZE = 50
+export const defaultDataprocDiskSize = 60 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
+export const defaultGceBootDiskSize = 70 // GCE boot disk size is not customizable by users. We use this for cost estimate calculations only.
+export const defaultGcePersistentDiskSize = 50
```
kyuksel, line 15: I'm personally not used to using camelCase for constants but since that appears to be the consensus for terra-ui, I wanted the code to be consistent.

zarsky-broad (approved):

petesantos (approved):
LGTMob w/ @dmohs, @zarsky-broad

dmohs (approved):
Review comments addressed via mobbing session.

PR #2591, opened 2021-08-23
AS-773: Switch snapshot endpoints to V2

ansingh7115:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/workspace-utils.js
+++ b/src/components/workspace-utils.js
@@ -73,7 +73,7 @@ const SnapshotLabeledInfo = ({ title, text }) => {
 }
 
 export const SnapshotInfo = ({
-  workspace: { workspace, workspace: { namespace, name } }, resource: { resourceId, description, snapshot } , snapshotName,
+  workspace: { workspace, workspace: { namespace, name } }, resource: { resourceId, description, snapshot: snapshotId } , snapshotName,
   onUpdate, onDelete
 }) => {
   // State
@@ -103,7 +103,7 @@ export const SnapshotInfo = ({
   // Lifecycle
   Utils.useOnMount(() => {
     const loadSnapshotInfo = async () => {
-      const snapshotInfo = await Ajax(signal).DataRepo.snapshot(snapshot).details()
+      const snapshotInfo = await Ajax(signal).DataRepo.snapshot(snapshotId).details()
       setSelectedSnapshotInfo(snapshotInfo)
     }
 
@@ -219,7 +219,7 @@ export const SnapshotInfo = ({
             Ajax().Metrics.captureEvent(Events.workspaceSnapshotDelete, {
               ...extractWorkspaceDetails(workspace),
               resourceId,
-              snapshot
+              snapshotId
             })
             onDelete()
           } catch (e) {
--- a/src/components/workspace-utils.js
+++ b/src/components/workspace-utils.js
@@ -73,7 +73,7 @@ const SnapshotLabeledInfo = ({ title, text }) => {
 }
 
 export const SnapshotInfo = ({
-  workspace: { workspace, workspace: { namespace, name } }, resource: { referenceId, description, reference: { snapshot: snapshotId } }, snapshotName,
+  workspace: { workspace, workspace: { namespace, name } }, resource: { resourceId, description, snapshot } , snapshotName,
   onUpdate, onDelete
 }) => {
   // State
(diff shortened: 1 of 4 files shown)


src/components/workspace-utils.js
```diff
@@ -103,7 +103,7 @@ export const SnapshotInfo = ({
   // Lifecycle
   Utils.useOnMount(() => {
     const loadSnapshotInfo = async () => {
-      const snapshotInfo = await Ajax(signal).DataRepo.snapshot(snapshotId).details()
+      const snapshotInfo = await Ajax(signal).DataRepo.snapshot(snapshot).details()
```
dmohs: This looks like a type mismatch to me—providing a value named "snapshot" to a function that names the parameter "snapshotId." It may be worth investigating where this mismatch is introduced and either commenting there to make the intention explicit or fixing the error if that is what has occurred.

src/libs/ajax.js
```diff
@@ -678,12 +678,12 @@ const Workspaces = signal => ({
       },
 
       listSnapshots: async (limit, offset) => {
-        const res = await fetchRawls(`${root}/snapshots?offset=${offset}&limit=${limit}`, _.merge(authOpts(), { signal }))
+        const res = await fetchRawls(`${root}/snapshots/v2?offset=${offset}&limit=${limit}`, _.merge(authOpts(), { signal }))
```
dmohs, line 681: This could be where the type mismatch is introduced. Rawls may be returning a snapshot ID in a field with the label "snapshot." If that were true, I might file a ticket against Rawls and make the correction here to contain the impact.

src/libs/ajax.js
```diff
@@ -678,12 +678,12 @@ const Workspaces = signal => ({
...
         return res.json()
```
dmohs:
```suggestion
        // Why are we doing this? e.g., "This endpoint returns snapshot ID in a field labeled 'snapshot.' Correcting here so usage in dependent code is consistent."
        return _.update('gcpDataRepoSnapshots', _.map(_.update('attributes', a => {...a, snapshotId: a.snapshot})), res.json())
```
I haven't tested this, but I think it might actually work. Allow me to explain.

First, we're using Lodash FP, so the arguments to the `update` method are: path, updateFn, object. You can see the path there and the object is last, but what is that update function? The array at the key `gcpDataRepoSnapshots` is a list of objects with `metadata` and `attributes` keys, so I want to map over the objects and update each one. `map` takes a function and an array.

The inner `update` function modifies the object at the `attributes` key by adding the new key we want. You can see that I am *not* passing a third argument (the object to modify). Since Lodash FP automatically curries that function for me, it becomes a function taking only one argument (the remaining object that I did not pass), which is perfect for passing to `map`. Likewise, the `map` itself is curried and becomes the function applied to the array.

The result is a fairly concise modification to exactly the place I want to modify in the nested structure.

MatthewBemis (approved):
LGTM; did not test

src/pages/workspaces/workspace/Data.js
```diff
@@ -553,7 +553,7 @@ const WorkspaceData = _.flow(
                         setSelectedDataType([snapshotName, tableName])
                         Ajax().Metrics.captureEvent(Events.workspaceSnapshotContentsView, {
                           ...extractWorkspaceDetails(workspace.workspace),
-                          referenceId,
+                          resourceId,
                           snapshotId: snapshot,
```
dmohs:
```suggestion
                          snapshotId,
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -505,7 +505,7 @@ const WorkspaceData = _.flow(
             error: snapshotMetadataError,
             retryFunction: loadSnapshotMetadata
           }, [
-            _.map(([snapshotName, { resource: { referenceId, reference: { snapshot } }, entityMetadata: snapshotTables, error: snapshotTablesError }]) => {
+            _.map(([snapshotName, { resource: { resourceId, snapshot }, entityMetadata: snapshotTables, error: snapshotTablesError }]) => {
```
dmohs:
```suggestion
            _.map(([snapshotName, { resource: { resourceId, snapshotId }, entityMetadata: snapshotTables, error: snapshotTablesError }]) => {
```

src/components/workspace-utils.js
```diff
@@ -73,7 +73,7 @@ const SnapshotLabeledInfo = ({ title, text }) => {
 }
 
 export const SnapshotInfo = ({
-  workspace: { workspace, workspace: { namespace, name } }, resource: { referenceId, description, reference: { snapshot: snapshotId } }, snapshotName,
+  workspace: { workspace, workspace: { namespace, name } }, resource: { resourceId, description, snapshot: snapshotId } , snapshotName,
```
dmohs:
```suggestion
  workspace: { workspace, workspace: { namespace, name } }, resource: { resourceId, description, snapshotId } , snapshotName,
```

dmohs (approved):
A couple of suggestions where we may be able to clean things up now that we fixed the name mismatch at its source. I may be wrong about these, so use your own judgment. Approving either way.

PR #2594, opened 2021-08-25
[CA-1587] Paint Icon for Status of Billing Account Change

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1487

When a user changes the Google Billing Account associated with a Terra Billing Project, Rawls updates the Billing Accounts of the Billing Project's Workspaces' Google Projects asynchronously.
While a user waits for these changes to be made, we will paint an icon next to each workspace that best matches the status of the change. See the mock-up in the ticket for more information.

In this change:
- Add padding to the Billing Project's Workspace table headings for an icon beside each workspace
- Conditionally add an icon beside each workspace based upon the status of the update. If all workspaces have been updated successfully, don't draw any icon.

before
after

### DIFF
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -26,10 +26,17 @@ import { billingRoles } from 'src/pages/billing/List'
 const workspaceLastModifiedWidth = 150
 const workspaceExpandIconSize = 20
 
+const workspaceBillingStatusIcon = (() => {
+  const size = 16
+  const blank = div({ style: { width: size } },
+    [div({ className: 'sr-only' }, ['Status'])])
+  return shape => shape ? icon(shape, { size }) : blank
+})()
+
 const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ sort, onSort }) => {
   return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem', marginBottom: '0.5rem' } }, [
-    // extra left-padding for workspace billing account status icons
-    div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '2rem' } }, [
+    workspaceBillingStatusIcon(null),
+    div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '1rem' } }, [
       h(HeaderRenderer, { sort, onSort, name: 'name' })
     ]),
     div({ 'aria-sort': ariaSort(sort, 'createdBy'), style: { flex: 1 } }, [
@@ -74,8 +81,8 @@ const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, billingA
   return div({ role: 'listitem', style: { ...Style.cardList.longCardShadowless, flexDirection: 'column' } }, [
     h(IdContainer, [id => h(Fragment, [
       div({ style: workspaceCardStyles.row }, [
-        billingAccountStatusIcon && icon(billingAccountStatusIcon, { size: 16 }),
-        div({ style: { ...workspaceCardStyles.field, display: 'flex', alignItems: 'center', paddingLeft: billingAccountStatusIcon ? '1rem' : '2rem' } }, [
+        workspaceBillingStatusIcon(billingAccountStatusIcon),
+        div({ style: { ...workspaceCardStyles.field, display: 'flex', alignItems: 'center', paddingLeft: '1rem' } }, [
           h(Link, {
             style: Style.noWrapEllipsis,
             href: Nav.getLink('workspace-dashboard', { namespace, name }),
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -22,12 +22,17 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 import { billingRoles } from 'src/pages/billing/List'
 
-
+const workspaceStatusIconWidth = 16
 const workspaceLastModifiedWidth = 150
 const workspaceExpandIconSize = 20
 
+const paddingForWorkspaceStatusIcon =
+  div({ style: { width: workspaceStatusIconWidth } },
+  [div({ className: 'sr-only' }, ['Status'])])
+
 const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ sort, onSort }) => {
   return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem', marginBottom: '0.5rem' } }, [
+    paddingForWorkspaceStatusIcon,
     div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '1rem' } }, [
       h(HeaderRenderer, { sort, onSort, name: 'name' })
     ]),
@@ -60,6 +65,10 @@ const ExpandedInfoRow = Utils.memoWithName('ExpandedInfoRow', ({ title, details,
   ])
 })
 
+const statusIcon = () => {
+  return paddingForWorkspaceStatusIcon
+}
+
 const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, isExpanded, onExpand }) => {
   const { namespace, name, createdBy, lastModified, googleProject, billingAccountName } = workspace
   const workspaceCardStyles = {
@@ -73,6 +82,7 @@ const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, isExpand
   return div({ role: 'listitem', style: { ...Style.cardList.longCardShadowless, flexDirection: 'column' } }, [
     h(IdContainer, [id => h(Fragment, [
       div({ style: workspaceCardStyles.row }, [
+        statusIcon(),
         div({ style: { ...workspaceCardStyles.field, display: 'flex', alignItems: 'center', paddingLeft: '1rem' } }, [
           h(Link, {
             style: Style.noWrapEllipsis,
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -1,5 +1,4 @@
 import _ from 'lodash/fp'
-import { any } from 'prop-types'
 import * as qs from 'qs'
 import { Fragment, useEffect, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
@@ -23,17 +22,18 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 import { billingRoles } from 'src/pages/billing/List'
 
+
 const workspaceStatusIconWidth = 16
 const workspaceLastModifiedWidth = 150
 const workspaceExpandIconSize = 20
 
-const paddingForWorkspaceStatusIcon = () =>
+const paddingForWorkspaceStatusIcon =
   div({ style: { width: workspaceStatusIconWidth } },
-  [div({ className: 'sr-only' }, ['Status'])])
+    [div({ className: 'sr-only' }, ['Status'])])
 
 const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ sort, onSort }) => {
   return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem', marginBottom: '0.5rem' } }, [
-    paddingForWorkspaceStatusIcon(),
+    paddingForWorkspaceStatusIcon,
     div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '1rem' } }, [
       h(HeaderRenderer, { sort, onSort, name: 'name' })
     ]),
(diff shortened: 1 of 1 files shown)


marctalbott (approved):

src/pages/billing/Project.js
```diff
@@ -23,11 +23,17 @@ import * as Utils from 'src/libs/utils'
...
 
 
+const workspaceStatusIconWidth = 16
 const workspaceLastModifiedWidth = 150
 const workspaceExpandIconSize = 20
 
+const paddingForWorkspaceStatusIcon =
+  div({ style: { width: workspaceStatusIconWidth } },
+    [div({ className: 'sr-only' }, ['Status'])])
+
```
petesantos: We don't typically add elements for padding purposes only. I have some suggestions below to eliminate this need

src/pages/billing/Project.js
```diff
@@ -23,11 +23,17 @@ import * as Utils from 'src/libs/utils'
...
 const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ sort, onSort }) => {
   return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem', marginBottom: '0.5rem' } }, [
+    paddingForWorkspaceStatusIcon,
     div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '1rem' } }, [
```
petesantos:
```suggestion
    div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '2rem' } }, [
```

src/pages/billing/Project.js
```diff
@@ -60,7 +66,7 @@ const ExpandedInfoRow = Utils.memoWithName('ExpandedInfoRow', ({ title, details,
   ])
 })
 
-const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, isExpanded, onExpand }) => {
+const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, statusIcon, isExpanded, onExpand }) => {
```
petesantos: It looks like we have the information we need to derive the `statusIcon` in this component without having to pass in additional props.
```suggestion
const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, isExpanded, onExpand }) => {
```

src/pages/billing/Project.js
```diff
@@ -145,23 +152,38 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
   const adminCanEdit = _.filter(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers).length > 1
 
+  const workspacesInProject = useMemo(() => _.flow(
+    _.map(({ workspace }) => workspace),
+    _.filter({ namespace: projectName })
+  )(workspaces), [workspaces, projectName])
+
+  const getBillingAccountStatusIcon = (() => {
+    const allWorkspacesHaveCorrectBillingAccount =
+      _.isEmpty(_.remove({ billingAccount: billingAccountName }, workspacesInProject))
```
petesantos: I think this is functionally the same as: 
```javascript
_.includes({ billingAccount: billingAccountName }, workspace)
```
We can directly invoke the `_.includes`

src/pages/billing/Project.js
```diff
@@ -145,23 +152,38 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
+    const getIcon = shape => icon(shape, { size: workspaceStatusIconWidth })
+    return workspace => allWorkspacesHaveCorrectBillingAccount ? paddingForWorkspaceStatusIcon :
+      billingAccountName === workspace.billingAccount ? getIcon('success-standard') :
+        'billingAccountErrorMessage' in workspace ? getIcon('error-standard') :
+          getIcon('loadingSpinner')
+  })()
```
petesantos: We have a utility function to handle these types of conditionals:

```suggestion
  const statusIcon = Utils.cond(
    [_.includes({ billingAccount: billingAccountName }, workspace), () => null],
    [billingAccountName === workspace.billingAccount, () => 'success-standard'],
    ['billingAccountErrorMessage' in workspace, () => 'error-standard'],
    [Utils.DEFAULT, () => 'loadingSpinner']
  )
```

Move this above line 79 where I annotated above.

Before this change I was seeing busy spinners on my accounts. Please double check the functionality to make sure the correct icons are appearing.

src/pages/billing/Project.js
```diff
@@ -145,23 +152,38 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
-          _.filter({ namespace: projectName }),
           _.orderBy([workspaceSort.field], [workspaceSort.direction]),
           _.map(workspace => {
             const isExpanded = expandedWorkspaceName === workspace.name
             return h(WorkspaceCard, {
-              workspace, key: workspace.workspaceId, isExpanded,
+              workspace,
+              statusIcon: getBillingAccountStatusIcon(workspace),
+              key: workspace.workspaceId,
+              isExpanded,
```
petesantos: This change can be removed

src/pages/billing/Project.js
```diff
@@ -73,6 +79,7 @@ const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, isExpand
   return div({ role: 'listitem', style: { ...Style.cardList.longCardShadowless, flexDirection: 'column' } }, [
```
petesantos, line 81: You can insert the `Utils.cond` statement before the return

src/pages/billing/Project.js
```diff
@@ -73,6 +79,7 @@ const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, isExpand
...
     h(IdContainer, [id => h(Fragment, [
       div({ style: workspaceCardStyles.row }, [
+        statusIcon,
```
petesantos: With the other changes in place try:
```suggestion
statusIcon && icon(statusIcon, { size: 16 }),
```

src/pages/billing/Project.js
```diff
@@ -73,6 +79,7 @@ const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, isExpand
...
         div({ style: { ...workspaceCardStyles.field, display: 'flex', alignItems: 'center', paddingLeft: '1rem' } }, [
```
petesantos, line 85:
```suggestion
        div({ style: { ...workspaceCardStyles.field, display: 'flex', alignItems: 'center', paddingLeft: statusIcon ? '1rem' : '2rem' } }, [
```

src/pages/billing/Project.js
```diff
@@ -23,11 +23,17 @@ import * as Utils from 'src/libs/utils'
 import { billingRoles } from 'src/pages/billing/List'
 
 
+const workspaceStatusIconWidth = 16
```
petesantos: With the other changes below we can remove this

src/pages/billing/Project.js
ehigham: Great - thank you very much!

src/pages/billing/Project.js
ehigham: That logic could be simplified, you're right. 
`includes` has the signature `T -> [T] -> Bool` whereas I'm using `{ billingAccount: billingAccountName }` as a predicate. Would `every` be acceptable - i think that better captures the logic here? 
FYI, it's `workspacesInProject`, not just `workspace`.

src/pages/billing/Project.js
ehigham: Thanks for `cond` - I was hoping we had something like that instead of ternaries!
The spinners are because of an issue in the UI this change exposes (also, it only works when using rawls v2_workspaces). I'm going to fix it in a subsequent change that needs to be double submitted to here and dev. The ticket is https://broadworkbench.atlassian.net/browse/CA-1491

src/pages/billing/Project.js
ehigham: Not exactly. We need to know if all the workspaces have the right billing account, not just if the current one has it. This function lacks the context to know that.

src/pages/billing/Project.js
s-rubenstein: @petesantos I think we still need the screenreader div. Would you suggest just having that with no width there? Just so screenreaders don't get confused at the extra column?

src/pages/billing/Project.js
s-rubenstein: I think you might be able to do something like `_.includes((workspace) => workspace.billingAccount != billingAccountName, workspaces)`, but I don't actually feel like that is necessarily simpler? It may be easier to read/understand though.

src/pages/billing/Project.js
andy7i: echoing sky's concerns of making this column screenreader-friendly

src/pages/billing/Project.js
```diff
@@ -28,7 +28,8 @@ const workspaceExpandIconSize = 20
 
 const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ sort, onSort }) => {
   return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem', marginBottom: '0.5rem' } }, [
-    div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '1rem' } }, [
+    // extra left-padding for workspace billing account status icons
```
dmohs:
```suggestion
```
I don't think this comment adds enough value to pay for the vertical space it consumes.

andy7i:
overall code looks good. +1 for sky's comment https://github.com/DataBiosphere/terra-ui/pull/2594#discussion_r697520369

I ran this locally (against dev) and everything has spinners, which is to be expected since this change relies on changes that are on the v2_workspaces feature branches. since it's a bit annoying to get this running with a fiab, screenshots of the changes would be nice (for next time)

src/pages/billing/Project.js
ehigham: I'm uncertain. When we look back at this, will be just remember that there's extra padding in the headers because each row may have an icon?

ehigham:
> overall code looks good. +1 for sky's comment [#2594 (comment)](https://github.com/DataBiosphere/terra-ui/pull/2594#discussion_r697520369)
> 
> I ran this locally (against dev) and everything has spinners, which is to be expected since this change relies on changes that are on the v2_workspaces feature branches. since it's a bit annoying to get this running with a fiab, screenshots of the changes would be nice (for next time)

Point taken. I'll add screenshots!

src/pages/billing/Project.js
ehigham: added the padding element back in.

andy7i (approved):
LGTM, thanks for the changes!

src/pages/billing/Project.js
```diff
@@ -26,8 +26,16 @@ import { billingRoles } from 'src/pages/billing/List'
...
+const workspaceBillingStatusIcon = (() => {
+  const size = 16
+  const blank = div({ style: { width: size } },
+    [div({ className: 'sr-only' }, ['Status'])])
+  return shape => shape ? icon(shape, { size }) : blank
+})()
+
 const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ sort, onSort }) => {
   return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem', marginBottom: '0.5rem' } }, [
+    workspaceBillingStatusIcon(null),
```
dmohs: Passing `null` implies that this takes an argument. Better to pass nothing.

src/pages/billing/Project.js
```diff
@@ -145,23 +154,38 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
...
+    _.filter({ namespace: projectName })
+  )(workspaces), [workspaces, projectName])
+
+  const getBillingAccountStatusIcon = (() => {
+    const allWorkspacesHaveCorrectBillingAccount =
+      _.every({ billingAccount: billingAccountName }, workspacesInProject)
+    return workspace => Utils.cond(
+      [allWorkspacesHaveCorrectBillingAccount, () => null],
+      [billingAccountName === workspace.billingAccount, () => 'success-standard'],
+      ['billingAccountErrorMessage' in workspace, () => 'error-standard'],
```
dmohs, line 168: Lodash's `_.has` is more idiomatic for this codebase.

dmohs (approved):

src/pages/billing/Project.js
ehigham: It does take an argument - see line 84 for an example.

src/pages/billing/Project.js
ehigham: Is there not an argument to be made in favour of js primitives vs library calls? Surely `in` is more idiomatic for the language?

dmohs: I generally agree with you, but I would imagine the reason `in` is avoided is because of it's sometimes confusing behavior:
```javascript
> 0 in [1, 2, 3]
true
```
My preference in this particular case is to match the existing style and then have the discussion elsewhere.

src/pages/billing/Project.js
dmohs: I had to look more carefully at the definition. Indeed I was wrong.

src/pages/billing/Project.js
ehigham: Thanks for the explanation - I'm new to JS and front end work so the extra context is really helpful.

ehigham: Also, thanks javascript...

ehigham: Wait a second, `_.has` seems to have the same behaviour... 
```javascript
> _.has(0, [1,2,3])
true
```

dmohs:
@ehigham [ANDI](https://www.ssa.gov/accessibility/andi/help/install.html) does not see the workspaces list on the billing projects page as a table. This likely needs to be corrected. Would you be open to creating a follow-up PR focused on fixing this particular issue?

ehigham:
> @ehigham [ANDI](https://www.ssa.gov/accessibility/andi/help/install.html) does not see the workspaces list on the billing projects page as a table. This likely needs to be corrected. Would you be open to creating a follow-up PR focused on fixing this particular issue?

Did this ever work? I'm not sure what I'm supposed to be seeing but looks the same in dev. Created https://broadworkbench.atlassian.net/browse/CA-1500.

s-rubenstein:
@dmohs @ehigham I discussed this back with Andrew Marcus back when we were initially doing this work, and decided that it made a bit more sense as a list due to the expandable rows. However, because we have headers as readable, it is probably clearer to have something explaining that there is something that might appear in that space even if it isn't semantically a table. I would also say it matches the way other menus here are treated, and provides a more consistent user experience.

But no, because of the expandable items, table doesn't quite make sense because of the nature of the rows/items. But it bears a lot of similarities, and so following some of those patterns will make things more clear.

TLDR: We don't actually want a table.

PR #2595, opened 2021-08-26
[DR-1986] Created table view for cataloge datasets

lu-c:
Tested the browser locally using the canned data supplied in the DataBrowser.js file

1. Built out this table view: https://projects.invisionapp.com/share/AC11LGEXBDJM#/screens
2. Columns on table view will sort
3. See more button will expand card
4. Selecting datasets will bring up card that summarizes selections


src/pages/library/common.js
```diff
@@ -235,22 +321,118 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
+                },
+                'aria-label': `Sorted by ${sortDir === 1 ? 'ascending' : 'descending'}`
+              }
+            )
+          ]
+        ) : div({ style: styles.table.header }, headerName)
+    ])
+  }
+
+  return div({ style: styles.table.table }, [
```
cahrens, line 380: I have concerns about how this table will behave with respect to screen readers, since it is all divs with just table styles being applied. I'm still new to terra-ui, but my understanding is that you will want to use one of table components (for example FlexTable), instead of trying to roll your own table from scratch.

If you haven't already, I recommend you test out your new components with a screen reader to make sure it behaves as expected-- if you are are using the Mac's built-in VoicerOver functionality (which you can enable in Settings), it has a tutorial that will help teach you the keystrokes and let you test them out with vanilla HTML components. You can use that to compare against how your UI behaves with VoiceOver.

src/pages/library/common.js
```diff
@@ -139,16 +165,55 @@ export const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSectio
...
+      div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', justifyContent: 'space-between' } }, [
+        `${length} dataset${length > 1 ? 's' : ''} (${fileSizeFormatted} - ${files} bam files) selected to be saved to a Terra Workspace`,
+        div({}, [
+          h(Link, {
+            style: { fontSize: 16, marginRight: 40 },
+            onClick: () => setSelectedData([])
+          }, 'Cancel'),
+          h(ButtonPrimary, {
+            style: { textTransform: 'none', fontSize: 14 },
+            onClick: () => {}
```
cahrens, line 194: I confess I have not checked out the branch, but I am confused as to what happens when you click "Save to a workspace" since onClick doesn't seem to do anything.

As far as Cancel, take a look at _ButtonSecondary_. A Link is used to go someplace, so I'm concerned that it will not have the correct semantics here.

cahrens:
@lu-c I am not an official reviewer-- I just added a couple of comments to help point you in directions that I think will be better from an accessibility standpoint. Feel free to wait for official reviewers to look at the code before making changes.

lu-c:
Moving code to this pr instead because of branching issues: https://github.com/DataBiosphere/terra-ui/pull/2596

src/pages/library/common.js
lu-c: Hi Cahrens, I'll definitely look at ButtonSecondary!

Right now, you are right, the button does not currently do anything, but since this page is hidden from the users and is still under development, I figured it would be okay to handle this chunk of code first and keep the PR a little more streamlined in its purpose. There are follow on tickets that should address functionality of this page, so we can address that there! Eventually, clicking on that button should eventually save them to a workspace, but I'm not sure right now if that's intended to be an existing or a new one.

PR #2596, opened 2021-08-26
[DR-1986] Added table for for catalog (second PR attempt)

lu-c:
The first one was branched off of not dev, which lead to extra files being in the pull request. This one just cherry picked out the relevant ones to my pr.

Tested the browser locally using the canned data supplied in the DataBrowser.js file

Built out this table view: https://projects.invisionapp.com/share/AC11LGEXBDJM#/screens
Columns on table view will sort
See more button will expand card
Selecting datasets will bring up card that summarizes selections

Previous PR: https://github.com/DataBiosphere/terra-ui/pull/2595


src/pages/library/common.js
```diff
@@ -130,7 +156,7 @@ export const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSectio
           }, [
             div({ style: { flex: 1 } }, [label]),
             h(Pill, {
-              count: _.size(workspacesByTag[tag]),
+              count: _.size(listdataByTag[tag]),
```
tcjiang9, line 160: thank you for renaming these!

src/pages/library/common.js
```diff
@@ -161,34 +226,55 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
+    if (_.some(data, selectedData)) {
+      setSelectedData(_.filter(d => !_.isEqual(d, data), selectedData))
+    } else {
+      setSelectedData([...selectedData, data])
+    }
+  }
+
+  const makeListDisplay = (listdataType, listdata) => {
+    switch (listdataType) {
+      case 'datasets': return makeTable(listdata, sort, setSort, sortDir, setSortDir, selectedData, toggleData)
```
tcjiang9: you should double check this with @pshapiro4broad , but I believe we have settled on the backend result being 'snapshots'

src/libs/utils.js
```diff
@@ -76,6 +76,20 @@ export const connectStore = (theStore, name) => WrappedComponent => {
...
+
+  const k = 1024
+  const dm = Math.floor(0, decimals)
+  const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
+
+  const i = Math.floor(Math.log(bytes) / Math.log(k))
+
+  return `${parseFloat((bytes / Math.pow(k, i)).toFixed(dm))} ${sizes[i]}`
+}
+
```
dmohs: This is really nice.

src/pages/library/common.js
pshapiro4broad: Yes, the backend API that we'll use to populate this page is the enumerate snapshots API. However, I think the UI concept will still be called "dataset" to match user expectations. So at one layer we have snapshots and another we have datasets; not sure where you'd want to transition from using the endpoint name vs UI name.

src/pages/library/common.js
```diff
@@ -161,34 +226,55 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
+      case 'workspaces': return _.map(makeCard(), listdata)
```
pshapiro4broad, line 289: I think this UI concept is actually a 'data collection' or 'dataset collection'.

src/pages/library/common.js
```diff
@@ -197,9 +283,9 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
         div({ style: { display: 'flex', margin: '1rem 1rem 0', alignItems: 'baseline' } }, [
           div({ style: { width: '19rem', flex: 'none' } }, [
             div({ style: styles.sidebarRow }, [
-              div({ style: styles.header }, 'Featured workspaces'),
+              div({ style: styles.header }, `Featured ${listdataType}`),
```
pshapiro4broad, line 302: I'm sure we'll have time to further refine the text here, but this isn't a "featured" dataset view, it will show all datasets or collections. This may be confusing to users since terra also has other workspace or dataset views. We may need some wording here to make it clear that these are data repo datasets, which are different than terra workspaces.

kyuksel:
Pending scheduling mobbing time for review with @lu-c per policy.

src/components/common.js
```diff
@@ -92,7 +92,7 @@ export const Link = Utils.forwardRefWithName('Link', ({ disabled, variant, child
     style: {
       color: disabled ? colors.dark(0.7) : colors.accent(variant === 'light' ? 0.3 : 1),
       cursor: disabled ? 'not-allowed' : 'pointer',
-      fontWeight: 500, display: 'inline'
+      fontWeight: _.getOr(500, 'style.fontWeight', props), display: 'inline'
```
dmohs: As discussed, we hope the merge takes care of this, but if not here is another option:
```suggestion
      fontWeight: props.style?.fontWeight || 500, display: 'inline'
```

src/libs/utils.js
dmohs: However, `filesize` lib might give you what you want here.

src/components/common.js
lu-c: Remove this

src/pages/library/common.js
```diff
@@ -139,16 +165,55 @@ export const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSectio
...
+          h(ButtonSecondary, {
+            style: { fontSize: 16, marginRight: 40, textTransform: 'none' },
+            onClick: () => setSelectedData([])
+          }, 'Cancel'),
+          h(ButtonPrimary, {
+            style: { textTransform: 'none', fontSize: 14 },
+            onClick: () => {}
+          }, [
+            'Save to a workspace'
+          ])
```
dmohs:
```suggestion
          }, ['Save to a workspace'])
```

src/pages/library/common.js
```diff
@@ -139,16 +165,55 @@ export const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSectio
...
+        position: 'sticky', bottom: 0, marginTop: '20px',
+        width: '100%', padding: '34px 60px',
+        backgroundColor: 'white', boxShadow: 'rgb(0 0 0 / 30%) 0px 0px 8px 3px',
+        fontSize: 17
+      }
+    },
+    [
+      div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', justifyContent: 'space-between' } }, [
+        `${length} dataset${length > 1 ? 's' : ''} (${fileSizeFormatted} - ${files} bam files) selected to be saved to a Terra Workspace`,
+        div({}, [
```
dmohs:
```suggestion
        div([
```

src/pages/library/common.js
```diff
@@ -161,34 +226,55 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
     } else {
       const tags = _.uniq(_.flatMap(section => section.tags, selectedSections))
-      return _.uniq(_.flatMap(tag => workspacesByTag[tag], tags))
+      return _.uniq(_.flatMap(tag => listdataByTag[tag], tags))
     }
   }
   // eslint-disable-next-line lodash-fp/no-single-composition
-  const filterByTags = workspaces => _.flow(_.map(tag => _.intersection(workspacesByTag[tag]), selectedTags))(workspaces)
-  const filterByText = workspaces => {
+  const filterByTags = listdata => _.flow(_.map(tag => _.intersection(listdataByTag[tag]), selectedTags))(listdata)
```
dmohs:
```suggestion
  const filterByTags = () => _.map(tag => _.intersection(listdataByTag[tag]), selectedTags)
```

src/pages/library/common.js
```diff
@@ -161,34 +226,55 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
     if (_.isEmpty(selectedSections)) {
-      return workspaces
+      return listdata
     } else {
       const tags = _.uniq(_.flatMap(section => section.tags, selectedSections))
-      return _.uniq(_.flatMap(tag => workspacesByTag[tag], tags))
+      return _.uniq(_.flatMap(tag => listdataByTag[tag], tags))
     }
   }
   // eslint-disable-next-line lodash-fp/no-single-composition
```
dmohs:
```suggestion
```

src/pages/library/common.js
```diff
@@ -161,34 +226,55 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
   }
-  const filteredWorkspaces = _.flow(
+  const filteredData = _.flow(
     filterBySections,
     filterByTags,
     filterByText
   )(featuredList)
 
+  const toggleData = data => {
+    if (_.some(data, selectedData)) {
```
dmohs, line 271: Please check order:
```suggestion
    if (_.includes(data, selectedData)) {
```

src/pages/library/common.js
```diff
@@ -161,34 +226,55 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
+    } else {
+      setSelectedData([...selectedData, data])
+    }
+  }
+
+  const makeListDisplay = (listdataType, listdata) => {
+    switch (listdataType) {
+      case 'datasets': return makeTable(listdata, sort, setSort, sortDir, setSortDir, selectedData, toggleData)
+      case 'workspaces': return _.map(makeCard(), listdata)
+      default: return div({}, '')
```
dmohs:
```suggestion
      default: return null // or div()
```
May also want to take a look at `Utils.switchCase`.

src/pages/library/common.js
```diff
@@ -235,22 +321,118 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
+  ])
+}
+
+const makeTable = (listData, sort, setSort, sortDir, setSortDir, selectedData, toggleData) => {
+  const makeTableHeader = (headerStyles, headerName, sortable = false) => {
+    return div({ style: { ...styles.table.header, ...headerStyles } }, [
+      sortable ?
+        h(
+          Link,
+          {
```
dmohs:
```suggestion
        h(Link, {
```

src/pages/library/common.js
```diff
@@ -235,22 +321,118 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
+                'aria-label': `Sorted by ${sortDir === 1 ? 'ascending' : 'descending'}`
+              }
+            )
+          ]
+        ) : div({ style: styles.table.header }, headerName)
+    ])
+  }
+
+  return div({ style: styles.table.table }, [
+    div({ style: { ...styles.table.flexTableRow, marginBottom: '-15px' } }, [
```
dmohs: Equivalent:
```suggestion
    div({ style: { ...styles.table.flexTableRow, marginBottom: -15 } }, [
```

src/pages/library/common.js
```diff
@@ -235,22 +321,118 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
...
+        ),
+        div({ style: { ...styles.table.flexTableRow, alignItems: 'flex-start' } }, [
+          div({ style: { ...styles.table.col, ...styles.table.firstElem } }, [
+            icon(listdatum.locked ? 'lock' : 'lock-o', { size: 12, style: { flex: 'none', color: listdatum.locked ? colors.accent() : colors.primary() } })
+          ]),
+          div({ style: { ...styles.table.col, width: '100%', fontSize: '12px' } }, [
+            h(
+              Link,
+              { onClick: () => setIsOpened(!isOpened) },
+              [
```
dmohs:
```suggestion
            h(Link, { onClick: () => setIsOpened(!isOpened) }, [
```

dmohs (approved):
Some suggestions, a few nits.

src/libs/utils.js
lu-c: haha, gotcha, tested this, and I'll be replacing this in my next commit!

src/pages/library/common.js
lu-c: When I switch to `_.includes`, the condition only ever returns as false - I don't understand the reasoning though, since according to the documentation, they should be pretty equivalent methods... For now, I'd like to propose we leave it as `_.some`

src/pages/library/common.js
lu-c: Sounds good - It might make more sense to pass in an object with overrides for labels / headers instead of a single string then

src/pages/library/common.js
lu-c: I think changing the workspace name to "data collection" is out of scope for this PR, but I'll make a ticket for it in jira to capture that for future work!

src/pages/library/common.js
dmohs: Thank you for checking this. Indeed, `_.some` is much more powerful than I realized. While `_.includes` does an equality comparison, `_.some` will use different comparison mechanisms depending on the argument. I'm glad to have learned this.

tcjiang9 (approved):

PR #2597, opened 2021-08-27
[CA-1491] Use v2 billing projects endpoint over user/billing

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1491

Change `Ajax -> Billing -> listProjects` to use `/api/billing/v2` and
fix any resulting breakages. Since v2 projects are created
synchronously, no longer need creation in progress logic.

Don't call Google to discover the billing project's billing account -
this does not work for v2 billing projects as we don't create google
projects to back terra billing projects anymore.


src/components/NewWorkspaceModal.js
```diff
@@ -100,10 +100,9 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
       Ajax(signal).Billing.listProjects(),
       Ajax(signal).Groups.list()
     ])
-    const usableProjects = _.filter({ creationStatus: 'Ready' }, billingProjects)
-    setBillingProjects(usableProjects)
+    setBillingProjects(billingProjects)
```
s-rubenstein, line 103: Does our new endpoint filter out unusable projects? Theoretically people can create projects using the v1 endpoint and we still want to block those from showing up here? (Although I guess if we use the ppw workspaces, they don't care about they state of the billing project)

src/libs/ajax.js
```diff
@@ -89,9 +89,10 @@ const withRequesterPays = wrappedFetch => (url, ...args) => {
   const { workspace, canCompute } = workspaceStore.get() || {}
   const workspaceProject = workspace?.namespace
 
-  const canUseWorkspaceProject = async () => {
-    return canCompute || _.some({ projectName: workspaceProject, role: 'Owner' }, await Ajax().Billing.listProjects())
-  }
+  const canUseWorkspaceProject = async () => canCompute || _.some(
```
s-rubenstein: Can we reuse the above code? I just noticed we have some code duplication between here and workspace-utils

src/pages/billing/Project.js
```diff
@@ -145,7 +144,7 @@ const ProjectDetail = ({ project, project: { projectName, creationStatus }, bill
 
   const signal = Utils.useCancellation()
 
-  const adminCanEdit = _.filter(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers).length > 1
+  const adminCanEdit = _.some(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers)
```
s-rubenstein: I don't think this is quite equivalent? The goal is `> 1`, and I think this will now return `>= 1`?

ehigham: I'll revert it. For my reference, what does this mean? Admin can edit if there's more then one project user with the `Owner` role?

src/components/NewWorkspaceModal.js
ehigham: Not sure - what's an unusable project?

ehigham: Just tested this out - v1 billing projects whose google projects are still being created are not filtered out. I guess this doesn't actually matter as you say as we won't use that project in PPW.

marctalbott: Terra billing project creation used to fail sometimes which would result in a `creationStatus` of `Error`. Those projects may show up in this list now. That's a product of the endpoint in Rawls though since we don't expose `creationStatus`

marctalbott (approved):

src/components/NewWorkspaceModal.js
ehigham: Is that failing to create the billing project or failing to create the google project that backs it? Do we care if it's the latter and it fails?

marctalbott: It's failing to create the google project that backs it. Given that, we're not going to be getting any _new_ errored Terra billing projects, so no need to really worry about that. However, any errored Terra billing projects from before PPW will show up in the UI. Couple potential issues with this:
1. It may confuse users to have projects that were once unusable suddenly appear to be fully functional (not just "appear" I suppose, they may in fact be fully functional). 
2. Users would potentially be able to create v2 workspaces in their errored Terra billing projects. Even though a Terra billing project with a errored Google project is functionally similar to a Terra billing project without a Google project at all, in the past errored Terra billing projects have never had workspaces in them and that _could_ cause issues somewhere, just hard to say exactly where.

ehigham: How should we address this then? We could make rawls not return these projects since we'd be exposing that sort of behaviour via the API anyway? Or I could just get the list of projects from both endpoints and intersect on ready or something?

marctalbott: We should probably prevent users from creating workspaces in errored Terra billing projects on the Rawls side of things if we don't already. 

As for errored Terra billing projects suddenly appearing OK in the UI, I'd say we should get product input on if that's okay or if we need to devise a way to exclude or otherwise indicate the errored project status for users

src/pages/billing/Project.js
s-rubenstein: I think it has to do with being able to remove admin users? I am not 100% clear without looking at the usages, but I think it is to ensure there is at least one admin.

s-rubenstein (approved):
LGTM

src/pages/billing/Project.js
ehigham: Thank you!

src/libs/ajax.js
```diff
@@ -80,22 +80,23 @@ const checkRequesterPaysError = async response => {
...
  * project if the user has access, retrying the request once if necessary.
  */
 const withRequesterPays = wrappedFetch => (url, ...args) => {
   const bucket = /\/b\/([^/?]+)[/?]/.exec(url)[1]
-  const { workspace, canCompute } = workspaceStore.get() || {}
-  const workspaceProject = workspace?.namespace
-
-  const canUseWorkspaceProject = async () => {
-    return canCompute || _.some({ projectName: workspaceProject, role: 'Owner' }, await Ajax().Billing.listProjects())
-  }
```
dmohs, line 94: Nice duplication removal here.

dmohs (approved):

src/components/NewWorkspaceModal.js
marctalbott: Ticket to prevent creating workspaces in errored Terra billing projects: https://broadworkbench.atlassian.net/browse/CA-1494

PR #2598, opened 2021-08-30
[CA-1479] Move to googleProject for pet SA calls

s-rubenstein:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/IGVBrowser.js
+++ b/src/components/IGVBrowser.js
@@ -33,7 +33,7 @@ const IGVBrowser = _.flow(
           return knownBucketStatus
         } else {
           try {
-            await Ajax(signal).Buckets.getObject(bucket, file, workspace.workspace.namespace, { fields: 'kind' })
+            await Ajax(signal).Buckets.getObject(workspace.workspace.googleProject, bucket, file, { fields: 'kind' })
             return false
           } catch (e) {
             if (e.requesterPaysError) {
@@ -63,7 +63,7 @@ const IGVBrowser = _.flow(
             }, selectedFiles))
           }
 
-          igv.setGoogleOauthToken(() => saToken(workspace.workspace.namespace))
+          igv.setGoogleOauthToken(() => saToken(workspace.workspace.googleProject))
           igv.createBrowser(containerRef.current, options)
         } catch (e) {
           reportError('Error loading IGV.js', e)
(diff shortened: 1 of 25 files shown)


ehigham (approved):
Looks good. Like the parameter reordering.

src/components/IGVBrowser.js
```diff
@@ -33,7 +33,7 @@ const IGVBrowser = _.flow(
           return knownBucketStatus
         } else {
           try {
-            await Ajax(signal).Buckets.getObject(bucket, file, workspace.workspace.namespace, { fields: 'kind' })
+            await Ajax(signal).Buckets.getObject(workspace.workspace.googleProject, bucket, file, { fields: 'kind' })
```
jacmrob, line 36: does order matter here? previously it went bucket, file, namespace (project), new one is project, bucket, file

jacmrob: looking at the rest of the code, it seems intentional. just curious why it could be re-ordered

gpolumbo-broad (approved):
LGTM, but how do we test all of this?

dmohs (approved):
Very well-contained PR, making it easy to review.

s-rubenstein:
@dmohs do you have any clear advice on testing all of this? I clicked around the UI some, and nothing seemed obviously broken (and it compiled), but especially with some of the data table pieces I wasn't sure how to test it.

jacmrob (commented):
I noticed there are still instances where `namespace` sticks around - is that intended?

breilly2:
There were previously some cases where we would use a general pet service account instead of a project-specific service account because, for popular public workspaces, it was too easy to hit the project's service account limit. That may be less of a problem in a PPW world, but I don't imagine it going away completely. Is there a plan to mitigate this?

dmohs:
@s-rubenstein

> @dmohs do you have any clear advice on testing all of this?

The integration tests cover critical use cases. It's not feasible to test every page manually. Spot check where you can (and get help if there is something you'd like to test but can't figure out how to do so). Then, take comfort in the fact that this codebase is built to minimize mean time to repair, so if something does break in this PR or another, we can usually fix it quickly.

s-rubenstein:
@jacmrob namespace is still used as a workspace identifier. So we still want it for things where:
A) We do routing
B) We log mixpanel events
C) We make calls to rawls to fetch data in a workspace.

@breilly2 I am not aware of any mitigation for this, but I don't think it affects this change, since this is only changing ones that are getting the SA token by passing in namespace. 

@dmohs good to know. Yeah, my inclination was to do some minimal smoke testing and then fix in post if things are broken, given the tests all passed.

breilly2:
Ah, my mistake. It looked to me like there were a couple of places where `googleProject` was added where there wasn’t previously `namespace`, but I think I tripped over argument reordering. Sorry about that.

PR #2599, opened 2021-08-30
Add the ability add/edit user comments on workflows [BW-759]

cahrens:
This PR allows the ability to specify a comment when launching a workflow, and the ability to view/edit the comment from either the Job History page or the Submission Details.

It introduces a reusable "ValidatedTextArea", based on our style guide. The existing "ValidatedInput" is for single-line input only.

It also does a bit of clean-up with regard to marking div's that display error messages as aria-live, and fixing some syntax that was throwing deprecation warnings.

See also https://github.com/broadinstitute/rawls/pull/1510


src/components/input.js
```diff
@@ -301,6 +305,32 @@ export const TextArea = ({ onChange, autosize = false, nativeOnChange = false, .
   }, props))
 }
 
+/**
+ * Utility function for wrapping text on periods.
+ */
+export const wrappableOnPeriods = _.flow(str => str?.split(/(\.)/), _.flatMap(sub => sub === '.' ? [wbr(), '.'] : sub))
```
cahrens: This was an existing line of code from LaunchAnalysisModal (I take no credit for it…).

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -155,13 +158,14 @@ const JobHistory = _.flow(
...
-  const makeHeaderRenderer = (name, label, sortable = true) => {
+  const makeHeaderRenderer = (name, label, sortable = true, ariaLabel = label) => {
     return () => h(HeaderRenderer, {
       sort: sortable ? sort : undefined,
       name,
       label,
       onSort: setSort,
-      style: { fontWeight: 400 }
+      style: { fontWeight: 400 },
+      'aria-label': ariaLabel
```
cahrens, line 168: @breilly2 I added the ability to specify an ariaLabel so that "No. of Workspaces" reads out correctly.

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -185,13 +189,13 @@ const JobHistory = _.flow(
 
   const sortedSubmissions = _.orderBy(
     Utils.cond(
-      [sort.field === 'entityName', ['submissionEntity.entityName']],
-      [sort.field === 'numberOfWorkflows', [s => _.sum(_.values(s.workflowStatuses))]],
-      [sort.field === 'status', [s => {
+      [sort.field === 'entityName', () => 'submissionEntity.entityName'],
```
cahrens: @breilly2 This code was throwing deprecation warnings.

src/pages/workspaces/workspace/JobHistory.js
cahrens: I also figured that for the first column, "click for details" was not helpful (and potentially confusing) in the spoken text.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -54,14 +58,14 @@ const LaunchAnalysisModal = ({
...
-        memoryRetryMultiplier: retryWithMoreMemory ? retryMemoryFactor : undefined,
+        memoryRetryMultiplier: retryWithMoreMemory ? retryMemoryFactor : undefined, userComment: _.trim(userComment),
         onProgress: stage => {
           setMessage({ createSet: 'Creating set...', launch: 'Launching analysis...', checkBucketAccess: 'Checking bucket access...' }[stage])
         }
       })
       onSuccess(submissionId)
     } catch (error) {
-      setLaunchError(await (error instanceof Response ? error.text() : error.message))
+      setLaunchError(await (error instanceof Response ? error.json().then(data => data.message) : error.message))
```
cahrens, line 68: This would show a pretty ugly error before, so pulling out just the message portion. Before:

image

src/style.css
```diff
@@ -88,6 +88,12 @@ a {
   outline: none !important;
 }
 
+.error-style {
```
cahrens, line 91: See the block of CSS immediately preceding this.

src/pages/workspaces/workspace/JobHistory.js
breilly2: Cool. Also, this could actually be `Utils.switchCase`, as Isaac pointed out to me in a similar situation in the showcase PR.

kyuksel:
Pending scheduling mobbing time for review with @cahrens per policy.

aednichols (changes requested):
First preference:

- Treat all unicode characters, including emoji, correctly & equally. We don't live in an ASCII world anymore 😃 

Second preference:

- User friendly error that does not quote raw SQL errors

---

Screenshot 2021-08-30 at 17-38-45 anichols-test - Workflows - echo_strings

dmohs:
Nice catch @aednichols. The returned string is truncated (e.g., "Incorrect string value: '\\xF0\\x9F\\x98\\x80\\xF0\\x9F...' for column 'USER_COMMENT' at row 1"), so I don't think we can reliably parse it to display the right thing. The request payload (as rendered by Chrome) looks good to me, so I'd look to Rawls as the source of the bug.

I judge the details in the error message are useful, so I don't think I'd change this code. It seems to me it is doing the best it can with what it gets.

cahrens:
Yes, thanks @aednichols for thinking to test with non-ASCII characters… my mind still defaults to an ASCII world.

It's interesting because some unicode characters, like ăѣ, work fine (and @salonishah11 included a unit test for them in https://github.com/broadinstitute/rawls/pull/1489), but others like 𝔠 or 😍, do not. As @dmohs commented, I think the terra-ui code is doing the best it can, unless we wanted to add client-side validation to exclude certain classes of characters. My thought is that we should probably create a new ticket for further investigation/changes (most likely on the rawls side). But I will continue to poke a bit more as well.

aednichols:
The more common/"normal" non-ASCII characters used by major languages have lower indexes in the unicode numbering system, causing them to require only one byte in a possible multi-byte encoding. So that explains why some non-ASCII works but not all.

I tried some Polish characters (my other language) and they work fine as expected.

Screen Shot 2021-08-31 at 2 45 16 PM

cahrens:
Right, I think the 4-byte characters are the problem… though supposedly mysql has the ability to support them. https://dev.mysql.com/doc/refman/5.6/en/charset-unicode.html

aednichols:
I could look into the MySQL encoding configuration with you if you like, I find text encoding problems fascinating.

On the other hand, to save time it makes me feel better knowing that it works just fine with most characters/languages people are actually going to use.

cahrens:
@aednichols I agree that character encoding is interesting, but I think we probably shouldn't get too distracted from the other things we are working on (and probably product should weigh in on how much they do or don't care). I'll create a separate ticket to capture this issue, and we can discuss with product whether it's worth spending time on it.

https://broadworkbench.atlassian.net/browse/BW-797

aednichols:
Agreed.

Even if we did figure out a fix, it ~would~ could be a potentially invasive DB reconfiguration that we would have to weight against the chance of introducing new, worse bugs.

aednichols (approved):
LGTM, with an optional suggestion:

I did some informal usability testing and the focus group (my boyfriend) preferred this version because it states explicitly that it's optional

Screenshot 2021-08-31 at 11-45-29 anichols-test - Workflows - echo_strings

src/components/input.js
```diff
@@ -301,6 +305,31 @@ export const TextArea = ({ onChange, autosize = false, nativeOnChange = false, .
...
+/**
+ * A TextArea that provides visual and textual indications when the content is invalid.
+ *
+ * @param {Object} inputProps input properties for the TextArea
+ * @param {String} error the message to display below the TextArea, or undefined if no error
+ */
+export const ValidatedTextArea = ({ inputProps, error }) => {
+  return h(Fragment, [
+    h(TextArea,
+      _.merge({ className: error ? 'error-style' : 'focus-style' }, inputProps)
```
cahrens:
```suggestion
      { className: error ? 'error-style' : 'focus-style', ...inputProps }
```

src/components/input.js
```diff
@@ -301,6 +305,31 @@ export const TextArea = ({ onChange, autosize = false, nativeOnChange = false, .
...
+    ),
+    div(
```
cahrens: 324-325 can be same line.

src/components/input.js
```diff
@@ -301,6 +305,31 @@ export const TextArea = ({ onChange, autosize = false, nativeOnChange = false, .
   }, props))
 }
 
+/**
+ * Utility function for wrapping text on periods.
```
dmohs:
```suggestion
 * Adds a WBR element before each period which advises the browser that a line-break here is acceptable.
```

src/components/input.js
cahrens:
```suggestion
const PeriodWrapper = ({ children }) => {
  const child = Children.only(children)
  const wrappableText = _.flow(
    _.split(/(\.)/),
    _.flatMap(sub => sub === '.' ? [wbr(), '.'] : sub)
  )(child)
  return h(Fragment, wrappableText)
}
```

src/components/input.js
```diff
@@ -301,6 +305,31 @@ export const TextArea = ({ onChange, autosize = false, nativeOnChange = false, .
...
+ * @param {String} error the message to display below the TextArea, or undefined if no error
+ */
+export const ValidatedTextArea = ({ inputProps, error }) => {
+  return h(Fragment, [
+    h(TextArea,
+      _.merge({ className: error ? 'error-style' : 'focus-style' }, inputProps)
+    ),
+    div(
+      {
+        style: { color: colors.danger(), overflowWrap: 'break-word', marginTop: '.75rem' },
```
cahrens: 0.75

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -121,6 +122,7 @@ const JobHistory = _.flow(
   const [abortingId, setAbortingId] = useState(undefined)
   const [textFilter, setTextFilter] = useState('')
   const [sort, setSort] = useState({ field: 'submissionDate', direction: 'desc' })
+  const [updatingComment, setUpdatingComment] = useState(undefined)
```
cahrens: updatingCommentId

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -306,34 +312,51 @@ const JobHistory = _.flow(
...
             {
-              size: { basis: 210, grow: 1 },
+              // Disable shrinking so that the ID does not render outside of the cell
+              size: { basis: 170, grow: 1, shrink: 0 },
               field: 'submissionId',
               headerRenderer: makeHeaderRenderer('submissionId', 'Submission ID'),
               cellRenderer: ({ rowIndex }) => {
                 const { submissionId } = sortedSubmissions[rowIndex]
-                return h(TooltipCell, { tooltip: submissionId }, [
+                return h(TextCell, { style: { whiteSpace: 'normal' } }, [
```
cahrens: Can just be a div.

cahrens: Font size of 12

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -306,34 +312,51 @@ const JobHistory = _.flow(
...
                   }, [submissionId])
                 ])
               }
             },
+            {
+              size: { basis: 250, grow: 1 },
+              headerRenderer: makeHeaderRenderer('userComment', 'Comment'),
+              cellRenderer: ({ rowIndex }) => {
+                const { userComment } = sortedSubmissions[rowIndex]
+                return userComment && h(TooltipCell, { tooltip: userComment }, [userComment])
```
cahrens: May not need userComment &&

And don't need tooltip: userComment

src/pages/workspaces/workspace/jobHistory/UpdateUserCommentModal.js
```diff
@@ -0,0 +1,74 @@
...
+ * @param {Function} updateError the method to display or clear an error message
+ */
+export const commentValidation = (newComment, updateComment, updateError) => {
+  // This is the server-side limit for the API. We are doing client-side validation also in order to present
+  // a friendlier user experience (in particular for the Launch Workflows dialog, where a server-side error can
+  // occur for a variety of reasons). If the server-side limit changes, it must also be updated here.
+  const maxCommentLength = 1000
+  updateComment(newComment)
+  updateError(newComment.length > maxCommentLength ?
+    `The comment can be at most ${maxCommentLength} characters` : undefined
```
kyuksel: Nit:
```suggestion
    `The comment can be at most ${maxCommentLength} characters.` : undefined
```

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -130,7 +146,11 @@ const LaunchAnalysisModal = ({
...
       message
     ]),
-    div({ style: { color: colors.danger(), overflowWrap: 'break-word' } }, [
+    div({
+      style: { color: colors.danger(), overflowWrap: 'break-word' },
+      'aria-live': 'assertive',
+      'aria-relevant': 'all'
+    }, [
       h(Fragment, wrappableOnPeriods(launchError))
     ])
```
dmohs: Maybe?
```suggestion
    }, [h(Fragment, wrappableOnPeriods(launchError))])
```

dmohs (approved):
Approved with changes as suggested via Mob review session.

src/components/input.js
cahrens: @dmohs we actually decided to remove the function completely (https://broadinstitute.slack.com/archives/C01EHNUM73R/p1630528774002800)

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
cahrens: This got further simplified.

src/components/input.js
cahrens: We ended up removing this because it wasn't actually useful for textual messages (since HTML wraps on whitespace by default).

src/pages/workspaces/workspace/JobHistory.js
cahrens: @breilly2 @zarsky-broad I decided to stop being lazy and changed this to Utils.switchCase.

cahrens:
FYI, I have pushed up all the changes from the mob review.

cahrens:
@petesantos @kyuksel I am hoping to merge this by EOD today. Please let me know if you have any concerns about that-- I am not sure if I am expected to wait for both of you to approve, or if the 2 checkmarks I have are considered sufficient.

kyuksel:
> I am hoping to merge this by EOD today. Please let me know if you have any concerns about that-- I am not sure if I am expected to wait for both of you to approve, or if the 2 checkmarks I have are considered sufficient.

@cahrens Two 👍🏼s are sufficient to merge (as long as at least one is from a "[terra devs](https://github.com/orgs/DataBiosphere/teams/terra-devs)" group member) so you're good to merge 😊  Thanks for making sure though!

PR #2602, opened 2021-08-31
[CA-1495] Poll for Billing Account Changes in Billing Project Page

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1495
When a user changes the billing account of a billing project, the billing account on the project's workspaces is updated asynchronously. We need to poll Rawls for changes to the billing account until all changes have been applied. Note that polling is disabled when the change billing modal is being displayed as this causes nasty UI jumping sadness.

### DIFF
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -125,7 +125,7 @@ const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, billingA
   ])
 })
 
-const ProjectDetail = ({ project, project: { projectName }, billingAccounts, authorizeAndLoadAccounts }) => {
+const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) => {
   // State
   const { query } = Nav.useRoute()
   // Rather than using a localized StateHistory store here, we use the existing `workspaceStore` value (via the `useWorkspaces` hook)
@@ -138,7 +138,7 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
   const [updating, setUpdating] = useState(false)
   const [updatingAccount, setUpdatingAccount] = useState(false)
   const [refreshing, setRefreshing] = useState(false)
-  const [billingAccountName, setBillingAccountName] = useState(project.billingAccount)
+  const [billingProject, setBillingProject] = useState(project)
   const [showBillingModal, setShowBillingModal] = useState(false)
   const [showSpendReportConfigurationModal, setShowSpendReportConfigurationModal] = useState(false)
   const [selectedBilling, setSelectedBilling] = useState()
@@ -154,8 +154,8 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
   const adminCanEdit = _.filter(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers).length > 1
 
   const getBillingAccountStatusIcon = workspace => Utils.cond(
-    [_.isEmpty(project.workspacesWithIncorrectBillingAccount), () => null],
-    [billingAccountName === workspace.billingAccount, () => 'success-standard'],
+    [_.isEmpty(billingProject.workspacesWithIncorrectBillingAccount), () => null],
+    [billingProject.billingAccount === workspace.billingAccount, () => 'success-standard'],
     ['billingAccountErrorMessage' in workspace, () => 'error-standard'],
     [Utils.DEFAULT, () => 'loadingSpinner']
   )
@@ -163,10 +163,10 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
   const tabToTable = {
     workspaces: h(Fragment, [
       h(WorkspaceCardHeaders, { sort: workspaceSort, onSort: setWorkspaceSort }),
-      div({ role: 'list', 'aria-label': `workspaces in billing project ${projectName}`, style: { flexGrow: 1, width: '100%' } }, [
+      div({ role: 'list', 'aria-label': `workspaces in billing project ${billingProject.projectName}`, style: { flexGrow: 1, width: '100%' } }, [
         _.flow(
           _.map(({ workspace }) => workspace),
-          _.filter({ namespace: projectName }),
+          _.filter({ namespace: billingProject.projectName }),
           _.orderBy([workspaceSort.field], [workspaceSort.direction]),
           _.map(workspace => {
             const isExpanded = expandedWorkspaceName === workspace.name
@@ -188,7 +188,7 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
         icon('plus-circle', { size: 14 }),
         div({ style: { marginLeft: '0.5rem' } }, ['Add User'])
       ]),
-      div({ role: 'table', 'aria-label': `users in billing project ${projectName}` }, [
+      div({ role: 'table', 'aria-label': `users in billing project ${billingProject.projectName}` }, [
         h(MemberCardHeaders, { sort, onSort: setSort }),
         div(_.map(member => {
           return h(MemberCard, {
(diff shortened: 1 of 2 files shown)


src/pages/billing/Project.js
```diff
@@ -276,12 +279,17 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
 
   useEffect(() => StateHistory.update({ projectUsers }), [projectUsers])
 
-  // Render
-  const { displayName = null } = _.find({ accountName: billingAccountName }, billingAccounts) || { displayName: 'No Access' }
+  Utils.usePollingEffect(
+    async () => _.isEmpty(getBillingProject().workspacesWithIncorrectBillingAccount) ||
```
ehigham: I'm holding the billing account in a ref as the value of `billingAccount` in the lambda never changed when i used 
```javascript
[billingProject, setBillingProject] = useState( ...
```
I have no idea why. I'd really welcome an explanation.

src/pages/billing/Project.js
```diff
@@ -148,31 +147,26 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
...
-    return workspace => Utils.cond(
-      [allWorkspacesHaveCorrectBillingAccount, () => null],
-      [billingAccountName === workspace.billingAccount, () => 'success-standard'],
-      ['billingAccountErrorMessage' in workspace, () => 'error-standard'],
-      [Utils.DEFAULT, () => 'loadingSpinner'])
-  })()
+  const getBillingAccountStatusIcon = workspace => Utils.cond(
+    [_.isEmpty(billingProject.workspacesWithIncorrectBillingAccount), () => null],
+    [billingProject.billingAccount === workspace.billingAccount, () => 'success-standard'],
+    ['billingAccountErrorMessage' in workspace, () => 'error-standard'],
```
dmohs, line 159: To avoid the surprises of `in`:
```suggestion
    [_.has(workspace, 'billingAccountErrorMessage'), () => 'error-standard'],
```

src/pages/billing/Project.js
```diff
@@ -148,31 +147,26 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
...
+    [Utils.DEFAULT, () => 'loadingSpinner']
+  )
 
   const tabToTable = {
     workspaces: h(Fragment, [
       h(WorkspaceCardHeaders, { sort: workspaceSort, onSort: setWorkspaceSort }),
-      div({ role: 'list', 'aria-label': `workspaces in billing project ${projectName}`, style: { flexGrow: 1, width: '100%' } }, [
+      div({ role: 'list', 'aria-label': `workspaces in billing project ${billingProject.projectName}`, style: { flexGrow: 1, width: '100%' } }, [
         _.flow(
+          _.map(({ workspace }) => workspace),
```
dmohs: Equivalently:
```suggestion
          _.map('workspace'),
```
We tend to take advantage of Lodash's shorthands when we can, which makes them less surprising when encountered.

src/pages/billing/Project.js
```diff
@@ -276,12 +276,23 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
 
   useEffect(() => StateHistory.update({ projectUsers }), [projectUsers])
 
-  // Render
-  const { displayName = null } = _.find({ accountName: billingAccountName }, billingAccounts) || { displayName: 'No Access' }
+  // There's some madness going on when using state variables in polling effects - the reference
+  // never updates (i.e. it's bound to the value that the component was first mounted with).
+  // `useGetter` works around this and I have no idea why.
```
dmohs, line 281: @zarsky-broad I'd appreciate your input here.

dmohs (approved):

src/pages/billing/Project.js
ehigham: This is good to know, thank you.

src/pages/billing/Project.js
ehigham: `has` exhibits the same behaviour as `in` for your [example](https://github.com/DataBiosphere/terra-ui/pull/2594#discussion_r699554389): 
```javascript
> 0 in [1, 2, 3] === _.has(0, [1, 2, 3])
true
```

marctalbott (approved):

src/pages/billing/Project.js
```diff
@@ -148,31 +147,26 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
...
-    const allWorkspacesHaveCorrectBillingAccount =
-      _.every({ billingAccount: billingAccountName }, workspacesInProject)
-    return workspace => Utils.cond(
-      [allWorkspacesHaveCorrectBillingAccount, () => null],
-      [billingAccountName === workspace.billingAccount, () => 'success-standard'],
-      ['billingAccountErrorMessage' in workspace, () => 'error-standard'],
-      [Utils.DEFAULT, () => 'loadingSpinner'])
-  })()
+  const getBillingAccountStatusIcon = workspace => Utils.cond(
+    [_.isEmpty(billingProject.workspacesWithIncorrectBillingAccount), () => null],
```
andy7i, line 157: nice improvement over using `allWorkspacesHaveCorrectBillingAccount`

src/pages/billing/Project.js
```diff
@@ -148,31 +147,26 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
   const [expandedWorkspaceName, setExpandedWorkspaceName] = useState()
   const [sort, setSort] = useState({ field: 'email', direction: 'asc' })
   const [workspaceSort, setWorkspaceSort] = useState({ field: 'name', direction: 'asc' })
+  const [billingProject, setBillingProject] = useState(project)
 
   const signal = Utils.useCancellation()
 
   const adminCanEdit = _.filter(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers).length > 1
 
-  const workspacesInProject = useMemo(() => _.flow(
```
andy7i, line 156: what led you to remove the use of `useMemo`? Is it because `workspaces` already caches the workspace state?

src/pages/billing/Project.js
```diff
@@ -231,25 +225,34 @@ const ProjectDetail = ({ project, project: { projectName }, billingAccounts, aut
...
+    await updateBillingProject()
   })
 
   const updateSpendConfiguration = _.flow(
     withErrorReporting('Error updating workflow spend report configuration'),
     Utils.withBusyState(setUpdating)
-  )(async () => {
-    await Ajax(signal).Billing.updateSpendConfiguration({ billingProjectName: projectName, datasetGoogleProject: selectedDatasetProjectName, datasetName: selectedDatasetName })
-  })
+  )(() => Ajax(signal).Billing.updateSpendConfiguration({
```
andy7i, line 251: why don't we need the async / await here anymore?

src/pages/billing/Project.js
andy7i: not sure, but is `Utils.usePollingEffect` compatible with using updated state values? I see it uses `useOnMount`

andy7i (approved):
a few questions but looks great otherwise. I ran it locally but not against a fiab. (would appreciate screenshots next time!)

src/pages/billing/Project.js
ehigham: very loosely speaking: `async` wraps a return value in a `Promise` and `await` unwraps it. It's usage here could be seen as unnecessary as we can accomplish the same thing by using the `then` continuation and eliminating unnecessary bindings.

src/pages/billing/Project.js
ehigham: I added it in without profiling in an earlier change. I'm correcting the error of my ways :D

src/pages/billing/Project.js
andy7i: thanks for the explanations!

PR #2605, opened 2021-09-01
[IA-2918] updating namespace to use google project in order to fix notebook launch bug

xuscrbw:
[Jira Ticket IA-2918](https://broadworkbench.atlassian.net/browse/IA-2918)

While testing PPW, we discovered that the namespace was not being set correctly, which would cause notebooks to not load. Updated the used variable in the Notebook Launcher.

Tested this locally: had no issues with creating notebook, reattaching a new notebook with PD, stopping and starting


src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -354,7 +354,7 @@ const NotebookPreviewFrame = ({ notebookName, workspace: { workspace: { googlePr
 
 const JupyterFrameManager = ({ onClose, frameRef, details = {} }) => {
   Utils.useOnMount(() => {
-    Ajax().Metrics.captureEvent(Events.notebookLaunch, { 'Notebook Name': details.notebookName, 'Workspace Name': details.name, 'Workspace Namespace': details.namespace })
+    Ajax().Metrics.captureEvent(Events.notebookLaunch, { 'Notebook Name': details.notebookName, 'Workspace Name': details.name, 'Workspace Namespace': details.googleProject })
```
rtitle: I think the `Metrics` stuff still should capture workspace namespace/name, so this should probably be left as-is.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -444,15 +444,15 @@ const NotebookEditorFrame = ({ mode, notebookName, workspace: { workspace: { nam
       }),
       h(JupyterFrameManager, {
         frameRef,
-        onClose: () => Nav.goToPath('workspace-notebooks', { namespace, name }),
-        details: { notebookName, name, namespace }
+        onClose: () => Nav.goToPath('workspace-notebooks', { googleProject, name }),
```
rtitle: Also think `Nav.goToPath(…)` should still use `namespace`. The `workspace-notebooks` navPaths are still based on namespace: https://github.com/DataBiosphere/terra-ui/blob/2b2d8ca1640ca4293d8a48c443fdc74a36ea15bd/src/pages/workspaces/workspace/Notebooks.js#L486

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -444,15 +444,15 @@ const NotebookEditorFrame = ({ mode, notebookName, workspace: { workspace: { nam
...
+        details: { notebookName, name, googleProject }
```
rtitle: These details would probably need to contain `googleProject` _and_ `namespace` if both are used in `JupyterFrameManager`.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -504,8 +504,8 @@ const WelderDisabledNotebookEditorFrame = ({ mode, notebookName, workspace: { wo
       }),
       h(JupyterFrameManager, {
         frameRef,
-        onClose: () => Nav.goToPath('workspace-notebooks', { namespace, name }),
-        details: { notebookName, name, namespace }
+        onClose: () => Nav.goToPath('workspace-notebooks', { googleProject, name }),
```
rtitle: Same comment as above re `navPaths`

rtitle (changes requested):
Few comments above.. the Leo API calls look correct (those should all be using `googleProject`) but there are a few places which need to remain using `namespace`, I think.

(Let me know if I'm misunderstanding anything)

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -445,14 +445,14 @@ const NotebookEditorFrame = ({ mode, notebookName, workspace: { workspace: { nam
       h(JupyterFrameManager, {
         frameRef,
         onClose: () => Nav.goToPath('workspace-notebooks', { namespace, name }),
-        details: { notebookName, name, namespace }
+        details: { notebookName, name, namespace, googleProject }
```
rtitle: Looking again, it seems like `JupyterFrameManager` doesn't actually use `googleProject` so maybe it doesn't need to be passed here?

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -505,7 +505,7 @@ const WelderDisabledNotebookEditorFrame = ({ mode, notebookName, workspace: { wo
...
```
rtitle: Same as above

rtitle:
1 more minor comment, otherwise looks good. I also tested locally against PPW and it seems to work.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
kyuksel: Good catch! I wonder if we _should_ capture `googleProject` as well now that it may be distinct from `workspace.namespace`  🤔

kyuksel (changes requested):

kyuksel:
@xuscrbw `AnalysisLauncher` (added as part of the UI revamp work) duplicates a lot of the `NotebookLauncher` code and also contains `Ajax` calls that need updating.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -445,14 +445,14 @@ const NotebookEditorFrame = ({ mode, notebookName, workspace: { workspace: { nam
       h(JupyterFrameManager, {
         frameRef,
         onClose: () => Nav.goToPath('workspace-notebooks', { namespace, name }),
-        details: { notebookName, name, namespace }
+        details: { notebookName, name, namespace}
```
rtitle: Maybe needs a lint -- from the repo root directory can do:
```
yarn run eslint src --fix
```

rtitle (approved):
This looks good to me

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -1,3 +1,4 @@
+
```
kyuksel: The linter seems to miss this one  🤷🏼 
```suggestion
```

kyuksel (approved):
LGTM with a linting fix 👍🏼

petesantos (approved):

PR #2606, opened 2021-09-02
SUP-540 update wording when user is trying to delete a rstudio VM

Qi77Qi:
https://broadworkbench.atlassian.net/browse/SUP-540

New warning will show as follows for RStudio:

Screen Shot 2021-09-02 at 5 12 22 PM

For Jupyter:
Screen Shot 2021-09-02 at 5 14 15 PM

cc @ac3eb-br

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -498,7 +498,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   }
 
   const willDeletePersistentDisk = () => {
-    const { persistentDisk: existingPersistentDisk} = getExistingEnvironmentConfig()
+    const { persistentDisk: existingPersistentDisk } = getExistingEnvironmentConfig()
     return existingPersistentDisk && !canUpdatePersistentDisk()
   }
 
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -10,8 +10,7 @@ import { NumberInput, TextInput, ValidatedInput } from 'src/components/input'
 import { withModalDrawer } from 'src/components/ModalDrawer'
 import { tools } from 'src/components/notebook-utils'
 import { InfoBox } from 'src/components/PopupTrigger'
-import { SaveFilesHelp } from 'src/components/runtime-common'
-import { SaveFilesHelpRStudio } from 'src/components/runtime-common'
+import { SaveFilesHelp, SaveFilesHelpRStudio } from 'src/components/runtime-common'
 import TitleBar from 'src/components/TitleBar'
 import TooltipTrigger from 'src/components/TooltipTrigger'
 import { cloudServices, machineTypes } from 'src/data/machines'
@@ -968,6 +967,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   }
 
   const renderDeleteDiskChoices = () => {
+    const { runtime: existingRuntime } = getExistingEnvironmentConfig()
+
     return h(Fragment, [
       h(RadioBlock, {
         name: 'keep-persistent-disk',
@@ -1000,7 +1001,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           'Also deletes your application configuration and cloud compute profile.'
         ])
       ]),
-      getExistingEnvironmentConfig().runtime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
+      existingRuntime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
     ])
   }
 
@@ -1047,7 +1048,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                   'Since the persistent disk is not attached, the application configuration and cloud compute profile will remain.'
                 ])
               ]),
-              getExistingEnvironmentConfig().runtime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
+              existingRuntime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
             ])
           }],
           [existingRuntime && existingPersistentDisk, () => renderDeleteDiskChoices()],
@@ -1066,7 +1067,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                   'If you want to permanently save some files from the disk before deleting it, you will need to create a new cloud environment to access it.'
                 ])
               ]),
-              getExistingEnvironmentConfig().runtime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
+              existingRuntime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
             ])
           }],
           () => {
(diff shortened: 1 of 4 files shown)


src/components/ComputeModal.js
```diff
@@ -958,6 +967,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           canUpdateRuntime: !!canUpdateRuntime(),
           willDeleteBuiltinDisk: !!willDeleteBuiltinDisk(),
           willDeletePersistentDisk: !!willDeletePersistentDisk(),
+          willDeletePersistentDisk: !!willDeletePersistentDiskForRstudio(),
```
rtitle: This seems like a duplicate key

src/components/ComputeModal.js
```diff
@@ -495,8 +498,14 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   }
 
   const willDeletePersistentDisk = () => {
-    const { persistentDisk: existingPersistentDisk } = getExistingEnvironmentConfig()
-    return existingPersistentDisk && !canUpdatePersistentDisk()
+    const { persistentDisk: existingPersistentDisk, runtime: existingRuntime } = getExistingEnvironmentConfig()
+    return existingPersistentDisk && !canUpdatePersistentDisk() && existingRuntime.tool !== 'RStudio'
+  }
+
+  const willDeletePersistentDiskForRstudio = () => {
```
rtitle: Seems like some slight duplication.. wondering if there could be `willDeletePersistentDisk()` and `isToolRStudio()` methods to control the logic.

src/components/ComputeModal.js
```diff
@@ -1126,6 +1136,13 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
             ]),
             h(SaveFilesHelp)
           ])],
+          [willDeletePersistentDiskForRstudio(), () => h(Fragment, [
```
rtitle: And this could be:
```
[willDeletePersistentDisk(), () => h(Fragment, [
            p([
              'Reducing the size of a persistent disk requires it to be deleted and recreated. This will ',
              span({ style: { fontWeight: 600 } }, ['delete all files on the disk.'])
            ]),
            isToolRStudio() ? h(SaveFilesHelp) : h(SaveFilesHelpRStudio)
          ])],
```

src/components/runtime-common.js
```diff
@@ -123,6 +123,19 @@ export const SaveFilesHelp = () => {
...
+export const SaveFilesHelpRStudio = () => {
+  return h(Fragment, [
+    p([
+      'If you want to save some files permanently, such as input data, analysis outputs, or installed packages, ',
+      h(Link, {
+        href: 'https://support.terra.bio/hc/en-us/articles/360026639112',
+        ...Utils.newTabLinkProps
+      }, ['move them to the workspace bucket.'])
+    ]),
+    p(['Note: Unlike Jupyter noteooks, RStudio files are not autosaved to workspace bucket. Autosaving *.Rmd files will be supported soon. Stay tuned.'])
```
rtitle: Typo `noteooks`. Also maybe run this text by Alessandro (does he have a GitHub handle?)

rtitle (changes requested):
see above

src/components/runtime-common.js
kyuksel: His GitHub handle should be [ac3eb](https://github.com/ac3eb) but not sure if he's part of the org. I just submitted a request to get him added to it 🤞🏼

src/components/ComputeModal.js
Qi77Qi: I was initially thinking the whole text will look a bit different..but I think we end up only changing the text in `SaveFileHelp` part..so we should be able to remove `willDeletePersistentDiskForRStudio`

src/components/ComputeModal.js
```diff
@@ -499,6 +502,11 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     return existingPersistentDisk && !canUpdatePersistentDisk()
   }
 
+  const isToolRStudio = () => {
+    const { runtime: existingRuntime } = getExistingEnvironmentConfig()
+    return existingRuntime.tool === 'RStudio'
```
dmohs: I think `getExistingEnvironmentConfig` is providing adequate information, so this is clearer when shortened to:
```suggestion
    return getExistingEnvironmentConfig().runtime.tool === 'RStudio'
```

src/components/ComputeModal.js
```diff
@@ -1121,10 +1129,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           ])],
           [willDeletePersistentDisk(), () => h(Fragment, [
             p([
-              'Reducing the size of a persistent disk requires it to be deleted and recreated. This will ',
+              'To reduce the size of the PD, the existing PD will be deleted and a new one will be created and attached to your virtual machine instance. This will ',
               span({ style: { fontWeight: 600 } }, ['delete all files on the disk.'])
             ]),
-            h(SaveFilesHelp)
+            isToolRStudio() ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
```
dmohs: With the above, it may be nice to prevent the need for a lookup, as in:
```suggestion
            getExistingEnvironmentConfig().runtime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp) 
```
That way, you don't have to wonder where `isToolRStudio` is obtaining its information.

dmohs (approved):

src/components/runtime-common.js
Qi77Qi: yea, I couldn't find his handle..so sent on [slack](https://broadinstitute.slack.com/archives/GB1VAM14J/p1630593811142600) directly

kyuksel:
@Qi77Qi Could you please run `yarn run eslint src --fix` from the root of your `terra-ui` repo to have the linting errors fixed?

src/components/ComputeModal.js
```diff
@@ -298,12 +299,14 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     const cloudService = runtimeConfig?.cloudService
     const numberOfWorkers = runtimeConfig?.numberOfWorkers || 0
     const gpuConfig = runtimeConfig?.gpuConfig
+    const tool = currentRuntimeDetails?.labels.tool
```
```suggestion
    const tool = currentRuntimeDetails?.labels?.tool
```
so it doesn't explode in case `currentRuntimeDetails` doesn't have `labels`.

src/components/ComputeModal.js
```diff
@@ -1121,10 +1124,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           ])],
           [willDeletePersistentDisk(), () => h(Fragment, [
             p([
-              'Reducing the size of a persistent disk requires it to be deleted and recreated. This will ',
+              'To reduce the size of the PD, the existing PD will be deleted and a new one will be created and attached to your virtual machine instance. This will ',
               span({ style: { fontWeight: 600 } }, ['delete all files on the disk.'])
             ]),
-            h(SaveFilesHelp)
+            getExistingEnvironmentConfig().runtime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
```
kyuksel: The logic makes sense. It seems like there are other places in `ComputeModal` where `SaveFilesHelp` is rendered. We should apply the logic above in those places as well?

kyuksel (changes requested):
Looks good overall. See my comment inline about applying the conditional to other places in `ComputeModal`.

src/components/ComputeModal.js
```diff
@@ -11,6 +11,7 @@ import { withModalDrawer } from 'src/components/ModalDrawer'
 import { tools } from 'src/components/notebook-utils'
 import { InfoBox } from 'src/components/PopupTrigger'
 import { SaveFilesHelp } from 'src/components/runtime-common'
+import { SaveFilesHelpRStudio } from 'src/components/runtime-common'
```
```suggestion
import { SaveFilesHelp, SaveFilesHelpRStudio } from 'src/components/runtime-common'
```

rtitle (approved):
LGTM

kyuksel (approved):
:shipit:

Qi77Qi:
@rtitle @dmohs @kyuksel thank you all for reviewing! think this is the biggest terra UI PR I've ever done lol

PR #2608, opened 2021-09-03
[CA-1497] Add Billing Account Summary Panel to Billing Project

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1497
Add workspace billing account summary panel to billing project page. This panel summarises the status of the billing account changes applied to each workspace in the billing project. When there are errors, a link appears to contact support.
Mock-up here: https://www.sketch.com/s/59da26e3-a1a4-48f4-9a35-f39e2de22da9/a/bgaxn85
screenshot
(yes - I'm aware it's meant to be fizzbuzz :D )

### DIFF
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -1,6 +1,6 @@
 import _ from 'lodash/fp'
 import * as qs from 'qs'
-import { Fragment, useEffect, useState } from 'react'
+import { Fragment, useEffect, useMemo, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, HeaderRenderer, IdContainer, Link, Select, spinnerOverlay } from 'src/components/common'
 import { DeleteUserModal, EditUserModal, MemberCard, MemberCardHeaders, NewUserCard, NewUserModal } from 'src/components/group-common'
@@ -28,22 +28,21 @@ import { billingRoles } from 'src/pages/billing/List'
 const workspaceLastModifiedWidth = 150
 const workspaceExpandIconSize = 20
 
-const BillingAccountStatusIcon = {
-  Updating: 'loadingSpinner',
-  Done: 'success-standard',
-  Error: 'error-standard'
+const BillingAccountIcon = {
+  updating: { shape: 'sync', color: 'orange' },
+  done: { shape: 'check', color: colors.accent() },
+  error: { shape: 'warning-standard', color: 'red' }
 }
 
-const getBillingAccountStatusIconOrEmpty = (() => {
+const getBillingAccountIconOrEmpty = (() => {
   const size = 16
-  const blank = div({ style: { width: size } },
-    [div({ className: 'sr-only' }, ['Status'])])
-  return shape => icon(shape, { size }) || blank
+  const blank = div({ style: { width: size } }, [div({ className: 'sr-only' }, ['Status'])])
+  return ({ shape, color }) => icon(shape, { size, color }) || blank
 })()
 
 const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ sort, onSort }) => {
   return div({ style: { display: 'flex', justifyContent: 'space-between', marginTop: '1.5rem', padding: '0 1rem', marginBottom: '0.5rem' } }, [
-    getBillingAccountStatusIconOrEmpty(null),
+    getBillingAccountIconOrEmpty({ }),
     div({ 'aria-sort': ariaSort(sort, 'name'), style: { flex: 1, paddingLeft: '1rem' } }, [
       h(HeaderRenderer, { sort, onSort, name: 'name' })
     ]),
@@ -76,7 +75,7 @@ const ExpandedInfoRow = Utils.memoWithName('ExpandedInfoRow', ({ title, details,
   ])
 })
 
-const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, billingAccountStatusIcon, isExpanded, onExpand }) => {
+const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, billingAccountIcon, isExpanded, onExpand }) => {
   const { namespace, name, createdBy, lastModified, googleProject, billingAccountName } = workspace
   const workspaceCardStyles = {
     field: {
@@ -89,7 +88,7 @@ const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, billingA
   return div({ role: 'listitem', style: { ...Style.cardList.longCardShadowless, flexDirection: 'column' } }, [
     h(IdContainer, [id => h(Fragment, [
       div({ style: workspaceCardStyles.row }, [
-        getBillingAccountStatusIconOrEmpty(billingAccountStatusIcon),
+        getBillingAccountIconOrEmpty(billingAccountIcon),
         div({ style: { ...workspaceCardStyles.field, display: 'flex', alignItems: 'center', paddingLeft: '1rem' } }, [
           h(Link, {
             style: Style.noWrapEllipsis,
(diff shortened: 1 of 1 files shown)


src/pages/billing/Project.js
```diff
@@ -26,16 +28,21 @@ import { billingRoles } from 'src/pages/billing/List'
 const workspaceLastModifiedWidth = 150
 const workspaceExpandIconSize = 20
 
-const workspaceBillingStatusIconOrEmpty = (() => {
+const BillingAccountIcon = {
+  updating: { shape: 'sync', color: 'orange' },
```
s-rubenstein: You can probably use `colors.warning` here.

src/pages/billing/Project.js
```diff
@@ -26,16 +28,21 @@ import { billingRoles } from 'src/pages/billing/List'
...
+  done: { shape: 'check', color: colors.accent() },
+  error: { shape: 'warning-standard', color: 'red' }
```
s-rubenstein: You can probably use `colors.danger` here

src/pages/billing/Project.js
```diff
@@ -26,16 +28,21 @@ import { billingRoles } from 'src/pages/billing/List'
...
+}
+
+const getBillingAccountIconOrEmpty = (() => {
```
s-rubenstein: It looks like this takes an object with a shape and a color? Could we make those args? Right now it looks like this isn't called with anything. You could even make it:
`const getBillingAccountIconOrEmpty = ((shape, color) => {` and then you wouldn't need to pass in an empty object, the args would just default to undefined.

If you really want to pass it in as an object, you could do:
`const getBillingAccountIconOrEmpty = ((icon) => {`, and then have it be a check on icon down below, with either short circuiting and destructuring the arg, or just pulling out the properties of the object when needed.

src/pages/billing/Project.js
```diff
@@ -26,16 +28,21 @@ import { billingRoles } from 'src/pages/billing/List'
...
   const size = 16
-  const blank = div({ style: { width: size } },
-    [div({ className: 'sr-only' }, ['Status'])])
-  return shape => shape ? icon(shape, { size }) : blank
+  const blank = div({ style: { width: size } }, [div({ className: 'sr-only' }, ['Status'])])
+  return ({ shape, color }) => icon(shape, { size, color }) || blank
```
s-rubenstein: I think this will always return an icon? It just won't have a shape sometimes and thus won't render? But not 100% sure

src/pages/billing/Project.js
```diff
@@ -125,6 +132,53 @@ const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, billingA
...
 })
 
+const BillingAccountSummaryPanel = (() => {
+  const StatusAndCount = ({ status, count }) => div({ style: { display: 'float' } }, [
+    div({ style: { float: 'left' } }, [getBillingAccountIconOrEmpty(BillingAccountIcon[status])]),
+    div({ style: { float: 'left', marginLeft: '0.5rem' } }, [`${status} (${count})`])
+  ])
+
+  const maybeAddStatus = (status, count) => count > 0 && div({ style: { marginRight: '2rem' } },
+    [h(StatusAndCount, { status, count })])
```
s-rubenstein: Nit: I would put the `[` on the line above, and then put a newline before the `]`, just to follow Isaac's rule of the indentation of the closing brace being on the same indentation as the opening. Alternatively, you could follow that rule by just putting a newline before the last `)`. At first glance I thought this might have more children than just the StatusAndCount component.

src/pages/billing/Project.js
```diff
@@ -26,16 +28,21 @@ import { billingRoles } from 'src/pages/billing/List'
...
+  done: { shape: 'check', color: colors.accent() },
+  error: { shape: 'warning-standard', color: 'red' }
+}
+
+const getBillingAccountIconOrEmpty = (() => {
   const size = 16
-  const blank = div({ style: { width: size } },
-    [div({ className: 'sr-only' }, ['Status'])])
-  return shape => shape ? icon(shape, { size }) : blank
+  const blank = div({ style: { width: size } }, [div({ className: 'sr-only' }, ['Status'])])
```
s-rubenstein: Because this is returning the actual information for status, it feels a little weird to also be used for the header like this. I looked further down and saw it was also used for the status column on the workspaces list when everything is up to date. I think my preference would be for the text to say something like: `done` or for us to completely hide the column. Other than serving as a header for this column (which we want if there are icons), I think that this will cause noise if there's not some information there. `done` would provide a bit of information that makes it clear what this is reporting at least.

That would require pulling the header out into its own component, which I think would be my slight preference, just to make the distinction a bit clear between header and component (which follows the general repo pattern a bit more)

src/pages/billing/Project.js
```diff
@@ -26,16 +28,21 @@ import { billingRoles } from 'src/pages/billing/List'
 const workspaceLastModifiedWidth = 150
 const workspaceExpandIconSize = 20
 
-const workspaceBillingStatusIconOrEmpty = (() => {
+const BillingAccountIcon = {
```
s-rubenstein, line 32: Nice. I like this structure a lot.

src/pages/billing/Project.js
```diff
@@ -153,32 +207,32 @@ const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) =
 
   const adminCanEdit = _.filter(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers).length > 1
 
-  const getBillingAccountStatusIcon = workspace => Utils.cond(
-    [_.isEmpty(billingProject.workspacesWithIncorrectBillingAccount), () => null],
-    [billingProject.billingAccount === workspace.billingAccount, () => 'success-standard'],
-    ['billingAccountErrorMessage' in workspace, () => 'error-standard'],
-    [Utils.DEFAULT, () => 'loadingSpinner']
-  )
+  const workspacesInProject = useMemo(() => _.flow(
```
s-rubenstein, line 213: Great abstraction. Love this.

marctalbott (approved):

s-rubenstein (commented):
A few comments around component structure and code styling, otherwise looks good. Happy to take another later, and feel free to ping with any questions.

src/pages/billing/Project.js
s-rubenstein: The short circuiting could look like a line of:
`if(!icon) return blank`
and then `const { shape, color } = icon` and then the `||` at the end goes away.

src/pages/billing/Project.js
ehigham: Thanks!-

src/pages/billing/Project.js
ehigham: Thanks.

src/pages/billing/Project.js
ehigham: Thanks.

src/pages/billing/Project.js
ehigham: My idea was that `getBillingAccountIconOrEmpty` would have the following signature: 
```haskell
getBillingAccountIconOrEmpty :: Maybe BillingAccountStatus -> Icon
```
But this doesn't seem to work as in JavaScript I can't bind `null` to `{ shape, color }`. A peculiar language inconsistency.
I don't really like the idea of taking the pair `(shape, color)` because this breaks composition. I might make it take a `Maybe String` and that preserves some of the composition that I'm going for.

s-rubenstein: So, if you do the destructuring in the signature, the problem you get is that the interpreter is going to try to reach into a non-existent object when it is undefined. What you want is something like this:
```
const fn = (icon) => {
 if (icon === undefined) {
  console.log('undefined')
  return;
 }
 const {foo, bar} = icon
 console.log(`${foo}/${bar}`)
}
```
What that will do is give you a function that logs ${foo}/${bar} when icon is defined, and logs 'undefined' if you call it as `fn()`.

s-rubenstein: The reason you got the compiler error, which I am assuming was something like: `Uncaught TypeError: Cannot destructure property 'foo' of 'undefined' as it is undefined.` was that javascript was protecting you against a null pointer if you called it with no arg. Because otherwise it will try to index into undefined, which wouldn't work.

(As a side note, it will be undefined as the above, not null. They're actually two different constructs in JS.)

src/pages/billing/Project.js
ehigham: `icon`  returns `undefined` when you pass call it with a `null` shape

src/pages/billing/Project.js
ehigham: Sure - I've conditionally hid the column (and padded the next) when no icons are required. Let me know what you think.

src/pages/billing/Project.js
ehigham: Thanks for the explanation. I assumed that something like this was happening. I wish it didn't do this though. JS bites your head off in so many ways, it feels odd that it would protect you here, expecially when binding `null` to an object is a reasonable thing to do IMO.

s-rubenstein (approved):
LGTM

src/pages/billing/Project.js
```diff
@@ -153,32 +210,36 @@ const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) =
...
+    _.map('workspace'),
+    _.filter({ namespace: billingProject.projectName })
+  )(workspaces), [billingProject, workspaces])
+
+  const groups = groupByBillingAccountStatus(billingProject, workspacesInProject)
+  const billingAccountsOutOfDate = !_.every(_.isEmpty, [groups.error, groups.updating])
+  const getBillingAccountStatus = workspace => _.findKey(g => g.has(workspace), groups)
```
dmohs, line 220: I might rename `g` to `ws` ("workspaces set") or `set` to make it clear that the members of `groups` are sets.

src/pages/billing/Project.js
```diff
@@ -125,6 +134,54 @@ const WorkspaceCard = Utils.memoWithName('WorkspaceCard', ({ workspace, billingA
...
+          ['contact us regarding unresolved errors']),
+        '.'
+      ])
+    ]))
+})()
+
+const groupByBillingAccountStatus = (billingProject, workspaces) => {
+  const group = workspace => Utils.cond(
+    [billingProject.billingAccount === workspace.billingAccount, () => 'done'],
+    ['billingAccountErrorMessage' in workspace, () => 'error'],
```
dmohs, line 178: Prefer `_.includes` to the somewhat unpredictable `in`?

dmohs (approved):
Apologies for the delay.

src/pages/billing/Project.js
ehigham: Is there a generic way of telling if a collection contains some element? The fact I'm using a set is an implementation detail (for `O(1)` lookup) - really this code should work for any collection.

src/pages/billing/Project.js
ehigham:
```javascript
> _.includes('foo', { foo: 'bar' })
false
```
?

src/pages/billing/Project.js
dmohs: Take a quick look at the Lodash documentation for `_.some`. It has some magic that may work for your use case here. That said, I think your usage of `Set` here clarifies the intent and may be the best choice.

src/pages/billing/Project.js
dmohs: I meant `_.has` :)

dmohs:
Just to be clear, you were right to merge this without waiting for my responses to your questions. Even if we find something we like better, the main value is in improving _future_ code.

PR #2610, opened 2021-09-03
[CA-1501] audit namespace usages

s-rubenstein:
The main thing I wasn't sure on was requester pays. That feels like maybe a google project thing, but I think ultimately it routes through rawls, which means I don't think it should change. Curious y'alls understanding.

There are also probably other props which can go, but I decided cleaning it all up didn't need to happen right now.

### DIFF
--- a/src/components/RequesterPaysModal.js
+++ b/src/components/RequesterPaysModal.js
@@ -65,11 +65,11 @@ const RequesterPaysModal = ({ onDismiss, onSuccess }) => {
           value: selectedGoogleProject,
           placeholder: 'Select a workspace',
           onChange: ({ value }) => setSelectedGoogleProject(value),
-          options: _.uniq(_.map(workspace => {
+          options: _.sortBy(option => option.label, _.uniq(_.map(workspace => {
             return {
               value: workspace.workspace.googleProject, label: `${workspace.workspace.namespace}/${workspace.workspace.name}`
             }
-          }, workspaceList)).sort()
+          }, workspaceList)))
         }),
         requesterPaysHelpInfo
       ])])
--- a/src/components/RequesterPaysModal.js
+++ b/src/components/RequesterPaysModal.js
@@ -12,27 +12,28 @@ import { requesterPaysProjectStore } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
 
-const billingHelpInfo = div({ style: { paddingTop: '1rem' } }, [
+const requesterPaysHelpInfo = div({ style: { paddingTop: '1rem' } }, [
   h(Link, {
     href: 'https://support.terra.bio/hc/en-us/articles/360029801491',
     ...Utils.newTabLinkProps
-  }, ['Why is billing required to access this data?', icon('pop-out', { style: { marginLeft: '0.25rem' }, size: 12 })])
+  }, ['Why is a workspace required to access this data?', icon('pop-out', { style: { marginLeft: '0.25rem' }, size: 12 })])
 ])
 
 const RequesterPaysModal = ({ onDismiss, onSuccess }) => {
   const [loading, setLoading] = useState(false)
-  const [billingList, setBillingList] = useState([])
-  const [selectedBilling, setSelectedBilling] = useState(requesterPaysProjectStore.get())
+  const [workspaceList, setWorkspaceList] = useState([])
+  const [selectedGoogleProject, setSelectedGoogleProject] = useState(requesterPaysProjectStore.get())
   const signal = Utils.useCancellation()
 
   Utils.useOnMount(() => {
-    const loadBillingProjects = _.flow(
+    const loadWorkspaces = _.flow(
       Utils.withBusyState(setLoading),
-      withErrorReporting('Error loading billing projects')
+      withErrorReporting('Error loading workspaces')
     )(async () => {
-      setBillingList(await Ajax(signal).Billing.listProjects())
+      const workspaces = await Ajax(signal).Workspaces.list()
+      setWorkspaceList(_.filter(workspace => workspace.accessLevel === 'OWNER' || workspace.accessLevel === 'PROJECT_OWNER', workspaces))
     })
-    loadBillingProjects()
+    loadWorkspaces()
   })
 
   return Utils.cond(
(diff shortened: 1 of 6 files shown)


andy7i (commented):
> The main thing I wasn't sure on was requester pays. That feels like maybe a google project thing, but I think ultimately it routes through rawls, which means I don't think it should change. Curious y'alls understanding.

Good question.. I'm not sure OTTOMH if/how requesterPays changes with PPW. We could verify this by looking through the rawls code. It might lead to discovering some bugs

s-rubenstein:
Yeah, I can go rawls diving on tuesday.

src/components/data/data-utils.js
```diff
@@ -52,7 +52,7 @@ export const getDownloadCommand = (fileName, gsUri, accessUrl) => {
     return `gsutil cp ${gsUri} ${fileName || '.'}`
   }
 }
-
+// should this one change?
```
gpolumbo-broad: Depends on where/how `getUserProjectForWorkspace` is used.

s-rubenstein:
Added a change to requester pays which has it now using google project. This means that users need to have a workspace to try out requester pays. We have no way to check `can-compute` on the workspace from list, so the UI is currently just showing `OWNER` and `PROJECT_OWNER` workspaces, since we can be sure those permissions have `can-compute`

src/components/RequesterPaysModal.js
```diff
@@ -12,27 +12,28 @@ import { requesterPaysProjectStore } from 'src/libs/state'
...
   Utils.useOnMount(() => {
-    const loadBillingProjects = _.flow(
+    const loadWorkspaces = _.flow(
       Utils.withBusyState(setLoading),
-      withErrorReporting('Error loading billing projects')
+      withErrorReporting('Error loading workspaces')
     )(async () => {
-      setBillingList(await Ajax(signal).Billing.listProjects())
+      const workspaces = await Ajax(signal).Workspaces.list()
+      setWorkspaceList(_.filter(workspace => workspace.accessLevel === 'OWNER' || workspace.accessLevel === 'PROJECT_OWNER', workspaces))
```
gpolumbo-broad, line 34: Are the `accessLevel`s here the equivalent of roles in Sam?  If so, have we verified that this is the complete list of roles we want to filter on here?

gpolumbo-broad (approved):
The UX is changing here for RequesterPays.  That's fine with me, I think this change is correct for making things work properly.  However, we should run this by Janki to at least let her know.  So long as these proposed changes are not seen as release blockers, then I propose we move ahead as is and any changes/improvements that Janki wants to make to this we can make after PPW launches.

src/components/RequesterPaysModal.js
```diff
@@ -12,27 +12,28 @@ import { requesterPaysProjectStore } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
 
-const billingHelpInfo = div({ style: { paddingTop: '1rem' } }, [
+const requesterPaysHelpInfo = div({ style: { paddingTop: '1rem' } }, [
   h(Link, {
     href: 'https://support.terra.bio/hc/en-us/articles/360029801491',
     ...Utils.newTabLinkProps
-  }, ['Why is billing required to access this data?', icon('pop-out', { style: { marginLeft: '0.25rem' }, size: 12 })])
+  }, ['Why is a workspace required to access this data?', icon('pop-out', { style: { marginLeft: '0.25rem' }, size: 12 })])
```
marctalbott, line 19: Is this article still going to be relevant?

src/components/RequesterPaysModal.js
marctalbott: Writers with compute access should also be able to use the workspace IMO since they have permission to use the workspace's Google project. Owners and project owners get compute access automatically

src/components/RequesterPaysModal.js
s-rubenstein: I let Allie know to change it.

src/components/RequesterPaysModal.js
s-rubenstein: So, I would love to be able to do that, but unfortunately 'can-compute' doesn't actually return with the workspace list response, and we don't necessarily want to query getWorkspace N times for that information. I think it could be a future change, but I am very worried about including that here for scope creep/releasing PPW in a reasonable timeframe. I agree that the preference would be for writers with compute access should absolutely be able to use the workspace.

s-rubenstein: https://github.com/broadinstitute/rawls/blob/develop/core/src/main/resources/swagger/api-docs.yaml#L6010 is what this accessLevel is referring to.

s-rubenstein:
@marctalbott I checked in with Janki, she is fine with both this UX and `OWNER`/`PROJECT_OWNER` for now.

src/components/RequesterPaysModal.js
```diff
@@ -44,42 +45,46 @@ const RequesterPaysModal = ({ onDismiss, onSuccess }) => {
...
-          onChange: ({ value }) => setSelectedBilling(value),
-          options: _.uniq(_.map('projectName', billingList)).sort()
+          value: selectedGoogleProject,
+          placeholder: 'Select a workspace',
+          onChange: ({ value }) => setSelectedGoogleProject(value),
+          options: _.uniq(_.map(workspace => {
+            return {
+              value: workspace.workspace.googleProject, label: `${workspace.workspace.namespace}/${workspace.workspace.name}`
+            }
+          }, workspaceList)).sort()
```
petesantos: I don't think this sort is doing anything anymore, can you confirm?
If we do want it to sort, can we use lodash's `sortBy`?

petesantos (approved):
Could you take another look at the sort before merging?

src/components/RequesterPaysModal.js
s-rubenstein: Yes, the sort doesn't seem to do anything. I will switch to using lodash's sort method, since I think we can do some sorting there.

src/components/RequesterPaysModal.js
```diff
@@ -44,42 +45,46 @@ const RequesterPaysModal = ({ onDismiss, onSuccess }) => {
...
-          onChange: ({ value }) => setSelectedBilling(value),
-          options: _.uniq(_.map('projectName', billingList)).sort()
+          value: selectedGoogleProject,
+          placeholder: 'Select a workspace',
+          onChange: ({ value }) => setSelectedGoogleProject(value),
+          options: _.sortBy(option => option.label, _.uniq(_.map(workspace => {
+            return {
+              value: workspace.workspace.googleProject, label: `${workspace.workspace.namespace}/${workspace.workspace.name}`
+            }
+          }, workspaceList)))
```
petesantos: nit: I think it is good to switch to `flow` once we hit three nested functions for clarity - it can help future readers by not making them "unwrap" the code from the inside out mentally.

nit: On the map/destructure - I prefer the destructure because it makes it a bit more obvious that all of the values are coming from the same place in the object

```suggestion
          options: _.flow(
            _.map(({ workspace: { workspace: { googleProject, namespace, name } } }) => ({
                value: googleProject, label: `${namespace}/${name}`
            })),
            _.uniq,
            _.sortBy('label')
          )(workspaceList)
```

petesantos (approved):

marctalbott (approved):

PR #2615, opened 2021-09-09
AS-879: Add support for tsv async imports 

ansingh7115:
This PR is a duplicate of https://github.com/DataBiosphere/terra-ui/pull/2561 which was merged prematurely and had to be reverted. See that PR for discussion. This PR should only be merged once necessary changes are released in the backend.

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -20,7 +20,8 @@ import colors from 'src/libs/colors'
 import { reportError } from 'src/libs/error'
 import Events from 'src/libs/events'
 import { FormLabel } from 'src/libs/forms'
-import { requesterPaysProjectStore } from 'src/libs/state'
+import { notify } from 'src/libs/notifications'
+import { pfbImportJobStore, requesterPaysProjectStore } from 'src/libs/state'
 import * as StateHistory from 'src/libs/state-history'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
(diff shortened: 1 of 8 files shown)


sagehen03 (approved):

src/components/data/data-utils.js
```diff
@@ -234,7 +242,9 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
     setUploading(true)
     try {
       const workspace = Ajax().Workspaces.workspace(namespace, name)
-      await (useFireCloudDataModel ? workspace.importEntitiesFile : workspace.importFlexibleEntitiesFile)(file)
+      const { jobId } = await (useFireCloudDataModel ? workspace.importEntitiesFile : workspace.importFlexibleEntitiesFileAsync)(file)
```
davidangb: I believe that right here, `file.size` will give you the size, in bytes, of the file-to-be-uploaded.

For a follow-on PR, or in this one if you feel like tackling it, I _thiiiink_ we can do something like:
```
      let filesize = Number.MAX_SAFE_INTEGER
      try {
        filesize = file.size
      } catch (err) {
        // noop
      }
      if (filesize < 524288) { // 512k, let's choose a good number, I chose this arbitrarily
        await (useFireCloudDataModel ? workspace.importEntitiesFile : workspace.importFlexibleEntitiesFile)(false, file)
      } else {
        const { jobId } = await (useFireCloudDataModel ? workspace.importEntitiesFile : workspace.importFlexibleEntitiesFile)(true, file)
        asyncImportJobStore.update(Utils.append({ targetWorkspace: { namespace, name }, jobId }))
        notifyDataImportProgress(jobId)
      }
```

dmohs (approved):
I appreciate that the previous suggested changes have been preserved in this version of the PR. Thank you for that.

davidangb (approved):

PR #2617, opened 2021-09-10
DR-2058 Snapshot export UI for data catalog

myessail:
snapshotmodal

Once we have actual snapshots on the explorer page, we can populate the link to the export with real UUIDs and it will actually work. Followup ticket is [DR-2090](https://broadworkbench.atlassian.net/browse/DR-2090)


src/pages/library/common.js
```diff
@@ -192,8 +195,32 @@ export const selectionActionComponent = (selectedData, setSelectedData) => {
...
+              setShowExportModal(true)
+            }
           }, ['Save to a workspace'])
+        ]),
+        showExportModal &&
+        h(Modal, {
+          title: 'Save to a Terra workspace',
+          onDismiss: () => setShowExportModal(false),
+          okButton: () => {
+            Nav.history.push({
```
lu-c: Would `Nav.goToPath` be a better choice? I'm not sure what the difference is between that and `Nav.history.push`

src/pages/library/common.js
```diff
@@ -354,7 +384,7 @@ export const SearchAndFilterComponent = (featuredList, sidebarSections, activeTa
         ])
       ]),
 
-    selectionActionComponent(selectedData, setSelectedData)
+    SelectionActionComponent(selectedData, setSelectedData)
```
kyuksel: We use components with the following format:

```suggestion
    h(SelectionActionComponent, { selectedData, setSelectedData })
```

src/pages/library/common.js
```diff
@@ -192,8 +195,32 @@ export const selectionActionComponent = (selectedData, setSelectedData) => {
...
           }, ['Save to a workspace'])
+        ]),
+        showExportModal &&
+        h(Modal, {
+          title: 'Save to a Terra workspace',
+          onDismiss: () => setShowExportModal(false),
+          okButton: () => {
+            Nav.history.push({
+              pathname: Nav.getPath('import-data'),
+              search: `?url=https://jade.datarepo-dev.broadinstitute.org/&snapshotId=idid&snapshotName=${datasetName}&format=snapshot`
```
kyuksel: Just to make it more obvious, and using the correct data repo url:
```suggestion
              search: `?url=${getConfig().dataRepoUrlRoot}&snapshotId=REPLACE_ME&snapshotName=${datasetName}&format=snapshot`
```

src/pages/library/common.js
```diff
@@ -427,7 +457,8 @@ const makeTable = (listData, sort, setSort, sortDir, setSortDir, selectedData, t
         ),
         div({ style: { ...styles.table.flexTableRow, alignItems: 'flex-start' } }, [
           div({ style: { ...styles.table.col, ...styles.table.firstElem } }, [
-            icon(listdatum.locked ? 'lock' : 'lock-o', { size: 12, style: { flex: 'none', color: listdatum.locked ? colors.accent() : colors.primary() } })
+            icon(listdatum.locked ? 'lock' : 'lock-o',
+              { size: 12, style: { flex: 'none', color: listdatum.locked ? colors.accent() : colors.primary() } })
```
kyuksel: Thanks for reformatting this the way we normally do :)

```suggestion
            icon(listdatum.locked ? 'lock' : 'lock-o', {
              size: 12, style: { flex: 'none', color: listdatum.locked ? colors.accent() : colors.primary() }
            })
```

src/pages/library/common.js
```diff
@@ -166,7 +167,9 @@ export const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSectio
   ])
 }
 
-export const selectionActionComponent = (selectedData, setSelectedData) => {
+export const SelectionActionComponent = (selectedData, setSelectedData) => {
```
kyuksel: 👍🏼  on capitalizing this correctly. We do also need to pass the inputs as part of an object in order to call it with `h`, like:

```suggestion
const SelectionActionComponent = ({ selectedData, setSelectedData }) => {
```

Removed `export` since this component is only used within this file.

src/pages/library/common.js
```diff
@@ -192,8 +195,32 @@ export const selectionActionComponent = (selectedData, setSelectedData) => {
           }, 'Cancel'),
           h(ButtonPrimary, {
             style: { textTransform: 'none', fontSize: 14 },
-            onClick: () => {}
+            onClick: () => {
+              setShowExportModal(true)
+            }
```
kyuksel:
```suggestion
            onClick: () => setShowExportModal(true)
```

src/pages/library/common.js
```diff
@@ -192,8 +195,32 @@ export const selectionActionComponent = (selectedData, setSelectedData) => {
...
           }, ['Save to a workspace'])
+        ]),
+        showExportModal &&
+        h(Modal, {
```
kyuksel: We don't newline for these cases:

```suggestion
        showExportModal && h(Modal, {
```

src/pages/library/common.js
```diff
@@ -192,8 +195,32 @@ export const selectionActionComponent = (selectedData, setSelectedData) => {
...
+          div([
+            div('Data with this name will be saved in Terra. If this name already exists, it will be overwritten.', []),
+            div({ style: { marginTop: '1rem' } }, [
+              h(TextInput, {
+                placeholder: 'Name of data',
+                value: datasetName,
+                onChange: setDatasetName
+              })
+            ])
+          ])
```
kyuksel: The diff below is to show some simplifications:

```suggestion
          'Data with this name will be saved in Terra. If this name already exists, it will be overwritten.',
          div({ style: { marginTop: '1rem' } }, [
            h(TextInput, {
              placeholder: 'Name of data',
              value: datasetName,
              onChange: setDatasetName
            })
          ])
```

Also, we require `TextInput` to have a `label` for accessibility. Please see other instances in the code for examples of it.

kyuksel (commented):
I like the animated description :)

src/pages/library/common.js
lu-c: I looked it up a bit because I was working on other tickets that require Navigation as well, nav.goToPath will call Nav.history.push, so they're essentially the same thing. I'd still recommend using GoToPath though, just in case we add some hooks to pathing at some point

lu-c (approved):
LGTM, Thanks for fixing up the formatting!

kyuksel (approved):
👍🏼   pending fixing the linting errors:

```
src/pages/library/common.js
  Line 9:30:    'TextInput' is defined but never used          
  Line 173:23:  'setDatasetName' is assigned a value but never used
```

kyuksel:
@myessail FYI that you have the PR review approvals. I noticed that an integration test run had failed, which should be irrelevant to your changes so I started a re-run. If that passes, you're good to merge it :)

src/pages/library/common.js
```diff
@@ -205,6 +206,7 @@ export const SearchAndFilterComponent = ({ featuredList, sidebarSections, active
   const [searchFilter, setSearchFilter] = useState('')
   const [sort, setSort] = useState({ field: 'created', direction: 'desc' })
   const [requestDatasetAccessList, setRequestDatasetAccessList] = useState()
+  const [datasetName] = useState('')
```
zarsky-broad: Mind sticking the setter on here for now, even if it's not yet used? Just so people don't get confused.
```suggestion
  const [datasetName, setDatasetName] = useState('')
```

PR #2620, opened 2021-09-13
Ensure Workflow Status on Job Submission Details is fully visible BW-806

cahrens:
This is a cleanup MR related to https://github.com/DataBiosphere/terra-ui/pull/2599. It addresses 2 things.

1) I was not aware that the Workflow Status could be more than 1 line. Consequently, when I made the Comments area on the Submission Details page line up with the other entries to its right, it meant that multi-line Workflow Status gets cut off/obscured. I have changed the layout to allow the Workflow Status to take as much room as necessary. This means that long, multi-line (with newline characters) comments may make the top portion of the Submission Details page grow, but given that the field is limited to 1000 characters it will not grow indefinitely. Here are some screen shots of how it looks for different data:

image
_______________

image

_______________

image

2) Although I didn't change the submission date column on the Job History page, the space taken by the new comment column means that the submission date moves to 3 lines sooner for smaller browser sizes. In that case, the lack of vertical padding between rows makes the text difficult to read. So similar to what we did for the submission ID field (which is typically 3 lines), I reduced the text size to 12 points (and made the default size consistent with the submission ID field).

image

_______________

image

### DIFF
--- a/src/pages/workspaces/workspace/JobHistory.js
+++ b/src/pages/workspaces/workspace/JobHistory.js
@@ -33,6 +33,9 @@ const styles = {
   statusDetailCell: {
     align: 'center',
     style: { padding: '0.5rem' }
+  },
+  multiLineCellText: {
+    fontSize: 12
   }
 }
 
@@ -300,13 +303,13 @@ const JobHistory = _.flow(
               }
             },
             {
-              size: { basis: 170, grow: 0 },
+              size: { basis: 150, grow: 0 },
               field: 'submissionDate',
               headerRenderer: makeHeaderRenderer('submissionDate', 'Submitted'),
               cellRenderer: ({ rowIndex }) => {
                 const { submissionDate } = sortedSubmissions[rowIndex]
                 const dateParts = Utils.makeCompleteDateParts(submissionDate)
-                return div([
+                return div({ style: styles.multiLineCellText }, [
                   div([dateParts[0]]),
                   div([dateParts[1]])
                 ])
@@ -320,7 +323,7 @@ const JobHistory = _.flow(
               cellRenderer: ({ rowIndex }) => {
                 const { submissionId } = sortedSubmissions[rowIndex]
                 return h(Link, {
-                  style: { fontSize: 12 },
+                  style: styles.multiLineCellText,
                   ...Utils.newTabLinkProps,
                   href: bucketBrowserUrl(`${bucketName}/${submissionId}`)
                 }, [submissionId])
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -158,7 +158,7 @@ const SubmissionDetails = _.flow(
     submissionDetailsBreadcrumbSubtitle(namespace, name, submissionId),
     _.isEmpty(submission) ? centeredSpinner() : h(Fragment, [
       div({ style: { display: 'grid', gridTemplateColumns: '1fr 4fr' } }, [
-        div({ style: { display: 'grid', gridAutoRows: 'min-content' } }, [
+        div({ style: { display: 'grid', gridTemplateRows: '1fr auto' } }, [
           makeSection('Workflow Statuses', [
             succeeded && makeStatusLine(successIcon, `Succeeded: ${succeeded.length}`, { marginTop: '0.5rem' }),
             failed && makeStatusLine(failedIcon, `Failed: ${failed.length}`, { marginTop: '0.5rem' }),
@@ -166,17 +166,19 @@ const SubmissionDetails = _.flow(
             submitted && makeStatusLine(submittedIcon, `Submitted: ${submitted.length}`, { marginTop: '0.5rem' })
           ]),
           div({
-            style: { padding: '0 0.5rem 0.5rem', marginTop: '0.5rem', whiteSpace: 'pre', overflow: 'hidden' }
+            style: { padding: '0 0.5rem 0.5rem', marginTop: '1.0rem', whiteSpace: 'pre', overflow: 'hidden' }
           }, [
             h4({ style: Style.elements.sectionHeader }, [
               'Comment',
               h(ButtonSecondary, {
-                style: { marginLeft: '0.50em', height: '1.0em' },
+                style: { marginLeft: '0.50rem', height: '1.0rem' },
                 tooltip: 'Edit Comment',
                 onClick: () => setUpdatingComment(true)
               }, [icon('edit')])
             ]),
-            div({ style: { textOverflow: 'ellipsis', overflow: 'hidden' } }, [firstLine(userComment)])
+            div({ style: { height: '1.0rem', textOverflow: 'ellipsis', overflow: 'hidden' } },
+              [firstLine(userComment)]
+            )
           ]),
           updatingComment && h(UpdateUserCommentModal, {
             workspace: { name, namespace }, submissionId, userComment,
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -150,7 +150,7 @@ const SubmissionDetails = _.flow(
     submissionDetailsBreadcrumbSubtitle(namespace, name, submissionId),
     _.isEmpty(submission) ? centeredSpinner() : h(Fragment, [
       div({ style: { display: 'grid', gridTemplateColumns: '1fr 4fr' } }, [
-        div({ style: { display: 'grid', gridTemplateRows: 'repeat(3, 33%)' } }, [
+        div({ style: { display: 'grid', gridAutoRows: 'min-content'} }, [
           makeSection('Workflow Statuses', [
             succeeded && makeStatusLine(successIcon, `Succeeded: ${succeeded.length}`, { marginTop: '0.5rem' }),
             failed && makeStatusLine(failedIcon, `Failed: ${failed.length}`, { marginTop: '0.5rem', marginBottom: '0.5rem' }),
@@ -158,10 +158,7 @@ const SubmissionDetails = _.flow(
             submitted && makeStatusLine(submittedIcon, `Submitted: ${submitted.length}`, { marginTop: '0.5rem' })
           ]),
           div({
-            style: {
-              padding: '0 0.5rem 0.5rem', marginTop: '1rem',
-              whiteSpace: 'pre', overflow: 'hidden', gridRow: '2 / 4'
-            }
+            style: { padding: '0 0.5rem 0.5rem', whiteSpace: 'pre', overflow: 'hidden' }
           }, [
             h4({ style: Style.elements.sectionHeader }, [
               'Comment',
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -143,6 +143,14 @@ const SubmissionDetails = _.flow(
     ])
   }
 
+  /**
+   * If the text has multiple lines, return just the first line of text with ellipses to indicate truncation.
+   */
+  const firstLine = text => {
+    const lines = text ? text.split('\n') : []
+    return lines.length > 1 ? `${lines[0]} ...` : lines.length > 0 ? lines[0] : undefined
+  }
+
   /*
    * Page render
    */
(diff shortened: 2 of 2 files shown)


src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -300,13 +303,13 @@ const JobHistory = _.flow(
               }
             },
             {
-              size: { basis: 170, grow: 0 },
+              size: { basis: 150, grow: 0 },
```
cjllanwarne, line 306: Am I reading this right? You made this field slightly narrower at the same time as making the font smaller?

I thought a `grow: 0` field would never resize, so I'm confused why this column was ever getting too narrow (or perhaps the problem was that it was never wide enough?)

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -150,18 +150,15 @@ const SubmissionDetails = _.flow(
...
             failed && makeStatusLine(failedIcon, `Failed: ${failed.length}`, { marginTop: '0.5rem', marginBottom: '0.5rem' }),
             running && makeStatusLine(runningIcon, `Running: ${running.length}`, { marginTop: '0.5rem' }),
             submitted && makeStatusLine(submittedIcon, `Submitted: ${submitted.length}`, { marginTop: '0.5rem' })
           ]),
           div({
-            style: {
-              padding: '0 0.5rem 0.5rem', marginTop: '1rem',
-              whiteSpace: 'pre', overflow: 'hidden', gridRow: '2 / 4'
-            }
+            style: { padding: '0 0.5rem 0.5rem', whiteSpace: 'pre', overflow: 'hidden' }
```
cjllanwarne: Just curious: what does omitting the `marginTop` and `gridRow` values do here?

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -150,18 +150,15 @@ const SubmissionDetails = _.flow(
     submissionDetailsBreadcrumbSubtitle(namespace, name, submissionId),
     _.isEmpty(submission) ? centeredSpinner() : h(Fragment, [
       div({ style: { display: 'grid', gridTemplateColumns: '1fr 4fr' } }, [
-        div({ style: { display: 'grid', gridTemplateRows: 'repeat(3, 33%)' } }, [
+        div({ style: { display: 'grid', gridAutoRows: 'min-content'} }, [
```
cjllanwarne: So previously we forced 3 equal sized rows, what happens now? Do we have as many rows as we need, each of which resizes up or down according to content?

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -33,6 +33,9 @@ const styles = {
   statusDetailCell: {
     align: 'center',
     style: { padding: '0.5rem' }
+  },
+  multiLineCellText: {
+    fontSize: 12
```
cjllanwarne, line 38: This doesn't seem very specifically multi-line? I'm admittedly very css-naive but isn't it better just to say `fontSize: 12` inline whenever we need to?

cjllanwarne:
My comments were more in the spirit of "help me understand" than necessarily implying that anything is wrong

src/pages/workspaces/workspace/JobHistory.js
cahrens: Because I made the font smaller, I thought that it was OK to also make the field smaller-- I decided to make the size consistent with the submission ID field next to it (also ran this by Brian). When the date wraps to 3 lines it still looks OK because of the additional vertical spacing.

There are separate variables for grow and shrink. This field will not grow, but it can shrink. In contrast, the submission ID field is not allowed to shrink because if it does, the ID starts rendering outside of the cell because it can only break at specific places.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cahrens: marginTop was adding some space to make the alignment work out better when I wanted the comments field to align with the elements on the right hand side. Now that I am not attempting that, I don't want the extra space.

gridRow 2 / 4 was saying that this field (comments) should take up rows 2 and 3. Again, this was in my attempt to align with the items on the right.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cahrens: Yes. I attached screenshots to the MR description.

src/pages/workspaces/workspace/JobHistory.js
cahrens: My rationale here was that we are using a smaller font size for elements that are expected to render over multiple lines. I want the font size for date and submission ID to be the same (I wouldn't want 12 for one and 13 for the other, for example), so I defined a variable.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
salonishah11: TOL: What if we truncate the comment after x number of lines / have a fixed height? Similar to what we do for the width of the comment?

cahrens: I don't think it's necessary for the following reasons:

1. I think that it is unlikely that users will put newlines in their comments.
2. Even if users do put newlines in their comments, I don't expect the total number to be incredibly large given that the comments field is limited to 1000 characters.
3. Even if users do something crazy like a newline every other character, all the information is still on the page… they just have to scroll down to it.

Trying to truncate the comment after x number of lines adds more complexity for very little gain.

salonishah11:
@cahrens - In Submission Details page, when the comment is too long (the one in screenshots would be almost 1000 characters) the size of the height of the table is significantly reduced (first image). And when I am viewing this page on my laptop (13inch laptop; not sure what is the minimum screen size we support), I can't see the table contents at all (second image). 

On Monitor:

Screen Shot 2021-09-13 at 5 12 09 PM
-------
On Laptop: 

Screen Shot 2021-09-13 at 5 15 53 PM

cahrens:
Thanks for the thorough testing, @salonishah11! I expected that the content of the page would be within a scrollbar, but apparently it is not. I'll take a look at this tomorrow.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
kyuksel: @cjllanwarne [This](https://bitsofco.de/understanding-the-difference-between-grid-template-and-grid-auto) might help clarify the distinction between `grid-template-*` and `grid-auto-*`.

cahrens:
@salonishah11 let me know what you think of the change I pushed up. This was my reasoning:

Since we don't put the contents of the page in a scroll pane, I do not want to grow the region above the table to accommodate the comment area. Originally that region was 230px high, and I need to plan for the worse case scenario, where there are 4 Workflow Status entries. In that case, a single line of comment text only slightly increases the region to 233px. Therefore I will only show the first line of comments, and give an indication of the truncation. 

Obviously when there are less than 4 Workflow Status entries, I could theoretically show more lines of comments. But given that newlines are likely to be fairly rare, and there is no automatic vertical truncation via CSS, going that route doesn't seem worth it to me (plus the Job History table also shows only the first line, so there is some precedence for that). But if you disagree, I could see allowing the number of lines of comments to be _5 - # of Workflow Status entries_.

image

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -143,35 +143,40 @@ const SubmissionDetails = _.flow(
...
    */
   return div({ style: { padding: '1rem 2rem 2rem', flex: 1, display: 'flex', flexDirection: 'column' } }, [
     submissionDetailsBreadcrumbSubtitle(namespace, name, submissionId),
     _.isEmpty(submission) ? centeredSpinner() : h(Fragment, [
       div({ style: { display: 'grid', gridTemplateColumns: '1fr 4fr' } }, [
-        div({ style: { display: 'grid', gridTemplateRows: 'repeat(3, 33%)' } }, [
+        div({ style: { display: 'grid', gridAutoRows: 'min-content' } }, [
           makeSection('Workflow Statuses', [
             succeeded && makeStatusLine(successIcon, `Succeeded: ${succeeded.length}`, { marginTop: '0.5rem' }),
-            failed && makeStatusLine(failedIcon, `Failed: ${failed.length}`, { marginTop: '0.5rem', marginBottom: '0.5rem' }),
```
cahrens, line 156: I don't know why there was a marginBottom set below the failed only.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -143,35 +143,40 @@ const SubmissionDetails = _.flow(
...
-              padding: '0 0.5rem 0.5rem', marginTop: '1rem',
-              whiteSpace: 'pre', overflow: 'hidden', gridRow: '2 / 4'
-            }
+            style: { padding: '0 0.5rem 0.5rem', marginTop: '0.5rem', whiteSpace: 'pre', overflow: 'hidden' }
           }, [
             h4({ style: Style.elements.sectionHeader }, [
               'Comment',
               h(ButtonSecondary, {
-                style: { marginLeft: '0.50em' },
+                style: { marginLeft: '0.50em', height: '1.0em' },
```
cahrens: Reducing the height (which doesn't change the size of the actual icon) so the amount of space between "Comment" and the comment text is consistent with other fields on the page.

zarsky-broad:
You could use a `Collapse` if you want, there are existing examples

cahrens:
@zarsky-broad thanks for the suggestion! I tried it out, and given where we are shoe-horning this feature in, I don't think the Collapse is a good option. When you expand/collapse it, everything shifts (and for comments that take up a lot of space you wouldn't want it initially expanded). We would still want horizontal truncation and newlines respected, but I don't know if that model will work with the Collapse given the grid/flex layout of the page.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -143,35 +143,40 @@ const SubmissionDetails = _.flow(
     ])
   }
 
+  /**
+   * If the text has multiple lines, return just the first line of text with ellipses to indicate truncation.
+   */
+  const firstLine = text => {
+    const lines = text ? text.split('\n') : []
+    return lines.length > 1 ? `${lines[0]} ...` : lines.length > 0 ? lines[0] : undefined
+  }
```
petesantos: nit: We tend to prefer using lodash functions for consistency and for error handling. In this case `_.split` is better at handling bad values (e.g. number) on and will not throw an error. Additionally, these functions are auto curried.
```suggestion
  const firstLine = _.flow(
    _.split('\n'),
    lines => lines.length > 1 ? `${lines[0]} ...` : lines[0]
  )
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -143,35 +143,40 @@ const SubmissionDetails = _.flow(
...
+
   /*
    * Page render
    */
   return div({ style: { padding: '1rem 2rem 2rem', flex: 1, display: 'flex', flexDirection: 'column' } }, [
     submissionDetailsBreadcrumbSubtitle(namespace, name, submissionId),
     _.isEmpty(submission) ? centeredSpinner() : h(Fragment, [
       div({ style: { display: 'grid', gridTemplateColumns: '1fr 4fr' } }, [
-        div({ style: { display: 'grid', gridTemplateRows: 'repeat(3, 33%)' } }, [
+        div({ style: { display: 'grid', gridAutoRows: 'min-content' } }, [
```
petesantos: I like your solution by restricting the comment to a single line in the display.
If UX and Product likes that as well, I would recommend:
```suggestion
        div({ style: { display: 'grid', gridTemplateRows: '1fr auto' } }, [
```
As this will give the status section and the comment section consistent heights. I think the UX of entering a multiline comment and then only seeing one line when there is a lot of whitespace underneath can make it feel like the comment wasn't saved correctly. If there is only enough space for 1 line, it may seem more natural to the user that their multiline comment was clipped.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -143,35 +143,40 @@ const SubmissionDetails = _.flow(
...
+            failed && makeStatusLine(failedIcon, `Failed: ${failed.length}`, { marginTop: '0.5rem' }),
             running && makeStatusLine(runningIcon, `Running: ${running.length}`, { marginTop: '0.5rem' }),
             submitted && makeStatusLine(submittedIcon, `Submitted: ${submitted.length}`, { marginTop: '0.5rem' })
           ]),
           div({
-            style: {
-              padding: '0 0.5rem 0.5rem', marginTop: '1rem',
-              whiteSpace: 'pre', overflow: 'hidden', gridRow: '2 / 4'
-            }
+            style: { padding: '0 0.5rem 0.5rem', marginTop: '0.5rem', whiteSpace: 'pre', overflow: 'hidden' }
```
petesantos: If you do reduce the height to one line - you will need to adjust the top margin:
```suggestion
            style: { padding: '0 0.5rem 0.5rem', marginTop: '1rem', whiteSpace: 'pre', overflow: 'hidden' }
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -143,35 +143,40 @@ const SubmissionDetails = _.flow(
...
               'Comment',
               h(ButtonSecondary, {
-                style: { marginLeft: '0.50em' },
+                style: { marginLeft: '0.50em', height: '1.0em' },
                 tooltip: 'Edit Comment',
                 onClick: () => setUpdatingComment(true)
               }, [icon('edit')])
             ]),
-            div({ style: { textOverflow: 'ellipsis', overflow: 'hidden' } }, [userComment])
+            div({ style: { textOverflow: 'ellipsis', overflow: 'hidden' } }, [firstLine(userComment)])
```
petesantos: Again, if you reduce the height to 1 line, add height to the div to keep it consistent
```suggestion
            div({ style: { height: '1rem', textOverflow: 'ellipsis', overflow: 'hidden' } }, [firstLine(userComment)])
```

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
petesantos: We use `rem` in the codebase rather than `em`
```suggestion
                style: { marginLeft: '0.50rem', height: '1.0rem' },
```

petesantos (commented):
@cahrens I personally prefer your solution of trimming the comment down to one line in the detail display. It keeps the view simple, consistent and predictable.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cahrens: Thanks for the suggestion-- definitely simpler! I'm still unfamiliar with these functions, but will try to remember to look through them next time that I'm defining a utility method.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cahrens: I like it! It solves both the problem of the comment field not aligning with anything, and any potential confusion about the truncation. Thanks for taking the time to play around with it!

salonishah11 (approved):
LGTM 👍  I like with the idea of just showing 1 line in Submission Details page and having a detailed view inside the Modal.

petesantos (approved):
LGTM

PR #2622, opened 2021-09-14
[Dr 2902] Set up routing for new page, and basic layout for the page

lu-c:
Screen Shot 2021-09-13 at 8 21 43 PM

This is a sanity check in to prevent future prs from getting too long!


src/pages/workspaces/workspace/Data.js
```diff
@@ -450,7 +450,7 @@ const WorkspaceData = _.flow(
   const getRunningImportJobs = async () => {
     try {
       const runningJobs = await Ajax(signal).Workspaces.workspace(namespace, name).listImportJobs(true)
-      _.map(job => {
+      _.forEach(job => {
```
lu-c: The linter was complaining about this map having no return - I figured if there was no return expected here, this would essentially be running a forEach.

Let me know if I should revert!

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+      name: 'CDC'
+    },
+    subjects: 10,
+    dataType: 'RNA Seq',
+    keepCollapsed: true,
+    locked: false,
+    files: 15,
+    fileSize: 4432
+  }
+}
```
kyuksel: `id` seems unused.
We can also save some space by removing the optional `return` and the braces around it.

```suggestion
const getSnapshot = () => ({
  namespace: 'harry-potter',
  name: 'Harry Potter',
  created: '2020-01-07T18:25:28.340Z',
  lastUpdated: '2020-01-07T18:25:28.340Z',
  tags: {
    itemsType: 'AttributeValue',
    items: ['1000 Genomes', 'CMG', 'Open Access', 'Exome']
  },
  description: 'The boy who lived',
  lowerName: 'harry potter',
  lowerDescription: 'lower description',
  project: {
    id: '112',
    name: 'CDC'
  },
  subjects: 10,
  dataType: 'RNA Seq',
  keepCollapsed: true,
  locked: false,
  files: 15,
  fileSize: 4432
})
```

kyuksel: If the object returned doesn't need to be parameterized, it doesn't have to be a function.

src/components/data/data-utils.js
```diff
@@ -259,7 +259,7 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
           notifyDataImportProgress(jobId)
         }
       }
-      
+
```
kyuksel: Thanks for cleaning up the spaces. Maybe add a newline before the if block at line 245?

src/pages/library/DataBrowserDetails.js
kyuksel: @petesantos and I were also chatting about this. If the use case for the code above is to mock an Ajax call, Pete may have suggestions for the structuring of it.

src/pages/workspaces/workspace/Data.js
dmohs: Thanks for this.

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+    libraryTopMatter(activeTab),
+    !snapshot ?
+      centeredSpinner() :
+      h(Fragment, [
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', width: '100%' } }, [
+          div({ style: styles.page }, [
+            icon('angle-left', { size: 35 })
+          ]),
+          getMainContent(snapshot),
+          getSidebar(snapshot)
```
dmohs: I like this style in languages that support it—start with main functions calling supporting functions which are defined later in the file. This opposes our current style, which is define before use.

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+    subjects: 10,
+    dataType: 'RNA Seq',
+    keepCollapsed: true,
+    locked: false,
+    files: 15,
+    fileSize: 4432
+  }
+}
+
+const DataBrowserDetails = routeParams => {
```
kyuksel: For function components, our convention is to use the term `props`:
```suggestion
const DataBrowserDetails = props => {
```

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+  const id = routeParams.id
+  const snapshot = getSnapshot(id)
+
+  return h(FooterWrapper, { alwaysShow: true }, [
+    libraryTopMatter(activeTab),
+    !snapshot ?
```
dmohs, line 63: This will never be true. I generally like the idea of skeletons like this, so I would update this to use a promise fulfilled on a delay so you can see the case with the spinner.

src/pages/library/DataBrowserDetails.js
kyuksel: And since `id` seems to be the only property used from `routeParams`, we could destructure it like below:

```suggestion
const DataBrowserDetails = ({ id }) => {
```

and `id` will be available in the code below.

src/components/data/data-utils.js
dmohs: I like this change, but not in a file that would otherwise be untouched.

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+          getSidebar(snapshot)
+        ])
+      ])
+  ])
+}
+
+const getMainContent = snapshot => {
+  return div({ style: { ...styles.page, width: '100%', marginTop: 0 } }, [
+    h1(snapshot.name, { style: { marginTop: 0 } }),
+    div([JSON.stringify(snapshot)])
```
dmohs: When you are sketching things out like this, I recommend adding a `whiteSpace: 'pre-wrap'` to format the JSON just to make it easier to read. You'd also need to `JSON.stringify(snapshot, nil, 2)`.

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+}
+
+const DataBrowserDetails = routeParams => {
+  const id = routeParams.id
+  const snapshot = getSnapshot(id)
+
+  return h(FooterWrapper, { alwaysShow: true }, [
+    libraryTopMatter(activeTab),
+    !snapshot ?
+      centeredSpinner() :
```
kyuksel, line 64: TOL: Expanding on a previous comment, `centeredSpinner()` won't take effect unless some mocking with delay is put in.

P.S. We use the `TOL` (i.e. Thinking Out Loudly) prefix to indicate that it may be something we could improve on but no immediate action is necessary.

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+      h(Fragment, [
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', width: '100%' } }, [
+          div({ style: styles.page }, [
+            icon('angle-left', { size: 35 })
+          ]),
+          getMainContent(snapshot),
+          getSidebar(snapshot)
+        ])
+      ])
```
kyuksel: TOL: This may be a good use case for CSS Grid, which may simplify mark-up and styling that gets rendered to the page.

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+}
+
+const getMainContent = snapshot => {
+  return div({ style: { ...styles.page, width: '100%', marginTop: 0 } }, [
+    h1(snapshot.name, { style: { marginTop: 0 } }),
+    div([JSON.stringify(snapshot)])
+  ])
+}
+
+const getSidebar = snapshot => {
```
kyuksel: Since `snapshot` is unused:

```suggestion
const Sidebar = () => {
```

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+            icon('angle-left', { size: 35 })
+          ]),
+          getMainContent(snapshot),
+          getSidebar(snapshot)
+        ])
+      ])
+  ])
+}
+
+const getMainContent = snapshot => {
```
kyuksel: Similarly as before, we want to specify inputs to a component in an object. Here, we also do some destructuring to make referencing easier later below.
```suggestion
const MainContent = ({ snapshot, snapshot: { name }}) => {
```

src/pages/library/DataBrowserDetails.js
lu-c: Ok, I'll move it up in the page then!

src/components/data/data-utils.js
lu-c: Should I revert it then? NP if so, just wanted to check

src/pages/library/DataBrowserDetails.js
kyuksel:
```suggestion
          h(MainContent, { snapshot }),
          h(Sidebar)
```
We're dropping the input argument `snapshot` for the `Sidebar` call since it isn't used at the moment. See comment further down on `Sidebar` implementation.

src/components/data/data-utils.js
dmohs: Yes. Sorry about not being clear.

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,77 @@
...
+  return div({ style: { ...styles.page, width: '100%', marginTop: 0 } }, [
+    h1(snapshot.name, { style: { marginTop: 0 } }),
```
kyuksel: Now that we did the destructuring above, we can use the property `name` directly as below:
```suggestion
    h1({ style: { marginTop: 0 } }, [name]),
```
Note that the child has to be the second parameter to `h1`.

src/pages/library/DataBrowserDetails.js
kyuksel: `div` isn't needed since `h1` on the line above it would create its own block on the page.
```suggestion
    JSON.stringify(snapshot)
```

src/pages/workspaces/workspace/Data.js
kyuksel: The conversion makes sense. Because what's happening within the `forEach` is an effect, it may make sense to utilize `useEffect()`. We can revisit this later.

src/components/data/data-utils.js
kyuksel: Feel free to ignore the comment above in the light of your thread with @dmohs .

src/pages/library/DataBrowserDetails.js
lu-c: Done!

src/pages/library/DataBrowserDetails.js
lu-c: I have a follow up PR that will use the snapshot in the Sidebar call - https://github.com/DataBiosphere/terra-ui/pull/2623/files

Since this was a PR focused on just setting up the basics, I thought it would be fine for this to be here since it will be used.

src/pages/library/DataBrowserDetails.js
lu-c: Updated to use the ({id}) suggestion

src/pages/library/DataBrowserDetails.js
lu-c: I'll make the change for the function names, but I will skip the destructuring because we do not know the format of the json yet

src/pages/library/DataBrowserDetails.js
lu-c: Im now applying styles to the snapshot, so I'll leave it in the div! Plus, it's a placeholder :)

src/components/data/data-utils.js
lu-c: I might have to do this anyway, I'm afraid the linter might be complaining again >.<

src/pages/library/DataBrowserDetails.js
lu-c: How can I tell the difference between when to minimize code? It seems like we shouldn't need to wrap these elements with the `h`, because it's rendering without being explicitly declared. The get-- methods are returning hyperscript objects that are rendering properly

petesantos: I can see how this can be confusing. The key idea is that react components conceptually are passed two parameters when they're created: 

1. Props - An object whose key/value pairs are the properties passed in to the component
2. Children - A list of child components that will be rendered by this component somewhere inside of it

You can create a function that returns a react component - which is what the `get` functions are doing. But since the intent of the code is to render something to the page (correct me if I am wrong), it would be best to define these functions as [react function components](https://reactjs.org/docs/hooks-state.html#hooks-and-function-components)

A react component needs to have a capitalized name, and should accept a list of props as a parameter - as a convention in this code base, we name this parameter `props`. Making this a function component will make it easy to add children to it later on and further enhance the functionality, for example adding state, later on without having to change anything.

**TL;DR** When in doubt, make a Component, rendered using `h`. This makes sure React will know what to do with it in any situation.

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,82 @@
...
+import { div, h, h1 } from 'react-hyperscript-helpers'
+import FooterWrapper from 'src/components/FooterWrapper'
+import { centeredSpinner, icon } from 'src/components/icons'
+import { libraryTopMatter } from 'src/components/library-common'
+import * as Utils from 'src/libs/utils'
+
+
+const activeTab = 'browse & explore'
+const styles = {
+  page: { padding: 20, marginTop: 15 }
```
petesantos: TOL: Would this be more accurate as "content"

lu-c: sure, i'll push up this change

src/pages/library/DataBrowserDetails.js
lu-c: Hey, I'm sorry, I wasnt able to make this change - This error popped up on this change

Screen Shot 2021-09-15 at 5 18 51 PM

src/pages/library/DataBrowserDetails.js
```diff
@@ -0,0 +1,82 @@
...
+    subjects: 10,
+    dataType: 'RNA Seq',
+    keepCollapsed: true,
+    locked: false,
+    files: 15,
+    fileSize: 4432
+  })
+}, 2000))
+
+const MainContent = snapshot => {
```
petesantos: This needs to be destructured to pull the snapshot out
```suggestion
const MainContent = ({ snapshot }) => {
```

petesantos (approved):
LGTM - just needs one minor correction on the MainContent parameterization

tcjiang9 (approved):

PR #2624, opened 2021-09-14
[IA-2930][IA-2701] Implement R file syncing and proper RStudio analysis preview/launch flow

jdcanas:
Testing this PR depends on manually updating leo conf file to point to a newer welder hash than exists in dev, and hotswapping it to your fiab (see linked PR for necessary code change). This means this PR **cannot be merged** until [this](https://github.com/DataBiosphere/leonardo/pull/2295/files) is.

This PR ensures storageLinks is called on RStudio application launch, which results in any `.Rmd` files in the users bucket in the `notebooks/` folder being localized for editing. Any files saved in the RStudio editor will then be delocalized back to the bucket.

Today, the only way to create and interact with .Rmd files is the RStudio editor. With the analysis tab, we will also support Rmd preview and launching from the preview page, where we also interface with localization. This PR updates this, and touches up the buttons to reflect that RStudio does not launch the editor with a particular file in mind, where as Jupyter does. See screenshots below.

RStudio Preview:
image

RStudio with files localized:
image

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -614,7 +614,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         p(['Terra attaches a persistent disk (PD) to your cloud compute in order to provide an option to keep the data on the disk after you delete your compute. PDs also act as a safeguard to protect your data in the case that something goes wrong with the compute.']),
         p(['A minimal cost per hour is associated with maintaining the disk even when the cloud compute is paused or deleted.']),
         p(['If you delete your cloud compute, but keep your PD, the PD will be reattached when creating the next cloud compute.']),
-        h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360047318551', ...Utils.newTabLinkProps }, [
+        h(Link, { 'aria-label': `Disk doc`, href: 'https://support.terra.bio/hc/en-us/articles/360047318551', ...Utils.newTabLinkProps }, [
           'Learn more about persistent disks',
           icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
         ])
@@ -687,14 +687,14 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
             ]),
             div([
               'Custom environments ', b(['must ']), 'be based off one of the ',
-              h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images'])
+              h(Link, { 'aria-label': `Terra base images`, href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images'])
             ])
           ])
         }],
         [Utils.DEFAULT, () => {
           return h(Fragment, [
             div({ style: { display: 'flex' } }, [
-              h(Link, { onClick: () => setViewMode('packages') }, ['What’s installed on this environment?']),
+              h(Link, { 'aria-label': `Image packages`, onClick: () => setViewMode('packages') }, ['What’s installed on this environment?']),
               makeImageInfo({ marginLeft: 'auto' })
             ])
           ])
@@ -771,6 +771,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                 enableGpusSpan
             ]),
             h(Link, {
+              'aria-label': `GPU doc`,
               style: { marginLeft: '1rem', verticalAlign: 'top' },
               href: 'https://support.terra.bio/hc/en-us/articles/4403006001947', ...Utils.newTabLinkProps
             }, [
@@ -934,10 +935,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         p([
           'You are about to create a virtual machine using an unverified Docker image. ',
           'Please make sure that it was created by you or someone you trust, using one of our ',
-          h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['base images.']),
+          h(Link, { 'aria-label': `Terra base images`, href: terraBaseImages, ...Utils.newTabLinkProps }, ['base images.']),
           ' Custom Docker images could potentially cause serious security issues.'
         ]),
-        h(Link, { href: safeImageDocumentation, ...Utils.newTabLinkProps }, ['Learn more about creating safe and secure custom Docker images.']),
+        h(Link, { 'aria-label': `Safe image doc`, href: safeImageDocumentation, ...Utils.newTabLinkProps }, ['Learn more about creating safe and secure custom Docker images.']),
         p(['If you\'re confident that your image is safe, you may continue using it. Otherwise, go back to select another image.'])
       ]),
       div({ style: { display: 'flex', justifyContent: 'flex-end', marginTop: '1rem' } }, [
@@ -951,7 +952,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     const makeJSON = value => div({ style: { whiteSpace: 'pre-wrap', fontFamily: 'Menlo, monospace' } }, [JSON.stringify(value, null, 2)])
     return showDebugger ?
       div({ style: { position: 'fixed', top: 0, left: 0, bottom: 0, right: '50vw', backgroundColor: 'white', padding: '1rem', overflowY: 'auto' } }, [
-        h(Link, { onClick: () => setShowDebugger(false), style: { position: 'absolute', top: 0, right: 0 } }, ['x']),
+        h(Link, { 'aria-label': `Show debugger`, onClick: () => setShowDebugger(false), style: { position: 'absolute', top: 0, right: 0 } }, ['x']),
         makeHeader('Old Environment Config'),
         makeJSON(getExistingEnvironmentConfig()),
         makeHeader('New Environment Config'),
@@ -964,7 +965,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           willRequireDowntime: !!willRequireDowntime()
         })
       ]) :
-      h(Link, { onClick: () => setShowDebugger(true), style: { position: 'fixed', top: 0, left: 0, color: 'white' } }, ['D'])
+      h(Link, { 'aria-label': `Show debugger`, onClick: () => setShowDebugger(true), style: { position: 'fixed', top: 0, left: 0, color: 'white' } }, ['D'])
   }
 
   const renderDeleteDiskChoices = () => {
@@ -1227,7 +1228,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
               ul({ style: { paddingLeft: '1rem', marginBottom: 0, lineHeight: 1.5 } }, [
                 li([
                   div([packageLabel]),
-                  h(Link, { onClick: () => setViewMode('packages') }, ['What’s installed on this environment?'])
+                  h(Link, { 'aria-label': `Packages`, onClick: () => setViewMode('packages') }, ['What’s installed on this environment?'])
                 ]),
                 li({ style: { marginTop: '1rem' } }, [
                   'Default compute size of ', span({ style: { fontWeight: 600 } }, [cpu, ' CPU(s)']), ', ',
@@ -1237,7 +1238,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                     h(Fragment, ['a ', renderDiskText(), ' to keep your data even after you delete your compute'])
                 ]),
                 li({ style: { marginTop: '1rem' } }, [
-                  h(Link, { onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
+                  h(Link, { 'aria-label': `Disk doc`, onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
                 ])
               ])
             ]),
@@ -1265,7 +1266,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           !sparkMode && !isPersistentDisk && div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
             div([
               'Time to upgrade your cloud environment. Terra’s new persistent disk feature will safeguard your work and data. ',
-              h(Link, { onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
+              h(Link, { 'aria-label': `Disk doc`, onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
             ]),
             h(ButtonOutline, {
               style: { marginTop: '1rem' },
@@ -1301,7 +1302,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           label({ htmlFor: id, style: styles.label }, ['Persistent disk size (GB)']),
           div({ style: { marginTop: '0.5rem' } }, [
             'Persistent disks store analysis data. ',
-            h(Link, { onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about persistent disks and where your disk is mounted.'])
+            h(Link, { 'aria-label': `Disk doc`, onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about persistent disks and where your disk is mounted.'])
           ]),
           h(NumberInput, {
             id,
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -934,7 +934,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         p([
           'You are about to create a virtual machine using an unverified Docker image. ',
           'Please make sure that it was created by you or someone you trust, using one of our ',
-          h(Link, { 'aria-label': 'Terra base images', href: terraBaseImages, ...Utils.newTabLinkProps }, ['base images.']),
+          h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra base images.']),
           ' Custom Docker images could potentially cause serious security issues.'
         ]),
         h(Link, { href: safeImageDocumentation, ...Utils.newTabLinkProps }, ['Learn more about creating safe and secure custom Docker images.']),
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -951,7 +951,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     const makeJSON = value => div({ style: { whiteSpace: 'pre-wrap', fontFamily: 'Menlo, monospace' } }, [JSON.stringify(value, null, 2)])
     return showDebugger ?
       div({ style: { position: 'fixed', top: 0, left: 0, bottom: 0, right: '50vw', backgroundColor: 'white', padding: '1rem', overflowY: 'auto' } }, [
-        h(Link, { 'aria-label': 'Show debugger', onClick: () => setShowDebugger(false), style: { position: 'absolute', top: 0, right: 0 } }, ['x']),
+        h(Link, { 'aria-label': 'Hide debugger', onClick: () => setShowDebugger(false), style: { position: 'absolute', top: 0, right: 0 } }, ['x']),
         makeHeader('Old Environment Config'),
         makeJSON(getExistingEnvironmentConfig()),
         makeHeader('New Environment Config'),
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -614,7 +614,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         p(['Terra attaches a persistent disk (PD) to your cloud compute in order to provide an option to keep the data on the disk after you delete your compute. PDs also act as a safeguard to protect your data in the case that something goes wrong with the compute.']),
         p(['A minimal cost per hour is associated with maintaining the disk even when the cloud compute is paused or deleted.']),
         p(['If you delete your cloud compute, but keep your PD, the PD will be reattached when creating the next cloud compute.']),
-        h(Link, { 'aria-label': `Disk doc`, href: 'https://support.terra.bio/hc/en-us/articles/360047318551', ...Utils.newTabLinkProps }, [
+        h(Link, { 'aria-label': 'Disk doc', href: 'https://support.terra.bio/hc/en-us/articles/360047318551', ...Utils.newTabLinkProps }, [
           'Learn more about persistent disks',
           icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
         ])
@@ -687,14 +687,14 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
             ]),
             div([
               'Custom environments ', b(['must ']), 'be based off one of the ',
-              h(Link, { 'aria-label': `Terra base images`, href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images'])
+              h(Link, { 'aria-label': 'Terra base images', href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images'])
             ])
           ])
         }],
         [Utils.DEFAULT, () => {
           return h(Fragment, [
             div({ style: { display: 'flex' } }, [
-              h(Link, { 'aria-label': `Image packages`, onClick: () => setViewMode('packages') }, ['What’s installed on this environment?']),
+              h(Link, { 'aria-label': 'Image packages', onClick: () => setViewMode('packages') }, ['What’s installed on this environment?']),
               makeImageInfo({ marginLeft: 'auto' })
             ])
           ])
@@ -771,7 +771,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                 enableGpusSpan
             ]),
             h(Link, {
-              'aria-label': `GPU doc`,
+              'aria-label': 'GPU doc',
               style: { marginLeft: '1rem', verticalAlign: 'top' },
               href: 'https://support.terra.bio/hc/en-us/articles/4403006001947', ...Utils.newTabLinkProps
             }, [
@@ -935,10 +935,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         p([
           'You are about to create a virtual machine using an unverified Docker image. ',
           'Please make sure that it was created by you or someone you trust, using one of our ',
-          h(Link, { 'aria-label': `Terra base images`, href: terraBaseImages, ...Utils.newTabLinkProps }, ['base images.']),
+          h(Link, { 'aria-label': 'Terra base images', href: terraBaseImages, ...Utils.newTabLinkProps }, ['base images.']),
           ' Custom Docker images could potentially cause serious security issues.'
         ]),
-        h(Link, { 'aria-label': `Safe image doc`, href: safeImageDocumentation, ...Utils.newTabLinkProps }, ['Learn more about creating safe and secure custom Docker images.']),
+        h(Link, { 'aria-label': 'Safe image doc', href: safeImageDocumentation, ...Utils.newTabLinkProps }, ['Learn more about creating safe and secure custom Docker images.']),
         p(['If you\'re confident that your image is safe, you may continue using it. Otherwise, go back to select another image.'])
       ]),
       div({ style: { display: 'flex', justifyContent: 'flex-end', marginTop: '1rem' } }, [
@@ -952,7 +952,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     const makeJSON = value => div({ style: { whiteSpace: 'pre-wrap', fontFamily: 'Menlo, monospace' } }, [JSON.stringify(value, null, 2)])
     return showDebugger ?
       div({ style: { position: 'fixed', top: 0, left: 0, bottom: 0, right: '50vw', backgroundColor: 'white', padding: '1rem', overflowY: 'auto' } }, [
-        h(Link, { 'aria-label': `Show debugger`, onClick: () => setShowDebugger(false), style: { position: 'absolute', top: 0, right: 0 } }, ['x']),
+        h(Link, { 'aria-label': 'Show debugger', onClick: () => setShowDebugger(false), style: { position: 'absolute', top: 0, right: 0 } }, ['x']),
         makeHeader('Old Environment Config'),
         makeJSON(getExistingEnvironmentConfig()),
         makeHeader('New Environment Config'),
@@ -965,7 +965,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           willRequireDowntime: !!willRequireDowntime()
         })
       ]) :
-      h(Link, { 'aria-label': `Show debugger`, onClick: () => setShowDebugger(true), style: { position: 'fixed', top: 0, left: 0, color: 'white' } }, ['D'])
+      h(Link, { 'aria-label': 'Show debugger', onClick: () => setShowDebugger(true), style: { position: 'fixed', top: 0, left: 0, color: 'white' } }, ['D'])
   }
 
   const renderDeleteDiskChoices = () => {
@@ -1228,7 +1228,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
               ul({ style: { paddingLeft: '1rem', marginBottom: 0, lineHeight: 1.5 } }, [
                 li([
                   div([packageLabel]),
-                  h(Link, { 'aria-label': `Packages`, onClick: () => setViewMode('packages') }, ['What’s installed on this environment?'])
+                  h(Link, { 'aria-label': 'Packages', onClick: () => setViewMode('packages') }, ['What’s installed on this environment?'])
                 ]),
                 li({ style: { marginTop: '1rem' } }, [
                   'Default compute size of ', span({ style: { fontWeight: 600 } }, [cpu, ' CPU(s)']), ', ',
@@ -1238,7 +1238,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                     h(Fragment, ['a ', renderDiskText(), ' to keep your data even after you delete your compute'])
                 ]),
                 li({ style: { marginTop: '1rem' } }, [
-                  h(Link, { 'aria-label': `Disk doc`, onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
+                  h(Link, { 'aria-label': 'Disk doc', onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
                 ])
               ])
             ]),
@@ -1266,7 +1266,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           !sparkMode && !isPersistentDisk && div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
             div([
               'Time to upgrade your cloud environment. Terra’s new persistent disk feature will safeguard your work and data. ',
-              h(Link, { 'aria-label': `Disk doc`, onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
+              h(Link, { 'aria-label': 'Disk doc', onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
             ]),
             h(ButtonOutline, {
               style: { marginTop: '1rem' },
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -1301,7 +1301,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           label({ htmlFor: id, style: styles.label }, ['Persistent disk size (GB)']),
           div({ style: { marginTop: '0.5rem' } }, [
             'Persistent disks store analysis data. ',
-            h(Link, { 'aria-label': `Disk doc`, onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about persistent disks and where your disk is mounted.'])
+            h(Link, { onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about persistent disks and where your disk is mounted.'])
           ]),
           h(NumberInput, {
             id,
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -614,7 +614,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         p(['Terra attaches a persistent disk (PD) to your cloud compute in order to provide an option to keep the data on the disk after you delete your compute. PDs also act as a safeguard to protect your data in the case that something goes wrong with the compute.']),
         p(['A minimal cost per hour is associated with maintaining the disk even when the cloud compute is paused or deleted.']),
         p(['If you delete your cloud compute, but keep your PD, the PD will be reattached when creating the next cloud compute.']),
-        h(Link, { 'aria-label': 'Disk doc', href: 'https://support.terra.bio/hc/en-us/articles/360047318551', ...Utils.newTabLinkProps }, [
+        h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360047318551', ...Utils.newTabLinkProps }, [
           'Learn more about persistent disks',
           icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
         ])
@@ -694,7 +694,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         [Utils.DEFAULT, () => {
           return h(Fragment, [
             div({ style: { display: 'flex' } }, [
-              h(Link, { 'aria-label': 'Image packages', onClick: () => setViewMode('packages') }, ['What’s installed on this environment?']),
+              h(Link, { onClick: () => setViewMode('packages') }, ['What’s installed on this environment?']),
               makeImageInfo({ marginLeft: 'auto' })
             ])
           ])
@@ -771,7 +771,6 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                 enableGpusSpan
             ]),
             h(Link, {
-              'aria-label': 'GPU doc',
               style: { marginLeft: '1rem', verticalAlign: 'top' },
               href: 'https://support.terra.bio/hc/en-us/articles/4403006001947', ...Utils.newTabLinkProps
             }, [
@@ -938,7 +937,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           h(Link, { 'aria-label': 'Terra base images', href: terraBaseImages, ...Utils.newTabLinkProps }, ['base images.']),
           ' Custom Docker images could potentially cause serious security issues.'
         ]),
-        h(Link, { 'aria-label': 'Safe image doc', href: safeImageDocumentation, ...Utils.newTabLinkProps }, ['Learn more about creating safe and secure custom Docker images.']),
+        h(Link, { href: safeImageDocumentation, ...Utils.newTabLinkProps }, ['Learn more about creating safe and secure custom Docker images.']),
         p(['If you\'re confident that your image is safe, you may continue using it. Otherwise, go back to select another image.'])
       ]),
       div({ style: { display: 'flex', justifyContent: 'flex-end', marginTop: '1rem' } }, [
@@ -1228,7 +1227,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
               ul({ style: { paddingLeft: '1rem', marginBottom: 0, lineHeight: 1.5 } }, [
                 li([
                   div([packageLabel]),
-                  h(Link, { 'aria-label': 'Packages', onClick: () => setViewMode('packages') }, ['What’s installed on this environment?'])
+                  h(Link, { onClick: () => setViewMode('packages') }, ['What’s installed on this environment?'])
                 ]),
                 li({ style: { marginTop: '1rem' } }, [
                   'Default compute size of ', span({ style: { fontWeight: 600 } }, [cpu, ' CPU(s)']), ', ',
@@ -1238,7 +1237,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                     h(Fragment, ['a ', renderDiskText(), ' to keep your data even after you delete your compute'])
                 ]),
                 li({ style: { marginTop: '1rem' } }, [
-                  h(Link, { 'aria-label': 'Disk doc', onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
+                  h(Link, { onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
                 ])
               ])
             ]),
@@ -1266,7 +1265,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           !sparkMode && !isPersistentDisk && div({ style: { ...styles.whiteBoxContainer, marginTop: '1rem' } }, [
             div([
               'Time to upgrade your cloud environment. Terra’s new persistent disk feature will safeguard your work and data. ',
-              h(Link, { 'aria-label': 'Disk doc', onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
+              h(Link, { onClick: handleLearnMoreAboutPersistentDisk }, ['Learn more about Persistent disks and where your disk is mounted'])
             ]),
             h(ButtonOutline, {
               style: { marginTop: '1rem' },
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -687,7 +687,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
             ]),
             div([
               'Custom environments ', b(['must ']), 'be based off one of the ',
-              h(Link, { 'aria-label': 'Terra base images', href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images'])
+              h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images'])
             ])
           ])
         }],
--- a/src/components/ContextBar.js
+++ b/src/components/ContextBar.js
@@ -102,7 +102,7 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
           }, [icon('ellipsis-v', { size: 24 })])
         ]),
         h(Clickable, {
-          'aria-label': `Cloud env config`,
+          'aria-label': 'Cloud env config',
           style: { ...contextBarStyles.contextBarButton, flexDirection: 'column', justifyContent: 'center', padding: '.75rem' },
           hover: contextBarStyles.hover,
           tooltipSide: 'left',
@@ -115,7 +115,7 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
           getEnvironmentStatusIcons()
         ]),
         h(Clickable, {
-          'aria-label': `Terminal button`,
+          'aria-label': 'Terminal button',
           style: { ...contextBarStyles.contextBarButton, color: !isTerminalEnabled ? colors.dark(0.7) : contextBarStyles.contextBarButton.color },
           hover: contextBarStyles.hover,
           tooltipSide: 'left',
--- a/src/components/ContextBar.js
+++ b/src/components/ContextBar.js
@@ -93,7 +93,7 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
       div({ style: contextBarStyles.contextBarContainer }, [
         h(WorkspaceMenuTrigger, { canShare, isOwner, setCloningWorkspace, setSharingWorkspace, setDeletingWorkspace }, [
           h(Clickable, {
-            'aria-label': 'Menu',
+            'aria-label': 'Workspace menu',
             style: contextBarStyles.contextBarButton,
             hover: contextBarStyles.hover,
             tooltipSide: 'left',
@@ -102,6 +102,7 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
           }, [icon('ellipsis-v', { size: 24 })])
         ]),
         h(Clickable, {
+          'aria-label': `Cloud env config`,
           style: { ...contextBarStyles.contextBarButton, flexDirection: 'column', justifyContent: 'center', padding: '.75rem' },
           hover: contextBarStyles.hover,
           tooltipSide: 'left',
@@ -114,6 +115,7 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
           getEnvironmentStatusIcons()
         ]),
         h(Clickable, {
+          'aria-label': `Terminal button`,
           style: { ...contextBarStyles.contextBarButton, color: !isTerminalEnabled ? colors.dark(0.7) : contextBarStyles.contextBarButton.color },
           hover: contextBarStyles.hover,
           tooltipSide: 'left',
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -59,6 +59,7 @@ const styles = {
 
 const RuntimeIcon = ({ shape, onClick, disabled, style, ...props }) => {
   return h(Clickable, {
+    'aria-label': `Runtime icon`,
     style: { color: onClick && !disabled ? colors.accent() : colors.dark(0.7), ...styles.verticalCenter, ...style },
     onClick, disabled, ...props
   }, [icon(shape, { size: 20 })])
@@ -103,6 +104,7 @@ const RuntimeErrorNotification = ({ runtime }) => {
 
   return h(Fragment, [
     h(Clickable, {
+      'aria-label': `Runtime error notification`,
       onClick: () => setModalOpen(true),
       style: {
         marginTop: '1rem',
@@ -147,6 +149,7 @@ const AppErrorNotification = ({ app }) => {
 
   return h(Fragment, [
     h(Clickable, {
+      'aria-label': `App error notification`,
       onClick: () => setModalOpen(true),
       style: {
         marginTop: '1rem',
@@ -215,7 +218,7 @@ export default class RuntimeManager extends PureComponent {
       notify('warn', 'Please Update Your Cloud Environment', {
         message: h(Fragment, [
           p(['Last year, we introduced important updates to Terra that are not compatible with the older cloud environment associated with this workspace. You are no longer able to save new changes to notebooks using this older cloud environment.']),
-          h(Link, { href: dataSyncingDocUrl, ...Utils.newTabLinkProps }, ['Read here for more details.'])
+          h(Link, { 'aria-label': `Welder cutoff link`, href: dataSyncingDocUrl, ...Utils.newTabLinkProps }, ['Read here for more details.'])
         ])
       })
     } else if (isAfter(createdDate, twoMonthsAgo) && !isToday(dateNotified)) {
@@ -348,6 +351,7 @@ export default class RuntimeManager extends PureComponent {
     return h(Fragment, [
       app && div({ style: { ...styles.container, borderRadius: 5, marginRight: '1.5rem' } }, [
         h(Clickable, {
+          'aria-label': `Galaxy clickable`,
           style: { display: 'flex' },
           disabled: appIsSettingUp(app),
           tooltip: appIsSettingUp(app) ?
@@ -366,11 +370,13 @@ export default class RuntimeManager extends PureComponent {
       ]),
       div({ style: styles.container }, [
         (activeRuntimes.length > 1 || activeDisks.length > 1) && h(Link, {
+          'aria-label': `Multiple cloud env warning`,
           style: { marginRight: '1rem' },
           href: Nav.getLink('environments'),
           tooltip: 'Multiple cloud environments found in this billing project. Click to select which to delete.'
         }, [icon('warning-standard', { size: 24, style: { color: colors.danger() } })]),
         h(Link, {
+          'aria-label': `Launch app`,
           href: applicationLaunchLink,
           onClick: window.location.hash === applicationLaunchLink && currentStatus === 'Stopped' ? () => this.startRuntime() : undefined,
           tooltip: canCompute ? `Open ${applicationName}` : noCompute,
@@ -382,6 +388,7 @@ export default class RuntimeManager extends PureComponent {
         h(IdContainer, [id => h(Fragment, [
           h(Clickable, {
             id,
+            'aria-label': `Create runtime clickable`,
             style: styles.button(isDisabled),
             tooltip: Utils.cond(
               [!canCompute, () => noCompute],
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -59,7 +59,7 @@ const styles = {
 
 const RuntimeIcon = ({ shape, onClick, disabled, style, ...props }) => {
   return h(Clickable, {
-    'aria-label': `Runtime icon`,
+    'aria-label': 'Runtime icon',
     style: { color: onClick && !disabled ? colors.accent() : colors.dark(0.7), ...styles.verticalCenter, ...style },
     onClick, disabled, ...props
   }, [icon(shape, { size: 20 })])
@@ -104,7 +104,7 @@ const RuntimeErrorNotification = ({ runtime }) => {
 
   return h(Fragment, [
     h(Clickable, {
-      'aria-label': `Runtime error notification`,
+      'aria-label': 'Runtime error notification',
       onClick: () => setModalOpen(true),
       style: {
         marginTop: '1rem',
@@ -149,7 +149,7 @@ const AppErrorNotification = ({ app }) => {
 
   return h(Fragment, [
     h(Clickable, {
-      'aria-label': `App error notification`,
+      'aria-label': 'App error notification',
       onClick: () => setModalOpen(true),
       style: {
         marginTop: '1rem',
@@ -351,7 +351,7 @@ export default class RuntimeManager extends PureComponent {
     return h(Fragment, [
       app && div({ style: { ...styles.container, borderRadius: 5, marginRight: '1.5rem' } }, [
         h(Clickable, {
-          'aria-label': `Galaxy clickable`,
+          'aria-label': 'Galaxy clickable',
           style: { display: 'flex' },
           disabled: appIsSettingUp(app),
           tooltip: appIsSettingUp(app) ?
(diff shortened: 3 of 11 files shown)


kyuksel:
When I created a new RStudio runtime, a pop-up with a `LAUNCH` button came up. When I clicked on that button, I got the error below.

image

The same error occurred when I clicked on `Launch RStudio` on the modal below:

image

The error may be due to the `502` I see on the browser console:

image

After dismissing the error pop-ups, I can see the RStudio session. Note that it has an error in red on the bottom:

image

kyuksel:
When I click on an RStudio analysis file where the runtime is paused, the `LAUNCH` button is disabled. It'd be helpful to have a tooltip there saying something like "You must have a running cloud environment.", which would also be more consistent with other buttons we have in similar situations.

image

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -11,9 +11,7 @@ import { ButtonOutline, Clickable, HeaderRenderer, Link, PageBox, spinnerOverlay
 import Dropzone from 'src/components/Dropzone'
 import { icon } from 'src/components/icons'
 import { DelayedSearchInput } from 'src/components/input'
-import {
-  AnalysisDeleter, AnalysisDuplicator, findPotentialNotebookLockers, getDisplayName, getFileName, getTool, notebookLockHash, stripExtension, tools
-} from 'src/components/notebook-utils'
```
kyuksel: The IntelliJ Reformatter splits this into three lines so I'd suggest not changing it. Please see the README snippet below to set it up.

image

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -63,16 +62,20 @@ const AnalysisCardHeaders = ({ sort, onSort }) => {
 }
 
 const AnalysisCard = ({
-  namespace, name, lastModified, metadata, application, wsName, onRename, onCopy, onDelete, onExport, canWrite, currentUserHash,
-  potentialLockers
+  currentRuntime, namespace, name, lastModified, metadata, application, wsName, onRename, onCopy, onDelete, onExport, canWrite, currentUserHash,
+  potentialLockers, currentTool
```
kyuksel: 1. `currentTool` doesn't seem to be used.
2. IntelliJ Reformatter splits lines a bit differently.

```suggestion
  currentRuntime, namespace, name, lastModified, metadata, application, wsName, onRename, onCopy, onDelete, onExport, canWrite,
  currentUserHash, potentialLockers
```

kyuksel:
When I launch RStudio using the modal below

image

I see the following on the console

image

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -85,17 +88,26 @@ const AnalysisCard = ({
...
-      h(MenuButton, {
-        href: analysisPlaygroundLink,
-        tooltip: canWrite && 'Open in playground mode',
-        tooltipSide: 'left'
-      }, [makeMenuIcon('chalkboard'), 'Playground']),
+      ...(toolLabel === tools.Jupyter.label ? [
+        h(MenuButton, {
+          href: analysisEditLink,
+          disabled: locked || !canWrite || currentRuntimeTool === tools.RStudio.label,
+          tooltip: !canWrite && noWrite,
```
kyuksel: When hovering over `Edit` below, should there be some text with tooltip stating why `Edit` is disabled?

image

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -85,17 +88,26 @@ const AnalysisCard = ({
...
+        h(MenuButton, {
+          href: analysisPlaygroundLink,
+          tooltip: canWrite && 'Open in playground mode',
+          tooltipSide: 'left'
+        }, [makeMenuIcon('chalkboard'), 'Playground'])
+      ] : [
+        h(MenuButton, {
+          href: rstudioLaunchLink,
+          disabled: !canWrite || currentRuntimeTool === tools.Jupyter.label,
+          tooltip: !canWrite && noWrite,
```
kyuksel: Similar tooltip comment as above

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -21,14 +21,39 @@ const ApplicationLauncher = _.flow(
     breadcrumbs: props => breadcrumbs.commonPaths.workspaceDashboard(props),
     title: _.get('application')
   })
-)(({ namespace, name, refreshRuntimes, runtimes, persistentDisks, application, workspace }, ref) => {
+)(({ namespace, name, refreshRuntimes, runtimes, persistentDisks, application, workspace, workspace: { workspace: { googleProject, bucketName } } }, ref) => {
```
kyuksel: `namespace` is not used

```suggestion
)(({ name, refreshRuntimes, runtimes, persistentDisks, application, workspace, workspace: { workspace: { googleProject, bucketName } } }, ref) => {
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -21,14 +21,39 @@ const ApplicationLauncher = _.flow(
...
   const cookieReady = Utils.useStore(cookieReadyStore)
   const [showCreate, setShowCreate] = useState(false)
   const [busy, setBusy] = useState(false)
+  const [isWelderInit, setIsWelderInit] = useState(application !== 'RStudio') //we've already init welder if app isn't rstudio
```
kyuksel: Since we also have Galaxy apps, would it be more precise to say the following?

```suggestion
  const [isWelderInit, setIsWelderInit] = useState(application !== 'RStudio') // We've already init Welder if app is Jupyter.
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -9,7 +9,15 @@ import { ButtonPrimary, ButtonSecondary, Clickable, LabeledCheckbox, Link, spinn
...
-import { AnalysisDuplicator, findPotentialNotebookLockers, getDisplayName, getTool, notebookLockHash } from 'src/components/notebook-utils'
+import {
+  AnalysisDuplicator,
+  findPotentialNotebookLockers,
+  getDisplayName,
+  getTool,
+  getToolFromRuntime,
+  notebookLockHash,
+  tools
+} from 'src/components/notebook-utils'
```
kyuksel:
```suggestion
import {
  AnalysisDuplicator, findPotentialNotebookLockers, getDisplayName, getTool, getToolFromRuntime, notebookLockHash, tools
} from 'src/components/notebook-utils'
```
for consistency

kyuksel (commented):
I'd like to spend a bit more time on a couple of the files but I'll submit the inline comments I have so far. 
I've also reported my findings from testing as separate comments.

src/components/notebook-utils.js
```diff
@@ -77,6 +77,7 @@ export const tools = {
 const toolToExtensionMap = { [tools.RStudio.label]: tools.RStudio.ext, [tools.Jupyter.label]: tools.Jupyter.ext }
 
 export const getTool = fileName => _.invert(toolToExtensionMap)[getExtension(fileName)]
+export const getToolFromRuntime = runtime => runtime?.labels?.tool
```
zarsky-broad:
```suggestion
export const getToolFromRuntime = _.get(['labels', 'tool'])
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -441,52 +464,57 @@ const AnalysisEditorFrame = ({
...
+
   const cloudStorageDirectory = `gs://${bucketName}/notebooks`
 
-  //TODO: see inline TODOs, may require call-site change too
-  const setUpNotebook = _.flow(
+  const pattern = Utils.switchCase(toolLabel,
+    [tools.RStudio.label, () => `.*\\.Rmd`],
+    [tools.Jupyter.label, () => `.*\\.ipynb`])
+
+  const setUpAnalysis = _.flow(
```
zarsky-broad: If this is only called in the `useOnMount`, this definition should be inside of it too.

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -441,52 +464,57 @@ const AnalysisEditorFrame = ({
...
+    [tools.Jupyter.label, () => `${name}/safe`],
+    [tools.RStudio.label, () => ''])
+
   const cloudStorageDirectory = `gs://${bucketName}/notebooks`
 
-  //TODO: see inline TODOs, may require call-site change too
-  const setUpNotebook = _.flow(
+  const pattern = Utils.switchCase(toolLabel,
+    [tools.RStudio.label, () => `.*\\.Rmd`],
+    [tools.Jupyter.label, () => `.*\\.ipynb`])
```
zarsky-broad: We try to keep any close braces on the same level of indentation as their matching open braces, so in this case they should be on new lines.
```suggestion
    [tools.Jupyter.label, () => `.*\\.ipynb`]
  )
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -441,52 +464,57 @@ const AnalysisEditorFrame = ({
...
-  const localBaseDirectory = `${name}/edit`
-  const localSafeModeBaseDirectory = `${name}/safe`
-  //TODO: should this vary on toolLabel? (As of right now, the answer is no because the VM path will be `/home/rstudio`, but the bucket path for all analyses will be `/notebooks`, but design still WIP)
+  const localBaseDirectory = Utils.switchCase(toolLabel,
+    [tools.Jupyter.label, () => `${name}/edit`],
+    [toolLabel === tools.RStudio.label, () => ''])
+
+  const localSafeModeBaseDirectory = Utils.switchCase(toolLabel,
+    [tools.Jupyter.label, () => `${name}/safe`],
+    [tools.RStudio.label, () => ''])
```
zarsky-broad:
```suggestion
    [tools.RStudio.label, () => '']
  )
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -441,52 +464,57 @@ const AnalysisEditorFrame = ({
...
-  const [notebookSetupComplete, setNotebookSetupComplete] = useState(false)
+  const [analysisSetupComplete, setAnalysisSetupComplete] = useState(false)
   const cookieReady = Utils.useStore(cookieReadyStore)
 
-  const localBaseDirectory = `${name}/edit`
-  const localSafeModeBaseDirectory = `${name}/safe`
-  //TODO: should this vary on toolLabel? (As of right now, the answer is no because the VM path will be `/home/rstudio`, but the bucket path for all analyses will be `/notebooks`, but design still WIP)
+  const localBaseDirectory = Utils.switchCase(toolLabel,
+    [tools.Jupyter.label, () => `${name}/edit`],
+    [toolLabel === tools.RStudio.label, () => ''])
```
zarsky-broad:
```suggestion
    [toolLabel === tools.RStudio.label, () => '']
  )
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -235,25 +244,39 @@ const PreviewHeader = ({
...
+                  chooseMode('edit')
+                }
+              }
+            }, [makeMenuIcon('edit'), 'Edit'])
+          ),
+          h(HeaderButton, {
+            onClick: () => getLocalPref('hidePlaygroundMessage') ? chooseMode('playground') : setPlaygroundModalOpen(true)
+          }, [makeMenuIcon('chalkboard'), 'Playground mode'])
+        ] :
+          [h(HeaderButton, {
```
zarsky-broad: If you move the open bracket on 270 up to the same line as the colon, the two branches of the ternary will have matching indentation, which should make this easier to parse.

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -235,25 +244,39 @@ const PreviewHeader = ({
...
+            onClick: () => {
+              if (currentRuntimeTool !== tools.RStudio.label) {
+                setCreateOpen(true)
+              } else {
+                Nav.goToPath('workspace-application-launch', { namespace, name, application: 'RStudio' })
+              }
+            }, disabled: runtimeStatus !== 'Running'
```
zarsky-broad: Could you move the `disabled` prop to its own line, and add a tooltip explaining to the user why the button is disabled when it is?

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -21,14 +21,39 @@ const ApplicationLauncher = _.flow(
...
+  useEffect(() => {
+    const runtime = getCurrentRuntime(runtimes)
+    const runtimeStatus = getConvertedRuntimeStatus(runtime)
+
+    const setupWelder = _.flow(
+      Utils.withBusyState(setBusy),
+      withErrorReporting('Error setting up analysis')
+    )(async () => {
+      const localBaseDirectory = ``
+      const localSafeModeBaseDirectory = ``
```
gpcarr, line 49: `localSafeModeBaseDirectory` is now optional in this API request once this [welder PR](https://github.com/DataBiosphere/welder/pull/68) is merged and integrated with Leo. We don't need `localSafeModeBaseDirectory` for Rmd file syncing so I would remove that from the Ajax call

src/pages/workspaces/workspace/Analyses.js
jdcanas: Edit is disabled for analyses if you don't have the relevant runtime to edit it. You can still go through the preview -> create runtime flow if you want.

jdcanas:
> When I created a new RStudio runtime, a pop-up with a `LAUNCH` button came up. When I clicked on that button, I got the error below.
> 
> image
> 
> The same error occurred when I clicked on `Launch RStudio` on the modal below:
> 
> image
> 
> The error may be due to the `502` I see on the browser console:
> 
> image
> 
> After dismissing the error pop-ups, I can see the RStudio session. Note that it has an error in red on the bottom:
> 
> image

Yeah, sorry localize won't work without the updated welder. Forgot to include setup, see PR desc again.

src/pages/workspaces/workspace/Analyses.js
jdcanas: misread, toolltip added

src/pages/workspaces/workspace/applications/AppLauncher.js
jdcanas: I'd prefer to include explicitly since we have tickets to possibly use it down the line, to simplify the API from the UI's perspective (same payload for all requests, since Jupyter still needs it)

jdcanas:
> When I launch RStudio using the modal below
> 
> image
> 
> I see the following on the console
> 
> image

Hmm.. seems this was introduced a few PRs ago, and this https://github.com/DataBiosphere/terra-ui/pull/2624/files#diff-44105412b7d7ce49a5ee9ac6b7202ea01d275dbd1054339f21abebdf163349e5R23 fixes it, but I'm a little confused as to what changed along the way as that tab never had to pass that prop (code is 2 years old) and the issue doesn't seem to exist in dev

src/components/notebook-utils.js
zarsky-broad: Were you going to go with this?

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -11,9 +11,7 @@ import { ButtonOutline, Clickable, HeaderRenderer, Link, PageBox, spinnerOverlay
 import Dropzone from 'src/components/Dropzone'
 import { icon } from 'src/components/icons'
 import { DelayedSearchInput } from 'src/components/input'
-import {
-  AnalysisDeleter, AnalysisDuplicator, findPotentialNotebookLockers, getDisplayName, getFileName, getTool, notebookLockHash, stripExtension, tools
-} from 'src/components/notebook-utils'
+import { AnalysisDeleter, AnalysisDuplicator, findPotentialNotebookLockers, getDisplayName, getFileName, getTool, getToolFromRuntime, notebookLockHash, stripExtension, tools } from 'src/components/notebook-utils'
```
kyuksel: As per the reformatter:

```suggestion
import {
  AnalysisDeleter, AnalysisDuplicator, findPotentialNotebookLockers, getDisplayName, getFileName, getTool, getToolFromRuntime, notebookLockHash,
  stripExtension, tools
} from 'src/components/notebook-utils'
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -85,17 +88,28 @@ const AnalysisCard = ({
...
-        href: analysisPlaygroundLink,
-        tooltip: canWrite && 'Open in playground mode',
-        tooltipSide: 'left'
-      }, [makeMenuIcon('chalkboard'), 'Playground']),
+      ...(toolLabel === tools.Jupyter.label ? [
+        h(MenuButton, {
+          href: analysisEditLink,
+          disabled: locked || !canWrite || currentRuntimeTool === tools.RStudio.label,
+          tooltip: Utils.cond([!canWrite, () => noWrite],
+            [currentRuntimeTool === tools.RStudio.label, () => 'Must have a runtime with Jupyter to edit']),
```
kyuksel: To be consistent with other tooltips:

```suggestion
            [currentRuntimeTool === tools.RStudio.label, () => 'You must have a runtime with Jupyter to edit.']),
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -85,17 +88,28 @@ const AnalysisCard = ({
...
+          href: analysisPlaygroundLink,
+          tooltip: canWrite && 'Open in playground mode',
+          tooltipSide: 'left'
+        }, [makeMenuIcon('chalkboard'), 'Playground'])
+      ] : [
+        h(MenuButton, {
+          href: rstudioLaunchLink,
+          disabled: !canWrite || currentRuntimeTool === tools.Jupyter.label,
+          tooltip: Utils.cond([!canWrite, () => noWrite],
+            [currentRuntimeTool === tools.RStudio.label, () => 'Must have a runtime with RStudio to launch']),
```
kyuksel:
```suggestion
            [currentRuntimeTool === tools.RStudio.label, () => 'You must have a runtime with RStudio to launch.']),
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -235,25 +239,42 @@ const PreviewHeader = ({
...
+          h(HeaderButton, {
+            onClick: () => {
+              if (currentRuntimeTool !== tools.RStudio.label) {
+                setCreateOpen(true)
+              } else {
+                Nav.goToPath('workspace-application-launch', { namespace, name, application: 'RStudio' })
+              }
+            },
+            disabled: runtimeStatus !== 'Running',
+            tooltip: runtimeStatus !== 'Running' ? 'You must have a running cloud environment' : ''
```
kyuksel: It'd be good to end user-facing sentences with a full stop.

```suggestion
            tooltip: runtimeStatus !== 'Running' ? 'You must have a running cloud environment.' : ''
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -382,6 +402,7 @@ const AnalysisPreviewFrame = ({ analysisName, toolLabel, workspace: { workspace:
 }
 
 //TODO: this ensures that navigating away from the Jupyter iframe results in a save via a custom extension located in `jupyter-iframe-extension`
+//Do we need functionality for rstudio?
```
kyuksel:
```suggestion
// Do we need a similar functionality for rstudio?
```

kyuksel: I vaguely remember some convo on this on Slack. If so, could you update the TODO here if applicable?  Ideally, we should have a ticket to track this question if still applicable, and remove the TODO either way?

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
kyuksel:
```suggestion
    [tools.RStudio.label, () => '']
  )
```

kyuksel:
When clicking on an RStudio analysis as shown below

image

I see an accessibility warning appear on the console as shown below

image

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -19,16 +19,42 @@ const ApplicationLauncher = _.flow(
...
-    title: _.get('application')
+    title: _.get('application'),
+    activeTab: appLauncherTabName
   })
-)(({ namespace, name, refreshRuntimes, runtimes, persistentDisks, application, workspace }, ref) => {
+)(({ name, refreshRuntimes, runtimes, persistentDisks, application, workspace, workspace: { workspace: { googleProject, bucketName } } }, ref) => {
   const cookieReady = Utils.useStore(cookieReadyStore)
   const [showCreate, setShowCreate] = useState(false)
   const [busy, setBusy] = useState(false)
+  const [isWelderInit, setIsWelderInit] = useState(application !== 'RStudio') // We've already init Welder if app is Jupyter.
```
kyuksel: Perhaps we shouldn't assume that `not 'RStudio' == 'Jupyter'`?

```suggestion
  const [isWelderInit, setIsWelderInit] = useState(application === 'Jupyter') // We've already init Welder if app is Jupyter.
```
Note: I couldn't be sure if `'Jupyter'` is the exact string for when the application is Jupyter. Please use the correct one :)

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -19,16 +19,42 @@ const ApplicationLauncher = _.flow(
...
   const runtime = getCurrentRuntime(runtimes)
   const runtimeStatus = getConvertedRuntimeStatus(runtime) // preserve null vs undefined
 
+  useEffect(() => {
+    const runtime = getCurrentRuntime(runtimes)
+    const runtimeStatus = getConvertedRuntimeStatus(runtime)
+
+    const setupWelder = _.flow(
+      Utils.withBusyState(setBusy),
+      withErrorReporting('Error setting up analysis')
```
kyuksel: In this case, the error message would specifically correspond to file syncing. Therefore, it may be better to reflect that in the message with something like

```suggestion
      withErrorReporting('Error setting up analysis file syncing')
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -19,16 +19,42 @@ const ApplicationLauncher = _.flow(
   Utils.forwardRefWithName('ApplicationLauncher'),
   wrapWorkspace({
     breadcrumbs: props => breadcrumbs.commonPaths.workspaceDashboard(props),
-    title: _.get('application')
+    title: _.get('application'),
+    activeTab: appLauncherTabName
   })
-)(({ namespace, name, refreshRuntimes, runtimes, persistentDisks, application, workspace }, ref) => {
+)(({ name, refreshRuntimes, runtimes, persistentDisks, application, workspace, workspace: { workspace: { googleProject, bucketName } } }, ref) => {
```
kyuksel: `name` is very generic. Can we be more specific with `workspaceName` so a reader wouldn't need context to figure out what name it is?

Same for other components like [AnalysisLauncher](https://github.com/DataBiosphere/terra-ui/pull/2624/files#diff-e19c52f5a9d3432ff7537aebcbb3d2a439a5fefc7346c1b0ac0017e2f9e8652bR65).

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -19,16 +19,42 @@ const ApplicationLauncher = _.flow(
...
+        .Runtimes
+        .fileSyncing(googleProject, runtime.runtimeName)
+        .setStorageLinks(localBaseDirectory, localSafeModeBaseDirectory, cloudStorageDirectory, `.*\\.Rmd`)
+    })
+
+    if (!isWelderInit && runtimeStatus === 'Running') {
+      setupWelder()
+      setIsWelderInit(true)
+    }
+  }, [googleProject, name, isWelderInit, runtimes, bucketName])
```
kyuksel: The `useEffect` hook would be run one more time unnecessarily when it changes `isWelderInit`, a state that is in its dependency array so we should take it out?

```suggestion
  }, [googleProject, name, runtimes, bucketName])
```

`isWelderInit` is only changed by that `useEffect` so it shouldn't miss any updates to that state due to the removal.

kyuksel (commented):
Testing looked good 😊  with one accessibility warning noted as a separate message. Some more comments are inline.

src/pages/workspaces/workspace/applications/AppLauncher.js
jdcanas: Ah that's a remnant of a previous refactor, thanks

edit: Oh I may have been mistaken with the above comment. React complains if it isn't listed as a dependency since its used in the `useEffect`. I could suppress it, but seems easier to just include since it makes no semantic difference.

jdcanas: image

zarsky-broad: Yup, it should be suppressed in this instance to accurately capture the intent of this line. Add this:
```
// eslint-disable-line react-hooks/exhaustive-deps
```

src/pages/workspaces/workspace/applications/AppLauncher.js
jdcanas: The boolean expression here ensures we only attempt to init welder in the app launcher if the application is RStudio (isWelderInit only resolves to false `if application === RStudio`). Semantically this makes more sense to me since welder is already initialized at this point in the flow if application !== RStudio.

The way this is set up, if another application is added, there's no path via which welder will be invoked for file syncing unless explicitly added to one of the two aforementioned paths (which is desirable in my opinion since we need to think about the .ext pattern)

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
jdcanas: We have a ticket to track it. I like the idea of having a set of TODOs to go through when cleaning up as a back-up checklist wen releasing

src/pages/workspaces/workspace/applications/AppLauncher.js
kyuksel: I see. Then how about we rename `isWelderInit` as `shouldWelderBeSetUp` since that's really why this state is and it feels clarity is worth the verbosity in this case?

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
kyuksel: > I like the idea of having a set of TODOs to go through when cleaning up as a back-up checklist wen releasing
I see. That sounds good.

I'd also add the ticket # to the comment for reference.

src/components/ComputeModal.js
```diff
@@ -771,6 +771,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                 enableGpusSpan
             ]),
             h(Link, {
+              'aria-label': `GPU docc`,
```
kyuksel: Typo in `docc`

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -235,31 +245,49 @@ const PreviewHeader = ({
       [readOnlyAccess, () => h(HeaderButton, { onClick: () => setExportingAnalysis(true) },
         [makeMenuIcon('export'), 'Copy to another workspace']
       )],
```
kyuksel: To be consistent with the other places in the code where the children array left bracket is on the same line as the component name and props:

```suggestion
      [readOnlyAccess, () => h(HeaderButton, { onClick: () => setExportingAnalysis(true) }, [
        makeMenuIcon('export'), 'Copy to another workspace'
      ])],
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -235,31 +245,49 @@ const PreviewHeader = ({
...
-      //TODO: THIS CONDITIONAL should look for Jupyter or Rstudio
-      [!!runtimeStatus && runtime.labels.tool !== 'Jupyter', () => h(StatusMessage, { hideSpinner: true }, [
-        'Your cloud compute doesn\'t appear to be running Jupyter. Create a new cloud environment with Jupyter on it to edit this notebook.'
-      ])],
       [!mode || [null, 'Stopped'].includes(runtimeStatus), () => h(Fragment, [
```
kyuksel, line 247: GitHub didn't let me make a suggestion selecting a large number of lines so I'll copy my styling suggestion below, with motivations being
- consistency with the rest of the codebase (e.g. left bracket for component children starting on the same line as the previous args to `h()`)
- conserving vertical space when the above isn't violated

```javascript
  return h(ApplicationHeader, {
    label: 'PREVIEW (READ-ONLY)',
    labelBgColor: colors.dark(0.2)
  }, [
    Utils.cond(
      [readOnlyAccess, () => h(HeaderButton, { onClick: () => setExportingAnalysis(true) }, [
        makeMenuIcon('export'), 'Copy to another workspace'
      ])],
      [!mode || [null, 'Stopped'].includes(runtimeStatus), () => h(Fragment, [
        ...(toolLabel === tools.Jupyter.label ? [
          Utils.cond(
            [runtime && !welderEnabled, () => h(HeaderButton, { onClick: () => setEditModeDisabledOpen(true) }, [
              makeMenuIcon('warning-standard'), 'Edit (Disabled)'
            ])],
            [locked, () => h(HeaderButton, { onClick: () => setFileInUseOpen(true) }, [
              makeMenuIcon('lock'), 'Edit (In use)'
            ])],
            () => h(HeaderButton, {
              onClick: () => { currentRuntimeTool !== tools.Jupyter.label ? setCreateOpen(true) : chooseMode('edit') }
            }, [
              makeMenuIcon('edit'), 'Edit'
            ])
          ),
          h(HeaderButton, {
            onClick: () => getLocalPref('hidePlaygroundMessage') ? chooseMode('playground') : setPlaygroundModalOpen(true)
          }, [
            makeMenuIcon('chalkboard'), 'Playground mode'
          ])
        ] : [
          h(HeaderButton, {
            onClick: () => {
              currentRuntimeTool !== tools.RStudio.label ?
                setCreateOpen(true) :
                Nav.goToPath('workspace-application-launch', { namespace, name, application: 'RStudio' })
            },
            disabled: runtimeStatus !== 'Running',
            tooltip: runtimeStatus !== 'Running' ? 'You must have a running cloud environment' : ''
          }, [
            makeMenuIcon('rocket'), 'Launch'
          ])
        ]),
        h(MenuTrigger, {
          closeOnClick: true,
          content: h(Fragment, [
            h(MenuButton, { 'aria-label': `Copy analysis`, onClick: () => setCopyingAnalysis(true) }, ['Make a Copy']),
            h(MenuButton, { 'aria-label': `Export analysis`, onClick: () => setExportingAnalysis(true) }, ['Copy to another workspace']),
            h(MenuButton, {
              'aria-label': `Copy to clipboard`,
              onClick: withErrorReporting('Error copying to clipboard', async () => {
                await clipboard.writeText(`${window.location.host}/${analysisLink}`)
                notify('success', 'Successfully copied URL to clipboard', { timeout: 3000 })
              })
            }, ['Copy URL to clipboard'])
          ]),
          side: 'bottom'
        }, [
          h(HeaderButton, {}, [icon('ellipsis-v')])
        ])
      ])],
      [_.includes(runtimeStatus, usableStatuses), () => {
        console.assert(false, `Expected cloud environment to NOT be one of: [${usableStatuses}]`)
        return null
      }],
      [runtimeStatus === 'Creating', () => h(StatusMessage, [
        'Creating cloud environment. You can navigate away and return in 3-5 minutes.'
      ])],
      [runtimeStatus === 'Starting', () => h(StatusMessage, [
        'Starting cloud environment, this may take up to 2 minutes.'
      ])],
      [runtimeStatus === 'Stopping', () => h(StatusMessage, [
        'Cloud environment is stopping, which takes ~4 minutes. You can restart it after it finishes.'
      ])],
      [runtimeStatus === 'LeoReconfiguring', () => h(StatusMessage, [
        'Cloud environment is updating, please wait.'
      ])],
      [runtimeStatus === 'Error', () => h(StatusMessage, { hideSpinner: true }, ['Cloud environment error.'])]
    ),
    div({ style: { position: 'relative' } }, [
      h(Clickable, {
        'aria-label': 'Exit preview mode',
        style: { opacity: 0.65, marginRight: '1.5rem' },
        hover: { opacity: 1 }, focus: 'hover',
        onClick: () => Nav.goToPath('workspace-analyses', { namespace, name })
      }, [icon('times-circle', { size: 30 })])
    ]),
    editModeDisabledOpen && h(EditModeDisabledModal, {
      onDismiss: () => setEditModeDisabledOpen(false),
      onRecreateRuntime: () => {
        setEditModeDisabledOpen(false)
        onCreateRuntime()
      },
      onPlayground: () => {
        setEditModeDisabledOpen(false)
        chooseMode('playground')
      }
    }),
    fileInUseOpen && h(FileInUseModal, {
      namespace, name, lockedBy, canShare, bucketName,
      onDismiss: () => setFileInUseOpen(false),
      onCopy: () => {
        setFileInUseOpen(false)
        setCopyingAnalysis(true)
      },
      onPlayground: () => {
        setFileInUseOpen(false)
        chooseMode('playground')
      }
    }),
    copyingAnalysis && h(AnalysisDuplicator, {
      printName: getDisplayName(analysisName),
      toolLabel: getTool(analysisName),
      fromLauncher: true,
      wsName: name, googleProject, namespace, bucketName, destroyOld: false,
      onDismiss: () => setCopyingAnalysis(false),
      onSuccess: () => setCopyingAnalysis(false)
    }),
    exportingAnalysis && h(ExportAnalysisModal, {
      printName: getDisplayName(analysisName),
      toolLabel: getTool(analysisName), workspace,
      fromLauncher: true,
      onDismiss: () => setExportingAnalysis(false)
    }),
    playgroundModalOpen && h(PlaygroundModal, {
      onDismiss: () => setPlaygroundModalOpen(false),
      onPlayground: () => {
        setPlaygroundModalOpen(false)
        chooseMode('playground')
      }
    })
  ])
```

jdcanas: I implemented most of it, save for one suggested ternary. I like using if/else over a ternary when there are heavy side effects (in this case one branch navigates away from the page entirely)

src/pages/workspaces/workspace/applications/AppLauncher.js
jdcanas: hmm.. that sounds good to me

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -22,11 +23,12 @@ const ApplicationLauncher = _.flow(
...
     activeTab: appLauncherTabName
   })
-)(({ name, refreshRuntimes, runtimes, persistentDisks, application, workspace, workspace: { workspace: { googleProject, bucketName } } }, ref) => {
+)(({ name: workspaceName, refreshRuntimes, runtimes, persistentDisks, application, workspace, workspace: { workspace: { googleProject, bucketName } } }, ref) => {
   const cookieReady = Utils.useStore(cookieReadyStore)
   const [showCreate, setShowCreate] = useState(false)
   const [busy, setBusy] = useState(false)
-  const [isWelderInit, setIsWelderInit] = useState(application !== 'RStudio') // We've already init Welder if app is Jupyter.
+  // We've already init Welder if app is Jupyter.
+  const [shouldSetupWelder, setShouldWelderSetup] = useState(application !== tools.RStudio.label)
```
kyuksel:
```suggestion
  const [shouldSetupWelder, setShouldSetupWelder] = useState(application !== tools.RStudio.label)
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -49,11 +51,12 @@ const ApplicationLauncher = _.flow(
         .setStorageLinks(localBaseDirectory, localSafeModeBaseDirectory, cloudStorageDirectory, `.*\\.Rmd`)
     })
 
-    if (!isWelderInit && runtimeStatus === 'Running') {
+    if (!shouldSetupWelder && runtimeStatus === 'Running') {
       setupWelder()
-      setIsWelderInit(true)
+      setShouldWelderSetup(true)
```
kyuksel:
```suggestion
      setShouldSetupWelder(true)
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
kyuksel: That makes sense. 

Sorry for the painful manual diff on my comment. I should've made a PR into your branch to make it easier to realize what I had suggested changing.

kyuksel:
@jdcanas Thank you for addressing my comments. Just three more nits: [1](https://github.com/DataBiosphere/terra-ui/pull/2624#discussion_r716817014), [2](https://github.com/DataBiosphere/terra-ui/pull/2624#discussion_r720374250), [3](https://github.com/DataBiosphere/terra-ui/pull/2624#discussion_r720374962)

kyuksel (approved):
👍🏼  May want to rebase with `dev` and see if the integration tests will pass on that time.

zarsky-broad (changes requested):
A couple things:
1. Links or buttons that already contain descriptive text shouldn't have additional aria labels. Only those whose only children are icons need them. Adding unnecessary labels adds a potential source of inconsistency.
2. We avoid backtick-delimited template strings when they don't contain any interpolation or newlines. There are quite a few of these in this PR, could you convert them to normal strings?
3. Adding aria labels to components that aren't mapped directly to a DOM node, like `PreviewHeader`, is a no-op, as the prop isn't used, so we shouldn't do that.

rtitle:
I did some testing:
1. Pointed local UI to dev Leo
2. Created RStudio runtime
3. Created Rmd file in RStudio
4. The file appeared in my workspace!
image

I then made some changes to the Rmd file in RStudio and verified it was reflected in the workspace (and I could preview it).

I did notice 1 oddity with `delete` and `rename` operations. If I delete an Rmd file in the workspace, it disappears… but if I wait a few seconds it reappears again. I think this is because the file is still on my persistent disk, and the running welder process re-syncs all Rmd files. Should `delete` in the UI also destroy the storage link / file on the PD?

Also I think we should test this scenario:
1. User A creates a runtime, creates a Rmd file, stops their runtime.
2. User B creates a runtime, makes some changes to the Rmd file, stops their runtime.
3. User A starts their runtime again, and overwrites User B's edits with a stale version of the Rmd file.

I'm not sure how welder behaves here, but it's important as it could lead to lost work. I can try and test this out..

jdcanas:
> A couple things:
> 
> 1. Links or buttons that already contain descriptive text shouldn't have additional aria labels. Only those whose only children are icons need them. Adding unnecessary labels adds a potential source of inconsistency.
> 2. We avoid backtick-delimited template strings when they don't contain any interpolation or newlines. There are quite a few of these in this PR, could you convert them to normal strings?
> 3. Adding aria labels to components that aren't mapped directly to a DOM node, like `PreviewHeader`, is a no-op, as the prop isn't used, so we shouldn't do that.

1. Not sure what you are calling attention to here. There are no Links/Buttons with a `description` field. Do you mean ones with tooltip? Ones with text as a child? Some examples would be helpful. Also, is it really a big deal to have an extra descriptive field on components? A big enough deal to hold up this already-gated-20-days-in-review-PR? 
2. Done. Seems like this should be something enforced by a linter if we care enough to block PRs for it, and if an industry-standard linter cannot enforce, it I'm of the opinion it is a bit too pedantic to block PRs for. 
3. This component was in the (unclear) stacktrace around accessibility labels. Removed.

zarsky-broad:
By "contain descriptive text", I mean that an element has useful descriptive text within its children.

By inspecting the page using [ANDI](https://www.ssa.gov/accessibility/andi/help/install.html), we can see that, in cases where there's both an aria label and text content, the aria label is read instead of the text content. Thus, for people with screen readers, they won't be able to access the text in a link at all. Unless the aria label is significantly more descriptive than the inner text (in which case we should probably change the inner text), we would prefer that they hear what we wrote.

The intention behind enforcing single quotes is consistency, especially with an eye for new developers in the codebase. If they see backticks, that should be a signal for them that the string includes interpolation. We have a number of things that the linter doesn't enforce, but we find useful for readability. For example, we don't enforce line lengths through a linter rule because we've encountered occasional instances where it's more readable to exceed a given length, but most of the time we try to break at ~120.

rtitle (approved):
I did a quick re-test.. the UI looks good and I verified that RStudio syncing does NOT happen where it did before when I last tested this. 👍

kyuksel (approved):

jdcanas:
@zarsky-broad I addressed all your changes could you remove the merge block?

cc @sjacobs146

dmohs:
@jdcanas I've been monitoring this PR and it makes me sad to see that there is still such a gap in communication or understanding. In my judgment, @zarsky-broad made a strong case for removing the `aria-label` attribute from links. You have not presented a counter argument, yet your PR still contains (at a glance) five or so instances of this. If we merge your PR as-is, someone else will need to take on the additional work of removing them. I ask that you either counter Isaac's argument so that we can understand your position or make the requested changes.

sjacobs146:
This particular breakdown in communication is my fault. This is the
conflicting feedback that I mentioned to you.

jdcanas:
His argument is that the aria-label field should be removed for links in which the label is not more descriptive than the text. All instance where the label is not more descriptive than the text have been removed. Those where it is more descriptive remain. 

If you disagree with specific instances please call them out and I will change them. @zarsky-broad

src/components/ComputeModal.js
```diff
@@ -687,7 +687,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
             ]),
             div([
               'Custom environments ', b(['must ']), 'be based off one of the ',
-              h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images'])
+              h(Link, { 'aria-label': 'Terra base images', href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images'])
```
zarsky-broad: This label is the same as the content, so we should drop it.
```suggestion
              h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra Jupyter Notebook base images'])
```

src/components/ComputeModal.js
```diff
@@ -934,7 +934,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         p([
           'You are about to create a virtual machine using an unverified Docker image. ',
           'Please make sure that it was created by you or someone you trust, using one of our ',
-          h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['base images.']),
+          h(Link, { 'aria-label': 'Terra base images', href: terraBaseImages, ...Utils.newTabLinkProps }, ['base images.']),
```
zarsky-broad: It seems like we should just normalize this text:
```suggestion
          h(Link, { href: terraBaseImages, ...Utils.newTabLinkProps }, ['Terra base images.']),
```

src/components/ComputeModal.js
```diff
@@ -951,7 +951,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     const makeJSON = value => div({ style: { whiteSpace: 'pre-wrap', fontFamily: 'Menlo, monospace' } }, [JSON.stringify(value, null, 2)])
     return showDebugger ?
       div({ style: { position: 'fixed', top: 0, left: 0, bottom: 0, right: '50vw', backgroundColor: 'white', padding: '1rem', overflowY: 'auto' } }, [
-        h(Link, { onClick: () => setShowDebugger(false), style: { position: 'absolute', top: 0, right: 0 } }, ['x']),
+        h(Link, { 'aria-label': 'Show debugger', onClick: () => setShowDebugger(false), style: { position: 'absolute', top: 0, right: 0 } }, ['x']),
```
zarsky-broad:
```suggestion
        h(Link, { 'aria-label': 'Hide debugger', onClick: () => setShowDebugger(false), style: { position: 'absolute', top: 0, right: 0 } }, ['x']),
```

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -333,16 +334,19 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
 export const WorkspaceMenuTrigger = ({ children, canShare, isOwner, setCloningWorkspace, setSharingWorkspace, setDeletingWorkspace }) => h(
   MenuTrigger, {
     closeOnClick: true,
+    'aria-label': 'Workspace menu',
     content: h(Fragment, [
-      h(MenuButton, { onClick: () => setCloningWorkspace(true) }, [makeMenuIcon('copy'), 'Clone']),
+      h(MenuButton, { 'aria-label': 'Workspace copy', onClick: () => setCloningWorkspace(true) }, [makeMenuIcon('copy'), 'Clone']),
```
zarsky-broad: I'm not sure that the word 'workspace' is necessary here, since this is inside something that screenreaders will now (thank you!) read as 'workspace menu':
```suggestion
      h(MenuButton, { onClick: () => setCloningWorkspace(true) }, [makeMenuIcon('copy'), 'Clone']),
```

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -333,16 +334,19 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
...
       h(MenuButton, {
+        'aria-label': 'Workspace share',
```
zarsky-broad: Same as above, this can just be read as 'share' since it's in the workspace menu
```suggestion
```

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -333,16 +334,19 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
...
         disabled: !canShare,
         tooltip: !canShare && 'You have not been granted permission to share this workspace',
         tooltipSide: 'left',
         onClick: () => setSharingWorkspace(true)
       }, [makeMenuIcon('share'), 'Share']),
-      h(MenuButton, { disabled: true }, [makeMenuIcon('export'), 'Publish', comingSoon]),
+      h(MenuButton, { 'aria-label': 'Workspace publish', disabled: true }, [makeMenuIcon('export'), 'Publish', comingSoon]),
```
zarsky-broad, line 346: Same:
```suggestion
      h(MenuButton, { disabled: true }, [makeMenuIcon('export'), 'Publish', comingSoon]),
```

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -333,16 +334,19 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
...
       h(MenuButton, {
+        'aria-label': 'Workspace delete',
```
zarsky-broad, line 348:
```suggestion
```

src/components/ContextBar.js
```diff
@@ -93,7 +93,7 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
       div({ style: contextBarStyles.contextBarContainer }, [
         h(WorkspaceMenuTrigger, { canShare, isOwner, setCloningWorkspace, setSharingWorkspace, setDeletingWorkspace }, [
           h(Clickable, {
-            'aria-label': 'Menu',
+            'aria-label': 'Workspace menu',
```
zarsky-broad, line 96: This `aria-label` is actually entirely unused; if you read the definition of `Clickable`, you'll see that when it has a tooltip but no text inside of it, the tooltip is attached via `aria-labelledby`. If you want to change the label to workspace menu, you should change the tooltip.

That said, double checking: this will replace the tab bar, right? Otherwise this needs a different name, since the tab bar we have is already labelled 'workspace menu'.

src/components/ContextBar.js
```diff
@@ -102,6 +102,7 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
           }, [icon('ellipsis-v', { size: 24 })])
         ]),
         h(Clickable, {
+          'aria-label': 'Cloud env config',
```
zarsky-broad, line 105: Already has a tooltip (see above)
```suggestion
```

src/components/ContextBar.js
```diff
@@ -114,6 +115,7 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
           getEnvironmentStatusIcons()
         ]),
         h(Clickable, {
+          'aria-label': 'Terminal button',
```
zarsky-broad, line 118: Already has a tooltip (see above)
```suggestion
```

src/components/RuntimeManager.js
```diff
@@ -59,6 +59,7 @@ const styles = {
 
 const RuntimeIcon = ({ shape, onClick, disabled, style, ...props }) => {
   return h(Clickable, {
+    'aria-label': 'Runtime icon',
```
zarsky-broad, line 62: This is given a tooltip when it's used, so this won't be read (see above)
```suggestion
```

src/components/RuntimeManager.js
```diff
@@ -103,6 +104,7 @@ const RuntimeErrorNotification = ({ runtime }) => {
 
   return h(Fragment, [
     h(Clickable, {
+      'aria-label': 'Runtime error notification',
```
zarsky-broad, line 107: The clickable isn't the notifications itself, it's a button to see the log info, so I'd say the contained text is actually more helpful for someone using a screen reader, since this doesn't tell them what clicking it will do.
```suggestion
```

src/components/RuntimeManager.js
```diff
@@ -147,6 +149,7 @@ const AppErrorNotification = ({ app }) => {
 
   return h(Fragment, [
     h(Clickable, {
+      'aria-label': 'App error notification',
```
zarsky-broad, line 152: See above
```suggestion
```

src/components/RuntimeManager.js
```diff
@@ -215,7 +218,7 @@ export default class RuntimeManager extends PureComponent {
       notify('warn', 'Please Update Your Cloud Environment', {
         message: h(Fragment, [
           p(['Last year, we introduced important updates to Terra that are not compatible with the older cloud environment associated with this workspace. You are no longer able to save new changes to notebooks using this older cloud environment.']),
-          h(Link, { href: dataSyncingDocUrl, ...Utils.newTabLinkProps }, ['Read here for more details.'])
+          h(Link, { 'aria-label': `Welder cutoff link`, href: dataSyncingDocUrl, ...Utils.newTabLinkProps }, ['Read here for more details.'])
```
zarsky-broad, line 221: Since we don't explain what Welder is, I think the "read here for more details" text is probably more helpful.
```suggestion
          h(Link, { href: dataSyncingDocUrl, ...Utils.newTabLinkProps }, ['Read here for more details.'])
```

src/components/RuntimeManager.js
```diff
@@ -348,6 +351,7 @@ export default class RuntimeManager extends PureComponent {
     return h(Fragment, [
       app && div({ style: { ...styles.container, borderRadius: 5, marginRight: '1.5rem' } }, [
         h(Clickable, {
+          'aria-label': 'Galaxy clickable',
```
zarsky-broad, line 354: This has a tooltip already
```suggestion
```

src/components/RuntimeManager.js
```diff
@@ -366,11 +370,13 @@ export default class RuntimeManager extends PureComponent {
       ]),
       div({ style: styles.container }, [
         (activeRuntimes.length > 1 || activeDisks.length > 1) && h(Link, {
+          'aria-label': 'Multiple cloud env warning',
```
zarsky-broad, line 373: This has a tooltip already
```suggestion
```

src/components/RuntimeManager.js
```diff
@@ -366,11 +370,13 @@ export default class RuntimeManager extends PureComponent {
...
           style: { marginRight: '1rem' },
           href: Nav.getLink('environments'),
           tooltip: 'Multiple cloud environments found in this billing project. Click to select which to delete.'
         }, [icon('warning-standard', { size: 24, style: { color: colors.danger() } })]),
         h(Link, {
+          'aria-label': 'Launch app',
```
zarsky-broad, line 379: This has a tooltip already
```suggestion
```

src/components/RuntimeManager.js
```diff
@@ -382,6 +388,7 @@ export default class RuntimeManager extends PureComponent {
         h(IdContainer, [id => h(Fragment, [
           h(Clickable, {
             id,
+            'aria-label': 'Create runtime clickable',
```
zarsky-broad, line 391: This has a tooltip already
```suggestion
```

src/components/runtime-common.js
```diff
@@ -115,6 +115,7 @@ export const SaveFilesHelp = (isGalaxyDisk = false) => {
     p([
       'If you want to save some files permanently, such as input data, analysis outputs, or installed packages, ',
       h(Link, {
+        'aria-label': 'Save file help',
```
zarsky-broad, line 118: Since this will be read as a sentence, 'move them to the workspace bucket' makes more sense to me than this label. If anything, we might insert the text 'learn how to' before the link, or something similar.
```suggestion
```

src/components/runtime-common.js
```diff
@@ -128,6 +129,7 @@ export const SaveFilesHelpRStudio = () => {
     p([
       'If you want to save files permanently, including input data, analysis outputs, installed packages or code in your session, ',
       h(Link, {
+        'aria-label': 'RStudio save help',
```
zarsky-broad, line 132: Since this will be read as a sentence, 'move them to the workspace bucket' makes more sense to me than this label. If anything, we might insert the text 'learn how to' before the link, or something similar.
```suggestion
```

src/components/runtime-common.js
```diff
@@ -145,6 +147,7 @@ export const SaveFilesHelpGalaxy = () => {
     ]),
     p(['If you want to save some files permanently, such as input data, analysis outputs, or installed packages, ',
       h(Link, {
+        'aria-label': 'Galaxy save help',
```
zarsky-broad, line 150: Since this will be read as a sentence, 'move them to the workspace bucket' makes more sense to me than this label. If anything, we might insert the text 'learn how to' before the link, or something similar.
```suggestion
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -63,49 +65,72 @@ const AnalysisCardHeaders = ({ sort, onSort }) => {
...
   const analysisEditLink = `${analysisLink}/?${qs.stringify({ mode: 'edit' })}`
   const analysisPlaygroundLink = `${analysisLink}/?${qs.stringify({ mode: 'playground' })}`
 
   const analysisMenu = h(MenuTrigger, {
+    'aria-label': `Analysis menu`,
     side: 'right',
     closeOnClick: true,
     content: h(Fragment, [
       h(MenuButton, {
+        'aria-label': `Preview`,
```
zarsky-broad, line 91: This has a tooltip already
```suggestion
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -63,49 +65,72 @@ const AnalysisCardHeaders = ({ sort, onSort }) => {
...
         href: analysisLink,
         tooltip: canWrite && 'Open without cloud compute',
         tooltipSide: 'left'
       }, [makeMenuIcon('eye'), 'Open preview']),
+      ...(toolLabel === tools.Jupyter.label ? [
+        h(MenuButton, {
+          'aria-label': `Edit`,
```
zarsky-broad, line 98: This has text already
```suggestion
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -63,49 +65,72 @@ const AnalysisCardHeaders = ({ sort, onSort }) => {
...
+        }, locked ? [makeMenuIcon('lock'), 'Edit (In Use)'] : [makeMenuIcon('edit'), 'Edit']),
+        h(MenuButton, {
+          'aria-label': `Playground`,
+          href: analysisPlaygroundLink,
+          tooltip: canWrite && 'Open in playground mode',
+          tooltipSide: 'left'
+        }, [makeMenuIcon('chalkboard'), 'Playground'])
+      ] : [
+        h(MenuButton, {
+          'aria-label': `Launch`,
```
zarsky-broad, line 113: This has text already
```suggestion
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -63,49 +65,72 @@ const AnalysisCardHeaders = ({ sort, onSort }) => {
...
-        tooltip: !canWrite && noWrite,
-        tooltipSide: 'left'
-      }, locked ? [makeMenuIcon('lock'), 'Edit (In Use)'] : [makeMenuIcon('edit'), 'Edit']),
-      h(MenuButton, {
-        href: analysisPlaygroundLink,
-        tooltip: canWrite && 'Open in playground mode',
-        tooltipSide: 'left'
-      }, [makeMenuIcon('chalkboard'), 'Playground']),
-      h(MenuButton, {
+        'aria-label': `Copy`,
```
zarsky-broad, line 122: This has text already
```suggestion
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -63,49 +65,72 @@ const AnalysisCardHeaders = ({ sort, onSort }) => {
...
         disabled: !canWrite,
         tooltip: !canWrite && noWrite,
         tooltipSide: 'left',
         onClick: () => onCopy()
       }, [makeMenuIcon('copy'), 'Make a copy']),
       h(MenuButton, {
+        'aria-label': `Export`,
```
zarsky-broad, line 129: This has text already
```suggestion
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -63,49 +65,72 @@ const AnalysisCardHeaders = ({ sort, onSort }) => {
...
         onClick: () => onExport()
       }, [makeMenuIcon('export'), 'Copy to another workspace']),
       h(MenuButton, {
+        'aria-label': `Copy`,
```
zarsky-broad, line 133: This has text already
```suggestion
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -116,12 +141,14 @@ const AnalysisCard = ({
         }
       }, [makeMenuIcon('copy-to-clipboard'), 'Copy analysis URL to clipboard']),
       h(MenuButton, {
+        'aria-label': `Rename`,
```
zarsky-broad, line 144: This has text already
```suggestion
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -116,12 +141,14 @@ const AnalysisCard = ({
...
         disabled: !canWrite,
         tooltip: !canWrite && noWrite,
         tooltipSide: 'left',
         onClick: () => onRename()
       }, [makeMenuIcon('renameIcon'), 'Rename']),
       h(MenuButton, {
+        'aria-label': `Delete`,
```
zarsky-broad, line 151: This has text already
```suggestion
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -169,6 +196,7 @@ const AnalysisCard = ({
     div({ style: { ...endColumnFlex, flexDirection: 'row' } }, [
       div({ style: { flex: 1, display: 'flex' } }, [
         locked && h(Clickable, {
+          'aria-label': `${artifactName} artifact label`,
```
zarsky-broad, line 199: This has a tooltip already, so this would never be read
```suggestion
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -19,16 +19,49 @@ const ApplicationLauncher = _.flow(
...
 
+  useEffect(() => {
+    const runtime = getCurrentRuntime(runtimes)
+    const runtimeStatus = getConvertedRuntimeStatus(runtime)
+
+    const setupWelder = _.flow(
+      Utils.withBusyState(setBusy),
+      withErrorReporting('Error setting up analysis file syncing')
+    )(async () => {
+      const localBaseDirectory = ``
```
zarsky-broad, line 48:
```suggestion
      const localBaseDirectory = ''
```

src/pages/workspaces/workspace/applications/AppLauncher.js
zarsky-broad:
```suggestion
      const localSafeModeBaseDirectory = ''
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -19,16 +19,49 @@ const ApplicationLauncher = _.flow(
...
+      const localSafeModeBaseDirectory = ``
+      const cloudStorageDirectory = `gs://${bucketName}/notebooks`
+
+      await Ajax()
+        .Runtimes
+        .fileSyncing(googleProject, runtime.runtimeName)
+        .setStorageLinks(localBaseDirectory, localSafeModeBaseDirectory, cloudStorageDirectory, `.*\\.Rmd`)
```
zarsky-broad, line 55:
```suggestion
        .setStorageLinks(localBaseDirectory, localSafeModeBaseDirectory, cloudStorageDirectory, '.*\\.Rmd')
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -145,19 +148,23 @@ const EditModeDisabledModal = ({ onDismiss, onRecreateRuntime, onPlayground }) =
     p('We’ve released important updates that are not compatible with the older cloud environment associated with this workspace. To enable Edit Mode, please delete your existing cloud environment and create a new cloud environment.'),
     p('If you have any files on your old cloud environment that you want to keep, you can access your old cloud environment using the Playground Mode option.'),
     h(Link, {
+      'aria-label': 'Data syncing doc',
       href: dataSyncingDocUrl,
       ...Utils.newTabLinkProps
     }, ['Read here for more details.']),
     div({ style: { marginTop: '2rem' } }, [
       h(ButtonSecondary, {
+        'aria-label': 'Launcher dismiss',
```
zarsky-broad, line 157: This has text already
```suggestion
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -145,19 +148,23 @@ const EditModeDisabledModal = ({ onDismiss, onRecreateRuntime, onPlayground }) =
...
         style: { padding: '0 1rem' },
         onClick: () => onDismiss()
       }, ['Cancel']),
       h(ButtonSecondary, {
+        'aria-label': 'Launcher playground',
```
zarsky-broad, line 162: This has text already
```suggestion
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -145,19 +148,23 @@ const EditModeDisabledModal = ({ onDismiss, onRecreateRuntime, onPlayground }) =
...
         style: { padding: '0 1rem', marginLeft: '1rem' },
         onClick: () => onPlayground()
       }, ['Run in playground mode']),
       h(ButtonPrimary, {
+        'aria-label': `Launcher create`,
```
zarsky-broad, line 167: This has text already
```suggestion
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -232,34 +241,49 @@ const PreviewHeader = ({
...
+            ])],
+            () => h(HeaderButton, { onClick: () => currentRuntimeTool !== tools.Jupyter.label ? setCreateOpen(true) : chooseMode('edit') }, [
+              makeMenuIcon('edit'), 'Edit'
+            ])
+          ),
+          h(HeaderButton, {
+            onClick: () => getLocalPref('hidePlaygroundMessage') ? chooseMode('playground') : setPlaygroundModalOpen(true)
+          }, [
+            makeMenuIcon('chalkboard'), 'Playground mode'
+          ])
```
zarsky-broad, line 264: I'd combine this into one line, personally
```suggestion
          }, [makeMenuIcon('chalkboard'), 'Playground mode'])
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -232,34 +241,49 @@ const PreviewHeader = ({
...
+                setCreateOpen(true)
+              } else {
+                Nav.goToPath('workspace-application-launch', { namespace, name, application: 'RStudio' })
+              }
+            },
+            disabled: runtimeStatus !== 'Running',
+            tooltip: runtimeStatus !== 'Running' ? 'You must have a running cloud environment.' : ''
+          }, [
+            makeMenuIcon('rocket'), 'Launch'
+          ])
```
zarsky-broad, line 278: I'd combine this into one line, personally:
```suggestion
          }, [makeMenuIcon('rocket'), 'Launch'])
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -232,34 +241,49 @@ const PreviewHeader = ({
...
+        ]),
         h(MenuTrigger, {
           closeOnClick: true,
           content: h(Fragment, [
-            h(MenuButton, { onClick: () => setCopyingAnalysis(true) }, ['Make a Copy']),
-            h(MenuButton, { onClick: () => setExportingAnalysis(true) }, ['Copy to another workspace']),
+            h(MenuButton, { 'aria-label': `Copy analysis`, onClick: () => setCopyingAnalysis(true) }, ['Make a Copy']),
```
zarsky-broad, line 283: This has text already
```suggestion
            h(MenuButton, { onClick: () => setCopyingAnalysis(true) }, ['Make a Copy']),
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -232,34 +241,49 @@ const PreviewHeader = ({
...
+            h(MenuButton, { 'aria-label': `Export analysis`, onClick: () => setExportingAnalysis(true) }, ['Copy to another workspace']),
```
zarsky-broad: This has text already
```suggestion
            h(MenuButton, { onClick: () => setExportingAnalysis(true) }, ['Copy to another workspace']),
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -232,34 +241,49 @@ const PreviewHeader = ({
...
             h(MenuButton, {
+              'aria-label': `Copy to clipboard`,
```
zarsky-broad: This has text already
```suggestion
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -441,52 +465,58 @@ const AnalysisEditorFrame = ({
...
+    [tools.RStudio.label, () => '']
+  )
 
   Utils.useOnMount(() => {
-    setUpNotebook()
+    const cloudStorageDirectory = `gs://${bucketName}/notebooks`
+
+    const pattern = Utils.switchCase(toolLabel,
+      [tools.RStudio.label, () => `.*\\.Rmd`],
+      [tools.Jupyter.label, () => `.*\\.ipynb`]
```
zarsky-broad:
```suggestion
      [tools.RStudio.label, () => '.*\\.Rmd'],
      [tools.Jupyter.label, () => '.*\\.ipynb']
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -520,13 +550,12 @@ const WelderDisabledNotebookEditorFrame = ({
       notify('error', 'Cannot Edit Notebook', {
         message: h(Fragment, [
           p(['Recent updates to Terra are not compatible with the older cloud environment in this workspace. Please recreate your cloud environment in order to access Edit Mode for this notebook.']),
-          h(Link, { href: dataSyncingDocUrl, ...Utils.newTabLinkProps }, ['Read here for more details.'])
+          h(Link, { 'aria-label': 'Data syncing doc', href: dataSyncingDocUrl, ...Utils.newTabLinkProps }, ['Read here for more details.'])
```
zarsky-broad: I think, in context, it makes sense to say 'read more here' rather than the name of a doc, since that's the action we want people to do.
```suggestion
          h(Link, { href: dataSyncingDocUrl, ...Utils.newTabLinkProps }, ['Read here for more details.'])
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -543,14 +572,14 @@ const WelderDisabledNotebookEditorFrame = ({
       b(['NOT ']),
       'being saved to the workspace. To save your changes, download the notebook using the file menu.',
       h(Link, {
+        'aria-label': 'Data syncing doc',
```
zarsky-broad: I think, in context, it makes sense to say 'read more here' rather than the name of a doc, since that's the action we want people to do.
```suggestion
```

zarsky-broad (approved):
Thumbing this with some comments, so that you'll be unblocked once you're able to address them. I've done all of these as change suggestions, so if you're ok with them, you can just accept them. Let me know if you have any questions!

Here's a good explanation to learn more about labeling links: https://usability.yale.edu/web-accessibility/articles/links#aria

PR #2625, opened 2021-09-14
[CA-1514] Disable invalid billing projects

s-rubenstein:
We need billing to set up Google Projects for PPW, so Billing Projects without a billing account can't be used anymore to create workspaces.

Screen Shot 2021-09-14 at 5 01 06 PM

### DIFF
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -172,7 +172,7 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
           options: _.uniq(_.map(({ projectName, invalidBillingAccount }) => {
             return {
               label: h(Fragment, [
-                invalidBillingAccount ? h(TooltipTrigger, { content: ['Billing project must have a valid billing account to be used.'], side: 'left' }, [div({ style: { padding: optionPadding } }, [projectName])]) : div({ style: { padding: optionPadding } }, [projectName])
+                invalidBillingAccount ? h(TooltipTrigger, { content: ['Workspaces may only be created in billing projects that have a Google billing account accessible in Terra'], side: 'left' }, [div({ style: { padding: optionPadding } }, [projectName])]) : div({ style: { padding: optionPadding } }, [projectName])
               ]),
               value: projectName,
               isDisabled: invalidBillingAccount
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -154,13 +154,20 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
       ])]),
       h(IdContainer, [id => h(Fragment, [
         h(FormLabel, { htmlFor: id, required: true }, ['Billing project']),
+        div({ style: { margin: '0.5rem 0' } }, ['A billing project must have a billing account which Terra has access to in order to be used.']),
         h(Select, {
           id,
           isClearable: false,
           placeholder: 'Select a billing project',
           value: namespace,
           onChange: ({ value }) => setNamespace(value),
-          options: _.uniq(_.map('projectName', billingProjects)).sort()
+          options: _.uniq(_.map(({ projectName, invalidBillingAccount }) => {
+            return {
+              label: projectName,
+              value: projectName,
+              isDisabled: invalidBillingAccount
+            }
+          }, billingProjects)).sort()
         })
       ])]),
       h(IdContainer, [id => h(Fragment, [
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -6,6 +6,7 @@ import { icon } from 'src/components/icons'
 import { TextArea, ValidatedInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
 import { InfoBox } from 'src/components/PopupTrigger'
+import TooltipTrigger from 'src/components/TooltipTrigger'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
@@ -116,6 +117,8 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
   const errors = validate({ namespace, name }, constraints, {
     prettify: v => ({ namespace: 'Billing project', name: 'Name' }[v] || validate.prettify(v))
   })
+  // We define option padding here in order to put the padding inside of the tooltip trigger, for more consistent tooltips
+  const optionPadding = 10
 
   return Utils.cond(
     [loading, () => spinnerOverlay],
@@ -154,16 +157,23 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
       ])]),
       h(IdContainer, [id => h(Fragment, [
         h(FormLabel, { htmlFor: id, required: true }, ['Billing project']),
-        div({ style: { margin: '0.5rem 0' } }, ['A billing project must have a billing account which Terra has access to in order to be used.']),
         h(Select, {
           id,
           isClearable: false,
           placeholder: 'Select a billing project',
           value: namespace,
           onChange: ({ value }) => setNamespace(value),
+          styles: {
+            option: provided => ({
+              ...provided,
+              padding: 0
+            })
+          },
           options: _.uniq(_.map(({ projectName, invalidBillingAccount }) => {
             return {
-              label: projectName,
+              label: h(Fragment, [
+                invalidBillingAccount ? h(TooltipTrigger, { content: ['Billing project must have a valid billing account to be used.'], side: 'left' }, [div({ style: { padding: optionPadding } }, [projectName])]) : div({ style: { padding: optionPadding } }, [projectName])
+              ]),
               value: projectName,
               isDisabled: invalidBillingAccount
             }
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -117,8 +117,6 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
   const errors = validate({ namespace, name }, constraints, {
     prettify: v => ({ namespace: 'Billing project', name: 'Name' }[v] || validate.prettify(v))
   })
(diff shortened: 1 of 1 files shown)


src/components/NewWorkspaceModal.js
```diff
@@ -154,13 +154,20 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
       ])]),
       h(IdContainer, [id => h(Fragment, [
         h(FormLabel, { htmlFor: id, required: true }, ['Billing project']),
+        div({ style: { margin: '0.5rem 0' } }, ['A billing project must have a billing account which Terra has access to in order to be used.']),
```
andy7i: I like the idea behind this message. I'm not a huge fan of the UX though, since this text is a bit separated from the disabled billing projects. Could we either:
* show this message on each disabled billing project (maybe with an info icon and a tooltip?) 
* or not display the disabled billing projects in the list.

src/components/NewWorkspaceModal.js
```diff
@@ -154,13 +154,20 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
...
           isClearable: false,
           placeholder: 'Select a billing project',
           value: namespace,
           onChange: ({ value }) => setNamespace(value),
-          options: _.uniq(_.map('projectName', billingProjects)).sort()
+          options: _.uniq(_.map(({ projectName, invalidBillingAccount }) => {
+            return {
+              label: projectName,
+              value: projectName,
+              isDisabled: invalidBillingAccount
```
andy7i: I'm also noticing that these billing projects are disabled:

```
aj-bp-test-1owner
aj-bp-test-1owner-1user
aj-bp-test-2owners
aj-bp-test-transfer-ownership
```
even though they all seem to have a valid billing account `Broad Institute - 8201528`
https://bvdp-saturn-dev.appspot.com/#billing?selectedName=aj-bp-test-1owner&type=project

andy7i: this might not be an issue with this PR though, since the DB is not showing the billing accounts either.

```
> select * from BILLING_PROJECT where NAME = 'aj-bp-test-1owner' \G
*************************** 1. row ***************************
                               NAME: aj-bp-test-1owner
                CROMWELL_BUCKET_URL: gs://cromwell-auth-aj-bp-test-1owner
                    CREATION_STATUS: Ready
                    BILLING_ACCOUNT: null
                            MESSAGE: NULL
                   CROMWELL_BACKEND: PAPIv2
                  SERVICE_PERIMETER: NULL
              GOOGLE_PROJECT_NUMBER: 576518341068
               INVALID_BILLING_ACCT:
               SPEND_REPORT_DATASET: NULL
                 SPEND_REPORT_TABLE: NULL
SPEND_REPORT_DATASET_GOOGLE_PROJECT: NULL
1 row in set (0.064 sec)
```

even though the UI shows a valid billing account
Screen Shot 2021-09-15 at 9 51 34 AM

andy7i: ^ not sure why the UI would show a billing account if the DB says the billing account is `null` though

src/components/NewWorkspaceModal.js
```diff
@@ -154,13 +154,20 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
...
         h(FormLabel, { htmlFor: id, required: true }, ['Billing project']),
+        div({ style: { margin: '0.5rem 0' } }, ['A billing project must have a billing account which Terra has access to in order to be used.']),
         h(Select, {
           id,
           isClearable: false,
           placeholder: 'Select a billing project',
           value: namespace,
           onChange: ({ value }) => setNamespace(value),
-          options: _.uniq(_.map('projectName', billingProjects)).sort()
+          options: _.uniq(_.map(({ projectName, invalidBillingAccount }) => {
```
andy7i: I was able to create a workspace in an errored billing project:
Screen Shot 2021-09-15 at 9 48 34 AM
Screen Shot 2021-09-15 at 9 49 29 AM
Screen Shot 2021-09-15 at 9 50 05 AM



```

*************************** 1. row ***************************
                               NAME: test-eh
                CROMWELL_BUCKET_URL: NULL
                    CREATION_STATUS: Error
                    BILLING_ACCOUNT: billingAccounts/00708C-45D19D-27AAFA
                            MESSAGE: project test-eh creation finished with errors: {"ResourceType":"cloudresourcemanager.v1.project","ResourceErrorCode":"403","ResourceErrorMessage":{"code":403,"message":"The caller does not have permission","status":"PERMISSION_DENIED","statusMessage":"Forbidden","requestPath":"https://cloudresourcemanager.googleapis.com/v1/projects/test-eh","httpMethod":"GET"}} - code RESOURCE_ERROR
                   CROMWELL_BACKEND: PAPIv2
                  SERVICE_PERIMETER: NULL
              GOOGLE_PROJECT_NUMBER: NULL
               INVALID_BILLING_ACCT:
               SPEND_REPORT_DATASET: NULL
                 SPEND_REPORT_TABLE: NULL
SPEND_REPORT_DATASET_GOOGLE_PROJECT: NULL
1 row in set (5.006 sec)
```

this one seems to be a regression, as I don't see this BP listed when I look at dev.

andy7i (changes requested):

src/components/NewWorkspaceModal.js
andy7i: I think that the backend should also not allow this to happen. We should create a new ticket for this.

src/components/NewWorkspaceModal.js
andy7i: I thought this would affect a lot of users since I see it affecting 9500 projects on dev. I took a look on prod, and this only affects 69 projects (most of which are firecloud projects), so I do think this message adds visual clutter for most users.

src/components/NewWorkspaceModal.js
s-rubenstein: This does not block errored billing projects, just ones without a valid billing account. Those are two separate things that I would strongly prefer to keep separate for PRs.

src/components/NewWorkspaceModal.js
s-rubenstein: Dev's UI checks google directly, whereas the create needs it in our DB. v2_workspaces branch on the UI does the right thing.

s-rubenstein:
New screenshot:
Screen Shot 2021-09-16 at 12 04 03 PM

src/components/NewWorkspaceModal.js
andy7i: @s-rubenstein can you link the ticket you created to address this regression?

ehigham:
unsolicited comment but how about "Workspaces may only be created in billing projects that have a Google billing account accessible in Terra"?
Also, did you consider just not showing invalid billing projects in this dropdown?

s-rubenstein:
I like that wording better!

I did consider not showing them, but I think our users may be confused to not see projects that they expect to be there, so wanted to show them. We can re-evaluate later if it is a problem?

ehigham (approved):
Looks good, thanks!

src/components/NewWorkspaceModal.js
```diff
@@ -116,6 +117,8 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
   const errors = validate({ namespace, name }, constraints, {
     prettify: v => ({ namespace: 'Billing project', name: 'Name' }[v] || validate.prettify(v))
   })
+  // We define option padding here in order to put the padding inside of the tooltip trigger, for more consistent tooltips
+  const optionPadding = 10
```
kyuksel: Not sure if I understand the comment above. It'd save reader some scrolling if we inline `padding: 10` down below. It is used only in two places, both of which are on the same line.

src/components/NewWorkspaceModal.js
```diff
@@ -160,7 +163,21 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
           placeholder: 'Select a billing project',
           value: namespace,
           onChange: ({ value }) => setNamespace(value),
-          options: _.uniq(_.map('projectName', billingProjects)).sort()
+          styles: {
+            option: provided => ({
+              ...provided,
+              padding: 0
+            })
+          },
```
kyuksel: We could save some vertical space as single-lining it seems easy to read and short enough? 
```suggestion
          styles: { option: provided => ({ ...provided, padding: 0 }) },
```

src/components/NewWorkspaceModal.js
```diff
@@ -160,7 +163,21 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
...
+          options: _.uniq(_.map(({ projectName, invalidBillingAccount }) => {
+            return {
+              label: h(Fragment, [
+                invalidBillingAccount ? h(TooltipTrigger, { content: ['Workspaces may only be created in billing projects that have a Google billing account accessible in Terra'], side: 'left' }, [div({ style: { padding: optionPadding } }, [projectName])]) : div({ style: { padding: optionPadding } }, [projectName])
+              ]),
+              value: projectName,
+              isDisabled: invalidBillingAccount
+            }
+          }, billingProjects)).sort()
```
kyuksel: Couldn't insert it as a suggestion because I couldn't select the entire block due to the comments above but when I remove the optional `return` piece along with enclosing braces, and run the IntelliJ reformatter, I get the following:

```javascript
          options: _.uniq(_.map(({ projectName, invalidBillingAccount }) => ({
            label: h(Fragment, [
              invalidBillingAccount ?
                h(TooltipTrigger, {
                  content: ['Workspaces may only be created in billing projects that have a Google billing account accessible in Terra'], side: 'left'
                }, [div({ style: { padding: optionPadding } }, [projectName])]) :
                div({ style: { padding: optionPadding } }, [projectName])
            ]),
            value: projectName,
            isDisabled: invalidBillingAccount
          }),
          billingProjects)).sort()
```

It might be a bit easier to read that way, and the splitting the ternary into separate lines would be consistent with the rest of the codebase.

s-rubenstein:
@kyuksel should be good for another review.

src/components/NewWorkspaceModal.js
s-rubenstein: https://broadworkbench.atlassian.net/browse/CA-1524

gpolumbo-broad (approved):

andy7i (approved):
Tested locally 👍

kyuksel (approved):
:shipit:

PR #2626, opened 2021-09-15
Added functionality to back button, added stackedIcons

lu-c:
Added utility method for stacking buttons ( https://fontawesome.com/v5.15/how-to-use/on-the-web/styling/stacking-icons )

Screen Shot 2021-09-15 at 4 18 27 PM

It's pretty basic, since we don't need more than that right now, but a lot of the values can be overridden for more use cases. I can potentially see us using this in the future for sorting (ie: fa-sort-up and fa-sort-down stacked together, where the sort direction has a different opacity applied)

Then added the back button functionality to the stacked icon


petesantos (commented):
The fontAwesome library has the capability to stack icons - take a look at this:
https://github.com/DataBiosphere/terra-ui/blob/d35905540c30f5b4a938bf4c13661df1676764fe/src/libs/icon-dict.js#L132

The `mask` property will achieve this effect:
https://fontawesome.com/v5.15/how-to-use/on-the-web/using-with/react

lu-c:
merged with the latest branch

PR #2627, opened 2021-09-15
[DR-2094] Added request access modal to the dataset details page

lu-c:
There is a button on the right side bar that says 'request access'. This links up the existing request access modal in this view.


src/pages/library/DataBrowserDetails.js
```diff
@@ -181,7 +183,11 @@ const DataBrowserDetails = ({ id }) => {
             icon('angle-left', { size: 35 })
           ]),
           MainContent(snapshot),
-          Sidebar(snapshot)
+          Sidebar(snapshot, setShowRequestAccessModal),
```
petesantos: Once this has been converted to a component, will these be props of the object argument, rather than positional parameters?

lu-c: What do you mean? Sorry

src/pages/library/DataBrowserDetails.js
```diff
@@ -190,7 +194,11 @@ const DataBrowserDetails = ({ id }) => {
             icon('angle-left', { size: 35 })
           ]),
           h(MainContent, { snapshot }),
-          h(Sidebar, { snapshot })
+          h(Sidebar, { snapshot, setShowRequestAccessModal }),
+          showRequestAccessModal && h(RequestDatasetAccessModal, {
+            datasets: [snapshot],
+            onDismiss: () => setShowRequestAccessModal(false)
+          })
```
petesantos: I would recommend moving this modal and the associated state to the `Sidebar` component since this is where the decision to show the modal is being made. This will also simplify the code a bit as the parent component will not need to be concerned about the modal state.

src/pages/library/DataBrowserDetails.js
```diff
@@ -72,13 +73,16 @@ const MainContent = ({ snapshot }) => {
         div({ style: styles.attributesColumn }, [
           h3({ style: styles.headers }, ['Data release policy']),
           div([snapshot.releasePolicy])
-        ]), div({ style: styles.attributesColumn }, [
+        ]),
```
petesantos, line 76: thanks for the styling update!

src/pages/library/DataBrowserDetails.js
```diff
@@ -103,7 +107,7 @@ const MainContent = ({ snapshot }) => {
   ])
 }
 
-const Sidebar = ({ snapshot }) => {
+const Sidebar = ({ snapshot, setShowRequestAccessModal }) => {
```
petesantos: If you move the state into this component you can remove the `setShowRequestAccessModal` prop.

petesantos (approved):

PR #2629, opened 2021-09-16
[CA-1521] [CA-1519] PPW Bash Fixes

ehigham:
[CA-1521] Fix Billing Account Summary Panel Overlap with Cookie Consent Banner
RR: https://broadworkbench.atlassian.net/browse/CA-1521

Use absolute positioning instead of fixed.
No longer interferes with banner

[CA-1519] Fix Expand Workspace Icon Visible Through Billing Project Summary Panel
RR: https://broadworkbench.atlassian.net/browse/CA-1519

Caused by `down-arrow' being a 180 degree transform of 'up-arrow'. The
transform seems to be rendered later and thus shows up over the panel.
Fixed by flipping the icon in inkscape and adding it as a new one.
arrow now hidden

[CA-1521]: https://broadworkbench.atlassian.net/browse/CA-1521
[CA-1519]: https://broadworkbench.atlassian.net/browse/CA-1519


gpolumbo-broad (approved):

marctalbott (approved):

src/pages/billing/Project.js
```diff
@@ -354,7 +354,7 @@ const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) =
   const { displayName = null } = _.find({ accountName: billingProject.billingAccount }, billingAccounts) || { displayName: 'No Access' }
 
   return h(Fragment, [
-    billingAccountsOutOfDate && div({}, [h(BillingAccountSummaryPanel, { counts: _.mapValues(_.size, groups) })]),
+    billingAccountsOutOfDate && h(BillingAccountSummaryPanel, { counts: _.mapValues(_.size, groups) }),
```
kyuksel: Thanks for fixing the styling here.

zarsky-broad: Instead of changing the icon, move this to the end of the Fragment, which will cause the z-index to fix itself.

ehigham: @zarsky-broad Is there not merit to changing the icon? Someone else could encounter this with some of our icons and will have to remember that the additional transforms cause this?

zarsky-broad: In this case, where the fix is to use the standard pattern of putting modals last in the render, I'm actually ok with it causing issues; we should assume that transforms could be used by other components.

ehigham: > the fix is to use the standard pattern 

Thanks - I had ordered them as I'd see them on the page, not according to modallity. Will update.

zarsky-broad (approved):

PR #2630, opened 2021-09-17
[IA-2902] Increase boot disk to 80G for GCE and Dataproc

rtitle:
We are adding some extra Docker caching to the Leo image, so we need to increase the boot disk slightly for GCE and Dataproc.

See https://github.com/DataBiosphere/leonardo/pull/2271. This PR should go out before the Leo PR.

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -90,7 +90,7 @@ const DataprocDiskSelector = ({ value, onChange }) => {
       label({ htmlFor: id, style: styles.label }, ['Disk size (GB)']),
       h(NumberInput, {
         id,
-        min: 60, // less than this size causes failures in cluster creation
+        min: 80, // less than this size causes failures in cluster creation
         max: 64000,
         isClearable: false,
         onlyInteger: true,
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -10,8 +10,8 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
 
-export const defaultDataprocDiskSize = 60 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
-export const defaultGceBootDiskSize = 70 // GCE boot disk size is not customizable by users. We use this for cost estimate calculations only.
+export const defaultDataprocDiskSize = 80 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
+export const defaultGceBootDiskSize = 80 // GCE boot disk size is not customizable by users. We use this for cost estimate calculations only.
 export const defaultGcePersistentDiskSize = 50
 
 export const defaultGceMachineType = 'n1-standard-1'


kyuksel (approved):
:shipit:

src/libs/runtime-utils.js
```diff
@@ -10,8 +10,8 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
 
-export const defaultDataprocDiskSize = 60 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
-export const defaultGceBootDiskSize = 70 // GCE boot disk size is not customizable by users. We use this for cost estimate calculations only.
+export const defaultDataprocDiskSize = 80 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
+export const defaultGceBootDiskSize = 80 // GCE boot disk size is not customizable by users. We use this for cost estimate calculations only.
```
Qi77Qi, line 14: I think UI does actually get these defaults from leo...these values are only used if leo doesn't return the value properly

src/components/ComputeModal.js
```diff
@@ -90,7 +90,7 @@ const DataprocDiskSelector = ({ value, onChange }) => {
       label({ htmlFor: id, style: styles.label }, ['Disk size (GB)']),
       h(NumberInput, {
         id,
-        min: 60, // less than this size causes failures in cluster creation
+        min: 80, // less than this size causes failures in cluster creation
```
Qi77Qi, line 93: is this for boot disk or user disk?

rtitle: This is dataproc so there is only 1 disk.. the custom image is generated with a 80g disk too: https://github.com/DataBiosphere/leonardo/pull/2271/files#diff-6c263717a24af2836fc1795f59bdea6b8616d61bd499edb331d6c9d253478e34R402

src/libs/runtime-utils.js
rtitle: Oh hm these aren't passed in the Leo `createRuntime` request?

In any case this should be okay since I also updated these to 80g in Leo `reference.conf`..

kyuksel: `defaultGceBootDiskSize` isn't passed in since it's internally set by Leo. It is used for cost estimate calculations so it's good that we're updating it to reflect the new value Leo will start using.

`defaultDataprocDiskSize` determines the default disk size displayed on the runtime modal when there is no existing runtime. It also sets the minimum. So, also good that we're updating it here.

It is true that when there _is_ an existing runtime, the UI uses the values coming from Leo.

src/components/ComputeModal.js
Qi77Qi: ah right

src/libs/runtime-utils.js
Qi77Qi: yea...definitely still good to update..

Qi77Qi (approved):

PR #2632, opened 2021-09-17
[POL-85] Make build fail if lint fails

zarsky-broad:
Without this change, builds that failed the linter were still passing CI.


kyuksel (approved):
Thank you!

dmohs (approved):

PR #2638, opened 2021-09-28
[CA-1538] Look Up `displayName` for Workspace `billingAccount`

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1538
We had been showing the unique billing account name - it makes more sense to show the display name to match that of billing project. To do this, I'm storing billing accounts indexed by their name rather than just the list of accounts. This gives us an efficient alogorithm to look up the display name.

While I was in the neighbourhood, I tried to fix up some code that gave me a headache for glory and profit.

image

### DIFF
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -217,14 +217,12 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
   const [billingProjects, setBillingProjects] = useState(StateHistory.get().billingProjects || null)
   const [creatingBillingProject, setCreatingBillingProject] = useState(false)
   const [billingAccounts, setBillingAccounts] = useState(null)
-  const [isOwner, setIsOwner] = useState(false)
   const [isLoadingProjects, setIsLoadingProjects] = useState(false)
   const [isAuthorizing, setIsAuthorizing] = useState(false)
   const [isLoadingAccounts, setIsLoadingAccounts] = useState(false)
 
   const signal = Utils.useCancellation()
   const interval = useRef()
-
 
   // Helpers
   const loadProjects = _.flow(
@@ -245,13 +243,11 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
     Utils.withBusyState(setIsLoadingAccounts)
   )(async () => {
     if (Auth.hasBillingScope()) {
-      setBillingAccounts(await Ajax(signal).Billing.listAccounts())
+      const accounts = await Ajax(signal).Billing.listAccounts()
+      const byAccountName = Object.fromEntries(_.map(acc => [acc.accountName, acc], accounts))
+      setBillingAccounts(byAccountName)
     }
   })
-
-  const checkOwner = () => {
-    !isOwner && _.map(project => _.includes(billingRoles.owner, project.role) ? setIsOwner(true) : null, billingProjects)
-  }
 
   const showCreateProjectModal = async () => {
     if (Auth.hasBillingScope()) {
@@ -262,12 +258,10 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
     }
   }
 
-
   // Lifecycle
   Utils.useOnMount(() => {
     loadProjects()
     loadAccounts()
-    checkOwner()
   })
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/pages/billing/List.js
```diff
@@ -285,12 +279,14 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
...
-
   // Render
   const hasBillingProjects = !_.isEmpty(billingProjects)
   const breadcrumbs = `Billing > Billing Project`
-
   const billingProjectListWidth = 330
+  const [projectsOwned, projectsShared] = _.partition(
+    ({ roles }) => _.includes(billingRoles.owner, roles),
+    billingProjects
+  )
```
dmohs, line 289: Beautiful usage of partition here.

src/pages/billing/List.js
```diff
@@ -245,14 +243,12 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
     Utils.withBusyState(setIsLoadingAccounts)
   )(async () => {
     if (Auth.hasBillingScope()) {
-      setBillingAccounts(await Ajax(signal).Billing.listAccounts())
+      const accounts = await Ajax(signal).Billing.listAccounts()
+      const byAccountName = Object.fromEntries(_.map(acc => [acc.accountName, acc], accounts))
```
dmohs, line 247: `reduce` is my go-to pattern for this, but yours reads nicely so feel free to ignore.
```suggestion
      const accountsByName = _.reduce((acc, x) => {...acc, [x.accountName]: x}, {}, accounts))
```

dmohs (approved):

src/pages/billing/List.js
ehigham: Indeed. I was hoping to not have to hand-roll this myself, though. lodash's `fromEntries` follows your suggestion almost exactly. No doubt `Object.fromEntries` does something similar.
I was looking for some function that would "index" this a list with a key-function 
```javascript
listAccounts().then(index('accountName')).then(setBillingAccounts)
```
Do you know of such a thing?

gpolumbo-broad (approved):

src/pages/billing/Project.js
```diff
@@ -235,7 +235,7 @@ const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) =
           _.map(workspace => {
             const isExpanded = expandedWorkspaceName === workspace.name
             return h(WorkspaceCard, {
-              workspace,
+              workspace: { ...workspace, billingAccount: billingAccounts[workspace.billingAccount]?.displayName },
```
andy7i: nit: maybe update `billingAccount` to `billingAccountName` for accuracy

andy7i (approved):
tested locally. 1 suggestion but not blocking -- LGTM

src/pages/billing/List.js
dmohs: I'm afraid I do not. Perhaps this function belongs in our Utils.

zarsky-broad: I think you might have been looking for [_.fromPairs](https://lodash.com/docs/4.17.15#fromPairs) here; however, the simplest way to do this would be [_.keyBy](https://lodash.com/docs/4.17.15#keyBy):
```js
const accountsByName = _.keyBy('accountName', accounts)
```

src/pages/billing/List.js
zarsky-broad: +1

src/pages/billing/List.js
```diff
@@ -372,7 +366,11 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
           billingAccounts,
           authorizeAndLoadAccounts
         })],
-        [isOwner && !selectedName && hasBillingProjects, () => div({ style: { margin: '1rem auto 0 auto' } }, ['Select a Billing Project'])],
+        [!_.isEmpty(projectsOwned) && !selectedName && hasBillingProjects, () => div(
+          { style: { margin: '1rem auto 0 auto' } }, [
+            'Select a Billing Project'
+          ]
+        )],
```
zarsky-broad, line 373: I'd linebreak this as:
```js
[predicate, () => div({
  style: {}
}, [children])]
```

PR #2639, opened 2021-09-28
[CA-1539] Update `BillingList` when user changes project billing account

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1539
Introduce `updateBillingProject` callback triggered when billing
account is changed. This updates the parent component's state (ie
billing project list) and redraws the `ProjectDetail` component.

Prevent users from seeing a specified billing project when they're not a billing project owner.

Add `reportErrorsAndRethrow` utility to preserve failed
state of Promises. `withErrorReporting` reports the error then returns a successful
promise. This breaks processing sequences like the following where
successive operations should only be performed if the previous succeeds:

```javascript
async () => {
    await doSomethingDangerous()
    await shouldNotBeCalledIfPreviousOperationFails()
}
```

By returning a failed promise in `reportErrorsAndRethrow` the intended
effect is restored.

### DIFF
--- a/src/libs/error.js
+++ b/src/libs/error.js
@@ -12,17 +12,10 @@ export const reportError = async (title, obj) => {
   }
 }
 
-// Transforms an async function so that it catches and reports errors using the provided text
-export const withErrorReporting = _.curry((title, fn) => async (...args) => {
-  try {
-    return await fn(...args)
-  } catch (error) {
-    await reportError(title, error)
-    throw error
-  }
-})
-
-// Transforms an async function so that it catches and ignores errors
+/**
+ * Return a Promise to the result of evaluating the async `fn` with `...args` or undefined if
+ * evaluation fails.
+ */
 export const withErrorIgnoring = fn => async (...args) => {
   try {
     return await fn(...args)
@@ -30,3 +23,24 @@ export const withErrorIgnoring = fn => async (...args) => {
     // ignore error
   }
 }
+
+/**
+ * Return a Promise to the result of evaluating the async `fn` with `...args`. If evaluation fails,
+ * report the error to the user with `title` as a side effect.
+ */
+export const reportErrorAndRethrow = _.curry((title, fn) => async (...args) => {
+  try {
+    return await fn(...args)
+  } catch (error) {
+    reportError(title, error)
+    throw error
+  }
+})
+
+/**
+ * Return a Promise to the result of evaluating the async `fn` with `...args` or undefined if
+ * evaluation fails. If evaluation fails, report the error to the user with `title`.
+ */
+export const withErrorReporting = _.curry((title, fn) => {
+  return withErrorIgnoring(reportErrorAndRethrow(title)(fn))
+})
--- a/src/libs/error.js
+++ b/src/libs/error.js
@@ -17,7 +17,8 @@ export const withErrorReporting = _.curry((title, fn) => async (...args) => {
   try {
     return await fn(...args)
   } catch (error) {
-    reportError(title, error)
+    await reportError(title, error)
+    throw error
   }
 })
 
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -13,7 +13,7 @@ import TopBar from 'src/components/TopBar'
 import { Ajax } from 'src/libs/ajax'
 import * as Auth from 'src/libs/auth'
 import colors from 'src/libs/colors'
-import { withErrorReporting } from 'src/libs/error'
+import { reportErrorAndRethrow } from 'src/libs/error'
 import Events from 'src/libs/events'
 import { formHint, FormLabel } from 'src/libs/forms'
 import * as Nav from 'src/libs/nav'
@@ -113,7 +113,7 @@ const NewBillingProjectModal = ({ onSuccess, onDismiss, billingAccounts, loadAcc
   const [chosenBillingAccount, setChosenBillingAccount] = useState('')
 
   const submit = _.flow(
-    withErrorReporting('Error creating billing project'),
+    reportErrorAndRethrow('Error creating billing project'),
     Utils.withBusyState(setIsBusy)
   )(async () => {
     try {
@@ -226,12 +226,12 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
 
   // Helpers
   const loadProjects = _.flow(
-    withErrorReporting('Error loading billing projects list'),
+    reportErrorAndRethrow('Error loading billing projects list'),
     Utils.withBusyState(setIsLoadingProjects)
   )(async () => setBillingProjects(_.sortBy('projectName', await Ajax(signal).Billing.listProjects())))
 
   const reloadBillingProject = _.flow(
-    withErrorReporting('Error loading billing project'),
+    reportErrorAndRethrow('Error loading billing project'),
     Utils.withBusyState(setIsLoadingProjects)
   )(async ({ projectName }) => {
     // evaluate first to error if project doesn't exist/user can't access
@@ -241,12 +241,12 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
   })
 
   const authorizeAccounts = _.flow(
-    withErrorReporting('Error setting up authorization'),
+    reportErrorAndRethrow('Error setting up authorization'),
     Utils.withBusyState(setIsAuthorizing)
   )(Auth.ensureBillingScope)
 
   const loadAccounts = _.flow(
-    withErrorReporting('Error loading billing accounts'),
+    reportErrorAndRethrow('Error loading billing accounts'),
     Utils.withBusyState(setIsLoadingAccounts)
   )(() => {
     if (Auth.hasBillingScope()) {
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -1,6 +1,6 @@
 import _ from 'lodash/fp'
 import * as qs from 'qs'
-import { Fragment, useCallback, useEffect, useRef, useState } from 'react'
+import { Fragment, useEffect, useRef, useState } from 'react'
 import { div, h, h2, p, span } from 'react-hyperscript-helpers'
 import Collapse from 'src/components/Collapse'
 import { ButtonOutline, ButtonPrimary, Clickable, IdContainer, Link, Select, spinnerOverlay } from 'src/components/common'
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -360,13 +360,20 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
               p(['It may not exist, or you may not have access to it.'])
             ])
           ])],
-        [selectedName && hasBillingProjects, () => h(ProjectDetail, {
-          key: selectedName,
-          billingProject: _.find({ projectName: selectedName }, billingProjects),
-          billingAccounts,
-          authorizeAndLoadAccounts,
-          invalidateProjectAndReload: loadProjects
-        })],
+        [selectedName && hasBillingProjects, () => {
+          const index = _.findIndex({ projectName: selectedName }, billingProjects)
+          return h(ProjectDetail, {
+            key: selectedName,
+            billingProject: billingProjects[index],
+            billingAccounts,
+            authorizeAndLoadAccounts,
+            invalidateProjectAndReload: async () => {
+              const projects = billingProjects.slice()
+              projects[index] = await Ajax(signal).Billing.billingProject(selectedName)
+              setBillingProjects(projects)
+            }
+          })
+        }],
         [!_.isEmpty(projectsOwned) && !selectedName && hasBillingProjects, () => div(
           { style: { margin: '1rem auto 0 auto' } }, [
             'Select a Billing Project'
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -367,7 +367,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
             billingProject: billingProjects[index],
             billingAccounts,
             authorizeAndLoadAccounts,
-            invalidateProjectAndReload: async () => {
+            updateProjectAndReload: async () => {
               const projects = billingProjects.slice()
               projects[index] = await Ajax(signal).Billing.billingProject(selectedName)
               setBillingProjects(projects)
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -362,9 +362,10 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
           ])],
         [selectedName && hasBillingProjects, () => h(ProjectDetail, {
           key: selectedName,
-          project: _.find({ projectName: selectedName }, billingProjects),
+          billingProject: _.find({ projectName: selectedName }, billingProjects),
           billingAccounts,
-          authorizeAndLoadAccounts
+          authorizeAndLoadAccounts,
+          invalidateProjectAndReload: loadProjects
         })],
         [!_.isEmpty(projectsOwned) && !selectedName && hasBillingProjects, () => div(
           { style: { margin: '1rem auto 0 auto' } }, [
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -366,7 +366,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
             billingProject: billingProjects[index],
             billingAccounts,
             authorizeAndLoadAccounts,
-            updateProject: _.flow(
+            updateBillingProject: _.flow(
               withErrorReporting('Error updating billing project'),
               Utils.withBusyState(setIsLoadingProjects)
             )(async () => {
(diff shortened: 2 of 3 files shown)


gpolumbo-broad (approved):

src/pages/billing/List.js
```diff
@@ -362,9 +362,10 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
           ])],
         [selectedName && hasBillingProjects, () => h(ProjectDetail, {
           key: selectedName,
-          project: _.find({ projectName: selectedName }, billingProjects),
+          billingProject: _.find({ projectName: selectedName }, billingProjects),
           billingAccounts,
-          authorizeAndLoadAccounts
+          authorizeAndLoadAccounts,
+          invalidateProjectAndReload: loadProjects
```
marctalbott: slightly confused about the naming here, why/how does this invalidate the project?

marctalbott (approved):
one question on naming,  but otherwise looks good to me

src/pages/billing/List.js
ehigham: I was thinking along the lines of marking an element in a cache as invalidand triggering a reload. Really what I'm trying to convey is that the whole component needs to be reloaded with an updated project. maybe `updateProjectAndReload` is more suitable.

src/pages/billing/Project.js
```diff
@@ -289,26 +287,21 @@ const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) =
   })
 
   // Helpers
-  const updateBillingProject = withErrorReporting('Error updating billing project')(
-    () => Ajax(signal).Billing.billingProject(billingProject.projectName).then(setBillingProject)
+  const setBillingAccount = withErrorReporting('Error updating billing account')(
```
petesantos: Does the user need to wait for this to return, and if so should we communicate that with the busy spinner?

src/pages/billing/Project.js
```diff
@@ -289,26 +287,21 @@ const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) =
...
+    async newAccountName => {
+      Ajax().Metrics.captureEvent(Events.changeBillingAccount, {
+        oldName: billingProject.billingAccount,
+        newName: newAccountName,
+        billingProjectName: billingProject.projectName
+      })
+      await Ajax(signal).Billing.changeBillingAccount({
```
zarsky-broad: TOL: If you `return` here instead of `await`ing, you wouldn't have to make the function `async`; I don't know which I prefer, though, or if I have a preference.

src/pages/billing/List.js
```diff
@@ -360,12 +360,20 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
...
+          const index = _.findIndex({ projectName: selectedName }, billingProjects)
+          return h(ProjectDetail, {
+            key: selectedName,
+            billingProject: billingProjects[index],
+            billingAccounts,
+            authorizeAndLoadAccounts,
+            updateProjectAndReload: async () => {
+              const projects = billingProjects.slice()
+              projects[index] = await Ajax(signal).Billing.billingProject(selectedName)
+              setBillingProjects(projects)
```
zarsky-broad: 1. Should the effect that watches for updated billing projects on 267 have a dependency on `billingProjects`, so that it's not re-evaluated on every render?
2. To make sure we're reading from the backend as the single source of truth (as opposed to modeling what we think it looks like locally), I'd recommend calling `loadProjects` here, instead of just loading the one. That would also let you get rid of the whole find-the-index pattern, and you could just use `_.find`.

src/pages/billing/Project.js
ehigham: I did have a spinner initally and it looked good when I was updating all the proejcts. When I made the change to update only one project, the spinner appeared and went so quickly that it looked like a weird rendering glitch so I removed it. What do you think?

src/pages/billing/Project.js
ehigham: I think I prefer this - I returned the promise and then chained off it in the onClick handler to `setBillingAccount(selectedBilling).then(updateProjectAndReload)`

zarsky-broad: To be clear, there's no functional difference between returning this promise and awaiting it, as creating an async function is just a promise itself.

src/pages/billing/List.js
```diff
@@ -360,12 +360,20 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
...
+            }
+          })
```
petesantos: I had added a suggestion in here yesterday, not sure why github didn't add the comment to the review.
This suggestion removes the need for mutation and tracking the index. But I was also considering Isaac's suggestion to just load all of the projects, this would simplify the code. I am not sure how many billing projects we are expecting users to have.

Also, the billing project name can change on the update which would require a new sort

```suggestion
          return h(ProjectDetail, {
            key: selectedName,
            billingProject: _.find({ projectName: selectedName }, billingProjects),
            billingAccounts,
            authorizeAndLoadAccounts,
            updateProjectAndReload: async () => _.flow(
              _.concat(_.reject({ projectName: selectedName }, billingProjects)),
              _.sortBy('projectName'),
              setBillingProjects
            )(await Ajax(signal).Billing.billingProject(selectedName))
          })
```

src/pages/billing/Project.js
petesantos: There can be some perceived glitchiness with busy spinners, but when it comes to async calls I prefer to keep them in. We never know how someone's internet connection is, or whether the server may take an extra second to respond.

src/pages/billing/List.js
ehigham: Thank you for your suggestion. This code is much more complicated for the sake of avoiding assignment. I really don't think it's worth it - JavaScript is not a pure functional language and we seem bending over backwards to make it appear so.

src/pages/billing/List.js
ehigham: We're changing one billing project so reloading all projects is a bit unnecessary. I understand what you mean about using the backend as a single source of truth but in this case it's overkill.

src/pages/billing/Project.js
ehigham: They look shit though lol

src/pages/billing/List.js
```diff
@@ -241,11 +241,11 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
...
     Utils.withBusyState(setIsLoadingAccounts)
-  )(async () => {
+  )(() => {
     if (Auth.hasBillingScope()) {
-      const accounts = await Ajax(signal).Billing.listAccounts()
-      const byAccountName = Object.fromEntries(_.map(acc => [acc.accountName, acc], accounts))
-      setBillingAccounts(byAccountName)
+      return Ajax(signal).Billing.listAccounts()
+        .then(_.keyBy('accountName'))
+        .then(setBillingAccounts)
```
zarsky-broad, line 255: Functionally the same, but without the `.then` chaining, which we tend to avoid in favor of `async`/`await`:

```js
async () => {
  if (Auth.hasBillingScope()) {
    return _.flow(
      _.keyBy('accountName'),
      setBillingAccounts
    )(await Ajax(signal).Billing.listAccounts())
  }
}
```

I can't decide if I think this looks ickier than what you have here, though.

src/pages/billing/List.js
```diff
@@ -360,18 +359,33 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
...
+            billingAccounts,
+            authorizeAndLoadAccounts,
+            updateBillingProject: _.flow(
+              withErrorReporting('Error updating billing project'),
+              Utils.withBusyState(setIsLoadingProjects)
+            )(async () => {
+              try {
+                const projects = billingProjects.slice()
+                projects[index] = await Ajax(signal).Billing.billingProject(selectedName)
+                setBillingProjects(projects)
```
zarsky-broad: For people seeing this:
@ehigham and I talked about this at length; there isn't a better way to do this that we could come up with.

src/pages/billing/List.js
```diff
@@ -360,18 +359,33 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
...
+              } catch (_) {
+                await loadProjects()
+              }
```
zarsky-broad: 1. Could you rename the argument to `catch` to `e` for consistency?
2. I think you'll have to report the error yourself in this catch, since it'll prevent the error reporting wrapper from seeing it.

src/pages/billing/Project.js
```diff
@@ -318,36 +310,28 @@ const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) =
...
+  )(() => Ajax(signal).Billing.project(billingProject.projectName).listUsers()
+    .then(_.flow(
       _.groupBy('email'),
       _.map(gs => ({ ..._.omit('role', gs[0]), roles: _.map('role', gs) })),
       _.sortBy('email')
-    )(rawProjectUsers)
-    setProjectUsers(projectUsers)
-  })
+    ))
+    .then(setProjectUsers))
```
zarsky-broad: Could you convert this to async/await? Also, I think setProjectUsers could go in the flow? You could pass the awaited ajax result directly to the flow if you want to avoid sticking it somewhere with a name.
Like:
```js
async () => {
  _.flow(
    // group, map, sort, etc
    setProjectUsers
  )(await Ajax...)
}

File: src/pages/billing/Project.js [hunk 10]
```diff
@@ -376,9 +360,9 @@ const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) =
           onDismiss: () => setShowBillingModal(false),
           okButton: h(ButtonPrimary, {
             disabled: !selectedBilling || billingProject.billingAccount === selectedBilling,
-            onClick: async () => {
+            onClick: () => {
               setShowBillingModal(false)
-              await setBillingAccount(selectedBilling)
+              setBillingAccount(selectedBilling).then(updateBillingProject)
```
zarsky-broad: Again, could you change this to an async/await?

src/pages/billing/Project.js
```diff
@@ -475,30 +459,38 @@ const ProjectDetail = ({ project, billingAccounts, authorizeAndLoadAccounts }) =
...
     }),
     !!deletingUser && h(DeleteUserModal, {
       userEmail: deletingUser.email,
       onDismiss: () => setDeletingUser(false),
-      onSubmit: _.flow(
-        withErrorReporting('Error removing member from billing project'),
-        Utils.withBusyState(setUpdating)
-      )(async () => {
+      onSubmit: async () => {
+        await _.flow(
```
zarsky-broad: Might this be one of those places where it's useful to name this function, since it has all these wrappers around it, and then invoke it by name? Thinking purely of ease of readability.

src/pages/billing/List.js
ehigham: I prefer mine - it reads more linearly.

src/pages/billing/List.js
ehigham: I'd be surprised if the linter doesn't complain about unused binds - `_` is a convention for not caring about that particular value.

src/pages/billing/List.js
ehigham: for what it's worth - I'd be surprised if the linter doesn't complain about unused binds - _ is a convention for not caring about that particular value.
I'll add in the error reporting.

src/pages/billing/Project.js
ehigham: it can't - you'll get a linter error about an unused return

ehigham: I'm hesitant to convert to async/await in these cases. I think that 
```javascript
return getMeAFoo()
    .then(makeMeABar)
    .then(makeMeABaz)
```
is less busy and consequently reads better than
```javascript
const foo = await getMeAFoo()
const bar = await makeMeABar(foo)
return await makeMeABaz(bar)
```
I also think that the names given to return values in such cases offer very little in terms of clarity, if only for breaking up the line.

I'm not saying I won't use async/await - it certainly has its place and can make code look much clearer, especially when interleaving control logic and requests. Where i've inlined a `_.flow`, you're right, these should be pulled out (and thank you for making the suggestions). In the simple cases, however, I don't agree that it helps.

Please keep calling me out on it - I think eventually we'll arrive at a happy medium.

src/pages/billing/Project.js
ehigham: I completely agree. Thanks.

src/pages/billing/List.js
ehigham: The billing project name can't change via the UI - in the case we can't access it, I've fallen back to just reloading everything

src/libs/error.js
```diff
@@ -17,7 +17,8 @@ export const withErrorReporting = _.curry((title, fn) => async (...args) => {
   try {
     return await fn(...args)
   } catch (error) {
-    reportError(title, error)
+    await reportError(title, error)
+    throw error
```
petesantos: This will have an effect everywhere we use `withErrorReporting`. Would you mind removing the `throw`.
In an earlier revision, I saw you had some retry logic - was this change related to that?

petesantos (commented):
This is looking good, thanks for working through the comments. The only outstanding commend I have is on the `throw` in `error.js`

src/libs/error.js
ehigham: Yes - this change simplies error handling a lot. Adding the throw is how to make `withErrorReporing` preserve the failed state of promises, thus subsequent operations will not be executed if an error is reported - you can recover using `.catch()` or `.finally()`.

petesantos: To prevent unexpected errors from occurring around the site, how would you feel about not using `withErrorReporting` in `reloadBillingProject` and handling that imperatively with a `try/catch/throw`? 

Or adding a new function called `withPreservedErrorReporting` (or a better name...). This will give us control in introducing the new wrapper where appropriate

src/pages/billing/List.js
```diff
@@ -230,30 +230,38 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
...
+    withErrorReporting('Error loading billing project'),
+    Utils.withBusyState(setIsLoadingProjects)
+  )(async ({ projectName }) => {
+    // evaluate first to error if project doesn't exist/user can't access
+    const project = await Ajax(signal).Billing.billingProject(selectedName)
+    const index = _.findIndex({ projectName }, billingProjects)
+    setBillingProjects(_.set([index], project))
+  })
+
+  const authorizeAccounts = _.flow(
```
petesantos, line 243: `authorizeAndLoadAccounts` is being used by the `ProjectDetail` component as well - is that component expecting the accounts to be loaded when calling this function?

src/pages/billing/Project.js
dmohs: This is a very good point that I'd like to discuss further. In the past, I've had spinners with logic to only appear after a slight delay in case it turns out the operation completed quickly and, once rendered, stay for a minimum duration to avoid jitter. Would it be worth making this a pattern in this codebase?

src/libs/error.js
ehigham: I could use add `withPreservedErrorReporting` sure. The reason i didn't initially do this is because I feel preserving the error state is really what we want to do (and it is for the billing stuff) as it'll force us to think about errors. I can take the risk-averse route for this.

petesantos (approved):
Chatted with Edmund and the reviewers around the error reporting.
Giving this a thumb in anticipation of a new util function for error reporting that we can begin applying to the rest of the codebase.

src/pages/billing/List.js
ehigham: I'm not sure. I assume it's left to which ever child component first needs it to call it.

src/libs/error.js
```diff
@@ -12,20 +12,35 @@ export const reportError = async (title, obj) => {
...
   }
-}
+})
+
+/**
+ * Return a Promise to the result of evaluating the async `fn` with `...args` or undefined if
+ * evaluation fails. If evaluation fails, report the error to the user with `title`.
+ */
+export const withErrorReporting = _.curry((title, fn) => {
```
ehigham, line 44: As a point on style - I don't share the need to not have a new line after `=`, eg:
```javascript
export const withErrorReporting = 
  _.curry((title, fn) => withErrorIgnoring(reportErrorAndRethrow(title)(fn))
```

src/libs/error.js
```diff
@@ -12,20 +12,35 @@ export const reportError = async (title, obj) => {
...
+  return withErrorIgnoring(reportErrorAndRethrow(title)(fn))
+})
```
ehigham, line 46: As a point on style - how do we feel about new lines after `=`, eg:
```javascript
export const withErrorReporting = 
  _.curry((title, fn) => withErrorIgnoring(reportErrorAndRethrow(title)(fn))
```

kyuksel: Personally, I'd just single-line the whole thing given it's only 108 chars long but I don't feel strongly. I would prefer your suggestion over the version with `{ return …. }` as it feels unnecessary to have `return` and the enclosing braces for a single line.

ehigham: "only 108 characters long" *cries into k&r*

ehigham: Point taken for next time!

PR #2640, opened 2021-09-29
[WIP] [QA-1585] increasing timeout for kernel wait

xuscrbw:
<!--
Hello, friend!
Thanks!
--!>

Addressing some flaky terra-ui automation tests. Seems like it takes time for the Kernel to open

Adding some more context:

Originally, we thought that the kernel was taking a longer time to get ready but it looks like the notebook actually enters a bad state in this instance. The notebook is unable to connect.

This seemed to happen less on a better VM config, but after some testing, it seems like this issue is not config dependent. 

I am not sure the current fix fixes what is needed, but I want to keep this PR open to test.


integration-tests/tests/run-notebook.js
```diff
@@ -46,7 +46,7 @@ const testRunNotebookFn = _.flow(
   await findElement(page, clickable({ textContains: 'Running' }), { timeout: 10 * 60 * 1000 })
 
   const frame = await findIframe(page)
-  await findElement(frame, '//*[@title="Kernel Idle"]')
+  await findElement(frame, '//*[@title="Kernel Idle"]', { timeout: 6 * 60 * 1000 })
```
petesantos: How could we check this without a timeout?

petesantos (commented):
Is this still a WIP?

integration-tests/tests/run-notebook.js
xuscrbw: Not entirely sure in this instance. I think this findElement was our way to check if the Kernel is ready to be used/ when we can execute a command on the notebook. There's an indicator which changes when the Kernel is ready which is what the xpath is checking for. 

The challenge is the notebook is usable in a "kernel not ready" state, but the query itself will not run until the kernel is ready. Meaning, the user will have no issues typing the following line: print(123456789099876543210990+9876543219), but it will not run. This will cause the test to fail at that step.

I do feel like there has to be a cleaner way to address this, especially if the timeout is exceeded because of a failure somewhere. Still trying to figure out something

petesantos: Is this still a WIP,or are you looking to merge? 
If you want to put the timeout in for now to prevent some immediate failures but are investigating further, I am ok with that. If that is the case, could you put a comment in the code with the ticket number?

xuscrbw: This is still a work in progress. I am testing some changes to the test itself. Will add the ticket number also

xuscrbw:
Closing. The issue is not a UI test problem

PR #207, opened 2018-05-30
apply min width to body

MattPutnam:

PR #2642, opened 2021-09-30
[PF-692] Support regionality in Terra

wnojopra:
# This PR:
1) Adds a dropdown to the "Create a New Workspace" dialog for Bucket location. It defaults to the current default setting, US multi-regional.
2) Adds a dropdown for choosing a compute region. Defaults to location of workspace bucket
3) Adds informative messaging to ComputeModal
4) Adds informative messaging to the Import Table Data dialog
5) Add informative messaging to Launch Analysis Modal

# I've tested:
- That creating a workspace with the default location works, and a non-default location.
- Cloud Enviroment creation testing:
  - Creating a GCE instance with the default us-multiregional bucket.
  - Creating a cluster with the default us-multiregional bucket.
  - Creating a GCE instance in a non default location (like australia-southeast1)
  - Creating a cluster in a non default location (like australia-southeast1)

# Release planning
We are targeting October 18 for these changes to go live. With this release we will also be working with comms team to release a set of Terra articles explaining regionality (when/why you want to use it, egress risks, etc). These articles are available to preview as Google Docs in [Working with Non-US data in Terra](https://docs.google.com/document/d/1I-RKjtc0REMDZPJANCO2HcmBtKXHxDQeGiwliaYzFFs/edit) and [Regional Storage and Compute now available in Terra](https://docs.google.com/document/d/13DAwRXVM-qX5jYYz6o5EOMcTUXXhKoOBby2TK8kx0xY/edit#).

# Screenshots of UI changes:
Create/clone workspace
image

Create Cloud Environment
image

ComputeModal update
image

Import Table Data update
image

Launch Analysis Modal update
image


src/components/region-common.js
```diff
@@ -1,74 +1,73 @@
...
+      ['EUROPE-NORTH1', () => ({ flag: '🇫🇮', regionDescription: `${regionDescription} (Finland)`, computeZone: 'europe-north1-a', computeRegion: 'europe-north1' })],
+      ['EUROPE-WEST1', () => ({ flag: '🇧🇪', regionDescription: `${regionDescription} (Belgium)`, computeZone: 'europe-west1-b', computeRegion: 'europe-west1' })],
+      ['EUROPE-WEST2', () => ({ flag: '🇬🇧', regionDescription: `${regionDescription} (London)`, computeZone: 'europe-west2-a', computeRegion: 'europe-west2' })],
+      ['EUROPE-WEST3', () => ({ flag: '🇩🇪', regionDescription: `${regionDescription} (Frankfurt)`, computeZone: 'europe-west3-a', computeRegion: 'europe-west3' })],
+      ['EUROPE-WEST4', () => ({ flag: '🇳🇱', regionDescription: `${regionDescription} (Netherlands)`, computeZone: 'europe-west4-a', computeRegion: 'europe-west4' })],
+      ['EUROPE-WEST6', () => ({ flag: '🇨🇭', regionDescription: `${regionDescription} (Zurich)`, computeZone: 'europe-west6-a', computeRegion: 'europe-west6' })],
+      ['NORTHAMERICA-NORTHEAST1', () => ({ flag: '🇨🇦', regionDescription: `${regionDescription} (Montreal)`, computeZone: 'northamerica-northeast1-a', computeRegion: 'northamerica-northeast1' })],
```
wnojopra: Realize there is a NORTHAMERICA-NORTHEAST2 region now. Willy to review and update entire list.

kyuksel:
Hi @wnojopra - I meant to write earlier that I had wanted to review the related design doc first, and hence the delay. Will get to it soon. Thank you for your patience.

src/components/NewWorkspaceModal.js
```diff
@@ -175,6 +195,30 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
...
       ])]),
+      h(IdContainer, [id => h(Fragment, [
+        h(FormLabel, { htmlFor: id }, [
+          'Bucket location',
+          h(InfoBox, { style: { marginLeft: '0.25rem' } }, [
+            'A bucket location can only be set when creating a workspace. ',
+            'Once set, it cannot be changed. ',
+            'A cloned workspace will automatically inherit the bucket location from the original workspace but this may be changed at clone time.',
+            p([
+              'By default, workflow and Cloud Environment VMs will run in the same region as the workspace bucket. ',
```
kyuksel: Nit: It feels like `Cloud Environments` would suffice as `VM` is not a term we mean to use a lot on Terra UI interactive analysis sections going forward, and may not be completely intuitive in the case of Spark clusters and Kubernetes clusters (where we deploy Galaxy). Similarly for line #207 below and also for [Compute Modal line #875](https://github.com/DataBiosphere/terra-ui/pull/2642/files#diff-149d7ad595448fb2f6bedcff6dde5f4cbcccd8d5148fed8820ce88190a3bc38bR875). 
```suggestion
              'By default, workflow and Cloud Environments will run in the same region as the workspace bucket. ',
```

src/components/region-common.js
yonghaoy: Can you try using NORTHAMERICA-NORTHEAST2 locally to create a runtime/PAPI jobs? I doubt it would work because Terra does not create VPC subnet in that region.

wnojopra: Leo will create a VPC subnet once https://github.com/DataBiosphere/leonardo/pull/2319 is deployed. But it doesn't matter anyway, the UI will only allow one region per country to create runtimes. For Canada, we're only showing NORTHAMERICA-NORTHEAST1. See comment in https://github.com/DataBiosphere/terra-ui/pull/2642/files#diff-1fcbff2d49cbfcbbbd2e6fc3f78f453a36ae84030cbddfec06885366f9c00bc8R55

src/components/region-common.js
```diff
@@ -1,74 +1,77 @@
...
+      ['US-WEST3', () => ({ flag: '🇺🇸', regionDescription: `${regionDescription} (Salt Lake City)`, computeZone: 'US-WEST3-A', computeRegion: 'US-WEST3' })],
+      ['US-WEST4', () => ({ flag: '🇺🇸', regionDescription: `${regionDescription} (Las Vegas)`, computeZone: 'US-WEST4-A', computeRegion: 'US-WEST4' })],
+      [Utils.DEFAULT, () => ({ flag: unknownRegionFlag, regionDescription, computeZone: 'UNKNOWN', computeRegion: 'UNKNOWN' })]
+    )],
+    [Utils.DEFAULT, () => ({ flag: unknownRegionFlag, regionDescription, computeZone: 'UNKNOWN', computeRegion: 'UNKNOWN' })]
+  )
 }
+
+export const allRegions = [
+  // In this list, us-east*, us-west*, northamerica-northeast2 and asia-northeast2 have purposefully been removed.
```
rtitle, line 56: Not a hard objection, just trying to understand. Is this restriction geared more toward workspace buckets? Should we allow the full list (us-east, etc) for cloud environments given we allow overriding from the "default" region anyway?

From my initial testing, I tried to create a cloud environment and was confused why there were so many `asia-` options and only 1 `us-` option. I imagine some users might be confused too.

src/components/region-common.js
yonghaoy: I see. Also I don't think Leo change works because there is no subnet exists in that region.
With PPW, Terra projects are all use VPC network, and it only create subnets in those regions: 
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java

. But I am not sure if Leo will create a subnet before creating cloud resources. I thought Leo will just take that reason to create dataproc/VMs directly.

cjllanwarne:
The changes in the workflow launch modal seem similar to an early mockup that was drafted but not approved... were the copied from sample screenshots in a ticket? 

I would suggest running the modal past UX because I think there were concerns at the time about over-filling the launch modal to the point of overwhelming users.

cjllanwarne:
We also need to be careful with workflows not to over-promise on regionality. The workflow itself will remain in Terra's single Cromwell instance, only the jobs we forward to the cloud will run in the region specified.

cjllanwarne:
The workflow modal also highlights an interesting discrepancy between the bucket region (with more information on what it means - ie "Montreal") and the compute region, which seems to be a raw region value from Google. Maybe we can route all region text through the "pretty printer" logic that is currently being used for that bucket region field?

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -79,7 +79,7 @@ const LaunchAnalysisModal = ({
     [type === chooseSetType, () => _.flow(mergeSets, _.uniqBy('entityName'))(selectedEntities).length]
   )
   const { location, locationType } = bucketLocation
-  const { flag, regionDescription } = regionInfo(location, locationType)
+  const { flag, regionDescription, computeRegion } = regionInfo(location, locationType)
```
cahrens: It doesn't seem like Cromwell actually respects the computeRegion-- it always uses us-central1. Therefore I don't think the changes in this file are helpful.

wnojopra:
@cjllanwarne @cahrens Hoping to clarify a few things (that are admittedly confusing):
1) The [lifesciences API has a concept of location](https://cloud.google.com/life-sciences/docs/concepts/locations). This location is used to store metadata about the pipeline operation. In fact, **it is only used to store metadata about the pipeline operation.**
2) The above location is **separate** from the region that the VM instances will actually run in. The locations where the VMs are allocated can be any of the Compute Engine regions and zones and are independent from where the pipeline metadata is stored. 
3) When launching a workflow in Terra, the location of the workspace bucket determines the default region that the VMs will run in. This has been true since Dec 2020 thanks to https://github.com/broadinstitute/rawls/pull/1319 . I have tested and confirmed today that if I have a workspace bucket in europe-north1 and launch a workflow from that workspace, then the VM will run in a europe-north1 zone.
https://github.com/broadinstitute/rawls/pull/1319

I hope that clarifies things. Earlier today on Slack I mentioned that the operation metadata was in us-central1 despite my workspace bucket existing in europe-north1. This is totally expected, and the reason why we state **compute** for the workflow will occur in the location. The [lifesciences API doesn't actually support all regions](https://cloud.google.com/life-sciences/docs/concepts/locations#regional_locations), so we would have had to choose a separate lifesciences metadata region anyway.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
wnojopra: us-central1 is always used for the [location of the lifesciences api metadata](https://cloud.google.com/life-sciences/docs/concepts/locations). This is separate from the location that the VMs will run in, which is what we're setting here. See my comment starting with "Hoping to clarify a few things..." for more details.

src/components/region-common.js
wnojopra: Its a good question, and one that we'll have answered in a Terra article (currently under review). Pasting the relevant part here:

> For many organizations and individual labs, the best long-term choice is to select a single region for bucket location. In the US, Terra makes available the us-central1 region(*).
> 
> (*)For data in the US, the Terra interface supports creating workspace buckets in US multi-region or us-central1. Terra's default region for Workflow and Cloud Environment VMs has historically been us-central1. The exclusion of us-east and us-west regions is intended to help the community avoid unintentionally creating data silos in different regions since cross-analysis of data in different regions is inherently more expensive due to network egress charges. If you have a specific need to create workspaces with buckets in these other US regions, please contact Terra Support.

TL;DR: We want to avoid silos of different regions within a country. 

I'm debating how important it is to put this information directly in the UI vs just linking to the proper articles.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
cahrens: Thanks for the information-- that definitely helps! I will talk it over with the Batch team to make sure everyone is in agreement.

As far as the dialog, to minimize the "wall of scary text" could you only show a warning in the case when the bucket region is different from the compute region? That seems to be the only time that the user should be concerned, and providing the warning messages _every_ time could have the effect of confusing users who don't have a problem, as well as making people numb to the messaging.

And please definitely do run the modal by UX.

kyuksel: > us-central1 is always used for the location of the lifesciences api metadata. This is separate from the location that the VMs will run in, which is what we're setting here

Would this be good to add as a code comment as well?

src/components/region-common.js
kyuksel: > I'm debating how important it is to put this information directly in the UI vs just linking to the proper articles.

Perhaps we can keep the existing comment and add to it referencing the proper articles for more info?

kyuksel:
When I created a workspace in `asia-southeast2 (Jakarta)` region and then attempted to select `europe-central2 (Warsaw)`:

image

the location pull-down seemed to remain on `Select…`

image

It didn't seem to happen when I tried selecting any of the other compute regions.

wnojopra:
Great catch. Fixed the europe-central2 (Warsaw) issue by adding it to the regionInfo list. It was missing because this region was added earlier this year https://cloud.google.com/blog/products/infrastructure/google-cloud-region-in-warsaw-poland-is-now-open

src/components/ComputeModal.js
```diff
@@ -188,6 +192,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     const runtimeConfig = desiredRuntime && {
       cloudService: desiredRuntime.cloudService,
       ...(desiredRuntime.cloudService === cloudServices.GCE ? {
+        zone: desiredRuntime.zone.toLowerCase(),
```
kyuksel, line 238: Could `computeZone` and `computeRegion` values be lower case so we wouldn't have `toLowerCase()` them here and below?

src/components/ComputeModal.js
```diff
@@ -511,6 +523,12 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     const { runtime: existingRuntime } = getExistingEnvironmentConfig()
     return existingRuntime && (!canUpdateRuntime() || isStopRequired())
   }
+
+  const onLocationChange = (location, locationType) => {
+    const { computeZone, computeRegion } = regionInfo(location, locationType)
```
kyuksel, line 579: For function names, we prefer verbs; e.g. `getRegionInfo`.

src/components/ComputeModal.js
```diff
@@ -844,6 +867,24 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+                h(Select, {
+                  id,
+                  isSearchable: false,
+                  value: computeConfig.computeRegion,
+                  onChange: ({ value, locationType }) => onLocationChange(value, locationType),
+                  options: _.flow(_.filter(l => l.value !== 'US'), _.sortBy('label'))(allRegions)
+                })
+              ])
+            ])
           ])
```
kyuksel: We talked about it offline but just wanted to call out here too that where this component is, will be occupied by a `Link` for Spark UI once the [corresponding PR](https://github.com/DataBiosphere/terra-ui/pull/2643) is in and we were hoping this component can be slated below `Compute type`?

src/components/ComputeModal.js
```diff
@@ -1068,7 +1109,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                   'If you want to permanently save some files from the disk before deleting it, you will need to create a new cloud environment to access it.'
                 ])
               ]),
-              existingRuntime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
+              // At this point there is no runtime to check the tool
+              h(SaveFilesHelpRStudio)
```
kyuksel: Could you explain the reason for this change a bit more? I'm trying to understand why there is no runtime at this point.

src/components/NewWorkspaceModal.js
```diff
@@ -69,6 +71,9 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
         attributes: { description },
         copyFilesWithPrefix: 'notebooks/'
       }
+      if (bucketLocation) {
+        body.bucketLocation = bucketLocation
+      }
```
kyuksel: It might be easier to read if we inlined this, like so

```javascript
const body = {
    namespace,
    name,
    authorizationDomain: _.map(v => ({ membersGroupName: v }), [...getRequiredGroups(), ...groups]),
    attributes: { description },
    copyFilesWithPrefix: 'notebooks/',
    ...(!!bucketLocation && { bucketLocation })
}
```

src/components/NewWorkspaceModal.js
```diff
@@ -106,8 +111,23 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
...
+    setBucketLocation(locationResponse.location)
+  })
+
   // Lifecycle
-  Utils.useOnMount(() => { loadProjectsGroups() })
+  Utils.useOnMount(() => {
+    loadProjectsGroups()
+    if (cloneWorkspace) {
+      loadBucketLocation()
+    }
```
kyuksel: We'd typically write this as below for conciseness:
```suggestion
    !!cloneWorkspace && loadBucketLocation()
```

src/components/NewWorkspaceModal.js
```diff
@@ -106,8 +111,23 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
     Ajax(signal).Groups.list().then(setAllGroups)
   ]))
 
+  const loadBucketLocation = _.flow(
+    withErrorReporting('Error loading data'),
+    Utils.withBusyState(setLoading)
+  )(async () => {
+    const [locationResponse] = await Promise.all([
+      Ajax(signal).Workspaces.workspace(namespace, cloneWorkspace.workspace.name).checkBucketLocation(cloneWorkspace.workspace.googleProject, cloneWorkspace.workspace.bucketName)
+    ])
```
kyuksel: Is `Promise.all` necessary since there seems to be a single `Promise` being input?

```suggestion
    const locationResponse = await Ajax(signal).Workspaces.workspace(namespace, cloneWorkspace.workspace.name).checkBucketLocation(cloneWorkspace.workspace.googleProject, cloneWorkspace.workspace.bucketName)
```

src/components/NewWorkspaceModal.js
```diff
@@ -175,6 +195,30 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
...
+            h(Link, {
+              href: 'https://support.terra.bio/hc/en-us/articles/360058964552',
+              ...Utils.newTabLinkProps
+            }, ['Read more about bucket locations'])
+          ])
+        ]),
+        h(Select, {
+          id,
+          value: bucketLocation,
+          onChange: ({ value }) => setBucketLocation(value),
```
kyuksel, line 207: `value` is a string, correct? If so, we wouldn't want to do object deconstruction (i.e. ` { value }`)?
Also,
```javascript
onChange: value => setBucketLocation(value),
```

can be written as
```javascript
onChange: setBucketLocation,
```

src/components/region-common.js
```diff
@@ -1,74 +1,78 @@
...
+
 // Get a { flag: ..., countryName: ... } object representing a google locationType/location input.
 // 'flag' will always be defined (even if it's a question mark.
 // 'regionDescription' is the same as location when locationType is 'multi-region', or a country name when locationType is 'region'.
+// computeZone is generally the 'a' zone for each region, except for those regions where it is not available.
+// The choice to use the 'a' zone is arbitrary, choosing 'b' zone would also work.
+// The region choice for multi-region locations is arbitrary as well.
+
+// When updating region list, please also update the list in
+// https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf
```
kyuksel, line 11: Thanks for the helpful comments!

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -110,6 +110,15 @@ const LaunchAnalysisModal = ({
           p(['Note that metadata about this run will be stored in the US.'])
         ])]) : 'Loading...'
     ]),
+    div(['Compute for this workflow will occur ', span({ style: { fontWeight: 'bold' } }, 'by default '), 'in the region:']),
+    div({ style: { margin: '1rem 0 1.5rem' } }, [
+      location ? h(Fragment, [span({ style: { marginRight: '0.5rem' } }, [flag]),
+        h(Fragment, [span({ style: { marginRight: '0.5rem' } }, [computeRegion])])]) : 'Loading...'
+    ]),
```
kyuksel: I had a little difficult time parsing the nesting above. I wonder if the following may be a bit easier:

```suggestion
    div({ style: { margin: '1rem 0 1.5rem' } }, [
      location ?
        h(Fragment, [
          span({ style: { marginRight: '0.5rem' } }, [flag]),
          h(Fragment, [span({ style: { marginRight: '0.5rem' } }, [computeRegion])])
        ]) :
        'Loading...'
    ]),
```

kyuksel:
What's the expected behaviour when a user creates compute in region X, and then chooses region Y on `ComputeModal`? I assume we won't allow an update of the compute in that case? If so, we should disable the `UPDATE` button on `ComputeModal`.

src/components/ComputeModal.js
wnojopra: The issue is the Google Cloud Storage API returns locations in uppercase, but the Leonardo APIs accept locations in lowercase. 

I had designed the Leonardo APIs to accept lowercase locations because the GCE metadata APIs return locations in lowercase 👎 . So we can blame Google for the inconsistency. 

Probably the best thing for us to do is update the Leonardo APIs to be case insensitive with locations. This would be fairly easy to implement. There's never going to be a time when US-CENTRAL1 is different from us-central1. Then the UI code here wouldn't need to convert to lowercase.

src/components/ComputeModal.js
wnojopra: On line 1097 we check !existingRuntime. This is the block where there is no existing runtime. Updated the comment to clarify.

src/components/NewWorkspaceModal.js
wnojopra: It's written this way because `value` is an object like `{ value: 'ASIA-EAST1', label: 'asia-east1 (Taiwan)', locationType: 'region' }`. See the const `allRegions` in `region-common.js`.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
wnojopra: I've run this by UX, and also an updated version of it. Awaiting feedback on it. 

I have added it as a code comment.

src/components/region-common.js
wnojopra: Updated the comment here approriately.

src/components/NewWorkspaceModal.js
wnojopra: Done. Thanks!

src/components/ComputeModal.js
wnojopra: Done. Thanks!

src/components/NewWorkspaceModal.js

src/components/NewWorkspaceModal.js

src/components/NewWorkspaceModal.js

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
wnojopra: Done.

wnojopra:
> What's the expected behaviour when a user creates compute in region X, and then chooses region Y on `ComputeModal`? I assume we won't allow an update of the compute in that case? If so, we should disable the `UPDATE` button on `ComputeModal`.

You're right, it doesn't make sense to update the location of an existing runtime. I think either disabling the update button or even disabling the location dropdown would work.

src/components/common.js
```diff
@@ -480,7 +480,7 @@ export const CromwellVersionLink = props => {
       href: `https://github.com/broadinstitute/cromwell/releases/tag/${version}`,
       ...Utils.newTabLinkProps,
       ...props
-    }, ['Cromwell ', version]) :
+    }, ['Cromwell ', version, icon('pop-out', { size: 12, style: { margin: '0 0.25rem 1px' } })]) :
```
cahrens: Let's go back to just setting marginLeft as 0.25rem. When I checked out the code and tried it, it looks OK to me, and it's consistent with all the other usages.

We definitely don't want 0.25rem for marginRight, as that spaces the period out too far.

Thanks!

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -110,6 +112,26 @@ const LaunchAnalysisModal = ({
           p(['Note that metadata about this run will be stored in the US.'])
         ])]) : 'Loading...'
     ]),
+    div(['Running workflows will generate cloud charges. ',
+      h(InfoBox, [
+        p(['If you run a large workflow or a large number of small workflows, you may generate significant cloud costs ',
+          'for compute, disks, and network egress charges. When your workflow stages complete, the intermediate/output results ',
+          'will generate storage costs.'])
+      ])]),
+    div([h(Link, {
```
cahrens: Can you add .25rem margin between "Running workflows…" and the link below it?

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -110,6 +112,26 @@ const LaunchAnalysisModal = ({
...
+    ])]),
+    div([h(Link, {
+      style: { verticalAlign: 'top', marginTop: '0.5rem' },
+      href: 'https://support.terra.bio/hc/en-us/articles/360057589931', ...Utils.newTabLinkProps
+    }, [
+      'Set up budget alert',
+      icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
+    ])]),
     h(IdContainer, [id => div({ style: { margin: '1rem 0' } }, [
       label({ htmlFor: id, style: { display: 'block', margin: '1rem 0' } }, ['Describe your submission (optional):']),
```
cahrens: Can you change this so that there is only .5 rem between "Describe your submission…" and the input field? I'm trying to get the 3 regions in this dialog more clearly delineated.

src/components/common.js
wnojopra: Done

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
wnojopra: Done

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
wnojopra: Done. Thanks for reviewing!

src/pages/Environments.js
```diff
@@ -505,7 +505,7 @@ const Environments = () => {
               return h(Sortable, { sort: diskSort, field: 'cost', onSort: setDiskSort }, [`Cost / month (${Utils.formatUSD(totalDiskCost)} total)`])
             },
             cellRenderer: ({ rowIndex }) => {
-              return Utils.formatUSD(persistentDiskCostMonthly(filteredDisks[rowIndex]))
+              return Utils.formatUSD(getPersistentDiskCostMonthly(filteredDisks[rowIndex], 'US-CENTRAL1'))
```
cahrens: Why is this set to US-CENTRAL1? Should it reference a constant (such as the one defined in runtime-utils)?

src/data/machines.js
```diff
@@ -46,22 +46,127 @@ export const gpuTypes = [
...
-export const monthlyStoragePrice = 0.04 // from https://cloud.google.com/compute/pricing
-export const storagePrice = monthlyStoragePrice / 730 // per GB hour using 730 hours per month from https://cloud.google.com/compute/pricing
 export const dataprocCpuPrice = 0.01 // dataproc costs $0.01 per cpu per hour
 export const ephemeralExternalIpAddressPrice = { // per hour in dollars for Iowa (us-central1) region per https://cloud.google.com/vpc/network-pricing#ipaddress
   standard: 0.004,
   preemptible: 0.002
 }
 
+export const regionToDiskPrice = [
+  { name: 'ASIA-EAST1', monthlyDiskPrice: 0.04 },
```
cahrens, line 144: How will prices (here and above) be kept up-to-date with Google's pricing?

src/libs/runtime-utils.js
```diff
@@ -32,13 +32,19 @@ export const defaultNumDataprocPreemptibleWorkers = 0
 export const defaultGpuType = 'nvidia-tesla-t4'
 export const defaultNumGpus = 1
 
+export const defaultLocation = 'US'
+export const defaultLocationType = 'multi-region'
```
cahrens, line 36: I don't see that defaultLocationType and defaultComputeZone are used. Are they here for documentation purposes only, or were they intended to be used?

cahrens: Actually, I see now that defaultComputeZone is used by ComputeModal, which was collapsed in the diff.

Also defaultLocationType.

src/pages/Environments.js
wnojopra: Good catch, it should have used those constants. Fixed.

src/data/machines.js
wnojopra: Ideally, we would have some job running periodically that could update our database of prices. I have written a script that can pull these prices. Outside of this PR we could use this script in a cronjob somewhere. 

But I think for now we're fine to have the prices hard-coded here. They are not expected to change often at all. Prior to this PR the price of us-central1 resources were hard coded and haven't changed at all.

src/components/ComputeModal.js
```diff
@@ -582,7 +582,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   }
 
   const isDifferentLocation = () => {
-    return bucketLocation === 'US' ? computeConfig.computeRegion !== 'US-CENTRAL1' : bucketLocation !== computeConfig.computeZone
+    return bucketLocation === 'US' ? computeConfig.computeRegion !== defaultComputeRegion : computeConfig.computeRegion !== bucketLocation
```
cahrens, line 589: Why is US treated differently?

wnojopra: Buckets are either regional or multi-regional. In production Terra today, you can only create a US multi-regional bucket.

If the bucket is regional, we can easily compare the bucketLocation with the compute region. 

Here, bucketLocation === 'US' means the bucket is US multi-regional. For a US multi-regional bucket, the computeRegion needs to be US-CENTRAL1 in order to be considered "in the same location". 

This is a good question so I'll add a comment to clarify this.

cahrens: But there will be other multi-regional buckets that we will need to handle, correct (not now, but later)?

wnojopra: Probably not, actually. For many organizations and individual labs, the best long-term choice is to use a single region for bucket location. Storage in a regional bucket is cheaper than a multi-regional bucket, and storage tends to be a majority of cloud costs for researchers. We've written up [US Regional or Multi-regional US buckets: tradeoffs](https://docs.google.com/document/d/1RMu8bxXAyP2q_85UHkPARW1Xpt_G_662pRi0LCML4yA/edit#) if you're interested in more details. 

Another argument that supports this is the Rawls createWorkspace API was updated in Dec 2020 to support specification of a Google Cloud region. If no region is specified, the default (US multi-regional) is used. No other multi-regional values (ASIA, EU) are currently supported.

cahrens: OK, as long as no other multi-regional values are going to be supported anytime in the planned future. Can you add to the comment that 'US' is the only multi-region is supported?

src/components/ComputeModal.js
```diff
@@ -1001,8 +1001,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           id => div({ style: { gridColumnEnd: 'span 3' } }, [
             label({ htmlFor: id, style: computeStyles.label }, ['Location']),
             h(InfoBox, { style: { marginLeft: '0.5rem' } }, [
-              'By default, workflow and cloud environments will run in the same region as the workspace bucket. ',
-              p(['Currently, only a limited amount of locations are available but more will become available over time.'])
+              'Cloud environments run in the same region as the workspace bucket and cannot be changed at this time.',
```
cahrens, line 1005: Suggestion: where you are changing things that will need to be updated with subsequent commits, it would be nice to include a comment with the JIRA ticket for the future work, so that it easy to search for the places that need to be updated. If there are not JIRA tickets for each of the planned "next steps", it would be good to create them.

src/pages/workspaces/workspace/workflows/LaunchAnalysisModal.js
```diff
@@ -1,7 +1,7 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
 import { b, div, h, label, p, span } from 'react-hyperscript-helpers'
-import { ButtonPrimary, CromwellVersionLink, IdContainer } from 'src/components/common'
+import { ButtonPrimary, CromwellVersionLink, IdContainer, Link } from 'src/components/common'
```
cahrens, line 4: Changes to this dialog LGTM. Thanks for all the iterations!

src/components/ComputeModal.js
cahrens: (It might be too late at this point to make this suggestion worthwhile, since you would be doing the same sort of searching through the changes for things that will need to be updated to add comments as you would at the time of making the changes)

cahrens (approved):
I am done reviewing and testing the changes to the Launch Submissions dialog. I did some spot checking of other areas (and left a couple of suggestions), but I am leaving it to the reviewers from other teams to fully review/test code in their areas.

src/components/ComputeModal.js
```diff
@@ -952,6 +995,27 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           }),
           h(DataprocDiskSelector, { value: computeConfig.workerDiskSize, onChange: updateComputeConfig('workerDiskSize') })
         ])
+      ]),
+      div({ style: { ...gridStyle, gridTemplateColumns: '0.25fr 8.5rem 1fr 5.5rem 1fr 5rem', marginTop: '1.5rem' } }, [
+        h(IdContainer, [
+          id => div({ style: { gridColumnEnd: 'span 3' } }, [
+            label({ htmlFor: id, style: computeStyles.label }, ['Location']),
+            h(InfoBox, { style: { marginLeft: '0.5rem' } }, [
+              'Cloud environments run in the same region as the workspace bucket and cannot be changed at this time.',
```
kyuksel, line 1005:
```suggestion
              'Cloud environments run in the same region as the workspace bucket and cannot be changed at this time.'
```

to fix

```
Compiled with warnings.

src/components/ComputeModal.js
  Line 1004:118:  Unexpected trailing comma  comma-dangle
```

src/components/region-common.js
```diff
@@ -1,74 +1,81 @@
+import * as Utils from 'src/libs/utils'
+
 // Get a { flag: ..., countryName: ... } object representing a google locationType/location input.
 // 'flag' will always be defined (even if it's a question mark.
```
kyuksel, line 4: Nit:

```suggestion
// 'flag' will always be defined (even if it's a question mark).
```

src/data/machines.js
```diff
@@ -46,22 +46,127 @@ export const gpuTypes = [
   { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 1, maxNumCpus: 12, maxMem: 78, price: 2.4800, preemptiblePrice: 0.7400 },
   { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 2, maxNumCpus: 24, maxMem: 156, price: 4.9600, preemptiblePrice: 1.4800 },
   { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 4, maxNumCpus: 48, maxMem: 312, price: 9.9200, preemptiblePrice: 2.9600 },
-  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 8, maxNumCpus: 96, maxMem: 624, price: 19.8400, preemptiblePrice: 5.9200 }
+  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 8, maxNumCpus: 96, maxMem: 624, price: 19.8400, preemptiblePrice: 5.9200 },
+  { name: 'NVIDIA Tesla VP00', type: 'nvidia-tesla-p100', numGpus: 1, maxNumCpus: 16, maxMem: 104, price: 1.4600, preemptiblePrice: 0.4300 },
+  { name: 'NVIDIA Tesla VP00', type: 'nvidia-tesla-p100', numGpus: 2, maxNumCpus: 32, maxMem: 208, price: 2.9200, preemptiblePrice: 0.8600 },
+  { name: 'NVIDIA Tesla VP00', type: 'nvidia-tesla-p100', numGpus: 4, maxNumCpus: 96, maxMem: 624, price: 5.8400, preemptiblePrice: 1.7200 }
```
kyuksel, line 52: Typo?

```suggestion
  { name: 'NVIDIA Tesla P100', type: 'nvidia-tesla-p100', numGpus: 1, maxNumCpus: 16, maxMem: 104, price: 1.4600, preemptiblePrice: 0.4300 },
  { name: 'NVIDIA Tesla P100', type: 'nvidia-tesla-p100', numGpus: 2, maxNumCpus: 32, maxMem: 208, price: 2.9200, preemptiblePrice: 0.8600 },
  { name: 'NVIDIA Tesla P100', type: 'nvidia-tesla-p100', numGpus: 4, maxNumCpus: 96, maxMem: 624, price: 5.8400, preemptiblePrice: 1.7200 }
```

src/components/ComputeModal.js
wnojopra: Done!

src/components/ComputeModal.js
wnojopra: Fixed in the new PR

src/components/region-common.js
wnojopra: Fixed in the new PR

src/data/machines.js
wnojopra: Fixed in the new PR. Good catch.

wnojopra:
Closing in favor of #2660

PR #2643, opened 2021-09-30
[IA-2888] Support Spark UIs

kyuksel:
### For Reviewers
Recommended view: [Unified diff without whitespaces](https://github.com/DataBiosphere/terra-ui/pull/2643/files?diff=unified&w=1)

### Ticket
[IA-2888](https://broadworkbench.atlassian.net/browse/IA-2888)

### Highlights
- Five of the Spark UIs made available by Google Dataproc are proxied by Leonardo. The Terra UI can now take advantage of this and make four of these UIs accessible to Terra users. The ticket for the remaining one (HDFS UI) is [here](https://broadworkbench.atlassian.net/browse/IA-2985). Please see the screenshots for more details. 
- Rename `details` `Link`s as `view` `Link`s on the tables on `Cloud Environment` page.

### Screenshots
Current (dev) version of the Compute (formerly Runtime) Modal:

image

Link added as highlighted below:

Screen Shot 2021-09-30 at 6 25 52 PM

Pressing the link when enabled displays

image

Clicking one of the `Launch` buttons creates a new tab with the selected Spark UI

image

The link is disabled with a tooltip when the runtime isn't a running Spark cluster or Spark master node.

image

### Testing
- Locally deployed the changes and ran against dev backends, using all three runtime types in various statuses.


src/components/ComputeModal.js
```diff
@@ -33,15 +34,6 @@ import validate from 'validate.js'
...
 
-// TODO Factor out common pieces with GalaxyModal.styles into runtime-utils
-const styles = {
-  label: { fontWeight: 600, whiteSpace: 'pre' },
-  titleBar: { marginBottom: '1rem' },
-  drawerContent: { display: 'flex', flexDirection: 'column', flex: 1, padding: '1.5rem' },
-  warningView: { backgroundColor: colors.warning(0.1) },
-  whiteBoxContainer: { padding: '1.5rem', borderRadius: 3, backgroundColor: 'white' }
-}
-
```
kyuksel, line 44: Opportunistically took care of this `TODO`  😊

src/components/ComputeModal.js
```diff
@@ -101,6 +93,34 @@ const DataprocDiskSelector = ({ value, onChange }) => {
...
+    ['yarn', () => 'YARN Resource Manager'],
+    ['apphistory', () => 'YARN Application Timeline'],
+    ['hdfs', () => 'HDFS NameNode'],
+    ['sparkhistory', () => 'Spark History Server'],
+    ['jobhistory', () => 'MapReduce History Server']
+  )
+
+  return div({ style: { ...styles.whiteBoxContainer, marginBottom: '1rem', backgroundColor: colors.accent(0.1), boxShadow: Style.standardShadow } }, [
+    div({ style: { flex: '1', lineHeight: '1.5rem', minWidth: 0, display: 'flex' } }, [
+      div([
```
kyuksel: It feels there should be an easy way to deem this `div` unnecessary but it isn't obvious to me how.

src/components/ComputeModal.js
```diff
@@ -101,6 +93,34 @@ const DataprocDiskSelector = ({ value, onChange }) => {
...
+        div({ style: { ...styles.headerText, marginTop: '0.5rem' } }, [interfaceDisplayName]),
+        p([synopsis]),
+        div({ style: { display: 'flex', marginTop: '1rem' } }, [
+          h(ButtonOutline, {
+            disabled: false,
+            href: Nav.getLink('workspace-spark-interface-launch', { namespace, name, application: 'spark', sparkInterface: interfaceName }),
```
kyuksel: Added `workspace-spark-interface-launch` to `navPaths` in `AppLauncher` to leverage the tab launching logic existed there for `Terminal` and `RStudio`.

src/components/ComputeModal.js
```diff
@@ -141,6 +161,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     numberOfPreemptibleWorkers: defaultNumDataprocPreemptibleWorkers,
     workerMachineType: defaultDataprocMachineType,
     workerDiskSize: defaultDataprocDiskSize,
+    componentGatewayEnabled: true, // We enable web interfaces (aka Spark console) for all new Dataproc clusters.
```
kyuksel, line 185: Since we're enabling [Dataproc Component Gateway](https://cloud.google.com/dataproc/docs/concepts/accessing/dataproc-gateways) on all Dataproc clusters, we shouldn't need to keep track of this state, may one rightfully think. We're not enabling it retroactively, though, so we may still encounter an old Dataproc clusters where it's not enabled, in which case we'll need to know not to enable the managing of Spark UIs on Compute (formerly Runtime) Modal.

src/components/ComputeModal.js
```diff
@@ -828,14 +857,17 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
             ])
           ]),
           h(IdContainer, [
-            id => div({ style: { gridColumnEnd: 'span 3', marginTop: '0.5rem' } }, [
+            id => div({ style: { gridColumnEnd: 'span 4', marginTop: '0.5rem' } }, [
```
kyuksel, line 884: The increase to `span 4` is to account for the length of the `Compute type` pull-down menu after adding the [span element below](https://github.com/DataBiosphere/terra-ui/pull/2643/files?w=1#diff-149d7ad595448fb2f6bedcff6dde5f4cbcccd8d5148fed8820ce88190a3bc38bR881). 

As you can see by comparing the old version

Screen Shot 2021-09-30 at 5 40 00 PM

vs. the new version

Screen Shot 2021-09-30 at 5 42 24 PM

the pull-down menu is still slightly shorter now but that is desirable to me since the old version spanned unnecessarily long even for the longest pull-down menu option (i.e. `Spark master node`).

src/components/GalaxyModal.js
```diff
@@ -15,20 +15,11 @@ import { withErrorReporting } from 'src/libs/error'
...
 
-// TODO Factor out common pieces with ComputeModal.styles into runtime-utils
-const styles = {
-  label: { fontWeight: 600, whiteSpace: 'pre' },
-  value: { fontWeight: 400, whiteSpace: 'pre' },
-  whiteBoxContainer: { padding: '1rem', borderRadius: 3, backgroundColor: 'white' },
-  drawerContent: { display: 'flex', flexDirection: 'column', flex: 1, padding: '1.5rem' },
-  headerText: { fontSize: 16, fontWeight: 600 },
-  warningView: { backgroundColor: colors.warning(0.1) }
-}
```
kyuksel, line 31: Opportunistically took care of this TODO 😊

src/components/GalaxyModal.js
```diff
@@ -209,7 +200,6 @@ export const GalaxyModalBase = Utils.withDisplayName('GalaxyModal')(
       ])
     }
 
-
```
kyuksel, line 212: Always thinking of you @zarsky-broad  😊

src/pages/Environments.js
```diff
@@ -234,7 +234,7 @@ const Environments = () => {
         div([span({ style: { fontWeight: '600' } }, ['Name: ']), appName]),
         disk && div([span({ style: { fontWeight: '600' } }, ['Persistent Disk: ']), disk.name])
       ])
-    }, [h(Link, ['details'])])
+    }, [h(Link, ['view'])])
```
kyuksel, line 237: These `details` => `view` changes aren't related to the ticket. They had been requested by UX and I forgot to change them in my previous PR, and wanted to squeeze them in now. Sorry about that.

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -15,13 +15,27 @@ import * as Utils from 'src/libs/utils'
 import { wrapWorkspace } from 'src/pages/workspaces/workspace/WorkspaceContainer'
 
 
+const getSparkInterfaceSource = (proxyUrl, sparkInterface) => {
+  console.assert(_.endsWith('/jupyter', proxyUrl), 'Unexpected ending for proxy URL')
+  const proxyUrlWithlastSegmentDropped = _.flow(_.split('/'), _.dropRight(1), _.join('/'))(proxyUrl)
+  return `${proxyUrlWithlastSegmentDropped}/${sparkInterface}`
+}
```
kyuksel, line 23: Why is the string manipulation shenanigans, you might wonder 🤔  Turns out for Jupyter runtimes (both GCE and Dataproc), Leonardo returns a `proxyUrl` mapping to the `Jupyter` endpoint and not the Spark UI endpoints. Therefore we need to strip off `jupyter` at the end of the URL before adding the Spark UI-specific segment. I'm planning on filing a Leo ticket for us to consider returning a map of proxy URLs instead, now that we have several of them. With that, this string manipulation would go away and the `switchCase` below would be simpler.

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -15,13 +15,27 @@ import * as Utils from 'src/libs/utils'
...
+
 const ApplicationLauncher = _.flow(
   Utils.forwardRefWithName('ApplicationLauncher'),
   wrapWorkspace({
     breadcrumbs: props => breadcrumbs.commonPaths.workspaceDashboard(props),
     title: _.get('application')
   })
-)(({ namespace, name, refreshRuntimes, runtimes, persistentDisks, application, workspace }, ref) => {
+)(({ sparkInterface, refreshRuntimes, runtimes, persistentDisks, application, workspace }, ref) => {
```
kyuksel: `namespace` and `name` were unused.

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -97,5 +111,11 @@ export const navPaths = [
     path: '/workspaces/:namespace/:name/applications/:application',
     component: ApplicationLauncher,
     title: ({ name, application }) => `${name} - ${application}`
+  },
+  {
+    name: 'workspace-spark-interface-launch',
+    path: '/workspaces/:namespace/:name/applications/:application/:sparkInterface',
+    component: ApplicationLauncher,
+    title: ({ name, application, sparkInterface }) => `${name} - ${application} - ${sparkInterface}`
   }
```
kyuksel, line 155: Added this path here in AppLauncher to leverage the tab launching logic existed there for Terminal and RStudio.
I wanted to roll the new path into the path above it via

```javascript
name: 'workspace-application-launch',
path: '/workspaces/:namespace/:name/applications/:application/:sparkInterface',
component: ApplicationLauncher,
title: ({ name, application }) => `${name} - ${application} - ${sparkInterface}`
```

But it doesn't seem to work for the cases where `application` is `Terminal` or `RStudio`, in which case there is no `sparkInterface`  🤷🏼

src/components/ComputeModal.js
jdcanas: Can it be replaced with `h(Fragment`? It's not apparent if the div is actually providing style, but since it only has children and no style attribute, it doesn't seem to be adding anything valuable to the DOM on the surface.

src/components/ComputeModal.js
```diff
@@ -572,16 +599,18 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         const isDataproc = (sparkMode, runtimeConfig) => !sparkMode && !runtimeConfig?.diskSize
         const runtimeConfig = currentRuntimeDetails?.runtimeConfig
         const gpuConfig = runtimeConfig?.gpuConfig
-
-        setSparkMode(runtimeConfig?.cloudService === cloudServices.DATAPROC ?
+        const newSparkMode = runtimeConfig?.cloudService === cloudServices.DATAPROC ?
```
jdcanas: I really dislike nested ternaries. Can we make this into a `Utils.cond` so the logic is easier to parse?

src/components/ComputeModal.js
```diff
@@ -844,6 +876,14 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                 })
               ])
             ])
+          ]),
+          // TODO: Is there a more robust way to center Link vertically wrt. the Select component above?
```
jdcanas: Yeah, the best way to do most alignment is via a flexbox. See this for a nice illustration (worth experimenting, it seems preferable over using padding to center, which is a lot less responsive to resizing.). The question in the link isn't the same, but the solution applies to elements other than text. I did this heavily in analysis tab components, if you want to search for the `align-items: center;` portion to find in-code examples.

https://stackoverflow.com/questions/5166574/vertical-align-text-in-block-element/9406982

src/components/ComputeModal.js
```diff
@@ -1157,9 +1197,11 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       value: selectedLeoImage,
       onChange: ({ value }) => {
         const requiresSpark = _.find({ image: value }, leoImages)?.requiresSpark
+        const newSparkMode = requiresSpark ? (sparkMode || 'master') : false
```
jdcanas, line 1226: It seems preferable to avoid multi-typed variables wherever possible despite the language supporting it. I don't like that this variable is overloaded with both the spark mode, and a boolean which I don't inherently understand the meaning of. 

I'd say either change `false` to `undefined` (which would be a more clear falsey value to indicate its not applicable. It is technically also a different type, but that's JavaScripts fault), or separate it into a flag that is checked before the (enum?) string

src/components/ComputeModal.js
```diff
@@ -1294,6 +1336,73 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+        onPrevious: () => setViewMode(undefined)
+      }),
+      p([
+        `Some of the Spark cluster components such as Apache Hadoop and Apache Spark
+         provide web interfaces. These interfaces can be used to manage and monitor cluster
+         resources and facilities, such as the YARN resource manager, the Hadoop Distributed
+         File System (HDFS), MapReduce, and Spark.`
+      ]),
+      h(SparkInterface, {
+        interfaceName: 'yarn',
```
jdcanas: Magic string alert. I'd consider making an array holding all the interface name strings and then mapping them to these SparkInterface components (this is what we do for most components that appear multiple times, i.e. notebooks/analyses). This would cut down on the number of lines considerably.

With that in mind, I'm also not sure why the caller (`ComputeModal` in this case) is responsible for passing the `synopsis`. It's derived state thats being derived too early which results in an unnecessary param. Seems like there should be an object of (k = interfaceName, v = synopsis) pairs in the SparkInterface  component itself.

src/libs/runtime-utils.js
```diff
@@ -10,6 +10,16 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
 
+export const styles = {
```
jdcanas: Can we have a slightly more descriptive name? I'm imagining a possible future world where we consolidate multiple large shared style objects here, and need to differentiate.

src/pages/workspaces/workspace/applications/AppLauncher.js
rtitle: Might be good to capture this in a code comment

src/pages/workspaces/workspace/applications/AppLauncher.js
jdcanas: The title is a simple string displayed at the top as I understand it. Should it be something like this:

```
title: ({ name, application, sparkInterface }) => `${name} - ${application}${sparkInterface ? ' -' + sparkInterface + ' ': ''}
```

(you can also just make it not a one line arrow function and add brackets... starting to look messy)

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -15,13 +15,27 @@ import * as Utils from 'src/libs/utils'
...
 
+const getSparkInterfaceSource = (proxyUrl, sparkInterface) => {
+  console.assert(_.endsWith('/jupyter', proxyUrl), 'Unexpected ending for proxy URL')
+  const proxyUrlWithlastSegmentDropped = _.flow(_.split('/'), _.dropRight(1), _.join('/'))(proxyUrl)
+  return `${proxyUrlWithlastSegmentDropped}/${sparkInterface}`
+}
+
+const getApplicationIFrameSource = (proxyUrl, application, sparkInterface) => {
+  return Utils.switchCase(application,
+    ['terminal', () => `${proxyUrl}/terminals/1`],
```
jdcanas: magic string alert... do we want to add this to the existing `tools` (its in runtime-common or utils I forget which) object with a `label` property in line with other tools? (I'm ok with just leaving this too since its not too egregious, but it does seem to fit within that existing tool/label pattern.

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -15,13 +15,27 @@ import * as Utils from 'src/libs/utils'
...
+    ['spark', () => getSparkInterfaceSource(proxyUrl, sparkInterface)],
+    [Utils.DEFAULT, () => proxyUrl] // RStudio
```
jdcanas: I'd prefer to explicitly have a branch for `tools.label.RStudio`, and have default give some sort of error.

rtitle:
I did some testing and it works well! A few comments/observations.

1. When I click a Spark UI it opens a new tab as expected, but it flickers a lot before the tool actually loads (not sure how to describe it better). It looks like the Terra app loads; then I get a cookie pop-up; then the Cloud Environment modal appears and disappears; then finally I get the Spark tool in an iframe. Is it possible to make this a little more seamless? Honestly a new tab without the Terra header would feel pretty nice as a user, I think.

2. A lot of the Spark UIs can be used to just view logs, which don't have any styling. It looks like this:
image
It feels a bit unintuitive navigating around. It seems like the browser Back button works, but it feels a bit wrong to use browser Back for an embedded UI. Should the iframe have its own "Back" button? This might also feel more natural if we didn't include the Terra header.

These are just personal observations, and may be out of scope.. thought I would share though. Comments appreciated.

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -15,13 +15,27 @@ import * as Utils from 'src/libs/utils'
...
 const ApplicationLauncher = _.flow(
   Utils.forwardRefWithName('ApplicationLauncher'),
   wrapWorkspace({
     breadcrumbs: props => breadcrumbs.commonPaths.workspaceDashboard(props),
     title: _.get('application')
   })
-)(({ namespace, name, refreshRuntimes, runtimes, persistentDisks, application, workspace }, ref) => {
+)(({ sparkInterface, refreshRuntimes, runtimes, persistentDisks, application, workspace }, ref) => {
   const cookieReady = Utils.useStore(cookieReadyStore)
   const [showCreate, setShowCreate] = useState(false)
```
jdcanas, line 43: huh... not sure why but the compute modal pops up in my face briefly when I navigate from a different page to the appLauncher sometimes...

It does it in dev too so not from this PR... might be worth a bug ticket (would be good to not hold this up)

src/components/ComputeModal.js
```diff
@@ -101,6 +93,34 @@ const DataprocDiskSelector = ({ value, onChange }) => {
   ])
 }
 
+const SparkInterface = ({ interfaceName, synopsis, namespace, name, onDismiss }) => {
+  const interfaceDisplayName = Utils.switchCase(interfaceName,
```
jdcanas: Why does it display this `Logged in as: dr. who` in the top right of the proxied iframe itself? Is this some backend configuration that can be changed?

image

src/components/ComputeModal.js
```diff
@@ -101,6 +93,34 @@ const DataprocDiskSelector = ({ value, onChange }) => {
...
+    ['yarn', () => 'YARN Resource Manager'],
+    ['apphistory', () => 'YARN Application Timeline'],
+    ['hdfs', () => 'HDFS NameNode'],
```
jdcanas: not sure how we would fix it off the top of my head, but the double sameish-color-but-slightly-off header is really jarring for hdfs..

image

src/components/ComputeModal.js
```diff
@@ -101,6 +93,34 @@ const DataprocDiskSelector = ({ value, onChange }) => {
...
+    ['sparkhistory', () => 'Spark History Server'],
```
jdcanas: Is this how it's supposed to look? It suggests we have an invalid path configuration, not enough context to confirm/deny (cc @rtitle ) 
image

jdcanas (approved):
LGTM code-wise.

Few nits here and there.

Would like some clarification on the questions around the pages themselves before merge, but approving and assuming they are correct given my lack of knowledge of the spark console.

src/components/ComputeModal.js
kyuksel: Yea I thought the same and tried `h(Fragment` but it looked like below

image

src/components/ComputeModal.js
kyuksel: Good question @jdcanas! I had assumed that output was because I didn't run anything on my cluster but I hadn't noticed the question there:

> Did you specify the correct logging directory?

@rtitle I see that the [Spark Console Design Doc](https://docs.google.com/document/d/1Zr6qzICStrrJwYkfoIE7fTf9Ziu3f6JmhdA9Ao3Y8Bo/edit#heading=h.b7doxousxooq) has the same screenshot. Do you have any thoughts on whether this requires an action?

rtitle: Yeah I think this is the expected screen if there isn't any spark history. If you run some Spark jobs it should be populated.

src/components/ComputeModal.js
kyuksel: I think @rtitle kept it because it sounded cool 😛 

Joke aside, I see a comment on it on the [related section](https://docs.google.com/document/d/1Zr6qzICStrrJwYkfoIE7fTf9Ziu3f6JmhdA9Ao3Y8Bo/edit#heading=h.93ivn47x7ed) of the design doc. Was there any follow-up on it?

src/components/ComputeModal.js
joyliu10: @kyuksel  is this green header from hadoop branding?

src/components/ComputeModal.js
rtitle: Haha.. yeah I think the UIs display that if there is no auth. I did not look into how to change/remove it. Perhaps we could make a follow-on ticket.

src/pages/workspaces/workspace/applications/AppLauncher.js
kyuksel: Yea, I was going to file a bug ticket on this. It happens for Terminal launching as well.

kyuksel:
@rtitle Thank you for the testing and the great comments!

> then the Cloud Environment modal appears and disappears

Addressed in [another comment](https://github.com/DataBiosphere/terra-ui/pull/2643#discussion_r721395828) 

> [UX comments here](https://github.com/DataBiosphere/terra-ui/pull/2643#issuecomment-932290157)

It'd be good to discuss with product and UX to determine how to proceed. I'll bring it up with them.

src/components/ComputeModal.js
kyuksel: Yea it's from the particular Spark UI. I was going to bring this up during the UX meeting scheduled for 4pm today.

src/components/ComputeModal.js
```diff
@@ -101,6 +93,34 @@ const DataprocDiskSelector = ({ value, onChange }) => {
...
+  )
+
+  return div({ style: { ...styles.whiteBoxContainer, marginBottom: '1rem', backgroundColor: colors.accent(0.1), boxShadow: Style.standardShadow } }, [
+    div({ style: { flex: '1', lineHeight: '1.5rem', minWidth: 0, display: 'flex' } }, [
+      div([
+        div({ style: { ...styles.headerText, marginTop: '0.5rem' } }, [interfaceDisplayName]),
+        p([synopsis]),
+        div({ style: { display: 'flex', marginTop: '1rem' } }, [
+          h(ButtonOutline, {
+            disabled: false,
```
petesantos: Is the `disabled: false` needed?

src/components/ComputeModal.js
```diff
@@ -247,9 +269,11 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
   })
 
-  const canUpdateNumberOfWorkers = () => {
-    return !currentRuntimeDetails || currentRuntimeDetails.status === 'Running'
-  }
+  const isRuntimeRunning = () => currentRuntimeDetails?.status === 'Running'
+
+  const canManageSparkConsole = () => !!sparkMode && isRuntimeRunning() && computeConfig.componentGatewayEnabled
+
+  const canUpdateNumberOfWorkers = () => !currentRuntimeDetails || isRuntimeRunning()
```
petesantos: Is there a reason these are functions, or could they be constants?

src/components/ComputeModal.js
petesantos: If you move this into the same div as the select, you should be able to align with flex.

src/components/ComputeModal.js
kyuksel: The outcome was to take this particular UI out and consider adding it back in with a differenting styling. The details are in the [respective ticket](https://broadworkbench.atlassian.net/browse/IA-2985).

src/pages/workspaces/workspace/applications/AppLauncher.js
kyuksel: Ticket: https://broadworkbench.atlassian.net/browse/IA-2985

src/components/ComputeModal.js
kyuksel: Because we want them to be re-evaluated when there are changes in the component's states. Does that make sense?

src/pages/workspaces/workspace/applications/AppLauncher.js
kyuksel: Did you mean for the case where we'd use `path: '/workspaces/:namespace/:name/applications/:application/:sparkInterface'` for both Spark UIs and applications? If so, the issue is, `sparkInterface` is expected to be input at all times, which doesn't work for non-Spark UI applications.

src/components/ComputeModal.js
kyuksel: I'd wanted to guard against a change in `ButtonOutline`'s default behaviour when `disabled` is `undefined` but I was probably being unnecessarily vigilant.

src/components/ComputeModal.js
kyuksel: Agreed. I started refactoring and realized it was going to add a lot of diff's to this PR, which is not desirable so I moved the changes I have to the branch [ky_ia_refactor_spark_mode_usages](https://github.com/DataBiosphere/terra-ui/tree/ky_ia_refactor_spark_mode_usages). Hoping to finish that up after merging this PR.

src/components/ComputeModal.js
kyuksel: Thanks for the great suggestion!

rtitle (approved):
Looks good! I launched a Hail image, ran a couple jobs, and opened each Spark UI and verified I could see the jobs. 

The flickering is still a bit annoying as a user, but glad it's tracked in a separate ticket.

:shipit:

src/components/ComputeModal.js
```diff
@@ -247,9 +290,13 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
-  const canUpdateNumberOfWorkers = () => {
-    return !currentRuntimeDetails || currentRuntimeDetails.status === 'Running'
-  }
+  const isRuntimeRunning = () => currentRuntimeDetails?.status === 'Running'
+
+  const shouldDisplaySparkConsoleLink = () => !!sparkMode && currentRuntimeDetails?.runtimeConfig?.componentGatewayEnabled
+
+  const canManageSparkConsole = () => shouldDisplaySparkConsoleLink() && isRuntimeRunning()
+
+  const canUpdateNumberOfWorkers = () => !currentRuntimeDetails || isRuntimeRunning()
```
petesantos, line 299: To clarify my earlier comment about these being constants:
`currentRuntimeDetails` is state
`isRuntimeRunning` and `shouldDisplaySparkConsoleLink` are derived from `currentRuntimeDetails`
`canManageSparkConsole` is derived from `isRuntimeRunning` and `shouldDisplaySparkConsoleLink` (see above)
`canUpdateNumberOfWorkers` is derived from `currentRuntimeDetails`

Being in the render method, `currentRuntimeDetails` should never change. If it does, then a state change should have occurred causing a re-render and new assignment to these cosntants.

If that is not reporting correctly there may be some mutation occurring in `currentRuntimeDetails`.

petesantos (approved):
A minor comment around constant vs function.

Overall, I think `ComputeModal` is getting too large and complex and it needs a refactor with functional decomposition. Perhaps we can put some time on the calendar to start chipping away at it a couple times a week ?

`GalaxyModal` may benefit from this as well.

src/components/ComputeModal.js
kyuksel: I see. Thanks for the explanation. I've put this on the TODO list for the follow-up ticket I'll create in a bit. There were a couple of other things I've punted on for various reasons as well.

kyuksel:
@petesantos 

> Overall, I think ComputeModal is getting too large and complex and it needs a refactor with functional decomposition. Perhaps we can put some time on the calendar to start chipping away at it a couple times a week ? GalaxyModal may benefit from this as well.

I'd appreciate that. Will follow up offline.

PR #2645, opened 2021-10-06
[DR-2160] Use Data Explorer Endpoint to populate data

petesantos:
This PR removes the test file that was being used by the data explorer to render the data tables.
The data is now retrieved from the server endpoint. The data is stored in an atom to prevent the need to re-fetch the data each time the user navigates around the catalog.

Tested in the browser locally.


src/pages/library/DataBrowser.js
```diff
@@ -119,25 +118,6 @@ const sidebarSections = [{
   labels: ['Homo sapiens', 'Mus musculus']
 }]
 
-const getRawList = async () => {
-  const list = await fetch('hca-sample.json').then(res => res.json())
-  return new Promise(resolve => setTimeout(resolve(list.data), 1000))
-}
-
-const extractTags = snapshot => {
```
petesantos, line 127: Moved to the utils file

src/pages/library/DataBrowser.js
```diff
@@ -273,40 +253,27 @@ const makeDataBrowserTableComponent = ({ sort, setSort, selectedData, toggleSele
 }
 
 const Browser = () => {
-  const [fullList, setFullList] = useState(() => StateHistory.get().catalogSnapshots)
   const [sort, setSort] = useState({ field: 'created', direction: 'desc' })
   const [showProjectFilters, setShowProjectFilters] = useState(false)
   const [selectedData, setSelectedData] = useState([])
   const [requestDatasetAccessList, setRequestDatasetAccessList] = useState()
-
-  Utils.useOnMount(() => {
```
petesantos, line 282: No longer needed, data is retrieved via the `useDataCatalog` hook

src/pages/library/DataBrowser.js
```diff
@@ -316,6 +283,6 @@ export const navPaths = [
     path: '/library/browser',
     component: Browser,
     title: 'Datasets',
-    public: true
```
petesantos, line 319: Had an issue with getting auth opts as a `true`

src/pages/library/DataBrowserDetails.js
```diff
@@ -175,13 +170,10 @@ const Sidebar = ({ snapshot, setShowRequestAccessModal }) => {
...
 
 const DataBrowserDetails = ({ id }) => {
-  const [snapshot, setSnapshot] = useState()
   const [showRequestAccessModal, setShowRequestAccessModal] = useState()
-
-  Utils.useOnMount(() => {
-    const loadData = async () => setSnapshot(normalizeSnapshot(await getSnapshot(id)))
-    loadData()
-  })
+  const { dataCatalog } = useDataCatalog()
```
petesantos, line 174: The hook will retrieve the data if it has not been loaded

src/pages/library/common.js
```diff
@@ -3,12 +3,18 @@ import { Fragment, useState } from 'react'
 import { div, h, label } from 'react-hyperscript-helpers'
 import Collapse from 'src/components/Collapse'
 import { Clickable, IdContainer, Link, Select } from 'src/components/common'
-import { centeredSpinner } from 'src/components/icons'
 import { DelayedSearchInput } from 'src/components/input'
 import colors from 'src/libs/colors'
-import * as Utils from 'src/libs/utils'
 
 
+export const commonStyles = {
```
petesantos, line 10: moved from the snapshots file

src/pages/library/dataBrowser-utils.js
```diff
@@ -1,16 +1,12 @@
...
-export const snapshotStyles = {
-  access: {
-    open: colors.success(1.5),
-    controlled: colors.accent(),
-    pending: '#F7981C'
-  }
-}
-
-export const normalizeSnapshot = snapshot => {
+const normalizeSnapshot = snapshot => {
```
petesantos, line 9: `normalizeSnapshot` and `extractTags` were moved from the `snapshots`

src/libs/ajax.js
```diff
@@ -919,6 +919,11 @@ const Workspaces = signal => ({
 
 
 const DataRepo = signal => ({
+  getMetadata: async () => {
+    const res = await fetchDataRepo(`repository/v1/search/metadata`, _.merge(authOpts(), { signal }))
```
dmohs: I think our convention is single quotes unless interpolating. I use the opposite convention in my Ruby code, so I'm open to always using backticks for strings.
```suggestion
    const res = await fetchDataRepo('repository/v1/search/metadata', _.merge(authOpts(), { signal }))
```

src/libs/ajax.js
```diff
@@ -919,6 +919,11 @@ const Workspaces = signal => ({
...
+    return res.json()
```
dmohs: Option per Edmund:
```suggestion
  getMetadata: () => {
    return fetchDataRepo(`repository/v1/search/metadata`, _.merge(authOpts(), { signal })).then(res => res.json())
```

src/pages/library/dataBrowser-utils.js
dmohs: Love these little notes.

dmohs (approved):
Nicely sized PR.

src/pages/library/DataBrowser.js
lu-c: My one concern about moving this to a utils file is that I don't think it's very reuseable - the format for the tags is specific to just the data browser search page view. The details view, and (eventually) the preview view shouldnt need these tags at all. These are only used for the search filters page, and I would advise us to keep one time use functions in separated files

src/libs/ajax.js
zarsky-broad: This entire file uses the `.json()` on the next line convention. We shouldn't introduce exceptions to this unless we're going to change them all.

src/pages/library/DataBrowser.js
petesantos: I agree. My rationale for putting them in a `dataBrowser-utils` file is that they are no longer exported, instead they are used locally in that file when loading the data. The definitions are close to the code that uses them and is exported - the react hook.

As we remove the need for these functions we can update the hook in the utils file and remove the functions in that file without having to touch the other files that were using those functions

src/pages/library/DataBrowserDetails.js
lu-c: Will `useDataCatalog` update the data if a new user logs in? The back end returns only the datasets you have permission to see, which might be different per user. I'm not sure if this is something we have to worry about right now.

(I don't think this is a problem we have to address in this pr, it could prob be its own ticket)

lu-c (approved):
Left some comments with questions, but there are no code changes i'd like to request.... so throwing an approval on the pile! (heave ho!)

src/pages/library/DataBrowser.js
lu-c: I see.... It makes sense if we're going to be using the extractTags inside the getDataCatalog function in a different file. How would you feel about extracting the tags in DataBrowser.js, after getting the contents of getDataCatalog? It would cost us another iteration of the array, but it would keep the code separate a little

petesantos: When I first looked at this comment I was thinking this was part of the snapshots file - sorry for the confusion.

I still prefer this in utils, where we are loading the data. If we only run `extractTags` in `DataBrowser`, and someone were to directly navigate to a details page (e.g via bookmark or shared link) then we would need to run this step and update the data once they hit the browser page. For me this adds complexity to the flow of loading the data: depending on when the data was loaded (on the details vs the browser) the data in the store could have a different shape and content.

src/pages/library/DataBrowserDetails.js
pshapiro4broad: Right now we don't have a no-auth endpoint to use to get this data so there's nothing we can do about this now. I think anonymous access will require many changes so I don't know it's useful to make issues for each case now, rather than creating issues when the team takes on the work.

PR #2648, opened 2021-10-07
[AS-975] 'select all in filter' for data tables

davidangb:
We have UX approval for this! All set as-is.

PR site https://pr-8-dot-bvdp-saturn-dev.appspot.com/ shows this in action!

-----

This PR changes the "select all" feature of a data table to follow the text filter currently in use. When there is no text filter, the "select all" says "All (nnn)" and selects all rows of that type:
Screenshot-nofilter

When there is a text filter, the menu item says "Filtered (nnn)" and selects only those rows that match the user's current filter:
Screenshot-filtered

Additionally, I've changed the backend API call powering this menu item away from the to-be-deprecated list-all-entities API, and instead it now uses the paginated-entities API … but specifying a page size that still returns all entities. Using the paginated API is necessary to support the text filter.

AS-956 tracks further optimizations for this feature, such as not needing to make an API call at all.

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -120,7 +120,7 @@ const DataTable = props => {
     Utils.withBusyState(setLoading),
     withErrorReporting('Error loading entities')
   )(async () => {
-    const params = _.merge({pageSize: filteredCount}, activeTextFilter ? {filterTerms: activeTextFilter} : {})
+    const params = _.merge({ pageSize: filteredCount }, activeTextFilter ? { filterTerms: activeTextFilter } : {})
     const queryResults = await Ajax(signal).Workspaces.workspace(namespace, name).paginatedEntitiesOfType(entityType, params)
     setSelected(entityMap(queryResults.results))
   })
@@ -217,7 +217,7 @@ const DataTable = props => {
                         content: h(Fragment, [
                           h(MenuButton, { onClick: selectPage }, ['Page']),
                           h(MenuButton, { onClick: selectAll },
-                            ((totalRowCount == filteredCount) ? [`All (${filteredCount})`] : [`Filtered (${filteredCount})`])),
+                            ((totalRowCount === filteredCount) ? [`All (${filteredCount})`] : [`Filtered (${filteredCount})`])),
                           h(MenuButton, { onClick: selectNone }, ['None'])
                         ]),
                         side: 'bottom'
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -216,8 +216,8 @@ const DataTable = props => {
                         closeOnClick: true,
                         content: h(Fragment, [
                           h(MenuButton, { onClick: selectPage }, ['Page']),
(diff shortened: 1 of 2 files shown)


davidangb:
huh, I sure did not mean to close this. reopening.

src/components/data/DataTable.js
```diff
@@ -120,8 +120,9 @@ const DataTable = props => {
     Utils.withBusyState(setLoading),
     withErrorReporting('Error loading entities')
   )(async () => {
-    const results = await Ajax(signal).Workspaces.workspace(namespace, name).entitiesOfType(entityType)
-    setSelected(entityMap(results))
+    const params = _.merge({ pageSize: filteredCount }, activeTextFilter ? { filterTerms: activeTextFilter } : {})
```
davidangb: is there a better way to do this? I never want to send an empty filterTerms query parameter, i.e. `?filterTerms=&pageSize=10` should never happen.

petesantos: When testing the page I am able to send a request with a pageSize of `0`, which results in this error:

image

I believe this will alleviate the error by always removing empty terms:
```suggestion
    const params = _.pickBy(_.identity(_.trim), { pageSize: filteredCount, filterTerms: activeTextFilter })
```

There is an additional bug on the page that is outside of the scope of this change. If you type some spaces in the search field and hit enter, it will result in a server error. We need to trim the input in the search box, but it appears there is a bug in the underlying component that is preventing this.

I created a bug to track this [here ](https://broadworkbench.atlassian.net/jira/software/c/projects/TUB/boards/110?modal=detail&selectedIssue=TUB-18)as I am not sure where else to track core component bugs.

petesantos (commented):
Added a suggestion for handling params. I will wait to thumb after UX approves.

src/components/data/DataTable.js
davidangb: * good find on the behavior with zero results! I've removed the menu item in that case, as it doesn't make sense.
* thank you for the suggestion of `_.pickBy`! In your suggested code, does it really need the extra `_.identity` step? I've added a commit without identity, how does that look?
* I've added a few tweaks to dealing with whitespace-only in the text filter. I doubt this is a comprehensive fix but it seems to help; I'll leave any further fixes as out of scope.

Back to you for re-review, thank you!

src/components/data/DataTable.js
```diff
@@ -178,7 +179,7 @@ const DataTable = props => {
             'aria-label': 'Search',
             placeholder: 'Search',
             onChange: v => {
-              setActiveTextFilter(v)
+              setActiveTextFilter(v.toString().trim())
```
davidangb, line 182: the `toString()` shouldn't be necessary here, but felt like safe defensive coding - we've seen other instances where we expected a string but got something else and caused page-killing errors

petesantos (approved):
Looks good - tested manually

ansingh7115 (approved):

PR #2650, opened 2021-10-12
[AS-951] Delete Entity Column 

ansingh7115:
New menu on each column
Screen Shot 2021-10-13 at 12 32 03 AM

Entity name column not deletable
Screen Shot 2021-10-13 at 12 32 11 AM

Screen Shot 2021-10-13 at 12 32 25 AM

Delete modal
Screen Shot 2021-10-13 at 12 39 28 AM
Screen Shot 2021-10-13 at 12 40 30 AM

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -3,7 +3,7 @@ import { Fragment, useEffect, useRef, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import { Checkbox, Clickable, fixedSpinnerOverlay, Link } from 'src/components/common'
-import { EditDataLink, EntityEditor, EntityRenamer, renderDataCell } from 'src/components/data/data-utils'
+import { DeleteEntityColumnModal, EditDataLink, EntityEditor, EntityRenamer, HeaderOptions, renderDataCell } from 'src/components/data/data-utils'
 import { icon } from 'src/components/icons'
 import { ConfirmedSearchInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
@@ -54,7 +54,8 @@ const DataTable = props => {
     childrenBefore,
     editable,
     persist, refreshKey, firstRender,
-    snapshotName
+    snapshotName,
+    loadMetadata
   } = props
 
   const persistenceId = `${namespace}/${name}/${entityType}`
@@ -95,6 +96,7 @@ const DataTable = props => {
 
   const [renamingEntity, setRenamingEntity] = useState()
   const [updatingEntity, setUpdatingEntity] = useState()
+  const [deletingColumn, setDeletingColumn] = useState()
 
   const table = useRef()
   const signal = Utils.useCancellation()
@@ -245,9 +247,8 @@ const DataTable = props => {
                       setColumnWidths(_.set('name', nameWidth + delta))
                     }
                   }, [
-                    h(Sortable, { sort, field: 'name', onSort: setSort }, [
-                      h(HeaderCell, [entityMetadata[entityType].idName])
-                    ])
+                    h(HeaderOptions, { field: 'name', onSort: setSort, isEntityName: true, },
+                      [ h(HeaderCell, [entityMetadata[entityType].idName]) ])
                   ]),
                   cellRenderer: ({ rowIndex }) => {
                     const { name: entityName } = entities[rowIndex]
@@ -261,16 +262,18 @@ const DataTable = props => {
                     ])
                   }
                 },
-                ..._.map(({ name }) => {
-                  const thisWidth = columnWidths[name] || 300
-                  const [, columnNamespace, columnName] = /(.+:)?(.+)/.exec(name)
+                ..._.map(({ name: attributeName }) => {
+                  const thisWidth = columnWidths[attributeName] || 300
+                  const [, columnNamespace, columnName] = /(.+:)?(.+)/.exec(attributeName)
                   return {
-                    field: name,
+                    field: attributeName,
                     width: thisWidth,
                     headerRenderer: () => h(Resizable, {
-                      width: thisWidth, onWidthChange: delta => setColumnWidths(_.set(name, thisWidth + delta))
+                      width: thisWidth, onWidthChange: delta => setColumnWidths(_.set(attributeName, thisWidth + delta))
                     }, [
-                      h(Sortable, { sort, field: name, onSort: setSort }, [
+                      h(HeaderOptions, { field: attributeName, onSort: setSort, isEntityName: false,
+                        beginDelete: () => setDeletingColumn({ entityType: entityType, attributeName: attributeName })
+                      }, [
                         h(HeaderCell, [
                           !!columnNamespace && span({ style: { fontStyle: 'italic', color: colors.dark(0.75), paddingRight: '0.2rem' } },
                             columnNamespace)
@@ -279,7 +282,7 @@ const DataTable = props => {
                       ])
                     ]),
                     cellRenderer: ({ rowIndex }) => {
-                      const { attributes: { [name]: dataInfo }, name: entityName } = entities[rowIndex]
+                      const { attributes: { [attributeName]: dataInfo }, name: entityName } = entities[rowIndex]
                       const dataCell = renderDataCell(Utils.entityAttributeText(dataInfo), googleProject)
                       return h(Fragment, [
                         (!!dataInfo && _.isArray(dataInfo.items)) ?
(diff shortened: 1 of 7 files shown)


src/components/data/DataTable.js
```diff
@@ -259,16 +260,19 @@ const DataTable = props => {
                     ])
                   }
                 },
-                ..._.map(({ name }) => {
-                  const thisWidth = columnWidths[name] || 300
-                  const [, columnNamespace, columnName] = /(.+:)?(.+)/.exec(name)
+                ..._.map(({ name: attributeName }) => {
```
dmohs, line 273: Nice rename here.

src/components/data/data-utils.js
```diff
@@ -695,6 +696,82 @@ export const ModalToolButton = ({ icon, text, disabled, ...props }) => {
...
+    }, 'Delete column')
+  }, [
+    div(['Are you sure you want to permanently delete the column ',
+      span({ style: { fontWeight: 600, wordBreak: 'break-word' } }, attributeName),
+      '?']),
+    div({
+      style: {
+        fontWeight: 500,
+        marginTop: '1rem'
+      }
```
dmohs: We like to conserve vertical space when we can to give the reader more context on a single screen.
```suggestion
      style: { fontWeight: 500, marginTop: '1rem' }
```

src/components/data/data-utils.js
```diff
@@ -695,6 +696,82 @@ export const ModalToolButton = ({ icon, text, disabled, ...props }) => {
...
+  }
+
+  return h(Modal, {
+    title: 'Delete Column',
+    onDismiss,
+    okButton: h(ButtonPrimary, {
+      onClick: deleteColumn,
+      tooltip: Utils.cond(
+        [_.toLower(deleteConfirmation) !== 'delete column', () => 'You must type the confirmation message'],
+        () => 'Delete Column')
```
dmohs: I think this might make more sense as a ternary:
```suggestion
      tooltip: _.toLower(deleteConfirmation) !== 'delete column' ? 'You must type the confirmation message' : undefined
```
In the fall-through case, I would hope you can leave it as undefined rather than duplicating the button text. Maybe the behavior bothered you, and if so, please ignore me.

src/components/data/data-utils.js
```diff
@@ -695,6 +696,82 @@ export const ModalToolButton = ({ icon, text, disabled, ...props }) => {
...
+      span({ style: { fontWeight: 600, wordBreak: 'break-word' } }, attributeName),
+      '?']),
+    div({
+      style: {
+        fontWeight: 500,
+        marginTop: '1rem'
+      }
+    }, 'This cannot be undone.'),
+    div({ style: { marginTop: '1rem' } }, [
+      label({ htmlFor: 'delete-column-confirmation' }, ['Please type \'Delete Column\' to continue:']),
```
dmohs: I think double quotes are more typical in instructions like this:
```suggestion
      label({ htmlFor: 'delete-column-confirmation' }, ['Please type "Delete Column" to continue:']),
```

src/components/data/data-utils.js
```diff
@@ -695,6 +696,82 @@ export const ModalToolButton = ({ icon, text, disabled, ...props }) => {
...
+    content: h(Fragment, [
+      h(MenuButton, { onClick: () => onSort({ field, direction: 'asc' }) }, ['Sort Ascending']),
+      h(MenuButton, { onClick: () => onSort({ field, direction: 'desc' }) }, ['Sort Descending']),
+      !isEntityName && h(MenuButton, { onClick: beginDelete }, ['Delete column'])
+    ])
+  }, [
+    h(Link, { 'aria-label': 'Workflow menu', onClick: e => e.stopPropagation() }, [
+      icon('cardMenuIcon', {
+        size: 16
+      })
```
dmohs: Similarly:
```suggestion
      icon('cardMenuIcon', { size: 16 })
```

src/components/data/data-utils.js
```diff
@@ -695,6 +696,82 @@ export const ModalToolButton = ({ icon, text, disabled, ...props }) => {
...
+    ])
+  ])
+
+  return h(IdContainer, [id => div({
+    style: { flex: 1, display: 'flex', alignItems: 'center', cursor: 'pointer', width: '100%', height: '100%' },
+    'aria-describedby': id
+  }, [
+    children,
+    div({ style: { marginRight: '1rem', marginLeft: 'auto' } }, [columnMenu])
```
dmohs: I'm surprised you need the marginLeft here. I think 'auto' is the default most of the time.

dmohs (approved):

src/components/data/data-utils.js
ansingh7115: Unfortunately, when I remove it, it does this:
Screen Shot 2021-10-13 at 12 30 42 AM

src/components/data/DataTable.js
```diff
@@ -349,6 +353,15 @@ const DataTable = props => {
       },
       onDismiss: () => setUpdatingEntity(undefined)
     }),
+    !!deletingColumn && h(DeleteEntityColumnModal, {
+      workspaceId: { namespace, name },
+      column: deletingColumn,
+      onSuccess: () => {
+        setDeletingColumn(undefined)
+        loadMetadata()
```
davidangb: the call to `loadMetadata()` here, because it occurs directly after deleting the column, is extremely unlikely to be cached in the back end. For workspaces with lots of rows/columns, this could cause a significant delay. Since the operation of deleting a column is targeted and well-known, would it be possible to inspect the current `entityMetadata` value, delete the column from that value in-memory and set it back into state, without ever needing to make this expensive ajax call? This feels important to try to optimize.

src/components/data/data-utils.js
```diff
@@ -695,6 +696,76 @@ export const ModalToolButton = ({ icon, text, disabled, ...props }) => {
...
+    }, 'Delete column')
+  }, [
+    div(['Are you sure you want to permanently delete the column ',
+      span({ style: { fontWeight: 600, wordBreak: 'break-word' } }, attributeName),
+      '?']),
+    div({
+      style: { fontWeight: 500, marginTop: '1rem' }
+    }, 'This cannot be undone.'),
+    div({ style: { marginTop: '1rem' } }, [
+      label({ htmlFor: 'delete-column-confirmation' }, ['Please type "Delete Column" to continue:']),
```
davidangb, line 733: my personal $.02 is that we should ask the user to type the column name, instead of a generic "Delete Column". Since the column name is included in the message above, they can always copy/paste it. If our UX team feels strongly about using the generic "Delete Column" then so be it, I won't push this

src/components/data/DataTable.js
```diff
@@ -243,9 +245,8 @@ const DataTable = props => {
                       setColumnWidths(_.set('name', nameWidth + delta))
                     }
                   }, [
-                    h(Sortable, { sort, field: 'name', onSort: setSort }, [
-                      h(HeaderCell, [entityMetadata[entityType].idName])
-                    ])
+                    h(HeaderOptions, { field: 'name', onSort: setSort, isEntityName: true },
+                      [h(HeaderCell, [entityMetadata[entityType].idName])])
```
davidangb, line 259: Two comments on this menu:
1. We've lost the up/down arrow that indicates which column is currently the active sort order - that icon was in `Sortable` but is not in `HeaderOptions`. I think we should have that back.


| old | new |
| --- | --- |
| sort-old | sort-new
 |




2. At least in the case where a column is not the active sort order, we've lost some characters of the column's header. Is it possible to move the menu a bit to the right to gain back any characters we can? Those are precious and valuable:


| old | new |
| --- | --- |
| colname-old | colname-new |

davidangb (commented):
A couple comments about UX, and one about ajax efficiency inline. Code looks good!

zarsky-broad:
Did we consider only showing the menu on hover over the column? This seems like a lot of visual noise. Just asking.

src/components/data/data-utils.js
ansingh7115: I modeled this after the delete workspace modal, which asks the user to type "Delete Workspace"

davidangb: oh, I know … I don't like that one either :)

ansingh7115:
@zarsky-broad 
I think there's some merit in keeping the dropdown icon present because removing it would harm discoverability and create some amount of noise when the user's cursor passes over the headers. I'd be open to trying this out, but in a separate ticket so we can get this merged and put the feature in the hands of the user quicker.

dmohs (approved):
Approved via in-person review.

ansingh7115:
Tests for `run-workflow-on-snapshot`, `run-workflow` and `run-notebook` are failing for reasons unrelated to this PR. It seems like they're failing on all PRs. The reason seems to be related to the workspace bucket.

kyuksel:
> @zarsky-broad
I think there's some merit in keeping the dropdown icon present because removing it would harm discoverability and create some amount of noise when the user's cursor passes over the headers. I'd be open to trying this out, but in a separate ticket so we can get this merged and put the feature in the hands of the user quicker.

@ansingh7115 Isaac is on vacation but your comment sounds sensible :) Feel free to merge it when integration tests pass.

davidangb (approved):

PR #2612, opened 2021-09-07
[CA-910] PPW changes

marctalbott:
<!--
Hello, friend!
Remember to mention what sort of testing/verification you performed in the course of building this PR, i.e. checking functionality in the browser locally.
Thanks!
--!>

These changes have been reviewed as they were merged into this feature branch. Some of the changes include (nonexhaustive sorry):

1. Using v2 billing endpoints in Rawls
2. Asynchronous billing account updates
3. Various PPW bugfixes

PR #2652, opened 2021-10-12
[CA-1541] Address PPW Feature Branch Feedback

ehigham:
RR: https://broadworkbench.atlassian.net/browse/CA-1541
In this PR I'm addressing some of @zarsky-broad's suggestions and feedback in #2612, namely merging the Project per Workspace (PPW) branch, `v2_workspaces` into `dev`.
Isaac and I spoke at length about #2612 and a few suggestions for tidy-ups were also raised in `ajax.js`. I've tried to address these as well.

### DIFF
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -50,12 +50,10 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
 
 
   // Helpers
-  const getRequiredGroups = () => {
-    return _.uniq([
-      ...(cloneWorkspace ? _.map('membersGroupName', cloneWorkspace.workspace.authorizationDomain) : []),
-      ...(requiredAuthDomain ? [requiredAuthDomain] : [])
-    ])
-  }
+  const getRequiredGroups = () => _.uniq([
+    ...(cloneWorkspace ? _.map('membersGroupName', cloneWorkspace.workspace.authorizationDomain) : []),
+    ...(requiredAuthDomain ? [requiredAuthDomain] : [])
+  ])
 
   const create = async () => {
     try {
@@ -161,18 +159,18 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
           value: namespace,
           onChange: ({ value }) => setNamespace(value),
           styles: { option: provided => ({ ...provided, padding: 0 }) },
-          options: _.uniq(_.map(({ projectName, invalidBillingAccount }) => ({
-            label: h(Fragment, [
-              invalidBillingAccount ?
+          options: _.map(({ projectName, invalidBillingAccount }) => ({
+            label: ((label = div({ style: { padding: 10 } }, [projectName])) => {
+              return invalidBillingAccount ?
                 h(TooltipTrigger, {
                   content: 'Workspaces may only be created in billing projects that have a Google billing account accessible in Terra',
                   side: 'left'
-                }, [div({ style: { padding: 10 } }, [projectName])]) :
-                div({ style: { padding: 10 } }, [projectName])
-            ]),
+                }, [label]) :
+                label
+            })(),
             value: projectName,
             isDisabled: invalidBillingAccount
-          }), billingProjects)).sort()
+          }), _.sortBy('projectName', _.uniq(billingProjects)))
         })
       ])]),
       h(IdContainer, [id => h(Fragment, [
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -160,14 +160,11 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
           onChange: ({ value }) => setNamespace(value),
           styles: { option: provided => ({ ...provided, padding: 0 }) },
           options: _.map(({ projectName, invalidBillingAccount }) => ({
-            label: ((label = div({ style: { padding: 10 } }, [projectName])) => {
-              return invalidBillingAccount ?
-                h(TooltipTrigger, {
-                  content: 'Workspaces may only be created in billing projects that have a Google billing account accessible in Terra',
-                  side: 'left'
-                }, [label]) :
-                label
-            })(),
+            label: h(TooltipTrigger, {
+              content: invalidBillingAccount && 'Workspaces may only be created in billing projects that have a Google billing account accessible in Terra',
+              side: 'left'
+            }, [div({ style: { padding: 10 } }, [projectName])]
+            ),
             value: projectName,
             isDisabled: invalidBillingAccount
           }), _.sortBy('projectName', _.uniq(billingProjects)))
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -442,15 +442,24 @@ const Billing = signal => ({
     return res.json()
   },
 
+  getProject: async projectName => {
+    const route = `billing/v2/${projectName}`
+    const res = await fetchRawls(route, _.merge(authOpts(), { signal, method: 'GET' }))
+    return res.json()
+  },
+
   listAccounts: async () => {
     const res = await fetchRawls('user/billingAccounts', _.merge(authOpts(), { signal }))
     return res.json()
   },
 
   createProject: async (projectName, billingAccount) => {
-    const res = await fetchRawls('billing/v2',
-      _.mergeAll([authOpts(), jsonBody({ projectName, billingAccount }), { signal, method: 'POST' }]))
-    return res
+    const res = await fetchRawls('billing/v2', _.mergeAll([
+      authOpts(),
+      jsonBody({ projectName, billingAccount }),
+      { signal, method: 'POST' }
+    ]))
+    return res.json()
   },
 
   changeBillingAccount: async ({ billingProjectName, newBillingAccountName }) => {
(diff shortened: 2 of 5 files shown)


src/pages/billing/Project.js
```diff
@@ -147,41 +147,39 @@ const BillingAccountSummaryPanel = (() => {
...
 
 const groupByBillingAccountStatus = (billingProject, workspaces) => {
-  const group = workspace => Utils.cond(
-    [billingProject.billingAccount === workspace.billingAccount, () => 'done'],
-    ['billingAccountErrorMessage' in workspace, () => 'error'],
-    [Utils.DEFAULT, () => 'updating']
-  )
+  const group = workspace => billingProject.billingAccount === workspace.billingAccount ? 'done' :
+    !!workspace.billingAccountErrorMessage ? 'error' :
+      'updating'
```
ehigham: I went with ternary instead of `Utils.cond` because this is more idomatic for the language. I believe that newcomers who don't know Clojure would be far more familiar with C-like ternaries.

src/pages/billing/Project.js
```diff
@@ -147,41 +147,39 @@ const BillingAccountSummaryPanel = (() => {
...
 
+  // return `Set`s of workspaces mapped by their status to reduce the complexity of looking up a
+  // workspace's status.
```
ehigham: This comment is terrible. What I'm trying to get at is I'm using this data structure for two things - to test if we need to draw icons to represent the status of billing account changes and if we do, what icon to draw for each workspace. I'm using sets to reduce the complexity of seaching for the status of any workspace from `O(N * W)` to `O(N * 1)`, where `N` is the number of statuses a billing account change could have and `W` is the number of workspaces in a billing project. Note I need to perform this `W` times for each billing project and `W` can be very large in the case of GP's billing projects. Thus reduces the complexity by an order of magnitute.

ehigham: (Suggestions welcome)

src/libs/ajax.js
```diff
@@ -471,43 +480,34 @@ const Billing = signal => ({
...
 
-  billingProject: async projectName => {
-    const route = `billing/v2/${projectName}`
-    const res = await fetchRawls(route, _.merge(authOpts(), { signal, method: 'GET' }))
+  listProjectUsers: async projectName => {
+    const res = await fetchRawls(`billing/v2/${projectName}/members`, _.merge(authOpts(), { signal }))
     return res.json()
   },
 
-  project: projectName => {
```
s-rubenstein, line 480: Why are we getting rid of the project wrapper here? That seemed more representative of the codebase's general practice. (Similar to analysis and notebooks under storage)

src/pages/billing/Project.js
```diff
@@ -60,7 +60,7 @@ const WorkspaceCardHeaders = Utils.memoWithName('WorkspaceCardHeaders', ({ needs
   ])
 })
 
-const ExpandedInfoRow = Utils.memoWithName('ExpandedInfoRow', ({ title, details, errorMessage }) => {
```
s-rubenstein, line 63: Can we confirm that without this it shows up with the right name in the React dev tools?

src/pages/billing/Project.js
s-rubenstein: Could you say that last bit that you had there? 
```
I'm using sets to reduce the complexity of seaching for the status of any workspace from O(N * W) to O(N * 1), where N is the number of statuses a billing account change could have and W is the number of workspaces in a billing project. Note I need to perform this W times for each billing project and W can be very large in the case of GP's billing projects. Thus reduces the complexity by an order of magnitute.
```

src/libs/ajax.js
ehigham: I did do that originally (renaming it to `projectUsers`) but then thought the wrapper served very little benefit. Merging with the `Billing` object and renaming seemed cleaner.

src/pages/billing/Project.js
ehigham: Sure!

ehigham: Confirmed - it's called the right thing.

src/pages/billing/Project.js
ehigham: Thanks!

marctalbott (approved):

src/components/NewWorkspaceModal.js
```diff
@@ -161,18 +159,18 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
...
           onChange: ({ value }) => setNamespace(value),
           styles: { option: provided => ({ ...provided, padding: 0 }) },
-          options: _.uniq(_.map(({ projectName, invalidBillingAccount }) => ({
-            label: h(Fragment, [
-              invalidBillingAccount ?
+          options: _.map(({ projectName, invalidBillingAccount }) => ({
+            label: ((label = div({ style: { padding: 10 } }, [projectName])) => {
+              return invalidBillingAccount ?
                 h(TooltipTrigger, {
                   content: 'Workspaces may only be created in billing projects that have a Google billing account accessible in Terra',
```
zarsky-broad: You can just make content `invalidBillingAccount && 'Workspaces...'` here, and remove the ternary altogether; a tooltip won't be rendered if content is falsy. Why did you make label an argument to this function if it's never called with arguments? I continue to not like the self-invoking function pattern.

src/libs/ajax.js
```diff
@@ -442,15 +442,24 @@ const Billing = signal => ({
...
 
   createProject: async (projectName, billingAccount) => {
-    const res = await fetchRawls('billing/v2',
-      _.mergeAll([authOpts(), jsonBody({ projectName, billingAccount }), { signal, method: 'POST' }]))
-    return res
+    const res = await fetchRawls('billing/v2', _.mergeAll([
+      authOpts(),
+      jsonBody({ projectName, billingAccount }),
+      { signal, method: 'POST' }
+    ]))
```
zarsky-broad: I don't love breaking this into 3 lines; I cold see this all on one line, but I think if we're adding linebreaks I might go for 
```js
fetchRawls(
  'path',
  _.mergeAll([])
)
```

src/libs/ajax.js
zarsky-broad: I'm with @s-rubenstein on this one, I find the consistency with other places to be extremely valuable. It also cuts down on the number of positional arguments to functions.

src/libs/ajax.js
```diff
@@ -471,43 +480,34 @@ const Billing = signal => ({
...
+  },
 
-      changeUserRoles: async (email, oldRoles, newRoles) => {
-        if (!_.isEqual(oldRoles, newRoles)) {
-          await Promise.all(_.map(role => addRole(role, email), _.difference(newRoles, oldRoles)))
-          return Promise.all(_.map(role => removeRole(role, email), _.difference(oldRoles, newRoles)))
-        }
-      }
+  changeUserRoles: async (projectName, email, oldRoles, newRoles) => {
+    const billing = Billing(signal)
```
zarsky-broad: This is a potentially destructive operation, so it shouldn't take `signal`, it should complete even if a modal or something is closed.

src/pages/billing/Project.js
```diff
@@ -29,14 +29,14 @@ const workspaceLastModifiedWidth = 150
 const workspaceExpandIconSize = 20
 const billingAccountIconSize = 16
 
-const BillingAccountIcon = {
+const billingAccountIcon = {
```
zarsky-broad: Would you consider `billingAccountIcons`, since it's effectively a dict?

src/pages/billing/Project.js
zarsky-broad: I hear you, but we really try to avoid nested ternaries. I'd prefer to see it as a cond because that's our standard in this codebase, and we use it everywhere else. I want to keep things consistent.

src/pages/billing/Project.js
```diff
@@ -320,15 +322,15 @@ const ProjectDetail = ({ billingProject, reloadBillingProject, billingAccounts,
...
+  )(() => Ajax(signal).Billing.listProjectUsers(billingProject.projectName)
     .then(collectUserRoles)
     .then(setProjectUsers)
   )
 
   const removeUserFromBillingProject = _.flow(
     reportErrorAndRethrow('Error removing member from billing project'),
     Utils.withBusyState(setUpdating)
-  )(Ajax().Billing.project(billingProject.projectName).removeUser)
+  )(_.partial(Ajax().Billing.removeProjectUser, [billingProject.projectName]))
```
zarsky-broad, line 335: So these are our first instances of `_.partial` in the codebase; we normally declare any function likely to be used like this with `_.curry` instead. That said, this feels to me like a reason to revert the structure in the ajax file to how it was before, which was done with the intention of how it was going to be used in mind.

zarsky-broad (commented):
I'm off on vacation after today, so I'm gonna nominate @petesantos to follow up. Here are my thoughts, though; thank you for sticking with this!

src/pages/billing/Project.js
ehigham: Cond is noisy. Cond does not preserve the evaluation order of ternaries. Ternaries are what programmers familiar with C syntax are familiar with. We should be consistent with the idoms of the language and stop pretending this is clojure.

src/libs/ajax.js
ehigham: I disagree strongly. This way keeps things simple and follows the underlying API whereas the other feels consistency for consistency's sake. I don't really know/understand what this is done for either.

src/components/NewWorkspaceModal.js
ehigham: I didn't like `div({ style: { padding: 10 } }, [projectName])` appearing twice. What I really wanted to do was use a let expression to generate a binding right next to where it's used. I didn't want to make a whole bunch of changes here like restructuring the code and adding a couple of `const`s a thousand lines above this as I prefer to keep helper definitions close to uses. Sadly, JavaScript lacks `let` expressions but it does have lambdas and `let`s are syntactic sugar for lambdas. Thus, I can encode
```
let x = <init> in
   f(x, x)
```
simply by
```javascript
((x = <init>) => f(x, x))()
```
I'm sorry you don't like self-invoking functions - they're such a useful tool.

ehigham: Thanks for the tip about tooltips not rendering if the content is falsy. I didn't realise that was a thing.

src/libs/ajax.js
ehigham: I don't love lines being 150 characters long. I don't particularly like lines exceeding 100 columns either. I'm just going to revert this.

src/pages/billing/Project.js
ehigham: If using the `project` wrapper was a means of achieving partial application then that's another reason against it. As a side note, I don't like this practice of defining curried functions. If it's a normal function then I can choose what to do with it. If I want to, I can curry it or partially apply arguments or just call it without a ton of parentheses. Let the caller decide.

src/libs/ajax.js
ehigham: Can you explain what this means? Why is this destructive? I'm not even sure what the `signal` is for...

src/pages/billing/Project.js
s-rubenstein: So, having now looked at the docs, I understand what this is doing but I found it less readable, personally, then the previous code. Especially in the case of the props below. However, in this case I _personally_ don't have strong opinions on it, but just wanted to throw my two cents in.

src/pages/billing/Project.js
s-rubenstein: General question: why should we be preferring C-style syntax over clojure-style syntax?

I am happy with the argument in this case that it preserves evaluation order, that is compelling to me personally, but I don't necessarily find C that readable and the nice thing about the clojure syntax is it makes it a little bit more clear to me what the possible outcomes are, whereas if we got more nested ternaries in here (two is okay, but I have seen it get much longer than that), I personally start to have trouble making that out. It takes what was a relatively easy task and makes it difficult.

src/libs/ajax.js
s-rubenstein: The benefit that it provides, in my mind, is similar to object layering. By abstracting out `projectUsers` (and I like that name a lot), we gain the ability to have a structure that groups more in a way that makes it easy to find what neighboring functions are, and keeps associated endpoints grouped. (If I see a `.projectUsers(project).add()` call, I can fairly easily follow `.projectUsers` to find all associated endpoints.

I would personally prefer those child objects pulled out into their own objects, so we get a cleaner view in our editors, which I think would more easily show the benefits here.

I think it does follow the underlying REST structure a bit more (where they are all a page) but it doesn't necessarily follow the route definitions in the service as cleanly. In Rawls (or Sam) these are grouped effectively by arg, so the original code here preserves that, IMO.

s-rubenstein: Thinking more about it, the difference between `projectUsers` and `notebooks`/`analysis` is that notebooks and analysis provide useful wrappers around some business logic (that they are making storage api calls and doing a bit of work around abstracting that), while `projectUsers` provides an abstraction that doesn't help much and adds potential for misunderstanding.

`workspaces` is a place which might be more comparable, and is an interesting case study, but that has multiple params and specific structure to it, whereas this just has one route param.

s-rubenstein (approved):
Pending the curry vs partial decision, I am happy with this as is.

src/pages/billing/Project.js
ehigham: Parking this conversation for a UI meeting. Lots of good comments.Reverting to cond.

src/libs/ajax.js
ehigham: @petesantos explained - thanks very much.

src/pages/billing/Project.js
ehigham: Follow-up - I didn't realise lodash's `curry` allowed you to pass more than one argument at a time (ie `f(a)(b)(c)` can be called `f(a, b, c)` when `f` is `_.curry`'ed.

src/libs/ajax.js
petesantos: In this particular case, I do think it comes down to how it is being used. It does seem odd that when we call this, it is always being called with with the billing project and then an immediate execution of one of the member functions or a reference to the member function. In other words, we never just store the result of the wrapper as `currentBillingProject` and invoke a call from there. And this is further complicated by the fact that we don't need a signal in one of the cases.

In this particular case, I am in favor of removing the wrapper. It does increase the arity of the functions, but it seems closer to what the function is doing and a bit simpler.

src/pages/billing/Project.js
petesantos: Looking through the code, while we do use `_.curry` we do not have a huge number of cases where we curry functions. Though we do take advantage of the auto currying in the `lodash/fp` variant, so it should become familiar to those using the codebase.

But it seems that most of the functions we curry in the codebase are wrappers (e.g. `withSomething`) or utilities intended to be used similar to an auto-curried lodash function. In this specific case, I like partial as it is clear that we are not fully executing a function - which is especially helpful in `addFunction` and `saveFunction`

Although this is tied to the concept of the `project` wrapper in the `Ajax` file.

Personally, I am ok with this. This would be another interesting topic for our next ui review discussion.

petesantos (approved):
Added the currying and ternary discussion to the next UI meeting

PR #2656, opened 2021-10-13
[AS-998] don't poll for the same import job twice

davidangb:
[AS-998](https://broadworkbench.atlassian.net/browse/AS-998). Tested locally via dev tools.

* when entering a workspace and finding outstanding import jobs, only add those jobs to the `asyncImportJobStore` if they don't already exist in `asyncImportJobStore`
* when creating the polling jobs, run `_.uniq` on `asyncImportJobStore` for safety. This shouldn't be necessary but exists as a safeguard in case some future code also adds duplicate jobs to `asyncImportJobStore`.

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -453,7 +453,7 @@ const WorkspaceData = _.flow(
       const currentJobIds = _.map('jobId', asyncImportJobStore.get())
       _.forEach(job => {
         const jobStatus = _.lowerCase(job.status)
-        if (!_.includes(jobStatus, ['success', 'error', 'done']) && _.indexOf(job.jobId, currentJobIds) === -1) {
+        if (!_.includes(jobStatus, ['success', 'error', 'done']) && !_.includes(job.jobId, currentJobIds)) {
           asyncImportJobStore.update(Utils.append({ targetWorkspace: { namespace, name }, jobId: job.jobId }))
         }
       }, runningJobs)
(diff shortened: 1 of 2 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -450,9 +450,10 @@ const WorkspaceData = _.flow(
   const getRunningImportJobs = async () => {
     try {
       const runningJobs = await Ajax(signal).Workspaces.workspace(namespace, name).listImportJobs(true)
+      const currentJobIds = _.map('jobId', asyncImportJobStore.get())
       _.forEach(job => {
         const jobStatus = _.lowerCase(job.status)
-        if (!_.includes(jobStatus, ['success', 'error', 'done'])) {
+        if (!_.includes(jobStatus, ['success', 'error', 'done']) && _.indexOf(job.jobId, currentJobIds) === -1) {
```
davidangb: could also have done a `_.find` directly against `asyncImportJobStore.get()` … it felt more efficient in cases where there exist multiple jobs inside `asyncImportJobStore` to call `.get()` once and transform its results, but I'm open to feedback on this approach

dmohs: My feedback is that I like your approach. Perhaps `includes` would make this slightly clearer:
```suggestion
        if (!_.includes(jobStatus, ['success', 'error', 'done']) && !_.includes(job.jobId, currentJobIds)) {
```

dmohs (approved):

src/pages/workspaces/workspace/Data.js
zarsky-broad: Remember that fp lodash has the arguments flipped, so `_.includes` takes haystack followed by needle; I don't think this would ever have been happy before...?

davidangb: @zarsky-broad standard lodash is haystack followed by needle (https://lodash.com/docs/4.17.15#includes), so I think this is correct as written? It works for me empirically in testing.

zarsky-broad: ugh, my bad, brain fart. carry on...

ansingh7115 (approved):

PR #2657, opened 2021-10-13
[AS-978] hide 'Use defaults' command for workflows on snapshots

davidangb:
[AS-978](https://broadworkbench.atlassian.net/browse/AS-978)

When configuring a workflow to run on a Terra Data Repo snapshot, hide the "Use defaults" command for workflow outputs, since those defaults do not typically apply to Terra Data Repo snapshots.

Tested manually in my browser!

### DIFF
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -91,8 +91,11 @@ const filterConfigIO = ({ inputs, outputs }) => {
   )
 }
 
-const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange, onSetDefaults, onBrowse, suggestions, readOnly }) => {
+const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange, onSetDefaults, onBrowse, suggestions, availableSnapshots, readOnly }) => {
   const [sort, setSort] = useState({ field: 'taskVariable', direction: 'asc' })
+
+  // will only match if the current root entity type comes from a snapshot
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -91,9 +91,12 @@ const filterConfigIO = ({ inputs, outputs }) => {
   )
 }
 
-const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange, onSetDefaults, onBrowse, suggestions, readOnly }) => {
+const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange, onSetDefaults, onBrowse, suggestions, availableSnapshots, readOnly }) => {
   const [sort, setSort] = useState({ field: 'taskVariable', direction: 'asc' })
 
+  // will only match if the current root entity type comes from a snapshot
+  const isSnapshot = !!_.find({ name: config.dataReferenceName }, availableSnapshots)
```
kyuksel: Would `_.some` be a better fit since we don't seem to care about the value `_.find` may return?

```suggestion
  const isSnapshot = _.some({ name: config.dataReferenceName }, availableSnapshots)
```

kyuksel (approved):
👍🏼  with a minor suggestion inline

src/pages/workspaces/workspace/workflows/WorkflowView.js
davidangb: good call, thank you!

MatthewBemis (approved):

PR #2659, opened 2021-10-19
[IA-3010] Fix compute type drop-down default

kyuksel:
This PR fixes the bug I introduced with the [Spark UI PR](https://github.com/DataBiosphere/terra-ui/pull/2643) such that 

Screen Shot 2021-10-19 at 5 54 48 PM

looks like

Screen Shot 2021-10-19 at 5 55 40 PM

again.

Tested locally.

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -623,7 +623,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         const gpuConfig = runtimeConfig?.gpuConfig
         const newSparkMode = Utils.switchCase(runtimeConfig?.cloudService,
           [cloudServices.DATAPROC, () => runtimeConfig.numberOfWorkers === 0 ? 'master' : 'cluster'],
-          [cloudServices.GCE, () => false]
+          [cloudServices.GCE, () => false],
+          [undefined, () => false] // for when there's no existing runtime
         )
         const isDataproc = !sparkMode && !runtimeConfig?.diskSize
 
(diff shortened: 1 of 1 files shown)


src/components/ComputeModal.js
```diff
@@ -623,7 +623,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         const gpuConfig = runtimeConfig?.gpuConfig
         const newSparkMode = Utils.switchCase(runtimeConfig?.cloudService,
           [cloudServices.DATAPROC, () => runtimeConfig.numberOfWorkers === 0 ? 'master' : 'cluster'],
-          [cloudServices.GCE, () => false]
+          [cloudServices.GCE, () => false],
+          [undefined, () => false] // for when there's no existing runtime
```
kyuksel: I preferred `undefined` over `Utils.DEFAULT` for clarity.

xuscrbw (approved):
LGTM

src/components/ComputeModal.js
petesantos: There is a subtle difference between `undefined` and `Utils.default`.

If `cloudService` is equivalent to `undefined` then `newSparkMode` will be set to false. However, if `cloudService` is something unexpected there will be no default case, resulting in `newSparkMode` getting set to `undefined`

Leaving `undefined` in may be ok, but do you may also want an explicit default case. If undefined is essentially the default, I would stick with `DEFAULT`.

kyuksel: Okay, that's sensible. I'll change to `Utils.DEFAULT`. I'm hoping to make a PR soon to refactor the overloaded usage of `sparkMode`, which may get rid of the ambiguity.

petesantos (approved):

PR #2660, opened 2021-10-20
[PF-692] Support regionality in Terra (starting with just one region, us-central1)

wnojopra:
This PR was originally in #2642, which was from a fork. The integration tests cannot run if PR was from a fork. Thus, we've created this one. 

# This PR:
1) Adds a dropdown to the "Create a New Workspace" dialog for Bucket location. It defaults to the current default setting, US multi-regional. The only other choice for this initial release is us-central1.
2) Adds a dropdown for choosing a compute region. For the initial release, the dropdown is disabled and set to us-central1 only.
3) Adds informative messaging to ComputeModal
4) Adds informative messaging to the Import Table Data dialog
5) Add informative messaging to Launch Analysis Modal
6) Adds a warning if you try to clone a workspace in a different location from the source. 

# I've tested:
- That creating a workspace with the default location works, and a non-default location.
- Cloud Environment creation testing:
  - Creating a GCE instance with the default us-multiregional bucket.
  - Creating a cluster with the default us-multiregional bucket.
  - Creating a GCE instance in a non default location (like australia-southeast1)
  - Creating a cluster in a non default location (like australia-southeast1)

# Release planning
We are targeting ~~October 18~~ October 26 for these changes to go live. The initial release will have us-central1 as the only region choice. Future release will slowly add more regions. With this release we will also be working with comms team to release a set of Terra articles explaining regionality (when/why you want to use it, egress risks, etc). These articles are available to preview as Google Docs in [Working with Non-US data in Terra](https://docs.google.com/document/d/1I-RKjtc0REMDZPJANCO2HcmBtKXHxDQeGiwliaYzFFs/edit) and [Regional Storage and Compute now available in Terra](https://docs.google.com/document/d/13DAwRXVM-qX5jYYz6o5EOMcTUXXhKoOBby2TK8kx0xY/edit#).

# Screenshots of UI changes:
Create/clone workspace
image

Warning in clone workspace
image

Create Cloud Environment
image

ComputeModal update
image

Import Table Data update
image

Launch Analysis Modal update
image


src/data/machines.js
```diff
@@ -46,22 +46,127 @@ export const gpuTypes = [
   { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 1, maxNumCpus: 12, maxMem: 78, price: 2.4800, preemptiblePrice: 0.7400 },
   { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 2, maxNumCpus: 24, maxMem: 156, price: 4.9600, preemptiblePrice: 1.4800 },
   { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 4, maxNumCpus: 48, maxMem: 312, price: 9.9200, preemptiblePrice: 2.9600 },
-  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 8, maxNumCpus: 96, maxMem: 624, price: 19.8400, preemptiblePrice: 5.9200 }
+  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 8, maxNumCpus: 96, maxMem: 624, price: 19.8400, preemptiblePrice: 5.9200 },
+  { name: 'NVIDIA Tesla P100', type: 'nvidia-tesla-p100', numGpus: 1, maxNumCpus: 16, maxMem: 104, price: 1.4600, preemptiblePrice: 0.4300 },
+  { name: 'NVIDIA Tesla P100', type: 'nvidia-tesla-p100', numGpus: 2, maxNumCpus: 32, maxMem: 208, price: 2.9200, preemptiblePrice: 0.8600 },
+  { name: 'NVIDIA Tesla P100', type: 'nvidia-tesla-p100', numGpus: 4, maxNumCpus: 96, maxMem: 624, price: 5.8400, preemptiblePrice: 1.7200 }
+
```
kyuksel: Let's remove the blank line:

```suggestion
```

src/components/region-common.js
```diff
@@ -1,74 +1,81 @@
...
+// 'flag' will always be defined (even if it's a question mark).
 // 'regionDescription' is the same as location when locationType is 'multi-region', or a country name when locationType is 'region'.
+// computeZone is generally the 'a' zone for each region, except for those regions where it is not available.
+// The choice to use the 'a' zone is arbitrary, choosing 'b' zone would also work.
+// The region choice for multi-region locations is arbitrary as well.
+
+// When updating region list, please also update the list in
+// https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf
 export const unknownRegionFlag = '❓'
 export const regionInfo = (location, locationType) => {
```
kyuksel: For consistency, can we use a verb (e.g. `getRegionInfo`) for the function name?

src/components/region-common.js
```diff
@@ -1,74 +1,81 @@
...
+      ['US-WEST2', () => ({ flag: '🇺🇸', regionDescription: `${regionDescription} (Los Angeles)`, computeZone: 'US-WEST2-A', computeRegion: 'US-WEST2' })],
+      ['US-WEST3', () => ({ flag: '🇺🇸', regionDescription: `${regionDescription} (Salt Lake City)`, computeZone: 'US-WEST3-A', computeRegion: 'US-WEST3' })],
+      ['US-WEST4', () => ({ flag: '🇺🇸', regionDescription: `${regionDescription} (Las Vegas)`, computeZone: 'US-WEST4-A', computeRegion: 'US-WEST4' })],
+      [Utils.DEFAULT, () => ({ flag: unknownRegionFlag, regionDescription, computeZone: 'UNKNOWN', computeRegion: 'UNKNOWN' })]
+    )],
+    [Utils.DEFAULT, () => ({ flag: unknownRegionFlag, regionDescription, computeZone: 'UNKNOWN', computeRegion: 'UNKNOWN' })]
+  )
 }
+
+export const allRegions = [
```
kyuksel, line 61: Should the name be `allLocations` since we also have `multi-region`s?

src/components/region-common.js
```diff
@@ -1,74 +1,81 @@
...
+  // In this list, us-east*, us-west*, northamerica-northeast2 and asia-northeast2 have purposefully been removed.
+  // This is to avoid creating within-country silos of life sciences community data.
+  // So for US, Canada and Japan, we are restricting to one region.
+  // For more information, see https://docs.google.com/document/d/1RMu8bxXAyP2q_85UHkPARW1Xpt_G_662pRi0LCML4yA/edit#heading=h.53j9ueoid5vt
+  // TODO(wnojopra): Replace the above link with the Terra support article link once published
+  { value: 'US', label: 'US multi-regional (default)', locationType: 'multi-region' },
+  { value: 'US-CENTRAL1', label: 'us-central1 (Iowa)', locationType: 'region' }
```
kyuksel: How would you feel about using enums for `locationType` so it's easier to see all possible values and a bit more difficult to accidentally insert a wrong string? We could also supply the default value there. For instance,

```javascript
export const locationTypes = {
  region: 'region',
  multiRegion: 'multi-region',
  default: 'multi-region'
}
```

src/libs/runtime-utils.js
```diff
@@ -53,16 +59,22 @@ export const normalizeRuntimeConfig = ({
...
   }
 }
 
 export const findMachineType = name => {
   return _.find({ name }, machineTypes) || { name, cpu: '?', memory: '?', price: NaN, preemptiblePrice: NaN }
 }
 
-export const getValidGpuTypes = (numCpus, mem) => {
-  const validGpuTypes = _.filter(({ maxNumCpus, maxMem }) => numCpus <= maxNumCpus && mem <= maxMem, gpuTypes)
+export const getValidGpusForZone = zone => {
```
kyuksel: Should this be
```suggestion
export const getValidGpuTypesForZone = zone => {
```
?

Similarly for `validGpusForZone` below.

src/components/ComputeModal.js
```diff
@@ -678,18 +711,29 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
     const { runtime: desiredRuntime } = getDesiredEnvironmentConfig()
-    const commonButtonProps = hasGpu && viewMode !== 'deleteEnvironmentOptions' ?
-      { disabled: true, tooltip: 'Cloud compute with GPU(s) cannot be updated. Please delete it and create a new one.' } :
-      { disabled: !hasChanges() || !!errors, tooltip: Utils.summarizeErrors(errors) }
-    const canShowCustomImageWarning = viewMode === undefined
+    const commonButtonProps = Utils.cond([
+      hasGpu && viewMode !== 'deleteEnvironmentOptions',
+      () => ({ disabled: true, tooltip: 'Cloud compute with GPU(s) cannot be updated. Please delete it and create a new one.' })
+    ], [
+      computeConfig.gpuEnabled && getValidGpusForZone(computeConfig.computeZone).length === 0 && viewMode !== 'deleteEnvironmentOptions',
```
kyuksel: Lodash FP's `_.isEmpty()` may fit nicely here:

```suggestion
      computeConfig.gpuEnabled && _.isEmpty(getValidGpusForZone(computeConfig.computeZone)) && viewMode !== 'deleteEnvironmentOptions',
```

src/libs/runtime-utils.js
```diff
@@ -32,13 +32,19 @@ export const defaultNumDataprocPreemptibleWorkers = 0
 export const defaultGpuType = 'nvidia-tesla-t4'
 export const defaultNumGpus = 1
 
+export const defaultLocation = 'US'
+export const defaultLocationType = 'multi-region'
+
+export const defaultComputeZone = 'US-CENTRAL1-A'
+export const defaultComputeRegion = 'US-CENTRAL1'
```
kyuksel: Could you remind me why these had to be uppercase? I've made mistakes in the past by mixing up case and would love to reduce chances of that if possible.

kyuksel (commented):
I ran out of time and wanted to submit what I have so far. Feel free to ignore until I submit the rest, hopefully early tomorrow (Eastern Time).

src/components/region-common.js
wnojopra: I've been using the term `location` informally to mean a zone or a region. Multi-regions are a set of regions, so I think its reasonable to call this allRegions.

Feel free to push back if you feel strongly about the name.

src/libs/runtime-utils.js
wnojopra: The value returned from Google's `storage/v1/b/${bucket}?fields=location` API returns locations in caps. There's a few places where we do comparisons with this value, so we make the default uppercase to match.

src/data/machines.js
wnojopra: Done

src/components/region-common.js
wnojopra: Done

src/components/region-common.js
wnojopra: Done

src/libs/runtime-utils.js
wnojopra: Yes, good catch. Done.

src/components/ComputeModal.js
wnojopra: Done

src/components/region-common.js
kyuksel: We actually use the term `location` in Leonardo to encompass `region` and `zone`. It'd be easier for me to not get confused if we make the distinction, especially given Javascript's lack of static typing.

src/data/machines.js
```diff
@@ -46,22 +46,126 @@ export const gpuTypes = [
...
 
 export const cloudServices = {
   GCE: 'GCE',
   DATAPROC: 'DATAPROC'
 }
 
-export const monthlyStoragePrice = 0.04 // from https://cloud.google.com/compute/pricing
-export const storagePrice = monthlyStoragePrice / 730 // per GB hour using 730 hours per month from https://cloud.google.com/compute/pricing
 export const dataprocCpuPrice = 0.01 // dataproc costs $0.01 per cpu per hour
 export const ephemeralExternalIpAddressPrice = { // per hour in dollars for Iowa (us-central1) region per https://cloud.google.com/vpc/network-pricing#ipaddress
```
kyuksel: It'd be good to make it clear here that we have a single constant defined because the ephemeral external ip address price is the same for all regions.

```suggestion
export const ephemeralExternalIpAddressPrice = { // per hour in dollars for all regions per https://cloud.google.com/vpc/network-pricing#ipaddress
```

src/libs/runtime-utils.js
```diff
@@ -32,13 +33,19 @@ export const defaultNumDataprocPreemptibleWorkers = 0
 export const defaultGpuType = 'nvidia-tesla-t4'
 export const defaultNumGpus = 1
 
+export const defaultLocation = 'US'
+export const defaultLocationType = locationTypes.default
```
kyuksel: Do we still need `defaultLocationType` since we can reference `locationTypes.default` directly?

src/libs/runtime-utils.js
```diff
@@ -102,19 +115,27 @@ export const runtimeConfigCost = config => {
...
+}
+
+export const getPersistentDiskCost = ({ size, status }, computeRegion) => {
+  const price = getStoragePriceForRegion(computeRegion)
+  return _.includes(status, ['Deleting', 'Failed']) ? 0.0 : size * price
+}
+export const getPersistentDiskCostMonthly = ({ size, status }, computeRegion) => {
+  const price = _.flow(_.find({ name: computeRegion }), _.get(['monthlyDiskPrice']))(regionToDiskPrice)
   return _.includes(status, ['Deleting', 'Failed']) ? 0.0 : size * price
 }
```
kyuksel: I know a lot of this predates your PR but I wonder if we can take the opportunity to eliminate some duplicate code and simplify things a bit, with something like

```javascript
// Per GB following https://cloud.google.com/compute/pricing
const getPersistentDiskPriceForRegionMonthly = computeRegion => {
  return _.flow(_.find({ name: computeRegion }), _.get(['monthlyDiskPrice']))(regionToDiskPrice)
}
const numberOfHoursPerMonth = 730
const getPersistentDiskPriceForRegionHourly = computeRegion => getPersistentDiskPriceForRegionMonthly(computeRegion) / numberOfHoursPerMonth

export const getPersistentDiskCostMonthly = ({ size, status }, computeRegion) => {
  const price = getPersistentDiskPriceForRegionMonthly(computeRegion)
  return _.includes(status, ['Deleting', 'Failed']) ? 0.0 : size * price
}
export const getPersistentDiskCostHourly = ({ size, status }, computeRegion) => getPersistentDiskCostMonthly({ size, status }, computeRegion) / numberOfHoursPerMonth
```

src/components/RuntimeManager.js
```diff
@@ -336,7 +337,7 @@ export default class RuntimeManager extends PureComponent {
           })
       }
     }
-    const totalCost = _.sum(_.map(runtimeCost, runtimes)) + _.sum(_.map(persistentDiskCost, persistentDisks))
+    const totalCost = _.sum(_.map(runtimeCost, runtimes)) + _.sum(_.map(getPersistentDiskCost(defaultComputeRegion), persistentDisks))
```
kyuksel: Don't we need to determine `computeRegion` dynamically here?

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -246,7 +247,7 @@ export const CloudEnvironmentModal = ({
     [toolLabel === tools.galaxy.label, () => getGalaxyCostTextChildren(currentApp, galaxyDataDisks)],
     [getRuntimeForTool(toolLabel), () => {
       const runtime = getRuntimeForTool(toolLabel)
-      const totalCost = runtimeCost(runtime) + _.sum(_.map(persistentDiskCost, persistentDisks))
+      const totalCost = runtimeCost(runtime) + _.sum(_.map(getPersistentDiskCost(defaultComputeRegion), persistentDisks))
```
kyuksel: 1. Don't we need to determine `computeRegion` dynamically here?
2. Are we missing the other input (`{ size, status }`)?

src/components/ComputeModal.js
```diff
@@ -561,6 +574,20 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+    updateComputeConfig('computeZone', computeZone)
+    updateComputeConfig('computeRegion', computeRegion)
+  }
+
+  const isDifferentLocation = () => {
+    // If the bucket is regional, we can easily compare the bucketLocation with the compute region.
+    // bucketLocation === 'US' means the bucket is US multi-regional.
+    // For a US multi-regional bucket, the computeRegion needs to be US-CENTRAL1 in order to be considered "in the same location".
+    // Currently, US is the only multi-region supported in Terra
+    return bucketLocation === 'US' ? computeConfig.computeRegion !== defaultComputeRegion : computeConfig.computeRegion !== bucketLocation
```
kyuksel: 1. Can we avoid using a magic string (i.e. `'US'`)?
2. My eyes had a little difficult time parsing the ternary. I wonder if it's worth a bit more verbosity in this case.

```javascript
    if (bucketLocation === defaultLocation) {
      return computeConfig.computeRegion !== defaultComputeRegion
    } else {
      return computeConfig.computeRegion !== bucketLocation
    }
```

src/components/ComputeModal.js
```diff
@@ -679,18 +712,29 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   const renderActionButton = () => {
     const { runtime: existingRuntime, hasGpu } = getExistingEnvironmentConfig()
     const { runtime: desiredRuntime } = getDesiredEnvironmentConfig()
```
kyuksel, line 720: There's some misalignment below and in a couple of other places in this file. 

If you use IntelliJ, would you be able to run the reformatter as described in `Code style and linting` of the README (screenshot below)?

If you don't, let me know and I can make a PR into your branch at some point.

image

src/components/region-common.js
```diff
@@ -1,74 +1,87 @@
...
+export const getRegionInfo = (location, locationType) => {
+  const regionDescription = `${locationType}: ${location}`.toLowerCase()
+  return Utils.switchCase(locationType,
+    ['multi-region', () => Utils.switchCase(location,
+      ['US', () => ({ flag: '🇺🇸', regionDescription, computeZone: 'US-CENTRAL1-A', computeRegion: 'US-CENTRAL1' })],
+      ['EU', () => ({ flag: '🇪🇺', regionDescription, computeZone: 'EUROPE-CENTRAL2-A', computeRegion: 'EUROPE-CENTRAL2' })],
+      ['ASIA', () => ({ flag: '🌏', regionDescription, computeZone: 'ASIA-EAST1-A', computeRegion: 'ASIA-EAST1' })],
+      [Utils.DEFAULT, () => ({ flag: unknownRegionFlag, regionDescription, computeZone: 'UNKNOWN', computeRegion: 'UNKNOWN' })]
+    )],
+    ['region', () => Utils.switchCase(location,
```
kyuksel, line 22: For when `locationType` is `region`, does it make sense to set `computeRegion` to `location` instead of the string literal? It'd be one less field to worry about in terms of typos, keeping up to date, etc.

src/components/ComputeModal.js
```diff
@@ -679,18 +712,29 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
-      { disabled: !hasChanges() || !!errors, tooltip: Utils.summarizeErrors(errors) }
-    const canShowCustomImageWarning = viewMode === undefined
+    const commonButtonProps = Utils.cond([
+      hasGpu && viewMode !== 'deleteEnvironmentOptions',
+      () => ({ disabled: true, tooltip: 'Cloud compute with GPU(s) cannot be updated. Please delete it and create a new one.' })
+    ], [
+      computeConfig.gpuEnabled && _.isEmpty(getValidGpuTypesForZone(computeConfig.computeZone)) && viewMode !== 'deleteEnvironmentOptions',
+      () => ({ disabled: true, tooltip: 'GPUs not available in this location.' })
+    ], [
+      currentPersistentDiskDetails && currentPersistentDiskDetails.zone.toUpperCase() !== computeConfig.computeZone && viewMode !== 'deleteEnvironmentOptions',
```
kyuksel: We prefer explicit casting to boolean when a non-boolean const is used in logical operations:

```suggestion
      !!currentPersistentDiskDetails && currentPersistentDiskDetails.zone.toUpperCase() !== computeConfig.computeZone && viewMode !== 'deleteEnvironmentOptions',
```

kyuksel (commented):
I'm mostly through but will be in meetings for a bit so wanted to submit what I have so far out.

src/components/region-common.js
wnojopra: Great. So to be clear, we're good on the name allRegions here, since this doesn't include zones?

src/data/machines.js
wnojopra: Done.

src/libs/runtime-utils.js
wnojopra: Done. I made some fixes to how you've written getPersistentDiskCostHourly:
```
export const getPersistentDiskCostHourly = ({ size, status }, computeRegion) => {
  const price = getPersistentDiskPriceForRegionHourly(computeRegion)
  return _.includes(status, ['Deleting', 'Failed']) ? 0.0 : size * price
}
```

src/components/ComputeModal.js
wnojopra: Done

src/components/region-common.js
wnojopra: Yes, this is a no-brainer. Thanks! Done.

src/components/ComputeModal.js
wnojopra: Done

src/libs/runtime-utils.js
wnojopra: We don't. Fixed.

src/components/RuntimeManager.js
wnojopra: Yes. In fact there was a bug here the disk params wasn't being passed to getPersistentDiskCost. This resulted in the price not being shown in the little runtime manager box (which I didn't notice until today). I've fixed this, and will update my test plan doc to be aware of this.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
wnojopra: Yes. In fact there was a bug here the disk params wasn't being passed to getPersistentDiskCost. This resulted in the price not being shown in cloud environment modal (which I didn't notice until today, when I figured out how to override the isAnalysisTabVisible config ). I've fixed this, and will update my test plan doc to be aware of this.

src/components/ComputeModal.js
wnojopra: Ok! Managed to get IntelliJ installed and followed those instructions to fix the spacing.

src/components/region-common.js
kyuksel: Yes, it's fine to keep as is for now. If we add zones at some point, we can rename it then.

src/components/ComputeModal.js
```diff
@@ -953,6 +1001,27 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+            ]),
+            div({ style: { marginTop: '0.5rem' } }, [
+              h(Select, {
+                id,
+                isDisabled: true, // Initially releasing regionality supporting only us-central1 region
+                // isDisabled: computeExists, // Can't update location of existing environments
+                isSearchable: false,
+                value: computeConfig.computeRegion,
+                onChange: ({ value, locationType }) => updateComputeLocation(value, locationType),
+                options: _.flow(_.filter(l => l.value !== 'US'), _.sortBy('label'))(allRegions)
```
kyuksel: One more magic string left:

```suggestion
                options: _.flow(_.filter(l => l.value !== defaultLocation), _.sortBy('label'))(allRegions)
```

src/components/ComputeModal.js
```diff
@@ -1012,6 +1082,33 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+        hideCloseButton: isAnalysisMode,
+        style: computeStyles.titleBar,
+        title: h(WarningTitle, ['Compute location differs from workspace bucket location']),
+        onDismiss,
+        onPrevious: () => setViewMode(undefined)
+      }),
+      div({ style: { lineHeight: 1.5 } }, [
+        p([
+          'This cloud environment will be created in the region ', computeConfig.computeRegion.toLowerCase(), '. ',
+          'Copying data from your workspace bucket in ', bucketLocation.toLowerCase(), ' may incur network egress charges.'
```
kyuksel: We usually use string extrapolation to embed code, which to me is a bit easier to read, a bit harder to mess up on an IDE :)
```suggestion
          `This cloud environment will be created in the region ${computeConfig.computeRegion.toLowerCase()}. `,
          `Copying data from your workspace bucket in ${bucketLocation.toLowerCase()} may incur network egress charges.`
```

src/components/ComputeModal.js
```diff
@@ -1012,6 +1082,33 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+        ]),
+        h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360058964552', ...Utils.newTabLinkProps }, [
+          'For more information please read the documentation.',
+          icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
+        ]),
+        p(['If you want your VM in ', computeConfig.computeRegion.toLowerCase(), ', continue. Otherwise, go back to select another location.'])
```
kyuksel: Similarly as above

```suggestion
        p([`If you want your VM in ${computeConfig.computeRegion.toLowerCase()}, continue. Otherwise, go back to select another location.`])
```

src/libs/runtime-utils.js
```diff
@@ -53,16 +58,22 @@ export const normalizeRuntimeConfig = ({
...
 
-export const getValidGpuTypes = (numCpus, mem) => {
-  const validGpuTypes = _.filter(({ maxNumCpus, maxMem }) => numCpus <= maxNumCpus && mem <= maxMem, gpuTypes)
+export const getValidGpuTypesForZone = zone => {
+  return _.flow(_.find({ name: zone }), _.get(['validTypes']))(zonesToGpus)
+}
+
+export const getValidGpuTypes = (numCpus, mem, zone) => {
+  const validGpuTypesForZone = getValidGpuTypesForZone(zone)
+  const validGpuTypes = _.filter(({ maxNumCpus, maxMem, type }) => numCpus <= maxNumCpus && mem <= maxMem && validGpuTypesForZone.includes(type), gpuTypes)
```
kyuksel, line 76: I couldn't see where `type` is coming from?

kyuksel (commented):
I'm done going through the code. Great work overall, and thank you very much for taking feedback so well  🙏🏼 

Tomorrow, hopefully with a fresher brain, I will:
- cross-check prices, GPU types, etc.
- review any code changes from today.

src/libs/runtime-utils.js
wnojopra: Its in the const gpuTypes imported from machines.js, along with maxNumCpus and maxMem.

```
export const gpuTypes = [
  { name: 'NVIDIA Tesla T4', type: 'nvidia-tesla-t4', numGpus: 1, maxNumCpus: 24, maxMem: 156, price: 0.3500, preemptiblePrice: 0.1100 },
...
```

src/components/ComputeModal.js
wnojopra: Done

src/components/ComputeModal.js
wnojopra: Done

src/components/ComputeModal.js
wnojopra: Done

src/libs/runtime-utils.js
kyuksel: oh right - thanks!

src/components/NewWorkspaceModal.js
```diff
@@ -31,6 +33,15 @@ const constraints = {
...
 }
 
+const warningStyle = {
+  border: `1px solid ${colors.warning(0.8)}`, borderLeft: 'none', borderRight: 'none',
+  backgroundColor: colors.warning(0.15),
+  padding: '1rem 1.25rem', margin: '0 -1.25rem',
+  fontWeight: 'bold', fontSize: 12,
+  display: 'flex',
+  alignItems: 'center'
+}
```
kyuksel: It looks like this styling is almost the same as the [warningStyle in ExportDataModal](https://github.com/DataBiosphere/terra-ui/blob/db2e08b0d66832cd784b3035bf904d2067d0d15b/src/components/data/ExportDataModal.js#L18). Can we factor out the common code into an existing util/helper file and reuse the common part here?

kyuksel:
@wnojopra have we cleared the newly added warning with the UX folks?


image

kyuksel:
While testing the new workspace cloning warning (which works well), when I clicked on the workspace's `Clone` button

image


I noticed that a pop-up appears very briefly prior to seeing 

image

I couldn't take a screenshot of it because it disappears too quickly but the message comes from [this line](https://github.com/DataBiosphere/terra-ui/blob/eef4b3ef5c8536f319a97a688e6741c41e65d944/src/components/NewWorkspaceModal.js#L285) of `NewWorkspaceModal` that says `You need a billing project to…`.

I don't see the same thing happen with the `dev` code (using exactly the same workspace) although I couldn't see how any of this PR's changes may have introduced this behaviour. 

While troubleshooting, I added this line

```javascript
console.log('billing projects: ', billingProjects)
```

right before `return Utils.cond(` in the `Render` section and noticed from the console that the modal gets rendered multiple times:

image

It gets rendered with `dev` multiple times too but fewer times. Perhaps the new changes led to a visible manifestation of an issue that had been there?

When I added ```

src/components/NewWorkspaceModal.js

wnojopra:
> I noticed that a pop-up appears very briefly

In the useOnMount function it was calling loadProjectsGroups() followed by loadBucketLocation(). Both functions were making two separate async calls changing the state independently. I've fixed this by putting the two methods inside one Promise.

src/components/NewWorkspaceModal.js
```diff
@@ -101,11 +106,22 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
         setBillingProjects(projects)
         setNamespace(_.some({ projectName: namespace }, projects) ? namespace : undefined)
       }),
-    Ajax(signal).Groups.list().then(setAllGroups)
+    Ajax(signal).Groups.list().then(setAllGroups),
+    !! cloneWorkspace && Ajax(signal).Workspaces.workspace(namespace, cloneWorkspace.workspace.name).checkBucketLocation(cloneWorkspace.workspace.googleProject, cloneWorkspace.workspace.bucketName)
```
kyuksel: For consistency:

```suggestion
    !!cloneWorkspace && Ajax(signal).Workspaces.workspace(namespace, cloneWorkspace.workspace.name).checkBucketLocation(cloneWorkspace.workspace.googleProject, cloneWorkspace.workspace.bucketName)
```

src/components/NewWorkspaceModal.js
```diff
@@ -101,11 +106,22 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
...
+      .then(locationResponse => {
+        setBucketLocation(locationResponse.location)
+        setSourceWorkspaceLocation(locationResponse.location)
+      })
   ]))
 
+  const showDifferentRegionWarning = () => {
+    return !!cloneWorkspace && bucketLocation !== sourceWorkspaceLocation
+  }
```
kyuksel: For control booleans like this, we try to have a name that makes it clear that the value is a boolean; e.g. `shouldShowDifferentRegionWarning`. 

Also, does it need to be a function or can it be a constant?

src/data/machines.js
```diff
@@ -46,22 +46,126 @@ export const gpuTypes = [
...
+  { name: 'ASIA-SOUTH1-C', validTypes: [] },
+  { name: 'ASIA-SOUTHEAST1-A', validTypes: ['nvidia-tesla-t4'] },
+  { name: 'ASIA-SOUTHEAST1-B', validTypes: ['nvidia-tesla-t4', 'nvidia-tesla-p4'] },
+  { name: 'ASIA-SOUTHEAST1-C', validTypes: ['nvidia-tesla-t4', 'nvidia-tesla-p4'] },
+  { name: 'ASIA-SOUTHEAST2-A', validTypes: ['nvidia-tesla-t4'] },
+  { name: 'ASIA-SOUTHEAST2-B', validTypes: ['nvidia-tesla-t4'] },
+  { name: 'ASIA-SOUTHEAST2-C', validTypes: [] },
+  { name: 'AUSTRALIA-SOUTHEAST1-A', validTypes: ['nvidia-tesla-t4', 'nvidia-tesla-p4'] },
+  { name: 'AUSTRALIA-SOUTHEAST1-B', validTypes: ['nvidia-tesla-p4'] },
+  { name: 'AUSTRALIA-SOUTHEAST1-C', validTypes: ['nvidia-tesla-p4'] },
```
kyuksel: I see a different GPU type on https://cloud.google.com/compute/docs/gpus/gpu-regions-zones (screenshot below)

```suggestion
  { name: 'AUSTRALIA-SOUTHEAST1-C', validTypes: ['nvidia-tesla-p100'] },
```

Screen Shot 2021-10-22 at 5 08 59 PM

src/data/machines.js
```diff
@@ -46,22 +46,126 @@ export const gpuTypes = [
...
+  { name: 'EUROPE-WEST6-C', validTypes: [] },
+  { name: 'NORTHAMERICA-NORTHEAST1-A', validTypes: ['nvidia-tesla-p4'] },
+  { name: 'NORTHAMERICA-NORTHEAST1-B', validTypes: ['nvidia-tesla-p4'] },
+  { name: 'NORTHAMERICA-NORTHEAST1-C', validTypes: ['nvidia-tesla-p4'] },
+  { name: 'SOUTHAMERICA-EAST1-A', validTypes: [] },
+  { name: 'SOUTHAMERICA-EAST1-B', validTypes: [] },
+  { name: 'SOUTHAMERICA-EAST1-C', validTypes: ['nvidia-tesla-t4'] },
+  { name: 'US-CENTRAL1-A', validTypes: ['nvidia-tesla-t4', 'nvidia-tesla-v100', 'nvidia-tesla-p4', 'nvidia-tesla-k80'] },
+  { name: 'US-CENTRAL1-B', validTypes: ['nvidia-tesla-t4', 'nvidia-tesla-v100'] },
+  { name: 'US-CENTRAL1-C', validTypes: ['nvidia-tesla-t4', 'nvidia-tesla-v100', 'nvidia-tesla-p100', 'nvidia-tesla-p4', 'nvidia-tesla-k80'] },
```
kyuksel: I couldn't see a `T4` on the Google page.

```suggestion
  { name: 'US-CENTRAL1-C', validTypes: ['nvidia-tesla-v100', 'nvidia-tesla-p100', 'nvidia-tesla-p4', 'nvidia-tesla-k80'] },
```

src/data/machines.js
```diff
@@ -46,22 +46,126 @@ export const gpuTypes = [
...
+  { name: 'US-CENTRAL1-F', validTypes: ['nvidia-tesla-v100', 'nvidia-tesla-p100'] },
```
kyuksel: I do see a `T4` for this one:

```suggestion
  { name: 'US-CENTRAL1-F', validTypes: ['nvidia-tesla-t4', 'nvidia-tesla-v100', 'nvidia-tesla-p100'] },
```

src/data/machines.js
```diff
@@ -46,22 +46,126 @@ export const gpuTypes = [
...
+  { name: 'US-EAST1-B', validTypes: ['nvidia-tesla-p100'] },
+  { name: 'US-EAST1-C', validTypes: ['nvidia-tesla-t4', 'nvidia-tesla-v100', 'nvidia-tesla-p4', 'nvidia-tesla-k80'] },
```
kyuksel:
```suggestion
  { name: 'US-EAST1-C', validTypes: ['nvidia-tesla-t4', 'nvidia-tesla-v100', 'nvidia-tesla-p100', 'nvidia-tesla-k80'] },
```

src/data/machines.js
```diff
@@ -46,22 +46,126 @@ export const gpuTypes = [
...
 }
 
-export const monthlyStoragePrice = 0.04 // from https://cloud.google.com/compute/pricing
-export const storagePrice = monthlyStoragePrice / 730 // per GB hour using 730 hours per month from https://cloud.google.com/compute/pricing
 export const dataprocCpuPrice = 0.01 // dataproc costs $0.01 per cpu per hour
-export const ephemeralExternalIpAddressPrice = { // per hour in dollars for Iowa (us-central1) region per https://cloud.google.com/vpc/network-pricing#ipaddress
+export const ephemeralExternalIpAddressPrice = { // per hour in dollars for all regions per https://cloud.google.com/vpc/network-pricing#ipaddress
   standard: 0.004,
   preemptible: 0.002
 }
```
kyuksel, line 141: Can we add a comment here to

1. say that the prices below are per GB
2. reference the source: https://cloud.google.com/compute/all-pricing#disk

src/data/machines.js
```diff
@@ -46,22 +46,126 @@ export const gpuTypes = [
   { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 1, maxNumCpus: 12, maxMem: 78, price: 2.4800, preemptiblePrice: 0.7400 },
   { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 2, maxNumCpus: 24, maxMem: 156, price: 4.9600, preemptiblePrice: 1.4800 },
   { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 4, maxNumCpus: 48, maxMem: 312, price: 9.9200, preemptiblePrice: 2.9600 },
-  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 8, maxNumCpus: 96, maxMem: 624, price: 19.8400, preemptiblePrice: 5.9200 }
+  { name: 'NVIDIA Tesla V100', type: 'nvidia-tesla-v100', numGpus: 8, maxNumCpus: 96, maxMem: 624, price: 19.8400, preemptiblePrice: 5.9200 },
+  { name: 'NVIDIA Tesla P100', type: 'nvidia-tesla-p100', numGpus: 1, maxNumCpus: 16, maxMem: 104, price: 1.4600, preemptiblePrice: 0.4300 },
+  { name: 'NVIDIA Tesla P100', type: 'nvidia-tesla-p100', numGpus: 2, maxNumCpus: 32, maxMem: 208, price: 2.9200, preemptiblePrice: 0.8600 },
+  { name: 'NVIDIA Tesla P100', type: 'nvidia-tesla-p100', numGpus: 4, maxNumCpus: 96, maxMem: 624, price: 5.8400, preemptiblePrice: 1.7200 }
+]
```
kyuksel, line 53: [GPU pricing](https://cloud.google.com/compute/all-pricing#gpus) varies not only by `type` but also `region`. Are we planning on adding that info?

kyuksel (commented):
I finished cross-checking resource availability and pricing data and noted the discrepancies I've seen inline.

On Monday morning, I'll go through the diff from yesterday and today. I think we're very close to done code review-wise 😊

src/components/NewWorkspaceModal.js
wnojopra: Done

src/components/NewWorkspaceModal.js

I believe it does need to be a function, because bucketLocation changes based on the dropdown choice.

src/data/machines.js
wnojopra: Fixed

src/data/machines.js
wnojopra: Fixed

src/data/machines.js
wnojopra: Fixed

src/data/machines.js
wnojopra: Fixed

src/data/machines.js
wnojopra: Done

src/data/machines.js
wnojopra: Yes. I thought this PR was large enough, so I chose not to add all that info here once we decided to do a us-central1 only release. 

I also have a script that can parse all that information out through an API, so there won't be any manual mistakes next time.

kyuksel: > I thought this PR was large enough, so I chose not to add all that info here once we decided to do a us-central1 only release.

👍🏼  Is the TO-DO tracked anywhere?

> I also have a script that can parse all that information out through an API, so there won't be any manual mistakes next time.

That sounds awesome! I wonder if we should add it to the repo (not necessarily as part of this PR)?

wnojopra: > 👍🏼 Is the TO-DO tracked anywhere?

Just filed [PF-1068](https://broadworkbench.atlassian.net/browse/PF-1068) for it

> I wonder if we should add it to the repo (not necessarily as part of this PR)?

Thats a good idea. I will clean it up, add it to the scripts dir, and create a PR. It is in Python though (my language of choice). I could try creating a bash version that uses jq to parse. But if that ends up being too hard I'll just submit the Python version.

integration-tests/tests/import-cohort-data.js
```diff
@@ -27,6 +27,8 @@ const testImportCohortDataFn = _.flow(
   await click(page, clickable({ textContains: 'an existing workspace' }))
   await select(page, 'Select a workspace', workspaceName)
   await click(page, clickable({ text: 'Import' }))
+  // Wait a bit for the workspace to load
```
cahrens: Instead of adding a sleep, it would be **much** better to poll for the item to be ready. I don't know what utilities we have in the code base to help with this (hopefully others do), but it is a much more robust testing pattern.

Adding sleeps causes both sporadic failures (not waiting long enough) and test length to be extended (waiting longer than necessary).

integration-tests/tests/preview-drs-uri.js
```diff
@@ -28,6 +28,8 @@ const testPreviewDrsUriFn = _.flow(
   await click(page, clickable({ textContains: workspaceName }))
 
   await click(page, navChild('data'))
+  // Wait a bit for the workspace to load
+  await delay(5000)
```
cahrens: Same comment here-- we should poll until the workspace is loaded instead of adding a delay.

cahrens (approved):
I am approving from the perspective of the Launch Submission dialog changes, but I have serious concerns about adding delays into the integration tests. Please check with members of the ui-champions team to see if there is a way to poll for the items to be present instead of picking an arbitrary sleep time.

cahrens:
Also, I know the flaky integration tests are **super** annoying. But I believe there is a team who is tasked with fixing them this quarter (and hopefully there are JIRA tickets representing them). I don't think you should try to fix the issue as part of this merge request by adding sleeps.

wnojopra:
>  I don't think you should try to fix the issue as part of this merge request by adding sleeps.

Totally understand the hesitancy about the sleeps. The sleeps aren't there to fix the flakiness - but actually my code change means the workspace now has to to make a Google API call to fetch the GCS bucket location. It looks like this is causing the test to see the workspace, but for some reason isn't actually clicking on the right thing. I can tell from the screenshot in the integration test that it polls for the element it needs to click on, and actually finds it! But I can't figure out why it doesn't click on it because the screenshot shows its obviously hovering over the thing it needs to click on. By letting the test sleep for a few seconds, I think its letting the API call finish, letting the React states be stable, and so it can actually now click.

integration-tests/tests/import-cohort-data.js
wnojopra: Totally understand the hesitancy about the sleeps. The sleeps aren't there to fix the flakiness - but actually my code change means the workspace now has to to make a Google API call to fetch the GCS bucket location. It looks like this is causing the test to see the workspace, but for some reason isn't actually clicking on the right thing. I can tell from the screenshot in the integration test that it polls for the element it needs to click on, and actually finds it! But I can't figure out why it doesn't click on it because the screenshot shows its obviously hovering over the thing it needs to click on. By letting the test sleep for a few seconds, I think its letting the API call finish, letting the React states be stable, and so it can actually now click.

cahrens:
@wnojopra Thank you for the explanation. If you keep the sleeps in, can you please add comments that explain more about the issue? The comment that is present made me think the issue was something that could be polled for.

wnojopra:
> @wnojopra Thank you for the explanation. If you keep the sleeps in, can you please add comments that explain more about the issue? The comment that is present made me think the issue was something that could be polled for.

@cahrens Actually, I figured out how to fix the tests without the explicit sleep. And added comments. Win/Win! Thank you so much for bringing up the issue.

kyuksel:
(I had meant to send this hours ago and just now realized I've never hit the `Comment` button :(

Re: the delays added to the two tests

Curious why this PRs changes are slowing down steps in those tests.
Also, I wonder if there is a way to wait without the explicit delays.

wnojopra:
> Curious why this PRs changes are slowing down steps in those tests.
> Also, I wonder if there is a way to wait without the explicit delays.

Loading the workspace now needs to make an API call to GCS, so perhaps some state is changing causing a re-render where it wasn't before. 

And I've just pushed a fix so the tests pass without explicit delays. Yay! 

I'm just re-running the tests now until the flaky ones pass (a different test has been failing each time)

kyuksel:
> I will clean it up, add it to the scripts dir, and create a PR. It is in Python though (my language of choice). I could try creating a bash version that uses jq to parse. But if that ends up being too hard I'll just submit the Python version.

Sounds good. Thank you in advance!

src/components/ComputeModal.js
```diff
@@ -618,7 +618,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       ])
       const filteredNewLeoImages = !!tool ? _.filter(image => _.includes(image.id, tools[tool].imageIds), newLeoImages) : newLeoImages
 
-      const { location, locationType } = await Ajax().Workspaces.workspace(namespace, name).checkBucketLocation(googleProject, bucketName)
+      const { location } = await Ajax().Workspaces.workspace(namespace, name).checkBucketLocation(googleProject, bucketName)
```
kyuksel, line 621: I wonder if we could eliminate this new API call. `ComputeModal` seems to be passed in a `workspace` object. Does that not have the necessary bucket info?

wnojopra: It does not contain the bucket location. It would make sense to update the workspace object to include the bucket location because the bucket location is a never-changing property of the bucket/workspace. I believe that includes a rawls and database update though, which is fairly outside the scope of this PR.

kyuksel: Yea. Another route might be to pass in `location` into `ComputeModal`. From a quick look, the parents of `ComputeModal` seems to have `location`. What do you think?

wnojopra: Looks like the following components create a `ComputeModal`:
- `AppLauncher`
- `RuntimeManager`
- `NotebookLauncher`
- `AnalysisModal`
- `CloudEnvModal`
- `GalaxyModal`

Of those, it seems like `AppLauncher` and `NotebookLauncher` don't have a way to pass `location` to `ComputeModal`. 

Could we run with this API call for now and investigate into passing in location later this week?

kyuksel: Sounds good.

kyuksel: Would you be able to create a ticket for it so we don't forget?

wnojopra: Yes, https://broadworkbench.atlassian.net/browse/PF-1070

kyuksel (approved):

PR #2661, opened 2021-10-22
[DR-2182] Add autocompletion

petesantos:
https://broadworkbench.atlassian.net/browse/DR-2182

This change adds the ability to populate the data explorer filter with auto completion.
Our current AutoComplete component was designed to be used with a list of strings. In this case the filter is looking at other properties on an object of values.

Some changes were made to the component to be able to support object filtering.

The other changes were to add rendering logic to highlight where the match was found.

This PR was reviewed by UX. I will be making[ follow up changes](https://broadworkbench.atlassian.net/browse/DR-2210) for styling

image

### DIFF
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -230,15 +230,17 @@ const AutocompleteSuggestions = ({ target: targetId, containerProps, children })
 }
 
 const withAutocomplete = WrappedComponent => ({
-  instructions, value, onChange, onPick, suggestions: rawSuggestions, style, id, labelId,
-  renderSuggestion = _.identity, openOnFocus = true, placeholderText, ...props
+  instructions, itemToString, value, onChange, onPick, suggestions: rawSuggestions, style, id, labelId,
+  renderSuggestion = _.identity, openOnFocus = true, suggestionFilter = Utils.textMatch, placeholderText, ...props
 }) => {
   Utils.useLabelAssert('withAutocomplete', { id, 'aria-labelledby': labelId, ...props, allowId: true })
 
-  const suggestions = _.filter(Utils.textMatch(value), rawSuggestions)
+  const suggestions = _.filter(suggestionFilter(value), rawSuggestions)
+  const controlProps = itemToString ? { itemToString: v => v ? itemToString(v) : value } : { selectedItem: value }
 
   return h(Downshift, {
-    selectedItem: value,
+    ...controlProps,
+    initialInputValue: value,
     onSelect: v => !!v && onPick?.(v),
     onInputValueChange: newValue => {
       if (newValue !== value) {
@@ -267,7 +269,7 @@ const withAutocomplete = WrappedComponent => ({
             } else if (_.includes(e.key, ['ArrowUp', 'ArrowDown']) && !suggestions.length) {
               e.nativeEvent.preventDownshiftDefault = true
             } else if (e.key === 'Enter') {
-              onPick && onPick(value)
+              onPick ? onPick(value) : toggleMenu()
             }
           },
           nativeOnChange: true,
@@ -283,7 +285,7 @@ const withAutocomplete = WrappedComponent => ({
           }, [placeholderText])]],
           [!_.isEmpty(suggestions), () => _.map(([index, item]) => {
             return div(getItemProps({
-              item, key: item,
+              item, key: index,
               style: styles.suggestion(highlightedIndex === index)
             }), [renderSuggestion(item)])
           }, Utils.toIndexPairs(suggestions))])
@@ -294,6 +296,8 @@ const withAutocomplete = WrappedComponent => ({
 }
 
 export const AutocompleteTextInput = withAutocomplete(TextInput)
+
+export const DelayedAutoCompleteInput = withDebouncedChange(AutocompleteTextInput)
 
 export const TextArea = ({ onChange, autosize = false, nativeOnChange = false, ...props }) => {
   Utils.useLabelAssert('TextArea', { ...props, allowId: true })
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -269,7 +269,8 @@ const withAutocomplete = WrappedComponent => ({
             } else if (_.includes(e.key, ['ArrowUp', 'ArrowDown']) && !suggestions.length) {
               e.nativeEvent.preventDownshiftDefault = true
             } else if (e.key === 'Enter') {
-              onPick ? onPick(value) : toggleMenu()
+              onPick && onPick(value)
+              toggleMenu()
             }
           },
           nativeOnChange: true,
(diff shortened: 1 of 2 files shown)


src/components/input.js
```diff
@@ -230,15 +230,17 @@ const AutocompleteSuggestions = ({ target: targetId, containerProps, children })
...
 const withAutocomplete = WrappedComponent => ({
-  instructions, value, onChange, onPick, suggestions: rawSuggestions, style, id, labelId,
-  renderSuggestion = _.identity, openOnFocus = true, placeholderText, ...props
+  instructions, itemToString, value, onChange, onPick, suggestions: rawSuggestions, style, id, labelId,
+  renderSuggestion = _.identity, openOnFocus = true, suggestionFilter = Utils.textMatch, placeholderText, ...props
 }) => {
   Utils.useLabelAssert('withAutocomplete', { id, 'aria-labelledby': labelId, ...props, allowId: true })
 
-  const suggestions = _.filter(Utils.textMatch(value), rawSuggestions)
+  const suggestions = _.filter(suggestionFilter(value), rawSuggestions)
```
petesantos, line 238: Added the ability to pass in a custom filter

src/components/input.js
```diff
@@ -230,15 +230,17 @@ const AutocompleteSuggestions = ({ target: targetId, containerProps, children })
...
+  const controlProps = itemToString ? { itemToString: v => v ? itemToString(v) : value } : { selectedItem: value }
```
petesantos: `itemToString` is a downshift property, you cannot have both `selectedItem` and `itemToString` otherwise you will get [react control](https://reactjs.org/docs/uncontrolled-components.html) warnings

src/components/input.js
```diff
@@ -283,7 +285,7 @@ const withAutocomplete = WrappedComponent => ({
           }, [placeholderText])]],
           [!_.isEmpty(suggestions), () => _.map(([index, item]) => {
             return div(getItemProps({
-              item, key: item,
+              item, key: index,
```
petesantos, line 291: Ran into collisions with the item. When selecting objects, there can be non-unique prop values (e.g. multiple items with the same title)

src/components/input.js
```diff
@@ -295,6 +297,8 @@ const withAutocomplete = WrappedComponent => ({
 
 export const AutocompleteTextInput = withAutocomplete(TextInput)
 
+export const DelayedAutoCompleteInput = withDebouncedChange(AutocompleteTextInput)
```
petesantos, line 303: Performance optimization. Being a user typed field, we want to try to keep the latency down in this case to provide a good UX

src/pages/library/common.js
```diff
@@ -85,11 +87,30 @@ const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSections, sel
...
 
+const reverseText = _.flow(_.split(''), _.reverse, _.join(''))
+const truncateLeftWord = _.curry((options, text) => _.flow(
+  reverseText,
+  _.truncate(options),
+  reverseText,
+  _.replace(/\.\.\.(\S+)/, '...')
+)(text))
+
+const getContextualSuggestion = ([leftContext, match, rightContext]) => {
```
petesantos, line 103: When searching a long description, we pull out the line where the match was found to give the user some context on the match

src/pages/library/common.js
```diff
@@ -106,40 +127,63 @@ export const SearchAndFilterComponent = ({ fullList, sidebarSections, customSort
...
+    return _.flow(
+      filterBySections,
+      filterByTags,
+      filterByText,
+      customSort ? _.orderBy([customSort.field], [customSort.direction]) : _.orderBy([sort.field], [sort.direction])
+    )(fullList)
+  }, [fullList, searchFilter, customSort, sort, listDataByTag, selectedTags, selectedSections])
+
+
+  const onSearchChange = filter => {
```
petesantos, line 174: Preserve the search in the URL so we can navigate between details without losing the search when we go back

src/components/input.js
```diff
@@ -267,7 +269,7 @@ const withAutocomplete = WrappedComponent => ({
             } else if (_.includes(e.key, ['ArrowUp', 'ArrowDown']) && !suggestions.length) {
               e.nativeEvent.preventDownshiftDefault = true
             } else if (e.key === 'Enter') {
-              onPick && onPick(value)
+              onPick ? onPick(value) : toggleMenu()
```
lu-c: Should we maybe run `toggleMenu` on any enter regardless of whether the onPick function was defined?

src/pages/library/common.js
```diff
@@ -85,11 +87,30 @@ const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSections, sel
   ])
 }
 
+const reverseText = _.flow(_.split(''), _.reverse, _.join(''))
+const truncateLeftWord = _.curry((options, text) => _.flow(
+  reverseText,
+  _.truncate(options),
+  reverseText,
```
lu-c: Why do we need to reverse the text twice?

lu-c (approved):
Had some questions, nothing blocking :)

src/pages/library/common.js
petesantos: This is to get the ellipses at the front of the line. Lodash truncate will truncate the string at the end and add an ellipses. In this case I reverse and truncate to get the ellipses where I want, then reverse again to bring it back. It isn't the cleanest solution. I tried with regex, but it seemed to be noticeably slower on the page.

src/components/input.js
petesantos: I like it - the UX feels better with your suggestion - thanks!

src/components/input.js
kyuksel: My eyes found it a bit difficult to parse the nested ternary on the same line. Something like the below would help:

```suggestion
  const controlProps = itemToString ?
    { itemToString: v => v ? itemToString(v) : value } :
    { selectedItem: value }
```

src/pages/library/common.js
```diff
@@ -85,11 +87,30 @@ const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSections, sel
   ])
 }
 
+const reverseText = _.flow(_.split(''), _.reverse, _.join(''))
```
kyuksel: Do we need to `_.split`?

```suggestion
const reverseText = _.flow(_.reverse, _.join(''))
```

src/pages/library/common.js
kyuksel: How would you feel about adding a comment, perhaps with an example?

kyuksel:
LGTM overall 👍🏼  Minor comments inline.
Also, could you please
1. reference the JIRA ticket as prefix to the PR description (required for compliance reasons)
2. run `input.js` through the IntelliJ reformatter
? 

Couldn't approve because of this weirdness. Will try again later.

image (2)

src/pages/library/common.js
petesantos: I didn't realize lodash `reverse` would automatically convert the string to an array - thanks!

src/pages/library/common.js
petesantos: I will add some comments to clarify

PR #2662, opened 2021-10-22
[IA-2812] Unique Galaxy disk per workspace

gpcarr:
[Bug](https://broadworkbench.atlassian.net/browse/IA-2812): If a user has workspaces that were created before the PPW migration, it is possible for that user to create galaxy instances in different workspaces but the same google project. [Leonardo does not allow the same disk to exist in 2 different workspaces](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala#L167). so the UI needs to create a new disk per workspace (if they are in the same project pre PPW migration). 

This PR fixes the bug mentioned. Tested by using two workspaces that were created pre PPW migration: `callisto-dev/gab-test` and `callisto-dev/gab-test copy`.  Created a galaxy in `gab-test`, deleted that galaxy but kept the disk. Then created a galaxy in `gab-test copy` and validated that the UI does not attach the disk from `gab-test` and creates a new disk and validated that Leo no longer throws an error and galaxy created successfully.

<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/GalaxyModal.js
+++ b/src/components/GalaxyModal.js
@@ -46,7 +46,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
     const [loading, setLoading] = useState(false)
     const [shouldDeleteDisk, setShouldDeleteDisk] = useState(false)
 
-    const currentDataDisk = getCurrentPersistentDisk(tools.galaxy.appType, apps, appDataDisks)
+    const currentDataDisk = getCurrentPersistentDisk(tools.galaxy.appType, apps, appDataDisks, workspaceName)
 
     const createGalaxy = _.flow(
       Utils.withBusyState(setLoading),
--- a/src/libs/_tests/runtime-utils.test.js
+++ b/src/libs/_tests/runtime-utils.test.js
@@ -1,5 +1,7 @@
 import { tools } from 'src/components/notebook-utils'
-import { getCurrentApp, getCurrentAppIncludingDeleting, getCurrentPersistentDisk, getDiskAppType } from 'src/libs/runtime-utils'
+import {
+  getCurrentApp, getCurrentAppIncludingDeleting, getCurrentPersistentDisk, getDiskAppType, workspaceHasMultipleApps, workspaceHasMultipleDisks
+} from 'src/libs/runtime-utils'
 
 
 const cromwellRunning = {
@@ -66,6 +68,53 @@ const galaxyDeleting = {
 
 const mockApps = [cromwellProvisioning, cromwellRunning, galaxyRunning, galaxyDeleting]
 
+const galaxy1Workspace1 = {
+  appName: 'terra-app-69200c2f-89c3-47db-874c-b770d8de858g',
+  appType: 'GALAXY',
+  auditInfo: {
+    creator: 'cahrens@gmail.com', createdDate: '2021-12-10T20:19:13.162484Z', destroyedDate: null, dateAccessed: '2021-12-11T20:19:13.162484Z'
+  },
+  diskName: 'saturn-pd-026594ac-d829-423d-a8df-87fe07f6b5e8', // galaxyDisk1Workspace1
+  errors: [],
+  googleProject: 'terra-test-e4000484',
+  kubernetesRuntimeConfig: { numNodes: 1, machineType: 'n1-highmem-8', autoscalingEnabled: false },
+  labels: { saturnWorkspaceName: 'test-workspace'},
+  proxyUrls: { galaxy: 'https://leonardo-fiab.dsde-dev.broadinstitute.org/a-app-69200c2f-89c3-47db-874c-b770d8de737f/galaxy' },
+  status: 'RUNNING'
+}
+
+const galaxy2Workspace1 = {
+  appName: 'terra-app-69200c2f-89c3-47db-874c-b770d8de656t',
+  appType: 'GALAXY',
+  auditInfo: {
+    creator: 'cahrens@gmail.com', createdDate: '2021-12-10T20:19:13.162484Z', destroyedDate: null, dateAccessed: '2021-12-11T20:19:13.162484Z'
+  },
+  diskName: 'saturn-pd-026594ac-d829-423d-a8df-98fe18f7b6e9', // galaxyDisk2Workspace1
+  errors: [],
+  googleProject: 'terra-test-e4000484',
+  kubernetesRuntimeConfig: { numNodes: 1, machineType: 'n1-highmem-8', autoscalingEnabled: false },
+  labels: { saturnWorkspaceName: 'test-workspace'},
+  proxyUrls: { galaxy: 'https://leonardo-fiab.dsde-dev.broadinstitute.org/a-app-69200c2f-89c3-47db-874c-b770d8de737f/galaxy' },
+  status: 'RUNNING'
+}
+
+const cromwell1Workspace1 = {
+  appName: 'terra-app-69200c2f-89c3-47db-874c-b770d8de656t',
+  appType: 'GALAXY',
+  auditInfo: {
+    creator: 'cahrens@gmail.com', createdDate: '2021-12-10T20:19:13.162484Z', destroyedDate: null, dateAccessed: '2021-12-11T20:19:13.162484Z'
+  },
+  diskName: 'saturn-pd-026594ac-d829-423d-a8df-55fe36f5b4e8', // cromwellDisk1Workspace1
+  errors: [],
+  googleProject: 'terra-test-e4000484',
+  kubernetesRuntimeConfig: { numNodes: 1, machineType: 'n1-highmem-8', autoscalingEnabled: false },
+  labels: { saturnWorkspaceName: 'test-workspace'},
+  proxyUrls: { galaxy: 'https://leonardo-fiab.dsde-dev.broadinstitute.org/a-app-69200c2f-89c3-47db-874c-b770d8de737f/galaxy' },
+  status: 'RUNNING'
+}
+
+const mockAppsSameWorkspace = [galaxy1Workspace1, galaxy2Workspace1, cromwell1Workspace1]
+
 const galaxyDisk = {
   auditInfo: {
     creator: 'cahrens@gmail.com', createdDate: '2021-11-29T20:19:13.162484Z', destroyedDate: null, dateAccessed: '2021-11-29T20:19:14.114Z'
@@ -74,7 +123,7 @@ const galaxyDisk = {
   diskType: 'pd-standard',
   googleProject: 'terra-test-e4000484',
   id: 10,
-  labels: { saturnApplication: 'galaxy' }, // Note 'galaxy' vs. 'GALAXY', to represent our older naming scheme
+  labels: { saturnApplication: 'galaxy', saturnWorkspaceName: 'test-workspace' }, // Note 'galaxy' vs. 'GALAXY', to represent our older naming scheme
   name: 'saturn-pd-026594ac-d829-423d-a8df-76fe96f5b4e7',
   size: 500,
   status: 'Ready',
@@ -90,7 +139,7 @@ const galaxyDeletingDisk = {
   diskType: 'pd-standard',
   googleProject: 'terra-test-e4000484',
   id: 10,
-  labels: { saturnApplication: 'GALAXY' },
+  labels: { saturnApplication: 'GALAXY', saturnWorkspaceName: 'test-workspace' },
   name: 'saturn-pd-1236594ac-d829-423d-a8df-76fe96f5897',
   size: 500,
   status: 'Deleting',
@@ -105,7 +154,7 @@ const cromwellUnattachedDisk = {
   diskType: 'pd-standard',
   googleProject: 'terra-test-e4000484',
   id: 12,
-  labels: { saturnApplication: 'CROMWELL' },
+  labels: { saturnApplication: 'CROMWELL', saturnWorkspaceName: 'test-workspace' },
   name: 'saturn-pd-7fc0c398-63fe-4441-aea5-1e794c961310',
   size: 500,
   status: 'Ready',
@@ -121,7 +170,7 @@ const cromwellProvisioningDisk = {
   diskType: 'pd-standard',
   googleProject: 'terra-test-e4000484',
   id: 11,
-  labels: { saturnApplication: 'CROMWELL' },
+  labels: { saturnApplication: 'CROMWELL', saturnWorkspaceName: 'test-workspace' },
   name: 'saturn-pd-693a9707-634d-4134-bb3a-cbb73cd5a8ce',
   size: 500,
   status: 'Creating',
@@ -145,6 +194,68 @@ const jupyterDisk = {
 }
 
 const mockAppDisks = [galaxyDisk, galaxyDeletingDisk, cromwellProvisioningDisk, cromwellUnattachedDisk]
+
+const galaxyDisk1Workspace1 = {
+  auditInfo: {
+    creator: 'cahrens@gmail.com', createdDate: '2021-11-30T20:19:13.162484Z', destroyedDate: null, dateAccessed: '2021-12-10T20:19:14.114Z'
+  },
+  blockSize: 4096,
+  diskType: 'pd-standard',
+  googleProject: 'terra-test-e4000484',
+  id: 13,
+  labels: { saturnApplication: 'GALAXY', saturnWorkspaceName: 'test-workspace' },
+  name: 'saturn-pd-026594ac-d829-423d-a8df-87fe07f6b5e8',
+  size: 500,
+  status: 'Ready',
+  zone: 'us-central1-a'
+}
+
+const galaxyDisk2Workspace1 = {
+  auditInfo: {
+    creator: 'cahrens@gmail.com', createdDate: '2021-11-28T20:19:13.162484Z', destroyedDate: null, dateAccessed: '2021-11-29T20:19:14.114Z'
+  },
+  blockSize: 4096,
+  diskType: 'pd-standard',
+  googleProject: 'terra-test-e4000484',
+  id: 14,
+  labels: { saturnApplication: 'GALAXY', saturnWorkspaceName: 'test-workspace' },
+  name: 'saturn-pd-026594ac-d829-423d-a8df-98fe18f7b6e9',
+  size: 500,
+  status: 'Ready',
+  zone: 'us-central1-a'
+}
+
+const galaxyDisk3Workspace2 = {
+  auditInfo: {
+    creator: 'cahrens@gmail.com', createdDate: '2021-11-26T20:19:13.162484Z', destroyedDate: null, dateAccessed: '2021-11-29T20:19:14.114Z'
+  },
+  blockSize: 4096,
+  diskType: 'pd-standard',
+  googleProject: 'terra-test-e4000484',
+  id: 15,
+  labels: { saturnApplication: 'GALAXY', saturnWorkspaceName: 'test-workspace-2' },
+  name: 'saturn-pd-026594ac-d829-423d-a8df-33fe36f5b4e4',
+  size: 500,
+  status: 'Ready',
+  zone: 'us-central1-a'
+}
+
+const cromwellDisk1Workspace1 = {
+  auditInfo: {
+    creator: 'cahrens@gmail.com', createdDate: '2021-11-26T20:19:13.162484Z', destroyedDate: null, dateAccessed: '2021-11-29T20:19:14.114Z'
+  },
+  blockSize: 4096,
+  diskType: 'pd-standard',
+  googleProject: 'terra-test-e4000484',
+  id: 16,
+  labels: { saturnApplication: 'CROMWELL', saturnWorkspaceName: 'test-workspace' },
+  name: 'saturn-pd-026594ac-d829-423d-a8df-55fe36f5b4e8',
+  size: 500,
+  status: 'Ready',
+  zone: 'us-central1-a'
+}
+
+const mockAppDisksSameWorkspace = [galaxyDisk1Workspace1, galaxyDisk2Workspace1, galaxyDisk3Workspace2, cromwellDisk1Workspace1]
 
 describe('getCurrentApp', () => {
   it('returns undefined if no instances of the app exist', () => {
@@ -179,11 +290,33 @@ describe('getCurrentPersistentDisk', () => {
     expect(getCurrentPersistentDisk(tools.galaxy.appType, [cromwellProvisioning], [cromwellProvisioningDisk])).toBeUndefined()
   })
   it('returns the newest attached disk, even if app is deleting', () => {
-    expect(getCurrentPersistentDisk(tools.galaxy.appType, mockApps, mockAppDisks)).toBe(galaxyDeletingDisk)
-    expect(getCurrentPersistentDisk(tools.cromwell.appType, mockApps, mockAppDisks)).toBe(cromwellProvisioningDisk)
+    expect(getCurrentPersistentDisk(tools.galaxy.appType, mockApps, mockAppDisks, 'test-workspace')).toBe(galaxyDeletingDisk)
+    expect(getCurrentPersistentDisk(tools.cromwell.appType, mockApps, mockAppDisks, 'test-workspace')).toBe(cromwellProvisioningDisk)
   })
   it('returns the newest unattached disk that is not deleting if no app instance exists', () => {
-    expect(getCurrentPersistentDisk(tools.galaxy.appType, [], mockAppDisks)).toBe(galaxyDisk)
-    expect(getCurrentPersistentDisk(tools.cromwell.appType, [galaxyRunning], mockAppDisks)).toBe(cromwellUnattachedDisk)
+    expect(getCurrentPersistentDisk(tools.galaxy.appType, [], mockAppDisks, 'test-workspace')).toBe(galaxyDisk)
+    expect(getCurrentPersistentDisk(tools.cromwell.appType, [galaxyRunning], mockAppDisks, 'test-workspace')).toBe(cromwellUnattachedDisk)
+  })
+  it('returns a galaxy disk only if it is in the same workspace as the previous app it was attached to', () => {
+    expect(getCurrentPersistentDisk(tools.galaxy.appType, [], mockAppDisks, 'test-workspace')).toBe(galaxyDisk)
+    expect(getCurrentPersistentDisk(tools.galaxy.appType, [], mockAppDisks, 'incorrect-workspace')).toBeUndefined()
+  })
+})
+
+describe('workspaceHasMultipleApps', () => {
+  it('returns true when there are multiple galaxy apps in the same project and workspace', () => {
+    expect(workspaceHasMultipleApps(mockAppsSameWorkspace, tools.galaxy.appType)).toBe(true)
+  })
+  it('returns false when there is not multiple cromwell apps', () => {
+    expect(workspaceHasMultipleApps(mockAppsSameWorkspace, tools.cromwell.appType)).toBe(false)
+  })
+})
+
+describe('workspaceHasMultipleDisks', () => {
+  it('returns true when there are multiple galaxy disks in the same project and workspace', () => {
+    expect(workspaceHasMultipleDisks(mockAppDisksSameWorkspace, tools.galaxy.appType)).toBe(true)
+  })
+  it('returns false when there is not multiple cromwell disks', () => {
+    expect(workspaceHasMultipleDisks(mockAppDisksSameWorkspace, tools.cromwell.appType)).toBe(false)
   })
 })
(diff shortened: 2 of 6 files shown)


kyuksel:
@gpcarr Before setting out to review and test the PR, I thought I’d reproduce the error on `dev` first and to my surprise, I was able to have two `Galaxy`s running concurrently in different workspaces in the same project without any errors 🤔

kyuksel:
@gpcarr Nevermind - got clarification that the error is seen only with pre-PPW workspaces, which makes sense.

kyuksel:
With this PR's changes, I see that I can now create two `Galaxy`s in the same project, using distinct workspaces. I've noticed two issues:

1. The Cloud Environments page still displays the warning triangles for having multiple cloud envs/disks.

image

2. When I had the first Galaxy app running and was going to create the second Galaxy app, the Galaxy modal seemed to have selected the existing disk as shown below although a new disk ended up being created. Perhaps there's still something to sort out with the logic that detects an existing disk?

image

For testing, I used Rob's pre-PPW workspaces, `test-rstudio` and `rstudio test`, and the `callisto-dev` project.

src/components/GalaxyModal.js
```diff
@@ -45,7 +45,7 @@ export const GalaxyModalBase = Utils.withDisplayName('GalaxyModal')(
     const [loading, setLoading] = useState(false)
     const [shouldDeleteDisk, setShouldDeleteDisk] = useState(false)
 
-    const currentDataDisk = currentPersistentDisk(apps, galaxyDataDisks)
+    const currentDataDisk = currentPersistentDisk(apps, galaxyDataDisks, workspace.workspace.name)
```
kyuksel: Given the deconstruction

```javascript
workspace: { workspace: { namespace, bucketName, name: workspaceName, googleProject } }
```

in `GalaxyModalBase` input definition

```javascript
export const GalaxyModalBase = Utils.withDisplayName('GalaxyModal')(
  ({
    onDismiss, onSuccess, apps, galaxyDataDisks, workspace, workspace: { workspace: { namespace, bucketName, name: workspaceName, googleProject } },
    isAnalysisMode = false
  }) => {…}
```

don't we already have `workspace.workspace.name`?

```suggestion
    const currentDataDisk = currentPersistentDisk(apps, galaxyDataDisks, workspaceName)
```

src/pages/Environments.js
```diff
@@ -130,7 +130,7 @@ const Environments = () => {
     const [newRuntimes, newDisks, galaxyDisks, newApps] = await Promise.all([
       Ajax(signal).Runtimes.list({ creator }),
       Ajax(signal).Disks.list({ creator }),
-      Ajax(signal).Disks.list({ creator, saturnApplication: 'galaxy' }),
+      Ajax(signal).Disks.list({ creator, saturnApplication: 'galaxy', includeLabels: 'saturnWorkspaceName' }),
```
kyuksel: Could you remind me why we still need to make the second `Disks.list` call instead of filtering through the results of the previous `Disks.list` call?

gpcarr:
> With this PR's changes, I see that I can now create two `Galaxy`s in the same project, using distinct workspaces. I've noticed two issues:
> 
> 1. The Cloud Environments page still displays the warning triangles for having multiple cloud envs/disks.
> 
> image
> 
> 1. When I had the first Galaxy app running and was going to create the second Galaxy app, the Galaxy modal seemed to have selected the existing disk as shown below although a new disk ended up being created. Perhaps there's still something to sort out with the logic that detects an existing disk?
> 
> image
> 
> For testing, I used Rob's pre-PPW workspaces, `test-rstudio` and `rstudio test`, and the `callisto-dev` project.

Taking a look at both of these issues now and will update the PR accordingly!

kyuksel:
@gpcarr FYI for when you get back to this PR:

@cahrens's [recent PRs](https://github.com/DataBiosphere/terra-ui/pulls?q=is%3Apr+draft%3Afalse+author%3Acahrens+) have likely touched pieces you'll be updating so you may want to rebase before making any changes 😊 

image

src/pages/Environments.js
cahrens: With my PR, we will just be making a single Disk.list call. See https://github.com/DataBiosphere/terra-ui/pull/2702/files#diff-db07f65dd0ba5edbd535b2a231709c1f5901f4e1c43f39c333165d97c1abacc4R133

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -180,7 +180,7 @@ const useCloudEnvironmentPolling = googleProject => {
       const [newDisks, newRuntimes, galaxyDisks, cromwellDisks] = googleProject ? await Promise.all([
         Ajax(signal).Disks.list({ googleProject, creator: getUser().email }),
         Ajax(signal).Runtimes.list({ googleProject, creator: getUser().email }),
-        Ajax(signal).Disks.list({ googleProject, creator: getUser().email, saturnApplication: tools.galaxy.appType }),
+        Ajax(signal).Disks.list({ googleProject, creator: getUser().email, saturnApplication: tools.galaxy.appType, includeLabels: 'saturnWorkspaceName' }),
```
cahrens, line 183: Here also there will be  single call. I don't think the workspace name will ever be used though by WorkspaceContainer… why are you including the label here?

cahrens: Here also this will be just a single call after my PR merges.

src/pages/Environments.js
cahrens: I don't think though that the Environments page needs the workspace name. It does not call `currentPersistentDisk`. The Environments page already partitions the disks by Google project, using disksByProject (the project property of the disk).

In general, why use saturnWorkspaceName instead of googleProject? With project per workspace, shouldn't they be the same thing?

cahrens:
@gpcarr my PR has merged. Let me know if you have questions about some of the changes I made.

cahrens:
@gpcarr You need to rebase this branch on dev and resolve the conflicts. I changed several of the lines you were also editing (we no longer do a separate AJAX call for galaxy disks, and do not want to reintroduce that).

src/pages/Environments.js
gpcarr: This PR is to fix a bug where in workspaces that were created before the project per workspace migration cause a Leo bug. Say a user has disk A in workspace A, deletes workspace A and then creates a new Galaxy in workspace B. Leo tries to attach the same disk from workspace A to the newly created Galaxy in workspace B. In Leo, two disks can't exist in the same workspace. So the fix is that there should be a new disk that gets created for workspace B which post PPW is the case for free

gpcarr:
> @gpcarr You need to rebase this branch on dev and resolve the conflicts. I changed several of the lines you were also editing (we no longer do a separate AJAX call for galaxy disks, and do not want to reintroduce that).

I rebased yesterday! Working on fixing some code before I push that up since a lot of logic around disks was changed while I had to put this PR on pause.

src/pages/Environments.js
cahrens: Thanks @gpcarr, but I am still a bit confused about the issue. Looking at IA-2812, it seems that the issue is that 2 disks cannot exist in the same Google project. With pre-PPW workspaces, you could have 2 workspaces using the same project.

If that is the bug, why isn't it enough to just use googleProject? 

I am also not clear where the limitation is coming from (and therefore I'm unsure if this is something we need to be concerned about with Cromwell also). Can you clarify what the limitation is in Leo code itself?

cahrens: I also am not following this:

_"Say a user has disk A in workspace A, deletes workspace A and then creates a new Galaxy in workspace B. Leo tries to attach the same disk from workspace A to the newly created Galaxy in workspace B. In Leo, two disks can't exist in the same workspace. So the fix is that there should be a new disk that gets created for workspace B which post PPW is the case for free"_

This seems to be a case of a single disk being used in 2 workspaces (one of which has been deleted). Why is this a problem? How is it related to the Leo limitation that 2 disks can't exist in the same workspace (we have only 1 disk in play)?

gpcarr: This is a Leo specific limitation. See [this line](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala#L167) in Leo - we do not allow a disk to exist in two different workspaces. which is currently possible to do with pre-ppw migration workspaces. so we need to force the frontend to create a new disk if a user has 2 workspaces in the same project (these would be older ie pre-ppw workspaces)

src/libs/runtime-utils.js
```diff
@@ -272,7 +272,7 @@ export const appIsSettingUp = app => {
   return app && (app.status === 'PROVISIONING' || app.status === 'PRECREATING')
 }
 
-export const getCurrentPersistentDisk = (appType, apps, appDataDisks) => {
+export const getCurrentPersistentDisk = (appType, apps, appDataDisks, workspaceName) => {
```
cahrens, line 287: Please add unit test coverage for this argument.

src/pages/Environments.js
```diff
@@ -424,9 +424,13 @@ const Environments = () => {
               const multipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
                 disksByProject[googleProject]).length > 1
               const forAppText = !!appType ? ` for ${_.capitalize(appType)}` : ''
+              // Galaxy can have multiple disks per project as long as they are in different workspaces
+              const galaxyDisksOnly = _.filter(disk => disk.labels.saturnApplication?.toLowerCase() === tools.galaxy.appType.toLowerCase() && disk.status !== 'DELETING', filteredDisks)
```
cahrens: If you use `getDiskAppType`, it takes care of the case sensitivity issue for you. So this would become 
`_.filter(disk => getDiskAppType(disk) === tools.galaxy.appType …`

src/pages/Environments.js
```diff
@@ -424,9 +424,13 @@ const Environments = () => {
...
+              const galaxyDiskWorkspaces = _.map(currentDisk => currentDisk.labels.saturnWorkspaceName, galaxyDisksOnly)
+              const multipleGalaxyDisks = _.uniq(galaxyDiskWorkspaces).length !== galaxyDiskWorkspaces.length
               return h(Fragment, [
                 googleProject,
-                diskStatus !== 'Deleting' && multipleDisksOfType &&
+                diskStatus !== 'Deleting' && (appType === tools.galaxy.appType.toLowerCase() ? multipleGalaxyDisks : multipleDisksOfType) &&
```
cahrens: appType here should be the upper-case version ("GALAXY"), so you do not want to convert tools.galaxy.appType to lower case.

gpcarr:
> With this PR's changes, I see that I can now create two `Galaxy`s in the same project, using distinct workspaces. I've noticed two issues:
> 
> 1. The Cloud Environments page still displays the warning triangles for having multiple cloud envs/disks.
> 
> image
> 
> 2. When I had the first Galaxy app running and was going to create the second Galaxy app, the Galaxy modal seemed to have selected the existing disk as shown below although a new disk ended up being created. Perhaps there's still something to sort out with the logic that detects an existing disk?
> 
> image
> 
> For testing, I used Rob's pre-PPW workspaces, `test-rstudio` and `rstudio test`, and the `callisto-dev` project.

Hi @kyuksel,

I've addressed all your feedback at this point. The only thing I was not able to recreate is what you have listed as #2 in this comment. Let me know if/how you're still able to!

src/pages/Environments.js
gpcarr: thanks for the feedback, will update!

src/libs/runtime-utils.js
gpcarr: will do

kyuksel:
@gpcarr could you update the PR description?

kyuksel:
When I have Galaxy apps `Provisioning` (2 in `pre-PPW` and 1 in `post-PPW` workspaces), I see the warning triangle and the tooltip, which I'd think shouldn't have happened?

image

gpcarr:
> @gpcarr could you update the PR description?

just updated it

src/pages/Environments.js
```diff
@@ -425,9 +425,13 @@ const Environments = () => {
               const multipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
                 disksByProject[googleProject]).length > 1
               const forAppText = !!appType ? ` for ${_.capitalize(appType)}` : ''
+              // Galaxy can have multiple disks per project as long as they are in different workspaces
+              const galaxyDisksOnly = _.filter(disk => getDiskAppType(disk) === tools.galaxy.appType && disk.status !== 'DELETING', filteredDisks)
```
kyuksel: Looks like we had already gotten the app type. Can we use that instead of getting it again?
```suggestion
              const galaxyDisksOnly = _.filter(disk => appType === tools.galaxy.appType && disk.status !== 'DELETING', filteredDisks)
```

src/pages/Environments.js
```diff
@@ -425,9 +425,13 @@ const Environments = () => {
...
+              const galaxyDiskWorkspaces = _.map(currentDisk => currentDisk.labels.saturnWorkspaceName, galaxyDisksOnly)
+              const multipleGalaxyDisks = _.uniq(galaxyDiskWorkspaces).length !== galaxyDiskWorkspaces.length
               return h(Fragment, [
                 googleProject,
-                diskStatus !== 'Deleting' && multipleDisksOfType &&
+                diskStatus !== 'Deleting' && (appType === tools.galaxy.appType ? multipleGalaxyDisks : multipleDisksOfType) &&
                 h(TooltipTrigger, {
                   content: `This billing project has multiple active persistent disks${forAppText}. Only the latest one will be used.`
                 }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
```
kyuksel, line 462: Within the `cellRenderer` function, while evaluating whether we have multiple disks, we seem to be running both of the evaluation logic (one for Galaxy apps and one for non-Galaxy apps). This should be unnecessary since an app cannot be Galaxy and non-Galaxy at the same time. In other words we could move the conditional from the rendering code to higher up and avoid the unnecessary computations, like so:

```javascript
              const { status: diskStatus, googleProject } = filteredDisks[rowIndex]
              const appType = getDiskAppType(filteredDisks[rowIndex])
              const haveMultipleDisks = () => {
                if (appType === tools.galaxy.appType) {
                  // Galaxy can have multiple disks per project as long as they are in different workspaces
                  const galaxyDisksOnly = _.filter(disk => appType === tools.galaxy.appType && disk.status !== 'DELETING', filteredDisks)
                  const galaxyDiskWorkspaces = _.map(currentDisk => currentDisk.labels.saturnWorkspaceName, galaxyDisksOnly)
                  return _.uniq(galaxyDiskWorkspaces).length !== galaxyDiskWorkspaces.length
                } else {
                  return _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting', disksByProject[googleProject]).length > 1
                }
              }

              const forAppText = !!appType ? ` for ${_.capitalize(appType)}` : ''

              return h(Fragment, [
                googleProject,
                diskStatus !== 'Deleting' && haveMultipleDisks() &&
                h(TooltipTrigger, {
                  content: `This billing project has multiple active persistent disks${forAppText}. Only the latest one will be used.`
                }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
              ])
            }
```

What do you think?

kyuksel: I have made the PR #2750 into your branch to better illustrate the change above and also pull the logic into its own function, which would also make it possible to write a unit test for it.

src/pages/Environments.js
cahrens: @gpcarr @kyuksel the Leo line of code linked above is not specific to Galaxy-- it is used by both Cromwell and Galaxy (see line 154). Therefore I think these changes should be made general for all apps, not Galaxy-specific. I have some questions about what a Custom app instance would do, but I don't think that is exposed to users, and I doubt the persistent disk logic is implemented for Custom. For simplicity, I think it would be best to make the restriction consistent for all apps until we have a use case where it would not be wanted.

src/components/GalaxyModal.js
```diff
@@ -46,7 +46,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
     const [loading, setLoading] = useState(false)
     const [shouldDeleteDisk, setShouldDeleteDisk] = useState(false)
 
-    const currentDataDisk = getCurrentPersistentDisk(tools.galaxy.appType, apps, appDataDisks)
+    const currentDataDisk = getCurrentPersistentDisk(tools.galaxy.appType, apps, appDataDisks, workspaceName)
```
cahrens, line 49: To make Cromwell consistent with Galaxy, it will be necessary to also make this change in CromwellModal.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -183,7 +183,11 @@ describe('getCurrentPersistentDisk', () => {
     expect(getCurrentPersistentDisk(tools.cromwell.appType, mockApps, mockAppDisks)).toBe(cromwellProvisioningDisk)
   })
   it('returns the newest unattached disk that is not deleting if no app instance exists', () => {
-    expect(getCurrentPersistentDisk(tools.galaxy.appType, [], mockAppDisks)).toBe(galaxyDisk)
+    expect(getCurrentPersistentDisk(tools.galaxy.appType, [], mockAppDisks, 'test-workspace')).toBe(galaxyDisk)
```
cahrens, line 297: I believe we will want to make the workspace name a required variable for getCurrentPersistentDisk (update all test usages as well).

src/pages/Environments.js
cahrens: I think this check should not be Galaxy-specific.

src/pages/Environments.js
cahrens: (I admit though that I still don't understand **why** Leo is enforcing this restriction)

kyuksel:
@gpcarr Posting here for you, I and @cahrens to get on the same page regarding the next steps about Christina's [PR comment](https://github.com/DataBiosphere/terra-ui/pull/2662/files#r778847503) on removing app specificity. You're welcome to take your time to address it but we're also available for mobbing, etc. if you feel it'd help at some point 😊

src/libs/_tests/runtime-utils.test.js
```diff
@@ -74,7 +76,52 @@ const galaxyDisk = {
...
+  labels: { saturnApplication: 'galaxy', saturnWorkspaceName: 'test-workspace' }, // Note 'galaxy' vs. 'GALAXY', to represent our older naming scheme
+  name: 'saturn-pd-026594ac-d829-423d-a8df-76fe96f5b4e7',
+  size: 500,
+  status: 'Ready',
+  zone: 'us-central1-a'
+}
+
+const galaxyDisk2 = {
+  auditInfo: {
+    creator: 'cahrens@gmail.com', createdDate: '2021-11-29T20:19:13.162484Z', destroyedDate: null, dateAccessed: '2021-11-29T20:19:14.114Z'
```
cahrens: Please make sure the dates are unique and don't impact the existing tests. The dates of disks are used in some of our logic.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -74,7 +76,52 @@ const galaxyDisk = {
...
+  },
+  blockSize: 4096,
+  diskType: 'pd-standard',
+  googleProject: 'terra-test-e4000484-2',
+  id: 10,
+  labels: { saturnApplication: 'galaxy', saturnWorkspaceName: 'test-workspace' }, // Note 'galaxy' vs. 'GALAXY', to represent our older naming scheme
```
cahrens: In general, `saturnApplication` should be "GALAXY". I had one with "galaxy" to represent a certain condition (which is why I had the comment after it), but we don't want to litter that all of the place because it makes it more confusing.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -74,7 +76,52 @@ const galaxyDisk = {
...
+  name: 'saturn-pd-026594ac-d829-423d-a8df-76fe96f5b4e7',
```
cahrens: I think the mocks should be realistic, so please make sure that names are unique.

src/libs/runtime-utils.js
```diff
@@ -268,11 +268,21 @@ export const getDiskAppType = disk => {
   return appType
 }
 
+export const projectHasMultipleDisks = disks => {
```
cahrens: I would add documentation to this method to clarify that it does not check the app types, and therefore it is expected that the disks passed in have already been filtered by app type. And I would suggest making the method name `workspaceHasMultipleDisks`.

src/pages/Environments.js
```diff
@@ -203,11 +203,13 @@ const Environments = () => {
   const appsByProject = _.groupBy('googleProject', apps)
 
   const renderBillingProjectApp = app => {
+    const appsByType = _.filter(currentApp => currentApp.appType === app.appType && !_.includes(currentApp.status, ['DELETING', 'PREDELETING']), appsByProject[app.googleProject])
+    const multipleApps = projectHasMultipleApps(appsByType)
     const inactive = !_.includes(app.status, ['DELETING', 'ERROR', 'PREDELETING']) &&
```
cahrens: Is inactive still needed given line 206?

src/libs/runtime-utils.js
```diff
@@ -268,11 +268,21 @@ export const getDiskAppType = disk => {
...
+  const diskWorkspaces = _.map(currentDisk => currentDisk.labels.saturnWorkspaceName, disks)
+  return _.uniq(diskWorkspaces).length < diskWorkspaces.length
+}
+
+export const projectHasMultipleApps = apps => {
```
cahrens: Same comment about documentation and name as above.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -179,11 +226,24 @@ describe('getCurrentPersistentDisk', () => {
...
+  })
+  it('returns a galaxy disk only if it is in the same workspace as the previous app it was attached to', () => {
+    expect(getCurrentPersistentDisk(tools.galaxy.appType, [], mockAppDisks, 'test-workspace')).toBe(galaxyDisk)
+    expect(getCurrentPersistentDisk(tools.galaxy.appType, [], mockAppDisks, 'incorrect-workspace')).toBeUndefined()
+  })
+})
+
+describe('projectHasMultipleDisks', () => {
+  it('returns true when there are multiple disks', () => {
+    expect(projectHasMultipleDisks(mockAppDisks + galaxyDisk3)).toBe(true)
```
cahrens: Note this is confusing because mockAppDisks contains disks related to different app types.

I really would be tempted to have `workspaceHasMultipleDisks` do the app filtering itself (so you would pass in the appType as a variable) because I think it makes the method much easier to understand.

src/libs/runtime-utils.js
cahrens: Please add unit test coverage for this as well.

Also, the implementation is the same for the 2 methods, so you could consider refactoring them to share the implementation, as long as you can make the contract of what is passed in clear.

(Note though that I'm somewhat in favor of making the methods do the appType filtering, which if you go down that route could make sharing the code more complicated.)

src/libs/_tests/runtime-utils.test.js
gpcarr: I like the idea of changing the method to do the app filtering, i'll update!

kyuksel:
Warning triangles seem to still appear when I create Galaxy apps in distinct workspaces that belong to the same billing project?

image

gpcarr:
> Warning triangles seem to still appear when I create Galaxy apps in distinct workspaces that belong to the same billing project?
> 
> image

I'm not seeing that on my end. Are you on this branch with the latest changes?

kyuksel:
> I'm not seeing that on my end. Are you on this branch with the latest changes?

Yes. I think the warnings were due to the disks in `Deleting` status left over from error'ed apps. The warnings went away once I marked the disks `Deleted` in Leo DB. This shouldn't have been happening though, right, since we're supposed to exclude `Deleting` disks from consideration? I wonder if we can replicate the issue in a unit test.

src/pages/Environments.js
```diff
@@ -313,6 +312,11 @@ const Environments = () => {
     })
   }
 
+  const multipleDisksError = (disks, appType) => {
+    return appType ? workspaceHasMultipleDisks(disks, appType) : _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
```
kyuksel: We use `!!` to be explicit that we're coercing a non-boolean (`appType`) onto a boolean.

```suggestion
    return !!appType ? workspaceHasMultipleDisks(disks, appType) : _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
```

kyuksel: Would `getDiskAppType(disk) !== appType` ever evaluate to `true` since it's in the second branch of the ternary where `!!appType` would be `false`?

cahrens: I think the code is iterating over all the disks, and we want to remove the ones that aren't associated with this specific appType.

kyuksel: Right. I was trying to understand, for the case where `!!appType` is `false` (meaning, `appType` is `undefined`?), how `getDiskAppType(disk) !== appType` could be `true`. Looking more closely, it looks like `getDiskAppType(disk)` returns `undefined` if `disk` is unattached. In other words, `getDiskAppType(disk) !== appType` is necessary to filter out unattached disks?

cahrens: Not just unattached disks, but disks related to a different app. Like Galaxy vs. Cromwell.

kyuksel: `_.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting'…` part of the ternary would not be executed if a disk has `appType === Galaxy/Cromwell`.

gpcarr: plan on discussing this comment on a call at noon today with @kyuksel if you'd like to join @cahrens

kyuksel: For posterity: Talked to Gabriela offline and sorted out my misunderstanding.

gpcarr:
@kyuksel @cahrens all feedback has been addressed, looking for a final review/thumb :) thanks!

src/pages/Environments.js
```diff
@@ -219,10 +219,9 @@ const Environments = () => {
     ])
   }
   const renderWorkspaceForApps = app => {
-    const { status, appType, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = app
-    const shouldWarn = !_.includes(status, ['DELETING', 'ERROR', 'PREDELETING']) &&
-      getCurrentApp(appType)(appsByProject[googleProject]) !== app
-    return getWorkspaceCell(saturnWorkspaceNamespace, saturnWorkspaceName, appType, shouldWarn)
+    const { appType, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = app
+    const multipleApps = workspaceHasMultipleApps(appsByProject[app.googleProject], app.appType)
```
cahrens: I think the argument can be just `appType` instead of `app.appType`.

src/pages/Environments.js
```diff
@@ -446,12 +450,12 @@ const Environments = () => {
             cellRenderer: ({ rowIndex }) => {
               const { status: diskStatus, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = filteredDisks[rowIndex]
               const appType = getDiskAppType(filteredDisks[rowIndex])
-              const multipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
-                disksByProject[googleProject]).length > 1
+              const forAppText = !!appType ? ` for ${_.capitalize(appType)}` : ''
```
cahrens: I believe `forAppText` is defined elsewhere now.

gpcarr: just updated this!

src/pages/Environments.js
```diff
@@ -321,6 +320,11 @@ const Environments = () => {
     })
   }
 
+  const multipleDisksError = (disks, appType) => {
+    return !!appType ? workspaceHasMultipleDisks(disks, appType) : _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
```
cahrens, line 326: I think a comment would be good to explain that we expect appType to be `undefined` in the case of runtimes (and therefore the second part of the conditional is filtering out disks that are app disks instead of runtime disks).

gpcarr: i'll add one

cahrens (approved):
The code LGTM. I'm not able to test the original issue (multiple disks in the same workspace), so I'm leaving that to others.

I love how the code got simpler in each iteration! And thanks for the tests.

kyuksel (approved):
Thanks for addressing our comments 😊

PR #2665, opened 2021-10-25
[PF-1071] Tweaks to clone workspace warning from UX feedback

wnojopra:
Now looks like:
image

### DIFF
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -24,6 +24,7 @@ const warningStyle = {
   borderRadius: '5px',
   backgroundColor: colors.warning(0.15),
   display: 'flex',
+  lineHeight: '22px',
   padding: '1rem 1rem', margin: '0.5rem 0 1rem',
   fontWeight: 'normal', fontSize: 14
 }
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -24,7 +24,7 @@ const warningStyle = {
   borderRadius: '5px',
   backgroundColor: colors.warning(0.15),
   display: 'flex',
-  lineHeight: '22px',
+  lineHeight: '18px',
   padding: '1rem 1rem', margin: '0.5rem 0 1rem',
   fontWeight: 'normal', fontSize: 14
 }
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -1,6 +1,6 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
-import { div, h, p, span } from 'react-hyperscript-helpers'
+import { div, h, p, strong } from 'react-hyperscript-helpers'
 import { ButtonPrimary, IdContainer, Link, Select, spinnerOverlay } from 'src/components/common'
 import { icon } from 'src/components/icons'
 import { TextArea, ValidatedInput } from 'src/components/input'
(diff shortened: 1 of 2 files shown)


src/components/NewWorkspaceModal.js
```diff
@@ -210,6 +222,17 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
...
         })
       ])]),
+      shouldShowDifferentRegionWarning() && div({ style: { ...warningStyle } }, [
+        icon('warning-standard', { size: 24, style: { color: colors.warning(), flex: 'none', marginRight: '0.5rem' } }),
+        div({ style: { flex: 1 } }, [
+          `Copying data from `,
+          span({ style: { fontWeight: 600 } }, [`${getRegionInfo(sourceWorkspaceLocation, sourceLocationType).regionDescription} `]),
+          `to `,
+          span({ style: { fontWeight: 600 } }, [`${getRegionInfo(bucketLocation, destLocationType).regionDescription} `]),
+          `may incur network egress charges. `,
```
petesantos: For A11y, it is best to use a `strong` tag rather than manually embolden the text.
Also, there are a few strings that do not need interpolation.

```suggestion
          strong([getRegionInfo(sourceWorkspaceLocation, sourceLocationType).regionDescription]),
          ' to ',
          strong([getRegionInfo(bucketLocation, destLocationType).regionDescription]),
          ' may incur network egress charges. ',
```

src/components/region-common.js
```diff
@@ -11,7 +11,9 @@ import * as Utils from 'src/libs/utils'
 // https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf
 export const unknownRegionFlag = '❓'
 export const getRegionInfo = (location, locationType) => {
-  const regionDescription = `${locationType}: ${location}`.toLowerCase()
+  const regionDescription = locationType === locationTypes.multiRegion ?
+    `${location} (${locationTypes.multiRegion})` :
+    `${location}`.toLowerCase()
```
petesantos: With the simpler string, interpolation is no longer needed
```suggestion
    location.toLowerCase()
```

petesantos (commented):
Looks good - see comments on A11y.

src/components/NewWorkspaceModal.js

src/components/region-common.js
wnojopra: Thanks! What this code ended up doing here was checking for `location`. If `location` is undefined, `location.toLowerCase()` throws an error, while `` `${location}`.toLowerCase() `` does not. 

I've fixed this to not use interpolation and also check for `location`.

src/components/NewWorkspaceModal.js
```diff
@@ -15,11 +15,20 @@ import Events from 'src/libs/events'
...
 
+const warningStyle = {
+  border: `1px solid ${colors.warning(0.8)}`,
+  borderRadius: '5px',
+  backgroundColor: colors.warning(0.15),
+  display: 'flex',
+  lineHeight: '18px',
+  padding: '1rem 1rem', margin: '0.5rem 0 1rem',
+  fontWeight: 'normal', fontSize: 14
+}
```
kyuksel, line 30: Can we re-use some of [warningStyle in style.js](https://github.com/DataBiosphere/terra-ui/blob/f574a99482a57e40c366f5bbf7fab7dd00636b70/src/libs/style.js#L155) and override where necessary?

kyuksel (approved):
I love the new look! One minor comment inline.

Manually tested locally.

src/components/NewWorkspaceModal.js
wnojopra: Literally the only thing that is similar now is the backgroundColor, so I didn't think it worth it. Is it still worth it just for the backgroundColor?

kyuksel: `border` is the same too but yea it may not be worth it in this case on a second thought.

petesantos (approved):

PR #2669, opened 2021-10-27
[DC-15] UI Update to "Import Data" page to show list of datasets if (instead of just one)

lu-c:
Before:
Screen Shot 2021-10-27 at 6 49 32 PM

After:
Screen Shot 2021-10-27 at 6 49 25 PM

### DIFF
--- a/src/libs/nav.js
+++ b/src/libs/nav.js
@@ -51,14 +51,13 @@ export const Redirector = ({ pathname, search }) => {
   return null
 }
 
-const parseRoute = (handlers, { pathname, search, state }) => {
+const parseRoute = (handlers, { pathname, search }) => {
   const handler = _.find(({ regex }) => regex.test(pathname), handlers)
   console.assert(handler, 'No handler found for path')
   return handler && {
     ...handler,
     params: _.zipObject(handler.keys, _.tail(handler.regex.exec(pathname))),
-    query: qs.parse(search, { ignoreQueryPrefix: true, plainObjects: true }),
-    state: state || {}
+    query: qs.parse(search, { ignoreQueryPrefix: true, plainObjects: true })
   }
 }
 
--- a/src/libs/nav.js
+++ b/src/libs/nav.js
@@ -51,13 +51,14 @@ export const Redirector = ({ pathname, search }) => {
   return null
 }
 
-const parseRoute = (handlers, { pathname, search }) => {
+const parseRoute = (handlers, { pathname, search, state }) => {
   const handler = _.find(({ regex }) => regex.test(pathname), handlers)
   console.assert(handler, 'No handler found for path')
   return handler && {
     ...handler,
     params: _.zipObject(handler.keys, _.tail(handler.regex.exec(pathname))),
-    query: qs.parse(search, { ignoreQueryPrefix: true, plainObjects: true })
+    query: qs.parse(search, { ignoreQueryPrefix: true, plainObjects: true }),
+    state: state || {}
   }
 }
 
--- a/src/pages/ImportData.js
+++ b/src/pages/ImportData.js
@@ -89,8 +89,6 @@ const ImportData = () => {
     loadTemplateWorkspaces()
   })
 
-  console.log('snapshots', snapshots)
-
   const onImport = _.flow(
     Utils.withBusyState(setIsImporting),
     withErrorReporting('Import Error')
--- a/src/pages/ImportData.js
+++ b/src/pages/ImportData.js
@@ -130,10 +130,21 @@ const ImportData = () => {
           div({ style: { marginTop: 20, marginBottom: 60 } }, [
             'Dataset(s):',
             ul({ style: { listStyle: 'none', position: 'relative', marginLeft: 0, paddingLeft: '2rem' } }, [
-              _.map.convert({ cap: false })(({ title, id }, mapindex) => li({ key: `snapshot_${id}`, style: { fontSize: 16, fontWeight: 'bold', marginTop: 20, paddingTop: mapindex ? 20 : 0, borderTop: `${mapindex ? 1 : 0}px solid #AAA` } }, [
-                icon('success-standard', { size: 18, style: { position: 'absolute', left: 0, color: colors.primary() } }),
-                title
-              ]), snapshots)
+              _.flow(
+                Utils.toIndexPairs,
+                _.map(([mapindex, { title, id }]) => li({
+                  key: `snapshot_${id}`,
+                  style: {
+                    fontSize: 16,
+                    fontWeight: 'bold',
+                    marginTop: 20,
+                    paddingTop: mapindex ? 20 : 0,
+                    borderTop: `${mapindex ? 1 : 0}px solid #AAA`
+                  }
+                }, [
+                  icon('success-standard', { size: 18, style: { position: 'absolute', left: 0, color: colors.primary() } }),
+                  title
+                ])))(snapshots)
             ])
           ]) :
           div({ style: { fontSize: 16 } }, ['From: ', new URL(url).hostname]),
--- a/src/pages/ImportData.js
+++ b/src/pages/ImportData.js
@@ -1,6 +1,6 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
-import { div, h, h2, img, p } from 'react-hyperscript-helpers'
+import { div, h, h2, img, li, p, ul } from 'react-hyperscript-helpers'
 import Collapse from 'src/components/Collapse'
 import { backgroundLogo, ButtonPrimary, ButtonSecondary, Clickable, IdContainer, RadioButton, spinnerOverlay } from 'src/components/common'
 import { notifyDataImportProgress } from 'src/components/data/data-utils'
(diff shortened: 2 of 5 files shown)


src/libs/nav.js
```diff
@@ -51,13 +51,14 @@ export const Redirector = ({ pathname, search }) => {
...
-const parseRoute = (handlers, { pathname, search }) => {
+const parseRoute = (handlers, { pathname, search, state }) => {
   const handler = _.find(({ regex }) => regex.test(pathname), handlers)
   console.assert(handler, 'No handler found for path')
   return handler && {
     ...handler,
     params: _.zipObject(handler.keys, _.tail(handler.regex.exec(pathname))),
-    query: qs.parse(search, { ignoreQueryPrefix: true, plainObjects: true })
+    query: qs.parse(search, { ignoreQueryPrefix: true, plainObjects: true }),
+    state: state || {}
```
lu-c: I vaguely remember having a conversation with someone about how we should not be junking up the URL with too much data, and there was a concern that linking to multiple snapshots (with their long titles) would be unwieldy... So I decided to resolve this by throwing it into the state variable.

My problem with this solution is that when you refresh the page, all of this state stuff disappears, and you also cant share the url with someone...

Happy to hear any suggestions for how to resolve this!

src/pages/ImportData.js
```diff
@@ -106,7 +108,9 @@ const ImportData = () => {
         notify('success', 'Data imported successfully.', { timeout: 3000 })
       }],
       ['snapshot', async () => {
-        await Ajax().Workspaces.workspace(namespace, name).importSnapshot(snapshotId, snapshotName)
+        supportMultipleImports ?
+          await Promise.allSettled(_.map(({ title, id }) => Ajax().Workspaces.workspace(namespace, name).importSnapshot(id, title), snapshots)) :
```
lu-c, line 122: The existing "Import Data" page would only ever import the first snapshot selected, and completely ignored any of the other selected objects. Fixed this bug in this pr so they should all be added now.

src/libs/nav.js
petesantos: How would you feel about storing the id's and title's in local storage, and only sending a list of id's as parameters.
If we do not find the title's in local storage, we could fetch the metadata with the list of ID's from the query param.

This also comes down to the use cases. Is this page something that users may bookmark or share? Is there a limit on the number of snapshot's they will request for import?

src/pages/ImportData.js
```diff
@@ -119,12 +121,22 @@ const ImportData = () => {
...
+        supportMultipleImports ?
+          div({ style: { marginTop: 20, marginBottom: 60 } }, [
+            'Dataset(s):',
+            ul({ style: { listStyle: 'none', position: 'relative', marginLeft: 0, paddingLeft: '2rem' } }, [
+              _.map.convert({ cap: false })(({ title, id }, mapindex) => li({ key: `snapshot_${id}`, style: { fontSize: 16, fontWeight: 'bold', marginTop: 20, paddingTop: mapindex ? 20 : 0, borderTop: `${mapindex ? 1 : 0}px solid #AAA` } }, [
+                icon('success-standard', { size: 18, style: { position: 'absolute', left: 0, color: colors.primary() } }),
+                title
+              ]), snapshots)
+            ])
+          ]) :
```
petesantos: We have a util that will get the index for an item when using map: `toIndexPairs`. The only other changes here are formatting:

```suggestion
          div({ style: { marginTop: 20, marginBottom: 60 } }, [
            'Dataset(s):',
            ul({ style: { listStyle: 'none', position: 'relative', marginLeft: 0, paddingLeft: '2rem' } }, [
              _.flow(
                Utils.toIndexPairs,
                _.map(([mapindex, { title, id }]) => li({
                  key: `snapshot_${id}`,
                  style: {
                    fontSize: 16,
                    fontWeight: 'bold',
                    marginTop: 20,
                    paddingTop: mapindex ? 20 : 0,
                    borderTop: `${mapindex ? 1 : 0}px solid #AAA`
                  }
                }, [
                  icon('success-standard', { size: 18, style: { position: 'absolute', left: 0, color: colors.primary() } }),
                  title
                ])))(snapshots)
            ])
          ]) :
```

petesantos (commented):
Had some thoughts on the multi import / page refresh issue, other than that this is looking good.

src/libs/nav.js
lu-c: 1. how about the other page variables title and header? I don't think those should be passed as a query parameters. The page already accepts a query parameter of snapshotId (before we started touching this code), and since this is being used by other people, I don't want to change how that behavior works... so I won't be removing the snapshotId and snapshotName query parameters. What would you think of adding a new one like 'snapshotIds' (plural)
2. I don't think importing the same snapshots would be a bookmarked feature, it seems like kind of a one time feature. Do you agree?
3. Can users share workspaces? If so, that might be more useful than sharing the import page
3. There is no limit on the number of snapshots they *can* request for import at the moment

petesantos: I like the idea of adding a new param and leaving the others untouched. This may allow us to have a cleaner url, and still have enough data to display the relevant information.

> I don't think importing the same snapshots would be a bookmarked feature, it seems like kind of a one time feature. Do you agree?

For the most part, yes. I was thinking of a case where another user wants the same data, imported to different workspaces over time. But it is unlikely they would know to bookmark the URL for this purpose.- and it is more of an edge case.

> Can users share workspaces? If so, that might be more useful than sharing the import page
Yes - that is the typical flow.

Either way, if we load the title data from localstorage or via a fetch when it is not available it should work with a bookmark.

lu-c: > Either way, if we load the title data from localstorage or via a fetch when it is not available it should work with a bookmark.
If we stored it in localstorage, would we have an issue here where the title and header show our header/title ("catalog, link datasets") if the user got to the same page from a different link?

Re: Loading in data and storing it in localstorage
Would this solution use the `useDataCatalog` method? if the values are NOT there? And if it does, we're going to have to add logic in the application to explicitly pull out the field "dct:title". Is this okay? Would we be tying this solution down too closely to our structure of a snapshot?

lu-c: What do you think about the current solution btw? Using state? Do you think it's bad?

dmohs: I think your instinct to avoid breaking refresh is a good one, so I'd lean toward putting the IDs in the URL. If you are worried about the implementation being coupled to the data catalog, just make that coupling explicit; for example, you could call the parameter `dcSnapshotIds`.

dmohs (approved):

src/libs/nav.js
```diff
@@ -51,13 +51,14 @@ export const Redirector = ({ pathname, search }) => {
   return null
 }
 
-const parseRoute = (handlers, { pathname, search }) => {
+const parseRoute = (handlers, { pathname, search, state }) => {
```
zarsky-broad: So, is this unused? Should we revert this file?

src/pages/ImportData.js
```diff
@@ -60,16 +61,29 @@ const ChoiceButton = ({ iconName, title, detail, style, ...props }) => {
...
   const [isCreateOpen, setIsCreateOpen] = useState(false)
   const [isCloneOpen, setIsCloneOpen] = useState(false)
   const [selectedWorkspaceId, setSelectedWorkspaceId] = useState(wid)
   const [selectedTemplateWorkspaceKey, setSelectedTemplateWorkspaceKey] = useState()
   const [allTemplates, setAllTemplates] = useState()
 
+  const { dataCatalog } = useDataCatalog()
+  const snapshots = _.flow(
+    _.filter(snapshot => _.includes(snapshot['dct:identifier'], snapshotIds)),
+    _.map(snapshot => { return { id: snapshot['dct:identifier'], title: snapshot['dct:title'] } })
```
zarsky-broad:
```suggestion
    _.map(snapshot => ({ id: snapshot['dct:identifier'], title: snapshot['dct:title'] }))
```

src/libs/nav.js
lu-c: This is unused, we can revert it if that's the standard for this repo!

src/pages/ImportData.js
```diff
@@ -60,16 +61,29 @@ const ChoiceButton = ({ iconName, title, detail, style, ...props }) => {
...
+
   const isDataset = format !== 'snapshot'
   const noteMessage = 'Note that the import process may take some time after you are redirected into your destination workspace.'
+  let title = `Import ${isDataset ? 'Data' : 'Snapshot'}`
+  let header = `Importing ${isDataset ? 'Data' : `Snapshot ${snapshotName}`}`
+
+  if (referrer === 'data-catalog') {
+    title = 'Catalog'
+    header = 'Linking data to a workspace'
+  }
```
petesantos: Would this work? It will prevent the need for mutation
```suggestion
  const [title, header] = Utils.cond(
    [referrer === 'data-catalog', () => ['Catalog', 'Linking data to a workspace']],
    [isDataset, () => ['Data', 'Data']],
    [Utils.DEFAULT, () => ['Snapshot', `Snapshot ${snapshotName}`]]
  )
```

petesantos (approved):
I have a suggestion to eliminate the mutation. Other than that, lgtm.

PR #2671, opened 2021-10-28
Updated to add an enum for snapshot access types

lu-c:
I was working on making changes to the request access button, and it wasn't working because it was case sensitive... so to try to address that problem, i made the value an enum.

### DIFF
--- a/src/pages/library/DataBrowser.js
+++ b/src/pages/library/DataBrowser.js
@@ -13,7 +13,7 @@ import { getConfig } from 'src/libs/config'
 import * as Nav from 'src/libs/nav'
 import * as Utils from 'src/libs/utils'
 import { commonStyles, SearchAndFilterComponent } from 'src/pages/library/common'
-import { useDataCatalog } from 'src/pages/library/dataBrowser-utils'
+import { SNAPSHOT_ACCESS_TYPES, useDataCatalog } from 'src/pages/library/dataBrowser-utils'
 import { RequestDatasetAccessModal } from 'src/pages/library/RequestDatasetAccessModal'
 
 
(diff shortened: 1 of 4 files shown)


src/pages/library/dataBrowser-utils.js
```diff
@@ -5,6 +5,11 @@ import { withErrorReporting } from 'src/libs/error'
 import { dataCatalogStore } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
+export const SNAPSHOT_ACCESS_TYPES = {
```
petesantos: What are your thoughts on naming the top level item in camelcase? Accessing the constants would look like:
`snapshotAccessTypes.CONTROLLED`. 

We don't have a strong precedent in the codebase, but it looks like most of our enums are all camelcase.
For constants, I prefer to hold off on the capital snake case until the end as I think it looks less noisy in the code and makes it clearer when identifying the actual constant vs namespacing.

src/pages/library/DataBrowser.js
```diff
@@ -42,33 +42,18 @@ const getUnique = (prop, data) => _.flow(
...
-          icon('lock', { style: { color: styles.access.pending, marginRight: 5 } }),
-          div(['Pending'])
-        ])
-      ]
-    }
+    labels: accessArray,
+    labelDisplays: _.zipObject(accessArray, _.map(accessKey => {
+      const lowerKey = _.toLower(accessKey)
+      return [div({ key: `access-filter-${lowerKey}`, style: { display: 'flex' } }, [
+        icon(SNAPSHOT_ACCESS_TYPES[accessKey] === SNAPSHOT_ACCESS_TYPES.OPEN ? 'unlock' : 'lock', { style: { color: styles.access[lowerKey], marginRight: 5 } }),
```
petesantos: Formatting only:
```suggestion
        icon(SNAPSHOT_ACCESS_TYPES[accessKey] === SNAPSHOT_ACCESS_TYPES.OPEN ? 'unlock' : 'lock', {
          style: { color: styles.access[lowerKey], marginRight: 5 }
        }),
```

petesantos (approved):
Looks good, minor nit on the casing for the constants

src/pages/library/dataBrowser-utils.js
lu-c: I always like using capital snake case for constants, and camel case for variables and functions, made the change tho, cause consistency in the project :)

tcjiang9 (approved):

PR #2673, opened 2021-10-29
[CA-1141] Add the ability to remove billing account and clean up

s-rubenstein:
Hi! I wanted to get this in given the Prod incident that happened, this should make it a lot easier for our billing project owners to re-remove their billing accounts, which seems like something that given the prod incident we should fix. This adds a red icon as shown. I have not run it by UX, but it seems in line with what we are already doing, so I figure we can try this and UX can weigh in with follow ups. (Just so this goes ASAP)
Screen Shot 2021-10-29 at 11 53 45 AM

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -468,6 +468,12 @@ const Billing = signal => ({
     return res
   },
 
+  removeBillingAccount: async ({ billingProjectName }) => {
+    const res = await fetchOrchestration(`api/billing/v2/${billingProjectName}/billingAccount`,
+      _.merge(authOpts(), { signal, method: 'DELETE' }))
+    return res
+  },
+
   updateSpendConfiguration: async ({ billingProjectName, datasetGoogleProject, datasetName }) => {
     const res = await fetchOrchestration(`api/billing/v2/${billingProjectName}/spendReportConfiguration`,
       _.mergeAll([
@@ -1166,19 +1172,6 @@ const GoogleBilling = signal => ({
     const response = await fetchGoogleBilling(`${billingAccountName}/projects`, _.merge(authOpts(), { signal }))
     const json = await response.json()
     return _.map('projectId', json.projectBillingInfo)
-  },
-  getBillingInfo: async project => {
-    const response = await fetchGoogleBilling(`projects/${project}/billingInfo`, _.merge(authOpts(), { signal }))
-    return response.json()
-  },
-  changeBillingAccount: async ({ projectId, newAccountName }) => {
-    const name = `projects/${projectId}/billingInfo`
-    const response = await fetchGoogleBilling(name,
-      _.mergeAll([
-        authOpts(), { signal, method: 'PUT' },
-        jsonBody({ billingEnabled: true, billingAccountName: newAccountName, name, projectId })
-      ]))
-    return response.json()
   }
 })
 
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -201,6 +201,7 @@ const ProjectDetail = ({ billingProject, reloadBillingProject, billingAccounts, 
   const [deletingUser, setDeletingUser] = useState(false)
   const [updating, setUpdating] = useState(false)
   const [showBillingModal, setShowBillingModal] = useState(false)
+  const [showBillingRemovalModal, setShowBillingRemovalModal] = useState(false)
   const [showSpendReportConfigurationModal, setShowSpendReportConfigurationModal] = useState(false)
   const [selectedBilling, setSelectedBilling] = useState()
   const [selectedDatasetProjectName, setSelectedDatasetProjectName] = useState(null)
@@ -305,6 +306,18 @@ const ProjectDetail = ({ billingProject, reloadBillingProject, billingAccounts, 
     })
   })
 
+  const removeBillingAccount = _.flow(
+    reportErrorAndRethrow('Error removing billing account'),
+    Utils.withBusyState(setUpdating)
+  )(() => {
+    Ajax().Metrics.captureEvent(Events.removeBillingAccount, {
+      billingProject: billingProject.projectName
+    })
+    return Ajax(signal).Billing.removeBillingAccount({
+      billingProjectName: billingProject.projectName
+    })
+  })
+
   const updateSpendConfiguration = _.flow(
     reportErrorAndRethrow('Error updating workflow spend report configuration'),
     Utils.withBusyState(setUpdating)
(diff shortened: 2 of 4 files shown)


ehigham (approved):
Looks great. I wonder if the icon might throw users into thinking they were deleting the billing account rather than just removing it?

kyuksel (approved):

s-rubenstein:
@ehigham it definitely could. I think it would be clearer if we had a real delete billing project endpoint, and thus had a second delete on the page. My guess is that this UX will change once it gets real UX input, but for the time being it will support our users and I am relying on the tooltip to guide users.

zarsky-broad:
try an x in a circle icon for remove rather than delete?

s-rubenstein:
@zarsky-broad done

s-rubenstein:
Screen Shot 2021-10-29 at 11 53 45 AM

s-rubenstein:
Screen Shot 2021-10-29 at 2 52 26 PM

src/libs/ajax.js
```diff
@@ -1166,19 +1172,6 @@ const GoogleBilling = signal => ({
...
-    return response.json()
-  },
-  changeBillingAccount: async ({ projectId, newAccountName }) => {
-    const name = `projects/${projectId}/billingInfo`
-    const response = await fetchGoogleBilling(name,
-      _.mergeAll([
-        authOpts(), { signal, method: 'PUT' },
-        jsonBody({ billingEnabled: true, billingAccountName: newAccountName, name, projectId })
-      ]))
-    return response.json()
```
andy7i, line 1181: why are we removing these?

src/pages/billing/Project.js
```diff
@@ -368,6 +381,19 @@ const ProjectDetail = ({ billingProject, reloadBillingProject, billingAccounts,
...
+        h(Link, {
+          tooltip: 'Remove Billing Account',
+          style: { marginLeft: '0.5rem' },
+          baseColor: colors.danger,
+          onClick: async () => {
+            if (Auth.hasBillingScope()) {
+              setShowBillingRemovalModal(true)
+            } else {
+              await authorizeAndLoadAccounts()
+              setShowBillingRemovalModal(Auth.hasBillingScope())
```
andy7i, line 395: do billing project users have billing scopes? / should billing project users be able to remove the billing account from a billing project?

src/libs/ajax.js
s-rubenstein: They are unused, and, more importantly, we actively don't want to use these. We would much rather use the `.Billing` endpoints

src/pages/billing/Project.js
s-rubenstein: I don't know if users have billing scopes, but currently they don't have access to the project details page. (All 'shared with me' projects are disabled.)

src/libs/ajax.js
s-rubenstein: The reason the one above is staying is that we don't currently have endpoints to check billing accounts in Terra. We would need that capability before removing the listBillingAccounts endpoint from here.

andy7i: aha, I see `Billing.changeBillingAccount` now.
> The reason the one above is staying is that we don't currently have endpoints to check billing accounts in Terra. We would need that capability before removing the listBillingAccounts endpoint from here.

should we have a ticket for this?

src/pages/billing/Project.js
andy7i: the remove button is still clickable when there is no billing account. Maybe we should disable or hide the button in this case?

Screen Shot 2021-11-01 at 12 26 22 PM

src/pages/billing/Project.js
```diff
@@ -391,6 +417,19 @@ const ProjectDetail = ({ billingProject, reloadBillingProject, billingAccounts,
...
               ['Note: Changing the billing account for this billing project will clear the workflow spend report configuration.'])
           ])])
+        ]),
+        showBillingRemovalModal && h(Modal, {
+          title: 'Remove Billing Account',
+          onDismiss: () => setShowBillingRemovalModal(false),
+          okButton: h(ButtonPrimary, {
+            onClick: () => {
+              setShowBillingModal(false)
+              removeBillingAccount(selectedBilling).then(reloadBillingProject)
```
andy7i: modal stays open when I click OK. I can click OK multiple times and it never goes away. loading dialogue opens in the background
Screen Shot 2021-11-01 at 12 24 21 PM

andy7i (changes requested):
tested locally and found some UX issues

src/libs/ajax.js
s-rubenstein: I don't know if it is something we want to handle? @gpolumbo-broad is talking to product about what we want to do to support billing account managers.

src/pages/billing/Project.js
```diff
@@ -384,6 +384,7 @@ const ProjectDetail = ({ billingProject, reloadBillingProject, billingAccounts,
         h(Link, {
           tooltip: 'Remove Billing Account',
           style: { marginLeft: '0.5rem' },
+          disabled: billingProject.billingAccount === 'null',
```
andy7i: As expected, I'm seeing some billing projects with the button disabled if the billing account is "No Access" (which is maybe not the best name for "no billing account", but that's separate issue) 

However, I'm also seeing that if I add and remove a billing account (or just remove an existing billing account) it will not disable the button, even after refreshing the page.

src/pages/billing/Project.js
```diff
@@ -384,7 +384,8 @@ const ProjectDetail = ({ billingProject, reloadBillingProject, billingAccounts,
         h(Link, {
           tooltip: 'Remove Billing Account',
           style: { marginLeft: '0.5rem' },
-          disabled: billingProject.billingAccount === 'null',
+          // For some reason the api sometimes returns string null, and sometimes returns no field, and soemtimes returns null. This is just to be complete.
```
andy7i, line 387: should we create a tech debt ticket to fix this in the backend and add that ticket # here?

andy7i (approved):
thanks for fixing all of the issues! One tech debt ticket question, not blocking for this PR.

src/pages/billing/Project.js
s-rubenstein: https://broadworkbench.atlassian.net/browse/CA-1586 done

andy7i: thanks for creating the ticket! What do you think about adding it to the comment in the code, for ease of reference? I'm worried someone might read this line and try to create a (duplicate) ticket or need to go fishing for it in the great ocean that is jira.

PR #2675, opened 2021-11-01
Support more than just Galaxy apps [BW-891]

cahrens:
I am adding appType to the listApps API method (and getApp). To make this change not have to wait for the Leo code to merge, I am checking whether or not appType is present.

Example with a Cromwell App:

image

### DIFF
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -17,8 +17,8 @@ import colors from 'src/libs/colors'
 import { withErrorIgnoring, withErrorReporting } from 'src/libs/error'
 import {
   defaultComputeRegion,
-  defaultComputeZone, getComputeStatusForDisplay, getCurrentApp, getCurrentRuntime, getGalaxyComputeCost, getGalaxyCost, getPersistentDiskCostMonthly, isComputePausable,
-  isResourceDeletable, runtimeCost
+  defaultComputeZone, getComputeStatusForDisplay, getCurrentApp, getCurrentRuntime, getGalaxyComputeCost, getGalaxyCost, getPersistentDiskCostMonthly,
+  isApp, isComputePausable, isResourceDeletable, runtimeCost
 } from 'src/libs/runtime-utils'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
@@ -338,7 +338,7 @@ const Environments = () => {
             headerRenderer: () => h(Sortable, { sort, field: 'created', onSort: setSort }, ['Type']),
             cellRenderer: ({ rowIndex }) => {
               const cloudEnvironment = filteredCloudEnvironments[rowIndex]
-              return cloudEnvironment.appName ?
+              return isApp(cloudEnvironment) ?
                 (cloudEnvironment.appType ? _.capitalize(cloudEnvironment.appType) : 'Galaxy') :
                 (cloudEnvironment.runtimeConfig.cloudService === 'DATAPROC' ? 'Dataproc' : cloudEnvironment.runtimeConfig.cloudService)
             }
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -454,8 +454,8 @@ const Environments = () => {
                   app && div([
                     span({ style: { fontWeight: 600 } },
                       [app.appType ? `${_.capitalize(app.appType)}: ` : 'Galaxy: ']
-                    ), app.appName]
-                  )
+                    ), app.appName
+                  ])
                 ])
               }, [h(Link, ['view'])])
             }
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -334,13 +334,12 @@ const Environments = () => {
             }
           },
           {
-            size: { basis: 90, grow: 0 },
+            size: { basis: 100, grow: 0 },
             headerRenderer: () => h(Sortable, { sort, field: 'created', onSort: setSort }, ['Type']),
             cellRenderer: ({ rowIndex }) => {
               const cloudEnvironment = filteredCloudEnvironments[rowIndex]
-              // TODO: update return logic once we support more app types (will need a backend change to return appType in list apps endpoint as well)
               return cloudEnvironment.appName ?
-                'Galaxy' :
+                (cloudEnvironment.appType ? _.capitalize(cloudEnvironment.appType) : 'Galaxy') :
                 (cloudEnvironment.runtimeConfig.cloudService === 'DATAPROC' ? 'Dataproc' : cloudEnvironment.runtimeConfig.cloudService)
             }
           },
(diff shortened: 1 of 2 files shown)


src/pages/Environments.js
```diff
@@ -334,13 +334,12 @@ const Environments = () => {
...
-            size: { basis: 90, grow: 0 },
+            size: { basis: 100, grow: 0 },
             headerRenderer: () => h(Sortable, { sort, field: 'created', onSort: setSort }, ['Type']),
             cellRenderer: ({ rowIndex }) => {
               const cloudEnvironment = filteredCloudEnvironments[rowIndex]
-              // TODO: update return logic once we support more app types (will need a backend change to return appType in list apps endpoint as well)
               return cloudEnvironment.appName ?
-                'Galaxy' :
+                (cloudEnvironment.appType ? _.capitalize(cloudEnvironment.appType) : 'Galaxy') :
                 (cloudEnvironment.runtimeConfig.cloudService === 'DATAPROC' ? 'Dataproc' : cloudEnvironment.runtimeConfig.cloudService)
```
cahrens, line 343: @kyuksel I see that you added the above line. Is there any reason to not just do _.capitalize(cloudEnvironment.runtimeConfig.cloudService)? I'm using the capitalize function to convert from CROMWELL to Cromwell.

(I don't know though what cloudEnvironment.runtimeConfig.cloudService will be if it is **not** DATAPROC)

cahrens:
Here is a screenshot with both Cromwell and Galaxy.

image

src/pages/Environments.js
```diff
@@ -334,13 +334,12 @@ const Environments = () => {
...
           {
-            size: { basis: 90, grow: 0 },
+            size: { basis: 100, grow: 0 },
             headerRenderer: () => h(Sortable, { sort, field: 'created', onSort: setSort }, ['Type']),
             cellRenderer: ({ rowIndex }) => {
               const cloudEnvironment = filteredCloudEnvironments[rowIndex]
-              // TODO: update return logic once we support more app types (will need a backend change to return appType in list apps endpoint as well)
               return cloudEnvironment.appName ?
-                'Galaxy' :
+                (cloudEnvironment.appType ? _.capitalize(cloudEnvironment.appType) : 'Galaxy') :
```
zarsky-broad, line 342: I think we should probably change this to a `Utils.cond` at this point for simplicity

cahrens: I hope this will eventually get down to just

`return _.capitalize(cloudEnvironment.appName ? cloudEnvironment.appType : cloudEnvironment.runtimeConfig.cloudService)`

src/pages/Environments.js
```diff
@@ -452,7 +451,11 @@ const Environments = () => {
                 content: div({ style: { padding: '0.5rem' } }, [
                   div([span({ style: { fontWeight: 600 } }, ['Name: ']), name]),
                   runtime && div([span({ style: { fontWeight: 600 } }, ['Runtime: ']), runtime.runtimeName]),
-                  app && div([span({ style: { fontWeight: 600 } }, ['Galaxy: ']), app.appName])
+                  app && div([
+                    span({ style: { fontWeight: 600 } },
+                      [app.appType ? `${_.capitalize(app.appType)}: ` : 'Galaxy: ']
+                    ), app.appName
+                  ])
```
cahrens, line 458: Just noticed this other place where Galaxy was hardcoded. Once the Leo changes merge, this can be simplified to just be `${_.capitalize(app.appType)}: `

image

src/pages/Environments.js
kyuksel: Yea I can't see why we couldn't do what you suggest @cahrens once Leo exposes the `appType` info.

As an aside:
1. Can we factor out the app vs. not app decision making it's contained in one place, and it's easier to see/comment on the assumptions we're making around it (i.e. apps have `appName` defined)?
2. Can we try to avoid magic strings around `appTypes`? It's our bad that we hadn't already been doing that with `Galaxy` but it'll become more important now that we'll have non-Galaxy apps.

(perhaps in `runtime-utils.js` which I'm hoping we'll rename as `compute-utils.js` soon)
```javascript
export const isApp = cloudEnvironment => !!cloudEnvironment?.appName

export const appTypes = {
  galaxy: 'Galaxy',
  cromwell: 'Cromwell',
  custom: 'Custom',
  default: 'Galaxy',
}
```

and then in this file

```javascript
              // TODO: Update the conditionals below as return _.capitalize(isApp(cloudEnvironment) ? cloudEnvironment.appType : cloudEnvironment.runtimeConfig.cloudService) once BW-890 is done.
              return isApp(cloudEnvironment) ?
                (cloudEnvironment.appType ? _.capitalize(cloudEnvironment.appType) : appTypes.default) :
                _.capitalize(cloudEnvironment.runtimeConfig?.cloudService)
```

And finally when the backend is ready, as you suggested:

```javascript
              return _.capitalize(isApp(cloudEnvironment) ? cloudEnvironment.appType : cloudEnvironment.runtimeConfig.cloudService)
```

src/pages/Environments.js
kyuksel: > Is there any reason to not just do _.capitalize(cloudEnvironment.runtimeConfig.cloudService)?

I can't see why we couldn't do what you suggested from looking at the code and running it through a couple of test scenarios.

> I don't know though what cloudEnvironment.runtimeConfig.cloudService will be if it is not DATAPROC

It can also be `GCE`, which should be fine with your code change.

kyuksel (commented):
Looks great overall.
Highlighted inline some opportunity for refactoring, which this file is very ripe for :)
Will test more once the code is finalized.

src/pages/Environments.js
cahrens: @kyuksel but would you want "GCE" to be "Gce"? I would imagine not (which means I can't just always use capitalize).

src/pages/Environments.js
cahrens: I like the isApp idea, but I'm not sold on the appTypes const yet because my hope is that we can get rid of any need to have the app names listed in terra-ui at all. I may be overly naive about that, but given that my goal with these changes is to get rid of "Galaxy" in these spots, it seems strange to temporarily introduce an appTypes const.

kyuksel: We use `'Galaxy'` in a lot of places outside this file too and with `appTypes` I was hoping we could slowly get rid of those as well.

src/pages/Environments.js
kyuksel: Ah yea good point! I have some ideas around how we may be able to better handle string cases by categorizing them as:

1. coming from API responses/input into API requests
2. used internally
3. displayed to users

and providing utilities to make conversions where necessary…but this could be a discussion for another time (perhaps the next Terra UI meeting :)

kyuksel (approved):

kyuksel:
Rerunning the notebook test as it failed with the flaky failure signature…

src/pages/Environments.js
cahrens: @kyuksel I promise that the first time I find myself putting "Cromwell" in some of the terra-ui code (in reference to an App) I will create appTypes.

kpierre13 (approved):

src/pages/Environments.js
kyuksel: 👍🏼

PR #2676, opened 2021-11-01
Updated to allow custom debounce rate

lu-c:
- fixed bug with the input[type:search] remove button not de-selecting items
-- to recreate the original bug, use chrome. Start on a fresh page. Run a search for anything. Hover on the input bar, and a blue x will appear in teh right side. Click the x, the search is removed, as expected. However, click onto the background (outside of the search input), and suddenly the original search is back in the search bar.
 
- fixed bug with the input[type:search] backspace delete not de-selecting items
-- to recreate the original bug, use chrome. Start on a fresh page, run a search for anything (and hit enter). Focus back in to the input bar, double click to select all, and hit "backspace". Data clears, now click on the background of the page again, and the original search is back in the search bar

- updated css for the search bar, so when it is focused, the outline is gone. There were two css rules being triggered for focus, (rule from `style.css#85`) intended to apply react-interactive focus styles on the input itself, and then also from the combobox rule (`style.css#137`)

- Added a customizable debounce parameter. I originally thought the ^ search issues was happening because the user was blurring the search content before the debounce time, but it ended up not fixing anything. However, I really like being able to override it, because 250ms seems like it's so long

ss

### DIFF
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -50,7 +50,7 @@ export const withDebouncedChange = WrappedComponent => {
     const [internalValue, setInternalValue] = useState()
     const getInternalValue = Utils.useGetter(internalValue)
     const getOnChange = Utils.useGetter(onChange)
-    const updateParent = Utils.useInstance(() => _.debounce(250, () => {
+    const updateParent = Utils.useInstance(() => _.debounce(props.debounce || 250, () => {
       getOnChange()(getInternalValue())
       setInternalValue(undefined)
     }))
@@ -240,6 +240,14 @@ const withAutocomplete = WrappedComponent => ({
     { itemToString: v => v ? itemToString(v) : value } :
     { selectedItem: value }
 
+  const inputEl = useRef()
+  let clearMethod
+  Utils.useOnMount(() => {
+    inputEl.current.addEventListener('search', () => {
+      clearMethod && clearMethod()
+    })
+  })
+
   return h(Downshift, {
     ...controlProps,
     initialInputValue: value,
@@ -252,10 +260,11 @@ const withAutocomplete = WrappedComponent => ({
     inputId: id,
     labelId
   }, [
-    ({ getInputProps, getMenuProps, getItemProps, isOpen, openMenu, toggleMenu, highlightedIndex }) => {
+    ({ getInputProps, getMenuProps, getItemProps, isOpen, openMenu, toggleMenu, clearSelection, highlightedIndex }) => {
+      clearMethod = clearSelection
       return div({
         onFocus: openOnFocus ? openMenu : undefined,
-        style: { width: style?.width || '100%', display: 'inline-flex', position: 'relative' }
+        style: { width: style?.width || '100%', display: 'inline-flex', position: 'relative', outline: 'none' }
       }, [
         inputIcon && icon(inputIcon, {
           style: { transform: 'translateX(1.5rem)', alignSelf: 'center', color: colors.accent(), position: 'absolute', ...iconStyle },
@@ -264,7 +273,7 @@ const withAutocomplete = WrappedComponent => ({
         h(WrappedComponent, getInputProps({
           style: inputIcon ? { ...style, paddingLeft: '3rem' } : style,
           type: 'search',
-          onKeyDown: e => {
+          onKeyUp: e => {
             if (e.key === 'Escape') {
               (value || isOpen) && e.stopPropagation() // prevent e.g. closing a modal
               if (!value || isOpen) { // don't clear if blank (prevent e.g. undefined -> '') or if menu is shown
@@ -277,10 +286,14 @@ const withAutocomplete = WrappedComponent => ({
             } else if (e.key === 'Enter') {
               onPick && onPick(value)
               toggleMenu()
+            } else if (e.key === 'Backspace' && !value) {
+              clearSelection()
+              openMenu()
             }
           },
           nativeOnChange: true,
-          ...props
+          ...props,
+          ref: inputEl
         })),
         isOpen && h(AutocompleteSuggestions, {
           target: getInputProps().id,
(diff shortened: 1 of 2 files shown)


src/components/input.js
```diff
@@ -240,6 +240,14 @@ const withAutocomplete = WrappedComponent => ({
     { itemToString: v => v ? itemToString(v) : value } :
     { selectedItem: value }
 
+  const inputEl = useRef()
```
lu-c, line 243: In order to use the x to clear the search, we need this. I used a reference that does this in the same file, for "ConfirmedSearchInput"

src/components/input.js
```diff
@@ -264,7 +273,7 @@ const withAutocomplete = WrappedComponent => ({
         h(WrappedComponent, getInputProps({
           style: inputIcon ? { ...style, paddingLeft: '3rem' } : style,
           type: 'search',
-          onKeyDown: e => {
+          onKeyUp: e => {
```
lu-c, line 276: This is a lifecycle difference, that caused the backspace bug - KeyUp runs after the value has updated, KeyDown runs before the input value updates

src/components/input.js
```diff
@@ -240,6 +240,14 @@ const withAutocomplete = WrappedComponent => ({
     { itemToString: v => v ? itemToString(v) : value } :
     { selectedItem: value }
 
+  const inputEl = useRef()
+  let clearMethod
```
dmohs: I would probably call this `clearFunction` or `clearFn` if I'm understanding the usage correctly. "Method" means "a function defined on an object" to me. Your call.

dmohs (approved):

src/components/input.js
zarsky-broad: This should be a ref instead of a let.

src/components/input.js
```diff
@@ -240,6 +240,14 @@ const withAutocomplete = WrappedComponent => ({
...
+  Utils.useOnMount(() => {
+    inputEl.current.addEventListener('search', () => {
+      clearMethod && clearMethod()
```
zarsky-broad:
```suggestion
      clearMethod?.()
```

src/components/input.js
```diff
@@ -252,10 +260,11 @@ const withAutocomplete = WrappedComponent => ({
     inputId: id,
     labelId
   }, [
-    ({ getInputProps, getMenuProps, getItemProps, isOpen, openMenu, toggleMenu, highlightedIndex }) => {
+    ({ getInputProps, getMenuProps, getItemProps, isOpen, openMenu, toggleMenu, clearSelection, highlightedIndex }) => {
+      clearMethod = clearSelection
```
zarsky-broad: Is it possible to attach a ref to the downshift and call `clearSelection` through that?

src/components/input.js
```diff
@@ -50,7 +50,7 @@ export const withDebouncedChange = WrappedComponent => {
     const [internalValue, setInternalValue] = useState()
     const getInternalValue = Utils.useGetter(internalValue)
     const getOnChange = Utils.useGetter(onChange)
-    const updateParent = Utils.useInstance(() => _.debounce(250, () => {
+    const updateParent = Utils.useInstance(() => _.debounce(props.debounce || 250, () => {
```
zarsky-broad: Should we destructure debounce from props? is it used somewhere further in?

src/components/input.js
lu-c: Sorry, can you expand on what you mean by that?

src/components/input.js
lu-c: I honestly don't know how to do that, since we're using this method in the event listener on the input, we need this method before the downshift block that exposes this method is available

src/components/input.js
dmohs: Isaac is right. `useRef` is the pattern. Take a look at `src/components/Collapse.js` for an example. It will work very similarly to what you have, but `useRef` is more idiomatic to our codebase.

lu-c: Did you mean you wanted the variable called `clearSelectionRef`? I updated the PR to do that, but I am only 20% sure that's what you guys meant.

> This should be a ref instead of a let

`let clearMethod` -> `ref clearMethod` is not a valid line of code, and that won't compile. So I'm just kind of flying off the cuff now, and I need more specificity to fix the issue.

src/components/input.js
lu-c: We are passing props in to other methods, so we should leave it in there

src/components/input.js
```diff
@@ -240,6 +240,14 @@ const withAutocomplete = WrappedComponent => ({
     { itemToString: v => v ? itemToString(v) : value } :
     { selectedItem: value }
 
+  const inputEl = useRef()
+  let clearSelectionRef
```
zarsky-broad: see: https://markoskon.com/the-difference-between-refs-and-variables/

zarsky-broad:
Rename debounce and destructure

zarsky-broad (approved):

petesantos:
Testing this, I am seeing a bug - when hitting enter it clears the search. Do we need the ref at all, or can we just call `clearSelection` on Backspace?

lu-c:
> Testing this, I am seeing a bug - when hitting enter it clears the search. Do we need the ref at all, or can we just call `clearSelection` on Backspace?

Unfortunately, we need the ref for when you hit the x button in chrome. It does not register as a keypress action. I think I pushed up a fix for this bug, would you mind trying it again?

petesantos:
> 
> 
> > Testing this, I am seeing a bug - when hitting enter it clears the search. Do we need the ref at all, or can we just call `clearSelection` on Backspace?
> 
> Unfortunately, we need the ref for when you hit the x button in chrome. It does not register as a keypress action. I think I pushed up a fix for this bug, would you mind trying it again?

This is working for me - thanks!

petesantos (approved):

PR #2678, opened 2021-11-04
[IA-3039] Analysis Tab Bug bash fixes 

jdcanas:
This PR addresses a bunch of misc issues identified in a [bug bashing](https://docs.google.com/spreadsheets/d/1pMHeNlZrAuK_O8HPEhk8XSxPsYosXAPLx5nBT1jWXBg/edit#gid=0)  session.

It also fixes an unrelated bug, where our modals (Compute and Galaxy) were not properly closing on error, and then hiding the error behind them. It does this by ensuring the error reporting function calls the modal onDismiss if an error is thrown via `withErrorReportingInModal`.


src/libs/error.js
```diff
@@ -37,6 +37,21 @@ export const reportErrorAndRethrow = _.curry((title, fn) => async (...args) => {
   }
 })
 
+/**
+ * This function is designed for use in modals
```
kyuksel: Nit:
```suggestion
 *  This function is designed for use in modals
```

src/components/ContextBar.js
```diff
@@ -58,6 +59,7 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
   const getColorForStatus = status => Utils.cond(
     [_.upperCase(status) === 'RUNNING', () => colors.success()],
     [_.upperCase(status) === 'ERROR', () => colors.danger()],
+    [_.includes('ING', _.upperCase(status)), () => colors.accent()],
```
kyuksel, line 62: I wonder if we should enumerate the possible status values here to be safe. We could have an array of them somewhere and still one-line it above and also on [this line](https://github.com/DataBiosphere/terra-ui/pull/2678/files#diff-4b313fc432ea524fbc7fdff4b0093404a4e65633ba99229f896fdfc15a0822c2L320).

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -311,17 +313,17 @@ const Analyses = _.flow(
...
+    div({ style: { display: 'flex', flexDirection: 'row', justifyContent: 'center', alignItems: 'center' } }, [
+      img({ src: jupyterLogo, style: { height: 120, width: 80, marginRight: '5rem' } }),
+      img({ src: rstudioBioLogo, style: { width: 406, height: 75, marginRight: '1rem' } }),
       div([
-        img({ src: galaxyLogo, style: { height: 75, width: 260, marginTop: '2.5rem' } })
+        img({ src: galaxyLogo, style: { height: 60, width: 208 } })
         // span({ style: { marginTop: '3.5rem'} }, ['Galaxy'])
       ])
     ]),
     //TODO, check with Joy, wording change (it may not be their first, just first in this workspace)
```
kyuksel: Is this `TODO` still applicable?

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -63,13 +64,10 @@ export const CloudEnvironmentModal = ({
...
-    onSuccess: _.flow(
-      withErrorReporting('Error creating cloud compute'),
-      Utils.withBusyState(setBusy)
-    )(async () => {
+    onSuccess: () => {
       setViewMode(undefined)
-      await refreshRuntimes(true)
-    })
+      onDismiss()
+    }
```
kyuksel, line 70: Could you help me understand the following?

2. Why are we removing the error reporting? What happens if `onDismiss()` throws an error?
2. Why are we no longer setting `busy` to `true`? 

Same questions for below.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -347,7 +342,7 @@ export const CloudEnvironmentModal = ({
             disabled: isCloudEnvForToolDisabled,
             onClick: () => setViewMode(toolLabel)
           }, [
-            img({ src: cloudIcon, style: { height: 20, width: 20, opacity: isCloudEnvForToolDisabled ? 0.4 : 1 } }), //TODO: why doesn't this icon disable properly color-wise?
```
kyuksel, line 350: What does it mean for an icon to "disable properly colour-wise"?

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -400,8 +395,7 @@ export const CloudEnvironmentModal = ({
     errorRuntimeId && h(RuntimeErrorModal, {
       runtime: _.find({ id: errorRuntimeId }, runtimes),
       onDismiss: () => setErrorRuntimeId(undefined)
-    }),
-    busy && spinnerOverlay
```
kyuksel: Why do we want to remove these?

kyuksel (commented):
Looks good overall. Mostly questions inline to better understand the changes.

kyuksel:
### Testing

Verified the fix for the bug not on the spreadsheet

> where our modals (Compute and Galaxy) were not properly closing on error, and then hiding the error behind them.

Verified the following from the spreadsheet are fixed:
1. Context bar isn't visible when notebook is open.
2. a. If you click on an Rmd file with no running cloud environment, the Context Bar on the right disappears.

I still see the following happen:
2. b. If you click on an Rmd file with no running cloud environment, the user has to navigate back to the analysis tab in order to create a cloud environment. Tabs are also not present so the user either has to click back or click on analyses in upper green section of UI after workspace.

Is there a ticket to fix 2.b.?

kyuksel:
Is there a ticket to ensure the following is fixed when the feature flag is removed?

> Context Bar not visible on page with Terminal

kyuksel:
What's the plan/resolution for the items on the spreadsheet that aren't highlighted in green or orange?

kyuksel:
Regarding the spreadsheet item `Can't delete a Jupyter runtime in Pausing state` and the note

> hmm.. this was a bug. I think it should be fixed (in terms of parity with current functionality for rstudio and jupyter). Would like secondary verification

I still see it. Clicked on the `pause` icon, and then went through the `delete` clickings, the modal remained with a spinner overlay and eventually got dismissed with an error pop-up. When I clicked on the error link, it displayed

image

src/components/ContextBar.js
jdcanas: I think its safer to use 'ING', since this approach allows the backend to add further transition statuses without needing to remember this specific UI place (thinking of when we added things like 'PreCreating')

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
jdcanas: So this error reporting is actually duplicative. The modal itself has its own error reporting wrapper, and this was interfering.

We are not setting busy to true here as this is no longer an async blocking call. This was also duplicative, and you'll notice that any busy spinner + overlay is handled by the inner modal.

This was part of an investigation to correctly display the spinner and any errors associated with async actions.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
jdcanas: Oh, originally the img was misbehaving and was not being grayed out when the disabled property was true (it was staying the same color)

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
jdcanas: Hmm.. I was trying to remove the layers of nested-blocking from this modal and the inner New(Runtime/Galaxy)Modals, but it seems that isn't possible since this modal can pause/stop. Essentially the behavior without this is, instead of blocking with a spinner when you click stop/start, it refreshes dynamically. This doesn't look the best when tested with simulated slow internet speed though... I'll add it back. Was kind of a rabbit hole to fix the issues I mention above so thanks for catching this :)

jdcanas:
> ### Testing
> Verified the fix for the bug not on the spreadsheet
> 
> > where our modals (Compute and Galaxy) were not properly closing on error, and then hiding the error behind them.
> 
> Verified the following from the spreadsheet are fixed:
> 
> 1. Context bar isn't visible when notebook is open.
> 2. a. If you click on an Rmd file with no running cloud environment, the Context Bar on the right disappears.
> 
> I still see the following happen: 2. b. If you click on an Rmd file with no running cloud environment, the user has to navigate back to the analysis tab in order to create a cloud environment. Tabs are also not present so the user either has to click back or click on analyses in upper green section of UI after workspace.
> 
> Is there a ticket to fix 2.b.?

Fixed

jdcanas:
> Is there a ticket to ensure the following is fixed when the feature flag is removed?
> 
> > Context Bar not visible on page with Terminal

Essentially, it does not work because terra ui uses `sessionStorage` over `localStorage` (the config is scoped to a tab). This will be fixed for free with no verification needed once the feature flag is removed (you can verify by manually re-setting the flag in the config in the new tab)

We talked about this and decided on the approach for upcoming user testing in this slack thread https://broadinstitute.slack.com/archives/GB1VAM14J/p1634923111060000

With the technical discussion here https://broadinstitute.slack.com/archives/C01EHNUM73R/p1634920178004000

jdcanas:
> What's the plan/resolution for the items on the spreadsheet that aren't highlighted in green or orange?

I added a few notes here and there to clarify. Overall, the outstanding non-bug items should be captured here https://broadworkbench.atlassian.net/browse/IA-3033

jdcanas:
> I still see it. Clicked on the `pause` icon, and then went through the `delete` clickings, the modal remained with a spinner overlay and eventually got dismissed with an error pop-up. When I clicked on the error link, it displayed

Oh I tested it without a disk. The original issue is that it physically would not let you click delete (or open then modal) when it was in this status. This seems to be a backend issue at this point, I'll make a follow-up ticket

kyuksel:
Could you compress `r-bio-logo` by dragging and dropping the file on [tinypng.com](https://tinypng.com), and check that in instead?

image

src/components/ContextBar.js
kyuksel: I guess it's safe in this case to assume that a status that a backend reports is fine to pass through to users.

src/libs/error.js
```diff
@@ -37,6 +37,21 @@ export const reportErrorAndRethrow = _.curry((title, fn) => async (...args) => {
   }
 })
 
+/**
+ *  This function is designed for use in modals
+ *  Modals can overlay any error reporting, with the `throw` in the default `withErrorReporting` preventing the modal itself from closing on error
+ *  As such, we must ensure we call the dismiss function if an error occurs
+ */
```
kyuksel: Sorry, I missed a few things on this last time:

```suggestion
/**
 * This function is designed for use in modals.
 * Modals can overlay any error reporting, with the `throw` in the default `withErrorReporting`
 * preventing the modal itself from closing on error.
 * As such, we must ensure we call the dismiss function if an error occurs
 */
```

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
kyuksel: Sounds great. Thank you.

kyuksel:
@jdcanas Thanks for addressing my comments. Would you mind replacing the current version of the new image with a compressed version? My original comment is [here](https://github.com/DataBiosphere/terra-ui/pull/2678#issuecomment-964581074).

kyuksel (approved):

Qi77Qi (approved):

src/components/ContextBar.js
cahrens: @jdcanas @kyuksel 
With this change, "Provisioning" does not show in the orange dot and instead appears to be green (though perhaps I just can't tell the difference visually between colors.accent() and colors.success()). To me, this seems confusing because it appears that the app is ready to interact with.

cahrens: I think the issue is that the accent and success colors are too similar. Here is Galaxy "resuming" and Cromwell "running". When you have both at the same time, you might notice a difference, but when there is only one showing I think it looks too similar to green.

image

PR #2680, opened 2021-11-04
[PF-692] Phase 2 regionality release

wnojopra:
Follow up from #2660 . Phase 2 allows Cloud Environments to be created outside of us-central1. This becomes available if your workspace bucket is outside of the US.

To test:
1) Go to a workspace with a non-US bucket. If you need to create one, modify the availableBucketRegions function in region-common.js.
2) Open the Cloud Environment modal and click `CUSTOMIZE`. The location dropdown should be enabled and should have two choices - the bucket region and the US.
3) If you create the environment in the non-US region, you should see this warning:
image

4) Create the environment!

This PR also updates the pricing estimates to be regional. Previously, price estimate data was looked up manually in the documentation and entered into machines.js. I have written a script that looks up the pricing data via Google's API in https://gist.github.com/wnojopra/be5be3ab7e6a09a351b89085992a940d (TODO: re-write this from Python to Node). You can copy-paste the output of that script into machines.js for updated pricing data. Follow the instructions in the script to run.


scripts/parse_pricing_api.py
```diff
@@ -0,0 +1,303 @@
...
+# See https://cloud.google.com/skus/sku-groups/on-demand-vms,
+# https://cloud.google.com/skus/sku-groups/on-demand-persistent-disk-hdd
+# on SKU groups
+GCE_SERVICE = '6F81-5844-456A'
+
+N1_STANDARD_GROUP = 'N1Standard'
+PD_STANDARD_GROUP = 'PDStandard'
+GPU_STANDARD_GROUP = 'GPU'
+
+NVIDIA_TESTLA_T4 = 'Nvidia Tesla T4'
```
kyuksel: Are these supposed to be `*TESLA*` instead of `*TESTLA*`?

src/components/ComputeModal.js
```diff
@@ -1006,19 +1009,19 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       div({ style: { ...gridStyle, gridTemplateColumns: '0.25fr 8.5rem 1fr 5.5rem 1fr 5rem', marginTop: '1.5rem' } }, [
         h(IdContainer, [
           id => div({ style: { gridColumnEnd: 'span 3' } }, [
-            label({ htmlFor: id, style: computeStyles.label }, ['Location']),
+            label({ htmlFor: id, style: computeStyles.label }, ['Location ']),
+            versionTag('Beta', { color: colors.primary(1.5), backgroundColor: 'white', border: `1px solid ${colors.primary(1.5)}` }),
```
```javascript
versionTag('Beta', { color: colors.primary(1.5), backgroundColor: 'white', border: `1px solid ${colors.primary(1.5)}` })
```

seems to be used in 2 other places. Can we factor it out?

src/components/ComputeModal.js
```diff
@@ -1006,19 +1009,19 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
             h(InfoBox, { style: { marginLeft: '0.5rem' } }, [
-              'Cloud environments run in the same region as the workspace bucket and cannot be changed at this time.'
+              'Cloud environments run in the same region as the workspace bucket and can be changed as a beta feature.'
             ]),
             div({ style: { marginTop: '0.5rem' } }, [
               h(Select, {
                 id,
-                isDisabled: true, // Initially releasing regionality supporting only us-central1 region
-                // isDisabled: computeExists, // Can't update location of existing environments
+                isDisabled: computeExists || isUSLocation(bucketLocation) || sparkMode === 'cluster', // Can't update location of existing environments
```
kyuksel: 1. Why do we need `isUSLocation(bucketLocation)`?
2. Should we add `sparkMode === 'master'` to exclude Dataproc clusters with a single node?

src/components/region-common.js
```diff
@@ -64,26 +65,38 @@ export const allRegions = [
...
+  { value: 'AUSTRALIA-SOUTHEAST1', label: 'australia-southeast1 (Sydney)', locationType: locationTypes.region }
 ]
+
+// For current phased release of regionality only supporting US and US-CENTRAL1 buckets.
+export const availableBucketRegions = allRegions.slice(0, 2)
+
+// For current phased release of regionality only supporting compute region in US-CENTRAL1
+// and the same region as your workspace bucket
+export const getAvailableComputeRegions = location => {
+  const UsCentralRegion = _.find({ value: 'US-CENTRAL1' }, allRegions)
```
kyuksel: Can we avoid magic strings so default locations are defined in a single place?

```suggestion
  const defaultComputeRegion = _.find({ value: defaultComputeRegion }, allRegions)
```

kyuksel (commented):
Couldn't finish going through the PR yet so to be continued…

src/components/ComputeModal.js
wnojopra: 1) For workspaces with buckets in us-central1 or us multi-regional, we don't want to enable a location choice. `isUSLocation` is a helper function returns true if the workspace bucket is us-central1 or us multi-regional.
2) My understanding is that single node dataproc clusters still work. The bug with dataproc clusters is caused by misaligned IP ranges preventing the VMs in the clusters from communicating properly.

I've added comments here that hopefully helps clarify this.

src/components/region-common.js
wnojopra: My thought was this function explicitly cares that the region is US, and doesn't necessarily care that we've chosen US-CENTRAL1 as the default. 

Its a minor point to me, however. I've made the change.

src/components/ComputeModal.js
wnojopra: Done.

scripts/parse_pricing_api.py
wnojopra: Yes. Thank you. Fixed.

scripts/parse_pricing_api.py
```diff
@@ -0,0 +1,303 @@
+"""Script that parses pricing data from Google's API.
```
zarsky-broad: Can you find some way to rewrite this as node.js or shell script? I'm really, really, really hesitant to introduce python to this repo.

wnojopra: Including the script in this PR isn't **necessary**, but I thought I'd add it here as a "show your work" kind of deal. I did make an attempt with Bash but it really ended up taking me too much effort for something that isn't necessary.

What I can do is host the Python script elsewhere, and add a pointer here to that script. How does that sound?

zarsky-broad: Would you consider just using node?

wnojopra: Sure, the script can be any language. I used Python just because I have more expertise in it. Re-writing this in Node would take me a while, though not impossible. 

For now I'll remove this script from the PR and put it somewhere else. I'll add a TODO task to re-write the script in Node for anyone to do.

wnojopra: I mentioned this in a code comment and the PR description but I have the script as a gist here: https://gist.github.com/wnojopra/be5be3ab7e6a09a351b89085992a940d

kyuksel: @wnojopra could you please add the script to [dsp-scripts/terra-ui](https://github.com/broadinstitute/dsp-scripts/tree/master/terra-ui)? Sorry for the trouble.

wnojopra: Done, https://github.com/broadinstitute/dsp-scripts/pull/96 .

src/components/region-common.js
kyuksel: Ah I see. In that case, we could define a `defaultUSComputeRegion` or such but probably overkill in this case.

src/components/ComputeModal.js
```diff
@@ -1102,7 +1108,35 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           'For more information please read the documentation.',
           icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
         ]),
-        p([`If you want your VM in ${computeConfig.computeRegion.toLowerCase()} continue. Otherwise, go back to select another location.`])
+        p(['If you want your VM in ',
+          strong(`${computeConfig.computeRegion.toLowerCase()}`),
```
kyuksel: Can we use `strong` styling on line [1104](https://github.com/DataBiosphere/terra-ui/pull/2680/files#diff-149d7ad595448fb2f6bedcff6dde5f4cbcccd8d5148fed8820ce88190a3bc38bR1104) and [1448](https://github.com/DataBiosphere/terra-ui/pull/2680/files#diff-149d7ad595448fb2f6bedcff6dde5f4cbcccd8d5148fed8820ce88190a3bc38bR1448) as well?

src/components/region-common.js
wnojopra: Agreed that would be overkill, but feel free to push back if you feel strongly otherwise.

src/libs/runtime-utils.js
```diff
@@ -77,14 +76,36 @@ export const getValidGpuTypes = (numCpus, mem, zone) => {
...
+  const machineType = _.find({ name: machineTypeName }, machineTypes)
+  const regionalPrices = _.find({ name: region }, regionToPrices)
+  return (machineType.cpu * regionalPrices.n1HourlyCpuPrice) + (machineType.memory * regionalPrices.n1HourlyGBRamPrice)
+}
+
+const getHourlyPreemptibleCostForMachineType = (machineTypeName, region) => {
+  const machineType = _.find({ name: machineTypeName }, machineTypes)
+  const regionalPrices = _.find({ name: region }, regionToPrices)
+  return (machineType.cpu * regionalPrices.preemptibleN1HourlyCpuPrice) + (machineType.memory * regionalPrices.preemptibleN1HourlyGBRamPrice)
+}
```
kyuksel: `regionalPrices` `name`s are uppercase whereas values passed into the two methods above are lowercase which causes a runtime error. I find it difficult not to make mistakes when using various cases throughout the code requiring knowing where to make the right conversion at the right place; e.g. 

image

Any reason why we can't default `name`s to lowercase and make conversions only at API call / display time if necessary?

kyuksel: Also, does it make sense to merge these two helper methods as below to avoid code duplication?

```javascript
const getHourlyCostForMachineType = (machineTypeName, region, isPreemptible) => {
  const { cpu, memory } = _.find({ name: machineTypeName }, machineTypes)
  const { n1HourlyCpuPrice, preemptibleN1HourlyCpuPrice, n1HourlyGBRamPrice, preemptibleN1HourlyGBRamPrice } = _.find({ name: _.toUpper(region) },
    regionToPrices)
  return !isPreemptible ?
    (cpu * n1HourlyCpuPrice) + (memory * n1HourlyGBRamPrice) :
    (cpu * preemptibleN1HourlyCpuPrice) + (memory * preemptibleN1HourlyGBRamPrice)
}
```

wnojopra: 1) Where do you see lowercase values passed in? Names are always uppercase, and we convert to lowercase only for the leo api and for display time. We default to uppercase because the values returned by the Google API are uppercase.
2) Thanks! I have merged them together.

src/components/ComputeModal.js
wnojopra: Done. Thanks.

src/components/region-common.js
kyuksel: No push back :)

scripts/parse_pricing_api.py
kyuksel: I'll review that PR separately.

src/libs/runtime-utils.js
kyuksel: > Where do you see lowercase values passed in?

While creating a Dataproc cluster.

wnojopra: There's a bug with runtimeConfigCost. Its trying to compute cost based on the region returned by Leo's API. For dataproc clusters, the region returned is in lowercase. For standard VMs, the region doesn't exist (it returns zone instead), so it uses the default US-CENTRAL1 region (even if the VM is outside of US).

Working on a fix.

src/components/ComputeModal.js
```diff
@@ -1006,19 +1009,22 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
             ]),
             div({ style: { marginTop: '0.5rem' } }, [
               h(Select, {
                 id,
-                isDisabled: true, // Initially releasing regionality supporting only us-central1 region
-                // isDisabled: computeExists, // Can't update location of existing environments
+                // Location dropdown is disabled for:
+                // 1) If editing an existing environment (can't update location of existing environments)
+                // 2) Workspace buckets that are either us-central1 or us multi-regional
+                isDisabled: computeExists || isUSLocation(bucketLocation),
```
kyuksel: It feels like we need a tooltip explaining why it is disabled when it is disabled. What do you think?

image

src/components/ComputeModal.js
```diff
@@ -1095,14 +1101,45 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+          strong(`${bucketLocation.toLowerCase()}`),
+          ' may incur network egress charges.'
+        ]),
+        h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/360058964552', ...Utils.newTabLinkProps }, [
+          'For more information please read the documentation.',
+          icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
         ]),
+        p(['If you want your VM in ',
+          strong(`${computeConfig.computeRegion.toLowerCase()}`),
+          ' continue. Otherwise, go back to select another location.'])
```
kyuksel: Should there be a comma before `continue`?

```suggestion
          ', continue. Otherwise, go back to select another location.'])
```

Same for [line 1140](https://github.com/DataBiosphere/terra-ui/pull/2680/files#diff-149d7ad595448fb2f6bedcff6dde5f4cbcccd8d5148fed8820ce88190a3bc38bR1142).

src/components/ComputeModal.js
```diff
@@ -1095,14 +1101,45 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       }),
       div({ style: { lineHeight: 1.5 } }, [
         p([
-          `This cloud environment will be created in the region ${computeConfig.computeRegion.toLowerCase()}. `,
-          `Copying data from your workspace bucket in ${bucketLocation.toLowerCase()} may incur network egress charges.`
+          'This cloud environment will be created in the region ',
+          strong(`${computeConfig.computeRegion.toLowerCase()}.`),
+          ' Copying data from your workspace bucket in ',
+          strong(`${bucketLocation.toLowerCase()}`),
+          ' may incur network egress charges.'
```
kyuksel: I wonder if we should also mention that network egress charges are not accounted for in cost estimates that are dynamically displayed, which may misleadingly give the impression otherwise.

Same for line 1452.

image

ghost: Agreed. Something along these lines…

"This cloud environment will be created in the region us-central1. Copying data from your workspace bucket in asia-southeast2 may incur network egress charges. Cost estimates displayed in the Terra cloud environment modal will not reflect these additional charges."

src/libs/runtime-utils.js
wnojopra: Pushed a fix. Where we read the config from Leonardo directly, we turn the region (if exists) uppercase. If it doesn't exist (because its a standard VM), we get a region from the zone and uppercase that.

src/components/ComputeModal.js
wnojopra: I agree, that makes total sense. I added these two messages depending on why its disabled:
>Cannot update the location of an existing cloud environment. Delete your cloud environment to create a new one in a different region.

>Currently US workspaces can only have US cloud environments.

src/components/ComputeModal.js
wnojopra: Yes, fixed.

src/components/ComputeModal.js
kyuksel: Sorry that I forgot to mention this earlier: 

React best practice for inputting children is as an array:

```javascript
strong([computeConfig.computeRegion.toLowerCase])
```

Similarly for the few other places in this file.

src/components/ComputeModal.js
wnojopra: Agreed! Added additional text:

> ...may incur network egress charges. Note that network egress charges are not accounted for in cost estimates.

src/components/ComputeModal.js
wnojopra: No worries - done.

kyuksel:
I'm seeing an issue when creating a runtime after deleting a runtime without refreshing the page manually, at least that's the only pattern I could identify so far. Using my test workspace with bucket in a non-US region

image

if I open the compute modal right after deleting a runtime, it displays location as `us-central1`, and if I click `CREATE`, it submits a runtime creation request to Leo with location as `us-central1` too with none of the warning wording due to locations differing. I didn't get to dig into it why. If I manually refresh the page, the modal seems to eventually pick up the right (Asia) location.

kyuksel:
Workspace cloning modal displays `Select…` for `Bucket location` when its bucket is in a non-US location:

image

kyuksel:
The warning message below

image

seems incorrect based on Google's documentation?

image

src/components/region-common.js
```diff
@@ -64,26 +66,38 @@ export const allRegions = [
...
-  // { value: 'ASIA-NORTHEAST1', label: 'asia-northeast1 (Tokyo)', locationType: 'region' },
-  // { value: 'ASIA-NORTHEAST3', label: 'asia-northeast3 (Seoul)', locationType: 'region' },
-  // { value: 'ASIA-SOUTH1', label: 'asia-south1 (Mumbai)', locationType: 'region' },
-  // { value: 'ASIA-SOUTHEAST1', label: 'asia-southeast1 (Singapore)', locationType: 'region' },
-  // { value: 'ASIA-SOUTHEAST2', label: 'asia-southeast2 (Jakarta)', locationType: 'region' },
-  // { value: 'AUSTRALIA-SOUTHEAST1', label: 'australia-southeast1 (Sydney)', locationType: 'region' }
+  { value: 'US-CENTRAL1', label: 'us-central1 (Iowa)', locationType: locationTypes.region },
+  { value: 'NORTHAMERICA-NORTHEAST1', label: 'northamerica-northeast1 (Montreal)', locationType: locationTypes.region },
+  { value: 'SOUTHAMERICA-EAST1', label: 'southamerica-east1 (Sao Paulo)', locationType: locationTypes.region },
+  { value: 'EUROPE-CENTRAL2', label: 'europe-central2 (Warsaw)', locationType: locationTypes.region },
```
kyuksel, line 74: I vaguely recall a problem with the location above. Has that been resolved?

wnojopra:
> Workspace cloning modal displays Select… for Bucket location when its bucket is in a non-US location:

Yeah... this one's a bit tricky. The UI here tries to set the clone workspace location to be the same as the source workspace location, but with this phased release, we still only allow the UI to create buckets in US. One has to use the API to create a non-US workspace bucket. 

I've pushed a fix that will set the bucketLocation to US if you have a non-US workspace bucket, with a comment explaining that in this phased release we're only allowing US workspace buckets. I think this is the right thing to do at this point.

src/components/region-common.js
wnojopra: Yes, resolved with https://github.com/DataBiosphere/terra-resource-buffer/pull/190

wnojopra:
> seems incorrect based on Google's documentation?

It's a bit confusing, but that says to a "different Google Cloud service". Bucket to bucket transfers fall under the "Data moves between different locations on the same continent (assuming none of the above free cases apply)" category.

wnojopra:
> I'm seeing an issue when creating a runtime after deleting a runtime without refreshing the page manually, at least that's the only pattern I could identify so far. Using my test workspace with bucket in a non-US region

Kaan, I'm not able to reproduce this. I've uploaded a screencast of me attempting to reproduce: https://www.youtube.com/watch?v=0P606VYfTFY . Am I doing anything differently than you? Can you help me reproduce?

src/components/region-common.js
```diff
@@ -64,26 +66,38 @@ export const allRegions = [
...
+  { value: 'ASIA-NORTHEAST1', label: 'asia-northeast1 (Tokyo)', locationType: locationTypes.region },
+  { value: 'ASIA-NORTHEAST3', label: 'asia-northeast3 (Seoul)', locationType: locationTypes.region },
+  { value: 'ASIA-SOUTH1', label: 'asia-south1 (Mumbai)', locationType: locationTypes.region },
+  { value: 'ASIA-SOUTHEAST1', label: 'asia-southeast1 (Singapore)', locationType: locationTypes.region },
+  { value: 'ASIA-SOUTHEAST2', label: 'asia-southeast2 (Jakarta)', locationType: locationTypes.region },
+  { value: 'AUSTRALIA-SOUTHEAST1', label: 'australia-southeast1 (Sydney)', locationType: locationTypes.region }
 ]
+
+// For current phased release of regionality only supporting US and US-CENTRAL1 buckets.
+export const availableBucketRegions = allRegions.slice(0, 2)
```
kyuksel: We typically recommend using Lodash functions over the native functions for the following reasons:

- Lodash/fp functions support auto currying. This is especially useful for code changes later. If someone wants to add functionality, it makes it easy to do something like add the function to a flow
- Lodash/fp will not mutate the array. In the case of native slice, it also returns a new array. But there are other functions that can mutate (e.g. push/pop)
-Mixing native and lodash functions can lead to confusion about behavior.

src/libs/runtime-utils.js
```diff
@@ -41,12 +41,26 @@ export const usableStatuses = ['Updating', 'Running']
 
 export const getDefaultMachineType = isDataproc => isDataproc ? defaultDataprocMachineType : defaultGceMachineType
 
+// GCP zones look like 'US-CENTRAL1-A'. To get the region, remove the last two characters.
+export const getRegionFromZone = zone => zone.slice(0, -2)
```
kyuksel, line 46: same comment as https://github.com/DataBiosphere/terra-ui/pull/2680#discussion_r752292241

kyuksel:
> Kaan, I'm not able to reproduce this. I've uploaded a screencast of me attempting to reproduce: https://www.youtube.com/watch?v=0P606VYfTFY . Am I doing anything differently than you? Can you help me reproduce?

Yea that's interesting. I wonder if there's something peculiar with my (Chrome) browser. It _is_ up to date for what it's worth.

I reproduced it again at first try and the screencast is below.

https://user-images.githubusercontent.com/5438223/142446630-6f13a968-4af4-4447-92b6-adc596990a55.mov

wnojopra:
> I reproduced it again at first try and the screencast is below.

Ok! I think I understand the issue. When WorkspaceContainer is first mounted it makes one api call to GCS for the bucket location. But if you do something that calls wrapWorkspace (such as navigating to a different tab, for example), it re-renders WorkspaceContainer _but doesn't keep the old state_. So when you opened the ComputeModal, you now opened it with the default location instead of the actual location. 

So the fix I pushed copies how WorkspaceContainer caches the workspace data. I create a new store for the bucket location, and fetch it when re-rendering WorkspaceContainer.

src/components/region-common.js
wnojopra: Fixed. Thanks! I have trouble remembering which functions are available in lodash, so appreciate the reminders.

src/libs/runtime-utils.js
wnojopra: Lodash's slice works on arrays, and here we're doing a string slice.

kyuksel:
> It's a bit confusing, but that says to a "different Google Cloud service".

Oh wow - good eye!

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -247,8 +246,10 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
       cachedWorkspace :
       undefined
     const [googleProject, setGoogleProject] = useState(workspace?.workspace.googleProject)
-    const [location, setLocation] = useState(defaultLocation)
-    const [locationType, setLocationType] = useState(locationTypes.default)
+
+    const cachedLocation = Utils.useStore(workspaceBucketLocationStore)
+    const location = cachedLocation?.location
+    const locationType = cachedLocation?.locationType
```
kyuksel: Is it possible for `location` or `locationType` to be undefined? If it is, what would the downstream implication be?

kyuksel (commented):
I think I've validated the fixes since my last round of review and reviewed the corresponding code. Good work! One question inline about the new store entry.

Next (Monday unless something unforeseen happens), I want to do some validation on price, etc. changes in `machines.js` and `region-common.js`. I'll so ask the team for someone else to do a review. 

I'll plan on reviewing the script PR later in the week since it's not a blocker.

src/pages/workspaces/workspace/WorkspaceContainer.js
wnojopra: Yes. Downstream components would need to fall back to the defaultLocation and defaultLocationType.

jdcanas (approved):
Looking good from a functional testing perspective,

tested it with disks from the same/different region, and cross-region + same-region workspace/VMs.

src/pages/workspaces/workspace/WorkspaceContainer.js
kyuksel: Saw some opportunity to remove state around bucket location and made a [PR into your branch](https://github.com/DataBiosphere/terra-ui/pull/2701) thinking it might be easier to explain rather than in comments. Let me know what you think.

wnojopra: The main change in that PR seems to undo the bug you discovered and I fixed in [this comment](https://github.com/DataBiosphere/terra-ui/pull/2680#issuecomment-973167744). See my comment that PR for details.

kyuksel:
I see a discrepancy in cost estimates that showed up when I enabled GPUs and started playing around with values.
I'm using the same workspace for both.

### Dev:

image

### PR:

image

kyuksel:
Another cost estimate discrepancy, this time with Dataproc clusters: Not sure if it could be a rounding difference but even if it is, it'd be good to understand why.

I'm using the same workspace for both.

### Dev:

image

### PR:

image

src/data/machines.js
```diff
@@ -139,32 +138,180 @@ export const ephemeralExternalIpAddressPrice = { // per hour in dollars for all
...
-  { name: 'US-EAST1', monthlyDiskPrice: 0.04 },
-  { name: 'US-EAST4', monthlyDiskPrice: 0.044 },
-  { name: 'US-WEST1', monthlyDiskPrice: 0.04 },
-  { name: 'US-WEST2', monthlyDiskPrice: 0.048 },
-  { name: 'US-WEST3', monthlyDiskPrice: 0.048 },
-  { name: 'US-WEST4', monthlyDiskPrice: 0.044 }
+// Disk prices below are per GB per https://cloud.google.com/compute/all-pricing#disk
+// GPU prices below are hourly and per GPU (https://cloud.google.com/compute/gpus-pricing).
+// The data below comes from Google's pricing API: https://cloud.google.com/billing/v1/how-tos/caltalog-api
+// There is a script available in https://gist.github.com/wnojopra/be5be3ab7e6a09a351b89085992a940d
```
kyuksel, line 144: Let's put a TODO above to update the URL after merging the `dsp-scripts` PR.

kyuksel:
Did some spot-checks on individual resource prices in `machines.js` vs. Google docs, and they looked good 👍🏼

wnojopra:
The discrepancy with the 4 P4 GPUs is due to a bug on dev. Take a look at the price for [4 P4 GPUs on dev](https://github.com/DataBiosphere/terra-ui/blob/dev/src/data/machines.js#L45). It is `1.8000`. Because the price for one P4 GPU is `0.6000`, the price for four should be 4x that, or `2.4000`. The price is correct in my PR. 

Taking a look at the dataproc discrepancy now.

wnojopra:
The dataproc discrepancy is 100% caused by rounding changes. I believe the cost estimates in my PR are more accurate. 

In your example, you have 16 `n1-highmem-96` preemtible workers. On [dev an `n1-highmem-96` preemptible machine](https://github.com/DataBiosphere/terra-ui/blob/dev/src/data/machines.js#L16) costs $1.20 per hour. This makes the total cost for preemptible workers 16 * $1.20 = $19.20

In my PR, the cost of the preemptible machine $1.195488 per hour. This makes the total cost for preemptible workers 16 * $1.195488 = $19.127808. This is nearly an 8 cent difference alone. Then you include the same discrepancies for the non-preemptible workers.

src/data/machines.js
wnojopra: Done.

src/components/NewWorkspaceModal.js
```diff
@@ -118,7 +118,8 @@ const NewWorkspaceModal = Utils.withDisplayName('NewWorkspaceModal', ({
     Ajax(signal).Groups.list().then(setAllGroups),
     !!cloneWorkspace && Ajax(signal).Workspaces.workspace(namespace, cloneWorkspace.workspace.name).checkBucketLocation(cloneWorkspace.workspace.googleProject, cloneWorkspace.workspace.bucketName)
       .then(locationResponse => {
-        setBucketLocation(locationResponse.location)
+        // For current phased regionality release, we only allow US workspace buckets.
+        isUSLocation(locationResponse.location) ? setBucketLocation(locationResponse.location) : setBucketLocation(defaultLocation)
```
zarsky-broad: Could you nest the position of this ternary, and also destructure location?
```js
      .then(({ location }) => {
        setBucketLocation(isUSLocation(location) ? location : defaultLocation)
        setSourceWorkspaceLocation(location)
      })

```

src/components/region-common.js
```diff
@@ -64,26 +66,38 @@ export const allRegions = [
   // In this list, us-east*, us-west*, northamerica-northeast2 and asia-northeast2 have purposefully been removed.
   // This is to avoid creating within-country silos of life sciences community data.
   // So for US, Canada and Japan, we are restricting to one region.
-  // For more information, see https://docs.google.com/document/d/1RMu8bxXAyP2q_85UHkPARW1Xpt_G_662pRi0LCML4yA/edit#heading=h.53j9ueoid5vt
-  // TODO(wnojopra): Replace the above link with the Terra support article link once published
+  // For more information, see https://support.terra.bio/hc/en-us/articles/360060777272-US-regional-versus-Multi-regional-US-buckets-trade-offs
```
zarsky-broad, line 69: Looks like this (and `unknownRegionFlag`) no longer needs to be exported?

src/components/region-common.js
```diff
@@ -64,26 +66,38 @@ export const allRegions = [
...
+  { value: 'ASIA-NORTHEAST1', label: 'asia-northeast1 (Tokyo)', locationType: locationTypes.region },
+  { value: 'ASIA-NORTHEAST3', label: 'asia-northeast3 (Seoul)', locationType: locationTypes.region },
+  { value: 'ASIA-SOUTH1', label: 'asia-south1 (Mumbai)', locationType: locationTypes.region },
+  { value: 'ASIA-SOUTHEAST1', label: 'asia-southeast1 (Singapore)', locationType: locationTypes.region },
+  { value: 'ASIA-SOUTHEAST2', label: 'asia-southeast2 (Jakarta)', locationType: locationTypes.region },
+  { value: 'AUSTRALIA-SOUTHEAST1', label: 'australia-southeast1 (Sydney)', locationType: locationTypes.region }
 ]
+
+// For current phased release of regionality only supporting US and US-CENTRAL1 buckets.
+export const availableBucketRegions = _.slice(0, 2, allRegions)
```
zarsky-broad: Based on the comment, would this (or something similar) be clearer about the intention?
```suggestion
export const availableBucketRegions = _.filter(({ value }) => _.startsWith('US', value), allRegions)
```

src/libs/runtime-utils.js
```diff
@@ -77,14 +87,33 @@ export const getValidGpuTypes = (numCpus, mem, zone) => {
...
   return cpuPrice * numInstances * dataprocCpuPrice
 }
 
+const getHourlyCostForMachineType = (machineTypeName, region, isPreemptible) => {
+  const { cpu, memory } = _.find({ name: machineTypeName }, machineTypes)
+  const { n1HourlyCpuPrice, preemptibleN1HourlyCpuPrice, n1HourlyGBRamPrice, preemptibleN1HourlyGBRamPrice } = _.find({ name: _.toUpper(region) },
+    regionToPrices)
+  return !isPreemptible ?
+    (cpu * n1HourlyCpuPrice) + (memory * n1HourlyGBRamPrice) :
+    (cpu * preemptibleN1HourlyCpuPrice) + (memory * preemptibleN1HourlyGBRamPrice)
```
zarsky-broad: Could you flip this? We like positive case first, when reasonably possible:
```suggestion
  return isPreemptible ?
    (cpu * preemptibleN1HourlyCpuPrice) + (memory * preemptibleN1HourlyGBRamPrice) :
    (cpu * n1HourlyCpuPrice) + (memory * n1HourlyGBRamPrice)
```

src/libs/runtime-utils.js
```diff
@@ -77,14 +87,33 @@ export const getValidGpuTypes = (numCpus, mem, zone) => {
...
+}
+
+const getGpuCost = (gpuType, numGpus, region) => {
+  const prices = _.find({ name: region }, regionToPrices)
+  const price = Utils.switchCase(gpuType,
```
zarsky-broad: Would this be assuming too much about naming?
```js
const price = prices[`${_.last(_.split('-', gpuType))}HourlyPrice`]
```

src/libs/runtime-utils.js
kyuksel: The reason for having the negative case first was because `isPreemptible` will be `false` most of the time this ternary is executed and I wanted the evaluation to be able to stop sooner. I understand that it's probably premature optimization and don't feel strongly about it.

zarsky-broad: All other things being equal, then, I'd prefer to avoid the premature optimization.

wnojopra:
Thanks @zarsky-broad . I have your changes implemented but I'm running into trouble testing because I'm suddenly getting 403s from sam. I've posted on slack and will update here asap.

kyuksel:
> The discrepancy with the 4 P4 GPUs is due to a bug on dev. Take a look at the price for 4 P4 GPUs on dev. It is 1.8000. Because the price for one P4 GPU is 0.6000, the price for four should be 4x that, or 2.4000. The price is correct in my PR.

💯 Thank you!

kyuksel (approved):
👍🏼  pending Isaac's comments, rebasing and having tests passed.
Thanks for sticking with this! 🏆

src/components/NewWorkspaceModal.js

src/components/region-common.js
wnojopra: Done.

src/components/region-common.js
wnojopra: Done.

src/libs/runtime-utils.js
wnojopra: Done.

src/libs/runtime-utils.js
wnojopra: Done. I think when gpus that aren't nvidia are supported by Google Cloud we'll have to do something else. But for now, this makes sense.

PR #2681, opened 2021-11-05
[AS-1009] Add metrics for different menu items accessible through entities table

jordansafer:
Added metrics for entity tables in the Workspace Data page. New metrics for: Delete selected rows, Export rows to another workspace, and Open with: IGV, Workflow, Notebook, and Data Explorer
Updated metric for downloading selected rows as a TSV, so we can differentiate from users selecting the Download All Rows button

Testing: Tested all operations work as expected after the change. Verified the new metrics are populated in MixPanel for terra dev.


src/components/data/EntitiesContent.js
```diff
@@ -146,7 +149,12 @@ const ToolDrawer = _.flow(
               text: 'Workflow'
             }),
             h(ModalToolButton, {
-              onClick: !openDataExplorerInSameTab ? onDismiss : undefined,
+              onClick: () => {
+                Ajax().Metrics.captureEvent(Events.workspaceDataOpenWithDataExplorer, {
+                  ...extractWorkspaceDetails(workspace.workspace)
+                })
+                if (!openDataExplorerInSameTab) onDismiss()
```
petesantos: nit: we typically use short circuiting for these changes
```suggestion
                !openDataExplorerInSameTab && onDismiss()
```

petesantos (approved):
Looks good - one minor nit

src/components/data/EntitiesContent.js
jordansafer: Fixing!

src/components/data/EntitiesContent.js
```diff
@@ -391,6 +399,9 @@ const EntitiesContent = ({
           setDeletingEntities(false)
           setSelectedEntities({})
           setRefreshKey(_.add(1))
+          Ajax().Metrics.captureEvent(Events.workspaceDataDelete, {
+            ...extractWorkspaceDetails(workspace.workspace)
+          })
```
sagehen03: I don't think it's necessary to use the spread operator to create a new object.
```suggestion
          Ajax().Metrics.captureEvent(Events.workspaceDataDelete, extractWorkspaceDetails(workspace.workspace))
```

sagehen03 (approved):
I don't see any bugs but I think we may be creating extra objects to pass to the payload for captureEvent.

src/components/data/EntitiesContent.js
jordansafer: fixing!

ansingh7115 (approved):

PR #2682, opened 2021-11-08
Add ability to create Cromwell instance from Analyses tab [BW-819, BW-910]

cahrens:
This allows one to create a Cromwell app on the new Analyses tab. It is feature-flagged off by default. To enable it, execute the following in the JavaScript console:

`configOverridesStore.set({ isAnalysisTabVisible: true, isCromwellAppVisible: true })`

This MR is deployed at https://pr-2-dot-bvdp-saturn-dev.appspot.com/. To enable Launch button and interact with the Cromwell Swagger page, you have to modify the headers to make the referer https://leonardo.dsde-dev.broadinstitute.org.

Things that work:
1) Should be able to create a Cromwell and bring up the Swagger page through Launch (once the app has finished provisioning).
2) You should not be allowed to create more than one Cromwell.
3) You are given the option* to Pause/Resume from the Compute modal dialog (within the Analyses tab), though the functionality does not reliably work
4) You can delete Cromwell from the Cloud Environments page. 
5) Hopefully nothing about Galaxy got broken (biggest risk is around assumptions that an app is always Galaxy and a persistent disk is always related to Galaxy).
6) Cromwell state is displayed in the Environment context widget on the right-hand side.

Things that don't work:
1) No ability to customize anything about the Cromwell deployment
2) Does not display under "Your Analyses".
3) No explanation of costs associated with a deployment.
4) No fancy graphics.
5) The data disk is not reused from a previously created Cromwell (new persistent disk every time).
6) Pause/Resume functionality is not working properly*-- see https://broadworkbench.atlassian.net/browse/BW-908
7) No tests. :(

image


src/components/AnalysisModal.js
```diff
@@ -28,7 +39,7 @@ const environmentMode = Symbol('environment')
 
 export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
   ({
-    isOpen, onDismiss, onSuccess, uploadFiles, openUploader, runtimes, apps, galaxyDataDisks, refreshRuntimes, refreshApps, refreshAnalyses,
+    isOpen, onDismiss, onSuccess, uploadFiles, openUploader, runtimes, apps, appDataDisks, refreshRuntimes, refreshApps, refreshAnalyses,
```
cahrens, line 42: These disks will include the persistent disk used by the Cromwell App.

src/components/AnalysisModal.js
```diff
@@ -40,7 +51,7 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
 
     const currentRuntime = getCurrentRuntime(runtimes)
     const currentRuntimeTool = currentRuntime?.labels?.tool
-    const currentApp = getCurrentApp(apps)
+    const currentApp = toolLabel => getCurrentAppForType(toolToAppTypeMap[toolLabel])(apps)
```
cahrens: We intend the limit the user to a single Cromwell per workspace, so extending the pattern in place for Galaxy (though IMO it would be cleaner to have the code be robust to any number of instances even if in practice we prevent users from creating multiple ones).

src/components/CromwellModal.js
```diff
@@ -0,0 +1,87 @@
...
+import Events, { extractWorkspaceDetails } from 'src/libs/events'
+import {
+  computeStyles,
+  getCurrentAppForType
+} from 'src/libs/runtime-utils'
+import * as Utils from 'src/libs/utils'
+
+
+const defaultDataDiskSize = 500 // GB
+const defaultKubernetesRuntimeConfig = { machineType: 'n1-highmem-8', numNodes: 1, autoscalingEnabled: false }
```
cahrens, line 19: @cjllanwarne what do you think of these as defaults?

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -267,10 +292,11 @@ export const CloudEnvironmentModal = ({
   )
 
   const getToolLaunchClickableProps = toolLabel => {
-    const doesCloudEnvForToolExist = currentRuntimeTool === toolLabel || (currentApp && toolLabel === tools.galaxy.label)
+    const app = currentApp(toolLabel)
+    const doesCloudEnvForToolExist = currentRuntimeTool === toolLabel || app
     // TODO what does cookieReady do? Found it in the galaxy app launch code, is it needed here?
```
cahrens, line 288: @kyuksel I would like to understand more about why this cookie is necessary (for the MR deployment, one has to modify the request header in order for the cookie access to work). I will add this to the Tuesday agenda.

src/components/CromwellModal.js
```diff
@@ -0,0 +1,88 @@
...
+        div({ style: { display: 'flex', marginTop: '2rem', justifyContent: 'flex-end' } }, [
+          renderActionButton()
+        ])
+      ])
+    }
+    return h(Fragment, [renderDefaultCase(), loading && spinnerOverlay])
+  }
+)
+
+export const CromwellModal = withModalDrawer({ width: 675, 'aria-labelledby': titleId })(CromwellModalBase)
```
cahrens, line 90: It doesn't need to be so wide at this point, but I was going with Galaxy's size for consistency.

src/components/ContextBar.js
```diff
@@ -102,7 +106,6 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
           }, [icon('ellipsis-v', { size: 24 })])
         ]),
         h(Clickable, {
-          'aria-label': 'Cloud env config',
```
cahrens, line 114: Aria label is unnecessary because the tooltip (which is being used as the label) will be read by the screen reader… I verified this behavior.

Note also that when it is necessary to add an aria label, it is good to use complete words so it makes more sense to the user.

src/pages/Environments.js
```diff
@@ -339,7 +339,7 @@ const Environments = () => {
             cellRenderer: ({ rowIndex }) => {
               const cloudEnvironment = filteredCloudEnvironments[rowIndex]
               return isApp(cloudEnvironment) ?
-                (cloudEnvironment.appType ? _.capitalize(cloudEnvironment.appType) : 'Galaxy') :
+                _.capitalize(cloudEnvironment.appType) :
```
cahrens, line 344: Changes in this file are cleanup from an earlier MR, now that the Leo changes are in production (https://broadworkbench.atlassian.net/browse/BW-890).

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -328,7 +328,7 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
       }, [
         workspace && h(WrappedComponent, {
           ref: child,
-          workspace, refreshWorkspace, refreshRuntimes, refreshApps, runtimes, persistentDisks, galaxyDataDisks,
+          workspace, refreshWorkspace, refreshRuntimes, refreshApps, runtimes, persistentDisks, galaxyDataDisks, apps,
```
cahrens: This is a bug fix that is also included in https://github.com/DataBiosphere/terra-ui/pull/2678

src/libs/ajax.js
```diff
@@ -1388,7 +1388,7 @@ const Apps = signal => ({
             name: diskName,
             size: diskSize,
             labels: {
-              saturnApplication: 'galaxy',
+              saturnApplication: appType,
```
cahrens, line 1396: I don't where this label is manifested, but hopefully changing it to GALAXY or CROMWELL should be fine.

src/components/AnalysisModal.js
```diff
@@ -118,12 +131,12 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
       })
     })
 
-    const renderGalaxyModal = () => h(GalaxyModalBase, {
-      isOpen: viewMode === tools.galaxy.label,
+    const renderAppModal = (AppModalBase, toolLabel) => h(AppModalBase, {
```
cjllanwarne: Sort of strange that we have to pass two dependent arguments in here... Could we calculate one from the other within this function, or bake the appropriate tool labels into the modal bases?

And should `AppModalBase` be lowercase?

src/components/ContextBar.js
```diff
@@ -72,17 +73,20 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
   }
 
   const getEnvironmentStatusIcons = () => {
+    const galaxyApp = getCurrentAppForType(tools.galaxy.appType)(apps)
+    const cromwellApp = getCurrentAppForType(tools.cromwell.appType)(apps)
```
cjllanwarne: "I suspect I can predict the answer to this, but... "

Could we reframe this as an arbitrary list of apps rather than hard-coding two that we know we care about today?

src/components/ContextBar.js
cjllanwarne: TIL

src/components/CromwellModal.js
```diff
@@ -0,0 +1,88 @@
...
+    })
+
+    const renderActionButton = () => {
+      return !app ? h(ButtonPrimary, { onClick: createCromwell }, ['Create']) :
+        h(ButtonPrimary, { onClick: onDismiss }, ['Close'])
+    }
+
+    const getEnvMessageBasedOnStatus = app => {
+      const waitMessage = 'This process will take up to a few minutes.'
+      const nonStatusSpecificMessage = 'A cloud environment for Cromwell consists of application configuration, cloud compute and persistent disk(s).'
```
cjllanwarne: "cloud compute" is potentially misleading. We do include running "the cromwell service in the cloud" but doesn't include "further VMs running jobs started by cromwell"

src/components/CromwellModal.js
```diff
@@ -0,0 +1,88 @@
...
+
+      return !app ?
+        nonStatusSpecificMessage :
+        Utils.switchCase(app.status,
+          ['PROVISIONING', () => `The cloud compute is provisioning.`],
```
cjllanwarne: This might also take up to a few minutes

src/components/notebook-utils.js
```diff
@@ -73,11 +73,18 @@ export const tools = {
...
   spark: { label: 'spark' },
-  galaxy: { label: 'galaxy' }
+  galaxy: { label: 'galaxy', appType: 'GALAXY' },
+  cromwell: { label: 'cromwell', appType: 'CROMWELL' }
 }
 
 const toolToExtensionMap = { [tools.RStudio.label]: tools.RStudio.ext, [tools.Jupyter.label]: tools.Jupyter.ext }
 
+// Mapping of tool label to appType (or undefined if tool is not an app).
+export const toolToAppTypeMap = { [tools.galaxy.label]: tools.galaxy.appType, [tools.cromwell.label]: tools.cromwell.appType }
```
cjllanwarne: Thought 1: Is there a way of `_.flow` ing this something like:
- Take tools
- Filter by existence of `appType`
- Map to `label -> appType`

cjllanwarne: Thought 2: Why are the keys in this object arrays?

cjllanwarne: Thought 3: Isn't this just a fancy way of rephrasing object lookups? eg aren't these the same:
```
const way1 = toolToAppTypeMap('cromwell')
const way2 = tools.cromwell.Apptype
```

src/libs/ajax.js
```diff
@@ -1388,7 +1388,7 @@ const Apps = signal => ({
...
               saturnWorkspaceNamespace: namespace,
               saturnWorkspaceName: workspaceName
```
cjllanwarne, line 1398: worth checking with the IA team but I think `saturn` is a ye-olde name for "terra" and might be worth migrating away from?

src/libs/runtime-utils.js
```diff
@@ -222,33 +223,37 @@ export const machineCost = machineType => {
   return _.find(knownMachineType => knownMachineType.name === machineType, machineTypes).price
 }
 
-export const getCurrentApp = _.flow(trimAppsOldestFirst, _.last)
+export const getCurrentAppForType = appType => _.flow(trimAppsOldestFirst, _.remove(_ => _.appType !== appType), _.last)
 
-export const currentAppIncludingDeleting = _.flow(_.sortBy('auditInfo.createdDate'), _.last)
+// DEPRECATED: this assumes that Galaxy is the only app that could be running.
+// Use getCurrentAppForType instead.
+export const getCurrentApp = getCurrentAppForType(tools.galaxy.appType)
```
cjllanwarne: Is it worth adding a console log along the lines of "you used a deprecated function" here?

src/libs/runtime-utils.js
```diff
@@ -222,33 +223,37 @@ export const machineCost = machineType => {
   return _.find(knownMachineType => knownMachineType.name === machineType, machineTypes).price
 }
 
-export const getCurrentApp = _.flow(trimAppsOldestFirst, _.last)
+export const getCurrentAppForType = appType => _.flow(trimAppsOldestFirst, _.remove(_ => _.appType !== appType), _.last)
```
cjllanwarne: Is `_.remove(badPredicate)` a negative rephrasing of an alternative `_.filter(positivePredicate)`? If so, for readability maybe the positive predicate version would be easier for humans to parse in this case?

src/libs/runtime-utils.js
```diff
@@ -222,33 +223,37 @@ export const machineCost = machineType => {
...
+export const currentAttachedDataDisk = (app, appDataDisks) => {
+  return _.find({ name: app?.diskName }, appDataDisks)
 }
 
 export const appIsSettingUp = app => {
   return app && (app.status === 'PROVISIONING' || app.status === 'PRECREATING')
 }
 
-export const currentPersistentDisk = (apps, galaxyDataDisks) => {
+export const currentGalaxyPersistentDisk = (apps, appDataDisks) => {
```
cjllanwarne, line 238: Is this kind of check not relevant to apps other than Galaxy too?

src/libs/utils.js
```diff
@@ -197,7 +197,7 @@ export const abandonedPromise = () => {
 
 export const generateRuntimeName = () => `saturn-${uuid()}`
 
-export const generateGalaxyAppName = () => `saturn-k8-${uuid()}`
+export const generateAppName = () => `saturn-k8-${uuid()}`
```
cjllanwarne: I don't think I've ever seen apps actually named like this, is it still used anywhere...?

src/components/AnalysisModal.js
cahrens: I will make AppModalBase lowercase.

I agree that it would be nicer to not have to pass in the tool label as well, and I experimented with getting rid of that. However, I'm not finding a nice solution without changing the input arguments to GalaxyModalBase/CromwellModalBase, and I believe that there is a desire to have a common set of input arguments to these classes and the similar tool constructors.

I think this issue would go away if we refactor to isolate the tool/app-specific logic in a single place and expose things like the compute modal via an API. Until then, I am of the opinion that it isn't worth trying to fix this specific case.

src/components/ContextBar.js
cahrens: I would love an arbitrary list of apps/tools, but for this first submission I was trying to minimize the refactoring.

In general, I think that each of the "tool" instances should implement a series of methods, and these infrastructure classes should iterate through them and call the method on each one. There is some complexity though in the difference in behavior between the "runtime/tool" things (RStudio/Jupyter) and the "apps" (Galaxy/Cromwell). Perhaps we could also push into the "tool" instances knowledge of whether or not they are running, and that might help.

src/components/CromwellModal.js
cahrens: I put very little thought into this message and am happy to change it to whatever you want! I probably took the Galaxy wording and substituted Cromwell.

src/components/notebook-utils.js
cahrens: I don't want to have to put "tools.cromwell" because I'm trying to generalize to whatever the current tool label is.

I wanted to do something like your first thought. But being new to lodash, I couldn't figure out how to make the flow stuff work correctly. After a couple of hours of frustration, I decided to follow the model of toolToExtensionMap and move on with my life. But I'm happy to replace this with a nicer implementation… if I can figure it out!

src/libs/ajax.js
cahrens: I wondered why Saturn was here. And I don't know how these labels are used.

src/libs/runtime-utils.js
cahrens: My hope was that we will get rid of all getCurrentApp usages as we fix the rest of the Cromwell issues. I will actually add a list to our "ui tasks" to go through all the remaining usages and fix them. They may be contributing to some of our issues (for instance, the app polling is using this method).

https://broadworkbench.atlassian.net/browse/BW-910

src/libs/runtime-utils.js
cahrens: It may be at some point… but not yet. With this ticket we are creating a new persistent disk for every Cromwell instance, instead of trying to reuse a persistent disk if it exists. We also aren't providing a control to delete the disk. 

The Galaxy code attempts to do management of the persistent disk that we haven't gotten to yet.

src/libs/utils.js
cahrens: Yes, this is the name that is associated with the app when it is created, and it is used for pause/resume/delete.

src/components/CromwellModal.js
cahrens: We might also just want to wait until we get some feedback from UX on what the wording should be (though I'm also very happy to temporarily put in whatever text you think is best).

src/components/CromwellModal.js
cahrens: Changing to "The cloud compute is provisioning, which may take several minutes".

src/libs/runtime-utils.js
cahrens: Perhaps it depends if you are a "glass half full or a glass half empty" person. :) But sure, I will make the change.

src/libs/utils.js
cahrens: This name also shows up on the Cloud Environments page under details.

src/components/notebook-utils.js
cahrens: I think I figured out something reasonable! I have pushed it up.

src/components/RuntimeManager.js
```diff
@@ -430,7 +430,7 @@ export default class RuntimeManager extends PureComponent {
         h(GalaxyModal, {
           workspace,
           apps,
-          galaxyDataDisks,
+          appDataDisks: galaxyDataDisks,
```
cahrens: I missed these usages before because GalaxyModal wraps around GalaxyModalBase.

src/pages/Environments.js
```diff
@@ -207,7 +207,7 @@ const Environments = () => {
 
   const renderBillingProjectApp = app => {
     const inactive = !_.includes(app.status, ['DELETING', 'ERROR', 'PREDELETING']) &&
-      getCurrentApp(appsByProject[app.googleProject]) !== app
+      getCurrentAppForType(app.appType)(appsByProject[app.googleProject]) !== app
```
cahrens, line 211: @cjllanwarne this fixes the issue you saw where you created a Cromwell instance (only) in a Workspace, and it showed a warning on the Cloud Environments page. Now warnings are only shown when there are 2 apps of the same type within a project.

cahrens:
@cjllanwarne FYI I did a force push because I rebased to resolve merge conflicts with dev.

src/components/AnalysisModal.js
```diff
@@ -118,12 +131,12 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
...
+    const renderAppModal = (appModalBase, toolLabel) => h(appModalBase, {
+      isOpen: viewMode === toolLabel,
       isAnalysisMode: true,
       workspace,
       apps,
-      galaxyDataDisks,
+      appDataDisks,
       onDismiss: () => {
         setViewMode(undefined)
         onDismiss()
```
cahrens, line 137: @jdcanas @kyuksel In the MR that merged this morning, the onSuccess callback was changed in the other location where GalaxyModalBase is created (CloudEnvironmentModal). However, it was not changed here. Was that intentional?

https://github.com/DataBiosphere/terra-ui/pull/2678/files#diff-91cb757d5be0abe83ebdf3f4d8b3f4d78a4262d118db2990d542d30a5c7fd15dL84

src/components/ContextBar.js
```diff
@@ -74,9 +75,12 @@ export const ContextBar = ({ setDeletingWorkspace, setCloningWorkspace, setShari
...
     return h(Fragment, [
       ...(currentRuntime ? [getIconForTool(currentRuntimeTool, currentRuntime.status)] : []),
-      ...(currentApp ? [getIconForTool(tools.galaxy.label, currentApp.status)] : [])
+      ...(galaxyApp ? [getIconForTool(tools.galaxy.label, galaxyApp.status)] : []),
+      ...(cromwellApp ? [getIconForTool(tools.cromwell.label, cromwellApp.status)] : [])
```
cahrens, line 83: I think the status indication difference now between provisioning/restarting (all the INGs besides running) and running is too subtle. Here is an example of Galaxy restarting and Cromwell running.
image

src/components/notebook-utils.js
```diff
@@ -80,11 +80,14 @@ export const tools = {
...
+// Returns appType for app with given label, or undefined if tool is not an app.
+export const getAppType = label => {
+  const app = _.find(_ => _.label === label)(tools)
+  return app?.appType
+}
 
-// Returns true iff the tool label corresponds to an app.
-export const isToolAnApp = label => label in toolToAppTypeMap
+// Does the tool label correspond to an app?
+export const isToolAnApp = label => getAppType(label)
```
cjllanwarne: Is this a type coercion relying on the "falsiness" of `undefined`? Could we make it explicit by saying (essentially)...?
```
export const isToolAnApp = label => (getAppType(label) != undefined)
```

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -218,8 +219,12 @@ const useAppPolling = (googleProject, workspaceName) => {
         await Ajax(signal).Apps.list(googleProject, { creator: getUser().email, saturnWorkspaceName: workspaceName }) :
         []
       setApps(newApps)
-      const app = getCurrentApp(newApps)
-      reschedule((app && _.includes(app.status, ['PROVISIONING', 'PREDELETING'])) ? 10000 : 120000)
+      _.forOwn(tool => {
+        if (tool.appType) {
+          const app = getCurrentAppForType(tool.appType)(newApps)
+          reschedule((app && _.includes(app.status, ['PROVISIONING', 'PREDELETING'])) ? 10000 : 120000)
+        }
```
cjllanwarne, line 228: I'm finding this hard to parse inline... could we extract it into a separate, named method to make the intention clearer? ... or maybe just add a comment?

cjllanwarne (approved):
LGTM from a Cromwell App point of view. I haven't tested them in full and so can't say for sure whether there are any unintended consequences of the changes for Galaxy in removing any of the "everything is Galaxy" assumptions.

src/components/notebook-utils.js
cahrens: Yep, this is falsiness in all of its glory! TBH, it seems no less mysterious than the Scala code (to me). :)

src/pages/workspaces/workspace/WorkspaceContainer.js
cahrens: Are you talking specifically about line 225 (the reschedule line)? That was existing code, so I would prefer to leave it.

I believe the intention is that if the app is in the provisioning or predeleting state statue, we want to do the app polling again more quickly so that we more accurately show when it leaves those states.

cahrens: And before this was only being done for Galaxy. I generalized it to have the same behavior for all apps.

src/components/notebook-utils.js
cahrens: I modified the method so that it returns a boolean.

cahrens (commented):
Sorry, I never intended to make this a review!

@kyuksel @cjllanwarne I have added a feature flag (disabled by default). To enable: configOverridesStore.set({ isAnalysisTabVisible: true, isCromwellAppVisible: true })

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -85,28 +87,47 @@ export const CloudEnvironmentModal = ({
...
+  const renderGalaxyModal = () => renderAppModal(GalaxyModalBase, NEW_GALAXY_MODE)
+  const renderCromwellModal = () => renderAppModal(CromwellModalBase, NEW_CROMWELL_MODE)
 
-  const renderDefaultPage = () => div({ style: { display: 'flex', flexDirection: 'column', flex: 1 } }, [
-    renderToolButtons(tools.Jupyter.label),
-    renderToolButtons(tools.RStudio.label),
-    renderToolButtons(tools.galaxy.label)
-  ])
+  const renderDefaultPage = () => div({ style: { display: 'flex', flexDirection: 'column', flex: 1 } },
+    _.map(tool => renderToolButtons(tool.label))(getToolsToDisplay)
```
cahrens, line 92: This is the sort of change that I'd like the code to move incrementally towards… removing case statements/listing of individual tools and instead iterating over tools to display their elements.

Some examples of things that could easily be moved into the tools object is tool images and editor default sizes.

cjllanwarne (approved):

src/components/CromwellModal.js
kyuksel: Probably not worth dwelling on it further but since Cromwell users currently have no control over `application configuration`, `cloud compute` or `persistent disk(s)`, it may not make sense to display this text at all. The original text this was copied from is meant to go along with a modal like

image

where there are corresponding sections.

If we want to display _something_, I'd suggest some text that says `Cromwell will run your jobs on the same compute where it will be deployed.` so users know it won't submit jobs to PAPI, etc. (unless submitting jobs is an implementation detail meant to be hidden from users today).

kyuksel:
image

1. Since the persistent disk cannot be reused with a new Cromwell app at this time, should we make it the only option to delete the disk while deleting the corresponding app?
2. We should change references to `Galaxy` in the text to `Cromwell`.

cahrens:
> image
> 
> 1. Since the persistent disk cannot be reused with a new Cromwell app at this time, should we make it the only option to delete the disk while deleting the corresponding app?
> 2. We should change references to `Galaxy` in the text to `Cromwell`.

We are already working on making it so that we can reuse the persistent disk, so it doesn't make sense to force the deletion.

However, we should change the text so that it does not hardcode Galaxy (I never bothered to read it!). I don't know yet what to do about the specifics of how long it takes to create a new environment, how to copy over things you might want to save yet, etc. This will probably end up as something we need to work with UX on, but I will see what is practical in the short term.

kyuksel:
> We are already working on making it so that we can reuse the persistent disk, so it doesn't make sense to force the deletion.

Gotcha. One thing I noticed around that is, when I did not delete my disk while deleting my Cromwell app, and then create another Cromwell app, I got a misleading error/danger triangle (since it says there are multiple _cloud environments_ whereas it's actually multiple _disks_) on the main page

image

where clicking on the error sign takes you to Cloud Environments page with warnings next to the two Cromwell disks

image

Because Cromwell will be behind a feature flag and you folks are working on implementing better support for it, I don't see this as a blocker but wanted to note anyway.

cahrens:
> > We are already working on making it so that we can reuse the persistent disk, so it doesn't make sense to force the deletion.
> 
> Gotcha. One thing I noticed around that is, when I did not delete my disk while deleting my Cromwell app, and then create another Cromwell app, I got a misleading error/danger triangle (since it says there are multiple _cloud environments_ whereas it's actually multiple _disks_) on the main page
> 
> image
> 
> where clicking on the error sign takes you to Cloud Environments page with warnings next to the two Cromwell disks
> 
> image
> 
> Because Cromwell will be behind a feature flag and you folks are working on implementing better support for it, I don't see this as a blocker but wanted to note anyway.

This is an existing bug/assumption in the [code](https://github.com/DataBiosphere/terra-ui/blob/c0ceb2e8ff4e7700e43f981126d3feb58ce61f97/src/components/RuntimeManager.js#L375). If I am reading the code correctly, you would see the same behavior if you create a Galaxy instance, deleted it without deleting the persistent disk, and then created a new Galaxy instance using a new disk. But perhaps that isn't possible because Galaxy will always use the already existing disk in that case?? Please let me know what the Galaxy behavior is.

I think this falls into the category of things that we will have to clean up if we decide to allow the user to create a Cromwell instance and NOT reuse an existing disk. In our next planned iteration, we plan to always reuse an existing disk.

kyuksel:
> But perhaps that isn't possible because Galaxy will always use the already existing disk in that case??

Correct.

> I think this falls into the category of things that we will have to clean up if we decide to allow the user to create a Cromwell instance and NOT reuse an existing disk. In our next planned iteration, we plan to always reuse an existing disk.

Sounds good.

src/components/AnalysisModal.js
```diff
@@ -3,20 +3,31 @@ import { Fragment, useState } from 'react'
...
+  analysisNameValidator,
+  getAppType,
+  getDisplayName,
+  getTool,
+  isToolAnApp,
+  notebookData,
+  tools
+} from 'src/components/notebook-utils'
 import TitleBar from 'src/components/TitleBar'
+import cromwellImg from 'src/images/cromwell-logo.png'
```
kyuksel, line 22: We typically pre-process images by dragging and dropping them on https://tinypng.com and then downloading the resulting compressed image.

image

src/libs/runtime-utils.js
```diff
@@ -222,33 +223,33 @@ export const machineCost = machineType => {
   return _.find(knownMachineType => knownMachineType.name === machineType, machineTypes).price
 }
 
-export const getCurrentApp = _.flow(trimAppsOldestFirst, _.last)
+export const getCurrentAppForType = appType => _.flow(trimAppsOldestFirst, _.filter(_ => _.appType === appType), _.last)
```
kyuksel: In Scala, we typically use `_` when we don't use/care about an input variable. In Javascript (at least in this code base) we usually use `_` for Lodash library methods only. Therefore, it may be confusing to use `_` like above.

```suggestion
export const getCurrentAppForType = appType => _.flow(trimAppsOldestFirst, _.filter(app => app.appType === appType), _.last)
```
Similarly for other places in this file and `notebook-utils`

cahrens:
> > But perhaps that isn't possible because Galaxy will always use the already existing disk in that case??
> 
> Correct.
> 
> > I think this falls into the category of things that we will have to clean up if we decide to allow the user to create a Cromwell instance and NOT reuse an existing disk. In our next planned iteration, we plan to always reuse an existing disk.
> 
> Sounds good.

Actually, thinking about this more the code should be changed to check if there is more than one app instance per app type (instead of looking at the number of persistent disks). That's consistent with the message, and it will fix the current issue that this error icon would show if you have 1 Cromwell and 1 Galaxy in the same workspace. I will make the fix and push it up in a bit.

kyuksel:
> it will fix the current issue that this error icon would show if you have 1 Cromwell and 1 Galaxy in the same workspace. I will make the fix and push it up in a bit.

I haven't seen the error when having 1 Cromwell + 1 Galaxy when they had no additional disks. I saw it only when I had 2 disks for Cromwell. Regardless, I agree with 

> …the code should be changed to check if there is more than one app instance per app type (instead of looking at the number of persistent disks). That's consistent with the message...

cahrens:
> > it will fix the current issue that this error icon would show if you have 1 Cromwell and 1 Galaxy in the same workspace. I will make the fix and push it up in a bit.
> 
> I haven't seen the error when having 1 Cromwell + 1 Galaxy when they had no additional disks. I saw it only when I had 2 disks for Cromwell. Regardless, I agree with
> 
> > …the code should be changed to check if there is more than one app instance per app type (instead of looking at the number of persistent disks). That's consistent with the message...

You are right, one doesn't see that error. That's because what the code is doing is taking all the persistent disks and removing the Galaxy ones… so if you have one Cromwell and one Galaxy, there's not an issue.

https://github.com/DataBiosphere/terra-ui/blob/c0ceb2e8ff4e7700e43f981126d3feb58ce61f97/src/pages/workspaces/workspace/WorkspaceContainer.js#L187

Now I don't think I should make the change until I better understand what we intend to show in that "Cloud Environment" widget. Which Cloud Environment is it showing statistics for? Why does it say None even though Galaxy is running? What is it meant to represent when you have say, a Jupyter environment, Galaxy, and Cromwell all at once? Will this widget go away with the new Analyses tab?

image

(I've also wondering if the Galaxy-specific widget to the right of it will go away with the new Analyses tab).

src/libs/runtime-utils.js
cahrens: Sure, fixed these (and did a little simplification at the same time).

src/components/AnalysisModal.js
cahrens: This is not the final image yet (Tiffany is still working on it), but I will compress the placeholder one.

kyuksel:
image

represents Jupyter/RStudio runtimes. It as well as

image

will go away (along with Notebooks tab) once Analyses Tab is released.

Does this clarify things a bit?

src/components/CromwellModal.js
cahrens: Simplified the message to just ask if you want to create a Cromwell instance to run workflows. Also added Cromwell to the modal title.

cahrens:
> > image
> > 
> > 1. Since the persistent disk cannot be reused with a new Cromwell app at this time, should we make it the only option to delete the disk while deleting the corresponding app?
> > 2. We should change references to `Galaxy` in the text to `Cromwell`.
> 
> We are already working on making it so that we can reuse the persistent disk, so it doesn't make sense to force the deletion.
> 
> However, we should change the text so that it does not hardcode Galaxy (I never bothered to read it!). I don't know yet what to do about the specifics of how long it takes to create a new environment, how to copy over things you might want to save yet, etc. This will probably end up as something we need to work with UX on, but I will see what is practical in the short term.

Made the Galaxy-specific message only displayed for Galaxy apps. In the Cromwell case it will only ask if you wish to also delete the persistent disk-- I expect we will want to add wording later about the impact of deleting the persistent disk, but we need to work with product/UX to figure out the wording. Perhaps we will be able to come up with something general enough that it works for both Galaxy and Cromwell, or perhaps we will have Cromwell-specific language.

kyuksel:
image

I'm guessing adding a `RESUME` and `DELETE` buttons above will be part of the subsequent work?

Relatedly and for my edification, can a Cromwell app be deleted if it's in paused state. For context, Galaxy cannot be.

cahrens:
> above

Probably… if we decide that it is worth allowing them to pause/resume vs. just deleting and recreating Cromwell. We (dev and product) have questions about the use case and costs related to pause.

For Cromwell, deleting and recreating an instance  is pretty fast. And pause is also not working properly for Cromwell at the moment, so we may remove it as an option for Cromwell specifically on the Cloud Environments page and in the Cloud Environments modal (https://broadworkbench.atlassian.net/browse/BW-908). I have it on my list to find out more about how Pause is implemented, and what the benefits are.

cahrens:
> image
> 
> I'm guessing adding a `RESUME` and `DELETE` buttons above will be part of the subsequent work?
> 
> Relatedly and for my edification, can a Cromwell app be deleted if it's in paused state. For context, Galaxy cannot be.

Deciding what we want to show in this modal for Cromwell is something we want to work out with UX. For this submission, it was really just about enabling the "Start" for Cromwell. I imagine Delete will show up here at some point, but TBD.

And I don't know about Deleting while it is Paused-- like I mentioned in my last comment, we don't yet understand the technical details of the Pause state.

kyuksel:
We have a [leonardo-cron-jobs](https://github.com/broadinstitute/leonardo-cron-jobs) repo for processes that intermittently run to detect/fix anomalies, and cleaning resources up. I didn't look into the code exhaustively but we made Galaxy app-specific assumptions there too (like [this](https://github.com/broadinstitute/leonardo-cron-jobs/blob/4d844ca56a96471800c7ff572796fbad7735c246/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedDiskChecker.scala#L31)) that may need updating and/or we may need to add additional cron jobs for Cromwell. Resource clean-up part of things may actually be a high priority since it'll cost money if they're left behind. Can we create a ticket for it?

cahrens:
> image
> 
> represents Jupyter/RStudio runtimes. It as well as
> 
> image
> 
> will go away (along with Notebooks tab) once Analyses Tab is released.
> 
> Does this clarify things a bit?

It's good to know that those will go away. Given your explanation, I think the code change we want here is to remove both Galaxy and Cromwell persistent disks from that array, since that widget is meant to only be considering Jupyter/Studio runtime information.

cahrens:
@kyuksel I have pushed up the changes to remove Pause for Cromwell apps, and I completed the generalization of galaxyDataDisks to appDataDisks.

src/components/AnalysisModal.js
jdcanas: Ah, we probably need to fix it here too. The issue that the PR you linked addressed was that we had an erroneous extra layer of error reporting, and the baseModals did not properly dismiss on error. The baseModals now correctly will dismiss to show the error, and the error reporting in the parent modal is no longer needed.

I believe you actually mentioned seeing this corrected behavior briefly in the meeting we had :)

src/libs/ajax.js
jdcanas: Yeah, these labels are used in the backend for filtering based on workspace. They are stored in a label map, and it would be a non-trivial to make this change while maintaining backwards compatibility (it would require updating the labels on all non-deleted disks)

src/components/ContextBar.js
jdcanas: Yeah, the coloring is an active item we are working on, along with a laundry list of other UX items here: https://broadworkbench.atlassian.net/browse/IA-3033. Feel free to add any UX comments you have here!

src/components/CromwellModal.js
```diff
@@ -0,0 +1,88 @@
+import _ from 'lodash/fp'
+import { Fragment, useState } from 'react'
+import { div, h } from 'react-hyperscript-helpers'
+import { ButtonPrimary, spinnerOverlay } from 'src/components/common'
+import { withModalDrawer } from 'src/components/ModalDrawer'
+import { tools } from 'src/components/notebook-utils'
+import TitleBar from 'src/components/TitleBar'
+import { Ajax } from 'src/libs/ajax'
+import { withErrorReporting } from 'src/libs/error'
```
jdcanas: Use `withErrorReportingInModal` to get this modal to properly dismiss when errors occur, just added in a PR recently :)

src/components/AnalysisModal.js
cahrens: @jdcanas do you want me to just make the change in this MR so that the code is consistent between the 2 locations?

src/components/AnalysisModal.js
```diff
@@ -162,11 +180,17 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
         }
       }, [img({ src: rstudioLogo, style: styles.image })]),
       div({
-        style: { opacity: currentApp ? '0.5' : '1', ...styles.toolCard }, onClick: () => {
+        style: { opacity: galaxyApp ? '0.5' : '1', ...styles.toolCard }, onClick: () => {
```
jdcanas, line 172: Hmm... If I'm a user and I create a galaxy app, and then create a cromwell app what is the intended behavior?

It seems you've attempted to allow the two to co-exist, which is I think what we want.

That being said, if the Galaxy/Cromwell app is still creating, it allows me to attempt to create another app of a different type and the backend kicks back and says the K8 cluster is still creating. We may want to consider adding a check to see if ANY app is Precreating/Provisioning to the disabled flag here and within the context bar. Thoughts?

cahrens: I do think we want both to co-exist. I think that you shouldn't get the error, and instead it should just wait until the cluster is created and then proceed with the creation of the 2nd app. I will create a ticket for the problem.

The fact that Cromwell and Galaxy share a cluster is interesting… and perhaps something that we want to change. We have questions about the cluster lifecycle and the cost to the user when a cluster has been created, but apps have been deleted. Starting up a cluster takes a LONG time, so I anticipate users are going to want us to pre-create the cluster if at all possible.

src/components/CromwellModal.js
```diff
@@ -0,0 +1,88 @@
...
+        defaultKubernetesRuntimeConfig, diskName: Utils.generatePersistentDiskName(), diskSize: defaultDataDiskSize,
+        appType: tools.cromwell.appType, namespace, bucketName, workspaceName
+      })
+      Ajax().Metrics.captureEvent(Events.applicationCreate, { app: tools.cromwell.appType, ...extractWorkspaceDetails(workspace) })
+      return onSuccess()
+    })
+
+    const renderActionButton = () => {
+      return !app ? h(ButtonPrimary, { onClick: createCromwell }, ['Create']) :
+        h(ButtonPrimary, { onClick: onDismiss }, ['Close'])
```
jdcanas, line 47: should there be a delete button if there is a running app, or is that out of scope?

src/components/CromwellModal.js
```diff
@@ -0,0 +1,88 @@
+import _ from 'lodash/fp'
```
jdcanas, line 1: Would love to see some UX done for this modal in the long term 😄. Happy to help/participate in any of those discussions. 

I'm also curious if its worth thinking about ticketing something like an `AppModal` in an eventual refactor. Seems like most apps would have similar concerns here once the disk management converges... thoughts?

src/components/GalaxyModal.js
```diff
@@ -32,28 +33,28 @@ const titleId = 'galaxy-modal-title'
...
+    const attachedDataDisk = currentAttachedDataDisk(app, appDataDisks)
 
     const [dataDiskSize, setDataDiskSize] = useState(attachedDataDisk?.size || defaultDataDiskSize)
     const [kubernetesRuntimeConfig, setKubernetesRuntimeConfig] = useState(app?.kubernetesRuntimeConfig || defaultKubernetesRuntimeConfig)
     const [viewMode, setViewMode] = useState(undefined)
     const [loading, setLoading] = useState(false)
     const [shouldDeleteDisk, setShouldDeleteDisk] = useState(false)
 
-    const currentDataDisk = currentPersistentDisk(apps, galaxyDataDisks)
+    const currentDataDisk = currentGalaxyPersistentDisk(apps, appDataDisks)
```
jdcanas, line 49: should we have a cromwell-related TODO here when to note this is a place that needs changing when we evolve the disk management past delete/re-create?

Is there a ticket to link?

src/components/CromwellModal.js
cahrens: It's out of scope at this point- you have to go to the Cloud Environments page to delete Cromwell at the moment. This ticket was really just about starting Cromwell (and we want to work with UX to design what we display in the "Cromwell modal").

src/components/CromwellModal.js
cahrens: Yes, that is the plan! That's why I put in the most bare bones, ugly thing I could. :)

jdcanas: This also brings up an interesting question we haven't hit yet in terms of actually displaying the resources in the cluster the user configured for app#1 and allowing them to adjust when they create app#2.

cc @kyuksel.

src/components/AnalysisModal.js
jdcanas: that would be nice if you could

src/components/AnalysisModal.js
jdcanas: Yeah we've had discussions around this. The problem is when the cluster pre-creation occurs in the workspace lifecycle and cost. There's not really a good place besides directly when the user requests an app, since not all workspace users will use apps, and we don't want them paying for the cluster premium before that point. Right now, clusters are per-project in the backend.

src/components/AnalysisModal.js
```diff
@@ -75,10 +87,10 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
         }],
         [Utils.DEFAULT, () => Utils.cond(
           [currentTool === tools.RStudio.label || currentTool === tools.Jupyter.label, () => setViewMode(analysisMode)],
-          [currentTool === tools.galaxy.label && !currentApp, () => setViewMode(environmentMode)],
-          [currentTool === tools.galaxy.label && currentApp, () => {
+          [isToolAnApp(currentTool) && !app, () => setViewMode(environmentMode)],
+          [isToolAnApp(currentTool) && app, () => {
```
kyuksel: When we use non-boolean values in logical operations as truthy, we typically prefix them with `!!` to make the intention clear. `truthy/falsy` concepts can be fairly foreign for people coming from non-Javascript backgrounds.

```suggestion
          [isToolAnApp(currentTool) && !!app, () => {
```

src/components/AnalysisModal.js
```diff
@@ -75,10 +87,10 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
...
             console.error(
-              'This shouldn\'t be possible, as you aren\'t allowed to create a galaxy analysis when one exists; the button should be disabled.')
+              `This shouldn't be possible, as you aren't allowed to create a ${_.capitalize(app.appType)} instance when one exists; the button should be disabled.`)
```
kyuksel, line 92: Thanks for cleaning this up!

src/components/AnalysisModal.js
```diff
@@ -137,6 +150,8 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
         await refreshApps(true)
       })
     })
+    const renderGalaxyModal = () => renderAppModal(GalaxyModalBase, tools.galaxy.label)
+    const renderCromwellModal = () => renderAppModal(CromwellModalBase, tools.cromwell.label)
```
kyuksel: Wondering why we need these functions. Can't we inline the body where they are called?

src/components/AnalysisModal.js
kyuksel: Thanks @jdcanas for catching this!

I'd also agree that we want them to co-exist but would like to get UX/Product thoughts on it too. @cahrens, could you add us + Alessandro and Tiffany on the ticket you'll create so we can get on the same page.

> We have questions about the cluster lifecycle

Happy to chat about it 😊 

> the cost to the user when a cluster has been created, but apps have been deleted

We (`leonardo-cron-jobs`) detect when that happens and if it's been over a few hours, delete the cluster.

src/libs/ajax.js
jdcanas: Hmm.. I don't think this is a valid change.

There are two things at play here. One is the backend appType, defined here (the all caps version) https://github.com/DataBiosphere/leonardo/blob/develop/core/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/kubernetesModels.scala#L304

One is the frontend label, used for filtering in list calls, here: https://github.com/DataBiosphere/terra-ui/blob/dev/src/pages/workspaces/workspace/WorkspaceContainer.js#L182

It would be nice if these were the same, but it seems different people worked on each 😵‍💫 

In other words, this value needs to remain the lowercase 'galaxy' (the `galaxy.label`) for the UI to properly function, as the list calls rely on existing disks with this label for backend filtering of disks returned.

src/components/AnalysisModal.js
cahrens: OK, I updated a couple of things to be consistent with the code at https://github.com/DataBiosphere/terra-ui/blob/c0ceb2e8ff4e7700e43f981126d3feb58ce61f97/src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js#L63.

src/components/RuntimeManager.js
```diff
@@ -196,8 +197,8 @@ export default class RuntimeManager extends PureComponent {
     const createdDate = new Date(runtime.createdDate)
     const dateNotified = getDynamic(sessionStorage, `notifiedOutdatedRuntime${runtime.id}`) || {}
     const rStudioLaunchLink = Nav.getLink(appLauncherTabName, { namespace, name, application: 'RStudio' })
-    const app = getCurrentApp(apps)
-    const prevApp = getCurrentApp(prevProps.apps)
+    const galaxyApp = getCurrentAppForType(tools.galaxy.appType)(apps)
+    const prevGalaxyApp = getCurrentAppForType(tools.galaxy.appType)(prevProps.apps)
```
jdcanas, line 201: perhaps a TODO here (maybe group these TODOS with some `generalize apps` tag or the like) to make this a list of apps. The intent being it would be nice to get the[ launch notification ](https://github.com/DataBiosphere/terra-ui/pull/2682/files#diff-aabe65a220cb1e02fd8b0e1a38c532eb92df64bcc9b0e244801141b6330d1ca5L234-L235 )relatively for free when adding new apps!

Same probably applies to [error notification](https://github.com/DataBiosphere/terra-ui/pull/2682/files#diff-aabe65a220cb1e02fd8b0e1a38c532eb92df64bcc9b0e244801141b6330d1ca5R245-R246)

In general it seems quite risky to diverge RuntimeManager to be specific about certain kinds of apps, when its more of a high-level controller-type component. We are planning a large refactor of it soon, for context: https://broadworkbench.atlassian.net/browse/IA-2848

src/components/ContextBar.js
kyuksel: Appreciate the accessibility edification 😊 
How about the `aria-label`s in `Clickable`s above and below?

src/components/AnalysisModal.js
```diff
@@ -105,38 +118,32 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
       persistentDisks,
       location,
       onDismiss: () => {
-        resetView()
+        setViewMode(undefined)
```
cahrens: I made this change to be consistent with https://github.com/DataBiosphere/terra-ui/blob/c0ceb2e8ff4e7700e43f981126d3feb58ce61f97/src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js#L63, but I don't know if resetView is more appropriate here (`setViewMode(undefined)`) is what we are also calling below in the similar code for GalaxyModal.

src/components/CromwellModal.js
```diff
@@ -0,0 +1,88 @@
...
+
+
+const defaultDataDiskSize = 500 // GB
+const defaultKubernetesRuntimeConfig = { machineType: 'n1-highmem-8', numNodes: 1, autoscalingEnabled: false }
+
+const titleId = 'cromwell-modal-title'
+
+export const CromwellModalBase = Utils.withDisplayName('CromwellModal')(
+  ({
+    onDismiss, onSuccess, apps, appDataDisks, workspace, workspace: { workspace: { namespace, bucketName, name: workspaceName, googleProject } },
```
kyuksel: Looks like an unused input?

```suggestion
    onDismiss, onSuccess, apps, workspace, workspace: { workspace: { namespace, bucketName, name: workspaceName, googleProject } },
```

src/components/CromwellModal.js
kyuksel: It'd be good to add a TODO in this file with a link to the ticket for refactoring.

src/libs/utils.js
kyuksel: `saturn` used to be the name of the UI team. I think we could remove it in favour of something like

```suggestion
export const generateAppName = () => `terra-app-${uuid()}`
```

src/libs/ajax.js
cahrens: Interestingly, this has been working with the case difference. It appears that the backend call is case-insensitive because I verified that the following 2 things return the same thing:

```
Ajax(signal).Disks.list({ googleProject, creator: getUser().email, saturnApplication: "galaxy" })

Ajax(signal).Disks.list({ googleProject, creator: getUser().email, saturnApplication: "GALAXY" })
```
I will make the front-end code consistent though, using the appType since that is what we have available in ajax.js.

kyuksel (commented):
I still have more files to go through but I'll make a drop for what I have until then.

src/components/RuntimeManager.js
cahrens: Can we add this the existing IA-2848 ticket? I agree it would be nice to get this support for Cromwell as well.

My changes didn't make the code any more app-specific than it was before. It's just that previously `app` was always a Galaxy app, and the code that used `app` showed Galaxy-specific messages and functionality. I renamed the variables so that it is more obvious what is going on.

src/components/CromwellModal.js
cahrens: It is unused at this point, but I was keeping the input args for this object the same as the input args for the others (for example, GalaxyModalBase). I do imagine we will use appDataDisks at some point.

Would you prefer I delete it anyway?

src/components/RuntimeManager.js
jdcanas: Yeah, I don't think it should go in the scope of this PR.

Maybe just a TODO for now. We would likely get to it when we add in other apps in future quarters, not with the analysis tab release which is what the epic that ticket is intended for.

src/libs/ajax.js
jdcanas: that's good to know!

src/components/CromwellModal.js
cahrens: Sure, I can create a ticket for the creation of AppModal to share between GalaxyModal and CromwellModal. Though I'm still somewhat unsure about the best way to generalize all of this… I would really prefer that each app/tool isolate its code in one place, and the infrastructure code gets the things it needs through an established API. But that doesn't necessarily mean that Galaxy and Modal won't expose a modal class as part of that API (and their modal classes could use inheritance).

src/components/ContextBar.js
cahrens: It is necessary for "Workspace Menu" because there is no tooltip that is being used as a label.

"Terminal button" is problematic. Because a tooltip IS being used there as a label, the aria-label does not get read out, and  instead the tooltip is read out. So what you get is a bit confusing.

src/components/CromwellModal.js
kyuksel: @cahrens I think your thoughts would be a good discussion topic for the meeting @jdcanas scheduled 😊

src/components/CromwellModal.js
cahrens: @jdcanas I need more information about how `withErrorReportingInModal` works. When I tried switching to it, the modal would remain open when I hit an error, but I would not see the error itself. And when I closed the modal, the popup that used to show the error message was not present.

What do I need to do to fully implement `withErrorReportingInModal`?

src/components/AnalysisModal.js
cahrens: I created https://broadworkbench.atlassian.net/browse/IA-3071

src/components/ContextBar.js
kyuksel: > It is necessary for "Workspace Menu" because there is no tooltip that is being used as a label.

Is there a reason why we couldn't add a `useTooltipAsLabel: true` and update `tooltip: 'Menu'` and `tooltip: 'Workspace Menu'` and get rid of `aria-label`?

> "Terminal button" is problematic. Because a tooltip IS being used there as a label, the aria-label does not get read out, and instead the tooltip is read out. So what you get is a bit confusing.

Agreed. Does it make sense to remove `useTooltipAsLabel: true` for that one?

src/components/RuntimeManager.js
cahrens: I created https://broadworkbench.atlassian.net/browse/BW-929.

cahrens:
> We have a [leonardo-cron-jobs](https://github.com/broadinstitute/leonardo-cron-jobs) repo for processes that intermittently run to detect/fix anomalies, and cleaning resources up. I didn't look into the code exhaustively but we made Galaxy app-specific assumptions there too (like [this](https://github.com/broadinstitute/leonardo-cron-jobs/blob/4d844ca56a96471800c7ff572796fbad7735c246/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedDiskChecker.scala#L31)) that may need updating and/or we may need to add additional cron jobs for Cromwell. Resource clean-up part of things may actually be a high priority since it'll cost money if they're left behind.

Created https://broadworkbench.atlassian.net/browse/IA-3072

src/components/CromwellModal.js
cahrens: I put a TODO in the code instead of a JIRA ticket.

src/components/AnalysisModal.js
cahrens: Totally don't need them… it just happened because of refactoring (originally there were 2 separate methods). Pushing up change.

src/components/ContextBar.js
cahrens: Sure, I made those changes. "Workspace Menu" is more useful than "Menu" for a tooltip anyway!

kyuksel: Much appreciated! To be clear, I wasn't asking you to make the changes since your PR didn't introduce any of that. I was just trying to make sure I understand what was desirable as far as a11y 😊

kyuksel:
### Summary
**TL;DR** With the latest code, there seems to be an issue with creating a Cromwell when there is a Galaxy app in `DELETING` status.

- Not sure if Galaxy's status being `DELETING` (as opposed to any non-`DELETED` status) particularly matters. 
- Also not sure whether this issue existed with the original version of the PR or not.

### Details
I deleted a Cromwell so my `#clusters` page looked like

image

and then on the Cloud Env Modal, I noticed I wasn't able to create a new Cromwell:

image

The issue went away once I deleted Galaxy.

kyuksel:
**Note:** I'm reporting this separately from [this](https://github.com/DataBiosphere/terra-ui/pull/2682#issuecomment-974682242) but the root cause for both issues may be the same; i.e. the code isn't distinguishing between Galaxy and Cromwell instances, and trying to assert cases like one app per workspace. 

When I have a `PROVISIONING` or `RUNNING` Galaxy:

image

I get the message below when hovering over Cromwell

image

If I go ahead and attempt to create a Cromwell, both Cromwell seems to be created successfully and Galaxy doesn't seem to be affected by it, which is good news as far as the backend.

In that last case, there's an issue on `#clusters` page with the warning displays as well:

image

FWIW, the disk warning occurs for not only the Cromwell disk but also the GCE disk.

kyuksel:
Now that we've removed the pause buttons for Cromwell, the right-most column on `#clusters` looks weird:

image

I understand that the UX details aren't determined but if it's a quick change to left-align Cromwell's delete button or align it with the delete buttons for other apps, it may be worth improving for the interim too.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
kyuksel: I'm guessing it's for `new tab`ing but not sure. Do you know, @jdcanas?

src/components/CromwellModal.js
kyuksel: Okay. What I was after mainly is, we not do any significant further development in this file without refactoring it first.

src/components/CromwellModal.js
kyuksel: I hear you but for me, it has usually turned out to be the best to not put in/leave things in place with future expectations…probably because the future is difficult to predict 😊

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -218,8 +221,12 @@ const useAppPolling = (googleProject, workspaceName) => {
         await Ajax(signal).Apps.list(googleProject, { creator: getUser().email, saturnWorkspaceName: workspaceName }) :
         []
       setApps(newApps)
-      const app = getCurrentApp(newApps)
-      reschedule((app && _.includes(app.status, ['PROVISIONING', 'PREDELETING'])) ? 10000 : 120000)
+      _.forOwn(tool => {
```
kyuksel, line 224: TIL

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -85,28 +87,48 @@ export const CloudEnvironmentModal = ({
       onDismiss()
     }
   })
+  const renderGalaxyModal = () => renderAppModal(GalaxyModalBase, NEW_GALAXY_MODE)
+  const renderCromwellModal = () => renderAppModal(CromwellModalBase, NEW_CROMWELL_MODE)
```
kyuksel: Same comment as https://github.com/DataBiosphere/terra-ui/pull/2682#discussion_r753457086

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -85,28 +87,48 @@ export const CloudEnvironmentModal = ({
...
+    color: colors.accent()
   }
 
   const currentRuntime = getCurrentRuntime(runtimes)
   const currentRuntimeStatus = getConvertedRuntimeStatus(currentRuntime)
   const currentRuntimeTool = currentRuntime?.labels?.tool
 
-  const currentApp = getCurrentApp(apps)
+  const currentApp = toolLabel => getCurrentAppForType(getAppType(toolLabel))(apps)
+  const showPause = toolLabel => !_.find(tool => tool.label === toolLabel)(tools).isPauseUnsupported
```
kyuksel: I find it clearer when functions like `showPause` is prefixed with a modal verb like `should` because they are meant to be used in logical operations. For instance, without looking at its implementation, it wouldn't be obvious to me whether we use `showApp` for control or actually show something (as in the case of `startApp` below).

kyuksel (commented):
Made at least one pass through all of the diff. Comments inline.

Reported issues from testing separately.

src/components/CromwellModal.js
cahrens: I think it depends on the form the work takes. It may not make sense to create an AppModalBase class if Galaxy and Cromwell modals end up being fairly different.

src/components/CromwellModal.js
cahrens: Sure, I will remove it. 

Can someone explain to me the `workspace` variables? To me, it appears that workspace is defined as an input twice, but trying to remove the first one (which seems less specific in form) causes a compilation error.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
cahrens: OK, switching to isPauseSupported for this variable.

(I only made the method on the tools object a isPauseUnsupported because I needed the default value to be False).

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
cahrens: Yep, forgot about the duplication here. Will do.

cahrens:
> I understand that the UX details aren't determined but if it's a quick change to left-align Cromwell's delete button or align it with the delete buttons for other apps, it may be worth improving for the interim too.

My gut feeling is that we will end up exposing Pause, but it's easy enough to move the spacing from the Delete button to the Pause button.

cahrens:
> I get the message below when hovering over Cromwell <(while Galaxy was Provisioning>)

The message "Create new Environment (may overwrite existing)" is what we always show when an instance of that type does not already exist. From your comment, I think you expect "may overwrite existing" to mean we think there is already an instance. When there is an instance, what we show is "Edit existing Environment".

cahrens:
> **TL;DR** With the latest code, there seems to be an issue with creating a Cromwell when there is a Galaxy app in `DELETING` status.

Excellent catch! Yes, there was a Galaxy-specific conditional we were hitting. I've modified it to check if the item being rendered is Galaxy.

(I do wonder what special logic we will need for handling Cromwell's persistent disk, but it's out of scope for this change)

cahrens:
> 
> FWIW, the disk warning occurs for not only the Cromwell disk but also the GCE disk.

The disk warning code needs some work. Right now it is pulling Galaxy disks out of the list of disks, and it shows the warning if there are multiple Galaxy disks or multiple tool disks (disks left after removing Galaxy's) in the workspace. But the Cromwell disks are **not** being pulled out of the total list of disks, so if you have a GCE and a Cromwell you get the warning.

Would you be OK if I create a follow-up ticket to fix this problem? I'd like to see if I can do it without adding a whole bunch of Cromwell variables/logic that duplicates the Galaxy logic. And since this MR is already pretty large, I would like to minimize further churn on it.

A couple of other thoughts-- we should share the code between Environment.js and WorkspaceContainer.js, which is doing similar things. And I think that @salonishah11's Leo MR will add a field to the disk that lets us know the app that has provisioned it, so that may simplify the code.

Update from Monday meeting: IA team is OK with doing this as a separate ticket. I have created https://broadworkbench.atlassian.net/browse/BW-931.

src/components/AnalysisModal.js
```diff
@@ -75,10 +86,10 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
         }],
         [Utils.DEFAULT, () => Utils.cond(
           [currentTool === tools.RStudio.label || currentTool === tools.Jupyter.label, () => setViewMode(analysisMode)],
-          [currentTool === tools.galaxy.label && !currentApp, () => setViewMode(environmentMode)],
-          [currentTool === tools.galaxy.label && currentApp, () => {
+          [isToolAnApp(currentTool) && !app, () => setViewMode(environmentMode)],
+          [isToolAnApp(currentTool) && !!app, () => {
```
cahrens, line 90: @kyuksel FYI since you asked about accessibility. These create tool controls have a couple of serious issues. Right now they are just divs with images inside of them (and the images have no alt text). So the controls are not focusable (meaning that they can't be triggered via the keyboard), and there is no way for someone using a screen reader to know what any of them would actually do. I have created https://broadworkbench.atlassian.net/browse/IA-3043.

src/components/ContextBar.js
cahrens: @kyuksel Sure, this was easy to change. But actually the ContextBar has far larger a11y issues, so I created https://broadworkbench.atlassian.net/browse/IA-3073 for it.

src/components/CromwellModal.js
cahrens: FYI, I now understand the workspace variables.

src/components/CromwellModal.js
cahrens: @jdcanas Passing in onDismiss did fix the issue. I have pushed up the change.

src/components/AnalysisModal.js
cahrens: As we discussed, I changed to call resetView here and the other similar locations in this file.

jdcanas (approved):
LGTM after our meeting

kyuksel:
👍🏼 
Thanks for diligently addressing all of the PR comments!

kyuksel (approved):

PR #2684, opened 2021-11-09
add editorconfig

zarsky-broad:
This should be automatically parsed by intellij and other editors, but it's got all of the intellij formatting rules in it, so people will get the right code style automatically (I hope)


s-rubenstein (approved):
LGTM

breilly2 (approved):
👍   I didn't try this out, but I trust that the rules enforced by `.editorconfig` are the same as those in `js-style.xml`.

PR #2687, opened 2021-11-10
Quick fix for issue in test

petesantos:
Quick fix for issue that came up in staging. The component was throwing an error on the workflows page where the ref was not present. Some additional investigation will be needed, but this unblocks teams immediately.

A follow up ticket has been created here:
https://broadworkbench.atlassian.net/browse/DC-67


lu-c (approved):

lu-c (commented):
If you want to remove the error message, you can also delete line 296, to remove the ref from the object

kyuksel (approved):

PR #2688, opened 2021-11-12
[QA-1652] Reduce run-notebook test flakiness

aweng98:
Reduce `run-notebook` test flaky failures.
https://broadworkbench.atlassian.net/browse/QA-1652

### DIFF
--- a/integration-tests/jest-puppeteer.config.js
+++ b/integration-tests/jest-puppeteer.config.js
@@ -13,6 +13,7 @@ const printWarning = message => {
 }
 
 module.exports = {
+  exitOnPageError: false,
   browserContext: 'incognito',
   launch: {
     devtools: process.env.DEVTOOLS === 'true',
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -53,7 +53,6 @@ const testRunNotebookFn = _.flow(
   await findText(frame, '123456789099886419754209')
   // Save notebook to avoid "unsaved changes" modal when test tear-down tries to close the window
   await click(frame, clickable({ text: 'Save and Checkpoint' }))
-  await delay(1000)
 })
 const testRunNotebook = {
   name: 'run-notebook',
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -41,18 +41,19 @@ const testRunNotebookFn = _.flow(
   // Therefore, we start with a 1000ms delay, then make sure there are no spinners just in case the
   // AJAX calls are unexpectedly slow.
   await delay(1000)
-  waitForNoSpinners(page)
+  await waitForNoSpinners(page)
   await click(page, clickable({ text: 'Create' }))
   await findElement(page, clickable({ textContains: 'Creating' }))
   await findElement(page, clickable({ textContains: 'Running' }), { timeout: 10 * 60 * 1000 })
 
   const frame = await findIframe(page)
   await findElement(frame, '//*[@title="Kernel Idle"]')
-  await fillIn(frame, '//textarea', 'print(123456789099876543210990+9876543219)')
+  await fillIn(frame, '//textarea', '100 + 1')
   await click(frame, clickable({ text: 'Run' }))
-  await findText(frame, '123456789099886419754209')
+  await findText(frame, '101')
   // Save notebook to avoid "unsaved changes" modal when test tear-down tries to close the window
   await click(frame, clickable({ text: 'Save and Checkpoint' }))
+  await delay(1000)
 })
 const testRunNotebook = {
   name: 'run-notebook',
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -48,9 +48,9 @@ const testRunNotebookFn = _.flow(
 
   const frame = await findIframe(page)
   await findElement(frame, '//*[@title="Kernel Idle"]')
-  await fillIn(frame, '//textarea', '100 + 1')
+  await fillIn(frame, '//textarea', 'print(123456789099876543210990+9876543219)')
   await click(frame, clickable({ text: 'Run' }))
-  await findText(frame, '101')
+  await findText(frame, '123456789099886419754209')
   // Save notebook to avoid "unsaved changes" modal when test tear-down tries to close the window
   await click(frame, clickable({ text: 'Save and Checkpoint' }))
   await delay(1000)
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -40,7 +40,7 @@ const clickable = ({ text, textContains }) => {
   }
 }
 
-const click = async (page, xpath, options) => {
+const click = async (page, xpath, options = { visible: true }) => {
   return (await page.waitForXPath(xpath, options)).click()
 }
 
@@ -85,7 +85,7 @@ const select = async (page, labelContains, text) => {
   return click(page, `//div[starts-with(@id, "react-select-") and contains(normalize-space(.),"${text}")]`)
 }
 
-const waitForNoSpinners = page => {
+const waitForNoSpinners = async (page) => {
   return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
 }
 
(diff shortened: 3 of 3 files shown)


integration-tests/jest-puppeteer.config.js
```diff
@@ -13,6 +13,7 @@ const printWarning = message => {
 }
 
 module.exports = {
+  exitOnPageError: false,
```
aweng98, line 16: `exitOnPageError`: Do not exits page on any JS page error message thrown by UI code. Defaults is true.

integration-tests/tests/run-notebook.js
```diff
@@ -41,18 +41,19 @@ const testRunNotebookFn = _.flow(
...
   await findElement(page, clickable({ textContains: 'Creating' }))
   await findElement(page, clickable({ textContains: 'Running' }), { timeout: 10 * 60 * 1000 })
 
   const frame = await findIframe(page)
   await findElement(frame, '//*[@title="Kernel Idle"]')
-  await fillIn(frame, '//textarea', 'print(123456789099876543210990+9876543219)')
+  await fillIn(frame, '//textarea', '100 + 1')
   await click(frame, clickable({ text: 'Run' }))
-  await findText(frame, '123456789099886419754209')
+  await findText(frame, '101')
```
aweng98: The original code looks for `123456789099886419754209` string in code output field to verify code executed successfully. But, the same string can be found in code input field as well. So we don't know what the `findText()` finds. 
Make code output uniquely different so that test finds and verify the right string.

integration-tests/utils/integration-utils.js
```diff
@@ -85,7 +85,7 @@ const select = async (page, labelContains, text) => {
   return click(page, `//div[starts-with(@id, "react-select-") and contains(normalize-space(.),"${text}")]`)
 }
 
-const waitForNoSpinners = page => {
+const waitForNoSpinners = async (page) => {
```
aweng98: Code bug: missing `async` modifier.

integration-tests/utils/integration-utils.js
```diff
@@ -40,7 +40,7 @@ const clickable = ({ text, textContains }) => {
   }
 }
 
-const click = async (page, xpath, options) => {
+const click = async (page, xpath, options = { visible: true }) => {
```
aweng98, line 43: Clicks on "visible" web element.

integration-tests/tests/run-notebook.js
breilly2:
```
123456789099876543210990+9876543219
123456789099886419754209
             ^
   difference starts at this digit
```

aweng98: aha, I missed it. I will undo this change.

integration-tests/utils/integration-utils.js
zarsky-broad: So, because the body of the function doesn't contain the keyword `await`, this shouldn't be marked `async`. The function just returns the promise from the `waitForXPath` directly, which happens to be async, but this function doesn't need to care.

zarsky-broad:
This needs a JIRA issue number before we can merge it, please and thank you

integration-tests/jest-puppeteer.config.js
cahrens: Why do we want to make this false? I assume it is true by default to maximize the chance of catching errors.

integration-tests/tests/run-notebook.js
```diff
@@ -53,6 +53,7 @@ const testRunNotebookFn = _.flow(
   await findText(frame, '123456789099886419754209')
   // Save notebook to avoid "unsaved changes" modal when test tear-down tries to close the window
   await click(frame, clickable({ text: 'Save and Checkpoint' }))
+  await delay(1000)
```
cahrens: What is this delay for?

integration-tests/utils/integration-utils.js
aweng98: Yes, I knew. I think the modifier makes function clear and syntactically correct. I've using typescript, without the modifier, it looked "odd". I'll remove the modifier.

integration-tests/tests/run-notebook.js
aweng98: A short delay when clicking a button/link in Jupyter notebook UI will make tests less flaky. It's a lesson learned in All-of-Us Puppeteer automation tests.
Half second is sufficient if one second "feels" too long.

integration-tests/jest-puppeteer.config.js
aweng98: `exitOnPageError: true` makes tests flaky: It's not a good practice to allow sudden/random page exits during running of tests. 

It would be good to log page errors (in future) so it would be easier to know what happened while letting tests finish from beginning to end.

integration-tests/utils/integration-utils.js
aweng98: Calling `waitForNoSpinners(page)` function without `await` could have been prevented at compile time if `async` is present in function declaration.

integration-tests/tests/run-notebook.js
cahrens: Thanks @aweng98 Did you learn anything about why it would make the test less flaky? Ideally one would want to wait for something specific to complete instead of putting in a sleep with an arbitrary time.

If this delay stays in, can you please add a comment in the code as to why the sleep is necessary (and how the timeframe of 1000 ms was chosen)?

kyuksel:
@aweng98 Thanks for the PR. FYI that I'm rerunning tests on this PR to make sure we don't encounter a failure signature we haven't seen before.

kyuksel:
Also curious if folks have any thoughts on retrying failures via something like [jest.retryTimes()](https://stackoverflow.com/a/60129634/2851999). Ideally we fix the underlying reasons for failures but until then we end up rerunning failed tests manually multiple times so it might make sense to automate that?

zarsky-broad:
> Also curious if folks have any thoughts on retrying failures via something like [jest.retryTimes()](https://stackoverflow.com/a/60129634/2851999). Ideally we fix the underlying reasons for failures but until then we end up rerunning failed tests manually multiple times so it might make sense to automate that?

We've brought that up before, historically there's huge pushback that if a test doesn't give meaningful signal it should either be fixed or disabled.

integration-tests/jest-puppeteer.config.js
cahrens: With the true option, is the error that caused the page exit present in the test artifacts?

aweng98: no, I don't think page errors are saved anywhere. Ideally, page console errors and networks requests and responses should be logged to aid with the failure troubleshooting.

integration-tests/tests/run-notebook.js
aweng98: We learned that Puppeteer clicks are fast. By slowing down the speed just a little (like with `slowmo: 10`), tests are less likely to fail randomly when actions and pages are "synchronized".
Ideally, we should wait for notebook saved message after click `save` button to verify save worked. But I didn't want to change more code.
A second thought, click is at the end of the test, so a sleep is not really necessary because there is no verification that save worked. So I will remove the sleep here.

aweng98 (commented):
This PR should reduces flakiness of `run-notebook` test but doesn't fix all flakiness.

aweng98:
> Also curious if folks have any thoughts on retrying failures via something like [jest.retryTimes()](https://stackoverflow.com/a/60129634/2851999). Ideally we fix the underlying reasons for failures but until then we end up rerunning failed tests manually multiple times so it might make sense to automate that?

Re-run failed tests is an easy way out but it hides flaky tests. We ought to eliminate test flakiness.

kyuksel (approved):
The code looks good to me. Pending tests passing. The last notebook test failure snapshot (below) looked unusual. Rerunning it...

image

integration-tests/jest-puppeteer.config.js
breilly2: > `exitOnPageError: true` makes tests flaky: It's not a good practice to allow sudden/random page exits during running of tests.

I'm not sure I follow this logic. I haven't personally experimented with this, but I assume this means that, if an exception bubbles up to the top, that would cause the test to exit. However, we generally catch these kinds of errors and display them as the error alerts that appear in the upper right. This generally results in the test failing. If we happen to _not_ catch an error and display a notification, then the page "crashes" and the test will fail. I can't think of a case where exiting after an exception would cause a test to fail while _not_ exiting after an exception might lead to the test passing.

PR #2689, opened 2021-11-12
Add redirect for RAS [POL-106]

zarsky-broad:
This isn't used yet. However!

**N.B. This changes routing behavior!**
If you go to **any** path that doesn't start with a hash, instead of the current behavior of unhelpfully dumping you to the landing page (or any hash path, if there also is one), it'll now go to the hash of that path, preserving the search if present.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,7 +2,7 @@ import { loadedConfigStore } from 'src/configStore'
 
 
 const loadApp = async () => {
-  const res = await fetch('/config.json')
+  const res = await fetch(`${process.env.PUBLIC_URL}/config.json`)
   loadedConfigStore.current = await res.json()
 
   import('src/appLoader')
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -2,7 +2,7 @@ import { loadedConfigStore } from 'src/configStore'
 
 
 const loadApp = async () => {
-  const res = await fetch('config.json')
+  const res = await fetch('/config.json')
```
zarsky-broad: Adding a slash so that if you go to a path that's more than one level deep, the app can still load and turn it into a usable one.

petesantos: Thanks! I was just looking at this issue in my review.

petesantos (approved):
Tested manually, tried out different combinations of manual URL entry and navigation.
LGTM

kyuksel (approved):

PR #2690, opened 2021-11-12
[QA-1645] [DC-6] Fixed the "forward ref" warning complaint

lu-c:
> index.js:1 Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Noticed the warning about how the forward-ref would be required would trigger on the workflow page, but not the catalog page. Digging in to the differences, discovered the workflow is attaching the autocomplete to a textarea, where the catalog is attaching autocomplete to an input.

I'm hesitant about this change, because it could affect a lot of the other unsuspecting code in the codebase, but Ive confirmed while running this branch that the error no longer shows up and the autocomplete will make suggestions for both the [catalog view](localhost:3000/#library/browser) and the [workflow page](http://localhost:3000/#workspaces/general-dev-billing-account/christine-test/workflows/gatk/haplotypecaller-gvcf-gatk4-configured). (Links may not work, but give an idea of where in the app I was playing around)


petesantos (approved):
I tested out the sites where the `TextArea` is used - I didn't see any issues

src/components/input.js
```diff
@@ -319,15 +319,15 @@ export const AutocompleteTextInput = withAutocomplete(TextInput)
...
+export const TextArea = Utils.forwardRefWithName('TextArea', ({ onChange, autosize = false, nativeOnChange = false, ...props }, ref) => {
   Utils.useLabelAssert('TextArea', { ...props, allowId: true })
 
   return h(autosize ? TextAreaAutosize : 'textarea', _.merge({
     className: 'focus-style',
     style: styles.textarea,
     onChange: onChange ? (e => onChange(nativeOnChange ? e : e.target.value)) : undefined
-  }, props))
-}
+  }, props), ref)
```
zarsky-broad: Should the ref be in the merged object? I don't think it's doing anything here.

lu-c: This allows us to attach the ref in line 296, otherwise, React complains about the lack of fowardRef

petesantos: I believe Isaac is correct on this. The ref needs to be added to the props otherwise it will not be set correctly:

```javascript  
return h(autosize ? TextAreaAutosize : 'textarea', {
    ..._.merge({
      className: 'focus-style',
      style: styles.textarea,
      onChange: onChange ? (e => onChange(nativeOnChange ? e : e.target.value)) : undefined
    }, props),
    ref
  })
```

You can take a look at [line 84](https://github.com/DataBiosphere/terra-ui/blob/0a2b86be12cae5e22783fae8714dc2925ea3ebc6/src/components/input.js#L84) for an example. You can also test this code out by adding a logging statement around line 243 to display the `current` value and visit the workflow page. Coded as-is, the ref does not get populated.

src/components/input.js
```diff
@@ -319,15 +319,18 @@ export const AutocompleteTextInput = withAutocomplete(TextInput)
...
-  }, props))
-}
+  return h(autosize ? TextAreaAutosize : 'textarea', {
+    ..._.merge({
+      className: 'focus-style',
+      style: styles.textarea,
+      onChange: onChange ? (e => onChange(nativeOnChange ? e : e.target.value)) : undefined
+    }, props),
+    ref
+  })
```
zarsky-broad: consider:
```suggestion
  return h(autosize ? TextAreaAutosize : 'textarea', _.merge({
    ref,
    className: 'focus-style',
    style: styles.textarea,
    onChange: onChange ? (e => onChange(nativeOnChange ? e : e.target.value)) : undefined
  }, props)
```

lu-c: Updated!

zarsky-broad (approved):

PR #2694, opened 2021-11-16
[DC-47] Added new filter catagory for release policy type

lu-c:
Screen Shot 2021-11-15 at 10 18 49 PM

To test, update `dataBrowser-utils.js.116` so you modify the ajax return from the server:
```
const metadata = await Ajax(signal).DataRepo.getMetadata()
metadata.result[0]['TerraDCAT_ap:hasDataUsePermission'] = 'TestingThis'
metadata.result[1]['TerraDCAT_ap:hasDataUsePermission'] = 'UnknownReleasePolicyListed'
metadata.result[2]['TerraDCAT_ap:hasDataUsePermission'] = 'TerraCore:PUB'
```


src/pages/library/dataBrowser-utils.js
```diff
@@ -38,10 +67,14 @@ const normalizeSnapshot = snapshot => {
     _.uniqBy(_.toLower)
   )(snapshot['prov:wasGeneratedBy'])
 
-  const dataReleasePolicy = _.flow(
-    _.replace('TerraCore:', ''),
-    _.startCase
-  )(snapshot['TerraDCAT_ap:hasDataUsePermission'])
+  const dataReleasePolicy = _.getOr(
+    {
+      ...snapshotReleasePolicies.releasepolicy_other,
```
petesantos, line 63: is `releasepolicy_other` used anywhere else? Could this just be:
```javascript
      tag: `${releasePolicyPrefix}OTHER`,
      label: 'Other',
      desc: _.flow(
        _.replace('TerraCore:', ''),
        _.startCase
      )(snapshot['TerraDCAT_ap:hasDataUsePermission'])
```

src/pages/library/DataBrowser.js
```diff
@@ -55,17 +56,26 @@ const extractCatalogFilters = dataCatalog => {
...
     }, _.keys(snapshotAccessTypes)))
+  }, {
+    name: 'Data use policy',
+    labels: releasePolicyArray,
+    labelDisplays: _.zipObject(releasePolicyArray, _.map(releasePolicy => {
+      return [div({ key: `releasePolicy-filter-${releasePolicy.tag.toLowerCase()}`, style: { display: 'flex', flexDirection: 'column' } }, [
+        releasePolicy.label,
+        releasePolicy.desc && div({ style: { fontSize: '0.625rem', lineHeight: '0.625rem' } }, [releasePolicy.desc])
+      ])]
+    }, snapshotReleasePolicies))
```
petesantos: nit: How do you feel about making this a `flow`? For me it is a bit easier to read, but it may just be me.
```suggestion
    labelDisplays: _.flow(
      _.map(({ label, desc, tag }) => {
        return [div({ key: `releasePolicy-filter-${tag.toLowerCase()}`, style: { display: 'flex', flexDirection: 'column' } }, [
          label,
          desc && div({ style: { fontSize: '0.625rem', lineHeight: '0.625rem' } }, [desc])
        ])]
      }),
      _.zipObject(releasePolicyArray)
    )(snapshotReleasePolicies)
```

src/pages/library/DataBrowser.js
```diff
@@ -55,17 +56,26 @@ const extractCatalogFilters = dataCatalog => {
         div([snapshotAccessTypes[accessKey]])
       ])]
     }, _.keys(snapshotAccessTypes)))
+  }, {
+    name: 'Data use policy',
+    labels: releasePolicyArray,
+    labelDisplays: _.zipObject(releasePolicyArray, _.map(releasePolicy => {
```
petesantos: If you don't want it to be a flow, could you destructure the props used in `releasePolicy`?
```suggestion
    labelDisplays: _.zipObject(releasePolicyArray, _.map(({label, tag, desc})=> {
```

src/pages/library/DataBrowser.js
```diff
@@ -42,9 +42,10 @@ const getUnique = (prop, data) => _.flow(
 // Description of the structure of the sidebar. Case is preserved when rendering but all matching is case-insensitive.
 const extractCatalogFilters = dataCatalog => {
   const accessArray = _.values(snapshotAccessTypes)
+  const releasePolicyArray = _.map('tag', snapshotReleasePolicies)
```
petesantos: When I change `tag` to label, I end up seeing more restrictions in the filter column.  I am not sure of the purpose of the tag property prefix.

src/pages/library/DataBrowser.js
lu-c: Will do

src/pages/library/dataBrowser-utils.js
lu-c: I like the current solution of `...snapshotReleasePolicies` so we dont have to make sure these tags and labels are the same if they ever change, keep it to the same variable instead of hardcoding it twice. It's used in the side filters and it's used in the display on the details page (only reason why the description really even comes into play)

src/pages/library/DataBrowser.js
lu-c: You do, but it's inaccurate - if you click on the restruction (like "CC"), and then select the one dataset that matches, the json returned for that dataset actually shows the data use policy to be "NRES (no restriction)" as well.

The reason why I had to create the tag field is because of how the filters work - Each dataset just has an array of filter attributes that it matches on. When we select a filter, it checks the string to see if that value exists, but it doesn't check if it matches for that same field.

So back to the example, the "CC" (release policy filter) will match on the dataset "Single Cell transcriptomic landscape of early human diabetic nephropathy" because the disease is "Enteroco**cc**us faecalis infection". It's only doing a simple string match.

I would recommend eventually rewriting how these filters work, but since we're building on code that already existed in the library, that felt like a bigger update

src/pages/library/DataBrowser.js
```diff
@@ -42,30 +42,46 @@ const getUnique = (prop, data) => _.flow(
...
+    labels: releasePolicyArray,
+    labelDisplays: _.flow(
+      _.map(({ label, tag, desc }) => {
+        return [div({ key: `releasePolicy-filter-${tag.toLowerCase()}`, style: { display: 'flex', flexDirection: 'column' } }, [
+          label,
+          desc && div({ style: { fontSize: '0.625rem', lineHeight: '0.625rem' } }, [desc])
+        ])]
+      }),
+      _.zipObject(releasePolicyArray)
+    )(snapshotReleasePolicies)
```
petesantos: What do you think of changing `labelDisplays` to a function? The reason being, it seems to cause some confusion around when  `labels` is used. With a function the dev can define the labels to use, and a function that will be used to render as it iterates over each label. This would also alleviate the need for lines 44-45

```javascript
    name: 'Access type',
    labels: _.keys(snapshotAccessTypes),
    labelRenderer: accessKey => {
      const lowerKey = _.toLower(accessKey)
      return [div({ key: `access-filter-${lowerKey}`, style: { display: 'flex' } }, [
        icon(snapshotAccessTypes[accessKey] === snapshotAccessTypes.OPEN ? 'unlock' : 'lock', {
          style: { color: styles.access[lowerKey], marginRight: 5 }
        }),
        div([snapshotAccessTypes[accessKey]])
      ])]
    }
  },
  {
    name: 'Data use policy',
    labels: getUnique('dataReleasePolicy.policy', dataCatalog),
    labelRenderer: rawPolicy => {
      const { label, desc } = snapshotReleasePolicies[rawPolicy] || snapshotReleasePolicies.releasepolicy_other
      return [div({ key: rawPolicy, style: { display: 'flex', flexDirection: 'column' } }, [
        label ? label : rawPolicy,
        desc && div({ style: { fontSize: '0.625rem', lineHeight: '0.625rem' } }, [desc])
      ])]
    }
```

petesantos (commented):
I have been diving in a bit more to better understand the filtering. I will follow up with a few more comments or we can pair

lu-c:
> I have been diving in a bit more to better understand the filtering. I will follow up with a few more comments or we can pair

Sure, whichever works better for you! Let me know if you want to pair today

src/pages/library/dataBrowser-utils.js
```diff
@@ -38,10 +67,14 @@ const normalizeSnapshot = snapshot => {
...
+      desc: _.flow(
+        _.replace('TerraCore:', ''),
+        _.startCase
+      )(snapshot['TerraDCAT_ap:hasDataUsePermission'])
+    }, snapshot['TerraDCAT_ap:hasDataUsePermission'], snapshotReleasePolicies)
```
petesantos: How about something like this. We can then reference the policy using the name with the `TerraCore` prefix

```suggestion
  const dataReleasePolicy = _.has(snapshot['TerraDCAT_ap:hasDataUsePermission'], snapshotReleasePolicies) ?
    { policy: snapshot['TerraDCAT_ap:hasDataUsePermission'], ...snapshotReleasePolicies[snapshot['TerraDCAT_ap:hasDataUsePermission']] } :
    { policy: 'releasepolicyOther', ...snapshotReleasePolicies.releasepolicyOther }
```

src/pages/library/dataBrowser-utils.js
```diff
@@ -65,7 +98,8 @@ const extractTags = snapshot => {
       snapshot.samples?.disease,
       snapshot.dataType,
       snapshot.dataModality,
-      _.map('dcat:mediaType', snapshot.files)
+      _.map('dcat:mediaType', snapshot.files),
+      snapshot.dataReleasePolicy.tag
```
petesantos: With the above change this would be:
```suggestion
      snapshot.dataReleasePolicy.policy
```

src/pages/library/dataBrowser-utils.js
```diff
@@ -15,6 +15,35 @@ export const snapshotAccessTypes = {
...
+  Filters will grab from an array of values without differentiating which filter it belongs to
+
+  For example:
+  if the user decides to filter by "data release policy = 'other'", it will filter the data
+  based off "{any filter field} = 'other'". Adding the prefix forces the filtering to recognize
+  which field is being filtered on.
+*/
+const releasePolicyPrefix = 'releasepolicytag_'
+export const snapshotReleasePolicies = {
+  'TerraCore:NoRestriction': { tag: `${releasePolicyPrefix}NRES`, label: 'NRES', desc: 'No restrictions' },
```
petesantos: If we go with the above suggestions I think we can remove the `tag` prop, but I may be missing something.

petesantos (approved):
LGTM - thanks!

pshapiro4broad (approved):

PR #2695, opened 2021-11-16
[DC-45] Fixed bug with importing data, where the title would cause error

lu-c:
…rors. Added UI messaging to indicate which datasets caused the error

Before, it would redirect the user to the selected workspace regardless of error

After
Screen Shot 2021-11-16 at 4 46 15 PM

Now shows statuses per dataset


src/libs/ajax.js
```diff
@@ -873,8 +873,8 @@ const Workspaces = signal => ({
         return res.json()
       },
 
-      importSnapshot: async (snapshotId, name) => {
-        const res = await fetchRawls(`${root}/snapshots`, _.mergeAll([authOpts(), jsonBody({ snapshotId, name }), { signal, method: 'POST' }]))
+      importSnapshot: async (snapshotId, name, description) => {
+        const res = await fetchRawls(`${root}/snapshots/v2`, _.mergeAll([authOpts(), jsonBody({ snapshotId, name, description }), { signal, method: 'POST' }]))
```
lu-c, line 882: The api states that the original endpoint is now deprecated, updated this to call the new endpoint as well

src/pages/ImportData.js
```diff
@@ -107,6 +108,7 @@ const ImportData = () => {
   )(async workspace => {
     const namespace = workspace.namespace
     const name = workspace.name
+    let success = true
```
dmohs: The logic using this boolean is spread pretty widely. I wonder if we can tighten it up.

src/pages/ImportData.js
```diff
@@ -119,17 +121,40 @@ const ImportData = () => {
...
+              const normalizedResponses = await Promise.all(_.map(async ({ status, reason }) => {
+                const reasonJson = reason ? await reason.json() : '{}'
+                const message = JSON.parse(reasonJson.message || '{}').message
+                return { status, message }
+              }, responses))
+              setSnapshotResponses(normalizedResponses)
+              notify('error', `There was a problem importing ${snapshots.length > 1 ? 'some of these snapshots' : 'this snapshot'}`, { timeout: 3000 })
+            }
+          }) :
+          await Ajax().Workspaces.workspace(namespace, name).importSnapshot(snapshotId, snapshotName).then(() => {
```
dmohs: This await appears to be a no-op.

dmohs (commented):
I think we can refactor this to make the flow a bit more clear. Do you want to pair on it?

src/pages/ImportData.js
lu-c: Pushed up a solution for it, just going to throw an error and let that catch it, please let me know what you think!

src/pages/ImportData.js
```diff
@@ -119,9 +121,33 @@ const ImportData = () => {
       }],
       ['snapshot', async () => {
         snapshots && snapshots.length > 0 ?
-          await Promise.allSettled(_.map(({ title, id }) => Ajax().Workspaces.workspace(namespace, name).importSnapshot(id, title), snapshots)) :
-          await Ajax().Workspaces.workspace(namespace, name).importSnapshot(snapshotId, snapshotName)
-        notify('success', 'Snapshot imported successfully.', { timeout: 3000 })
+          await Promise.allSettled(
+            _.map(({ title, id, description }) => {
+              // Replace all whitespace characters with _
+              // Then replace all non alphanumeric characters with nothing
```
dmohs, line 143: This is what the code does, but my preference would be a comment here describing the why instead of the what.

lu-c: Updated!

src/pages/ImportData.js
lu-c: I think it should continue to await to confirm whether or not this succeeds or fails before redirecting the user to the workspace

src/pages/ImportData.js
```diff
@@ -154,8 +182,31 @@ const ImportData = () => {
...
                   }
                 }, [
-                  icon('success-standard', { size: 18, style: { position: 'absolute', left: 0, color: colors.primary() } }),
-                  title
+                  !snapshotResponses ?
+                    div([
+                      icon('success-standard', { size: 18, style: { position: 'absolute', left: 0, color: colors.primary() } }),
+                      title
+                    ]) :
+                    Utils.switchCase(snapshotResponses[mapindex].status,
```
ehanna4: This might be overkill but I just wanted to offer it anyway - not a blocker just a thought on DRYing this up a little since each option in the switch uses the same Fragment with icon and message template. So essentially making a function to return the filled out Fragment and then using the status to populate an object with the needed different things and passing that to the Fragment builder function

(this is not working code more just pseudo code)

```
> // A fragment builder  using the repeated style parts and taking as args the things that differ
>
>                       getMessageFragment( ({iconType,color, messageArray} = ...responseOptions )) {
>                         return h(Fragment, [
>                             icon(iconType, { size: 18, style: { position: 'absolute', left: 0, color } }),
>                             title,
>                             div({ style: { color, fontWeight: 'normal', fontSize: '0.625rem', marginTop: 5 } }, messageArray)
>                           ])
>                       }
>
>// the object of pieces specific to each response type
>                       const responseOptions = snapshotResponses[mapindex].status === 'fulfilled' ?
>                         {iconType: 'success-standard', color: colors.primary(), messageArray:  [strong(['Success: ']), 'Snapshot successfully imported']} : 
>                         {iconType: 'rejected', color: colors.danger(), messageArray:  [strong(['Error: ']),snapshotResponses[mapindex].message]
>                 } 
>
> // calling and returning the built fragment
>                     return getMessageFragment(responseOptions)`
```

lu-c: Made this update, please let me know what you think

src/pages/ImportData.js
```diff
@@ -119,9 +136,35 @@ const ImportData = () => {
...
+              throw new Error(_.flow(
+                _.filter(({ status }) => status === 'rejected'),
+                _.map('message'),
+                _.join(', ')
+              )(normalizedResponses))
+            }
+          }) :
+          await Ajax().Workspaces.workspace(namespace, name).importSnapshot(snapshotId, snapshotName).then(() => {
+            notify('success', 'Snapshot imported successfully.', { timeout: 3000 })
+          })
```
petesantos: The long ternary here can be hard to follow, along with the `.then` chain. I think `await` cleans this up a bit. Also, since we are displaying the error message beneath each entry, I think we can keep the error notification short.

```suggestion
        if (snapshots?.length > 0) {
          const responses = await Promise.allSettled(
            _.map(({ title, id, description }) => {
              // Normalize the title:
              // Importing snapshot will throw an "enum" error if the title has any spaces or special characters
              const normalizedTitle = title.replace(/\s/g, '_').replace(/[^A-Za-z0-9\s-_]/g, '')
              return Ajax().Workspaces.workspace(namespace, name).importSnapshot(id, normalizedTitle, description)
            }, snapshots)
          )

          if (_.some({ status: 'rejected' }, responses)) {
            const normalizedResponses = await Promise.all(_.map(async ({ status, reason }) => {
              const reasonJson = reason ? await reason.json() : '{}'
              const message = JSON.parse(reasonJson.message || '{}').message
              return { status, message }
            }, responses))

            setSnapshotResponses(normalizedResponses)

            const numFailures = _.size(_.filter(({ status }) => status === 'rejected', normalizedResponses))
            throw new Error(`${numFailures} workspace${numFailures > 1 ? 's' : ''} failed to import. ` +
              'See details in the "Linking to Workspace" section')
          }
        } else {
          await Ajax().Workspaces.workspace(namespace, name).importSnapshot(snapshotId, snapshotName).then(() => {
            notify('success', 'Snapshot imported successfully.', { timeout: 3000 })
          })
        }
```

petesantos (commented):
I do have a general concern about the UX in this case. The user can get into a situation where they are repeatedly seeing error messages. If some of the data has imported and we have a few errors, the initially successful imports will fail on subsequent retries, never allowing the user to get redirected to their workspace.

I am ok getting this in to fix the immediate issue, but we should work with UX to get a clean flow for the user when importing and moving to the workspace.

src/pages/ImportData.js
lu-c: It's really hard for me to see the difference between this code and the one I have - Did you just mkae one change to replace the ternary with an if/else?

petesantos: Sorry - the diff is too big. 
1. Change to the ternary
2. Removed the `then` in favor of capturing the responses in the `await`
3. Changed the error notification message

lu-c: ok updated 👍

lu-c:
> I do have a general concern about the UX in this case. The user can get into a situation where they are repeatedly seeing error messages. If some of the data has imported and we have a few errors, the initially successful imports will fail on subsequent retries, never allowing the user to get redirected to their workspace.
> 
> I am ok getting this in to fix the immediate issue, but we should work with UX to get a clean flow for the user when importing and moving to the workspace.

I think a solution could be to add a "go to workspace anyway" button in the error details view, but that seems like it would be kind of hard to add.

src/pages/ImportData.js
```diff
@@ -118,10 +135,37 @@ const ImportData = () => {
...
       }],
       ['snapshot', async () => {
-        snapshots && snapshots.length > 0 ?
-          await Promise.allSettled(_.map(({ title, id }) => Ajax().Workspaces.workspace(namespace, name).importSnapshot(id, title), snapshots)) :
+        if (snapshots?.length > 0) {
+          const responses = await Promise.allSettled(
+            _.map(({ title, id, description }) => {
+              // Normalize the title:
+              // Importing snapshot will throw an "enum" error if the title has any spaces or special characters
+              // Replace all whitespace characters with _
```
petesantos, line 142: nit could you remove the comments on line 143 and 144. The comment Mohs had earlier was more about adding the "why" and removing the "how" of normalize code.

src/pages/ImportData.js
```diff
@@ -118,10 +135,37 @@ const ImportData = () => {
...
+              // Then replace all non alphanumeric characters with nothing
+              const normalizedTitle = title.replace(/\s/g, '_').replace(/[^A-Za-z0-9\s-_]/g, '')
```
petesantos: Could you convert this to a lodash flow/replace for consistency within the codebase?

petesantos (approved):
Thanks for working through this. I have a two minor comments, other than that, LGTM

src/pages/ImportData.js
lu-c: I honestly suck at reading regex, it's my preference to leave any comments that explain what the regex is targeting XD; Is that okay if we keep it?

src/pages/ImportData.js
```diff
@@ -58,6 +58,21 @@ const ChoiceButton = ({ iconName, title, detail, style, ...props }) => {
   ])
 }
 
+const ResponseFragment = ({ title, snapshotResponses, responseIndex }) => {
+  const { status, message } = snapshotResponses ? snapshotResponses[responseIndex] : {}
+  const [color, iconKey, children] = Utils.switchCase(status,
```
zarsky-broad, line 63: If you make this return children as a `Fragment` and render it with `[]` below, you can just do `children &&` for the condition

src/pages/ImportData.js
```diff
@@ -118,10 +135,40 @@ const ImportData = () => {
         notify('success', 'Data imported successfully.', { timeout: 3000 })
       }],
       ['snapshot', async () => {
-        snapshots && snapshots.length > 0 ?
-          await Promise.allSettled(_.map(({ title, id }) => Ajax().Workspaces.workspace(namespace, name).importSnapshot(id, title), snapshots)) :
+        if (snapshots?.length > 0) {
```
zarsky-broad: consider `!_.isEmpty(snapshots)`

src/pages/ImportData.js
```diff
@@ -118,10 +135,40 @@ const ImportData = () => {
...
+          const responses = await Promise.allSettled(
+            _.map(({ title, id, description }) => {
+              // Normalize the title:
+              // Importing snapshot will throw an "enum" error if the title has any spaces or special characters
+              // Replace all whitespace characters with _
+              // Then replace all non alphanumeric characters with nothing
+              const normalizedTitle = _.flow(
+                _.replace(/\s/g, '_'),
+                _.replace(/[^A-Za-z0-9\s-_]/g, '')
+              )(title)
```
zarsky-broad: What if:
```suggestion
              const normalizedTitle = _.replace(/[^\w-]/g, char => /\s/.test(char) ? '_' : '', title)
```

src/pages/ImportData.js
```diff
@@ -118,10 +135,40 @@ const ImportData = () => {
...
+              return Ajax().Workspaces.workspace(namespace, name).importSnapshot(id, normalizedTitle, description)
+            }, snapshots)
+          )
+
+          if (_.some(({ status }) => status === 'rejected', responses)) {
```
zarsky-broad: can you just:
```suggestion
          if (_.some({ status: 'rejected' }, responses)) {
```

src/pages/ImportData.js
```diff
@@ -118,10 +135,40 @@ const ImportData = () => {
...
+            const normalizedResponses = await Promise.all(_.map(async ({ status, reason }) => {
+              const reasonJson = reason ? await reason.json() : '{}'
+              const message = JSON.parse(reasonJson.message || '{}').message
```
zarsky-broad: Does this work?
```suggestion
              const reasonJson = await reason?.json()
              const { message } = JSON.parse(reasonJson?.message || '{}')
```

src/pages/ImportData.js
```diff
@@ -118,10 +135,40 @@ const ImportData = () => {
...
+              return { status, message }
+            }, responses))
+            setSnapshotResponses(normalizedResponses)
+
+            // Consolidate the multiple errors into a single error message
+            const numFailures = _.flow(
+              _.filter(({ status }) => status === 'rejected'),
```
zarsky-broad:
```suggestion
              _.filter({ status: 'rejected' }),
```

src/pages/ImportData.js
```diff
@@ -107,6 +123,7 @@ const ImportData = () => {
   )(async workspace => {
     const namespace = workspace.namespace
     const name = workspace.name
```
zarsky-broad: While you're in here, could you just
```suggestion
    const { namespace, name } = workspace
```

src/pages/ImportData.js
lu-c: I think this is a lot harder to read, going to vote to leave it as the flow! Any other opinions?

src/pages/ImportData.js
lu-c: yes it does!

src/pages/ImportData.js
lu-c: I pushed a diff update that would shorten the condition too. Is this okay?

zarsky-broad: I'd still like to avoid passing children to `div()` not in an array, so I'd like if it could be a Fragment coming out of the cond. That said, couldn't you just use `!!children` as the condition now? It's never present but empty.

src/pages/ImportData.js
zarsky-broad: If you could remove `\s` from the second regex, since it should be gone by then, that would work for me.

lu-c: done

src/pages/ImportData.js
petesantos: That works for me

tcjiang9 (approved):

PR #2696, opened 2021-11-17
[DC-46] Allow filter data to persist when navigating pages

petesantos:
This change allows the filters to be preserved when navigating between pages.
For instance: A user filters the catalog down to the data they are interested in and browses to the details page. When they go back, they will lose their filters. This change fixes this


lu-c (approved):
Tested this locally, confirm it works. Bonus is it works for the featured workspaces tab as well, which is also a useful feature!

dmohs (approved):
LGTM.

PR #2698, opened 2021-11-18
AS-910: fetch TOS markdown from SAM

sagehen03:


jordansafer (approved):

src/libs/ajax.js
```diff
@@ -235,6 +235,11 @@ const User = signal => ({
     }
   },
 
+  getTos: async () => {
+    const response = await fetchSam('tos/text', _.merge(authOpts(), { signal, method: 'GET' }))
+    return await response.text()
```
zarsky-broad: because it's already a promise, and this whole function will be awaited:
```suggestion
    return response.text()
```

src/libs/ajax.js
```diff
@@ -235,6 +235,11 @@ const User = signal => ({
     }
   },
 
+  getTos: async () => {
+    const response = await fetchSam('tos/text', _.merge(authOpts(), { signal, method: 'GET' }))
```
zarsky-broad: `GET` is the default
```suggestion
    const response = await fetchSam('tos/text', _.merge(authOpts(), { signal }))
```

zarsky-broad (commented):
Take a look at some places where we use `withBusyState` and `withErrorReporting`, you should be able to simplify the `useOnMount` here.

sagehen03:
> Take a look at some places where we use `withBusyState` and `withErrorReporting`, you should be able to simplify the `useOnMount` here.

Thanks! I took a stab at it.

src/pages/TermsOfService.js
```diff
@@ -1,307 +1,34 @@
...
-
-Terms as of February 12, 2020.
-`
-
 const TermsOfServicePage = () => {
   const [busy, setBusy] = useState()
   const { isSignedIn, acceptedTos } = authStore.get() // can't change while viewing this without causing it to unmount, so doesn't need to subscribe
   const needToAccept = isSignedIn && !acceptedTos
+  const [termsOfService, setTermsOfService] = useState('')
+  const [tosLoaded, setTosLoaded] = useState(false)
```
zarsky-broad: You can get rid of this state entirely if you use `!!termsOfService` as the render condition

src/pages/TermsOfService.js
```diff
@@ -1,307 +1,34 @@
...
-implemented the Clinical Features.
-
-Terms as of February 12, 2020.
-`
-
 const TermsOfServicePage = () => {
   const [busy, setBusy] = useState()
   const { isSignedIn, acceptedTos } = authStore.get() // can't change while viewing this without causing it to unmount, so doesn't need to subscribe
   const needToAccept = isSignedIn && !acceptedTos
+  const [termsOfService, setTermsOfService] = useState('')
```
zarsky-broad: Consider leaving the initial value as undefined; we often use undefined in the UI to represent data that specifically hasn't been loaded from the backend yet.

src/pages/TermsOfService.js
```diff
@@ -1,307 +1,34 @@
...
+
+  Utils.useOnMount(() => {
+    const renderTos = _.flow(
+      Utils.withBusyState(setBusy),
+      withErrorReporting('There was an error retrieving our terms of service.')
+    )(async () => {
+      setTermsOfService(await Ajax().User.getTos())
+      setTosLoaded(true)
+    })
+    renderTos()
```
zarsky-broad: nit: I'd call this `loadTos`, since that's what it's doing

zarsky-broad (approved):
Code looks good, pending some comments. Didn't test.

src/pages/TermsOfService.js
sagehen03: When I tried that initially, the markdown viewer threw an exception.  I'll look again to see whether that's still the case.

src/pages/TermsOfService.js
sagehen03: I initially called it that, but it also updates the state that drives the UI according to the response from `Ajax().User.getTos()` so I felt like `loadTos` was misleading.

src/pages/TermsOfService.js
sagehen03: Yes, still the case.  Let me know if you have other suggestions. image

zarsky-broad: Ah, I didn't realize there wasn't a spinner on this page. I'd put the markdown renderer in a ternary with a spinner as the other branch, with `!!termsOfService` as the condition

src/pages/TermsOfService.js
zarsky-broad: I hear you; that said, this doesn't actually render the text, it just puts it into state, which independently causes the render. Generally, we think of saving to state as part of loading, so I'd say load is the more idiomatic name.

src/pages/TermsOfService.js
sagehen03: I don't think I totally follow you, but I did make the change to include a spinner and set the initial `termsOfService` state to undefined or null or whatever `useState()` does and then use a ternary for including the markdown renderer or not depending upon state.  If my latest change isn't what you were looking for can you send me a code snippet?

zarsky-broad: Perfect, that's exactly it

PR #2699, opened 2021-11-22
[SUP-585] Increase GCE VM boot disk size

kyuksel:
https://broadworkbench.atlassian.net/browse/SUP-585

This UI change is for cost estimate display purposes only by reflecting the fix for the Leo prod issue where GPU VM creation fails due to insufficient boot disk size.


Qi77Qi (approved):
thanks @kyuksel !

zarsky-broad (approved):
Code change is definitely safe enough, trusting the team review for behavior being correct.

PR #2700, opened 2021-11-23
[PROD-623] Hid dataCatalog information behind the feature flag

lu-c:
I think this should fix the problem on prod, but I'm unable to completely reproduce locally.

### DIFF
--- a/src/pages/library/dataBrowser-utils.js
+++ b/src/pages/library/dataBrowser-utils.js
@@ -2,7 +2,7 @@ import _ from 'lodash/fp'
 import qs from 'qs'
 import { useState } from 'react'
 import { Ajax } from 'src/libs/ajax'
-import { getConfig } from 'src/libs/config'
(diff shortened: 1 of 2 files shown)


src/pages/library/dataBrowser-utils.js
```diff
@@ -71,6 +71,7 @@ const extractTags = snapshot => {
 }
 
 export const useDataCatalog = () => {
+  const skip = !isDataBrowserVisible()
```
dmohs: Var name suggestion?
```suggestion
  const skipMetadata = !isDataBrowserVisible()
```

dmohs (approved):
One naming suggestion, but not a big deal if getting this merged is urgent.

zarsky-broad (approved):
Could you inline `skip`?

lu-c:
Nice thank you everyone!

PR #2701, opened 2021-11-23
Remove state

kyuksel:
### State
- Remove global (store) state since it wasn't used outside the scope of runtime creation
- Remove ComputeModal state

### Misc
- Favour `Utils.cond` over `if-else` for consistency and a slight bit more concision

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -176,7 +176,6 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   const [customEnvImage, setCustomEnvImage] = useState('')
   const [jupyterUserScriptUri, setJupyterUserScriptUri] = useState('')
   const [sparkMode, setSparkMode] = useState(false)
-  const [bucketLocation, setBucketLocation] = useState(defaultLocation)
   const [computeConfig, setComputeConfig] = useState({
     selectedPersistentDiskSize: defaultGcePersistentDiskSize,
     masterMachineType: defaultGceMachineType,
@@ -588,10 +587,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     // bucketLocation === 'US' means the bucket is US multi-regional.
     // For a US multi-regional bucket, the computeRegion needs to be US-CENTRAL1 in order to be considered "in the same location".
     // Currently, US is the only multi-region supported in Terra
-    if (bucketLocation === defaultLocation) {
+    if (location === defaultLocation) {
       return computeConfig.computeRegion !== defaultComputeRegion
     } else {
-      return computeConfig.computeRegion !== bucketLocation
+      return computeConfig.computeRegion !== location
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/libs/state.js
```diff
@@ -41,8 +41,6 @@ export const knownBucketRequesterPaysStatuses = Utils.atom({})
 
 export const requesterPaysProjectStore = Utils.atom()
 
-export const workspaceBucketLocationStore = Utils.atom()
```
wnojopra, line 44: By removing this state, it re-introduces the bug you found that I fixed with WorkspaceContainer. See this comment: https://github.com/DataBiosphere/terra-ui/pull/2680#issuecomment-973167744 

To reproduce the bug:

1) Navigate to a workspace with a non-US bucket. Open the ComputeModal. Notice it says `This cloud environment will be created in the region $region.` Click customize. Notice the location dropdown says `$region`.
2) Close the compute modal, and switch to a different tab (say data). Open the Cloud Environment modal here. Notice it now says  `This cloud environment will be created in the region us-central1.` Click customize. Notice the location dropdown says `us-central1`. It shouldn't say us-central1, it should say $region. 

The issue is switching tabs creates a new WorkspaceContainer, but doesn't call refreshWorkspace. Thus, the need for the workspaceBucketLocationStore state.

wnojopra (changes requested):

src/components/ComputeModal.js
```diff
@@ -176,7 +176,6 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   const [customEnvImage, setCustomEnvImage] = useState('')
   const [jupyterUserScriptUri, setJupyterUserScriptUri] = useState('')
   const [sparkMode, setSparkMode] = useState(false)
-  const [bucketLocation, setBucketLocation] = useState(defaultLocation)
```
wnojopra, line 179: Thanks, the changes in this file look right to me. We have the computeRegion and computeZone states holding the location of the to-be created cloud env, and we have the location prop to hold the location of the workspace bucket. Thus, there is no need for the bucketLocation state here.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -258,6 +257,11 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
       refreshApps()
     }
 
+    const loadBucketLocation = async (googleProject, bucketName) => {
+      const bucketLocation = await Ajax(signal).Workspaces.workspace(namespace, name).checkBucketLocation(googleProject, bucketName)
```
wnojopra, line 261: Why is making more ajax calls preferred over adding global state?

wnojopra: We discussed over slack that adding global state would cause more state subscriptions than we'd care for. The real fix here is to add bucketLocation to the workspace object in Rawls.

wnojopra (approved):

PR #2702, opened 2021-11-23
Reuse a Cromwell data disk if it exists, and fix spurious warnings about multiple disks [BW-897, BW-931].

cahrens:
When creating a Cromwell App, if a disk already exists from a previously deleted Cromwell app in that workspace, reuse it.

This also fixes a bug where if one tries to create a Galaxy instance in a workspace where an unattached disk originally created by Cromwell exists, then we attempt to reuse that Cromwell disk for Galaxy.

NOTE: to test this ticket you must test against dev.

To test BW-897:
1) Create a Workspace
2) Create a Crowell instance
3) Create a Galaxy instance.
4) Delete the Cromwell instance leaving the disk behind.
5) Delete the Galaxy instance, leaving the disk behind.
6) Create new Galaxy and Cromwell instances and verify they attach to the correct pre-existing disks (looking at information displayed in Cloud Environments/clusters page).

To test BW-931:
It is easy to verify that we don't show the warning when you have both a Jupyter disk and a Cromwell disk in the same project. To test that we DO show the warning if you > 1 disk for a given app type is somewhat harder. To test that, I removed the modification to CromwellModal so that we always create a new disk instance when creating a Cromwell instance. Then I could get 2 Cromwell disks in the same workspace, and I verified that we correctly show the warning. (Note: in this screenshot there is also a Galaxy disk left over from a deleted Galaxy instance… so 1 Jupyter, 1 Galaxy, and 2 Cromwell).

image


src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -183,6 +183,10 @@ const useCloudEnvironmentPolling = googleProject => {
         Ajax(signal).Disks.list({ googleProject, creator: getUser().email, saturnApplication: tools.galaxy.appType }),
         Ajax(signal).Disks.list({ googleProject, creator: getUser().email, saturnApplication: tools.cromwell.appType })
       ]) : [[], [], [], []]
+      _.forEach(disk => disk.appType = tools.galaxy.appType)(galaxyDisks)
```
cahrens: @salonishah11 if you were to add "formattedBy" to the information returned by the Disks list method, then I could key off of that value instead of injecting appType here.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -183,6 +183,10 @@ const useCloudEnvironmentPolling = googleProject => {
         Ajax(signal).Disks.list({ googleProject, creator: getUser().email, saturnApplication: tools.galaxy.appType }),
```
cahrens: @kyuksel Thanks for picking up this review! What I would like to do here is remove the explicit mentioning of Galaxy and Cromwell by iterating through the tools and adding a Disks.list call for each tool that is an app. Then I need to inject the app type into the returned response for each.

Is there any downside to doing each of those calls as its own await so I can easily do it in a loop? Or can you think of some way to do it with the existing Promise.all?

src/pages/Environments.js
```diff
@@ -426,16 +429,23 @@ const Environments = () => {
...
-              const runtimeDisks = _.remove(disk => _.includes(disk.name, galaxyDiskNames), disks)
-              const runtimeDiskNames = _.map(disk => disk.name, runtimeDisks)
-              const multipleRuntimeDisks = _.remove(disk => _.includes(disk.name, galaxyDiskNames) || disk.status === 'Deleting',
-                disksByProject[disk.googleProject]).length > 1
-              const multipleGalaxyDisks = _.remove(disk => _.includes(disk.name, runtimeDiskNames) || disk.status === 'DELETING',
+              const appDiskNames = _.map(disk => disk.name, appDisks)
+              const multipleRuntimeDisks = _.remove(disk => _.includes(disk.name, appDiskNames) || disk.status === 'Deleting',
                 disksByProject[disk.googleProject]).length > 1
+              const getAppType = diskName => {
+                // appDisks has appType populated on it if the disk was related to an app, but
```
cahrens: @salonishah11 this is another place where our life would be easier of "provisionedBy" was returned in the disk information. Then I would be able to tell if a disk was provisioned by an app without having to search in appDisks.

src/pages/workspaces/workspace/WorkspaceContainer.js
salonishah11: If it is non-trivial, I will incorporate it in [my current PR](https://github.com/DataBiosphere/leonardo/pull/2401) 👍

src/components/CromwellModal.js
```diff
@@ -24,18 +21,19 @@ const titleId = 'cromwell-modal-title'
...
     const [loading, setLoading] = useState(false)
+    const currentDataDisk = currentPersistentDisk(tools.cromwell.appType, apps, appDataDisks)
 
     const createCromwell = _.flow(
       Utils.withBusyState(setLoading),
       withErrorReportingInModal('Error creating Cromwell', onDismiss)
     )(async () => {
       await Ajax().Apps.app(googleProject, Utils.generateAppName()).create({
-        defaultKubernetesRuntimeConfig, diskName: Utils.generatePersistentDiskName(), diskSize: defaultDataDiskSize,
+        defaultKubernetesRuntimeConfig, diskName: currentDataDisk ? currentDataDisk.name : Utils.generatePersistentDiskName(), diskSize: defaultDataDiskSize,
```
zarsky-broad: What would you think about:
```suggestion
        defaultKubernetesRuntimeConfig, diskName: currentDataDisk?.name ?? Utils.generatePersistentDiskName(), diskSize: defaultDataDiskSize,
```
(`??` is [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator))

src/libs/runtime-utils.js
```diff
@@ -235,17 +235,19 @@ export const appIsSettingUp = app => {
...
-  // if the disk is attached to an app (or being detached from a deleting app), return that disk. otherwise,
-  // return the newest galaxy disk that the user has unattached to an app
+  // If the disk is attached to an app (or being detached from a deleting app), return that disk. Otherwise,
+  // return the newest unattached disk that was provisioned by the desired appType.
+  const desiredAppType = appType
   return currentDataDiskName ?
     _.find({ name: currentDataDiskName }, appDataDisks) :
     _.last(_.sortBy('auditInfo.createdDate',
-      _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), appDataDisks)))
+      _.filter(({ name, status, appType }) => appType === desiredAppType && status !== 'Deleting' && !_.includes(name, attachedDataDiskNames),
```
zarsky-broad: This case feels like it should be a flow at this point...

src/pages/Environments.js
```diff
@@ -129,17 +129,21 @@ const Environments = () => {
 
   const refreshData = Utils.withBusyState(setLoading, async () => {
     const creator = getUser().email
-    const [newRuntimes, newDisks, galaxyDisks, newApps] = await Promise.all([
+    const [newRuntimes, newDisks, galaxyDisks, cromwellDisks, newApps] = await Promise.all([
       Ajax(signal).Runtimes.list({ creator }),
       Ajax(signal).Disks.list({ creator }),
-      Ajax(signal).Disks.list({ creator, saturnApplication: 'galaxy' }),
+      Ajax(signal).Disks.list({ creator, saturnApplication: tools.galaxy.appType }),
+      Ajax(signal).Disks.list({ creator, saturnApplication: tools.cromwell.appType }),
```
zarsky-broad: Is there a reason to call `Disks.list` 3 times instead of once and splitting it up locally? It seems to my untrained eye that all app types would be in that first call.

src/pages/Environments.js
```diff
@@ -129,17 +129,21 @@ const Environments = () => {
...
       Ajax(signal).Apps.listWithoutProject({ creator })
     ])
     setRuntimes(newRuntimes)
     setDisks(newDisks)
-    setGalaxyDisks(galaxyDisks)
     setApps(newApps)
 
+    _.forEach(disk => disk.appType = tools.galaxy.appType)(galaxyDisks)
+    _.forEach(disk => disk.appType = tools.cromwell.appType)(cromwellDisks)
+    setAppDisks(_.concat(galaxyDisks, cromwellDisks))
```
zarsky-broad: Can I sell you on a functional version? With potential replacement for multiple calls above?
```js
setAppDisks(_.mapKeys(key => key === 'saturnApplication' ? 'appType' : key, newDisks)
```
or if you don't like renaming the keys
```js
setAppDisks(_.map(disk => _.set('appType', disk.saturnApplication, disk), newDisks)
```

src/pages/Environments.js
```diff
@@ -426,16 +429,23 @@ const Environments = () => {
...
-              const multipleRuntimeDisks = _.remove(disk => _.includes(disk.name, galaxyDiskNames) || disk.status === 'Deleting',
-                disksByProject[disk.googleProject]).length > 1
-              const multipleGalaxyDisks = _.remove(disk => _.includes(disk.name, runtimeDiskNames) || disk.status === 'DELETING',
+              const appDiskNames = _.map(disk => disk.name, appDisks)
+              const multipleRuntimeDisks = _.remove(disk => _.includes(disk.name, appDiskNames) || disk.status === 'Deleting',
                 disksByProject[disk.googleProject]).length > 1
+              const getAppType = diskName => {
+                // appDisks has appType populated on it if the disk was related to an app, but
+                // disksByProject (which is populated from disks) does not.
+                return _.find(disk => disk.name === diskName)(appDisks)?.appType
```
zarsky-broad: In general, we like to avoid the double-call when possible:
```suggestion
                return _.find({ name: diskName }, appDisks)?.appType
```

src/pages/workspaces/workspace/WorkspaceContainer.js
zarsky-broad: see above for another way to do this

src/components/CromwellModal.js
cahrens: Huh, that looks like magic to me. :)

src/libs/runtime-utils.js
cahrens: What is the advantage of making it a flow?

src/pages/Environments.js
cahrens: I believe the issue is that we don't know which apps or runtimes the disks returned by the first call are associated with. We could simplify this if the information returned from the server included the provisionedBy information-- that's what I was suggesting to @salonishah11.

The use case is that we wish to warn if 2 or more disks in a workspace are associated with the same app or runtime.

src/pages/Environments.js
cahrens: Unfortunately I can't really follow the suggestions, but the disk information returned does not include saturnApplication as a field. I definitely would like to remove the multiple calls, but I think that requires more information to be returned from the server so that we can tell which app or runtime a disk was associated with.

src/components/CromwellModal.js
cahrens: Changed both locations.

src/pages/Environments.js
```diff
@@ -426,16 +429,23 @@ const Environments = () => {
             headerRenderer: () => h(Sortable, { sort: diskSort, field: 'project', onSort: setDiskSort }, ['Billing project']),
             cellRenderer: ({ rowIndex }) => {
               const disk = filteredDisks[rowIndex]
-              const galaxyDiskNames = _.map(disk => disk.name, galaxyDisks)
-              const runtimeDisks = _.remove(disk => _.includes(disk.name, galaxyDiskNames), disks)
-              const runtimeDiskNames = _.map(disk => disk.name, runtimeDisks)
-              const multipleRuntimeDisks = _.remove(disk => _.includes(disk.name, galaxyDiskNames) || disk.status === 'Deleting',
-                disksByProject[disk.googleProject]).length > 1
-              const multipleGalaxyDisks = _.remove(disk => _.includes(disk.name, runtimeDiskNames) || disk.status === 'DELETING',
```
cahrens, line 434: Note that there was a bug here with DELETING vs. Deleting. These string status comparisons are quite fragile, especially since apparently the app status strings use all upper case letters.

src/pages/Environments.js
cahrens: This is what we get back from the server.

image

src/components/CromwellModal.js
cahrens: I do wonder though if this goes against our policy of trying to make the code familiar/friendly to folks who aren't JavaScript experts. Tertiary operators are a common pattern in many different languages, but this nullish coalescing, while elegant, is likely to be unfamiliar to pretty much everyone.

kyuksel:
I

1. created a Galaxy and a Cromwell app in the same workspace
2. deleted the Cromwell app while keeping its disk
3. attempted to create another Cromwell app

Got 

image

Note that the disk mentioned in the error belongs to the Galaxy disk so I'm guessing the logic that's supposed to differentiate between Cromwell and Galaxy disks has an issue. I haven't looked at the code yet.

kyuksel:
More like a UX question and not directly related to this PR: Was the default Cromwell disk size of `500GB` intentionally chosen? Asking in case there's room for cost savings (especially since there's currently no way to reduce it manually either).

cahrens:
> More like a UX question and not directly related to this PR: Was the default Cromwell disk size of `500GB` intentionally chosen? Asking in case there's room for cost savings (especially since there's currently no way to reduce it manually either).

We have tickets coming up soon to work out what sensible defaults are.

cahrens:
> Note that the disk mentioned in the error belongs to the Galaxy disk so I'm guessing the logic that's supposed to differentiate between Cromwell and Galaxy disks has an issue. I haven't looked at the code yet.

You must run against the FIAB with a Leo branch as mentioned in the description.

src/components/CromwellModal.js
kyuksel: I think your question is valid @cahrens and we should at least clarify what dev audience exactly we optimize for.

That being said, I had been wanting `??` all along not knowing it's existed 🤩 , for which, thank you @zarsky-broad!

src/libs/runtime-utils.js
```diff
@@ -235,17 +235,19 @@ export const appIsSettingUp = app => {
   return app && (app.status === 'PROVISIONING' || app.status === 'PRECREATING')
 }
```
kyuksel, line 269: (sorry that I missed this in earlier reviews 😞 )

1. `_.filter(app => app.appType === appType)` in the two places above could be written as `_.filter({ appType })`?
2. Can we use verbs for function names (e.g. `getCurrent*` instead of `current*`)? I know we have other places that we haven't converted yet but I'm hoping we convert them all albeit gradually.
3. How would you feel about refactoring `getCurrentAppForType()` so it would allow for specifying what statuses to trim (if any) such that we can reuse it in lieu of `currentAppIncludingDeleting` as well?

src/pages/Environments.js
kyuksel: Leonardo persists values for a field called `formattedBy` that's applicable for apps. Does it make sense to expose that through the API?

src/pages/Environments.js
```diff
@@ -426,16 +429,23 @@ const Environments = () => {
...
+              const appDiskNames = _.map(disk => disk.name, appDisks)
```
kyuksel: Could it be
```suggestion
              const appDiskNames = _.get('name', appDisks)
```
?

src/pages/Environments.js
```diff
@@ -426,16 +429,23 @@ const Environments = () => {
...
+              const multipleRuntimeDisks = _.remove(disk => _.includes(disk.name, appDiskNames) || disk.status === 'Deleting',
```
kyuksel: Consider
```suggestion
              const multipleRuntimeDisks = _.remove( ({ name, status }) => _.includes(name, appDiskNames) || status === 'Deleting',
```

src/pages/Environments.js
```diff
@@ -426,16 +429,23 @@ const Environments = () => {
             headerRenderer: () => h(Sortable, { sort: diskSort, field: 'project', onSort: setDiskSort }, ['Billing project']),
             cellRenderer: ({ rowIndex }) => {
               const disk = filteredDisks[rowIndex]
```
kyuksel: If we deconstruct disk via

```suggestion
              const { name, status, googleProject } = filteredDisks[rowIndex]
```

we can remove all `disk.`s below.

Another advantage of the latter, besides being less code, is that the reader doesn't need to keep seeing what props belong to as they already see it at the beginning (where the deconstruction is done) at once.

src/pages/Environments.js
```diff
@@ -426,16 +429,23 @@ const Environments = () => {
...
+                // appDisks has appType populated on it if the disk was related to an app, but
+                // disksByProject (which is populated from disks) does not.
+                return _.find({ name: diskName }, appDisks)?.appType
+              }
+              const hasMultipleAppDisks = diskName => {
+                const desiredAppType = getAppType(diskName)
+                const disks = _.remove(disk => getAppType(disk.name) !== desiredAppType || disk.status === 'Deleting',
+                  disksByProject[disk.googleProject])
+                return disks.length > 1
+              }
```
kyuksel: I wonder if it'd be easier to scan through the table renderer code (starting with `return h(FooterWrapper, ...`) if we didn't inline `hasMultipleAppDisks` which is sort of a low-level detail. It could be moved outside that block and  `getAppType` could be inlined within `hasMultipleAppDisks`. What do you think?

src/pages/workspaces/workspace/WorkspaceContainer.js
kyuksel: I _think_ I see what you're proposing. I feel like we may want to, instead, invest the time to add the necessary info to Leo's `listDisks()` so we don't need to make multiple `Ajax(signal).Disks.list` calls at the first place.

kyuksel (commented):
Comments inline.
Haven't tested yet.

src/pages/Environments.js
cahrens: Yes, I think that is what @salonishah11 is adding.

src/pages/workspaces/workspace/WorkspaceContainer.js
cahrens: I agree, I think we will be able to do it in a single .list call once the formatted by information is present, so it's not worth trying to do my proposed intermediary improvement.

src/pages/Environments.js
zarsky-broad: It'll still need to me `map`, `get` would just be for one object. That said,
```suggestion
              const appDiskNames = _.map('name', appDisks)
```

src/components/CromwellModal.js
zarsky-broad: Would you consider my suggestion with `||` instead of `??`, then?

src/libs/runtime-utils.js
zarsky-broad: It lets us un-nest the `last`, `sortBy`, and `filter`

src/libs/runtime-utils.js
cahrens: I will make the first 2 suggested changes, and I am going to rename "getCurrentAppForType" to just "getCurrentApp". All the methods now take appType as an argument, and the original (temporarily deprecated) getCurrentApp that assumed Galaxy is gone, so I think it is best to drop the "ForType" in the name.

For the 3rd suggestion, I would prefer not to touch that, especially with the lack of tests to make sure that I'm actually implementing the right behavior for the different statuses. But the currentAppIncludingDeleting method is only used within runtime-utils itself, so I will change it so that is not exported (and it is renamed getCurrentAppIncludingDeleting).

src/components/CromwellModal.js
cahrens: @zarsky-broad I don't understand what that would look like. But I already made the change you proposed with ??, and I don't feel like it's worth backing it out. But I do think it's worth discussing who our dev audience is, as there is tension between using the newer JavaScript language constructs and making the code easily understandable to people who aren't JS experts.

src/pages/Environments.js
cahrens: Ugh, to me `_.map('name', appDisks)` is so much harder to understand than the more `explicit _.map(disk => disk.name, appDisks)`.

src/pages/Environments.js
cahrens: Will do.

src/pages/Environments.js
cahrens: I will do this, but with more explicit names for "name" and "status" as it gets confusing to see where those are coming from. So `const { name: diskName, status: diskStatus, googleProject } = filteredDisks[rowIndex]`.

src/pages/Environments.js
cahrens: Sure.

src/libs/runtime-utils.js
cahrens: I would be so much more inclined to make this change if we had unit tests for these type of methods so I could be confident that I'm not introducing bugs.

src/pages/Environments.js
zarsky-broad:
```suggestion
    setAppDisks(_.concat(
      _.map(_.set('appType', tools.galaxy.appType), galaxyDisks),
      _.map(_.set('appType', tools.cromwell.appType), cromwellDisks)
    ))
```

cahrens:
@salonishah11 @kyuksel hold off on continuing your reviews. I am still addressing review feedback and trying to get some basic unit tests in place. I haven't had a chance to manually test the changes I've made thus far, so it is possible that I have broken some things.

I'll give you a heads-up when this PR is ready for re-review.

src/pages/Environments.js
```diff
@@ -129,17 +128,14 @@ const Environments = () => {
 
   const refreshData = Utils.withBusyState(setLoading, async () => {
     const creator = getUser().email
-    const [newRuntimes, newDisks, galaxyDisks, newApps] = await Promise.all([
+    const [newRuntimes, newDisks, newApps] = await Promise.all([
       Ajax(signal).Runtimes.list({ creator }),
       Ajax(signal).Disks.list({ creator }),
-      Ajax(signal).Disks.list({ creator, saturnApplication: 'galaxy' }),
       Ajax(signal).Apps.listWithoutProject({ creator })
     ])
```
cahrens, line 135: @kyuksel and @salonishah11 So it turns out that instead of relying on formattedBy, I could ask Disks.list to include the saturnApplication label in the return value-- I just found out about this by looking at https://github.com/DataBiosphere/terra-ui/pull/2662/files#diff-db07f65dd0ba5edbd535b2a231709c1f5901f4e1c43f39c333165d97c1abacc4R133

So I could still have a single Disks.list call like this:
`Ajax(signal).Disks.list({ creator, includeLabels: 'saturnApplication'}),`

And then apps would have their appType as `label.saturnApplication` (because we set that value at creation time), and I wouldn't need to use formattedBy to tell if multiple disks exist for the same app. And this approach does not have the current limitation of formattedBy not being set yet for disks that are in the Creating status.

Example disk info:
```
auditInfo: {creator: 'cahrens@gmail.com', createdDate: '2021-12-01T19:37:31.391603Z', destroyedDate: null, dateAccessed: '2021-12-01T19:37:32.255Z'}
blockSize: 4096
cloudContext: {cloudProvider: 'GCP', cloudResource: 'terra-test-f828b4cd'}
diskType: "pd-standard"
formattedBy: "CROMWELL"
googleProject: "terra-test-f828b4cd"
id: 14
labels: {saturnApplication: 'CROMWELL'}
name: "saturn-pd-141a3db4-6ce6-4896-b61b-6e85fdf38608"
size: 500
status: "Ready"
zone: "us-central1-a"

File: src/components/CromwellModal.js (hunk 4 above)
Comment (cahrens) [reply src/components/CromwellModal.js]: I ended up going back the ternary structure as we discussed in the mob.

File: src/libs/runtime-utils.js (hunk 5 above)
Comment (cahrens) [reply src/libs/runtime-utils.js]: I wrote a unit test and made the change to a flow.

File: src/libs/runtime-utils.js (hunk 10 above)
Comment (cahrens) [reply src/libs/runtime-utils.js:269]: I wrote a unit test and made the third proposed change as well.

File: src/pages/Environments.js (hunk 15 above)
Comment (kyuksel) [reply src/pages/Environments.js:135]: The ability to label disks was added recently (by [this PR](https://github.com/DataBiosphere/leonardo/pull/2340)) so we may not be able to rely on it for disks created prior to then, for backwards compatibility?

Comment (cahrens) [reply src/pages/Environments.js:135]: I did not know that labels were a recent thing. `formattedBy` seems like the better solution because it allows us to know who formatted non-app disks, but thus far the assumption has just been that there is only a single non-app disk that we care about (since Jupyter and RStudio can't exist in the same workspace at the same time). I'll wait until tomorrow and check in with others before making a decision about whether to stick with formattedBy or go with labels.

(Note though that the ability to retrieve labels may have been implemented later than when labels were sent with app creation)

Comment (kyuksel) [reply src/pages/Environments.js:135]: Is `formattedBy` populated for runtime (`Jupyter` | `RStudio`) disks? Even if it isn't, it should be okay to make the assumption that disk belongs to a runtime when `formattedBy` is `undefined`.

I also wonder if we should start populating `formattedBy` for `Jupyter` and `RStudio` as well so we could move to not relying on the assumption above in the future 🤔

Comment (cahrens) [reply src/pages/Environments.js:135]: `formattedBy` is "GCE" for Jupyter (and I assume RStudio). I have seen it be null in two cases-- when the disk is in "Creating" (always), and once when I tried creating a Galaxy instance and it errored due to permissions.

In the latter case, I got a Galaxy instance with an error indication on the Cloud Environments page, and the disk with null formattedBy. After I deleted the Galaxy instance, I was unable to delete the disk… if I remember correctly, delete was disabled because it said that I needed to delete the cloud environment it was attached to (even though there was none left). However, my fiab is no longer in that state and I don't recall what I later did.

I am trying to reproduce the bug where Jupyter tries to attached to Galaxy's disk in dev. I am pretty sure I saw it back when I was giving the Cromwell demo to the IA team, but it is tricky to reproduce because if I understand the cause of the problem, Jupyter has to try to get a disk instance while Galaxy's is still in "Creating".

File: src/pages/Environments.js (hunk 14 above)
Comment (cahrens) [reply src/pages/Environments.js]: This code simplified considerably, so now it is just a variable again.

File: src/pages/Environments.js (hunk 15 above)
Comment (cahrens) [reply src/pages/Environments.js:135]: I have gone back to using the saturnApplication label (though still with just a single call to Disks.list).

File: src/pages/Environments.js (hunk 7 above)
Comment (cahrens) [reply src/pages/Environments.js]: This code is no longer needed.

File: src/_tests/libs/runtime-utils.test.js [hunk 16]
```diff
@@ -0,0 +1,194 @@
...
+  kubernetesRuntimeConfig: { numNodes: 1, machineType: 'n1-highmem-8', autoscalingEnabled: false },
+  labels: {},
+  proxyUrls: { galaxy: 'https://leonardo-fiab.dsde-dev.broadinstitute.org/a-app-69200c2f-89c3-47db-874c-b770d8de737f/galaxy' },
+  status: 'DELETING'
+}
+
+const mockApps = [cromwellProvisioning, cromwellRunning, galaxyRunning, galaxyDeleting]
+
+const galaxyDisk = {
+  formattedBy: 'GALAXY',
```
cahrens: @salonishah11 for now I left formattedBy in the mock objects for the unit test (to help explain why I wasn't using it)… I included disk with status "Creating"). If you decide to remove formattedBy from the Disks API, I will delete it from here as well for consistency.

src/pages/Environments.js
```diff
@@ -306,10 +302,9 @@ const Environments = () => {
   }
 
   const renderDeleteDiskModal = disk => {
-    const diskName = disk.name
     return h(DeleteDiskModal, {
       disk,
-      isGalaxyDisk: _.some({ name: diskName }, galaxyDisks),
+      isGalaxyDisk: getDiskAppType(disk) === tools.galaxy.appType,
```
cahrens, line 307: The reason we were searching before was that the disk passed in was from the collection of all disks, and the only way we had to know if it was a Galaxy disk was to see if the disk also existed in the smaller galaxyDisks collection.

cahrens:
@salonishah11 @kyuksel this is ready for re-review. I am no longer using formattedBy, so it would be OK to remove it from the Leo API in you PR @salonishah11 .

Note that to test the Cromwell disk reattaching behavior, it is necessary to run against a fiab with the branch from https://github.com/DataBiosphere/leonardo/pull/2401.

src/pages/Environments.js
kyuksel: Follow-up [Slack thread](https://broadinstitute.slack.com/archives/C02H8KALYMD/p1638450691023700) for posterity

src/_tests/libs/runtime-utils.test.js
kyuksel: If there is no longer a use case for exposing `formattedBy` through Leo's get/list disk APIs, we'd prefer to not expose it.

cahrens: I have removed it.

src/components/RuntimeManager.js
```diff
@@ -197,8 +197,8 @@ export default class RuntimeManager extends PureComponent {
     const createdDate = new Date(runtime.createdDate)
     const dateNotified = getDynamic(sessionStorage, `notifiedOutdatedRuntime${runtime.id}`) || {}
     const rStudioLaunchLink = Nav.getLink(appLauncherTabName, { namespace, name, application: 'RStudio' })
-    const galaxyApp = getCurrentAppForType(tools.galaxy.appType)(apps)
-    const prevGalaxyApp = getCurrentAppForType(tools.galaxy.appType)(prevProps.apps)
+    const galaxyApp = getCurrentApp(tools.galaxy.appType)(apps)
+    const prevGalaxyApp = getCurrentApp(tools.galaxy.appType)(prevProps.apps)
```
salonishah11, line 201: What is `prevProps`?

src/pages/Environments.js
```diff
@@ -425,19 +420,16 @@ const Environments = () => {
...
+                disksByProject[googleProject]).length > 1
+              const forAppText = !!appType ? ` for ${_.capitalize(appType)}` : ''
               return h(Fragment, [
-                disk.googleProject,
-                disk.status !== 'Deleting' && (_.includes(disk.name, galaxyDiskNames) ? multipleGalaxyDisks : multipleRuntimeDisks) &&
+                googleProject,
+                diskStatus !== 'Deleting' && multipleDisksOfType &&
                 h(TooltipTrigger, {
-                  content: 'This billing project has multiple active persistent disks. Only one will be used.'
+                  content: `This billing project has multiple active persistent disks${forAppText}. Only one will be used.`
```
salonishah11: Would it make sense to add that only the **latest** one will be used?

salonishah11:
@cahrens While testing it locally, when I went to create an App (Galaxy or Cromwell) from the Create button, it failed with the below error. Interestingly I get a different error when I do this on Terra Dev (screenshot 3)

Screen Shot 2021-12-02 at 7 37 03 PM

Error from Local:

Screen Shot 2021-12-02 at 7 37 35 PM

Error on Terra Dev:

Screen Shot 2021-12-02 at 7 42 08 PM

cahrens:
> @cahrens While testing it locally, when I went to create an App (Galaxy or Cromwell) from the Create button, it failed with the below error. Interestingly I get a different error when I do this on Terra Dev (screenshot 3)

@salonishah11 Thanks for finding this bug! I obviously never use the "Create" button anymore. 

The bug was caused by a late refactoring in my previous PR, and I just pushed up a fix for it.

src/components/AnalysisModal.js
```diff
@@ -137,7 +137,7 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
         onDismiss()
       },
       onSuccess: () => {
-        resetView()
+        setViewMode(undefined)
```
cahrens, line 140: @jdcanas Justin, in my previous PR we decided to change these to `resetView` because the method seemed to "do more". Unfortunately, in the case of the app modals (Galaxy and Cromwell) it does "too much"…. you end up with a blank modal when you are trying to create a new instance. 

I don't understand the intended use case for `resetView`, so I am reverting to `setViewMode(undefined)` in the places where it was previously used. (Interestingly though, in the `renderComputeModal` case, resetView did not seem to cause a problem.)

src/components/RuntimeManager.js
cahrens: This was pre-existing code (I only changed the name of the method). `prevProps` is passed into the `componentDidUpdate` method, and it seems that it allows you to tell what changed. It is a React method.

```
/**
         * Called immediately after updating occurs. Not called for the initial render.
         *
         * The snapshot is only present if getSnapshotBeforeUpdate is present and returns non-null.
         */
        componentDidUpdate?(prevProps: Readonly<P>, prevState: Readonly<S>, snapshot?: SS): void;

File: src/pages/Environments.js (hunk 19 above)
Comment (cahrens) [reply src/pages/Environments.js]: Sure, done.

File: src/components/AnalysisModal.js [hunk 21]
```diff
@@ -104,8 +104,8 @@ export const AnalysisModal = Utils.withDisplayName('AnalysisModal')(
     const getEnvironmentView = () => Utils.switchCase(currentTool,
       [tools.Jupyter.label, renderComputeModal],
       [tools.RStudio.label, renderComputeModal],
-      [tools.galaxy.label, renderAppModal(GalaxyModalBase, tools.galaxy.label)],
-      [tools.cromwell.label, renderAppModal(CromwellModalBase, tools.cromwell.label)]
+      [tools.galaxy.label, () => renderAppModal(GalaxyModalBase, tools.galaxy.label)],
```
cahrens, line 107: FYI @kyuksel @jdcanas this fixes a current bug introduced by my last PR (a late refactoring in the review process that I didn't adequately test), where the "Create" button does not work for Galaxy and Cromwell. Creating an instance through the side cloud environment modal DOES work.

Just want to make sure you guys don't waste time filing a bug/investigating this if you see it. Hopefully this change will be out early next week, and I could also submit it as a separate PR if this one gets further delayed and folks are experiencing the bug.

src/components/notebook-utils.js
```diff
@@ -90,6 +90,9 @@ export const getAppType = label => _.find(tool => tool.label === label)(tools)?.
 // Does the tool label correspond to an app?
 export const isToolAnApp = label => getAppType(label) !== undefined
 
+// Returns registered appTypes.
+export const getAllAppTypes = _.flow(_.map('appType'), _.compact)(tools)
```
kyuksel: This seems to be a constant. Can we name it accordingly?
```suggestion
export const allAppTypes = _.flow(_.map('appType'), _.compact)(tools)
```

src/libs/runtime-utils.js
```diff
@@ -241,43 +241,55 @@ export const getCurrentRuntime = runtimes => {
...
+export const getCurrentAttachedDataDisk = (app, appDataDisks) => {
   return _.find({ name: app?.diskName }, appDataDisks)
 }
 
+// If the disk was attached to an app, return the appType. Otherwise return undefined.
+export const getDiskAppType = disk => {
+  const saturnApp = disk.labels.saturnApplication
+  // Do a case-insensitive match as disks have been created with both "galaxy" and "GALAXY".
+  const appType = _.find(type => type.toLowerCase() === saturnApp?.toLowerCase(), getAllAppTypes)
+  return appType
```
kyuksel: Especially since the function name nicely describes what it'd return:
```suggestion
  return _.find(type => type.toLowerCase() === saturnApp?.toLowerCase(), getAllAppTypes)
```

src/pages/Environments.js
```diff
@@ -425,19 +420,16 @@ const Environments = () => {
...
-              const galaxyDiskNames = _.map(disk => disk.name, galaxyDisks)
-              const runtimeDisks = _.remove(disk => _.includes(disk.name, galaxyDiskNames), disks)
-              const runtimeDiskNames = _.map(disk => disk.name, runtimeDisks)
-              const multipleRuntimeDisks = _.remove(disk => _.includes(disk.name, galaxyDiskNames) || disk.status === 'Deleting',
-                disksByProject[disk.googleProject]).length > 1
-              const multipleGalaxyDisks = _.remove(disk => _.includes(disk.name, runtimeDiskNames) || disk.status === 'DELETING',
-                disksByProject[disk.googleProject]).length > 1
+              const { status: diskStatus, googleProject } = filteredDisks[rowIndex]
+              const appType = getDiskAppType(filteredDisks[rowIndex])
+              const multipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
```
kyuksel, line 425: `diskStatus` instead of `disk.status`

Also, a bit nitty but since it took me some time to parse, I figured I'd mention that it helps me when a constant/function of type boolean has a name that starts with a modal or auxiliary verb, and perhaps even `there` at times.
```suggestion
              const areThereMultipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || diskStatus === 'Deleting',
```

If that's too cumbersome, `isMultipleDisksOfType` would do too even though it's grammatically incorrect.

kyuksel (commented):
Will resume re-reviewing after 4pm today.

src/components/notebook-utils.js
cahrens: I thought you wanted the get prefix. I'm personally used to all caps for constants… but that's not our convention here, correct? I feel like I'm getting confused about how things should be named. Is "get" and "is" reserved for functions/methods?

src/pages/Environments.js
cahrens: `disk.status` is actually correct instead of `diskStatus`. We are iterating over the disks for the project, and removing any that are Deleting or are of a different "type". `diskStatus` is for the specific disk represented by the row we are rendering.

Honestly, I'm not a fan of either `areThereMultipleDisksOfType` or `isMultipleDisksOfType`. But being a Python person, I'm not used to sticking `get` or `is` in front of everything (which is why I'm now confused!). To me, `multipleDisksOfType` seems clear.

src/components/notebook-utils.js
cahrens: (I did make the suggested change, but would still like clarification on our naming standards)

salonishah11 (approved):
Tested it locally against my Leo branch in a FIAB. LGTM 👍  
Thank you for adding tests and fixing the bug! 💯

cahrens:
FYI Saloni's Leo PR has merged so this can now be tested against dev.

src/pages/Environments.js
kyuksel: > disk.status is actually correct instead of diskStatus. We are iterating over the disks for the project, and removing any that are Deleting or are of a different "type". diskStatus is for the specific disk represented by the row we are rendering.

Ah right sorry, I'd misread it.

> Honestly, I'm not a fan of either areThereMultipleDisksOfType or isMultipleDisksOfType. But being a Python person, I'm not used to sticking get or is in front of everything (which is why I'm now confused!). To me, multipleDisksOfType seems clear.

Especially in absence of static typing where the IDE can help more, clear naming conventions act as type annotation to me. Not a blocker for this PR but I will make a note to discuss offline.

src/components/notebook-utils.js
kyuksel: > I did make the suggested change, but would still like clarification on our naming standards

I would too. See https://github.com/DataBiosphere/terra-ui/pull/2702#discussion_r762449605.

kyuksel:
> src/_tests/components/notebook-utils.test.js

Why is the test directory name (`_tests`) prefixed with an underscore? Is it a Javascript convention?

src/components/RuntimeManager.js
kyuksel: It's one of the [React lifecycle methods](https://www.w3schools.com/react/react_lifecycle.asp) for `class` components. We have mostly moved to using `function` (instead of `class`) components so we don't have a lot of `componentDidUpdate`s left. [useEffect hook](https://reactjs.org/docs/hooks-effect.html) for `function` components replaces its functionality (and more).

src/libs/runtime-utils.js
```diff
@@ -241,43 +241,55 @@ export const getCurrentRuntime = runtimes => {
...
-  const currentDataDiskName = currentGalaxyApp?.diskName
-  const attachedDataDiskNames = _.without([undefined], _.map(app => app.diskName, apps))
-  // if the disk is attached to an app (or being detached from a deleting app), return that disk. otherwise,
-  // return the newest galaxy disk that the user has unattached to an app
-  return currentDataDiskName ?
-    _.find({ name: currentDataDiskName }, appDataDisks) :
-    _.last(_.sortBy('auditInfo.createdDate',
-      _.filter(({ name, status }) => status !== 'Deleting' && !_.includes(name, attachedDataDiskNames), appDataDisks)))
+  const currentApp = getCurrentAppIncludingDeleting(appType)(apps)
+  const currentDiskName = currentApp?.diskName
```
kyuksel, line 273:
```suggestion
  const { diskName: currentDiskName } = getCurrentAppIncludingDeleting(appType)(apps)
```

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -177,17 +177,13 @@ const useCloudEnvironmentPolling = googleProject => {
...
-        Ajax(signal).Runtimes.list({ googleProject, creator: getUser().email }),
-        Ajax(signal).Disks.list({ googleProject, creator: getUser().email, saturnApplication: tools.galaxy.appType }),
-        Ajax(signal).Disks.list({ googleProject, creator: getUser().email, saturnApplication: tools.cromwell.appType })
-      ]) : [[], [], [], []]
-      const appDisks = _.concat(galaxyDisks, cromwellDisks)
-      const appDiskNames = _.map(disk => disk.name, appDisks)
+      const [newDisks, newRuntimes] = googleProject ? await Promise.all([
+        Ajax(signal).Disks.list({ googleProject, creator: getUser().email, includeLabels: 'saturnApplication' }),
+        Ajax(signal).Runtimes.list({ googleProject, creator: getUser().email })
+      ]) : [[], []]
```
kyuksel, line 183: Serenity 😍

cahrens:
> > src/_tests/components/notebook-utils.test.js
> 
> Why is the test directory name (`_tests`) prefixed with an underscore? Is it a Javascript convention?

Not as far as I know… but it already was `_tests`. I assume ci scripts and such have that name in it, so I don't wish to change it.

src/libs/runtime-utils.js
```diff
@@ -241,43 +241,55 @@ export const getCurrentRuntime = runtimes => {
...
   return _.find({ name: app?.diskName }, appDataDisks)
 }
 
+// If the disk was attached to an app, return the appType. Otherwise return undefined.
+export const getDiskAppType = disk => {
+  const saturnApp = disk.labels.saturnApplication
+  // Do a case-insensitive match as disks have been created with both "galaxy" and "GALAXY".
+  const appType = _.find(type => type.toLowerCase() === saturnApp?.toLowerCase(), allAppTypes)
+  return appType
+}
```
kyuksel, line 264: Consider
```suggestion
export const getDiskAppType = ({ labels: { saturnApplication } }) => {
  // Do a case-insensitive match as disks have been created with both "galaxy" and "GALAXY".
  return _.find(type => type.toLowerCase() === saturnApplication?.toLowerCase(), allAppTypes)
}
```

kyuksel (commented):
Finished re-reviewing the non-test code. Next, I'll test the code and then review the test code.

P.S. Thank you very much for taking the time to include a clear PR description and annotating your PR with helpful inline comments  🙏🏼

kyuksel:
I seem to be able to delete a `Cromwell` only on `#clusters` page, and not through `Cromwell Modal`. Is that intentional?

image

kyuksel:
Done testing various combinations of apps/tools with/without disks and various statuses. All looks good with one question above to make sure the behaviour was intentional.

cahrens:
> I seem to be able to delete a `Cromwell` only on `#clusters` page, and not through `Cromwell Modal`. Is that intentional?

Yes, that is intentional. All we have implemented at this point in the Cromwell modal is the ability to create an instance.

src/libs/runtime-utils.js
cahrens: That doesn't work because `getCurrentAppIncludingDeleting(appType)(apps)` can return undefined (my unit test failed when I tried your suggestion!).

src/libs/runtime-utils.js
cahrens: I would prefer to stay with what I had. `saturnApplication` won't exist in labels for notebook disks, and in that case the object deconstruction seems confusing to me.

src/_tests/libs/runtime-utils.test.js
```diff
@@ -0,0 +1,189 @@
...
+  diskName: 'saturn-pd-026594ac-d829-423d-a8df-76fe96f5b4e7',
+  errors: [],
+  googleProject: 'terra-test-e4000484',
+  kubernetesRuntimeConfig: { numNodes: 1, machineType: 'n1-highmem-8', autoscalingEnabled: false },
+  labels: {},
+  proxyUrls: { galaxy: 'https://leonardo-fiab.dsde-dev.broadinstitute.org/a-app-69200c2f-89c3-47db-874c-b770d8de737f/galaxy' },
+  status: 'RUNNING'
+}
+
+// Newer than galaxyRunning
```
kyuksel, line 51: Did you mean
```suggestion
// Older than galaxyRunning
```
?

Why is this comment (and the similar one above) necessary?

kyuksel: Ah I think I now understand what you mean by `older/newer` (`createdDate` timestamps). I initially thought you were referring to the lifecycle statuses of compute. Since the comment threw me off, perhaps we could explain a bit more in the comment what we mean by `older/newer` and why it's relevant.

src/_tests/libs/runtime-utils.test.js
```diff
@@ -0,0 +1,189 @@
...
+  errors: [],
+  googleProject: 'terra-test-e4000484',
+  kubernetesRuntimeConfig: { numNodes: 1, machineType: 'n1-highmem-8', autoscalingEnabled: false },
+  labels: {},
+  proxyUrls: { galaxy: 'https://leonardo-fiab.dsde-dev.broadinstitute.org/a-app-69200c2f-89c3-47db-874c-b770d8de737f/galaxy' },
+  status: 'DELETING'
+}
+
+const mockApps = [cromwellProvisioning, cromwellRunning, galaxyRunning, galaxyDeleting]
+
```
kyuksel, line 68: We tend to have a naming scheme like below

galaxyDiskReady
galaxyDiskDeleting
cromwellDiskUnattached

for more consistency. What do you think?

kyuksel (approved):
Thank you very much for taking the time to figure out how to add the first tests, and writing them 🏅

src/_tests/libs/runtime-utils.test.js
cahrens: These are the apps, not the disks. The disks have their own statuses which are already in their names.

src/_tests/libs/runtime-utils.test.js
cahrens: Yes, the comments were in reference to the start dates. I was attempting to make some of the test assertions easier to understand so that one didn't have to think through the dates. I'm not sure what else I would say though, as the tests call out the dependency on date newest).

src/_tests/libs/runtime-utils.test.js
kyuksel: Oh sorry, I was referring to the disk objects below.

PR #2705, opened 2021-11-30
[PF-1232] Fix for galaxy compute cost estimates

wnojopra:
Tested locally

Before:
image

After:
image

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -195,7 +195,8 @@ export const getGalaxyCost = (app, dataDiskSize) => {
  */
 export const getGalaxyComputeCost = app => {
   const appStatus = app?.status?.toUpperCase()
-  const defaultNodepoolComputeCost = getHourlyCostForMachineType('n1-standard-1', defaultComputeRegion, false)
+  // Galaxy uses defaultComputeRegion because we're not yet enabling other locations for Galaxy apps.
+  const defaultNodepoolComputeCost = getHourlyCostForMachineType(defaultGceMachineType, defaultComputeRegion, false)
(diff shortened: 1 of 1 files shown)


kyuksel:
Verified the config below (that I arbitrarily chose) now displays the same costs (+- a small rounding difference) as `dev` used to prior to the regression.

image

Qi77Qi (approved):

src/libs/runtime-utils.js
```diff
@@ -195,11 +195,12 @@ export const getGalaxyCost = (app, dataDiskSize) => {
  */
 export const getGalaxyComputeCost = app => {
   const appStatus = app?.status?.toUpperCase()
-  const defaultNodepoolComputeCost = machineCost('n1-standard-1')
+  const defaultNodepoolComputeCost = getHourlyCostForMachineType('n1-standard-1', defaultComputeRegion, false)
```
kyuksel: I know it's not your PR but I wonder if we can opportunistically unhardcode `n1-standard-1` and instead define it with something like `defaultNodepoolMachineType`?

kyuksel (approved):
Thanks for the quick fix! The nit inline is not a blocker.

src/libs/runtime-utils.js
kyuksel: It may also help to have a comment saying we use `defaultComputeRegion` because we're not yet enabling other locations for Galaxy apps.

wnojopra: Done. Turns out theres already a constant named `defaultGceMachineType` that is equal to `n1-standard-1`

PR #2707, opened 2021-12-02
[DC-73] Added beta disclaimer

lu-c:
After:
Screen Shot 2021-12-02 at 5 00 28 PM

On clicking OK, the val is set in localstorage

To test, go to the databrowser, click on OK, confirm it's gone. Refresh the page, confirm it's no loading anymore.


fboulnois (approved):

src/pages/library/DataBrowser.js
```diff
@@ -221,11 +224,33 @@ const Browser = () => {
   const [selectedData, setSelectedData] = useState([])
   const [requestDatasetAccessList, setRequestDatasetAccessList] = useState()
   const { dataCatalog, loading } = useDataCatalog()
+  const [acknowledged, setAcknowledged] = useState(getDynamic(localStorage, acknowledgmentStorageKey))
```
petesantos: For this we may want to use a `staticStorageSlot` as `dynamicStorage` is meant to be used for more transient settings (e.g. navigating between routes and preserving some data). Dynamic storage will be the first to get deleted. Also, by using a `staticStorageSlot` we can utilize the `useStore` hook.

Also, when assigning a key to this value, we should add the user id so we can fetch the correct key if the user logs out and in again as another user, or another user signs in, etc... You can retrieve this with the `getUser` function from `src/libs/auth` via `getUser().id`

lu-c: i think i've updated this to fix this now, please let me know if this works!

src/pages/library/DataBrowser.js
```diff
@@ -16,6 +18,8 @@ import { importDataToWorkspace, snapshotAccessTypes, useDataCatalog } from 'src/
 import { RequestDatasetAccessModal } from 'src/pages/library/RequestDatasetAccessModal'
 
 
+export const acknowledgmentStore = staticStorageSlot(localStorage, `catalog-beta-acknowledgment-${getUser().id}`)
```
petesantos: Apologies for the churn on this - when testing, it looks like the userId isn't populated when this code is initially run.
I have a few changes that should correct the issue and allow us to capture the response for a specific user:
```suggestion
export const acknowledgmentStore = staticStorageSlot(localStorage, 'catalog-beta-acknowledgment')
```

src/pages/library/DataBrowser.js
```diff
@@ -221,11 +225,30 @@ const Browser = () => {
   const [selectedData, setSelectedData] = useState([])
   const [requestDatasetAccessList, setRequestDatasetAccessList] = useState()
   const { dataCatalog, loading } = useDataCatalog()
+  const acknowledged = Utils.useStore(acknowledgmentStore)
```
petesantos: This will give us the user ID when it populates
```suggestion
  const acknowledged = Utils.useStore(acknowledgmentStore) || {}
  const { user: { id } } = Utils.useStore(authStore)
```

src/pages/library/DataBrowser.js
```diff
@@ -221,11 +225,30 @@ const Browser = () => {
...
 
   const toggleSelectedData = data => setSelectedData(_.xor([data]))
 
   return h(FooterWrapper, { alwaysShow: true }, [
     libraryTopMatter('browse & explore'),
+    !acknowledged && div({
```
petesantos: Check on the id within the store
```suggestion
    !acknowledged[id] && div({
```

src/pages/library/DataBrowser.js
```diff
@@ -221,11 +225,30 @@ const Browser = () => {
...
+        display: 'flex', flexDirection: 'row', alignItems: 'center', justifyContent: 'space-between'
+      }
+    }, [
+      div({ style: { lineHeight: '1.4rem' } }, [
+        div(['Thank you for participating in the BETA Catalog test. The HCA data is for the sole use of the BETA testing of the Terra Data Catalog.']),
+        div(['Not for research purposes.'])
+      ]),
+      h(ButtonPrimary, {
+        style: { minWidth: 88, minHeight: 40 },
+        onClick: () => acknowledgmentStore.set(true)
```
petesantos: Set the pref with the ID
```suggestion
        onClick: () => acknowledgmentStore.set({ [id]: true })
```

petesantos (commented):
Apologies for the churn. There are some async issues with `getUser` and the `authStore` that are not capturing the user id. I added a few one liner suggestions that should correct the issue.

Ultimately, we should have a custom hook in the code to handle prefs like this.

petesantos (approved):

kyuksel (approved):
Looks like GitHub doesn't seem to have recognized one of the approvals and is showing `Merge Blocked` for some reason so adding an approval in hopes of fixing that.

lu-c:
> 

Thank you Kyuksel!

PR #2708, opened 2021-12-02
[DC-82] add metric for request access

tcjiang9:
separate metrics for opening the modal and actually requesting access

### DIFF
--- a/src/pages/library/DataBrowser.js
+++ b/src/pages/library/DataBrowser.js
@@ -202,7 +202,13 @@ const makeDataBrowserTableComponent = ({ sort, setSort, selectedData, toggleSele
               Utils.switchCase(access,
                 [snapshotAccessTypes.CONTROLLED, () => h(ButtonOutline, {
                   style: { height: 'unset', textTransform: 'none', padding: '.5rem' },
-                  onClick: () => setRequestDatasetAccessList([datum])
+                  onClick: () => {
+                    setRequestDatasetAccessList([datum])
+                    Ajax().Metrics.captureEvent(`${Events.catalogRequestAccess}:popUp`, {
+                      snapshotId: _.get('dct:identifier', datum),
+                      snapshotName: datum['dct:title']
+                    })
+                  }
                 }, [icon('lock'), div({ style: { paddingLeft: 10, fontSize: 12 } }, ['Request Access'])])],
                 [snapshotAccessTypes.PENDING, () => div({ style: { color: styles.access.pending, display: 'flex' } }, [
                   icon('lock'),
--- a/src/pages/library/DataBrowser.js
+++ b/src/pages/library/DataBrowser.js
@@ -173,7 +173,6 @@ const makeDataBrowserTableComponent = ({ sort, setSort, selectedData, toggleSele
       useHover: false,
       underRowKey: 'underRow',
       rows: _.map(datum => {
-        datum['access'] = 'Controlled'
         const { project, dataType, access } = datum
 
         return {
--- a/src/pages/library/DataBrowser.js
+++ b/src/pages/library/DataBrowser.js
@@ -173,6 +173,7 @@ const makeDataBrowserTableComponent = ({ sort, setSort, selectedData, toggleSele
       useHover: false,
       underRowKey: 'underRow',
       rows: _.map(datum => {
+        datum['access'] = 'Controlled'
         const { project, dataType, access } = datum
 
         return {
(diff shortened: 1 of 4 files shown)


src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -60,7 +61,7 @@ export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
           ]),
           td([
             Utils.switchCase(access,
-              ['Controlled', () => h(RequestDatasetAccessButton, { datasetName: title })],
+              ['Controlled', () => h(RequestDatasetAccessButton, datasets[0] )],
```
lu-c:
```suggestion
              ['Controlled', () => h(RequestDatasetAccessButton, { title, id } )],
```
This way it will change with which button the user clicks, so you can make requests for all of the datasets

src/pages/library/DataBrowserDetails.js
```diff
@@ -198,7 +204,8 @@ const DataBrowserDetails = ({ id }) => {
           h(Sidebar, { snapshot, id, setShowRequestAccessModal }),
           showRequestAccessModal && h(RequestDatasetAccessModal, {
             datasets: [snapshot],
-            onDismiss: () => setShowRequestAccessModal(false)
+            onDismiss: () => setShowRequestAccessModal(false),
+            isSnapshot: true
```
lu-c: Why did you add this field?

kyuksel:
@tcjiang9 UI PRs need to have a ticket number in the PR title for compliance reasons.

src/pages/library/DataBrowserDetails.js
tcjiang9: accident, forgot to delete

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -71,22 +72,26 @@ export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
...
-
   return h(ButtonPrimary, {
     disabled: requesting || requested,
-    onClick: requestAccess
+    onClick: () => {
+      _.flow(
+        Utils.withBusyState(setRequesting),
+        withErrorReporting('Error requesting dataset access')
+      )(async () => {
+        await Ajax(signal).DataRepo.requestAccess(dataset.id)
```
petesantos: Does this need to be feature flagged?

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -71,22 +72,26 @@ export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
   ])
 }
 
-const RequestDatasetAccessButton = ({ dataset }) => {
+const RequestDatasetAccessButton = (dataset) => {
```
petesantos: Being a react component `dataset` is essentially the list of props being passed in to this component. We destructure the props on the component. Also, using Christine's suggestion above, this could be changed to `({title, id}) => {`
I would prefer this approach in the spirit of using the least powerful abstraction / simplicity. Given the amount of data in the dataset, I would prefer only passing in the necessary information for a small component like this.

src/pages/library/DataBrowserDetails.js
lu-c: oh haha, been there :) Can we delete it?

src/pages/library/RequestDatasetAccessModal.js
tcjiang9: not sure what you mean, @lu-c ?

src/pages/library/DataBrowser.js
```diff
@@ -173,6 +173,7 @@ const makeDataBrowserTableComponent = ({ sort, setSort, selectedData, toggleSele
       useHover: false,
       underRowKey: 'underRow',
       rows: _.map(datum => {
+        datum['access'] = 'Controlled'
```
lu-c: Please remove this line! Phil has a code change to set some of the datasets to Controlled

src/pages/library/DataBrowserDetails.js
```diff
@@ -199,7 +205,7 @@ const DataBrowserDetails = ({ id }) => {
           h(Sidebar, { snapshot, id, setShowRequestAccessModal }),
           showRequestAccessModal && h(RequestDatasetAccessModal, {
             datasets: [snapshot],
-            onDismiss: () => setShowRequestAccessModal(false)
+            onDismiss: () => setShowRequestAccessModal(false),
```
lu-c: The linter is complaining about this line having a comma dangle

lu-c (approved):
approved pending changes

src/pages/library/RequestDatasetAccessModal.js
lu-c: I think pete is asking, "Does it need a "is dataCatalog enabled" check before you run this request?"

The requestAccess endpoint doesnt exist ATM, so this request will just fail. That error is being caught and showing "error requesting dataset access" to the user. I don't think we need to feature flag this, esp since access to this page should be feature flagged, but if there is something we should do, let me know!

petesantos (approved):

PR #2712, opened 2021-12-06
IA-2924 update dataproc machine type and worker disk size

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2924

With latest dataproc images, we'll need a bit more powerful machine type. `n1-standard-2` will work in a sense the runtime will be able to run, but user won't be able to do much analysis (I tried a simple hail tutorial, and it's really slow on n1-standard-2)

Tested locally
Screen Shot 2021-12-06 at 11 21 50 AM

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -26,7 +26,7 @@ export const defaultGceBootDiskSize = 100 // GCE boot disk size is not customiza
 export const defaultGcePersistentDiskSize = 50
 
 export const defaultGceMachineType = 'n1-standard-1'
-export const defaultDataprocMachineType = 'n1-standard-2'
+export const defaultDataprocMachineType = 'n1-standard-4'
 export const defaultNumDataprocWorkers = 2
 export const defaultNumDataprocPreemptibleWorkers = 0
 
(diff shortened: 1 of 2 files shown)


kyuksel (approved):
👍🏼 with a minor comment inline

src/libs/runtime-utils.js
```diff
@@ -21,12 +21,13 @@ export const computeStyles = {
   warningView: { backgroundColor: colors.warning(0.1) }
 }
 
-export const defaultDataprocDiskSize = 100 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
+export const defaultDataprocMasterDiskSize = 100 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
+export const defaultDataprocWorkerDiskSize = 150 // For both main and worker machine disks. Dataproc clusters don't have persistent disks.
```
```suggestion
// Dataproc clusters don't have persistent disks.
export const defaultDataprocMasterDiskSize = 100 
export const defaultDataprocWorkerDiskSize = 150
```

jdcanas (approved):

Qi77Qi:
thanks @kyuksel (also the extra formatting commit :) ) @jdcanas for the reviews

PR #2714, opened 2021-12-06
DC-74: For the beta test, artificially mark some snapshots as controlled access

pshapiro4broad:
All snapshots in our current test set are public, and we can't modify their permissions in TDR. This change modifies snapshots in memory so any snapshot with an even number of characters in its title will be shown as controlled access in the UI.

Once the beta test cycle is complete this change will be reverted.

### DIFF
--- a/src/pages/library/dataBrowser-utils.js
+++ b/src/pages/library/dataBrowser-utils.js
@@ -57,12 +57,6 @@ const normalizeSnapshot = snapshot => {
     _.uniqBy(_.toLower)
   )(snapshot['prov:wasGeneratedBy'])
 
-  //  For beta test: if the title length is even, pretend that this is a controlled snapshot.
-  if (snapshot['dct:title'].length % 2 === 0) {
-    snapshot.roles = ['discoverer']
-    snapshot['TerraDCAT_ap:hasDataUsePermission'] = 'TerraCore:CC'
-  }
-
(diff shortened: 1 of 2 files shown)


src/pages/library/dataBrowser-utils.js
```diff
@@ -57,6 +57,12 @@ const normalizeSnapshot = snapshot => {
     _.uniqBy(_.toLower)
   )(snapshot['prov:wasGeneratedBy'])
 
+  //  For beta test: if the title length is even, pretend that this is a controlled snapshot.
+  if (snapshot['dct:title'].length % 2 === 0) {
+    snapshot.roles = ['discoverer']
+    snapshot['TerraDCAT_ap:hasDataUsePermission'] = 'TerraCore:CC'
+  }
+
```
petesantos: We typically try to avoid mutation in the code - you could do this and put it as the first line in the function:

```suggestion
  //  For beta test: if the title length is even, pretend that this is a controlled snapshot.
  const snapshot = rawSnapshot['dct:title'].length % 2 === 0 ? {
    ...rawSnapshot,
    roles: ['discoverer'],
    'TerraDCAT_ap:hasDataUsePermission': 'TerraCore:CC'
  } : rawSnapshot

```

`rawSnapshot` will be the new name of the input parameter of the function
I really like using the title length to introduce randomness!

lu-c (approved):
Looks good to me!

petesantos (approved):

PR #2716, opened 2021-12-06
Add documentation to dynamic and static storage methods [BW-946].

cahrens:
Documentation only change to explain the difference between dynamic and static storage.

### DIFF
--- a/src/libs/browser-storage.js
+++ b/src/libs/browser-storage.js
@@ -41,7 +41,7 @@ export const getStatic = (storage, key) => {
  *
  * See also `staticStorageSlot`, which can be passed to the `useStore` hook function.
  *
- * @param storage either `localStorage` or `sessionStorage`
+ * @param storage either one of these `window` properties: `localStorage` or `sessionStorage`
  * @param key the key for storing the value
  * @param value the new JSON-serializable value to be stored, or `undefined` to remove the key.
  */
@@ -73,7 +73,7 @@ export const getDynamic = (storage, key) => {
  * Dynamic storage will be automatically deleted in the case of space overflow, deleting oldest stored
  * data first. If your storage usage is not transient, instead use `setStatic`.
  *
- * @param storage either `localStorage` or `sessionStorage`
+ * @param storage either one of these `window` properties: `localStorage` or `sessionStorage`
  * @param key the key for storing the value. Note that it will be prefixed by a value to allow separating
  * static from dynamic storage.
  * @param value the new JSON-serializable value to be stored, or `undefined` to remove the key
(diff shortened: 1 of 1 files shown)


src/libs/browser-storage.js
```diff
@@ -32,6 +32,19 @@ export const getStatic = (storage, key) => {
   return maybeParseJSON(storage.getItem(key))
 }
```
kyuksel, line 33: While at it, can we add the missing `*` to line 5 so it is `/**` instead of `/*`, which would also help IntelliJ parse and colour it appropriately?

zarsky-broad (approved):

kyuksel (approved):
Looks great to me with one nit inline. Would be good to have @petesantos's  👀  on it too.

Thanks for taking the time to make this improvement  🙏🏼

src/libs/browser-storage.js
cahrens: That comment block isn't attached to any particular method. Is `/**` correct there? I guess it would be file-level documentation.

kyuksel: I see your point. I think it is still correct but I don't feel strongly.

cahrens: OK, I made the change.

src/libs/browser-storage.js
```diff
@@ -32,6 +32,19 @@ export const getStatic = (storage, key) => {
...
+/**
+ * Sets a key/value in browser storage, or removes the key from storage if `value` is undefined.
+ *
+ * Static storage will not be automatically deleted in the case of space overflow. If your storage usage
+ * is transient (for example, navigating between routes and preserving data where the loss of it is not
+ * critical), instead use `setDynamic`.
+ *
+ * See also `staticStorageSlot`, which can be passed to the `useStore` hook function.
+ *
+ * @param storage either `localStorage` or `sessionStorage`
```
petesantos: nit: Could we call this out as something like (I leave it to your discretion if you prefer what you already have): 
```
either one of these `window` properties: `localStorage` or `sessionStorage`
```

src/libs/browser-storage.js
```diff
@@ -76,8 +100,11 @@ export const listenDynamic = (storage, key, fn) => {
 /**
  * Returns a stateful object that manages the given storage location.
  * Implements the Store interface, and can be passed to useStore.
+ *
+ * Note that this method will use static storage, meaning that it will not
+ * automatically be deleted in the case of space overflow. If the data you are using
+ * is transient, consider using `setDynamic` instead.
```
petesantos, line 105: Should we add `@param` here for the param documentation to be consistent? 
Also, I would suggest moving line 101 to the end and annotating with `@returns`
Perhaps something like:
```
@returns A stateful object that manages the given storage location.
@see atom for more {@link https://github.com/DataBiosphere/terra-ui/blob/999656a2cb159d73f0342130ccdb37215bb171c3/src/libs/utils.js#L32}
```
I'm not sure how folks feel about adding links to other parts of the code. We could leave that out.

petesantos (approved):

src/libs/browser-storage.js
cahrens: I added the `params` and the `returns`, but didn't do the `see` because I didn't want the file to need the atom import. I don't like putting in a URL link because it is sure to get out-of-date.

src/libs/browser-storage.js
cahrens: Done

PR #2717, opened 2021-12-07
[DC-44] Added preview page

lu-c:
screen_shot_2021-12-07_at_10 27 16_am
screen_shot_2021-12-07_at_10 27 10_am
>
screen_shot_2021-12-07_at_10 34 03_am

Added the initial pass of this page - I think there are some changes that may still need to happen, but I would like to make follow on tickets to address any additional features or large changes! This gets the data in, which is a nice start :)


src/pages/library/DataBrowserPreview.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+    loadData()
+  })
+
+  return h(FooterWrapper, { alwaysShow: true }, [
+    libraryTopMatter(activeTab),
+    !snapshot ?
+      centeredSpinner() :
+      div({ style: { padding: 20 } }, [
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', width: '100%', lineHeight: '26px' } }, [
```
dmohs, line 70: Implicit pixels here?
```suggestion
        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', width: '100%', lineHeight: 26 } }, [
```

src/libs/ajax.js
```diff
@@ -944,6 +944,16 @@ const DataRepo = signal => ({
     //TODO: Update this link to hit the real endpoint
     const res = await fetchRawls(`dunno/what/this/is/${id}/requestAccess`, _.merge(authOpts(), { signal }))
     return res.json()
+  },
+
+  getPreviewMetadata: async id => {
+    const res = await fetchDataRepo(`repository/v1/snapshots/${id}?include=TABLES,DATA_PROJECT`, _.merge(authOpts(), { signal }))
+    return res.json()
+  },
```
dmohs, line 952: An alternative pattern you might consider to make these flow a bit better:
```suggestion
  getPreviewMetadata: id => {
    return fetchDataRepo(`repository/v1/snapshots/${id}?include=TABLES,DATA_PROJECT`, _.merge(authOpts(), { signal })).then(res => res.json())
  },
```

src/pages/library/DataBrowserPreview.js
```diff
@@ -0,0 +1,109 @@
...
+      color: colors.accent(),
+      height: '1rem',
+      textTransform: 'uppercase', fontWeight: 600, fontSize: '0.75rem'
+    }
+  }
+}
+
+const activeTab = 'browse & explore'
+let metadata
+let tableMap
```
dmohs: These deserve explanation, either in PR comments or code comments.

src/pages/library/DataBrowserPreview.js
```diff
@@ -0,0 +1,109 @@
...
+      ])
+  ])
+}
+
+export const navPaths = [{
+  name: 'library-catalog-preview',
+  path: '/library/browser/:id/preview',
+  component: DataBrowserPreview,
+  title: ({ id }) => `Catalog - Dataset Preview`,
+  public: false
```
dmohs: I think the convention is to leave this out—not public is the norm.

src/libs/ajax.js
petesantos: I would prefer this to remain as-is, since the convention of the file is the way Christine coded it. If we want to change the convention, I would rather see a PR that goes through the whole file and changes it.

src/pages/library/DataBrowserPreview.js
petesantos: I think we can avoid the use of `let` - metadata seems to only be used in `useOnMount` and I think  `tableMap` can be set as state

src/libs/ajax.js
```diff
@@ -944,6 +944,16 @@ const DataRepo = signal => ({
     //TODO: Update this link to hit the real endpoint
     const res = await fetchRawls(`dunno/what/this/is/${id}/requestAccess`, _.merge(authOpts(), { signal }))
     return res.json()
+  },
+
+  getPreviewMetadata: async id => {
```
petesantos, line 949: Do these need to be feature flagged?

src/pages/library/DataBrowserPreview.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+}
+
+const activeTab = 'browse & explore'
+let metadata
+let tableMap
+
+const selectTable = ({ id, value, signal, setSelectedTable, setPreviewData, loading, setLoading }) => {
+  const loadTable = _.flow(
```
petesantos: Could this be moved to the `useOnMount` below? I think it may simplify the code a bit.

src/pages/library/DataBrowserPreview.js
```diff
@@ -0,0 +1,109 @@
...
+  const snapshot = dataMap[id]
+  const [tables, setTables] = useState()
+  const [selectedTable, setSelectedTable] = useState('')
+  const [previewData, setPreviewData] = useState()
+
+  Utils.useOnMount(() => {
+    const loadData = async () => {
+      metadata = await Ajax(signal).DataRepo.getPreviewMetadata(id)
+      tableMap = _.keyBy('name', metadata.tables)
+      setTables(_.map('name', metadata.tables))
```
petesantos: How would you feel about doing something like:
`setTables(metadata.tables)`

And then around line 54 you could do:
```javascript
const tableMap = _.keyBy('name', tables)
const tableNames = _.map('name', metadata.tables)
```

petesantos (commented):
I have a few comments that may help simplify the code a bit.
In a meeting now, I will take another quick pass later.

src/libs/ajax.js
lu-c: It could be a bigger change in the future if we want to do that. I also just kind of think this reads better, but that's just me!

src/pages/library/DataBrowserPreview.js
lu-c: Actually, i made this change, but I need to change it back. For whatever reason, the line-height field does not add the pixels to the value, and the view is dramatically different by excluding this

src/pages/library/DataBrowserPreview.js
lu-c: Made these updates!

src/pages/library/DataBrowserPreview.js
lu-c: i moved the whole selectTable block into the DataBrowserPreview method - the selectTable method needs to be called in the autocomplete block too, so I just want that method to be reuseable. If I'm misunderstanding, please let me know!

src/pages/library/DataBrowserPreview.js
lu-c: Ok, i like it, haha, making this change now!

src/libs/ajax.js
lu-c: According to FB, only the preview table endpoint needs to be feature flagged. The convention for the feature flagging has been to catch it in the part of the code that runs the feature (and not here in the ajax.js file). And the part of the preview page that calls the previewEndpoint is wrapped in a "withErrorReporting" method that will catch the error. Then the page itself should be hidden from view (unless they manage to type it into the search bar correctly, and even so, it should still catch the error and report it to the user)

Is there any further action that I should do?

petesantos (approved):

dmohs (approved):
Recent version is much improved. Nicely done.

PR #2720, opened 2021-12-08
[DC-91] Add ability to hide filters that have no content, and enable this for the catalog filters

pshapiro4broad:
For a Sidebar element, if `hideEmpty` is `true`, hide it if its size is 0. Enable this setting for all filters used by the catalog page.

The `keepCollapsed` property appeared to be unused so I removed it. I can add it back if it's needed.

I did look at setting the property on the parent object instead of the filter item, but couldn't get that to work.


src/pages/library/common.js
```diff
@@ -48,35 +48,22 @@ const groupByFeaturedTags = (workspaces, sidebarSections) => _.flow(
...
 )(sidebarSections)
 
-const Sidebar = ({ onSectionFilter, onTagFilter, sections, selectedSections, selectedTags, listDataByTag }) => {
-  const unionSectionWorkspacesCount = ({ tags }) => _.flow(
-    _.flatMap(tag => listDataByTag[tag]),
-    _.uniq,
-    _.size
-  )(tags)
-
+const Sidebar = ({ onTagFilter, sections, selectedTags, listDataByTag }) => {
```
lu-c: There might be some missing functionality, do we have to get rid of `onSectionFilter`, `selectedSections`? I think that might be used somewhere

pshapiro4broad: I can add them back, but they appear to only be used in the `keepCollapsed` code block, and I couldn't find any uses of `keepCollapsed` in the code. If it's really unused I should also remove the places where `onSectionFilter` and `selectedSections` is passed in.

Can we be sure something is unused by looking at the code? If not, is there a manual test I can do to see if I've broken something?

lu-c (approved):
Approved, but curious what @petesantos thinks too!

src/pages/library/DataBrowser.js
```diff
@@ -264,7 +264,7 @@ const Browser = () => {
       }, ['OK'])
     ]),
     h(SearchAndFilterComponent, {
-      fullList: dataCatalog, sidebarSections: extractCatalogFilters(dataCatalog),
+      fullList: dataCatalog, sidebarSections: _.map(filter => ({ ...filter, hideEmpty: true }), extractCatalogFilters(dataCatalog)),
```
zarsky-broad: Consider:
```suggestion
      fullList: dataCatalog, sidebarSections: _.map(_.set('hideEmpty', true), extractCatalogFilters(dataCatalog)),
```

src/pages/library/common.js
zarsky-broad: +1 to removing if unused; if the param doesn't show up in any invocations of the component, it's unused.

src/pages/library/common.js
```diff
@@ -48,35 +48,22 @@ const groupByFeaturedTags = (workspaces, sidebarSections) => _.flow(
...
+        key: name,
+        style: styles.nav.navSection,
+        buttonStyle: styles.nav.title,
+        titleFirst: true, initialOpenState: true,
+        title: h(Fragment, [name, span({ style: { marginLeft: '0.5rem', fontWeight: 400 } }, [`(${_.size(labels)})`])])
+      }, [_.map(label => {
+        const tag = _.toLower(label)
+        const size = _.size(listDataByTag[tag])
+        return (hideEmpty && size === 0) ? null :
+          h(Clickable, {
```
zarsky-broad: could you just
```js
        return !(hideEmpty && size === 0) && h(Clickable, {})
// or
        return (!hideEmpty || size > 0) && h(Clickable, {})
```

src/pages/library/common.js
pshapiro4broad: It turns out it was removed by accident. Adding it back wasn't easy because (of course) the code that was temporarily unreachable no longer works. I (or @petesantos ) will make another PR to add `keepCollapsed` back in.

src/pages/library/DataBrowser.js
pshapiro4broad: Yes, that seems like a better way. We may end up deleting this but it's good to know for next time.

src/pages/library/common.js
pshapiro4broad: Sure. I thought about it but wasn't sure if it was preferred to explicitly return ~`nil`~ `null` or not.

petesantos:
After discussing the UX with the team, we have decided to not proceed with this change for now.

PR #2721, opened 2021-12-08
[AS-944] importPFB -> importJob for listing

davidangb:
per broadinstitute/firecloud-orchestration#904, the `GET /api/workspaces/{workspaceNamespace}/{workspaceName}/importPFB` API is deprecated and replaced with `GET /api/workspaces/{workspaceNamespace}/{workspaceName}/importJob`.

This PR updates the UI to use the not-deprecated endpoint.

Do not merge this until broadinstitute/firecloud-orchestration#904 is released to prod!

Tested manually by running a local UI. The API call in question occurs when entering the Data tab of a workspace.

N.B. even though the `POST /api/workspaces/{workspaceNamespace}/{workspaceName}/importPFB` is not deprecated, it may be soon, so I've updated that ajax call as well.

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -863,7 +863,7 @@ const Workspaces = signal => ({
       },
 
       importPFB: async url => {
-        const res = await fetchOrchestration(`api/${root}/importPFB`, _.mergeAll([authOpts(), jsonBody({ url }), { signal, method: 'POST' }]))
+        const res = await fetchOrchestration(`api/${root}/importJob`, _.mergeAll([authOpts(), jsonBody({ url, filetype: 'pfb' }), { signal, method: 'POST' }]))
         return res.json()
       },
 
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -874,7 +874,7 @@ const Workspaces = signal => ({
 
       listImportJobs: async isRunning => {
         // ToDo: This endpoint should be deprecated in favor of more generic "importJob" endpoint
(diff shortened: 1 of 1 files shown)


src/libs/ajax.js
```diff
@@ -874,7 +874,7 @@ const Workspaces = signal => ({
 
       listImportJobs: async isRunning => {
         // ToDo: This endpoint should be deprecated in favor of more generic "importJob" endpoint
-        const res = await fetchOrchestration(`api/${root}/importPFB?running_only=${isRunning}`, _.merge(authOpts(), { signal }))
+        const res = await fetchOrchestration(`api/${root}/importJob?running_only=${isRunning}`, _.merge(authOpts(), { signal }))
         return res.json()
```
dmohs, line 878: If you'd like the concision, we've also been using:
```suggestion
        return fetchOrchestration(`api/${root}/importJob?running_only=${isRunning}`, _.merge(authOpts(), { signal })).then(res => res.json())
```

dmohs (approved):

davidangb:
@dmohs thank you for the thumb … I've re-requested review because I ended up changing the `POST` api as well.

I like your suggestion about concision but I've left it as-is for consistency within the file, please speak up if you feel strongly about changing that.

dmohs (approved):
Minor process note: In my judgment, this addition isn't materially different from the original, so my original approval could have stuck.

src/libs/ajax.js
zarsky-broad: We've been doing it in the two line format traditionally; I'd **really** like to keep this file consistent.

MatthewBemis (approved):

PR #2723, opened 2021-12-09
[DC-60] Correct the behavior of hitting escape on autocomplete dropdown

petesantos:
Prior to this change, when hitting the escape key the suggestions in the autocomplete field would not dismiss the drop down or clear the text. This fix adjusts the behavior as follows:

1. When the user hits the escape key it dismisses the drop down

### DIFF
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -280,7 +280,6 @@ const withAutocomplete = WrappedComponent => ({
                 e.nativeEvent.preventDownshiftDefault = true
                 e.preventDefault()
               }
-              toggleMenu()
             } else if (_.includes(e.key, ['ArrowUp', 'ArrowDown']) && !suggestions.length) {
               e.nativeEvent.preventDownshiftDefault = true
             } else if (e.key === 'Enter') {
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -248,8 +248,14 @@ const withAutocomplete = WrappedComponent => ({
     })
   })
 
+  // Unused param denoted as _u
+  const stateReducer = (_u, action) => {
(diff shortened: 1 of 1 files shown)


lu-c:
> When the user hits the escape key the second time, it clears the text from the input field

I don't think this is expected behavior for a search input, this would be a surprising part of the user experience. Can we just have the second escape do nothing?

lu-c:
code looks fine as is, haha

petesantos:
> > When the user hits the escape key the second time, it clears the text from the input field
> 
> I don't think this is expected behavior for a search input, this would be a surprising part of the user experience. Can we just have the second escape do nothing?

I updated the code to not clear the text

src/components/input.js
```diff
@@ -248,8 +248,14 @@ const withAutocomplete = WrappedComponent => ({
     })
   })
 
+  // Unused param denoted as _u
```
fboulnois: Is this comment necessary?

fboulnois (approved):
Overall lgtm, one question:

src/components/input.js
```diff
@@ -248,8 +248,14 @@ const withAutocomplete = WrappedComponent => ({
...
+  const stateReducer = (_u, action) => {
```
kyuksel:
```suggestion
  const stateReducer = (_unused, action) => {
```

kyuksel (approved):
Neat 👍🏼  One convention comment inline.

src/components/input.js
petesantos: resolved with a variable rename - thanks!

fboulnois (approved):

PR #2724, opened 2021-12-09
[DC-90] Updated beta disclaimer modal to replace original request access modal

lu-c:
Before
screen_shot_2021-12-09_at_12 53 23_pm

After
screen_shot_2021-12-09_at_12 53 10_pm

### DIFF
--- a/src/pages/library/RequestDatasetAccessModal.js
+++ b/src/pages/library/RequestDatasetAccessModal.js
@@ -9,6 +9,7 @@ import { withErrorReporting } from 'src/libs/error'
 import Events from 'src/libs/events'
 import { FormLabel } from 'src/libs/forms'
 import * as Utils from 'src/libs/utils'
+import { snapshotAccessTypes } from 'src/pages/library/dataBrowser-utils'
 
 
 const sendCopyEnabled = false
@@ -64,7 +65,7 @@ export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
             checked: sendCopy,
             onChange: setSendCopy
           }, [
-            label({ style: { margin: '0 2rem 0 0.25rem' } }, [`Also send me a copy`])
+            label({ style: { margin: '0 2rem 0 0.25rem' } }, ['Also send me a copy'])
           ])
         ])
       ])]),
@@ -82,8 +83,8 @@ export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
             ]),
             td([
               Utils.switchCase(access,
-                ['Controlled', () => h(RequestDatasetAccessButton, { title, id, setShowWipModal })],
-                ['Pending', () => span({ style: { fontWeight: 600 } }, ['Request Pending'])],
+                [snapshotAccessTypes.CONTROLLED, () => h(RequestDatasetAccessButton, { title, id, setShowWipModal })],
+                [snapshotAccessTypes.PENDING, () => span({ style: { fontWeight: 600 } }, ['Request Pending'])],
                 [Utils.DEFAULT, () => span({ style: { fontWeight: 600 } }, ['Permission Granted'])]
               )
             ])
--- a/src/pages/library/RequestDatasetAccessModal.js
+++ b/src/pages/library/RequestDatasetAccessModal.js
@@ -6,6 +6,7 @@ import { TextArea } from 'src/components/input'
 import Modal from 'src/components/Modal'
 import { Ajax } from 'src/libs/ajax'
 import { withErrorReporting } from 'src/libs/error'
+import Events from 'src/libs/events'
 import { FormLabel } from 'src/libs/forms'
 import * as Utils from 'src/libs/utils'
 
@@ -61,7 +62,7 @@ export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
           ]),
           td([
             Utils.switchCase(access,
-              ['Controlled', () => h(RequestDatasetAccessButton, { datasetName: title })],
+              ['Controlled', () => h(RequestDatasetAccessButton, { title, id, onDismiss })],
               ['Pending', () => span({ style: { fontWeight: 600 } }, ['Request Pending'])],
               [Utils.DEFAULT, () => span({ style: { fontWeight: 600 } }, ['Permission Granted'])]
             )
@@ -72,20 +73,11 @@ export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
   ])
 }
 
-const RequestDatasetAccessButton = ({ dataset }) => {
+const RequestDatasetAccessButton = ({ title, id, onDismiss }) => {
   const [requesting, setRequesting] = useState(false)
   const [requested, setRequested] = useState(false)
   const [showWipModal, setShowWipModal] = useState(false)
   const signal = Utils.useCancellation()
-
-  const requestAccess = _.flow(
-    Utils.withBusyState(setRequesting),
-    withErrorReporting('Error requesting dataset access')
-  )(async () => {
-    requestAccessEnabled && await Ajax(signal).DataRepo.requestAccess(dataset.id)
-    setRequested(true)
-    setShowWipModal(true)
-  })
 
   return h(Fragment, [
     showWipModal && h(Modal, {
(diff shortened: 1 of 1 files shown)


src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -92,31 +33,96 @@ const RequestDatasetAccessButton = ({ title, id }) => {
...
+
+const RequestDatasetAccessButton = ({ title, id, setShowWipModal }) => {
+  const [requesting, setRequesting] = useState(false)
+  const [requested, setRequested] = useState(false)
+  const signal = Utils.useCancellation()
+
+  return h(ButtonPrimary, {
+    disabled: requesting || requested,
+    onClick: () => {
+      withErrorReporting('Error requesting dataset access', async () => {
```
zarsky-broad: You can promote this a level:
```js
onClick: withErrorReporting('error', async () => {})
```

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -12,71 +12,12 @@ import * as Utils from 'src/libs/utils'
...
-  ])
-}
-
-const RequestDatasetAccessButton = ({ title, id }) => {
-  const [requesting, setRequesting] = useState(false)
-  const [requested, setRequested] = useState(false)
   const [showWipModal, setShowWipModal] = useState(false)
-  const signal = Utils.useCancellation()
 
   return h(Fragment, [
```
zarsky-broad: The traditional method to do this would be, instead of having both of these modals in a fragment, to either write it as a ternary. I also wouldn't expect to have to call `setShowWipModal(false)` as well as `onDismiss`, since the latter should unmount the entire component.

src/pages/library/RequestDatasetAccessModal.js
lu-c: Updated!

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -92,31 +33,96 @@ const RequestDatasetAccessButton = ({ title, id }) => {
...
+          ]), datasets)
+        )
+      ])
+    ])
+  ])
+}
+
+const RequestDatasetAccessButton = ({ title, id, setShowWipModal }) => {
+  const [requesting, setRequesting] = useState(false)
+  const [requested, setRequested] = useState(false)
```
petesantos: How would you feel about either merging these into one state, where it switches the value. The values can be stored in an enum. Having two mutually exclusive states opens the possibility for the code to get into what should be an impossible state. Or we could remove the `requesting` state for now, it looks like it is not used.

src/pages/library/RequestDatasetAccessModal.js
lu-c: good point, thank you!

src/pages/library/RequestDatasetAccessModal.js
lu-c: That's a great point! Let me take a look

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -12,111 +12,104 @@ import * as Utils from 'src/libs/utils'
...
-        () => 'Request Access'
-      )
-    ])
+  return h(ButtonPrimary, {
+    disabled: status,
+    onClick: withErrorReporting('Error requesting dataset access', async () => {
+      requestAccessEnabled ?
+        await Ajax(signal).DataRepo.requestAccess(id) :
+        setShowWipModal(true)
+      setStatus('Request Sent')
```
petesantos, line 107: nit: Now that I am seeing it this way, I am thinking a better name for this may just be `buttonText`, with a default of 'Request Access'

petesantos (approved):
LGTM. One minor nit on the state naming

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -12,111 +12,104 @@ import * as Utils from 'src/libs/utils'
...
 export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
   const [reason, setReason] = useState('')
   const [sendCopy, setSendCopy] = useState(false)
+  const [showWipModal, setShowWipModal] = useState(false)
 
-  return h(Modal, {
-    title: 'Request Access',
-    width: '40rem',
+  return showWipModal ? h(Modal, {
+    title: `Keep in mind we're still in BETA`,
```
nmalfroy, line 24: Any reason to use a template string here?

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -12,111 +12,104 @@ import * as Utils from 'src/libs/utils'
...
-        }, [
-          label({ style: { margin: '0 2rem 0 0.25rem' } }, [`Also send me a copy`])
-        ])
-      ])
-    ])]),
-    table({ style: { margin: '1rem', width: '100%' } }, [
-      thead([
-        tr({ style: { height: '2rem' } }, [th({ style: { textAlign: 'left' } }, ['Datasets']), th({ style: { textAlign: 'left', width: '15rem' } }, ['Access'])])
+    div({ style: { marginTop: 30, display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems: 'center' } }, [
+      `Let's continue testing the BETA Data Catalog!`,
```
nmalfroy, line 35: same with the backticks here

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -12,111 +12,104 @@ import * as Utils from 'src/libs/utils'
...
+  ]) :
+    h(Modal, {
+      title: 'Request Access',
+      width: '40rem',
+      showButtons: false,
+      showX: true,
+      onDismiss
+    }, [
+      div([
+        'You cannot access this dataset because it is protected by an Authorization Domain. ',
```
nmalfroy, line 50: I'm not super familiar with the hyperscript helpers lib but do we not need indexes for child elements?  Wondering if maybe it adds them when the children are text

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -12,111 +12,104 @@ import * as Utils from 'src/libs/utils'
...
+          placeholder: 'Enter your reason',
+          value: reason,
+          onChange: setReason
+        }),
+        sendCopyEnabled && div([
+          h(LabeledCheckbox, {
+            checked: sendCopy,
+            onChange: setSendCopy
+          }, [
+            label({ style: { margin: '0 2rem 0 0.25rem' } }, [`Also send me a copy`])
```
nmalfroy: backticks...not sure if it's a problem...figured there may be a suuuper tiny perf gain

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -12,111 +12,104 @@ import * as Utils from 'src/libs/utils'
...
+          _.map(({ 'dct:title': title, access, 'dct:identifier': id, contacts }) => tr({ key: id, style: { height: '2rem' } }, [
+            td({ style: { paddingRight: 20 } }, [
+              title,
+              div({ style: { fontSize: '.7rem', marginTop: 5, width: 'fit-content' } }, [
+                _.map(({ email }) => email && h(Link, { key: email, href: `mailto:${email}`, style: { marginTop: 5, display: 'block' } }, [email]), contacts)
+              ])
+            ]),
+            td([
+              Utils.switchCase(access,
+                ['Controlled', () => h(RequestDatasetAccessButton, { title, id, setShowWipModal })],
```
nmalfroy: Shouldn't these strings be constants?

nmalfroy (approved):
Looks good from where I stand but I'm not sure how much my approval means in this repo.  Couple minor things

src/pages/library/RequestDatasetAccessModal.js
lu-c: no, i can change them back! :o
EDIT: Wait, it's because of the ' in "we're"! I'd rather backtick those than escape them!

src/pages/library/RequestDatasetAccessModal.js
lu-c: We have to uniquely key them if the object is part of a list, but otherwise, it's okay!

src/pages/library/RequestDatasetAccessModal.js
lu-c: i'll change this one, but please let me know what you think about the backticks in the other sections!

src/pages/library/RequestDatasetAccessModal.js
lu-c: nice catch!

PR #2727, opened 2021-12-10
[IA-3128] Display total disk cost estimate accurately on Cloud Environments page

kyuksel:
### Before

unknown_total

### After

image

### DIFF
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -17,9 +17,8 @@ import { getUser } from 'src/libs/auth'
 import colors from 'src/libs/colors'
 import { withErrorIgnoring, withErrorReporting } from 'src/libs/error'
 import {
-  defaultComputeRegion,
   defaultComputeZone, getComputeStatusForDisplay, getCurrentApp, getCurrentRuntime, getDiskAppType, getGalaxyComputeCost, getGalaxyCost,
-  getPersistentDiskCostMonthly, isApp, isComputePausable, isResourceDeletable, runtimeCost
+  getPersistentDiskCostMonthly, getRegionFromZone, isApp, isComputePausable, isResourceDeletable, runtimeCost
 } from 'src/libs/runtime-utils'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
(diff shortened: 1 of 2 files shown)


src/pages/Environments.js
```diff
@@ -196,7 +195,7 @@ const Environments = () => {
   const totalRuntimeCost = _.sum(_.map(runtimeCost, runtimes))
   const totalAppCost = _.sum(_.map(getGalaxyComputeCost, apps))
   const totalCost = totalRuntimeCost + totalAppCost
-  const totalDiskCost = _.sum(_.map(getPersistentDiskCostMonthly, disks))
```
kyuksel, line 199: We were not passing in the second (`region`) parameter into `getPersistentDiskCostMonthly` function.

wnojopra (approved):
Thanks! Sorry I missed this. Tested locally and LGTM
image

Qi77Qi (approved):

src/pages/Environments.js
```diff
@@ -196,7 +195,7 @@ const Environments = () => {
...
+  const totalDiskCost = _.sum(_.map(disk => getPersistentDiskCostMonthly(disk, getRegionFromZone(disk.zone)), disks))
```
petesantos, line 198: Out of scope for this PR - but the `getPersistentDiskCostMonthly` and `getRegionFromZone` is triggering a code smell for me since both of these functions are working off of data from `disk`. I think we can refactor some of these utility functions to work a bit better with composition. This may be a good pairing opportunity

src/pages/Environments.js
```diff
@@ -503,7 +502,9 @@ const Environments = () => {
               return h(Sortable, { sort: diskSort, field: 'cost', onSort: setDiskSort }, [`Cost / month (${Utils.formatUSD(totalDiskCost)} total)`])
             },
             cellRenderer: ({ rowIndex }) => {
-              return Utils.formatUSD(getPersistentDiskCostMonthly(filteredDisks[rowIndex], defaultComputeRegion))
+              const disk = filteredDisks[rowIndex]
+              const diskRegion = getRegionFromZone(disk.zone)
```
petesantos, line 506: Is there always a region? Do we need a default?

petesantos (approved):

cahrens (approved):

src/pages/Environments.js
kyuksel: Every `disk` should have a `zone` and every `zone` is in a `region` so yes. The bigger assumption is there is a `disk` at `filteredDisks[rowIndex]`, which is consistent with the rest of the file. If I'm missing something, let me know.

src/pages/Environments.js
kyuksel: Sounds good and thank you for the offer. Added to rolling agenda :)

PR #2728, opened 2021-12-13
[DC-89] added column selector

lu-c:
Added column selector to the simple table so you can select and reorder columns.


fboulnois (approved):
this is super awesome!

petesantos (approved):
LGTM

PR #2729, opened 2021-12-14
[DC-94] Added "view json" drawer view on the preview data

lu-c:
Screen Shot 2021-12-14 at 6 15 21 PM
Screen Shot 2021-12-14 at 6 15 27 PM

I might want to change how the drawers work (or add config settings), so you can continue interacting with the page behind it.

### DIFF
--- a/src/pages/library/DataBrowserPreview.js
+++ b/src/pages/library/DataBrowserPreview.js
@@ -1,10 +1,11 @@
 import _ from 'lodash/fp'
 import { useState } from 'react'
 import { div, h, h1 } from 'react-hyperscript-helpers'
-import { Select } from 'src/components/common'
+import { ButtonPrimary, Select } from 'src/components/common'
 import FooterWrapper from 'src/components/FooterWrapper'
 import { centeredSpinner, spinner } from 'src/components/icons'
 import { libraryTopMatter } from 'src/components/library-common'
+import ModalDrawer from 'src/components/ModalDrawer'
 import { ColumnSelector, SimpleTable } from 'src/components/table'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
@@ -35,6 +36,7 @@ const DataBrowserPreview = ({ id }) => {
   const [selectedTable, setSelectedTable] = useState('')
   const [previewData, setPreviewData] = useState()
   const [columnSettings, setColumnSettings] = useState([])
+  const [viewJSON, setViewJSON] = useState()
 
   const tableMap = _.keyBy('name', tables)
   const tableNames = _.map('name', tables)
@@ -49,13 +51,26 @@ const DataBrowserPreview = ({ id }) => {
         id, limit: 50, offset: 0,
         table: value
       })
-      setPreviewData(previewTableData?.result || [])
+
+      const columnNames = _.map('name', tableMap[selectedTable]?.columns)
+
+      setPreviewData(_.flow([
+        _.getOr([], 'result'),
+        _.map(row => {
+          return _.reduce((obj, param) => {
+            obj[param] = formatTableCell(row[param])
+            return obj
+          }, {}, columnNames)
+        })
+      ])(previewTableData))
 
       setColumnSettings(
         _.flow([
           _.toPairs,
           _.map(([index, col]) => {
             return {
+              // name field is used in the column selector
+              // key field is used in the Simple Table
               name: col.name, key: col.name,
               visible: index < 6,
               header: div({ style: styles.table.header }, [col.name])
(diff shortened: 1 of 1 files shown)


src/pages/library/DataBrowserPreview.js
```diff
@@ -108,7 +138,18 @@ const DataBrowserPreview = ({ id }) => {
           columnSettings.length > 0 && h(ColumnSelector, { onSave: setColumnSettings, columnSettings })
         ]),
         previewData?.length === 0 && div({ style: { width: '100%', height: '100%', display: 'flex', alignItems: 'center', justifyContent: 'center' } }, ['(No Data)'])
-      ])
+      ]),
+    viewJSON && h(ModalDrawer, {
```
petesantos, line 148: Could we add a title to the drawer? Perhaps something like "Viewing JSON for version _verionID_"

src/pages/library/DataBrowserPreview.js
```diff
@@ -67,6 +82,21 @@ const DataBrowserPreview = ({ id }) => {
     loadTable()
   }
 
+  const formatTableCell = cellContent => {
+    return Utils.cond(
+      [!Utils.cantBeNumber(cellContent), () => cellContent],
+      [Utils.maybeParseJSON(cellContent), () => {
+        const contentAsJSON = Utils.maybeParseJSON(cellContent)
+        const contentAsPrettyString = JSON.stringify(contentAsJSON, null, 4)
```
petesantos: We have a library - `react-json-view` that does a nice job of prettifying JSON for display. Here is an example with some of our data:
image

petesantos (commented):
Looks good! A few display nits. 
Could you run this by @lstrano-broad to get his feedback?

src/pages/library/DataBrowserPreview.js
lu-c: wow i love how that looks, oky, let me  update it to this!

src/pages/library/DataBrowserPreview.js
lu-c: I added a title that is `(tablename) row (row#) - (column name)` - im just not sure that version will be a field available for all data types! Please let me know if you think this is okay!

petesantos: That works. Based on our conversation with Lou I think we will be iterating on this. But having a title is a good cue for the user to let them know which JSON they are looking at. Thanks!

petesantos (approved):

src/pages/library/DataBrowserPreview.js
```diff
@@ -49,6 +50,19 @@ const DataBrowserPreview = ({ id }) => {
...
       })
       setPreviewData(previewTableData?.result || [])
+
+      setColumnSettings(
+        _.flow([
+          _.toPairs,
+          _.map(([index, col]) => {
+            return {
+              name: col.name, key: col.name,
+              visible: index < 6,
```
nmalfroy, line 77: I'm curious about the 6 index.  This feels like it could break if more columns are added early on.  It's probably worth a comment or maybe look up the column where things start showing up (assuming that the 6th column is meaningful).  Also, what happens if a user re-orders columns?  Just not sure what the rational is for hiding all columns that are at higher indexes

src/pages/library/DataBrowserPreview.js
```diff
@@ -67,6 +82,21 @@ const DataBrowserPreview = ({ id }) => {
     loadTable()
   }
 
+  const formatTableCell = cellContent => {
+    return Utils.cond(
+      [!Utils.cantBeNumber(cellContent), () => cellContent],
+      [Utils.maybeParseJSON(cellContent), () => {
```
nmalfroy: I wonder about this approach.  this may be a little over-eager.  For instance, `"test"` would pass as a json string (since it's quoted).  I do also wonder about a potential perf impact of doing this in the cell since you could have a table full of big json fields.  Might make sense to do it lazily

nmalfroy (approved):
I think this should work but I had a couple of question on perf. If behavior/perf is ok, I think it'd be cool to merge.  I also wonder about the column visibility thing

src/pages/library/DataBrowserPreview.js
lu-c: I picked a random index, just because it would've looked weird with more columns than that, the table gets really squished! There is a column selector that allows the user to reorder the columns and choose the visibility as well, so the 6 column limit is an initial setting, totally adjustable. Since the visibility field is tied with the columns, if they reorder them, they wont disappear (if the new index  > 6)

nmalfroy: Ah ok, I gotcha

src/pages/library/DataBrowserPreview.js
lu-c: Running `JSON.parse("test")` will fail - I'm trying to catch cases where they would successfully render as a smaller field above it ('cant be number'), because `JSON.parse(123)` will succeed... I can add extra cases as necessary!

I was a little concerned about the perf after it was brought up today in stand up, but then, I'm not sure how to lazily test for json - We would have to run JSON.parse for all fields to test them to determine if a json button should be shown, right? We never run the jsonparsing again after that. If there's a better way to test if a variable is a json object, we could try that, but from what I found on stackoverflow, it seems like parsing is the best way

nmalfroy: Oh, I was thinking about `JSON.parse('"test"')`

That being said, it's not a huge deal.  I think to do it lazily would require a different ux.  Maybe it's something like a context menu or something that happens on hover.

src/pages/library/DataBrowserPreview.js
kyuksel: It'd be great to avoid [magic numbers](https://codeburst.io/software-anti-patterns-magic-numbers-7bc484f40544) like `6` above :)

PR #2730, opened 2021-12-15
[SATURN-1909] Free-form text description fields in workspace data UI

johntbates:
[SATURN-1909] Free-form text description fields in workspace data UI (#2730)

[Link to Terra Feature Request](https://support.terra.bio/hc/en-us/community/posts/4411817354011-FR-Add-a-description-field-to-the-Workspace-data-tab)

* Descriptions for workspace attribute 'key' are stored in a new key: '__DESCRIPTION__key'.
* Descriptions are free-form text, limited to 200 characters.
* Verified that key-value pairs with descriptions can be imported and exported as TSV files.
* No server-side modification is required, and existing code that depends on workspace attributes should be unaffected.

Testing
* In-browser verification, including download and re-upload of workspace data TSVs.
* Unit tests.

image
image

[SATURN-1909]: https://broadworkbench.atlassian.net/browse/SATURN-1909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kyuksel:
Hi @johntbates, thank you for your PR. UI PRs are supposed to have a JIRA ticket for compliance reasons. Could you create one and prefix the PR title with it (e.g. `[PF-123] Free-form text…`)?

[PF-123]: https://broadworkbench.atlassian.net/browse/PF-123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/components/data/LocalVariablesContent.js
```diff
@@ -47,25 +48,60 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
...
+        description: v
+      }
+    }
+    return { key: k, value: v }
+  }
+
+  function toDisplayedAttribute(arr) {
+    const obj = Object.assign({}, ...arr)
+    return [obj.key, obj.value, obj.description]
+  }
```
kyuksel: We use [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#comparing_traditional_functions_to_arrow_functions) instead of traditional functions. Could you please convert them for consistency?

kyuksel: Nit: We typically use `verb`s when naming functions (e.g. `getDisplayedAttribute` instead of `toDisplayedAttribute`).

src/components/data/LocalVariablesContent.js
```diff
@@ -179,7 +225,29 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
                     value: editType,
                     onChange: ({ value }) => setEditType(value),
                     options: ['string', 'number', 'boolean', 'string list', 'number list', 'boolean list']
-                  }),
+                  })
+                ]) :
+                div({ className: 'hover-only' }, [])
```
cahrens: What is this for?

johntbates: There were warnings about a dangling comma on that line in the Chrome JavaScript console and removing that comma silenced those warnings.

src/components/data/LocalVariablesContent.js
johntbates: Done

johntbates:
Created a JIRA ticket:
[SATURN-1909 Add a free-form text Description field to the Workspace Data UI](https://broadworkbench.atlassian.net/browse/SATURN-1909)

wnojopra (changes requested):
Awesome!

1) I get an error whenever I try to edit an attribute that does not contain a description. `TypeError: Cannot read properties of undefined (reading 'length')`

2) Can you please include a screenshot to the description of what changes? I tend to use snippets and draw arrows to the relevant parts. I find it helps reviewers know what to look for in general.

src/components/data/LocalVariablesContent.js
```diff
@@ -47,25 +48,57 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
...
+      {
+        key: k.substring(0, k.length - DESCRIPTION_SUFFIX.length),
+        description: v
+      } : {
+        key: k,
+        value: v
+      }
+  }
+
+  const getDisplayedAttribute = arr => {
```
wnojopra: This function might be a good candidate to write a unit test for. See the unit test files in [terra-ui/src/_tests/](https://github.com/DataBiosphere/terra-ui/tree/dev/src/_tests). Use `yarn test` in the _tests directory to run them.

src/components/data/LocalVariablesContent.js
cahrens: I meant line 227 (the empty div).

src/components/data/LocalVariablesContent.js
johntbates: Done - wrote unit tests for this function and renameAttribute.  Also fixed the error you encountered and added screenshots to the PR description.

johntbates: Also fixed the error you encountered.

src/components/data/LocalVariablesContent.js
```diff
@@ -17,6 +17,27 @@ import { withErrorReporting } from 'src/libs/error'
...
+  const obj = Object.assign({}, ...arr)
+  return [obj.key, obj.value, obj.description ? obj.description : '']
+}
+
+const DESCRIPTION_SUFFIX = '__DESCRIPTION__'
+const isDescriptionKey = k => k.endsWith(DESCRIPTION_SUFFIX)
+
+// Exported for testing
+export const renameAttribute = attr => {
+  const [k, v] = attr
```
kyuksel: We usually use object/array deconstructing on input arguments for

- concision
- explicitly pulling out only what's needed
- making it easier to assign default arguments when necessary

```suggestion
export const renameAttribute = ([k, v]) => {
```

src/components/data/LocalVariablesContent.js
johntbates: Ah - got it.  I removed the empty div and replaced it with a call to renderDataCell, as is done for the 'Key' column.

src/components/data/LocalVariablesContent.js
johntbates: Done - thanks.

src/components/data/LocalVariablesContent.js
```diff
@@ -85,9 +115,22 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
       Utils.convertValue(newBaseType, editValue)
 
     await Ajax().Workspaces.workspace(namespace, name).shallowMergeNewAttributes({ [editKey]: parsedValue })
+    const originalDescription = attributes[toDescriptionKey(originalKey)]
+    if (originalDescription || editDescription) {
```
wnojopra: I like having comments when there's a few if statements in a block. Something like "When editing description, do X. When deleting a description, do Y." This becomes helpful when readers are figuring out what actions enter which if blocks. 

Btw, can we minimize the number of ajax calls we're making? For example, I feel like it should be possible to edit the key and description of a row in just one ajax call. Am I wrong about that?

wnojopra (changes requested):
I'm seeing two issues:

1) If I attempt to edit the key and description at the same time, it appears that only the key actually gets updated.
2) The value seems to be appearing an extra time. In this example, I create a new row all 3 of Key, Value, and Description equal to 'a'. For some reason 'a' appears twice in the Value row.
image

src/components/data/LocalVariablesContent.js
johntbates: * Fixed the duplicate value issue - turns out I should have returned null instead of attempting to renderCellContents.
* Totally restructured the saveAttribute logic, adding comments where applicable - your key-only update issue should be resolved.
* We now make only two Ajax calls - one to delete and one to merge attributes.

kyuksel:
@johntbates I took the liberty to push a commit to fix linting issues to not burden you with getting set up with linter and reformatter. If you're planning on making more PRs, please refer to the section `Code style and linting` in [Terra UI README](https://github.com/DataBiosphere/terra-ui).

src/components/data/LocalVariablesContent.js
```diff
@@ -17,6 +17,24 @@ import { withErrorReporting } from 'src/libs/error'
 import * as StateHistory from 'src/libs/state-history'
 import * as Utils from 'src/libs/utils'
 
+// Exported for testing
```
kyuksel: Appreciate the clarification. I don't think we need it though since IDEs make it easy to see why something is exported plus we don't comment on other exported `const`s.

```suggestion
```

src/components/data/LocalVariablesContent.js
```diff
@@ -17,6 +17,24 @@ import { withErrorReporting } from 'src/libs/error'
...
+export const getDisplayedAttribute = arr => {
+  const obj = Object.assign({}, ...arr)
+  return [obj.key, obj.value, obj.description ? obj.description : '']
+}
+
+const DESCRIPTION_SUFFIX = '__DESCRIPTION__'
+const isDescriptionKey = k => k.endsWith(DESCRIPTION_SUFFIX)
+
+// Exported for testing
```
kyuksel:
```suggestion
```

johntbates:
Thanks very much for the lint and pointers to the linter.  Apologies for just diving in - this is my first time writing any JS to speak of.

src/components/data/LocalVariablesContent.js
```diff
@@ -17,6 +17,24 @@ import { withErrorReporting } from 'src/libs/error'
 import * as StateHistory from 'src/libs/state-history'
 import * as Utils from 'src/libs/utils'
 
+// Exported for testing
+export const getDisplayedAttribute = arr => {
+  const obj = Object.assign({}, ...arr)
+  return [obj.key, obj.value, obj.description ? obj.description : '']
+}
```
kyuksel: We try to
- not mutate `const`s whenever possible
- use [Lodash FP](https://lodash.com/docs/4.17.15) methods whenever possible since they tend to handle corner cases better than native JS counterparts, and have desirable features like being auto-curried
- explicitly pull out only what's necessary

```suggestion
export const getDisplayedAttribute = arr => {
  const { key, value, description } = _.assign(...arr, {})
  return [key, value, description ?? '']
}
```

I like `description ?? ''` because it's nicely concise but if you feel it's too cryptique, I'm fine with the original (`description ? description : ''`)

kyuksel (commented):
Thanks for addressing the comments from the previous round.
A few more comments inline.
I ❤️  the tests. Thank you very much for taking the time to write them 🙏🏼

src/components/data/LocalVariablesContent.js
```diff
@@ -80,17 +110,34 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
...
+    const editDescriptionKey = toDescriptionKey(editKey)
+    if (editDescription) {
+      // New description or change to existing one =>
+      // merge editDescription
+      attributesToMerge[editDescriptionKey] = editDescription
+    } else {
+      // Change to description resulting in empty description =>
+      // delete description attribute
+      attributesToDelete.push(editDescriptionKey)
+    }
```
kyuksel: I wonder if we can avoid the mutations above via something like

```javascript
    const editDescriptionKey = toDescriptionKey(editKey)

    const attributesToMerge = {
      editKey: parsedValue,
      ...(!!editDescription && { editDescriptionKey: editDescription })
    }

    const attributesToDelete = _.flatten(
      editKey !== originalKey && [originalKey, toDescriptionKey(originalKey)],
      !editDescription && [editDescriptionKey]
    )
```

johntbates: Almost - turns out we needed something like this:

```javascript
    const editDescriptionKey = toDescriptionKey(editKey)

    const attributesToMerge = {
      [editKey]: parsedValue,
      ...(!!editDescription && { [editDescriptionKey]: editDescription })
    }

    const attributesToDelete = _.compact(_.flatten([
      editKey !== originalKey && [originalKey, toDescriptionKey(originalKey)],
      !editDescription && [editDescriptionKey]
    ]
    ))
```

src/components/data/LocalVariablesContent.js
johntbates: Didn't realize destructuring could work here - neat.  Also, the `??` operator definitely makes things more readable.

src/components/data/LocalVariablesContent.js
johntbates: Done.

src/components/data/LocalVariablesContent.js
johntbates: Done

johntbates:
No worries!  Tests keep me sane(-ish.)

src/components/data/LocalVariablesContent.js
kyuksel: Why does `attributesToMerge` need to have an array key; i.e. `[editKey]` instead of `editKey`?

kyuksel:
I've observed two behaviours with the `Description` text of length around 200 where

1. I can't get the mouse pointer at the end of the text (to the right of the last `8` on the screenshot below).
2. I _can_ move my mouse pointer to right before that `8`. However, I can't then use my keyboard's `right arrow` to move it to the end. 

In other words there doesn't seem to be a way to move the pointer to the end of a text of length over 200.

image

src/components/data/LocalVariablesContent.js
johntbates: It's not actually an array key - it's a computed property name (see details [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer) in the "Computed property names" section.) Without it, the name of the key would literally be "editKey", as seen below:

image

The same applies to `editDescriptionKey`.

johntbates:
I've not been able to reproduce either of the behaviors you describe in your last comment.  Could you please paste the strings you're using for the key, value, and description into the conversation?

Also, FWIW, there doesn't appear to be a limit on key or value length. I wonder if there should be one, as this might be a possible attack surface (overflow attack, etc.)

src/components/data/LocalVariablesContent.js
kyuksel: Ah I see. Seems like a valid use of array key but since I haven't seen it in use much, I'll tag @zarsky-broad in case there's a more preferable way to accomplish the goal.

kyuksel:
> I've not been able to reproduce either of the behaviors you describe in your last comment. Could you please paste the strings you're using for the key, value, and description into the conversation?

Sure. Here they are:

image

The string was just a string of length `200`: `01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789`

**Edit:**
While writing the message above, I've realized that the behaviour is not unique to `Description` and occurs for both `Key` and `Value` fields. When I re-size my browser window, it goes away in some cases. Given that, addressing it is not a blocker for this PR.

kyuksel:
> Also, FWIW, there doesn't appear to be a limit on key or value length. I wonder if there should be one, as this might be a possible attack surface (overflow attack, etc.)

It makes sense to me to have one but I lack context on the development of that page. I wonder if @zarsky-broad can chime in.

johntbates:
This appears to be a Mac/Linux discrepancy.

I wrote and tested this PR on a Google Cloudtop instance running Ubuntu where I've not been able to reproduce the bug.

I then installed node and yarn on my MacBook Pro, and I've been able to reproduce the issue as described.

My observations:

* This issue manifests itself with any string that is as long as or longer than the description edit field, well below the 200 character limit.
* The cursor disappears from the screen, but it still behaves as though it is there: you can delete a character immediately to its left and add new characters to its right.
* "key"s and "value"s longer than their cell's viewable width are truncated, with ellipses added at the end, but this does not happen with descriptions.
* After clicking the edit icon on a row with a long "key", the cursor is immediately placed at the end of the key.  This is not the case for either the "value" or "description".

I'll make the behaviors of the "value" and "description" fields consistent with the "key" field, as the "key" field behavior seems correct.

src/components/data/LocalVariablesContent.js
```diff
@@ -18,13 +18,31 @@ import * as StateHistory from 'src/libs/state-history'
 import * as Utils from 'src/libs/utils'
 
 
+export const getDisplayedAttribute = arr => {
+  const { key, value, description } = _.assign(...arr, {})
+  return [key, value, description ?? '']
+}
```
zarsky-broad: Consider:
```suggestion
export const getDisplayedAttribute = arr => {
  const { key, value, description = '' } = _.mergeAll(arr)
  return [key, value, description]
}
```

src/components/data/LocalVariablesContent.js
```diff
@@ -18,13 +18,31 @@ import * as StateHistory from 'src/libs/state-history'
...
+
+const DESCRIPTION_SUFFIX = '__DESCRIPTION__'
```
zarsky-broad: Historically we've done this sort of UI feature overlaying trickery with prefixes, so if someone were to download the data directly from the API it would be obvious; would you consider prefixing instead of suffixing? (see `referenceData_`) Obviously, you'd then have to flip the `endsWith` type things to `startsWith` type things, but that should be pretty straightforward, I think?

src/components/data/LocalVariablesContent.js
```diff
@@ -18,13 +18,31 @@ import * as StateHistory from 'src/libs/state-history'
...
+const isDescriptionKey = k => k.endsWith(DESCRIPTION_SUFFIX)
```
zarsky-broad: you could let the auto-currying do the work for you:
```suggestion
const isDescriptionKey = _.endsWith(DESCRIPTION_SUFFIX)
```

src/components/data/LocalVariablesContent.js
```diff
@@ -18,13 +18,31 @@ import * as StateHistory from 'src/libs/state-history'
...
+
+export const renameAttribute = ([k, v]) => isDescriptionKey(k) ?
+  {
+    key: k.substring(0, k.length - DESCRIPTION_SUFFIX.length),
+    description: v
+  } : {
+    key: k,
+    value: v
+  }
```
zarsky-broad: Perhaps a bit easier?
```suggestion
export const renameAttribute = ([k, v]) => isDescriptionKey(k) ?
  { key: k.slice(0, -DESCRIPTION_SUFFIX.length), description: v } :
  { key: k, value: v }
```

src/components/data/LocalVariablesContent.js
```diff
@@ -47,25 +65,36 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
...
+    arr => [...arr, ...(creatingNewVariable ? [['', '', '']] : [])]
   )(initialAttributes)
 
+  const DESCRIPTION_MAX_LENGTH = 200
   const inputErrors = editIndex !== undefined && [
     ...(_.keys(_.unset(amendedAttributes[editIndex][0], attributes)).includes(editKey) ? ['Key must be unique'] : []),
     ...(!/^[\w-]*$/.test(editKey) ? ['Key can only contain letters, numbers, underscores, and dashes'] : []),
     ...(editKey === 'description' ? ['Key cannot be \'description\''] : []),
+    ...(isDescriptionKey(editKey) ? [`Key cannot end with '${DESCRIPTION_SUFFIX}'`] : []),
+    ...(editDescription && editDescription.length > DESCRIPTION_MAX_LENGTH ? [`Description cannot be longer than ${DESCRIPTION_MAX_LENGTH} characters`] : []),
```
zarsky-broad:
```suggestion
    ...(editDescription?.length > DESCRIPTION_MAX_LENGTH ? [`Description cannot be longer than ${DESCRIPTION_MAX_LENGTH} characters`] : []),
```

src/components/data/LocalVariablesContent.js
```diff
@@ -80,17 +109,26 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
...
+    const attributesToMerge = {
+      [editKey]: parsedValue,
+      ...(!!editDescription && { [editDescriptionKey]: editDescription })
     }
 
+    const attributesToDelete = _.compact(_.flatten([
+      editKey !== originalKey && [originalKey, toDescriptionKey(originalKey)],
+      !editDescription && [editDescriptionKey]
+    ]
+    ))
```
zarsky-broad: Without the compacting or flattening:
```suggestion
    const attributesToDelete = [
      ...(editKey !== originalKey ? [originalKey, toDescriptionKey(originalKey)] : []),
      ...(!editDescription ? [editDescriptionKey] : [])
    ]
```

src/components/data/LocalVariablesContent.js
```diff
@@ -80,17 +109,26 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
...
+
+    await Ajax().Workspaces.workspace(namespace, name).deleteAttributes(attributesToDelete)
+    await Ajax().Workspaces.workspace(namespace, name).shallowMergeNewAttributes(attributesToMerge)
```
zarsky-broad: Could we flip these, as they were before, so we don't try to delete until we've created?
```suggestion
    await Ajax().Workspaces.workspace(namespace, name).shallowMergeNewAttributes(attributesToMerge)
    await Ajax().Workspaces.workspace(namespace, name).deleteAttributes(attributesToDelete)
```

src/components/data/LocalVariablesContent.js
```diff
@@ -151,11 +189,12 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
               onChange: setEditKey
             }) :
             renderDataCell(amendedAttributes[rowIndex][0], googleProject)
-        }, {
+        },
+        {
```
zarsky-broad: Could you revert this line?
```suggestion
        }, {
```

src/components/data/LocalVariablesContent.js
```diff
@@ -179,7 +218,28 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
                     value: editType,
                     onChange: ({ value }) => setEditType(value),
                     options: ['string', 'number', 'boolean', 'string list', 'number list', 'boolean list']
-                  }),
+                  })
+                ]) : null
```
zarsky-broad: Instead of a ternary with null, you could just use `&&`

src/components/data/LocalVariablesContent.js
```diff
@@ -236,7 +298,8 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
           Utils.withBusyState(setBusy)
         )(async () => {
           setDeleteIndex()
-          await Ajax().Workspaces.workspace(namespace, name).deleteAttributes([amendedAttributes[deleteIndex][0]])
+          await Ajax().Workspaces.workspace(namespace, name).deleteAttributes([amendedAttributes[deleteIndex][0],
+            toDescriptionKey(amendedAttributes[deleteIndex][0])])
```
zarsky-broad: This might be easier to read if we tweak where the linebreak lands, what do you think?
```suggestion
          await Ajax().Workspaces.workspace(namespace, name)
            .deleteAttributes([amendedAttributes[deleteIndex][0], toDescriptionKey(amendedAttributes[deleteIndex][0])])
```

src/components/data/LocalVariablesContent.js
```diff
@@ -47,25 +65,36 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
...
+
   const initialAttributes = _.flow(
     _.toPairs,
-    _.remove(([key]) => /^description$|:|^referenceData_/.test(key))
+    _.remove(([key]) => /^description$|:|^referenceData_/.test(key)),
+    _.map(renameAttribute),
+    _.groupBy('key'),
+    _.values,
+    _.map(getDisplayedAttribute)
   )(attributes)
```
zarsky-broad: Now that this flow is so long, would you consider extracting this into `convertInitialAttributes` outside of the component and writing a test for it? Partially to make sure it doesn't get messed up in the future, and partly to give some concrete documentation of how it's expected to work?

zarsky-broad (commented):
Hello! Some comments for you, with a bunch of easily commitable suggestions if you like them 😄

src/components/data/LocalVariablesContent.js
zarsky-broad: This is exactly how I'd want to see this used, A+

zarsky-broad:
Also, re long string displays, I think this is probably a browser weirdness thing, as all is well on firefox on mac for me; however, we should probably think about field widths. Terra has a minimum width of 1200px, under which it has a horizontal scrollbar, but even then it's not a good situation for the value column currently:
(both value and description are set to kaan's 200 character string here; I got an error from rawls that it was too long when trying to set it for the key, which means the UI should have stopped me sooner, so we need to handle that too, not part of this PR tho)

Screen Shot 2021-12-17 at 15 51 42

src/components/data/LocalVariablesContent.js
kyuksel: Oh nice! I had hoped there'd be something like `mergeAll` but it doesn't seem to show up on the [page I use as reference](https://lodash.com/docs/4.17.15), probably because it doesn't list the newest version's. Where do you look for them?

zarsky-broad: It's actually an [fp-specific function](https://github.com/lodash/lodash/wiki/FP-Guide), created to handle the fact that non-fp-merge can take any number of arguments, but all fp functions are fixed arity.

zarsky-broad: (unfortunately, it doesn't show up on the [generated fp docs](https://gist.github.com/jfmengels/6b973b69c491375117dc) for that reason, as well)

kyuksel: > It's actually an fp-specific function, created to handle the fact that non-fp-merge can take any number of arguments, but all fp functions are fixed arity.

aaahhh…that explains the weirdness I was observing when playing with the FP version and comparing the result to the examples from the non-FP version 🤦🏻‍♂️

johntbates: Made the change - thanks!

src/components/data/LocalVariablesContent.js
johntbates: No worries - all done.

src/components/data/LocalVariablesContent.js
johntbates: Done.

src/components/data/LocalVariablesContent.js
johntbates: Done.

src/components/data/LocalVariablesContent.js
johntbates: Done.

src/components/data/LocalVariablesContent.js
johntbates: Done.

src/components/data/LocalVariablesContent.js
johntbates: Done.

src/components/data/LocalVariablesContent.js
johntbates: Done.

src/components/data/LocalVariablesContent.js
johntbates: Done.

src/components/data/LocalVariablesContent.js
johntbates: Done.

johntbates:
Thanks for all your comments!  As far as the description text box goes, how do you style it such that it truncates and adds ellipses when the description string exceeds the text box width?  I've not been able to figure this out so far.

UPDATE: I must have been referring to a prior version - the text box is being styled correctly now, with long results truncated and ellipses appended.

src/components/data/LocalVariablesContent.js
johntbates: Done.

src/components/data/LocalVariablesContent.js
```diff
@@ -179,7 +213,28 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
                     value: editType,
                     onChange: ({ value }) => setEditType(value),
                     options: ['string', 'number', 'boolean', 'string list', 'number list', 'boolean list']
-                  }),
+                  })
+                ])
```
kyuksel: Can we re-align as follows for consistency?

```javascript
              editIndex === rowIndex && h(Fragment, [
                h(Select, {
                  'aria-label': 'data type',
                  styles: { container: base => ({ ...base, marginLeft: '1rem', width: 150 }) },
                  isSearchable: false,
                  isClearable: false,
                  menuPortalTarget: document.getElementById('root'),
                  getOptionLabel: ({ value }) => _.startCase(value),
                  value: editType,
                  onChange: ({ value }) => setEditType(value),
                  options: ['string', 'number', 'boolean', 'string list', 'number list', 'boolean list']
                })
              ])
```

johntbates: Done.

wnojopra (approved):

wnojopra:
Testing and code LGTM. And thanks for adding the tests!

kyuksel:
Looks great! Will let @zarsky-broad give the official 👍🏼 though since he's also made comments.

johntbates:
Thanks, all!  I really appreciate your comments and suggestions.

src/components/data/LocalVariablesContent.js
```diff
@@ -47,25 +70,27 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
...
+
+  const initialAttributes = convertInitialAttributes(attributes)
 
   const creatingNewVariable = editIndex === initialAttributes.length
   const amendedAttributes = _.flow(
-    _.filter(([key, value]) => Utils.textMatch(textFilter, `${key} ${value}`)),
-    _.sortBy(_.first),
-    arr => [...arr, ...(creatingNewVariable ? [['', '']] : [])]
+    _.filter(([key, value, description]) => Utils.textMatch(textFilter, `${key} ${value} ${description}`)),
+    arr => [...arr, ...(creatingNewVariable ? [['', '', '']] : [])]
```
zarsky-broad: Just noticing, I think this could actually be
```suggestion
    creatingNewVariable ? Utils.append(['', '', '']) : _.identity
```
But I don't know if that's actually easier to read, so it's your call!

src/components/data/LocalVariablesContent.js
```diff
@@ -236,7 +292,9 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
           Utils.withBusyState(setBusy)
         )(async () => {
           setDeleteIndex()
-          await Ajax().Workspaces.workspace(namespace, name).deleteAttributes([amendedAttributes[deleteIndex][0]])
+          await Ajax().Workspaces.workspace(namespace, name)
+            .deleteAttributes([amendedAttributes[deleteIndex][0],
+              toDescriptionKey(amendedAttributes[deleteIndex][0])])
```
zarsky-broad: Any reason for this to be on two lines instead of one? Feel free to ignore if you think it's easier to read this way.
```suggestion
            .deleteAttributes([amendedAttributes[deleteIndex][0], toDescriptionKey(amendedAttributes[deleteIndex][0])])
```

zarsky-broad (approved):
Code looks good, couple of optional comments, I genuinely don't know if they're helpful, so feel free to ignore them if you don't think they make things better!
If you want extra credit, you could move your test from `src/_tests/...` to `src/components/data/_tests/`, which is the structure I'm trying to encourage us toward.

src/components/data/LocalVariablesContent.js
johntbates: Made it a single line.

src/components/data/LocalVariablesContent.js
johntbates: Done.

johntbates:
Thanks for your comments and suggestions.  I moved the test s.t. it matches the directory structure you're aiming for.  I really appreciate the review!

PR #2731, opened 2021-12-15
[AS-712] Use orch for saving workspace acl

davidangb:
[AS-712](https://broadworkbench.atlassian.net/browse/AS-712)

When sharing a workspace, make the ajax call to Orchestration instead of directly to Rawls. Orch contains additional logic to persist metrics about who you've shared with. These metrics power the suggestions offered when sharing.

Tested manually by running a local UI.


kyuksel (approved):
Code change looks good to me. 
Haven't tested it.

MatthewBemis (approved):
👍

PR #2732, opened 2021-12-15
[AS-643] Always treat workspace description as a string

davidangb:
[AS-643](https://broadworkbench.atlassian.net/browse/AS-643)

It is possible, by calling APIs directly, to create or edit a workspace such that its description attribute is not a string (it could be an array, or a number, etc).

In these cases, the UI broke in the workspace list and workspace dashboard.

These are now fixed.

Tested manually by running a local UI.


kyuksel (approved):

MatthewBemis (approved):
Awesome!

PR #2734, opened 2021-12-15
Entering a special character on search no longer crashes app

petesantos:
Fixes a bug where entering a character like `\` will crash the app.
This is the result of an unescaped regex.


fboulnois (approved):
Thanks for the quick fix, change is straightforward

lu-c (approved):
Easy peasy review! Thanks for the update 👍

tcjiang9 (approved):

kyuksel (approved):

PR #2736, opened 2021-12-16
[DC-94] Updating preview table styles so they have more table definition

lu-c:
Before
Screen Shot 2021-12-14 at 6 15 21 PM

After
Screen Shot 2021-12-16 at 3 47 22 PM

### DIFF
--- a/src/components/table.js
+++ b/src/components/table.js
@@ -231,6 +231,7 @@ export const FlexTable = ({
   }, [
     div({
       style: {
+        ...styles.headerRow,
         width: width - scrollbarSize,
         height: headerHeight,
         display: 'flex'
@@ -563,7 +564,11 @@ GridTable.propTypes = {
 }
 
 export const SimpleTable = ({
-  columns, rows, 'aria-label': ariaLabel, rowStyle = {}, cellStyle: cellStyleOverrides = {}, useHover = true, underRowKey
+  columns, rows, 'aria-label': ariaLabel,
+  rowStyle = {}, evenRowStyle = {}, oddRowStyle = {},
+  cellStyle: cellStyleOverrides = {},
+  headerRowStyle = {},
+  useHover = true, underRowKey
 }) => {
   Utils.useLabelAssert('SimpleTable', { 'aria-label': ariaLabel, allowLabelledBy: false })
 
@@ -572,7 +577,7 @@ export const SimpleTable = ({
     role: 'table',
     'aria-label': ariaLabel
   }, [
-    div({ role: 'row', style: { display: 'flex', alignItems: 'center' } }, [
+    !_.isEmpty(columns) && div({ role: 'row', style: { display: 'flex', alignItems: 'center', ...headerRowStyle } }, [
       _.map(({ key, header, size }) => {
         return div({
           key,
@@ -586,7 +591,7 @@ export const SimpleTable = ({
         key: i,
         role: 'row',
         as: 'div',
-        style: rowStyle, className: 'table-row',
+        style: { ...rowStyle, ...(i % 2 ? evenRowStyle : oddRowStyle) }, className: 'table-row',
         hover: useHover && { backgroundColor: colors.light(0.4) }
       }, [
         div({ style: { display: 'flex' } }, [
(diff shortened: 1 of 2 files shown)


src/components/table.js
```diff
@@ -586,7 +591,7 @@ export const SimpleTable = ({
         key: i,
         role: 'row',
         as: 'div',
-        style: rowStyle, className: 'table-row',
+        style: { ...rowStyle, ...(i % 2 ? evenRowStyle : oddRowStyle) }, className: 'table-row',
```
dmohs:
```
> 1 % 2 ? 'even' : 'odd'
'even'
> 2 % 2 ? 'even' : 'odd'
'odd'
> 3 % 2 ? 'even' : 'odd'
'even'
> 
```

src/pages/library/DataBrowserPreview.js
```diff
@@ -19,8 +19,20 @@ const styles = {
   table: {
     header: {
       color: colors.accent(),
-      height: '1rem',
+      height: '2rem', lineHeight: '2rem',
       textTransform: 'uppercase', fontWeight: 600, fontSize: '0.75rem'
+    },
+    headerRow: {
+      borderTop: '1px solid rgba(0,0,0,.3)', borderBottom: '1px solid rgba(0,0,0,.3)'
```
dmohs: Note to self: We are inconsistent about "0.3" vs. ".3" for decimals in our styles. Might be worth a cleanup PR to have the same style everywhere.

dmohs (approved):
I think the even/odd ternary doesn't match your intent, but code looks good to me otherwise.

src/components/table.js
lu-c: haha! Gosh, nice catch, updated to flip this around!

nmalfroy (approved):
LGTM.  I'm guessing it's ok that we use zebra stripes for this table but not other tables in Terra?

src/components/table.js
```diff
@@ -701,9 +706,10 @@ const SortableHandleDiv = SortableHandle(props => div(props))
  * @param {Object[]} columnSettings - current column list, in order
  * @param {string} columnSettings[].name
  * @param {bool} columnSettings[].visible
+ * @param {bool} style - style override for the button
```
zarsky-broad: Wouldn't this be an object instead of a boolean?

src/pages/library/DataBrowserPreview.js
zarsky-broad: The standard in the codebase is `0.x`, so we should use that.

src/pages/library/DataBrowserPreview.js
```diff
@@ -133,15 +145,19 @@ const DataBrowserPreview = ({ id }) => {
...
+        tableMap && tableMap[selectedTable] && div({ style: { position: 'relative', padding: '0 15px' } }, [
           h(SimpleTable, {
             'aria-label': `${_.startCase(selectedTable)} Preview Data`,
             columns: _.filter('visible', columnSettings),
-            cellStyle: { border: 'none', paddingRight: 15, wordBreak: 'break-all' },
+            cellStyle: { border: 'none', paddingRight: 15, wordBreak: 'break-all', display: 'flex', alignItems: 'center' },
+            headerRowStyle: styles.table.headerRow,
+            rowStyle: styles.table.rowStyle,
+            evenRowStyle: styles.table.evenRow,
+            oddRowStyle: styles.table.oddRow,
```
zarsky-broad: If we make the names match, could we just do `...styles.table` here?

lu-c: done!

src/components/table.js
lu-c: thank you!

zarsky-broad:
my only other q right now: is there a reason to have the height of the buttons `unset` so that they're so tall? just feels a bit weird to me compared to all our other ones.

src/pages/library/DataBrowserPreview.js
```diff
@@ -19,8 +19,20 @@ const styles = {
   table: {
     header: {
       color: colors.accent(),
-      height: '1rem',
+      height: '2rem', lineHeight: '2rem',
       textTransform: 'uppercase', fontWeight: 600, fontSize: '0.75rem'
+    },
+    headerRowStyle: {
+      borderTop: '1px solid rgba(0,0,0,0.3)', borderBottom: '1px solid rgba(0,0,0,0.3)'
```
zarsky-broad: I lied when I said that was my last comment; could we change all these to use `colors.dark`?  Reason being that we don't actually use transparency there, we calculate a similar effect, so that there isn't a render cost or change based on background color.

lu-c:
> my only other q right now: is there a reason to have the height of the buttons `unset` so that they're so tall? just feels a bit weird to me compared to all our other ones.

That would be the result of copy pasting! I'll remove it, it looks better without it

petesantos (approved):
Looks great- thanks!

PR #2737, opened 2021-12-17
Add status information for cause of failure [QA-1679, BW-977].

cahrens:
The run-workflow test has been failing often recently nightly integration runs because the call to Workspace's `checkBucketAccess` method fails (as we can see in the test screenshot). However, the message displayed is static text, with no information about the actual cause. I am attempting to get more information about the cause of the failure.

Example of how this looks when the URL is incorrect (which I do **not** expect is the cause of the actual problem!).

image

Note also that these failures are not terribly new… the oldest I found was 25 days ago. However, the frequency with which they have been seen recently has increased dramatically.

### DIFF
--- a/src/libs/analysis.js
+++ b/src/libs/analysis.js
@@ -27,7 +27,8 @@ export const launch = async ({
   try {
     await Ajax().Workspaces.workspace(namespace, name).checkBucketAccess(googleProject, bucketName, accessLevel)
   } catch (error) {
-    throw new Error('Error confirming workspace bucket access. This may be a transient problem. Please try again in a few minutes. If the problem persists, please contact support.')
+    console.error(error)
+    throw new Error(`Error confirming workspace bucket access. This may be a transient problem, please try again in a few minutes. If the problem persists, please contact support and provide this information: status ${error.status}, status text '${error.statusText}'.`)
   }
   const { entityName, processSet = false } = await Utils.cond(
     [isSnapshot || (selectedEntityType === undefined), () => ({})],
--- a/src/libs/analysis.js
+++ b/src/libs/analysis.js
@@ -27,8 +27,7 @@ export const launch = async ({
   try {
     await Ajax().Workspaces.workspace(namespace, name).checkBucketAccess(googleProject, bucketName, accessLevel)
   } catch (error) {
-    console.error(error)
-    throw new Error(`Error confirming workspace bucket access. This may be a transient problem, please try again in a few minutes. If the problem persists, please contact support and provide this information: status ${error.status}, status text '${error.statusText}'.`)
+    throw new Error('Error confirming workspace bucket access. This may be a transient problem. Please try again in a few minutes. If the problem persists, please contact support.')
   }
   const { entityName, processSet = false } = await Utils.cond(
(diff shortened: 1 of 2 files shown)


src/libs/analysis.js
```diff
@@ -27,7 +27,8 @@ export const launch = async ({
   try {
     await Ajax().Workspaces.workspace(namespace, name).checkBucketAccess(googleProject, bucketName, accessLevel)
   } catch (error) {
-    throw new Error('Error confirming workspace bucket access. This may be a transient problem. Please try again in a few minutes. If the problem persists, please contact support.')
+    console.error(error)
```
cahrens: As far as I know, there is no way to see the console output emitted by production code in the integration test logs (I tried `console.info`, `console.error`, etc.). From the test code itself, `console.info` is printed into the logs. Please let me know if I am wrong!

kyuksel: I couldn't tell why it's being swallowed  :(

kyuksel (approved):
LGTM and fingers crossed for the debug :)

src/libs/analysis.js
zarsky-broad: What if, just before we click the button in the test, we register a listener? https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#event-requestfinished (see further https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#class-httprequest)

cahrens: Thanks @zarsky-broad! I found it difficult to get the details of the object when using the console listener, but the request listener is giving the same level of detail that I would have gotten with my previous commit. So now this debugging only impacts the test.

cahrens: Example logging: https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/7553/workflows/20c242b2-0ecb-41b5-805f-759ca7bc48de/jobs/35473/parallel-runs/0/steps/0-109

kyuksel: Where exactly are we looking on the link above for the example of the logging?

kyuksel (approved):
Still looks good to me but I'll defer to @zarsky-broad.

src/libs/analysis.js
cahrens: 1. Click "Details" next to the integration job in checks passed section.
2. Find the parallel run that included the run-workspace job (Usually the longest running one).
3. Expand the stage above "Uploading test results".
image

cjllanwarne:
LGTM too. I'll wait for Isaac for any thumbs though. Incidentally, is it possible to just keep re-running the tests against this PR until an error appears and then use _that_ for debugging without having to commit this?

cahrens:
> LGTM too. I'll wait for Isaac for any thumbs though. Incidentally, is it possible to just keep re-running the tests against this PR until an error appears and then use _that_ for debugging without having to commit this?

The test hasn't been failing on branch runs, only on nightly integration test runs on alpha and staging. And even there, it's sporadic. Given that the logging only impacts the test code (and is innocuous), I think it is simpler to just merge this and keep an eye on what happens on nightly runs.

zarsky-broad (approved):

PR #2742, opened 2021-12-20
Clean up tests [POL-150]

zarsky-broad:
Changes in this PR:
1. add a manual linting step in circle, since test files aren't part of the app, so they weren't being linted
2. remove legacy test setup cruft, to clarify what's actually doing something
3. restructure `utils.js`, moving react-specific functions to `react-utils.js`
    - as part of this migration, my IDE cleaned up imports in effected files
    - this is specifically to allow cleaner mocking in the future
4. return to intended test folder structure: rather than a central `_tests` folder with a hierarchy inside it, there's now a `_tests` folder in the same directory as the file being tested

I was wondering why the existing tests took so long; because the codebase is tightly coupled, the tests all wind up partially starting the app itself (although without actually trying to render anything). Splitting up react utils from logic utils is a start to handling this; ideally, utils-type files would be divided between pure logic functions, which can be unit tested, and react/component functions, which need to be tested differently.


kyuksel (approved):
LGTM with two questions: Why do we want to `import *` from `src/libs/utils`? And why not from `src/libs/react-utils`?

Thanks for the clean-up!

zarsky-broad:
`import * as Utils`:
This is historical. A lot of the things exported from Utils are named very generically; we instituted this so that they would always be namespaced when used, to avoid confusion.

For `react-utils`, the names are react-specific, so they're less generic. Most are hooks, so `useX`, and the rest are wrappers, either `withX` or things that add `withDisplayName` to generic react functions.

cahrens:
LGTM. I definitely like having the tests live directly with code, though I would drop the underscore in the folder name unless it is required by the testing framework (`tests` instead of `_tests`).

In addition to creating a JIRA ticket, please also announce this change in Slack when it is merged so folks know where their unit tests moved to (and to further advertise the support for writing unit tests).

And thanks for doing this cleanup!!

cahrens (approved):

zarsky-broad:
@cahrens So the underscore definitely isn't required - tests can either be in a folder called `__tests__` or have `.test.js` or `.spec.js` on the end, and we have full control over the regex that finds them. As I recall, the thinking behind the naming was that, since the test folder was going to be next to other structural subfolders and files, the underscore was purely to make it appear apart from other contents of a folder when listed alphabetically, in this case at the top of the list. I personally like that; your thoughts?

cahrens:
> @cahrens So the underscore definitely isn't required - tests can either be in a folder called `__tests__` or have `.test.js` or `.spec.js` on the end, and we have full control over the regex that finds them. As I recall, the thinking behind the naming was that, since the test folder was going to be next to other structural subfolders and files, the underscore was purely to make it appear apart from other contents of a folder when listed alphabetically, in this case at the top of the list. I personally like that; your thoughts?

I personally would vote for the directory being simply `test`, which I think is more consistent with the naming structure in other repos. But I don't feel strongly about it, and it's an easy thing to change later.

zarsky-broad:
@cahrens Legit. Can we talk more about it when we get back? I'll merge for now.

PR #2744, opened 2021-12-21
[TUB-20] Clearer error handling during screenshot save

dmohs:
This doesn't change the logic, but hopefully it makes the logic for handling errors a bit more obvious.

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -141,35 +141,37 @@ const openError = async page => {
   return !!errorDetails.length
 }
 
+const maybeSaveScreenshot = async page => {
+  if( !screenshotDir) { return }
+  try {
+    const path = `${screenshotDir}/failure-${Date.now()}-${testName}.png`
+    const failureNotificationDetailsPath = `${screenshotDir}/failureDetails-${Date.now()}-${testName}.png`
+
+    await page.screenshot({ path, fullPage: true })
+
+    const errorsPresent = await openError(page)
+
+    if (errorsPresent) {
+      await page.screenshot({ path: failureNotificationDetailsPath, fullPage: true })
+    }
+
+    if (screenshotBucket) {
+      const storage = new Storage()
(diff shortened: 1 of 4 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -141,35 +142,37 @@ const openError = async page => {
   return !!errorDetails.length
 }
 
+const maybeSaveScreenshot = async (page, testName) => {
+  if (!screenshotDir) { return }
```
kyuksel: Not sure how/if ever `screenshotDir` may be `falsy` but I wonder if we should log info in case it is. Perhaps something like:

```suggestion
  if (!screenshotDir) { 
    console.log(`Cannot take screenshot as screenshot directory doesn't exist.`) 
    return
  }
```

kyuksel (approved):
LGTM with a minor comment inline.
Needs a ticket for compliance reasons.
Took the liberty to push a couple of fixes. Hope that was okay.

kyuksel:
Needs a ticket for compliance reasons.

dmohs:
> Took the liberty to push a couple of fixes. Hope that was okay.

Thank you!

integration-tests/utils/integration-utils.js
dmohs: I think I should rename this variable to `screenshotDirPath`, since it is only a string. The directory is created before the tests are started: https://github.com/DataBiosphere/terra-ui/blob/dev/.circleci/config.yml#L157

The environment variable `SCREENSHOT_DIR` that gets assigned to this variable is serving two purposes: To tell us where to put the screenshots and as a flag to enable screenshot capturing via this conditional.

Would the rename help clarify?

kyuksel: Renaming the variable as `screenshotDirPath` makes sense.

Are you also saying, we don't need to log a message? If so, I'm not sure I understand why we don't.

cahrens (approved):

PR #2745, opened 2021-12-21
[TUB-19] Add HTTP response logging to integration tests

dmohs:
While tests are flaky, it might be helpful to log all HTTP responses.

### DIFF
--- a/integration-tests/tests/run-workflow.js
+++ b/integration-tests/tests/run-workflow.js
@@ -39,16 +39,13 @@ const testRunWorkflowFn = _.flow(
   await click(page, clickable({ text: 'OK' }))
   await click(page, clickable({ text: 'Run analysis' }))
 
-  // Get request status for sporadically failing checkBucketAccess call.
-  const handler = request => {
-    console.log(`${request.url()} : ${request.status()} : ${request.statusText()}`)
-  }
-  page.on('response', handler)
+  // Uncomment the following to get request status for sporadically failing checkBucketAccess call.
+  // const stopLoggingPageAjaxResponses = logPageAjaxResponses(page)
   await Promise.all([
     page.waitForNavigation(),
     click(page, clickable({ text: 'Launch' }))
   ])
-  page.off('response', handler)
+  // stopLoggingPageAjaxResponses()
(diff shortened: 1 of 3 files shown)


kyuksel (approved):
Sounds reasonable to have some potentially noisy logging since we have flakiness going on. We can plan on turning it off once tests stabilize.

integration-tests/tests/run-workflow.js
```diff
@@ -39,16 +39,13 @@ const testRunWorkflowFn = _.flow(
   await click(page, clickable({ text: 'OK' }))
   await click(page, clickable({ text: 'Run analysis' }))
 
-  // Get request status for sporadically failing checkBucketAccess call.
-  const handler = request => {
-    console.log(`${request.url()} : ${request.status()} : ${request.statusText()}`)
-  }
-  page.on('response', handler)
+  // Uncomment the following to get request status for sporadically failing checkBucketAccess call.
```
cahrens: This comment is confusing to me because the general logging is in effect, so there is no need to uncomment line 43. This is here in case we turn off the general logging?

cahrens: If that is the case, I would suggest

```If general ajax logging is disabled, uncomment the following to debug the sporadically failing checkBucketAccess call.```

**Review by cahrens — APPROVED** [review]
Suggested comment change, otherwise LGTM.

File: integration-tests/tests/run-workflow.js (hunk 1 above)
Comment (dmohs) [reply integration-tests/tests/run-workflow.js]: I was worried about the confusion here. Thanks for the improved wording.

File: integration-tests/utils/integration-utils.js [hunk 2]
```diff
@@ -174,6 +174,28 @@ const withScreenshot = _.curry((testName, fn) => async options => {
   }
 })
 
+const logPageConsoleMessages = page => {
```
cahrens, line 177: @dmohs how is one supposed to use this method? I tried using it [here](https://github.com/DataBiosphere/terra-ui/pull/2790#discussion_r793928434), but the console.info calls are not going through line 178.

PR #2747, opened 2022-01-04
[DC-105] Added close button to preview page

lu-c:
BEFORE
Screen Shot 2022-01-04 at 2 08 51 PM

AFTER
Screen Shot 2022-01-04 at 2 07 59 PM

### DIFF
--- a/src/pages/library/DataBrowserPreview.js
+++ b/src/pages/library/DataBrowserPreview.js
@@ -134,11 +134,11 @@ const DataBrowserPreview = ({ id }) => {
         div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', justifyContent: 'space-between', width: '100%', lineHeight: '26px' } }, [
           h1([snapshot['dct:title']]),
           h(Link, {
-            onClick: () => { Nav.goToPath('library-details', { id: Nav.getCurrentRoute().params.id }) },
+            href: Nav.getLink('library-details', { id: Nav.getCurrentRoute().params.id }),
             'aria-label': 'Close',
             style: { marginTop: '1rem' }
           }, [
-            icon('times', { size: 30, style: { color: colors.primary('light') } })
+            icon('times', { size: 30 })
           ])
         ]),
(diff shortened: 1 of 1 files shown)


src/pages/library/DataBrowserPreview.js
```diff
@@ -130,8 +131,15 @@ const DataBrowserPreview = ({ id }) => {
     !snapshot ?
       centeredSpinner() :
       div({ style: { padding: 20 } }, [
-        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', width: '100%', lineHeight: '26px' } }, [
-          h1([snapshot['dct:title']])
+        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', justifyContent: 'space-between', width: '100%', lineHeight: '26px' } }, [
+          h1([snapshot['dct:title']]),
+          h(Link, {
+            onClick: () => { Nav.goToPath('library-details', { id: Nav.getCurrentRoute().params.id }) },
```
lu-c: Toyed with the idea of making this `Nav.history.goBack()`, but decided to go with a new path instead just in case the user got to the page by copy-pasting the link from somewhere else

zarsky-broad: This should get the href from the link if you're going to a known path, rather than the onClick.

src/pages/library/DataBrowserPreview.js
```diff
@@ -130,8 +131,15 @@ const DataBrowserPreview = ({ id }) => {
...
+            'aria-label': 'Close',
+            style: { marginTop: '1rem' }
+          }, [
+            icon('times', { size: 30, style: { color: colors.primary('light') } })
```
zarsky-broad: The color functions only work with a number.

petesantos (approved):

pshapiro4broad (approved):

fboulnois (approved):
lgtm, thank you!

PR #2749, opened 2022-01-04
[DC-111] Hide/show button/checkbox based off of access status

lu-c:
**BEFORE**
Screen Shot 2022-01-04 at 2 51 21 PM
Screen Shot 2022-01-04 at 2 51 29 PM

**AFTER**
Screen Shot 2022-01-04 at 2 50 52 PM
Screen Shot 2022-01-04 at 2 50 44 PM

If the user tries to link to the workspace without access, it should error for hte user. This PR disables the checkbox and hides the "Link to Workspace" link on the details page

### DIFF
--- a/src/pages/library/DataBrowserDetails.js
+++ b/src/pages/library/DataBrowserDetails.js
@@ -168,7 +168,7 @@ const Sidebar = ({ snapshot, id, setShowRequestAccessModal }) => {
         'Preview data'
       ])
     ]),
-    h(ButtonPrimary, {
+    snapshot.access === snapshotAccessTypes.GRANTED && h(ButtonPrimary, {
       style: { fontSize: 16, textTransform: 'none', height: 'unset', width: 230, marginTop: 20 },
       onClick: () => {
         Ajax().Metrics.captureEvent(`${Events.catalogWorkspaceLink}:detailsView`, {
--- a/src/pages/library/DataBrowserDetails.js
+++ b/src/pages/library/DataBrowserDetails.js
@@ -153,7 +153,7 @@ const Sidebar = ({ snapshot, id, setShowRequestAccessModal }) => {
         ])
       ])
     ]),
-    h(ButtonOutline, {
+    snapshot.access === snapshotAccessTypes.GRANTED && h(ButtonOutline, {
       style: { fontSize: 16, textTransform: 'none', height: 'unset', width: 230, marginTop: 20 },
       onClick: () => {
         Ajax().Metrics.captureEvent(`${Events.catalogView}:previewData`, {
--- a/src/pages/library/DataBrowserDetails.js
+++ b/src/pages/library/DataBrowserDetails.js
@@ -153,7 +153,9 @@ const Sidebar = ({ snapshot, id, setShowRequestAccessModal }) => {
         ])
       ])
     ]),
-    snapshot.access === snapshotAccessTypes.GRANTED && h(ButtonOutline, {
+    h(ButtonOutline, {
+      disabled: snapshot.access !== snapshotAccessTypes.GRANTED,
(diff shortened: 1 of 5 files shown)


pshapiro4broad:
Preview Data is also not available if a user hasn't been granted access to the dataset, so it should be disabled or removed as well.

petesantos (approved):

lu-c:
> Preview Data is also not available if a user hasn't been granted access to the dataset, so it should be disabled or removed as well.

Updated the PR so it removes the "Preview Data" button from the details page. The page is still accessible if the user types +"/preview" into the url bar, at which point, they will see this:

Screen Shot 2022-01-04 at 4 08 12 PM

I think normal dataset preview isnt impacted by this:
Screen Shot 2022-01-04 at 4 08 36 PM

src/pages/library/DataBrowserDetails.js
```diff
@@ -169,6 +171,8 @@ const Sidebar = ({ snapshot, id, setShowRequestAccessModal }) => {
       ])
     ]),
     h(ButtonPrimary, {
+      disabled: snapshot.access !== snapshotAccessTypes.GRANTED,
+      tooltip: snapshot.access === snapshotAccessTypes.GRANTED ? '' : 'You do not have access to this dataset. Please request access to unlock this feature.',
```
pshapiro4broad: Is there a way to share text to avoid duplication? (Or maybe this isn't something that we do in terra-ui right now?)

src/pages/library/DataBrowserPreview.js
```diff
@@ -118,9 +126,11 @@ const DataBrowserPreview = ({ id }) => {
 
   useOnMount(() => {
     const loadData = async () => {
-      const metadata = await Ajax(signal).DataRepo.getPreviewMetadata(id)
-      setTables(metadata.tables)
-      selectTable({ value: metadata.tables[0].name })
+      if (snapshot && snapshot.access === snapshotAccessTypes.GRANTED) {
```
pshapiro4broad: can this be written using `?.`?
```suggestion
      if (snapshot?.access === snapshotAccessTypes.GRANTED) {
```

pshapiro4broad (approved):
Looks good, just a few minor comments.

src/pages/library/DataBrowserPreview.js
```diff
@@ -129,11 +139,30 @@ const DataBrowserPreview = ({ id }) => {
...
     !snapshot ?
       centeredSpinner() :
-      div({ style: { padding: 20 } }, [
+      div({ style: { padding: 20, display: 'flex', flexDirection: 'column', height: '100%' } }, [
         div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', width: '100%', lineHeight: '26px' } }, [
           h1([snapshot['dct:title']])
         ]),
-        div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', width: '100%', marginBottom: 30 } }, [
+        snapshot.access === snapshotAccessTypes.CONTROLLED && div({ style: { display: 'flex', flexDirection: 'row', backgroundColor: 'white', fontSize: '1.1rem', lineHeight: '1.7rem', padding: '20px 30px 25px', width: 'fit-content', margin: 'auto' } }, [
+          icon('lock', { size: 27, style: { marginRight: 20, marginTop: 5, color: commonStyles.access.controlled } }),
```
petesantos: Given this morning's conversation, will this be getting removed?

lu-c: yes!

src/pages/library/DataBrowserPreview.js
lu-c: yes, i'll make that change 👍

src/pages/library/DataBrowserDetails.js
lu-c: We don't currently do that, but I can see the value... I'll add it to the databrowser-utils constants

PR #2750, opened 2022-01-04
Refactor multipleDisks and getDiskAppType functions

kyuksel:
1. Wanted to make it easier to explain my comment on #2662
2. Refactored another method a bit for concision while at it
3. Ran reformatter

**Note:** Haven't tested it.

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -269,7 +269,7 @@ export const getDiskAppType = ({ labels: { saturnApplication } }) => {
 export const haveTooManyDisks = (appType, filteredDisks, disksByProject) => {
   if (appType === tools.galaxy.appType) {
     // Galaxy can have multiple disks per project as long as they are in different workspaces
-    const galaxyDisksOnly = _.filter(disk => appType === tools.galaxy.appType && disk.status !== 'DELETING', filteredDisks)
+    const galaxyDisksOnly = _.filter(disk => disk.status !== 'DELETING', filteredDisks)
     const galaxyDiskWorkspaces = _.map(currentDisk => currentDisk.labels.saturnWorkspaceName, galaxyDisksOnly)
     return _.uniq(galaxyDiskWorkspaces).length !== galaxyDiskWorkspaces.length
   } else {
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -261,11 +261,20 @@ export const getCurrentAttachedDataDisk = (app, appDataDisks) => {
 }
 
 // If the disk was attached to an app, return the appType. Otherwise return undefined.
-export const getDiskAppType = disk => {
-  const saturnApp = disk.labels.saturnApplication
+export const getDiskAppType = ({ labels: { saturnApplication } }) => {
   // Do a case-insensitive match as disks have been created with both "galaxy" and "GALAXY".
-  const appType = _.find(type => type.toLowerCase() === saturnApp?.toLowerCase(), allAppTypes)
-  return appType
+  return _.find(type => type.toLowerCase() === saturnApplication?.toLowerCase(), allAppTypes)
+}
+
+export const haveTooManyDisks = (appType, filteredDisks, disksByProject) => {
+  if (appType === tools.galaxy.appType) {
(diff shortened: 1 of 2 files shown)


src/libs/runtime-utils.js
```diff
@@ -282,7 +291,8 @@ export const getCurrentPersistentDisk = (appType, apps, appDataDisks, workspaceN
   return !!currentDiskName ?
     _.find({ name: currentDiskName }, appDataDisks) :
     _.flow(
-      _.filter(disk => getDiskAppType(disk) === appType && disk.status !== 'Deleting' && !_.includes(disk.name, attachedDiskNames) && disk.labels.saturnWorkspaceName === workspaceName),
+      _.filter(disk => getDiskAppType(disk) === appType && disk.status !== 'Deleting' && !_.includes(disk.name, attachedDiskNames) &&
+        disk.labels.saturnWorkspaceName === workspaceName),
```
kyuksel, line 295: reformatting change

src/libs/runtime-utils.js
```diff
@@ -261,11 +261,20 @@ export const getCurrentAttachedDataDisk = (app, appDataDisks) => {
...
   // Do a case-insensitive match as disks have been created with both "galaxy" and "GALAXY".
-  const appType = _.find(type => type.toLowerCase() === saturnApp?.toLowerCase(), allAppTypes)
-  return appType
+  return _.find(type => type.toLowerCase() === saturnApplication?.toLowerCase(), allAppTypes)
+}
+
+export const haveTooManyDisks = (appType, filteredDisks, disksByProject) => {
+  if (appType === tools.galaxy.appType) {
+    // Galaxy can have multiple disks per project as long as they are in different workspaces
+    const galaxyDisksOnly = _.filter(disk => appType === tools.galaxy.appType && disk.status !== 'DELETING', filteredDisks)
```
cahrens: As I mentioned in the other PR, I do not think we should be treating Galaxy as a special case here.

kyuksel: How would that work? If we use the pre-existing predicate

```javascript
_.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting', disksByProject).length > 1
```

it can validly evaluate to `true` for Galaxy apps because there _can_ be two Galaxy apps, and hence two disks, for a given `googleProject`. What would be invalid is for there to be two Galaxy apps (and hence two disks) for a given `googleProject` _and_ a given `workspace`. 

Cc: @gpcarr

cahrens: My point is that the restriction in [Leo](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/http/service/AppServiceInterp.scala#L167) is not specific to Galaxy, so the check that is being done should be the same for all app types.

src/libs/runtime-utils.js
```diff
@@ -261,11 +261,20 @@ export const getCurrentAttachedDataDisk = (app, appDataDisks) => {
 }
 
 // If the disk was attached to an app, return the appType. Otherwise return undefined.
-export const getDiskAppType = disk => {
-  const saturnApp = disk.labels.saturnApplication
+export const getDiskAppType = ({ labels: { saturnApplication } }) => {
```
cahrens, line 264: I went down this route of unpacking in my previous PRs and found myself not liking it. I think the API is much clearer when it takes a disk.

(I also thought the unpacking failed when saturnApplication is undefined, but I have a unit test for that case, so presumably the method must be working.)

src/libs/runtime-utils.js
```diff
@@ -261,11 +261,20 @@ export const getCurrentAttachedDataDisk = (app, appDataDisks) => {
...
   // Do a case-insensitive match as disks have been created with both "galaxy" and "GALAXY".
-  const appType = _.find(type => type.toLowerCase() === saturnApp?.toLowerCase(), allAppTypes)
-  return appType
+  return _.find(type => type.toLowerCase() === saturnApplication?.toLowerCase(), allAppTypes)
+}
+
+export const haveTooManyDisks = (appType, filteredDisks, disksByProject) => {
```
cahrens, line 269: Since this is being exposed as a method, and it's complicated, please add unit test coverage.

PR #2751, opened 2022-01-05
[DC-93] Test link to workspace page

samanehsan:
Update the `run-catalog-workflow.js` puppeteer test to click on the first item in the dataset list and then click "Link to a workspace." From there, it checks that both options (start with an existing workspace and a new workspace) look as expected.

### DIFF
--- a/integration-tests/tests/run-catalog-workflow.js
+++ b/integration-tests/tests/run-catalog-workflow.js
@@ -37,6 +37,8 @@ const testCatalogFlowFn = _.flow(
   await click(page, input({ labelContains: 'Select a workspace' }))
   await click(page, `//*[@role="combobox"][contains(normalize-space(.), "${workspaceName}")]`)
   await click(page, clickable({ textContains: 'Import' }))
+  await waitForNoSpinners(page)
+  await page.url().includes(workspaceName)
 })
 
 const testCatalog = {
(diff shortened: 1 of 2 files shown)


petesantos (commented):
This looks good. 
Could we have this test link to a new workspace?
There is a [wrapper ](https://github.com/DataBiosphere/terra-ui/blob/5e131c14cb1244b2accb51e66fd8b442acd56ae2/integration-tests/utils/integration-helpers.js#L44)that will create and delete a workspace for the test. If you need help, let me know.

fboulnois (approved):
This is great! I think @petesantos 's suggestion is good.

samanehsan:
@petesantos, I updated the test to use the wrapper you mentioned and link the dataset to that test workspace!

integration-tests/tests/run-catalog-workflow.js
```diff
@@ -20,6 +22,21 @@ const testCatalogFlowFn = _.flow(
...
+  await waitForNoSpinners(page)
+
+  await click(page, clickable({ textContains: 'Start with a new workspace' }))
+  await findText(page, 'Create a New Workspace')
+  await click(page, clickable({ textContains: 'Cancel' }))
+
+  await click(page, clickable({ textContains: 'Start with an existing workspace' }))
+  await click(page, input({ labelContains: 'Select a workspace' }))
+  await click(page, `//*[@role="combobox"][contains(normalize-space(.), "${workspaceName}")]`)
+  await click(page, clickable({ textContains: 'Import' }))
```
petesantos, line 39: To close out the journey, could we add two more items:
1. A few more steps to ensure we made it to the workspace page
    1. Something like another `waitForNoSpinners` and a check to ensure we made it to the workspace page without an error
2. Wrap it in a `withScreenshot` - this will take a screenshot in case an error is encountered.

samanehsan: It looks like every test calls `withScreenshot` here: https://github.com/DataBiosphere/terra-ui/blob/dev/integration-tests/src/server.js#L50

petesantos (approved):

PR #2753, opened 2022-01-06
AS-521: don't break on lists-of-nulls

davidangb:
[AS-521](https://broadworkbench.atlassian.net/browse/AS-521)

if an entity attribute is an AttributeValueList containing only AttributeNulls, don't break the entire UI when attempting to edit that attribute.

viewing the attribute works ok, both before and after this PR:
view

attempting to edit the attribute breaks before this PR:
before

and does not break after this PR:
after

Note that the edit functionality is not ideal - simply editing the attribute and clicking save will translate it into a list of empty strings instead of a list of nulls. However, this feels like positive forward movement, better off than we were before, so I'm submitting this as-is.

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -71,7 +71,7 @@ export const renderDataCell = (data, googleProject) => {
 
   const renderArray = items => {
     return _.map(([i, v]) => h(Fragment, { key: i }, [
-      renderCell(v.toString()), i < (items.length - 1) && div({ style: { marginRight: '0.5rem', color: colors.dark(0.85) } }, ',')
+      renderCell(v?.toString()), i < (items.length - 1) && div({ style: { marginRight: '0.5rem', color: colors.dark(0.85) } }, ',')
     ]), Utils.toIndexPairs(items))
   }
 
@@ -473,7 +473,9 @@ export const EntityEditor = ({ entityType, entityName, attributeName, attributeV
   const initialIsList = _.isObject(attributeValue) && attributeValue.items
   const initialType = Utils.cond(
     [initialIsReference, () => 'reference'],
-    [(initialIsList ? attributeValue.items[0] : attributeValue) === undefined, () => 'string'],
+    // explicit double-equal to check for null and undefined, since entity attribute lists can contain nulls
+    // eslint-disable-next-line eqeqeq
+    [(initialIsList ? attributeValue.items[0] : attributeValue) == undefined, () => 'string'],
     [initialIsList, () => typeof attributeValue.items[0]],
     () => typeof attributeValue
   )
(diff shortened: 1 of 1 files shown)


src/components/data/data-utils.js
```diff
@@ -71,7 +71,7 @@ export const renderDataCell = (data, googleProject) => {
 
   const renderArray = items => {
     return _.map(([i, v]) => h(Fragment, { key: i }, [
-      renderCell(v.toString()), i < (items.length - 1) && div({ style: { marginRight: '0.5rem', color: colors.dark(0.85) } }, ',')
+      renderCell(v?.toString()), i < (items.length - 1) && div({ style: { marginRight: '0.5rem', color: colors.dark(0.85) } }, ',')
```
davidangb: the change from `v.toString()` to `v?.toString()` here isn't strictly necessary for this PR; viewing attributes worked before this change. But it felt like a good, safe change to make.

src/components/data/data-utils.js
```diff
@@ -491,7 +493,9 @@ export const EntityEditor = ({ entityType, entityName, attributeName, attributeV
   ))
   const [editType, setEditType] = useState(() => Utils.cond(
     [initialIsReference, () => 'reference'],
-    [(initialIsList ? attributeValue.items[0] : attributeValue) === undefined, () => 'string'],
+    // explicit double-equal to check for null and undefined, since entity attribute lists can contain nulls
+    // eslint-disable-next-line eqeqeq
+    [(initialIsList ? attributeValue.items[0] : attributeValue) == undefined, () => 'string'],
```
davidangb, line 498: here and above, we were previously comparing `null === undefined`, which is false, so this condition would be bypassed and we fell down to the default condition. The default condition returns `typeof attributeValue` … and `typeof(null)` is `object`.

When the UI attempted to edit these, it expected an object but actually got null, and the UI broke.

In this PR, by changing `===` to `==`, this condition returns true because `null == undefined`. Thus, this condition would be activated, returning 'string', and the UI will now attempt to edit the nulls as strings.

src/components/data/data-utils.js
kyuksel: The main guideline we try to follow whether to use optional chaining (i.e. `?` above) is…use it if data is fetched from an API and may not exist, and don't use otherwise as it can unintentionally hide issues and make code assumptions less clear.

kyuksel (approved):
Code LGTM with one inline comment to consider while evaluating optional chaining usage

src/components/data/data-utils.js
davidangb: I have now reverted that optional chain!

sagehen03 (approved):

PR #2754, opened 2022-01-06
Attempt waiting for bucket access before trying to launch workflow [QA-1679].

cahrens:
Add polling to make sure the user has bucket write location before we try to launch a workflow. Otherwise the launch will fail with a message in the dialog suggesting that they refresh and try again, and that is a difficult thing to work around in the integration test.

<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/integration-tests/tests/run-workflow.js
+++ b/integration-tests/tests/run-workflow.js
@@ -1,6 +1,6 @@
 const _ = require('lodash/fp')
 const pRetry = require('p-retry')
-const { withWorkspace, createEntityInWorkspace } = require('../utils/integration-helpers')
+const { checkBucketAccess, withWorkspace, createEntityInWorkspace } = require('../utils/integration-helpers')
 const { click, clickable, dismissNotifications, findElement, fillIn, input, signIntoTerra, waitForNoSpinners, findInGrid, navChild, findInDataTableRow } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
@@ -17,6 +17,8 @@ const testRunWorkflowFn = _.flow(
   await dismissNotifications(page)
 
   await createEntityInWorkspace(page, billingProject, workspaceName, testEntity)
+  // Wait for bucket access to avoid sporadic failure when launching workflow.
+  await checkBucketAccess(page, billingProject, workspaceName)
 
   await click(page, clickable({ textContains: 'View Workspaces' }))
   await waitForNoSpinners(page)
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -57,6 +57,21 @@ const createEntityInWorkspace = (page, billingProject, workspaceName, testEntity
   }, billingProject, workspaceName, testEntity)
 }
 
+const checkBucketAccess = async (page, billingProject, workspaceName, accessLevel = 'OWNER') => {
+  const details = await page.evaluate((billingProject, workspaceName) => {
+    return window.Ajax().Workspaces.workspace(billingProject, workspaceName).details()
+  }, billingProject, workspaceName)
+  const bucketName = details.workspace.bucketName
+  console.info(`Checking workspace access for ${billingProject}, ${workspaceName}, ${bucketName}.`)
+  // Try polling for workspace bucket access to be available.
+  await page.waitForFunction(async (billingProject, workspaceName, bucketName, accessLevel) => {
+    try {
+      await window.Ajax().Workspaces.workspace(billingProject, workspaceName).checkBucketAccess(billingProject, bucketName, accessLevel)
+      return true
+    } catch (e) { return false }
+  }, { timeout: 60000, polling: 10000 }, billingProject, workspaceName, bucketName, accessLevel)
+}
+
 const makeUser = async () => {
   const { email } = await fetchLyle('create')
   const { accessToken: token } = await fetchLyle('token', email)
@@ -131,7 +146,9 @@ const registerUser = withSignedInPage(async ({ page, token }) => {
         if (e instanceof Response) {
           const text = await e.text()
           const headers = e.headers
-          const headerAuthToken = headers.get('authorization') ? `...${clipToken(headers.get('authorization').toString())}` : headers.get('authorization')
+          const headerAuthToken = headers.get('authorization') ?
+            `...${clipToken(headers.get('authorization').toString())}` :
+            headers.get('authorization')
           throw new Error(`Failed to Ajax: ${e.url} authorization header was: ${headerAuthToken} and status of: ${e.status}: ${text}`)
         } else {
           throw e
@@ -153,6 +170,7 @@ const withRegisteredUser = test => withUser(async options => {
 })
 
 module.exports = {
+  checkBucketAccess,
   createEntityInWorkspace,
   defaultTimeout,
   withWorkspace,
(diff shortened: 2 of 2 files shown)


integration-tests/utils/integration-helpers.js
```diff
@@ -57,6 +57,21 @@ const createEntityInWorkspace = (page, billingProject, workspaceName, testEntity
...
+const checkBucketAccess = async (page, billingProject, workspaceName, accessLevel = 'OWNER') => {
+  const details = await page.evaluate((billingProject, workspaceName) => {
+    return window.Ajax().Workspaces.workspace(billingProject, workspaceName).details()
+  }, billingProject, workspaceName)
+  const bucketName = details.workspace.bucketName
+  console.info(`Checking workspace access for ${billingProject}, ${workspaceName}, ${bucketName}.`)
+  // Try polling for workspace bucket access to be available.
+  await page.waitForFunction(async (billingProject, workspaceName, bucketName, accessLevel) => {
+    try {
+      await window.Ajax().Workspaces.workspace(billingProject, workspaceName).checkBucketAccess(billingProject, bucketName, accessLevel)
```
cahrens, line 69: This is the same call we make in the [launch workflow code](https://github.com/DataBiosphere/terra-ui/blob/795eacf63f30f9dd092529610d5ff1f2b164a73e/src/libs/analysis.js#L28).

integration-tests/utils/integration-helpers.js
```diff
@@ -131,7 +146,9 @@ const registerUser = withSignedInPage(async ({ page, token }) => {
         if (e instanceof Response) {
           const text = await e.text()
           const headers = e.headers
-          const headerAuthToken = headers.get('authorization') ? `...${clipToken(headers.get('authorization').toString())}` : headers.get('authorization')
+          const headerAuthToken = headers.get('authorization') ?
```
cahrens, line 149: I didn't mean to change this… IntellIj must have done some formatting.

integration-tests/tests/run-workflow.js
```diff
@@ -17,6 +17,8 @@ const testRunWorkflowFn = _.flow(
   await dismissNotifications(page)
 
   await createEntityInWorkspace(page, billingProject, workspaceName, testEntity)
+  // Wait for bucket access to avoid sporadic failure when launching workflow.
+  await checkBucketAccess(page, billingProject, workspaceName)
```
breilly2, line 21: Another option would be to call this from `makeWorkspace` (in `integration-helpers.js`). There are other tests that have shown similar failures. For example, running a notebook requires loading the notebooks tab which lists the current notebooks which fails if the bucket can't be read.

breilly2 (approved):
👍 Looks good to me. I think this is a slightly different approach than my failed attempt in the past, so I'm hopeful the outcome will be better.

integration-tests/tests/run-workflow.js
cahrens: I was wondering if we should make it in a more central location. My concern though is that we want to make the integration tests as close to a user experience as possible. In general, this problem is indicative of something that actually does impact users, and I would like us to consider on a test-by-test basis if the fix should be something in the product code itself (to handle the failure condition better), or if it is OK to make the test do something that is different from what the user would experience.

cahrens: Also, it may not be true that every time you create a workspace you need to wait for the write access, and I don't wish to change the timing in general.

integration-tests/utils/integration-helpers.js
```diff
@@ -57,6 +57,21 @@ const createEntityInWorkspace = (page, billingProject, workspaceName, testEntity
   }, billingProject, workspaceName, testEntity)
 }
 
+const checkBucketAccess = async (page, billingProject, workspaceName, accessLevel = 'OWNER') => {
+  const details = await page.evaluate((billingProject, workspaceName) => {
+    return window.Ajax().Workspaces.workspace(billingProject, workspaceName).details()
+  }, billingProject, workspaceName)
```
dmohs, line 63: Is it necessary to pass these values through or could you do this?
```suggestion
  const details = await page.evaluate(() => {
    return window.Ajax().Workspaces.workspace(billingProject, workspaceName).details()
  })
```

integration-tests/tests/run-workflow.js
dmohs: Thank you both for this excellent discussion.

integration-tests/utils/integration-helpers.js
```diff
@@ -57,6 +57,21 @@ const createEntityInWorkspace = (page, billingProject, workspaceName, testEntity
...
+  const bucketName = details.workspace.bucketName
+  console.info(`Checking workspace access for ${billingProject}, ${workspaceName}, ${bucketName}.`)
+  // Try polling for workspace bucket access to be available.
+  await page.waitForFunction(async (billingProject, workspaceName, bucketName, accessLevel) => {
```
dmohs, line 67: Same question as above. I would really like to know whether these values are closed-over or not.

dmohs (approved):

integration-tests/utils/integration-helpers.js
cahrens: It seemed like I needed to, but I would love it if I do not… I will experiment. 

What do you mean by closed-over?

cahrens: @dmohs Unfortunately without explicitly passing in the values, the code fails (variables aren't defined). I don't understand though why it is necessary.

integration-tests/utils/integration-helpers.js
cahrens: That fails with billingProject not being defined.

zarsky-broad: passing arguments via through like this is the way we get them into scope for the function evaluation on the page; the function itself is really just a string until it's run on the page.

integration-tests/utils/integration-helpers.js
aednichols: closed over = captured and made accessible in the scope of a closure

integration-tests/utils/integration-helpers.js
```diff
@@ -57,6 +57,21 @@ const createEntityInWorkspace = (page, billingProject, workspaceName, testEntity
...
+    try {
+      await window.Ajax().Workspaces.workspace(billingProject, workspaceName).checkBucketAccess(billingProject, bucketName, accessLevel)
+      return true
+    } catch (e) { return false }
+  }, { timeout: 60000, polling: 10000 }, billingProject, workspaceName, bucketName, accessLevel)
```
aednichols:
```suggestion
  }, { timeout: 60000, polling: 500 }, billingProject, workspaceName, bucketName, accessLevel)
```
Assuming the unit is milliseconds, 10 seconds is a long time IMO. If the bucket becomes available at 3 seconds, we can and should save the users those 7 seconds.

aednichols (changes requested):
Requesting changes based on my understanding of a 10s poll interval

integration-tests/utils/integration-helpers.js
cahrens: It is a 10s polling interval. My rational was that this is something that apparently can take quite a long time to grant (given that the test does several other things before launching the workflow, and yet sometime the access is sometimes not resolved during that time). So I was going for fairly coarse polling intervals and a longer-than-usual timeout to avoid flooding the log with tons of API calls. But sure, we can cut it to 5 seconds.

aednichols (approved):
Very happy to see this flakiness fixed with a production-code solution that also helps users 👍

cahrens:
> Very happy to see this flakiness fixed with a production-code solution that also helps users

Oh no, this change is just test-only. In the test, we are waiting for the bucket write access before we get to the point of launching the workflow.

aednichols (approved):
Ah, didn't realize the helpers file was under a tests directory. 👍 nevertheless

PR #2755, opened 2022-01-07
[DC-102] Add test for search and filter in data explorer

petesantos:
Integration test to ensure search and filter functionality

### DIFF
--- a/integration-tests/tests/run-catalog-filter.js
+++ b/integration-tests/tests/run-catalog-filter.js
@@ -1,0 +1,51 @@
+const _ = require('lodash/fp')
+const { signIntoTerra, click, clickable, waitForNoSpinners, input, findText, fillIn, getNumRowsInTable } = require('../utils/integration-utils')
+const { withUserToken } = require('../utils/terra-sa-utils')
+const { dismissNotifications } = require('../utils/integration-utils')
+
+
+const testCatalogFilterFn = withUserToken(async ({ testUrl, page, token }) => {
+  const searchText = 'stem cell'
+  const filterItem = 'Granted'
+
+  await page.goto(testUrl)
+  await waitForNoSpinners(page)
+
+  await findText(page, 'Browse Data')
+
+  await page.evaluate(() => window.configOverridesStore.set({ isDataBrowserVisible: true }))
+  await page.reload({ waitUntil: ['networkidle0', 'domcontentloaded'] })
+
+  await click(page, clickable({ textContains: 'Browse Data' }))
+  await signIntoTerra(page, token)
+  await dismissNotifications(page)
+
+  await click(page, clickable({ textContains: 'browse & explore' }))
+
+  const totalRows = await getNumRowsInTable(page, 'dataset list')
+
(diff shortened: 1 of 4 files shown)


src/pages/library/common.js
```diff
@@ -237,7 +238,7 @@ export const SearchAndFilterComponent = ({
         alignItems: 'baseline'
       }
     }, [
-      div({ style: styles.sidebarRow }, [
+      h2({ style: { ...styles.sidebarRow, margin: 0 } }, [
```
lu-c: Can we move the `margin: 0` declaration to the styles.sidebarRow instead?

lu-c (approved):
Looks good to me! one nitpick, but non blocking!

integration-tests/tests/run-catalog-filter.js
```diff
@@ -0,0 +1,56 @@
+const _ = require('lodash/fp')
+const { signIntoTerra, click, clickable, waitForNoSpinners, input, findText, fillIn, heading, findHeading } = require('../utils/integration-utils')
+const { withUserToken } = require('../utils/terra-sa-utils')
+const { dismissNotifications } = require('../utils/integration-utils')
+
+
+const getDatasetCount = async page => {
```
dmohs, line 7: Love this helper.

integration-tests/tests/run-catalog-filter.js
```diff
@@ -0,0 +1,56 @@
...
+const testCatalogFilterFn = withUserToken(async ({ testUrl, page, token }) => {
+  const searchText = 'stem cell'
+  const filterItem = 'Granted'
+
+  await page.goto(testUrl)
+  await waitForNoSpinners(page)
+
+  await findText(page, 'Browse Data')
+
+  await page.evaluate(() => window.configOverridesStore.set({ isDataBrowserVisible: true }))
```
dmohs, line 21: Will this throw an error when `isDataBrowserVisible` is no longer a valid configuration option? It would be nice if it did so this code would get cleaned up naturally one day.

integration-tests/utils/integration-utils.js
```diff
@@ -31,8 +31,8 @@ const findInGrid = (page, textContains, options) => {
   return page.waitForXPath(`//*[@role="table"][contains(normalize-space(.),"${textContains}")]`, options)
 }
 
-const clickable = ({ text, textContains }) => {
-  const base = '(//a | //button | //*[@role="button"] | //*[@role="link"])'
+const clickable = ({ text, textContains, isDescendant = false }) => {
```
dmohs, line 34: Fantastic parameter name, thank you.

dmohs (approved):

integration-tests/tests/run-catalog-filter.js
petesantos: No - but this is a good point. I create [DC-176](https://broadworkbench.atlassian.net/browse/DC-176) to address this issue.

I like the idea of these tests failing once the config setting is no longer needed.

PR #2756, opened 2022-01-07
[DC-103] Test request access buttons

samanehsan:
Add a puppeteer test for the Request Access buttons in the Data Catalog.

### DIFF
--- a/integration-tests/src/server.js
+++ b/integration-tests/src/server.js
@@ -9,7 +9,6 @@ const { testImportDockstoreWorkflow } = require('../tests/import-dockstore-workf
 const { testRegisterUser } = require('../tests/register-user')
 const { testRunNotebook } = require('../tests/run-notebook')
 const { testRunWorkflow } = require('../tests/run-workflow')
-const { testRequestAccess } = require('../tests/request-access')
 const { testRunWorkflowOnSnapshot } = require('../tests/run-workflow-on-snapshot')
 const { defaultTimeout } = require('../utils/integration-helpers')
 const { delay, withScreenshot } = require('../utils/integration-utils')
@@ -82,7 +81,6 @@ _.forEach(registerTestEndpoint, [
   testRegisterUser,
   testRunNotebook,
   testRunWorkflow,
-  testRequestAccess,
   testRunWorkflowOnSnapshot
 ])
 console.info('Ready')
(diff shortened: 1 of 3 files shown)


integration-tests/src/server.js
```diff
@@ -81,6 +82,7 @@ _.forEach(registerTestEndpoint, [
   testRegisterUser,
   testRunNotebook,
   testRunWorkflow,
+  testRequestAccess,
```
samanehsan: I'm actually not sure if I need to update this file 🤔

integration-tests/tests/request-access.js
```diff
@@ -0,0 +1,45 @@
...
+  await waitForNoSpinners(page)
+  await click(page, clickable({ textContains: 'Request Access' }))
+  await findText(page, 'Request Access')
+
+})
+
+const testRequestAccess = {
+  name: 'request-access',
+  fn: testRequestAccessFn,
+  timeout: 2 * 60 * 1000
```
fboulnois: I'm guessing this is in ms?

fboulnois (approved):
Overall, lgtm. I wonder if there's a way to express the timeout more clearly (or perhaps a comment?)

lu-c:
Ran through this PR, confirmed it works. The test successfully checks that the modal shows up! Thanks for helping me with this too :)

lu-c (approved):

integration-tests/tests/request-access.js
```diff
@@ -0,0 +1,45 @@
...
+}
```
petesantos, line 44: Can you add:
`targetEnvironments: ['local', 'dev']`
to prevent this from running on environments where the back end is not ready?

integration-tests/src/server.js
petesantos: You can remove this.
This file starts a server that was intended for perf testing. QA could call the tests in rapid succession for testing. I don't think it has been used in a few years.

petesantos (commented):
Looks good! I tested with debug on to view the test steps.
Once the target envs are set this should be ready to go.

integration-tests/tests/request-access.js
```diff
@@ -0,0 +1,45 @@
...
+  await waitForNoSpinners(page)
+  await click(page, clickable({ textContains: 'Controlled' }))
+
+  // Request access from the browse & explore page
+  await click(page, clickable({ textContains: 'Request Access' }))
+  await findText(page, 'Request Access')
+  await click(page, clickable({ textContains: 'Close modal' }))
+
+  // Request access from the dataset details page
+  await clickTableCell(page, "dataset list", 2, 2)
```
petesantos: nit: 
```suggestion
  await clickTableCell(page, 'dataset list', 2, 2)
```

integration-tests/tests/request-access.js
samanehsan: Yes, it's in ms

samanehsan:
I made the changes you mentioned @petesantos! :octocat:

petesantos (approved):
LGTM - thanks!

PR #2757, opened 2022-01-07
[DC-102] Allow tests to run in specific environments

petesantos:
Data explorer is building integration tests, but the application is not flagged to run on production.
This change allows developers to write tests and choose which environments to run the tests on.
Default testing behavior will not change.

### DIFF
--- a/integration-tests/jest/jest-utils.js
+++ b/integration-tests/jest/jest-utils.js
@@ -23,10 +23,7 @@ const registerTest = ({ fn, name, timeout = defaultTimeout, targetEnvironments =
     timeout
   ) : test(
     name,
-    () => {
-      console.log(`Skipping ${name} as it is not configured to run on the ${environment} environment`)
-      new Promise(resolve => resolve())
-    },
+    () => console.log(`Skipping ${name} as it is not configured to run on the ${environment} environment`),
     timeout)
 }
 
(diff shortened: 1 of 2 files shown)


samanehsan (approved):
Lgtm! ☃️

integration-tests/jest/jest-utils.js
```diff
@@ -16,12 +16,18 @@ const {
...
+  return _.includes(environment, targetEnvironments) ? test(
     name,
     () => withPageLogging(withScreenshot(name)(fn))({ context, page, ...targetEnvParams }),
     timeout
-  )
+  ) : test(
+    name,
+    () => {
+      console.log(`Skipping ${name} as it is not configured to run on the ${environment} environment`)
+      new Promise(resolve => resolve())
```
cahrens: Why do you need this promise?

petesantos: Jest is expecting a promise to be returned - I will add a comment

petesantos: Actually, this is doing nothing as I am not returning it.
The error message I originally saw was: 
```    
test functions can only return Promise or undefined.
Returned value: 1
```
I had added the promise because I missed the `undefined` part of the message.
Removed the line - thanks!

cahrens (approved):

kyuksel:
@petesantos I prefixed the PR title with `[DC-102]` for compliance reasons. If that's not the right ticket, please correct it.

[DC-102]: https://broadworkbench.atlassian.net/browse/DC-102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2628, opened 2021-09-16
[IA-2919] Label runtimes with workspace namespace value

kyuksel:
This is the first PR for the ticket [IA-2919](https://broadworkbench.atlassian.net/browse/IA-2919). There will be another PR for the remainder of the work once the [requisite backend work](https://broadworkbench.atlassian.net/browse/IA-2958) is in place. I wanted to get this PR's changes in so we start capturing `workspace namespace` (aka `billing project`) values for runtimes before the One Project Per Workspace (PPW) work is released to `prod` since `workspace namespace` may no longer be the same as `google project` after that release, which would cause Cloud Environments page to display wrong information in `Billing project` column.

Tested against `dev` using a local UI deploy.

PR #2759, opened 2022-01-08
[IA-2919] Enhancements to Cloud Environments (#clusters) page

kyuksel:
This is the part-2 PR for [IA-2919](https://broadworkbench.atlassian.net/browse/IA-2919). The part-1 PR is [here](https://github.com/DataBiosphere/terra-ui/pull/2628).

### Glossary
**Workspace namespace:**
- Terra term
- parent of a `workspace`
- holds billing information and other metadata (e.g. security perimeters)
- not user-facing
- used primarily in backend services

**Billing project:**
- Terra term
- synonymous with _workspace namespace_
- user-facing

**Google project:**
- Google Cloud Platform (GCP) term
- user-facing 

### Changes
- Add `Workspace` column to both (`YOUR CLOUD ENVIRONMENTS` and `YOUR PERSISTENT DISKS`) tables on `Cloud Environments` page.
- Made `Workspace` values link to `Notebooks` tabs of corresponding workspaces for users to more easily navigate.
- Shrink widths of `Created` and `Last accessed` to allow more room to the user-input fields of `Billing project` and `Workspace` columns.
- Add `Google Project` info to `view` pop-ups in `Details` columns for both tables.
- Use `strong` consistently instead of custom-styled `span`s.
- Factored out some duplicated code

#### Before
image

image

#### After
image

image

### DIFF
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -212,7 +212,7 @@ const Environments = () => {
 
   const getWorkspaceCell = (namespace, name, appType, shouldWarn) => {
     return h(Fragment, [
-      h(Link, { onClick: () => Nav.goToPath('workspace-notebooks', { namespace, name }) }, [name]),
+      h(Link, { href: Nav.getLink('workspace-dashboard', { namespace, name }) }, [name]),
       shouldWarn && h(TooltipTrigger, {
         content: `This workspace has multiple active cloud environments${forAppText(appType)}. Only the latest one will be used.`
       }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
@@ -224,6 +224,7 @@ const Environments = () => {
       getCurrentApp(appType)(appsByProject[googleProject]) !== app
     return getWorkspaceCell(saturnWorkspaceNamespace, saturnWorkspaceName, appType, shouldWarn)
   }
+
   const renderWorkspaceForRuntimes = runtime => {
     const { status, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = runtime
     const shouldWarn = !_.includes(status, ['Deleting', 'Error']) &&
@@ -236,7 +237,7 @@ const Environments = () => {
       content: div({ style: { padding: '0.5rem' } }, [
         div([strong(['Name: ']), cloudEnvName]),
         div([strong(['Google Project: ']), googleProject]),
-        disk && div([strong(['Persistent Disk: ']), disk.name])
+        !!disk && div([strong(['Persistent Disk: ']), disk.name])
       ])
     }, [h(Link, ['view'])])
   }
@@ -447,13 +448,14 @@ const Environments = () => {
               const appType = getDiskAppType(filteredDisks[rowIndex])
               const multipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
                 disksByProject[googleProject]).length > 1
-              return h(Fragment,
-                [h(Link, { onClick: () => Nav.goToPath('workspace-notebooks', { namespace: saturnWorkspaceNamespace, name: saturnWorkspaceName }) },
+              return h(Fragment, [
+                h(Link, { href: Nav.getLink('workspace-dashboard', { namespace: saturnWorkspaceNamespace, name: saturnWorkspaceName }) },
                   [saturnWorkspaceName]),
                 diskStatus !== 'Deleting' && multipleDisksOfType &&
-                  h(TooltipTrigger, {
-                    content: `This workspace has multiple active persistent disks${forAppText(appType)}. Only the latest one will be used.`
-                  }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])])
+                h(TooltipTrigger, {
+                  content: `This workspace has multiple active persistent disks${forAppText(appType)}. Only the latest one will be used.`
+                }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
+              ])
             }
           },
           {
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -202,23 +202,26 @@ const Environments = () => {
   const disksByProject = _.groupBy('googleProject', disks)
   const appsByProject = _.groupBy('googleProject', apps)
 
-  const getBillingProjectCell = (billingProject, shouldWarn) => {
+  const forAppText = appType => !!appType ? `for ${_.capitalize(appType)}` : ''
+
+  const getWorkspaceCell = (workspace, appType, shouldWarn) => {
     return h(Fragment, [
-      billingProject,
+      workspace,
       shouldWarn && h(TooltipTrigger, {
-        content: 'This billing project has multiple active cloud environments. Only the most recently created one will be used.'
+        content: `This workspace has multiple active cloud environments ${forAppText(appType)}. Only the latest one will be used.`
       }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
     ])
   }
-  const renderBillingProjectForApps = app => {
-    const shouldWarn = !_.includes(app.status, ['DELETING', 'ERROR', 'PREDELETING']) &&
-      getCurrentApp(app.appType)(appsByProject[app.googleProject]) !== app
-    return getBillingProjectCell(app.labels.saturnWorkspaceNamespace, shouldWarn)
+  const renderWorkspaceForApps = app => {
+    const { status, appType, googleProject, labels: { saturnWorkspaceName } } = app
+    const shouldWarn = !_.includes(status, ['DELETING', 'ERROR', 'PREDELETING']) &&
+      getCurrentApp(appType)(appsByProject[googleProject]) !== app
+    return getWorkspaceCell(saturnWorkspaceName, appType, shouldWarn)
   }
-  const renderBillingProjectForRuntimes = runtime => {
+  const renderWorkspaceForRuntimes = runtime => {
     const shouldWarn = !_.includes(runtime.status, ['Deleting', 'Error']) &&
       getCurrentRuntime(runtimesByProject[runtime.googleProject]) !== runtime
-    return getBillingProjectCell(runtime.labels.saturnWorkspaceNamespace, shouldWarn)
+    return getWorkspaceCell(runtime.labels.saturnWorkspaceName, shouldWarn)
   }
 
   const getDetailsPopup = (cloudEnvName, googleProject, disk) => {
@@ -324,7 +327,7 @@ const Environments = () => {
             headerRenderer: () => h(Sortable, { sort, field: 'billing-project', onSort: setSort }, ['Billing project']),
             cellRenderer: ({ rowIndex }) => {
               const cloudEnvironment = filteredCloudEnvironments[rowIndex]
-              return !!cloudEnvironment.appName ? renderBillingProjectForApps(cloudEnvironment) : renderBillingProjectForRuntimes(cloudEnvironment)
+              return cloudEnvironment.labels.saturnWorkspaceNamespace
             }
           },
           {
@@ -332,7 +335,7 @@ const Environments = () => {
             headerRenderer: () => h(Sortable, { sort, field: 'workspace', onSort: setSort }, ['Workspace']),
             cellRenderer: ({ rowIndex }) => {
               const cloudEnvironment = filteredCloudEnvironments[rowIndex]
-              return cloudEnvironment.labels.saturnWorkspaceName
+              return !!cloudEnvironment.appName ? renderWorkspaceForApps(cloudEnvironment) : renderWorkspaceForRuntimes(cloudEnvironment)
             }
           },
           {
@@ -425,26 +428,25 @@ const Environments = () => {
             field: 'billing-project',
             headerRenderer: () => h(Sortable, { sort: diskSort, field: 'billing-project', onSort: setDiskSort }, ['Billing project']),
             cellRenderer: ({ rowIndex }) => {
-              const { status: diskStatus, googleProject, labels: { saturnWorkspaceNamespace } } = filteredDisks[rowIndex]
-              const appType = getDiskAppType(filteredDisks[rowIndex])
-              const multipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
-                disksByProject[googleProject]).length > 1
-              const forAppText = !!appType ? `for ${_.capitalize(appType)}` : ''
-              return h(Fragment, [
-                saturnWorkspaceNamespace,
-                diskStatus !== 'Deleting' && multipleDisksOfType &&
-                h(TooltipTrigger, {
-                  content: `This billing project has multiple active persistent disks ${forAppText}. Only the latest one will be used.`
-                }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
-              ])
+              const { labels: { saturnWorkspaceNamespace } } = filteredDisks[rowIndex]
+              return saturnWorkspaceNamespace
             }
           },
           {
             field: 'workspace',
             headerRenderer: () => h(Sortable, { sort: diskSort, field: 'workspace', onSort: setDiskSort }, ['Workspace']),
             cellRenderer: ({ rowIndex }) => {
-              const { labels: { saturnWorkspaceName } } = filteredDisks[rowIndex]
-              return h(Fragment, [saturnWorkspaceName])
+              const { status: diskStatus, googleProject, labels: { saturnWorkspaceName } } = filteredDisks[rowIndex]
+              const appType = getDiskAppType(filteredDisks[rowIndex])
+              const multipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
+                disksByProject[googleProject]).length > 1
+              return h(Fragment, [
+                saturnWorkspaceName,
+                diskStatus !== 'Deleting' && multipleDisksOfType &&
+                h(TooltipTrigger, {
+                  content: `This workspace has multiple active persistent disks ${forAppText(appType)}. Only the latest one will be used.`
+                }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
+              ])
             }
           },
           {
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -441,17 +441,17 @@ const Environments = () => {
             field: 'workspace',
             headerRenderer: () => h(Sortable, { sort: diskSort, field: 'workspace', onSort: setDiskSort }, ['Workspace']),
             cellRenderer: ({ rowIndex }) => {
-              const { status: diskStatus, googleProject, labels: { saturnWorkspaceName } } = filteredDisks[rowIndex]
+              const { status: diskStatus, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = filteredDisks[rowIndex]
               const appType = getDiskAppType(filteredDisks[rowIndex])
               const multipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
                 disksByProject[googleProject]).length > 1
-              return h(Fragment, [
-                saturnWorkspaceName,
+              return h(Fragment,
+                [h(Link, { onClick: () => Nav.goToPath('workspace-notebooks', { namespace: saturnWorkspaceNamespace, name: saturnWorkspaceName }) },
+                  [saturnWorkspaceName]),
                 diskStatus !== 'Deleting' && multipleDisksOfType &&
-                h(TooltipTrigger, {
-                  content: `This workspace has multiple active persistent disks ${forAppText(appType)}. Only the latest one will be used.`
-                }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
-              ])
+                  h(TooltipTrigger, {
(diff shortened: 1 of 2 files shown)


kyuksel:
@gpcarr Heads-up for potential merge conflicts if this PR goes in before yours but they should hopefully be straightforward :)

src/pages/Environments.js
```diff
@@ -202,48 +202,43 @@ const Environments = () => {
...
-  const renderBillingProjectApp = app => {
-    const inactive = !_.includes(app.status, ['DELETING', 'ERROR', 'PREDELETING']) &&
-      getCurrentApp(app.appType)(appsByProject[app.googleProject]) !== app
+  const getBillingProjectCell = (billingProject, shouldWarn) => {
     return h(Fragment, [
-      app.googleProject,
-      inactive && h(TooltipTrigger, {
+      billingProject,
+      shouldWarn && h(TooltipTrigger, {
         content: 'This billing project has multiple active cloud environments. Only the most recently created one will be used.'
```
cahrens: All of your workspaces can be linked to the same billing project, so this message doesn't make sense. Did we decide it should read "This Workspace has multiple active…"?

cahrens: Looking at https://github.com/DataBiosphere/terra-ui/pull/2759/files#diff-db07f65dd0ba5edbd535b2a231709c1f5901f4e1c43f39c333165d97c1abacc4R215, I guess the message should technically be "This Google Project has multiple active cloud environments", but the Google Project is a detail that most users are not aware of. And with PPW, they are usually the same thing.

src/pages/Environments.js
```diff
@@ -417,39 +420,44 @@ const Environments = () => {
...
                 disksByProject[googleProject]).length > 1
-              const forAppText = !!appType ? ` for ${_.capitalize(appType)}` : ''
+              const forAppText = !!appType ? `for ${_.capitalize(appType)}` : ''
               return h(Fragment, [
-                googleProject,
+                saturnWorkspaceNamespace,
                 diskStatus !== 'Deleting' && multipleDisksOfType &&
                 h(TooltipTrigger, {
-                  content: `This billing project has multiple active persistent disks${forAppText}. Only the latest one will be used.`
+                  content: `This billing project has multiple active persistent disks ${forAppText}. Only the latest one will be used.`
```
cahrens: This message should be updated to "This Workspace…".

src/pages/Environments.js
cahrens: This message is also confusing because it's fine to have "multiple active cloud environments" in the same workspace… or Google Project. The problem comes when you have multiple ones of the same TYPE. Like, 2 Galaxy, 2 Cromwell, or 2 GCE in the same Workspace/Google Project. Can we change the message to include that detail like I did in the the multiple persistent disk message?

src/pages/Environments.js
```diff
@@ -202,48 +202,43 @@ const Environments = () => {
...
-
-  const renderBillingProjectRuntime = runtime => {
-    const inactive = !_.includes(runtime.status, ['Deleting', 'Error']) &&
+  const renderBillingProjectForApps = app => {
+    const shouldWarn = !_.includes(app.status, ['DELETING', 'ERROR', 'PREDELETING']) &&
+      getCurrentApp(app.appType)(appsByProject[app.googleProject]) !== app
+    return getBillingProjectCell(app.labels.saturnWorkspaceNamespace, shouldWarn)
+  }
+  const renderBillingProjectForRuntimes = runtime => {
+    const shouldWarn = !_.includes(runtime.status, ['Deleting', 'Error']) &&
```
cahrens: Is the case really 'Deleting' for runtimes (instead of all caps like it is for apps)?

We really need some constants or case-insensitive helper methods to check statuses. It's really confusing because the disk statuses are mixed case, but the App ones are all uppercase.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -179,7 +179,7 @@ const useCloudEnvironmentPolling = googleProject => {
   const load = async maybeStale => {
     try {
       const [newDisks, newRuntimes] = googleProject ? await Promise.all([
-        Ajax(signal).Disks.list({ googleProject, creator: getUser().email, includeLabels: 'saturnApplication' }),
+        Ajax(signal).Disks.list({ googleProject, creator: getUser().email, includeLabels: 'saturnApplication,saturnWorkspaceNamespace,saturnWorkspaceName' }),
```
cahrens: Are these actually needed here? Environments.js does its own GET.

src/pages/Environments.js
cahrens: I double-checked in the debugger, and yep, it's "Deleting…" here. Ugh, the inconsistency is infuriating.

src/pages/Environments.js
```diff
@@ -417,39 +420,44 @@ const Environments = () => {
...
                 h(TooltipTrigger, {
-                  content: `This billing project has multiple active persistent disks${forAppText}. Only the latest one will be used.`
+                  content: `This billing project has multiple active persistent disks ${forAppText}. Only the latest one will be used.`
                 }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
               ])
             }
           },
+          {
+            field: 'workspace',
+            headerRenderer: () => h(Sortable, { sort: diskSort, field: 'workspace', onSort: setDiskSort }, ['Workspace']),
```
cahrens: The sorting isn't working correctly for me on this page. I think the problem for "Type" may be that the sorting is only being done within apps and runtimes, which is very unexpected. And for Workspace, it doesn't seem to be doing anything at all.

image

image

cahrens:
I'm wondering of Workspace should be promoted to the first column. Would most users have all their workspaces within the same billing project? Do they even really think about billing project, or do they think primarily about Workspaces?

I suppose that a counterargument is the breadcrumb/URL ordering, which has billing project first, so I don't feel strongly about this.

However, it would be AWESOME if we could make the Workspace entry a link to the Workspace. Going back and forth between the Cloud Environments page and the Workspaces is pretty tedious.

src/pages/Environments.js
```diff
@@ -202,48 +202,43 @@ const Environments = () => {
...
         content: 'This billing project has multiple active cloud environments. Only the most recently created one will be used.'
       }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
     ])
   }
-
-  const renderBillingProjectRuntime = runtime => {
-    const inactive = !_.includes(runtime.status, ['Deleting', 'Error']) &&
+  const renderBillingProjectForApps = app => {
+    const shouldWarn = !_.includes(app.status, ['DELETING', 'ERROR', 'PREDELETING']) &&
+      getCurrentApp(app.appType)(appsByProject[app.googleProject]) !== app
```
petesantos: This is a bit out of scope for this PR, but this is confusing for me. I had to trace this through a few levels of code to find out where `app` was coming from, and `getCurrentApp` is not clear to me

src/pages/Environments.js
```diff
@@ -202,48 +202,43 @@ const Environments = () => {
...
-        div([span({ style: { fontWeight: '600' } }, ['Name: ']), appName]),
-        disk && div([span({ style: { fontWeight: '600' } }, ['Persistent Disk: ']), disk.name])
+        div([strong(['Name: ']), cloudEnvName]),
+        div([strong(['Google Project: ']), googleProject]),
+        disk && div([strong(['Persistent Disk: ']), disk.name])
       ])
     }, [h(Link, ['view'])])
   }
-
+  const renderDetailsApp = (app, disks) => {
```
petesantos, line 245: nit: add line break

src/pages/Environments.js
```diff
@@ -202,48 +202,43 @@ const Environments = () => {
...
+    const { appName, diskName, googleProject } = app
+    const disk = _.find({ name: diskName }, disks)
+    return getDetailsPopup(appName, googleProject, disk)
+  }
```
petesantos, line 249: nit: add line break after this line

Qi77Qi (approved):
this is great...was testing yesterday but got sidetracked by prod issue...but really excited for this!

src/pages/Environments.js
petesantos: I am seeing the same thing with sorting. It seems to be incorrect on several fields. It also seems like the last row is not being included in the sort

src/pages/Environments.js
kyuksel: Thanks for pointing this out. Does the following sound correct to you?

> `The workspace has multiple active cloud environments ${forAppText}. Only the most recently created one will be used.`

Assuming I'm not missing something, I think we'd also want to move the warning to the newly added `Workspace` column.

Note that a `Google project` _can_ have multiple active cloud environments pre-PPW so I don't think we'd want to warn users for that case.

src/pages/Environments.js
kyuksel: 💯

src/pages/workspaces/workspace/WorkspaceContainer.js
kyuksel: No. Thanks for the catch!

kyuksel:
> I'm wondering of Workspace should be promoted to the first column. Would most users have all their workspaces within the same billing project? Do they even really think about billing project, or do they think primarily about Workspaces? I suppose that a counterargument is the breadcrumb/URL ordering, which has billing project first, so I don't feel strongly about this.

Good questions. I could see both ways too so I'll keep it unchanged for the time being.

> it would be AWESOME if we could make the Workspace entry a link to the Workspace. Going back and forth between the Cloud Environments page and the Workspaces is pretty tedious.

Great suggestion! I'll add it to this PR unless I encounter something unexpected doing that.

src/pages/Environments.js
kyuksel: Yea I see what you're saying. I think part of it is due to the complicated history of how many apps/runtimes/disks we've allowed per billing project/google project/workspace post- and pre-PPW (one project per workspace). If we can deprecate some cases in the future, it'd be good to try to simplify this logic.

src/pages/Environments.js
kyuksel: I meant to cluster the closely related functions of `getDetailsPopup`, `renderDetailsApp`, `renderDetailsRuntime` for easier parsing, thinking that we do in other places in the codebase too but maybe not. I don't feel strongly either way so happy to add the line breaks.

src/pages/Environments.js
cahrens: Yes, I think that it would move to the Workspace column. And this is one of the places where "cloud environments" gets confusing for apps vs. runtimes. For apps (Galaxy, Cromwell), we are really talking about multiple instances of the app itself… they technically both run in the same kubernetes cluster, which I guess is the cloud environment? For Jupyter/RStudio, we are talking about multiple GCE environments. Coming up with a single wording that works for both might be a challenge.

src/pages/Environments.js
cahrens: And it would be good to move it to the Warning column.

src/pages/Environments.js
kyuksel: `Cloud Environment` is meant to be an abstraction over backend services like `GKE` and `GCE` to encompass application configuration, cloud compute and persistent disk(s)…so multiple instances of an app running on the same K8s cluster would still be multiple cloud environments.

src/pages/Environments.js
kyuksel: Ugh yea…we seem to have accumulated regressions over time. I've added sorting for the new `Workspace` column, but would prefer to file a bug to address the pre-existing issues properly (with unit testing if possible) since sorting is no worse as part of this PR.

kyuksel:
@cahrens @petesantos @Qi77Qi 
Thank you for your comments. I think I've addressed them. Could you take another look?

src/pages/Environments.js
kyuksel: Ticket created: https://broadworkbench.atlassian.net/browse/IA-3166

Qi77Qi (approved):
looks great :shipit:

src/pages/Environments.js
```diff
@@ -202,23 +202,26 @@ const Environments = () => {
...
+    const shouldWarn = !_.includes(status, ['DELETING', 'ERROR', 'PREDELETING']) &&
+      getCurrentApp(appType)(appsByProject[googleProject]) !== app
+    return getWorkspaceCell(saturnWorkspaceName, appType, shouldWarn)
   }
-  const renderBillingProjectForRuntimes = runtime => {
+  const renderWorkspaceForRuntimes = runtime => {
     const shouldWarn = !_.includes(runtime.status, ['Deleting', 'Error']) &&
       getCurrentRuntime(runtimesByProject[runtime.googleProject]) !== runtime
-    return getBillingProjectCell(runtime.labels.saturnWorkspaceNamespace, shouldWarn)
+    return getWorkspaceCell(runtime.labels.saturnWorkspaceName, shouldWarn)
```
cahrens: Here you are not passing appType (which makes sense because it's not an app. But that means your arguments to getWorkspaceCell are incorrect.

src/pages/Environments.js
```diff
@@ -202,23 +202,26 @@ const Environments = () => {
   const disksByProject = _.groupBy('googleProject', disks)
   const appsByProject = _.groupBy('googleProject', apps)
 
-  const getBillingProjectCell = (billingProject, shouldWarn) => {
+  const forAppText = appType => !!appType ? `for ${_.capitalize(appType)}` : ''
```
cahrens: Note that appType will be undefined for runtimes (Jupyter/RStudio). That's why I had the space before for originally, so that this would work for both apps and runtimes. A comment to this effect would be a good idea.

src/pages/Environments.js
```diff
@@ -420,36 +433,40 @@ const Environments = () => {
...
+              return h(Fragment,
+                [h(Link, { onClick: () => Nav.goToPath('workspace-notebooks', { namespace: saturnWorkspaceNamespace, name: saturnWorkspaceName }) },
+                  [saturnWorkspaceName]),
                 diskStatus !== 'Deleting' && multipleDisksOfType &&
-                h(TooltipTrigger, {
-                  content: `This billing project has multiple active persistent disks${forAppText}. Only the latest one will be used.`
-                }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
-              ])
+                  h(TooltipTrigger, {
+                    content: `This workspace has multiple active persistent disks ${forAppText(appType)}. Only the latest one will be used.`
```
cahrens: As I mentioned in the other comment, the lack of space before "${forAppText" was intentional to handle the non-app case.

src/pages/Environments.js
```diff
@@ -202,48 +206,49 @@ const Environments = () => {
...
+    const { status, appType, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = app
+    const shouldWarn = !_.includes(status, ['DELETING', 'ERROR', 'PREDELETING']) &&
+      getCurrentApp(appType)(appsByProject[googleProject]) !== app
+    return getWorkspaceCell(saturnWorkspaceNamespace, saturnWorkspaceName, appType, shouldWarn)
+  }
+  const renderWorkspaceForRuntimes = runtime => {
+    const { status, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = runtime
+    const shouldWarn = !_.includes(status, ['Deleting', 'Error']) &&
+      getCurrentRuntime(runtimesByProject[googleProject]) !== runtime
+    return getWorkspaceCell(saturnWorkspaceNamespace, saturnWorkspaceName, shouldWarn)
```
cahrens: I think my comment got lost, but note that the third argument to getWorkspaceCell should be appType.

kyuksel:
@cahrens Thank you for the catches. Hopefully I got them right this time.

src/pages/Environments.js
cahrens: Even in the new Workspace column, sorting isn't working properly between runtimes and apps. I think that is expected and will be covered by IA-3166, but just wanted to confirm.

image

src/pages/Environments.js
```diff
@@ -202,48 +206,51 @@ const Environments = () => {
...
 
-  const renderBillingProjectRuntime = runtime => {
-    const inactive = !_.includes(runtime.status, ['Deleting', 'Error']) &&
-      getCurrentRuntime(runtimesByProject[runtime.googleProject]) !== runtime
+  const getWorkspaceCell = (namespace, name, appType, shouldWarn) => {
     return h(Fragment, [
-      runtime.googleProject,
-      inactive && h(TooltipTrigger, {
-        content: 'This billing project has multiple active cloud environments. Only the most recently created one will be used.'
+      h(Link, { onClick: () => Nav.goToPath('workspace-notebooks', { namespace, name }) }, [name]),
```
cahrens: I expected this to just go to the Workspace (dashboard), not the Notebooks tab.

src/pages/Environments.js
```diff
@@ -202,48 +206,51 @@ const Environments = () => {
...
       }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
     ])
   }
+  const renderWorkspaceForApps = app => {
+    const { status, appType, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = app
+    const shouldWarn = !_.includes(status, ['DELETING', 'ERROR', 'PREDELETING']) &&
+      getCurrentApp(appType)(appsByProject[googleProject]) !== app
+    return getWorkspaceCell(saturnWorkspaceNamespace, saturnWorkspaceName, appType, shouldWarn)
+  }
+  const renderWorkspaceForRuntimes = runtime => {
```
cahrens, line 228: Nit: a blank line in-between would be nice.

src/pages/Environments.js
```diff
@@ -202,48 +206,51 @@ const Environments = () => {
...
-    const { appName, diskName } = app
-    const disk = _.find({ name: diskName }, disks)
+  const getDetailsPopup = (cloudEnvName, googleProject, disk) => {
     return h(PopupTrigger, {
       content: div({ style: { padding: '0.5rem' } }, [
-        div([span({ style: { fontWeight: '600' } }, ['Name: ']), appName]),
-        disk && div([span({ style: { fontWeight: '600' } }, ['Persistent Disk: ']), disk.name])
+        div([strong(['Name: ']), cloudEnvName]),
+        div([strong(['Google Project: ']), googleProject]),
+        disk && div([strong(['Persistent Disk: ']), disk.name])
```
cahrens: I personally prefer this syntax over `!!disk &&…`, but I know you have been trying to move us in the direction of using the !!.

src/pages/Environments.js
cahrens: Also, instead of suppling onClick, can you do the following?

`h(Link, { href: Nav.getLink('workspace-notebooks', { namespace, name }) }, [name]),`

The reason is that my suggested code will render an anchor tag, whereas yours renders a div with aria-role button. This link is taking you someplace, not opening a popup/dialog. Therefore the anchor tag is more meaningful and will behave correctly with assistive technologies.

src/pages/Environments.js
```diff
@@ -420,36 +435,40 @@ const Environments = () => {
...
+            cellRenderer: ({ rowIndex }) => {
+              const { status: diskStatus, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = filteredDisks[rowIndex]
               const appType = getDiskAppType(filteredDisks[rowIndex])
               const multipleDisksOfType = _.remove(disk => getDiskAppType(disk) !== appType || disk.status === 'Deleting',
                 disksByProject[googleProject]).length > 1
-              const forAppText = !!appType ? ` for ${_.capitalize(appType)}` : ''
-              return h(Fragment, [
-                googleProject,
+              return h(Fragment,
+                [h(Link, { onClick: () => Nav.goToPath('workspace-notebooks', { namespace: saturnWorkspaceNamespace, name: saturnWorkspaceName }) },
```
cahrens: Same comments as above about going to the notebooks tab vs. the Workspace dashboard, and the desire for an a-tag instead of a "button".

src/pages/Environments.js
kyuksel: That is correct.

src/pages/Environments.js
kyuksel: My assumption was, if someone is on `Cloud Environments` page then the chances are, they're using interactive analysis tools, and would want to go to the corresponding (`Notebooks`) tab. It feels `Dashboard` tends to be less relevant for interactive analysis purposes. I can take this up with our UX/PO and change it if they also suggest going to `Dashboard`. What do you think?

src/pages/Environments.js
petesantos: Should this be `disk` or  `disk?.name`
I prefer the boolean coercion as it will prevent accidental rendering of some falsey values

kyuksel: Thanks. Changing for consistency.

src/pages/Environments.js
cahrens: Sure, run it by UX. My thought is that I don't use the Notebooks tab… I only use the hidden Analysis tab. To me, just getting to the correct Workspace is a big win (though before part of the pain was that the names were obscured).

For the purposes of getting this to "done", feel free to remove the link altogether.

src/pages/Environments.js
cahrens: I think it should be `disk` because disks should always have names.

petesantos (approved):
+1 to Christina's comments on the links/a tags vs onClick

src/pages/Environments.js
kyuksel: I love the link! I'll have them point to `Dashboard` for the time being. 

Thanks for the `href` comment.

kyuksel:
@cahrens I made all of the suggested changes. Any chance you could take another look?

cahrens (approved):
These are some great improvements. Thank you!

PR #2760, opened 2022-01-11
add copy RAS passport button [post- POL-143]

dvoet:
This loads the passport of a user who has linked a ras account and adds a button to copy it to the clipboard on the profile page. Again, this is only for devs to use; it won't be there once this makes it to real users.


dmohs (approved):
Love tools like this. I wish our UIs had more of them in general.

zarsky-broad (approved):

PR #2762, opened 2022-01-12
[DC-177] Make filter items accessible

petesantos:
No visible change. This changes the filter items to have a role of `checkbox` as this is closer to what the application is doing and will announce to a screen reader whether a filtered item has been selected.

### DIFF
--- a/src/pages/library/common.js
+++ b/src/pages/library/common.js
@@ -68,7 +68,11 @@ const FilterSection = ({ onTagFilter, labels, selectedTags, labelRenderer, listD
   return h(Fragment, [
     _.map(label => {
       const lowerTag = _.toLower(label)
+      const checked = _.includes(lowerTag, lowerSelectedTags)
+      const numMatches = _.size(listDataByTag[lowerTag])
       return h(Clickable, {
+        'aria-checked': checked,
+        role: 'checkbox',
         key: label,
         style: {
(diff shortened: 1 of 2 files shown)


src/pages/library/common.js
```diff
@@ -77,7 +81,7 @@ const FilterSection = ({ onTagFilter, labels, selectedTags, labelRenderer, listD
         onClick: () => onTagFilter({ lowerTag, label })
       }, [
         div({ style: { lineHeight: '1.375rem', flex: 1 } }, [...(labelRenderer ? labelRenderer(label) : label)]),
-        div({ style: styles.pill(_.includes(lowerTag, lowerSelectedTags)) }, [_.size(listDataByTag[lowerTag])])
+        div({ 'aria-label': `${numMatches} matches`, style: styles.pill(isChecked) }, [numMatches])
```
petesantos, line 84: Added a label for a better reader experience. When navigating the page, the reader will announce (as an example):

"Epigenomic 4 matches unchecked"
rather than 
"Epigenomic 4 unchecked"

lu-c: this makes sense!

lu-c (approved):
Looks good to me!

dmohs (approved):
Very nice.

PR #2764, opened 2022-01-12
[DC-110] Updated style on selected filters to have higher contrast

lu-c:
Before
Screen Shot 2022-01-12 at 1 43 47 PM

After
Screen Shot 2022-01-12 at 1 43 37 PM

Increased contrast, bolded text on select

### DIFF
--- a/integration-tests/tests/run-catalog-filter.js
+++ b/integration-tests/tests/run-catalog-filter.js
@@ -1,5 +1,5 @@
 const _ = require('lodash/fp')
-const { signIntoTerra, click, clickable, waitForNoSpinners, input, findText, fillIn, heading, findHeading } = require('../utils/integration-utils')
+const { signIntoTerra, click, clickable, waitForNoSpinners, input, findInDataTableRow, findText, fillIn, heading, findHeading } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 const { dismissNotifications } = require('../utils/integration-utils')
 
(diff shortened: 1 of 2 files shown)


petesantos (approved):
LGTM - I am assuming the test diff will go away once 2763 is merged

lu-c:
> LGTM - I am assuming the test diff will go away once 2763 is merged

Oh shoot, i didnt notice i didnt create a new branch for this >.< Thanks! Yeah, sorry about that!

integration-tests/tests/run-catalog-filter.js
```diff
@@ -38,12 +39,27 @@ const testCatalogFilterFn = withUserToken(async ({ testUrl, page, token }) => {
...
+  await click(page, clickable({ text: 'Controlled', isDescendant: true }))
+  const datasetSizeAfterFilter2 = await getDatasetCount(page)
+  if (datasetSizeAfterFilter2 !== 0) {
+    throw new Error(`Filters for 'Controlled' should be ANDed between the same facet category in the table'`)
+  }
+
+  // Testing clearing filters
+  await click(page, clickable({ textContains: 'clear' }))
+  const datasetSizeAfterClear = await getDatasetCount(page)
+  if (datasetSizeAfterClear !== datasetSizeAfterSearch) {
```
samanehsan, line 60: Should this throw an error if `datasetSizeAfterClear !== totalDatasetSize` instead?

lu-c: The text filter is still applied at this point of the unit tests, so this test matches the behavior of the application. I guess the question we should ask is whether or not the clear button should also clear out the search text too?

samanehsan: Oh I see! Then in the error message, should it include the `datasetSizeAfterSearch` instead of the `totalDatasetSize`?

samanehsan: > I guess the question we should ask is whether or not the clear button should also clear out the search text too?

Good point! Maybe @lstrano-broad has some insight?

lu-c: Ah, good point, I'll update the error message!

fboulnois (approved):

PR #2765, opened 2022-01-12
[No ticket] Aria-live update to regions of the catalog that experience data updates without loading a new page

lu-c:
No changes in the UI, but adds the aria label to assist accessibility users in letting them know the content has been updated when changes occur

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

### DIFF
--- a/src/pages/library/DataBrowserPreview.js
+++ b/src/pages/library/DataBrowserPreview.js
@@ -169,7 +169,7 @@ const DataBrowserPreview = ({ id }) => {
         loading ?
           centeredSpinner() :
           div({ style: { position: 'relative', padding: '0 15px' } }, [
-            div({ role: 'status', 'aria-label': `${selectedTable} Preview Data Table ${loading ? 'loading' : 'loaded'}` }, []),
+            div({ role: 'status', 'aria-label': `${selectedTable} Preview Data Table loaded` }, []),
             h(SimpleTable, {
               'aria-label': `${_.startCase(selectedTable)} Preview Data`,
               columns: _.filter('visible', columns),
--- a/src/pages/library/DataBrowserPreview.js
+++ b/src/pages/library/DataBrowserPreview.js
@@ -171,7 +171,7 @@ const DataBrowserPreview = ({ id }) => {
           div({ style: { position: 'relative', padding: '0 15px' } }, [
             div({ role: 'status', 'aria-label': `${selectedTable} Preview Data Table ${loading ? 'loading' : 'loaded'}` }, []),
             h(SimpleTable, {
-              // 'aria-label': `${_.startCase(selectedTable)} Preview Data`,
+              'aria-label': `${_.startCase(selectedTable)} Preview Data`,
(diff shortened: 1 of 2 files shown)


cahrens:
I am very happy to see you paying attention to this level of detail! If you haven't already tested this with a screen reader, please do. I have found in the past that it can be tricky to get aria-live regions to behave correctly.

lu-c:
> I am very happy to see you paying attention to this level of detail! If you haven't already tested this with a screen reader, please do. I have found in the past that it can be tricky to get aria-live regions to behave correctly.

Thanks for the recommendation! I turned on the screenreader, and it's.... kind of messy. Im looking at how other people handle this, and the general solution is to read out how the search results have changed, but not the search results themselves. I'll update this PR to reflect that!

samanehsan (approved):

samanehsan:
Should we create a ticket for this change? I know that's a requirement on the data repo core team, but I'm not sure about the data catalog! 🎟️

src/pages/library/common.js
```diff
@@ -240,7 +240,7 @@ export const SearchAndFilterComponent = ({
     }, [
       h2({ style: { ...styles.sidebarRow } }, [
         div({ style: styles.header }, [searchType]),
-        div({ style: styles.pill(_.isEmpty(selectedSections) && _.isEmpty(selectedTags)) }, [_.size(filteredData)])
+        div({ style: styles.pill(_.isEmpty(selectedSections) && _.isEmpty(selectedTags)), role: 'status', 'aria-label': `${_.size(filteredData)} Results found` }, [_.size(filteredData)])
```
petesantos, line 243: This works nicely in the screen reader!

petesantos (approved):

src/pages/library/DataBrowserPreview.js
```diff
@@ -169,6 +169,7 @@ const DataBrowserPreview = ({ id }) => {
         loading ?
           centeredSpinner() :
           div({ style: { position: 'relative', padding: '0 15px' } }, [
+            div({ role: 'status', 'aria-label': `${selectedTable} Preview Data Table loaded` }, []),
```
petesantos, line 172: I found putting this as a sibling next to [h1](https://github.com/DataBiosphere/terra-ui/blob/4fb7efc68855a84eafdc4cbcbd7ac336348afb4f/src/pages/library/DataBrowserPreview.js#L137)yielded a better screen reader result - the announcement seemed to occur with greater consistency.

It looks like most of the noise is coming from our combo box.

petesantos:
> Should we create a ticket for this change? I know that's a requirement on the data repo core team, but I'm not sure about the data catalog! 🎟️

I think we do need a ticket

src/pages/library/DataBrowserPreview.js
lu-c: That combobox is REALLY noisy, but I don't really know what we can do about it!

PR #2766, opened 2022-01-13
[DC-104] Add test for dataset preview page

samanehsan:
Add a puppeteer test for the dataset preview page.

### DIFF
--- a/integration-tests/tests/preview-dataset.js
+++ b/integration-tests/tests/preview-dataset.js
@@ -22,9 +22,10 @@ const testPreviewDatasetFn = _.flow(
 
   await click(page, clickable({ textContains: 'browse & explore' }))
   await waitForNoSpinners(page)
-  await click(page, clickable({ textContains: 'Granted' }))
-  await click(page, clickable({ textContains: datasetName }))
+  await click(page, clickable({ text: 'Granted', isDescendant: true, isCheckbox: true }))
+  await click(page, clickable({ textContains: `${datasetName}` }))
   await waitForNoSpinners(page)
+  await findText(page, 'Contributors')
   await click(page, clickable({ textContains: 'Preview data' }))
   await waitForNoSpinners(page)
 
--- a/integration-tests/tests/preview-dataset.js
+++ b/integration-tests/tests/preview-dataset.js
@@ -1,5 +1,5 @@
 const _ = require('lodash/fp')
-const { signIntoTerra, click, clickable, findText, findTableCellText, getTableCellPath, getTableHeaderPath, waitForNoSpinners } = require('../utils/integration-utils')
+const { signIntoTerra, checkbox, click, clickable, findText, findTableCellText, getTableCellPath, getTableHeaderPath, waitForNoSpinners } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 const { dismissNotifications } = require('../utils/integration-utils')
 
@@ -22,7 +22,7 @@ const testPreviewDatasetFn = _.flow(
 
   await click(page, clickable({ textContains: 'browse & explore' }))
   await waitForNoSpinners(page)
-  await click(page, clickable({ text: 'Granted', isDescendant: true, isCheckbox: true }))
+  await click(page, checkbox({ text: 'Granted', isDescendant: true}))
   await click(page, clickable({ textContains: `${datasetName}` }))
   await waitForNoSpinners(page)
   await findText(page, 'Contributors')
--- a/integration-tests/tests/preview-dataset.js
+++ b/integration-tests/tests/preview-dataset.js
@@ -1,0 +1,75 @@
+const _ = require('lodash/fp')
+const { signIntoTerra, click, clickable, clickTableCell, findElement, findText, findTableCellText, getTableCellPath, getTableHeaderPath, input, waitForNoSpinners } = require('../utils/integration-utils')
+const { withWorkspace } = require('../utils/integration-helpers')
+const { withUserToken } = require('../utils/terra-sa-utils')
+const { dismissNotifications } = require('../utils/integration-utils')
+
+const datasetName = "Cell hashing with barcoded antibodies enables multiplexing and doublet detection for single cell genomics";
+
+const assertEquals = (expected, actual) => {
+    if (expected !== actual) {
+      throw new Error("Expected value " + expected + " does not equal actual " + actual)
+    }
+}
+
+const testPreviewDatasetFn = _.flow(
+  withWorkspace,
+  withUserToken
+)(async ({ testUrl, page, token, workspaceName }) => {
+  await page.goto(testUrl)
+  await waitForNoSpinners(page)
+
+  await findText(page, 'Browse Data')
+
+  await page.evaluate(() => window.configOverridesStore.set({ isDataBrowserVisible: true }))
+  await page.reload({ waitUntil: ['networkidle0', 'domcontentloaded'] })
+
+  await click(page, clickable({ textContains: 'Browse Data' }))
+  await signIntoTerra(page, token)
+  await dismissNotifications(page)
+
+  await click(page, clickable({ textContains: 'browse & explore' }))
+  await waitForNoSpinners(page)
+  await click(page, clickable({ textContains: 'Granted' }))
+  await click(page, clickable({ textContains: datasetName }))
+  await waitForNoSpinners(page)
+  await click(page, clickable({ textContains: 'Preview data' }))
+  await waitForNoSpinners(page)
+
+  // Verify number of tables
+  const tableName = 'Cell Suspension'
+
+  await click(page, '//*[@role="combobox"]//div')
+  const tablesWithData = await page.$x('//*[@role="combobox"]//*[contains(@id, "option-0")]')
+  const countTablesWithData = _.size(tablesWithData)
+  assertEquals(countTablesWithData, 11)
+  const tablesWithNoData = await page.$x('//*[@role="combobox"]//*[contains(@id, "option-1")]')
+  const countTablesWithNoData = _.size(tablesWithNoData)
+  assertEquals(countTablesWithNoData, 16)
+
+  //  const previewTableName = 'Cell Suspension Preview Data'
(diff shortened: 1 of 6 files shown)


integration-tests/tests/preview-dataset.js
```diff
@@ -0,0 +1,74 @@
...
+  const tableNameNoData = "Analysis File"
+  await click(page, `//*[@role="listbox"]//*[@role="option" and contains(normalize-space(.),"${tableNameNoData}")]`)
+  await waitForNoSpinners(page)
+  await findText(page, 'No Data')
+
+  // Verify table counts
+  await click(page, '//*[@role="combobox"]//div')
+  const tablesWithData = await page.$x('//*[@role="combobox"]//*[contains(@id, "option-0")]')
+  const countTablesWithData = _.size(tablesWithData)
+  assertEquals(countTablesWithData, 11)
```
dmohs: I think this usage feels intuitive, which means your parameters need to be reversed in the definition above.

integration-tests/tests/preview-dataset.js
```diff
@@ -0,0 +1,74 @@
...
+const { withUserToken } = require('../utils/terra-sa-utils')
+const { dismissNotifications } = require('../utils/integration-utils')
+
+const datasetName = "Cell hashing with barcoded antibodies enables multiplexing and doublet detection for single cell genomics";
+
+const assertEquals = (expected, actual) => {
+    if (expected !== actual) {
+      throw new Error("Expected value " + expected + " does not equal actual " + actual)
+    }
+}
```
dmohs:
```suggestion
const assertEquals = (expected, actual) => {
  if (expected !== actual) {
    throw new Error("Expected value " + expected + " does not equal actual " + actual)
  }
}
```

integration-tests/tests/preview-dataset.js
```diff
@@ -0,0 +1,74 @@
+const _ = require('lodash/fp')
+const { signIntoTerra, click, clickable, clickTableCell, findElement, findText, findTableCellText, getTableCellPath, getTableHeaderPath, input, waitForNoSpinners } = require('../utils/integration-utils')
+const { withWorkspace } = require('../utils/integration-helpers')
+const { withUserToken } = require('../utils/terra-sa-utils')
+const { dismissNotifications } = require('../utils/integration-utils')
+
+const datasetName = "Cell hashing with barcoded antibodies enables multiplexing and doublet detection for single cell genomics";
```
dmohs:
```suggestion
const datasetName = 'Cell hashing with barcoded antibodies enables multiplexing and doublet detection for single cell genomics'
```

integration-tests/tests/preview-dataset.js
```diff
@@ -0,0 +1,74 @@
...
+  await click(page, clickable({ textContains: datasetName }))
+  await waitForNoSpinners(page)
+  await click(page, clickable({ textContains: 'Preview data' }))
+  await waitForNoSpinners(page)
+
+  const previewTableName = 'Cell Suspension Preview Data'
+  await findTableCellText(page, getTableHeaderPath(previewTableName, 1), 'cell_suspension_id')
+  await findTableCellText(page, getTableHeaderPath(previewTableName, 2), 'version')
+  await findTableCellText(page, getTableHeaderPath(previewTableName, 3), 'content')
+  await findTableCellText(page, getTableCellPath(previewTableName, 2, 1), 'f0caec4a-2a37-4895-8304-83d0fd0da588')
```
dmohs, line 36: If this data were to change, would it be appropriate for the test to fail, or does this make it a bit fragile?

integration-tests/tests/preview-dataset.js
```diff
@@ -0,0 +1,74 @@
...
+  await findTableCellText(page, getTableCellPath(previewTableName, 2, 2), '1558104905.862')
```
dmohs, line 37: Same question as previous line.

integration-tests/tests/preview-dataset.js
```diff
@@ -0,0 +1,74 @@
...
+  await findTableCellText(page, getTableCellPath(previewTableName, 2, 3), 'View JSON')
+  await click(page, clickable({ textContains: 'View JSON'  }))
+  await findElement(page, '//div[contains(@class, "react-json-view")]')
```
dmohs: We try as often as possible to make our tests check for things the way a user would check for them to aid accessibility. That is why you shouldn't see checks in our tests that verify a class name. Is there something more user-visible that you can verify here?

dmohs: `findText` may be the right choice here, for example.

dmohs (commented):
Pete and I paired on this, he fill follow-up with some additional suggestions.

integration-tests/tests/preview-dataset.js
samanehsan: Good catch!

integration-tests/tests/preview-dataset.js
```diff
@@ -0,0 +1,74 @@
...
+  await click(page, `//*[@role="listbox"]//*[@role="option" and contains(normalize-space(.),"${tableNameNoData}")]`)
+  await waitForNoSpinners(page)
+  await findText(page, 'No Data')
+
+  // Verify table counts
+  await click(page, '//*[@role="combobox"]//div')
+  const tablesWithData = await page.$x('//*[@role="combobox"]//*[contains(@id, "option-0")]')
+  const countTablesWithData = _.size(tablesWithData)
+  assertEquals(countTablesWithData, 11)
+  const tablesWithNoData = await page.$x('//*[@role="combobox"]//*[contains(@id, "option-1")]')
```
petesantos: Looking at this xpath and the page content, I think we need to add a label to the combobox. Once we have that, we can use the label to fetch the combobox.

integration-tests/tests/preview-dataset.js
samanehsan: That's a good point! I wanted to confirm that the preview page looks as expected here, and while the snapshot data can't be modified, if anyone deletes the snapshot the test would fail. Is it possible to mock data in this test?

integration-tests/tests/preview-dataset.js
```diff
@@ -0,0 +1,74 @@
...
+const { signIntoTerra, click, clickable, clickTableCell, findElement, findText, findTableCellText, getTableCellPath, getTableHeaderPath, input, waitForNoSpinners } = require('../utils/integration-utils')
+const { withWorkspace } = require('../utils/integration-helpers')
+const { withUserToken } = require('../utils/terra-sa-utils')
+const { dismissNotifications } = require('../utils/integration-utils')
+
+const datasetName = 'Cell hashing with barcoded antibodies enables multiplexing and doublet detection for single cell genomics'
+
+const assertEquals = (expected, actual) => {
+  if (expected !== actual) {
+    throw new Error("Expected value " + expected + " does not equal actual " + actual)
```
kyuksel: We use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) when we want to substitute values (like Python's f-strings).
```suggestion
    throw new Error(`Expected value ${expected} does not equal actual ${actual}`)
```

integration-tests/tests/preview-dataset.js
```diff
@@ -0,0 +1,56 @@
...
+  await findTableCellText(page, getTableCellPath(previewTableName, 2, 1), 'f0caec4a-2a37-4895-8304-83d0fd0da588')
+  await findTableCellText(page, getTableCellPath(previewTableName, 2, 2), '1558104905.862')
+  await findTableCellText(page, getTableCellPath(previewTableName, 2, 3), 'View JSON')
+  await click(page, clickable({ textContains: 'View JSON'  }))
+  await findText(page, 'describedBy')
+  await page.keyboard.press('Escape')
+
+  // Click on a table with no data
+  await click(page, clickable({ text: 'data type', isDescendant: true }))
+  await click(page, clickable({ text: 'Analysis File'}))
```
petesantos, line 45: This looks nice! I think it makes it easier to debug too. If you are looking at the execution, it is easier to read the test and see what should be clicked.

My only reservation now (upon seeing it) is that it isn't clear in the test that we are opening a combo box and clicking an option. But I think this is something we can improve on later

petesantos (approved):
Looks good!
Could you create two tickets? 
1. Investigate a test data strategy
2. Improve A11y in the grouped combo box

samanehsan:
Here are the follow-on tickets:
- https://broadworkbench.atlassian.net/browse/DC-186
- https://broadworkbench.atlassian.net/browse/DC-187

integration-tests/tests/preview-dataset.js
samanehsan: I ended up removing this function, but that is good to know!

integration-tests/utils/integration-utils.js
```diff
@@ -33,16 +33,29 @@ const findInGrid = (page, textContains, options) => {
 }
 
 const clickable = ({ text, textContains, isDescendant = false }) => {
-  const base = `(//a | //button | //input[@type="checkbox"] | //*[@role="button" or @role="checkbox"] | //*[@role="link"])${isDescendant ? '//*' : ''}`
+  const base = `(//a | //button | //input[@type="checkbox"] | //*[@role="button" or @role="checkbox"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])${isDescendant ? '//*' : ''}`
```
samanehsan: The `or @role="checkbox"` here is causing the test to fail to click the "Preview data" button and changing it to `| [@role="checkbox"]` doesn't make a difference 🤔

samanehsan: It turns out the test does not reach this page because this function clicks the checkbox element in the list of datasets instead of the dataset title!

samanehsan:
The `run-workflow-on-snapshot` test that failed succeeds for me locally. Is this flakiness expected?

petesantos (approved):
Took a look at the rebase - LGTM!

integration-tests/tests/preview-dataset.js
kyuksel: 👍🏼

lu-c (approved):

PR #2771, opened 2022-01-19
[BW-1035] Upgrades follows-redirects (high severity dependabot alert)

cahrens:
Updates follows-redirects.

https://github.com/DataBiosphere/terra-ui/security/dependabot

### DIFF
--- a/integration-tests/yarn.lock
+++ b/integration-tests/yarn.lock
@@ -1297,7 +1297,7 @@ __metadata:
   version: 0.21.4
   resolution: "axios@npm:0.21.4"
   dependencies:
-    follow-redirects: ^1.14.7
+    follow-redirects: ^1.14.0
   checksum: 44245f24ac971e7458f3120c92f9d66d1fc695e8b97019139de5b0cc65d9b8104647db01e5f46917728edfc0cfd88eb30fc4c55e6053eef4ace76768ce95ff3c
   languageName: node
   linkType: hard
@@ -2562,7 +2562,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"follow-redirects@npm:^1.14.7":
+"follow-redirects@npm:^1.14.0":
   version: 1.14.7
   resolution: "follow-redirects@npm:1.14.7"
   peerDependenciesMeta:
--- a/integration-tests/yarn.lock
+++ b/integration-tests/yarn.lock
@@ -1294,11 +1294,11 @@ __metadata:
   linkType: hard
 
 "axios@npm:^0.21.1":
-  version: 0.21.1
-  resolution: "axios@npm:0.21.1"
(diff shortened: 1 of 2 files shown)


cjllanwarne (approved):

aednichols (approved):
Seems reasonable, did not test

integration-tests/yarn.lock
```diff
@@ -1294,11 +1294,11 @@ __metadata:
   linkType: hard
 
 "axios@npm:^0.21.1":
-  version: 0.21.1
-  resolution: "axios@npm:0.21.1"
+  version: 0.21.4
```
cahrens, line 1297: @dmohs is hand-editing these files legitimate?

cahrens: And how do we know what to test to make sure we haven't broken anything?

yarn.lock
```diff
@@ -7236,12 +7236,12 @@ __metadata:
   linkType: hard
 
 "follow-redirects@npm:^1.0.0":
```
cahrens, line 7220: `follow-redirects` is a dependency of `http-proxy`, which is a dependency of `http-proxy-middleware`, which is needed by `webpack-dev-server`, which is part of `react-scripts`. I could try upgrading `react-scripts` from 4.0.3 to 5.0.0 if we want.

cahrens: I tried that upgrade, and as one would expect, it introduces tons of dependency changes. Also 5.0.0 was released about a month ago, so it won't necessarily update follow-redirects (security fix is more recent).

zarsky-broad: Webpack dev server is only used for local dev, so we should really dismiss this alert, since it's not exposed, rather than messing with it. That said, the way I normally handle this sort of thing is to delete the entry entirely and see what version `yarn install` puts when we run it.

zarsky-broad: (and then run `yarn dedupe`)

cahrens: Thanks, I took that approach in my most recent commit:

1. Delete entries in yarn.lock that reference follow-redirects
2. yarn install
3. yarn dedupe

kyuksel: Would it be possible to add documentation around these steps to [Security and Maintenance](https://github.com/DataBiosphere/terra-ui/wiki/Security-and-Maintenance)?

cahrens: @kyuksel definitely! That would have helped me immensely yesterday as I floundered around trying to figure out what to do.

cahrens: Updated https://github.com/DataBiosphere/terra-ui/wiki/Security-and-Maintenance#responsibilities

kyuksel (approved):
Also thank you for the documentation 🙏🏼

PR #2748, opened 2022-01-04
[IA-3152] Add one more PR site

dmohs:
Circle configuration updated to add NUM_PR_SITES as an environment variable. Usage of that variable is added in this PR. Any change to that variable would need a corresponding change in the web client authorized JavaScript origins:
https://console.cloud.google.com/apis/credentials/oauthclient/500025638838-s2v23ar3spugtd5t2v1vgfa2sp7ppg0d.apps.googleusercontent.com?project=bvdp-saturn-dev

PR #2772, opened 2022-01-19
[IA-3175] Use Google project if workspace namespace label is undefined

kyuksel:
This PR is to fix the bug I introduced with #2759. 

Older apps, runtimes and disks may not have the label `saturnWorkspaceNamespace` defined, which was causing the Cloud Environments (#clusters) page to give error and not load:

image

The fix is to post-process those resources retrieved from API calls to add the missing label using `googleProject` value. When such resources were created, `googleProject` used to be synonymous with `workspace namespace` (a.k.a. `billing project`).

### Testing
The issue can be reproduced on [dev Terra UI #clusters page](https://bvdp-saturn-dev.appspot.com/#clusters) when logged in as `hermione.owner@test.firecloud.org`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,11 +55,13 @@ commands:
             - run:
                 name: Update Github deployment status
                 command: |
+                  set +e -x # don't allow a failure here to fail the job, print steps when running them
                   <<# parameters.pr >>
                   PR_SLUG="pr-$(expr "${CIRCLE_PR_NUMBER:-${CIRCLE_PULL_REQUEST##*/}}" % $NUM_PR_SITES)-dot-"
                   <</ parameters.pr >>
-                  gh api "repos/databiosphere/terra-ui/deployments/$(cat /tmp/build_num)/statuses" -H "Authorization: token $GITHUB_TOKEN" \
-                    -F state=success -F environment_url="https://${PR_SLUG}bvdp-saturn-<< parameters.env >>.appspot.com" -F log_url="$CIRCLE_BUILD_URL"
+                  RESPONSE=$(gh api "repos/databiosphere/terra-ui/deployments/$(cat /tmp/build_num)/statuses" -H "Authorization: token $GITHUB_TOKEN" \
+                    -F state=success -F environment_url="https://${PR_SLUG}bvdp-saturn-<< parameters.env >>.appspot.com" -F log_url="$CIRCLE_BUILD_URL")
+                  echo $RESPONSE
             - run:
                 name: Update Github deployment status
                 when: on_fail
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -111,14 +111,6 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
   ])
 }
 
-const addNamespaceLabelIfAbsent = resource => {
-  const { googleProject, labels: { saturnWorkspaceNamespace } } = resource
-  // Use googleProject if workspace namespace label is not defined
-  return !!saturnWorkspaceNamespace ?
-    resource :
-    _.set(['labels', 'saturnWorkspaceNamespace'], googleProject, resource)
-}
-
 const Environments = () => {
   const signal = useCancellation()
   const [runtimes, setRuntimes] = useState()
@@ -143,14 +135,9 @@ const Environments = () => {
       Ajax(signal).Disks.list({ creator, includeLabels: 'saturnApplication,saturnWorkspaceNamespace,saturnWorkspaceName' }),
       Ajax(signal).Apps.listWithoutProject({ creator, includeLabels: 'saturnWorkspaceNamespace,saturnWorkspaceName' })
     ])
-
-    // Old apps, runtimes and disks may not have 'saturnWorkspaceNamespace' label defined. When they were
-    // created, workspace namespace (a.k.a billing project) value used to equal the google project.
-    // Therefore we use google project if the namespace label is not defined.
-    setRuntimes(_.map(addNamespaceLabelIfAbsent, newRuntimes))
-    setDisks(_.map(addNamespaceLabelIfAbsent, newDisks))
-    setApps(_.map(addNamespaceLabelIfAbsent, newApps))
-
+    setRuntimes(newRuntimes)
+    setDisks(newDisks)
+    setApps(newApps)
     if (!_.some({ id: getErrorRuntimeId() }, newRuntimes)) {
       setErrorRuntimeId(undefined)
     }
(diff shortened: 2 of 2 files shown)


src/pages/Environments.js
```diff
@@ -20,7 +20,8 @@ import * as Nav from 'src/libs/nav'
 import { useCancellation, useGetter, useOnMount, usePollingEffect } from 'src/libs/react-utils'
 import {
   defaultComputeZone, getComputeStatusForDisplay, getCurrentRuntime, getDiskAppType, getGalaxyComputeCost, getGalaxyCost,
-  getPersistentDiskCostMonthly, getRegionFromZone, isApp, isComputePausable, isResourceDeletable, runtimeCost, workspaceHasMultipleApps, workspaceHasMultipleDisks
+  getPersistentDiskCostMonthly, getRegionFromZone, isApp, isComputePausable, isResourceDeletable, runtimeCost, workspaceHasMultipleApps,
+  workspaceHasMultipleDisks
```
kyuksel, line 24: reformatting only

src/pages/Environments.js
```diff
@@ -134,9 +140,14 @@ const Environments = () => {
...
-    setRuntimes(newRuntimes)
-    setDisks(newDisks)
-    setApps(newApps)
+
+    // Old apps, runtimes and disks may not have 'saturnWorkspaceNamespace' label defined. When they were
+    // created, workspace namespace (a.k.a billing project) value used to equal the google project.
+    // Therefore we use google project if the namespace label is not defined.
+    setRuntimes(_.map(addNamespaceLabelIfAbsent, newRuntimes))
+    setDisks(_.map(addNamespaceLabelIfAbsent, newDisks))
+    setApps(_.map(addNamespaceLabelIfAbsent, newApps))
```
kyuksel: This is the entry point that felt right to handle any missing workspace namespace labels at, to contain the logic in a single place.

src/pages/Environments.js
```diff
@@ -219,8 +230,8 @@ const Environments = () => {
     ])
   }
   const renderWorkspaceForApps = app => {
-    const { appType, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = app
-    const multipleApps = workspaceHasMultipleApps(appsByProject[app.googleProject], appType)
+    const { appType, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = app
+    const multipleApps = workspaceHasMultipleApps(appsByProject[googleProject], appType)
```
kyuksel: Non-functional and unrelated to the bug this PR is aimed to fix. Just noticed that I'd forgotten to pull out `googleProject`.

src/pages/Environments.js
```diff
@@ -110,6 +111,11 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
   ])
 }
 
+const addNamespaceLabelIfAbsent = resource => {
+  // Use googleProject if workspace namespace label is not defined
+  return _.merge({ labels: { saturnWorkspaceNamespace: resource.googleProject } }, resource)
```
gpcarr: if a resource has `labels.saturnWorkspaceNamespace` then this will not do anything correct? would it be possible to only call this method in the case that `saturnWorkspaceNamespace` is undefined?

kyuksel: Yea, totally. Thanks! Updated the code.

gpcarr (approved):
LGTM! Approving assuming this was tested manually. I know this is a prod issue so maybe we can create a ticket for a follow-up PR to add unit tests for the helper method that was added?

.circleci/config.yml
```diff
@@ -55,11 +55,13 @@ commands:
...
                 command: |
+                  set +e -x # don't allow a failure here to fail the job, print steps when running them
                   <<# parameters.pr >>
                   PR_SLUG="pr-$(expr "${CIRCLE_PR_NUMBER:-${CIRCLE_PULL_REQUEST##*/}}" % $NUM_PR_SITES)-dot-"
                   <</ parameters.pr >>
-                  gh api "repos/databiosphere/terra-ui/deployments/$(cat /tmp/build_num)/statuses" -H "Authorization: token $GITHUB_TOKEN" \
-                    -F state=success -F environment_url="https://${PR_SLUG}bvdp-saturn-<< parameters.env >>.appspot.com" -F log_url="$CIRCLE_BUILD_URL"
+                  RESPONSE=$(gh api "repos/databiosphere/terra-ui/deployments/$(cat /tmp/build_num)/statuses" -H "Authorization: token $GITHUB_TOKEN" \
+                    -F state=success -F environment_url="https://${PR_SLUG}bvdp-saturn-<< parameters.env >>.appspot.com" -F log_url="$CIRCLE_BUILD_URL")
+                  echo $RESPONSE
```
kyuksel, line 64: The changes in this file are unrelated to the bug but the PR was failing because the step above was exiting with a non-zero code. This recently happened with the previous step as well and was remediated by #2748 so I followed that approach to prevent the step from failing the entire flow.

src/pages/Environments.js
```diff
@@ -110,6 +111,14 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
   ])
 }
 
+const addNamespaceLabelIfAbsent = resource => {
+  const { googleProject, labels: { saturnWorkspaceNamespace } } = resource
+  // Use googleProject if workspace namespace label is not defined
+  return !!saturnWorkspaceNamespace ?
+    resource :
+    _.set(['labels', 'saturnWorkspaceNamespace'], googleProject, resource)
```
zarsky-broad: I'd suggest `_.defaultsDeep` here

src/pages/Environments.js
zarsky-broad: If it's only used here, should this be `addNamespaceLabelsIfAbsent`, with the map inside the function?

src/pages/Environments.js
```diff
@@ -219,8 +233,8 @@ const Environments = () => {
     ])
   }
   const renderWorkspaceForApps = app => {
-    const { appType, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = app
-    const multipleApps = workspaceHasMultipleApps(appsByProject[app.googleProject], appType)
+    const { appType, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = app
```
zarsky-broad: Instead of all of this, could you just do this?
```suggestion
    const { appType, googleProject, labels: { saturnWorkspaceNamespace = googleProject, saturnWorkspaceName } } = app
```

zarsky-broad (approved):
Some comments but functionality makes sense

src/pages/Environments.js
kyuksel: Yes! I didn't know we could use the value of a property whose object is deconstructed on the same line right away like that 🤩 

There's still the issue of having to do that in 3 different places but I don't think avoiding that is worth all the previous complexity.

PR #2773, opened 2022-01-19
[DC-180] Added modal for see more filters in the data catalog

lu-c:
Screen Shot 2022-01-19 at 3 16 24 PM

^ This is the after. The before is just a list that expands when you hit "see more"


src/components/Modal.js
```diff
@@ -55,7 +55,7 @@ const Modal = ({ onDismiss, title, titleExtras, children, width = 450, showCance
       previouslyFocusedNode.current = modalNode.current.contains(document.activeElement) ? previouslyFocusedNode.current : document.activeElement
       nodeToFocus.focus()
     },
-    style: { overlay: styles.overlay, content: { ...styles.modal, width } },
+    style: { overlay: styles.overlay, content: { ...styles.modal, width, ...props.styles.modal } },
```
samanehsan: I checked this out locally and hit an error in this line: `TypeError: Cannot read properties of undefined (reading 'modal')`

samanehsan:
Lgtm after the tests pass! :octocat:

src/pages/library/common.js
```diff
@@ -86,8 +99,132 @@ const FilterSection = ({ onTagFilter, labels, selectedTags, labelRenderer, listD
...
+      },
+      onDismiss: () => {
+        setShowAll(false)
+        _.forEach(onTagFilter, filterChanges)
+        setFilterChanges({})
+        setFilterSearchText('')
+        setFilteredLabels(labels)
+        setFilterType(filterOptions.alpha)
+      }
+    }, [
```
petesantos: How would you feel about factoring out the filter index bar into its own component (e.g. PartitionBar)?
It could support events like:

1. onPartition
2. onTopPartition (or onSpecial)
3. onSearch
4. onClear

I think this would help separate the state concerns between managing the partition state (which partition is selected / whether the search bar is displayed, etc..

And the results displayed by the partition. I think this would reduce the complexity of this component and provide Terra UI with a nice re-usable component.

petesantos (commented):
This is looking great! I played around with it a bit and it is working nicely.
My only concern is the added size/complexity of the section filter. I have a comment below. If you want to chat about it, let me know.

src/pages/library/common.js
lu-c: Let me try splitting it out into a separate component - I think the terminology confuses me a little with "partition", but let me give it an initial crack and see what you think!

petesantos: Thanks! I was thinking of partition because we are viewing segments of the data. I could see a use case where we use the bar but with a different partition scheme (e.g. numbers or something like "A-D", "E-H", "I-M"). But no need to worry about other "partition" types right now. 

I don't feel strongly about the name - feel free to use something clearer. 
Other suggestions: "SegmentSelector" or "SegmentBar"

src/pages/library/common.js
```diff
@@ -86,8 +174,74 @@ const FilterSection = ({ onTagFilter, labels, selectedTags, labelRenderer, listD
...
+
+        const _tagFilters = {}
+        _.forEach(label => {
+          const lowerTag = _.toLower(label)
+          _tagFilters[lowerTag] = listDataByTag[lowerTag]
+        }, labels)
+        setFilteredTags(_tagFilters)
+      }
+    }, [`See more`]),
+    showAll && h(Modal, {
```
petesantos: I was hoping that moving the filter bar would be enough to simplify `FilterSection`, but I think we should factor the modal out as well. 

My main reasoning for this is the state for `filterTags` and `filterLabels` seems to only be used in the modal, so we could remove this state from `FilterSection`which would reduce the complexity of this component and hopefully make it easier to maintain.

src/pages/library/common.js
```diff
@@ -86,8 +174,74 @@ const FilterSection = ({ onTagFilter, labels, selectedTags, labelRenderer, listD
...
+      onClick: () => {
+        setShowAll(!showAll)
+        setFilterChanges({})
+
+        const _tagFilters = {}
+        _.forEach(label => {
+          const lowerTag = _.toLower(label)
+          _tagFilters[lowerTag] = listDataByTag[lowerTag]
+        }, labels)
+        setFilteredTags(_tagFilters)
```
petesantos: To avoid mutation we can create key/value pairs (basically tuples) and use lodash's `fromPairs` function

```suggestion
        setFilteredTags(_.flow(
          _.map(label => {
            const lowerTag = _.toLower(label)
            return [lowerTag, listDataByTag[lowerTag]]
          }),
          _.fromPairs
        )(labels))
```

src/pages/library/common.js
```diff
@@ -86,8 +174,74 @@ const FilterSection = ({ onTagFilter, labels, selectedTags, labelRenderer, listD
...
+          }
+        }, _.map(label => {
+          const lowerTag = _.toLower(label)
+          return div({ className: 'label', style: { width: '25%', margin: '0 15px 10px 30px', position: 'relative', minHeight: 30 } }, [
+            h(LabeledCheckbox, {
+              checked: !!filterChanges[lowerTag] ^ _.includes(lowerTag, lowerSelectedTags),
+              onChange: v => {
+                filterChanges[lowerTag] ? delete filterChanges[lowerTag] : filterChanges[lowerTag] = { lowerTag, label }
+                setFilterChanges({ ...filterChanges })
+              },
```
petesantos: We tend to avoid mutation where we can, especially on react state values. Mutating the state outside of a set state could have unintended effects.

The react `useState` hook allows us to set state with a function, where the parameter passed in is the previous state. I think this will do the same thing without the mutation:

```suggestion
onChange: () => {
                setFilterChanges(prevFilter => prevFilter[lowerTag] ? _.omit(lowerTag, prevFilter) : _.set(lowerTag, { lowerTag, label }, prevFilter))
              },
```

petesantos (commented):
Another decomposition suggestion.
I am going through this a bit more to get a better handle on the state being used. But I do think moving the modal out will help

src/pages/library/common.js
lu-c: tht's fine! I pushed up n update to move the code around to its own function, let me know if that's more what you were thinking!

src/pages/library/common.js
```diff
@@ -58,11 +61,158 @@ const computeLabels = (allLabels, selectedLabels) => _.flow(
   _.uniq
 )(selectedLabels)
 
-const FilterSection = ({ onTagFilter, labels, selectedTags, labelRenderer, listDataByTag }) => {
+const FilterBar = ({ name, onSearch, onClear, onFilterByLetter, onFilterBySearchText, onSortBy }) => {
```
petesantos: nit: it looks like `onSearch` is unused. I am not sure why the linter didn't catch this

src/pages/library/common.js
```diff
@@ -58,11 +61,158 @@ const computeLabels = (allLabels, selectedLabels) => _.flow(
...
+  const filterOptions = { alpha: 1, input: 2 }
+  const [filterSearchText, setFilterSearchText] = useState('')
+  const [filteredBy, setFilteredBy] = useState()
+  const [filterType, setFilterType] = useState(filterOptions.alpha)
+
+  return h(Fragment, [
+    filterType === filterOptions.alpha && div({ style: { backgroundColor: colors.dark(0.1), paddingTop: 6, borderRadius: 8, display: 'flex', flexDirection: 'row', alignItems: 'center', justifyContent: 'space-between' } }, [
```
petesantos: nit: can you break this style up into multiple lines? Or add it to the styles at the top of the file?

src/pages/library/common.js
```diff
@@ -58,11 +61,158 @@ const computeLabels = (allLabels, selectedLabels) => _.flow(
...
+          marginLeft: 20, padding: 5,
+          fontWeight: 'bold', fontSize: '1rem', textTransform: 'capitalize',
+          background: filteredBy === 'top' ? colors.accent(0.3) : 'transparent', borderRadius: 10
+        },
+        onClick: () => {
+          setFilteredBy('top')
+          onSortBy('top')
+        }
+      }, [`Top ${pluralize(name)}`]),
+      div({ style: { width: '100%', maxWidth: 600, display: 'flex', flexDirection: 'row', alignItems: 'center', justifyContent: 'space-around', fontSize: '1rem' } },
```
petesantos: Same style nit as above

src/pages/library/common.js
```diff
@@ -58,11 +61,158 @@ const computeLabels = (allLabels, selectedLabels) => _.flow(
...
+      return [lowerTag, listDataByTag[lowerTag]]
+    }),
+    _.fromPairs
+  )(labels))
+  const labelsByFirstChar = _.groupBy(label => _.capitalize(label)[0], labels)
+  const [filterChanges, setFilterChanges] = useState({})
+  const [filteredLabels, setFilteredLabels] = useState(labels)
+
+  return h(Modal, {
+    title: ``,
```
petesantos: I think for the moment we need to put the title in here. The reason being, title uses an `aria-labelledby` for the title and it is not rendering the associated element.

The title can be a component so we could the following to match the styling:
```suggestion
    title: div({ style: { fontSize: '1.325rem', fontWeight: 700, marginTop: 0 } }, [`Filter by: "${name}"`]),
```
I discovered this issue using the [ANDI](https://www.ssa.gov/accessibility/andi/help/install.html) tool/bookmarklet

src/pages/library/common.js
```diff
@@ -58,11 +61,158 @@ const computeLabels = (allLabels, selectedLabels) => _.flow(
...
+    styles: {
+      modal: { maxWidth: 900, padding: 30 },
+      buttonRow: { width: '100%', borderTop: `6px solid ${colors.dark(0.1)}`, paddingTop: 20 }
+    },
+    onDismiss: () => {
+      setShowAll(false)
+      _.forEach(onTagFilter, filterChanges)
+    }
+  }, [
+    h2({ style: { marginTop: 0 } }, [`Filter by: "${name}"`]),
```
petesantos: I would recommend removing this in favor of using the modal title

petesantos (approved):
A few nits, but looks good - thanks!

samanehsan:
The run-catalog-filter test is failing with: `Error: Rows not filtered after searching for 'stem cell'` that looks different from the usual flakiness 🤔

lu-c:
> The run-catalog-filter test is failing with: `Error: Rows not filtered after searching for 'stem cell'` that looks different from the usual flakiness 🤔

Shoot, let me look into that, sorry.... I saw the failed test, and assumed flakiness T_T

EDIT: I tried running this locally, and am not getting the same failing test... so I'm going to try pushing a new commit to try to rerun this pipeline!

samanehsan:
> I tried running this locally, and am not getting the same failing test... so I'm going to try pushing a new commit to try to rerun this pipeline!

Oh okay! It's not failing in this run so it looks like it was actually flaky!

samanehsan (approved):
Lgtm! :octocat:

PR #2774, opened 2022-01-19
[DC-173] Test linking to new workspace

samanehsan:
- Add UI test to link a dataset from the data catalog page to a new workspace
- Move the data catalog test set-up into a util function (could also pull this out into it's own PR if that's easier!)

### DIFF
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -1,24 +1,13 @@
-const _ = require('lodash/fp')
-const { signIntoTerra, checkbox, click, clickable, clickTableCell, dismissNotifications, fillIn, findText, waitForNoSpinners } = require('../utils/integration-utils')
-const { testWorkspaceName } = require('../utils/integration-helpers')
+const { checkbox, click, clickable, clickTableCell, fillIn, waitForNoSpinners } = require('../utils/integration-utils')
+const { enableDataCatalog, testWorkspaceName } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
 const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) => {
-  await page.goto(testUrl)
-  await waitForNoSpinners(page)
-
-  await findText(page, 'Browse Data')
-
-  await page.evaluate(() => window.configOverridesStore.set({ isDataBrowserVisible: true }))
-  await page.reload({ waitUntil: ['networkidle0', 'domcontentloaded'] })
-
-  await click(page, clickable({ textContains: 'Browse Data' }))
-  await signIntoTerra(page, token)
-  await dismissNotifications(page)
-
+  await enableDataCatalog(page, testUrl, token)
   await click(page, clickable({ textContains: 'browse & explore' }))
   await waitForNoSpinners(page)
+
   await click(page, checkbox({ text: 'Granted', isDescendant: true }))
   await clickTableCell(page, "dataset list", 2, 2)
   await waitForNoSpinners(page)
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -9,7 +9,7 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) 
   await waitForNoSpinners(page)
 
   await click(page, checkbox({ text: 'Granted', isDescendant: true }))
-  await clickTableCell(page, "dataset list", 2, 2)
+  await clickTableCell(page, 'dataset list', 2, 2)
   await waitForNoSpinners(page)
   await click(page, clickable({ textContains: 'Link to a workspace' }))
   await waitForNoSpinners(page)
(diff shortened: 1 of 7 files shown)


kyuksel:
1. Could you please add the new integration test to [this ownership map](https://broadworkbench.atlassian.net/wiki/spaces/QA/pages/109248571/Triage+Processes#Terra-UI-Test-Inventory-%26-Release-Criteria) so the QA team can be aware of it? While at it, it'd be great if you could make sure any other integration tests owned by the Data Catalog team are there as well 🙏🏼 
2. Could you please run the test several times to catch any flakiness before the PR is merged?

integration-tests/tests/link-to-new-workspace.js
```diff
@@ -0,0 +1,42 @@
...
+const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) => {
+  await enableDataCatalog(page, testUrl, token)
+  await click(page, clickable({ textContains: 'browse & explore' }))
+  await waitForNoSpinners(page)
+
+  await click(page, checkbox({ text: 'Granted', isDescendant: true }))
+  await clickTableCell(page, "dataset list", 2, 2)
+  await waitForNoSpinners(page)
+  await click(page, clickable({ textContains: 'Link to a workspace' }))
+  await waitForNoSpinners(page)
```
kyuksel: `run-catalog-workflow` test also has these 9 lines, which makes me wonder if it'd make sense to factor them out.

integration-tests/tests/link-to-new-workspace.js
```diff
@@ -0,0 +1,42 @@
...
+    await click(page, clickable({ text: 'Select a billing project' }))
+    await click(page, clickable({ text: `${newWorkspaceBillingAccount}` }))
+    await click(page, clickable({ text: 'Create Workspace' }))
+    await waitForNoSpinners(page)
+    await page.url().includes(newWorkspaceName)
+  } finally {
+    await page.evaluate((name, billingProject) => {
+        return window.Ajax().Workspaces.workspace(billingProject, name).delete()
+      }, `${newWorkspaceName}`, `${newWorkspaceBillingAccount}`)
+  }
```
kyuksel: There seems to be some mis-indentation above. Could you [run the reformatter](https://github.com/DataBiosphere/terra-ui/wiki/Developing#code-style-and-linting)?

integration-tests/utils/integration-helpers.js
```diff
@@ -22,9 +22,10 @@ const withSignedInPage = fn => async options => {
 
 const clipToken = str => str.toString().substr(-10, 10)
 
-const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
-  const workspaceName = `test-workspace-${Math.floor(Math.random() * 100000)}`
+const testWorkspaceName = () => `test-workspace-${Math.floor(Math.random() * 100000)}`
```
kyuksel: We typically use verbs for function names (and nouns for others) (e.g. `getTestWorkspaceName`) to make it easier to know at a glance if a name belongs to a function.

kyuksel (commented):
Thanks for factoring out common code to `enableDataCatalog`.

integration-tests/tests/link-to-new-workspace.js
samanehsan: The link to the `js-style.xml` file is broken. Is there a different way to configure the code style?

kyuksel: @zarsky-broad do you have that file at hand, and/or do you have a different recommendation?

zarsky-broad: Ah, the docs are outdated; intellij should pick up the editorconfig file at the project root and apply it...

zarsky-broad: but you can also just right click and fix eslint problems

integration-tests/utils/integration-helpers.js
```diff
@@ -170,10 +172,25 @@ const withRegisteredUser = test => withUser(async options => {
   await test(options)
 })
 
+const enableDataCatalog = async (page, testUrl, token) => {
+  await page.goto(testUrl)
+  await waitForNoSpinners(page)
+
+  await findText(page, 'Browse Data')
+  await page.evaluate(() => window.configOverridesStore.set({ isDataBrowserVisible: true }))
+  await page.reload({ waitUntil: ['networkidle0', 'domcontentloaded'] })
```
petesantos, line 181: @dmohs had made a comment in a previous PR about throwing an error when we no longer need this flag to ensure it gets cleaned up. I thought that is a good idea, so created [this ticket](https://broadworkbench.atlassian.net/browse/DC-176) to track this. This looks like a good step in that direction.

I would prefer not making any changes for this PR because I think it is a good iterative improvement. We can follow up with changing how we handle the feature flag in a small future PR.

Also, once the flag handling has been adjusted, we can make this function more generic to apply it to other tests that share some of this boilerplate. I looked at some of the existing tests, it looks like this helper function would be of good use in some of those as well.

petesantos (approved):
Thanks for factoring the boilerplate out - it is nice to see a lot of code removed!
I have a comment around the helper function, but I think we can address it in a future PR, as this looks like a good iterative improvement!

kyuksel (approved):

PR #2775, opened 2022-01-20
[BW-1037] Upgrades related to marked

cahrens:
There is a high severity security alert related to marked.

https://github.com/DataBiosphere/terra-ui/security/dependabot/yarn.lock/marked/open

Upgrading the libraries that have marked as dependencies.


package.json
```diff
@@ -25,7 +25,7 @@
     "igv": "2.3.5",
     "jszip": "^3.7.1",
     "lodash": "^4.17.21",
-    "marked": "^2.1.3",
+    "marked": "^4.0.10",
```
cahrens, line 28: A pretty big leap, but this is the version required for the security alert.

package.json
```diff
@@ -46,7 +46,7 @@
     "react-paginating": "^1.4.0",
     "react-scripts": "4.0.3",
     "react-select": "^4.3.1",
-    "react-simplemde-editor": "^4.1.5",
+    "react-simplemde-editor": "^5.0.2",
```
cahrens, line 49: Because of the changes in version 5 that are marked as "breaking", we should either test our usages of this or consider staying at version 4.1.5.

https://www.npmjs.com/package/react-simplemde-editor

kyuksel (approved):
LGTM but would love to have someone like @zarsky-broad to double-check.
Tested some basic functionality (mostly IA), which looked good.
Thank you for taking care of this.

package.json
cahrens: These are the two places we use the editor, and our usages are extremely straightforward:

1) Editing the description of a workspace. I am able to test that.
2) Editing a snapshot in Data. I do not know how to get to that point in the UI.

Screenshot from fist usage.
image

cahrens: @zarsky-broad it looks like you wrote the code for editing snapshots. Can you help me with testing that the editor is still working correctly in that location?

src/components/markdown.js
```diff
@@ -1,6 +1,6 @@
 import DOMPurify from 'dompurify'
 import _ from 'lodash/fp'
-import marked from 'marked'
+import { marked } from 'marked'
```
cahrens, line 3: Without this change, attempts to reference `marked.Renderer` failed with marked as undefined. This was caught by our register user integration test (which shows the terms of service by calling `renderAndSanitizeMarkdown`.

breilly2: Sweet! _This_ is an _excellent_ example of a *Critical* User Journey. The consequences of new users not being able sign up (or existing users not being able to accept a new ToS) is really bad, and it's something that we might easily miss otherwise because it's not a commonly walked path for developers.

package.json
zarsky-broad: As long as you can edit and save, and the saving works, that's really all there is to see.

cahrens: @zarsky-broad can you provide steps on how I get to the Data Snapshot UI where this editor is present? I searched around in the Data tab but couldn't find anything about snapshots.

zarsky-broad: Go to any workspace with a snapshot imported, click the info button for the snapshot in the data tab in the left bar, and you can edit the description.

cahrens: @zarsky-broad shared a workspace with me that included a snapshot. I was able to use the editor to update the snapshot description and save my changes.

kyuksel (approved):

cjllanwarne (approved):
Code changes seem reasonable and appropriate validation steps were taken. 👍

PR #2779, opened 2022-01-21
[IA-2959] Disabled dataproc for RStudio

cstrandw:
https://broadworkbench.atlassian.net/browse/IA-2959

Small change to disable both of the dataproc options when an RStudio image is selected.

Tested by selecting the RStudio option to check that the options were disabled. I also made sure that for other images dataproc is not disabled.

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -8,7 +8,7 @@ import { icon } from 'src/components/icons'
 import { ImageDepViewer } from 'src/components/ImageDepViewer'
 import { NumberInput, TextInput, ValidatedInput } from 'src/components/input'
 import { withModalDrawer } from 'src/components/ModalDrawer'
-import { tools } from 'src/components/notebook-utils'
+import { tools, getToolForImage } from 'src/components/notebook-utils'
 import { InfoBox } from 'src/components/PopupTrigger'
 import { getAvailableComputeRegions, getRegionInfo, isUSLocation, locationTypes } from 'src/components/region-common'
 import { SaveFilesHelp, SaveFilesHelpRStudio } from 'src/components/runtime-common'
@@ -197,7 +197,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   const isPersistentDisk = shouldUsePersistentDisk(sparkMode, currentRuntimeDetails, upgradeDiskSelected)
 
   const isCustomImage = selectedLeoImage === customMode
-  const { version, updated, packages, requiresSpark, isRStudio,label: packageLabel } = _.find({ image: selectedLeoImage }, leoImages) || {}
+  const { version, updated, packages, requiresSpark, label: packageLabel } = _.find({ image: selectedLeoImage }, leoImages) || {}
 
   const minRequiredMemory = sparkMode ? 7.5 : 3.75
   const validMachineTypes = _.filter(({ memory }) => memory >= minRequiredMemory, machineTypes)
@@ -303,6 +303,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   const canManageSparkConsole = () => shouldDisplaySparkConsoleLink() && isRuntimeRunning()
 
   const canUpdateNumberOfWorkers = () => !currentRuntimeDetails || isRuntimeRunning()
+
+  const requiresStandard = () => getToolForImage(_.find({ image: selectedLeoImage }, leoImages)?.id) === tools.RStudio.label
 
   const canUpdateRuntime = () => {
     const { runtime: existingRuntime } = getExistingEnvironmentConfig()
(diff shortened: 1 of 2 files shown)


src/components/ComputeModal.js
```diff
@@ -953,8 +952,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
                     },
                     options: [
                       { value: false, label: 'Standard VM', isDisabled: requiresSpark },
-                      { value: 'master', label: 'Spark master node' },
-                      { value: 'cluster', label: 'Spark cluster' }
+                      { value: 'master', label: 'Spark master node', isDisabled: isRStudio },
```
jdcanas: Hmm... After thinking about it a bit more, I don't think I like the image object itself having a flag for `isRStudio`. 

This isn't a scalable approach. For example, say in the future we need to disable/enable options for Jupyter images. I think a better approach is to leverage this [tools dictionary](https://github.com/DataBiosphere/terra-ui/blob/dev/src/components/notebook-utils.js#L75), and check if the `selectedLeoImage` id is present in the `imageIds` array. This approach is more flexible too, as it doesn't require any terra-docker changes and building of the .json file for re-classification of images. 

I would prefer to use this tools dictionary for all current uses of `isRStudio` I think. `requiresSpark` is a little different in that a Jupyter image can require spark, so it's needed still. What are your thoughts? cc @gpcarr

cahrens: I agree that it would be great to expose this via a property/method in the tools dictionary instead of sprinkling isRStudio around!

kyuksel:
With this change, `Compute type` drop down now has the `Dataproc` options greyed out:

image


which is great. I was wondering if we should also add an `infobox` similarly to the one we have for `Location` right below with a `tooltip` informing the user why the Dataproc options are greyed out.


image

src/components/ComputeModal.js
```diff
@@ -305,6 +304,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
 
   const canUpdateNumberOfWorkers = () => !currentRuntimeDetails || isRuntimeRunning()
 
+  const requiresStandard = () => getToolForImage(_.find({ image: selectedLeoImage }, leoImages)?.id) === tools.RStudio.label
```
jdcanas: nit: maybe rename this to `requiresGCE`. People may have different perspectives on what's standard, especially if they're coming into this code without too much background knowledge of google compute. The drop down option uses the 'standard' wording because if a user has no knowledge of the backing compute we want them to keep that option selected by default

jdcanas (approved):

src/components/ComputeModal.js
```diff
@@ -194,7 +194,6 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     computeZone: defaultComputeZone
   })
   // State -- end
-
```
kyuksel: Nit: It helps to have the blank line to more easily differentiate the section above from the rest.

src/components/ComputeModal.js
```diff
@@ -941,6 +942,9 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           h(IdContainer, [
             id => div({ style: { gridColumnEnd: 'span 4', marginTop: '0.5rem' } }, [
               label({ htmlFor: id, style: computeStyles.label }, ['Compute type']),
+              h(InfoBox, { style: { marginLeft: '0.5rem', display: requiresStandard() || requiresSpark ? '' : 'none' } }, [
+                'Only the compute types compatible with the selected application configuration are made available below.'
```
jdcanas, line 947: maybe just check quickly with our UX partner, Tiffany Chan, about the wording here if you haven't :)

jdcanas:
Overall looks good! 🎉 

Just a few nits.

After you look into those, :shipit: from me (Note you need 2 approval though)

src/components/ComputeModal.js
```diff
@@ -1377,15 +1381,15 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+          options: getImages(({ isCommunity, id }) => (!isCommunity && !(getToolForImage(id) === tools.RStudio.label)))
         },
         {
           label: 'COMMUNITY-MAINTAINED JUPYTER ENVIRONMENTS (verified partners)',
           options: getImages(_.get(['isCommunity']))
         },
         {
           label: 'COMMUNITY-MAINTAINED RSTUDIO ENVIRONMENTS (verified partners)',
-          options: getImages(_.get(['isRStudio']))
+          options: getImages((image) => getToolForImage(image.id) === tools.RStudio.label)
```
kyuksel: Can be
```suggestion
          options: getImages(image => getToolForImage(image.id) === tools.RStudio.label)
```
since the function input takes a single input (i.e. `image`)

kyuksel:
There seem to be a couple of linting errors:

image

src/components/ComputeModal.js
```diff
@@ -941,6 +942,9 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           h(IdContainer, [
             id => div({ style: { gridColumnEnd: 'span 4', marginTop: '0.5rem' } }, [
               label({ htmlFor: id, style: computeStyles.label }, ['Compute type']),
+              h(InfoBox, { style: { marginLeft: '0.5rem', display: requiresStandard() || requiresSpark ? '' : 'none' } }, [
```
kyuksel: We tend to pull the predicate out and write it like below
```suggestion
              (requiresStandard() || requiresSpark) && h(InfoBox, { style: { marginLeft: '0.5rem' } }, [
```

This makes the predicate easier to notice when reading the code plus eliminates the need for `? '' : 'none'`.

src/components/ComputeModal.js
kyuksel: I agree that `requiresStandard` can be ambiguous.

One downside of `requiresGCE` is that it is Google Cloud Platform (GCP)-specific, which would renaming when we introduce Azure but I guess at that point, that'd be the least of renamings we'd have to do :)

Nit: I like `requiresGce` better for consistency with how we've used `gce` in variable/method names.

kyuksel (approved):
Looks great  👍🏼

PR #2780, opened 2022-01-24
[IA-3175] Use googleProject when a disk doesn't have workspace namespace defined

kyuksel:
The last PRs missed the corner case where a disk does _not_ have workspace _namespace_  defined but _does_ have workspace _name_ defined. 

Currently in this scenario, Cloud Environments page gives the error below:

image

With this fix, we'll get the expected page view

image

Reproduced the issue and by manually updating Leonardo's `dev` `LABEL` table and verified the fix on a local deployment.

### DIFF
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -457,7 +457,7 @@ const Environments = () => {
             field: 'workspace',
             headerRenderer: () => h(Sortable, { sort: diskSort, field: 'workspace', onSort: setDiskSort }, ['Workspace']),
             cellRenderer: ({ rowIndex }) => {
-              const { status: diskStatus, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = filteredDisks[rowIndex]
+              const { status: diskStatus, googleProject, labels: { saturnWorkspaceNamespace = googleProject, saturnWorkspaceName } } = filteredDisks[rowIndex]
               const appType = getDiskAppType(filteredDisks[rowIndex])
               const multipleDisks = multipleDisksError(disksByProject[googleProject], appType)
               return !!saturnWorkspaceName ?


jdcanas (approved):

src/pages/Environments.js
```diff
@@ -457,7 +457,7 @@ const Environments = () => {
             field: 'workspace',
             headerRenderer: () => h(Sortable, { sort: diskSort, field: 'workspace', onSort: setDiskSort }, ['Workspace']),
             cellRenderer: ({ rowIndex }) => {
-              const { status: diskStatus, googleProject, labels: { saturnWorkspaceNamespace, saturnWorkspaceName } } = filteredDisks[rowIndex]
```
Qi77Qi, line 460: would `information unavaible` make more sense?

src/pages/Environments.js
```diff
@@ -457,7 +457,7 @@ const Environments = () => {
...
+              const { status: diskStatus, googleProject, labels: { saturnWorkspaceNamespace = googleProject, saturnWorkspaceName } } = filteredDisks[rowIndex]
```
cahrens, line 460: Assuming this just populates a default value if saturnWorkspaceNamespace doesn't exist (I'm still a newbie at this syntax), this LGTM.

cahrens (approved):

src/pages/Environments.js
kyuksel: Exactly.

src/pages/Environments.js
kyuksel: For old disks where `namespace` (aka `billing project`) label may be missing, `googleProject` value is the same as `namespace/billing project` so we use that. `googleProject` and `namespace/billing project` values started diverging only after PPW.

Qi77Qi: okay...if we're confident in that..just wonder if there's a case where workspace name doesn't exist after PPW....

Qi77Qi (approved):

src/pages/Environments.js
kyuksel: I think that'd be an anomaly if it happened, in which case I think we'd want to know about and debug it.

kyuksel:
Will merge despite the integration test failure because

1. The integration test failure has the known signature of

image

which is worked on via #2781 .

2. The change in this PR is very unrelated to the failing test.

3. Prod bug this PR fixes has the potential to break Cloud Environments page for users.

PR #2781, opened 2022-01-24
[QA-1696] Fix click create flakiness

xuscrbw:
<!--
Hello, friend!
Thanks!
--!>
Testing this fix out to see if this helps to resolve the terra-ui tests.

After testing locally, it seems like the test gets stuck when clicking the create button after: creating notebook > edit >Create runtime.

It might pass more consistently if we slightly increase the delay before the page opens.


xuscrbw:
multi-test

integration-tests/tests/run-notebook.js
```diff
@@ -25,6 +25,7 @@ const testRunNotebookFn = _.flow(
   await select(page, 'Language', 'Python 3')
   await click(page, clickable({ text: 'Create Notebook' }))
   await click(page, clickable({ textContains: notebookName }))
+  await delay(2000)
```
breilly2: Do you need this new delay or is increasing the delay below enough? Adding delays in a test is not ideal. When necessary, it should be justified like the comment below justifies the existing delay. Without that, someone coming across this in the future has no idea if the amount of delay is still appropriate or if it's even needed at all.

integration-tests/tests/run-notebook.js
```diff
@@ -40,7 +41,7 @@ const testRunNotebookFn = _.flow(
   // 500ms delay in the test is not enough to guarantee that the UI elements have finished moving.
   // Therefore, we start with a 1000ms delay, then make sure there are no spinners just in case the
```
breilly2: Nit: update comment to reflect new learning about the delay length.

kyuksel: +1

integration-tests/tests/run-notebook.js
kyuksel: 💯

kyuksel (approved):
LGTM pending @breilly2's comments.

I'd also suggest running the test ~10 times to have a statistically better idea about the stability of the test. In that vein, I just took the liberty to kick off a re-run.

kyuksel:
@xuscrbw I've rerun `run-notebook` several times through the CI and it passed every time  🤩

integration-tests/tests/run-notebook.js
xuscrbw: Addressed

integration-tests/tests/run-notebook.js
xuscrbw: I've been having more success with this new delay locally, but going to try a few runs without it on this PR.

xuscrbw:
@breilly2 Re-requested a review. I was able to get two successful test passes without the first delay and only adding an increase to the second. Also addressed your nit comment

breilly2 (approved):
👍  Thanks for addressing my feedback!

breilly2:
Hmmm... I wouldn't expect this delay to need to be 3 seconds. This delay was just to cover the gap between the drawer sliding in and the spinner to render (while loading cloud environment details). Makes me think that something else is going on here.

xuscrbw:
> Hmmm... I wouldn't expect this delay to need to be 3 seconds. This delay was just to cover the gap between the drawer sliding in and the spinner to render (while loading cloud environment details). Makes me think that something else is going on here.

I reverted the increase and it looks like it managed to pass the last 6 runs. I am going to merge this, just to unblock things for now. May keep looking for other answers in the meantime

kyuksel:
> I reverted the increase and it looks like it managed to pass the last 6 runs. I am going to merge this, just to unblock things for now. May keep looking for other answers in the meantime

Sounds good. Thanks for looking into it.

PR #2782, opened 2022-01-24
[WOR-62] Add call to Sam to check that user is member of given group

marctalbott:
Ticket: [WOR-62](https://broadworkbench.atlassian.net/browse/WOR-62)
* We will be restricting access to alpha spend reports to a subset of users that will be members of a specific Sam group. We'll use `isMember` to conditionally display spend reports to users depending on if they're in the Sam group in question. The group that the UI checks will be specified in config (which will be added in a separate PR)

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -479,6 +479,10 @@ const Groups = signal => ({
 
       setPolicy: (policyName, value) => {
         return fetchSam(`${resourceRoot}/policies/${policyName}/public`, _.mergeAll([authOpts(), { signal, method: 'PUT' }, jsonBody(value)]))
+      },
+
+      isMember: async () => {
+        await fetchSam(`${resourceRoot}/action/use`, _.merge(authOpts(), { signal }))
       }
     }
   }
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -482,7 +482,8 @@ const Groups = signal => ({
       },
 
       isMember: async () => {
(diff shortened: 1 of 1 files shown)


src/libs/ajax.js
```diff
@@ -479,6 +479,11 @@ const Groups = signal => ({
 
       setPolicy: (policyName, value) => {
         return fetchSam(`${resourceRoot}/policies/${policyName}/public`, _.mergeAll([authOpts(), { signal, method: 'PUT' }, jsonBody(value)]))
+      },
+
+      isMember: async () => {
+        const res = await fetchSam(`${resourceRoot}/action/use`, _.merge(authOpts(), { signal }))
+        return res.json()
```
marctalbott, line 486: This is going to be a boolean, is there anything I should do to make this boolean/results of this fn easier to use in conditional statements or will javascript just handle it?

My goal is to be able to use it like this

```
if (Ajax(signal).Groups.group(spendReportGroup).isMember) {
    do spend report stuff
}

Comment (dmohs) [reply src/libs/ajax.js:486]: It would technically work (and is allowed by the standard) if the endpoint returns a boolean value as the response:
```javascript
> JSON.parse('true') ? 'yes' : 'no'
'yes'
> JSON.parse('false') ? 'yes' : 'no'
'no'
```
Because it's not a common pattern, it would not surprise me if the encoding library screwed this up. For the record, Node does the right thing:
```javascript
> JSON.stringify(false)
'false'
```

dmohs (approved):

ungwudik (approved):
lgtm

MatthewBemis (approved):

PR #2784, opened 2022-01-24
Upgrade nanoid (BW-1045).

cahrens:
Poked around in terra-ui and didn't see any CSS isssues.

roberts@wm365-d20 terra-ui % yarn why nanoid        
└─ postcss@npm:8.3.6
   └─ nanoid@npm:3.2.0 (via npm:^3.1.23)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10477,11 +10477,11 @@ fsevents@^1.2.7:
   linkType: hard
 
 "nanoid@npm:^3.1.23":
-  version: 3.1.23
-  resolution: "nanoid@npm:3.1.23"
+  version: 3.2.0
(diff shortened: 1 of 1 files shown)


cjllanwarne (approved):

kyuksel (approved):
`nanoid` seems to be a `react-scripts` dependency ~5 levels up, which I'd think would affect a lot of the UI if it were problematic, due to which I assumed a quick walkthrough of the UI pages should give confidence, which is what I did.

The [change log 3.1.23 to 3.2.0](https://github.com/ai/nanoid/blob/main/CHANGELOG.md#3123) also looked safe.

PR #2785, opened 2022-01-25
Fix bug with cursor jumping to end (BW-1046).

cahrens:
Downgrading to the earlier version, which does not require memoization. The editor is being recreated on every keystroke, but that was true originally (created AS-1099).

This is fallout from https://github.com/DataBiosphere/terra-ui/pull/2775

I tested with both the Workspace Description editor and the Data Snapshot Description editor.

### DIFF
--- a/src/components/markdown.js
+++ b/src/components/markdown.js
@@ -1,7 +1,7 @@
 import DOMPurify from 'dompurify'
 import _ from 'lodash/fp'
 import { marked } from 'marked'
-import { lazy, Suspense, useMemo } from 'react'
+import { lazy, Suspense } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
 import { centeredSpinner } from 'src/components/icons'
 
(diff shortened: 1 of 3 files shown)


src/components/markdown.js
```diff
@@ -48,15 +48,19 @@ export const newWindowLinkRenderer = (href, title, text) => {
...
+        autofocus: true,
+        renderingConfig: {
+          singleLineBreaks: false
+        },
+        previewClass: ['editor-preview', 'markdown-body'],
+        previewRender: renderAndSanitizeMarkdown,
+        status: false
+      }
+    }, props)
+  }, [])
```
cahrens: I get a warning here about needing to pass in `props`. However, when I do pass in `props` then I see the original bug. I think I don't quite understand how to use `useMemo` and `_.merge` together.

(React Hook useMemo has a missing dependency: 'props'. Either include it or remove the dependency array)

Here is a description of the issue I'm trying to fix:
https://github.com/RIP21/react-simplemde-editor#options

Note that with this code change, the bug goes away in both places where the editor is used. However, I'm noticing that in the snapshot data description usage only, if you delete all the text in the description and press Save the deletion is not honored-- it is if you have **at least one character left in the editor**.

cahrens: I went back to an old version of dev (before my library upgrade) and verified that the bug where deleting all text in the snapshot data description field is not honored predated my work. I will create a separate bug for that (AS-1098).

cahrens:
@kyuksel only integration failure was run-notebook. Can you take a look (and manually test) in hopes I can merge this in time for the release?

kyuksel (approved):
Tested via pairing.

cjllanwarne (approved):

cahrens:
Integration test failures were known flaky issues. Merging in hopes of making the 11 AM terra UI release train.

PR #2789, opened 2022-01-27
[QA-1696] Wait for spinner to appear (before waiting for it to disappear)

xuscrbw:
We normally `waitForNoSpinners` prior to clicking `Edit`. However, it's possible that we do that before the spinner actually appears causing synchronization issues to manifest themselves in a later step. Therefore, we are adding a micro sleep to make sure that the spinner is rendered by the time we start waiting for the spinner to go away.


integration-tests/tests/run-notebook.js
```diff
@@ -26,6 +26,7 @@ const testRunNotebookFn = _.flow(
   await click(page, clickable({ text: 'Create Notebook' }))
   await click(page, clickable({ textContains: notebookName }))
   await waitForNoSpinners(page)
+  await findElement(page, clickable({ text: 'Edit' }))
   await click(page, clickable({ text: 'Edit' }))
```
cahrens, line 32: Doesn't a click first wait for the element to appear?

dmohs: It does:
```javascript
const click = async (page, xpath, options) => {
  return (await page.waitForXPath(xpath, options)).click()
}
```
The `findElement` should be redundant.

dmohs (approved):

integration-tests/tests/run-notebook.js
breilly2: There is a blocking spinner while the notebook preview loads. It's possible that the spinner hasn't yet rendered when the `await waitForNoSpinners(page)` call happens and then the test clicks (or at least thinks it clicks) "Edit" while the spinner is there, which means the click gets ignored.

This might be a case for a micro-sleep (probably no more than 500ms) in the test (including a comment describing the gap that the sleep is bridging).

kyuksel: What Brian described sounds plausible to me.

Pre-mature follow-up question: If  ^^^ (hopefully) is the culprit indeed, would it make sense to insert the micro sleep in `waitForNoSpinners` since that function is called right after clicking something, or is it that whether we'd need the micro sleep really depends on what happens when something is clicked prior?

breilly2: I think the need for a micro-sleep is an exception. Usually, a click would update state in a way that would cause the spinner to appear right away. In some cases, though, there's an intermediate state change... click causes state change which causes another state change which renders the spinner. It's that extra step (which I haven't 100% confirmed is the case here but, knowing that this is having to call out to calhoun to render a preview and display it in an iframe, I wouldn't be surprised) that I think could be creating the gap that a micro-sleep could bridge.

integration-tests/tests/run-notebook.js
```diff
@@ -26,6 +26,9 @@ const testRunNotebookFn = _.flow(
   await click(page, clickable({ text: 'Create Notebook' }))
   await click(page, clickable({ textContains: notebookName }))
   await waitForNoSpinners(page)
+  //Adding a short microsleep to make sure that we finish waiting for the spinner
+  //Make sure that the edit button is clicked
+  await delay(500)
```
breilly2: My hypothesis is that the problem is caused by this sequence of events:

1. Puppeteer test clicks notebook
2. Terra renders Edit button
3. Puppeteer test sees Edit button
4. Terra renders spinner
5. Puppeteer test clicks where the Edit button is located, but the click is swallowed by the spinner
6. Terra removes spinner and renders notebook preview
7. Test times out and screenshot shows notebook preview

If that's the case, then the test needs to micro-sleep _before_ `waitForNoSpinners`.

I know that I'm being very particular about this, but I think we need to be very precise in both the execution and documentation of these fixes. We have to leave enough context so that the next person who has to change the test (because something about the page behavior changed) understands whether or not the reason for a micro-sleep is still valid. If they have to guess and guess wrong, they will either re-introduce the flakiness or leave unnecessary sleeps in the tests, both of which lead to test suite pain down the road. (Believe me, this is the 4th time I've witnessed/participated in the evolution of a suite of browser-driven tests.)

integration-tests/tests/run-notebook.js
kyuksel: Makes sense. Thanks.

integration-tests/tests/run-notebook.js
```diff
@@ -25,6 +25,8 @@ const testRunNotebookFn = _.flow(
   await select(page, 'Language', 'Python 3')
   await click(page, clickable({ text: 'Create Notebook' }))
   await click(page, clickable({ textContains: notebookName }))
+  //Adding a short microsleep to make sure that we finish waiting for the spinner to stop blocking the click
```
kyuksel: My understanding of why we add the micro sleep was more like

```suggestion
  // We are adding a short micro sleep to make sure that the spinner is
  // rendered by the time we start waiting for the spinner to go away.
```

Am I missing something?

Cc: @breilly2

kyuksel:
FYI @xuscrbw that it could be very helpful to take advantage of the tool introduced by [this PR](https://github.com/DataBiosphere/terra-ui/pull/2792) while testing your change.

integration-tests/tests/run-notebook.js
breilly2: Yes, that's exactly it. Thank you for boiling it down to the simplest explanation!

cahrens (approved):

PR #2790, opened 2022-01-27
[QA-1701] Increase timout and log when workflows start Running

cahrens:
See comments and timing information in [JIRA](https://broadworkbench.atlassian.net/browse/QA-1701?focusedCommentId=55559).

Since we are actively working on submission speed, the Batch team decided it made the most sense to increase the timeout in the short term, and then examine the timeout interval again once our Q1 performance work is complete.

### DIFF
--- a/integration-tests/tests/run-workflow-on-snapshot.js
+++ b/integration-tests/tests/run-workflow-on-snapshot.js
@@ -1,8 +1,8 @@
 const _ = require('lodash/fp')
 const fetch = require('node-fetch')
-const pRetry = require('p-retry')
+const { launchWorkflowAndWaitForSuccess } = require('./run-workflow')
 const { withWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, delay, dismissNotifications, fillInReplace, findElement, findText, input, select, signIntoTerra, waitForNoSpinners, findInGrid, navChild } = require('../utils/integration-utils')
+const { click, clickable, delay, dismissNotifications, fillInReplace, findElement, findText, input, select, signIntoTerra, waitForNoSpinners, navChild } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
@@ -61,19 +61,9 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await delay(1000) // Without this delay, the input field sometimes reverts back to its default value
   await click(page, clickable({ text: 'Save' }))
 
-  await delay(1000) // The Run Analysis button requires time to become enabled after hitting the save button
-  await click(page, clickable({ textContains: 'Run analysis' }))
+  await delay(1000) // The Run Analysis button (launchWorkflowAndWaitForSuccess) requires time to become enabled after hitting the save button
 
-  await click(page, clickable({ text: 'Launch' }))
-
-  // CHECK WORKFLOW SUCCEEDED AND RESULT IS WRITTEN
-  await pRetry(async () => {
-    try {
-      await findInGrid(page, 'Succeeded', { timeout: 65 * 1000 }) // long enough for the submission details to refresh
-    } catch (e) {
-      throw new Error(e)
-    }
-  }, { retries: 10, factor: 1 })
+  await launchWorkflowAndWaitForSuccess(page)
 
   await click(page, navChild('data'))
   await click(page, clickable({ textContains: 'Workspace Data' }))
--- a/integration-tests/tests/run-workflow.js
+++ b/integration-tests/tests/run-workflow.js
@@ -66,9 +66,9 @@ const launchWorkflowAndWaitForSuccess = async page => {
     } catch (e) {
       try {
         await findInGrid(page, 'Running', { timeout: 1000 })
-        console.log(`Workspace is running, elapsed time (minutes): ${(Date.now() - start) / (1000 * 60)}`)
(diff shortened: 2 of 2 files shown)


integration-tests/tests/run-workflow.js
```diff
@@ -50,23 +58,28 @@ const testRunWorkflowFn = _.flow(
...
+      }
       throw new Error(e)
     }
-  }, { retries: 10, factor: 1 })
-
-  await click(page, navChild('data'))
-  await click(page, clickable({ textContains: 'test_entity' }))
-  await findInDataTableRow(page, testEntity.name, testEntity.attributes.input)
-})
+  }, { retries: 15, factor: 1 })
```
cahrens, line 75: Increases timeout for workflow to finish to 15 minutes (from 10). Note that we hope to ultimately decrease this after we have improved the submission speed to Cromwell (this is captured in https://broadworkbench.atlassian.net/browse/BW-1057).

To understand better what percentage of the time waiting is in "Submitted" vs. "Running", I added the log messages. In a few of the failures that we have seen, the workflow doesn't even make it to "Running" within the previous 10 minute timeout.

cahrens: Example logging message in test run:

````
page.http.res 200 GET https://rawls.dsde-dev.broadinstitute.org/api/workspaces/saturn-integration-test-dev/test-workspace-22765/snapshots/v2?offset=0&limit=1000
 PASS  jest/1-run-workflow-on-snapshot.integration-test.js (360.182 s)
  ● Console

    console.info
      Workflow not yet running, elapsed time (minutes): 1.1000333333333334

      at pRetry.retries (tests/run-workflow.js:71:17)

    console.info
      Workflow is running, elapsed time (minutes): 2.20035

      at pRetry.retries (tests/run-workflow.js:69:17)

    console.info
      Workflow is running, elapsed time (minutes): 3.3006333333333333

      at pRetry.retries (tests/run-workflow.js:69:17)

File: integration-tests/tests/run-workflow-on-snapshot.js [hunk 2]
```diff
@@ -61,19 +61,9 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await delay(1000) // Without this delay, the input field sometimes reverts back to its default value
   await click(page, clickable({ text: 'Save' }))
 
-  await delay(1000) // The Run Analysis button requires time to become enabled after hitting the save button
```
cahrens, line 64: FYI @ungwudik, I am sharing this increase in timeout and logging between the 2 run-workflow tests. This is certainly not the only cause of run-workflow-on-snapshot failing (https://broadworkbench.atlassian.net/browse/QA-1700), but it might help reduce the number of failures.

ungwudik: Thanks @cahrens!
I have not started on QA-1700, but will keep your updates here (and in the ticket itself) in mind.

integration-tests/tests/run-workflow.js
```diff
@@ -50,23 +58,28 @@ const testRunWorkflowFn = _.flow(
...
+  disablePageLogging()
```
aednichols, line 76: It's not immediately clear to me why we're storing this function in a value and then calling it - does `page` from line 61 go out of scope?

aednichols (approved):

integration-tests/tests/run-workflow.js
cahrens: `logPageConsoleMessages` enables logging and returns a function that you can use to turn it off later. At least that's what the code looks like to me.

cjllanwarne (approved):

integration-tests/tests/run-workflow.js
kyuksel: Is there a way to not print out references to code lines like `at pRetry.retries (tests/run-workflow.js:71:17)`? They don't seem helpful and make the output more verbose.

kyuksel:
> See comments and timing information in JIRA. Since we are actively working on submission speed, the Batch team decided it made the most sense to increase the timeout in the short term, and then examine the timeout interval again once our Q1 performance work is complete.

This sounds reasonable except that I've been seeing a stark discrepancy between test runs against `dev` vs. `alpha/staging`, which makes me wonder if it's the backend slow, why doesn't it cause the same timeout issues in `alpha/staging`  🤔

integration-tests/tests/run-workflow.js
cahrens: @kyuksel honestly I don't know, but since this is only test output, I don't think it's worth the investigation time.

cahrens:
> This sounds reasonable except that I've been seeing a stark discrepancy between test runs against `dev` vs. `alpha/staging`, which makes me wonder if it's the backend slow, why doesn't it cause the same timeout issues in `alpha/staging` 🤔

I do wonder what is different about the 2 environments. I have heard that they are supposed to have similar resources/configuration, but maybe that is not true. I could try to ask around some more but would like to go ahead and get this change in.

kyuksel (approved):

integration-tests/tests/run-workflow.js
cahrens: Following up on this, `logPageConsoleMessages` doesn't seem to be intercepting the console.info calls. I will follow up with @dmohs about how I am supposed to use the method.

PR #2793, opened 2022-01-31
[DC-181] Updated filter logic for the data catalog

lu-c:
Updated the filter logic for data catalog so:

- Filters among the same filter category are ORed
- Filters across different categories are ANDed

Previous behavior was universal ANDing.


petesantos (commented):
Just to give some quick feedback: I tested this by clicking around a bit and am seeing some issues:
1. Using 'see more' no longer adds the item to the filtered list in the left column
2. Within the modal, specifically File Type, if I click an entry like "Rds.gz" it ends up selecting "Rds"

I'll continue looking to see if I can find the issue

lu-c:
> Just to give some quick feedback: I tested this by clicking around a bit and am seeing some issues:
> 
> 1. Using 'see more' no longer adds the item to the filtered list in the left column
> 2. Within the modal, specifically File Type, if I click an entry like "Rds.gz" it ends up selecting "Rds"
> 
> I'll continue looking to see if I can find the issue

Oh boy, these are big problems, I'm not sure how this PR touches those functions @.@ I'll take a look at it today!

lu-c:
> Just to give some quick feedback: I tested this by clicking around a bit and am seeing some issues:
> 
> 1. Using 'see more' no longer adds the item to the filtered list in the left column
> 2. Within the modal, specifically File Type, if I click an entry like "Rds.gz" it ends up selecting "Rds"
> 
> I'll continue looking to see if I can find the issue

Should be fixed now!

src/pages/library/common.js
```diff
@@ -150,12 +154,12 @@ const FilterBar = ({ name, onClear, onFilterByLetter, onFilterBySearchText, onSo
...
 }
 
-const FilterModal = ({ name, labels, setShowAll, onTagFilter, listDataByTag, lowerSelectedTags }) => {
+const FilterModal = ({ name, labels, setShowAll, onTagFilter, listDataByTag, lowerSelectedTags, filteredData }) => {
   // Filter Modal Vars
   const filteredTags = (_.flow(
     _.map(label => {
       const lowerTag = _.toLower(label)
-      return [lowerTag, listDataByTag[lowerTag]]
+      return [lowerTag, getFacetCount({ listDataByTag, filteredData, section: name, label: lowerTag })]
```
samanehsan: Should the order of `filteredData` and `listDataByTag` be swapped here?

lu-c: The parameter for this function takes an object, so it doesnt matter what order it's provided in - Would you recommend we still swap the order?

samanehsan: Got it! Yeah I think that would make it a bit clearer!

petesantos (approved):

samanehsan (approved):
Lgtm! :octocat:

PR #2794, opened 2022-01-31
[BW-1015] Display "Waiting for Cloud Quota" status for tasks within a workflow

cahrens:
This exposes "Awaiting Cloud Quota" as a Cromwell task status on the Workflows Dashboard (only). Previously this would be displayed as "Running", which was not helpful for users who wondered why tasks appeared to be "stuck". Example display (CPU limit has been reduced in a specific region that this test WDL is set to run on):

image

Things to note: 

1) There is a "learn more" link that should go to https://support.terra.bio/hc/en-us/articles/360029071251
2) There is a unique tooltip in the table for this status. A tooltip was already being shown, but just duplicating the cell content.

To test:
1) I can share a workspace/WDL that will hit the cloud quota issue (CPU quota has been reduced to 1 in a particular zone). 
2) cromwell-dev has to be set to develop, as the new version of Cromwell supporting this hasn't yet been released.

### DIFF
--- a/src/components/job-common.js
+++ b/src/components/job-common.js
@@ -89,7 +89,7 @@ export const collapseCromwellStatus = (executionStatus, backendStatus) => {
     case 'Running':
       return backendStatus === 'AwaitingCloudQuota' ? statusType.waitingForCloudQuota : statusType.running
     default:
-      return `Unexpected status (${status})`
+      return statusType.unknown
   }
 }
 
--- a/src/components/job-common.js
+++ b/src/components/job-common.js
@@ -16,27 +16,27 @@ export const addCountSuffix = (label, count = undefined) => {
 export const statusType = {
   succeeded: {
     id: 'succeeded',
-    label: _ => 'Succeeded',
+    label: () => 'Succeeded',
     icon: style => icon('check', { size: iconSize, style: { color: colors.success(), ...style } })
   },
   failed: {
     id: 'failed',
-    label: _ => 'Failed',
+    label: () => 'Failed',
     icon: style => icon('warning-standard', { size: iconSize, style: { color: colors.danger(), ...style } })
   },
   running: {
     id: 'running',
-    label: _ => 'Running',
+    label: () => 'Running',
     icon: style => icon('sync', { size: iconSize, style: { color: colors.dark(), ...style } })
   },
   submitted: {
     id: 'submitted',
-    label: _ => 'Submitted',
+    label: () => 'Submitted',
     icon: style => icon('clock', { size: iconSize, style: { color: colors.dark(), ...style } })
   },
   waitingForCloudQuota: {
     id: 'waitingForQuota',
-    label: _ => 'Submitted, Awaiting Cloud Quota',
+    label: () => 'Submitted, Awaiting Cloud Quota',
     icon: style => icon('error-standard', { size: iconSize, style: { color: colors.warning(), ...style } }),
     moreInfoLink: 'https://support.terra.bio/hc/en-us/articles/360029071251',
     moreInfoLabel: 'Learn more about cloud quota',
@@ -94,7 +94,7 @@ export const collapseCromwellStatus = (executionStatus, backendStatus) => {
 }
 
 /**
- * Returns the icon for display with workflow or submission status.
+ * Returns the icon for display with workflow or submission status, using `collapseStatus` to determine the appropriate icon.
  */
 export const statusIcon = (status, style) => {
   return collapseStatus(status).icon(style)
--- a/src/components/job-common.js
+++ b/src/components/job-common.js
@@ -43,7 +43,7 @@ export const statusType = {
   },
   unknown: {
     id: 'unknown',
-    label: rawStatus => `Unexpected status (${rawStatus})`,
+    label: executionStatus => `Unexpected status (${executionStatus})`,
     icon: style => icon('question', { size: iconSize, style: { color: colors.dark(), ...style } })
   }
 }
--- a/src/components/job-common.js
+++ b/src/components/job-common.js
@@ -39,6 +39,7 @@ export const statusType = {
     label: _ => 'Submitted, Awaiting Cloud Quota',
     icon: style => icon('error-standard', { size: iconSize, style: { color: colors.warning(), ...style } }),
     moreInfoLink: 'https://support.terra.bio/hc/en-us/articles/360029071251',
+    moreInfoLabel: 'Learn more about cloud quota',
     tooltip: 'Delayed by Google Cloud Platform (GCP) quota limits. Contact Terra Support to request a quota increase.'
   },
   unknown: {
@@ -103,7 +104,7 @@ export const statusIcon = (status, style) => {
  * Returns the rendered status line, based on icon function, label, and style.
  */
 export const makeStatusLine = (iconFn, label, style) => div(
-  { style: { display: 'flex', alignItems: 'center', fontSize: 14, textTransform: 'capitalize', ...style } },
+  { style: { display: 'flex', alignItems: 'center', fontSize: 14, ...style } },
   [iconFn({ marginRight: '0.5rem' }), label]
 )
 
@@ -112,7 +113,7 @@ export const makeStatusLine = (iconFn, label, style) => div(
  */
 export const makeCromwellStatusLine = (executionStatus, backendStatus) => {
   const collapsedStatus = collapseCromwellStatus(executionStatus, backendStatus)
-  return h(TooltipCell, { tooltip: collapsedStatus.tooltip },
+  return h(TooltipCell, { tooltip: collapsedStatus.tooltip }, // Note that if the tooltip is undefined, a default will be shown
     [makeStatusLine(style => collapsedStatus.icon(style), collapsedStatus.label(executionStatus), { marginLeft: '0.5rem' }
     )]
   )
--- a/src/components/job-common.js
+++ b/src/components/job-common.js
@@ -14,38 +14,37 @@ export const addCountSuffix = (label, count = undefined) => {
 }
 
 export const statusType = {
-  // TODO: We probably don't need the aria labels (if the icon is always used beside text). They should be aria-hidden.
   succeeded: {
     id: 'succeeded',
     label: _ => 'Succeeded',
-    icon: style => icon('check', { size: iconSize, style: { color: colors.success(), ...style }, 'aria-label': 'success' })
+    icon: style => icon('check', { size: iconSize, style: { color: colors.success(), ...style } })
   },
   failed: {
     id: 'failed',
     label: _ => 'Failed',
-    icon: style => icon('warning-standard', { size: iconSize, style: { color: colors.danger(), ...style }, 'aria-label': 'failed' })
+    icon: style => icon('warning-standard', { size: iconSize, style: { color: colors.danger(), ...style } })
   },
   running: {
     id: 'running',
     label: _ => 'Running',
-    icon: style => icon('sync', { size: iconSize, style: { color: colors.dark(), ...style }, 'aria-label': 'running' })
+    icon: style => icon('sync', { size: iconSize, style: { color: colors.dark(), ...style } })
   },
   submitted: {
     id: 'submitted',
     label: _ => 'Submitted',
-    icon: style => icon('clock', { size: iconSize, style: { color: colors.dark(), ...style }, 'aria-label': 'submitted' })
+    icon: style => icon('clock', { size: iconSize, style: { color: colors.dark(), ...style } })
   },
   waitingForCloudQuota: {
     id: 'waitingForQuota',
     label: _ => 'Submitted, Awaiting Cloud Quota',
-    icon: style => icon('error-standard', { size: iconSize, style: { color: colors.warning(), ...style }, 'aria-label': 'waiting for cloud quota' }),
+    icon: style => icon('error-standard', { size: iconSize, style: { color: colors.warning(), ...style } }),
     moreInfoLink: 'https://support.terra.bio/hc/en-us/articles/360029071251-Google-Cloud-quotas-What-are-they-and-how-do-you-request-more-',
     tooltip: 'Delayed by Google Cloud Platform (GCP) quota limits. Contact Terra Support to request a quota increase.'
   },
   unknown: {
     id: 'unknown',
     label: rawStatus => `Unexpected status (${rawStatus})`,
-    icon: style => icon('question', { size: iconSize, style: { color: colors.dark(), ...style }, 'aria-label': 'unknown' })
+    icon: style => icon('question', { size: iconSize, style: { color: colors.dark(), ...style } })
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -89,7 +98,7 @@ const WorkflowDashboard = _.flow(
       setWorkflow(metadata)
       setFetchTime(Date.now() - timeBefore)
 
-      if (_.includes(wf.status, ['Running', 'Submitted'])) {
+      if (_.includes(collapseStatus(metadata.status), [statusType.running, statusType.submitted])) {
```
cahrens, line 96: `wf.status` is always undefined, so auto-refresh was not working on this page.

src/components/job-common.js
```diff
@@ -2,76 +2,120 @@ import { Fragment } from 'react'
...
-      return 'succeeded'
+      return statusType.succeeded
     case 'Aborting':
     case 'Aborted':
     case 'Failed':
-      return 'failed'
+      return statusType.failed
     case 'Running':
-      return 'running'
-    case 'Submitted':
```
cahrens, line 36: Removed `Submitted` because there is no Cromwell `Submitted` status, and this caused confusion when trying to understand the code. Here are the possible Cromwell executionStatuses: https://github.com/broadinstitute/cromwell/blob/17203758fa4608fdfb9e77194f67579b67343e39/core/src/main/scala/cromwell/core/ExecutionStatus.scala#L11-L26

src/components/job-common.js
```diff
@@ -2,76 +2,120 @@ import { Fragment } from 'react'
...
+const iconSize = 24
+export const addCountSuffix = (label, count = undefined) => {
+  return label + (count === undefined ? '' : `: ${count}`)
+}
+
+export const statusType = {
+  succeeded: {
+    id: 'succeeded',
+    label: _ => 'Succeeded',
+    icon: style => icon('check', { size: iconSize, style: { color: colors.success(), ...style } })
```
cahrens, line 20: Note that I removed the aria-labels from the icons. You only want to put an aria-label on an icon when the information represented by the icon is not otherwise available; otherwise, you want to mark it aria-hidden (which our icon class does if a label is not present). In all the places where we show these icons, we represent the icon status through a label or through summary text, so the aria-label was just adding extra noise when navigating through the page.

Example locations:
image
image

image

The only tricky place is the one below, and the information being displayed in the popup is already being summarized on the cell entry via an aria-label.

image

image

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -116,7 +116,8 @@ const SubmissionDetails = _.flow(
     sort.direction === 'asc' ? _.identity : _.reverse
   )(workflows)
 
-  const { succeeded, failed, running, submitted } = _.groupBy(wf => collapseStatus(wf.status), workflows)
+  // Note: these variable names match the id values of statusType.
+  const { succeeded, failed, running, submitted } = _.groupBy(wf => collapseStatus(wf.status).id, workflows)
```
cahrens: It would be nice if these could be declared as  `{ statusType.succeeded.id, statusType.failed.id …}`, but as far as I can tell that is not possible.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -55,7 +55,7 @@ const SubmissionDetails = _.flow(
   useEffect(() => {
     const initialize = withErrorReporting('Unable to fetch submission details',
       async () => {
-        if (_.isEmpty(submission) || _.some(({ status }) => _.includes(collapseStatus(status), ['running', 'submitted']), submission.workflows)) {
+        if (_.isEmpty(submission) || _.some(({ status }) => _.includes(collapseStatus(status), [statusType.running, statusType.submitted]), submission.workflows)) {
           if (!_.isEmpty(submission)) {
             await Utils.delay(60000)
```
cahrens, line 60: @zarsky-broad it looks like you originally wrote this code. Do you recall what this delay is meant to do?

cahrens:
Note to reviewers: I'm being plagued by unrelated integration test failures. I've been updating the JIRA tickets for the individual tests, but am going to give up continuously re-running as I expect that I will have code review feedback to address before merging.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
zarsky-broad: If there are any in-progress submissions, it'll refresh the list every minute.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -30,25 +29,33 @@ const styles = {
...
-    makeRow(running, runningIcon(), 'Running'),
-    makeRow(succeeded, successIcon(), 'Succeeded'),
-    makeRow(failed, failedIcon(), 'Failed'),
-    _.map(([name, count]) => makeRow(count, unknownIcon(), name), _.toPairs(others))
+    makeRow(submitted, statusType.submitted),
+    makeRow(waitingForQuota, statusType.waitingForCloudQuota),
+    makeRow(running, statusType.running),
+    makeRow(succeeded, statusType.succeeded),
+    makeRow(failed, statusType.failed),
+    h(Fragment, _.map(([label, count]) => makeRow(count, statusType.unknown, label), _.toPairs(others)))
```
cahrens: Note the `Fragment` usage here to avoid this issue: `index.js:1 Warning: Each child in a list should have a unique "key" prop.`. `._map` returns an array.

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -92,17 +92,15 @@ const CallTable = ({ namespace, name, submissionId, workflowId, callName, callOb
...
-              return failures ? [
-                h(Link, {
-                  style: { marginLeft: '0.5rem' },
-                  onClick: () => setFailuresModalParams({ index, attempt, failures })
-                }, [
-                  div({ style: { display: 'flex', alignItems: 'center' } }, [
-                    icon('warning-standard', { size: 18, style: { color: colors.warning(), marginRight: '0.5rem' } }),
-                    `${failureCount} Message${failureCount > 1 ? 's' : ''}`
-                  ])
+              return !!failureCount && h(Link, {
```
cahrens, line 95: The `Link` was previously in an array, which causes this warning: `index.js:1 Warning: Each child in a list should have a unique "key" prop.`

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cahrens: Discussed in mob, and now I understand this much better. Thank you!

src/components/job-common.js
dmohs: This is an absolutely fantastic PR note. Thank you for using your PR to teach me something.

aednichols (approved):
👍 👍 👍 

Also, I read the linked support article and it is really good, so glad we have that reference.

src/components/job-common.js
```diff
@@ -2,76 +2,121 @@ import { Fragment } from 'react'
...
+  },
+  failed: {
+    id: 'failed',
+    label: _ => 'Failed',
```
dmohs: The common pattern in the codebase is `() => 'Failed'`. I do like the `_` pattern because it devotes even less space to the arguments, which are ignored. That makes sense to me. In the interest of consistency, I would be in favor of defining something like:
```javascript
const val = v => () => v
```
So that the above could become:
```javascript
label: val('Failed')
```
For this PR, I'd probably advocate for keeping the common style, since consistency on trivial details is a quality signal for people who are new to the codebase. I would definitely support a PR that changed this pattern to be consistent throughout. Happy with whatever you decide.

src/components/job-common.js
dmohs: Good call.

src/components/job-common.js
```diff
@@ -2,76 +2,121 @@ import { Fragment } from 'react'
...
-      return failedIcon(style)
-    case 'running':
-      return runningIcon(style)
-    case 'submitted':
-      return submittedIcon(style)
-    default:
-      return unknownIcon(style)
-  }
+/**
+ * Returns the icon for display with workflow or submission status.
```
dmohs: This style of comment is controversial, so I'll share my take, for what it's worth. First, a function should not make judgments on where it is to be used. Instead, it should specify its behavior and let the user of the function decide if its behavior is appropriate for their use case.

I almost always start with a verb. If this were called `getStatusIcon(status, style)`, the signature tells me just as much as the comment, and will be less likely to become incorrect as the code evolves. More generally, if I am inclined to add a comment to describe what something does, I first see if I can refactor the code to make the comment unnecessary.

src/pages/workspaces/workspace/JobHistory.js
```diff
@@ -66,10 +66,10 @@ const statusCell = (workflowStatuses, status) => {
...
       tbody({}, [
         tr({}, [
-          td(styles.statusDetailCell, [successIcon()]),
-          td(styles.statusDetailCell, [failedIcon()]),
-          td(styles.statusDetailCell, [runningIcon()]),
-          td(styles.statusDetailCell, [submittedIcon()])
+          td(styles.statusDetailCell, [statusType.succeeded.icon()]),
+          td(styles.statusDetailCell, [statusType.failed.icon()]),
+          td(styles.statusDetailCell, [statusType.running.icon()]),
+          td(styles.statusDetailCell, [statusType.submitted.icon()])
```
dmohs, line 72: Very nice.

src/pages/workspaces/workspace/jobHistory/CallTable.js
dmohs: Thank you!!!

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -160,10 +161,10 @@ const SubmissionDetails = _.flow(
...
         div({ style: { display: 'grid', gridTemplateRows: '1fr auto' } }, [
           makeSection('Workflow Statuses', [
-            succeeded && makeStatusLine(successIcon, `Succeeded: ${succeeded.length}`, { marginTop: '0.5rem' }),
-            failed && makeStatusLine(failedIcon, `Failed: ${failed.length}`, { marginTop: '0.5rem' }),
-            running && makeStatusLine(runningIcon, `Running: ${running.length}`, { marginTop: '0.5rem' }),
-            submitted && makeStatusLine(submittedIcon, `Submitted: ${submitted.length}`, { marginTop: '0.5rem' })
+            succeeded && makeStatusLine(statusType.succeeded.icon, addCountSuffix(statusType.succeeded.label(), succeeded.length), { marginTop: '0.5rem' }),
+            failed && makeStatusLine(statusType.failed.icon, addCountSuffix(statusType.failed.label(), failed.length), { marginTop: '0.5rem' }),
+            running && makeStatusLine(statusType.running.icon, addCountSuffix(statusType.running.label(), running.length), { marginTop: '0.5rem' }),
+            submitted && makeStatusLine(statusType.submitted.icon, addCountSuffix(statusType.submitted.label(), submitted.length), { marginTop: '0.5rem' })
```
dmohs: Similarly here, if you did not decompose the groups:
```javascript
'succeeded failed running submitted'.split(' ').filter(s => groups[s])
  .map(s =>
  makeStatusLine(statusType[s].icon, addCountSuffix(statusType[s].label(), groups[s].length)))
```

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -30,25 +29,33 @@ const styles = {
...
-    makeRow(submitted, submittedIcon(), 'Submitted'),
-    makeRow(running, runningIcon(), 'Running'),
-    makeRow(succeeded, successIcon(), 'Succeeded'),
-    makeRow(failed, failedIcon(), 'Failed'),
-    _.map(([name, count]) => makeRow(count, unknownIcon(), name), _.toPairs(others))
+    makeRow(submitted, statusType.submitted),
+    makeRow(waitingForQuota, statusType.waitingForCloudQuota),
+    makeRow(running, statusType.running),
+    makeRow(succeeded, statusType.succeeded),
+    makeRow(failed, statusType.failed),
```
dmohs:
```javascript
  'submitted waitingForQuota running succeeded failed'.split(' ').map(s => makeRow(others[s], statusType[s]))
```

dmohs (approved):

src/components/job-common.js
cahrens: Thanks, I forgot that "_" is not the style in this codebase. I will change it to "()".

src/components/job-common.js
cahrens: The thing I was trying to help clarify was that there are 2 types of status: workflow/submission status, and Cromwell status. They have different "collapse" methods, and this method relies on `collapseStatus`, which is specific to workflow/submission status.

This probably isn't what you are hoping for, but I'm going to make the comment more explicit.

src/components/job-common.js
zarsky-broad: @dmohs you're describing `_.identity`, so maybe a util with a shorter alias for that, like `iden(val)`?

zarsky-broad: That said, there's something I wrote in utils but never exported called `maybeCall`, which would allow whoever's ingesting this to handle cases where something is only sometimes a function; maybe the place using this could use this, and then label wouldn't have to be a function at all? Disclaimer that I'm not looking too deeply, just reading this and thinking.

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
cahrens: Thank you! This simplifies the code quite a bit.

src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js
```diff
@@ -30,26 +29,31 @@ const styles = {
...
-
-  return h(Fragment, [
-    makeRow(submitted, submittedIcon(), 'Submitted'),
-    makeRow(running, runningIcon(), 'Running'),
-    makeRow(succeeded, successIcon(), 'Succeeded'),
-    makeRow(failed, failedIcon(), 'Failed'),
-    _.map(([name, count]) => makeRow(count, unknownIcon(), name), _.toPairs(others))
-  ])
+  return h(Fragment, _.concat(
+    [statusType.submitted.id, statusType.waitingForCloudQuota.id, statusType.running.id, statusType.succeeded.id, statusType.failed.id].filter(
```
cahrens, line 53: @dmohs had you thought of a way to do this that got rid of the need for the decomposition entirely (given "unknownStatus", formerly known as "others")? I wasn't clear from your comment, and this was the best I could come up with. I still think it's an improvement over what I had previously.

FWIW, here's an example with "unknownStatuses":

image

src/components/job-common.js
cahrens: I actually came to the conclusion that the method is no longer helpful, so I have deleted it (before my refactoring this method had more functionality).

src/components/job-common.js
cahrens: @zarsky-broad `maybeCall` is clever, and I'm somewhat tempted. But for simplicity I think I will stick with what I have. FWIW, `statusType.unknown` does use an argument passed to `label`, which is why I made it a function.

PR #2796, opened 2022-02-01
Updates for node-fetch (BW-1062).

cahrens:
Per security vulnerability, upgrade to node-fetch to version 2.6.7.

node-fetch is used by react-json-view in Terra to display JSON. For example,

image-20220201-214139

In the integration tests, it is used by Puppeteer. I upgraded Puppeteer, and given that I managed to get all tests passing, I assume the new version of Puppeteer is OK; the only breaking change was the name of a method that had a typo (and we never called it).

### DIFF
--- a/integration-tests/package.json
+++ b/integration-tests/package.json
@@ -32,6 +32,6 @@
     "lodash": "^4.17.21",
     "nodemon": "^2.0.12",
     "p-retry": "^4.6.1",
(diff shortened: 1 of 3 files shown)


cjllanwarne (approved):

salonishah11 (approved):
LGTM 👍

dmohs (approved):

PR #2797, opened 2022-02-01
Upgrade tmpl (BW-1063).

cahrens:
This is a JavaScript template library that I doubt anything actually in Terra actually uses (but it is a transitive dependency). It hasn't been under active deployment for about 7 years… https://www.npmjs.com/package/tmpl

Security vulnerability update: 

Remediation
Upgrade tmpl to version 1.0.5 or later. For example:

tmpl@^1.0.5:
  version "1.0.5"
Always verify the validity and compatibility of suggestions with your codebase.

Details
CVE-2021-3777
Moderate severity
Vulnerable versions: < 1.0.5
Patched version: 1.0.5
nodejs-tmpl is vulnerable to Inefficient Regular Expression Complexity which may lead to resource exhaustion.

### DIFF
--- a/integration-tests/yarn.lock
+++ b/integration-tests/yarn.lock
@@ -6005,9 +6005,9 @@ resolve@^1.20.0:
   linkType: hard
 
 "tmpl@npm:1.0.x":
-  version: 1.0.4
-  resolution: "tmpl@npm:1.0.4"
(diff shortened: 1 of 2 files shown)


cjllanwarne (approved):

petesantos (approved):
LGTM - thanks for doing this! 
Based on `yarn why` this is used by jest. Ran some tests - they passed.

PR #2798, opened 2022-02-02
Remove delays in run-notebook test

petesantos:
## Overview
This PR removes the delay's from the `run-notebook` test.
In order to achieve this I did the following:

1. Created a utility that better handle's waiting on busy spinners
2. Added a small improvement to the a11y of the slide in drawer
3. Added a test utility to utilize this a11y improvement to detect when the drawer is open

I ran this test using the flake check, running the test over 100 times. The only issues I saw were related to actual system instability rather than flakiness with puppeteer.

## Results
In the last set of 100 the failures I saw were (all failures I saw on other runs were of these signatures): 

### Error checking TOS - not puppeteer related (2 errors)
1. The error thrown was: ` waiting for XPath `//*[@data-icon="loadingSpinner"]` to be hidden failed: timeout 30000ms exceeded`
    1. I consider this is an accurate description of the error. Along with the screenshot it shows the problem

### Error Checking Registration - not puppeteer related (1 error)
1. The error thrown was: ` waiting for XPath `//*[@data-icon="loadingSpinner"]` to be hidden failed: timeout 30000ms exceeded`
    1. Again, I consider this is an accurate description of the error. Along with the screenshot it shows the problem

### Blank page - app didn't load
1. The error thrown was: `Error: Evaluation failed: Response`
    1. I will see if I can improve the logging on this within the flake checker in a future PR
    2. From seeing this before, I believe the workspace was not created, preventing the test from getting off the launchpad

### DIFF
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -1,6 +1,6 @@
 const _ = require('lodash/fp')
 const { withRegisteredUser, withBilling, withWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, getAnimatedDrawer, signIntoTerra, findElement, navChild, waitForNoSpinners, waitForNoSpinnersAfterAction, select, fillIn, input, findIframe, findText, dismissNotifications } = require('../utils/integration-utils')
+const { click, clickable, getAnimatedDrawer, signIntoTerra, findElement, navChild, waitForNoSpinners, noSpinnersAfter, select, fillIn, input, findIframe, findText, dismissNotifications } = require('../utils/integration-utils')
 
 
 const notebookName = 'TestNotebook'
@@ -16,15 +16,15 @@ const testRunNotebookFn = _.flow(
   await dismissNotifications(page)
   await waitForNoSpinners(page)
 
-  await waitForNoSpinnersAfterAction(page, () => click(page, clickable({ textContains: workspaceName })))
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: workspaceName })) })
   await click(page, navChild('notebooks'))
   await click(page, clickable({ textContains: 'Create a' }))
   await fillIn(page, input({ placeholder: 'Enter a name' }), notebookName)
   await select(page, 'Language', 'Python 3')
 
-  await waitForNoSpinnersAfterAction(page, () => click(page, clickable({ text: 'Create Notebook' })))
-  await waitForNoSpinnersAfterAction(page, () => click(page, clickable({ textContains: notebookName })))
-  await waitForNoSpinnersAfterAction(page, () => click(page, clickable({ text: 'Edit' })))
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Create Notebook' })) })
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: notebookName })) })
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Edit' })) })
 
   await findElement(page, getAnimatedDrawer('Cloud Environment'))
   await click(page, clickable({ text: 'Create' }))
--- a/integration-tests/tests/run-notebook.js
+++ b/integration-tests/tests/run-notebook.js
@@ -1,6 +1,6 @@
 const _ = require('lodash/fp')
 const { withRegisteredUser, withBilling, withWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, delay, signIntoTerra, findElement, navChild, waitForNoSpinners, select, fillIn, input, findIframe, findText, dismissNotifications } = require('../utils/integration-utils')
+const { click, clickable, getAnimatedDrawer, signIntoTerra, findElement, navChild, waitForNoSpinners, waitForNoSpinnersAfterAction, select, fillIn, input, findIframe, findText, dismissNotifications } = require('../utils/integration-utils')
 
 
 const notebookName = 'TestNotebook'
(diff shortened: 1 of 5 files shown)


src/components/ModalDrawer.js
```diff
@@ -34,7 +34,7 @@ const ModalDrawer = ({ isOpen, onDismiss, width = 450, children, ...props }) =>
     mountOnEnter: true,
     unmountOnExit: true
   }, [transitionState => h(RModal, {
-    aria: { label: props['aria-label'], labelledby: props['aria-labelledby'], modal: true },
+    aria: { label: props['aria-label'], labelledby: props['aria-labelledby'], modal: true, hidden: transitionState !== 'entered' },
```
petesantos, line 37: I use `aria-hidden` in this case based on [recommendations for carousel widgets](https://www.w3.org/WAI/tutorials/carousels/animations/). While this is not a carousel, it shares the animation concept and I think should remain hidden to a screen reader until the animation is complete.

This also helps reduce testing complexity.

integration-tests/tests/run-notebook.js
```diff
@@ -14,37 +14,19 @@ const testRunNotebookFn = _.flow(
...
-  // spinner being rendered that simply waiting for no spinners does not work; the test recognizes
-  // that there aren't any spinners before the spinner has a chance to render. Additionally, even
-  // though the slide-in animation is supposedly only 200ms, experimentation has shown that even a
-  // 500ms delay in the test is not enough to guarantee that the UI elements have finished moving.
-  // Therefore, we start with a 2000ms delay, then make sure there are no spinners just in case the
-  // AJAX calls are unexpectedly slow.
-  await delay(2000)
-  waitForNoSpinners(page)
+
+  await waitForNoSpinnersAfterAction(page, () => click(page, clickable({ text: 'Create Notebook' })))
```
petesantos: Each of these actions is followed up by a busy spinner, so I pair taking the action with waiting on the spinner.

src/components/ModalDrawer.js
kyuksel: It may be good to add a comment to the code as well.

integration-tests/utils/integration-utils.js
```diff
@@ -119,6 +123,15 @@ const waitForNoSpinners = page => {
   return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
 }
 
+// waitForNoSpinnersAfterAction
```
kyuksel: Is the line above necessary?

integration-tests/utils/integration-utils.js
```diff
@@ -119,6 +123,15 @@ const waitForNoSpinners = page => {
...
+// This takes advantage of the mutationObserver in puppeteer by initializing the wait before taking an action.
+// This helps us guarantee that we will catch the spinner before it disappears
+const waitForNoSpinnersAfterAction = async (page, action) => {
+  const foundSpinner = page.waitForXPath('//*[@data-icon="loadingSpinner"]')
+  await Promise.all([foundSpinner, action()])
+  return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
```
kyuksel: Should we factor out the common code?

```suggestion
  return waitForNoSpinners(page)
```

kyuksel (approved):
Looks great 🚀 
A couple of nits inline.
Got 99/100 pass locally. The one failure seemed unrelated to test 😊

src/components/ModalDrawer.js
xuscrbw: I haven't seen the "transitionState !== 'entered'" before. How does this work?

integration-tests/utils/integration-utils.js
```diff
@@ -119,6 +123,15 @@ const waitForNoSpinners = page => {
   return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
 }
 
+// waitForNoSpinnersAfterAction
+// This takes advantage of the mutationObserver in puppeteer by initializing the wait before taking an action.
```
breilly2: [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) does look like it could be useful. This function is using `Promise.all` which may also be effective in this case. I'm still a little suspicious that there could be a gap here, but I also can't think of a case where this wouldn't work. 🤔  Anyway, I just wanted to make sure the code and comment are consistent.

integration-tests/tests/run-notebook.js
```diff
@@ -14,37 +14,19 @@ const testRunNotebookFn = _.flow(
   await click(page, clickable({ textContains: 'View Workspaces' }))
   await signIntoTerra(page, token)
   await dismissNotifications(page)
-  await findElement(page, clickable({ textContains: workspaceName }))
   await waitForNoSpinners(page)
-  await click(page, clickable({ textContains: workspaceName }))
-  await delay(5000)
+
+  await waitForNoSpinnersAfterAction(page, () => click(page, clickable({ textContains: workspaceName })))
```
breilly2: Nit: This reads oddly to me because of the back-to-back "wait for no spinner" statements. I wonder if `waitForNoSpinnersAfterAction` should be renamed to be more action-centric instead of spinner-centric. I haven't been able to think of a good name myself. 🤔

src/components/ModalDrawer.js
breilly2: This is fantastic! And probably an a11y win, too!

integration-tests/tests/run-notebook.js
zarsky-broad: I wonder if something more along the lines of `doThenSpinners` is better? I feel like `doAfterSpinners`reads better, but that's obviously putting the spinner wait on the next thing, so maybe just `await noSpinnersAfter(page, action)`?

integration-tests/utils/integration-utils.js
petesantos: The comment is meant to highlight that puppeteer is using `MutationObserver` under the hood. We are basically setting up the observer before the action, this should guarantee that we catch the spinner (though I hate using "guarantee")

Is this comment clearer?

``` 
Puppeteer works by internally using MutationObserver. We are setting up the listener before the 
action to ensure that the spinner rendering is captured by the observer, followed by waiting for the 
spinner to be removed
```

integration-tests/tests/run-notebook.js
petesantos: I like `await noSpinnersAfter(page, action)` with a slight tweak:

```
await noSpinnersAfter(page, { action })
```
The reason being `action` will need to be explicitly added, making it read consistently

kyuksel: +1 on `await noSpinnersAfter(page, action)`

petesantos: @kyuksel which signature do you prefer?

`await noSpinnersAfter(page, action)`

`await noSpinnersAfter(page, { action: foo})`

src/components/ModalDrawer.js
petesantos: @xuscrbw 
This expression will return a boolean in Javascript:
```
transitionState !== 'entered'
```
The `ModalDrawer` has a few states is sets internally as it is animating in and out. I am using the existing state logic to set the `aria-hidden` attribute on the component by using the boolean returned from the above expression.

integration-tests/tests/run-notebook.js
kyuksel: I think I'd prefer `await noSpinnersAfter(page, { action: foo})` because the call site would read more clearly; e.g.

```javascript
await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Create Notebook' }))})
```

vs.

```javascript
await noSpinnersAfter(page, () => click(page, clickable({ text: 'Create Notebook' })))
```

but I don't feel strongly.

samanehsan (approved):
Lgtm!

PR #2800, opened 2022-02-02
[DC-207] Fix link-to-new-workspace flakiness

samanehsan:


petesantos (approved):
LGTM!
I ran this test 500 times during our team's mobbing session. The only failures were on the workspace deletion Ajax call (workspace service was not responding).

integration-tests/tests/link-to-new-workspace.js
```diff
@@ -1,16 +1,16 @@
...
 const { enableDataCatalog, testWorkspaceName } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
 const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) => {
   await enableDataCatalog(page, testUrl, token)
   await click(page, clickable({ textContains: 'browse & explore' }))
-  await waitForNoSpinners(page)
 
+  await findText(page, 'Granted')
```
cahrens: So does this mean that the click operation does not wait for the text 'Granted' to be present? I would think that findText is unnecessary. (The click does do `await page.waitForXPath(xpath, options)`)

(Also, it looks like you have introduced some imports that are not needed-- findHeading and heading).

kyuksel: I was wondering the same.

petesantos: I do agree this is something that needs investigation.
I would like to get this fix in since it removes the flakiness of the existing test, preventing devs from seeing failures.

I created [this ticket](https://broadworkbench.atlassian.net/browse/DC-216) to track the issue.
Are you ok with allowing this ticket to merge with a follow up?

cahrens: Yes, but please put a comment in the code as well. I don't really want folks to see this pattern and think they need to do it everywhere.

integration-tests/tests/link-to-new-workspace.js
```diff
@@ -1,16 +1,16 @@
-const { checkbox, click, clickable, clickTableCell, fillIn, waitForNoSpinners } = require('../utils/integration-utils')
+const { checkbox, click, clickable, clickTableCell, fillIn, findHeading, findText, heading, waitForNoSpinners } = require('../utils/integration-utils')
```
cahrens: Are `findHeading` and `heading` used?

samanehsan: I'll take those out! They're not needed anymore

integration-tests/tests/link-to-new-workspace.js
samanehsan: > So does this mean that the click operation does not wait for the text 'Granted' to be present? I would think that findText is unnecessary.

@cahrens thats what I expected, but when I added this in it reduced the test flakiness from 3/100 (failing to find "Granted") to 0/100) 🤔

samanehsan: I just ran this two more times without the additional findText lines and got 0/100 failures, so it looks like it isn't necessary

breilly2: Yeah, anything that is sensitive to timing can come and go. It could be that some external factor made this particularly sensitive at the time you were working on it, and the extra check effectively acted as a micro-sleep which temporarily alleviated the problem. However, as you discovered, that sensitivity can be transient.

kyuksel: I also wonder if a change of 3% is statistically insignificant in this case but don't feel strongly.

kyuksel: Sorry, I realize I'd missed Saman's last message somehow.

cahrens (approved):

PR #2805, opened 2022-02-04
[WOR-57] Use Sam ToS status if possible, otherwise rely on existing ToS cloud function

marctalbott:
Ticket: [WOR-57](https://broadworkbench.atlassian.net/browse/WOR-57)
Testing:
* against a fiab
  * spin up a fiab with ToS feature branches ([sam](https://github.com/broadinstitute/sam/pull/568), [firecloud-develop](https://github.com/broadinstitute/firecloud-develop/pull/2771), [orchestration](https://github.com/broadinstitute/firecloud-orchestration/pull/913))
  * sign in to Terra with users that have not registered at all and users that are registered but haven't accepted the current ToS version. Verified that user was prompted with ToS in both situations and that user was no longer prompted after accepting. 
* against dev
  * I also ran this branch against dev to make sure it would be safe to merge now before the other ToS branches were merged. Verified that I was not prompted to accept ToS as I have already accepted the ToS via existing Cloud Function

<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/AuthContainer.js
+++ b/src/components/AuthContainer.js
@@ -14,7 +14,6 @@ import TermsOfService from 'src/pages/TermsOfService'
 const AuthContainer = ({ children }) => {
   const { name, public: isPublic } = useRoute()
   const { isSignedIn, registrationStatus, acceptedTos, acceptedSamTos, profile } = useStore(authStore)
-  const needToAccept = !(acceptedSamTos === null ? acceptedTos : acceptedSamTos) // acceptedSamTos will be null if Sam ToS enforcement is not yet live. Rely on acceptedTos (from Cloud Function) until Sam ToS is live
   const authspinner = () => h(centeredSpinner, { style: { position: 'fixed' } })
 
   return Utils.cond(
@@ -22,9 +21,9 @@ const AuthContainer = ({ children }) => {
     [isSignedIn === false && !isPublic, () => h(SignIn)],
     [registrationStatus === undefined && !isPublic, authspinner],
     [registrationStatus === 'unregistered', () => h(Register)],
+    [(acceptedTos === undefined || acceptedSamTos === undefined) && !isPublic, authspinner],
+    [!(acceptedSamTos === null ? acceptedTos : acceptedSamTos) && name !== 'privacy', () => h(TermsOfService)],
     [registrationStatus === 'disabled', () => h(Disabled)],
-    [acceptedTos === undefined && !isPublic, authspinner],
-    [needToAccept === true && name !== 'privacy', () => h(TermsOfService)],
     [_.isEmpty(profile) && !isPublic, authspinner],
     () => children
   )
--- a/src/components/AuthContainer.js
+++ b/src/components/AuthContainer.js
@@ -13,7 +13,7 @@ import TermsOfService from 'src/pages/TermsOfService'
 
 const AuthContainer = ({ children }) => {
   const { name, public: isPublic } = useRoute()
-  const { isSignedIn, registrationStatus, acceptedTos, acceptedSamTos, profile } = useStore(authStore)
+  const { isSignedIn, registrationStatus, acceptedTos, profile } = useStore(authStore)
   const authspinner = () => h(centeredSpinner, { style: { position: 'fixed' } })
 
   return Utils.cond(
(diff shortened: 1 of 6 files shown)


src/libs/ajax.js
```diff
@@ -236,6 +236,21 @@ const User = signal => ({
     }
   },
 
+  getSamTosAccepted: async () => {
+    try {
+      const res = await fetchSam('register/user/v1/termsofservice/status', _.merge(authOpts(), { signal }))
+      return await res.json()
```
marctalbott: Is `await` necessary both here and on the line above? It seems redundant to me, but maybe it takes time to get the json?

src/libs/ajax.js
```diff
@@ -248,6 +263,13 @@ const User = signal => ({
     )
   },
 
+  acceptSamTos: async () => {
+    await fetchSam(
+      'register/user/v1/termsofservice',
+      _.mergeAll([authOpts(), { signal, method: 'POST' }, jsonBody('app.terra.bio/#terms-of-service')])
```
marctalbott: Isaac raised this question on a past ToS PR and I don't know the answer and kept forgetting to bring it up so raising it here

> Should the URL come from the current location, since the app is deployed at multiple URLs?

src/libs/auth.js
```diff
@@ -171,7 +172,7 @@ authStore.subscribe(withErrorReporting('Error checking registration', async (sta
       }
     }
   }
-  if (!oldState.isSignedIn && state.isSignedIn) {
+  if ((!oldState.isSignedIn && state.isSignedIn) || (!oldState.acceptedSamTos && state.acceptedSamTos)) {
```
marctalbott: accepting the ToS in Sam may result in the user being re-enabled, so we need to get the user's registration status again to prevent them from seeing the disabled user page

src/libs/ajax.js
MatthewBemis: Sam only knows about this URL, so any other URLs sent in will be rejected. This is what was decided by legal and also it just makes things easier on the dev env side of things.

src/components/AuthContainer.js
```diff
@@ -13,17 +13,17 @@ import TermsOfService from 'src/pages/TermsOfService'
...
-  const { isSignedIn, registrationStatus, acceptedTos, profile } = useStore(authStore)
+  const { isSignedIn, registrationStatus, acceptedTos, acceptedSamTos, profile } = useStore(authStore)
   const authspinner = () => h(centeredSpinner, { style: { position: 'fixed' } })
 
   return Utils.cond(
     [isSignedIn === undefined && !isPublic, authspinner],
     [isSignedIn === false && !isPublic, () => h(SignIn)],
     [registrationStatus === undefined && !isPublic, authspinner],
     [registrationStatus === 'unregistered', () => h(Register)],
+    [(acceptedTos === undefined || acceptedSamTos === undefined) && !isPublic, authspinner],
```
cahrens: Some comments as to what these cases represent would be helpful. As someone who is new to this code, I don't know why `isPublic` matters, and what the name `privacy` signifies.

src/libs/ajax.js
cahrens: It theoretically could take a long time. But it's also just architected to return a promise.

petesantos: I would ditch the `await`. In this case we are essentially unboxing and then immediately re-boxing the promise. Functions that are `async` always return a promise. In the code above the `res.json` is await-ed because it is pulling a specific value out and only returning that.

MatthewBemis (approved):
I didn't test this myself, but the code looks good and the test cases you described sound thorough

src/components/AuthContainer.js
```diff
@@ -13,17 +13,17 @@ import TermsOfService from 'src/pages/TermsOfService'
 
 const AuthContainer = ({ children }) => {
   const { name, public: isPublic } = useRoute()
-  const { isSignedIn, registrationStatus, acceptedTos, profile } = useStore(authStore)
+  const { isSignedIn, registrationStatus, acceptedTos, acceptedSamTos, profile } = useStore(authStore)
```
petesantos: Do we need to add this additional ToS state? Storing and reasoning between these two states makes the ToS flow a bit harder to follow. Also, if we ever need to consume ToS state in other pages and components we will repeatedly need to perform some compute logic to derive the correct state.

There are a get/set Ajax calls. Could we just compute the ToS state at the time of the get call and store the result of the computation as our state?

marctalbott: Perhaps, I'll see what I can do. There was a time where we needed to keep track of which ToS wasn't yet accepted, but I think you're right that we can compute the ToS state and just store that value rather than two separate values.

src/components/AuthContainer.js
marctalbott: To be honest, I don't really know why `isPublic` matters or what `privacy` is either -- I tried to leave those alone as much as possible as I assumed that that part of the conditional was independent of whether or not a user has accepted the ToS. I can try to figure out what those things are if you think some additional comments would be useful

petesantos: To clear things up - `isPublic` or `public` is an attribute we can assign to some pages in the app. Those are pages that you do not need to be signed in to see (e.g. ToS, Landing page, privacy policy). `privacy` is referring to the privacy policy page. 

@cahrens - I agree, either some comments or better naming would help clarify this code

petesantos (approved):
Looks good - thanks for taking care of the state issue!

PR #2806, opened 2022-02-07
[QA-1700] Saving twice results in flakiness related to the Save button enabling after second edit.

cahrens:
This does not address all causes of failure for this test, but it does address the most commonly seen one (Save button not being present after second edit operation). I would like this to be merged while the other issues continue to be investigated.

In 200 runs locally (with the portion of the test after the Save disabled), the "Save" button never caused a failure with this change.

I do periodically see this failure (especially when running locally), so it will need future investigation:

TimeoutError: waiting for XPath `(//input | //textarea)[contains(@aria-label,"Select a workspace") or @id=//label[contains(normalize-space(.),"Select a workspace")]/@for or @aria-labelledby=//*[contains(normalize-space(.),"Select a workspace")]/@id]` failed: timeout 30000ms exceeded

### DIFF
--- a/integration-tests/tests/run-workflow-on-snapshot.js
+++ b/integration-tests/tests/run-workflow-on-snapshot.js
@@ -2,7 +2,7 @@ const _ = require('lodash/fp')
 const fetch = require('node-fetch')
 const { launchWorkflowAndWaitForSuccess } = require('./run-workflow')
 const { withWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, delay, dismissNotifications, fillInReplace, findElement, findText, input, select, signIntoTerra, waitForNoSpinners, navChild } = require('../utils/integration-utils')
+const { click, clickable, dismissNotifications, enabledClickable, fillInReplace, findElement, findText, input, select, signIntoTerra, waitForNoSpinners, navChild } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
@@ -58,9 +58,8 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await fillInReplace(page, input({ labelContains: 'echo_to_file out attribute' }), 'workspace.result')
 
   await click(page, clickable({ text: 'Save' }))
-
-  // TODO: change this to wait for the button to become enabled.
-  await delay(5000) // The Run Analysis button (launchWorkflowAndWaitForSuccess) requires time to become enabled after hitting the save button
+  // Wait for "Run Analysis" to become enabled.
+  await page.waitForXPath(enabledClickable({ text: 'Run analysis' }))
 
   await launchWorkflowAndWaitForSuccess(page)
 
--- a/integration-tests/tests/run-workflow-on-snapshot.js
+++ b/integration-tests/tests/run-workflow-on-snapshot.js
@@ -53,12 +53,10 @@ const testRunWorkflowOnSnapshotFn = _.flow(
 
   await click(page, clickable({ textContains: 'Inputs' }))
   await fillInReplace(page, input({ labelContains: 'echo_to_file input1 attribute' }), `this.${snapshotColumnName}`)
-  await delay(1000) // Without this delay, the input field sometimes reverts back to its default value
-  await click(page, clickable({ text: 'Save' }))
 
   await click(page, clickable({ textContains: 'Outputs' }))
   await fillInReplace(page, input({ labelContains: 'echo_to_file out attribute' }), 'workspace.result')
-  await delay(1000) // Without this delay, the input field sometimes reverts back to its default value
+
   await click(page, clickable({ text: 'Save' }))
 
   await delay(1000) // The Run Analysis button (launchWorkflowAndWaitForSuccess) requires time to become enabled after hitting the save button
--- a/integration-tests/tests/run-workflow-on-snapshot.js
+++ b/integration-tests/tests/run-workflow-on-snapshot.js
@@ -59,7 +59,8 @@ const testRunWorkflowOnSnapshotFn = _.flow(
 
   await click(page, clickable({ text: 'Save' }))
 
-  await delay(1000) // The Run Analysis button (launchWorkflowAndWaitForSuccess) requires time to become enabled after hitting the save button
+  // TODO: change this to wait for the button to become enabled.
+  await delay(5000) // The Run Analysis button (launchWorkflowAndWaitForSuccess) requires time to become enabled after hitting the save button
 
   await launchWorkflowAndWaitForSuccess(page)
 
--- a/integration-tests/tests/run-workflow-on-snapshot.js
+++ b/integration-tests/tests/run-workflow-on-snapshot.js
@@ -32,6 +32,7 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await page.goto(`${testUrl}/#import-data?url=${dataRepoUrlRoot}&snapshotId=${snapshotId}&snapshotName=${snapshotName}&format=snapshot`)
   await signIntoTerra(page, token)
   await dismissNotifications(page)
+  await waitForNoSpinners(page)
   await click(page, clickable({ textContains: 'Start with an existing workspace' }))
   await select(page, 'Select a workspace', workspaceName)
   await click(page, clickable({ text: 'Import' }))
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -50,6 +50,11 @@ const clickable = ({ text, textContains, isDescendant = false }) => {
   return getClickablePath(base, text, textContains, isDescendant)
 }
 
+const enabledClickable = ({ text, textContains, isDescendant = false }) => {
+  const base = `(//a[@aria-disabled="false"] | //button[@aria-disabled="false"] | //*[@role="button" and @aria-disabled="false"] | //*[@role="link" and @aria-disabled="false"] | //*[@role="combobox" and @aria-disabled="false"] | //*[@role="option" and @aria-disabled="false"])`
+  return getClickablePath(base, text, textContains, isDescendant)
+}
+
 const checkbox = ({ text, textContains, isDescendant = false }) => {
   const base = `(//input[@type="checkbox"] | //*[@role="checkbox"])`
   return getClickablePath(base, text, textContains, isDescendant)
@@ -265,6 +270,7 @@ module.exports = {
   click,
   clickable,
   clickTableCell,
+  enabledClickable,
   dismissNotifications,
   findIframe,
   findInGrid,


cahrens:
I'm hitting an issue immediately after my fix (I had commented out that code while running locally), so I'm making this a draft while I try to work it out.

integration-tests/utils/integration-utils.js
```diff
@@ -50,6 +50,11 @@ const clickable = ({ text, textContains, isDescendant = false }) => {
   return getClickablePath(base, text, textContains, isDescendant)
 }
 
+const enabledClickable = ({ text, textContains, isDescendant = false }) => {
+  const base = `(//a[@aria-disabled="false"] | //button[@aria-disabled="false"] | //*[@role="button" and @aria-disabled="false"] | //*[@role="link" and @aria-disabled="false"] | //*[@role="combobox" and @aria-disabled="false"] | //*[@role="option" and @aria-disabled="false"])`
```
cahrens, line 54: I don't think that a native button will have "aria-disabled" on it. However, we never create a native button. I did see that we set "aria-disabled" on anchor tags.

I'm not sure what the correct xpath would be for a disabled native button, but I figure that it's not worth figuring out at this time. If we do start using native buttons, and if someone tries to call this method on one, this method will always fail and they can figure out the correct xpath at that time.

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -53,15 +53,13 @@ const testRunWorkflowOnSnapshotFn = _.flow(
...
-  await delay(1000) // Without this delay, the input field sometimes reverts back to its default value
-  await click(page, clickable({ text: 'Save' }))
 
   await click(page, clickable({ textContains: 'Outputs' }))
   await fillInReplace(page, input({ labelContains: 'echo_to_file out attribute' }), 'workspace.result')
-  await delay(1000) // Without this delay, the input field sometimes reverts back to its default value
-  await click(page, clickable({ text: 'Save' }))
 
-  await delay(1000) // The Run Analysis button (launchWorkflowAndWaitForSuccess) requires time to become enabled after hitting the save button
+  await click(page, clickable({ text: 'Save' }))
```
cahrens, line 61: When running 100s of times locally, it was clear that the test only was failing on the second Save attempt.

Save will persist both the inputs and outputs, so there was no reason to press Save twice. There may be a race condition between the page update after the Save response is received and editing the output field. I doubt that a user would ever encounter it though because they would not be able to interact with the elements that quickly. I was not able to reproduce it in my manual tests.

kyuksel: Feels great to see those delays removed  🤩

petesantos (commented):
Running this 100 times, I saw 7 failures with 3 signatures. I think only one of them is related to this test, but I am putting all of them here so you can take a look.

## One (2 instances) - looks like possible issue with test
```
TimeoutError: waiting for XPath `(//input | //textarea)[contains(@aria-label,"Select a workspace") or @id=//label[contains(normalize-space(.),"Select a workspace")]/@for or @aria-labelledby=//*[contains(normalize-space(.),"Select a workspace")]/@id]` failed: timeout 30000ms exceeded
    at new WaitTask (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/DOMWorld.ts:798:28)
    at DOMWorld.waitForXPath (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/DOMWorld.ts:695:22)
    at Frame.waitForXPath (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/FrameManager.ts:1304:47)
    at Page.waitForXPath (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/Page.ts:3269:29)
    at click (/Users/psantos/projects/terra-ui/integration-tests/utils/integration-utils.js:83:22)
    at select (/Users/psantos/projects/terra-ui/integration-tests/utils/integration-utils.js:123:9)
    at /Users/psantos/projects/terra-ui/integration-tests/tests/run-workflow-on-snapshot.js:36:9
    at /Users/psantos/projects/terra-ui/integration-tests/utils/integration-helpers.js:51:5
```
### Screenshot



## Two (1 instance) - bucket access - this may be an actual issue
This seems to be more of a sporadic systemic issue 
```
TimeoutError: Navigation timeout of 30000 ms exceeded
    at /Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/LifecycleWatcher.ts:205:18
    at FrameManager.waitForFrameNavigation (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/FrameManager.ts:239:19)
    at Frame.waitForNavigation (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/FrameManager.ts:786:12)
    at Page.waitForNavigation (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/Page.ts:1833:12)
    at async Promise.all (index 0)
    at launchWorkflowAndWaitForSuccess (/Users/psantos/projects/terra-ui/integration-tests/tests/run-workflow.js:55:3)
    at /Users/psantos/projects/terra-ui/integration-tests/tests/run-workflow-on-snapshot.js:64:3
    at /Users/psantos/projects/terra-ui/integration-tests/utils/integration-helpers.js:51:5
    at /Users/psantos/projects/terra-ui/integration-tests/jest/jest-utils.js:75:24
```
### Screenshot

eenshot


## Three (1 instance) - Looks like the issue where a workspace was not created

```
Error: Evaluation failed: Response
    at ExecutionContext._evaluateInternal (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/ExecutionContext.ts:273:13)
    at ExecutionContext.evaluate (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/ExecutionContext.ts:140:12)
    at /Users/psantos/projects/terra-ui/integration-tests/utils/integration-helpers.js:29:3
    at /Users/psantos/projects/terra-ui/integration-tests/utils/integration-helpers.js:17:12
    at /Users/psantos/projects/terra-ui/integration-tests/utils/integration-helpers.js:48:25
    at /Users/psantos/projects/terra-ui/integration-tests/jest/jest-utils.js:75:24
```
### Screenshot (Blank)


## Four (3 instances) Login error - probably not related to the test
I haven't seen this before, but I don't think it is related to this test specifically
```
Error: Evaluation failed: Response
    at ExecutionContext._evaluateInternal (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/ExecutionContext.ts:273:13)
    at ExecutionContext.evaluate (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/ExecutionContext.ts:140:12)
    at /Users/psantos/projects/terra-ui/integration-tests/utils/integration-helpers.js:40:3
    at /Users/psantos/projects/terra-ui/integration-tests/utils/integration-helpers.js:17:12
    at /Users/psantos/projects/terra-ui/integration-tests/utils/integration-helpers.js:53:5
    at /Users/psantos/projects/terra-ui/integration-tests/jest/jest-utils.js:75:24
```
### Screenshot

cahrens:
Thanks @petesantos for reporting the failures you saw. I definitely agree there is more work to be done on this test, but I wanted to get this change in first because I think it will drastically reduce the number of failures folks are seeing. As far as the specific failures you mentioned:

1) The "Select a workspace" issue is definitely one I saw, and I mentioned it in the PR description. I plan to look into that one next.
2) The "bucket access" issue is something that I added a workaround for to the `run-workflow` test (I have it wait for the bucket permissions to be granted before proceeding with the test). I didn't add it to this test at the same time because this test works on an existing workspace, and I wasn't sure how the bucket permissions worked in this case (if we would expect the user to already have access to this bucket). It appears though that we do want to explore that workaround for this test as well, and I can do that. The issue for `run-workflow` at least was that it takes some time for the bucket permission to be granted, and if you try to run a workflow during that time you see this expected message. We expect users to refresh/wait and try again. This is definitely not a great user experience, and is one we see in the Analyses tab as well where the user can get a bunch of popup errors while permissions are sorting themselves out.
3) The workspace not getting created is a sporadic problem we have seen across many different tests. I can bring it up with my team. I know there have been a lot of issues with workspace creation, but I am not yet up-to-speed on the various causes.
4) Login error. This isn't one I've been before, but I can keep an eye out for it while I'm working on 1 and 2 above.

petesantos (commented):
To address test issue #1 I highlighted in my earlier comment, I think adding a `waitForNoSpinners` before line 35 may solve the issue.
I suspect the button appears, but there is an spinner overlay, so the click may not be working.

petesantos (approved):
LGTM! Thanks
As discussed in PR, other issues will be addressed in follow ups.

cahrens:
> To address test issue #1 I highlighted in my earlier comment, I think adding a `waitForNoSpinners` before line 35 may solve the issue.
> I suspect the button appears, but there is an spinner overlay, so the click may not be working.

Thanks for the suggestion! I didn't realize that "click" would not error if a spinner was present. I made that change and have run that portion of the test 180 times with none of those failures. So I might as well just add it to this PR.

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -32,6 +32,7 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await page.goto(`${testUrl}/#import-data?url=${dataRepoUrlRoot}&snapshotId=${snapshotId}&snapshotName=${snapshotName}&format=snapshot`)
   await signIntoTerra(page, token)
   await dismissNotifications(page)
+  await waitForNoSpinners(page)
```
cahrens, line 35: As suggested by @petesantos, verified with 180+ local runs.

integration-tests/utils/integration-utils.js
breilly2: I think we could roll this into `clickable`. I can't imagine any test case that, when asking for something clickable, would be okay with it being disabled.

cahrens: Possibly, though I definitely would like to do that in a follow-up PR because I really want to merge this one ASAP. Personally, the way I think of this is "I want to wait for this button to become enabled, and then I want to click on it". I would expect the "wait for the button to become enabled" to be an opt-in because the test-writer should know that in the scenario being tested, it is expected that the button will be disabled initially. This is not normally the case. So I'm still trying to work through in my head if it is OK to make this the default behavior of `clickable`.

integration-tests/tests/run-workflow-on-snapshot.js
ungwudik: Good find! 
I wonder if something changed in the product leading to the dramatical increase of the failure rate, and if there are implications for other tests. But that's obvs out of scope for this PR.

ungwudik (approved):
lgtm

integration-tests/utils/integration-utils.js
breilly2: I understand the urgency to merge this PR and I'm okay with that.

After that, I'm very confident that we can and should roll this into `clickable`. The original intent was for the tests to read simply and clearly, so the extra cognitive overhead of having to be aware of `clickable` and `enabledClickable` feels cumbersome. Looking through the code, of all the calls to `clickable` there are 6 cases that don't match `click (page, clickable(...))`. 3 of those are just inserting a `waitForNoSpinners` between finding and clicking. 2 (consecutive) ares looking for status updates to the cloud environment widget and are presumably using `clickable` to avoid finding the words "Creating" or "Running" elsewhere on the page. The last one is coming from FireCloud and needs to hack a navigation back to the Terra UI under test (localhost, PR test deploy, etc) instead of necessarily going to the shared Terra UI (dev, alpha, etc).

cahrens: @breilly2 I'm coming around to it… I will see about rolling it into my next "QA-1700 PR" that is forthcoming.

petesantos: Personally, I am ok with merging this as-is to help alleviate immediate developer pain and improve the testing signal. But taking another look - I believe this xpath will do the equivalent:

```suggestion
const base = '(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])[@aria-disabled="false"]'
```

My first reaction was the same as @breilly2. But I was thinking about the use cases that @cahrens pointed out and didn't comment. Would re-naming help, along with an explicit wait? Perhaps something like: 

```javascript
await clickableEnabled(page, {text: 'foo'})
await click(page, clickable({text: 'foo'})
```

When combined with the xpath change, this could just be an option to `clickable` - perhaps `isDisabled`

I do like the explicit nature of checking the clickable as this is something a user would do and it makes the test read nicely. I can follow along in the app this way and see that the test will wait for the appropriate state before proceeding - it looks very deliberate to me even if a little verbose. I could see a case where we may want to ensure something is not clickable until the inputs are verified.

cahrens: I personally like having "clickableEnabled" as a separate, explicit call. I took a crack at what I think @petesantos is suggesting in https://github.com/DataBiosphere/terra-ui/pull/2811/commits/948a201efcbe4be09b940239255bae2db8e74d7f. @breilly2 what do you think?

cahrens: FWIW, I really don't think that one would ever have "aria-disabled" on a native button because buttons have disabled as an attribute built-in. So including "aria-disabled" by default on an xpath that is supposed to handle a button feels wrong to me, though in reality I don't think we ever use native buttons so it is probably a non-issue (😞 ).

PR #2807, opened 2022-02-07
[DC-207][QA-1706] Fix link-to-new-workspace test

samanehsan:
This test has failed a few times due to a 403 during workspace creation:
```
page.http.res 403 GET https://storage.googleapis.com/storage/v1/b/fc-9e71db22-ed13-4bdc-82fe-81df090de356?fields=location%2ClocationType
 FAIL  jest/link-to-new-workspace.integration-test.js (44.787 s)
  ● link-to-new-workspace
```

I was not able to reproduce the failure locally, but waiting for the bucket access to propagate might fix the issue (this is what the [run-workflow](https://github.com/DataBiosphere/terra-ui/blob/dev/integration-tests/tests/run-workflow.js#L21) test does)

### DIFF
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -1,5 +1,5 @@
 const { checkbox, click, clickable, clickTableCell, fillIn, waitForNoSpinners } = require('../utils/integration-utils')
-const { enableDataCatalog, testWorkspaceName } = require('../utils/integration-helpers')
+const { checkBucketAccess, enableDataCatalog, testWorkspaceName } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
@@ -21,6 +21,8 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) 
     await click(page, clickable({ text: `${newWorkspaceBillingAccount}` }))
     await click(page, clickable({ text: 'Create Workspace' }))
     await waitForNoSpinners(page)
+    // Wait for bucket access to avoid sporadic failures
+    await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
     await page.url().includes(newWorkspaceName)
   } finally {
     await page.evaluate((name, billingProject) => {
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -14,19 +14,23 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) 
 
   const newWorkspaceName = testWorkspaceName()
   const newWorkspaceBillingAccount = 'general-dev-billing-account'
+  await click(page, clickable({ textContains: 'Start with a new workspace' }))
+  await fillIn(page, '//*[@placeholder="Enter a name"]', `${newWorkspaceName}`)
+  await click(page, clickable({ text: 'Select a billing project' }))
+  await click(page, clickable({ text: `${newWorkspaceBillingAccount}` }))
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Link to a workspace' })) })
   try {
-    await click(page, clickable({ textContains: 'Start with a new workspace' }))
-    await fillIn(page, '//*[@placeholder="Enter a name"]', `${newWorkspaceName}`)
-    await click(page, clickable({ text: 'Select a billing project' }))
-    await click(page, clickable({ text: `${newWorkspaceBillingAccount}` }))
-    await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Link to a workspace' })) })
     // Wait for bucket access to avoid sporadic failures
     await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
     await page.url().includes(newWorkspaceName)
   } finally {
-    await page.evaluate((name, billingProject) => {
-      return window.Ajax().Workspaces.workspace(billingProject, name).delete()
-    }, `${newWorkspaceName}`, `${newWorkspaceBillingAccount}`)
+    try {
+      await page.evaluate((name, billingProject) => {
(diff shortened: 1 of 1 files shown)


pshapiro4broad (approved):

fboulnois (approved):

integration-tests/tests/link-to-new-workspace.js
```diff
@@ -21,6 +21,8 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token })
     await click(page, clickable({ text: `${newWorkspaceBillingAccount}` }))
     await click(page, clickable({ text: 'Create Workspace' }))
     await waitForNoSpinners(page)
```
petesantos: When running this test through the flake checker I was seeing I am seeing some (4/200) of these errors as well:

```
TimeoutError: waiting for XPath `//*[@data-icon="loadingSpinner"]` to be hidden failed: timeout 30000ms exceeded
    at new WaitTask (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/DOMWorld.ts:798:28)
    at DOMWorld.waitForXPath (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/DOMWorld.ts:695:22)
    at Frame.waitForXPath (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/FrameManager.ts:1304:47)
    at Page.waitForXPath (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/Page.ts:3269:29)
    at waitForNoSpinners (/Users/psantos/projects/terra-ui/integration-tests/utils/integration-utils.js:123:15)
    at /Users/psantos/projects/terra-ui/integration-tests/tests/link-to-new-workspace.js:23:11
    at /Users/psantos/projects/terra-ui/integration-tests/jest/jest-utils.js:75:24
```

I ran it with this code and I did not see the issue in 200 runs (this is a new util I added last week to address flaky spinner failures):
```suggestion
  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Link to a workspace' })) })
```

integration-tests/tests/link-to-new-workspace.js
```diff
@@ -1,5 +1,5 @@
 const { checkbox, click, clickable, clickTableCell, fillIn, waitForNoSpinners } = require('../utils/integration-utils')
-const { enableDataCatalog, testWorkspaceName } = require('../utils/integration-helpers')
+const { checkBucketAccess, enableDataCatalog, testWorkspaceName } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
```
petesantos, line 5: Could you replace lines 12-13 with the following?

```javascript
await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Link to a workspace' })) })
```

I cannot make the direct comment/suggestion on those lines in gh since they were not changed in the diff.

petesantos (commented):
I have two suggestions, but after implementing these it resulted in a different error:
```
Error: Evaluation failed: Response
    at ExecutionContext._evaluateInternal (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/ExecutionContext.ts:273:13)
    at ExecutionContext.evaluate (/Users/psantos/projects/terra-ui/integration-tests/node_modules/puppeteer/src/common/ExecutionContext.ts:140:12)
    at /Users/psantos/projects/terra-ui/integration-tests/tests/link-to-new-workspace.js:27:5
    at /Users/psantos/projects/terra-ui/integration-tests/jest/jest-utils.js:75:24
```

With the screenshot functionality in place in the flake tool I am seeing:


What I think is happening is the following:
1. We are trying to make a new workspace
2. This times out, throwing an error (the code should not get past `noSpinnersAfter`)
3. The `finally` block is executed before the error in the try block statement is thrown
4. An error occurs in the `finally` block because there is no workspace to delete
5. The error in `finally` is what gets thrown, reporting line 27 as the problem 
    1. I believe this is a red herring, and the actual error is occurring before deletion, at creation time

You can observe the behavior or `try` / `catch` with this code (you can run it in a chrome browser snippet):
```
function main () {
    try {
        console.log('Thing')
        throw new Error('Thing error')
    } finally {
        console.log('Finally')
        throw new Error('Finally error')
    }
}

main()
```

I think we need to make sure the errors in the try block are appropriately propagated so we are getting an accurate error signal from the tests. Could you take a look at this? If you need help, let me know.

integration-tests/tests/link-to-new-workspace.js
```diff
@@ -19,8 +19,9 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token })
     await fillIn(page, '//*[@placeholder="Enter a name"]', `${newWorkspaceName}`)
     await click(page, clickable({ text: 'Select a billing project' }))
     await click(page, clickable({ text: `${newWorkspaceBillingAccount}` }))
-    await click(page, clickable({ text: 'Create Workspace' }))
-    await waitForNoSpinners(page)
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Link to a workspace' })) })
+    // Wait for bucket access to avoid sporadic failures
+    await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
```
petesantos, line 23: Could you move the `try` to this line - right above `checkBucketAccess`?
If I understand correctly, we want to clean up/delete the workspace, but we should only need to do that if the workspace was created correctly (line 22). After that, no matter what happens we want to clean up. 

My concern here is I want to get the correct error signal reported if the test fails, but the finally is intercepting the creation error right now and reporting all errors in the `try` block as `evaluate` errors since the deletion will fail if any part of the test fails at or before line 22.

samanehsan: Oh I see what you mean! Thanks for clarifying

samanehsan: I added a try/catch around deleting the workspace so the error in the try block will get thrown now

samanehsan:
@petesantos lmk if the changes look good to you!

petesantos (approved):

PR #2809, opened 2022-02-07
Fix for issue with id not matching the actual variable name (BW-1078).

cahrens:
In [this code review commit](https://github.com/DataBiosphere/terra-ui/pull/2794/commits/9e63392c1b41f11f7abd2c5b12659dc9b337073f) I collapsed some code that was explicitly listing every possible status type with indexing into statusTypes. Somewhat unfortunately, this means that the `id` value of statusType must now match its variable name. There was only one statusType where this was not the case, and of course that was the one that I didn't test after I made the code change….

I really don't love the fact that the `id` has to match the variable name, but I did love the code that I deleted in the cleanup commit. So I am inclined to fix this bug by changing the one variable name (and adding comments to hopefully avoid introducing this bug in the future). However, I could be talked into restoring the previous code (as long as I can do it in time to get in tomorrow's terra-ui release, when this bug becomes customer-visible).

Screenshot showing the problematic status:
image

### DIFF
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -160,7 +160,7 @@ const SubmissionDetails = _.flow(
       div({ style: { display: 'grid', gridTemplateColumns: '1fr 4fr' } }, [
         div({ style: { display: 'grid', gridTemplateRows: '1fr auto' } }, [
           makeSection('Workflow Statuses',
-            [statusType.succeeded.id, statusType.failed.id, statusType.running.id, statusType.submitted.id].filter(s => statusGroups[s]).map(
+            ['succeeded', 'failed', 'running', 'submitted'].filter(s => statusGroups[s]).map(
               s => makeStatusLine(statusType[s].icon, addCountSuffix(statusType[s].label(), statusGroups[s].length), { marginTop: '0.5rem' })
             )),
           div({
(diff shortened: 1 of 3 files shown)


src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -160,7 +160,7 @@ const SubmissionDetails = _.flow(
       div({ style: { display: 'grid', gridTemplateColumns: '1fr 4fr' } }, [
         div({ style: { display: 'grid', gridTemplateRows: '1fr auto' } }, [
           makeSection('Workflow Statuses',
-            [statusType.succeeded.id, statusType.failed.id, statusType.running.id, statusType.submitted.id].filter(s => statusGroups[s]).map(
+            ['succeeded', 'failed', 'running', 'submitted'].filter(s => statusGroups[s]).map(
```
cahrens, line 163: This is not actually necessary to fix the bug, but I think it makes it more obvious what is going on. We are accessing  statusGroups.succeeded by doing statusGroups['succeeded']. @dmohs actually advocated for me to put the explicit strings like this, but I was trying to avoid potential typos….

cahrens:
Integration test failure, `run-workflow-on-snapshot`, is a known (unrelated) flaky test.

aednichols (approved):

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
petesantos: I think I would prefer something like an enumerated type here as well. But I am ok with it as-is for this PR

petesantos (approved):

PR #2810, opened 2022-02-08
[IA-3209] Enable Rmd file syncing

gpcarr:
This PR enables .Rmd file syncing. 

The modal that was added `FileOutdatedModal` is meant to handle when a user runs into a Generation Mismatch error. A generation mismatch error happens when two users who have access to the same workspace bucket are editing the same file. User 1 creates a fileA.Rmd, it gets delocalized to the workspace bucket, and then localized from bucket to user 2's VM. user 2 then edits that same file and user 2's changes get delocalized to the bucket. User 2's changes do not get localized back to user 1's VM because the file already exists on user 1's VM. So user 1 now has an outdated version of fileA.Rmd. User 1 will have 2 options: 1) make a copy of their outdated file or 2) do nothing and keep their outdated version.

Screen Shot 2022-04-08 at 4 47 35 PM

Screen Shot 2022-04-08 at 5 01 43 PM

This PR goes along with this [Welder PR](https://github.com/DataBiosphere/welder/pull/76) that handles the generation mismatch in the backend. And this [Leo PR ](https://github.com/DataBiosphere/leonardo/pull/2530)that enables the Welder background process.

~~This PR was tested using two users in a FIAB.~~
This PR is now also tested in dev. In order to get the backend changes that go with this PR, change `const [shouldSetupWelder, setShouldSetupWelder] = useState(false)` to `const [shouldSetupWelder, setShouldSetupWelder] = useState(application == tools.RStudio.label)` in line 60 of `AppLauncher` which will post the storageLinks to welder and enable the backend
Use two users: draco.malfoy@test.firecloud.org and b.admin.firec@gmail.com. I created a workspace as Bartholomew and shared that workspace with Draco. I created an RStudio runtime as Bartholomew and some Rmd files. I waited for those files to delocalize to the workspace bucket. Then I created an Rstudio runtime as Draco in that same shared workspace and waited for those same files to localize to Draco's VM. I then edited those files as Draco and waited for those edits to delocalize to the bucket. I then went back to Bartholomew (who would not get Draco's changes since those files already exist on Bartholomew's VM) and made edits to the same files. I saved those edits and got the generation mismatch error and validated that the modal comes up and handles the error appropriately. I then validated that Draco can still make edits to those files and they still get delocalized.


kyuksel:
@gpcarr Let's keep the PR in draft state until it's ready for review.

Qi77Qi (approved):
@gpcarr demoed the work during last IA sprint review...See https://drive.google.com/file/d/1HIQiuipQt3M2b_CXbNENEbJa47_pS0DZ/view?usp=sharing (probably started around 15 min in), and things looked good

cstrandw (approved):
Tested by following the instructions laid out in the PR description. I verified that the modal pops up at the expected time, and that the original and copy files behave as expected. I did not see any error messages, and verified the time stamps in the bucket for when the file should have been changed as I went along.

src/libs/ajax.js
```diff
@@ -1249,6 +1249,22 @@ const Buckets = signal => ({
       )
     }
 
+    const copyWithMetadata = async (newName, newBucket, copyMetadata) => {
+      const body = { metadata: copyMetadata }
+      return fetchBuckets(
+        `${bucketUrl}/${encodeFileName(name)}/copyTo/b/${newBucket}/o/${encodeFileName(newName)}`,
+        _.mergeAll([authOpts(await saToken(googleProject)), jsonBody(body), { signal, method: 'POST' }])
+      )
+    }
```
kyuksel: How would you feel about factoring out the common piece between `copy` and `copyWithMetadata` with something like:

```javascript
const doCopy = async (newName, newBucket, body) => fetchBuckets(
  `${bucketUrl}/${encodeFileName(name)}/copyTo/b/${newBucket}/o/${encodeFileName(newName)}`,
   _.mergeAll([authOpts(await saToken(googleProject)), jsonBody(body), { signal, method: 'POST' }])  
)

const copy = async (newName, newBucket, clearMetadata) => {
  const body = clearMetadata ? { metadata: { lastLockedBy: '' } } : {}
  doCopy(newName, newBucket, body)
}

const copyWithMetadata = async (newName, newBucket, clearMetadata) => {
  const body = { metadata: copyMetadata }
  doCopy(newName, newBucket, body)
}
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -41,9 +42,15 @@ const ApplicationLauncher = _.flow(
...
   const cookieReady = useStore(cookieReadyStore)
+  const signal = useCancellation()
+  const interval = useRef()
+  const { user: { email } } = useStore(authStore)
   const [showCreate, setShowCreate] = useState(false)
   const [busy, setBusy] = useState(false)
   const [location, setLocation] = useState(defaultLocation)
+  const [outdatedAnalyses, setOutdatedAnalyses] = useState()
+  const [fileOutdatedOpen, setFileOutdatedOpen] = useState(false)
+  const [hashedOwnerEmail, setHashedOwnerEmail] = useState()
```
kyuksel: We typically list the `state`s on top for easier parsing since they tend to be the first part to look at when reading a component code:

```suggestion
  // State
  const [showCreate, setShowCreate] = useState(false)
  const [busy, setBusy] = useState(false)
  const [location, setLocation] = useState(defaultLocation)
  const [outdatedAnalyses, setOutdatedAnalyses] = useState()
  const [fileOutdatedOpen, setFileOutdatedOpen] = useState(false)
  const [hashedOwnerEmail, setHashedOwnerEmail] = useState()
  
  const cookieReady = useStore(cookieReadyStore)
  const signal = useCancellation()
  const interval = useRef()
  const { user: { email } } = useStore(authStore)
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -90,6 +176,24 @@ const ApplicationLauncher = _.flow(
       setupWelder()
       setShouldSetupWelder(true)
     }
+
+    const findOutdatedAnalyses = withErrorReporting('Error loading outdated analyses', async () => {
+      const outdatedRAnalyses = await checkForOutdatedAnalyses({ googleProject, bucketName })
+      setOutdatedAnalyses(outdatedRAnalyses)
+      if (outdatedRAnalyses?.length >= 1) {
+        setFileOutdatedOpen(true)
+      }
```
kyuksel:
```suggestion
      _.isEmpty(outdatedRAnalyses) && setFileOutdatedOpen(true)
```
We typically use the `<predicate> && doSomething()` pattern for simple conditionals like this, for concision.

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +62,85 @@ const ApplicationLauncher = _.flow(
...
+      const end = analyses?.length
+      for (const analysis of analyses) {
+        res += analysis?.name.split('/')[1]
+        if (i < end) {
+          res += ', '
+        }
+        i++
+      }
+      return res
+    }
```
kyuksel: Similarly to how we try not to use mutables like `var` and imperative constructs like loops in our Scala codebases, let's try to avoid that here as well. For instance, using Lodash/FP (for reason explained in a comment below) something like below could do?

```javascript
const getDisplayList = _.flow(
  _.map(
    _.flow(
      _.get('name'),
      _.split('/'),
      _.nth(1)
    )
  ),
  _.without([undefined]),
  _.join(', ')
)
```

I don't know the exact content of an `analysis` but assuming some object that contains a `name` property:

image

I'd also suggest writing a unit test since this function is complicated enough. A unit test in this case would help with documentation as well 😉 

We prefer Lodash (FP) functions over native functions for the following reasons:
- Lodash/fp functions support currying. This is especially useful for code changes later. If someone wants to add functionality, it makes it easy to do something like add the function to a flow.
- Lodash/fp will not mutate the array. In the case of native slice, it also returns a new array. But there are other functions that can mutate (e.g. push/pop)
- Mixing native and lodash functions can lead to confusion about behavior.

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +62,85 @@ const ApplicationLauncher = _.flow(
   const runtime = getCurrentRuntime(runtimes)
   const runtimeStatus = getConvertedRuntimeStatus(runtime) // preserve null vs undefined
 
+  const FileOutdatedModal = ({ onDismiss, bucketName }) => {
+    const handleChoice = _.flow(
+      withErrorReportingInModal('Error setting up analysis file syncing')(onDismiss),
+      Utils.withBusyState(setBusy)
+    )(async shouldCopy => {
+      await Promise.all(_.flatMap(async analysis => {
```
kyuksel: As far as I know `Promise.all` is used to kick off multiple Ajax requests in parallel. It looks like that's not the case here so we don't need it?

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +62,85 @@ const ApplicationLauncher = _.flow(
...
+        const file = analysis.name.split('/')[1]
+        if (shouldCopy) {
+          currentMetadata[hashedOwnerEmail] = ''
+          await Ajax().Buckets.analysis(googleProject, bucketName, stripExtension(file), tools.RStudio.label).copyWithMetadata(stripExtension(file).concat(`_copy${Date.now().toString()}.`).concat(tools.RStudio.ext), bucketName, currentMetadata)
+        }
+        currentMetadata[hashedOwnerEmail] = 'doNotSync'
+        await Ajax().Buckets.analysis(googleProject, bucketName, stripExtension(file), tools.RStudio.label).updateMetadata(file, currentMetadata)
+      }, outdatedAnalyses))
+      onDismiss()
+    })
```
kyuksel, line 87: I'm having a little difficult time parsing the body of `handleChoice` (the part starting with line 69). Can we break it down a bit? For instance, `stripExtension(file).concat(`_copy${Date.now().toString()}.`).concat(tools.RStudio.ext)` could be pulled out into its own function (using Lodash for reasons explained in the comment below).

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +62,85 @@ const ApplicationLauncher = _.flow(
...
+  const FileOutdatedModal = ({ onDismiss, bucketName }) => {
+    const handleChoice = _.flow(
+      withErrorReportingInModal('Error setting up analysis file syncing')(onDismiss),
+      Utils.withBusyState(setBusy)
+    )(async shouldCopy => {
+      await Promise.all(_.flatMap(async analysis => {
+        const currentMetadata = analysis.metadata
+        const file = analysis.name.split('/')[1]
+        if (shouldCopy) {
+          currentMetadata[hashedOwnerEmail] = ''
```
kyuksel: Can we use separate `const`s instead of mutating `currentMetadata`?

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +62,85 @@ const ApplicationLauncher = _.flow(
...
+        }
+        i++
+      }
+      return res
+    }
+
+    return h(Modal, {
+      onDismiss,
+      width: 530,
+      title: outdatedAnalyses?.length > 1 ? 'R Markdown Files In Use' : `R Markdown File Is In Use`,
```
kyuksel: Can we use Lodash like so 

```suggestion
      title: _.isEmpty(outdatedAnalyses) ? 'R Markdown Files In Use' : `R Markdown File Is In Use`,
```

for reasons explained in the comment above?

Similarly for other occurences of `outdatedAnalyses?.length > 1`.

kyuksel: Have we checked with UX as far as the casing of the text; i.e., are we sure it shouldn't be `R markdown files in use`?

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +62,85 @@ const ApplicationLauncher = _.flow(
...
+      showButtons: false
+    }, [
+      Utils.cond(
+        [outdatedAnalyses?.length > 1, () => [p(`These R markdown files are being edited by another user and your versions are now outdated. Your files will no longer sync with the workspace bucket.`),
+          p(getDisplayList(outdatedAnalyses)),
+          p('You can'),
+          p(['1) ', b('save your changes as new copies'), ' of your files which will enable file syncing on the copies']),
```
kyuksel: We use `strong` for emphasizing text semantically:

```suggestion
          p(['1) ', strong(['save your changes as new copies']), ' of your files which will enable file syncing on the copies']),
```

Similarly below.

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +62,85 @@ const ApplicationLauncher = _.flow(
...
+      ])
+    ])
+  }
+
+  const checkForOutdatedAnalyses = async ({ googleProject, bucketName }) => {
+    const analyses = await Ajax(signal).Buckets.listAnalyses(googleProject, bucketName)
+    const rAnalyses = _.filter(({ name }) => name.endsWith(`.${tools.RStudio.ext}`), analyses)
+    return _.filter(
+      analysis => analysis && Object.keys(analysis).includes('metadata') && Object.keys(analysis.metadata).includes(hashedOwnerEmail) &&
+        analysis.metadata[hashedOwnerEmail] === 'outdated', rAnalyses)
```
kyuksel: Would the following be simpler?
```suggestion
    return _.filter(analysis =>
      _.endsWith(`.${tools.RStudio.ext}`, analysis?.name) &&
      analysis?.metadata?.hashedOwnerEmail === 'outdated', analyses)
```

src/libs/ajax.js
gpcarr: I like this approach, will update!

src/pages/workspaces/workspace/applications/AppLauncher.js
gpcarr: ah yes i remember seeing this pattern, will update

src/pages/workspaces/workspace/applications/AppLauncher.js
gpcarr: very good points! i'll update and add unit tests

src/pages/workspaces/workspace/applications/AppLauncher.js
gpcarr: It would be the case if `outdatedAnalyses` (which is kept track of in state) has more than one analysis in it

src/pages/workspaces/workspace/applications/AppLauncher.js
gpcarr: I can pull this out into its own function and add some helpful comments

src/pages/workspaces/workspace/applications/AppLauncher.js
gpcarr: so this check is actually checking if outdatedAnalyses has 1 or more than 1 entry in it so `_.isEmpty` would not work. maybe let's sync up when we pair on a better implementation.

And yeah, UX has approved this modal. We have a similar modal for notebooks whose title is `Notebook Is In Use`

src/pages/workspaces/workspace/applications/AppLauncher.js
gpcarr: we resolved this offline - marking as resolved

src/pages/workspaces/workspace/applications/AppLauncher.js
kyuksel: oh I see  👍🏼

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +63,95 @@ const ApplicationLauncher = _.flow(
...
   const runtimeStatus = getConvertedRuntimeStatus(runtime) // preserve null vs undefined
 
+  const FileOutdatedModal = ({ onDismiss, bucketName }) => {
+    const handleChoice = _.flow(
+      withErrorReportingInModal('Error setting up analysis file syncing')(onDismiss),
+      Utils.withBusyState(setBusy)
+    )(async shouldCopy => {
+      await Promise.all(_.flatMap(async analysis => {
+        const currentMetadata = analysis.metadata
+        const file = getFileName(analysis.name)
```
kyuksel: FYI that you can write the above as

```suggestion
      await Promise.all(_.flatMap(async ({ name, metadata: currentMetadata }) => {
        const file = getFileName(name)
```
which helps with concision. Not a blocker though.

Note that `{ name, metadata: currentMetadata }` is short-hand for `{ name: name, metadata: currentMetadata }` which can only be used when the `key` and `value` are the same.

kyuksel (commented):
Will come back later this afternoon to finish the re-review.

src/pages/workspaces/workspace/applications/AppLauncher.js
gpcarr: i like this pattern more, updated my code!

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +63,89 @@ const ApplicationLauncher = _.flow(
...
+    const handleChoice = _.flow(
+      withErrorReportingInModal('Error setting up analysis file syncing')(onDismiss),
+      Utils.withBusyState(setBusy)
+    )(async shouldCopy => {
+      // this modal only opens when the state variable outdatedAnalyses is non empty (keeps track of a user's outdated RStudio files). it gives users two options when their files are in use by another user
+      // 1) make copies of those files and continue working on the copies or 2) do nothing.
+      // in either case, their original version of the analysis is outdated and we will no longer sync that file to the workspace bucket for the current user
+      await Promise.all(_.flatMap(async ({ name, metadata: currentMetadata }) => {
+        const file = getFileName(name)
+        const newMetadata = currentMetadata
```
kyuksel, line 76: Looks like we can eliminate one more `const`?

```suggestion
      await Promise.all(_.flatMap(async ({ name, metadata: newMetadata }) => {
        const file = getFileName(name)
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +63,89 @@ const ApplicationLauncher = _.flow(
...
+      onDismiss,
+      width: 530,
+      title: _.size(outdatedAnalyses) > 1 ? 'R Markdown Files In Use' : `R Markdown File Is In Use`,
+      showButtons: false
+    }, [
+      Utils.cond(
+        // if user has more than one outdated rstudio analysis, display plural phrasing
+        [_.size(outdatedAnalyses) > 1, () => [p(`These R markdown files are being edited by another user and your versions are now outdated. Your files will no longer sync with the workspace bucket.`),
+          p(getAnalysesDisplayList(outdatedAnalyses)),
+          p('You can'),
```
kyuksel:
```suggestion
          p(['You can']),
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +63,89 @@ const ApplicationLauncher = _.flow(
...
+          p(['1) ', strong(['save your changes as new copies']), ' of your files which will enable file syncing on the copies']),
+          p([strong('or')]),
```
kyuksel:
```suggestion
          p([strong(['or'])]),
```

kyuksel:
```suggestion
          p([strong(['or'])]),
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +63,89 @@ const ApplicationLauncher = _.flow(
...
+          p(['2) ', strong(['continue working on your versions']), ` of ${getAnalysesDisplayList(outdatedAnalyses)} with file syncing disabled.`])]],
+        // if user has one outdated rstudio analysis, display singular phrasing
+        [_.size(outdatedAnalyses) === 1, () => [p(`${getAnalysisNameFromList(outdatedAnalyses)} is being edited by another user and your version is now outdated. Your file will no longer sync with the workspace bucket.`),
```
kyuksel:
```suggestion
        [_.size(outdatedAnalyses) === 1, () => [p([`${getAnalysisNameFromList(outdatedAnalyses)} is being edited by another user and your version is now outdated. Your file will no longer sync with the workspace bucket.`]),
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +63,89 @@ const ApplicationLauncher = _.flow(
...
+          p('You can'),
```
kyuksel:
```suggestion
          p(['You can']),
```

src/pages/workspaces/workspace/applications/AppLauncher.js
```diff
@@ -55,6 +63,89 @@ const ApplicationLauncher = _.flow(
...
+          p(['1) ', strong(['save your changes as a new copy']), ` of ${getAnalysisNameFromList(outdatedAnalyses)} which will enable file syncing on the copy`]),
+          p([strong('or')]),
```
kyuksel:
```suggestion
          p([strong(['or'])]),
```

kyuksel (approved):
A few more minor comments but looks good overall 👍🏼
Thanks for working through all the comments. 
Note: Didn't get to test the code.

kyuksel:
Merging since the integration test runs have failed due to unrelated tests. Jira tickets tracking failures of those tests will be updated.

PR #2811, opened 2022-02-08
[QA-1700] Further cleanup related to run-workflow-on-snapshot

cahrens:
This builds on top of https://github.com/DataBiosphere/terra-ui/pull/2806 and is mostly cleanup.

### DIFF
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -1,4 +1,4 @@
-const { checkbox, click, clickable, clickTableCell, fillIn, noSpinnersAfter, waitForNoSpinners } = require('../utils/integration-utils')
+const { checkbox, click, clickable, clickTableCell, fillIn, findText, noSpinnersAfter, select, waitForNoSpinners } = require('../utils/integration-utils')
 const { checkBucketAccess, enableDataCatalog, testWorkspaceName } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
@@ -15,13 +15,12 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) 
   const newWorkspaceBillingAccount = 'general-dev-billing-account'
   await click(page, clickable({ textContains: 'Start with a new workspace' }))
   await fillIn(page, '//*[@placeholder="Enter a name"]', `${newWorkspaceName}`)
-  await click(page, clickable({ text: 'Select a billing project' }))
-  await click(page, clickable({ text: `${newWorkspaceBillingAccount}` }))
-  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Create Workspace' })) })
+  await select(page, 'Billing project', `${newWorkspaceBillingAccount}`)
   try {
+    await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Create Workspace' })) })
     // Wait for bucket access to avoid sporadic failures
     await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
-    await page.url().includes(newWorkspaceName)
+    await findText(page, `${newWorkspaceBillingAccount}/${newWorkspaceName}`)
   } finally {
     try {
       await page.evaluate((name, billingProject) => {
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -16,8 +16,8 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) 
   await click(page, clickable({ textContains: 'Start with a new workspace' }))
   await fillIn(page, '//*[@placeholder="Enter a name"]', `${newWorkspaceName}`)
   await select(page, 'Billing project', `${newWorkspaceBillingAccount}`)
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Create Workspace' })) })
   try {
-    await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Create Workspace' })) })
     // Wait for bucket access to avoid sporadic failures
     await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
     await findText(page, `${newWorkspaceBillingAccount}/${newWorkspaceName}`)
@@ -26,7 +26,7 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) 
       await page.evaluate((name, billingProject) => {
         return window.Ajax().Workspaces.workspace(billingProject, name).delete()
       }, `${newWorkspaceName}`, `${newWorkspaceBillingAccount}`)
-    } catch(e) {
+    } catch (e) {
       console.error(`Error deleting workspace: ${e.message}`)
     }
   }
--- a/integration-tests/tests/run-catalog-workflow.js
+++ b/integration-tests/tests/run-catalog-workflow.js
@@ -21,9 +21,12 @@ const testCatalogFlowFn = _.flow(
   await click(page, clickable({ textContains: 'Start with an existing workspace' }))
   await click(page, input({ labelContains: 'Select a workspace' }))
   await click(page, `//*[@role="combobox"][contains(normalize-space(.), "${workspaceName}")]`)
-  await click(page, clickable({ textContains: 'Import' }))
-  await waitForNoSpinners(page)
-  await page.url().includes(workspaceName)
+  // The click call above is not actually selecting the workspace, so the Import button actually never becomes enabled.
+  // The "page.url().includes(workspaceName)" call returns false, but that doesn't actually fail the test.
+  // Instead of trying to check the URL, the test should check that the correct page loaded by looking for an expected element on the page.
+  // await click(page, clickable({ textContains: 'Import' }))
+  // await waitForNoSpinners(page)
+  // await page.url().includes(workspaceName)
 })
 
 const testCatalog = {
--- a/integration-tests/tests/run-workflow-on-snapshot.js
+++ b/integration-tests/tests/run-workflow-on-snapshot.js
@@ -1,7 +1,7 @@
 const _ = require('lodash/fp')
 const fetch = require('node-fetch')
 const { launchWorkflowAndWaitForSuccess } = require('./run-workflow')
-const { withWorkspace } = require('../utils/integration-helpers')
+const { checkBucketAccess, withWorkspace } = require('../utils/integration-helpers')
 const { click, clickable, dismissNotifications, enabledClickable, fillInReplace, findElement, findText, input, select, signIntoTerra, waitForNoSpinners, navChild } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
@@ -24,7 +24,7 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   withWorkspace,
   withUserToken,
   withDataRepoCheck
-)(async ({ dataRepoUrlRoot, page, testUrl, snapshotColumnName, snapshotId, snapshotTableName, token, workflowName, workspaceName }) => {
+)(async ({ billingProject, dataRepoUrlRoot, page, testUrl, snapshotColumnName, snapshotId, snapshotTableName, token, workflowName, workspaceName }) => {
   if (!snapshotId) {
     return
   }
@@ -36,6 +36,8 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await click(page, clickable({ textContains: 'Start with an existing workspace' }))
   await select(page, 'Select a workspace', workspaceName)
   await click(page, clickable({ text: 'Import' }))
+  // Wait for bucket access to avoid sporadic failure when launching workflow.
+  await checkBucketAccess(page, billingProject, workspaceName)
 
   // ADD WORKFLOW
   await click(page, navChild('workflows'))
--- a/integration-tests/tests/run-workflow-on-snapshot.js
+++ b/integration-tests/tests/run-workflow-on-snapshot.js
@@ -61,7 +61,6 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await fillInReplace(page, input({ labelContains: 'echo_to_file out attribute' }), 'workspace.result')
 
   await click(page, clickable({ text: 'Save' }))
-  await findElement(page, clickable({ text: 'Run analysis', isDisabled: false }))
 
   await launchWorkflowAndWaitForSuccess(page)
 
--- a/integration-tests/tests/run-workflow-on-snapshot.js
+++ b/integration-tests/tests/run-workflow-on-snapshot.js
@@ -2,7 +2,7 @@ const _ = require('lodash/fp')
 const fetch = require('node-fetch')
 const { launchWorkflowAndWaitForSuccess } = require('./run-workflow')
 const { checkBucketAccess, withWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, clickableEnabled, dismissNotifications, fillInReplace, findElement, findText, input, select, signIntoTerra, waitForNoSpinners, navChild } = require('../utils/integration-utils')
+const { click, clickable, dismissNotifications, fillInReplace, findElement, findText, input, select, signIntoTerra, waitForNoSpinners, navChild } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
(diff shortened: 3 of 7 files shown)


integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -36,6 +36,8 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await click(page, clickable({ textContains: 'Start with an existing workspace' }))
   await select(page, 'Select a workspace', workspaceName)
   await click(page, clickable({ text: 'Import' }))
+  // Wait for bucket access to avoid sporadic failure when launching workflow.
+  await checkBucketAccess(page, billingProject, workspaceName)
```
cahrens, line 40: Unfortunately this likely will not fix all instances of bucket access not being available, as I posted in [Slack](https://broadinstitute.slack.com/archives/C01EHNUM73R/p1644354756201859).

integration-tests/tests/run-workflow.js
```diff
@@ -58,22 +59,27 @@ const launchWorkflowAndWaitForSuccess = async page => {
   ])
   // stopLoggingPageAjaxResponses()
 
-  const disablePageLogging = logPageConsoleMessages(page)
```
cahrens, line 61: I misunderstood the point of this method… it is not for logging related to tests, but console logs from Terra UI itself. So removing this usage and switching to `rawConsole` to lessen noise in the tests.

integration-tests/tests/run-workflow.js
```diff
@@ -58,22 +59,27 @@ const launchWorkflowAndWaitForSuccess = async page => {
...
     }
-  }, { retries: 15, factor: 1 })
-  disablePageLogging()
+  },
+  // Note about retries value:
+  // If there is a large backlog of workflows in the environment (for example, if there has been a large submission), workflows
+  // might sit in "Submitted" for a very long time. In that situation, it is possible that this test will fail even with a 12-minute
+  // retries value. The test runner will kill any test that goes over 15 minutes though, so keep retries under that threshold.
+  // BW-1057 should further reduce this retries value after changes have been implemented to move workflows through Cromwell more quickly.
+  { retries: 12, factor: 1 }
```
cahrens, line 81: I had increased this from 5 retries to 15. But then when performance testing was going on in dev, we were seeing test failures because this test went over 15 minutes. There were log messages from this loop up until 13.x minutes, so reducing to 12 so that the cause of failure is more obvious (Workflow does not get to Succeeded in allotted time, vs. test runner killing the test for being too long).

integration-tests/utils/integration-utils.js
```diff
@@ -244,7 +246,7 @@ const withScreenshot = _.curry((testName, fn) => async options => {
 })
 
 const logPageConsoleMessages = page => {
-  const handle = msg => console.log('page.console', msg.text(), msg)
+  const handle = msg => rawConsole.log('page.console', msg.text(), msg)
```
cahrens, line 244: No longer using this method (in fact, nobody is), but changing to `rawConsole` for the future.

integration-tests/tests/run-workflow-on-snapshot.js
cahrens: I do think it is worth adding this check to address the more common problem of access not being granted fast enough for it to be present at the time that the workflow is launched. The fact that access can fail after having previously succeeded is a different (and very disappointing) issue.

integration-tests/utils/integration-utils.js
```diff
@@ -45,14 +45,16 @@ const getAnimatedDrawer = textContains => {
   return `//*[@role="dialog" and @aria-hidden="false"][contains(normalize-space(.), "${textContains}") or contains(@aria-label,"${textContains}") or @aria-labelledby=//*[contains(normalize-space(.),"${textContains}")]]`
 }
 
-const clickable = ({ text, textContains, isDescendant = false }) => {
-  const base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])`
+const clickable = ({ text, textContains, isDescendant = false, isDisabled = undefined }) => {
+  let base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])`
```
petesantos: How do you feel about this?

```suggestion
  const checkDisabled = isDisabled !== undefined ? `[@aria-disabled=${isDisabled}]` : ''
  const base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])${checkDisabled}`
```

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -59,8 +61,7 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await fillInReplace(page, input({ labelContains: 'echo_to_file out attribute' }), 'workspace.result')
 
   await click(page, clickable({ text: 'Save' }))
-  // Wait for "Run Analysis" to become enabled.
-  await page.waitForXPath(enabledClickable({ text: 'Run analysis' }))
+  await clickableEnabled(page, { text: 'Run analysis' })
```
petesantos: With the optional `isDisabled`
we could do:
```javascript
await click(page, clickable({text: 'Run analysis', isDisabled: false}))
```
or in this case

```javascript
await findElement(page, clickable({text: 'Run analysis', isDisabled: false}))
```
Just thinking about the conversation we had with @breilly2. This may converge on both points - being deliberate about disabled buttons when we want to, but not having to worry about it in the general case. I don't feel very strongly either way (this or an explicit function), but I do like calling out in the test: "I know this button is coming from a disabled state and have accounted for that"

We also could put the wait for enabled directly in the launch helper function so it is not decoupled from that code.

petesantos (commented):
Looks good, a few minor comments.
I did tag @breilly2 to get his eyes since he had some thoughts on the disabled buttons

integration-tests/tests/run-workflow-on-snapshot.js
cahrens: The reason I didn't put it in the launch helper function is because in the other test, `run-workflow`, the button should never be disabled. 

My slight preference is to go with `await findElement(page, clickable({text: 'Run analysis', isDisabled: false}))` or keep the separate method… but I don't have very strong feelings as long as there is a way to indicate that we are expecting the button to be initially disabled.

integration-tests/utils/integration-utils.js
```diff
@@ -45,14 +45,17 @@ const getAnimatedDrawer = textContains => {
...
 }
 
-const clickable = ({ text, textContains, isDescendant = false }) => {
-  const base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])`
+const clickable = ({ text, textContains, isDescendant = false, isDisabled = undefined }) => {
+  const checkDisabled = isDisabled !== undefined ? `[@aria-disabled=${isDisabled}]` : ''
+  const base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])${checkDisabled}`
+  if (isDisabled !== undefined) {
+    base += `[@aria-disabled="${isDisabled}"]`
+  }
```
petesantos: I should have included these lines in my earlier suggestion - sorry!

```suggestion
```

cahrens: No problem!

ungwudik (commented):
Test changes look good to me. I'd defer approval to Pete and Brian, though, since they had inputs earlier.

integration-tests/tests/run-workflow-on-snapshot.js
breilly2: I still feel that, 99% of the time, asking for something `clickable` is because you want to `click` it. Therefore, `isDisabled: false` should be the default.

I don't see a lot of value in worrying about whether or not the button is or was disabled here. This test is to assert that a person will be able to successfully run a workflow against a snapshot. It is not to assert that, when running a workflow against a snapshot, the button goes from disabled to enabled (and that it's never disabled when running a workflow against entity data).

(BTW, I was never fully convinced that running a workflow against a snapshot was sufficiently different than running a workflow against entity data to warrant a separate test, though I'll admit that I never questioned it aloud or thought too hard about it.)

petesantos (approved):

integration-tests/tests/run-workflow-on-snapshot.js
petesantos: I see - that makes sense.

I think I conflated two points of my argument.
1. The possibility of wanted to check, during a journey, that some functionality is working - like verifying user inputs before allowing them to proceed.
2. A code maintenance / dev ergonomics perspective (not really a "we need to test for this" concept). If a test fails, one of the first steps I take is to run through the journey myself. While I do so I like to look at what the test is doing. If I were new to the codebase I could see coming across a disabled button as part of that journey and thinking that is related to the issue. Seeing something in the test code as a hint that the test dev accounted for this may save me some time... But this may not be a huge win.

I am ok proceeding either way.

cahrens: Unfortunately disabled always present is causing a number of test failures. I suspect anchor tags, as I really wouldn't expect aria-disabled on either anchor or button elements (although we do support setting aria-disabled to true on an anchor tag).

I really am having trouble feeling excited about figuring out an xpath that will work for all of our potentially clickable items. I personally would rather kick that down the road until it is truly needed.

cahrens: It turns out that there are a lot of times when we will not have the `aria-disabled` property present, even on things with role="button". Therefore I'm going back to the explicit call with the expectation that it is only used when you know that the element you are waiting on has the expectation of being initially disabled.

breilly2: Oh, I see. I should have been expressing my thoughts in terms of behavior instead of implementation. I was trying to say that `clickable` should default to only returning things that aren't disabled. However, this is tricky to query for because the element might or might not have an explicit `aria-disabled` attribute.

cahrens: Yes, exactly. We have to take into account firecloud UI "buttons" (method repository), which are anchor tags without aria, and "Save cohort" which is an actual button, and I don't know what else. And then we have to look for the absence of attributes/properties that would indicate disabled. I think this is out-of-scope at this point.

cahrens:
Only test failure is preview-drs-uri, which is failing everywhere. I believe this is now ready for re-review.

integration-tests/tests/run-catalog-workflow.js
```diff
@@ -21,9 +21,12 @@ const testCatalogFlowFn = _.flow(
   await click(page, clickable({ textContains: 'Start with an existing workspace' }))
   await click(page, input({ labelContains: 'Select a workspace' }))
   await click(page, `//*[@role="combobox"][contains(normalize-space(.), "${workspaceName}")]`)
-  await click(page, clickable({ textContains: 'Import' }))
-  await waitForNoSpinners(page)
-  await page.url().includes(workspaceName)
+  // The click call above is not actually selecting the workspace, so the Import button actually never becomes enabled.
```
cahrens, line 24: @petesantos can you create a ticket for this test to be fixed?

integration-tests/tests/link-to-new-workspace.js
```diff
@@ -15,19 +15,18 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token })
...
   await fillIn(page, '//*[@placeholder="Enter a name"]', `${newWorkspaceName}`)
-  await click(page, clickable({ text: 'Select a billing project' }))
-  await click(page, clickable({ text: `${newWorkspaceBillingAccount}` }))
+  await select(page, 'Billing project', `${newWorkspaceBillingAccount}`)
   await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Create Workspace' })) })
   try {
     // Wait for bucket access to avoid sporadic failures
     await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
-    await page.url().includes(newWorkspaceName)
+    await findText(page, `${newWorkspaceBillingAccount}/${newWorkspaceName}`)
```
cahrens, line 23: @samanehsan FYI, removing the page.url call, which won't actually fail the test even if it returns `false`, with looking for the expected path in the top context bar.

integration-tests/utils/integration-utils.js
```diff
@@ -45,13 +45,10 @@ const getAnimatedDrawer = textContains => {
...
-const clickable = ({ text, textContains, isDescendant = false }) => {
-  const base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])`
-  return getClickablePath(base, text, textContains, isDescendant)
-}
-
-const enabledClickable = ({ text, textContains, isDescendant = false }) => {
-  const base = `(//a[@aria-disabled="false"] | //button[@aria-disabled="false"] | //*[@role="button" and @aria-disabled="false"] | //*[@role="link" and @aria-disabled="false"] | //*[@role="combobox" and @aria-disabled="false"] | //*[@role="option" and @aria-disabled="false"])`
+// Note: isEnabled is not supported for native anchor and button elements.
+const clickable = ({ text, textContains, isDescendant = false, isEnabled = true }) => {
+  const checkEnabled = isEnabled !== undefined ? ` and @aria-disabled!="${isEnabled}"` : ''
```
breilly2: This might need a more explicit conditional. If `isEnabled` is `false`, I think the check should be `@aria-disabled="true"`. `@aria-disabled!="false"` will succeed for enabled elements that don't have an `aria-disabled` attribute.

cahrens: I agree and did try that more straightforward implementation. I then got some test failures which indicate that we can't assume aria attributes to always be present on some of these elements-- probably combobox and option. We can either remove the aria check from those elements and see what happens, or we can keep the code as-is and it at least ensures that an element isn't disabled when you expect it to be enabled. I prefer to keep the code as-is, recognizing that there is a weakness here, but it works fine for elements that we always expect to have the aria-disabled property on them.

integration-tests/tests/link-to-new-workspace.js
samanehsan: Thanks!

integration-tests/tests/run-catalog-workflow.js
cahrens: Actually, it looks like @samanehsan is working on this now, so a ticket probably isn't necessary. It is possible though that the [new clickable calls ](https://github.com/DataBiosphere/terra-ui/pull/2814/files#diff-178f4d99efcf77626c2b99bb6a26712a04e14fa0503cd326823561090ecee1beR19) will not behave nicely with the "isEnabled" change on this branch, but if necessary "isEnabled" can be set to undefined.

breilly2 (approved):
👍  Thanks for sticking with this!

integration-tests/tests/run-catalog-workflow.js
petesantos: We can take a look and get it corrected - thanks!

integration-tests/utils/integration-utils.js
```diff
@@ -45,13 +45,10 @@ const getAnimatedDrawer = textContains => {
...
-  const base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])`
-  return getClickablePath(base, text, textContains, isDescendant)
-}
-
-const enabledClickable = ({ text, textContains, isDescendant = false }) => {
-  const base = `(//a[@aria-disabled="false"] | //button[@aria-disabled="false"] | //*[@role="button" and @aria-disabled="false"] | //*[@role="link" and @aria-disabled="false"] | //*[@role="combobox" and @aria-disabled="false"] | //*[@role="option" and @aria-disabled="false"])`
+// Note: isEnabled is not supported for native anchor and button elements.
+const clickable = ({ text, textContains, isDescendant = false, isEnabled = true }) => {
+  const checkEnabled = isEnabled !== undefined ? (isEnabled === false ? ` and @aria-disabled="${!isEnabled}"` : ` and @aria-disabled!="${isEnabled}"`) : ''
+  const base = `(//a | //button | //*[@role="button"${checkEnabled}] | //*[@role="link"${checkEnabled}] | //*[@role="combobox"${checkEnabled}] | //*[@role="option"${checkEnabled}])`
```
petesantos: No need to change this, we have spent a lot of time on it plus testing.
This is the xpath I was suggesting in slack:

```javascript
const checkEnabled = isEnabled === false ?  '[@aria-disabled="true"]' : '[not(@aria-disabled="true")]' 
const base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])${checkEnabled}`
```

Using `not` means the `@aria-disabled` attribute can be missing and it will still resolve the element.

PR #2814, opened 2022-02-10
[DC-208] Fix run-catalog-workflow test

samanehsan:
Changes:
- Check for text on the workspace page instead of the url
- Fix selecting a workspace
- Throw a more descriptive error when creating and deleting a workspace (otherwise it just says `Evaluation failed: Response` and it's difficult to tell what caused it)

### DIFF
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -1,16 +1,11 @@
-const { checkbox, click, clickable, clickTableCell, fillIn, findText, noSpinnersAfter, select, waitForNoSpinners } = require('../utils/integration-utils')
-const { checkBucketAccess, enableDataCatalog, testWorkspaceName } = require('../utils/integration-helpers')
+const { click, clickable, fillIn, findText, noSpinnersAfter, select } = require('../utils/integration-utils')
+const { checkBucketAccess, enableDataCatalog, goToLinkWorkspacePage, testWorkspaceName } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
 const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) => {
   await enableDataCatalog(page, testUrl, token)
-  await click(page, clickable({ textContains: 'browse & explore' }))
-
-  await click(page, checkbox({ text: 'Granted', isDescendant: true }))
-  await clickTableCell(page, 'dataset list', 2, 2)
-  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Link to a workspace' })) })
-
+  await goToLinkWorkspacePage(page)
   const newWorkspaceName = testWorkspaceName()
   const newWorkspaceBillingAccount = 'general-dev-billing-account'
   await click(page, clickable({ textContains: 'Start with a new workspace' }))
@@ -21,6 +16,7 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) 
     // Wait for bucket access to avoid sporadic failures
     await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
     await findText(page, `${newWorkspaceBillingAccount}/${newWorkspaceName}`)
+    await findText(page, 'Select a data type')
   } finally {
     try {
       await page.evaluate((name, billingProject) => {
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -1,11 +1,11 @@
 const { click, clickable, fillIn, findText, noSpinnersAfter, select } = require('../utils/integration-utils')
-const { checkBucketAccess, enableDataCatalog, goToLinkWorkspacePage, testWorkspaceName } = require('../utils/integration-helpers')
+const { checkBucketAccess, testWorkspaceName } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
+const { linkDataToWorkspace } = require('./run-catalog-workflow')
 
 
-const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token }) => {
-  await enableDataCatalog(page, testUrl, token)
-  await goToLinkWorkspacePage(page)
+const testLinkToNewWorkspaceFn = withUserToken(async ({ page, testUrl, token }) => {
+  await linkDataToWorkspace(page, testUrl, token)
   const newWorkspaceName = testWorkspaceName()
   const newWorkspaceBillingAccount = 'general-dev-billing-account'
   await click(page, clickable({ textContains: 'Start with a new workspace' }))
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -1,20 +1,9 @@
+const { linkDataToWorkspace, eitherThrow } = require('../utils/catalog-utils')
 const { click, clickable, fillIn, findText, noSpinnersAfter, select } = require('../utils/integration-utils')
 const { checkBucketAccess, testWorkspaceName } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
-const { linkDataToWorkspace } = require('./run-catalog-workflow')
 
 
-// Helper
-const eitherThrow = (testFailure, { cleanupFailure, cleanupMessage }) => {
-  if (testFailure) {
-    cleanupFailure && console.error(`${cleanupMessage}: ${cleanupFailure.message}`)
-    throw testFailure
-  } else if (cleanupFailure) {
-    throw new Error(`${cleanupMessage}: ${cleanupFailure.message}`)
-  }
-}
-
-// Test
 const testLinkToNewWorkspaceFn = withUserToken(async ({ page, testUrl, token }) => {
   await linkDataToWorkspace(page, testUrl, token)
   const newWorkspaceName = testWorkspaceName()
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -4,6 +4,17 @@ const { withUserToken } = require('../utils/terra-sa-utils')
 const { linkDataToWorkspace } = require('./run-catalog-workflow')
 
 
+// Helper
+const eitherThrow = (testFailure, { cleanupFailure, cleanupMessage }) => {
+  if (testFailure) {
+    cleanupFailure && console.error(`${cleanupMessage}: ${cleanupFailure.message}`)
+    throw testFailure
+  } else if (cleanupFailure) {
+    throw new Error(`${cleanupMessage}: ${cleanupFailure.message}`)
+  }
+}
+
+// Test
 const testLinkToNewWorkspaceFn = withUserToken(async ({ page, testUrl, token }) => {
   await linkDataToWorkspace(page, testUrl, token)
   const newWorkspaceName = testWorkspaceName()
@@ -12,19 +23,35 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ page, testUrl, token }) 
   await fillIn(page, '//*[@placeholder="Enter a name"]', `${newWorkspaceName}`)
   await select(page, 'Billing project', `${newWorkspaceBillingAccount}`)
   await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Create Workspace' })) })
-  try {
-    // Wait for bucket access to avoid sporadic failures
-    await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
-    await findText(page, `${newWorkspaceBillingAccount}/${newWorkspaceName}`)
-    await findText(page, 'Select a data type')
-  } finally {
+
+  const waitForWorkspacePage = async () => {
+    try {
+      await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
+      await findText(page, `${newWorkspaceBillingAccount}/${newWorkspaceName}`)
+      await findText(page, 'Select a data type')
+    } catch (error) {
+      return error
+    }
+  }
+
+  const cleanupFn = async () => {
     try {
       await page.evaluate((name, billingProject) => {
         return window.Ajax().Workspaces.workspace(billingProject, name).delete()
       }, `${newWorkspaceName}`, `${newWorkspaceBillingAccount}`)
-    } catch (e) {
-      console.error(`Error deleting workspace: ${e.message}`)
+    } catch (error) {
+      return error
     }
+  }
+
+  const workspaceFailure = await waitForWorkspacePage()
+  const cleanupFailure = await cleanupFn()
+
+  if (workspaceFailure || cleanupFailure) {
+    eitherThrow(workspaceFailure, {
+      cleanupFailure,
+      cleanupMessage: `Failed to delete workspace: ${newWorkspaceName} with billing project ${newWorkspaceBillingAccount}`
+    })
   }
 })
 
--- a/integration-tests/tests/run-catalog-workflow.js
+++ b/integration-tests/tests/run-catalog-workflow.js
@@ -1,32 +1,27 @@
 const _ = require('lodash/fp')
-const { checkbox, click, clickable, clickTableCell, input, waitForNoSpinners } = require('../utils/integration-utils')
-const { enableDataCatalog, withWorkspace } = require('../utils/integration-helpers')
+const { checkbox, click, clickable, clickTableCell, input, noSpinnersAfter, waitForNoSpinners } = require('../utils/integration-utils')
+const { checkBucketAccess, enableDataCatalog, withWorkspace } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
 const testCatalogFlowFn = _.flow(
   withWorkspace,
   withUserToken
-)(async ({ page, testUrl, token, workspaceName }) => {
+)(async ({ page, testUrl, token, workspaceName, billingProject }) => {
   await enableDataCatalog(page, testUrl, token)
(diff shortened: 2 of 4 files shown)


integration-tests/tests/run-catalog-workflow.js
```diff
@@ -1,28 +1,26 @@
...
-  await click(page, clickable({ textContains: 'Link to a workspace' }))
-  await waitForNoSpinners(page)
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Link to a workspace' })) })
 
   await click(page, clickable({ textContains: 'Start with an existing workspace' }))
   await click(page, input({ labelContains: 'Select a workspace' }))
   await click(page, `//*[@role="combobox"][contains(normalize-space(.), "${workspaceName}")]`)
   await click(page, clickable({ textContains: 'Import' }))
   await waitForNoSpinners(page)
+  await checkBucketAccess(page, billingProject, workspaceName)
```
samanehsan: This check actually might not be necessary since the workspace is created at the start of the test, but we could leave it in case it takes a while for the access to propagate

cahrens: It's definitely not useful here because the "Import" isn't actually getting clicked on (that's what I see when I run the test on dev locally with Chrome visible). The page URL check won't actually fail the test.

See [my PR](https://github.com/DataBiosphere/terra-ui/pull/2811/files#r803768256) where I'm commenting out these lines because the clickable xpath being used for "Import" is changing in my PR to require the button to be enabled. It then causes this test to fail because Import is not actually enabled (and thus the click is a no-op).

integration-tests/utils/integration-helpers.js
```diff
@@ -37,11 +37,15 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
...
-  }, workspaceName, billingProject)
+  try {
+    await page.evaluate((name, billingProject) => {
+      return window.Ajax().Workspaces.workspace(billingProject, name).delete()
+    }, workspaceName, billingProject)
 
-  rawConsole.info(`deleted workspace: ${workspaceName}`)
+    rawConsole.info(`deleted workspace: ${workspaceName}`)
+  } catch (e) {
+    console.error(`Failed to delete workspace: ${workspaceName}. Error: ${e.message}`)
```
cahrens: How would someone know to look for this? Should we fail the test?

Note also we typically use rawConsole, but perhaps the usage of console here was intentional to make this stand out more (you will get a bit of a stack trace)?

integration-tests/tests/run-catalog-workflow.js
cahrens: I'm also curious what failure signature you are trying to solve with waiting for bucket access.

samanehsan: Thanks, I'll take a look! Since we're creating the workspace at the start of the test, it seems like there is enough time for the bucket access to propagate but I'm not sure how reliable that is.

cahrens: @samanehsan my PR has merged, so you should rebase your branch to pick up the changes and test your new code with the changes that have been made to clickable.

samanehsan:
I'm running some more scale tests now 👀

integration-tests/utils/integration-helpers.js
samanehsan: I added this in for debugging since I thought the delete was failing and obscuring an actual test error. It turns out the `Evaluation failed` error was in creating a workspace (seen 1/100 times) so I updated that to throw a more descriptive error message and changed the delete workspace function to do the same thing.

integration-tests/tests/run-catalog-workflow.js
```diff
@@ -10,23 +10,15 @@ const testCatalogFlowFn = _.flow(
 )(async ({ page, testUrl, token, workspaceName }) => {
   await enableDataCatalog(page, testUrl, token)
```
cahrens: The first 6 lines of this test are identical to lines 7-12 from `link-to-new-workspace`. I would recommend that you share the code between the two tests.

integration-tests/tests/run-catalog-workflow.js
```diff
@@ -10,23 +10,15 @@ const testCatalogFlowFn = _.flow(
...
-  await click(page, `//*[@role="combobox"][contains(normalize-space(.), "${workspaceName}")]`)
-  // The click call above is not actually selecting the workspace, so the Import button actually never becomes enabled.
-  // The "page.url().includes(workspaceName)" call returns false, but that doesn't actually fail the test.
-  // Instead of trying to check the URL, the test should check that the correct page loaded by looking for an expected element on the page.
-  // await click(page, clickable({ textContains: 'Import' }))
-  // await waitForNoSpinners(page)
-  // await page.url().includes(workspaceName)
+  await select(page, 'Select a workspace', `${workspaceName}`)
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Import' })) })
+  await findText(page, workspaceName)
```
cahrens: I would suggest looking for something a bit more specific than just the workspace name, because I would imagine that the workspace name can appear on a lot of pages. One possibility is the combination of the billing project and workspace name, which is what we show in the top context bar (though even that doesn't feel highly specific to the page to make sure loaded). `await findText(page, `${billingProject}/${workspaceName}`)`, which requires you to pass the `billingProject` into the test.

Is there any text or element on the page you expect to load that is definitely unique to that page?

cahrens:
I'm running this with the flaky runner and getting a lot of this error signature (7/40):

````
TimeoutError: waiting for XPath `(//input[@type="checkbox"] | //*[@role="checkbox"])//*[normalize-space(.)="Granted" or @title="Granted" or @aria-label="Granted" or @aria-labelledby=//*[normalize-space(.)="Granted"]/@id]` failed: timeout 30000ms exceeded

File: integration-tests/tests/run-catalog-workflow.js (hunk 4 above)
Comment (samanehsan) [reply integration-tests/tests/run-catalog-workflow.js]: There isn't much else on the page:
Screen Shot 2022-02-10 at 4 09 49 PM

Comment (samanehsan) [reply integration-tests/tests/run-catalog-workflow.js]: How about `await findText(page, ${billingProject}/${workspaceName})` and `await findText(page, 'Workspace Data')` since we expect to be in the data tab?

Comment (cahrens) [reply integration-tests/tests/run-catalog-workflow.js]: The combination of `await findText(page, ${billingProject}/${workspaceName})` and looking for "Select a data type" would be good.

**Comment by samanehsan** [comment]
@cahrens what kind of computer are you using? I'm on a 2017 macbook pro with 16GB of memory so I've been running the tests with `CONCURRENCY=5`. If I use the default setting to run 10 concurrent tests a bunch of them timeout with that error.

**Comment by cahrens** [comment]
I have a 2021 MacBook Pro 32GB of memory. I'm using the default concurrency option of 10, which is what I've been using for all of my flaky testing. I've been able to get clean runs (out of 100) on other tests, so I don't necessarily think this is related to the computer's capabilities.

**Comment by samanehsan** [comment]
> I have a 2021 MacBook Pro 32GB of memory. I'm using the default concurrency option of 10, which is what I've been using for all of my flaky testing. I've been able to get clean runs (out of 100) on other tests, so I don't necessarily think this is related to the computer's capabilities.

Okay, that's good to know 🤔

File: integration-tests/utils/integration-helpers.js [hunk 5]
```diff
@@ -26,22 +26,29 @@ const getTestWorkspaceName = () => `test-workspace-${Math.floor(Math.random() *
...
-    return window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {} })
-  }, workspaceName, billingProject)
-
-  rawConsole.info(`created workspace: ${workspaceName}`)
+  try {
+    await page.evaluate((name, billingProject) => {
+      return window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {} })
+    }, workspaceName, billingProject)
 
+    rawConsole.info(`created workspace: ${workspaceName}`)
```
kyuksel: Nit:

```suggestion
    rawConsole.info(`Created workspace: ${workspaceName}`)
```

since we typically initialize the first letter of a sentence.

Same for `deleteWorkspace` below.

integration-tests/utils/integration-helpers.js
```diff
@@ -26,22 +26,29 @@ const getTestWorkspaceName = () => `test-workspace-${Math.floor(Math.random() *
...
+  } catch (e) {
+    throw Error(`Failed to create workspace: ${workspaceName}`)
```
kyuksel: Would it be helpful to include `billingProject` in the error messages above and below?

kyuksel (commented):
Looks good overall. Pending [Christina's comment](https://github.com/DataBiosphere/terra-ui/pull/2814#discussion_r804100542).

kyuksel:
> what kind of computer are you using?

FWIW, the specs below are for my Mac, which is more similar to @samanehsan's. I have also been able to run `100` tests with concurrency set to `10`.

image

integration-tests/utils/integration-helpers.js
```diff
@@ -192,11 +192,19 @@ const enableDataCatalog = async (page, testUrl, token) => {
   await dismissNotifications(page)
 }
 
+const goToLinkWorkspacePage = async (page) => {
```
cahrens: I was imagining you would keep this within one of the tests that uses it (I don't think it's general enough to go in integration-helpers). And you can put `await enableDataCatalog(page, testUrl, token)` in it also.

integration-tests/tests/run-catalog-workflow.js
```diff
@@ -1,24 +1,20 @@
...
+const { click, clickable, findText, noSpinnersAfter, select } = require('../utils/integration-utils')
+const { enableDataCatalog, goToLinkWorkspacePage, withWorkspace } = require('../utils/integration-helpers')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
 const testCatalogFlowFn = _.flow(
   withWorkspace,
   withUserToken
-)(async ({ page, testUrl, token, workspaceName }) => {
+)(async ({ page, testUrl, token, workspaceName, billingProject }) => {
```
cahrens: I think we try to keep the variables in alphabetical order.

integration-tests/utils/integration-helpers.js
samanehsan: Oh I see! I find keeping `enableDataCatalog` separate makes the test read better but I'm fine either way

cahrens: Maybe call it something like "linkDataToWorkspace"? Then it would make sense that it includes enabling the data catalog (which it is interacting with in the test anyway).

samanehsan:
@kyuksel that's good to know! I have a 3.1 GHz Quad-Core Intel Core i7 processor

samanehsan:
@kyuksel @cahrens let me know if the changes look good!

integration-tests/tests/link-to-new-workspace.js
```diff
@@ -21,6 +16,7 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ testUrl, page, token })
     // Wait for bucket access to avoid sporadic failures
     await checkBucketAccess(page, newWorkspaceBillingAccount, newWorkspaceName)
     await findText(page, `${newWorkspaceBillingAccount}/${newWorkspaceName}`)
+    await findText(page, 'Select a data type')
   } finally {
     try {
       await page.evaluate((name, billingProject) => {
```
cahrens, line 28: Can't you just call `deleteWorkspace` here?

cahrens: If for some reason you cannot, you probably want to throw an error if the workspace can't be deleted, like you did in `deleteWorkspace`.

cahrens:
> I'm running this with the flaky runner and getting a lot of this error signature (7/40):
> 
> ```
> TimeoutError: waiting for XPath `(//input[@type="checkbox"] | //*[@role="checkbox"])//*[normalize-space(.)="Granted" or @title="Granted" or @aria-label="Granted" or @aria-labelledby=//*[normalize-space(.)="Granted"]/@id]` failed: timeout 30000ms exceeded
> ```

@samanehsan will you continue to investigate this failure signature ^ for a future PR? I agree with iteratively improving instead of trying to figure out every possible issue at once, but just want to make sure this does get followed up on.

samanehsan:
@cahrens yes, I'll make a follow-up ticket for that!

integration-tests/tests/link-to-new-workspace.js
samanehsan: If the test fails to create a workspace, this delete throws an error. If it's not handled that error obscures the original test error and makes it difficult to debug. Although now that I think of it I could check if the workspace exists before calling `deleteWorkspace` to avoid that situation.

cahrens: If a workspace isn't created, won't the test fail earlier? checkBucketAccess will error I believe, and you shouldn't be able to navigate to the workspace page.

samanehsan: The finally block runs even if there's an error thrown in the try block, so the original error gets lost

samanehsan: This comment from Pete has a useful code snippet to help observe this behavior: https://github.com/DataBiosphere/terra-ui/pull/2807#pullrequestreview-875241605

cahrens: I still think you can call `deleteWorkspace` and just catch any errors that it throws. That way you would not duplicate the deleteWorkspace code.

integration-tests/tests/run-catalog-workflow.js
```diff
@@ -36,4 +31,7 @@ const testCatalog = {
   targetEnvironments: ['local', 'dev']
 }
 
-module.exports = { testCatalog }
+module.exports = {
+  linkDataToWorkspace,
```
petesantos: I would rather not have the test export a utility function. I think this confuses the purpose of the file - it becomes both test and utility. We could create a `catalog-utils.js` file for this type of code

integration-tests/utils/integration-helpers.js
```diff
@@ -26,22 +26,29 @@ const getTestWorkspaceName = () => `test-workspace-${Math.floor(Math.random() *
...
-
-  rawConsole.info(`created workspace: ${workspaceName}`)
+  try {
+    await page.evaluate((name, billingProject) => {
+      return window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {} })
+    }, workspaceName, billingProject)
 
+    rawConsole.info(`Created workspace: ${workspaceName}`)
+  } catch (e) {
+    throw Error(`Failed to create workspace: ${workspaceName} with billing project ${billingProject}`)
```
petesantos, line 36: I think these error messages will be much clearer when they fail - thanks!

petesantos (commented):
This is looking good.
I ran run-catalog-workflow 100 times with 1 error, a ToS error, that is more systemic and out of scope for the test.
I ran link-to-new-workspace with the same result

petesantos:
> > I'm running this with the flaky runner and getting a lot of this error signature (7/40):
> > ```
> > TimeoutError: waiting for XPath `(//input[@type="checkbox"] | //*[@role="checkbox"])//*[normalize-space(.)="Granted" or @title="Granted" or @aria-label="Granted" or @aria-labelledby=//*[normalize-space(.)="Granted"]/@id]` failed: timeout 30000ms exceeded
> > ```
> 
> @samanehsan will you continue to investigate this failure signature ^ for a future PR? I agree with iteratively improving instead of trying to figure out every possible issue at once, but just want to make sure this does get followed up on.

FWIW I got that as well, but it looks like a ToS error from the screenshot:

integration-tests/tests/link-to-new-workspace.js
petesantos: @cahrens Are you referring to [this deleteWorkspace](https://github.com/DataBiosphere/terra-ui/blob/bf9423b28bcf8d0098c076e61e8dd28f5b7f62f3/integration-tests/utils/integration-helpers.js#L42) function?

If so, I tried out this function. It doesn't look like a super-clean abstraction and ends up running some additional code around sign in that I don't think we want to do. We could add a ticket to investigate creating a cleaner utility for this.

samanehsan:
I created a ticket for the timeout error here: https://broadworkbench.atlassian.net/browse/DC-228

integration-tests/tests/run-catalog-workflow.js
samanehsan: Ah okay! I think that's a good compromise here

integration-tests/utils/integration-helpers.js
samanehsan: I saw this test fail 2/100 times because it failed to create a workspace. Is this something that would get covered by the retry fix you wanted to do? If not I can make a separate ticket for this.

integration-tests/tests/link-to-new-workspace.js
cahrens: OK, I was hoping it would be straightforward to use, but if not feel free to punt it.

samanehsan: @petesantos and I paired on improving the error handling here! Now if the test fails it will throw the test failure (and log the cleanup error if that fails too). If the test succeeds and the cleanup fails, it throws the cleanup error.

integration-tests/utils/integration-helpers.js
petesantos: Yes, but I think retry is out of scope for this ticket. I think the improvements here will reduce a lot of failure noise / flakiness

kyuksel: I wonder if it makes sense to investigate why workspace creation fails before adding a retry, to make sure retry wouldn't conceal a back-end issue.

integration-tests/tests/run-catalog-workflow.js
cahrens: You forgot to remove the export of `linkDataToWorkspace` here.

samanehsan: Ah good catch! Just removed it

petesantos (approved):

cahrens (approved):

PR #2815, opened 2022-02-14
[IA-2967] Added UI components to make auto pause configurable from UI

cstrandw:
Related to: https://github.com/DataBiosphere/leonardo/pull/2548

- Added inputs to allow user to enable/disable autopause and set autopause threshold from ComputeModal
- Added autopauseThreshold value to runtime objects as higher level values
- Added default value for autopauseThreshold in runtime-utils

Testing:
- Created runtime with autopause enabled at a variety of thresholds and checked that the values were perpetuated to the db
- Tested turning on and off autopause and changing the threshold to check that the correct value was in db
- Created runtime with autopause disabled and that value was correct in db
- Created runtime with autopause enabled and threshold of 10 minutes and checked that runtime paused after 10 minutes of no activity


kyuksel:
@cstrandw Have you tested the following `patch` cases?
1. We can create a runtime with auto-pause **disabled**, and then **change** it to have auto-pause **enabled** with a custom threshold.
2. We can create a runtime with auto-pause **enabled**, and then **disable** it.

cstrandw:
> @cstrandw Have you tested the following `patch` cases?
> 
> 1. We can create a runtime with auto-pause **disabled**, and then **change** it to have auto-pause **enabled** with a custom threshold.
> 2. We can create a runtime with auto-pause **enabled**, and then **disable** it.

Yes I did, and it was all successful

src/components/ComputeModal.js
```diff
@@ -190,6 +190,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     hasGpu: false,
     gpuType: defaultGpuType,
     numGpus: defaultNumGpus,
+    autopauseEnabled: defaultAutoPause,
+    autopauseThreshold: defaultAutoPauseLength,
```
kyuksel: I see that we use both `threshold` and `length`. Can we pick one and use that so there's less terminology? I'd vote for `threshold` but it's your call :)

src/components/ComputeModal.js
```diff
@@ -626,6 +651,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
           .then(res => res.json()),
         currentPersistentDisk ? Ajax().Disks.disk(currentPersistentDisk.googleProject, currentPersistentDisk.name).details() : null
       ])
+
```
kyuksel: Can we remove the extra line?
```suggestion
```

src/components/ComputeModal.js
```diff
@@ -664,13 +691,16 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       const { computeZone, computeRegion } = getRegionInfo(location || defaultLocation, locationType)
       const runtimeConfig = currentRuntimeDetails?.runtimeConfig
       const gpuConfig = runtimeConfig?.gpuConfig
+      const autopauseConfig = currentRuntimeDetails ? { enabled: currentRuntimeDetails?.autopauseEnabled, threshold: currentRuntimeDetails?.autopauseThreshold } : null
```
kyuksel: We don't need to use optional chaining for `currentRuntimeDetails` since we know it isn't `undefined` if we're in the first part of the ternary.

When we use a non-boolean (e.g. `currentRuntimeDetails`) as a boolean value, we use `!!` to coerce for explicitness about the intention.

Also, applied formatting.

```suggestion
      const autopauseConfig = currentRuntimeDetails ?
        { enabled: currentRuntimeDetails?.autopauseEnabled, threshold: currentRuntimeDetails?.autopauseThreshold } :
        null
```

cahrens:
@cstrandw judging from the integration test failures, this branch probably is not up-to-date with dev. If you rebase, you should see cleaner test runs.

src/components/ComputeModal.js
```diff
@@ -664,13 +691,16 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
       const newSparkMode = Utils.switchCase(runtimeConfig?.cloudService,
         [cloudServices.DATAPROC, () => runtimeConfig.numberOfWorkers === 0 ? 'master' : 'cluster'],
         [cloudServices.GCE, () => false],
         [Utils.DEFAULT, () => false] // for when there's no existing runtime
       )
       const isDataproc = !sparkMode && !runtimeConfig?.diskSize
 
+      const calculateAutopauseThreshold = () => autopauseConfig ? (autopauseConfig.enabled ? autopauseConfig.threshold : 0) : defaultAutoPauseLength
```
kyuksel: We could write it a bit more concisely using `optional chaining` and `nullish coalescence` like below unless you feel it looks too complicated.
```suggestion
      const calculateAutopauseThreshold = () => autopauseConfig?.enabled ? (autopauseConfig.threshold ?? 0) : defaultAutoPauseLength
```

Also, do we need for it to be a function? It appears to be evaluated only once per `useOnMount()`.

src/components/ComputeModal.js
```diff
@@ -686,6 +716,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         hasGpu: !!gpuConfig,
         gpuType: gpuConfig?.gpuType || defaultGpuType,
         numGpus: gpuConfig?.numOfGpus || defaultNumGpus,
+        autopauseEnabled: autopauseConfig ? autopauseConfig.enabled : true,
```
```suggestion
        autopauseEnabled: autopauseConfig?.enabled || true,
```

src/components/ComputeModal.js
```diff
@@ -975,6 +1010,37 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
       ]),
+      !sparkMode && div({ style: { gridColumnEnd: 'span 6', marginTop: '1.5rem' } }, [
+        h(LabeledCheckbox, {
+          checked: computeConfig.autopauseEnabled,
+          disabled: !autoPauseCheckboxEnabled,
+          onChange: v => {
+            updateComputeConfig('autopauseEnabled', v)
+            // if auto pause is being turned on set threshold to default length,
+            // if auto pause is being disabled it sets threshold to 0 as this is consistent with how threshold is currently managed
+            v ? updateComputeConfig('autopauseThreshold', defaultAutoPauseLength) : updateComputeConfig('autopauseThreshold', 0)
```
kyuksel: Why are we updating `autopauseThreshold` here since we don't get the user input for it until `NumberInput` below?

kyuksel (commented):
Some comments inline. Will take a deeper look and test once the [UX questions](https://broadworkbench.atlassian.net/browse/IA-2967?focusedCommentId=56458) are sorted out.

cstrandw:
> @cstrandw judging from the integration test failures, this branch probably is not up-to-date with dev. If you rebase, you should see cleaner test runs.

Thanks for the heads up! I did that, but one test is still not passing

src/components/ComputeModal.js
cstrandw: I originally had it the way you're suggesting, but that caused it to default to true when enabled was false because of the or. I don't think this syntax works when both values are boolean.. Is there a better way I could do this logic?

kyuksel:
> Thanks for the heads up! I did that, but one test is still not passing

The test failures seem unrelated but let's not worry about it much until the PR is finalized since any further change will have to re-pass tests.

src/components/ComputeModal.js
cstrandw: On the back end, when autopause is disabled the threshold has historically been set to 0 so I thought just keeping this consistent in the front end would be a good way to go. I tried to add comments to explain this, but let me know if there is anything I can add to make this more clear or if I should not do this.

src/components/ComputeModal.js
cstrandw: Yes agreed! Making the change for that default value now.

cahrens:
> > @cstrandw judging from the integration test failures, this branch probably is not up-to-date with dev. If you rebase, you should see cleaner test runs.
> 
> Thanks for the heads up! I did that, but one test is still not passing

Yes, now you are hitting a different problem. :( We are seeing that sometimes the bucket location access "goes way" after it previously succeeded.

src/components/ComputeModal.js
kyuksel: Oh my bad! This is actually a good use case for the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) (`??`) instead of `||` since it uses the default value only if the predicate is nullish (i.e. `undefined` or `null`).

```suggestion
        autopauseEnabled: autopauseConfig?.enabled ?? true,
```

src/components/ComputeModal.js
kyuksel: Ah I see. The thing is, we typically want to keep React state updates to a minimum. Do you see any issues when you remove that line? I'd think it would work the same way since `autopauseThreshold` is already initialized to the default value on the top of the component and React re-renders once we `updateComputeConfig('autopauseEnabled', v)`.

cstrandw: Yeah that makes sense! I'll make that change now

src/components/ComputeModal.js
```diff
@@ -23,7 +23,7 @@ import { betaVersionTag } from 'src/libs/logos'
 import * as Nav from 'src/libs/nav'
 import { useOnMount } from 'src/libs/react-utils'
 import {
-  computeStyles, defaultComputeRegion, defaultComputeZone, defaultDataprocMachineType, defaultDataprocMasterDiskSize, defaultDataprocWorkerDiskSize,
+  computeStyles, defaultAutoPause, defaultAutoPauseThreshold, defaultComputeRegion, defaultComputeZone, defaultDataprocMachineType, defaultDataprocMasterDiskSize, defaultDataprocWorkerDiskSize,
```
dmohs: Consider reducing the number of explicit imports here, which is pretty noisy:
1. `import * as ru from 'src/libs/runtime-utils'` and use as `ru.computeStyles` etc.
2. Create a `defaults` object in runtime-utils and use that, i.e.,
```javascript
const [computeConfig, setComputeConfig] = useState(_.merge(ru.defaultComputeConfig, {/* overrides */}))
```

dmohs:
For your consideration: https://github.com/DataBiosphere/terra-ui/pull/2818

src/components/ComputeModal.js
```diff
@@ -664,6 +690,9 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       const { computeZone, computeRegion } = getRegionInfo(location || defaultLocation, locationType)
       const runtimeConfig = currentRuntimeDetails?.runtimeConfig
       const gpuConfig = runtimeConfig?.gpuConfig
+      const autopauseConfig = !! currentRuntimeDetails ?
```
kyuksel: We typically don't put space between unary operators (like `!`) and their operands. Note that `!!` is not a separate operator but two `!` operators, one after another.
```suggestion
      const autopauseConfig = !!currentRuntimeDetails ?
```

src/components/ComputeModal.js
```diff
@@ -686,6 +715,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         hasGpu: !!gpuConfig,
         gpuType: gpuConfig?.gpuType || defaultGpuType,
         numGpus: gpuConfig?.numOfGpus || defaultNumGpus,
+        autopauseEnabled: autopauseConfig?.enabled ?? true,
+        autopauseThreshold: autopauseConfig?.enabled ? (autopauseConfig.threshold ?? 0) : defaultAutoPauseThreshold,
```
kyuksel: I like @dmohs' [comment](https://github.com/DataBiosphere/terra-ui/pull/2818/files#diff-149d7ad595448fb2f6bedcff6dde5f4cbcccd8d5148fed8820ce88190a3bc38bR710) for the line above; i.e. using `Utils.switchCase` to flatten the nested ternaries, which to me makes it easier to read:

```javascript
const autopauseThreshold = Utils.switchCase(autopauseConfig?.enabled,
  [undefined, () => defaultAutoPauseLength],
  [false, () => 0],
  [true, () => autopauseConfig.threshold]
)
```

P.S. Utility methods in `utils.js` are in general worth getting familiar with as we tend to utilize them a lot in this codebase 😊

src/components/ComputeModal.js
```diff
@@ -975,6 +1013,37 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+            max: 999,
+            isClearable: false,
+            onlyInteger: true,
+            value: computeConfig.autopauseThreshold,
+            disabled: !computeConfig.autopause,
+            tooltip: !computeConfig.autopause ? 'Autopause must be enabled to configure pause time.' : undefined,
+            onChange: updateComputeConfig('autopauseThreshold')
+          }),
+          span({style: {width: '300px'}}, ['minutes of inactivity'])
+        ])
```
```suggestion
        div({ style: { ...gridStyle, gridGap: '0.8rem', gridTemplateColumns: '4.5rem 9.5rem', marginTop: '0.75rem' } }, [
          h(NumberInput, {
            min: 1,
            max: 999,
            isClearable: false,
            onlyInteger: true,
            value: computeConfig.autopauseThreshold,
            disabled: !computeConfig.autopause,
            tooltip: !computeConfig.autopause ? 'Autopause must be enabled to configure pause time.' : undefined,
            onChange: updateComputeConfig('autopauseThreshold')
          }),
          span(['minutes of inactivity'])
        ])
```


would achieve the following:
- remove unnecessary `gridTemplateColumns` since only the first two are relevant in this `div` (because we only have the number box and the text);
- use `rem` for both so it's easier to compare, etc.;
- override `gridGap` inherited from the parent `div` (through `…gridStyle`) to shorten it, which allows addressing the UX request to move the text closer to the number box. I put `0.8rem` but I'd check with UX to be sure;
- get rid of the custom `style`ing in `span()` since it's taken care of by the previous changes and it's good practice to minimize lower-level `style`ing.

With that change, the gap looks like the following:

image

src/libs/runtime-utils.js
```diff
@@ -40,6 +40,9 @@ export const defaultLocation = 'US'
 export const defaultComputeZone = 'US-CENTRAL1-A'
 export const defaultComputeRegion = 'US-CENTRAL1'
 
+export const defaultAutoPause = true
```
kyuksel: `defaultAutoPause` (and `autopause` by itself) sounds non-descriptive me. I liked the previous version (`autopauseEnabled`) which made it clear that it held boolean values. What do you think?

src/components/ComputeModal.js
```diff
@@ -1015,9 +993,9 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       ]),
       div({ style: { gridColumnEnd: 'span 6', marginTop: '1.5rem' } }, [
         h(LabeledCheckbox, {
-          checked: computeConfig.autopause,
+          checked: computeConfig.autopauseThreshold !== 0,
```
kyuksel: It'd be good to encapsulate the logic used to decide if autopause is enabled as well as the special value (`0`) via something like

```javascript
const autopauseDisabledValue = 0
const isAutopauseEnabled = threshold => threshold > autopauseDisabledValue
const getAutopauseThreshold = isEnabled => isEnabled ? defaultAutopauseThreshold : autopauseDisabledValue
```

and then we can use those `const`s above and below.

It may make sense to put them where `defaultAutoPauseThreshold` is defined so they're all in one place.

cstrandw: I like that, I'll make that change

src/components/ComputeModal.js
```diff
@@ -846,6 +859,9 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     const gpuCheckboxDisabled = computeExists ? !computeConfig.gpuEnabled : isDataproc(runtimeType)
     const enableGpusSpan = span(
       ['Enable GPUs ', betaVersionTag])
+    const autoPauseCheckboxEnabled = true
+    const enableAutoPauseSpan = span(
+      ['Enable autopause'])
```
kyuksel: Should be able to one-line this?

```suggestion
    const enableAutoPauseSpan = span(['Enable autopause'])
```

kyuksel:
Nit: I see `autopause`-related naming (a lot of which pre-date your PR) where `p` is sometimes capitalized and sometimes isn't. Can we pick one of the schemes and use it consistently? I'd vote for not capitalizing `p` since we tend to write it as `autopause` in documentation, comments etc. but it's up to you.

src/libs/runtime-utils.js
```diff
@@ -42,7 +42,10 @@ export const defaultLocation = 'US'
 export const defaultComputeZone = 'US-CENTRAL1-A'
 export const defaultComputeRegion = 'US-CENTRAL1'
 
-export const defaultAutoPauseThreshold = 30
+export const defaultAutopauseThreshold = 30
+export const defaultAutopauseDisabledValue = 0
```
kyuksel: It's always the same value so we could drop `default` for concision:
```suggestion
export const autopauseDisabledValue = 0
```

src/libs/runtime-utils.js
```diff
@@ -42,7 +42,10 @@ export const defaultLocation = 'US'
...
+export const isAutopauseEnabled = threshold => threshold > defaultAutopauseDisabledValue
```
kyuksel: A comment could be good here, something like:
```suggestion
// Leonardo considers autopause disabled when the threshold is set to 0
export const isAutopauseEnabled = threshold => threshold > defaultAutopauseDisabledValue
```

src/components/ComputeModal.js
```diff
@@ -304,11 +307,15 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       await Ajax().Disks.disk(googleProject, currentPersistentDiskDetails.name).delete()
     }
     if (shouldUpdateRuntime) {
-      await Ajax().Runtimes.runtime(googleProject, currentRuntimeDetails.runtimeName).update({ runtimeConfig })
+      await Ajax().Runtimes.runtime(googleProject, currentRuntimeDetails.runtimeName).update({
+        runtimeConfig,
+        autopauseThreshold: computeConfig.autopauseThreshold
```
jdcanas, line 311: this seems like it is not needed. It's added to the runtimeConfig object on line 266, it shouldn't need to also be a top-level field

src/components/ComputeModal.js
```diff
@@ -304,11 +307,15 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+      })
     }
     if (shouldCreateRuntime) {
       await Ajax().Runtimes.runtime(googleProject, Utils.generateRuntimeName()).create({
         runtimeConfig,
+        autopauseThreshold: computeConfig.autopauseThreshold,
```
jdcanas, line 317: Same comment as the update call. Is this supposed to be a field in the runtimeConfig, or a top-level field?

src/components/ComputeModal.js
```diff
@@ -415,6 +424,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
 
     return {
       hasGpu: computeConfig.hasGpu,
+      autopauseThreshold: computeConfig.autopauseThreshold,
```
jdcanas, line 426: can this be a field in the runtime as opposed to a top-level field? Not sure what exactly the distinction is, but I view it as part of a runtime. Curious to hear your thoughts

src/components/ComputeModal.js
cstrandw: It's supposed to be a top-level field, that is what Gabriela recommended when I paired with her

kyuksel: I'd missed this. By "top-level", do you mean outside `compute/runtimeConfig` state? If so, I'd like to understand why `autopauseThreshold` is supposed to be top-level. Cc: @gpcarr

cstrandw: Screen Shot 2022-02-28 at 9 30 36 AM
autopauseThreshold is returned from the API call outside of the runtimeConfig, so we wanted to maintain that format on the Terra side as well

cstrandw: @kyuksel @jdcanas  I can easily remove it from runtimeConfig since it's not being used, I thought I had done that but must have forgotten. Otherwise I can refactor it into runtimeConfig and test that out, thoughts?

kyuksel: I wonder if we may be conflating two distinct runtime configs. One is the state variable [computeConfig](https://github.com/DataBiosphere/terra-ui/blob/fa8d4a58fc9b6e0e25606ba1e63a6380c3c71a6d/src/components/ComputeModal.js#L213), which includes `autopauseThreshold` and there are also local `runtimeConfig`s like [this](https://github.com/DataBiosphere/terra-ui/blob/fa8d4a58fc9b6e0e25606ba1e63a6380c3c71a6d/src/components/ComputeModal.js#L266) that already include `autopauseThreshold`. We use the latter (`runtimeConfig`) to construct API request bodies so we shouldn't need to separately add `autopauseThreshold` to the response. Does that make sense?

cstrandw: Yes, I think so! Will let you know if I have any questions

cstrandw: @kyuksel Leonardo is expecting the autopauseThreshold value to be outside of the runtimeConfig in the API calls so I believe it does still need to be separately added to the response

kyuksel: Oh I think I see what you mean now. The duplicate is actually [line #266](https://github.com/DataBiosphere/terra-ui/blob/fa8d4a58fc9b6e0e25606ba1e63a6380c3c71a6d/src/components/ComputeModal.js#L266), which we should be able to remove?

cstrandw: @kyuksel Yes, I just pushed that change. The backend leo changes have been approved and merged, so this should be testable now. Thanks!

jdcanas (approved):

kyuksel:
The functionality works great! For testing, I took the following actions on the UI while verifying that the Leo DB has the expected values:
- create a runtime with autopause disabled
- patch the runtime to enable the autopause with threshold set to a custom value
- observe that the runtime is paused after that much time
- patch the runtime to set autopause to a different custom value
- patch the runtime to disable autopause

kyuksel:
Looks like there's one corner case we may want to guard against:
- I clicked on Pause on the compute modal.
- While the runtime was in `Pausing` status, I attempted to patch the runtime to disable autopause and got an error from Leo

image

It doesn't seem to be specific to patching a runtime for autopause-related reasons, though, so I've created the ticket below to handle it more holistically:

https://broadworkbench.atlassian.net/browse/IA-3230

kyuksel (approved):
:shipit:

PR #2816, opened 2022-02-14
Show "Spend report" tab under Billing for alpha users only (WOR-42).

cahrens:
This adds an (empty) `Spend report` tab in Billing for billing projects owned by the current user iff the user is a member of the alpha testing group.

To test:
1. Log in with hermione.owner and you should see the tab.
2. Log in with dumbledore.admin and you should not see the tab.

image

### DIFF
--- a/src/pages/billing/List.js
+++ b/src/pages/billing/List.js
@@ -13,6 +13,7 @@ import TopBar from 'src/components/TopBar'
 import { Ajax } from 'src/libs/ajax'
 import * as Auth from 'src/libs/auth'
 import colors from 'src/libs/colors'
+import { getConfig } from 'src/libs/config'
 import { reportErrorAndRethrow } from 'src/libs/error'
 import Events from 'src/libs/events'
 import { formHint, FormLabel } from 'src/libs/forms'
@@ -221,6 +222,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
   const [isLoadingProjects, setIsLoadingProjects] = useState(false)
   const [isAuthorizing, setIsAuthorizing] = useState(false)
   const [isLoadingAccounts, setIsLoadingAccounts] = useState(false)
+  const [isAlphaSpendReportUser, setIsAlphaSpendReportUser] = useState(false)
 
   const signal = useCancellation()
   const interval = useRef()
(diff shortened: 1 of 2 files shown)


src/pages/billing/List.js
```diff
@@ -271,6 +279,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
   useOnMount(() => {
     loadProjects()
     loadAccounts()
+    loadAlphaSpendReportMember()
```
cahrens, line 282: I put the loading of this value in this file so that we don't make the Ajax call every time a project is loaded (since this value does not change by project).

src/pages/billing/Project.js
```diff
@@ -190,7 +190,7 @@ const groupByBillingAccountStatus = (billingProject, workspaces) => {
   return _.mapValues(ws => new Set(ws), _.groupBy(group, workspaces))
 }
 
-const ProjectDetail = ({ billingProject, reloadBillingProject, billingAccounts, authorizeAndLoadAccounts }) => {
+const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProject, isAlphaSpendReportUser, reloadBillingProject }) => {
```
cahrens, line 193: Changed to alphabetical order.

src/pages/billing/Project.js
```diff
@@ -487,6 +488,7 @@ const ProjectDetail = ({ billingProject, reloadBillingProject, billingAccounts,
         style: { marginTop: '2rem', textTransform: 'none', padding: '0 1rem', height: '1.5rem' },
         tabStyle: { borderBottomWidth: 4 },
         value: tab,
+        key: `tabBarKey${isAlphaSpendReportUser}`, // The Spend report tab is only present for alpha users, so force recreation.
```
aherbst-broad, line 491: I am not grokking the comment here, can you elaborate on what this is recreating?

aherbst-broad (approved):
Looks straightforward.

src/pages/billing/Project.js
cahrens: Sure. The issue is that the array of tabs within `SimpleTabBar` cannot change on re-renders. Thus we have to force React to create a new instance of `SimpleTabBar` when the value of `isAlphaSpendReportUser` changes, since we only show the "Spend report" tab when that value is true (and we don't hear back from the isMember Ajax call until after the UI is initially rendered).

If that still isn't making sense, this [Slack thread](https://broadinstitute.slack.com/archives/C01EHNUM73R/p1644873326437099) might help.

dmohs (approved):
Looks great. No notes.

PR #2817, opened 2022-02-14
[DC-209] Improved logging in testing flake checker

lu-c:
Before:

Screen Shot 2022-02-14 at 6 38 02 PM

After:

Screen Shot 2022-02-14 at 6 29 45 PM


integration-tests/jest/jest-utils.js
```diff
@@ -36,6 +36,11 @@ const registerTest = ({ fn, name, timeout = defaultTimeout, targetEnvironments =
     timeout)
 }
 
+const processResults = results => {
+  const errors = _.filter(_.isError, results)
+  return [errors, _.countBy('stack', errors)]
+}
+
```
kyuksel: Could be

```suggestion
const processResults = _.flow(_.filter(_.isError), _.countBy('stack'))
```

and the call-site below can remain the same, thanks to Lodash FP currying :)

May also want to consider making the function name more specific (e.g. `groupByErrors`).

A comment around why you chose `'stack'` to `countBy` may be helpful.

integration-tests/jest/jest-utils.js
```diff
@@ -73,26 +78,30 @@ const flakeShaker = ({ fn, name }) => {
...
     const numErrors = _.size(errors)
 
     await cluster.idle()
     await cluster.close()
 
     if (!!numErrors) {
-      _.forEach(err => rawConsole.log(err.stack), errors)
-      throw new Error(`${numErrors} failures out of ${testRuns}. See above for specifics.`)
+      _.forEach(key => {
+        rawConsole.log(`\t\x1b[31m\x1b[1mError encountered ${errorCounts[key]} times`)
```
kyuksel: Looking at the screenshot on the PR description, it may be helpful to append to the log message `out of how many` as well:

```suggestion
        rawConsole.log(`\t\x1b[31m\x1b[1mError encountered ${errorCounts[key]} times (out of ${numErrors} in total)`)
```

kyuksel:
@lu-c Thank you for this improvement! It'll be very helpful for prioritizing errors 😊

cahrens:
Why are there so many unrelated commits on this branch?

integration-tests/jest/jest-utils.js
```diff
@@ -73,26 +78,30 @@ const flakeShaker = ({ fn, name }) => {
       const { taskFn, taskParams, runId } = data
       try {
         const result = await taskFn({ context, page, ...taskParams })
-        rawConsole.log(`Test number ${runId} passed`)
+        process.stdout.write('.')
```
petesantos: When I run the tests I am not seeing the dot printed out. If I add a newline, then I do see the dots get printed.
I tried changing this to ```process.stdout.write(`Reached test ${runId}`)``` to reduce log noise, but some tests console out other information which ends up overwriting this line, resulting in it not getting seen.

Is the intention to reduce the log noise?

I originally added this logging to let the user know the tests are running and give them a sense of the progress. I do like the idea of removing all of this noise and updating the same line with the progress.

We could could capture `stdout` to a variable and redirect `process.stdout.write` to a log while the tests are running and then restore `stdout` when the tests are done. I think this would significantly reduce command line noise.

samanehsan (approved):
Lgtm!

integration-tests/jest/jest-utils.js
lu-c: I think this would just return the countBy result - but the logging requires both - the filtered error list and the grouped counts - I think this should stick around! But I can definitely add the comment about why "stack"

integration-tests/jest/jest-utils.js
lu-c: Yeah, I was hoping to reduce the log noise, and just have the dots indicate something was happening - Let me see if I can get it a little more consistent. I was erasing them  after the cluster.task await finished, but that might be the wrong time?

lu-c: Found the problem, we also needed to be resetting the cursor for stdout! updated the messaging to give you a percentage competed now instead, cause I think that's probably better than adding dots too

lu-c:
> Why are there so many unrelated commits on this branch?

Im not sure, those commits were merged to develop too @.@

samanehsan:
> Why are there so many unrelated commits on this branch?
>
>Im not sure, those commits were merged to develop too @.@

I've seen this happen if I branch off of another branch and then the original branch gets squashed & merged into dev :octocat:

cahrens:
> I've seen this happen if I branch off of another branch and then the original branch gets squashed & merged into dev :octocat:

Perhaps just rebase the branch on dev?

kyuksel:
LGTM but I'll let @petesantos 👍🏼 when ready.

integration-tests/jest/jest-utils.js
```diff
@@ -69,30 +75,38 @@ const flakeShaker = ({ fn, name }) => {
...
       try {
         const result = await taskFn({ context, page, ...taskParams })
-        rawConsole.log(`Test number ${runId} passed`)
         return result
       } catch (e) {
         await page.screenshot({ path: `${screenshotDir}/${runId}.jpg`, fullPage: true })
-        rawConsole.log(`Test number ${runId} failed: ${e}`)
         return e
+      } finally {
+        process.stdout.clearLine()
```
petesantos: This looks better, but when running this against a test like `link-to-new-workspace` I still see the 'Running tests' line disappear.

I have an idea to correct this, let's pair tomorrow.

.gitignore
```diff
@@ -32,6 +32,7 @@ saturn-ui.iml
 
 /integration-tests/failures
 /integration-tests/screenshots
```
pshapiro4broad: Is `screenshots` still necessary now that it's under `results`? Also is `failures` still needed?

pshapiro4broad:
> Perhaps just rebase the branch on dev?

Is that necessary? The branch's commit history won't be preserved when the PR is merged because of Squash and merge. So I don't think there's a need to rebase before merging. As long as the diff in the PR looks good then the results should be fine. @cahrens @lu-c

cahrens:
> Is that necessary? The branch's commit history won't be preserved when the PR is merged because of Squash and merge. So I don't think there's a need to rebase before merging. As long as the diff in the PR looks good then the results should be fine. @cahrens @lu-c

Probably not necessary, but having had rebase/merge issues in the past I get worried when I see a PR like this. It also makes it much easier from a reviewer's point of view to understand the history. But thankfully yes, the expectation is that a squash will be done at merge time (please delete all the commit history that it will try to preserve on that commit).

.gitignore
lu-c: Oh i suppose not! Unless there are screenshots still lingering around. I dont know when /failures is written to tho, I don't think it's part of this script. Let me just remove the screenshots at least, and we'll go from there? :)

fboulnois (approved):
lgtm!

petesantos (approved):
LGTM, thanks for all of the work on this!

PR #2819, opened 2022-02-15
[DC-211] Fix back button on data browser details page

samanehsan:
Fix behavior of the back button on the data browser details page so that it navigates back to the list of datasets after visiting the preview page.

Previous behavior:
navBug

New behavior:
fixedNav

### DIFF
--- a/src/pages/library/DataBrowserPreview.js
+++ b/src/pages/library/DataBrowserPreview.js
@@ -136,7 +136,7 @@ const DataBrowserPreview = ({ id }) => {
         div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', justifyContent: 'space-between', width: '100%', lineHeight: '26px' } }, [
           h1([snapshot['dct:title']]),
           h(Link, {
-            href: Nav.getLink('library-details', { id: Nav.getCurrentRoute().params.id }),
+            onClick: Nav.history.goBack,
             'aria-label': 'Close',
             style: { marginTop: '1rem' }
           }, [


src/pages/library/DataBrowserPreview.js
```diff
@@ -136,7 +136,7 @@ const DataBrowserPreview = ({ id }) => {
         div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'top', justifyContent: 'space-between', width: '100%', lineHeight: '26px' } }, [
           h1([snapshot['dct:title']]),
           h(Link, {
-            href: Nav.getLink('library-details', { id: Nav.getCurrentRoute().params.id }),
+            onClick: Nav.history.goBack,
             'aria-label': 'Close',
```
cahrens, line 140: Why is the aria label for this Close?

samanehsan: I think @lu-c or @petesantos would have more context about that

cahrens:
@samanehsan how do I enable the "Browse & Explore" tab? I'm finding the before/after videos a little hard to understand (they are probably just too fast for me!), and I'd like to try this out.

src/pages/library/DataBrowserPreview.js
cahrens: I'm trying to understand if it should be "Back" instead of "Close" since the action is to go back in the history. But I'm also having trouble understanding in the video which button this is.

petesantos: This doesn't look right - we should change this to 'go back'.

petesantos (approved):
LGMT once the `aria-label` is corrected

src/pages/library/DataBrowserPreview.js
cahrens: Having watched this video a few more times, I think this is dismissing the Data Preview view and going back to the Data Set details. So I can see "Close" given that the icon is an X, and perhaps the fact that it is implemented as going back in the history isn't that important. What do you think @petesantos?

(It seems like this might be better as a Back (arrow) icon instead of X, but I am still new to how we manage/expose our internal navigation)

samanehsan:
> How do I enable the "Browse & Explore" tab? I'm finding the before/after videos a little hard to understand (they are probably just too fast for me!), and I'd like to try this out.

Sorry about that @cahrens! You can enable it by entering  `window.configOverridesStore.set({ isDataBrowserVisible: true });` into the javascript console

fboulnois (approved):
The change lgtm modulo the decision around the `aria-label`!

src/pages/library/DataBrowserPreview.js
samanehsan: A back arrow icon with the aria-label "go back" would make it consistent with the details page, so that makes sense to me. What do you think @petesantos?

samanehsan: After discussing this with my team, the consensus was that we can leave it as an "X" for now since users haven't been confused by this behavior, but we can revisit later if needed.

PR #2821, opened 2022-02-16
[WOR-69] Add mix panel events when tabs of SimpleTabBar are clicked

cahrens:
This is in support of the "Spend report" that will be under development/testing. But I thought it was good to make this more general and allow any user of SimpleTabBar to opt-in.

To see the event for the "Spend report" you need to log in as hermione (since that is the only account currently in the alpha testers group in dev). However, you can see the event when clicking between Workspaces and Users for any account that owns a billing project (for example, dumbledore).

image

Example of new event:

image

### DIFF
--- a/src/components/tabBars.js
+++ b/src/components/tabBars.js
@@ -134,13 +134,13 @@ export const SimpleTabBar = ({
   useLabelAssert('SimpleTabBar', props)
   const tabIds = _.map(useUniqueId, _.range(0, tabs.length))
   const panelRef = useRef()
-  const emitViewMetric = key => {
+  const maybeEmitViewMetric = key => {
     !!metricsPrefix && Ajax().Metrics.captureEvent(`${metricsPrefix}:view:${_.replace(/\s/g, '-', key)}`, metricsData)
   }
 
   // Determine the index of the selected tab, or choose the first one
   const selectedId = Math.max(0, _.findIndex(({ key }) => key === value, tabs))
-  useOnMount(() => { emitViewMetric(tabs[selectedId].key) })
+  useOnMount(() => { maybeEmitViewMetric(tabs[selectedId].key) })
 
   return h(Fragment, [
     h(HorizontalNavigation, {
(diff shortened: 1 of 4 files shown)


src/pages/billing/List.js
```diff
@@ -233,9 +233,7 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
     Utils.withBusyState(setIsLoadingProjects)
   )(async () => setBillingProjects(_.sortBy('projectName', await Ajax(signal).Billing.listProjects())))
 
-  const loadAlphaSpendReportMember = _.flow(
-    reportErrorAndRethrow('Error loading user group membership')
-  )(async () => {
+  const loadAlphaSpendReportMember = reportErrorAndRethrow('Error loading user group membership')(async () => {
```
cahrens, line 236: This is cleanup from a previous PR-- I didn't notice the lint warning about calling `flow` with a single argument.

src/components/tabBars.js
```diff
@@ -123,17 +124,23 @@ TabBar.propTypes = {
...
-  value, onChange, tabs, tabProps = {}, panelProps = {}, style = {}, tabStyle = {}, children, ...props
+  value, onChange, tabs, tabProps = {}, panelProps = {}, style = {}, tabStyle = {}, metricsPrefix = undefined, metricsData = {}, children, ...props
 }) => {
   useLabelAssert('SimpleTabBar', props)
-
   const tabIds = _.map(useUniqueId, _.range(0, tabs.length))
   const panelRef = useRef()
+  const emitViewMetric = key => {
+    !!metricsPrefix && Ajax().Metrics.captureEvent(`${metricsPrefix}:view:${_.replace(/\s/g, '-', key)}`, metricsData)
+  }
```
dmohs: Since the name of this isn't `maybeEmitViewMetric`, I might:
```suggestion
  const emitViewMetric = key => {
    Ajax().Metrics.captureEvent(`${metricsPrefix}:view:${_.replace(/\s/g, '-', key)}`, metricsData)
  }
```

dmohs (approved):

src/components/tabBars.js
cahrens: If I do that, I have to be careful not to call `emitViewMetric` if the prefix key has not been provided (which is what I'm using to "opt-in" to the metrics). I will change the name to `maybeEmitViewMetric`.

aherbst-broad (approved):

PR #2823, opened 2022-02-16
[SU-20] Fix cancel button in delete workspace data confirmation modal

nawatts:
Currently, the confirmation modal shown when deleting a variable from Workspace Data cannot be dismissed.

Screen Shot 2022-02-17 at 11 37 11 AM

This is because `setDeleteIndex` is called with the event that triggers the `onDismiss` callback instead of `undefined`.

### DIFF
--- a/src/components/data/LocalVariablesContent.js
+++ b/src/components/data/LocalVariablesContent.js
@@ -285,9 +285,7 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
       }
     }),
     deleteIndex !== undefined && h(Modal, {
-      onDismiss: () => {
-        setDeleteIndex(undefined)
-      },
+      onDismiss: () => { setDeleteIndex(undefined) },
       title: 'Are you sure you wish to delete this variable?',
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

cahrens:
This seems reasonable, but I can't find a workspace with Workspace Data to test it.

src/components/data/LocalVariablesContent.js
```diff
@@ -285,7 +285,9 @@ const LocalVariablesContent = ({ workspace, workspace: { workspace: { googleProj
       }
     }),
     deleteIndex !== undefined && h(Modal, {
-      onDismiss: setDeleteIndex,
+      onDismiss: () => {
+        setDeleteIndex(undefined)
+      },
```
dmohs: This is so simple that I would still try to constrain the vertical space used to express it, i.e.,
```suggestion
      onDismiss: () => { setDeleteIndex(undefined) },
```
Feel free to ignore if the linter complains.

dmohs (approved):

PR #2824, opened 2022-02-16
[IA-3002] Refactor ComputeModal around runtimeTypes (formerly sparkMode)

kyuksel:
Please see the [JIRA ticket](https://broadworkbench.atlassian.net/browse/IA-3002) for context.

The only user-facing change is highlighted on the screenshot below where `Spark master node` is renamed as `Spark single node`.

image

### Note for Reviewers
Further refactoring is out of scope. Undoubtedly `ComputeModal` could use a lot more refactoring but this PR is an attempt at not making perfect the enemy of good 😊 

### Testing
Flipped through the `ComputeModal` (shown above) menu options for when there is and there isn't an existing runtime (Dataproc as well as GCE), and compared results vs. their `dev` version.


src/components/ComputeModal.js
```diff
@@ -195,16 +215,22 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
   // State -- end
 
-  const isPersistentDisk = shouldUsePersistentDisk(sparkMode, currentRuntimeDetails, upgradeDiskSelected)
+  const isPersistentDisk = shouldUsePersistentDisk(runtimeType, currentRuntimeDetails, upgradeDiskSelected)
 
   const isCustomImage = selectedLeoImage === customMode
   const { version, updated, packages, requiresSpark, label: packageLabel } = _.find({ image: selectedLeoImage }, leoImages) || {}
 
-  const minRequiredMemory = sparkMode ? 7.5 : 3.75
+  const minRequiredMemory = isDataproc(runtimeType) ? 7.5 : 3.75
```
sjacobs146: Is there any information about why these two values were chosen for minRequiredMemory?

kyuksel: Good question. I've added the comment below:

```
  // The memory sizes below are the minimum required to launch Terra-supported GCP runtimes, based on experimentation.
```

Does that clarify?

src/components/ComputeModal.js
```diff
@@ -195,16 +214,22 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
-  const mainMachineType = _.find({ name: computeConfig.masterMachineType }, validMachineTypes)?.name || getDefaultMachineType(sparkMode)
+  const mainMachineType = _.find({ name: computeConfig.masterMachineType }, validMachineTypes)?.name || getDefaultMachineType(isDataproc(runtimeType))
   const machineTypeConstraints = { inclusion: { within: _.map('name', validMachineTypes), message: 'is not supported' } }
 
+  const isRuntimeRunning = currentRuntimeDetails?.status === 'Running'
+  const shouldDisplaySparkConsoleLink = isDataproc(runtimeType) && currentRuntimeDetails?.runtimeConfig?.componentGatewayEnabled
+  const canManageSparkConsole = shouldDisplaySparkConsoleLink && isRuntimeRunning
+
+  const canUpdateNumberOfWorkers = !currentRuntimeDetails || isRuntimeRunning
+
```
kyuksel, line 232: The functions right below didn't need to be functions so I converted them to constants. Moved them up so they are in the right section.

jdcanas (approved):
Looks good to me code-wise and from manual testing

src/components/ComputeModal.js
sjacobs146: Yes, thank you! Can't say how many times in my career that I've seen magic numbers in code and had no clue how someone arrived at them.

kyuksel: 💯

src/components/ComputeModal.js
```diff
@@ -393,8 +410,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
 
   const getDesiredEnvironmentConfig = () => {
     const { persistentDisk: existingPersistentDisk, runtime: existingRuntime } = getExistingEnvironmentConfig()
-    const cloudService = sparkMode ? cloudServices.DATAPROC : cloudServices.GCE
-    const desiredNumberOfWorkers = sparkMode === 'cluster' ? computeConfig.numberOfWorkers : 0
+    const cloudService = isDataproc(runtimeType) ? cloudServices.DATAPROC : cloudServices.GCE
```
cahrens, line 413: You could have gone a different way with this refactoring and have had the different instances of runtimeType implement properties to provide this information so that you don't need conditionals. In this case, they would each have a "getCloudService" property and a "isDataproc" property, and you would not need "isGce", "isDataproc", and "isDataprocCluster". This is how we went with `tools` in `notebook-utils`, and it is more extensible as additional properties of types get added.

I'm not really advocating that you rework this PR, but just mentioning it as an alternative approach, especially if the properties of a runtimeType continue to grow.

src/components/ComputeModal.js
```diff
@@ -664,25 +682,25 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
-      setSparkMode(newSparkMode)
+      setRuntimeType(newRuntimeType)
       setComputeConfig({
         selectedPersistentDiskSize: currentPersistentDiskDetails?.size || defaultGcePersistentDiskSize,
         masterMachineType: runtimeConfig?.masterMachineType || runtimeConfig?.machineType,
-        masterDiskSize: runtimeConfig?.masterDiskSize || runtimeConfig?.diskSize ||
-          (isDataproc ? defaultDataprocMasterDiskSize : defaultGceBootDiskSize),
+        masterDiskSize: runtimeConfig?.masterDiskSize || runtimeConfig?.diskSize || isDataproc(newRuntimeType) ?
+          defaultDataprocMasterDiskSize :
+          defaultGceBootDiskSize,
```
cahrens, line 697: Nit: I think this would be easier to read if the ternary expression were all one line.

src/components/ComputeModal.js
```diff
@@ -951,22 +970,22 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
-                      setSparkMode(value)
-                      updateComputeConfig('componentGatewayEnabled', !!value)
+                      setRuntimeType(value)
+                      updateComputeConfig('componentGatewayEnabled', isDataproc(value))
                     },
                     options: [
-                      { value: false, label: 'Standard VM', isDisabled: requiresSpark },
-                      { value: 'master', label: 'Spark master node', isDisabled: requiresGCE() },
-                      { value: 'cluster', label: 'Spark cluster', isDisabled: requiresGCE() }
+                      { value: runtimeTypes.gceVm, label: runtimeTypes.gceVm, isDisabled: requiresSpark },
```
cahrens: Nit: I believe you don't need to specify label if it is the same as value.

src/components/ComputeModal.js
```diff
@@ -1372,11 +1394,15 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       value: selectedLeoImage,
       onChange: ({ value }) => {
         const requiresSpark = _.find({ image: value }, leoImages)?.requiresSpark
-        const newSparkMode = requiresSpark ? (sparkMode || 'master') : false
+        const newRuntimeType = Utils.cond(
+          [!!requiresSpark && isDataproc(runtimeType), () => runtimeType],
+          [!!requiresSpark && !isDataproc(runtimeType), () => runtimeTypes.dataprocSingleNode],
```
cahrens, line 1399: Can't the above 2 lines be simplified to just return runtimeType (without checking isDataproc)?

cahrens (approved):
A couple nits, but can choose whether to address them. I did not actually run the code, as the refactoring looked pretty straightforward and I am not familiar with this area.

src/components/ComputeModal.js
kyuksel: Thank you for the thoughts. I started with a scheme similar to what you described but I didn't like the verbosity it resulted in at call sites. Also, having `isDataproc` as a property did not seem to be scalable as we add more cloud services like `Azure VM` where we'd also need `isAzureVm`, etc.

src/components/ComputeModal.js
kyuksel: These line breaks are added by the reformatter  🤷🏼

src/components/ComputeModal.js
kyuksel: Thanks!

PR #2825, opened 2022-02-16
[SUP-565] Sort list of packages installed in cloud environments

nawatts:
Currently, the list of packages shown in the "What’s installed on this environment?" panel when creating a cloud environment are not sorted. Some environments do happen to have a sorted list of packages, but an example of an unordered list can be seen with the Python packages for the Legacy GATK environment. This makes it difficult to scan the list and determine whether or not a package is available.

This sorts the list of packages installed in the environment by package name.

### DIFF
--- a/src/components/ImageDepViewer.js
+++ b/src/components/ImageDepViewer.js
@@ -56,12 +56,16 @@ export const ImageDepViewer = ({ packageLink }) => {
           ])
         ]),
         tbody(
-          _.map(({ name, version }) => {
-            return tr({ key: name }, [
-              td({ style: { paddingRight: '1rem' } }, [name]),
-              td([version])
-            ])
-          }, _.filter({ tool: selectedTool }, packages))
+          _.flow(
+            _.filter({ tool: selectedTool }),
+            _.sortBy(({ name }) => _.toLower(name)),
+            _.map(({ name, version }) => {
+              return tr({ key: name }, [
+                td({ style: { paddingRight: '1rem' } }, [name]),
+                td([version])
+              ])
+            })
+          )(packages)
         )
       ])
     ])


src/components/ImageDepViewer.js
```diff
@@ -56,12 +56,16 @@ export const ImageDepViewer = ({ packageLink }) => {
...
         tbody(
-          _.map(({ name, version }) => {
-            return tr({ key: name }, [
-              td({ style: { paddingRight: '1rem' } }, [name]),
-              td([version])
-            ])
-          }, _.filter({ tool: selectedTool }, packages))
+          _.flow(
+            _.filter({ tool: selectedTool }),
+            _.sortBy(({ name }) => _.toLower(name)),
```
kyuksel, line 61: Have we gotten the UX team's approval on sorting such that packages with uppercase letters are interleaved with packages with lowercase letters? We don't seem to be doing that in other places in the codebase so wanted to double-check. I don't feel strongly either way but it'd be nice to be consistent :)

src/components/ImageDepViewer.js
```diff
@@ -56,12 +56,16 @@ export const ImageDepViewer = ({ packageLink }) => {
           ])
         ]),
         tbody(
-          _.map(({ name, version }) => {
-            return tr({ key: name }, [
-              td({ style: { paddingRight: '1rem' } }, [name]),
-              td([version])
-            ])
-          }, _.filter({ tool: selectedTool }, packages))
```
kyuksel, line 64: Thanks for making the code below `flow`y. It reads more easily despite the added function :)

kyuksel (commented):
Looks good. Just a UX question inline.

Welcome to Terra UI  🎉

src/components/ImageDepViewer.js
nawatts: I had not asked them. Just posted about it in Slack though.

nawatts: Approved: https://broadinstitute.slack.com/archives/C34GAKQ5R/p1645111855628929?thread_ts=1645046899.125539&cid=C34GAKQ5R

kyuksel (approved):
Thanks for checking with the UX team.

MatthewBemis (approved):

PR #2828, opened 2022-02-18
[SU-11] Limit the height of the confirmation modal for deleting rows from data tables

nawatts:
Currently, when deleting rows from a data table, the confirmation modal displays the entity IDs for all rows being deleted. The modal expands to fit the full list, so when deleting hundreds or thousands of rows, users have to scroll far down to find the Delete button.

This limits the height of the list of entity IDs, so that the list can be scrolled within the modal body without moving the Delete button out of the viewport. This is consistent with how the Share Workspace modal works with a long list of collaborators.

Before:
Screen Shot 2022-02-18 at 1 25 00 PM

After:
Screen Shot 2022-02-18 at 1 12 23 PM

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -206,7 +206,9 @@ export const EntityDeleter = ({ onDismiss, onSuccess, namespace, name, selectedE
       icon('warning-standard', { size: 36, style: { flex: 'none', marginRight: '0.5rem' } }),
       'In order to delete the selected data entries, the following entries that reference them must also be deleted.'
     ]),
-    div({ style: { maxHeight: 550, overflowY: 'auto', margin: '0 -1.25rem' } },
+    // Size the scroll container to cut off the last row to hint that there's more content to be scrolled into view
+    // Row height calculation is font size * line height + padding + border
+    div({ style: { maxHeight: 'calc((1em * 1.15 + 1.2rem + 1px) * 10.5)', overflowY: 'auto', margin: '0 -1.25rem' } },
       _.map(([i, entity]) => div({
         style: {
           borderTop: (i === 0 && runningSubmissionsCount === 0) ? undefined : Style.standardLine,
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -206,13 +206,15 @@ export const EntityDeleter = ({ onDismiss, onSuccess, namespace, name, selectedE
       icon('warning-standard', { size: 36, style: { flex: 'none', marginRight: '0.5rem' } }),
       'In order to delete the selected data entries, the following entries that reference them must also be deleted.'
     ]),
-    ..._.map(([i, entity]) => div({
-      style: {
-        borderTop: (i === 0 && runningSubmissionsCount === 0) ? undefined : Style.standardLine,
-        padding: '0.6rem 1.25rem', margin: '0 -1.25rem'
-      }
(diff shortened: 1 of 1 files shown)


src/components/data/data-utils.js
```diff
@@ -206,13 +206,15 @@ export const EntityDeleter = ({ onDismiss, onSuccess, namespace, name, selectedE
...
       'In order to delete the selected data entries, the following entries that reference them must also be deleted.'
     ]),
-    ..._.map(([i, entity]) => div({
-      style: {
-        borderTop: (i === 0 && runningSubmissionsCount === 0) ? undefined : Style.standardLine,
-        padding: '0.6rem 1.25rem', margin: '0 -1.25rem'
-      }
-    }, moreToDelete ? `${entity.entityName} (${entity.entityType})` : entity),
-    Utils.toIndexPairs(moreToDelete ? additionalDeletions : selectedKeys)),
+    div({ style: { maxHeight: 550, overflowY: 'auto', margin: '0 -1.25rem' } },
```
dmohs: Negative margins often produce surprising behavior. I wonder if there are any alternatives to get the look you want.

src/components/data/data-utils.js
```diff
@@ -206,13 +206,15 @@ export const EntityDeleter = ({ onDismiss, onSuccess, namespace, name, selectedE
...
+      _.map(([i, entity]) => div({
+        style: {
+          borderTop: (i === 0 && runningSubmissionsCount === 0) ? undefined : Style.standardLine,
+          padding: '0.6rem 1.25rem'
```
dmohs, line 215: I wonder if you could tweak this value so the yellow bar vertically splits the row and cuts off the text? This is a common affordance that demonstrates there is content that must be scrolled into view. This can be tricky to implement, so I will leave it to your discretion.

dmohs (approved):
Address what you can, but nothing blocking. Grab pairing time with me if you'd like help investigating ideas.

src/components/data/data-utils.js
nawatts: This was to made the row dividers span the full width of the modal, overriding the padding on the modal content. I kept it to preserve the existing style, but I don't know if there's a strong preference for one or the other.

Screen Shot 2022-02-18 at 1 12 23 PM copy
vs 
Screen Shot 2022-02-18 at 1 57 47 PM copy

src/components/data/data-utils.js
nawatts: Good point! Will attempt.

src/components/data/data-utils.js
nawatts: Ah, a more compelling argument for the negative margin than full width dividers: it puts the scroll bar flush with the edge of the modal vs inset from it.

Without negative margin:
Screen Shot 2022-02-18 at 2 06 14 PM

Passing `styles: { modal: { padding: '1.5rem 0' } }` to the Modal component almost removes the need for the negative margin, but that approach changes the alignment of the modal title and footer buttons. Modal exposes enough styling options to restore the footer but not the title.

src/components/data/data-utils.js
nawatts: Adjusted the scroll container height to do this in [b8f76a5](https://github.com/DataBiosphere/terra-ui/pull/2828/commits/b8f76a57cf9d2db39b8ff6917b6cbe4d74c0c2ef).

MatthewBemis (approved):

PR #2829, opened 2022-02-18
[SUP-560] Adjust width of billing account and workspace columns in cloud environments table

nawatts:
Currently, the billing project and workspace columns on the cloud environments page use SimpleFlexTable's default flex basis of 0. Thus, they can be collapsed down to nothing in smaller windows. This adds a flex basis to those columns so that they get some space allocated.

https://github.com/DataBiosphere/terra-ui/blob/d62d5b0ffd1892a5d41ca40dfe06873248cc7b74/src/components/table.js#L137

Before
Screen Shot 2022-02-18 at 2 48 32 PM

After
Screen Shot 2022-02-18 at 2 48 45 PM

### DIFF
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -344,6 +344,7 @@ const Environments = () => {
         rowCount: filteredCloudEnvironments.length,
         columns: [
           {
+            size: { basis: 250 },
             field: 'project',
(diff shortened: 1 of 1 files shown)


kyuksel (approved):
Looks great. Thanks for this improvement!

MatthewBemis (approved):

PR #2830, opened 2022-02-18
[DC-235] Fixed bug with filtering on the workspace page

lu-c:
The problem was the unique identifier field did not exist for all data types, added parameter to allow this to work for everyone using the common code!

No screenshots this time, cause it just shows the filters working the way you'd think they would. :P Before, the workspace filtering only ever showed a single value - because it was uniqueing the objects on a field that doesnt exist (and uniqueBy all null values meant we only ever got one result!)

### DIFF
--- a/src/pages/library/DataBrowser.js
+++ b/src/pages/library/DataBrowser.js
@@ -247,7 +247,8 @@ const Browser = () => {
       customSort: sort,
       searchType: 'Datasets',
       titleField: 'dct:title',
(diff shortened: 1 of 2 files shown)


kyuksel (approved):
Not familiar with the functionality but the code change LGTM.

petesantos (approved):

samanehsan (approved):
Lgtm!

PR #2831, opened 2022-02-18
[SUP-525] Remove Publish button from workspace menu

MatthewBemis:
This button has been marked as "Coming Soon" for 4 years, but it's not actually coming at all so we're getting rid of the button to avoid causing any more confusion.

Before
Screen Shot 2022-02-18 at 4 33 22 PM

After
Screen Shot 2022-02-18 at 4 33 35 PM

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -210,15 +210,6 @@ export const transparentSpinnerOverlay = makeBaseSpinner({ innerStyles: { backgr
 
 export const topSpinnerOverlay = makeBaseSpinner({ innerStyles: { marginTop: 150 } })
 
-export const comingSoon = span({
-  style: {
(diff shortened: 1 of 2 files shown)


nawatts (approved):

cahrens:
I suppose "soon" is relative. Nothing else is using the `comingSoon` element, so I would be tempted to delete it.

cahrens (approved):

MatthewBemis:
@cahrens I just pushed a commit removing the element. Always happy to delete more code :)

cahrens:
> @cahrens I just pushed a commit removing the element. Always happy to delete more code :)

Thanks @MatthewBemis ! I also find joy in deleting code.

PR #2832, opened 2022-02-22
[SU-10] Suggest set member attributes for workflow inputs

nawatts:
Currently, if a workflow is configured to run on a set entity type, the drop down menu for suggested workflow inputs contains the member list (for example, `this.samples` for a `sample_set` table). However, it does not include the attributes of those members (for example, `this.samples.sample_id`). Thus, to run a workflow using an attribute of members of a set as input, users must type in the expression for the input. This is an error prone process, since a spelling mistake in the input expression can cause the workflow to fail.

This adds attributes of the member entity type to the input suggestions menu when a set entity type is selected.

Screen Shot 2022-02-22 at 3 41 01 PM

Screen Shot 2022-02-22 at 3 40 42 PM


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -986,9 +986,30 @@ const WorkflowView = _.flow(
     const selectedTableName = modifiedConfig.dataReferenceName ? modifiedConfig.rootEntityType : undefined
     const selectionMetadata = selectedTableName ? selectedSnapshotEntityMetadata : entityMetadata
     const attributeNames = _.get([modifiedConfig.rootEntityType, 'attributeNames'], selectionMetadata) || []
+
+    const getSuggestionsForAttributesOfSetMembers = () => {
+      if (!modifiedConfig.rootEntityType.endsWith('_set')) {
+        return []
+      }
+      const memberEntityType = modifiedConfig.rootEntityType.slice(0, modifiedConfig.rootEntityType.length - 4)
```
dmohs: Is this correct? If so, it perhaps makes the intent more clear.
```suggestion
      const memberEntityType = modifiedConfig.rootEntityType.slice(0, modifiedConfig.rootEntityType.length - '_set'.length)
```

dmohs (approved):

src/pages/workspaces/workspace/workflows/WorkflowView.js
nawatts: This was based on Terra's behavior when uploading a membership TSV. I don't know if it's formally documented anywhere, but it looks like there's some magic based on the `_set` suffix. I can upload a TSV with column headers `entity:test_set attribute` and the resulting table gets a `tests` column even if there is not a `test` table in the workspace. Thus, an additional check after this to see if the member entity type derived here actually exists in the workspace.

MatthewBemis (approved):

nawatts:
Rebased to pull in test changes from #2835.

nawatts:
@MatthewBemis pointed out a more robust way to identify sets of entities using the selected input data vs relying on the table name. Switched to that method in 774b903accc614630079d6eb2410787b543cda12.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -297,6 +297,32 @@ const findPossibleSets = listOfExistingEntities => {
   }, [], listOfExistingEntities)
 }
 
+const getWorkflowInputSuggestionsForAttributesOfSetMembers = (selectedEntities, entityMetadata) => {
```
kyuksel: This method does _a lot_, which may make it a good candidate to write a unit test for?

dmohs (approved):

src/pages/workspaces/workspace/workflows/WorkflowView.js
nawatts: Added some tests in [49e7346](https://github.com/DataBiosphere/terra-ui/pull/2832/commits/49e734642bf6d6b8373d2bffea7fcdfdf88629f3). I had to move the function to another file to get Jest to run. I think it couldn't handle the circular import libs/auth => components/common => components/icons => libs/auth.

kyuksel: The tests look great ❤️ 

How would you feel about moving `getWorkflowInputSuggestionsForAttributesOfSetMembers` inside `src/libs/workflow-utils.js` instead of creating a dedicated file for it? Accordingly, the new test file could be moved to `src/libs/_tests/workflow-utils.test.js`.

nawatts: Moved in [47eee8c](https://github.com/DataBiosphere/terra-ui/pull/2832/commits/47eee8ce209d0cdf2739726e8961cbb0231440e9).

MatthewBemis (approved):
Nice!

kyuksel (approved):
Looks great!   :shipit:

Issue #426, opened 2018-07-24
The first column in the submissions history tab is super long


bradtaylor:
Consequently the dropdown is way far away from the workflow name. This should be shorter. The invision designs had a more balanced look:
https://projects.invisionapp.com/d/main#/console/13746874/304975026/preview

Issue #414, opened 2018-07-23
Cluster widget should indicate when you're not allowed to access a project's runtime


bradtaylor:
For billing-projects in which a user has no workspaces wherein they have "can compute" permission, they are prevented by Leo from creating clusters or running terminals (or, ought to be. Confirm). For such cases the runtime widget should display it's current greyed out look:
<img width="275" alt="screen shot 2018-07-24 at 12 44 39 am" src="https://user-images.githubusercontent.com/3492466/43117578-84e3d342-8edb-11e8-92e4-4afa83ea6ef8.png">

However, the widget dropdown should be disabled! There should also be a hover message that reads "You do not have to run notebooks on any workspaces in {billing project name}"

PR #2834, opened 2022-02-23
[IA-2962][IA-3087] Add metrics for beta and beta toggle

jdcanas:
https://broadworkbench.atlassian.net/browse/IA-3087
https://broadworkbench.atlassian.net/browse/IA-2962

This ticket adds metrics for the beta, and adds a hidden button to toggle between beta/current functionality.

To test the toggle, you will need to set `hidden: false` in `Notebook.js` (should be only occurrence of `hidden` in that file).

There are a few places where metrics use both `source` and `application` with the same value. This is because source seems to be a special category for visual display/aggregation in mixpanel.


src/components/ContextBar.js
```diff
@@ -133,7 +134,8 @@ export const ContextBar = ({
           tooltipSide: 'left',
           disabled: !isTerminalEnabled,
           href: terminalLaunchLink,
-          onClick: window.location.hash === terminalLaunchLink && currentRuntime?.status === 'Stopped' ? () => startCurrentRuntime() : undefined,
+          onClick: window.location.hash === terminalLaunchLink && currentRuntime?.status === 'Stopped' ? () => startCurrentRuntime() : Ajax().Metrics.captureEvent(Events.analysisLaunch,
+            { origin: 'contextBar', source: 'terminal', application: 'terminal', workspaceName, namespace }),
```
kyuksel: Could you run the formatter? It should reformat the two lines above as below:

```suggestion
          onClick: window.location.hash === terminalLaunchLink && currentRuntime?.status === 'Stopped' ?
            () => startCurrentRuntime() :
            Ajax().Metrics.captureEvent(Events.analysisLaunch,
              { origin: 'contextBar', source: 'terminal', application: 'terminal', workspaceName, namespace }),
```

kyuksel: When I clicked on the terminal icon on the context bar while having a stopped runtime, I noticed that the new tab that opens up has the legacy layout (currently in `prod`):

image

and if the user uses the RuntimeManager on that tab to resume their runtime, we would end up not collecting the metric even though the user had attempted to launch terminal on the new tab. Am I missing something?

src/components/runtime-common.js
```diff
@@ -180,4 +180,4 @@ export const GalaxyLaunchButton = ({ app, key = app.status, onClick, ...props })
 export const appLauncherTabName = 'workspace-application-launch'
 export const analysisLauncherTabName = 'workspace-analysis-launch'
 export const analysisTabName = 'workspace-analyses'
-export const contextBarTabs = [appLauncherTabName, analysisLauncherTabName, analysisTabName]
+export const contextBarTabs = [appLauncherTabName, analysisLauncherTabName, 'analyses']
```
kyuksel, line 183: Why was this change (i.e. not using `analysisTabName`) necessary?

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -330,7 +330,10 @@ export const CloudEnvironmentModal = ({
           onClick: () => {
             ((toolLabel === tools.Jupyter.label || toolLabel === tools.RStudio.label) && currentRuntime?.status === 'Stopped' ?
               () => startApp(toolLabel) :
-              () => {})()
+              () => {
+                Ajax().Metrics.captureEvent(Events.analysisLaunch,
+                  { origin: 'contextBar', source: toolLabel, application: toolLabel, workspaceName, namespace })
+              })()
             onDismiss()
           }
```
kyuksel: Could this be simplified as

```javascript
          onClick: () => {
            toolLabel === tools.Jupyter.label || toolLabel === tools.RStudio.label && currentRuntime?.status === 'Stopped' ?
              startApp(toolLabel) :
              Ajax().Metrics.captureEvent(Events.analysisLaunch,
                { origin: 'contextBar', source: toolLabel, application: toolLabel, workspaceName, namespace })
            onDismiss()
          }
```

?

kyuksel:
> There are a few places where metrics use both source and application with the same value. This is because source seems to be a special category for visual display/aggregation in mixpanel.

Could you elaborate a bit more why we need to specify both `source` and `application`? I don't see that in other parts of the codebase.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -407,6 +409,28 @@ const Notebooks = _.flow(
     notebooks && h(PageBox, [
       div({ style: { display: 'flex', alignItems: 'center', justifyContent: 'space-between', marginBottom: '1rem' } }, [
         div({ style: { ...Style.elements.sectionHeader, textTransform: 'uppercase' } }, ['Notebooks']),
+        div({ hidden: true }, //Will be released with this ticket https://broadworkbench.atlassian.net/browse/IA-3225
+          [
```
kyuksel: Can we single-line these two like we do in other places and adjust the end of the block accordingly?

```javascript
// Will be released with ticket https://broadworkbench.atlassian.net/browse/IA-3225
div({ hidden: true }, [
```

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
kyuksel: Also the [same comment as above](https://github.com/DataBiosphere/terra-ui/pull/2834/files#r814352929)

kyuksel (commented):
Looks good overall.
Some comments and questions inline.
Also asked a [UX question on Slack](https://broadinstitute.slack.com/archives/GB1VAM14J/p1645788038913329).

src/components/ContextBar.js
jdcanas: hmm... I just refactored this whole block so its more clear and avoids the situation you describe. I feel like we have a tendency to over-use ternary statements and avoid the dreaded if, but oftentimes if makes it a lot clearer 😂

src/components/runtime-common.js
jdcanas: So there's currently a bug where, if the analysis tab is the active tab, it wont have the green underline like in the picture below. The tab name passed into the tabbar needs to match the wording in the tab to accomplish this unfortunately (which is different from the nav keyword in this case, hence the change)

jdcanas: image

jdcanas:
> > There are a few places where metrics use both source and application with the same value. This is because source seems to be a special category for visual display/aggregation in mixpanel.
> 
> Could you elaborate a bit more why we need to specify both `source` and `application`? I don't see that in other parts of the codebase.

I can't tell if its a property of mixpanel or a tracked field of a few metrics 
image

I put it in to see to be safe rather than sorry when we go to display these metrics

src/components/ContextBar.js
kyuksel: Sure, when `if` is more readable and gets the job done, no need to be dogmatic about it.

kyuksel:
`Notebooks.js` lines #414 and #426 need to be broken up. Could you run the IntelliJ reformatter?

image

kyuksel:
`Sort By` is misaligned

image

Contrast that to the `dev` version:

image

kyuksel:
Can we horizontally align the newly added rectangle below

image

with the stack of notebooks, as in the UX mock below?

Screen Shot 2022-02-25 at 11 01 57 AM

src/components/runtime-common.js
kyuksel: Hmm…I think I'm observing the opposite of what you wrote; i.e. I _do_ see the green underline with the original code

```javascript
export const contextBarTabs = [appLauncherTabName, analysisLauncherTabName, analysisTabName]
```

so I don't understand why we'd change it. What am I missing?

kyuksel:
> I put it in to see to be safe rather than sorry when we go to display these metrics

I don't know MixPanel enough to make a good suggestion so I [posted on the MixPanel channel](https://broadinstitute.slack.com/archives/C015QCR0SKY/p1646171143500489).

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -386,18 +386,29 @@ const Analyses = _.flow(
...
+              Ajax().Metrics.captureEvent(Events.analysisDisableBeta, {
+                workspaceName: wsName,
+                workspaceNamespace: namespace
+              })
+              window.configOverridesStore.set({ isAnalysisTabVisible: false })
+              Nav.goToPath('workspace-notebooks', { namespace, name: wsName })
+            },
+            tooltip: 'Exit the analysis tab beta feature'
+          }, ['Revert layout'])
+        ]),
```
kyuksel: Consider reformatting:

```suggestion
        div({ style: { flex: 1.5 } }),
        div({
          style: {
            display: 'flex', flexDirection: 'column', marginLeft: '.5rem', backgroundColor: colors.secondary(0.1), padding: '1rem',
            border: `1px solid ${colors.accent()}`, borderRadius: 3
          }, hidden: false
        }, [
          // Will be released with this ticket https://broadworkbench.atlassian.net/browse/IA-3225
          div({ style: { maxWidth: 300 } }, [
            span(['What did you think? We\'d love to hear your thoughts. ']),
            // TODO href when user ed makes documentation, see: https://broadworkbench.atlassian.net/browse/IA-3085
            h(Link, { href: '', ...Utils.newTabLinkProps }, ['Submit feedback'])
          ]),
          h(ButtonPrimary, {
            style: { marginTop: '.5rem', maxWidth: 200, alignSelf: 'left' },
            onClick: () => {
              Ajax().Metrics.captureEvent(Events.analysisDisableBeta, { workspaceName: wsName, workspaceNamespace: namespace })
              window.configOverridesStore.set({ isAnalysisTabVisible: false })
              Nav.goToPath('workspace-notebooks', { namespace, name: wsName })
            },
            tooltip: 'Exit the analysis tab beta feature'
          }, ['Revert layout'])
        ]),
```

kyuksel (commented):
@jdcanas I've resolved/answered/added comments (inline as well as out of line) as applicable.

jdcanas:
> Can we horizontally align the newly added rectangle below
> 
> image
> 
> with the stack of notebooks, as in the UX mock below?
> 
> Screen Shot 2022-02-25 at 11 01 57 AM

Not really very easily without ripping apart the whole layout in that page since the top row isn't connected to the flex boxes below. I'd prefer not to heavily modify the layout for this component since it will be going away in the near future. I already discussed this limitation with Tiffany.

kyuksel (approved):
@jdcanas It seemed like there was an unrelated test failure and I was going to rerun it. Figured I'd take the liberty to update your PR branch with `dev` while at it. Hope that was okay.

Qi77Qi (approved):
`Learn more` and `Submit feedback` from the prompts dont' seem to work on the PR site. Other than that lgtm

PR #2840, opened 2022-02-25
[SUP-501] Ensure that a submission's workflows table is visible

nawatts:
Currently, when viewing a submission in the Job History tab, the table of workflows in the submission can be cut off if the window is not tall enough. This is because the workflows table is auto sized based on the amount of space available after the other submission information. If the window is just tall enough to fit the submission information, there is no space for the workflows table.

This reserves enough space for a table with 5.5 rows, or one row per workflow if there are fewer than 6 workflows.

Screen Shot 2022-02-25 at 4 11 57 PM

### DIFF
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -242,6 +242,7 @@ const SubmissionDetails = _.flow(
           })
         ])
       ]),
+      // 48px is based on the default row height of FlexTable
       div({ style: { flex: `1 0 calc(${1 + Math.min(filteredWorkflows.length, 5.5)} * 48px)` } }, [
         h(AutoSizer, [({ width, height }) => h(FlexTable, {
           'aria-label': 'submission details',
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -243,7 +243,7 @@ const SubmissionDetails = _.flow(
         ])
       ]),
       // 48px is based on the default row height of FlexTable
-      div({ style: { flex: `1 0 calc(${1 + Math.min(filteredWorkflows.length, 5.5)} * 48px)` } }, [
+      div({ style: { flex: `1 0 ${(1 + Math.min(filteredWorkflows.length, 5.5)) * 48}px` } }, [
         h(AutoSizer, [({ width, height }) => h(FlexTable, {
           'aria-label': 'submission details',
           width, height, sort,
--- a/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
+++ b/src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
@@ -242,7 +242,7 @@ const SubmissionDetails = _.flow(
           })
         ])
       ]),
-      div({ style: { flex: 1 } }, [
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -242,7 +242,7 @@ const SubmissionDetails = _.flow(
           })
         ])
       ]),
-      div({ style: { flex: 1 } }, [
+      div({ style: { flex: `1 0 calc(${1 + Math.min(filteredWorkflows.length, 5.5)} * 48px)` } }, [
```
cahrens: This seems to be working pretty well, but we prefer using rems instead of pixels for sizing things based on text size. rems are linked to the root font size, which is 16 in Terra. So 3 rems should be equivalent to 48 px. The advantage to using rems is that it will scale appropriately as the user zooms in and out.

Also, we prefer to use the lodash utility functions, so please use `_.min` instead of `Math.min`.

nawatts: I should have put a comment here, but the 48px is based on the row height of the FlexTable component.
https://github.com/DataBiosphere/terra-ui/blob/85d11a9972f2f1fb09f21a0534204c2983a6da98/src/components/table.js#L210-L214

cahrens: Ah, I see. A comment in the code would be useful, and then you can just do the multiplication in JavaScript and avoid the use of css `calc`.

cahrens (approved):

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
kyuksel: (Sorry to be late to the party 😛)

> 48px is based on the row height of the FlexTable component

I wonder if we should pull out that `48` onto a dedicated `const` and have `FlexTable` and the newly edited code reference it. The concern would be that if `FlexTable` spec changes, it'd be tough to know to also change the other place.

nawatts: A few options. We could...
* Export a constant with the default row height of FlexTable and use it here. However, that could be overridden by setting the row height on the submissions table here.
* Use a row height constant in this file for this and the submissions table row height. However, if the default row height of FlexTable changes, this table would be non-standard.

In any case, the stakes are low. If the FlexTable default changes, the consequence would be that this table doesn't show exactly 5.5 rows in browser windows less than a certain height.

kyuksel: To clarify, I was suggesting
- `export`ing a `const` equal to `48` in `table.js`, 
- using the `const` in `FlexTable` definition in `table.js`
- replacing `48` in `SubmissionDetails.js` with the `const`

If someone consciously wants to override the default, they can, which is the current situation and is fine. I just wanted to avoid having to change multiple places when a value changes.

kyuksel: To clarify, I was suggesting
- `export`ing a `const` equal to `48` in `table.js`, 
- using the `const` in `FlexTable` definition in `table.js`
- replacing `48` in `SubmissionDetails.js` with the `const`

If someone consciously wants to override the default, they can, which is the current situation and is fine. I just wanted to avoid having to change multiple places when a value changes.

nawatts: Done in #2845.

PR #2843, opened 2022-03-01
[SUP-611] Add more information to delete workspace confirmation

nawatts:
To clarify the impact of deleting a workspace, this adds additional information (the size of the workspace bucket and the list of users with access to the workspace) to the delete workspace confirmation prompt.

## Before
Screen Shot 2022-03-01 at 9 38 39 AM

## After
Screen Shot 2022-03-01 at 9 46 51 AM

### DIFF
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -343,3 +343,18 @@ export const sha256 = async message => {
     _.join('')
   )(new Uint8Array(hashBuffer))
 }
+
+export const formatBytes = bytes => {
+  if (bytes < 2 ** 10) {
+    return `${bytes} B`
+  }
+
+  const [prefix, divisor] = [
+    ['P', 2 ** 50],
+    ['T', 2 ** 40],
+    ['G', 2 ** 30],
+    ['M', 2 ** 20],
+    ['K', 2 ** 10]
+  ].find(([p, d]) => bytes >= d)
+  return `${(bytes / divisor).toPrecision(3)} ${prefix}iB`
+}
--- a/src/pages/workspaces/workspace/DeleteWorkspaceModal.js
+++ b/src/pages/workspaces/workspace/DeleteWorkspaceModal.js
@@ -1,4 +1,5 @@
 import _ from 'lodash/fp'
+import pluralize from 'pluralize'
 import { useState } from 'react'
 import { div, h, label, p, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, Link, spinnerOverlay } from 'src/components/common'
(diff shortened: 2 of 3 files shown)


MatthewBemis (approved):

src/libs/utils.js
```diff
@@ -343,3 +343,18 @@ export const sha256 = async message => {
...
+
+  const [prefix, divisor] = [
+    ['P', 2 ** 50],
+    ['T', 2 ** 40],
+    ['G', 2 ** 30],
+    ['M', 2 ** 20],
+    ['K', 2 ** 10]
+  ].find(([p, d]) => bytes >= d)
+  return `${(bytes / divisor).toPrecision(3)} ${prefix}iB`
+}
```
kyuksel, line 360: I wonder if we should use an existing library (e.g. [filesize.js](https://filesizejs.com/)) instead of rolling our own.

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -92,14 +98,23 @@ const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucke
...
+      '.'
     ]),
     hasApps() && div({ style: { marginTop: '1rem' } }, [
       p(['Deleting it will also delete any associated applications:']),
       getAppDeletionMessage()
     ]),
+    collaboratorEmails && collaboratorEmails.length > 0 && div({ style: { marginTop: '1rem' } }, [
+      p(`${pluralize('collaborator', collaboratorEmails.length, true)} will lose access to this workspace.`),
+      div(collaboratorEmails.map(
+        email => div({ key: email, style: { overflow: 'hidden', textOverflow: 'ellipsis' } }, [h(Link, { href: `mailto:${email}` }, [email])])
```
cahrens, line 112: What happens if there is a very long list of collaborators?

cahrens:
While you are in this area, I'd love it if you put a little padding below "Please type 'Delete Workspace'….".

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
cahrens: I have a suggestion-- we show the first X emails (maybe 5?), and then after that we include a line like "23 additional collaborators". I think the point here is to show that it will impact a lot of people, not to necessarily be able to see all of their email addresses, and we want to make sure that the dialog doesn't grow in an uncontrolled way.

nawatts: Done in https://github.com/DataBiosphere/terra-ui/pull/2843/commits/9b17fc95d44c3f984a4dc9516a508765ebd78427.

nawatts:
> While you are in this area, I'd love it if you put a little padding below "Please type 'Delete Workspace'….".

Added in [c203998](https://github.com/DataBiosphere/terra-ui/pull/2843/commits/c203998621b1a5734e980629005a969269e65e22).

src/libs/utils.js
nawatts: I don't know... If this was more complex, I'd agree. But is it worth vetting and maintaining a new dependency to replace 3 lines?

kyuksel: I'm okay with not using an external library in this case.

petesantos (approved):

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
cahrens: Thanks for making this change! You have to checkmarks, so you are good to go.

PR #2844, opened 2022-03-01
[SUP-366] Clarify process for sharing a workspace

nawatts:
Currently, users sometimes fail to share a workspace because they enter a collaborator's email in the "User email" field and click save.

Screen Shot 2022-03-01 at 11 42 17 AM

However, sharing a workspace is actually a two step process. They must either press "Enter" in the "User email" field or click the collaborator's email in the combobox's drop down list. This makes a few changes to clarify that:

Show a hint to press "enter" after entering the email while typing an email address. And add an "Add" button for a more obvious interaction to enter the email address.

Screen Shot 2022-03-01 at 11 41 40 AM

When a valid email address is entered in the "User email" field, disable the "Save" button and show a message.

Screen Shot 2022-03-01 at 11 41 26 AM

### DIFF
--- a/src/pages/workspaces/workspace/ShareWorkspaceModal.js
+++ b/src/pages/workspaces/workspace/ShareWorkspaceModal.js
@@ -200,10 +200,16 @@ const ShareWorkspaceModal = ({ onDismiss, workspace, workspace: { workspace: { n
 
   const remainingSuggestions = _.difference(suggestions, _.map('email', acl))
 
+  const addUserReminder = `Did you mean to add ${searchValue} as a collaborator? Add them or clear the "User email" field to save changes.`
+
   return h(Modal, {
     title: 'Share Workspace',
     width: 550,
-    okButton: h(ButtonPrimary, { onClick: save }, ['Save']),
+    okButton: h(ButtonPrimary, {
+      disabled: searchValueValid,
+      tooltip: searchValueValid && addUserReminder,
+      onClick: save
+    }, ['Save']),
     onDismiss
   }, [
     h(IdContainer, [id => h(Fragment, [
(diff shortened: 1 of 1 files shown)


cahrens:
@nawatts did you work with someone in UX to design this workflow? I agree that adding an explicit Add button is helpful (and I personally don't think we need to stress that they press Enter… we should just have them use the Add button). But to me the red validation error seems wrong because this isn't an error condition. It is expected that you will type in an address and then press Add. I would expect to see an error message if the email address is not registered, and think the tooltip is sufficient on the disabled Save button.

But mostly, I want to make sure that UX is in the loop in making these design choices.

nawatts:
> did you work with someone in UX to design this workflow?

Yes, these changes were discussed with UX.

> and I personally don't think we need to stress that they press Enter… we should just have them use the Add button

Yeah. The message text was written before the idea to add the "Add" button.

> But to me the red validation error seems wrong because this isn't an error condition. It is expected that you will type in an address and then press Add. I would expect to see an error message if the email address is not registered, and think the tooltip is sufficient on the disabled Save button.

I thought UX suggested the red text, but looking back at the design, I may have misunderstood a comment. On the other hand, I think it could be argued that, though not field-level , it's still a form-level validation error.

cahrens:
@nawatts Thanks Nick. I think it would be good to just run the final design by Joy. And to be clear, Return should still add the user, but I don't think we need to call it out in the help text since we have a visible Add button.

joyliu10:
hi Nick, the red text mean for a error if user didn't put in correct email format, this email address looks like correct.. so it should just be normal text, The red text comment I mentioned in a different ticket- I think is the delete workspace ticket. If there is a error when user enter the email address, the add button should be disabled until user enter correct address.

nawatts:
Removed the red text color in [3ad388f](https://github.com/DataBiosphere/terra-ui/pull/2844/commits/3ad388f1d1a4c2202a2ae32f792c3e0d09c9a2bd).

petesantos (approved):
LGTM, tried it out on PR site.

MatthewBemis (approved):

PR #2846, opened 2022-03-02
[WOR-71] Show total cost in Spend report

cahrens:
This displays total spend for a billing project under the following conditions:

1. User is billing project owner
2. User is part of the alpha spending group. "hermione.owner" is a member in dev.
3. The billing project has spend data stored. The "CARBillingTest" project has $.11 for the last 90 days (it's slowly increasing).

image

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -564,6 +564,21 @@ const Billing = signal => ({
     return res
   },
 
+  /**
+   * Returns the spend report for the given billing project, from 12 AM on the startDate to 11:59 PM on the endDate (UTC).
+   * @param billingProjectName
+   * @param startDate, a string of the format YYYY-MM-DD, representing the start date of the report.
+   * @param endDate a string of the format YYYY-MM-DD, representing the end date of the report.
+   * @returns {Promise<*>}
+   */
+  getSpendReport: async ({ billingProjectName, startDate, endDate }) => {
+    const res = await fetchRawls(
+      `billing/v2/${billingProjectName}/spendReport?${qs.stringify({ startDate, endDate })}`,
+      _.merge(authOpts(), { signal })
+    )
+    return res.json()
+  },
+
   listProjectUsers: async projectName => {
     const res = await fetchRawls(`billing/v2/${projectName}/members`, _.merge(authOpts(), { signal }))
     return res.json()
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -213,7 +213,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
   const [workspaceSort, setWorkspaceSort] = useState({ field: 'name', direction: 'asc' })
   const [totalCost, setTotalCost] = useState(null)
   const [updatingTotalCost, setUpdatingTotalCost] = useState(false)
-  const [spendReportLengthInDays, setSpendReportLengthInDays] = useState(90)
+  const [spendReportLengthInDays, setSpendReportLengthInDays] = useState(30)
 
   const signal = useCancellation()
 
@@ -315,7 +315,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
           options: _.map(days => ({
             label: `Last ${days} days`,
             value: days
-          }), [90]),
+          }), [7, 30, 90]),
           onChange: ({ value: selectedDays }) => {
             setSpendReportLengthInDays(selectedDays)
             setTotalCost(null) // This will force the report to be recalculated based on selectedDays
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -238,10 +238,6 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
   })
   maybeLoadTotalCost()
 
-  const groups = groupByBillingAccountStatus(billingProject, workspacesInProject)
-  const billingAccountsOutOfDate = !(_.isEmpty(groups.error) && _.isEmpty(groups.updating))
-  const getBillingAccountStatus = workspace => _.findKey(g => g.has(workspace), groups)
-
   const CostCard = ({ title, amount, ...props }) => {
     return div({
       ...props,
@@ -260,6 +256,10 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
       ])
     ])
   }
+
+  const groups = groupByBillingAccountStatus(billingProject, workspacesInProject)
+  const billingAccountsOutOfDate = !(_.isEmpty(groups.error) && _.isEmpty(groups.updating))
+  const getBillingAccountStatus = workspace => _.findKey(g => g.has(workspace), groups)
 
   const tabToTable = {
     workspaces: h(Fragment, [
@@ -414,6 +414,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
     () => !getShowBillingModal() && getBillingAccountsOutOfDate() && refreshWorkspaces(),
     { ms: 5000 }
   )
+
   const { displayName = null } = _.find({ accountName: billingProject.billingAccount }, billingAccounts) || { displayName: 'No Access' }
 
   return h(Fragment, [
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -13,7 +13,6 @@ import { useWorkspaces } from 'src/components/workspace-utils'
 import { Ajax } from 'src/libs/ajax'
 import * as Auth from 'src/libs/auth'
 import colors from 'src/libs/colors'
-import { getConfig } from 'src/libs/config'
 import { reportErrorAndRethrow } from 'src/libs/error'
 import Events from 'src/libs/events'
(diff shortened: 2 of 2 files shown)


src/pages/billing/Project.js
```diff
@@ -221,10 +224,42 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+        ...Style.elements.card.container,
+        backgroundColor: 'white',
+        width: 250,
+        padding: undefined,
+        boxShadow: undefined,
+        gridRowStart: 2
+      }
+    }, [
+      div({ style: { flex: 'none', padding: '10px 20px' } }, [
+        h3({ style: { fontSize: 16, color: colors.accent(), margin: '4px 0px', fontWeight: 'normal' } }, title),
```
cahrens: On this page, `h1` is "Terra Billing" (title bar) and `h2` is "Billing Projects" (in the nav). Therefore h3 would be the next heading level, though I'm not really sure if we want the cost information to be at a header level. I am definitely interested in what folks think about this. My thought with making it a header is that it helps with navigation, and it makes it clear that the cost value is associated with "Total cost".

src/pages/billing/Project.js
```diff
@@ -221,10 +224,42 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
     _.map('workspace', workspaces)
   ), [billingProject, workspaces])
 
+  const spendReportKey = 'spend report'
+  const maybeLoadTotalCost = reportErrorAndRethrow('Unable to retrieve spend report data')(async () => {
```
cahrens, line 228: Currently we get a 404 with no error message if no billing report data is available. I think we will want the billing API to return a more helpful error message in that case.

Note that there are other reasons that one can get a 404 (for example, startDate or endDate are missing), so we can't necessarily assume what a 404 means.

src/pages/billing/Project.js
```diff
@@ -221,10 +224,42 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+    if (!updatingTotalCost && totalCost === null && tab === spendReportKey) {
+      setUpdatingTotalCost(true)
+      const endDate = new Date().toISOString().slice(0, 10)
+      const startDate = new Date((Date.now() - spendReportLengthInDays * 24 * 60 * 60 * 1000)).toISOString().slice(0, 10)
+      const spend = await Ajax(signal).Billing.getSpendReport({ billingProjectName: billingProject.projectName, startDate, endDate })
+      const costFormatter = new Intl.NumberFormat(navigator.language, { style: 'currency', currency: spend.spendSummary.currency })
+      setTotalCost(costFormatter.format(spend.spendSummary.cost))
+      setUpdatingTotalCost(false)
```
cahrens, line 236: Since I am not changing `setUpdatingTotalCost` in a finally, if one gets an error from the API and want to try it again one must refresh the page. I think that is the behavior we want though, as we don't want repeated error attempts.

src/pages/billing/Project.js
```diff
@@ -271,7 +306,24 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+            label: `Last ${days} days`,
+            value: days
+          }), [90]),
+          onChange: ({ value: selectedDays }) => {
+            setSpendReportLengthInDays(selectedDays)
+            setTotalCost(null) // This will force the report to be recalculated based on selectedDays
+          }
+        })
+      ])])]),
+      CostCard({ title: 'Total spend', amount: (!!totalCost ? totalCost : '$__.__') })
```
cahrens, line 324: We may want to show a more explicit "Retrieving data" message in the future.

src/pages/billing/Project.js
```diff
@@ -221,10 +224,42 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+      style: {
+        ...Style.elements.card.container,
+        backgroundColor: 'white',
+        width: 250,
+        padding: undefined,
+        boxShadow: undefined,
+        gridRowStart: 2
+      }
+    }, [
+      div({ style: { flex: 'none', padding: '10px 20px' } }, [
```
cahrens: I am using pixels since these are just padding values and don't necessarily need to grow. I am trying to approximate UX mockups with respect to padding and font size.

src/pages/billing/Project.js
```diff
@@ -221,10 +224,42 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
     _.map('workspace', workspaces)
   ), [billingProject, workspaces])
 
+  const spendReportKey = 'spend report'
+  const maybeLoadTotalCost = reportErrorAndRethrow('Unable to retrieve spend report data')(async () => {
+    if (!updatingTotalCost && totalCost === null && tab === spendReportKey) {
+      setUpdatingTotalCost(true)
+      const endDate = new Date().toISOString().slice(0, 10)
+      const startDate = new Date((Date.now() - spendReportLengthInDays * 24 * 60 * 60 * 1000)).toISOString().slice(0, 10)
+      const spend = await Ajax(signal).Billing.getSpendReport({ billingProjectName: billingProject.projectName, startDate, endDate })
```
marctalbott, line 233: Would it be possible to display a spinning wheel while we're waiting for this to return? Even though it only appeared for a moment, I thought something was wrong when I opened the spend report and saw underscores.
Screen Shot 2022-03-02 at 5 27 49 PM

src/pages/billing/Project.js
marctalbott: I don't know enough about what should or should not be at a header level to have a strong opinion on this. When you say it helps with navigation, does that mean it's easier for users using screen readers to navigate the page this way? If this is the more accessible option, then that seems great to me

src/pages/billing/Project.js
```diff
@@ -221,10 +224,42 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+      const costFormatter = new Intl.NumberFormat(navigator.language, { style: 'currency', currency: spend.spendSummary.currency })
+      setTotalCost(costFormatter.format(spend.spendSummary.cost))
+      setUpdatingTotalCost(false)
+    }
+  })
+  maybeLoadTotalCost()
+
+  const CostCard = ({ title, amount, ...props }) => {
```
marctalbott, line 241: The cost card can only handle 9 digits before it overflows onto the next line like this
Screen Shot 2022-03-02 at 5 45 36 PM

This feels like a non-issue as our users aren't spending millions of dollars every 3 months as far as I know, but wanted to raise it in case you feel like I'm underestimating the issue

marctalbott (approved):
Couple of comments, but nothing blocking 👍

src/pages/billing/Project.js
cahrens: I created https://broadworkbench.atlassian.net/browse/WOR-171 for us to decide what we want to show to the user while the report is generating. This particular page already has 2 spinning wheels in play when the page first loads, and I'm somewhat reluctant to introduce a third.

src/pages/billing/Project.js
cahrens: Thanks, I didn't think to try this. I will see if it's possible to make the card grow in this case, though I agree it's probably a non-issue.

cahrens: I fixed this by allowing the card to grow if necessary. You see it grow, which isn't all that pretty, but the chances that any user actually has costs this high are extremely low.

image

marctalbott (approved):
looks great!

src/pages/billing/Project.js
cahrens: @marctalbott Sorry, I forgot to respond to this. Folks using screen readers may use headers to help navigate/understand the page (screen readers support navigation by header). But also it is good to make page content as semantic as possible, and I was trying to think through how we avoid having "a whole bunch of div content" on the Spend reporting tab. @dmohs what are your thoughts?

src/libs/ajax.js
```diff
@@ -564,6 +564,21 @@ const Billing = signal => ({
     return res
   },
 
+  /**
+   * Returns the spend report for the given billing project, from 12 AM on the startDate to 11:59 PM on the endDate (UTC).
```
dmohs, line 568: I love the precision in this comment.

src/pages/billing/Project.js
dmohs: Or shrink? '0.625rem 1.25rem' would get you the same initial values.

src/pages/billing/Project.js
dmohs: These pages are not articles, so the use of headings is going to be a bit ambiguous. You've given this some thought, it sounds good to me, and I think that is as much as we should expect from ourselves.

src/pages/billing/Project.js
dmohs: Alternatively:
```javascript
'$ ...'
```

dmohs (approved):

src/pages/billing/Project.js
cahrens: That's a possibility. We have a ticket in our backlog to figure out the best thing to show while loading the data.

src/pages/billing/Project.js
cahrens: Good point. I changed all the pixels to rems.

PR #2847, opened 2022-03-02
[IA-2172] Added links to copy image url and for image changelogs

cstrandw:
See [IA-2172](https://broadworkbench.atlassian.net/browse/IA-2172) for mockups and details

Changes:
- Added clipboard icon to ComputeModal to copy image url to clipboard (see screenshots for where)
- Clipboard icon has tooltip saying "Copy the image version of the runtime"
- Added links to github changelogs to version info for Terra maintained images

Testing:
- Tested locally by going through each environment config option to make sure that all links worked correctly and correct image url was copied

Screen Shot 2022-03-02 at 2 39 00 PM
Screen Shot 2022-03-02 at 2 39 17 PM
Screen Shot 2022-03-02 at 2 39 31 PM

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -634,7 +634,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     }
   }
 
-  const copyImageLinkTooltipText = 'Copy the image version of the runtime'
+  const copyImageLinkTooltipText = 'Copy the image version'
   const isTerraSupported = runtime => (!runtime.isCommunity && !(getToolForImage(runtime.id) === tools.RStudio.label) && !(runtime.id.includes('legacy')))
 
   const getLocationTooltip = (computeExists, bucketLocation) => Utils.cond(
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -2,7 +2,7 @@ import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
 import { b, br, code, div, fieldset, h, label, legend, li, p, span, strong, ul } from 'react-hyperscript-helpers'
 import {
-  ButtonOutline, ButtonPrimary, GroupedSelect, IdContainer, LabeledCheckbox, Link, Select, spinnerOverlay, WarningTitle
+  ButtonOutline, ButtonPrimary, ClipboardButton, GroupedSelect, IdContainer, LabeledCheckbox, Link, Select, spinnerOverlay, WarningTitle
 } from 'src/components/common'
 import { icon } from 'src/components/icons'
 import { ImageDepViewer } from 'src/components/ImageDepViewer'
@@ -547,7 +547,14 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
 
   const makeImageInfo = style => div({ style: { whiteSpace: 'pre', ...style } }, [
     div({ style: Style.proportionalNumbers }, ['Updated: ', updated ? Utils.makeStandardDate(updated) : null]),
-    div(['Version: ', version || null])
+    div([h(isTerraSupported(_.find({ image: selectedLeoImage }, leoImages)) ? Link : span, { href: `https://github.com/DataBiosphere/terra-docker/blob/master/${_.find({ image: selectedLeoImage }, leoImages)?.id }/CHANGELOG.md`, ...Utils.newTabLinkProps }, [
+      'Version: ', version || null
+    ]), h(ClipboardButton, {
+      text: selectedLeoImage,
+      style: { marginLeft: '0.5rem' },
+      tooltip: copyImageLinkTooltipText
+    })
+    ])
   ])
 
   const sendCloudEnvironmentMetrics = () => {
@@ -623,6 +630,9 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       return computeConfig.computeRegion !== location
     }
   }
+
+  const copyImageLinkTooltipText = 'Copy the image version of the runtime'
+  const isTerraSupported = (runtime) => (!runtime.isCommunity && !(getToolForImage(runtime.id) === tools.RStudio.label) && !(runtime.id.includes('legacy')))
 
   const getLocationTooltip = (computeExists, bucketLocation) => Utils.cond(
     [computeExists,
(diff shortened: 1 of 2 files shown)


src/components/ComputeModal.js
```diff
@@ -624,6 +634,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     }
   }
 
+  const copyImageLinkTooltipText = 'Copy the image version'
+  const isTerraSupported = runtime => (!runtime.isCommunity && !(getToolForImage(runtime.id) === tools.RStudio.label))
```
```suggestion
  const isTerraSupported = ({ isCommunity, id }) => !isCommunity && !(getToolForImage(id) === tools.RStudio.label)
```
- Object deconstruction at input time makes it more clear what props (from `runtime` in this case) the function uses.
- The function body can often be written more concisely that way since `runtime` doesn't have to repeated.
- The outer parentheses should be redundant.

src/components/ComputeModal.js
```diff
@@ -624,6 +634,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+  const getChangelogUrl = runtime => `https://github.com/DataBiosphere/terra-docker/blob/master/${runtime.id.replace('_legacy', '')}/CHANGELOG.md`
```
```suggestion
  const getChangelogUrl = ({ id }) => `https://github.com/DataBiosphere/terra-docker/blob/master/${_.replace('_legacy', '', id)}/CHANGELOG.md`
```

We usually prefer Lodash FP functions over native functions for several reasons:

{}.slice(2,4) // throws an error
_.slice(2, 4, {}) // returns an empty array
- When running through your code, if you pulled data in from an outside source and got unexpected garbage, the application will not have an error thrown, and can continue along with the value / type it was expecting when using Lodash.
- Lodash/fp functions support currying. This is especially useful for code changes later. If someone wants to add functionality, it makes it easy to do something like add the function to a flow.
- Lodash/fp will not mutate an array. In the case of native slice, it also returns a new array. But there are other functions that can mutate (e.g. push/pop)
- Mixing native and lodash functions can lead to confusion about behavior.

src/components/ComputeModal.js
```diff
@@ -624,6 +634,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     }
   }
 
+  const copyImageLinkTooltipText = 'Copy the image version'
```
kyuksel: I don't think it's worth separating out the text string in this case.
```suggestion
```

src/components/ComputeModal.js
```diff
@@ -1516,7 +1530,11 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
               div({ style: { fontSize: 16, fontWeight: 600 } }, ['Use default environment']),
               ul({ style: { paddingLeft: '1rem', marginBottom: 0, lineHeight: 1.5 } }, [
                 li([
-                  div([packageLabel]),
+                  div([packageLabel, h(ClipboardButton, {
+                    text: selectedLeoImage,
+                    style: { marginLeft: '0.5rem' },
+                    tooltip: copyImageLinkTooltipText
```
kyuksel: We could just inline the text here:
```suggestion
                    tooltip: 'Copy the image version'
```

src/components/common.js
```diff
@@ -540,7 +540,7 @@ export const ClipboardButton = ({ text, onClick, children, ...props }) => {
   const [copied, setCopied] = useState(false)
   return h(Link, {
     ...props,
-    tooltip: copied ? 'Copied to clipboard' : 'Copy to clipboard',
+    tooltip: props?.tooltip ?? (copied ? 'Copied to clipboard' : 'Copy to clipboard'),
```
kyuksel: Can we do away with the new logic if we swap the line `…props,` with the line starting with `tooltip:` since `props.tooltip` would override the default `tooltip` if there exists any `props.tooltip`?

src/components/ComputeModal.js
```diff
@@ -547,7 +547,17 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+      {
+        href: getChangelogUrl(_.find({ image: selectedLeoImage }, leoImages)),
+        ...Utils.newTabLinkProps
+      }, [
+        'Version: ', version || null
+      ]), h(ClipboardButton, {
+      text: selectedLeoImage,
+      style: { marginLeft: '0.5rem' },
+      tooltip: copyImageLinkTooltipText
+    })])
```
kyuksel: - `h(` needs to be followed by React components (whose initials are capitalized for distinction).
- I wonder if we could just disable the `Link` for the case of community-maintained images.
- Could remove the outer `div`.

```suggestion
    h(Link, {
      disabled: !isTerraSupported(_.find({ image: selectedLeoImage }, leoImages)),
      href: getChangelogUrl(_.find({ image: selectedLeoImage }, leoImages)),
      ...Utils.newTabLinkProps
    }, ['Version: ', version || null]),
    h(ClipboardButton, {
      text: selectedLeoImage,
      style: { marginLeft: '0.5rem' },
      tooltip: copyImageLinkTooltipText
    })
```

Feel free to run the `disabled` `Link` change by UX.

src/components/ComputeModal.js
cstrandw: I get this error when I try this formatting, any idea why?
Screen Shot 2022-03-03 at 11 23 18 AM

kyuksel: Ah yea…`isCommunity` being `undefined` in the original version wasn't causing an error but it does in the proposed way. Assigning a default value should make it behave the same way it did before, and the intention would be more explicit:

```javascript
const isTerraSupported = ({ isCommunity = undefined, id }) => !isCommunity && !(getToolForImage(id) === tools.RStudio.label)
```

src/components/ComputeModal.js
```diff
@@ -547,7 +547,17 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
 
   const makeImageInfo = style => div({ style: { whiteSpace: 'pre', ...style } }, [
     div({ style: Style.proportionalNumbers }, ['Updated: ', updated ? Utils.makeStandardDate(updated) : null]),
-    div(['Version: ', version || null])
+    h(Link, {
+      href: getChangelogUrl(_.find({ image: selectedLeoImage }, leoImages)),
+      disabled: !isTerraSupported(_.find({ image: selectedLeoImage }, leoImages)),
+      ...Utils.newTabLinkProps
+    },
+    ['Version: ', version || null]),
```
kyuksel: Styling-wise, `[` usually comes after `},`, which would also save a little vertical space:
```suggestion
    }, ['Version: ', version || null]),
```

kyuksel (approved):
Looks good! One very minor comment inline.

Qi77Qi:
lgtm on PR site

Qi77Qi (approved):

cstrandw:
> lgtm on PR site

what does lgtm mean?

PR #2850, opened 2022-03-03
[SUP-469] Handle requester pays errors when fetching workspace bucket location

nawatts:
Currently, when a user views a workspace with a requester pays bucket and they only have reader access to the workspace (a common scenario with public/featured workspaces), they see an "Error loading bucket location data" notification and the bucket's location says "Loading..." forever.

Screen Shot 2022-03-03 at 5 03 28 PM

Screen Shot 2022-03-03 at 5 04 10 PM

This catches that error and prompts the user for a project to bill to.


MatthewBemis (approved):
looks good to me!

petesantos (approved):

PR #2851, opened 2022-03-04
[DC-257] Fixed "Link to new workspace" integration test being flakey

lu-c:
I think I found the problem - Narrowed it down to the cleanup method, and the page.evaluate method. I think the problem is that the delete method was not being awaited on, and was possibly ending after the test was, causing issues.

At the very least, I am making this PR after running it locally 10 times without error. Here's to hoping it wasn't just weird timing issues!

To test, I ran it locally using
`yarn test-local link-to-new-workspace`

### DIFF
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -25,9 +25,9 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ page, testUrl, token }) 
 
   const cleanupFn = async () => {
     try {
-      await page.evaluate((name, billingProject) => {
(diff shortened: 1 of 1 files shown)


kyuksel:
> I am making this PR after running it locally 10 times without error.

To have more confidence, can we have `flake-checker` run the test at least 100 times?

lu-c:
Testing with test-flakes, 
Screen Shot 2022-03-04 at 12 50 39 PM

After 100 runs, it failed 1 time on a timeout - this is definitely an improvement on how it works right now on dev!

kyuksel (approved):

petesantos (approved):
Thanks for catching this!

PR #2854, opened 2022-03-04
Add first browser-based test for spend reports (WOR-142, WOR-179).

cahrens:
This adds a browser-based test that is a hybrid of running against actual services and mocked AJAX responses. Specifically, I log a user in with existing integration utility methods, but then my interactions with the Billing Projects page to test the new "Spend report" (which is only accessible to users in the alpha group) are completely mocked.

This test passed 200 times with the flaky test runner.

### DIFF
--- a/integration-tests/tests/billing-projects.js
+++ b/integration-tests/tests/billing-projects.js
@@ -21,7 +21,7 @@ class BillingProjectsPage {
       projectName: this.mockValues.ownedBillingProjectName,
       billingAccount: 'billingAccounts/fake-id', invalidBillingAccount: false, roles: ['Owner'], status: 'Ready'
     }]
-    const setAjaxMock = async () => {
+    const setAjaxOverrides = async () => {
       await this.testPage.evaluate((spendReturnResult, projectListResult) => {
         window.ajaxOverridesStore.set([
           {
@@ -43,7 +43,7 @@ class BillingProjectsPage {
         ])
       }, spendReturnResult, projectListResult)
     }
-    await setAjaxMock()
+    await setAjaxOverrides()
   }
 
   async visit() {
--- a/integration-tests/tests/billing-projects.js
+++ b/integration-tests/tests/billing-projects.js
@@ -3,7 +3,7 @@ const { click, clickable, dismissNotifications, findText, noSpinnersAfter, selec
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
-class BillingAccountPage {
+class BillingProjectsPage {
   constructor(testPage, testUrl) {
     this.testPage = testPage
     this.testUrl = testUrl
@@ -74,7 +74,7 @@ const testBillingSpendReportFn = withUserToken(async ({ page, testUrl, token }) 
   await dismissNotifications(page)
 
   // Interact with the Billing Page via mocked AJAX responses.
-  const billingPage = new BillingAccountPage(page, testUrl)
+  const billingPage = new BillingProjectsPage(page, testUrl)
   await billingPage.setAjaxMock({ ownedBillingProjectName: 'OwnedBillingProject', spendCost: '90.13' })
   await billingPage.visit()
   await billingPage.selectSpendReport()
--- a/integration-tests/tests/billing-projects.js
+++ b/integration-tests/tests/billing-projects.js
@@ -3,68 +3,54 @@ const { click, clickable, dismissNotifications, findText, noSpinnersAfter, selec
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
-class BillingProjectsPage {
-  constructor(testPage, testUrl) {
-    this.testPage = testPage
-    this.testUrl = testUrl
-    this.mockValues = {}
-  }
+const billingProjectsPage = (testPage, testUrl) => {
+  return {
+    visit: async () => await testPage.goto(`${testUrl}/#billing`),
 
-  async setAjaxMock(mockValues) {
-    this.mockValues = _.assign(this.mockValues, mockValues)
-    const spendReturnResult = {
-      spendSummary: {
-        cost: this.mockValues.spendCost, credits: '2.50', currency: 'USD', endTime: '2022-03-04T00:00:00.000Z', startTime: '2022-02-02T00:00:00.000Z'
+    selectSpendReport: async billingProjectName => {
+      await noSpinnersAfter(
+        testPage,
+        { action: () => click(testPage, clickable({ text: billingProjectName })) }
+      )
+      await click(testPage, clickable({ text: 'Spend report' }))
+    },
+
+    setSpendReportDays: async days => await select(testPage, 'Date range', `Last ${days} days`),
+
+    assertText: async expectedText => await findText(testPage, expectedText)
+  }
+}
+
+const setAjaxMockValues = async (testPage, ownedBillingProjectName, spendCost) => {
+  const spendReturnResult = {
+    spendSummary: {
+      cost: spendCost, credits: '2.50', currency: 'USD', endTime: '2022-03-04T00:00:00.000Z', startTime: '2022-02-02T00:00:00.000Z'
+    }
+  }
+  const projectListResult = [{
+    projectName: ownedBillingProjectName,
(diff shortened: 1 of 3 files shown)


integration-tests/tests/billing-projects.js
```diff
@@ -0,0 +1,92 @@
+const _ = require('lodash/fp')
+const { click, clickable, dismissNotifications, findText, noSpinnersAfter, select, signIntoTerra } = require('../utils/integration-utils')
+const { withUserToken } = require('../utils/terra-sa-utils')
+
+
+class BillingProjectsPage {
```
cahrens: I imagine my choice of using a class and calling methods on it in the test will be controversial. I am trying to move my tests toward the page pattern, which I have found to be a very nice way to write browser-based tests in the past.

https://martinfowler.com/bliki/PageObject.html

integration-tests/tests/billing-projects.js
```diff
@@ -0,0 +1,92 @@
...
+    this.mockValues = _.assign(this.mockValues, mockValues)
+    const spendReturnResult = {
+      spendSummary: {
+        cost: this.mockValues.spendCost, credits: '2.50', currency: 'USD', endTime: '2022-03-04T00:00:00.000Z', startTime: '2022-02-02T00:00:00.000Z'
+      }
+    }
+    const projectListResult = [{
+      projectName: this.mockValues.ownedBillingProjectName,
+      billingAccount: 'billingAccounts/fake-id', invalidBillingAccount: false, roles: ['Owner'], status: 'Ready'
+    }]
```
dmohs: This is the hazard that needs careful consideration. When the behavior of the target service changes, the confidence generated from this test should change, but I don't see anything that tells me we will become knowledgable of such an event.

cahrens: @dmohs That's true, but it's not a goal of the test. This is not an actual integration test (and perhaps we want to put these sort of tests in a different directory to make it clear). The goal of a test like this is to test the JavaScript code itself, so that we can be confident when we change utility methods or upgrade JavaScript 3rd party libraries, we have not broken anything. I explicitly do NOT want to be dependent on other services. I discuss this in the ticket itself-- https://broadworkbench.atlassian.net/browse/WOR-142

This test is actually a great example of why an integration test **cannot** work for this functionality. When getting the cost for the last X days, we depend on the BigQuery export of cost data from Google. There is no way I can keep actual data at an expected value from day-to-day, and depending on BigQuery data to exist introduces a ton of dependencies. I am trusting Rawls unit tests to enforce the contract from the API, and these tests are written against that contract. This type of test does not replace true integration tests, but it allows us to test _so many more things_, without the overhead and test flakiness we have seen with the existing integration tests.

integration-tests/tests/billing-projects.js
kyuksel: I added the PR on the UI Review Mobbing agenda, with one reason being I'd like to understand why the goal cannot be accomplished without introducing a class component, which we've been trying to move away from. Haven't read the link you mentioned yet.

petesantos (commented):
Overall I like the abstraction. I think it is a nice usage of the overrides and can be a good test of the functionality on a page.

What would you think of pulling the Ajax state out of the class? I view the Ajax result as the state we want to test and pulling that out into its own state management code may allow the page abstractions to be pure functions with the state passed in.

We can discuss it at the reviewers meeting.

integration-tests/tests/billing-projects.js
andy7i: asking out of curiosity, why are classes undesirable for the UI code?

kyuksel: One thing to clarify: I wrote the previous comment after too quick of a look at the PR, with the misunderstanding that `BillingProjectsPage` is a (React) class component, which it isn't. It's just a class. If your question is, why use function components over class components, the answer may be something like:
- class components come pre-loaded with stuff that you may not need
- function components are more light-weight and help you minimize state (which is a major pillar of UI development and then some)
- can be hard to reuse stateful logic between components
- I read things here and there that there may be reduced support for class components

Even though `BillingProjectsPage` is not a component, I'd be curious if we could do away with it by having mocked values stored somewhere and pure functions retrieve them as state makes debugging and reasoning about programmes more difficult.

cahrens:
@petesantos and others, I refactored the code to keep the mock values separate and to avoid using a class. Please take another look and let me know what you think.

aherbst-broad (approved):
lgtm

petesantos:
> @petesantos and others, I refactored the code to keep the mock values separate and to avoid using a class. Please take another look and let me know what you think.

Thanks for doing this! I like the functional purity of `BillingProjectPage` with the different creation pattern. I think this is a nice pattern we can use for testing UI functionality on a page while preventing flakiness.

integration-tests/utils/integration-utils.js
```diff
@@ -48,7 +48,7 @@ const getAnimatedDrawer = textContains => {
 // Note: isEnabled is not fully supported for native anchor and button elements (only aria-disabled is examined).
 const clickable = ({ text, textContains, isDescendant = false, isEnabled = true }) => {
   const checkEnabled = isEnabled === false ? '[@aria-disabled="true"]' : '[not(@aria-disabled="true")]'
-  const base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"])${checkEnabled}`
+  const base = `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"] | //*[@role="tab"])${checkEnabled}`
```
kyuksel, line 51: @petesantos and I were talking about this earlier. Adding another `or` to that logic makes it even more difficult to know which regular expression piece was targeted when a test fails due to timing out while waiting for that XPath. Can we pull out `//*[@role="tab"]` into its own function (i.e. `const tab = …`)?

petesantos (approved):

integration-tests/utils/integration-utils.js
petesantos: +1 ... I missed this when I thumbed the PR. I was too excited about the new pattern!

cahrens: @kyuksel Sorry, but I'm not following what you are requesting. Can you provide more of an example?

cahrens: I will include this in my next PR (once I understand it)… I wanted to go ahead and get this first test in.

kyuksel: Below is a screenshot of a diff for an example:

image

cahrens: Thanks @kyuksel. 

I am struggling though to understand why we want a different method for "tab" because I thought that "clickable" is meant to encompass all elements that are clickable. When I wrote the test, I fully expected "clickable" to work, and I don't think I would have found a "tab" method.

kyuksel: @petesantos I have a feeling you'll articulate it better 😊

kyuksel: But basically, you get errors like

```
TimeoutError: waiting for XPath `(//a | //button | //*[@role="button"] | //*[@role="link"] | //*[@role="combobox"] | //*[@role="option"] | //*[@role="tab"])[not(@aria-disabled="true")][contains(normalize-space(.),"Create new notebook") or contains(@title,"Create new notebook") or contains(@alt,"Create new notebook") or contains(@aria-label,"Create new notebook") or @aria-labelledby=//*[contains(normalize-space(.),"Create new notebook")]/@id]` failed: timeout 30000ms exceeded
```

when a test fails with a timeout because it couldn't find what it was looking and it's more difficult to know what exactly it was the test was looking for because it can be one of the many components of the `or` logic.

cahrens: I read it as "I'm looking for a clickable thing with text 'Create new notebook'". I don't think it makes sense to split out `tab` by itself… if we want to go that route, we should also split out all the others into individual methods (combobox, select, etc.).

I **do** agree that we shouldn't add `img` here, because images are generally not clickable.

PR #2856, opened 2022-03-04
[PF-693] For alpha users, make all regions available for workspace bucket location

wnojopra:
This makes all regions available as an option for workspace bucket location When a user is part of the Alpha group (named Alpha_Regionality_Users). When a user is NOT a part of the Alpha group, there is no change. This is a part of "Phase 3" of regionality - where [Phase 2](https://github.com/DataBiosphere/terra-ui/pull/2680) previously added Montreal as an available region. 

image

### DIFF
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -71,7 +71,7 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
     ...(requiredAuthDomain ? [requiredAuthDomain] : [])
   ])
 
-  const loadAlphaRegionalityUser = reportErrorAndRethrow('Error loading user group membership')(async () => {
+  const loadAlphaRegionalityUser = reportErrorAndRethrow('Error loading regionality group membership')(async () => {
     setIsAlphaRegionalityUser(await Ajax(signal).Groups.group(getConfig().alphaRegionalityGroup).isMember())
   })
 
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -6,11 +6,12 @@ import { icon } from 'src/components/icons'
 import { TextArea, ValidatedInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
 import { InfoBox } from 'src/components/PopupTrigger'
-import { availableBucketRegions, getRegionInfo, isSupportedBucketLocation, locationTypes } from 'src/components/region-common'
+import { allRegions, availableBucketRegions, getRegionInfo, isSupportedBucketLocation, locationTypes } from 'src/components/region-common'
 import TooltipTrigger from 'src/components/TooltipTrigger'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
-import { withErrorReporting } from 'src/libs/error'
+import { getConfig } from 'src/libs/config'
+import { reportErrorAndRethrow, withErrorReporting } from 'src/libs/error'
 import Events from 'src/libs/events'
 import { FormLabel } from 'src/libs/forms'
 import * as Nav from 'src/libs/nav'
@@ -60,6 +61,7 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
   const [createError, setCreateError] = useState()
   const [bucketLocation, setBucketLocation] = useState(defaultLocation)
   const [sourceWorkspaceLocation, setSourceWorkspaceLocation] = useState(defaultLocation)
+  const [isAlphaRegionalityUser, setIsAlphaRegionalityUser] = useState(false)
   const signal = useCancellation()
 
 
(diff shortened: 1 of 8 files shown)


src/components/NewWorkspaceModal.js
```diff
@@ -132,6 +138,7 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
   // Lifecycle
   useOnMount(() => {
     loadData()
+    loadAlphaRegionalityUser()
```
kyuksel, line 141: Does this work as intended? Asking since `loadAlphaRegionalityUser` doesn't seem to be a function? Could we inline like below?
```suggestion
    reportErrorAndRethrow('Error loading user group membership')(async () => {
    setIsAlphaRegionalityUser(await Ajax(signal).Groups.group(getConfig().alphaRegionalityGroup).isMember())
  })
```

kyuksel (commented):
The code changes look good overall. One question inline.

I verified the case where if a user is not part of the alpha group, there's no change in `NewWorkspaceModal` options. How can I test the other case? Do I need to be added to the alpha group, temporarily perhaps?

src/components/NewWorkspaceModal.js
wnojopra: loadAlphaRegionalityUser is defined on lines 74-77. Am I missing something?

wnojopra:
> I verified the case where if a user is not part of the alpha group, there's no change in `NewWorkspaceModal` options. How can I test the other case? Do I need to be added to the alpha group, temporarily perhaps?

Kaan, I've added kyuksel@broadinstitute.org as an admin to Alpha_Regionality_Users@dev.test.firecloud.org (note this is dev). LMK if that doesn't work for you.

kyuksel:
> Kaan, I've added [kyuksel@broadinstitute.org](mailto:kyuksel@broadinstitute.org) as an admin to [Alpha_Regionality_Users@dev.test.firecloud.org](mailto:Alpha_Regionality_Users@dev.test.firecloud.org) (note this is dev). LMK if that doesn't work for you.

Would you mind removing `kyuksel@broadinstitute.org` and adding `kyuksel.dev@gmail.com` instead? We try to not log into non-prod systems with our primary Broad account for security reasons.

src/components/NewWorkspaceModal.js
kyuksel: Ah never mind...I'd misread the definition of `reportErrorAndRethrow` which is called by `loadAlphaRegionalityUser`.

wnojopra:
Just added kyuksel.dev@gmail.com as an admin to Alpha_Regionality_Users@dev.test.firecloud.org

src/components/NewWorkspaceModal.js
```diff
@@ -69,6 +71,10 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
     ...(requiredAuthDomain ? [requiredAuthDomain] : [])
   ])
 
+  const loadAlphaRegionalityUser = reportErrorAndRethrow('Error loading user group membership')(async () => {
```
cahrens: I would suggest making this error more specific so we know that it was encountered when trying to obtain the regionality group membership (`Error loading regionality group membership`). I think you copied this from my code in List.js, so I will make mine more specific as well. :)

cahrens (approved):
This looks good to me based on the code (I didn't try running it, but it seems like @kyuksel is doing that testing). I had one suggestion with the error wording.

src/components/NewWorkspaceModal.js
wnojopra: Done! And thank you for being the first example of this pattern

kyuksel:
@wnojopra Didn't we use to display a warning text in orange when selecting a different runtime location than the workspace buckets? I don't see it anymore.

image

wnojopra:
@kyuksel If you click NEXT, you should see a warning that looks like this: 
image

Is this what you're thinking of? You may also be thinking of the warning that shows up in the clone workspace dialog:
image

kyuksel:
Below is while cloning a workspace whose bucket is in Jakarta. Shouldn't there be more options?

image

kyuksel:
> If you click NEXT, you should see a warning that looks like this:

Ah I see

image


It's different than the one in your message but that's probably okay?

wnojopra:
> Below is while cloning a workspace whose bucket is in Jakarta. Shouldn't there be more options?

Yes! I'm not able to reproduce your issue:
image
The list is determined by a groups check when the dialog is created. Any chance that the API call is failing for you for some reason? If you refresh and try again, are you able to reproduce the issue?

> It's different than the one in your message but that's probably okay?

Yes, the error you're seeing is the correct one. Sorry, I copy pasted a screenshot from an older PR.

kyuksel:
> Any chance that the API call is failing for you for some reason? If you refresh and try again, are you able to reproduce the issue?

I don't think the API call was failing since I didn't see any errors being reported. I tried again a couple of times and saw the same issue, with another workspace as well. Then I modified the code a bit to add a print statement to debug the issue. When I ran the modified code, I saw the complete list. Perhaps that forced flush of some stale stuff on my browser  🤷🏼

kyuksel (approved):

PR #2861, opened 2022-03-07
[AJ-154] Preview drs uris

davidangb:
Allow previews for drs:// uris

Before:
drs-uri-preview-before

After:
drs-uri-preview-after

### DIFF
--- a/src/components/UriViewer.js
+++ b/src/components/UriViewer.js
@@ -80,13 +80,13 @@ const PreviewContent = ({ uri, metadata, metadata: { bucket, name }, googleProje
     }
   }
   useOnMount(() => {
-    if (isGs(uri) && isFilePreviewable(metadata)) {
+    if (isFilePreviewable(metadata)) {
       loadPreview()
     }
   })
   return els.cell([
     Utils.cond(
-      [isGs(uri) && isFilePreviewable(metadata), () => h(Fragment, [
+      [isFilePreviewable(metadata), () => h(Fragment, [
         els.label('Preview'),
         Utils.cond(
           [preview === null, () => 'Unable to load preview.'],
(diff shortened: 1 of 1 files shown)


src/components/UriViewer.js
```diff
@@ -80,13 +80,13 @@ const PreviewContent = ({ uri, metadata, metadata: { bucket, name }, googleProje
     }
   }
   useOnMount(() => {
-    if (isGs(uri) && isFilePreviewable(metadata)) {
+    if (isFilePreviewable(metadata)) {
       loadPreview()
     }
   })
```
kyuksel: We could save a good amount of vertical space with

```javascript
  useOnMount(() => isFilePreviewable(metadata) && loadPreview())
```

`someFlag && doSomething()` is also the pattern we usually follow for simple conditional statements.

src/components/UriViewer.js
```diff
@@ -80,13 +80,13 @@ const PreviewContent = ({ uri, metadata, metadata: { bucket, name }, googleProje
...
   return els.cell([
```
kyuksel, line 83: Not your PR but I was wondering what `els` is, which makes me want to rename the variable to something more explanatory  🤷🏼

kyuksel (commented):
Code and the screenshot look good to me with a couple of comments inline. Didn't test.

jordansafer (approved):

src/components/UriViewer.js
davidangb: good call; adopted. I hadn't looked at refactoring the existing code like this, I was just making the minimal changes necessary.

src/components/UriViewer.js
davidangb: I'm going to leave this alone … I am also wondering what `els` is supposed to mean so I don't know what to change it to.

src/components/UriViewer.js
kyuksel: very understandable - thank you!

kyuksel (approved):

PR #2862, opened 2022-03-08
AJ-313: handle stale metadata

davidangb:
Recent perf/scaling work has enabled - on a per-workspace feature-flag basis - caching for entity type metadata. This results in the possibility that metadata is stale.

The biggest UX impact of stale metadata we've discovered so far is missing columns from a data table. The use case is:
* start with a workspace that has one data table "participant"
* enable the "alwaysCacheAttributes" feature flag for this workspace
* upload a TSV to create a new "sample" table; this TSV has 10 columns in it
* navigate to the "sample" table, note that it only displays "sample_id" and no other columns

This happens because in the first ~2 minutes after the TSV upload, the cache is stale. Since the cache previously knew nothing about the "sample" table, it returns no attributeNames for this table.

This PR changes the behavior of the data table so that it inspects the current page of results, finds all the attributes in that page of results, and adds any attributes it finds in the results to the metadata. This should prevent stale cache from hiding columns that exist in the page of results.

Impact to end user:
* if the feature flag is not enabled (the default), no change in behavior
* if the feature flag *is* enabled, data tables are more resilient to stale cache

Tested by running locally and manually modifying the backend Rawls db's cache entries for my workspace to be missing some columns. Verified that those columns do not display before this PR, but do display after this PR.

**UPDATE:** I did the back end tweaking to set up a workspace to demonstrate this PR. LMK if you'd like me to add you to the workspace.
https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/da-invalid-cache/data is missing entries in its cache for the "sample" type. It does not display the proper columns for samples.
https://pr-2-dot-bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/da-invalid-cache/data (the PR site for this PR) does display the proper columns for samples.

Dev:
Screenshot (4)

PR Site:
Screenshot (5)

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -122,7 +122,7 @@ const DataTable = props => {
     // For the time being, the uniquness check MUST be case-insensitive (e.g. { sensitivity: 'accent' })
     // in order to prevent case-divergent columns from being displayed, as that would expose some other bugs
     const newAttrsForThisType = concatenateAttributeNames(entityMetadata[entityType]?.attributeNames, attrNamesFromResults)
-    if (newAttrsForThisType !== entityMetadata[entityType].attributeNames) {
+    if (!_.isEqual(newAttrsForThisType, entityMetadata[entityType].attributeNames)) {
       const newMetadata = _.cloneDeep(entityMetadata)
       newMetadata[entityType].attributeNames = newAttrsForThisType
       setEntityMetadata(newMetadata)
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -3,7 +3,7 @@ import { Fragment, useEffect, useRef, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import { Checkbox, Clickable, fixedSpinnerOverlay, Link } from 'src/components/common'
-import { DeleteEntityColumnModal, EditDataLink, EntityEditor, EntityRenamer, HeaderOptions, renderDataCell } from 'src/components/data/data-utils'
+import { concatenateAttributeNames, DeleteEntityColumnModal, EditDataLink, EntityEditor, EntityRenamer, HeaderOptions, renderDataCell } from 'src/components/data/data-utils'
 import { icon } from 'src/components/icons'
 import { ConfirmedSearchInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
@@ -116,16 +116,17 @@ const DataTable = props => {
           { filterTerms: activeTextFilter })
       }))
     // find all the unique attribute names contained in the current page of results
-    const allAttrNames = _.uniq(_.flatMap(_.keys, _.map('attributes', results)))
+    const attrNamesFromResults = _.uniq(_.flatMap(_.keys, _.map('attributes', results)))
     // add any attribute names from the current page of results to those found in metadata.
     // This allows for the stale metadata (e.g. the metadata cache is out of date).
     // For the time being, the uniquness check MUST be case-insensitive (e.g. { sensitivity: 'accent' })
     // in order to prevent case-divergent columns from being displayed, as that would expose some other bugs
-    const newAttrsForThisType = _.uniqWith((left, right) => !left.localeCompare(right, undefined, { sensitivity: 'accent' }),
-      _.concat(entityMetadata[entityType].attributeNames, allAttrNames)).sort()
-    const newMetadata = _.cloneDeep(entityMetadata)
-    newMetadata[entityType].attributeNames = newAttrsForThisType
-    setEntityMetadata(newMetadata)
+    const newAttrsForThisType = concatenateAttributeNames(entityMetadata[entityType]?.attributeNames, attrNamesFromResults)
+    if (newAttrsForThisType !== entityMetadata[entityType].attributeNames) {
+      const newMetadata = _.cloneDeep(entityMetadata)
+      newMetadata[entityType].attributeNames = newAttrsForThisType
+      setEntityMetadata(newMetadata)
+    }
     setEntities(results)
     setFilteredCount(filteredCount)
     setTotalRowCount(unfilteredCount)
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -123,9 +123,7 @@ const DataTable = props => {
     // in order to prevent case-divergent columns from being displayed, as that would expose some other bugs
     const newAttrsForThisType = concatenateAttributeNames(entityMetadata[entityType]?.attributeNames, attrNamesFromResults)
     if (!_.isEqual(newAttrsForThisType, entityMetadata[entityType].attributeNames)) {
-      const newMetadata = _.cloneDeep(entityMetadata)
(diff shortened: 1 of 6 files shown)


src/components/data/DataTable.js
```diff
@@ -114,6 +115,17 @@ const DataTable = props => {
...
+    const allAttrNames = _.uniq(_.flatMap(_.keys, _.map('attributes', results)))
+    // add any attribute names from the current page of results to those found in metadata.
+    // This allows for the stale metadata (e.g. the metadata cache is out of date).
+    // For the time being, the uniquness check MUST be case-insensitive (e.g. { sensitivity: 'accent' })
+    // in order to prevent case-divergent columns from being displayed, as that would expose some other bugs
+    const newAttrsForThisType = _.uniqWith((left, right) => !left.localeCompare(right, undefined, { sensitivity: 'accent' }),
+      _.concat(entityMetadata[entityType].attributeNames, allAttrNames)).sort()
+    const newMetadata = _.cloneDeep(entityMetadata)
+    newMetadata[entityType].attributeNames = newAttrsForThisType
+    setEntityMetadata(newMetadata)
```
davidangb, line 128: question for the UI experts: do I need to explicitly call `setEntityMetadata()` to persist my changes to state, or could I just use:
```entityMetadata[entityType].attributeNames = newAttrsForThisType```
both seem to function the same as I click through the app, but I don't know if there is an underlying difference.

File: src/pages/workspaces/workspace/Data.js [hunk 2]
```diff
@@ -288,7 +288,7 @@ const BucketContent = _.flow(
       )
     ]),
     div({ style: { margin: '1rem -1rem 1rem -1rem', borderBottom: `1px solid ${colors.dark(0.25)}` } }),
-    div({ style: { flex: '1 1 0', overflow: 'auto' }}, [
+    div({ style: { flex: '1 1 0', overflow: 'auto' } }, [
```
davidangb, line 291: this spacing change is unrelated to this PR but I was seeing warnings so I fixed it.

davidangb:
putting this back to draft, there's a little more I want to do on it

src/components/data/DataTable.js
```diff
@@ -114,6 +115,18 @@ const DataTable = props => {
...
+    // add any attribute names from the current page of results to those found in metadata.
+    // This allows for the stale metadata (e.g. the metadata cache is out of date).
+    // For the time being, the uniquness check MUST be case-insensitive (e.g. { sensitivity: 'accent' })
+    // in order to prevent case-divergent columns from being displayed, as that would expose some other bugs
+    const newAttrsForThisType = concatenateAttributeNames(entityMetadata[entityType]?.attributeNames, attrNamesFromResults)
+    if (newAttrsForThisType !== entityMetadata[entityType].attributeNames) {
+      const newMetadata = _.cloneDeep(entityMetadata)
+      newMetadata[entityType].attributeNames = newAttrsForThisType
+      setEntityMetadata(newMetadata)
+    }
```
davidangb, line 127: question for the UI experts: do I need to explicitly call `setEntityMetadata()` to persist my changes to state, or could I just use:
```entityMetadata[entityType].attributeNames = newAttrsForThisType```
both seem to function the same as I click through the app, but I don't know if there is an underlying difference.

Comment (jordansafer) [reply src/components/data/DataTable.js:127]: From the react FAQ, I see them doing this: `      setState(state => ({ ...state, left: e.pageX, top: e.pageY }));`. The only actual difference I can see is that react has a mechanism to look at previous state compared to current state, and if we mutated current state than that wouldn't work, which would be confusing if someone was trying to use that mechanism. That being said, we aren't using that mechanism :)

My vote is for simplifying a little by getting rid of the clone, and using the setter like in the react FAQ. Should be something like this: `setEntityMetadata(entityMetadata => ({ ...entityMetadata, attributeNames: newAttrsForThisType }))`

**Review by jordansafer — APPROVED** [review]

File: src/components/data/DataTable.js (hunk 3 above)
Comment (dmohs) [reply src/components/data/DataTable.js:127]: This is the way. Something like:
```javascript
setEntityMetadata(_.update(['attributeNames', entityType], newAttrsForThisType))
```
Note that Lodash's auto-currying is aiding concision here.

src/components/data/EntitiesContent.js
```diff
@@ -369,7 +369,7 @@ const EntitiesContent = ({
     h(Fragment, [
       h(DataTable, {
         persist: true, firstRender, refreshKey, editable: !snapshotName && !Utils.editWorkspaceError(workspace),
-        entityType: entityKey, entityMetadata, columnDefaults, googleProject, workspaceId: { namespace, name },
+        entityType: entityKey, initialEntityMetadata: entityMetadata, columnDefaults, googleProject, workspaceId: { namespace, name },
```
dmohs, line 372: Very nice clarity through naming here.

src/components/data/_tests/data-utils.test.js
```diff
@@ -0,0 +1,63 @@
+import { concatenateAttributeNames } from 'src/components/data/data-utils.js'
+
+
```
dmohs, line 3: If you had, for example:
```suggestion
const mkEmpty = () => []
```

src/components/data/_tests/data-utils.test.js
```diff
@@ -0,0 +1,63 @@
...
+describe('concatenateAttributeNames', () => {
+  it('works with two empty arrays', () => {
+    const attrList1 = []
+    const attrList2 = []
+    const expected = []
+    expect(concatenateAttributeNames(attrList1, attrList2)).toEqual(expected)
```
dmohs, line 9: Then you could:
```suggestion
    expect(concatenateAttributeNames(mkEmpty(), mkEmpty())).toEqual(mkEmpty())
```

src/components/data/_tests/data-utils.test.js
```diff
@@ -0,0 +1,63 @@
...
+  })
+  it('works with left empty array', () => {
+    const attrList1 = []
+    const attrList2 = ['aaa', 'bbb']
+    const expected = ['aaa', 'bbb']
+    expect(concatenateAttributeNames(attrList1, attrList2)).toEqual(expected)
```
dmohs, line 15: And even:
```suggestion
    const asAndBs = () => ['aaa', 'bbb']
    expect(concatenateAttributeNames(mkEmpty(), asAndBs())).toEqual(asAndBs())
```
Regardless of whether you value the concision, to me this reads a bit clearer.

dmohs (approved):

jordansafer (approved):

davidangb:
@dmohs @jordansafer I've added a commit, if you care to re-review.

@dmohs thank you for the assist on setting values into state, that is much better code and I've adopted it. On the other hand, I have not updated the unit tests with your dry/concise code … I like to see unit tests written out because it makes it easier for a human to see inputs and expected results without having to consult other functions - though I agree if it was runtime code it would be much better.

jordansafer (approved):

nawatts:
I suspect it might be related to this change... if you delete a column from a data table now, the values are all cleared but the column remains in the table.

davidangb:
@nawatts I can reproduce this, you are correct. I will address it

PR #2863, opened 2022-03-08
Bug Fix: Version link causing error when ComputeModal not fully loaded

cstrandw:
The root issue ended up being that the leoImages array was not fully loaded upon opening the modal, which caused the error. I added checks that the leoImages array is not empty before the id is checked, and moved the isTerraSupported and getChangelogUrl functions into the makeImageInfo function since this is the only place where those functions are used

Tested by:

- Checked that the version link was functional for all images while there was a runtime running, and also if there was not a runtime running

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -546,14 +546,17 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
   }
 
   const makeImageInfo = style => {
-    const isTerraSupported = ({ isCommunity = false, id }) => !isCommunity && !(getToolForImage(id) === tools.RStudio.label)
-    const getChangelogUrl = ({ id }) => `https://github.com/DataBiosphere/terra-docker/blob/master/${_.replace('_legacy', '', id)}/CHANGELOG.md`
+    const selectedImage = _.find({ image: selectedLeoImage }, leoImages)
+    const shouldDisable = _.isEmpty(leoImages) ? true : selectedImage.isCommunity || getToolForImage(selectedImage.id) === tools.RStudio.label
+    const changelogUrl = _.isEmpty(leoImages) ?
+      '' :
+      `https://github.com/DataBiosphere/terra-docker/blob/master/${_.replace('_legacy', '', selectedImage.id)}/CHANGELOG.md`
 
     return div({ style: { whiteSpace: 'pre', ...style } }, [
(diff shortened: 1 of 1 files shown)


jdcanas (approved):
LGTM, tested locally. I like that we moved the functions to their call-site too, makes reading the code easier :)

src/components/ComputeModal.js
```diff
@@ -545,19 +545,24 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+        disabled: leoImages.length > 0 ? !isTerraSupported(_.find({ image: selectedLeoImage }, leoImages)) : true,
+        ...Utils.newTabLinkProps
+      }, ['Version: ', version || null]),
+      h(ClipboardButton, {
+        text: selectedLeoImage,
+        style: { marginLeft: '0.5rem' },
+        tooltip: 'Copy the image version'
+      })
+    ])
+  }
```
kyuksel: Consider 

```suggestion
  const makeImageInfo = style => {
    const selectedImage = _.find({ image: selectedLeoImage }, leoImages)
    const shouldDisable = _.isEmpty(leoImages) ? true : selectedImage.isCommunity || getToolForImage(selectedImage.id) === tools.RStudio.label
    const changelogUrl = _.isEmpty(leoImages) ?
      '' :
      `https://github.com/DataBiosphere/terra-docker/blob/master/${_.replace('_legacy', '', selectedImage.id)}/CHANGELOG.md`

    return div({ style: { whiteSpace: 'pre', ...style } }, [
      div({ style: Style.proportionalNumbers }, ['Updated: ', updated ? Utils.makeStandardDate(updated) : null]),
      h(Link, {
        href: changelogUrl,
        disabled: shouldDisable,
        ...Utils.newTabLinkProps
      }, ['Version: ', version || null]),
      h(ClipboardButton, {
        text: selectedLeoImage,
        style: { marginLeft: '0.5rem' },
        tooltip: 'Copy the image version'
      })
    ])
  }
```

mainly since `isTerraSupported` and `getChangelogUrl` no longer need to be functions.

kyuksel (approved):
A suggestion inline. Can be a follow-up change too in the interest of time.

src/components/ComputeModal.js
cstrandw: Thank you! Those changes make a lot of sense to me

PR #2864, opened 2022-03-08
[DC-241] Add mixpanel events for the Data Library page

samanehsan:
Record an event in mixpanel when a user clicks any of the "browse data" buttons on the [dataset library page](https://app.terra.bio/#library/datasets). I verified that the events appear in the [dev mixpanel](https://mixpanel.com/project/2085496/view/19055/app/events#~(dateRange~(from~'!start_of_current_day~to~'!end_of_current_day~type~'between)~entries~(~(aggregationOperator~'total~aggregationOperatorPerUser~null~dataGroupId~null~event~(label~'library*3abrowseData~value~'library*3abrowseData~tags~(~)~isRecentlyUsed~true)~filters~(~)~filtersOperator~'and~property~null~type~'event))~filters~(~)~search~'~colWidths~(*2524event_name~203~*2524time~203~*2524distinct_id~203~*2524city~203~mp_country_code~203~*2524os~203)~selectedProperties~(~(value~'!event_name~resourceType~'event~label~'Event*20Name~type~'string~propertyDefaultType~'string)~(value~'!time~resourceType~'event~label~'!time~type~'string~propertyDefaultType~'string)~(value~'!distinct_id~resourceType~'event~label~'Distinct*20ID~type~'string~propertyDefaultType~'string)~(value~'!city~resourceType~'event~label~'City~type~'string~propertyDefaultType~'string)~(value~'mp_country_code~resourceType~'event~label~'Country~type~'string~propertyDefaultType~'string)~(value~'!os~resourceType~'event~label~'Operating*20System~type~'string~propertyDefaultType~'string))~showQueryBuilder~true)).

### DIFF
--- a/src/pages/library/Datasets.js
+++ b/src/pages/library/Datasets.js
@@ -109,9 +109,13 @@ const Participant = ({ logo, title, shortDescription, description, sizeText, mod
 
 const browseTooltip = 'Look for the Export to Terra icon to export data from this provider.'
 
-const captureEventOnClick = (datasetName) => {
+const captureEventMetrics = (eventName, metrics) => {
+  Ajax().Metrics.captureEvent(eventName, metrics)
+}
+
+const captureBrowseDataEvent = (datasetName) => {
   const eventName = `${Events.datasetLibraryBrowseData}`
-  Ajax().Metrics.captureEvent(eventName, {
+  captureEventMetrics(eventName, {
     datasetName: datasetName
   })
 }
@@ -127,7 +131,7 @@ const thousandGenomesHighCoverage = () => h(Participant, {
     'aria-label': 'Browse 1000 Genomes High Coverage data',
     tooltip: 'Visit the workspace',
     href: Nav.getLink('workspace-dashboard', { namespace: 'anvil-datastorage', name: '1000G-high-coverage-2019' }),
-    onClick: () => captureEventOnClick('1000G-high-coverage-2019')
+    onClick: () => captureBrowseDataEvent('1000G-high-coverage-2019')
   }, ['Browse data'])
 ])
 
@@ -146,7 +150,7 @@ const thousandGenomesLowCoverage = () => h(Participant, {
     'aria-label': 'Browse 1000 Genomes Low Coverage data',
     tooltip: browseTooltip,
     href: Nav.getLink('data-explorer-public', { dataset: '1000 Genomes' }),
-    onClick: () => captureEventOnClick('1000 Genomes')
+    onClick: () => captureBrowseDataEvent('1000 Genomes')
   }, ['Browse data'])
 ])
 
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
nice work

lu-c:
I actually really like this solution, but I thought this was looking for this button here:

Screen Shot 2022-03-08 at 7 45 54 PM

src/pages/library/Datasets.js
```diff
@@ -118,7 +130,8 @@ const thousandGenomesHighCoverage = () => h(Participant, {
   h(ButtonPrimary, {
     'aria-label': 'Browse 1000 Genomes High Coverage data',
     tooltip: 'Visit the workspace',
-    href: Nav.getLink('workspace-dashboard', { namespace: 'anvil-datastorage', name: '1000G-high-coverage-2019' })
+    href: Nav.getLink('workspace-dashboard', { namespace: 'anvil-datastorage', name: '1000G-high-coverage-2019' }),
+    onClick: () => captureBrowseDataEvent('1000G-high-coverage-2019')
```
pshapiro4broad: Would it be possible to pass in the current object and then use the `aria-label` property? (Or maybe some other way to do it that doesn't involve duplication?)

samanehsan: I could pass it in here, but I'm not sure if that would make it more readable: https://github.com/DataBiosphere/terra-ui/blob/86af332a65d3e27e9955288e9b72a619ae13cfbd/src/pages/library/Datasets.js#L442

samanehsan: What do you think about passing in just the aria-label, like `thousandGenomesHighCoverage('Browse' 1000 Genomes High Coverage data')`?

samanehsan:
@lu-c the acceptance criteria in the ticket says "Working dashboard in production with Browse Data metrics separated by dataset." Maybe the "browse data" button on the welcome page could be another event (if it's not already recorded)?

src/pages/library/Datasets.js
```diff
@@ -107,6 +109,16 @@ const Participant = ({ logo, title, shortDescription, description, sizeText, mod
...
+const captureEventMetrics = (eventName, metrics) => {
+  Ajax().Metrics.captureEvent(eventName, metrics)
+}
+
+const captureBrowseDataEvent = (datasetName) => {
+  const eventName = `${Events.datasetLibraryBrowseData}`
+  captureEventMetrics(eventName, {
+    datasetName: datasetName
+  })
+}
```
dmohs: Equivalent, and doesn't swallow the promise returned by captureEvent (which may be useful to the caller):
```suggestion
const captureBrowseDataEvent = datasetName =>
  Ajax().Metrics.captureEvent(Events.datasetLibraryBrowseData, {datasetName})
```

dmohs (approved):

petesantos (approved):

PR #2865, opened 2022-03-09
[IA-3011][IA-3043] Integration testing for creating an interactive analysis and Analysis Modal improvements

kyuksel:
Thank you @cahrens for the very thorough reviews and great ideas, and thank you @petesantos for your help with flake testing.

JIRA ticket: [IA-3011](https://broadworkbench.atlassian.net/browse/IA-3012)

This PR:
1. adds a new test named `create-interactive-analysis` which tests one of the flows that we are planning on releasing as part of the `Analysis Tab` launch. The term `analysis` refers to artifacts like Jupyter notebook (`.ipynb`) and RStudio markdown (`.Rmd`) files, which also includes:
- a test utility for clicking on images with alt text
- a test utility to override config which is helpful for turning on feature flags
2. makes usability improvements to `AnalysisModal` which include:
  - adding `alt text` for the `Jupyter`, `RStudio`, `Galaxy` and `Cromwell` logo images so the test can select and click on them using those `alt text`s. With this change, `AnalysisModal` can be keyboard-navigated and screen readers can inform users about what each logo does
  - converting `div`s into `Clickable`s which allow for mousing over to show a hand icon
  - colour changing when hovering
  - clearer UX around disabled 'button's
3. minor reformattings

Screen recording that (roughly) goes through the steps the test takes:
https://user-images.githubusercontent.com/5438223/157454640-f59ca31a-89ad-49c3-935c-25bb5d68b9f1.mov

Tested locally by running the test individually. Couldn't get reliable runs with `flake-checker`, likely due to set-up issues. Getting help from @petesantos to run it in his environment. No test issues reported.


src/pages/workspaces/workspace/Data.js
```diff
@@ -583,8 +583,9 @@ const WorkspaceData = _.flow(
                       buttonStyle: { borderBottom: 0, height: 40, ...(canCompute ? {} : { color: colors.dark(0.25) }) },
                       tooltip: canCompute ?
                         tableName ? `${tableName} (${count} row${count === 1 ? '' : 's'})` : undefined :
-                        [div({ key: `${tableName}-tooltip`, style: { whiteSpace: 'pre-wrap' } }, 'You must be an owner, or a writer with compute permission, to view this snapshot.\n\n' +
-                        'Contact the owner of this workspace to change your permissions.')],
+                        [div({ key: `${tableName}-tooltip`, style: { whiteSpace: 'pre-wrap' } },
+                          'You must be an owner, or a writer with compute permission, to view this snapshot.\n\n' +
```
kyuksel, line 587: Reformatting only

src/components/AnalysisModal.js
```diff
@@ -84,7 +84,8 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
           [isToolAnApp(currentTool) && !app, () => setViewMode(environmentMode)],
           [isToolAnApp(currentTool) && !!app, () => {
             console.error(
-              `This shouldn't be possible, as you aren't allowed to create a ${_.capitalize(app.appType)} instance when one exists; the button should be disabled.`)
+              `This shouldn't be possible, as you aren't allowed to create a ${_.capitalize(
+                app.appType)} instance when one exists; the button should be disabled.`)
```
kyuksel, line 88: Reformatting only

src/components/AnalysisModal.js
```diff
@@ -156,19 +157,19 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
           setCurrentTool(tools.Jupyter.label)
           enterNextViewMode(tools.Jupyter.label)
         }
-      }, [img({ src: jupyterLogoLong, style: _.merge(styles.image, { width: '30%' }) })]),
+      }, [img({ src: jupyterLogoLong, alt: 'Create new notebook', style: _.merge(styles.image, { width: '30%' }) })]),
```
cahrens, line 163: It is true that the images need alt text for accessibility. However, these controls are not keyboard accessible-- to see that, try selecting one of them using the keyboard only (using Tab to move between components). Please fix https://broadworkbench.atlassian.net/browse/IA-3043 at the same time.

I would recommend that the components are buttons containing the images, which have alt text. You don't need to have labels on the buttons as long as the alt text is clear.

And it would be good to pull the name of the app/tool from the `tools` variable, so that we aren't duplicating app names/tool names here.

cahrens: Please also add alt text to the Cromwell "button" (line 178).

integration-tests/tests/create-interactive-analysis.js
```diff
@@ -0,0 +1,40 @@
...
+  await overrideConfig(page, { isAnalysisTabVisible: true })
+
+  await click(page, clickable({ textContains: 'View Workspaces' }))
+  await signIntoTerra(page, token)
+  await dismissNotifications(page)
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: workspaceName })) })
+  await click(page, navChild('analyses'))
+  await click(page, clickable({ textContains: 'Create' }))
+  await findElement(page, getAnimatedDrawer('Select an application'))
+  await click(page, image({ textContains: 'Create new notebook' }))
```
cahrens: Clicking an image is a strange thing-- images themselves are not clickable/focusable. You will be clicking on the button containing the image. Once you have an element that is actually focusable, we can figure out how best to express that.

src/components/AnalysisModal.js
kyuksel: Addressed in [ff9583c](https://github.com/DataBiosphere/terra-ui/pull/2865/commits/ff9583c967be11c71edcd96c97a4a0273a22e06e)

src/components/AnalysisModal.js
```diff
@@ -148,33 +149,34 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
...
-      div({
+      }, [img({ src: rstudioBioLogo, alt: 'Create new R markdown file', style: styles.image })]),
+      h(Clickable, {
         style: { opacity: galaxyApp ? '0.5' : '1', ...styles.toolCard }, onClick: () => {
           setCurrentTool(tools.galaxy.label)
           enterNextViewMode(tools.galaxy.label)
         }, disabled: !galaxyApp, title: galaxyApp ? 'You already have a galaxy environment' : ''
-      }, [img({ src: galaxyLogo, style: _.merge(styles.image, { width: '30%' }) })]),
+      }, [img({ src: galaxyLogo, alt: 'Create new Galaxy app', style: _.merge(styles.image, { width: '30%' }) })]),
       !tools.cromwell.isAppHidden && div({
```
cahrens: Please make this a Clickable too. And then I think you can link [IA-3043](https://broadworkbench.atlassian.net/browse/IA-3043) to this PR as well.

integration-tests/utils/integration-utils.js
```diff
@@ -37,7 +37,7 @@ const getClickablePath = (path, text, textContains, isDescendant = false) => {
   if (text) {
     return `${base}[normalize-space(.)="${text}" or @title="${text}" or @aria-label="${text}" or @aria-labelledby=//*[normalize-space(.)="${text}"]/@id]`
   } else if (textContains) {
-    return `${base}[contains(normalize-space(.),"${textContains}") or contains(@title,"${textContains}") or contains(@aria-label,"${textContains}") or @aria-labelledby=//*[contains(normalize-space(.),"${textContains}")]/@id]`
+    return `${base}[contains(normalize-space(.),"${textContains}") or contains(@title,"${textContains}") or contains(@alt,"${textContains}") or contains(@aria-label,"${textContains}") or @aria-labelledby=//*[contains(normalize-space(.),"${textContains}")]/@id]`
```
cahrens, line 40: I think we leave out the alt text since we are doing the separate img method.

kyuksel: That piece is needed actually because `image` function still calls `getClickablePath` and the resulting XPath doesn't match the alt text without `contains(@alt,"${textContains}")`.

src/components/AnalysisModal.js
kyuksel: Ugh…I don't know why I keep missing the Cromwell one.
Just pushed a commit and linked to the other ticket.

integration-tests/utils/integration-helpers.js
```diff
@@ -180,13 +180,17 @@ const withRegisteredUser = test => withUser(async options => {
   await test(options)
 })
 
+const overrideConfig = async (page, config) => {
+  await page.evaluate(passedInConfig => window.configOverridesStore.set(passedInConfig), config)
```
petesantos: nit: Is this more readable, or just cruft?

```suggestion
const overrideConfig = async (page, configToPassIn) => {
  await page.evaluate(passedInConfig => window.configOverridesStore.set(passedInConfig), configToPassIn)
```

kyuksel: I don't feel strongly either way so made the change in [4baa27c2](https://github.com/DataBiosphere/terra-ui/pull/2865/commits/4baa27c20c239185ec2a99c1fc6a69b2f8fc2024).

src/components/AnalysisModal.js
```diff
@@ -148,33 +149,34 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
...
           enterNextViewMode(tools.RStudio.label)
         }
-      }, [img({ src: rstudioBioLogo, style: styles.image })]),
-      div({
+      }, [img({ src: rstudioBioLogo, alt: 'Create new R markdown file', style: styles.image })]),
+      h(Clickable, {
         style: { opacity: galaxyApp ? '0.5' : '1', ...styles.toolCard }, onClick: () => {
           setCurrentTool(tools.galaxy.label)
           enterNextViewMode(tools.galaxy.label)
         }, disabled: !galaxyApp, title: galaxyApp ? 'You already have a galaxy environment' : ''
```
cahrens: For some reason the Galaxy and Cromwell buttons are disabled when I have nothing in my Workspace. And there is no indication of **why** they are disabled (the title is not displayed… I'm not sure how it is supposed to be).

If you don't want to fix this bug, please create a ticket for it. I do recall though that these buttons USED to be enabled, so it feels like something odd is going on (you should at least verify that this is not a regression).

cahrens: For some reason the Galaxy and Cromwell buttons are disabled when I have nothing in my Workspace. And there is no indication of **why** they are disabled (the title is not displayed… I'm not sure how it is supposed to be).

If you don't want to fix this bug, please create a ticket for it. I do recall though that these buttons USED to be enabled, so it feels like something odd is going on.

integration-tests/tests/create-interactive-analysis.js
```diff
@@ -0,0 +1,40 @@
...
+  await fillIn(page, input({ placeholder: 'Enter a name' }), notebookName)
+  await select(page, 'Language', 'Python 3')
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Create Analysis' })) })
+  await findText(page, 'A cloud environment consists of application configuration, cloud compute and persistent disk(s).')
+})
+const testCreateInteractiveAnalysis = {
+  name: 'create-interactive-analysis',
+  fn: testCreateInteractiveAnalysisFn,
+  timeout: 20 * 60 * 1000
```
cahrens: Why is it necessary to change the default timeout? Given that this test isn't actually creating a runtime, it should be fast.

cahrens: Why is it necessary to change the default timeout? Given that this test isn't actually creating a runtime, it should be fast.

integration-tests/utils/integration-utils.js
cahrens: I'm still feeling uneasy about this because images are not generally clickable. I'd like to see what some other folks things-- @petesantos , @zarsky-broad ?

cahrens: I'm still feeling uneasy about this because images are not generally clickable. I'd like to see what some other folks think-- @petesantos , @zarsky-broad ?

cahrens: I also think that if you are going to make this change, there is really no point in having a separate `image` method.

If we keep the separate image method, then I think it should not rely on `getClickablePath`.

cahrens: And one more comment, you only changed the "textContains" path to handle the alt text.

xuscrbw: Just as a functionality question, is the alt text something we could potentially request on specific components/elements or is it solely for accessibility right now?

cahrens: alt text is specific to images, and it is required for screen readers to interpret images.

All images ​must​ include an ​alt​ attribute. There are some cases when an empty ​alt​ attribute (​alt=""​) is desirable (if the image is purely decorative and there is already text on the page that describes what the image is displaying, for example if there is a button with both a word and an image). However, the ​alt​ attribute must never be omitted. An empty alt attribute has special significance to screen reader users. When a screen reader encounters an image with an empty alt attribute, it remains silent. It essentially skips over the image as if it weren't there. If an author fails to include an alt attribute at all, the screen reader will attempt a series of fallback behaviors in an effort to give the user any helpful information it can about the image. The last of these fallbacks is to read the path to the image. This might actually be helpful if the path to an image that links to the "About us" page is www.example.com/aboutus.gif. However, that is rarely the case. It is often something like www.example.com/theme/thYJ3JIeKD79_45_2H.jpg, which isn't very helpful at all.

src/components/AnalysisModal.js
```diff
@@ -148,33 +149,34 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
...
-      }, [img({ src: galaxyLogo, style: _.merge(styles.image, { width: '30%' }) })]),
-      !tools.cromwell.isAppHidden && div({
+      }, [img({ src: galaxyLogo, alt: 'Create new Galaxy app', style: _.merge(styles.image, { width: '30%' }) })]),
+      !tools.cromwell.isAppHidden && h(Clickable, {
```
xuscrbw, line 179: Could you explain why we do the div -> h(clickable change?

xuscrbw (approved):
lgtm

src/components/AnalysisModal.js
cahrens: A `div` is not keyboard focusable. Therefore, there is no way for someone to "click" it without a mouse. You can see this by attempting to use those components in prod via the keyboard only. 

One should never have an `onClick` method on a `div` without appropriate aria labels and non-zero tabIndex (which `Clickable` adds). Styling something as a button does not make it actually a button as far as screen readers/keyboard accessibility. (If one uses an actual HTML button element, there is no need to use aria labels and everything "just works". However, that is not how our shared components are designed in this repo.).

This problem was described in IA-3043.

Note that there are still usability issues here-- normally "buttons" have some visual indication when they are hovered over to indicate that they can be clicked. The only indication with this change is that the pointer changes to a finger (this is better than the situation in prod, where there is zero indication).

petesantos (approved):
I ran this test on the flake checker - only one failure on workspace creation.
I did see failures on run-notebooks - but they were general flakiness issues.

cahrens (changes requested):
Please address my comments before merging. I'm very concerned about Galaxy and Cromwell buttons not functioning.

src/components/AnalysisModal.js
xuscrbw: Thank you for the explanation!

integration-tests/utils/integration-utils.js
cahrens: @kyuksel since the button with just the image w/alt text is sounding good with the screen reader, I'm OK with adding to `getClickablePath`. However, you need to also modify the "text" conditional… in fact, that's actually the one you would be using since you are providing the full alt text.

FWIW, I would still prefer we don't create a separate image method, but that's up to you.

zarsky-broad: I strongly agree with @cahrens - just because it works doesn't mean it's a good idea. The more we follow semantic best practices, the less we'll have to build (and maintain!) custom tooling for our workarounds.

cahrens: I think at this point it comes down to UX. Do we want "buttons" with only images on them, or should there be text on the button as well as the image (in which case the image is decorative and should have an empty alt attribute so that the information is not read out twice via the screen reader)? In the case of these "buttons", there is also the UX issue that they do not have a visible hover state. I really expected that this would have all been sorted out through the process of user testing, as the controls just feel really weird, but given that it hasn't been sorted out, I can create a ticket for these issues.

kyuksel: Thank you for the very thorough reviews, @cahrens. Please do create tickets as needed. The team's goal _is_ to iterate and improve Analysis Tab UX 🤞🏼 

Addressed XPath stuff in [66bb9d7](https://github.com/DataBiosphere/terra-ui/pull/2865/commits/66bb9d75112ddd1ac43c9a74ab9d55f1685fad17).

src/components/AnalysisModal.js
kyuksel: I've made some improvements in consultation with UX, which hopefully address your comment above. If they don't, please create a ticket or add to one you may have already. Thank you.

integration-tests/tests/create-interactive-analysis.js
kyuksel: It isn't. It was my mindless copy/pasta'ing. Thanks for the catch!

kyuksel: Addressed in [PR feedback: Use default test timeout](https://github.com/DataBiosphere/terra-ui/pull/2865/commits/1dfa47f49fb1db69bc3248ff5d3cd54ddf3b7ca2).

cahrens (approved):
Thanks for the improvements @kyuksel!! This is much more usable now.

src/components/AnalysisModal.js
cahrens: This is now working-- thank you!

integration-tests/utils/integration-utils.js
cahrens: This is better with the hover color. Thanks!

PR #2868, opened 2022-03-09
[DC-224] Added fields to the workspace details page.

lu-c:
Added a section for the data source, and a section for the legacy dataset attributes.

Both of these fields will only show up if they exist on the dataset, so uncomment 245-267 on the dataBrowser-details.js file to see the changes.

According to the ticket, this is just a preliminary "skeleton" start, since the data format and the fields are not yet determined, so please keep that in mind while reviewing this PR! :)

### DIFF
--- a/src/pages/library/DataBrowserDetails.js
+++ b/src/pages/library/DataBrowserDetails.js
@@ -1,6 +1,7 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
 import { div, h, h1, h2, h3, span, table, tbody, td, tr } from 'react-hyperscript-helpers'
+import Collapse from 'src/components/Collapse'
 import { ButtonOutline, ButtonPrimary, ButtonSecondary, Link } from 'src/components/common'
 import FooterWrapper from 'src/components/FooterWrapper'
 import { centeredSpinner, icon } from 'src/components/icons'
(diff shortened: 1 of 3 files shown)


src/pages/library/DataBrowserDetails.js
```diff
@@ -58,142 +57,218 @@ const MainContent = ({ snapshot }) => {
...
+        _.flatMap(([key, value]) => {
+          return div({ style: { width: '30%', margin: '0 10px' } }, [
+            h3({ style: { textTransform: 'capitalize' } }, [
+              key.split(/(?=[A-Z])/).join(' ')
+            ]),
+            div([value])
+          ])
+        })
+      )(dataObj.legacyDatasetAttributes)
+      // _.map(attr => {
```
cahrens: I would strongly prefer that we do not check in commented out or non-functional code. It's great to iteratively develop features, but IMO the things that are checked into dev should be working. In progress work can be kept on a branch. Otherwise it is very easy for cruft to build up as things are inadvertently left behind.

samanehsan (approved):
This looks awesome! :octocat:

src/pages/library/DataBrowserDetails.js
samanehsan: I think we can leave this as a feature branch since we don't have the data format finalized

src/pages/library/DataBrowserDetails.js
```diff
@@ -58,142 +57,218 @@ const MainContent = ({ snapshot }) => {
...
 }
 
 const DataBrowserDetails = ({ id }) => {
-  const [showRequestAccessModal, setShowRequestAccessModal] = useState()
   const { dataCatalog } = useDataCatalog()
   const dataMap = _.keyBy('dct:identifier', dataCatalog)
-  const snapshot = dataMap[id]
+  const dataObj = dataMap[id]
+
+  // if (dataObj) {
```
petesantos: How would you feel about adding this object to the [ajax request](https://github.com/DataBiosphere/terra-ui/blob/5b00dbf78f3b55a2d09a00c6e1af85db528f8d25/src/libs/ajax.js#L1030) under the modulo operation?

lu-c: went ahead and did this - probably better to keep all that type of stuff in the same place!

petesantos (approved):

PR #2869, opened 2022-03-10
[WOR-44] Show most expensive workspaces in Spend Report

cahrens:
This total spend by workspace (for 10 most expensive workspaces) for a billing project under the following conditions:

1. User is billing project owner
2. User is part of the alpha spending group. "hermione.owner" is a member in dev.
3. The billing project has spend data stored. `galaxy-anvil-dev` is a good one to look at. http://localhost:3000/#billing?selectedName=galaxy-anvil-dev&type=project&tab=spend%20report

Also adds Highcharts attribution per WOR-141.

image

### DIFF
--- a/integration-tests/tests/billing-projects.js
+++ b/integration-tests/tests/billing-projects.js
@@ -17,16 +17,34 @@ const billingProjectsPage = (testPage, testUrl) => {
 
     setSpendReportDays: async days => await select(testPage, 'Date range', `Last ${days} days`),
 
-    assertText: async expectedText => await findText(testPage, expectedText)
+    assertText: async expectedText => await findText(testPage, expectedText),
+
+    assertChartValue: async (expectedNumber, expectedText) => {
+      // This checks the accessible text for chart values.
+      await testPage.waitForXPath(`(//*[@role="img"])[contains(@aria-label,"${expectedNumber}. ${expectedText}")]`)
+    }
   }
 }
 
-const setAjaxMockValues = async (testPage, ownedBillingProjectName, spendCost) => {
+const setAjaxMockValues = async (testPage, ownedBillingProjectName, spendCost, numExtraWorkspaces = 0) => {
   const spendReturnResult = {
     spendSummary: {
       cost: spendCost, credits: '2.50', currency: 'USD', endTime: '2022-03-04T00:00:00.000Z', startTime: '2022-02-02T00:00:00.000Z'
-    }
+    },
+    spendDetails: [{
+      aggregationKey: 'Workspace',
+      spendData: [
+        { cost: '100', workspace: { name: 'Second Most Expensive Workspace' } },
+        { cost: '1000', workspace: { name: 'Most Expensive Workspace' } },
+        { cost: '10', workspace: { name: 'Third Most Expensive Workspace' } }
+      ]
+    }]
   }
+
+  _.forEach(() => {
+    spendReturnResult.spendDetails[0].spendData.push({ cost: '0.01', workspace: { name: 'Extra Inexpensive Workspace' } })
+  }, _.range(0, numExtraWorkspaces))
+
   const projectListResult = [{
     projectName: ownedBillingProjectName,
     billingAccount: 'billingAccounts/fake-id', invalidBillingAccount: false, roles: ['Owner'], status: 'Ready'
@@ -68,11 +86,19 @@ const testBillingSpendReportFn = withUserToken(async ({ page, testUrl, token }) 
   await billingPage.visit()
   await billingPage.selectSpendReport(billingProjectName)
   await billingPage.assertText('$90.13')
+  // Check that chart loaded, and workspaces are sorted by cost.
+  await billingPage.assertText('Spend By Workspace')
+  await billingPage.assertChartValue(1, 'Most Expensive Workspace, $1,000')
+  await billingPage.assertChartValue(2, 'Second Most Expensive Workspace, $100')
+  await billingPage.assertChartValue(3, 'Third Most Expensive Workspace, $10')
 
-  // Change the returned mock cost to mimic different date ranges
-  await setAjaxMockValues(page, billingProjectName, '135')
+  // Change the returned mock cost to mimic different date ranges.
+  await setAjaxMockValues(page, billingProjectName, '135', 20)
   await billingPage.setSpendReportDays(90)
   await billingPage.assertText('$135.00')
+  // Check that title updated to reflect truncation.
+  await billingPage.assertText('Top 10 Spending Workspaces')
+  await billingPage.assertChartValue(10, 'Extra Inexpensive Workspace')
 })
 
 const testBillingSpendReport = {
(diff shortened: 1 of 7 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -288,7 +288,7 @@ const BucketContent = _.flow(
       )
     ]),
     div({ style: { margin: '1rem -1rem 1rem -1rem', borderBottom: `1px solid ${colors.dark(0.25)}` } }),
-    div({ style: { flex: '1 1 0', overflow: 'auto' }}, [
+    div({ style: { flex: '1 1 0', overflow: 'auto' } }, [
```
cahrens, line 291: I was getting a lint warning for this. I don't know how it crept in.

src/pages/billing/Project.js
```diff
@@ -305,23 +356,26 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
-          }), [7, 30, 90]),
-          onChange: ({ value: selectedDays }) => {
-            setSpendReportLengthInDays(selectedDays)
-            setTotalCost(null) // This will force the report to be recalculated based on selectedDays
-          }
-        })
-      ])])]),
-      CostCard({ title: 'Total spend', amount: (!!totalCost ? totalCost : '$__.__') })
+    [spendReportKey]: div({ style: { display: 'grid', rowGap: '1.25rem' } }, [
+      div({ style: { display: 'grid', gridTemplateColumns: 'minmax(15.625rem, max-content)', rowGap: '1.25rem' } }, [
```
cahrens, line 356: The first chunk is whitespace difference only.

src/pages/billing/Project.js
```diff
@@ -305,23 +356,26 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+            }), [7, 30, 90]),
+            onChange: ({ value: selectedDays }) => {
+              setSpendReportLengthInDays(selectedDays)
+              setTotalCost(null) // This will force the report to be recalculated based on selectedDays
+            }
+          })
+        ])])]),
+        CostCard({ title: 'Total spend', amount: (!!totalCost ? totalCost : '$__.__') })
+      ]),
+      div({ style: { gridRowStart: 2 } }, [h(HighchartsReact, { highcharts: Highcharts, options: spendChartOptions })])
```
cahrens: The chart is new.

src/pages/billing/Project.js
```diff
@@ -233,6 +262,23 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
       const spend = await Ajax(signal).Billing.getSpendReport({ billingProjectName: billingProject.projectName, startDate, endDate })
       const costFormatter = new Intl.NumberFormat(navigator.language, { style: 'currency', currency: spend.spendSummary.currency })
       setTotalCost(costFormatter.format(spend.spendSummary.cost))
+
+      const workspaceDetails = _.find(details => details.aggregationKey === 'Workspace')(spend.spendDetails)
+      console.assert(workspaceDetails !== undefined, 'Spend report details do not include aggregation by Workspace')
```
cahrens, line 268: If this assertion fails, we will not render the chart, but the user will see no error (that's the magic of `workspaceDetails?.spendData`). The error will appear in the JavaScript console.

integration-tests/tests/billing-projects.js
```diff
@@ -17,16 +17,34 @@ const billingProjectsPage = (testPage, testUrl) => {
 
     setSpendReportDays: async days => await select(testPage, 'Date range', `Last ${days} days`),
 
-    assertText: async expectedText => await findText(testPage, expectedText)
+    assertText: async expectedText => await findText(testPage, expectedText),
+
+    assertChartValue: async (expectedNumber, expectedText) => {
+      // This checks the accessible text for chart values.
+      await testPage.waitForXPath(`(//*[@role="img"])[contains(@aria-label,"${expectedNumber}. ${expectedText}")]`)
```
aherbst-broad, line 24: I like this assertion

marctalbott (approved):
I think it looks great. Only thing I found while testing was that the chart doesn't shrink well when the window is resized. Making the window bigger works great -- the chart expands to fit the new window size. However, it doesn't shrink as you shrink your window. Instead, it juts off the side of the page and you have to scroll horizontally to see it. Screenshot attached. It's kind of hard to tell what's going on, but the grey vertical line is where the background of the spend report stops.
Screen Shot 2022-03-10 at 2 41 35 PM

Approving as I'm not sure if this is a big deal or if it's even something we can fix or just a bug in Highcharts

src/pages/billing/Project.js
```diff
@@ -233,6 +262,23 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+      // Get the most expensive workspaces, sorted from most to least expensive.
+      const mostExpensiveWorkspaces = _.flow(
+        _.sortBy(({ cost }) => { return parseFloat(cost) }),
```
aherbst-broad, line 271: 🤔 We should ticket up API-side sorting + limiting for the spend detail array if we haven't already.

aherbst-broad (approved):
Looks straightforward.

cahrens:
> I think it looks great. Only thing I found while testing was that the chart doesn't shrink well when the window is resized. Making the window bigger works great -- the chart expands to fit the new window size. However, it doesn't shrink as you shrink your window. Instead, it juts off the side of the page and you have to scroll horizontally to see it. Screenshot attached. It's kind of hard to tell what's going on, but the grey vertical line is where the background of the spend report stops.

Thanks for nothing that @marctalbott. I set a min size on the chart (smaller than it will actually be) so that it will shrink as you resize the window. Note though that the parent div will only shrink to a certain point though, and after that you get the horizontal scrollbar. This behavior is the same on the other tabs as well. I believe that we set a minimum size across Terra.

image

src/pages/billing/Project.js
cahrens: I guess… I can certainly see sorting on the server side, but I'm not sure about limiting. If we limit, I feel like there has to be a parameter to specify the limit.

I think ultimately we are going to allow users to select the Workspaces they want to see, and in that case I am going to need all the Workspaces anyway, and I will filter what I expose to the chart.

src/pages/billing/Project.js
dmohs: I got the impression from your PR description that this could be fairly common, thus noisy. Would it be better to display this information in place of the chart so that if I were expecting to see spending details I would have some indication of why they are not there?

dmohs (approved):

src/pages/billing/Project.js
cahrens: No, this is not common. This would only happen if the API response was not formatted as we expected.

In the case of billing data not existing at all, we wouldn't be getting to this point.

cahrens:
FYI to reviewers, I made a couple small changes in consultation with Adam (PO), and he verified them from the PR site.

PR #2848, opened 2022-03-02
[SUP-627] Fix scrolling in files tab

nawatts:
#2837 broke the ability to scroll through long lists of files. This restores it.

Before:
Screen Shot 2022-03-02 at 2 58 48 PM

After:
Screen Shot 2022-03-02 at 2 58 11 PM

PR #2870, opened 2022-03-10
Fail CI on ESLint warnings

nawatts:
Currently, our custom ESLint rules are configured as warnings instead of errors.

https://github.com/DataBiosphere/terra-ui/blob/6af15805aeed7adc03be7101867d5eb501654b42/.eslintrc.js#L5-L94

Thus, violations of these rules can still be committed because ESLint run in CI only fails on errors. This can add unrelated churn to PRs when other devs notice the warnings and resolve them (for example: https://github.com/DataBiosphere/terra-ui/pull/2862#discussion_r822077233, https://github.com/DataBiosphere/terra-ui/pull/2848#discussion_r823743234).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -195,7 +195,7 @@ jobs:
           paths:
             - node_modules
             - .yarn/cache
-      - run: yarn eslint src
+      - run: yarn eslint --max-warnings=0 src
       - run: yarn test
       - run: DISABLE_ESLINT_PLUGIN=true yarn build # already linted above
       - run: tar -czf build.tgz .gcloudignore app.yaml build config
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   "scripts": {
     "analyze": "yarn build && source-map-explorer 'build/static/js/*.js' --gzip",
     "build": "env $(yarn setenv) react-app-rewired build && rm build/config.json",
-    "lint": "eslint src --fix",
+    "lint": "eslint src --fix --max-warnings=0",
     "optimize-image-svgs": "svgo --enable=inlineStyles,prefixIds --config '{ \"plugins\": [ { \"inlineStyles\": { \"onlyMatchedOnce\": false } }] }' --pretty -f src/images -r -p 1 --multipass",
     "preinstall": "node .hooks/check-engine-light.js",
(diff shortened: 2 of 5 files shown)


cahrens (approved):
Thanks @nawatts !

.circleci/config.yml
```diff
@@ -195,7 +195,7 @@ jobs:
           paths:
             - node_modules
             - .yarn/cache
-      - run: yarn eslint src
+      - run: yarn eslint --max-warnings=0 src
```
kyuksel, line 198: @zarsky-broad is this the place we could append `integration-test/src` so test files are linted as well?

```suggestion
      - run: yarn eslint --max-warnings=0 src integration-test/src
```

I also recall seeing another place to update. Was it this [package.json](https://github.com/DataBiosphere/terra-ui/blob/3ada1b41e239c7c954c8ef1e69c68bc218a6ef74/package.json#L62) line?

nawatts: If we're adding to the files checked by ESLint, do we want to stop at `integration-tests/src`? Or check all JS files in the project?

cahrens: There was concern about things taking a long time if we are linting dependency libraries. I would say lint as much of our own code as you can without introducing speed issues.

zarsky-broad: It seemed like we could just put `integration-tests` and it didn't take too long when we tried it

zarsky-broad: (or we could be grownups and add an .eslintignore file for node modules and the yarn cache)

zarsky-broad (approved):
I'm preemptively thumbing this because it's an improvement as is, but please also add it to the script in `package.json`, and feel free to go beyond that as feels appropriate.

.circleci/config.yml
nawatts: I tried a few runs on my machine. With node_modules and minified JS files ignored, running ESLint on the entire project vs src added ~0.5s to its ~7s second runtime.

kyuksel: FWIW I'd be happy to take that penalty 😊

nawatts: Great. I'll go ahead and merge this PR (to reduce the chance of additional warnings getting committed to dev while this stays open) and open another one to check everything with ESLint.

nawatts: Opened #2873.

PR #2875, opened 2022-03-11
[BW-1143] Update ansi-html

kpierre13:
<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3228,9 +3228,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"ansi-html@npm:0.0.7":
-  version: 0.0.7
-  resolution: "ansi-html@npm:0.0.7"
+"ansi-html@npm:0.0.8":
+  version: 0.0.8
+  resolution: "ansi-html@npm:0.0.8"
   bin:
     ansi-html: ./bin/ansi-html
   checksum: 9b839ce99650b4c2d83621d67d68622d27e7948b54f7a4386f2218a3997ee4e684e5a6e8d290880c3f3260e8d90c2613c59c7028f04992ad5c8d99d3a0fcc02c
@@ -15953,7 +15953,7 @@ resolve@^2.0.0-next.3:
   version: 3.11.1
   resolution: "webpack-dev-server@npm:3.11.1"
   dependencies:
-    ansi-html: 0.0.7
+    ansi-html: 0.0.8
     bonjour: ^3.5.0
     chokidar: ^2.1.8
     compression: ^1.7.4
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -3228,7 +3228,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"ansi-html@npm:0.0.7, ansi-html@npm:^0.0.7":
+"ansi-html@npm:0.0.7":
```
cahrens: I see that the 0.0.7 version is still used by "webpack-dev-server@npm:3.11.1", which should be fine since it is the dev server.

cahrens (approved):
LGTM, but I did not run with it.

yarn.lock
cahrens: But this means that the dependabot alert will need to be dismissed, as it will still complain about this usage.

cjllanwarne (approved):
LGTM assuming you can get the tests to pass 🤞

yarn.lock
```diff
@@ -15944,7 +15953,7 @@ resolve@^2.0.0-next.3:
   version: 3.11.1
   resolution: "webpack-dev-server@npm:3.11.1"
   dependencies:
-    ansi-html: 0.0.7
+    ansi-html: 0.0.8
```
cahrens, line 15956: @kpierre13 how did you get it to make this update? When I deleted the whole "webpack-dev-server@npm:3.11.1": block, it did not update in ansi-html after I did `yarn install` (following the directions in https://github.com/DataBiosphere/terra-ui/wiki/Security-and-Maintenance). I don't think we can just edit dependency versions ourselves, as the libraries define what versions they define on (and theoretically there could be an incompatibility with 0.0.8).

cahrens (changes requested):

yarn.lock
cahrens: In fact, when I go back to what is on dev and either follow the documented process or run `yarn up -R ansi-html` as suggested by @aednichols, neither of these usages update. So I don't think the upgrade can be done without upgrading the libraries that depend on ansi-html.

See the wiki document for options in this case.

kpierre13: I also ran that command and didn't see any changes, so I assumed that I must have to change it manually. Are the circleci checks not in place for version changes? How would I go about testing to make sure that these version upgrades don't break anything?

zarsky-broad: Notably, both instances of this are only used for serving a local dev environment, so we could close this warning as not used. Upgrading would require going to the new major version of `react-scripts`, which is a nontrivial undertaking.

cahrens: Correct, and this is documented in the wiki page (**A lot of the Dependabot alerts can be safely dismissed. These are all valid reasons for dismissing the alerts...**).

kpierre13: I understand. So if everyone agrees, I'm going to close this PR with the resolution to simply dismiss this alert due to its non-usage in production.

PR #2877, opened 2022-03-11
[IA-2954] Added dropdown to select SSD as PD

cstrandw:
- Added select dropdown to allow the user to choose the type of persistent disk on both the `ComputeModal` and `GalaxyModal`
- Added enum `pdTypes` to `runtime-utils.js` to keep track of valid pd types and display names
- Disabled disk type dropdown after runtime has been created 
- Updated `regionToPrices` object in `machines.js` file to include SSD values as per [this script](https://github.com/broadinstitute/dsp-scripts/blob/master/terra-ui/regionality/parse_pricing_api.py)
- Refactored cost calculation functions to accommodate `diskType` to get correct values from the `regionToPrices` object

Tested by:
- Created runtimes with standard and SSD persistent disks, and checked that the expected values were returned in API calls and displayed in UI
- Checked that the disk size could be updated with an SSD disk after runtime creation
- Created Galaxy apps with both standard and SSD disks, and checked that the expected values were returned in API calls and displayed in UI

Screenshots:
Screen Shot 2022-03-31 at 10 24 05 AM
Screen Shot 2022-03-31 at 10 23 35 AM
Screen Shot 2022-03-31 at 10 43 09 AM


src/components/ComputeModal.js
```diff
@@ -1051,7 +1065,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
             value: computeConfig.autopauseThreshold,
             hidden: !isAutopauseEnabled(computeConfig.autopauseThreshold),
             tooltip: !isAutopauseEnabled(computeConfig.autopauseThreshold) ? 'Autopause must be enabled to configure pause time.' : undefined,
-            onChange: updateComputeConfig('autopauseThreshold')
+            onChange: () => updateComputeConfig('autopauseThreshold')
```
kyuksel: I know we talked offline but submitting this review just so the PR triage automation doesn't consider the PR not-reviewed :)
```suggestion
            onChange: updateComputeConfig('autopauseThreshold')
```

kyuksel: I know we talked offline but submitting this review just so the PR triage automation doesn't consider the PR not-reviewed :)
```suggestion
            onChange: updateComputeConfig('autopauseThreshold')
```

kyuksel (commented):
Talked offline about adding the SSD option to `Galaxy Modal` as well.

kyuksel:
Let's also update the following three lines in `machine.js`

```
// There is a script available in https://gist.github.com/wnojopra/be5be3ab7e6a09a351b89085992a940d
// that parses the API and outputs this data.
// TODO: Update the above URL in the comment to the proper script location
```

as

```
// There is a script available at https://github.com/broadinstitute/dsp-scripts/blob/master/terra-ui/regionality/parse_pricing_api.py.
// It parses the API and outputs the response data.
```

now that the script PR is merged  😊

kyuksel:
Let's also update the following three lines in `machine.js`

```
// There is a script available in https://gist.github.com/wnojopra/be5be3ab7e6a09a351b89085992a940d
// that parses the API and outputs this data.
// TODO: Update the above URL in the comment to the proper script location
```

as

```
// There is a script available at https://github.com/broadinstitute/dsp-scripts/blob/master/terra-ui/regionality/parse_pricing_api.py.
// It parses the API and outputs the response data.
```

now that the script PR is merged  😊

kyuksel:
I
1. created a standard (GCE) VM with SSD PD
2. kept the SSD PD while deleting the VM
3. opened the compute modal
4. noticed that I could select the standard PD option while creating a new standard VM
5. hit CREATE
6. observed that the new VM was attached to the existing SSD PD even though the UI made it seem like I was going to have a VM with a standard VM

See the screencast below showing the behaviour. I assume what should happen is, at step 4 above, we should see that SSD PD is the only option and it's greyed out.

https://user-images.githubusercontent.com/5438223/161281406-5a06ac21-e0a2-4d0e-a5f4-4bfe2c224add.mov

**EDIT:**
Not shown on the recording above is something I noticed later when I opened the compute modal after step 6 above. As you can see below, the new disk seems the cloud environment to be incorrectly recognized as an old cloud environment that needs upgrading.

image

src/components/GalaxyModal.js
```diff
@@ -56,7 +44,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
 
     const [dataDisk, setDataDisk] = useState({
       size: attachedDataDisk?.size || defaultDataDiskSize,
-      diskType: attachedDataDisk?.diskType || defaultDataDiskType
+      diskType: attachedDataDisk?.diskType || defaultPersistentDiskType
```
kyuksel: I don't think we want to make this change. Galaxy is special in that it requires provisioning of two persistent disks. One (referred to as postgres or metadata disk) is not visible to users. The other is referred to as data disk, which is presented to users as persistent disk but it feels helpful to make the distinction in the code base.

kyuksel (commented):
Mainly spent time testing and reporting issues. One code comment inline. Haven't looked at the cost-related changes. Will come back.

src/components/GalaxyModal.js
```diff
@@ -324,7 +312,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
                         id,
                         value: dataDisk.diskType,
                         isDisabled: true,
-                        onChange: value => setDataDisk(_.set('diskType', value.value)),
+                        onChange: value => setDataDisk(_.set('diskType', value)),
```
kyuksel: I'd suggest adding a helper for disk size/type now that it's an object, similarly to 

```javascript
  const updateComputeConfig = _.curry((key, value) => setComputeConfig(_.set(key, value)))
```

in `ComputeModal`. With that, we should be able to have the following

```suggestion
                        onChange: updateDisk('type'),
```

src/components/GalaxyModal.js
cstrandw: What would you recommend doing instead here to store the user selected PD type? Should I make a separate state variable for the disk type as well?

src/components/ComputeModal.js
```diff
@@ -1647,25 +1650,65 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
...
+                      value: computeConfig.selectedPersistentDiskType,
+                      isDisabled: diskExists || false,
+                      onChange: ({ value }) => updateComputeConfig('selectedPersistentDiskType', value),
+                      options: [
+                        { label: pdTypes.standard.displayName, value: pdTypes.standard.label },
+                        { label: pdTypes.ssd.displayName, value: pdTypes.ssd.label }
+                      ]
+                    })
+                  ])
+                ])
```
kyuksel: It may be good to pull this snippet out to a method like 

```javascript
const renderPersistentDiskType = …
```

since it's duplicated right below.

src/components/GalaxyModal.js
```diff
@@ -293,40 +298,84 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
...
                     min: 250, // Galaxy doesn't come up with a smaller data disk
                     max: 64000,
                     isClearable: false,
                     onlyInteger: true,
-                    value: dataDiskSize,
-                    style: { marginTop: '0.5rem', width: '5rem' },
-                    onChange: v => setDataDiskSize(v)
+                    style: { marginTop: '0.5rem' },
+                    value: dataDisk.size,
+                    onChange: v => setDataDisk(_.set('size', v))
```
kyuksel:
```suggestion
                    onChange: updateDataDisk('size')
```

src/components/GalaxyModal.js
```diff
@@ -397,7 +446,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
           getEnvMessageBasedOnStatus(app)
         ]),
         div({ style: { paddingBottom: '1.5rem', borderBottom: `1px solid ${colors.dark(0.4)}` } }, [
-          renderGalaxyCostBreakdown(kubernetesRuntimeConfig, dataDiskSize)
+          renderGalaxyCostBreakdown(kubernetesRuntimeConfig, dataDisk.size, dataDisk.diskType)
```
kyuksel: We could have a bit more concision by having

```suggestion
          renderGalaxyCostBreakdown(kubernetesRuntimeConfig, dataDisk)
```

if we change the definition of the function as something like

```javascript
const renderGalaxyCostBreakdown = (kubernetesRuntimeConfig, { size, diskType }) => { … }
```

src/components/RuntimeManager.js
```diff
@@ -335,7 +335,8 @@ export default class RuntimeManager extends PureComponent {
           })
       }
     }
-    const totalCost = _.sum(_.map(runtimeCost, runtimes)) + _.sum(_.map(disk => getPersistentDiskCostHourly(disk, computeRegion), persistentDisks))
+
+    const totalCost = () => _.sum(_.map(runtimeCost, runtimes)) + _.sum(_.map(disk => getPersistentDiskCostHourly(disk, computeRegion), persistentDisks))
```
kyuksel: Why was this change necessary?

src/data/machines.js
```diff
@@ -141,173 +141,172 @@ export const ephemeralExternalIpAddressPrice = { // per hour in dollars for all
...
 // The data below comes from Google's pricing API: https://cloud.google.com/billing/v1/how-tos/caltalog-api
-// There is a script available in https://gist.github.com/wnojopra/be5be3ab7e6a09a351b89085992a940d
-// that parses the API and outputs this data.
-// TODO: Update the above URL in the comment to the proper script location
+// There is a script available at https://github.com/broadinstitute/dsp-scripts/blob/master/terra-ui/regionality/parse_pricing_api.py.
+// It parses the API and outputs the response data.
 export const regionToPrices = [
   {
-    name: 'ASIA-EAST1', monthlyDiskPrice: 0.04,
+    name: 'ASIA-EAST1', monthlyDiskPrice: 0.04, monthlySSDPrice: 0.17,
```
kyuksel: It may be clearer and more consistent to use the keys `monthlyStandardDiskPrice` and `monthlySSDDiskPrice` (`D` in `SSD` stands for `Drive` and not `Disk`).

src/libs/runtime-utils.js
```diff
@@ -161,18 +175,19 @@ export const runtimeConfigCost = config => {
 }
 
 // Per GB following https://cloud.google.com/compute/pricing
-const getPersistentDiskPriceForRegionMonthly = computeRegion => {
-  return _.flow(_.find({ name: _.toUpper(computeRegion) }), _.get(['monthlyDiskPrice']))(regionToPrices)
+const getPersistentDiskPriceForRegionMonthly = (computeRegion, diskType) => {
+  const selectedRegionToPricesValue = pdTypes[_.replace('pd-', '', diskType)]?.regionToPricesName || pdTypes.standard.regionToPricesName
```
kyuksel: Can we eliminate the need for the string manipulation on the line above, by having the input parameter `diskType` to have value `pdTypes.standard` or `pdTypes.ssd` so we can directly get `regionToPricesName` on it?

src/data/machines.js
cstrandw: Yes that's true, it will require going in and changing the script again and I can check in with Willy about the change if you think it would be best! When changing the script I wanted to keep the changes more minimal and just add the SSD cost value without renaming other values in the output because I did not initially know how the output was being used.

src/libs/runtime-utils.js
cstrandw: I've looked into this a bit, but the issue is that the API call only returns the diskType string (ie 'pd-ssd') so the string manipulation always has to happen at some point to get the other values in the enum. Would it be better if getting the enum value happened before this function is called? It will require some refactoring but I can test that out.

src/data/machines.js
kyuksel: Yes, I think it'd be good to prioritize readability here.

> I wanted to keep the changes more minimal and just add the SSD cost value without renaming other values in the output because I did not initially know how the output was being used

Good idea to check use cases first. Searching for `monthlyDiskPrice`, it's only referred to once in `runtime-utils.js` so I can't see any issues with renaming it.

src/libs/runtime-utils.js
```diff
@@ -21,13 +21,32 @@ export const computeStyles = {
...
+  },
+  ssd: {
+    label: 'pd-ssd',
+    displayName: 'Solid state drive (SSD)',
+    regionToPricesName: 'monthlySSDDiskPrice'
+  },
+  fromString: str => Utils.switchCase(str,
+    [pdTypes.standard.label, () => pdTypes.standard],
+    [pdTypes.ssd.label, () => pdTypes.ssd],
+    [Utils.DEFAULT, () => console.error('Should not be calling pdTypes. fromString for $(str), invalid disk type')]
```
```suggestion
    [Utils.DEFAULT, () => console.error(`Invalid disk type: Should not be calling pdTypes.fromString for ${str}`)]
```

to use back-tick since we substitute the value of `str`. May also be more readable with the `Invalid disk type` piece at the beginning?

kyuksel:
Spot-checked the new SSD prices against what GCP advertises on their pages and they matched  😊

src/components/ComputeModal.js
```diff
@@ -184,6 +184,7 @@ const shouldUsePersistentDisk = (runtimeType, runtimeDetails, upgradeDiskSelecte
 
 export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDisks, tool, workspace, location, isAnalysisMode = false, shouldHideCloseButton = isAnalysisMode }) => {
   // State -- begin
+
```
kyuksel: Can the blank line be removed?
```suggestion
```

src/libs/runtime-utils.js
```diff
@@ -21,13 +21,33 @@ export const computeStyles = {
...
+  )
+}
+export const mapDisksToPDTypes = disks => _.map(_.update('diskType', pdTypes.fromString), disks)
```
```suggestion
export const mapToPdTypes = _.map(_.update('diskType', pdTypes.fromString))
```

since
1. we've been treating `Pd` as a word as far as camel-casing
2. we can take advantage of Lodash/FP's auto-curried nature and generalize the function for concision and flexibility. Its call sites shouldn't have to change other than the renaming.

src/libs/runtime-utils.js
```diff
@@ -233,14 +253,15 @@ export const getGalaxyComputeCost = app => {
...
-export const getGalaxyDiskCost = dataDiskSize => {
+export const getGalaxyDiskCost = ({ size, diskType }) => {
   const metadataDiskSize = 10 // GB
   const defaultNodepoolBootDiskSize = 100 // GB
   const appNodepoolBootDiskSize = 100 // GB
 
   return getPersistentDiskCostHourly({
     status: 'Running',
-    size: dataDiskSize + metadataDiskSize + defaultNodepoolBootDiskSize + appNodepoolBootDiskSize
+    size: size + metadataDiskSize + defaultNodepoolBootDiskSize + appNodepoolBootDiskSize,
```
```suggestion
    size: dataDiskSize + metadataDiskSize + defaultNodepoolBootDiskSize + appNodepoolBootDiskSize,
```

may be more consistent and clearer (in that `size` on the left and the right are different). I understand that it'll come at the cost of more verboseness at input deconstruction time; i.e.

```javascript
export const getGalaxyDiskCost = ({ size: dataDiskSize, diskType }) => {
```

src/libs/runtime-utils.js
```diff
@@ -265,7 +286,8 @@ export const getCurrentApp = appType => getCurrentAppExcludingStatuses(appType,
 export const getCurrentAppIncludingDeleting = appType => getCurrentAppExcludingStatuses(appType, [])
 
 export const getCurrentAttachedDataDisk = (app, appDataDisks) => {
-  return _.find({ name: app?.diskName }, appDataDisks)
+  const disk = _.find({ name: app?.diskName }, appDataDisks)
+  return disk && _.set('diskType', pdTypes.fromString(disk.diskType), _.find({ name: app?.diskName }, appDataDisks))
```
kyuksel: Can we use `disk`?
```suggestion
  const disk = _.find({ name: app?.diskName }, appDataDisks)
  return disk && _.set('diskType', pdTypes.fromString(disk.diskType), disk)
```

src/libs/runtime-utils.js
```diff
@@ -265,7 +286,8 @@ export const getCurrentApp = appType => getCurrentAppExcludingStatuses(appType,
 export const getCurrentAppIncludingDeleting = appType => getCurrentAppExcludingStatuses(appType, [])
 
 export const getCurrentAttachedDataDisk = (app, appDataDisks) => {
-  return _.find({ name: app?.diskName }, appDataDisks)
+  const disk = _.find({ name: app?.diskName }, appDataDisks)
+  return disk && _.set('diskType', pdTypes.fromString(disk.diskType), disk)
```
kyuksel: Sorry that I missed this earlier. Can we be explicit about type coercion in cases like this? 
 
```suggestion
  return !!disk && _.set('diskType', pdTypes.fromString(disk.diskType), disk)
```

kyuksel (approved):
@cstrandw Thank you for working through the comments. There are two minor inline comments left but I just played around with the UI and it looks great  👍🏼

kyuksel (changes requested):
@cstrandw On further testing, I seem to have run into a bug. Steps to reproduce:

1. Create a Galaxy app with a PD of type `SSD` and size `520` (anything different than the default should do)
2. Once it's created, delete the Galaxy app while keeping the PD.
3. Once disk detachment is complete, open the compute modal
4. Note that the modal displays incorrect disk type and size as shown below even though the disk seems to be created with the desired `type` and `size` in the backend (based on what I see on API responses on the Cloud Environments page)

image

kyuksel:
@cstrandw I'm noticing a different issue. Notice how the `Select`s are displayed on the Persistent Disk section below:

**Steps to reproduce:**

- Create a Galaxy app with a PD of type SSD
- Once it's created, delete the Galaxy app while keeping the PD
- Click the Delete button for the PD separately from Cloud Environments page
- Open the Galaxy modal

image

cstrandw:
> @cstrandw I'm noticing a different issue. Notice how the `Select`s are displayed on the Persistent Disk section below:
> 
> **Steps to reproduce:**
> 
> * Create a Galaxy app with a PD of type SSD
> * Once it's created, delete the Galaxy app while keeping the PD
> * Click the Delete button for the PD separately from Cloud Environments page
> * Open the Galaxy modal
> 
> image

This should no longer be an issue!

kyuksel (approved):
🥳

Qi77Qi (approved):

PR #2879, opened 2022-03-14
Handle stale metadata (but do it better than in #2862) [AJ-313]

davidangb:
Recent perf/scaling work has enabled - on a per-workspace feature-flag basis - caching for entity type metadata. This results in the possibility that metadata is stale.

The biggest UX impact of stale metadata we've discovered so far is missing columns from a data table. The use case is:
* start with a workspace that has one data table "participant"
* enable the "alwaysCacheAttributes" feature flag for this workspace
* upload a TSV to create a new "sample" table; this TSV has 10 columns in it
* navigate to the "sample" table, note that it only displays "sample_id" and no other columns

This happens because in the first ~2 minutes after the TSV upload, the cache is stale. Since the cache previously knew nothing about the "sample" table, it returns no attributeNames for this table.

This PR changes the behavior of the data table so that it inspects the current page of results, finds all the attributes in that page of results, and adds any attributes it finds in the results to the metadata. This should prevent stale cache from hiding columns that exist in the page of results.

Impact to end user:
* if the feature flag is not enabled (the default), no change in behavior
* if the feature flag *is* enabled, data tables are more resilient to stale cache

Tested by running locally and manually modifying the backend Rawls db's cache entries for my workspace to be missing some columns. Verified that those columns do not display before this PR, but do display after this PR.

Additional click paths to test - these failed in #2862 but should work in this PR:
* when viewing a data table affected by stale cache, click the "Data" tab to refresh the view. In #2862 we had a race condition - if the ajax request for the page of results returned before the ajax request for metadata, we would not display the extra columns.
* when deleting a column from a data table, the column would not disappear until you refreshed or navigated away and back. This was an artifact of the incorrect state handling in #2862.

**REVIEWERS:** I did the back end tweaking to set up a workspace to demonstrate this PR. LMK if you'd like me to add you to the workspace.
https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/da-invalid-cache/data is missing entries in its cache for the "sample" type. It does not display the proper columns for samples.
https://pr-8-dot-bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/da-invalid-cache/data (the PR site for this PR) does display the proper columns for samples.

Dev:
Screenshot (4)

PR Site:
Screenshot (5)

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -103,7 +103,7 @@ const DataTable = props => {
   const signal = useCancellation()
 
   // Helpers
-  const loadData = _.flow(
+  const loadData = entityMetadata && _.flow(
     Utils.withBusyState(setLoading),
     withErrorReporting('Error loading entities')
   )(async () => {
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -3,7 +3,7 @@ import { Fragment, useEffect, useRef, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import { Checkbox, Clickable, fixedSpinnerOverlay, Link } from 'src/components/common'
-import { DeleteEntityColumnModal, EditDataLink, EntityEditor, EntityRenamer, HeaderOptions, renderDataCell } from 'src/components/data/data-utils'
+import { concatenateAttributeNames, DeleteEntityColumnModal, EditDataLink, EntityEditor, EntityRenamer, HeaderOptions, renderDataCell } from 'src/components/data/data-utils'
 import { icon } from 'src/components/icons'
 import { ConfirmedSearchInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
@@ -49,7 +49,7 @@ const displayData = ({ itemsType, items }) => {
 
 const DataTable = props => {
   const {
-    entityType, entityMetadata, workspaceId, googleProject, workspaceId: { namespace, name },
+    entityType, entityMetadata, setEntityMetadata, workspaceId, googleProject, workspaceId: { namespace, name },
     onScroll, initialX, initialY,
     selectionModel: { selected, setSelected },
     childrenBefore,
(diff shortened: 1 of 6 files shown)


src/components/data/DataTable.js
```diff
@@ -103,7 +103,7 @@ const DataTable = props => {
   const signal = useCancellation()
 
   // Helpers
-  const loadData = _.flow(
+  const loadData = entityMetadata && _.flow(
```
kyuksel:
```suggestion
  const loadData = !!entityMetadata && _.flow(
```

to be more explicit about the intention of coercing `entityMetadata` into a boolean (assuming `entityMetadata` does not boolean values).

src/components/data/DataTable.js
```diff
@@ -114,6 +114,16 @@ const DataTable = props => {
...
+    // find all the unique attribute names contained in the current page of results
+    const attrNamesFromResults = _.uniq(_.flatMap(_.keys, _.map('attributes', results)))
+    // add any attribute names from the current page of results to those found in metadata.
+    // This allows for the stale metadata (e.g. the metadata cache is out of date).
+    // For the time being, the uniquness check MUST be case-insensitive (e.g. { sensitivity: 'accent' })
+    // in order to prevent case-divergent columns from being displayed, as that would expose some other bugs
+    const newAttrsForThisType = concatenateAttributeNames(entityMetadata[entityType]?.attributeNames, attrNamesFromResults)
+    if (!_.isEqual(newAttrsForThisType, entityMetadata[entityType].attributeNames)) {
+      setEntityMetadata(_.set([entityType, 'attributeNames'], newAttrsForThisType))
+    }
```
kyuksel: Consider

```suggestion
    // Find all the unique attribute names contained in the current page of results.
    const attrNamesFromResults = _.uniq(_.flatMap(_.keys, _.map('attributes', results)))
    // Add any attribute names from the current page of results to those found in metadata.
    // This allows for the stale metadata (e.g. the metadata cache is out of date).
    // For the time being, the uniqueness check MUST be case-insensitive (e.g. { sensitivity: 'accent' })
    // in order to prevent case-divergent columns from being displayed, as that would expose some other bugs
    const  attrNames = entityMetadata[entityType]?.attributeNames
    const newAttrsForThisType = concatenateAttributeNames(attrNames, attrNamesFromResults)
    if (!_.isEqual(newAttrsForThisType, attrNames)) {
      setEntityMetadata(_.set([entityType, 'attributeNames'], newAttrsForThisType))
    }
```

kyuksel: GitHub didn't diff this the way I was hoping. The diff includes:
- correction of a typo
- capitalization of first letter of comment lines
- pulling out `attrNames` to be on the subsequent lines

src/components/data/_tests/data-utils.test.js
```diff
@@ -0,0 +1,63 @@
...
+    const expected = ['case', 'foo']
+    expect(concatenateAttributeNames(attrList1, attrList2)).toEqual(expected)
+  })
+  it('sorts attribute names', () => {
+    const attrList1 = ['namespace:ccc', 'namespace:aaa', 'ddd']
+    const attrList2 = ['namespace:bbb', 'aaa']
+    const expected = ['aaa', 'ddd', 'namespace:aaa', 'namespace:bbb', 'namespace:ccc']
+    expect(concatenateAttributeNames(attrList1, attrList2)).toEqual(expected)
+  })
+})
```
kyuksel, line 63: Thank you for adding these unit tests! They help with visualizing the functionality too :)

kyuksel (approved):
The code changes LGTM overall. Minor comments inline.

davidangb:
@kyuksel thank you for the feedback, I have updated based on your suggestions, if you have any desire to re-review. If I don't hear from you I'll assume your approval stands!

jordansafer (approved):
Looks good! Everything working as expected for me :)

kyuksel (approved):
> I have updated based on your suggestions, if you have any desire to re-review. If I don't hear from you I'll assume your approval stands!

Yea totally good assumption 😊

PR #2881, opened 2022-03-14
Load files asynchronously, add aria-live region (WOR-193).

cahrens:
This prevents HighCharts JS from being downloaded unless it is needed (it was being downloaded on all calls, not just when the Billing Project page was loaded).

The aria-live text ensures that the user knows something has changed after modifying the "Last X days" select.

### DIFF
--- a/src/components/Chart.js
+++ b/src/components/Chart.js
@@ -1,0 +1,15 @@
+import Highcharts from 'highcharts' // Highcharts is being used under the Creative Commons Attribution-NonCommercial 3.0 license
+import highchartsAccessibility from 'highcharts/modules/accessibility'
+import highchartsExporting from 'highcharts/modules/exporting'
+import HighchartsReact from 'highcharts-react-official'
+import { h } from 'react-hyperscript-helpers'
+
+
+highchartsAccessibility(Highcharts)
+highchartsExporting(Highcharts)
+
+const Chart = ({ options }) => {
+  return h(HighchartsReact, { highcharts: Highcharts, options })
+}
+
+export default Chart
--- a/src/components/Chart.js
+++ b/src/components/Chart.js
@@ -8,8 +8,6 @@ import { h } from 'react-hyperscript-helpers'
 highchartsAccessibility(Highcharts)
 highchartsExporting(Highcharts)
 
-const Chart = ({ options }) => {
-  return h(HighchartsReact, { highcharts: Highcharts, options })
-}
+const Chart = ({ options }) => h(HighchartsReact, { highcharts: Highcharts, options })
 
 export default Chart
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -292,7 +292,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
         gridRowStart: 2
       }
     }, [
-      div({ style: { flex: 'none', padding: '0.625rem 1.25rem' }, 'aria-live': !!totalCost ? 'polite' : 'off', 'aria-atomic': true }, [
+      div({ style: { flex: 'none', padding: '0.625rem 1.25rem' }, 'aria-live': totalCost !== null ? 'polite' : 'off', 'aria-atomic': true }, [
         h3({ style: { fontSize: 16, color: colors.accent(), margin: '0.25rem 0.0rem', fontWeight: 'normal' } }, title),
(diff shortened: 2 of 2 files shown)


src/pages/billing/Project.js
```diff
@@ -296,7 +292,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
         gridRowStart: 2
       }
     }, [
-      div({ style: { flex: 'none', padding: '0.625rem 1.25rem' } }, [
+      div({ style: { flex: 'none', padding: '0.625rem 1.25rem' }, 'aria-live': totalCost !== null ? 'polite' : 'off', 'aria-atomic': true }, [
```
kyuksel, line 295: TIL  🚀

src/pages/billing/Project.js
```diff
@@ -369,11 +365,11 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
             }
           })
         ])])]),
-        CostCard({ title: 'Total spend', amount: (!!totalCost ? totalCost : '$__.__') })
+        CostCard({ title: 'Total spend', amount: (totalCost !== null ? totalCost : '$__.__') })
```
kyuksel, line 368: Consider

```suggestion
        CostCard({ title: 'Total spend', amount: (totalCost ?? '$__.__') })
```

for concision and also because (I'd think) we'd want to display the default when `totalCost` is `undefined` as well.

kyuksel (approved):
Looks great to me  🚀 
One nit inline.
Didn't test.

src/pages/billing/Project.js
cahrens: The default value is null, it should never be undefined.

src/pages/billing/Project.js
```diff
@@ -369,11 +365,11 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
       ]),
-      costPerWorkspace.numWorkspaces > 0 && div({ style: { gridRowStart: 2, minWidth: '500px' } }, // Set minWidth so chart will shrink on resize
-        [h(HighchartsReact, { highcharts: Highcharts, options: spendChartOptions })]
-      )
+      costPerWorkspace.numWorkspaces > 0 && div({ style: { gridRowStart: 2, minWidth: 500 } }, [ // Set minWidth so chart will shrink on resize
+        h(Suspense, { fallback: null }, [h(LazyChart, { options: spendChartOptions })])
```
aherbst-broad, line 371: Fun, I did not know about Suspense.

aherbst-broad (approved):

src/pages/billing/Project.js
cahrens: Because of the lazy load, I have to use a Suspense with a fallback. Typically a spinner would be the fallback, but as we've talked about I'm trying to avoid another spinner. Since there is nothing below on the page at this point, having no fallback looks fine to me.

src/components/Chart.js
```diff
@@ -0,0 +1,15 @@
...
+import HighchartsReact from 'highcharts-react-official'
+import { h } from 'react-hyperscript-helpers'
+
+
+highchartsAccessibility(Highcharts)
+highchartsExporting(Highcharts)
+
+const Chart = ({ options }) => {
+  return h(HighchartsReact, { highcharts: Highcharts, options })
+}
```
zarsky-broad: Do we want to one-line this to make it a little more obvious that it's a straight wrapper?
```suggestion
const Chart = ({ options }) => h(HighchartsReact, { highcharts: Highcharts, options })
```

src/pages/billing/Project.js
zarsky-broad, line 295: what about:
```suggestion
      div({ style: { flex: 'none', padding: '0.625rem 1.25rem' }, 'aria-live': !!totalCost ? 'polite' : 'off', 'aria-atomic': true }, [
```

zarsky-broad (approved):
Some optional comments.

src/pages/billing/Project.js
cahrens: I originally had this, but totalCost could be 0, and I would still want it to be read out. To me, the check for not null is much more obvious because the default value is null, and when I clear it on reset, I set it to null.

What I'm trying to avoid is the aria region reading out "$__.__". With this null check it seems to be working well.

zarsky-broad: Looking at the API response, it seems like the value is a string though? The only falsy string would be `""`, I'd expect a cost of 0 to return `"0.00"`

PR #2882, opened 2022-03-15
Upgrade react-select (WOR-194).

cahrens:
Change notes: https://react-select.com/upgrade

I was testing the upgrade to see if it helps with accessibility (the change notes call out accessibility improvements). I'm not seeing much of a difference with VoiceOver, but given that it's good to be up-to-date and I've already tested this upgrade, it seems worth doing.

Testing: we use Selects all over the place. I've tried out several different ones, and I reviewed the release notes and looked at our code for risks. I created WOR-195 for the VoiceOver issues I'm having with our Selects, which I think are likely due to the particular ways we are using them (the Select demo page works fine).

Examples of selects for testing-- selects in "Create Workspace" dialog, at top of "Workspaces List" for filtering, etc.

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -882,7 +882,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
     return div({ style: { ...computeStyles.whiteBoxContainer, marginTop: '1rem' } }, [
       div({ style: { fontSize: '0.875rem', fontWeight: 600 } }, ['Cloud compute profile']),
       div([
-        div({ style: { ...gridStyle, gridTemplateColumns: '0.25fr 4.5rem 1fr 5.5rem 1fr 5rem' } }, [
+        div({ style: { ...gridStyle, gridTemplateColumns: '0.25fr 5rem 1fr 6rem 1fr 5rem' } }, [
           // CPU & Memory Selection
           h(IdContainer, [
             id => h(Fragment, [
@@ -1051,7 +1051,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
             value: computeConfig.autopauseThreshold,
             hidden: !isAutopauseEnabled(computeConfig.autopauseThreshold),
             tooltip: !isAutopauseEnabled(computeConfig.autopauseThreshold) ? 'Autopause must be enabled to configure pause time.' : undefined,
-            onChange: updateComputeConfig('autopauseThreshold')
+            onChange: updateComputeConfig('autopauseThreshold'),
+            'aria-label': 'Minutes of inactivity before autopausing'
           }),
           span({ hidden: !isAutopauseEnabled(computeConfig.autopauseThreshold) }, ['minutes of inactivity'])
         ])
--- a/src/pages/workspaces/workspace/WorkspaceContainer.js
+++ b/src/pages/workspaces/workspace/WorkspaceContainer.js
@@ -87,8 +87,8 @@ const WorkspaceContainer = ({
         div({ style: Style.noWrapEllipsis }, breadcrumbs),
         h2({ style: Style.breadcrumb.textUnderBreadcrumb }, [
           title || `${namespace}/${name}`,
-          workspace && !Utils.canWrite(workspace.accessLevel) && span({ style: { paddingLeft: '0.5rem', color: colors.dark(0.85) } }, '(read only)'),
-          workspace && workspace.workspace.isLocked && span({ style: { paddingLeft: '0.5rem', color: colors.dark(0.85) } }, '(locked)')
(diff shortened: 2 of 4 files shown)


aherbst-broad (approved):

kyuksel (changes requested):
At least one [Select used in Compute Modal](https://github.com/DataBiosphere/terra-ui/blob/7557fba7bea826a23d47b808130d60385ecbeb46/src/components/ComputeModal.js#L107) seems to be affected:

image

kyuksel:
Updating [Compute Modal line 885](https://github.com/DataBiosphere/terra-ui/blob/7557fba7bea826a23d47b808130d60385ecbeb46/src/components/ComputeModal.js#L885) as 

div({ style: { ...gridStyle, gridTemplateColumns: '0.25fr **5**rem 1fr **6**rem 1fr 5rem' } }, [

seems to fix the truncation issues in `CPUs` and `Memory (GB)` `Select`s.

This may be a sign that all `Select` usages need double-checking though.

src/components/ComputeModal.js
```diff
@@ -1051,7 +1051,8 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
             value: computeConfig.autopauseThreshold,
             hidden: !isAutopauseEnabled(computeConfig.autopauseThreshold),
             tooltip: !isAutopauseEnabled(computeConfig.autopauseThreshold) ? 'Autopause must be enabled to configure pause time.' : undefined,
-            onChange: updateComputeConfig('autopauseThreshold')
+            onChange: updateComputeConfig('autopauseThreshold'),
+            'aria-label': 'Minutes of inactivity before autopausing'
```
cahrens, line 1055: This spinner had no label. Normally the text above would be linked as the label, but it's a checkbox, so a bit complicated.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -87,8 +87,8 @@ const WorkspaceContainer = ({
         div({ style: Style.noWrapEllipsis }, breadcrumbs),
         h2({ style: Style.breadcrumb.textUnderBreadcrumb }, [
           title || `${namespace}/${name}`,
-          workspace && !Utils.canWrite(workspace.accessLevel) && span({ style: { paddingLeft: '0.5rem', color: colors.dark(0.85) } }, '(read only)'),
-          workspace && workspace.workspace.isLocked && span({ style: { paddingLeft: '0.5rem', color: colors.dark(0.85) } }, '(locked)')
+          workspace && !Utils.canWrite(workspace.accessLevel) && span({ style: { paddingLeft: '0.5rem' } }, '(read only)'),
```
cahrens, line 90: This text is not legible and doesn't meet color contrast. Before:

Workspaces ›

After:
image

cahrens:
> This may be a sign that all `Select` usages need double-checking though.

Thanks for catching that, @kyuksel! I have included your change (and fixed a couple a11y issues I noticed while I was testing Selects). I have looked through all the Selects in code and viewed as many as I can figure out how to in Terra. I think that the Analysis modal ones were pretty unique in that they were constraining the Select size to be quite small; that is not our pattern in general, so hopefully there won't be other truncation issues. If I missed something in my testing though, it would be easy to fix as a follow-up change.

src/components/ComputeModal.js
kyuksel: > Normally the text above would be linked as the label

Which text? Asking for my edification.

src/pages/workspaces/workspace/WorkspaceContainer.js
kyuksel: Oh wow - great catch! 

Does it need a UX ticket for follow-up in case they may know a better way of emphasizing the `read-only`ness?

kyuksel (approved):
Thank you for the thorough work here, testing and improving accessibility 🙏🏼 

One question and one comment inline that aren't blockers for merging the PR.

src/components/ComputeModal.js
cahrens: As an example, look further down in this file for "Persistent Disk size" https://github.com/DataBiosphere/terra-ui/blob/51d9c6042d74261b3f6bfa1576ed4d1dbf020e2b/src/components/ComputeModal.js#L1649. The label there is linked to the NumberInput below it.

src/pages/workspaces/workspace/WorkspaceContainer.js
cahrens: I personally don't think it's worth creating a UX ticket. I know Lou is looking at revamping this part of the UI. I'll send him a note about it.

cahrens: Lou came up with something much nicer-- https://broadworkbench.atlassian.net/browse/WOR-217

PR #2883, opened 2022-03-15
[SUP-449] Remove requirement to enter 3 characters to search workspace tags

nawatts:
Currently, workspace tags menu do not show any options until at least 3 characters are entered. This removes that requirement, and shows tag options when the user first clicks on the menu.


cahrens (approved):
I agree, this is much nicer!

MatthewBemis (approved):

PR #2884, opened 2022-03-15
[AJ-343] Fix alignment of long snapshot table names

nawatts:
When a data table name is too long to fit in the sidebar, the table name is truncated.

Screen Shot 2022-03-15 at 11 00 21 AM

However, long table names in snapshots are not aligned correctly.

Screen Shot 2022-03-15 at 11 10 53 AM

The issue is that the max width of the row/button is hard coded based on the sidebar width, and does not account for the [left padding added to the list of tables in a snapshot](https://github.com/DataBiosphere/terra-ui/blob/440f3745f9571dd03418792a3af6eed37e896132/src/pages/workspaces/workspace/Data.js#L552). Changing the max width from 232px to 100% makes everything line up.

Screen Shot 2022-03-15 at 11 11 31 AM

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -63,7 +63,7 @@ const DataTypeButton = ({ selected, entityName, children, entityCount, iconName 
     role: 'listitem'
   }, [
     h(Clickable, {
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

cahrens (approved):
LGTM-- I found a workspace I could test it with.

PR #2885, opened 2022-03-16
[AJ-157] Validate snapshot reference name at point of import into workspace

vitzlo:
As per AJ-157, creating a snapshot reference via the #import-data landing page allows you to form an illegal name for the reference. Any illegal names for a snapshot reference are only detected once the snapshot's name is edited from within the workspace (i.e. not changing any values marks the name as illegal). This PR uses the pre-existing regular expressions to automatically clean the snapshot reference names upon import. One case involving various prohibited characters, `snapshotName=t3st%09%2B%0A%60%20%2E%0Band-_break`, is shown below, having been imported after this PR's changes. Tested locally within browser.

Screen Shot 2022-03-16 at 9 13 07 AM

Screen Shot 2022-03-16 at 9 16 22 AM

### DIFF
--- a/src/pages/ImportData.js
+++ b/src/pages/ImportData.js
@@ -108,6 +108,15 @@ const ImportData = () => {
       _.filter(({ name, namespace }) => _.some({ workspace: { namespace, name } }, workspaces))
     )(_.castArray(template))
 
+  // Normalize the snapshot name:
+  // Importing snapshot will throw an "enum" error if the name has any spaces or special characters
+  // Replace all whitespace characters with _
+  // Then replace all non alphanumeric characters with nothing
+  const normalizeSnapshotName = input => _.flow(
+    _.replace(/\s/g, '_'),
+    _.replace(/[^A-Za-z0-9-_]/g, '')
+  )(input)
+
   useOnMount(() => {
     const loadTemplateWorkspaces = _.flow(
       Utils.withBusyState(setIsImporting),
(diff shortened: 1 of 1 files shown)


cahrens (approved):
This looks fine from a code point-of-view. I don't know how to test this, nor do I know anything about the implications of changing the snapshot name. It seems that it could be nice to let the user choose the snapshot name (and enforce that the user follows the appropriate conventions), but I really don't know the use case that is being supported.

Please re-run the integration tests. You can add comments to our tracking tickets for the sporadic failures-- see tickets linked to this epic: https://broadworkbench.atlassian.net/browse/QA-1665

src/pages/ImportData.js
```diff
@@ -140,17 +140,18 @@ const ImportData = () => {
...
           const responses = await Promise.allSettled(
             _.map(({ title, id, description }) => {
-              // Normalize the title:
-              // Importing snapshot will throw an "enum" error if the title has any spaces or special characters
-              // Replace all whitespace characters with _
-              // Then replace all non alphanumeric characters with nothing
-              const normalizedTitle = _.flow(
-                _.replace(/\s/g, '_'),
-                _.replace(/[^A-Za-z0-9-_]/g, '')
-              )(title)
```
davidangb, line 153: This refactor would break importing from Data Catalog.

Note that here, in the deleted code, it is normalizing the variable `title` … but in the new code at line 150, it is normalizing the variable `snapshotName`. Thus - here inside the if clause - it'll end up with the wrong value.

My suggestion is to extract a new _function_ for normalization. This could go at the root of the `ImportData` component, parallel to `const filteredTemplates`, `const onImport`, etc:
```javascript
  // Normalize a snapshot name:
  // Importing snapshot will throw an "enum" error if the name has any spaces or special characters
  // Replace all whitespace characters with _
  // Then replace all non alphanumeric characters with nothing
  const normalizeSnapshotName = input => _.flow(
    _.replace(/\s/g, '_'),
    _.replace(/[^A-Za-z0-9-_]/g, '')
  )(input)
```

In this suggestion, note that I've renamed a few variables … but most importantly, it adds the `input => ` syntax, which turns this into a function instead of procedural code. Then, in the top half (Data Catalog) of the if clause you can use: 
```javascript
Ajax().Workspaces.workspace(namespace, name).importSnapshot(id, normalizeSnapshotName(title), description)
```
and in the bottom half (TDR import-by-copy) you can use:
```javascript
Ajax().Workspaces.workspace(namespace, name).importSnapshot(snapshotId, normalizeSnapshotName(snapshotName))
```

davidangb (approved):
:+1: I ran it locally and it properly normalized the snapshot name for TDR import-by-reference

PR #2878, opened 2022-03-11
[SU-5] Add option to save multiple column settings for data tables

nawatts:
Currently, changes the order and visibility of columns in data tables are only stored in the browser's local storage. This provides the option for users to create multiple column configurations that are persisted in workspace attributes and shared between all users of the workspace.

Screen Shot 2022-03-14 at 9 25 57 AM

Screen Shot 2022-03-14 at 9 26 12 AM

PR #2886, opened 2022-03-16
[SU-5] Disable save column settings if workspace is read only or locked

nawatts:
Fixup for #2878.

Saving column settings requires updating workspace attributes. If the user does not have write access to the workspace or the workspace is locked, that API request will fail. This disables the save button and adds a tooltip for those cases.

Screen Shot 2022-03-16 at 11 33 00 AM

### DIFF
--- a/src/components/data/SavedColumnSettings.js
+++ b/src/components/data/SavedColumnSettings.js
@@ -178,6 +178,7 @@ const SavedColumnSettings = ({ workspaceId, snapshotName, entityType, entityMeta
 
   const selectedSettingsNameExists = _.has(selectedSettingsName, savedColumnSettings)
   const editWorkspaceErrorMessage = workspace ? editWorkspaceError(workspace) : undefined
+  const canSaveSettings = workspace && !editWorkspaceErrorMessage
 
   return div({ style: { display: 'flex', flexDirection: 'column', height: '100%' } }, [
     div(showSaveForm ? [
@@ -216,7 +217,7 @@ const SavedColumnSettings = ({ workspaceId, snapshotName, entityType, entityMeta
       }, selectedSettingsNameExists ? 'Update' : 'Save')
     ] : [
       h(ButtonOutline, {
-        disabled: Boolean(editWorkspaceErrorMessage),
+        disabled: !canSaveSettings,
         tooltip: editWorkspaceErrorMessage,
         onClick: () => { setShowSaveForm(true) }
       }, 'Save this column selection')
--- a/src/components/data/SavedColumnSettings.js
+++ b/src/components/data/SavedColumnSettings.js
@@ -120,7 +120,8 @@ const useSavedColumnSettings = ({ workspaceId, snapshotName, entityMetadata, ent
   }
 }
 
-const SavedColumnSettings = ({ workspaceId, workspace, snapshotName, entityType, entityMetadata, columnSettings, onLoad }) => {
+const SavedColumnSettings = ({ workspace, snapshotName, entityType, entityMetadata, columnSettings, onLoad }) => {
+  const { workspace: { namespace, name } } = workspace
   const [loading, setLoading] = useState(true)
   const [savedColumnSettings, setSavedColumnSettings] = useState([])
 
@@ -128,7 +129,7 @@ const SavedColumnSettings = ({ workspaceId, workspace, snapshotName, entityType,
     getSavedColumnSettings,
     saveColumnSettings,
     deleteSavedColumnSettings
-  } = useSavedColumnSettings({ workspaceId, snapshotName, entityType, entityMetadata })
+  } = useSavedColumnSettings({ workspaceId: { namespace, name }, snapshotName, entityType, entityMetadata })
 
   useOnMount(() => {
     const loadSavedColumnSettings = _.flow(
--- a/src/components/data/SavedColumnSettings.js
+++ b/src/components/data/SavedColumnSettings.js
@@ -7,6 +7,7 @@ import { AutocompleteTextInput } from 'src/components/input'
 import { MenuButton, MenuTrigger } from 'src/components/PopupTrigger'
 import { ColumnSettings } from 'src/components/table'
 import TooltipTrigger from 'src/components/TooltipTrigger'
+import { useWorkspaceDetails } from 'src/components/workspace-utils'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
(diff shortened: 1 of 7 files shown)


src/components/workspace-utils.js
```diff
@@ -38,6 +38,25 @@ export const useWorkspaces = () => {
   return { workspaces, refresh, loading }
 }
 
+export const useWorkspaceDetails = ({ namespace, name }, fields) => {
```
cahrens, line 41: It looks like we have a variety of components that are calling `await Ajax(signal).Workspaces.workspace(namespace, name).details`. Could this new code be used in those places as well to help clean things up? 

FYI @dmohs -- this PR is a follow-up from one you reviewed, so it might make sense for you to review it. Also, you might have a perspective on how this should be factored.

src/components/workspace-utils.js
```diff
@@ -38,6 +38,25 @@ export const useWorkspaces = () => {
...
+  const [workspace, setWorkspace] = useState()
+
+  const [loading, setLoading] = useState(true)
+  const signal = useCancellation()
+
+  const refresh = _.flow(
+    withErrorReporting('Error loading workspace'),
```
cahrens: It would be good to make the error message more descriptive (perhaps 'Error loading workspace details').

nawatts: Revised in [84f28aa](https://github.com/DataBiosphere/terra-ui/pull/2886/commits/84f28aad7dcbc78fae67b1496b0f6522496e5d9e)

src/components/workspace-utils.js
nawatts: I think so. On the other hand, using `useThing` hooks for API requests might be undesirable for components that make multiple API requests if each hook has its own cancellation signal and busy state vs sharing one for all API requests.

cahrens: I looked through the other calls to `details` and only saw one that was also making the call `onMount`. I went ahead and replaced it-- let me know what you think.

src/components/data/SavedColumnSettings.js
```diff
@@ -213,6 +216,8 @@ const SavedColumnSettings = ({ workspaceId, snapshotName, entityType, entityMeta
       }, selectedSettingsNameExists ? 'Update' : 'Save')
     ] : [
       h(ButtonOutline, {
+        disabled: Boolean(editWorkspaceErrorMessage),
```
cahrens: You should assume disabled until you hear back from WorkspaceDetails. Otherwise there is a race condition during which the user could press the button before we AJAX call returns.

nawatts: Done in [33ba467](https://github.com/DataBiosphere/terra-ui/pull/2886/commits/33ba4675194fa98c6bccb66b0b3e62ead7dc72a7).

cahrens:
@nawatts it looks like there is a deployment/test issue

cahrens (approved):

cahrens:
@nawatts I approved this because I think it solves the problem. However, the spinning overlay and disabled button are a bit ugly when you open the dialog. I could see changing the code so that the Data tab loads the WorkspaceDetails as an async call, and passes the Workspace Details to the dialog (if they have returned). Or perhaps some caching, as it seems this data is loaded every time the dialog is displayed.

nawatts:
The spinner overlay seemed to be a pattern in most every modal I've seen so far (for example, share workspace or delete workspace). Likewise with loading data from the API close to where it's used. Alternatively, here, the workspace could be plumbed through from EntitiesContent or DataStepContent.

cahrens:
I personally have spinner fatigue (for example, see https://broadworkbench.atlassian.net/browse/WOR-185 for what happens when we use spinners on every Ajax load), but it's true, you are following existing examples. If the relevant data is already available in the container that spans the dialog, definitely just pass it in. But if it's going to be more complicated than that, I'm fine with you sticking with your current solution.

nawatts:
Plumbed the workspace through for getting the access level and locked status. That at least takes care of the save button switching from disabled to enabled. Still need to refresh the saved column settings to make sure they're up to date though.

One thing we could possibly try is changing the spinner component to only render after ~100ms so that they stay out of the way when API requests are fast.

MatthewBemis (approved):
LGTM. Apologies for the delay reviewing this one

src/pages/workspaces/workspace/workflows/DataStepContent.js
```diff
@@ -17,7 +17,7 @@ import validate from 'validate.js'
 const DataStepContent = ({
   entitySelectionModel, onDismiss, onSuccess,
   entityMetadata, rootEntityType,
-  workspace: { namespace, name, googleProject, attributes: { 'workspace-column-defaults': columnDefaults } }
+  workspace, workspace: { workspace: { namespace, name, googleProject, attributes: { 'workspace-column-defaults': columnDefaults } } }
```
cahrens: Instead of passing in the whole workspace, could you add the additional things you need from workspace to `workspace: { workspace: { namespace, name, googleProject, attributes: { 'workspace-column-defaults': columnDefaults } }`? It always seems icky to me when both the whole object and a selected subset of the same object are passed into something.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -930,7 +930,7 @@ const WorkflowView = _.flow(
           this.setState({ selectingData: false })
         },
         onSuccess: model => this.setState({ entitySelectionModel: model, selectingData: false }),
-        workspace,
+        workspace: ws,
```
cahrens, line 933: So before workspace was undefined because the wrong variable name was being referenced?

nawatts: The workspace details API returns a response like:
```
{ accessLevel: ..., workspace: { ... }, ... }
```

In WorkflowView's renderSummary function, `workspace` refers to that inner `workspace` object and `ws` to the whole thing.
https://github.com/DataBiosphere/terra-ui/blob/38b2a5535925ab2dd08e20cbe8b6b4c100228323/src/pages/workspaces/workspace/workflows/WorkflowView.js#L588

This changes the type of DataStepContent's `workspace` prop. DataStepContent is updated to match here: https://github.com/DataBiosphere/terra-ui/pull/2886/files#diff-53608f0ffab957297bc01f1fd7a6bd48d53fc2315a86b575f1ebfc5d9fa0322bR20

src/pages/workspaces/workspace/workflows/DataStepContent.js
nawatts: Split this up in f37a36db147d20b47d0306872540eba609b9a5e6.

nawatts:
@cahrens you approved earlier, but there have been several changes since. What do think of this in its current state?

cahrens:
> @cahrens you approved earlier, but there have been several changes since. What do think of this in its current state?

@nawatts Watching how this PR has evolved, it's an excellent example of the trade-offs between passing state through multiple components to avoid an extra fetch vs. isolating the state needed in a component. I really am on the fence about which approach is better, and like you said, we have to have a spinner on the dialog anyway (the initially disabled button might not be a big deal given the spinner).

So if you would prefer to revert to https://github.com/DataBiosphere/terra-ui/pull/2886/commits/33ba4675194fa98c6bccb66b0b3e62ead7dc72a7, that's fine with me-- go with whichever approach you think is cleanest.

nawatts:
I think I prefer the current approach with passing the workspace down from WorkspaceContainer. It saves an API request and the save button is always in the desired state.

PR #2887, opened 2022-03-17
[SU-2] Validate collection names in data uploader

nawatts:
Collection names in the data uploader (http://localhost:3000/#upload) are used to create "directories" in the workspace's GCS bucket. Thus, some characters cause issues. Forward slashes create nested directories that break Terra's assumptions about collections. Spaces aren't handled by well by Cromwell / workflows. And Google recommends excluding a few other characters (https://cloud.google.com/storage/docs/naming-objects#objectnames).

This adds validation to the collection name input to disallow using those characters.

### DIFF
--- a/src/pages/Upload.js
+++ b/src/pages/Upload.js
@@ -403,7 +403,7 @@ const CollectionSelectorPanel = _.flow(
     ]),
     isCreating && h(NameModal, {
       thing: 'Collection',
-      validator: /^[^ /#*?\[\]]+$/, // eslint-disable-line no-useless-escape
+      validator: /^[^\s/#*?\[\]]+$/, // eslint-disable-line no-useless-escape
       validationMessage: 'Collection name may not contain spaces, forward slashes, or any of the following characters: # * ? [ ]',
       onDismiss: () => setCreating(false),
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

src/pages/Upload.js
```diff
@@ -403,6 +403,8 @@ const CollectionSelectorPanel = _.flow(
     ]),
     isCreating && h(NameModal, {
       thing: 'Collection',
+      validator: /^[^ /#*?\[\]]+$/, // eslint-disable-line no-useless-escape
```
kyuksel: Thinking out loudly (TOL - which implies it's not a blocker): I wonder if we should use `\s` instead of a plain space in the regex so it'd match `\t\r\n\f` as well.

kyuksel (approved):
LGTM with one TOL inline.
Didn't test.

src/pages/Upload.js
nawatts: Done in [c5f6edb](https://github.com/DataBiosphere/terra-ui/pull/2887/commits/c5f6edb954c7c08c536bf27ccaa945e9afc3891f).

PR #2888, opened 2022-03-17
[DC-164] Updated eslint to warn on unused arguments as well as unused variables

lu-c:
Now, unused variables in function parameters will throw errors
Screen Shot 2022-03-17 at 12 18 19 PM

One of the eslint changes to note: I updated the PR so that variables with "_" as the first character of the name will be exempted from this eslint rule. This is to allow for our use of external library functions without having to eslint:ignore all of the lines.

Some of the conditions to consider:
- I added a bunch of "ignore-next-line"s to cases where I wasn't sure if it would break something. The "Ref" example happens a LOT and I don't know if removing that unused variable would affect anything?


src/pages/Upload.js
```diff
@@ -703,7 +703,7 @@ const DonePanel = ({ workspace, workspace: { workspace: { namespace, name } }, t
     }, [
       p([
         h(Link, {
-          href: Nav.getLink('workspace-data', { namespace: workspace.workspace.namespace, name: workspace.workspace.name }),
+          href: Nav.getLink('workspace-data', { namespace, name }),
```
petesantos, line 706: Did the linter catch this specifically, or did it catch something close and you corrected it?

petesantos (commented):
This looks good.
I found the [PR](https://github.com/DataBiosphere/terra-ui/pull/2723) with the unused variable. After some conversation I used `_.unused`.
What are your thoughts on renaming all unused variables to `_.unused`?
The thought on the other PR is that it becomes a bit more obvious that the variable is not used.

src/pages/Upload.js
lu-c: The linter caught this, but in the function args - Said that the variables `namespace` and `name` were unused, so i flipped this to use it!

lu-c:
> This looks good. I found the [PR](https://github.com/DataBiosphere/terra-ui/pull/2723) with the unused variable. After some conversation I used `_.unused`. What are your thoughts on renaming all unused variables to `_.unused`? The thought on the other PR is that it becomes a bit more obvious that the variable is not used.

I think I want to keep the variable names what they are - here are some of my thoughts:
- We might have multiple unused vars in function names
- If we want to suddenly start using them, we can just delete the _
- when debugging, having a name can help us identify what it's SUPPOSED to be for (esp for things like the `_ref`s)

cahrens:
> I think I want to keep the variable names what they are - here are some of my thoughts:
> 
> * We might have multiple unused vars in function names
> * If we want to suddenly start using them, we can just delete the _
> * when debugging, having a name can help us identify what it's SUPPOSED to be for (esp for things like the `_ref`s)

I agree. I have seen the convention of putting a prefix of "_" on variable names that are unused in other projects/languages, so it feels very familiar and understandable. I would rather not use the same name for all unused variables.

src/pages/Upload.js
petesantos: Ah, nice - thanks!

petesantos (approved):

fboulnois (approved):
overall, lgtm!

PR #2891, opened 2022-03-18
[WOR-185] Spinner update on Billing Projects page.

cahrens:
There are a couple of things here. First, because a spinner was being displayed relative to a portion of the screen, you would get 2 visible spinners when first loading the billing page because the parent list page also used a spinner (See [WOR-185](https://broadworkbench.atlassian.net/browse/WOR-185)). I was reluctant to add a spinner to the loading of the spend report data because I was afraid of adding more spinner-craziness, but now that I understand that I can just make sure the spinners appear in the same location, I went ahead and enabled the spinner in that case also.

Second, I have attempted to clean up the code some for fetching the spend report data (which then causes the chart to re-render) by moving it into an effect.

To test:

1. Log in as "hermione.owner" (must be billing report owner, and must be in the alpha testing group).
2. Select galaxy-anvil-dev billing project. http://localhost:3000/#billing?selectedName=galaxy-anvil-dev&type=project&tab=spend%20report

### DIFF
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -191,6 +191,8 @@ const groupByBillingAccountStatus = (billingProject, workspaces) => {
 }
 
 const LazyChart = lazy(() => import('src/components/Chart'))
+const maxWorkspacesInChart = 10
+const spendReportKey = 'spend report'
 
 const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProject, isAlphaSpendReportUser, reloadBillingProject }) => {
   // State
@@ -226,8 +228,6 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
     { namespace: billingProject.projectName },
     _.map('workspace', workspaces)
   ), [billingProject, workspaces])
-
-  const maxWorkspacesInChart = 10
   const spendChartOptions = {
     chart: { type: 'bar', style: { fontFamily: 'inherit' } },
     credits: { enabled: false },
@@ -249,39 +249,6 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
     },
     exporting: { buttons: { contextButton: { x: -15 } } }
   }
-
-  const spendReportKey = 'spend report'
-  const maybeLoadTotalCost = _.flow(
-    reportErrorAndRethrow('Unable to retrieve spend report data'),
-    Utils.withBusyState(setUpdating)
-  )(async () => {
-    if (!updatingTotalCost && totalCost === null && tab === spendReportKey) {
-      setUpdatingTotalCost(true)
-      const endDate = new Date().toISOString().slice(0, 10)
-      const startDate = new Date((Date.now() - spendReportLengthInDays * 24 * 60 * 60 * 1000)).toISOString().slice(0, 10)
-      const spend = await Ajax(signal).Billing.getSpendReport({ billingProjectName: billingProject.projectName, startDate, endDate })
-      const costFormatter = new Intl.NumberFormat(navigator.language, { style: 'currency', currency: spend.spendSummary.currency })
-      setTotalCost(costFormatter.format(spend.spendSummary.cost))
-
-      const workspaceDetails = _.find(details => details.aggregationKey === 'Workspace')(spend.spendDetails)
-      console.assert(workspaceDetails !== undefined, 'Spend report details do not include aggregation by Workspace')
-      // Get the most expensive workspaces, sorted from most to least expensive.
-      const mostExpensiveWorkspaces = _.flow(
-        _.sortBy(({ cost }) => { return parseFloat(cost) }),
-        _.reverse,
-        _.slice(0, maxWorkspacesInChart)
-      )(workspaceDetails?.spendData)
-      // Pull out names and costs.
-      const costsPerWorkspace = { workspaceNames: [], workspaceCosts: [], numWorkspaces: workspaceDetails?.spendData.length }
-      _.forEach(workspaceCostData => {
-        costsPerWorkspace.workspaceNames.push(workspaceCostData.workspace.name)
-        costsPerWorkspace.workspaceCosts.push(parseFloat(workspaceCostData.cost))
-      })(mostExpensiveWorkspaces)
-      setCostPerWorkspace(costsPerWorkspace)
-
-      setUpdatingTotalCost(false)
-    }
-  })
 
   const CostCard = ({ title, amount, ...props }) => {
     return div({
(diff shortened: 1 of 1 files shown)


src/pages/billing/Project.js
```diff
@@ -645,7 +651,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
       }
     }),
     billingAccountsOutOfDate && h(BillingAccountSummaryPanel, { counts: _.mapValues(_.size, groups) }),
-    updating && spinnerOverlay
+    updating && absoluteSpinnerOverlay
```
cahrens, line 651: This causes the spinner to render in the same location as the one for the parent page.

src/pages/billing/Project.js
```diff
@@ -361,7 +363,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
             }), [7, 30, 90]),
             onChange: ({ value: selectedDays }) => {
               setSpendReportLengthInDays(selectedDays)
-              setTotalCost(null) // This will force the report to be recalculated based on selectedDays
```
cahrens, line 364: Changing the total cost itself does not cause the report to be recalculated-- `spendReportLengthInDays` does through the effect dependency.

src/pages/billing/Project.js
```diff
@@ -452,6 +454,11 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
   useOnMount(() => { reloadBillingProjectUsers() })
 
   useEffect(() => { StateHistory.update({ projectUsers }) }, [projectUsers])
+  // Update cost data only if report date range changes, or if spend report tab was selected.
+  useEffect(() => { maybeLoadTotalCost() },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
cahrens: I had to disable `exhaustive-deps` to stop the warning about calling `maybeLoadTotalCost`. If I add `maybeLoadTotalCost` to the dependencies things break because a lot of state is being used within that method. The point though is that I don't want to fetch the data unless the user has requested a different report length (or the tab was initially selected).

src/pages/billing/Project.js
```diff
@@ -596,7 +603,6 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
           } else {
             setTab(newTab)
           }
-          maybeLoadTotalCost()
```
cahrens, line 599: I was trying to make sure that we loaded the spend report data when the tab was first selected (we don't want to load the report when the parent page is loaded because the query costs money, and the user may never choose to view the spend report). However, this code wasn't doing what I thought because the new value of `tab` isn't actually persisted yet. Now `tab` is one of the dependencies for the effect.

src/pages/billing/Project.js
```diff
@@ -251,7 +251,10 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
   }
 
   const spendReportKey = 'spend report'
-  const maybeLoadTotalCost = reportErrorAndRethrow('Unable to retrieve spend report data')(async () => {
+  const maybeLoadTotalCost = _.flow(
+    reportErrorAndRethrow('Unable to retrieve spend report data'),
+    Utils.withBusyState(setUpdating)
+  )(async () => {
     if (!updatingTotalCost && totalCost === null && tab === spendReportKey) {
```
cahrens: One other tricky thing going on here is that if the ajax call to get the spend report throws an error, I don't want to keep calling it and get a stack of error notifications…. the call is expected to fail if billing permission information is not set up. This is why I'm managing `updatingTotalCost` here myself, so it remains `true` if the report errors.

We will probably want to handle this more elegantly before the feature is released (and show explanatory text vs. an error), but we are still early in development of this report and iteratively building up the UI.

cahrens:
@zarsky-broad @panentheos this is the code I was asking about at the end of training.

src/pages/billing/Project.js
```diff
@@ -250,37 +250,6 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
     exporting: { buttons: { contextButton: { x: -15 } } }
   }
 
-  const spendReportKey = 'spend report'
-  const maybeLoadTotalCost = reportErrorAndRethrow('Unable to retrieve spend report data')(async () => {
```
cahrens, line 254: maybeLoadTotalCost is moved into `useEffect`.

src/pages/billing/Project.js
kyuksel: I wonder if we should note this as a comment above the definition of `absoluteSpinnerOverlay`.

src/pages/billing/Project.js
```diff
@@ -452,6 +421,41 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+  // Update cost data only if report date range changes, or if spend report tab was selected.
+  useEffect(() => {
+    const maybeLoadTotalCost = _.flow(
+      reportErrorAndRethrow('Unable to retrieve spend report data'),
+      Utils.withBusyState(setUpdating)
+    )(async () => {
+      if (!updatingTotalCost && totalCost === null && tab === spendReportKey) {
+        setUpdatingTotalCost(true)
+        const endDate = new Date().toISOString().slice(0, 10)
+        const startDate = new Date((Date.now() - spendReportLengthInDays * 24 * 60 * 60 * 1000)).toISOString().slice(0, 10)
```
aherbst-broad, line 433: This is going to be flaky with DST (however long that stays with us ;) ). Is there a more idiomatic way in JS to add/subtract days?

aherbst-broad (approved):
lgtm , had a comment re: date arithmetic.

src/pages/billing/Project.js
cahrens: I can look into it.

zarsky-broad: We have https://date-fns.org/ as a dependency already

cahrens: Thanks, I will check it out, but would like to go ahead and merge this PR as-is since it doesn't introduce this problem.

@zarsky-broad or @dmohs if I can get a thumbs on this PR it will help me to move the review forward on a dependent PR (it will make it easier to see what has actually changed in the next PR if this refactoring is merged first).

cahrens: It looks like @dmohs is on vacation, so I'm removing him as a reviewer. @zarsky-broad, it's all up to you now. :)

src/pages/billing/Project.js
zarsky-broad: My only issue with this is that it would prevent selecting another project if your spend data is taking a long time to load.

src/pages/billing/Project.js
```diff
@@ -226,8 +228,6 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
     { namespace: billingProject.projectName },
     _.map('workspace', workspaces)
   ), [billingProject, workspaces])
```
zarsky-broad, line 230: nit: re-add spacer line
```suggestion
  ), [billingProject, workspaces])

```

zarsky-broad (approved):
One comment, and at some point I'd love to address the flicker that happens on load, causing the chart's bars to restart their animation

cahrens:
> One comment, and at some point I'd love to address the flicker that happens on load, causing the chart's bars to restart their animation

@zarsky-broad Oh no, there shouldn't be a flicker/restart! I definitely don't want that, and I'm not seeing it myself. Are there specific steps that cause it to happen?

src/pages/billing/Project.js
cahrens: Good point-- as we get some usability testing on this report, we can keep that in mind. So far the reports seem to be pretty fast, but we have had limited "real world" data.

src/pages/billing/Project.js
cahrens: Will put this in my next PR to avoid merge issues.

PR #2895, opened 2022-03-21
[SU-25] Refactor EntityEditor

nawatts:
The data table allows editing the value of an attribute using the EntityEditor component. This splits up EntityEditor and extracts the form for entering the attribute's value/type so that they can be reused for adding new rows (in a future PR).

### DIFF
--- a/src/components/data/_tests/data-utils.test.js
+++ b/src/components/data/_tests/data-utils.test.js
@@ -72,9 +72,9 @@ describe('getAttributeType', () => {
       entityName: 'thing_one'
     })).toEqual({ type: 'reference', isList: false })
 
-    expect(getAttributeType({ items: ['a', 'b', 'c'] })).toEqual({ type: 'string', isList: true })
-    expect(getAttributeType({ items: [1, 2, 3] })).toEqual({ type: 'number', isList: true })
-    expect(getAttributeType({ items: [true, false] })).toEqual({ type: 'boolean', isList: true })
+    expect(getAttributeType({ items: ['a', 'b', 'c'], itemsType: 'AttributeValue' })).toEqual({ type: 'string', isList: true })
+    expect(getAttributeType({ items: [1, 2, 3], itemsType: 'AttributeValue' })).toEqual({ type: 'number', isList: true })
+    expect(getAttributeType({ items: [true, false], itemsType: 'AttributeValue' })).toEqual({ type: 'boolean', isList: true })
     expect(getAttributeType({
       items: [
         { entityType: 'thing', entityName: 'thing_one' },
@@ -86,8 +86,8 @@ describe('getAttributeType', () => {
 
   it('returns string for null values', () => {
     expect(getAttributeType(null)).toEqual({ type: 'string', isList: false })
-    expect(getAttributeType({ items: [] })).toEqual({ type: 'string', isList: true })
-    expect(getAttributeType({ items: [null] })).toEqual({ type: 'string', isList: true })
+    expect(getAttributeType({ items: [], itemsType: 'AttributeValue' })).toEqual({ type: 'string', isList: true })
+    expect(getAttributeType({ items: [null], itemsType: 'AttributeValue' })).toEqual({ type: 'string', isList: true })
   })
 })
 
@@ -116,7 +116,7 @@ describe('convertAttributeValue', () => {
   })
 
   it('converts each value of lists', () => {
-    expect(convertAttributeValue({ items: ['42', 'value'] }, 'number')).toEqual({ items: [42, 0] })
+    expect(convertAttributeValue({ items: ['42', 'value'], itemsType: 'AttributeValue' }, 'number')).toEqual({ items: [42, 0], itemsType: 'AttributeValue' })
 
     expect(convertAttributeValue({
       items: [
@@ -124,11 +124,11 @@ describe('convertAttributeValue', () => {
         { entityType: 'thing', entityName: 'thing_two' }
       ],
       itemsType: 'EntityReference'
-    }, 'string')).toEqual({ items: ['thing_one', 'thing_two'] })
+    }, 'string')).toEqual({ items: ['thing_one', 'thing_two'], itemsType: 'AttributeValue' })
   })
 
   it('adds itemsType to reference lists', () => {
-    expect(convertAttributeValue({ items: ['thing_one', 'thing_two'] }, 'reference', 'thing')).toEqual({
+    expect(convertAttributeValue({ items: ['thing_one', 'thing_two'], itemsType: 'AttributeValue' }, 'reference', 'thing')).toEqual({
       items: [
         { entityType: 'thing', entityName: 'thing_one' },
         { entityType: 'thing', entityName: 'thing_two' }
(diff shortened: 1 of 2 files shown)


cahrens:
@nawatts is this PR a refactor in support of SU-25, or does it also add the ability to add rows/columns to the data table? I'm running the deployment but am not sure I know what I'm looking or.

nawatts:
Sorry, should have clarified. This is in support of SU-25 and adding new rows would be added in a future PR. This PR should not make any visible changes.

src/components/data/_tests/data-utils.test.js
```diff
@@ -61,3 +61,79 @@ describe('concatenateAttributeNames', () => {
...
+    expect(getAttributeType('value')).toEqual({ type: 'string', isList: false })
+    expect(getAttributeType(3)).toEqual({ type: 'number', isList: false })
+    expect(getAttributeType(false)).toEqual({ type: 'boolean', isList: false })
+    expect(getAttributeType({
+      entityType: 'thing',
+      entityName: 'thing_one'
+    })).toEqual({ type: 'reference', isList: false })
+
+    expect(getAttributeType({ items: ['a', 'b', 'c'] })).toEqual({ type: 'string', isList: true })
+    expect(getAttributeType({ items: [1, 2, 3] })).toEqual({ type: 'number', isList: true })
```
cahrens: What does it do with mixed types (`['a', 2, false]`, for example)?

src/components/data/_tests/data-utils.test.js
```diff
@@ -61,3 +61,79 @@ describe('concatenateAttributeNames', () => {
...
+      items: [
+        { entityType: 'thing', entityName: 'thing_one' },
+        { entityType: 'thing', entityName: 'thing_two' }
+      ],
+      itemsType: 'EntityReference'
+    })).toEqual({ type: 'reference', isList: true })
+  })
+
+  it('returns string for null values', () => {
+    expect(getAttributeType(null)).toEqual({ type: 'string', isList: false })
```
cahrens, line 88: These tests are great! Some of these conversions are not at all obvious, and it is great to have documentation of what they do.

src/components/data/_tests/data-utils.test.js
nawatts: I _think_ that a list is not allowed to contain items of mixed types.

However, I'm not sure if it's possible to create a generic JSON value that looks like a list. Generic JSON values currently break this component ([AJ-342](https://broadworkbench.atlassian.net/browse/AJ-342)).

nawatts: Rawls does reject creating a list with mixed types. `batchUpsert` with:
```
[
  {
    "name": "thing_one",
    "entityType": "things",
    "operations": [
      {
        "op": "CreateAttributeValueList",
        "attributeName": "new_attr"
      },
      {
        "op": "AddListMember",
        "attributeListName": "new_attr",
        "newMember": "a"
      },
      {
        "op": "AddListMember",
        "attributeListName": "new_attr",
        "newMember": 2
      },
      {
        "op": "AddListMember",
        "attributeListName": "new_attr",
        "newMember": false
      }
    ]
  }
]
```
returns an error:
> inconsistent attributes for list: attribute lists must consist of a single data type. For attribute 'new_attr', found types: [AttributeString, AttributeNumber, AttributeBoolean]. Sample values for these types: [AttributeString(a), AttributeNumber(2), AttributeBoolean(false)]

There is still the issue of creating a raw JSON attribute that looks like a list, but I think maybe we can defer that to AJ-342 since this is the same as the current behavior of EntityEditor: https://github.com/DataBiosphere/terra-ui/blob/9053e0a383676f8ac183d278204c3e2ae35066be/src/components/data/data-utils.js#L478-L485

src/components/data/data-utils.js
```diff
@@ -472,68 +472,217 @@ export const EntityRenamer = ({ entityType, entityName, workspaceId: { namespace
...
+        value,
+        onChange: v => onChange(_.trim(v)),
+        ...otherProps
+      })
+    }],
+    ['reference', () => {
+      const { value, onChange, ...otherProps } = props
+      return h(TextInput, {
+        autoFocus: true,
+        placeholder: 'Enter a value',
```
cahrens: This used to be `Enter a ${linkedEntityType}_id`. Was the change intentional?

nawatts: Oops, that was not intentional.

nawatts: Fixed in [5342971](https://github.com/DataBiosphere/terra-ui/pull/2895/commits/5342971e15085e1c826f60e9c87852c1c551f3df).

src/components/data/data-utils.js
```diff
@@ -472,68 +472,217 @@ export const EntityRenamer = ({ entityType, entityName, workspaceId: { namespace
...
+      return _.isNaN(numberVal) ? 0 : numberVal
+    }],
+    [Utils.DEFAULT, () => Utils.convertValue(newType)]
   )
+  const convertFn = type === 'reference' ? _.flow(_.get('entityName'), baseConvertFn) : baseConvertFn
+
+  if (isList) {
+    return _.flow(
+      _.update('items', _.map(convertFn)),
+      newType === 'reference' ? _.set('itemsType', 'EntityReference') : _.omit('itemsType')
```
cahrens: I don't understand why `itemsType` would be present (the omit part of this statement). Can you add a comment about what this code is doing?

nawatts: Lists of entity references look like:
```
{
   items: [
     { entityType: 'thing', entityName: 'thing_1' },
     { entityType: 'thing', entityName: 'thing_1' }
  ],
  itemsType: 'EntityReference'
}
```

This was dropping that `itemsType` field when converting from a list of references to a list of some other type. However, I missed that lists of other types also have an `itemsType` field, but it's set to `AttributeValue` instead of `EntityReference`.

[e8870fd](https://github.com/DataBiosphere/terra-ui/pull/2895/commits/e8870fdb45a008185b03a88c1cb6db2bfd6ad678) adds `itemsType` for non-reference types and makes this conversion clearer.

src/components/data/data-utils.js
```diff
@@ -472,68 +472,215 @@ export const EntityRenamer = ({ entityType, entityName, workspaceId: { namespace
...
+  )
+}
+
+const AttributeInput = ({ value: attributeValue, onChange, entityTypes = [] }) => {
+  const { type: attributeType, isList } = getAttributeType(attributeValue)
+
+  const renderInput = renderInputForAttributeType(attributeType)
+
+  const defaultReferenceEntityType = Utils.cond(
+    [attributeType === 'reference' && isList, () => !_.isEmpty(attributeValue.items) ? attributeValue.items[0].entityType : entityTypes[0]],
```
cahrens, line 572: It looks like this used to return undefined. Was this change intentional?

```
const [linkedEntityType, setLinkedEntityType] = useState(() => Utils.cond(
    [initialIsReference && initialIsList, () => attributeValue.items[0] ? attributeValue.items[0].entityType : undefined],
    [initialIsReference, () => attributeValue.entityType],
    () => entityTypes[0]
  ))

nawatts: This one was intentional. The new `defaultReferenceEntityType` variable is used slightly differently from the old `linkedEntityType` variable. While an empty entity reference list does not specify an entity type, we still need some entity type to use for added list items. So if we can't determine an entity type from the attribute, we pick one from the available entity types in the workspace.

cahrens (approved):
LGTM, though I have not tested the code. @MatthewBemis please do some bashing.

MatthewBemis (approved):
This seems good to me. I played around with it, no issues.

PR #2897, opened 2022-03-21
[SU-44] Allow resizing sidebar in data pane

nawatts:
With long table/snapshot names, it can be difficult to see what's in the data pane's sidebar. This allows resizing the sidebar. 

Style and interaction inspired by JIRA. ARIA attributes based on [MDN's article on the separator role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/separator_role#focusable_separator).

Had to add a bit to Interactive.js and styles.css to get the unique hover/focus style to work. Interactive applies an outline and currently the only options for overriding it are a blue border or a red border using using `focus-style` or `error-style` classes.

https://user-images.githubusercontent.com/1156625/159368376-396fed8d-de4e-47d4-bc44-5b4a54d1816b.mov

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -1,6 +1,7 @@
 import filesize from 'filesize'
 import _ from 'lodash/fp'
-import { Fragment, useEffect, useImperativeHandle, useState } from 'react'
+import { Fragment, useCallback, useEffect, useImperativeHandle, useState } from 'react'
+import { DraggableCore } from 'react-draggable'
 import { div, h, h3 } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import * as breadcrumbs from 'src/components/breadcrumbs'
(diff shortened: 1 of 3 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -41,14 +42,21 @@ const styles = {
...
   dataTypeSelectionPanel: {
-    flex: 'none', width: 280, backgroundColor: 'white',
+    flex: 'none', backgroundColor: 'white',
     overflow: 'auto',
-    boxShadow: '0 2px 5px 0 rgba(0,0,0,0.25)'
+    boxShadow: '0 2px 5px 0 rgba(0,0,0,0.25)',
+    transition: 'width 100ms'
+  },
+  sidebarSeparator: {
+    width: '0.75rem',
```
cahrens, line 51: I wondered if this should be wider, but change the width doesn't seem to make any difference. Not a big deal, but odd.

cahrens (approved):
@nawatts this is _really_ nice! It looks and works well, and thank you for taking the time to research the aria roles.

I recommend you post about this new component in the terra-ui Slack channel so that everyone is aware of it. We may wish to add it to more of our side bars for consistency.

src/pages/workspaces/workspace/Data.js
nawatts: I was trying to maintain the appearance of the 1rem padding around the main panel's content (the left padding in `styles.tableViewPanel` is reduced to 0.25rem to compensate). The width doesn't have much of a visual effect, it just controls the size of the separator's hover target / drag handle.

cahrens: Oh, that makes more sense. I thought it was the width of the separator itself.

MatthewBemis (approved):
This is sweet! Very fun to play around with. Code LGTM

PR #2902, opened 2022-03-23
[SU-8] Easily share workspace with support

nawatts:
Currently, it's not obvious how to share a workspace with Terra Support. They need to know the Terra Support email. Sharing a workspace is frequently the first step after a user contacts Support, and currently Support has to spend time explaining how to share the workspace.

This adds a "Share with support" switch to the share workspace modal.

Potentially controversial... this removes Terra Support from the collaborators list in the share workspace modal, which prevents a user from seeing or modifying Support's level of access to the workspace. The plan is discourage granting the Support team higher than read access. In cases where they do need additional access, the user will be asked to add an individual Support team member to the workspace.

For cases where Support already has additional access to a workspace or additional access has been granted through an API request, etc. the "Share with support" switch has a tooltip that includes the level of access granted to Support.

Screen Shot 2022-03-23 at 8 59 56 AM

Screen Shot 2022-03-23 at 9 04 16 AM

### DIFF
--- a/src/components/Modal.js
+++ b/src/components/Modal.js
@@ -10,7 +10,7 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
 
-const styles = {
+export const styles = {
   overlay: {
     backgroundColor: 'rgba(0, 0, 0, 0.5)', padding: '2rem 1rem',
     display: 'flex', justifyContent: 'center', alignItems: 'flex-start',
--- a/src/pages/workspaces/workspace/ShareWorkspaceModal.js
+++ b/src/pages/workspaces/workspace/ShareWorkspaceModal.js
@@ -260,7 +260,11 @@ const ShareWorkspaceModal = ({ onDismiss, workspace, workspace: { workspace: { n
     searchValueValid && !searchHasFocus && p(addUserReminder),
     h2({ style: { ...Style.elements.sectionHeader, margin: '1rem 0 0.5rem 0' } }, ['Current Collaborators']),
     div({ ref: list, role: 'list', style: styles.currentCollaboratorsArea }, [
-      h(Fragment, _.map(renderCollaborator, Utils.toIndexPairs(acl))),
+      h(Fragment, _.flow(
+        _.remove(aclEntryIsTerraSupport),
+        Utils.toIndexPairs,
+        _.map(renderCollaborator)
+      )(acl)),
       !loaded && centeredSpinner()
     ]),
     updateError && div({ style: { marginTop: '1rem' } }, [
--- a/src/pages/workspaces/workspace/ShareWorkspaceModal.js
+++ b/src/pages/workspaces/workspace/ShareWorkspaceModal.js
@@ -5,6 +5,7 @@ import { ButtonPrimary, ButtonSecondary, IdContainer, LabeledCheckbox, Link, Sel
 import { centeredSpinner, icon } from 'src/components/icons'
 import { AutocompleteTextInput } from 'src/components/input'
 import Modal, { styles as modalStyles } from 'src/components/Modal'
+import TooltipTrigger from 'src/components/TooltipTrigger'
 import { Ajax } from 'src/libs/ajax'
 import { getUser } from 'src/libs/auth'
 import colors from 'src/libs/colors'
@@ -218,7 +219,7 @@ const ShareWorkspaceModal = ({ onDismiss, workspace, workspace: { workspace: { n
   // The email they entered will appear in the ACL stored in this component's state until updates are saved.
   // We want the share with support switch to function with any variation of the support email.
   const aclEntryIsTerraSupport = ({ email }) => _.toLower(email) === _.toLower(terraSupportEmail)
-  const isTerraSupportInAcl = !!_.find(aclEntryIsTerraSupport, acl)
+  const terraSupportAccess = _.find(aclEntryIsTerraSupport, acl)
   const addTerraSupportToAcl = () => addCollaborator(terraSupportEmail)
   const removeTerraSupportFromAcl = () => setAcl(_.remove(aclEntryIsTerraSupport))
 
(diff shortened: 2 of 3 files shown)


cahrens:
@nawatts can you create a PR deployment for this?

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -211,14 +214,19 @@ const ShareWorkspaceModal = ({ onDismiss, workspace, workspace: { workspace: { n
     }
   }
 
+  // The get workspace ACL API endpoint returns emails capitalized as they are registered in Terra.
```
cahrens, line 217: Thanks for the excellent comment!

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -253,13 +261,52 @@ const ShareWorkspaceModal = ({ onDismiss, workspace, workspace: { workspace: { n
...
     ]),
     updateError && div({ style: { marginTop: '1rem' } }, [
       div(['An error occurred:']),
       updateError
     ]),
+    div({ style: { ...modalStyles.buttonRow, justifyContent: 'space-between' } }, [
+      h(IdContainer, [
+        id => h(TooltipTrigger, {
+          content: terraSupportAccess ?
+            `Terra Support has ${_.toLower(terraSupportAccess.accessLevel)} access to this workspace` :
```
cahrens: I found the tooltip a bit confusing when granting access. When I toggle the switch to True, the tooltip immediately says that Terra Support has reader access. I was confused about whether I needed to press Save, or if the operation truly happened when the switch was changed (I experimented to confirm that I actually do need to press Save).

I think it would make more sense for the tooltip to reflect the actual state, or have a different message for this intermediary state (perhaps "Saving will grant Terra Support reader access….", or "Terra Support will have reader access…").

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -253,13 +261,52 @@ const ShareWorkspaceModal = ({ onDismiss, workspace, workspace: { workspace: { n
...
+              checked: !!terraSupportAccess,
+              onChange: checked => {
+                if (checked) {
+                  addTerraSupportToAcl()
+                } else {
+                  removeTerraSupportFromAcl()
+                }
+              }
+            }),
+            span({ style: { marginLeft: '1ch' } }, 'Share with support')
```
cahrens: Did UX give the design with the label after the switch? Personally, I think it would be more standard to have the label before the control.

nawatts: Yes, the label is after the switch in the UX design attached to [SU-8](https://broadworkbench.atlassian.net/browse/SU-8).

I could go either way here. I agree that "Share with support? True/False" reads more naturally. On the other hand, checkboxes usually appear before their label and the switch is a fancy checkbox. We don't currently have an established pattern for this in Terra since the Switch component is only used in one other place (entity editor in data tables) and there it doesn't have a label other than the True/False in the switch itself.

src/pages/workspaces/workspace/ShareWorkspaceModal.js
nawatts: Good call. The necessity of clicking "Save" has caused confusion in the past ([SUP-366](https://broadworkbench.atlassian.net/browse/SUP-366)).

src/pages/workspaces/workspace/ShareWorkspaceModal.js
nawatts: Order changed in 596956da4fd55460df7c7de70c786934bba40962.

src/pages/workspaces/workspace/ShareWorkspaceModal.js
nawatts: Done in https://github.com/DataBiosphere/terra-ui/pull/2902/commits/e83afd6027c1654a15850e0ad8288ad8e9859947.

src/pages/workspaces/workspace/ShareWorkspaceModal.js
nawatts: Screen Shot 2022-03-24 at 10 44 22 AM

cahrens (approved):
Thanks, I think this looks much better with the label first, and the more descriptive tooltips are excellent.

nawatts:
Thanks for the suggestion on the tooltips. I think that prevented a lot of future confusion.

src/pages/workspaces/workspace/ShareWorkspaceModal.js
```diff
@@ -253,13 +262,59 @@ const ShareWorkspaceModal = ({ onDismiss, workspace, workspace: { workspace: { n
     searchValueValid && !searchHasFocus && p(addUserReminder),
     h2({ style: { ...Style.elements.sectionHeader, margin: '1rem 0 0.5rem 0' } }, ['Current Collaborators']),
     div({ ref: list, role: 'list', style: styles.currentCollaboratorsArea }, [
-      h(Fragment, _.map(renderCollaborator, Utils.toIndexPairs(acl))),
+      h(Fragment, _.flow(
+        _.remove(aclEntryIsTerraSupport),
```
petesantos, line 266: I like the naming, very clear!

src/components/Modal.js
```diff
@@ -10,7 +10,7 @@ import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 
 
-const styles = {
+export const styles = {
```
petesantos, line 13: nit: This is my only concern in this change. Exporting the style makes it a bit more difficult to change in the future if it is used in a lot of places. But I don't have a better solution in mind right now.

petesantos (approved):
LGTM. Tested by clicking around the modal and trying different states

src/components/Modal.js
nawatts: To add the switch to the modal footer, I had to set `showButtons: false` on the Modal and render a replacement footer/buttons. I added this export thinking that, this way, the footer here would stay up to date with any changes to the base modal footer styles. I can remove this export and copy and paste the styles, but that has its own drawbacks.

PR #2903, opened 2022-03-23
[WOR-182] Use UUID on workspace name to help avoid collisions with existing wor…

cahrens:
We have been seeing sporadic integration tests failures due to the inability to create a workspace. Looking in kibana, this is happening because the workspace we are attempting to create already exists. It seems that we are "leaking" workspaces, and our logic for getting a random workspace name was pretty weak.

`message:[ERROR] [13:26:54.249] [scala-execution-context-global-74] o.b.d.r.dataaccess.SlickDataSource - Transaction with temporary tables failed for (Workspace). Message: ErrorReport(rawls,Workspace saturn-integration-test-dev/test-workspace-77212 already exists,Some(409 Conflict),List(),List(),None)`

Switching to a UUID to alleviate the immediate issue, but we also need to clean up existing old workspaces and find a way to avoid this issue in the future. It is possible that people running the "test-flaky" in the last few months have exacerbated this problem-- I know that I often kill a running test process, and I imagine workspaces don't get cleaned up in that case.

````
MySQL [rawls]> select concat(extract(year from created_date),'/', extract(month from created_date)) as dt, count(*) from WORKSPACE where name like 'test-workspace-%' group by dt order by dt desc;
+---------+----------+
| dt      | count(*) |
+---------+----------+
| 2022/3  |      252 |
| 2022/2  |      381 |
| 2022/1  |      183 |
| 2021/9  |       54 |
| 2021/8  |       31 |
| 2021/7  |       17 |
| 2021/6  |       27 |
| 2021/5  |       31 |
| 2021/4  |       23 |
| 2021/3  |       18 |
| 2021/2  |       16 |
| 2021/12 |      105 |
| 2021/11 |       15 |
| 2021/10 |       35 |
| 2021/1  |       31 |
| 2020/9  |       53 |
| 2020/8  |       12 |
| 2020/7  |        9 |
| 2020/6  |       12 |
| 2020/5  |       18 |
| 2020/4  |        1 |
| 2020/3  |       13 |
| 2020/2  |       43 |
| 2020/12 |       29 |
| 2020/11 |        5 |
| 2020/10 |        7 |
| 2020/1  |       79 |
| 2019/12 |       20 |
| 2019/11 |       18 |
| 2019/10 |        1 |
+---------+----------+

### DIFF
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -1,5 +1,6 @@
 const rawConsole = require('console')
 const _ = require('lodash/fp')
+const uuid = require('uuid')
 
 const { click, clickable, dismissNotifications, findText, signIntoTerra, waitForNoSpinners } = require('./integration-utils')
 const { fetchLyle } = require('./lyle-utils')
@@ -22,7 +23,7 @@ const withSignedInPage = fn => async options => {
 
 const clipToken = str => str.toString().substr(-10, 10)
 
-const getTestWorkspaceName = () => `test-workspace-${Math.floor(Math.random() * 100000)}`
+const getTestWorkspaceName = () => `int-test-workspace-${uuid.v4()}`
 
 const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
   const workspaceName = getTestWorkspaceName()
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -23,7 +23,7 @@ const withSignedInPage = fn => async options => {
 
 const clipToken = str => str.toString().substr(-10, 10)
 
-const getTestWorkspaceName = () => `int-test-workspace-${uuid.v4()}`
+const getTestWorkspaceName = () => `terra-ui-test-workspace-${uuid.v4()}`
 
 const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
   const workspaceName = getTestWorkspaceName()

### CODE REVIEW

File: integration-tests/utils/integration-helpers.js [hunk 1]
```diff
@@ -22,7 +23,7 @@ const withSignedInPage = fn => async options => {
 
 const clipToken = str => str.toString().substr(-10, 10)
 
-const getTestWorkspaceName = () => `test-workspace-${Math.floor(Math.random() * 100000)}`
+const getTestWorkspaceName = () => `int-test-workspace-${uuid.v4()}`
```
cahrens: Added "int" prefix to add a clue that this workspace was created by integration tests. Of course, when we do cleanup we will also want to take into account the namespace.

````
MySQL [rawls]> select namespace, count(*) from WORKSPACE where name like 'test-workspace-%' group by namespace;
+-----------------------------+----------+
| namespace                   | count(*) |
+-----------------------------+----------+
| aa-test0401917-a            |        2 |
| general-dev-billing-account |      379 |
| saturn-integration-test-dev |     1155 |
| werqerweeer                 |        3 |
+-----------------------------+----------+

cahrens: Example workspace name: `int-test-workspace-cedafbc4-e7a2-4069-9031-2ff99857de6f`

ungwudik: Probably worth further noting that it's a _UI_ int test workspace, if that's not an overkill.

ungwudik (approved):
LGTM

integration-tests/utils/integration-helpers.js
kyuksel: 1. Do we need the entire ID like `cedafbc4-e7a2-4069-9031-2ff99857de6f`? Just six hex digits (e.g. `cedafbc4`) would create a space of ~4 billion IDs, which sounds sufficient to me.
2. FWIW my brain interpreted `int` as integer and kept thinking what integer has to do with it  😊  I'd vote for something like `terra-ui-test-workspace-cedafbc4`. Since a UI unit test wouldn't typically create a persisted workspace, I don't think we need to specify `integration`.

kyuksel (approved):
LGTM. A couple of nits inline. Didn't test.

integration-tests/utils/integration-helpers.js
cahrens: I made it `terra-ui-test-worskapce-<UUID>`. I don't see a compelling reason to truncate the UUID (we don't elsewhere where it is used) as long as we are not hitting the maximum allowed length of the workspace name.

PR #2904, opened 2022-03-23
[IA-3262] New eslint rules around hoisting

kyuksel:
Turn on the following eslint rules to the extent possible

- [no-use-before-define](https://eslint.org/docs/rules/no-use-before-define)

- [vars-on-top](https://eslint.org/docs/rules/vars-on-top)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -49,6 +49,8 @@ module.exports = {
     'arrow-parens': ['warn', 'as-needed'],
     'arrow-spacing': 'warn',
     'no-duplicate-imports': 'warn',
+    // TODO: Set 'variables' to 'true' after fixing the existing issues
+    'no-use-before-define': ['warn', { 'functions': true, 'classes': true, 'variables': false }],
     'no-useless-rename': 'warn',
     'no-var': 'warn',
     'object-shorthand': 'warn',
@@ -59,6 +61,7 @@ module.exports = {
     'prefer-spread': 'warn',
     'rest-spread-spacing': 'warn',
     'template-curly-spacing': 'warn',
+    'vars-on-top': 'warn',
 
 
     'lodash-fp/consistent-name': ['warn', '_'],


.eslintrc.js
```diff
@@ -49,6 +49,8 @@ module.exports = {
     'arrow-parens': ['warn', 'as-needed'],
     'arrow-spacing': 'warn',
     'no-duplicate-imports': 'warn',
+    // TODO: Set 'variables' to 'true' after fixing the existing issues
+    'no-use-before-define': ['warn', { 'functions': true, 'classes': true, 'variables': false }],
```
kyuksel, line 53: Setting `variables` to `true` results in ~90 warnings so couldn't do that at this time and figured I wouldn't make perfect the enemy of good  🤷🏼

pshapiro4broad (approved):
LGTM

.eslintrc.js
```diff
@@ -59,6 +61,7 @@ module.exports = {
     'prefer-spread': 'warn',
     'rest-spread-spacing': 'warn',
     'template-curly-spacing': 'warn',
+    'vars-on-top': 'warn',
```
cahrens, line 64: Having come from a place where we had to define all variables at the top of function blocks, I've been enjoying the freedom/clarity to define my variables closer to where they are used. So I personally am not a fan of this warning, if it means what I think it does.

kyuksel: It requires variable declarations to be at the top of their scope, not necessarily at the top of the file. All of the codebase complies with it as is so you should be able to continue enjoying the freedom to define variables closer to where they are used?  😊

cahrens: Oh great! I was thinking it would be more disruptive.

petesantos: I think this is more about preventing potential issues with hoisting. There was some confusion around this in a [slack thread](https://broadinstitute.slack.com/archives/C01EHNUM73R/p1647618956137379) a few days ago.

petesantos (approved):

PR #2905, opened 2022-03-23
[IA-3203] Add sars-cov-2 reference to igv.js

davidangb:
adds ASM985889v3 - the sars-cov-2 reference - to our embedded IGV.js.

This does not upgrade IGV.js, which we really should do too, but that's out of scope of this PR.

Tested by running locally:
Screenshot (12)

I can't vouch for the science, but I see a genome showing up.


kyuksel (approved):

Qi77Qi (approved):
thanks @davidangb ..I don't have much context to this..but where do you get `ASM985889v3`? :)

davidangb:
@Qi77Qi I looked at the genomes available in IGV.js by loading https://igv.org/app/, and copy-pasted from there.

PR #2906, opened 2022-03-24
Use yarn pnp [POL-215]

zarsky-broad:
We should talk through this synchronously. See https://yarnpkg.com/features/pnp and https://yarnpkg.com/features/workspaces

[Here's a diff with the zip files filtered out](https://github.com/DataBiosphere/terra-ui/pull/2906/files?file-filters%5B%5D=.cjs&file-filters%5B%5D=.css&file-filters%5B%5D=.html&file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.mjs&file-filters%5B%5D=.yml&file-filters%5B%5D=dotfile&show-deleted-files=true&show-viewed-files=true)


zarsky-broad:
The LGTM alerts are from generated pnp files, I added config to ignore them but it doesn't seem to take effect until it's merged

kyuksel (commented):
Will be group-reviewed online.

kyuksel (approved):
Approval through mob review

.circleci/config.yml
```diff
@@ -186,17 +180,17 @@ jobs:
...
-            - deps8-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
-            - deps8-
-      - run: yarn install --immutable
-      - save_cache:
-          key: deps8-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
-          paths:
-            - node_modules
-            - .yarn/cache
+            - deps-{{ .Branch }}-{{ checksum ".pnp.cjs" }}
+            - deps-{{ .Branch }}-
```
cahrens, line 184: What is line 184 for?

cahrens:
The integration file yarn.lock file is now gone, and everything was moved into the top-level yarn.lock file, correct? So does this mean that integration tests and source code cannot depend on different versions of the same library (direct dependencies)?

.circleci/config.yml
zarsky-broad: To allow for a partial restore if there isn't an exact hit. It'll then run the new install and save the updated cache entry with the new key.

zarsky-broad:
> The integration file yarn.lock file is now gone, and everything was moved into the top-level yarn.lock file, correct? So does this mean that integration tests and source code cannot depend on different versions of the same library (direct dependencies)?

No, yarn.lock allows multiple versions of a dependency if there are incompatible resolution ranges.

cahrens (approved):

PR #2837, opened 2022-02-25
[SU-4] Limit height of data table page and scroll data selection list

nawatts:
Currently, when a workspace has many data tables or snapshots, the selected table's horizontal scroll bar and pagination controls are pushed to the bottom of the data selection list.

This limits the height of the page so that the scroll bar and pagination controls are visible in the window and makes the data selection list independently scrollable.

## Before

Have to scroll the window to see the horizontal scroll bar and pagination controls. When those are scrolled into view, column headers and controls not visible.

Screen Shot 2022-02-25 at 7 56 57 AM

Screen Shot 2022-02-25 at 7 57 06 AM

## After

Horizontal scroll bar and pagination controls are always visible.

Screen Shot 2022-02-25 at 7 58 33 AM

Screen Shot 2022-02-25 at 7 59 40 AM

Data selection list scrolls independently.

Screen Shot 2022-02-25 at 7 59 43 AM

PR #2908, opened 2022-03-24
[IA-3269] allow scrolling in igv viewer

jordansafer:
For viewing IGVs with multiple sequence files, since there's no scrolling in the igv viewer, it's not currently possible to scroll down and view all the sequence files. This change adds y-scrolling.

https://user-images.githubusercontent.com/8800717/160001114-55602c7e-d0d8-4712-81e5-684970c34420.mov


nawatts:
This looks like another unintended consequence of #2837 😬

jordansafer:
> This looks like another unintended consequence of #2837 😬

Yup, just tested and confirmed, the page scrolling was used here before the change, but now it runs past the limit-

davidangb (approved):
thank you!

petesantos (approved):

PR #2909, opened 2022-03-24
[IA-3270] Upgrade IGV version to 2.11.2 (latest)

jordansafer:
Upgrade IGV version to latest- Tested in local UI

Screen Shot 2022-03-24 at 4 11 27 PM


davidangb (approved):
works for me running locally, though I don't know enough about IGV to detect any subtle issues.

kyuksel (approved):
LGTM. Didn't test.
Thanks for taking care of this!

PR #2910, opened 2022-03-24
[IA-3080] ux feedback

jdcanas:
The PR addresses the following bullet points in this ticket https://broadworkbench.atlassian.net/browse/IA-3080
- We need to add messaging for the individual hovers during different states when env creation or resume (ex. Precreating, starting, running, etc. See “Env status tooltip” screenshot attached
Env status tooltip

- Update side panel icons so that they get highlighted when hovering, individual hovers for each app env icon (so that they look like buttons)
- Clicking on each app icon brings up only its corresponding cloud env page only, and if user clicks on the cloud icon itself, then it will bring up the entire cloud env modal with all apps displayed
- Make side panel icons slightly bigger to improve usability
- Change the “Launch ______” name to “Open ______” (Note: this exists in the edit notebook view, also “do you want to launch” prompt, Context bar change only?)
- Change the name of Launch button with .rmd files to Open, and also the .ipynb from Edit to Open so that they both match
- Also change from Launch to Open when creating RStudio and Galaxy instances (pop up window at the right hand side of the page) - “Open X” where X = Galaxy, RStudio or other future apps

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -26,8 +26,8 @@ import {
   computeStyles, defaultAutopauseThreshold, defaultComputeRegion, defaultComputeZone, defaultDataprocMachineType, defaultDataprocMasterDiskSize,
   defaultDataprocWorkerDiskSize, defaultGceBootDiskSize, defaultGceMachineType, defaultGcePersistentDiskSize, defaultGpuType, defaultLocation,
   defaultNumDataprocPreemptibleWorkers, defaultNumDataprocWorkers, defaultNumGpus, displayNameForGpuType, findMachineType, getAutopauseThreshold,
-  getCurrentRuntime, getDefaultMachineType, getPersistentDiskCostMonthly, getValidGpuTypes, getValidGpuTypesForZone, isAutopauseEnabled, RadioBlock,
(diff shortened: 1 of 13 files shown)


src/components/ComputeModal.js
```diff
@@ -783,6 +783,9 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       ],
       () => ({ disabled: !hasChanges() || !!errors, tooltip: Utils.summarizeErrors(errors) })
     )
+
+    const isUpdateDisabled = getIsRuntimeBusy(currentRuntimeDetails)
```
jdcanas, line 787: this is a moderate paradigm-shift that's not captured in bullet points at the top. Essentially, we want to start shifting responsibility on the compute modal to disable updating/deleting if the runtime is in a 'busy' state, and allow the modal to open in a read-only state still. Today, we disable viewing the compute configuration entirely and don't offer this read-only view.

jdcanas:
closing temporarily, got some product feedback i need to incorporate. Should be ready for review again on monday

cahrens:
> closing temporarily, got some product feedback i need to incorporate. Should be ready for review again on monday

Good timing-- I was just starting to look at the PR! Ping me when it's ready.

PR #2911, opened 2022-03-24
[IA-3271] fix memory leak in igv viewer

jordansafer:
Fixed a memory leak coming in and out of the IGV viewer page.

### Before (3 visits to IGV viewer create 3 allocation for IGV lib, about 10 MB each time)
Screen Shot 2022-03-24 at 4 50 37 PM

### After (3 visits to the IGV viewer, no allocations left after the viewer is exited each time)
Screen Shot 2022-03-24 at 4 51 01 PM

### DIFF
--- a/src/components/IGVBrowser.js
+++ b/src/components/IGVBrowser.js
@@ -20,7 +20,7 @@ const IGVBrowser = _.flow(
   const containerRef = useRef()
   const signal = useCancellation()
   const [loadingIgv, setLoadingIgv] = useState(true)
-  const [igvLibrary, setIgvLibrary] = useState(null)
+  const igvLibrary = useRef()
 
   useOnMount(() => {
     const igvSetup = async () => {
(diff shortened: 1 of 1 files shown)


zarsky-broad:
Instead of calling it from the onClick, you can return a cleanup function in the onMount that'll be run at unmount time.
(incidentally, I think you want to use a ref instead of state to avoid an extra render)

davidangb (approved):
nice find on this!

jordansafer:
> Instead of calling it from the onClick, you can return a cleanup function in the onMount that'll be run at unmount time. (incidentally, I think you want to use a ref instead of state to avoid an extra render)

Thanks! Addressed, latest memory profile test post-fix:
Screen Shot 2022-03-24 at 7 16 41 PM

(looks like there may be a much smaller ~100kb leak on the workspace data page, but I haven't investigated deeply to see if this clears out)

kyuksel (commented):
LGTM but I'll let @zarsky-broad re-review and 👍🏼 
Thanks for taking care of this!

zarsky-broad (approved):
Code looks good.

PR #2912, opened 2022-03-24
Add janitor "test" to delete orphaned workspaces (WOR-210).

cahrens:
This adds a janitor "test" to the integration test suite so that we can clean up leaked resources. I am using it to delete workspaces created over 2 days ago that were generated by running integration tests.

Workspaces can be leaked by when tests are killed prematurely, for example when folks run the tests locally and kill the running process (something I've certainly been known to do!!). They also apparently happen on occasion when ci runs tests, as evidenced by the fact that we have a small number in both the alpha and staging environments (I will be cleaning them up soon… ones with the new name prefix will automatically get cleaned up when this test runs in alpha/staging, and ones with the [older name prefix](https://github.com/DataBiosphere/terra-ui/pull/2903) I will clean up by running a modified version of the test directly against those environments). More details are available in https://broadworkbench.atlassian.net/browse/WOR-210.

When I first starting deleting large numbers of workspaces (700ish), I was hitting problems with delete workspace returning an error. It turns out that we were receiving 404s from Sam's deleteResource method (due to database contention/timing out/retries), and the Rawls delete workspace method considered that an error. Therefore I held this PR [until that issue was fixed](https://github.com/broadinstitute/rawls/pull/1673), to prevent introducing more flakiness to our tests.

### DIFF
--- a/integration-tests/tests/janitor.js
+++ b/integration-tests/tests/janitor.js
@@ -5,7 +5,7 @@ const { dismissNotifications, signIntoTerra } = require('../utils/integration-ut
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
-const olderThanDays = 200
+const olderThanDays = 100
 const workspacePrefix = 'test-workspace-' // TODO: share with other location
 
 const runJanitor = withUserToken(async ({ billingProject, page, testUrl, token }) => {
@@ -30,7 +30,7 @@ const runJanitor = withUserToken(async ({ billingProject, page, testUrl, token }
       await page.evaluate((namespace, name) => window.Ajax().Workspaces.workspace(namespace, name).delete(), namespace, name)
       rawConsole.info(`Deleted old workspace: ${name}`)
     } catch (e) {
-      throw Error(`Failed to delete old workspace: ${name} with billing project ${namespace}`)
+      rawConsole.info(`Failed to delete old workspace: ${name} with billing project ${namespace}`)
     }
   }, oldWorkspaces))
 })
--- a/integration-tests/tests/janitor.js
+++ b/integration-tests/tests/janitor.js
@@ -6,7 +6,7 @@ const { dismissNotifications, signIntoTerra } = require('../utils/integration-ut
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
-const olderThanDays = 0
+const olderThanDays = 2
 
 const runJanitor = withUserToken(async ({ billingProject, page, testUrl, token }) => {
   // Sign into Terra so we have the correct credentials.
@@ -15,7 +15,8 @@ const runJanitor = withUserToken(async ({ billingProject, page, testUrl, token }
   await dismissNotifications(page)
 
   // Delete old orphaned workspaces. These can result from local integration test runs where the
-  // process was prematurely terminated.
+  // process was prematurely terminated, but a few also appear in alpha and staging indicating that
+  // ci test runs also sometimes leak workspaces.
   const workspaces = await page.evaluate(async () => await window.Ajax().Workspaces.list())
   const oldWorkspaces = _.filter(({ workspace: { namespace, name, createdDate } }) => {
     const age = dateFns.differenceInDays(new Date(createdDate), new Date())
--- a/integration-tests/tests/janitor.js
+++ b/integration-tests/tests/janitor.js
@@ -4,6 +4,7 @@ const _ = require('lodash/fp')
 const { dismissNotifications, signIntoTerra } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
+
 const olderThanDays = 365
 const workspacePrefix = 'test-workspace-' // TODO: share with other location
 
@@ -16,8 +17,8 @@ const runJanitor = withUserToken(async ({ billingProject, page, testUrl, token }
   // Delete old orphaned workspaces. These can result from local integration test runs where the
   // process was prematurely terminated.
   const workspaces = await page.evaluate(async () => await window.Ajax().Workspaces.list())
-  const oldWorkspaces = _.filter( ( { workspace: {namespace, name, lastModified} } ) => {
-      return namespace === billingProject &&
+  const oldWorkspaces = _.filter(({ workspace: { namespace, name, lastModified } }) => {
+    return namespace === billingProject &&
       _.startsWith(workspacePrefix, name) &&
       dateFns.differenceInDays(new Date(lastModified), new Date()) > olderThanDays
   }, workspaces)
--- a/integration-tests/tests/janitor.js
+++ b/integration-tests/tests/janitor.js
@@ -1,12 +1,12 @@
 const rawConsole = require('console')
 const dateFns = require('date-fns/fp')
 const _ = require('lodash/fp')
+const { testWorkspaceNamePrefix } = require('../utils/integration-helpers')
 const { dismissNotifications, signIntoTerra } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
-const olderThanDays = 100
-const workspacePrefix = 'test-workspace-' // TODO: share with other location
+const olderThanDays = 0
 
 const runJanitor = withUserToken(async ({ billingProject, page, testUrl, token }) => {
   // Sign into Terra so we have the correct credentials.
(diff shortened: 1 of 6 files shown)


cahrens:
The janitor passed in commit [36ef72f](https://github.com/DataBiosphere/terra-ui/pull/2912/commits/36ef72f6c9db0fa3a00c972850114daaa383bf16), deleting all 13 workspaces from 3/23. 2 other tests failed, but they did not fail with workspace deletion and at least one of the tests finished before the janitor ran.

````
…
Deleted old workspace: terra-ui-test-workspace-8e3c033a-5e82-4b63-b6c9-6cd7d2fcc5ad
page.http.res 202 DELETE https://rawls.dsde-dev.broadinstitute.org/api/workspaces/saturn-integration-test-dev/terra-ui-test-workspace-60b55a27-0456-40f7-a2b3-a1d79f9a7650
Deleted old workspace: terra-ui-test-workspace-60b55a27-0456-40f7-a2b3-a1d79f9a7650
page.http.res 202 DELETE https://rawls.dsde-dev.broadinstitute.org/api/workspaces/saturn-integration-test-dev/terra-ui-test-workspace-cb416730-8e3c-46d7-946d-06dd37e4e8c3
Deleted old workspace: terra-ui-test-workspace-cb416730-8e3c-46d7-946d-06dd37e4e8c3
page.http.res 202 DELETE https://rawls.dsde-dev.broadinstitute.org/api/workspaces/saturn-integration-test-dev/terra-ui-test-workspace-030efaa4-d394-4025-b895-af23b026b780
Deleted old workspace: terra-ui-test-workspace-030efaa4-d394-4025-b895-af23b026b780
page.http.res 202 DELETE https://rawls.dsde-dev.broadinstitute.org/api/workspaces/saturn-integration-test-dev/terra-ui-test-workspace-c16dbd18-6d38-4546-b350-3c663a77776e
Deleted old workspace: terra-ui-test-workspace-c16dbd18-6d38-4546-b350-3c663a77776e
 PASS  jest/janitor.integration-test.js (45.761 s)

**Comment by cahrens** [comment]
Most recent deletions can be seen [here](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/8620/workflows/92cce0f0-bb8f-48dd-9a3d-d5eb9018c691/jobs/39464/parallel-runs/3?filterBy=ALL&invite=true#step-108-480).

File: integration-tests/tests/janitor.js [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+  }, workspaces)
+
+  rawConsole.log(`Deleting ${oldWorkspaces.length} workspaces with prefix "${testWorkspaceNamePrefix}" created more than ${olderThanDays} days ago.`)
+
+  return Promise.all(_.map(async ({ workspace: { namespace, name } }) => {
+    try {
+      await page.evaluate((namespace, name) => window.Ajax().Workspaces.workspace(namespace, name).delete(), namespace, name)
+      rawConsole.info(`Deleted old workspace: ${name}`)
+    } catch (e) {
+      rawConsole.info(`Failed to delete old workspace: ${name} with billing project ${namespace}`)
```
cahrens, line 34: Note that I don't fail this if we can't delete the workspace. There is a chance that some workspaces may build up that we can't delete. For example, there are 5 workspaces in dev with the old prefix that I can't delete despite repeated attempts-- when I try to delete them, I got a google api permissions error.

If that happens going forward, we will try to delete those bad workspaces every time the test runs. If this retrying becomes an issue or is too noisy, we can either figure out how to clean up those workspaces manually, or add some sort of tracking mechanism so that we don't continue attempting to retry the delete. A simple heuristic could be that we only attempt to delete workspaces within a certain window of time (for instance, those created between 2 and 9 days ago).

integration-tests/tests/janitor.js
```diff
@@ -0,0 +1,44 @@
...
+  const workspaces = await page.evaluate(async () => await window.Ajax().Workspaces.list())
+  const oldWorkspaces = _.filter(({ workspace: { namespace, name, createdDate } }) => {
+    const age = dateFns.differenceInDays(new Date(createdDate), new Date())
+    // rawConsole.info(`${namespace} ${name}, age ${age} days`)
+    return namespace === billingProject && _.startsWith(testWorkspaceNamePrefix, name) && age > olderThanDays
+  }, workspaces)
+
+  rawConsole.log(`Deleting ${oldWorkspaces.length} workspaces with prefix "${testWorkspaceNamePrefix}" created more than ${olderThanDays} days ago.`)
+
+  return Promise.all(_.map(async ({ workspace: { namespace, name } }) => {
```
nawatts, line 29: Is it worth making these requests in batches vs all at once in case there's a large number of old workspaces?

nawatts (approved):
I like the simplicity of running this cleanup as a test.

integration-tests/tests/janitor.js
cahrens: @nawatts unfortunately there's no batch delete API endpoint.

kyuksel:
Didn't get to review the code but just a couple of high-level thoughts/questions:

1. I like the simplicity and unforgettableness of running this cleanup as a test. However, I also wonder if it will cause the cleanup to be run too many times since tests are run for every PR commit pushed, etc.
2. Have we been able to flake-test this with RUNS set to something like 100?

integration-tests/tests/janitor.js
nawatts: I meant like making the individual requests in batches of a limited size vs sending them all at once. Though I guess that since the test runs regularly, there's unlikely to be a very large number of workspaces that need to be deleted.

cahrens:
> 

Thanks for the thoughts, @kyuksel.

1. I don't think there is any problem with running the test on every commit. While not strictly necessary, the test is so fast (when not deleting hundreds of workspaces) that it doesn't bloat the test suite. And by running it very often, we are guaranteed to not get into the state of having hundreds of workspaces to delete.  On the last commit, it finished in 12.204 seconds, deleting 4 workspaces.
2. Thanks for the reminder about flakes! I just ran it and the test passed with RUNS set to 100. It passed in 60.484 seconds (with no workspaces to delete).

integration-tests/tests/janitor.js
cahrens: I don't think there's any harm in sending them all at once. As you noted, the number is unlikely to ever be very large since we run it often (the number of workspaces we are leaking per day in dev is in the range of 3-30). I did run this to delete the old workspaces, and then I was deleting 100s at a time. I hit the 404 issue from sam that I addressed in the deleteWorkspace endpoint, but otherwise there weren't problems. And it was actually good that it revealed the 404 issue, which had caused very sporadic failures both on swatomation tests and integration tests in the past.

marctalbott (approved):

kyuksel:
> It turns out that we were receiving 404s from Sam's deleteResource method (due to database contention/timing out/retries)

It's interesting that Sam returned 404s due to database contention/timing out/retries. I'd have expected a different error code in those cases.

cahrens:
> It's interesting that Sam returned 404s due to database contention/timing out/retries. I'd have expected a different error code in those cases.

It's because rawls retries with the timeout, and then at some point the item was deleted (resulting in a 404). See https://broadinstitute.slack.com/archives/CNXF2BAPJ/p1648237846362279 for details if you are interested.

integration-tests/tests/janitor.js
```diff
@@ -0,0 +1,44 @@
...
+
+const runJanitor = withUserToken(async ({ billingProject, page, testUrl, token }) => {
+  // Sign into Terra so we have the correct credentials.
+  await page.goto(testUrl)
+  await signIntoTerra(page, token)
+  await dismissNotifications(page)
+
+  // Delete old orphaned workspaces. These can result from local integration test runs where the
+  // process was prematurely terminated, but a few also appear in alpha and staging indicating that
+  // ci test runs also sometimes leak workspaces.
```
kyuksel, line 19: It might be good to move this comment higher, after the `require`s, and perhaps to also include a sentence about why clean-up was implemented as a test.

cahrens: I can imagine the test being used to clean up other things in the future, so I think it's best to keep the comment near the related code. And the history of all of this is accessible via git (ticket, PR, etc.).

kyuksel: It helps to be explicit about reasonings, especially in codebases contributed to by so many people but I won't push further 😊

integration-tests/tests/janitor.js
kyuksel: Should it be a `rawConsole.warning(…)`?

kyuksel (approved):
Thanks for taking the initiative on this! One question but not a blocker.

cahrens:
@kyuksel there is no rawConsole.warning method.

PR #2913, opened 2022-03-28
Add lgtm config [pre- POL-215]

zarsky-broad:
Merging this before https://github.com/DataBiosphere/terra-ui/pull/2906 should handle the lgtm warnings.

PR #2914, opened 2022-03-28
[IA-3080] UX feedback

jdcanas:
I had to re-create this because I rebased the branch... It wouldn't let me re-open the old PR for some reason

The PR addresses the following bullet points in this ticket https://broadworkbench.atlassian.net/browse/IA-3080
- We need to add messaging for the individual hovers during different states when env creation or resume (ex. Precreating, starting, running, etc. See “Env status tooltip” screenshot attached
Env status tooltip

- Update side panel icons so that they get highlighted when hovering, individual hovers for each app env icon (so that they look like buttons)
- Clicking on each app icon brings up only its corresponding cloud env page only, and if user clicks on the cloud icon itself, then it will bring up the entire cloud env modal with all apps displayed
- Make side panel icons slightly bigger to improve usability
- Change the “Launch ______” name to “Open ______” (Note: this exists in the edit notebook view, also “do you want to launch” prompt, Context bar change only?)
- Change the name of Launch button with .rmd files to Open, and also the .ipynb from Edit to Open so that they both match
- Also change from Launch to Open when creating RStudio and Galaxy instances (pop up window at the right hand side of the page) - “Open X” where X = Galaxy, RStudio or other future apps

### DIFF
--- a/src/components/AnalysisModal.js
+++ b/src/components/AnalysisModal.js
@@ -108,7 +108,6 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
     const renderComputeModal = () => h(ComputeModalBase, {
       isOpen: currentTool === tools.Jupyter.label || currentTool === tools.RStudio.label,
       isAnalysisMode: true,
-      shouldHideCloseButton: true,
       workspace,
       tool: currentTool,
       runtimes,
@@ -126,7 +125,6 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
 
     const renderAppModal = (appModalBase, toolLabel) => h(appModalBase, {
       isOpen: viewMode === toolLabel,
-      shouldHideCloseButton: true,
       workspace,
       apps,
       appDataDisks,
--- a/src/components/AnalysisModal.js
+++ b/src/components/AnalysisModal.js
@@ -101,13 +101,14 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
     const getEnvironmentView = () => Utils.switchCase(currentTool,
       [tools.Jupyter.label, renderComputeModal],
       [tools.RStudio.label, renderComputeModal],
-      [tools.galaxy.label, () => renderAppModal(GalaxyModalBase, tools.galaxy.label)],
-      [tools.cromwell.label, () => renderAppModal(CromwellModalBase, tools.cromwell.label)]
+      [tools.Galaxy.label, () => renderAppModal(GalaxyModalBase, tools.Galaxy.label)],
+      [tools.Cromwell.label, () => renderAppModal(CromwellModalBase, tools.Cromwell.label)]
     )
 
     const renderComputeModal = () => h(ComputeModalBase, {
       isOpen: currentTool === tools.Jupyter.label || currentTool === tools.RStudio.label,
       isAnalysisMode: true,
+      shouldHideCloseButton: true,
       workspace,
       tool: currentTool,
       runtimes,
@@ -125,7 +126,7 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
 
     const renderAppModal = (appModalBase, toolLabel) => h(appModalBase, {
       isOpen: viewMode === toolLabel,
-      isAnalysisMode: true,
+      shouldHideCloseButton: true,
       workspace,
       apps,
       appDataDisks,
@@ -148,8 +149,8 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
       hover: { backgroundColor: colors.accent(0.3) }
     }
 
-    const galaxyApp = currentApp(tools.galaxy.label)
-    const cromwellApp = currentApp(tools.cromwell.label)
+    const galaxyApp = currentApp(tools.Galaxy.label)
+    const cromwellApp = currentApp(tools.Cromwell.label)
 
     // TODO: Try to move app/tool-specific info into tools (in notebook-utils.js) so the function below can just iterate over tools instead of duplicating logic
     const renderToolButtons = () => div({
@@ -171,16 +172,16 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
       }, [img({ src: rstudioBioLogo, alt: 'Create new R markdown file', style: styles.image })]),
       h(Clickable, {
         style: { opacity: galaxyApp ? '0.5' : '1', ...styles.toolCard }, onClick: () => {
-          setCurrentTool(tools.galaxy.label)
-          enterNextViewMode(tools.galaxy.label)
+          setCurrentTool(tools.Galaxy.label)
+          enterNextViewMode(tools.Galaxy.label)
         },
         hover: !galaxyApp ? styles.hover : undefined,
         disabled: !!galaxyApp, tooltip: galaxyApp ? 'You already have a galaxy environment' : ''
       }, [img({ src: galaxyLogo, alt: 'Create new Galaxy app', style: _.merge(styles.image, { width: '30%' }) })]),
-      !tools.cromwell.isAppHidden && h(Clickable, {
+      !tools.Cromwell.isAppHidden && h(Clickable, {
         style: { opacity: cromwellApp ? '0.5' : '1', ...styles.toolCard }, onClick: () => {
-          setCurrentTool(tools.cromwell.label)
-          enterNextViewMode(tools.cromwell.label)
+          setCurrentTool(tools.Cromwell.label)
+          enterNextViewMode(tools.Cromwell.label)
         },
         hover: !cromwellApp ? styles.hover : undefined,
         disabled: !!cromwellApp, tooltip: cromwellApp ? 'You already have a Cromwell instance' : ''
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -784,7 +784,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       () => ({ disabled: !hasChanges() || !!errors, tooltip: Utils.summarizeErrors(errors) })
     )
 
-    const isUpdateDisabled = getIsRuntimeBusy(currentRuntimeDetails)
+    const isUpdateDisabled = getIsRuntimeBusy(currentRuntimeDetails, existingRuntime?.toolDockerImage === desiredRuntime?.toolDockerImage)
 
     const canShowWarning = viewMode === undefined
     const canShowEnvironmentWarning = _.includes(viewMode, [undefined, 'customImageWarning'])
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -659,6 +659,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       const { googleProject } = getWorkspaceObject()
       const currentRuntime = getCurrentRuntime(runtimes)
       const currentPersistentDisk = getCurrentPersistentDisk(runtimes, persistentDisks)
+      console.log('currentPersistentDisk', currentPersistentDisk)
 
       Ajax().Metrics.captureEvent(Events.cloudEnvironmentConfigOpen, {
         existingConfig: !!currentRuntime, ...extractWorkspaceDetails(getWorkspaceObject())
@@ -784,7 +785,9 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       () => ({ disabled: !hasChanges() || !!errors, tooltip: Utils.summarizeErrors(errors) })
     )
 
-    const isUpdateDisabled = getIsRuntimeBusy(currentRuntimeDetails, existingRuntime?.toolDockerImage === desiredRuntime?.toolDockerImage)
+    const isRuntimeError = existingRuntime?.status === 'Error'
+    const shouldErrorDisableUpdate = existingRuntime?.toolDockerImage === desiredRuntime?.toolDockerImage
+    const isUpdateDisabled = getIsRuntimeBusy(currentRuntimeDetails) || (shouldErrorDisableUpdate && isRuntimeError)
 
     const canShowWarning = viewMode === undefined
     const canShowEnvironmentWarning = _.includes(viewMode, [undefined, 'customImageWarning'])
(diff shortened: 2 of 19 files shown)


cahrens:
@jdcanas Can you take a look at the build error?

nawatts:
The unused variable ESLint is complaining about was removed in #2913. Rebasing on `dev` should resolve it.

cahrens:
@jdcanas Please ping me when this PR is ready for review and has a deployment. Note that when you rebase on dev, you will pick up the yarn change. See https://broadinstitute.slack.com/archives/C01EHNUM73R/p1648504581839079 for details.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -354,7 +359,8 @@ export const CloudEnvironmentModal = ({
         div({ style: toolLabelStyles }, [
           img({
             src: getToolIcon(toolLabel),
-            style: { height: 20 }
+            style: { height: 20 },
+            alt: `${toolLabel} image`
```
cahrens: Just put the name of the tool, as now it is reading "Jupyter image, image" (the screen reader adds on image).

It is good to add the alt text here, but it's not actually the images I was talking about. The place that needs to be updated for the buttons on the context bar is `getImgForTool`, and [line 133](https://github.com/DataBiosphere/terra-ui/blob/b3668a8045d41d1f098f60d2954d3f7cfb2790f5/src/components/ContextBar.js#L114) for the general cloud icon.

image

cahrens: @jdcanas what is working the best here is to put an empty alt text attribute on these images. Then the screenreader will use the tooltip text as the name of the label.

If you put the tool's label as the alt text, you don't get to hear to tooltip text and lose the status information.

For example,
`[tools.Jupyter.label, () => img({ alt: '', src: jupyterLogo, style: { height: 45, width: 45 } })],`

src/components/AnalysisModal.js
```diff
@@ -101,13 +101,14 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
...
-      [tools.galaxy.label, () => renderAppModal(GalaxyModalBase, tools.galaxy.label)],
-      [tools.cromwell.label, () => renderAppModal(CromwellModalBase, tools.cromwell.label)]
+      [tools.Galaxy.label, () => renderAppModal(GalaxyModalBase, tools.Galaxy.label)],
+      [tools.Cromwell.label, () => renderAppModal(CromwellModalBase, tools.Cromwell.label)]
     )
 
     const renderComputeModal = () => h(ComputeModalBase, {
       isOpen: currentTool === tools.Jupyter.label || currentTool === tools.RStudio.label,
       isAnalysisMode: true,
+      shouldHideCloseButton: true,
```
cahrens: I don't understand why we would want to hide a close button here. Which close button is this referring to?

src/components/ComputeModal.js
```diff
@@ -802,7 +805,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         ...commonButtonProps,
         onClick: () => {
           applyChanges()
-        }
+        },
+        disabled: isUpdateDisabled,
+        tooltipSide: 'left',
+        tooltip: isUpdateDisabled ? `Cannot perform change on environment in status ( ${currentRuntimeDetails.status} )` : 'Update Environment'
```
nawatts: I'm not sure what the possible values of `currentRuntimeDetails.status` are, but would this read a little more clearly as something like `Cannot update environment in ${currentRuntimeDetails.status} state`?

src/components/AnalysisModal.js
cahrens: Ah, I think I understand what it is now. It's an _inner_ close button. So now, my question is, when would you ever want it to be false?

(image with it set to false)
image

src/components/ContextBar.js
```diff
@@ -65,58 +69,74 @@ export const ContextBar = ({
...
-      div({ style: { display: 'flex', justifyContent: 'center', alignItems: 'center', marginTop: '1rem' } }, [
+    return h(Clickable, {
+      style: { display: 'flex', flexDirection: 'column', justifyContent: 'center', ...contextBarStyles.contextBarButton, borderBottom: '0px' },
+      hover: contextBarStyles.hover,
+      onClick: () => {
+        setSelectedToolIcon(toolLabel)
+        setCloudEnvOpen(true)
+      },
+      tooltipSide: 'left',
+      tooltip: `${toolLabel} Environment ( ${getComputeStatusForDisplay(status)} )`,
```
nawatts, line 80: Is the extra space around the parentheses needed?

```suggestion
      tooltip: `${toolLabel} Environment (${getComputeStatusForDisplay(status)})`,
```

src/components/RuntimeManager.js
```diff
@@ -81,14 +81,14 @@ export const RuntimeErrorModal = ({ runtime, onDismiss }) => {
           .then(res => res.text()))
       setUserscriptError(true)
     } else {
-      setError(runtimeErrors[0].errorMessage)
+      setError(runtimeErrors[0]?.errorMessage)
```
nawatts: Since it looks like `runtimeErrors` may be empty, does this need a default error?

src/libs/runtime-utils.js
```diff
@@ -381,8 +381,8 @@ export const RadioBlock = ({ labelText, children, name, checked, onChange, style
 }
 
 export const getIsAppBusy = app => app?.status !== 'RUNNING' && _.includes('ING', app?.status)
-export const getIsRuntimeBusy = runtime => {
-  const { Creating: creating, Updating: updating, LeoReconfiguring: reconfiguring } = _.countBy(getConvertedRuntimeStatus, [runtime])
-  return creating || updating || reconfiguring
+export const getIsRuntimeBusy = (runtime, isErrorBusy = false) => {
```
nawatts: With `isErrorBusy` set to true, this function's name is no longer descriptive. Especially since it looks like this is only used in one place, I think it would be clearer to use something like `isRuntimeBusy || isRuntimeError` or `isRuntimeInState(["creating", "updating", ...])` instead of adding this flag.

https://alexkondov.com/should-you-pass-boolean-to-functions/

src/components/CromwellModal.js
```diff
@@ -23,21 +23,21 @@ const titleId = 'cromwell-modal-title'
 export const CromwellModalBase = withDisplayName('CromwellModal')(
   ({
     onDismiss, onSuccess, apps, appDataDisks, workspace, workspace: { workspace: { namespace, bucketName, name: workspaceName, googleProject } },
-    isAnalysisMode = false
+    shouldHideCloseButton = false
```
cahrens: There is no reason for this default value to be false-- the CromwellModal is only ever shown in the Analyses tab (AFAIK). 

As we chatted, I think that the default value of `shouldHideCloseButton` should be true, and it should only be specified as an option in the very rare cases where it should be false (cases which are going away soon anyway).

Part of my confusion here is that I expected this "close button" to be the dialog close button, as I'm only used to seeing these modals within the parent dialog (from the new Analyses tab). I am wondering if there is also a clearer name… perhaps `hideInnerCloseButton`? I'm not crazy about that, but with the current name I was quite confused.

src/components/AnalysisModal.js
cahrens: It turns out all this code is related to the old cloud environment widget, which opens these inner "modals" outside of the parent dialog.

src/components/GalaxyModal.js
```diff
@@ -115,7 +114,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
             ['RUNNING', () => h(Fragment, [
               deleteButton,
               pauseButton,
-              h(ButtonPrimary, { disabled: false, onClick: () => setViewMode('launchWarn') }, ['Launch Galaxy'])
+              h(ButtonPrimary, { disabled: false, onClick: () => setViewMode('launchWarn') }, ['Open Galaxy'])
```
cahrens, line 117: Should the default text be updated for consistency?

image

cahrens: This seems to be line 391 in `CloudEnvironmentModal`. There are also some tooltips/aria text things in that file that should be changed from Launch -> Open.

src/components/ComputeModal.js
```diff
@@ -149,7 +149,7 @@ const SparkInterface = ({ sparkInterface, namespace, name, onDismiss }) => {
               style: { marginRight: 'auto' },
               onClick: onDismiss,
               ...Utils.newTabLinkProps
-            }, ['Launch', icon('pop-out', { size: 12, style: { marginLeft: '0.5rem' } })])
+            }, ['Open', icon('pop-out', { size: 12, style: { marginLeft: '0.5rem' } })])
```
cahrens, line 152: Can you add a screenshot of where this is? I can't figure it out.

src/components/ContextBar.js
```diff
@@ -65,58 +69,74 @@ export const ContextBar = ({
...
+          h(Clickable, {
+            style: { flexDirection: 'column', justifyContent: 'center', padding: '.75rem', ...contextBarStyles.contextBarButton, borderBottom: '0px' },
+            hover: contextBarStyles.hover,
+            tooltipSide: 'left',
+            onClick: () => setCloudEnvOpen(!isCloudEnvOpen),
+            tooltip: 'Environment Configuration',
+            tooltipDelay: 100,
+            useTooltipAsLabel: true
+          }, [
+            img({ src: cloudIcon, style: { display: 'flex', margin: 'auto', height: 40, width: 40 } })
```
cahrens: As mentioned, set `href=''` here to pick up tooltip.

src/components/ContextBar.js
```diff
@@ -52,10 +56,10 @@ export const ContextBar = ({
   }
 
   const getImgForTool = toolLabel => Utils.switchCase(toolLabel,
-    [tools.Jupyter.label, () => img({ src: jupyterLogo, style: { height: 30, width: 30 } })],
-    [tools.galaxy.label, () => img({ src: galaxyLogo, style: { height: 12, width: 35 } })],
-    [tools.cromwell.label, () => img({ src: cromwellImg, style: { width: 30 } })],
-    [tools.RStudio.label, () => img({ src: rstudioSquareLogo, style: { height: 30, width: 30 } })]
+    [tools.Jupyter.label, () => img({ src: jupyterLogo, style: { height: 45, width: 45 } })],
```
cahrens: As mentioned, set alt='' here to pick up tooltip.

src/components/GalaxyModal.js
cahrens: Here is some more inconsistency.

image

src/components/ComputeModal.js
kyuksel: ...or `Cannot update environment in ${currentRuntimeDetails.status} status` since `status` is the term we typically use for cloud environments, disks, etc.

jdcanas: It's not only updates that are prohibited though based on state, sometimes it's deletion. I'll move the status after though

src/components/ContextBar.js
jdcanas: That's what the UX team requested

src/components/CromwellModal.js
jdcanas: It'll help a lot to make the changes where it only appears in the legacy cases, and it should eventually be phased out (and not be needed in any new code). That should eliminate most confusion I think :)

src/components/ComputeModal.js
jdcanas: Its for the spark interface associated with dataproc clusters (you need to select something other than the default for a jupyter environment in this drop down, and then click `Manage and monitor spark console`
image

Here's what it looks like:
image

src/components/GalaxyModal.js
jdcanas: Huh... that occurrence just doesn't show up with ctrl+f for me... interesting. Good catch :). I updated the `Cloud Environment Details` text as well to be context-specific.

src/components/RuntimeManager.js
jdcanas: This shouldn't happen in the wild, I was only running into it because I was manually modifying my runtime in the database... It can't hurt though may as well

src/libs/runtime-utils.js
jdcanas: Hmm... that's a good point I hadn't thought of this as a code smell but it's totally true. I refactored a bit. It's a bit of a messy comparison regardless, but I hope its clearer now. 

The original thought was that we have so many places where there's a concept of a busy runtime, and I'd like to start using this function, and build the complexity here with various params, instead of the numerous complicated examples scattered throughout the code. For example, this: https://github.com/DataBiosphere/terra-ui/blob/dev/src/components/RuntimeManager.js#L343. I think it will help once we remove the legacy instances though, since this PR touches all instances of 'busy' that will exist in the new tab.

nawatts: Thanks. I find the use of this function in ComputeModal.js much easier to understand after d10f106f9a95f0fb78894d8c8c3aa08af9540d8c.

src/components/RuntimeManager.js
nawatts: For what it's worth, if it shouldn't happen in the wild, I think the former `setError(runtimeErrors[0].errorMessage)` would be fine. Adding the `?` operator made it seem like an empty array was a case that was expected and needed to be handled.

src/components/ComputeModal.js
```diff
@@ -659,6 +659,7 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
       const { googleProject } = getWorkspaceObject()
       const currentRuntime = getCurrentRuntime(runtimes)
       const currentPersistentDisk = getCurrentPersistentDisk(runtimes, persistentDisks)
+      console.log('currentPersistentDisk', currentPersistentDisk)
```
cahrens: Can you delete this console.log?

src/components/ComputeModal.js
cahrens: Thanks!

src/components/ComputeModal.js
```diff
@@ -802,7 +808,10 @@ export const ComputeModalBase = ({ onDismiss, onSuccess, runtimes, persistentDis
         ...commonButtonProps,
         onClick: () => {
           applyChanges()
-        }
+        },
+        disabled: isUpdateDisabled,
+        tooltipSide: 'left',
+        tooltip: isUpdateDisabled ? `Cannot perform change on environment in ( ${currentRuntimeDetails.status} ) status` : 'Update Environment'
```
cahrens: I'm sure UX didn't want it to look like this. I think the spaces need to go, but you can certainly run this by them!

image

cahrens (approved):
There is a console.log to remove, and I think the "can't update" tooltip needs a change to avoid splitting at an awkward place. But otherwise LGTM.

cahrens:
Also note that the Notebook integration test failed. If there is an open ticket for that test it should be updated with this failure, else a new ticket should be created (assuming that the failure is flaky and not an indication of an actual problem).

src/components/ComputeModal.js
jdcanas: Ah, it looks better in the tooltip from the context bar. I'll remove the spaces here.

Qi77Qi (approved):
:shipit: once tests pass

PR #2915, opened 2022-03-28
[SU-64] Allow typing spaces in attribute editor

nawatts:
This fixes a bug introduced in #2895.

When editing attributes, string values and entity names for references need to be trimmed after editing is finished. Currently, they are trimmed in the input's onChange handler, which makes typing a value that includes a space difficult.

https://user-images.githubusercontent.com/1156625/160475145-dde95f75-102d-45de-a4e9-9368d37ec356.mov


kyuksel (approved):
The code looks great. I found the test you added helpful for documentation. Didn't test.

I assume the screen recording was taken with the new version.

nawatts:
Yes, the screen recording is of the new version.

MatthewBemis (approved):
Looks great!

PR #2916, opened 2022-03-28
[SU-66] Prevent removing last list item in attribute editor

nawatts:
Another bug from #2895.

The attribute editor should prevent removing the last item in a list. However, the remove item button is currently disabled based on the length of the value for that list item, rather than the length of the attribute's `items` list.

## Before
Screen Shot 2022-03-28 at 4 35 58 PM
Screen Shot 2022-03-28 at 4 35 43 PM

## After
Screen Shot 2022-03-28 at 4 35 02 PM
Screen Shot 2022-03-28 at 4 35 09 PM

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -647,7 +647,7 @@ const AttributeInput = ({ value: attributeValue, onChange, entityTypes = [] }) =
           }),
           h(Link, {
             'aria-label': `Remove list value ${i + 1}`,
-            disabled: value.length === 1,
+            disabled: attributeValue.items.length === 1,
             onClick: () => {
               const newAttributeValue = _.update('items', _.pullAt(i), attributeValue)
               onChange(newAttributeValue)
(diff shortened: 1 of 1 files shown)


src/components/data/data-utils.js
```diff
@@ -647,7 +647,7 @@ const AttributeInput = ({ value: attributeValue, onChange, entityTypes = [] }) =
           }),
           h(Link, {
             'aria-label': `Remove list value ${i + 1}`,
-            disabled: value.length === 1,
+            disabled: attributeValue.items.length === 1,
```
kyuksel: It would be nice to use Lodash, in this case for consistency.
```suggestion
            disabled: _.size(attributeValue.items) === 1,
```

kyuksel (approved):
LGTM with a minor comment inline. 
Haven't tested it.

MatthewBemis (approved):

PR #2917, opened 2022-03-29
[IA-3080] Usability testing feedback (top 2 nice-to-have points)

cstrandw:
- Increased the size of the RStudio logo in the `Select an application` modal after the `+Create` button and in the `Cloud Environment Details` modals
- Removed unnecessary arrow from the `Select an application` modal after clicking the `+Create` button
- Added new `r-bio-logo.svg` that has larger text to use instead of the PNG version and deleted the old PNG image. The SVG was optimized before being committed.

Bigger/clearer RStudio Bioconductor logo:
Screen Shot 2022-04-08 at 2 10 34 PM
Screen Shot 2022-04-08 at 2 10 48 PM
Screen Shot 2022-04-08 at 2 13 12 PM

Before `Select an application` modal :
Screen Shot 2022-03-29 at 10 03 02 AM
After:
Screen Shot 2022-04-08 at 2 15 23 PM


cahrens:
@cstrandw please take a look at the test failures and get a deployment for the PR. Thanks!

src/components/AnalysisModal.js
```diff
@@ -316,7 +316,7 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
           resetView()
           onDismiss()
         },
-        onPrevious: () => !!viewMode && resetView()
+        onPrevious: !!viewMode ? () => setViewMode(undefined) : undefined
```
jdcanas: Why do we not call reset view here anymore?

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -255,6 +255,10 @@ export const CloudEnvironmentModal = ({
     [tools.RStudio.label, () => rstudioBioLogo],
     [tools.cromwell.label, () => cromwellImg])
 
+  const getToolIconHeight = toolLabel => Utils.switchCase(toolLabel,
+    [tools.RStudio.label, () => 30],
+    [Utils.DEFAULT, () => 20])
```
cahrens: It would be nice to see the tool icon/tool icon height move into the tools data structure so that each tool is describes its desired behavior vs. having switch statements like `getToolIcon` and `getToolIconHeight`.

cahrens: Actually, I wonder if this would be better handled by changing the RStudio logo itself to have the desired font size. It's sort of strange to change the height of the container to handle variable size of text within the logos.

(Also, now the heights/placement of "None" aren't completely consistent)

cstrandw: Yeah, I agree that changing the logo font itself would be ideal, but how would you recommend going about changing the font size?

cahrens: I would suggest asking UX to help you. They have edited svgs for me in the past.

kyuksel: There's the #dsp-ui-ux channel which would be great to use for this purpose :)

cahrens:
@cstrandw is this ready for review again?

cstrandw:
> @cstrandw is this ready for review again?

Yes, it is. Thanks!

cahrens:
@cstrandw you don't need to check in the PNGs anymore, do you? And can you delete the PNG that is already checked in?

Also, did you optimize the SVG?

cstrandw:
> @cstrandw you don't need to check in the PNGs anymore, do you? And can you delete the PNG that is already checked in?
> 
> Also, did you optimize the SVG?

Deleted the PNGs, sorry I thought I had not pushed them but must have added them to the commit accidentally. And yes I did optimize the SVG, I just updated the PR description to include this.

src/components/AnalysisModal.js
```diff
@@ -315,7 +315,7 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
           resetView()
           onDismiss()
         },
-        onPrevious: () => !!viewMode && resetView()
+        onPrevious: !!viewMode ? () => resetView() : undefined
```
cahrens, line 318: There is no functional difference here, correct?

cstrandw: With the old version, onPrevious was never undefined, it was still a function object even if !!viewMode was false which allowed the arrow in the AnalysisModal TitleBar to show up when it shouldn't. With the second version, this is fixed as onPrevious can be undefined. This logic is consistent with other implementations of the TitleBar in the GalaxyModal as well.

cahrens (approved):

Qi77Qi (approved):

PR #2918, opened 2022-03-29
[SU-25][SU-14] Allow editing attribute for multiple rows at once

nawatts:
Currently, a single attribute can be edited by hovering over a cell in a data table and clicking the edit button.

This allows editing an attribute for multiple rows at once using an interface similar to the existing attribute editor. It also allows adding a new column to the data table (by entering a new attribute name instead of selecting an existing one) and clearing a range of cells (by selecting rows and setting an attribute to an empty string).

https://user-images.githubusercontent.com/1156625/160714104-2fbd839f-32bf-4c44-b8fa-224b3e3169bc.mov

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -1,6 +1,6 @@
 import _ from 'lodash/fp'
 import pluralize from 'pluralize'
-import { Fragment, useState } from 'react'
+import { Fragment, useEffect, useRef, useState } from 'react'
 import { div, fieldset, h, img, label, legend, p, span } from 'react-hyperscript-helpers'
 import {
   ButtonOutline, ButtonPrimary, ButtonSecondary, Clickable, IdContainer, LabeledCheckbox, Link, RadioButton, Select, spinnerOverlay, Switch
@@ -563,7 +563,7 @@ const defaultValueForAttributeType = (attributeType, referenceEntityType) => {
   )
 }
 
-const AttributeInput = ({ value: attributeValue, onChange, entityTypes = [] }) => {
+const AttributeInput = ({ autoFocus = false, value: attributeValue, onChange, entityTypes = [] }) => {
   const { type: attributeType, isList } = getAttributeType(attributeValue)
 
   const renderInput = renderInputForAttributeType(attributeType)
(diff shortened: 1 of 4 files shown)


src/components/data/data-utils.js
```diff
@@ -659,14 +673,15 @@ const AttributeInput = ({ value: attributeValue, onChange, entityTypes = [] }) =
...
+            focusLastListItemInput.current = true
             const newAttributeValue = _.update('items', Utils.append(defaultValue), attributeValue)
             onChange(newAttributeValue)
           }
         }, [icon('plus', { style: { marginRight: '0.5rem' } }), 'Add item'])
       ]) : div({ style: { marginTop: '1.5rem' } }, [
         renderInput({
           'aria-label': 'New value',
-          autoFocus: true,
+          autoFocus,
```
nawatts, line 684: With autoFocus set, the attribute value input would take focus as soon as one letter was typed in the attribute name input.

src/components/data/data-utils.js
```diff
@@ -574,6 +575,18 @@ const AttributeInput = ({ value: attributeValue, onChange, entityTypes = [] }) =
   )
   const defaultValue = defaultValueForAttributeType(attributeType, defaultReferenceEntityType)
 
+  const focusLastListItemInput = useRef(false)
+  const lastListItemInput = useRef(null)
+  useEffect(() => {
```
nawatts, line 580: Moving focus to the last input for a list attribute relied on autoFocus.

src/components/data/data-utils.js
```diff
@@ -763,6 +779,138 @@ export const EntityEditor = ({ entityType, entityName, attributeName, attributeV
   ])
 }
 
+export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames, entityTypes, workspaceId: { namespace, name }, onDismiss, onSuccess }) => {
```
nawatts, line 782: This is very similar to `SingleEntityEditor` (formerly `EntityEditor`). At first, I moved `Modal` out of `EntityEditor` and implemented this component as the attribute input and `EntityEditor`. However, the number of props needed for `EntityEditor` to support both cases started to feel unwieldy, so I copied much of `EntityEditor`'s code into this component.

kyuksel: Yea it's not clear to me how well common code could be factored out in this case either.

kyuksel (approved):
Code and the screencast LGTM. Couldn't test much due to lacking setup.

nawatts:
> Couldn't test much due to lacking setup.

Re-ran the CI workflow to get a PR site in case you want to take a look.

MatthewBemis (approved):
Code looks good. Played around with it a bit too

nawatts:

PR #2920, opened 2022-03-30
[SU-14] Add option to clear a column

nawatts:
Currently, you can delete a data table column from the menu in the column heading. This adds an option to clear the column, setting its value to an empty string for all rows. This can be useful if users want to erase some data (for example, a workflow output), but preserve the column's place in their view settings for when the column is repopulated later.

https://user-images.githubusercontent.com/1156625/160873281-449643fb-f2f6-4071-8f20-a8b9136193a8.mov

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -9,6 +9,7 @@ import RAsyncCreatableSelect from 'react-select/async-creatable'
 import RSwitch from 'react-switch'
 import FooterWrapper from 'src/components/FooterWrapper'
 import { centeredSpinner, containsUnlabelledIcon, icon } from 'src/components/icons'
+import { TextInput } from 'src/components/input'
 import Interactive from 'src/components/Interactive'
 import Modal from 'src/components/Modal'
 import { MiniSortable } from 'src/components/table'
@@ -555,3 +556,43 @@ export const ClipboardButton = ({ text, onClick, children, ...props }) => {
 export const HeaderRenderer = ({ name, label, sort, onSort, style, ...props }) => h(MiniSortable, { sort, field: name, onSort }, [
   div({ style: { fontWeight: 600, ...style }, ...props }, [label || Utils.normalizeLabel(name)])
 ])
+
+export const DeleteConfirmationModal = ({ title, children, confirmationPrompt = 'Delete', buttonText = 'Delete', onConfirm, onDismiss }) => {
+  const [busy, setBusy] = useState(false)
+  const [confirmation, setConfirmation] = useState('')
+
+  const del = async () => {
+    try {
+      setBusy(true)
+      await onConfirm()
+    } finally {
+      onDismiss()
+    }
+  }
+
+  const isConfirmed = _.toLower(confirmation) === _.toLower(confirmationPrompt)
+
+  return h(Modal, {
+    title,
+    onDismiss,
+    okButton: h(ButtonPrimary, {
+      onClick: del,
+      disabled: !isConfirmed,
+      tooltip: isConfirmed ? undefined : 'You must type the confirmation message'
+    }, buttonText)
+  }, [
+    children,
+    div({ style: { display: 'flex', flexDirection: 'column', marginTop: '1rem' } }, [
+      h(IdContainer, [id => h(Fragment, [
+        label({ htmlFor: id, style: { marginBottom: '0.25rem' } }, [`Type "${confirmationPrompt}" to continue:`]),
+        h(TextInput, {
+          id,
+          placeholder: confirmationPrompt,
+          value: confirmation,
+          onChange: setConfirmation
+        })
+      ])])
+    ]),
+    busy && spinnerOverlay
+  ])
+}
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -1,7 +1,7 @@
 import * as clipboard from 'clipboard-polyfill/text'
 import _ from 'lodash/fp'
 import * as qs from 'qs'
-import { Fragment, useState } from 'react'
+import { Fragment, useRef, useState } from 'react'
 import FocusLock from 'react-focus-lock'
 import { b, div, h, h1, img, input, label, span } from 'react-hyperscript-helpers'
 import RSelect, { components as RSelectComponents } from 'react-select'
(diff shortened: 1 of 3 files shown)


cahrens:
Can you make the "Clear Column" and "Delete Column" dialog consistent? Perhaps it is worth sharing the code.

I like the spacing better for "Clear Column". I don't have a preference for whether or not we include "Please", but it seems that it should be consistent.

Also, can you put the keyboard focus in the text field when the dialog is opened?

image
image

src/components/common.js
```diff
@@ -555,3 +556,43 @@ export const ClipboardButton = ({ text, onClick, children, ...props }) => {
 export const HeaderRenderer = ({ name, label, sort, onSort, style, ...props }) => h(MiniSortable, { sort, field: name, onSort }, [
   div({ style: { fontWeight: 600, ...style }, ...props }, [label || Utils.normalizeLabel(name)])
 ])
+
+export const DeleteConfirmationModal = ({ title, children, confirmationPrompt = 'Delete', buttonText = 'Delete', onConfirm, onDismiss }) => {
```
cahrens: Ha, looking at the code now and I see you called this `DeleteConfirmationModal`, but then didn't call it from the delete location.

nawatts: I wanted a reusable component for the pattern currently used for confirming deleting columns and workspaces. I figured I'd keep this PR minimal and change the existing delete modal later. But as you pointed out, since there's a few inconsistencies, better to go ahead and change it here.

src/components/common.js
```diff
@@ -579,7 +581,12 @@ export const DeleteConfirmationModal = ({ title, children, confirmationPrompt =
       onClick: del,
       disabled: !isConfirmed,
       tooltip: isConfirmed ? undefined : 'You must type the confirmation message'
-    }, buttonText)
+    }, buttonText),
+    onAfterOpen: () => {
```
cahrens: It looks like inputs have an `autoFocus property`. Did you try that?

nawatts: I tried `autoFocus` but it did not work. I don't exactly know why, but I suspect there's an issue with focus management when modals are opened from popup menus.

nawatts: More evidence supporting that theory. I can get `autoFocus` to work if I change the column menu to:
```
extraActions: [
  { label: 'Delete Column', onClick: () => setTimeout(() => setDeletingColumn(attributeName), 0) },
  { label: 'Clear Column', onClick: () => setTimeout(() => setClearingColumn(attributeName), 0) }
]
```

src/components/common.js
cahrens: What do you think about calling the dialog "ConfirmationModal"? I'm afraid with the current name people won't realize it can be customized for other purposes.

src/components/common.js
cahrens: ^ Seems to be a better solution. Without the setTimeout, I'm seeing that focus does not return to the column menu after the dialog is dismissed… it seems to go nowhere. To see this, tab to the column menu, trigger the delete or clear choice, tab to Cancel, and dismiss the dialog.

It might be worth adding documentation to ConfirmationModal explaining the need for setTimeout from popup menus for proper focus behavior.

src/components/common.js
nawatts: I'm not opposed to renaming. I thought ConfirmationModal might suggest that it was a simple "Are you sure? Yes or No." without the prompt to type in a confirmation message.

cahrens: PromptedConfirmationDialog?

nawatts: Sounds good to me.

src/components/common.js
nawatts: Nice catch. Do you think we should add a `setTimeout` wrapper around `onClick` in the `MenuButton` component to handle this everywhere? At least in the data tab, there are several popup menus that open modals.

src/components/common.js
```diff
@@ -555,3 +556,44 @@ export const ClipboardButton = ({ text, onClick, children, ...props }) => {
 export const HeaderRenderer = ({ name, label, sort, onSort, style, ...props }) => h(MiniSortable, { sort, field: name, onSort }, [
   div({ style: { fontWeight: 600, ...style }, ...props }, [label || Utils.normalizeLabel(name)])
 ])
+
+export const PromptedConfirmationModal = ({ title, children, confirmationPrompt = 'Delete', buttonText = 'Delete', onConfirm, onDismiss }) => {
+  const [busy, setBusy] = useState(false)
+  const [confirmation, setConfirmation] = useState('')
+
+  const del = async () => {
```
cahrens: Nit: "del" isn't a good name for this anymore. Perhaps `handleClick` or `onClick`?

nawatts: Renamed to "confirm" in [efa7c43](https://github.com/DataBiosphere/terra-ui/pull/2920/commits/efa7c4393e689704d88f2a80283f538288b314d5).

cahrens (approved):

src/components/data/DataTable.js
```diff
@@ -131,21 +132,41 @@ const DataTable = props => {
     setTotalRowCount(unfilteredCount)
   })
 
-  const deleteColumnUpdateEntities = attributeName => {
+  const getAllEntities = async () => {
+    const params = _.pickBy(_.trim, { pageSize: filteredCount, filterTerms: activeTextFilter })
+    const queryResults = await Ajax(signal).Workspaces.workspace(namespace, name).paginatedEntitiesOfType(entityType, params)
```
MatthewBemis, line 137: I realize that this is an API limitation and that we do this same thing in several other places already, but I do want to call out that this `getAllEntities` call adds another path to hit an OOM vulnerability in Rawls which can cause production outages if the user is trying to clear the column for many rows. Analysis Journeys has done some work recently to make sure that this number is extremely high (I think 300k rows is the max per page) so I think this is fine at the moment.

In a similar vein to some of the other changes we've been discussing lately for supporting table-wide operations, I think we should look into modifying [this Rawls](https://rawls.dsde-prod.broadinstitute.org/#/entities/delete_entity_column) endpoint to accept a `retainColumnHeader=true` query param that would do this same work without having to load all of the entities

MatthewBemis (approved):

src/components/data/DataTable.js
nawatts: Yeah, I figured this is another place that will have to be cleaned up eventually, but isn't creating a problem that we don't have already. I like the idea of optimizing this with a query parameter to the delete column endpoint.

MatthewBemis: I will file a ticket for it!

MatthewBemis: Filed as SU-77

nawatts:

PR #2921, opened 2022-03-30
[SUP-345] Allow for more than 1000 files to be viewed in the workspace file viewer

MatthewBemis:
Tested with http://localhost:3000/#workspaces/general-dev-billing-account/10k_files/data , which has 1359 files in it. Let me know if you'd like access to test it (just slack me your dev email)

<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1165,7 +1165,7 @@ const Buckets = signal => ({
    * @returns {Promise<*>}
    * See https://cloud.google.com/storage/docs/json_api/v1/objects/list for additional documentation for underlying GCS API
    */
-  listAll: async (googleProject, bucket, prefix = null, pageToken = null, delimiter = null) => {
+  listAll: async (googleProject, bucket, { prefix = null, pageToken = null, delimiter = null } = {}) => {
     const res = await fetchBuckets(
       `storage/v1/b/${bucket}/o?${qs.stringify({ prefix, delimiter, pageToken })}`,
       _.merge(authOpts(await saToken(googleProject)), { signal })
@@ -1176,7 +1176,7 @@ const Buckets = signal => ({
 
     // Get the next page recursively if there is one
     if (body.nextPageToken) {
-      const next = await Buckets(signal).listAll(googleProject, bucket, prefix, body.nextPageToken, delimiter)
+      const next = await Buckets(signal).listAll(googleProject, bucket, { prefix, pageToken: body.nextPageToken, delimiter })
       return { items: _.concat(items, next.items), prefixes: _.concat(prefixes, next.prefixes) }
     }
     return { items, prefixes }
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1153,9 +1153,9 @@ const Buckets = signal => ({
     return res.json()
   },
 
-  listAll: async (googleProject, bucket, prefix = null, pageToken = null) => {
+  listAll: async (googleProject, bucket, prefix = null, pageToken = null, delimiter = null) => {
     const res = await fetchBuckets(
-      `storage/v1/b/${bucket}/o?${qs.stringify({ prefix, delimiter: '/', pageToken, maxResults: '1' })}`,
+      `storage/v1/b/${bucket}/o?${qs.stringify({ prefix, delimiter, pageToken, maxResults: '1' })}`,
       _.merge(authOpts(await saToken(googleProject)), { signal })
     )
     const body = await res.json()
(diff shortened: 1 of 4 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -216,7 +216,7 @@ const BucketContent = _.flow(
     withErrorReporting('Error loading bucket data'),
     Utils.withBusyState(setLoading)
   )(async (targetPrefix = prefix) => {
-    const { items, prefixes } = await Ajax(signal).Buckets.list(googleProject, bucketName, targetPrefix)
```
MatthewBemis, line 219: This is the key change. Previously, `list` was capped at 1k and did not support pagination, so this switches to use `listAll` which deals with pagination.

src/libs/ajax.js
```diff
@@ -1153,20 +1153,21 @@ const Buckets = signal => ({
...
-  listAll: async (googleProject, bucket, prefix = null, pageToken = null) => {
+  listAll: async (googleProject, bucket, prefix = null, pageToken = null, delimiter = null) => {
     const res = await fetchBuckets(
-      `storage/v1/b/${bucket}/o?${qs.stringify({ prefix, pageToken })}`,
+      `storage/v1/b/${bucket}/o?${qs.stringify({ prefix, delimiter, pageToken })}`,
       _.merge(authOpts(await saToken(googleProject)), { signal })
     )
     const body = await res.json()
     const items = body.items || []
+    const prefixes = body.prefixes || []
```
MatthewBemis, line 1176: A few different parts of Terra UI use `listAll`. None of the existing usages care about the `prefixes` that can be collected while iterating through the pages, but because the workspace file browser displays files in "directory mode" we want to collect the prefixes so we can display them as folders and allow users to click into them

src/pages/workspaces/workspace/Data.js
cahrens: How is the pagination exposed in the UI?

MatthewBemis: It's handled internally by gathering all of the results for the available pages. It's not displayed to the user.

MatthewBemis: For additional context, I think exposing the pagination is the right thing to do long-term. But this fix is motivated by a fairly high-priority blocker for GP so that's why I chose the quicker approach here

src/libs/ajax.js
```diff
@@ -1153,20 +1153,21 @@ const Buckets = signal => ({
     return res.json()
   },
 
-  listAll: async (googleProject, bucket, prefix = null, pageToken = null) => {
+  listAll: async (googleProject, bucket, prefix = null, pageToken = null, delimiter = null) => {
```
cahrens: What is the delimiter for?

cahrens: I know we have very little method documentation, but I personally think it would be nice to document the input arguments and output of the methods (like I did for `getSpendReport`). Given this is not generally our style, it's up to you.

cahrens (approved):
I haven't tested it as I don't have a workspace with files in it. But I'm trusting that Nick will be able to test it.

src/libs/ajax.js
MatthewBemis: Because directories are "fake" in GCS, supplying a delimiter allows us to list objects in a way that's easier to preserve the directory structure that users are expecting. From the [GCS documentation](https://cloud.google.com/storage/docs/json_api/v1/objects/list):

Returns results in a directory-like mode, with / being a common value for the delimiter.

`items[]` contains object metadata for objects whose names do not contain `delimiter`, or whose names only have instances of `delimiter` in their `prefix`.

`prefixes[]` contains truncated object names for objects whose names contain `delimiter` after any `prefix`. Object names are truncated beyond the first applicable instance of the delimiter. If multiple objects have the same truncated name, duplicates are omitted.


I'll add some documentation to this code, and a link to the GCS documentation.

nawatts: With the large number of arguments for this function, it might be nice to move the optional ones into an "options object".

```js
listAll: async (googleProject, bucket, { prefix = null, pageToken = null, delimiter = null } = {}) => {
...
}
```

So that something like `listAll('project', 'bucket', null, null, '/')` becomes `listAll('project', 'bucket', { delimiter: '/' })`.

src/pages/workspaces/workspace/Data.js
nawatts: I'm a bit concerned about what happens when there's a very large number of objects in the bucket. Like if there's 100k workflow folders or something like that. I feel like it may not be long until we get a ticket for the files tab being unresponsive. We also reload the objects list every time a file is uploaded or deleted, so those may become less usable with a large number of files.

But that seems a less severe issue than hiding objects, so I'm on board with starting with the quick fix.

nawatts (approved):
Tested with the (misleadingly named) 10k_files workspace and another workspace with "folders".

MatthewBemis:
@nawatts hah sorry for the misleading workspace name. The script to create 10k files was slow so I killed it after it had enough to at least test the fix

src/libs/ajax.js
MatthewBemis: I like it. Done in https://github.com/DataBiosphere/terra-ui/pull/2921/commits/15d46e356aefc877cf2cd0b1cf12e7e1617d8085

PR #2925, opened 2022-04-01
[SU-65] Improve visual indication of read only and locked workspaces

nawatts:
Currently, when a workspace is read only or locked, that status is displayed in the top bar next to the workspace name. If the workspace name is long enough, it will be hidden. This moves the status indicator to the tab bar next to the workspace menu.

## Before
Screen Shot 2022-04-01 at 9 24 11 AM
Screen Shot 2022-04-01 at 9 24 23 AM

## After
Screen Shot 2022-04-01 at 9 31 26 AM
Screen Shot 2022-04-01 at 9 31 36 AM

### DIFF
--- a/src/pages/workspaces/workspace/WorkspaceContainer.js
+++ b/src/pages/workspaces/workspace/WorkspaceContainer.js
@@ -36,6 +36,33 @@ const navIconProps = {
   hover: { opacity: 1 }, focus: 'hover'
 }
 
+const WorkspacePermissionNotice = ({ workspace }) => {
+  const isReadOnly = !Utils.canWrite(workspace.accessLevel)
+  const isLocked = workspace.workspace.isLocked
+
+  return (isReadOnly || isLocked) && span({
+    style: {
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -36,6 +36,33 @@ const navIconProps = {
...
+      alignItems: 'center',
+      height: '2rem',
+      padding: '0 1rem',
+      borderRadius: '1rem',
+      marginRight: '2rem',
+      backgroundColor: colors.dark(0.15),
+      textTransform: 'none'
+    }
+  }, [
+    isLocked ? icon('lock', { size: 16 }) : icon('eye', { size: 20 }),
```
nawatts, line 55: With these icons the same size, either the eye looked too small or the lock looked too big.

cahrens (approved):
Looks nice, thanks!

MatthewBemis (approved):

PR #2926, opened 2022-04-01
remove hot-refresh, we have react-refresh

zarsky-broad:
<!--
Hello, friend!
Thanks!
--!>


src/index.js
```diff
@@ -1,3 +1,5 @@
+import 'src/outdated-browser-message'
```
nawatts, line 1: My understanding from the UI training was that `outdated-browser-message` was a separate entry point so that it would be loaded even if an older browser failed to parse the main bundle. But since we already split off most of the application into a separate chunk by dynamically importing `appLoader`, this seems like it would be fine as well.

nawatts (approved):
Very nice to have this feature back.

src/index.js
zarsky-broad: That's what I figured, yeah...

ehanna4 (approved):

PR #2928, opened 2022-04-01
[SU-248] Add option to delete empty cells when importing tsv

MatthewBemis:
Permutations tested:

TSV <512k, type doesn't exist yet, no empty cells (don't show new warning, execute sync endpoint)
TSV <512k, type doesn't exist yet, empty cells (don't show new warning, execute sync endpoint)
TSV <512k, type exists, no empty cells (don't show new warning, execute sync endpoint)
TSV <512k, type exists, empty cells (show new warning, execute sync endpoint)
TSV >512k, type doesn't exist yet, no empty cells (don't show new warning, execute async endpoint)
TSV >512k, type doesn't exist yet, empty cells (don't show new warning, execute async endpoint)
TSV >512k, type exists, no empty cells (don't show new warning, execute async endpoint)
TSV >512k, type exists, empty cells (show new warning, execute async endpoint)

Also tested same permutations when checking the 'Create participant, sample, and pair associations' checkbox, which calls the old import endpoint for the FireCloud schema.

Screen Shot 2022-04-01 at 5 10 28 PM

<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -382,7 +382,7 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
         }, [
           icon('warning-standard', { size: 19, style: { color: colors.warning(), flex: 'none', marginRight: '0.5rem', marginLeft: '-0.5rem' } }),
           'We have detected empty cells in your TSV. Please choose an option:',
-          div({ style: { paddingTop: '0.5rem' } }, [
+          div({ style: { paddingTop: '0.5rem' }, role: 'radiogroup', 'aria-label': 'We have detected empty cells in your TSV. Please choose an option.' }, [
             h(RadioButton, {
               text: 'Ignore empty cells (default)',
               name: 'ignore-empty-cells',
@@ -394,7 +394,7 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
           div({ style: { paddingTop: '0.5rem' } }, [
             h(RadioButton, {
               text: 'Overwrite existing cells with empty cells',
-              name: 'overwrite-existing-cells',
+              name: 'ignore-empty-cells',
               checked: deleteEmptyValues,
               onChange: () => setDeleteEmptyValues(true),
               labelStyle: { padding: '0.5rem', fontWeight: 'normal' }
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -274,7 +274,7 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
   const isInvalid = isFileImportCurrMode === isFileImportLastUsedMode && file && !match
   const newEntityType = match?.[1]
   const currentFile = isFileImportCurrMode === isFileImportLastUsedMode ? file : undefined
-  const containsNullValues = fileContents.match('\t\t')
+  const containsNullValues = fileContents.match(/^\t|\t\t+|\t$|\n\n+/gm)
 
   return h(Dropzone, {
     multiple: false,
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -253,7 +253,7 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
       } else {
         const filesize = file?.size || Number.MAX_SAFE_INTEGER
         if (filesize < 524288) { // 512k
-          await workspace.importFlexibleEntitiesFileSynchronous(file)
+          await workspace.importFlexibleEntitiesFileSynchronous(file, deleteEmptyValues)
         } else {
           const { jobId } = await workspace.importFlexibleEntitiesFileAsync(file, deleteEmptyValues)
           asyncImportJobStore.update(Utils.append({ targetWorkspace: { namespace, name }, jobId }))
@@ -284,7 +284,6 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
       setFile(file)
       setFileContents(await Utils.readFileAsText(file.slice(0, 1000)))
       setIsFileImportLastUsedMode(true)
-      setDeleteEmptyValues(false)
     }
   }, [
     ({ dragging, openUploader }) => h(Fragment, [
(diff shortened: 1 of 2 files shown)


src/components/data/data-utils.js
```diff
@@ -375,6 +377,30 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
           `Data with the type '${newEntityType}' already exists in this workspace. `,
           'Uploading more data for the same type may overwrite some entries.'
         ]),
+        currentFile && containsNullValues && _.includes(_.toLower(newEntityType), entityTypes) && div({
+          style: { ...warningBoxStyle, margin: '1rem 0 0.5rem' }
+        }, [
+          icon('warning-standard', { size: 19, style: { color: colors.warning(), flex: 'none', marginRight: '0.5rem', marginLeft: '-0.5rem' } }),
+          'We have detected empty cells in your TSV. Please choose an option:',
+          div({ style: { paddingTop: '0.5rem' } }, [
+            h(RadioButton, {
```
nawatts: Since these two radio buttons are mutually exclusive options, they should have the same `name` and different `value`s. Using the same name for both will make them part of a group and allow using the arrow keys to switch between options when the group is focused and will cause a screen reader to say the proper number of options.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#defining_a_radio_group

src/libs/ajax.js
```diff
@@ -955,23 +955,24 @@ const Workspaces = signal => ({
         return upsertEntities(payload)
       },
 
-      importEntitiesFile: file => {
+      importEntitiesFile: (file, deleteEmptyValues = false) => {
```
nawatts: Nit: Even though there's only one argument, an options object might improve clarity.

src/components/data/data-utils.js
```diff
@@ -273,6 +274,7 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
   const isInvalid = isFileImportCurrMode === isFileImportLastUsedMode && file && !match
   const newEntityType = match?.[1]
   const currentFile = isFileImportCurrMode === isFileImportLastUsedMode ? file : undefined
+  const containsNullValues = fileContents.match('\t\t')
```
nawatts: This doesn't identify null values in the last column of a TSV.

nawatts (changes requested):

src/libs/ajax.js
```diff
@@ -955,23 +955,24 @@ const Workspaces = signal => ({
...
         const formData = new FormData()
         formData.set('entities', file)
-        const res = await fetchOrchestration(`api/${root}/flexibleImportEntities?async=false`, _.merge(authOpts(), { body: formData, signal, method: 'POST' }))
+        const res = await fetchOrchestration(`api/${root}/flexibleImportEntities?${qs.stringify({ deleteEmptyValues, async: false })}`, _.merge(authOpts(), { body: formData, signal, method: 'POST' }))
         return res
       },
 
-      importFlexibleEntitiesFileAsync: async file => {
+      importFlexibleEntitiesFileAsync: async (file, deleteEmptyValues = false) => {
+        console.log(deleteEmptyValues)
```
dmohs: Omit?

src/components/data/data-utils.js
MatthewBemis: 😭 nice catch

src/components/data/data-utils.js
cahrens: I think that you either want to put these in a fieldset, or you need to declare a role of group per https://stackoverflow.com/questions/48265742/how-to-group-form-inputs-accessibly-without-a-fieldset.

cahrens: To clarify, using the same `name` is necessary for the screen reader to properly state the number of options. But you still need to link the legend/title to the choices by either using a fieldset with legend, or by using a group with `aria-labelledby` or `aria-label`.

src/components/data/data-utils.js
dmohs: Perhaps:
```javascript
fileContents.match(/^\t|\t\t+|\t$|\n\n+/gm)
```

dmohs (approved):

src/libs/ajax.js
MatthewBemis: At this stage in my career I think I need to accept that one of these will always make it through, even if I thought I Command-F'd to find them all 😆 

Fixed in b4d782f929f86605a6850b303078da98bf60a59a

src/components/data/data-utils.js
MatthewBemis: Thanks all, this should be resolved in 17827ccdb918bfcda60d80c3d9d885de878ee386

Tested by tabbing through the page, and the up/down arrow keys allowed me to cycle through the radio options

src/components/data/data-utils.js
MatthewBemis: Thank you for saving me the regex-headache. That seemed to do the trick.

Resolved in e4484f2688c78cc7e1477f1717cc54f71027a9fe

src/libs/ajax.js
MatthewBemis: Done 78827900191851b47b77850d9fbd2481e4dafc6f

src/libs/ajax.js
nawatts: Perhaps we should turn on [no-console](https://eslint.org/docs/rules/no-console) in ESLint config?

src/components/data/data-utils.js
```diff
@@ -375,6 +377,32 @@ export const EntityUploader = ({ onSuccess, onDismiss, namespace, name, entityTy
           `Data with the type '${newEntityType}' already exists in this workspace. `,
           'Uploading more data for the same type may overwrite some entries.'
         ]),
+        currentFile && containsNullValues && _.includes(_.toLower(newEntityType), entityTypes) && div({
+          style: { ...warningBoxStyle, margin: '1rem 0 0.5rem' }
+        }, [
+          icon('warning-standard', { size: 19, style: { color: colors.warning(), flex: 'none', marginRight: '0.5rem', marginLeft: '-0.5rem' } }),
+          'We have detected empty cells in your TSV. Please choose an option:',
+          div({ role: 'radiogroup', 'aria-label': 'we have detected empty cells in your tsv. please choose an option.' }, [
```
nawatts, line 385: Since this text is also displayed on the page, `aria-labelledby` could be used here instead of `aria-label` to avoid repeating the label text.

nawatts (approved):

PR #2929, opened 2022-04-04
[WOR-230] Refactor firecloud bucket code.

cahrens:
Some Friday afternoon refactoring while I was looking over our code to understand how much depends on the Google Project ID (in preparation for Azure work).

I manually tested as many of these usages as possible… a couple were difficult to access or don't return anything meaningful in dev.

Rational: all of the other methods in “Buckets” require the Google Project ID (for the service account token), and they will not work with Azure workspaces. However, these firecloudBucketRoot methods do not require the Google Project ID.

I have moved all the firecloudBucketRoot JSON object methods into a new Ajax object, called FirecloudBucket to consolidate them and make it clearer that they do not depend on a Google Project ID.

### DIFF
--- a/src/components/NewWorkspaceModal.js
+++ b/src/components/NewWorkspaceModal.js
@@ -91,7 +91,7 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
       onSuccess(await Utils.cond(
         [cloneWorkspace, async () => {
           const workspace = await Ajax().Workspaces.workspace(cloneWorkspace.workspace.namespace, cloneWorkspace.workspace.name).clone(body)
-          const featuredList = await Ajax().Buckets.getFeaturedWorkspaces()
+          const featuredList = await Ajax().FirecloudBucket.getFeaturedWorkspaces()
           Ajax().Metrics.captureEvent(Events.workspaceClone, {
             public: cloneWorkspace.public,
             featured: _.some({ namespace: cloneWorkspace.workspace.namespace, name: cloneWorkspace.workspace.name }, featuredList),
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1119,21 +1119,6 @@ const Buckets = signal => ({
     )
   },
 
-  getServiceAlerts: async () => {
-    const res = await fetchOk(`${getConfig().firecloudBucketRoot}/alerts.json`, { signal })
-    return res.json()
-  },
-
-  getFeaturedWorkspaces: async () => {
-    const res = await fetchOk(`${getConfig().firecloudBucketRoot}/featured-workspaces.json`, { signal })
-    return res.json()
-  },
-
-  getShowcaseWorkspaces: async () => {
-    const res = await fetchOk(`${getConfig().firecloudBucketRoot}/showcase.json`, { signal })
-    return res.json()
-  },
-
   listNotebooks: async (googleProject, name) => {
     const res = await fetchBuckets(
       `storage/v1/b/${name}/o?prefix=notebooks/`,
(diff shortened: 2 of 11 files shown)


src/pages/library/Code.js
```diff
@@ -96,7 +96,7 @@ const Code = () => {
       withBusyState(setLoading)
     )(async () => {
       const [newFeaturedList, newMethods] = await Promise.all([
-        fetch(`${getConfig().firecloudBucketRoot}/featured-methods.json`, { signal }).then(res => res.json()),
+        Ajax(signal).FirecloudBucket.getFeaturedMethods(),
```
cahrens, line 99: I am passing signal to this one since Methods call right below is already using signal.

src/libs/ajax.js
```diff
@@ -1338,6 +1323,35 @@ const Buckets = signal => ({
 })
 
 
+const FirecloudBucket = signal => ({
+  getServiceAlerts: async () => {
+    const res = await fetchOk(`${getConfig().firecloudBucketRoot}/alerts.json`, { signal })
+    debugger
```
nawatts: Looks like this was left in from development?

cahrens: Oops! Will fix.

src/pages/library/Code.js
marctalbott: What is `signal` for?

marctalbott (approved):
It looks good overall, I just had one question for my own learning

src/pages/library/Code.js
cahrens: It allows the operation to be cancelled if you navigate away from the page. So typically you use it for GET operations, but definitely not for an operation that should complete even if you navigate away (such as a PUT/POST to update server state).

cahrens: Most of these calls weren't using `signal`, and I didn't attempt to decide if they actually should be except for the 2 cases where it was very obvious because there were 2 different pieces of information being obtained from the server, and the other was already using a cancellation signal.

cahrens: Oh, and actually this call was already using `signal`… I meant to make the comment below on the usage in the Workflows file.

aherbst-broad (approved):
Nice refactor!

src/components/NewWorkspaceModal.js
```diff
@@ -91,7 +91,7 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
       onSuccess(await Utils.cond(
         [cloneWorkspace, async () => {
           const workspace = await Ajax().Workspaces.workspace(cloneWorkspace.workspace.namespace, cloneWorkspace.workspace.name).clone(body)
-          const featuredList = await Ajax().Buckets.getFeaturedWorkspaces()
+          const featuredList = await Ajax().FirecloudBucket.getFeaturedWorkspaces()
```
kyuksel, line 94: I wonder if we should deprecate the term `Firecloud` as we refactor things. It's one more thing to find out what it means for newcomers.

kyuksel, line 94: I wonder if we should deprecate the term `Firecloud` as we refactor things. It's one more thing to find out what it means for newcomers.

kyuksel (approved):
Code LGTM with one comment inline.
Didn't test.

src/components/NewWorkspaceModal.js
kyuksel: Not sure why this comment got duplicated.

cahrens: I chose `FirecloudBucket` because the config that is being referenced is `firecloudBucketRoot`. I agree that better names would be good, but for now consistency with the config seemed like my best option.

src/libs/ajax.js
kyuksel: Ah I'd missed it. Thanks for the catch @nawatts!

nawatts: Perhaps we should turn on ESLint's [no-debugger](https://eslint.org/docs/rules/no-debugger) rule?

cahrens: @nawatts Yes, that would be good.

PR #2930, opened 2022-04-04
[SU-75] Limit width of data tab sidebar

nawatts:
This limits the width of the data tab's sidebar based on the browser's window width. This ensures that part of the main content and the sidebar's drag handle are always visible. It also fulfills UX's request for the sidebar width to be limited to 1200px.

## Before
https://user-images.githubusercontent.com/1156625/161556097-e5e66616-e7d0-470e-a781-77c05a7c23da.mov

## After
https://user-images.githubusercontent.com/1156625/161556115-dd087d8c-69ec-408b-8fbb-1e32876d8aaa.mov

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -391,10 +391,8 @@ const DataTypeSection = ({ title, titleExtras, error, retryFunction, children })
 ])
 
 const SidebarSeparator = ({ sidebarWidth, setSidebarWidth }) => {
-  const windowWidth = useRef(window.innerWidth)
-
   const minWidth = 280
-  const getMaxWidth = useCallback(() => _.clamp(minWidth, 1200, windowWidth.current - 200), [])
+  const getMaxWidth = useCallback(() => _.clamp(minWidth, 1200, window.innerWidth - 200), [])
 
   // eslint-disable-next-line react-hooks/exhaustive-deps
   const onDrag = useCallback(_.throttle(100, e => {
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):
:shipit:

nawatts:
Removed the ref that was used to cache `window.innerWidth`. I couldn't tell much of a difference between performance profiles with and without it. In either case, most of the time was used by scripting, not rendering.

src/pages/workspaces/workspace/Data.js
```diff
@@ -391,10 +391,8 @@ const DataTypeSection = ({ title, titleExtras, error, retryFunction, children })
 ])
 
 const SidebarSeparator = ({ sidebarWidth, setSidebarWidth }) => {
-  const windowWidth = useRef(window.innerWidth)
-
   const minWidth = 280
-  const getMaxWidth = useCallback(() => _.clamp(minWidth, 1200, windowWidth.current - 200), [])
+  const getMaxWidth = useCallback(() => _.clamp(minWidth, 1200, window.innerWidth - 200), [])
```
kyuksel, line 395: `useCallback` is said to return a memoized version of the callback that only changes if one of the `inputs` has changed. Since there are no inputs in the above case, won't it always return the same function with `window.innerWidth` set to its initial value at all times?

kyuksel, line 395: `useCallback` is said to return a memoized version of the callback that only changes if one of the `inputs` has changed. Since there are no inputs in the above case, won't it always return the same function with `window.innerWidth` set to its initial value at all times?

nawatts: That would be the case with something like:
```js
const windowWidth = window.innerWidth
const getMaxWidth = useCallback(() => _.clamp(minWidth, 1200, windowWidth - 200), [])
```

As is, the callback function holds a reference to the `window` object from when the callback function was first created. That's ok though, because the `window` object does not change.

kyuksel: Ah interesting.

cahrens (approved):

PR #2932, opened 2022-04-04
[IA-3301] Revert IGV upgrade due to covid genome issues

jordansafer:
Reverting IGV upgrade changes after finding that covid bam/bai files are no longer rendering.

Tested update:
Screen Shot 2022-04-04 at 2 07 59 PM


petesantos (approved):

gpcarr (approved):

kyuksel:
@jordansafer Did [this](https://github.com/DataBiosphere/terra-ui/pull/2908/files) need reverting too?

jordansafer:
> @jordansafer Did [this](https://github.com/DataBiosphere/terra-ui/pull/2908/files) need reverting too?

It shouldn't, the problem we're seeing is with the IGV viewer failing to load Covid .bam files. The IGV viewer was slightly broken before 2908 from a previous change, so it will be a problem if we revert that one.

PR #2933, opened 2022-04-04
[IA-3012][IA-3291] Context bar integration test and small misc UX feedback

jdcanas:
https://broadworkbench.atlassian.net/browse/IA-3012
https://broadworkbench.atlassian.net/browse/IA-3291

The main thing this PR introduces is a test around the Context Bar. The test:
1. creates a jupyter runtime via the context bar
2. ensures it can see the creating jupyter icon/details and the disabled terminal icon
3. ensures the icon changes to running eventually 
4. ensures it can pause the runtime once it is running, and that the terminal is enabled once its running

It also addresses a few one-liner style and wording changes in the second ticket listed


kyuksel:
@jdcanas
Running 

```
HEADLESS=false yarn test --verbose analysis-context-bar
```

locally failed with

```
TimeoutError: waiting for XPath `//*[@role="dialog" and @aria-hidden="false"][contains(normalize-space(.), "Cloud Environment Details") or contains(@aria-label,"Cloud Environment Details") or @aria-labelledby=//*[contains(normalize-space(.),"Cloud Environment Details")]]` failed: timeout 30000ms exceeded
```

for me 1 out of 3 times. What pass rate did you get from running [test-flakes](https://github.com/DataBiosphere/terra-ui/tree/dev/integration-tests#detecting-flakiness) (ideally with `RUNS` set to `100` or more)?

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -384,18 +384,18 @@ const Analyses = _.flow(
           tooltip: !Utils.canWrite(accessLevel) ? noWrite : undefined
         }, [
           icon('plus', { size: 14, style: { color: colors.accent() } }),
-          div({ style: { marginLeft: '0.5rem' } }, ['Create'])
+          div({ style: { marginLeft: '0.5rem' } }, ['Start'])
         ]),
-        div({ style: { flex: 1.5 } }),
+        div({ style: { flex: 1 } }),
```
kyuksel, line 389: I couldn't tell why the div above is necessary?

integration-tests/tests/analysis-context-bar.js
```diff
@@ -0,0 +1,60 @@
...
+)(async ({ workspaceName, page, testUrl, token }) => {
+  // Navigate to appropriate part of UI (the analysis tab)
+  await page.goto(testUrl)
+  await findText(page, 'View Workspaces')
+  await overrideConfig(page, { isAnalysisTabVisible: true })
+  await click(page, clickable({ textContains: 'View Workspaces' }))
+  await signIntoTerra(page, token)
+  await dismissNotifications(page)
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: workspaceName })) })
+  await click(page, navChild('analyses'))
```
kyuksel: Can we factor this out of here and `create-interactive-analysis` test?

kyuksel (commented):
Nice test! I found the comments in the test file helpful.

A couple of comments inline and one comment on the main PR page about having a test failure. 

In general, please report results from `test-flakes` when touching any test.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -140,9 +140,7 @@ const WorkspaceContainer = ({
       setSharingWorkspace, setShowLockWorkspaceModal
     }),
     div({ role: 'main', style: Style.elements.pageContentContainer },
-
-      // TODO: When we switch this over to all tabs, ensure other workspace tabs look the same when inside these divs
-      (isAnalysisTabVisible() && _.includes(activeTab, contextBarTabs) ?
+      (isAnalysisTabVisible() ?
         [div({ style: { flex: 1, display: 'flex' } }, [
           div({ style: { flex: 1, display: 'flex', flexDirection: 'column' } }, [
             children
```
cahrens, line 146: @jdcanas could you remove `setDeletingWorkspace, setCloningWorkspace, setSharingWorkspace, and setShowLockWorkspaceModal` from being passed to `ContextBar` below at line 149? `ContextBar` does not use them… I think it crept in through refactoring of the workspace menu.

src/pages/workspaces/workspace/Analyses.js
jdcanas: This div is responsible for the horizontal spacing between the (Your analyses +start button) to the components (Revert search bar) on the right

image

jdcanas:
> @jdcanas Running
> 
> ```
> HEADLESS=false yarn test --verbose analysis-context-bar
> ```
> 
> locally failed with
> 
> ```
> TimeoutError: waiting for XPath `//*[@role="dialog" and @aria-hidden="false"][contains(normalize-space(.), "Cloud Environment Details") or contains(@aria-label,"Cloud Environment Details") or @aria-labelledby=//*[contains(normalize-space(.),"Cloud Environment Details")]]` failed: timeout 30000ms exceeded
> ```
> 
> for me 1 out of 3 times. What pass rate did you get from running [test-flakes](https://github.com/DataBiosphere/terra-ui/tree/dev/integration-tests#detecting-flakiness) (ideally with `RUNS` set to `100` or more)?

I ran it a bunch of times during during development and it didn't flake once. When I try to run the test-flakes with any concurrency it fails with a bunch of workspace-level errors, so I didn't view it as very reliable.

When I ran it three times at 10 runs each (30 in total), 

- it failed twice the first run of 10, one with the issue you mention and one with an unrelated error in runtime creation (I couldn't get the result though because I forgot to clear the results from the previous concurrent run)
- in the second run of 10, 2 failures because of runtime creation failure
- In the third run of 10, 2 failures because of runtime creation failure.

So, runtime creation stability aside, (which was seeing some issues today), the failure you mention seems pretty rare (1/30). I've been trying to re-create it, but seem unable to reliably.

cahrens:
We expect to get clean runs with test-flakes (100 runs, concurrency set to 10), unless the failure is unrelated to the test being executed (for example, when we had workspace creation/deletion issues). It sounds like more work needs to be done to make the test robust.

I just checked out the branch and ran the test twice locally (not with flakes, with headless set to false so I could see what the test was doing). Both times it failed with this error:

    TimeoutError: waiting for XPath `//*[@data-icon="loadingSpinner"]` failed: timeout 30000ms exceeded
      at noSpinnersAfter (utils/integration-utils.js:140:29)
      at tests/analysis-context-bar.js:28:9

kyuksel:
@jdcanas, seconding @cahrens, please let us know when you get a successful `test-flakes` run (i.e., with `RUNS` set to `100` and not having a `TimeoutError: waiting for XPath…`).

jdcanas:
@cahrens I'm fairly sure that issue is environment-specific, I ran into it when first setting up and never again... I'd expect that to happen for any test, as I have not seen it in hundreds of runs since the first few. Are you sure this isn't affecting all tests for you (i.e. try the create-interactive-analysis test)?

@kyuksel As I've said, test-flakes fails with workspace related errors every run 20/20 with concurrency and is unreliable. However, I did find the issue related to the `waiting for cloud env` you mentioned without encountering it myself by running the test in headless mode and noticing there's a spinner we should be waiting for (i fixed this).

**In conclusion**, I ran it again for 6 runs of 10 (another 60), and saw no issues related to timeouts. So, in my env, we are **82/83 pass rate** (with 7 omitted due to unrelated backend runtime errors), and an attempted mitigation for the 1 failure.  I believe this is a good threshold for stability. Additionally, this test has yet to fail (except a few random 'compilation' errors I pushed while addressing feedback 😅 ) for any of the runs on this PR.

cahrens:
@jdcanas what are the workspace-related errors you see with concurrency? We are able to run all other tests with concurrency, so we definitely want to make sure this isn't indicating a problem.

I'll check out the branch again and try running the test.

cahrens:
@jdcanas I figured out my issue-- I had set the environment variable a couple days ago to run against dev vs. local, and this branch changes the name of the button. So now it passed for me on a single run.

I'll try flakes and see what happens.

integration-tests/src/server.js
```diff
@@ -76,6 +77,7 @@ const registerTestEndpoint = ({ fn, name, timeout = defaultTimeout }) => {
 }
 
 _.forEach(registerTestEndpoint, [
+  testAnalysisContextBar,
   testCreateInteractiveAnalysis,
   testFindWorkflow,
```
cahrens, line 82: Interesting… I haven't registered either of my new tests here, but they run fine in ci. @kyuksel do you know what this file is used for?

src/pages/workspaces/workspace/WorkspaceContainer.js
cahrens: Thanks!

integration-tests/src/server.js
cahrens: I think this is this alternate way of running the tests through Bueller, that nobody actually does…. I wonder if it's time to delete this stuff. I missed the last day of the training, when perhaps Brett talked about this part of the test infrastructure.

(Normally we run the tests through jest)

integration-tests/tests/analysis-context-bar.js
```diff
@@ -0,0 +1,52 @@
...
+  withWorkspace,
+  withBilling,
+  withRegisteredUser
+)(async ({ page, token, testUrl, workspaceName }) => {
+  // Navigate to appropriate part of UI (the analysis tab)
+  await performAnalysisTabSetup(page, token, testUrl, workspaceName)
+  // Create a runtime
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Environment Configuration' })) })
+  await findElement(page, getAnimatedDrawer('Cloud Environment Details'))
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Settings' })) })
```
cahrens, line 20: It probably would be good to make sure that this is the Jupyter one. I guess it's working just because it is the first?

integration-tests/src/server.js
kyuksel: I think so but I'm not sure either. @breilly2 do you know?

As for the possibility of deprecating Bueller, I think it'll be good to discuss that with the QA team since they'll take over the ownership of the testing infrastructure. Until then I'd suggest we keep the list up to date.

cahrens:
@jdcanas I was thinking about the conversation we had on Slack at the end of day yesterday. With a couple of spinner changes, I was able to get this test to run with concurrency and get to the point where it did launch Jupyter, but Jupyter failed to create due to a server-side error. You mentioned that you would like the test to not fail if the server has an issue on nightly runs, but you would like it to fail on PRs so the author can rerun the test to verify that there is no real issue.

Unfortunately this is not functionality that we currently have. Currently, the tests are running against dev on every commit, and they run on every merge to dev against staging/alpha (as well as nightly runs). We do not want any flaky tests, so unless the server side issue is a "once in a blue moon" event, we need to come up with a solution before this test merges. These are the options that come to my mind:

1. If the goal of the test is primarily to ensure that the JavaScript code itself is responding appropriately to user actions and server responses, mock out the Ajax calls related to launching Jupyter and getting its status. The test will no longer be subject to server instability and will run much faster.
2. Change the test so that it handles Juptyer either creating or failing to create, without failing the test.
3. Extend the test infrastructure so that it is possible to mark certain tests as disabled in ci. Have IA team members make sure to manually run the test before submitting/merging PRs. For now, perhaps the test is commented out and IA team members know they should uncomment it while working on their code?
4. Extend the test infrastructure so that it is possible to indicate whether or not test failures should be reported by job type.

integration-tests/src/server.js
cahrens: I am not inclined to try to keep the list up-to-date. If this infrastructure becomes useful at some point, all the missing tests could be added at that time (there are a number of them). I think it is more likely that we will ultimately delete this capability.

kyuksel (commented):
Pending @cahrens's comments to be addressed

integration-tests/src/server.js
kyuksel: I'm fine with that.

jdcanas:
I've only seen the server-side error twice in 300 runs. If it becomes an issue for nightly runs, we could slim down the test or look into disabling it for nightly runs. I would be extremely hesitant to remove it from CI entirely as opposed to making a custom yarn task for nightly runs, which I think is a bit premature. A little worried about making the perfect the enemy of the good here, as this test will be extremely valuable to ensure we don't break the context bar as we move forward.

Additionally, its worth mentioning these issues are only showing up under concurrency, which adds the conflating issue of my CPU being able to handle the runs.

Regardless, it seems infrequent enough that I think it's worth including.

integration-tests/tests/analysis-context-bar.js
```diff
@@ -0,0 +1,54 @@
...
+const testAnalysisContextBarFn = _.flow(
+  withWorkspace,
+  withBilling,
+  withRegisteredUser
+)(async ({ page, token, testUrl, workspaceName }) => {
+  // Navigate to appropriate part of UI (the analysis tab)
+  await performAnalysisTabSetup(page, token, testUrl, workspaceName)
+
+  // Create a runtime
+  await delay(30000)
```
cahrens: If the issue is that the timeout on the next line is sometimes not long enough, please increase the default delay for that call instead of adding a delay that will increase the run time for every test run.

(I realize you are probably already trying this out, but adding a note since @kyuksel was not part of our Slack conversation)

Qi77Qi (approved):
@jdcanas demoed this during sprint review last Fri...seems working as intended

integration-tests/tests/analysis-context-bar.js
```diff
@@ -0,0 +1,56 @@
...
+  withBilling,
+  withRegisteredUser
+)(async ({ page, token, testUrl, workspaceName }) => {
+  // Navigate to appropriate part of UI (the analysis tab)
+  await performAnalysisTabSetup(page, token, testUrl, workspaceName)
+
+  // Create a runtime
+  // This delay is because the button in this tab is enabled right away, but the page isn't fully loaded
+  // It should not be used commonly, and is only necessary here to support the test in a high-concurrency environment
+  await delay(10000)
```
cahrens: I'm not convinced that it is worth having the 10-second delay in. I ran this test with line 19 commented out 220 times, with concurrency set to 10. The first 100 times I commented out lines 24 to the end, the other times I ran the complete test.

I got 3 failures, only one of which was the issue that the delay is meant to prevent (the other 2 was one around the spinner not appearing after going to the analyses tab, and the other was a server-side error related to Jupyter starting, "Error modifying cloud environment").

I would advocate that we don't introduce the 10-second delay to the test until we see actual failures on PR/integration test runs. @jdcanas said he has never seen this problem without concurrency, so it seems to be related to having high load, and it's not clear that on dev we will experience close to the type of load that we get on our laptops while running the test concurrently.

@kyuksel what are your thoughts?

Qi77Qi: >I would advocate that we don't introduce the 10-second delay to the test until we see actual failures on PR/integration test runs

I'd agree with that..introducing sleep is easy enough if we do see flakiness in CIs

kyuksel: I also agree with not including the delay at this time.

Thank you @cahrens and @jdcanas for having been improving the test to reduce the probability of flakiness in the future as flakiness has cost us so much time and pain in the past.

jdcanas: works for me

integration-tests/tests/analysis-context-bar.js
```diff
@@ -0,0 +1,53 @@
+const _ = require('lodash/fp')
```
cahrens, line 2: Would you mind adding a comment at the top indicating team ownership like the ones I'm adding [here](https://github.com/DataBiosphere/terra-ui/pull/2953/files)?

// This test is owned by the Interactive Analysis (IA) Team.

cahrens (approved):

PR #2935, opened 2022-04-05
[SUP-490] Allow billing project users to view billing project details

MatthewBemis:
Users can now click on a billing project and see certain details, including the billing account associated with the project, the workspaces that they have access to within the project, and the owners of the project (per these changes [sam](https://github.com/broadinstitute/sam/pull/594), [rawls](https://github.com/broadinstitute/rawls/pull/1676).)

<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/integration-tests/jest/billing-projects.integration-test.js
+++ b/integration-tests/jest/billing-projects.integration-test.js
@@ -1,5 +1,7 @@
 const { registerTest } = require('./jest-utils')
-const { testBillingSpendReport } = require('../tests/billing-projects')
+const { testBillingMembers, testBillingSpendReport, testBillingWorkspaces } = require('../tests/billing-projects')
 
 
+registerTest(testBillingMembers)
 registerTest(testBillingSpendReport)
+registerTest(testBillingWorkspaces)
--- a/integration-tests/tests/billing-projects.js
+++ b/integration-tests/tests/billing-projects.js
@@ -250,9 +250,14 @@ const testBillingMembersFn = withUserToken(async ({ page, testUrl, token }) => {
   await billingPage.visit()
   await billingPage.selectProject(ownedBillingProjectName)
   await billingPage.selectMembers()
-  // The billing project members tab should be titled "Members"
+
+  // The test user has the Owner role, so the billing project members tab should be titled "Members"
   await billingPage.assertText('Members')
+
+  // The Owner role should see the Add User button
   await billingPage.assertText('Add User')
+
+  // The test user has the Owner role, so they should see all members
   await billingPage.assertText('testuser1@example.com')
   await billingPage.assertText('testuser3@example.com')
 
@@ -260,9 +265,14 @@ const testBillingMembersFn = withUserToken(async ({ page, testUrl, token }) => {
   await billingPage.visit()
   await billingPage.selectProject(notOwnedBillingProjectName)
   await billingPage.selectOwners()
-  // The billing project members tab should be titled "Owners"
+
+  // The test user has the User role, so the billing project members tab should be titled "Owners"
   await billingPage.assertText('Owners')
+
+  // The User role should not see the Add User button
   await assertTextNotFound(billingPage, 'Add User')
+
+  // The test user has the User role, so they should see members with the Owner role, but not with the User role
   await billingPage.assertText('testuser1@example.com')
   await assertTextNotFound(billingPage, 'testuser3@example.com')
 })
--- a/integration-tests/tests/billing-projects.js
+++ b/integration-tests/tests/billing-projects.js
@@ -144,19 +144,26 @@ const setAjaxMockValues = async (testPage, ownedBillingProjectName, notOwnedBill
   }, spendReturnResult, projectListResult, ownedProjectMembersListResult, notOwnedProjectMembersListResult)
 }
 
-const testBillingSpendReportFn = withUserToken(async ({ page, testUrl, token }) => {
-  // Sign in. This portion of the test is not mocked.
-  await page.goto(testUrl)
-  await signIntoTerra(page, token)
-  await dismissNotifications(page)
+const setUpBillingTest = async (page, testUrl, token) => {
+    // Sign in. This portion of the test is not mocked.
+    await page.goto(testUrl)
+    await signIntoTerra(page, token)
+    await dismissNotifications(page)
 
-  // Interact with the Billing Page via mocked AJAX responses.
-  const ownedBillingProjectName = 'OwnedBillingProject'
-  const notOwnedBillingProjectName = 'NotOwnedBillingProject'
-  await setAjaxMockValues(page, ownedBillingProjectName, notOwnedBillingProjectName, '1110')
+    // Interact with the Billing Page via mocked AJAX responses.
+    const ownedBillingProjectName = 'OwnedBillingProject'
+    const notOwnedBillingProjectName = 'NotOwnedBillingProject'
+    await setAjaxMockValues(page, ownedBillingProjectName, notOwnedBillingProjectName, '1110')
+
+    const billingPage = billingProjectsPage(page, testUrl)
+
+    return { ownedBillingProjectName, notOwnedBillingProjectName, billingPage }
+}
+
+const testBillingSpendReportFn = withUserToken(async ({ page, testUrl, token }) => {
+  const { ownedBillingProjectName, notOwnedBillingProjectName, billingPage } = await setUpBillingTest(page, testUrl, token)
 
   // Select spend report and verify cost for default date ranges
-  const billingPage = billingProjectsPage(page, testUrl)
   await billingPage.visit()
   await billingPage.selectProject(ownedBillingProjectName)
   await billingPage.selectSpendReport()
@@ -197,17 +204,7 @@ const testBillingSpendReport = {
 }
 
 const testBillingWorkspacesFn = withUserToken(async ({ page, testUrl, token }) => {
-  // Sign in. This portion of the test is not mocked.
-  await page.goto(testUrl)
-  await signIntoTerra(page, token)
-  await dismissNotifications(page)
-
-  // Interact with the Billing Page via mocked AJAX responses.
-  const ownedBillingProjectName = 'OwnedBillingProject'
-  const notOwnedBillingProjectName = 'NotOwnedBillingProject'
-  await setAjaxMockValues(page, ownedBillingProjectName, notOwnedBillingProjectName, '0')
-
-  const billingPage = billingProjectsPage(page, testUrl)
+  const { ownedBillingProjectName, notOwnedBillingProjectName, billingPage } = await setUpBillingTest(page, testUrl, token)
 
   // Select a billing project that is owned by the user
   await billingPage.visit()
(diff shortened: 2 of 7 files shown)


src/pages/billing/Project.js
```diff
@@ -225,7 +225,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
 
   const signal = useCancellation()
 
-  const adminCanEdit = _.filter(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers).length > 1
+  const projectHasMultipleOwners = _.filter(({ roles }) => _.includes(billingRoles.owner, roles), projectUsers).length > 1
```
MatthewBemis, line 228: The previous name for this was misleading and I think resulted in it being used in some ways that it shouldn't have been. The previous name implied that the user had admin privileges on the project. In reality it just checks that the project won't be orphaned if you remove the only owner.

src/pages/billing/Project.js
```diff
@@ -336,9 +336,11 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
-            member, adminCanEdit,
+            member,
+            adminCanEdit: (projectHasMultipleOwners && isOwner),
             onEdit: () => setEditingUser(member),
-            onDelete: () => setDeletingUser(member)
+            onDelete: () => setDeletingUser(member),
+            isOwner
           })
         }, _.orderBy([sort.field], [sort.direction], projectUsers))
         )
```
MatthewBemis, line 346: @cahrens are there any concerns or considerations that I need to be aware of for the spend reporting tab? aside from the feature flag, is that also limited only to people with the owner role on the project? if so, I can add that logic into this PR

src/components/group-common.js
```diff
@@ -87,7 +87,7 @@ export const MemberCard = memoWithName('MemberCard', ({ member: { email, roles }
   }, [
     div({ role: 'rowheader', style: { flex: '1', whiteSpace: 'nowrap', textOverflow: 'ellipsis', overflow: 'hidden', height: '1rem' } }, [email]),
     div({ role: 'cell', style: { flex: '1', textTransform: 'capitalize', height: '1rem' } }, [_.includes(adminLabel, roles) ? adminLabel : userLabel]),
-    div({ role: 'cell', style: { flex: 'none' } }, [
+    isOwner && div({ role: 'cell', style: { flex: 'none' } }, [
```
nawatts, line 90: ~This menu is now only shown if `isOwner` is true, but was previously enabled when `canEdit` is true. Is there a need for `canEdit` now? Will this menu ever be shown but disabled?~

Looks like this happens when the current user is the only owner of a billing project.

nawatts: Also, since this div is a table cell (`role: 'cell'`), it may be better to conditionally render the menu instead of the entire cell. That way, the number of columns would be consistent between table rows and match the number of columns in the table header. Though I'm not sure if it matters since this is the last column in the row.

nawatts (approved):
I'm not an owner of any billing projects, but as a member, I could see my workspaces and the owners of the billing project.

One minor thing... when expanding a workspace on the billing projects page, there's nothing shown for "Billing Account". I assume that information is not returned unless you're an owner of the billing project, but it would be nice to only render that section if the information is available or to show some placeholder message.

src/pages/billing/Project.js
cahrens: At this point, we need to limit the spend report visibility to owners. Otherwise, they will just get an error. We know that it would be good to open the report to non-owners in the future, but that will require us to change the query.

image

cahrens:
We have no automated test coverage of the Workspaces and Members tabs. Would you be up for adding coverage? AJAX calls can be mocked out as needed/desired-- see https://github.com/DataBiosphere/terra-ui/blob/dev/integration-tests/tests/billing-projects.js, which tests the spend report.

src/pages/billing/Project.js
cahrens: Please extend the [spend report test](https://github.com/DataBiosphere/terra-ui/blob/dev/integration-tests/tests/billing-projects.js) to verify that the spend report does not show for non-owners, even if they are in the alpha spend group.

src/pages/billing/List.js
```diff
@@ -67,8 +67,10 @@ const ProjectListItem = ({ project, project: { roles, status }, isActive }) => {
     ])
   }
 
+  const viewerRoles = _.intersection(roles, _.values(billingRoles))
```
cahrens, line 70: I am seeing some odd behavior when logged in as Hermione (hermione.owner@test…). Why are no members showing for this billing project? 

image

Note also that Hermione is in the alpha spend group, if you want to test that functionality.

cahrens: Also, I would expect to see myself listed in the members. For example, the billing project shown below was owned by Hermione. She made my account a User, so now it shows under "Shared with You". But it doesn't show me as a User in the members list.

image

cahrens: Logged in as Hermione:

image

MatthewBemis: Regarding the empty Members list, I'm going to guess that it's because that project is orphaned and doesn't have any owners. You can't get into that state from Terra UI, but if a user is extremely determined- they can manipulate the policy directly in Sam to get into this state.

Regarding not seeing yourself- that's another Sam quirk. A `User` on a project can't read the `User` policy so that's why it's not returned by the backend. That said, I can insert the entry manually in the UI code to improve the UX of this.

src/pages/billing/Project.js
MatthewBemis: Functionality resolved in 45e4330b, test extension to come soon

MatthewBemis:
@nawatts 
> 
> One minor thing... when expanding a workspace on the billing projects page, there's nothing shown for "Billing Account". I assume that information is not returned unless you're an owner of the billing project, but it would be nice to only render that section if the information is available or to show some placeholder message.

This is actually the case for owners too if they don't have access to the billing account. Personally I don't even understand why the option to expand the workspace is there at all, it doesn't display any information that isn't already on the page. I will file a ticket to re-evaluate this part of the page but it feels out of scope of this PR

src/pages/billing/List.js
cahrens: I think it is worth manually adding the entry. That guarantees you won't get the empty state (which looks broken), and it is less confusing/more consistent with what you see as a billing project owner.

The thing that's interesting is that it means you won't see what other users have access to the workspace, correct? You will only see yourself and any owners? If so, this is definitely worth touching base with product about.

MatthewBemis: Yes, that's correct that you won't see other members with the `User` role on the billing project. fwiw, this design has already been approved by both product and UX

cahrens: OK, from my conversations with Adam it wasn't clear to me that he was aware. I have posted in [Slack](https://broadinstitute.slack.com/archives/C02L9L96YB1/p1649338082330969) about this to make sure we are all on the same page.

nawatts:
> This is actually the case for owners too if they don't have access to the billing account. ...

Sounds good. I thought maybe this was a case where different permissions for owners/non-owners hadn't been accounted for. But if it also happens for owners, agreed that it's out of scope.

MatthewBemis:
Converting this to draft as I may be pushing some commits that aren't quite ready for review. Will undraft once it's ready for re-review.

MatthewBemis:
@cahrens @nawatts this is ready for re-review now. I've added some test coverage for the billing project pages. There is certainly still room to expand on it, but it provides some basic level of testing now.

Additionally, there is one thing that I could use some guidance on. I changed the name of the members tab to either "Members" or "Owners" based on the user's project role, per Product's feedback. However, this creates an issue with the URL- the query param is different for each case so if a User shared a link to the tab with an Owner, their links wouldn't be compatible (`&tab=members` vs `&tab=owners`). I've messed around with it a bunch trying to harmonize the query params, but can't quite get it. Is there a straightforward way to do this? I worry that I've been overcomplicating it so I'm gonna stop pouring time into it and ask for help :)

nawatts:
The ProjectDetail component updates query parameters based on the selected tab's key.
https://github.com/DataBiosphere/terra-ui/blob/1b41c96d8fe56f6baa416ea45f51d9a75ed757cc/src/pages/billing/Project.js#L388-L397

The issue is that the key for that tab changes based on whether or not the current user is an owner of the project.
https://github.com/DataBiosphere/terra-ui/blob/1b41c96d8fe56f6baa416ea45f51d9a75ed757cc/src/pages/billing/Project.js#L327
https://github.com/DataBiosphere/terra-ui/blob/1b41c96d8fe56f6baa416ea45f51d9a75ed757cc/src/pages/billing/Project.js#L230

Using the same key in either case and instead changing the tab's title and tableName should do the trick (currently both are based on `key`).
https://github.com/DataBiosphere/terra-ui/blob/1b41c96d8fe56f6baa416ea45f51d9a75ed757cc/src/pages/billing/Project.js#L381-L386

Alternatively, map a `tab` value of `owners` to `members` or vice versa when setting the query parameters.
https://github.com/DataBiosphere/terra-ui/blob/1b41c96d8fe56f6baa416ea45f51d9a75ed757cc/src/pages/billing/Project.js#L388-L397

nawatts (approved):

cahrens:
I agree that the title should be separated from the key so that the URL routing is consistent.

integration-tests/utils/integration-utils.js
```diff
@@ -89,6 +89,17 @@ const findText = (page, textContains, options) => {
   return page.waitForXPath(`//*[contains(normalize-space(.),"${textContains}")]`, options)
 }
 
+const assertTextNotFound = async (page, text) => {
```
cahrens, line 92: It would be good to allow the specification of the timeout interval so that the test completes faster. Right now it is using the default timeout period, which slows down the tests a lot when looking for text to not be present.

MatthewBemis:
Consistent URL routing handled in 02a20bd7dc2b8c0d6746febeae2d13773c235e03

integration-tests/jest/billing-projects.integration-test.js
```diff
@@ -1,5 +1,7 @@
 const { registerTest } = require('./jest-utils')
-const { testBillingSpendReport } = require('../tests/billing-projects')
+const { testBillingMembers, testBillingSpendReport, testBillingWorkspaces } = require('../tests/billing-projects')
 
 
+registerTest(testBillingMembers)
```
cahrens, line 5: This is the first time we have multiple tests being registered in the same jest wrapper. I think this means that we won't get metrics on the 3 tests individually. I'm not sure it is a big deal though… we can choose to split these up in the future if it ends up causing pain.

integration-tests/tests/billing-projects.js
```diff
@@ -131,17 +174,116 @@ const testBillingSpendReportFn = withUserToken(async ({ page, testUrl, token })
...
+
+  // Interact with the Billing Page via mocked AJAX responses.
+  const ownedBillingProjectName = 'OwnedBillingProject'
+  const notOwnedBillingProjectName = 'NotOwnedBillingProject'
+  await setAjaxMockValues(page, ownedBillingProjectName, notOwnedBillingProjectName, '0')
+
+  const billingPage = billingProjectsPage(page, testUrl)
+
+  // Select a billing project that is owned by the user
+  await billingPage.visit()
```
cahrens, line 210: Nit: to remove the duplication put this code in a block that you loop over for the 2 billing project names.

integration-tests/tests/billing-projects.js
```diff
@@ -131,17 +174,116 @@ const testBillingSpendReportFn = withUserToken(async ({ page, testUrl, token })
...
+})
+
+const testBillingWorkspaces = {
+  name: 'billing-workspaces',
+  fn: testBillingWorkspacesFn
+}
+
+const testBillingMembersFn = withUserToken(async ({ page, testUrl, token }) => {
+  // Sign in. This portion of the test is not mocked.
+  await page.goto(testUrl)
```
cahrens: I would suggest a helper method for lines 238-247 that is shared between the 3 tests.

integration-tests/utils/integration-utils.js
cahrens: Please do look into shortening the timeout period so the test doesn't take as long.

cahrens:
@MatthewBemis The functionality LGTM. I just have a couple requests for the tests.

integration-tests/utils/integration-utils.js
MatthewBemis: Addressed in 996157c6c37c540cd47781a143a90f61cc2992b5

integration-tests/tests/billing-projects.js
MatthewBemis: Test setup extracted in a9c945847d892cfbcebc852875507738dead4f59

cahrens (approved):

PR #2937, opened 2022-04-05
[SU-41] Display notice in Notebooks tab when notebooks are actively being copied over

MatthewBemis:
Screen Shot 2022-04-06 at 6 10 31 PM

Also plays nicely with the page if you happen to create a notebook before the transfer is complete:

Screen Shot 2022-04-06 at 6 06 32 PM

<!--
Hello, friend!
Thanks!
--!>

### DIFF
--- a/src/pages/workspaces/workspace/Analyses.js
+++ b/src/pages/workspaces/workspace/Analyses.js
@@ -344,7 +344,7 @@ const Analyses = _.flow(
       { backgroundColor: colors.warning(0.15), flexDirection: 'none', justifyContent: 'start', alignItems: 'center' })
   }, [
     icon('warning-standard', { size: 19, style: { color: colors.warning(), flex: 'none', marginRight: '1rem' } }),
-    'Copying 1 or more analysis files from another workspace.',
+    'Copying 1 or more interactive analysis files from another workspace.',
     span({ style: { fontWeight: 'bold', marginLeft: '0.5ch' } }, ['This may take a few minutes.'])
   ])
 
--- a/src/pages/workspaces/workspace/Analyses.js
+++ b/src/pages/workspaces/workspace/Analyses.js
@@ -344,7 +344,7 @@ const Analyses = _.flow(
       { backgroundColor: colors.warning(0.15), flexDirection: 'none', justifyContent: 'start', alignItems: 'center' })
   }, [
     icon('warning-standard', { size: 19, style: { color: colors.warning(), flex: 'none', marginRight: '1rem' } }),
-    'Copying 1 or more notebooks from another workspace.',
+    'Copying 1 or more analysis files from another workspace.',
     span({ style: { fontWeight: 'bold', marginLeft: '0.5ch' } }, ['This may take a few minutes.'])
   ])
 
--- a/src/pages/workspaces/workspace/Analyses.js
+++ b/src/pages/workspaces/workspace/Analyses.js
@@ -241,6 +241,7 @@ const Analyses = _.flow(
   const [analyses, setAnalyses] = useState(() => StateHistory.get().analyses || undefined)
   const [currentUserHash, setCurrentUserHash] = useState(undefined)
   const [potentialLockers, setPotentialLockers] = useState(undefined)
+  const [activeFileTransfers, setActiveFileTransfers] = useState(false)
 
   const authState = useStore(authStore)
   const signal = useCancellation()
@@ -273,6 +274,14 @@ const Analyses = _.flow(
     setAnalyses(_.reverse(_.sortBy('lastModified', analyses)))
   })
 
+  const getActiveFileTransfers = _.flow(
+    withErrorReporting('Error loading file transfer status for notebooks in the workspace.'),
+    Utils.withBusyState(setBusy)
+  )(async () => {
+    const fileTransfers = await Ajax(signal).Workspaces.workspace(namespace, wsName).listActiveFileTransfers()
+    setActiveFileTransfers(!_.isEmpty(fileTransfers))
+  })
+
   const uploadFiles = Utils.withBusyState(setBusy, async files => {
     try {
       await Promise.all(_.map(async file => {
@@ -303,6 +312,7 @@ const Analyses = _.flow(
         [notebookLockHash(bucketName, authState.user.email), findPotentialNotebookLockers({ canShare, namespace, wsName, bucketName })])
       setCurrentUserHash(currentUserHash)
       setPotentialLockers(potentialLockers)
+      getActiveFileTransfers()
       refreshAnalyses()
     }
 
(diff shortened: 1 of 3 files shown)


src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -69,14 +69,23 @@ const noNotebooksMessage = div({ style: { fontSize: 20 } }, [
...
   ])
 ])
 
+const activeFileTransferMessage = div({
+  style: _.merge(
+    Style.elements.card.container,
+    { backgroundColor: colors.warning(0.15), flexDirection: 'none', justifyContent: 'start' })
+}, [
+  icon('warning-standard', { size: 19, style: { color: colors.warning(), flex: 'none', marginRight: '0.5rem', marginLeft: '-0.5rem' } }),
+  'Copying 1 or more notebooks from another workspace. ', span({ style: { fontWeight: 'bold' } }, ['This may take a few minutes.'])
```
MatthewBemis: to reviewer: How do I prevent the space after the first sentence from getting clobbered? I tried a bunch of things but clearly I did not try the right thing. Hopefully it's embarrassingly easy to fix

see screenshot in PR description to see the effect of this

nawatts: Wrapping both the "Copying..." sentence and the bold span in another span works. I think the explanation for this is that wrapping them in a span puts both the text element and bold span in an inline formatting context, whereas currently they're in the outer `div`'s flex formatting context. Whitespace is handled differently depending on the formatting context.

* https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#html_largely_ignores_whitespace
* https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flow_Layout/Intro_to_formatting_contexts

Alternatively, you could add a left margin on the bold span to create some space. `ch` units can be useful for cases like this.

* https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#lengths

MatthewBemis: Thank you!

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -253,6 +264,14 @@ const Notebooks = _.flow(
     setNotebooks(_.reverse(_.sortBy('updated', notebooks)))
   })
 
+  const getActiveFileTransfers = _.flow(
+    withErrorReporting('Error loading file transfer status'),
```
nawatts: UX says to end sentences with a period.
https://broadinstitute.slack.com/archives/C34GAKQ5R/p1648570756312039

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -69,14 +69,24 @@ const noNotebooksMessage = div({ style: { fontSize: 20 } }, [
   div([
     'To get started, click ', span({ style: { fontWeight: 600 } }, ['Create a New Notebook'])
   ]),
-  div({ style: { marginTop: '1rem', fontSize: 16 } }, [
+  div({ style: { margin: '1rem 0 1rem 0', fontSize: 16 } }, [
```
nawatts: FYI: since the top and bottom and left and right are the same, this can be written with a shorthand `1rem 0`.
https://developer.mozilla.org/en-US/docs/Web/CSS/margin#syntax

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -69,14 +69,24 @@ const noNotebooksMessage = div({ style: { fontSize: 20 } }, [
...
       ...Utils.newTabLinkProps,
       href: `https://support.terra.bio/hc/en-us/sections/360004143932`
     }, [`What's a notebook?`])
   ])
 ])
 
+const activeFileTransferMessage = div({
+  style: _.merge(
+    Style.elements.card.container,
+    { backgroundColor: colors.warning(0.15), flexDirection: 'none', justifyContent: 'start' })
```
nawatts: Nit: `align-items: center` here would make the icon and text line up a little better.

src/pages/workspaces/workspace/Notebooks.js
```diff
@@ -69,14 +69,24 @@ const noNotebooksMessage = div({ style: { fontSize: 20 } }, [
...
+}, [
+  icon('warning-standard', { size: 19, style: { color: colors.warning(), flex: 'none', marginRight: '1rem' } }),
+  'Copying 1 or more notebooks from another workspace.',
+  span({ style: { fontWeight: 'bold', marginLeft: '1ch' } }, ['This may take a few minutes.'])
```
nawatts: Nit: The `1ch` gap looks noticeably wider than other spaces in the sentences. `0.5ch` might blend in a little better.

nawatts (approved):
A few very minor comments, nothing that would block merging.

src/pages/workspaces/workspace/Notebooks.js
MatthewBemis: Thank you, as you can certainly tell- I am still getting my bearings when it comes to styling. All of these nits are appreciated 😄

src/pages/workspaces/workspace/Notebooks.js
kyuksel: I wonder if we should make the error message more specific like 'Error loading file transfer status for notebooks in the workspace' or such.

MatthewBemis: I think that would be better!

kyuksel:
@MatthewBemis `Notebooks.js` will be replaced by `Analyses.js` once the Analysis Tab feature is released (soon™) so can we apply the changes to `Analyses.js` too? You can enable the Analysis Tab feature by running `configOverridesStore.set({ isAnalysisTabVisible: true }).

MatthewBemis:
@kyuksel I've added it to the Analyses tab in 3d29888cbdedf55b1c87833a807b37e944183437. I violated DRY there but it seemed like some of the new Analyses tab violates DRY already. My assumption is that it's intentionally done this way to make throwing away the old tab much easier, rather than breaking everything out into a common file solely for the sake of migration.

Screenshot:
Screen Shot 2022-04-07 at 1 28 46 PM

kyuksel (approved):
Code changes LGTM. Didn't test.

Qi77Qi (approved):
this is great!

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -328,6 +338,16 @@ const Analyses = _.flow(
     ])
   ])
 
+  const activeFileTransferMessage = div({
+    style: _.merge(
+      Style.elements.card.container,
+      { backgroundColor: colors.warning(0.15), flexDirection: 'none', justifyContent: 'start', alignItems: 'center' })
+  }, [
+    icon('warning-standard', { size: 19, style: { color: colors.warning(), flex: 'none', marginRight: '1rem' } }),
+    'Copying 1 or more notebooks from another workspace.',
```
Qi77Qi: we actually have more artifacts now..it might `.Rmd` file and `.R` files in the future, maybe use some other more generic term?

MatthewBemis: Would `Copying 1 or more analysis files from another workspace.` work? I am not sure what the best generic term is for this but am open to suggestions!

Qi77Qi: ah, I like that 👍

kyuksel: When the context isn't obvious and verbosity isn't a major concern, I prefer to use the term `interactive analysis` since `batch analysis` is also supported by Terra but I don't feel strongly here.

PR #2938, opened 2022-04-06
[IA-3270][IA-3271] Reapply igvjs update and memory leak fix

jordansafer:
After rolling back IA-3270 and IA-3271, we found that these changes were not the source of the problem. This PR is to reapply these changes.

Initial PRs:
https://github.com/DataBiosphere/terra-ui/pull/2909
https://github.com/DataBiosphere/terra-ui/pull/2911

Changes:
- Fix a memory leak in loading the IGV viewer
- Upgrade IGVjs to the latest version


nawatts (approved):

src/components/IGVBrowser.js
```diff
@@ -75,8 +77,9 @@ const IGVBrowser = _.flow(
     }
 
     igvSetup()
-  })
 
+    return () => igvLibrary.current.removeAllBrowsers()
```
nawatts: Since `igvLibrary.current = igv` is set after an async import, it's probably worth checking if `igvLibrary.current` is set before calling `removeAllBrowsers` in case someone quickly closes the IGV browser before igv.js loads.

jordansafer: Ooh, that wouldn't be great. Updated-

kyuksel: I wonder if this would be a good use case for the `useGetter()` hook.

nawatts: `useGetter` stores a state variable in a ref so that a function can capture the ref and, through that ref, have access to the latest value of the state variable. Without the ref, the function would capture the value of the state variable at the time when the function was created. Since `igvLibrary` is a ref, and thus stable across renders, there's no need to wrap it in `useGetter`.

kyuksel: Ah right; that makes sense. Thanks for the clarification!

kyuksel:
@jordansafer did you want to post around for an approval from a relevant team?

jordansafer:
> @jordansafer did you want to post around for an approval from a relevant team?

@kyuksel, I shared in my team channel last week, but it looks like it slipped by :) . I'll find another pair of eyes-

davidangb (approved):
lgtm, though recent happenings have shown I don't know how to test all the use cases for IGV

PR #2939, opened 2022-04-06
[SU-25] Add option to add a row to a data table

nawatts:
Currently, adding rows to a data table requires uploading or copying and pasting a TSV. This adds the ability to add a row and enter attribute values through the UI.

To try this out, enable the feature flag for the data tab redesign (run `configOverridesStore.set({ isDataTabRedesignEnabled: true })` in the console) and select "Add row" from the "Edit" menu.

Screen Shot 2022-04-06 at 2 53 00 PM

### DIFF
--- a/src/components/data/EntitiesContent.js
+++ b/src/components/data/EntitiesContent.js
@@ -384,7 +384,7 @@ const EntitiesContent = ({
           disabled: !entitiesSelected,
           tooltip: !entitiesSelected && 'Select rows to edit in the table',
           onClick: () => setEditingEntities(true)
-        }, 'Edit attribute'),
+        }, ['Edit attribute']),
         h(MenuButton, {
           disabled: !entitiesSelected,
           tooltip: !entitiesSelected && 'Select rows to delete in the table',
@@ -494,9 +494,7 @@ const EntitiesContent = ({
         entityTypes: _.keys(entityMetadata),
         workspaceId: { namespace, name },
         onDismiss: () => setAddingEntity(false),
-        onSuccess: () => {
-          setRefreshKey(_.add(1))
-        }
+        onSuccess: () => setRefreshKey(_.add(1))
       }),
       editingEntities && h(MultipleEntityEditor, {
         entityType: entityKey,
--- a/src/components/data/EntitiesContent.js
+++ b/src/components/data/EntitiesContent.js
@@ -7,7 +7,7 @@ import { Fragment, useRef, useState } from 'react'
 import { div, form, h, input } from 'react-hyperscript-helpers'
 import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
 import { ButtonPrimary, ButtonSecondary, Link } from 'src/components/common'
-import { EntityDeleter, ModalToolButton, MultipleEntityEditor, saveScroll } from 'src/components/data/data-utils'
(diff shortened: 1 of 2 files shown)


src/components/data/EntitiesContent.js
```diff
@@ -377,18 +378,24 @@ const EntitiesContent = ({
...
       content: h(Fragment, [
         h(MenuButton, {
+          onClick: () => setAddingEntity(true)
+        }, 'Add row'),
+        h(MenuButton, {
+          disabled: !entitiesSelected,
+          tooltip: !entitiesSelected && 'Select rows to edit in the table',
           onClick: () => setEditingEntities(true)
-        }, ['Edit Attribute']),
+        }, 'Edit attribute'),
```
kyuksel: The general React best practice we follow for children is to contain them in an array. Since there's a single child and it's a string, it'd work fine here without containment but it'd be good to be consistent.

```suggestion
        }, ['Edit attribute']),
```

src/components/data/EntitiesContent.js
```diff
@@ -481,6 +488,16 @@ const EntitiesContent = ({
...
       }),
+      addingEntity && h(AddEntityModal, {
+        entityType: entityKey,
+        attributeNames: entityMetadata[entityKey].attributeNames,
+        entityTypes: _.keys(entityMetadata),
+        workspaceId: { namespace, name },
+        onDismiss: () => setAddingEntity(false),
+        onSuccess: () => {
+          setRefreshKey(_.add(1))
+        }
```
kyuksel: Nit: We could save some vertical space:

```suggestion
        onSuccess: () => setRefreshKey(_.add(1))
```

src/components/data/data-utils.js
```diff
@@ -939,6 +940,111 @@ export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames,
...
 
+export const AddEntityModal = ({ workspaceId: { namespace, name }, entityType, attributeNames, entityTypes, onDismiss, onSuccess }) => {
+  const [entityName, setEntityName] = useState('')
+  const [entityNameInputTouched, setEntityNameInputTouched] = useState(false)
+  const [takenNames, setTakenNames] = useState([])
+
+  // Default all attribute values to empty strings
+  const [attributeValues, setAttributeValues] = useState(_.fromPairs(_.map(attributeName => [attributeName, ''], attributeNames)))
+
+  const [isBusy, setIsBusy] = useState(false)
```
kyuksel: Nit: From other components, my eyes are used to easily locating state thanks to having chunks like:

```suggestion
  // Default all attribute values to empty strings
  const [attributeValues, setAttributeValues] = useState(_.fromPairs(_.map(attributeName => [attributeName, ''], attributeNames)))
  const [isBusy, setIsBusy] = useState(false)
```

src/components/data/data-utils.js
```diff
@@ -939,6 +940,111 @@ export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames,
...
+  }
+
+  return h(Modal, {
+    onDismiss,
+    title: `Add ${entityType}`,
+    okButton: h(ButtonPrimary, {
+      disabled: !!entityNameErrors,
+      tooltip: Utils.summarizeErrors(entityNameErrors),
+      onClick: createEntity
+    }, 'Add')
```
kyuksel:
```suggestion
    }, ['Add'])
```

See comment above for a similar case.

kyuksel (approved):
Code LGTM with a few minor comments inline.
Didn't get to test.

MatthewBemis (approved):

PR #2940, opened 2022-04-06
[SU-83] Configure ESLint to warn about debugger and console logs

nawatts:
`debugger` statements and console logs used in development sometime make their way into terra-ui PRs. Console logs just add unnecessary noise, but `debugger` statements could be very disruptive if merged/deployed. Currently, we rely on PR reviews to remove these but we can automate that with ESLint.


.eslintrc.js
```diff
@@ -101,6 +101,16 @@ module.exports = {
...
     'import/newline-after-import': ['warn', { 'count': 2 }],
-    'import/no-anonymous-default-export': ['warn', { 'allowObject': true }]
-  }
+    'import/no-anonymous-default-export': ['warn', { 'allowObject': true }],
+
+    'no-debugger': 'warn'
+  },
+  'overrides': [
+    {
+      'files': 'src/**/*.js',
```
nawatts, line 110: Only enabling `no-console` for the contents of the `src` directory since there is more likely to be valid uses for console logs in integration tests, configuration, scripts, etc.

.eslintrc.js
```diff
@@ -101,6 +101,16 @@ module.exports = {
...
+      'rules': {
+        'no-console': ['warn', { 'allow': ['assert', 'error'] }]
```
nawatts, line 112: There were enough `console.assert` and `console.error`s that I left them enabled instead of going through and adding `eslint-disable-line` comments to each one.

cahrens: I think console.assert is fine in code.

src/pages/billing/NewAccountModal.js
```diff
@@ -330,6 +330,7 @@ const NewAccountModal = ({ onDismiss }) => {
         h(ButtonSecondary, { onClick: () => setPage(page - 1), disabled: onFirstPage }, ['Back']),
         h(ButtonPrimary, {
           style: { marginLeft: '2rem' },
+          // eslint-disable-next-line no-console
```
cahrens: Congrats, you found dead code! Please just delete this file.

kyuksel (approved):
Thanks for this!

src/libs/utils.js
```diff
@@ -89,7 +89,7 @@ export const isOwner = hasAccessLevel('OWNER')
 export const workflowStatuses = ['Queued', 'Launching', 'Submitted', 'Running', 'Aborting', 'Succeeded', 'Failed', 'Aborted']
 
 export const log = (...args) => {
-  console.log.apply(null, args)
+  console.log.apply(null, args) // eslint-disable-line no-console
```
cahrens, line 98: I believe this is unused code-- please delete it.

src/pages/billing/NewAccountModal.js
nawatts: I was wondering about that "TODO: add submit action here", but I didn't look too closely. Deleted in https://github.com/DataBiosphere/terra-ui/pull/2940/commits/ef5c5859ce1280f870cb5892831fd32e1fb19cb4.

src/libs/utils.js
nawatts: I think this function is intended to help with debugging `_.flow` pipelines. I'm not opposed to deleting though, since it is easy to rewrite and if its existence isn't widely known, it's likely to get rewritten anyway.

cahrens: How?

cahrens: If it's useful, please add documentation.

nawatts: To log an intermediate value in a pipeline, it's a shortcut to write:

```js
_.flow(
  step1,
  Utils.log,
  step2
)
```

instead of

```js
_.flow(
  step1,
  value => {
    console.log(value)
    return value
  },
  step2
)
```

Will add docs.

nawatts: Then again, I guess that's a route for console logs to go undetected by ESLint.

nawatts: Deleted in 794d78c66d224af885a6a2b51bb231af4c60d713.

cahrens (approved):

src/libs/utils.js
kyuksel: It provides a neat way to wrap a function to log something, like so:

image

It's there for debug purposes so the fact that there is no usage of it in the codebase is not an indicator of lack of usefulness. 

I'd prefer to keep it. Documenting sounds helpful too.

nawatts: That seems like more of a quirk of the implementation than intended usage.

If we want something that wraps a function with a console log, I'd suggest we add something like:
```js
const withLog = (message, fn) => ...
```
that wouldn't log the function itself.

nawatts: But that would still provide a route for console logs to go undetected by ESLint, when it seems like we always want to remove console logs before merging to dev.

kyuksel: > That seems like more of a quirk of the implementation than intended usage.

Yea your `_.flow` example was a better one. 

> But that would still provide a route for console logs to go undetected by ESLint, when it seems like we always want to remove console logs before merging to dev.

I think the benefit outweighs the risk of that happening but I'm okay with letting it go.

nawatts:
Since there's disagreement on what to do with `Utils.log`, I'll err on the side of no change. Restored the function and added some documentation about its purpose.

MatthewBemis (approved):

PR #2941, opened 2022-04-07
[SU-52] Improve support for long column names in column settings modal

nawatts:
Currently, the data table's column settings modal does not handle long attribute names very well. In the list of columns on the left, the column names are truncated (with no indication that they are) and there's no way to view the full column name.

This makes a few improvements:
- Use text-overflow: ellipsis to indicate when column names are truncated
- Add tooltips to view the full column name
- Widen the column settings modal
- Reduce the width used for the saved column settings

## Before
Screen Shot 2022-04-07 at 2 00 52 PM

## After
Screen Shot 2022-04-07 at 2 01 33 PM

### DIFF
--- a/src/components/data/SavedColumnSettings.js
+++ b/src/components/data/SavedColumnSettings.js
@@ -267,13 +267,13 @@ const SavedColumnSettings = ({ workspace, snapshotName, entityType, entityMetada
 
 export const ColumnSettingsWithSavedColumnSettings = ({ columnSettings, onChange, ...otherProps }) => {
   return div({ style: { display: 'flex', justifyContent: 'space-between' } }, [
-    div({ style: { width: 'calc(50% - 1rem)' } }, [
+    div({ style: { flex: '1 1 0' } }, [
       h(ColumnSettings, {
         columnSettings,
         onChange
       })
     ]),
-    div({ style: { width: 'calc(50% - 1rem)', marginTop: '2rem' } }, [
+    div({ style: { flex: '0 0 265px', marginLeft: '1rem', marginTop: '2rem' } }, [
       h(SavedColumnSettings, {
         ...otherProps,
         columnSettings,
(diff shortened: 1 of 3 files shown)


src/components/data/SavedColumnSettings.js
```diff
@@ -222,9 +222,8 @@ const SavedColumnSettings = ({ workspace, snapshotName, entityType, entityMetada
       }, 'Save this column selection')
     ]),
 
-    hr({ style: { margin: '1rem 0' } }),
-
     _.size(savedColumnSettings) > 0 && h(Fragment, [
+      hr({ style: { margin: '1rem 0' } }),
```
nawatts, line 226: This is unrelated, but there's no need to show this divider if there is nothing below it.

src/components/table.js
```diff
@@ -746,16 +741,30 @@ export const ColumnSettings = ({ columnSettings, onChange }) => {
               h(SortableHandleDiv, { style: styles.columnHandle }, [
                 icon('columnGrabber', { style: { transform: 'rotate(90deg)' } })
               ]),
-              div({ style: { display: 'flex', alignItems: 'center' } }, [
-                h(LabeledCheckbox, {
-                  checked: visible,
-                  onChange: () => toggleVisibility(index)
+              h(IdContainer, [id => h(Fragment, [
+                h(TooltipTrigger, {
+                  // Since entity names don't contain spaces, word-break: break-all is necessary to
```
MatthewBemis: nit: should this comment read "Since attribute names..." and "...wrap the attribute name..." instead?

MatthewBemis (approved):

src/components/table.js
nawatts: Yes. Fixed in https://github.com/DataBiosphere/terra-ui/pull/2941/commits/a6512a11293a66b9fe936fb9ce055888291ccc61.

petesantos (commented):
This looks good, I do want to spend some time testing it out some more since this is a common component, I will test some more and approve tomorrow.

petesantos (approved):

PR #2942, opened 2022-04-07
[DC-275] Move data browser to call catalog rather than data repo

s-rubenstein:
Tested by opening up the list page, clicking a dataset, exporting the dataset to a new workspace, and going to the preview page.

There is some functionality that is likely broken for some, but that is expected as the data catalog doesn't yet have full feature parity with the TDR. Given that it is a beta feature, we have deemed that acceptable risk.


src/components/workspace-utils.js
```diff
@@ -123,6 +123,7 @@ export const SnapshotInfo = ({
   // Lifecycle
   useOnMount(() => {
     const loadSnapshotInfo = async () => {
+      // This is the only .DataRepo.snapshot call. Should it use the catalog?
```
s-rubenstein: Open question! Thread here:

src/libs/ajax.js
```diff
@@ -1041,38 +1042,15 @@ const Workspaces = signal => ({
...
 
+const Catalog = signal => ({
+  getDatasets: async () => {
+    const res = await fetchCatalog('v1/datasets', _.merge(authOpts(), { signal }))
+    return res.json()
+  }
+})
 
-const DataRepo = signal => ({
-  getSnapshots: async () => {
```
s-rubenstein, line 1046: This was now unused, so I just pulled it. I am unsure the value of the hard coded extra fields though.

src/pages/library/DataBrowser.js
```diff
@@ -113,6 +113,8 @@ const SelectedItemsDisplay = ({ selectedData, setSelectedData }) => {
         style: { textTransform: 'none', fontSize: 14 },
         onClick: () => {
           Ajax().Metrics.captureEvent(`${Events.catalogWorkspaceLink}:tableView`, {
+            // These are still using snapshot as a relic to ensure backwards search
```
s-rubenstein, line 116: I didn't want to change the name of the fields for these, because it would ruin searching and metrics in mixpanel, but added a comment just so it is clear.

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -103,6 +103,8 @@ const RequestDatasetAccessButton = ({ title, id, setShowWipModal }) => {
   return h(ButtonPrimary, {
     disabled: status,
     onClick: withErrorReporting('Error requesting dataset access', async () => {
+      // Should this point to catalog? My understanding is that we are fully public?
```
s-rubenstein: This is the only thing I am really not sure on the behavior of. The modal is only used in the Data Browser, so from what I can tell it should probably point to catalog, but I wasn't sure because I feel like someone said all catalog datasets were public for now.

s-rubenstein (commented):
There was a big rename from snapshot to dataset as part of this, to help reduce confusion. From my understanding, snapshots are a subset of datasets which can show up here, so the variable should be the broadest and most accurate for clarity.

src/components/workspace-utils.js
nawatts: Was there supposed to be a link in this comment?

src/libs/ajax.js
nawatts: I'm also unsure, but from #2714 and #2868, they seem like hard-coded development/demo data. #2714 says the attributes added there would be reverted "once the beta test cycle is complete". The attributes added in #2868 look like they were just for the purposes of development/review and possibly should never have been committed at all.

nawatts (approved):

nawatts:
Sounds like there's a few unresolved product question, but the code changes thus far look like mostly straightforward renames and swapping out endpoints.

cahrens:
Please make sure to look into the test failures.

src/components/workspace-utils.js
s-rubenstein: No, it is referring to a method in the ajax file. (Would have been more clear to say `Ajax().DataRepo.snapshot.details` call, I can change it)

nawatts: Sorry, "comment" was ambiguous. I was referring to the [GitHub comment](https://github.com/DataBiosphere/terra-ui/pull/2942#discussion_r845479137), not the code comment.

 > Open question! Thread here:

s-rubenstein:
I can't rerun, but this test is marked flaky, and didn't fail before. (I just made specific test changes, so would be surprised if it wasn't just flakiness causing the failure)

cahrens:
@s-rubenstein you should be able to re-run tests in ci. Perhaps your permissions aren't correct?

image

None of our tests are very flaky at the moment, and we want to keep them that way! Please make sure to run any tests that you touch locally with test-flakes so that you can identify/fix any flakiness.

s-rubenstein:
Yeah, I discovered I wasn't logged in!

s-rubenstein:
@cahrens some of them are tagged in circle-ci as flaky, and rerunning seems to produce different failures? Maybe there's something going on if that's not expected?
Screen Shot 2022-04-08 at 2 26 52 PM

I'd also add that re-running the tests caused different test failures tagged as flaky. This was against the same commit. I definitely didn't touch any code relating to drs uris, as well. Catalog could be new, but it isn't failing consistently, and is passing locally.
Screen Shot 2022-04-08 at 2 28 18 PM
Screen Shot 2022-04-08 at 2 31 43 PM

If there's something special going on here that I don't know about or some way to isolate these down, let me know. I am hesitant to do this as part of this PR to avoid scope creep, if these are existing flakiness, but if we think it is new introduced by this, I can't figure out where that would come from. I am happy to pair. (Maybe @petesantos would know something here I don't? Since he's been in this code much more, and is a reviewer from the data-explorer team and in the UI reviewers group)

petesantos:
On the test flakiness, it looks like the `preview-drs-uri` test and `run-catalog` test are flakes. I have seen these tests sporadically fail in the past.

On a related note, there is an issue in Circle where the stack trace is not logged and there are no artifacts. I have been meaning to investigate this, but have not had the time. I think most of our remaining flakes are having this issue in Circle.

@s-rubenstein would you mind putting a ticket on our board to investigate this issue in circle?

s-rubenstein:
@petesantos done https://broadworkbench.atlassian.net/browse/DC-306

fboulnois (approved):
overall lgtm 👍

src/components/workspace-utils.js
s-rubenstein: Ah, no link intended! The `thread here:` is referring to the comment. I am just looking to open up discussion on it to see what people think.

petesantos: Given that this component is being used in workspaces, I would not change this to use the catalog.

petesantos (approved):

src/pages/library/RequestDatasetAccessModal.js
lu-c: yes, this should point towards the catalog! Right now, it shoudl also be pointing *nowhere* haha.

lu-c (approved):

PR #1544, opened 2019-05-02
 Add select all / select none on the IGV file selector, enforce max number of selected files

wnojopra:
Screenshot from 2019-05-02 13-13-13

PR #2943, opened 2022-04-07
[SU-76] Remove limit on the number of files that can be opened with IGV

nawatts:
Currently, we only allow opening up to 10 files with IGV. This removes that limit.

#1544 didn't provide any information about why this limit was added, but users have requested the ability to view more than 10.

### DIFF
--- a/src/components/IGVFileSelector.js
+++ b/src/components/IGVFileSelector.js
@@ -16,8 +16,6 @@ const getStrings = v => {
   )
 }
 
-const MAX_CONCURRENT_IGV_FILES = 10
-
 const IGVFileSelector = ({ selectedEntities, onSuccess }) => {
   const [refGenome, setRefGenome] = useState('hg38')
   const [selections, setSelections] = useState(() => {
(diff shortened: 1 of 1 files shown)


kyuksel (approved):
Code LGTM. Didn't test.

MatthewBemis (approved):

cahrens:
@nawatts FYI, I was curious so I asked Willy about why the limit of 10 was introduced. You can read his reply [here](https://github.com/DataBiosphere/terra-ui/pull/1544#discussion_r846192179).

nawatts:
Thanks for checking into that. For what it's worth, I tested this with 50 (small) BAM files and, while it took a while to load them all, it seemed to work fine. I don't know what all optimization IGV.js does to reduce the amount of data loaded. For indexed files, it at least makes requests with Range headers to only get data in the visible region. I'd guess that it would likely be difficult or impossible to come up with a limit on number of files that made sense for all file types, references, region sizes, etc. If it becomes an issue, maybe we could add a softer limit like a warning or a prompt for confirmation.

PR #2944, opened 2022-04-07
[SU-56] Add confirmation prompt for deleting analyses/notebooks

nawatts:
This adds a prompt to type a confirmation before deleting an analysis/notebook. Eventually, all delete confirmations will be moved to this pattern.

## Before
Screen Shot 2022-04-07 at 5 38 49 PM

## After
Screen Shot 2022-04-07 at 5 37 50 PM

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -557,7 +557,20 @@ export const HeaderRenderer = ({ name, label, sort, onSort, style, ...props }) =
   div({ style: { fontWeight: 600, ...style }, ...props }, [label || Utils.normalizeLabel(name)])
 ])
 
-export const PromptedConfirmationModal = ({ title, children, confirmationPrompt = 'Confirm', buttonText = 'Confirm', onConfirm, onDismiss }) => {
+export const DeleteConfirmationModal = ({
+  objectType,
+  objectName,
+  title: titleProp,
+  children,
+  confirmationPrompt: confirmationPromptProp,
+  buttonText: buttonTextProp,
+  onConfirm,
+  onDismiss
+}) => {
+  const title = titleProp || `Delete ${objectType}`
+  const confirmationPrompt = confirmationPromptProp || `Delete ${objectType}`
+  const buttonText = buttonTextProp || `Delete ${objectType}`
+
   const [busy, setBusy] = useState(false)
   const [confirmation, setConfirmation] = useState('')
 
@@ -581,7 +594,11 @@ export const PromptedConfirmationModal = ({ title, children, confirmationPrompt 
       tooltip: isConfirmed ? undefined : 'You must type the confirmation message'
     }, buttonText)
   }, [
-    children,
+    children || h(Fragment, [
+      div([`Are you sure you want to delete the ${objectType} `,
+        span({ style: { fontWeight: 600, wordBreak: 'break-word' } }, [objectName]), '?']),
+      div({ style: { fontWeight: 500, marginTop: '1rem' } }, 'This cannot be undone.')
+    ]),
     div({ style: { display: 'flex', flexDirection: 'column', marginTop: '1rem' } }, [
       h(IdContainer, [id => h(Fragment, [
         label({ htmlFor: id, style: { marginBottom: '0.25rem' } }, [`Type "${confirmationPrompt}" to continue:`]),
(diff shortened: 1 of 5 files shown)


src/pages/workspaces/workspace/Analyses.js
```diff
@@ -477,14 +477,23 @@ const Analyses = _.flow(
...
-        deletingAnalysisName && h(AnalysisDeleter, {
-          printName: getDisplayName(deletingAnalysisName), googleProject, bucketName,
-          toolLabel: getTool(deletingAnalysisName),
-          onDismiss: () => setDeletingAnalysisName(undefined),
-          onSuccess: () => {
-            setDeletingAnalysisName(undefined)
-            refreshAnalyses()
-          }
+        deletingAnalysisName && h(DeleteConfirmationModal, {
+          objectType: 'analysis',
```
jdcanas: Hmm.. not critical but would it be possible to use the `getTool` intelligently here so it doesn't say 'delete analysis'? 

Something like `Delete RStudio analysis` or `Delete Jupyter analysis`

jdcanas (approved):

src/components/common.js
```diff
@@ -597,3 +597,18 @@ export const PromptedConfirmationModal = ({ title, children, confirmationPrompt
     busy && spinnerOverlay
   ])
 }
+
+export const DeleteConfirmationModal = ({ objectName, objectType, children, ...props }) => {
```
cahrens: @nawatts can you change the DataTable `PromptedConfirmationModal` usage to call this new method?

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -477,14 +477,25 @@ const Analyses = _.flow(
...
+          objectName: getDisplayName(deletingAnalysisName),
+          onConfirm: async () => {
+            try {
+              await Ajax().Buckets.analysis(
+                googleProject,
+                bucketName,
+                getDisplayName(deletingAnalysisName),
+                getTool(deletingAnalysisName)
+              ).delete()
+              refreshAnalyses()
```
cahrens, line 493: There is no need to call `setDeletingAnalysisName(undefined)` in the `onConfirm` case? How does the deletingAnalysisName get cleared?

nawatts: PromptedConfirmationModal calls `onDismiss` after `onConfirm`.

https://github.com/DataBiosphere/terra-ui/blob/f1ad9fbaafb4c8f6513bb5fe1485895679ab868f/src/components/common.js#L564-L571

src/components/common.js
nawatts: Will do.

cahrens (approved):

nawatts:
Sorry about the post-approval changes...
- https://github.com/DataBiosphere/terra-ui/pull/2944/commits/17117c14737af8f9eb8ec6a453e1b05db04c8c52 uses the DeleteConfirmationModal component for confirming clearing a data table column. This has no visible changes now, but eventually DeleteConfirmationModal will get a yellow / warning color background.
- https://github.com/DataBiosphere/terra-ui/pull/2944/commits/e5a410b3226a36f12fc4a508f9bbc366687121cd adds the object type to the confirmation message. For example, "Are you sure you want to delete the notebook foo?" vs "Are you sure you want to delete foo?".

src/components/common.js
```diff
@@ -606,7 +606,7 @@ export const DeleteConfirmationModal = ({ objectName, objectType, children, ...p
     ...props
   }, [
     children || h(Fragment, [
-      div(['Are you sure you want to delete ',
+      div(['Are you sure you want to delete the ', objectType, ' ',
```
cahrens: Can you not use a templatized string here?

nawatts: Yes, a template string would work here. Changed in https://github.com/DataBiosphere/terra-ui/pull/2944/commits/3844470ed67a6e6890b4d72cdb4d97d8d8063fa9.

src/components/data/DataTable.js
```diff
@@ -410,18 +410,13 @@ const DataTable = props => {
...
       onConfirm: () => deleteColumn(deletingColumn),
       onDismiss: () => setDeletingColumn(undefined)
-    }, [
-      div(['Are you sure you want to permanently delete the column ',
-        span({ style: { fontWeight: 600, wordBreak: 'break-word' } }, deletingColumn), '?']),
-      div({ style: { fontWeight: 500, marginTop: '1rem' } }, 'This cannot be undone.')
-    ]),
-    !!clearingColumn && h(PromptedConfirmationModal, {
+    }),
+    !!clearingColumn && h(DeleteConfirmationModal, {
```
cahrens, line 419: What was the goal of moving from `PromptedConfirmationModal` to the delete one? It's pretty confusing to call something called "DeleteConfirmationModal" to do an operation other than deleting.

It seems like you have pretty much made `PromptedConfirmationModal` unnecessary, and the 2 things should just be collapsed into 1.

nawatts: The goal of [SU-56](https://broadworkbench.atlassian.net/browse/SU-56) is a consistent style of confirmation prompt for deleting things. Though the label here is "Clear column", it is deleting data from the table.

The reason for a component specific to confirming deletes is that, once more confirmation prompts are moved to `DeleteConfirmationModal`, `DeleteConfirmationModal` will get a yellow/warning color background, which wouldn't really make sense for a generic confirmation prompt.

Now, `PromptedConfirmationModal` exists as a slightly more generic version of `DeleteConfirmationModal`, but since it is not used anywhere else, it could be absorbed into `DeleteConfirmationModal`.

cahrens: I see, that makes sense. If you want to keep `PromptedConfirmationModal`, I'm fine with that.

nawatts: I think this sort of purposely high-friction confirmation is unlikely to be used elsewhere. I'll combine them.

PR #2946, opened 2022-04-11
[SU-87] Fix auto focus on last item in list in attribute editor

nawatts:
When editing an attribute in a data table, if the attribute is a list, clicking the "Add item" button should add another input and focus on that element.

Screen Shot 2022-04-11 at 9 19 01 AM

The AttributeInput component attaches a ref to the last input in the list and calls `focus` on it.

https://github.com/DataBiosphere/terra-ui/blob/f1ad9fbaafb4c8f6513bb5fe1485895679ab868f/src/components/data/data-utils.js#L683

https://github.com/DataBiosphere/terra-ui/blob/f1ad9fbaafb4c8f6513bb5fe1485895679ab868f/src/components/data/data-utils.js#L613-L616

Currently, this only works for string and reference attributes. The TextInput component used for those attribute types forwards its ref to the underlying `input` element. For number and boolean attributes, it causes an error because the `lastListItemInput ` ref references an instance of the `NumberInput` or `Switch` component, which do not expose a `focus` method.

This updates `NumberInput` and `Switch` to forward refs to their underlying `input` elements. This is straightforward for `NumberInput`. For `Switch`, we have to first get a reference to the `ReactSwitch` component instance and then grab the reference to the input element that it stores in `$inputRef`.

### DIFF
--- a/src/components/input.js
+++ b/src/components/input.js
@@ -146,44 +146,48 @@ export const SearchInput = ({ value, onChange, ...props }) => {
 
 export const DelayedSearchInput = withDebouncedChange(SearchInput)
 
-export const NumberInput = ({ onChange, onBlur, min = -Infinity, max = Infinity, onlyInteger = false, isClearable = true, tooltip, value, ...props }) => {
+export const NumberInput = forwardRefWithName('NumberInput', ({ onChange, onBlur, min = -Infinity, max = Infinity, onlyInteger = false, isClearable = true, tooltip, value, ...props }, ref) => {
   // If the user provided a tooltip but no other label, use the tooltip as the label for the input
   useLabelAssert('NumberInput', { tooltip, ...props, allowId: true, allowTooltip: true })
 
   const [internalValue, setInternalValue] = useState()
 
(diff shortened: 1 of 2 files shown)


src/components/input.js
```diff
@@ -146,44 +146,48 @@ export const SearchInput = ({ value, onChange, ...props }) => {
...
-      width: '100%',
-      paddingLeft: '1rem',
-      paddingRight: '0.25rem',
-      fontWeight: 400,
-      fontSize: 14,
-      backgroundColor: props.disabled ? colors.dark(0.25) : undefined
-    }
-  }, props))])
+  const numberInputChild = div([input({
+    ..._.merge({
```
petesantos, line 156: Do we need `merge` at all in this case? Could we just use the spread operator?

petesantos (approved):

src/components/input.js
nawatts: It looks like `merge` is used for overriding the component's default styles with the style prop, but that could also be done with the spread operator.

MatthewBemis (approved):

PR #2923, opened 2022-03-31
[SU-51] Data tab redesign part 2

nawatts:
Continuing from #2907, this moves the current "Upload .tsv" and "Add reference data" buttons in the data tab sidebar into a single "Import data" button at the top of the sidebar. This prepares for adding additional import options (the "data uploader") and making the sidebar sections collapsible.

To enable the feature flag, run `configOverridesStore.set({ isDataTabRedesignEnabled: true })` in the console.

## Before
Screen Shot 2022-03-31 at 3 55 48 PM

## After
Screen Shot 2022-03-31 at 3 55 30 PM
Screen Shot 2022-03-31 at 3 55 33 PM

PR #2907, opened 2022-03-24
[SU-51] Data tab redesign

nawatts:
This is behind a future flag so that we can collect feedback and iterate on the new design before releasing. To enable, run `configOverridesStore.set({ isDataTabRedesignEnabled: true })` in the console.

# Data table menu

Add menus to provide a place for additional future functionality.

This also makes the scope of actions more consistent. Currently, the first button downloads the entire table, the second button copies the current page, and the more actions menu operates on the selected rows. With this change, everything under export, edit, and open with operate on the selected rows.

## Before
Screen Shot 2022-03-24 at 11 25 15 AM

## After
Screen Shot 2022-03-24 at 4 16 25 PM

# Sidebar

Add a menu to take actions on the entire table. Current options are download or export to another workspace.

## Before
Screen Shot 2022-03-24 at 11 25 26 AM

## After
Screen Shot 2022-03-24 at 11 08 42 AM

PR #2947, opened 2022-04-11
[SU-3] Add link to data uploader from workspace data tab

nawatts:
This adds a link to the data uploader (https://app.terra.bio/#upload) to the workspace data tab.

This is part of the data tab redesign from #2907 and #2923. To see the new design, run `configOverridesStore.set({ isDataTabRedesignEnabled: true })` in the console and refresh.

Screen Shot 2022-04-11 at 11 39 36 AM

The data uploader is a multi-step process, where the user is prompted to select a workspace, then a "collection", then upload files, and finally upload metadata for those files. Since this is linking from a workspace's data tab, we want to be able to skip the select workspace step. While the `UploadData` component can accept an initial selection for workspace and collection via query parameters, it currently always starts out on the select workspaces step. This also updates `UploadData`'s initial state to start at a later step if some data is already provided through query parameters.

### DIFF
--- a/src/pages/Upload.js
+++ b/src/pages/Upload.js
@@ -739,9 +739,21 @@ const UploadData = _.flow( // eslint-disable-line lodash-fp/no-single-compositio
 
   // State
   const { query } = Nav.useRoute()
-  const [currentStep, setCurrentStep] = useState(StateHistory.get().currentStep || 'workspaces')
   const [workspaceId, setWorkspaceId] = useState(query.workspace)
   const [collection, setCollection] = useState(query.collection)
+  const [currentStep, setCurrentStep] = useState(() => {
+    if (StateHistory.get().currentStep) {
+      return StateHistory.get().currentStep
+    }
+    // If workspace and/or collection were provided in the URL, start at a later step
+    if (workspaceId) {
+      if (collection) {
+        return 'data'
+      }
+      return 'collection'
+    }
+    return 'workspaces'
+  })
   const [creatingNewWorkspace, setCreatingNewWorkspace] = useState(false)
   const [numFiles, setNumFiles] = useState(StateHistory.get().numFiles)
   const [tableName, setTableName] = useState(StateHistory.get().tableName)
@@ -775,7 +787,7 @@ const UploadData = _.flow( // eslint-disable-line lodash-fp/no-single-compositio
 
   // Make sure we have a valid step once the workspaces have finished loading
   useEffect(() => {
-    if (!stepIsEnabled(currentStep) && !loadingWorkspaces) {
+    if (!stepIsEnabled(currentStep) && workspaces) {
       let last = steps[0]
       for (const step of steps) {
         if (!step.test()) {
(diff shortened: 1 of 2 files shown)


src/pages/Upload.js
```diff
@@ -739,9 +739,21 @@ const UploadData = _.flow( // eslint-disable-line lodash-fp/no-single-compositio
...
+      return StateHistory.get().currentStep
+    }
+    // If workspace and/or collection were provided in the URL, start at a later step
+    if (workspaceId) {
+      if (collection) {
+        return 'data'
+      }
+      return 'collection'
+    }
+    return 'workspaces'
```
dmohs, line 755:
```suggestion
    return worspaceId ? collection ? 'data' : 'collection' : 'workspaces'
```

src/pages/Upload.js
```diff
@@ -775,7 +787,7 @@ const UploadData = _.flow( // eslint-disable-line lodash-fp/no-single-compositio
 
   // Make sure we have a valid step once the workspaces have finished loading
   useEffect(() => {
-    if (!stepIsEnabled(currentStep) && !loadingWorkspaces) {
+    if (!stepIsEnabled(currentStep) && workspaces) {
```
nawatts: This is an alternative way of determining if the workspaces have finished loading. `useWorkspaces` initially returns `false` for `loadingWorkspaces`, which caused this validation to happen before the workspaces list even begins loading.

src/pages/Upload.js
```diff
@@ -739,9 +739,21 @@ const UploadData = _.flow( // eslint-disable-line lodash-fp/no-single-compositio
...
+    }
+    // If workspace and/or collection were provided in the URL, start at a later step
+    if (workspaceId) {
+      if (collection) {
+        return 'data'
+      }
+      return 'collection'
+    }
+    return 'workspaces'
+  })
```
kyuksel: `Utils.cond` would help us write this without multiple `return`s, which may make it easier to reason about.

```suggestion
  const [currentStep, setCurrentStep] = useState(Utils.cond(
    [!!StateHistory.get().currentStep, () => StateHistory.get().currentStep],
    [!!workspaceId && !!collection, () => 'data'],
    [!!workspaceId && !collection, () => 'collection'],
    () => 'workspaces'
  ))
```

Note that we removed the laziness (the `() =>` piece) because it didn't seem needed. 

What do you think?

src/pages/Upload.js
kyuksel:
```suggestion
    if (!stepIsEnabled(currentStep) && !!workspaces) {
```

src/pages/workspaces/workspace/Data.js
```diff
@@ -690,6 +692,9 @@ const WorkspaceData = _.flow(
                 'aria-haspopup': 'dialog',
                 onClick: () => setUploadingFile(true)
               }, 'Upload TSV'),
+              h(MenuButton, {
+                href: `${Nav.getLink('upload')}?${qs.stringify({ workspace: workspaceId })}`
+              }, 'Open data uploader'),
```
kyuksel:
```suggestion
              }, ['Open data uploader']),
```

kyuksel (approved):
Looks great overall. A few comments inline.

MatthewBemis (approved):

PR #2951, opened 2022-04-12
[SU-56] Update confirmation prompts for deleting workspace data

nawatts:
Use common component for confirming deleting workspace data variables and reference data.

## Before
Screen Shot 2022-04-12 at 4 23 21 PM

Screen Shot 2022-04-12 at 4 23 31 PM

## After
Screen Shot 2022-04-15 at 7 12 42 AM

Screen Shot 2022-04-15 at 7 12 56 AM

Workspace data can be deleted from the "Workspace Data" panel of the Data tab. Hover over a table row to see the delete button.

Screen Shot 2022-04-12 at 4 22 47 PM

Reference data can be deleted from the sidebar on the Data tab.

Screen Shot 2022-04-12 at 4 21 53 PM

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -131,25 +131,26 @@ export const ReferenceDataImporter = ({ onSuccess, onDismiss, namespace, name })
 export const ReferenceDataDeleter = ({ onSuccess, onDismiss, namespace, name, referenceDataType }) => {
   const [deleting, setDeleting] = useState(false)
 
-  return h(Modal, {
-    onDismiss,
-    title: 'Confirm Delete',
-    okButton: h(ButtonPrimary, {
-      disabled: deleting,
-      onClick: async () => {
-        setDeleting(true)
-        try {
-          await Ajax().Workspaces.workspace(namespace, name).deleteAttributes(
-            _.map(key => `referenceData_${referenceDataType}_${key}`, _.keys(ReferenceData[referenceDataType]))
-          )
-          onSuccess()
-        } catch (error) {
-          await reportError('Error deleting reference data', error)
-          onDismiss()
-        }
+  return h(DeleteConfirmationModal, {
+    objectType: 'reference',
+    objectName: referenceDataType,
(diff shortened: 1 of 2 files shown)


MatthewBemis (approved):

src/components/data/data-utils.js
```diff
@@ -131,25 +131,26 @@ export const ReferenceDataImporter = ({ onSuccess, onDismiss, namespace, name })
...
+      } catch (error) {
+        reportError('Error deleting reference data', error)
+        onDismiss()
       }
-    }, ['Delete'])
-  }, [`Are you sure you want to delete ${referenceDataType}?`])
+    },
+    onDismiss
+  }, [
+    div(['Are you sure you want to delete the ', span({ style: { fontWeight: 600 } }, [referenceDataType]), ' reference data?']),
```
petesantos, line 151: I am not sure how to get to the reference deletion, but how would you feel about:
```suggestion
    div([`Are you sure you want to delete the ${b([referenceDataType])} reference data?`]),
```

petesantos (approved):

src/components/data/data-utils.js
nawatts: If you have an entity type `thing` for example, then you can upload a TSV that looks like:
```
membership:thing_set_id	thing
set1_name	thing1_name
set1_name	thing2_name
```

Then when you delete thing 1 or thing 2, it will prompt you to delete the set as well.

nawatts: Re: the `b` tag, sounds good to me. I'll go ahead and merge this as is and update #2962 to change that everywhere.

nawatts: Never mind https://github.com/DataBiosphere/terra-ui/pull/2951#discussion_r851443652, I got mixed up with something else. To delete reference data, first add some from the sidebar of the data tab.

Screen Shot 2022-04-15 at 3 18 53 PM

Screen Shot 2022-04-15 at 3 19 01 PM

PR #2953, opened 2022-04-13
[WOR-258] Move tests into team-based directories to make ownership clear.

cahrens:
I am trying to make test ownership easier to maintain/easier to find by adding a comment to each test.

I got the ownership from https://broadworkbench.atlassian.net/wiki/spaces/QA/pages/109248571/Triage+Processes#Terra-UI-Test-Inventory-%26-Release-Criteria


integration-tests/tests/link-to-new-workspace.js
```diff
@@ -1,3 +1,4 @@
+// This test is owned by the Data Explorer Team.
```
kyuksel: @petesantos Is `Data Explorer` the official name of the team? I've also seen `Data Catalog`.

kyuksel (approved):
LGTM with a question inline to be sure. 

Were you thinking of deprecating the table on the wiki? I like being able to see all ownership at one glance but I'm not sure if it's worth the additional maintenance.

integration-tests/tests/link-to-new-workspace.js
cahrens: Yeah, I was wondering if it should be `Data Catalog` given the JIRA Board name is `Data Catalog`.

cahrens:
> Were you thinking of deprecating the table on the wiki? I like being able to see all ownership at one glance but I'm not sure if it's worth the additional maintenance.

I'm not sure… I guess that somewhat depends on what QA wants. My hope though is that if QA wants to maintain the table, they can easily add new tests by looking at the comments in the test themselves (as opposed to posting in Slack as they do now to figure out ownership).

As a developer who is running the tests, it's much easier for me to just look at the comment in the code then to navigate to to the Confluence page, but that may not be true for QA.

aweng98:
Hi, may I suggest another approach? Create a visual hierarchical view that represents different teams. something like this:
```
integration-tests
  /tests
    /teams
      /interactive-anaysis
      /workspaces
      /batch-workflows
      /data-catglog
```     
This hierarchy can be useful. Names of teams can be programmed to send test fail notification to the right team's slack.

kyuksel:
> Create a visual hierarchical view that represents different teams

@aweng98 One issue is that we've frequently re-organized teams with different names.

aweng98:
Same problem applies to the inline code comments, right?

kyuksel:
> Same problem applies to the inline code comments, right?

Yes but updating comments is a lot easier. I'm not saying it's a deal breaker. Getting automated Slack notifications does sound quite helpful :)

cahrens:
@kyuksel @aweng98 I went ahead and moved the tests (leaving out the "teams" directory because I didn't think another level of nesting was useful). I also left in the comments at the top of each file, but could remove them if folks would prefer.

The directory structure has some shortcomings, namely tests that could be owned between teams will be more difficult to deal with. The example that immediately comes to my mind is "janitor", which is run as a test to clean up leaked resources. Currently it is only used to delete workspaces created by integration tests, so it is owned by the Workspaces team, but if "janitor" ends up being used more widely it may be "owned" by multiple teams.

But the advantage of automated alerts seems to outweigh that potential problem, and I do like how it more easily groups together related tests.

petesantos (approved):
This looks good.
I wonder if we could have a top level file with a key/value mapping (e.g. a config like file) to the location of each test. The require statements could then use this file rather than having the hard coded strings.

If a tests changes ownership, or a team name changes it may be a smaller change.

nawatts:
> I wonder if we could have a top level file with a key/value mapping (e.g. a config like file) to the location of each test. The require statements could then use this file rather than having the hard coded strings.
>
> If a tests changes ownership, or a team name changes it may be a smaller change.

It doesn't matter to me which way we go, but if we wanted to optimize for smaller changes in case of team changes, that would be easier with the original method of comments instead of folders. I'd vote against adding the complexity of indirect require paths.

cahrens:
> > I wonder if we could have a top level file with a key/value mapping (e.g. a config like file) to the location of each test. The require statements could then use this file rather than having the hard coded strings.
> > If a tests changes ownership, or a team name changes it may be a smaller change.
> 
> It doesn't matter to me which way we go, but if we wanted to optimize for smaller changes in case of team changes, that would be easier with the original method of comments instead of folders. I'd vote against adding more complexity in the form of indirect require paths.

I think indirect require paths is adding way more complexity than we need. Moving a test from one directory to another is pretty simple, and I really doubt it will happen that often (it's more likely that team directory names will change, but that refactoring is very straight-forward).

cahrens:
Putting this into draft status while we work out "ownership". See https://broadinstitute.slack.com/archives/C01EHNUM73R/p1649955538612209?thread_ts=1649876706.444109&cid=C01EHNUM73R for details.

PR #2954, opened 2022-04-13
[DC-290] Fixed Cookie banner so it flickers IN instead of OUT

lu-c:
Not sure if this is an okay solution. The problem was that when you refresh the page as a logged in user, the "Accept / Reject Cookies?" banner would flicker for a second before remembering who you were and disappearing. Now, the banner will be gone, but will pop in on top if you havent accepted it instead.

I think this works better, and would like to restyle the banner if that's okay too, so it's a pop over on existing content - I think this would make the banner appearing after a second feel better to the user. Wanted to get some opinions first tho!

### DIFF
--- a/src/components/CookieWarning.js
+++ b/src/components/CookieWarning.js
@@ -23,7 +23,11 @@ const CookieWarning = () => {
   const rejectCookies = async () => {
     const cookies = document.cookie.split(';')
     acceptCookies(false)
-    await Ajax(signal).Runtimes.invalidateCookie()
+    try {
+      await Ajax(signal).Runtimes.invalidateCookie()
+    } catch (err) {
+      // Do Nothing
+    }
     // Expire all cookies
     _.forEach(cookie => {
       // Find an equals sign and uses it to grab the substring of the cookie that is its name
@@ -33,7 +37,7 @@ const CookieWarning = () => {
     }, cookies)
     signOut()
   }
-  return !cookiesAccepted && aside({
+  return cookiesAccepted === false && aside({
     'aria-label': 'cookie consent banner',
     style: {
       flex: 0, height: 100, width: '100%',
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -110,7 +110,7 @@ export const initializeAuth = _.memoize(async () => {
         fenceStatus: isSignedIn ? state.fenceStatus : {},
         // Load whether a user has input a cookie acceptance in a previous session on this system,
         // or whether they input cookie acceptance previously in this session
-        cookiesAccepted: isSignedIn ? state.cookiesAccepted || getLocalPrefForUserId(user.getId(), cookiesAcceptedKey) : undefined,
+        cookiesAccepted: isSignedIn ? state.cookiesAccepted || getLocalPrefForUserId(user.getId(), cookiesAcceptedKey) : false,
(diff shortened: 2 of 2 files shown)


src/components/CookieWarning.js
```diff
@@ -23,7 +23,11 @@ const CookieWarning = () => {
   const rejectCookies = async () => {
     const cookies = document.cookie.split(';')
     acceptCookies(false)
-    await Ajax(signal).Runtimes.invalidateCookie()
+    try {
```
s-rubenstein, line 26: Why this change?

src/libs/auth.js
```diff
@@ -110,7 +110,7 @@ export const initializeAuth = _.memoize(async () => {
         fenceStatus: isSignedIn ? state.fenceStatus : {},
         // Load whether a user has input a cookie acceptance in a previous session on this system,
         // or whether they input cookie acceptance previously in this session
-        cookiesAccepted: isSignedIn ? state.cookiesAccepted || getLocalPrefForUserId(user.getId(), cookiesAcceptedKey) : undefined,
+        cookiesAccepted: isSignedIn ? state.cookiesAccepted || getLocalPrefForUserId(user.getId(), cookiesAcceptedKey) : false,
```
s-rubenstein, line 113: Just so I understand, the approach here is to wait until GAPI responds with the users sign in credentials, and defers making the decision until after we have a response from GAPI?

My question is (and the guiding philosophy of the original design): what if gapi is slow to respond? During that time, the user could navigate around our open for use portion of the site, and would not necessarily be aware that there are cookies in use. During that period (which admittedly would be small for 99.9% of cases), we would be in violation of GDPR guidelines. We could delay loading the site until after we have a response from GAPI, which would block users from navigating around until we have that response, but in that case, if signin is slow, we would just be having users staring at a blank screen while we wait for GAPI.

s-rubenstein:
Re: the pop over existing content, that seems fine. I think I was bouncing around in my head whether to do that or not when initially implementing, and decided not to so that if someone is leaving the banner there, they aren't impaired in their use of the site, but I don't think that is by any means a requirement of the banner, and am not sure that is a good thing or not.

src/components/CookieWarning.js
lu-c: I added the try catch because when it does error, the console bubbles that message back up, and it looks like something errored in the app. But the way the code is structured, it looks like it doesn't matter to us whether it errors or not, so catching this exception and moving on feels like the right approach.

If we want to actually handle it, we could add some sort of messaging in the ui?

s-rubenstein: Makes sense! Thank you!

src/libs/auth.js
lu-c: hm. That's a good point! I would rather not delay loading the site til GAPI, but that means having the cookie warning, and then potentially having it flicker away again, bringing back the original problem.

Right now, we're setting cookiesAccepted from this:
`cookiesAccepted: isSignedIn ? state.cookiesAccepted || getLocalPrefForUserId(...) : false`

What about if we move the check for state before the gapi checks, then have processUser update from localPrefForUserId if it's not set? I think this would still result in a flicker, but hopefully not as frequent as the current scenario

s-rubenstein: That seems fine to me. I don't know how much it will help, since state is ephemeral, so it will reset on refresh. (so the local pref is really the important thing, I think), but it might help some!

PR #2950, opened 2022-04-12
[SU-56] Update confirmation for deleting a workflow

nawatts:
Require typing a confirmation when deleting a workflow from a workspace.

## Before
Screen Shot 2022-04-12 at 1 37 46 PM

## After
Screen Shot 2022-04-12 at 2 40 15 PM

A workflow can be deleted from two places: the workflows list or the individual workflow page

Screen Shot 2022-04-12 at 1 38 55 PM

Screen Shot 2022-04-12 at 1 39 31 PM

PR #2955, opened 2022-04-13
[SU-56] Update confirmation prompt for deleting rows from a data table

nawatts:
Switched the confirmation prompt for deleting rows from a data table to the shared DeleteConfirmationModal component.

This also refactors the DeleteConfirmationModal to narrow its scope. It no longer awaits its `onConfirm` callback and displays loading spinner while that is going on. Thus, every component that uses it had to be updated to manage a spinner for the delete request. This affects clearing or deleting a column from a data table, deleting a workflow, deleting a notebook/analysis, and deleting a group.

## Before
Screen Shot 2022-04-13 at 12 14 44 PM

## After
Screen Shot 2022-04-14 at 12 23 37 PM

Note for testing: if deleting entities that are referenced by another entity (for example, deleting samples that are part of a sample set), the initial delete request will fail and the confirmation prompt will update to include also deleting any entities that refer to the selected entities.


src/components/data/data-utils.js
```diff
@@ -213,16 +207,12 @@ export const EntityDeleter = ({ onDismiss, onSuccess, namespace, name, selectedE
     div({ style: { maxHeight: 'calc((1em * 1.15 + 1.2rem + 1px) * 10.5)', overflowY: 'auto', margin: '0 -1.25rem' } },
       _.map(([i, entity]) => div({
         style: {
-          borderTop: (i === 0 && runningSubmissionsCount === 0) ? undefined : Style.standardLine,
+          borderTop: (i === 0 && runningSubmissionsCount === 0) ? undefined : `1px solid ${colors.light()}`,
```
nawatts, line 214: Unrelated. UX said they prefer this lighter color for dividers.

src/components/data/EntitiesContent.js
```diff
@@ -511,11 +511,14 @@ const EntitiesContent = ({
...
-          setDeletingEntities(false)
           setSelectedEntities({})
           setRefreshKey(_.add(1))
           Ajax().Metrics.captureEvent(Events.workspaceDataDelete, extractWorkspaceDetails(workspace.workspace))
-          loadMetadata()
+
+          // Avoid React warning about state update on unmounted component.
+          // Loading metadata will cause EntitiesComponent to be unmounted and replaced with a spinner while
+          // metadata is loading. Defer loading metadata until after the confirmation modal is dismissed.
+          Utils.onNextTick(loadMetadata)
```
nawatts: `DeleteConfirmationModal` calls `onDismiss` after `onConfirm`. This was meant to avoid having to duplicate code (usually some form of `setShouldModalDisplay(false)`) in both handlers. In retrospect, overall it may not have been a convenience since there are turning out to be more cases than I expected where the component that renders the confirmation modal gets unmounted after the deletion.

dmohs: As discussed, think about whether the two state transitions (and associated renders) in a row will be visually surprising to the user. If so, this is a reasonable approach. If not, you'll need to make some kind of change to the pattern.

src/components/data/data-utils.js
```diff
@@ -191,13 +186,12 @@ export const EntityDeleter = ({ onDismiss, onSuccess, namespace, name, selectedE
...
 
   const total = selectedKeys.length + additionalDeletions.length
-  return h(Modal, {
-    onDismiss,
-    title: 'Confirm Delete',
-    okButton: h(ButtonPrimary, {
-      disabled: deleting,
-      onClick: doDelete
-    }, ['Delete'])
+  return h(DeleteConfirmationModal, {
```
dmohs, line 194: Per our conversation, consider whether this is providing the right amount of friction in each case.

dmohs (approved):

src/components/data/EntitiesContent.js
nawatts: I took the need for using `Utils.onNextTick` here and in #2950, along with the need for adding the `dismissOnError` option, to be evidence that this was the wrong abstraction. Refactored DeleteConfirmationModal so that it only calls `onConfirm` and dismissing the modal and managing busy state are left to the parent component.

src/components/data/data-utils.js
nawatts: Removed the requirement to type a confirmation message from all the current confirmations that use this component (it still appears when deleting a workspace). Left it as an option (by passing a `confirmationPrompt` prop) since we may decide to use it when deleting an entire data table.

MatthewBemis (approved):

PR #2957, opened 2022-04-14
[DC-290] Updated cookie banner so to reduce the obtrusiveness of the flicker

lu-c:
Problem: Cookie banner flickers when you refresh the page (as a logged in user)

Attempted Solutions:
1. Moved the cookie banner to the bottom of the page
2. Added fade-in / fade-out (Though I think the code could probably be cleaner with some react thing I don't know about)
3. Added 3 second delay to fading in the first instance, so if a flicker would happen, it ... doesn't. This change is kind of potentially a controversial one, so please take a look at the generated branch and see if it feels okay!


src/components/CookieWarning.js
```diff
@@ -13,17 +14,55 @@ import { authStore } from 'src/libs/state'
...
 
+  useEffect(() => {
+    if (cookiesAccepted === undefined) {
+      timeout.current = setTimeout(showComponent, 3000)
+    } else {
+      cookiesAccepted ? hideComponent() : showComponent()
+    }
+  })
+
+  const showComponent = () => {
```
nawatts: Could this be simplified using a Transition or CSSTransition component?

https://reactcommunity.org/react-transition-group/

src/components/CookieWarning.js
```diff
@@ -13,17 +14,55 @@ import { authStore } from 'src/libs/state'
...
+  }
+
   const rejectCookies = async () => {
     const cookies = document.cookie.split(';')
     acceptCookies(false)
-    await Ajax(signal).Runtimes.invalidateCookie()
+    try {
+      await Ajax(signal).Runtimes.invalidateCookie()
+    } catch (err) {
+      // Do Nothing
```
nawatts: If this error is going to be ignored, is there a need to `await` the request vs sending it and moving on?

src/components/CookieWarning.js
```diff
@@ -13,17 +14,55 @@ import { authStore } from 'src/libs/state'
...
   const { cookiesAccepted } = useStore(authStore)
+  const timeout = useRef()
 
   const acceptCookies = acceptedCookies => {
     authStore.update(_.set('cookiesAccepted', acceptedCookies))
   }
 
+  useEffect(() => {
+    if (cookiesAccepted === undefined) {
+      timeout.current = setTimeout(showComponent, 3000)
```
s-rubenstein: I personally would check with Sarah Tahiri to see what she thinks about it from a compliance perspective. I'm worried that is too long still.

src/components/CookieWarning.js
```diff
@@ -33,27 +72,34 @@ const CookieWarning = () => {
...
     signOut()
   }
-  return !cookiesAccepted && aside({
+  return showWarning && aside({
     'aria-label': 'cookie consent banner',
     style: {
-      flex: 0, height: 100, width: '100%',
-      display: 'flex', alignItems: 'center', justifyContent: 'space-between',
-      backgroundColor: colors.dark(0.15), borderBottom: `6px solid ${colors.primary()}`
+      display: 'block', width: '100%',
```
s-rubenstein, line 69: With block, if the page is really tall, does it show up on the page? I would worry that we might need to do something like `position: 'absolute'`?

src/components/CookieWarning.js
```diff
@@ -13,17 +14,55 @@ import { authStore } from 'src/libs/state'
...
+  const [opacity, setOpacity] = useState(0)
   const { cookiesAccepted } = useStore(authStore)
+  const timeout = useRef()
 
   const acceptCookies = acceptedCookies => {
     authStore.update(_.set('cookiesAccepted', acceptedCookies))
   }
 
+  useEffect(() => {
+    if (cookiesAccepted === undefined) {
```
nawatts, line 36: When is `cookiesAccepted` expected to be undefined? (Asking since this PR changes the state for non-signed in users from undefined to false).

lu-c: cookiesAccepted is undefined when it is first declared, then set to be false after we check the gapi and see the user status. I'm using "undefined" to mean "still loading"

src/components/CookieWarning.js
lu-c: Yeah, it could be! I basically decided on this number because it just "felt good" to me. Have you tried it in the deployed pr branch?

https://pr-9-dot-bvdp-saturn-dev.appspot.com/

lu-c: https://broadinstitute.slack.com/archives/C01EHNUM73R/p1649949210579149?thread_ts=1649947007.669689&cid=C01EHNUM73R

Sarah thumbs uped this!

src/components/CookieWarning.js
lu-c: Ayy, yes, definitely. I didn't realize this was already in terra-ui, and didn't want to add new dependencies / packages. Let me try it out :)

lu-c: I just pushed an update that will use this now. The one issue I have with this is that it seems like the exit animation isnt working anymore - the component unmounts before it animates. Any ideas what I'm doing wrong?

src/components/CookieWarning.js
lu-c: Im sorry, I dont know how to answer this question, this was already there before - I guess we could remove the await!

EDIT: Or actually, wait, when I remove the await, there is an additional rejection error here for an uncaught error. I think the try catch doesnt work without it. Would you be okay if we leave it as is?

src/components/CookieWarning.js
lu-c: Im sorry, I don't understand the question! How does block get affected by the page length? 

I'm using `position: 'fixed'` already, which i prefer over `absolute`

src/components/CookieWarning.js
nawatts: Ah, I see. Thanks for the explanation.

src/components/CookieWarning.js
nawatts: It looks like opacity should be set to 0 for `exiting` in `transitionStyle`. Currently, opacity is set to 0 only when the `Transition` component reaches the `exited` state, so the CSS transition doesn't have a chance to run before `Transition` unmounts it child component.

src/components/CookieWarning.js
nawatts: To clarify, I was suggesting something like
```js
Ajax(signal).Runtimes.invalidateCookie().catch(() => {})
```
to avoid the `await` while still silencing the error. I'm also fine with leaving the `await` as is.

src/components/CookieWarning.js
```diff
@@ -33,29 +61,43 @@ const CookieWarning = () => {
...
   }
-  return !cookiesAccepted && aside({
+
+  return h(Transition, {
+    in: showWarning,
+    timeout: { exit: animTime * 1000 },
+    mountOnEnter: true,
+    unmountOnExit: true
+  }, [transitionState => aside({
     'aria-label': 'cookie consent banner',
```
nawatts: Since the banner is no longer present when the page first loads, should this have an alert role or aria-live attribute to make sure that it's read by a screen reader?

nawatts (approved):
One question about accessibility. Other than that, looks good.

src/components/CookieWarning.js
lu-c: Oh I like that a lot better, I'll make this change!

src/components/CookieWarning.js
lu-c: I'll give it an alert role - I think aria-live has to be applied to parent containing the element change, and technically, that would mean attaching it to the root div, which is probably too much to watch for changes on

src/components/CookieWarning.js
lu-c: Thank you! That is totally it :)

nawatts (dismissed):
`className : 'alert'` has no effect on accessibility. `role: 'alert'` would be needed for the content to be announced.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role

nawatts (approved):

s-rubenstein (approved):
LGTM

lu-c:
Updated cookie banner!

PR #2960, opened 2022-04-15
[SU-90, SU-54, SU-55] Workspace dashboard sidebar redesign 

MatthewBemis:
Mocks: https://projects.invisionapp.com/d/main#/projects/prototypes/21986264
Tickets: 
https://broadworkbench.atlassian.net/browse/SU-54
https://broadworkbench.atlassian.net/browse/SU-55
https://broadworkbench.atlassian.net/browse/SU-90

Redesign of the Workspace dashboard sidebar. 

* Obvious restyling per the above mock
* Sections are collapsable
* Browser will remember which sections were open/collapsed if the user leaves the workspace and comes back
* Two smaller SUE tickets included:
  * One to display the total bucket size
  * Another to display the date that the storage cost estimate was last updated

Before:

Screen Shot 2022-04-14 at 11 58 42 PM

After:
Screen Shot 2022-04-15 at 4 05 37 PM

Screen Shot 2022-04-15 at 4 05 54 PM

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -136,12 +136,16 @@ export const dashboard = {
     flex: 1, padding: '0 2rem 2rem 2rem'
   },
   rightBox: {
-    flex: 'none', width: 350, backgroundColor: colors.light(0.4),
-    padding: '0 1rem 2rem'
+    flex: 'none', width: 450, backgroundColor: colors.light(0.4),
+    padding: '1rem 1rem 2rem'
   },
   header: {
     ...elements.sectionHeader, textTransform: 'uppercase',
-    margin: '2.5rem 0 1rem 0', display: 'flex'
+    margin: '2.5rem 1rem 1rem 0', display: 'flex'
+  },
+  newHeader: {
+    ...elements.sectionHeader, color: colors.accent(0.8), textTransform: 'uppercase',
+    margin: '0.5rem', display: 'flex'
   },
   infoTile: {
     backgroundColor: colors.dark(0.15), color: 'black',
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -139,6 +139,9 @@ export const dashboard = {
     flex: 'none', width: 450, backgroundColor: colors.light(0.4),
     padding: '1rem 1rem 2rem'
   },
+  rightBoxContainer: {
+    borderRadius: 5, backgroundColor: 'white', padding: '0.5rem'
+  },
   header: {
     ...elements.sectionHeader, textTransform: 'uppercase',
     margin: '2.5rem 0 1rem 0', display: 'flex'
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -136,7 +136,7 @@ export const dashboard = {
     flex: 1, padding: '0 2rem 2rem 2rem'
   },
   rightBox: {
-    flex: 'none', width: 450, backgroundColor: colors.light(0.4),
+    flex: 'none', width: 470, backgroundColor: colors.light(0.4),
     padding: '1rem 1rem 2rem'
   },
   rightBoxContainer: {
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -146,7 +146,7 @@ export const dashboard = {
     ...elements.sectionHeader, textTransform: 'uppercase',
     margin: '2.5rem 0 1rem 0', display: 'flex'
   },
-  collapsableHeader: {
+  collapsibleHeader: {
     ...elements.sectionHeader, color: colors.accent(), textTransform: 'uppercase',
     margin: '0.5rem 0.5rem 0 0.5rem', display: 'flex', fontSize: 14
   },
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -141,9 +141,9 @@ export const dashboard = {
   },
   header: {
     ...elements.sectionHeader, textTransform: 'uppercase',
-    margin: '2.5rem 1rem 1rem 0', display: 'flex'
+    margin: '2.5rem 0 1rem 0', display: 'flex'
   },
-  newHeader: {
+  collapsableHeader: {
     ...elements.sectionHeader, color: colors.accent(), textTransform: 'uppercase',
     margin: '0.5rem 0.5rem 0 0.5rem', display: 'flex', fontSize: 14
   },
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -148,7 +148,7 @@ export const dashboard = {
   },
   collapsibleHeader: {
     ...elements.sectionHeader, color: colors.accent(), textTransform: 'uppercase',
-    margin: '0.5rem 0.5rem 0 0.5rem', display: 'flex', fontSize: 14
+    padding: '0.5rem 0.5rem 0 0.5rem', display: 'flex', fontSize: 14
   },
   infoTile: {
     backgroundColor: colors.dark(0.15), color: 'black',
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -145,7 +145,7 @@ export const dashboard = {
   },
   newHeader: {
     ...elements.sectionHeader, color: colors.accent(0.8), textTransform: 'uppercase',
-    margin: '0.5rem', display: 'flex'
+    margin: '0.5rem 0.5rem 0 0.5rem', display: 'flex', fontSize: 14
   },
   infoTile: {
     backgroundColor: colors.dark(0.15), color: 'black',
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -144,7 +144,7 @@ export const dashboard = {
     margin: '2.5rem 1rem 1rem 0', display: 'flex'
   },
   newHeader: {
-    ...elements.sectionHeader, color: colors.accent(0.8), textTransform: 'uppercase',
+    ...elements.sectionHeader, color: colors.accent(), textTransform: 'uppercase',
     margin: '0.5rem 0.5rem 0 0.5rem', display: 'flex', fontSize: 14
   },
   infoTile: {
--- a/src/pages/workspaces/workspace/Dashboard.js
+++ b/src/pages/workspaces/workspace/Dashboard.js
@@ -1,6 +1,6 @@
 import _ from 'lodash/fp'
 import { Fragment, useEffect, useImperativeHandle, useState } from 'react'
-import { dd, div, dl, dt, h, h2, i, span } from 'react-hyperscript-helpers'
+import { dd, div, dl, dt, h, h3, i, span } from 'react-hyperscript-helpers'
 import * as breadcrumbs from 'src/components/breadcrumbs'
 import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
 import Collapse from 'src/components/Collapse'
@@ -112,7 +112,7 @@ const RightBoxSection = ({ title, initialOpenState, onClick, children }) => {
   return div({ style: { paddingTop: '1rem' } }, [
     div({ style: Style.dashboard.rightBoxContainer }, [
       h(Collapse, {
-        title: h2({ style: Style.dashboard.collapsibleHeader }, [title]),
+        title: h3({ style: Style.dashboard.collapsibleHeader }, [title]),
         initialOpenState,
         titleFirst: true,
         onClick
--- a/src/pages/workspaces/workspace/Dashboard.js
+++ b/src/pages/workspaces/workspace/Dashboard.js
@@ -308,7 +308,7 @@ const WorkspaceDashboard = _.flow(
       div({ style: { paddingTop: '1rem' } }, [
         div({ style: { borderRadius: 5, backgroundColor: 'white', padding: '0.5rem' } }, [
           h(Collapse, {
-            title: h2({ style: Style.dashboard.newHeader }, ['Workspace information']),
+            title: h2({ style: Style.dashboard.collapsableHeader }, ['Workspace information']),
             initialOpenState: workspaceInfoPanelOpen !== undefined ? workspaceInfoPanelOpen : true,
             titleFirst: true,
             onClick: () => setWorkspaceInfoPanelOpen(workspaceInfoPanelOpen === undefined ? false : !workspaceInfoPanelOpen)
@@ -323,7 +323,7 @@ const WorkspaceDashboard = _.flow(
       div({ style: { paddingTop: '1rem' } }, [
         div({ style: { borderRadius: 5, backgroundColor: 'white', padding: '0.5rem' } }, [
           h(Collapse, {
-            title: h2({ style: Style.dashboard.newHeader }, ['Cloud information']),
+            title: h2({ style: Style.dashboard.collapsableHeader }, ['Cloud information']),
             initialOpenState: cloudInfoPanelOpen,
             titleFirst: true,
             onClick: () => setCloudInfoPanelOpen(!cloudInfoPanelOpen)
@@ -358,7 +358,7 @@ const WorkspaceDashboard = _.flow(
       div({ style: { paddingTop: '1rem' } }, [
         div({ style: { borderRadius: 5, backgroundColor: 'white', padding: '0.5rem' } }, [
           h(Collapse, {
-            title: h2({ style: Style.dashboard.newHeader }, ['Owners']),
+            title: h2({ style: Style.dashboard.collapsableHeader }, ['Owners']),
             initialOpenState: ownersPanelOpen,
             titleFirst: true,
             onClick: () => setOwnersPanelOpen(!ownersPanelOpen)
@@ -375,7 +375,7 @@ const WorkspaceDashboard = _.flow(
       !_.isEmpty(authorizationDomain) && div({ style: { paddingTop: '1rem' } }, [
         div({ style: { borderRadius: 5, backgroundColor: 'white', padding: '0.5rem' } }, [
           h(Collapse, {
-            title: h2({ style: Style.dashboard.newHeader }, ['Authorization domain']),
+            title: h2({ style: Style.dashboard.collapsableHeader }, ['Authorization domain']),
             initialOpenState: authDomainPanelOpen,
             titleFirst: true,
             onClick: () => setAuthDomainPanelOpen(!authDomainPanelOpen)
@@ -395,7 +395,7 @@ const WorkspaceDashboard = _.flow(
       div({ style: { paddingTop: '1rem' } }, [
         div({ style: { borderRadius: 5, backgroundColor: 'white', padding: '0.5rem' } }, [
           h(Collapse, {
-            title: h2({ style: Style.dashboard.newHeader }, ['Tags',
+            title: h2({ style: Style.dashboard.collapsableHeader }, ['Tags',
               h(InfoBox, { style: { marginLeft: '0.25rem' } }, [
                 `${getAppName()} is not intended to host personally identifiable information. Do not use any patient identifier including name,
                 social security number, or medical record number.`
--- a/src/pages/workspaces/workspace/Dashboard.js
+++ b/src/pages/workspaces/workspace/Dashboard.js
@@ -51,7 +51,7 @@ const roleString = {
 }
 
 const InfoRow = ({ title, subtitle, children }) => {
-  return dl({ role: 'row', style: { display: 'flex', justifyContent: 'space-between', margin: '1rem 0.5rem' } }, [
+  return div({ style: { display: 'flex', justifyContent: 'space-between', margin: '1rem 0.5rem' } }, [
     dt({ style: { width: 225 } }, [
       div({ style: { fontWeight: 500 } }, [title]),
       subtitle && div({ style: { fontWeight: 400, fontSize: 12 } }, [subtitle])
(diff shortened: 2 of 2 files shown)


src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -277,90 +305,132 @@ const WorkspaceDashboard = _.flow(
...
+            ]),
+            h(InfoRow, { title: 'Bucket Name' }, [
+              h(TooltipCell, [bucketName]),
+              h(ClipboardButton, { text: bucketName, style: { marginLeft: '0.25rem' } })
+            ]),
+            Utils.canWrite(accessLevel) && h(InfoRow, {
+              title: 'Estimated Storage Cost',
+              subtitle: storageCostEstimate ? `Updated on ${new Date(storageCostEstimateUpdated).toLocaleDateString()}` : 'Loading last updated...'
+            }, [storageCostEstimate || '$ ...']),
+            Utils.canWrite(accessLevel) && h(InfoRow, { title: 'Bucket Size' }, [bucketSize]),
```
nawatts: Should the last updated time for the bucket size also be shown here?

src/libs/style.js
```diff
@@ -136,13 +136,20 @@ export const dashboard = {
...
+    padding: '1rem 1rem 2rem'
+  },
+  rightBoxContainer: {
+    borderRadius: 5, backgroundColor: 'white', padding: '0.5rem'
   },
   header: {
     ...elements.sectionHeader, textTransform: 'uppercase',
     margin: '2.5rem 0 1rem 0', display: 'flex'
   },
+  collapsableHeader: {
```
nawatts: Nit on spelling: `collapsible` vs `collapsable`.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -133,18 +141,30 @@ const WorkspaceDashboard = _.flow(
...
     loadStorageCost()
+    loadBucketSize()
     loadConsent()
     loadWsTags()
     loadBucketLocation()
   }
 
   useImperativeHandle(ref, () => ({ refresh }))
 
+  const [workspaceInfoPanelOpen, setWorkspaceInfoPanelOpen] = useState(() => getLocalPref(persistenceId)?.workspaceInfoPanelOpen)
```
nawatts, line 173: Should this have a default boolean value like the others?

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -277,90 +305,132 @@ const WorkspaceDashboard = _.flow(
...
-            h(TooltipCell, [googleProject]),
-            h(ClipboardButton, { text: googleProject, style: { marginLeft: '0.25rem' } })
+      div({ style: { paddingTop: '1rem' } }, [
+        div({ style: Style.dashboard.rightBoxContainer }, [
+          h(Collapse, {
+            title: h2({ style: Style.dashboard.collapsableHeader }, ['Workspace information']),
+            initialOpenState: workspaceInfoPanelOpen !== undefined ? workspaceInfoPanelOpen : true,
+            titleFirst: true,
+            onClick: () => setWorkspaceInfoPanelOpen(workspaceInfoPanelOpen === undefined ? false : !workspaceInfoPanelOpen)
+          }, [
```
nawatts: It looks like this is repeated for each section. Could this be factored out into a component?

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -44,13 +47,16 @@ const roleString = {
...
-  PROJECT_OWNER: 'Proj. Owner'
+  PROJECT_OWNER: 'Project Owner'
 }
 
-const InfoTile = ({ title, children }) => {
-  return div({ style: Style.dashboard.infoTile }, [
-    div({ style: Style.dashboard.tinyCaps }, [title]),
-    div({ style: { fontSize: 12 } }, [children])
+const InfoRow = ({ title, subtitle, children }) => {
+  return div({ role: 'row', style: { display: 'flex', justifyContent: 'space-between', margin: '1rem 0.5rem' } }, [
```
nawatts: The rows of key-value pairs might be good place for a description list.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

nawatts (approved):

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Good question- I'll check with Lou on Tuesday

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Most definitely, I will take care of that

src/libs/style.js
MatthewBemis: Fixed https://github.com/DataBiosphere/terra-ui/pull/2960/commits/4a4c1ac7ddf1e49de8aee0a2f9dd76af07fda6bf

MatthewBemis:
Drafting until I have a chance to address the rest of the feedback. No need for review at the moment, I'll re-request once ready

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Added in a8c9e15e2570b00c02d8ac385cf2df0d7825b843

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Unfortunately this got split across two commits, but handled it in 57237d930f7eed1a44aad34d6e1e3ca4aa9dd452 and 73cc5135086c05c8eff9514095f460b30400af1a

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Changed to use description lists in b56fa95b0fc4cef797a80dd80b429799e1f37732

cahrens:
@MatthewBemis make sure to address this console error (by making the key unique):

````
react_devtools_backend.js:3973 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `RightBoxSection`. It was passed a child from WorkspaceDashboard. See https://reactjs.org/link/warning-keys for more information.

File: src/pages/workspaces/workspace/Dashboard.js [hunk 6]
```diff
@@ -102,6 +108,19 @@ const DashboardAuthContainer = props => {
   )
 }
 
+const RightBoxSection = ({ title, initialOpenState, onClick, children }) => {
+  return div({ style: { paddingTop: '1rem' } }, [
+    div({ style: Style.dashboard.rightBoxContainer }, [
+      h(Collapse, {
+        title: h2({ style: Style.dashboard.collapsibleHeader }, [title]),
```
cahrens: I think this should be h3 because the workspace path in the titlebar area is already an h2.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -277,90 +320,110 @@ const WorkspaceDashboard = _.flow(
...
-        ]) : 'Loading...'
+        ..._.map(({ membersGroupName }) => div({ style: { margin: '0.5rem', fontWeight: 500 } }, [membersGroupName]), authorizationDomain)
       ]),
-      h(Link, {
-        ...Utils.newTabLinkProps,
-        href: bucketBrowserUrl(bucketName)
-      }, ['Open in browser', icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })])
+      h(RightBoxSection, {
+        title: ['Tags',
+          h(InfoBox, { style: { marginLeft: '0.25rem' } }, [
```
cahrens: When you click on the info icon, the section expands/collapses. I think you need to stop the event propagation or the info icon needs to go somewhere else.

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Resolved 6a5f1d8a0a821002d4eb96b5196628912a2e0c25

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: This was on my docket to fix before undrafting :) Fixed in 560038c10892851e8a103c3d8974cc11a6a87155

cahrens: Sure, no problem. I just wanted to start getting a handle on the redesign as I'm also going to be working in this file momentarily.

cahrens:
Can you do something about the height of the title so that the focus indicator looks betters (solid rectangle)?

image

MatthewBemis:
@cahrens focus indicator is fixed in 9db8ffe83cc2bc7eae4841adbaf7fcb2d891ab7d, render warning is fixed in 800a7db2ad98b9bca152973906eeb7bc156d81cc

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: This one is a bit of a special case because it's always expanded by default. I preferred to leave it undefined and do the handling for this in the component props instead

nawatts (approved):

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -44,13 +47,16 @@ const roleString = {
...
   OWNER: 'Owner',
-  PROJECT_OWNER: 'Proj. Owner'
+  PROJECT_OWNER: 'Project Owner'
 }
 
-const InfoTile = ({ title, children }) => {
-  return div({ style: Style.dashboard.infoTile }, [
-    div({ style: Style.dashboard.tinyCaps }, [title]),
-    div({ style: { fontSize: 12 } }, [children])
+const InfoRow = ({ title, subtitle, children }) => {
```
cahrens, line 53: I did not know about `dl`! Right now you are outputting a `dl` for each row, but I think that the items should all be within the same `dl` per this example: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl
Otherwise it's not actually a list. I also wonder about the role of "row"; how did you decide on that?

BTW, when using the screenreader the elements within the collapse ALL get read out when the button is in the expanded state. I think that is because of the aria "owns" attribute. It seems excessively verbose to me, and I can't figure out how to actually navigate into the div contents to read the individually. But that is all within the existing Collapse element, so if there's an issue to be fixed there I don't expect it to be part of this PR.

cahrens:
Having the tags at the bottom causes some undesirable behavior with the select. If you have some sections expanded, when you click on the select for Tags it causes the whole window to shift so as to make the select contents completely visible. When you collapse the select, the window shifts again. Can you see if you can do something about that? I don't know if we could possibly put the Tags section up higher so that it is less likely to cause this problem, or if there is some way to change the shifting behavior.

image

MatthewBemis:
@cahrens the shifting should be corrected in e93378a5a03165bdd7da9800c61051459c0031e3

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: I'm not actually sure how I landed on the row role, it might just be an artifact of some old code that I scrapped. 

Agreed that the screenreader readout is a bit much, I can file a followup ticket to look into improving the behavior of that within a Collapse element

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -277,90 +320,118 @@ const WorkspaceDashboard = _.flow(
...
-        href: bucketBrowserUrl(bucketName)
-      }, ['Open in browser', icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })])
+      h(RightBoxSection, {
+        title: 'Tags',
+        info: span({ onClick: e => e.stopPropagation() }, [
+          h(InfoBox, { style: { marginLeft: '0.25rem' } }, [
+            `${getAppName()} is not intended to host personally identifiable information. Do not use any patient identifier including name,
+            social security number, or medical record number.`
+          ]),
+          (busy || !tagsList) && tagsPanelOpen && spinner({ size: '1rem', style: { marginLeft: '0.5rem' } })
```
cahrens: This spinner is slightly taller than the elements beside it, so it causes the select field to shift down when it is visible. This is easiest to reproduce by deleting a tag. Can you make it so there is no shifting?

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -277,90 +320,118 @@ const WorkspaceDashboard = _.flow(
...
+        ]),
+        initialOpenState: tagsPanelOpen,
+        onClick: () => setTagsPanelOpen(!tagsPanelOpen)
+      }, [
+        div({ style: { margin: '0.5rem' } }, [
+          !Utils.editWorkspaceError(workspace) && div({ style: { marginBottom: '0.5rem' } }, [
+            h(WorkspaceTagSelect, {
```
cahrens, line 414: There is a problem with this input regarding aria-expanded.

````
Elements must only use allowed ARIA attributes
1 of 1
[more info](https://dequeuniversity.com/rules/axe/4.4/aria-allowed-attr?application=AxeChrome)
Impact:
critical
Found on:
4/20/2022 at 9:35 AM
Found:
Automatically
Issue Tags:
cat.aria
wcag2a
wcag412
Issue Description
Ensures ARIA attributes are allowed for an element's role

Element location
#react-select-2-input
Element source
<input class="" autocapitalize="none" autocomplete="off" autocorrect="off" id="react-select-2-input" spellcheck="false" tabindex="0" type="text" aria-autocomplete="list" aria-expanded="false" aria-haspopup="true" aria-controls="react-select-2-listbox" aria-owns="react-select-2-listbox" aria-label="Add a tag" role="textbox" aria-describedby="react-select-2-placeholder" aria-multiline="false" value="" style="color: inherit; background: 0px center; opacity: 1; width: 100%; grid-area: 1 / 2 / auto / auto; font: inherit; min-width: 2px; border: 0px; margin: 0px; outline: 0px; padding: 0px;">
To solve this issue, you need to...
Fix the following:
ARIA attribute is not allowed: aria-expanded="false"

File: src/pages/workspaces/workspace/Dashboard.js [hunk 11]
```diff
@@ -277,90 +320,118 @@ const WorkspaceDashboard = _.flow(
...
+        dl({}, [
+          googleProject && h(InfoRow, { title: 'Cloud Name' }, [
+            h(GcpLogo, { title: 'Google Cloud Platform', role: 'img', style: { height: 16, width: 132, marginLeft: -15 } })
+          ]),
+          h(InfoRow, { title: 'Location' }, [bucketLocation ? h(Fragment, [
+            h(TooltipCell, [flag, ' ', regionDescription])
+          ]) : 'Loading...']),
+          h(InfoRow, { title: 'Google Project ID' }, [
+            h(TooltipCell, [googleProject]),
+            h(ClipboardButton, { text: googleProject, style: { marginLeft: '0.25rem' } })
```
cahrens: These clipboard buttons are missing labels.

```
Impact:
serious
Found on:
4/20/2022 at 9:35 AM
Found:
Automatically
Issue Tags:
cat.aria
wcag2a
wcag412
Issue Description
Ensures every ARIA button, link and menuitem has an accessible name

Element location
#unique-id-23
Element source
<div class="hover-style " role="button" tabindex="0" aria-disabled="false" id="unique-id-23" aria-describedby="unique-id-25" style="color: var(--hover-color, #4D72AA); cursor: pointer; font-weight: 500; display: inline; margin-left: 0.25rem; --app-hover-color:#718EBB;">
To solve this issue, you need to...
Fix at least one (1) of these issues:
Element does not have text that is visible to screen readers

aria-label attribute does not exist or is empty

aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty

Element has no title attribute

File: src/pages/workspaces/workspace/Dashboard.js (hunk 7 above)
Comment (cahrens) [inline src/pages/workspaces/workspace/Dashboard.js]: Unfortunately the button within a button is an accessibility issue.

```
Impact:
serious
Found on:
4/20/2022 at 9:35 AM
Found:
Automatically
Issue Tags:
cat.keyboard
wcag2a
wcag412
Issue Description
Ensures interactive controls are not nested as they are not always announced by screen readers or can cause focus problems for assistive technologies

Element location
div[aria-controls="unique-id-33"]
Element source
<div class="hover-style " role="button" tabindex="0" aria-disabled="false" aria-expanded="true" aria-controls="unique-id-33" aria-owns="unique-id-33" style="color: var(--hover-color, #4D72AA); cursor: pointer; font-weight: 500; display: flex; flex: 1 1 0%; align-items: center; margin-bottom: 0.5rem; --app-hover-color:#718EBB;">
To solve this issue, you need to...
Fix the following:
Element has focusable descendants

Related node:
#unique-id-34

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Fixed in a83252812c0421623a713abc1df2e643344fc4c4

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Resolved in 8cbc023befed1ead955996469629c27184da5b68

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Which tool are you using to find this accessibility warning? I am using the ANDI bookmarklet but I don't see this warning listed. I'd like to be able to test that any potential changes that I make are actually going to fix the problem

cahrens: I am using Axe Dev Tools (free). I find ANDI to be super clunky, and I can't get it to work at all on the PR site.

MatthewBemis: Thank you! I will give this one a try

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Filed as https://broadworkbench.atlassian.net/browse/SU-92

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: I have filed a followup ticket for this one because it is not an accessibility regression that was introduced in this PR and I'd like to limit the scope of this PR. Please see [SU-93](https://broadworkbench.atlassian.net/browse/SU-93).

src/pages/workspaces/workspace/Dashboard.js
MatthewBemis: Also, is it publicized anywhere internally that ANDI is not exhaustive when identifying accessibility issues? I believe ANDI was recommended during the training, but Axe seems much better

MatthewBemis: Fixed in f564bfd338526847393eccca9722105b5f991caa

MatthewBemis:
@cahrens I believe all comments are either addressed or have an associated Jira ticket

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -396,17 +395,17 @@ const WorkspaceDashboard = _.flow(
...
-          (busy || !tagsList) && tagsPanelOpen && spinner({ size: '1.5ch', style: { marginLeft: '0.5rem' } })
+        info: span({}, [
+          (busy || !tagsList) && tagsPanelOpen && spinner({ size: '1ch', style: { marginLeft: '0.5rem' } })
         ]),
         initialOpenState: tagsPanelOpen,
         onClick: () => setTagsPanelOpen(!tagsPanelOpen)
       }, [
         div({ style: { margin: '0.5rem' } }, [
+          div({ style: { marginBottom: '0.5rem' } }, [
+            `${getAppName()} is not intended to host personally identifiable information. Do not use
```
cahrens: Was this the suggestion from UX? I thought that perhaps we could put the info button near the search field, or at least make the text smaller, but defer to UX.

MatthewBemis: Lou has been out this week so we were making a judgment call here. I would prefer to not block the entire PR until he returns and instead choose to iterate on this small piece once he is back and can provide feedback.

That said, I am happy to make the text smaller.

MatthewBemis: eb625345380f2d8eecd01070e75707292994d3e4

cahrens: I'm out this afternoon anyway… so I won't get back to this until tomorrow unfortunately. You can check in with Ean Moody, who is the new UX person working with the Workspaces team.

MatthewBemis: Ean was fine with it but we ultimately landed on this abbreviated version:

Screen Shot 2022-04-21 at 1 32 26 PM

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -154,8 +188,17 @@ const WorkspaceDashboard = _.flow(
 
   const loadStorageCost = withErrorReporting('Error loading storage cost data', async () => {
     if (Utils.canWrite(accessLevel)) {
-      const { estimate } = await Ajax(signal).Workspaces.workspace(namespace, name).storageCostEstimate()
+      const { estimate, lastUpdated } = await Ajax(signal).Workspaces.workspace(namespace, name).storageCostEstimate()
       setStorageCostEstimate(estimate)
```
cahrens, line 192: Somewhat of a nit, but I think it would make sense to store these 2 things together since they are always fetched together.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -154,8 +188,17 @@ const WorkspaceDashboard = _.flow(
...
+      setStorageCostEstimateUpdated(lastUpdated)
+    }
+  })
+
+  const loadBucketSize = withErrorReporting('Error loading bucket size.', async () => {
+    if (Utils.canWrite(accessLevel)) {
+      const { usageInBytes, lastUpdated } = await Ajax(signal).Workspaces.workspace(namespace, name).bucketUsage()
+      setBucketSize(Utils.formatBytes(usageInBytes))
```
cahrens, line 200: Somewhat of a nit, but I think it would make sense to store these 2 things together since they are always fetched together.

cahrens (approved):
Good work, Matt! Thanks for taking so much time with this… I think it looks great.

MatthewBemis:
@cahrens Thank you! I also appreciate the time that you put into reviewing this and helping me improve it. This was my first significant UI contribution and I definitely learned a lot from the hands-on experience

PR #2961, opened 2022-04-15
[SU-56] Update confirmation prompt for deleting files

nawatts:
Use common component to confirm deleting objects from workspace bucket.

This is used in two places: one in the Files panel of a workspace's Data tab and the other in the data uploader (http://localhost:3000/#upload). If testing the data uploader, note that there's a bug ([SU-78](https://broadworkbench.atlassian.net/browse/SU-78)) where files don't immediately show up when they're uploaded.

This also fixes a bug. Currently, the delete object confirmation modal in the data uploader cannot be dismissed.

## Before
Screen Shot 2022-04-15 at 7 17 43 AM

## After
Screen Shot 2022-04-15 at 12 59 21 PM

### DIFF
--- a/src/components/data/FileBrowser.js
+++ b/src/components/data/FileBrowser.js
@@ -22,6 +22,23 @@ import { uploadFiles, useUploader } from 'src/libs/uploads'
 import * as Utils from 'src/libs/utils'
 
 
+export const DeleteObjectConfirmationModal = ({ object, ...props }) => {
+  const objectSize = _.toNumber(object.size)
+  return h(DeleteConfirmationModal, {
+    ...props,
+    objectType: 'file',
+    objectName: object.name
+  }, [
+    div(['Are you sure you want to delete the file ',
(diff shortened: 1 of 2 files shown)


dmohs (approved):
Showing the file name is a nice improvement. As a potential future enhancement, you could also show, for example, the file size. The logic behind this would be that very small files can be easily recreated, whereas large files less so. The confirmation dialog would be giving the user some indication as to how irreversible the action is.

nawatts:
> As a potential future enhancement, you could also show, for example, the file size. The logic behind this would be that very small files can be easily recreated, whereas large files less so. The confirmation dialog would be giving the user some indication as to how irreversible the action is.

Nice idea! Done in https://github.com/DataBiosphere/terra-ui/pull/2961/commits/569ab9bde5cc50619d89a53e5f735b2cba4dfd1f.

MatthewBemis (approved):

PR #2963, opened 2022-04-15
[DC-222] Use test specific entries for catalog preview

fboulnois:
The HCA snapshots and catalog entries are not available in all test environments. Use test catalog entries that specifically exist in all test environments.


petesantos (approved):

lu-c (approved):
lgtm

kyuksel:
@fboulnois I see that you just requested a review from me but your PR seems to be good to merge already?

cahrens:
Note that another option for situations like this is to mock out the Ajax responses so that you don't have to worry about environmental differences and test data.

PR #2964, opened 2022-04-15
[SU-71] Use select instead of radio buttons to choose change attribute types

nawatts:
Use a select menu instead of radio buttons to change the type of an attribute in a data table. This was requested by UX. It also avoids having to find space for another radio button when we add support for editing JSON attributes in the UI (SU-57).

## Before
Screen Shot 2022-04-15 at 4 56 09 PM

## After
Screen Shot 2022-04-15 at 4 55 34 PM

To get to this modal, hover over a cell in a data table and click the edit button.
Screen Shot 2022-04-15 at 4 58 25 PM

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -1,7 +1,7 @@
 import _ from 'lodash/fp'
 import pluralize from 'pluralize'
 import { Fragment, useEffect, useRef, useState } from 'react'
-import { div, fieldset, h, img, label, legend, li, p, span, ul } from 'react-hyperscript-helpers'
+import { div, fieldset, h, img, label, li, p, span, ul } from 'react-hyperscript-helpers'
 import Collapse from 'src/components/Collapse'
 import {
   absoluteSpinnerOverlay, ButtonOutline, ButtonPrimary, ButtonSecondary, Clickable, DeleteConfirmationModal, IdContainer, LabeledCheckbox, Link, RadioButton, Select, spinnerOverlay, Switch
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

nawatts:
Rebased to get the fix for the run-catalog-filter integration test.

dmohs (approved):
I am obligated by professional ethics to reject this PR. I am not going to to that, since I could be wrong. Instead, I will lay out my case and let the PR's author make the final decision.

Usability guidelines allow a dropdown to replace radio buttons when the number of options exceeds four [1]. While this may become true in the future, it is not true today, so this would be a usability regression. While there is a plan to include an additional option here, plans can change, and such a change would cause this regression to have an indefinite duration.

[1] https://www.nngroup.com/articles/listbox-dropdown/

Since dropdowns are significantly more cumbersome than radio buttons, I'd prefer to switch only when doing otherwise is prohibitive in some way. In this case, I think there is room for two columns: (1) String, Number, Boolean, and  (2) Reference, JSON. If a dropdown is the final decision, I'd prefer this be changed in the PR that adds the additional option.

kyuksel:
Converted to draft since the original author will be on vacation for a while.

nawatts:
Closing this as "won't do".

> While there is a plan to include an additional option here, plans can change, and such a change would cause this regression to have an indefinite duration.

Time has proven this statement wise. After discussion in #3049, we decided to only show the JSON option if the cell contains JSON. So most cells will still only have four options.

PR #2965, opened 2022-04-15
[WOR-211][WOR-168] Updates to footer

marctalbott:
Tickets: [WOR-211](https://broadworkbench.atlassian.net/browse/WOR-211), [WOR-168](https://broadworkbench.atlassian.net/browse/WOR-211)
* add terra.bio link to footer
* fix bug where footer would not stay anchored to the bottom of the page, specifically on the `SignIn` page

<!--
Hello, friend!
Thanks!

--!>


src/components/FooterWrapper.js
```diff
@@ -88,7 +95,7 @@ const FooterWrapper = ({ children, alwaysShow, fixedHeight }) => {
   const expandedFooterVisible = isExpanded || alwaysShow
 
   return div({ style: { display: 'flex', flexDirection: 'column', height: fixedHeight ? `calc(100vh - ${scrollBarHeight}px)` : '100%', flexGrow: 1 } }, [
-    div({ style: { display: 'flex', flexDirection: 'column', flexGrow: 1, height: `calc(100% - ${footerExists ? (!alwaysShow ? shrunkFooterHeight : 0) + (expandedFooterVisible ? expandedFooterHeight : 0) : 0}px)` } }, [children]),
+    div({ style: { display: 'flex', flexDirection: 'column', flexGrow: 1, height: 'max-content' } }, [children]),
```
marctalbott: I'm a bit confused by this. The old calculation takes many more parameters into account and I'm not sure if `max-content` really captures the complexity of the calculation. That said, `max-content` does keep the footer anchored to the bottom of the page and I couldn't find any other pages in Terra (logged in or not) where the footer looked wrong with my change.

Does someone who understands why we were doing this calculation in the first place see a problem with using `max-content` here? It also seems like we don't really use `max-content` often in the codebase, is there a reason for that?

cahrens: @s-rubenstein looking at the git history, I think you added this code a year ago. Do you happen to recall what the height complexity was about?

s-rubenstein: Oh gosh uhhh… maybe? I think it had something to do with screen jitter as it resized, but I am not 100% certain. There's some comments on the original PR here: https://github.com/DataBiosphere/terra-ui/pull/2418 from Brett, which may help clarify? It may also have to do with the BDC skin, which is far more complex.

nawatts: The complexity of the calculation looks related to the BioData Catalyst footer. Generally, setting the height to `100% - footerHeight` looks like maybe at one point the footer was intended to always be visible vs scrolling out of view?

nawatts: For what it's worth, removing the `height` style altogether also seems to fix the sign in page.

src/components/FooterWrapper.js
```diff
@@ -78,6 +78,13 @@ const FooterWrapper = ({ children, alwaysShow, fixedHeight }) => {
     }, [
       'Documentation', icon('pop-out', { size: 12, style: { marginLeft: '0.5rem' } })
     ]),
+    div({ style: styles.item }, '|'),
```
cahrens: @marctalbott can you refactor this code to loop over the similar items to remove the duplication? I am thinking of an "item" as the separator div followed by the anchor tag with popout styling (Security, Documentation, Terra.bio).

cahrens:
Note that you can rebase on dev to pick up the integration test fixes.

src/components/FooterWrapper.js
cahrens: Yes, I think we have come to the conclusion that there was an intention at one point for the footer to always be visible, even when there is a vertical scrollbar. Adam Mullen (PO) agrees with dropping that behavior and having the footer remain at the bottom of the scrollbar view in both views where the footer is present.

cahrens: I did some testing with the BioDataCatalyst (`configOverridesStore.set({isBioDataCatalyst: true })`), and the change doesn't seem to impact the behavior there. FWIW, it's odd that we don't persist the "expanded" state of the footer across page views.

marctalbott:
> Note that you can rebase on dev to pick up the integration test fixes.

👍 didn't realize there had been fixes, I just assumed I was dealing with the standard flakiness

cahrens (approved):

aherbst-broad (approved):

kyuksel:
Hi @marctalbott,

1. `import-cohort-data` is deterministically failing due to an issue the CI team is looking at. In the meantime we've been merging PRs that fail only due to that test so I kicked off another run for this PR as there was another test failure. Would you want me to merge your PR if only `import-cohort-data` ends up failing or would you prefer to wait for a clean pass?
2. Nick is on vacation so I've unassigned him. You already have the necessary approvals.

marctalbott:
> Hi @marctalbott,
> 
> 1. `import-cohort-data` is deterministically failing due to an issue the CI team is looking at. In the meantime we've been merging PRs that fail only due to that test so I kicked off another run for this PR as there was another test failure. Would you want me to merge your PR if only `import-cohort-data` ends up failing or would you prefer to wait for a clean pass?
> 2. Nick is on vacation so I've unassigned him. You already have the necessary approvals.

@kyuksel I can merge it if only `import-cohort-data` fails. I've been tracking the slack thread and just waiting for the fix so I could get a passing run, but if it's okay to merge with that test failing then I'll do so

marctalbott:
Merging with `import-cohort-data` failure. See comments above for justification
Screen Shot 2022-04-21 at 10 56 28 AM

PR #2966, opened 2022-04-20
[WOR-245, 246] Viewing Workspace dashboard for Azure workspace

cahrens:
This allows the Workspace Dashboard to be viewed for Azure workspaces (which currently can only be created by running Rawls locally against dev and using curl commands). Components created in WorkspaceContainer that require Google projects IDs or bucket locations are not instantiated, and all tabs except the Workspace Dashboard tab are not displayed (as currently they all depend on the Google project ID).

There are still things on the Workspace Dashboard that do not render with information because of the missing Google project ID, but those will be addressed in subsequent tickets.

image

### DIFF
--- a/integration-tests/tests/workspace-dashboard.js
+++ b/integration-tests/tests/workspace-dashboard.js
@@ -1,0 +1,116 @@
+// This test is owned by the Workspaces Team.
+const _ = require('lodash/fp')
+const { clickNavChildAndLoad, viewWorkspaceDashboard, withWorkspace } = require('../utils/integration-helpers')
+const { assertNavChildNotFound, findText } = require('../utils/integration-utils')
+const { withUserToken } = require('../utils/terra-sa-utils')
+
+
+const testGoogleWorkspace = _.flow(
+  withWorkspace,
+  withUserToken
+)(async ({ page, token, testUrl, workspaceName }) => {
+  await page.goto(testUrl)
+  await viewWorkspaceDashboard(page, token, workspaceName)
+  await findText(page, 'About the workspace')
+  // Click on each of the expected tabs
+  await clickNavChildAndLoad(page, 'data')
+  await clickNavChildAndLoad(page, 'notebooks')
+  await clickNavChildAndLoad(page, 'workflows')
+  await clickNavChildAndLoad(page, 'job history')
+})
+
+const googleWorkspaceDashboard = {
+  name: 'google-workspace',
+  fn: testGoogleWorkspace
+}
+
+const setAjaxMockValues = async (testPage, namespace, name, workspaceDescription) => {
+  const workspaceInfo = {
+    attributes: { description: workspaceDescription },
+    authorizationDomain: [],
+    bucketName: '',
+    createdBy: 'dummy@email.com',
+    createdDate: '2022-04-12T18:12:25.912Z',
+    googleProject: '',
+    isLocked: false,
+    lastModified: '2022-04-12T18:12:26.199Z',
+    name,
+    namespace,
+    workspaceId: '95accxxx-4c68-4e60-9c2e-c0863af11xxx',
+    workspaceVersion: 'v2'
+  }
+  const azureWorkspacesListResult = [{
+    accessLevel: 'OWNER',
+    public: false,
+    workspace: workspaceInfo,
+    workspaceSubmissionStats: { runningSubmissionsCount: 0 }
+  }]
+
+  const azureWorkspaceDetailsResult = {
+    azureContext: {
+      managedResourceGroupId: 'dummy-mrg-id',
+      subscriptionId: 'dummy-subscription-id',
+      tenantId: 'dummy-tenant-id'
+    },
+    workspaceSubmissionStats: { runningSubmissionsCount: 0 },
+    accessLevel: 'OWNER',
+    owners: ['dummy@email.comm'],
+    workspace: workspaceInfo,
+    canShare: true,
+    canCompute: true
+  }
+
+  return await testPage.evaluate((azureWorkspacesListResult, azureWorkspaceDetailsResult, namespace, name) => {
+    const detailsUrl = new RegExp(`api/workspaces/${namespace}/${name}[^/](.*)`, 'g')
+    const submissionsUrl = new RegExp(`api/workspaces/${namespace}/${name}/submissions(.*)`, 'g')
+    const tagsUrl = new RegExp(`api/workspaces/${namespace}/${name}/tags(.*)`, 'g')
+
+    window.ajaxOverridesStore.set([
+      {
+        filter: { url: /api\/workspaces\/saturn-integration-test-dev(.*)/ },
+        fn: () => () => Promise.resolve(new Response(JSON.stringify([azureWorkspaceDetailsResult]), { status: 200 }))
+      },
+      {
+        filter: { url: tagsUrl },
+        fn: () => () => Promise.resolve(new Response(JSON.stringify([]), { status: 200 }))
+      },
+      {
+        filter: { url: submissionsUrl },
+        fn: () => () => Promise.resolve(new Response(JSON.stringify([]), { status: 200 }))
+      },
+      {
+        filter: { url: detailsUrl },
+        fn: () => () => Promise.resolve(new Response(JSON.stringify(azureWorkspaceDetailsResult), { status: 200 }))
+      },
+      {
+        filter: { url: /api\/workspaces[^/](.*)/ },
+        fn: () => () => Promise.resolve(new Response(JSON.stringify(azureWorkspacesListResult), { status: 200 }))
+      }
+    ])
+  }, azureWorkspacesListResult, azureWorkspaceDetailsResult, namespace, name)
+}
+
+const testAzureWorkspace = withUserToken(async ({ page, token, testUrl }) => {
+  const workspaceDescription = 'azure workspace description'
+  const workspaceName = 'azure-workspace'
+
+  // Must load page before setting mock responses.
+  await page.goto(testUrl)
+  await findText(page, 'View Workspaces')
+  await setAjaxMockValues(page, 'azure-workspace-ns', workspaceName, workspaceDescription)
+
+  await viewWorkspaceDashboard(page, token, workspaceName)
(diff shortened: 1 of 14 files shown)


src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -212,7 +214,7 @@ const useCloudEnvironmentPolling = googleProject => {
   }
   const load = async maybeStale => {
     try {
-      const [newDisks, newRuntimes] = googleProject ? await Promise.all([
+      const [newDisks, newRuntimes] = !!googleProject ? await Promise.all([
```
cahrens, line 219: Handle empty string as false (the case for Azure workspace).

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -224,8 +224,9 @@ const Analyses = _.flow(
   }),
   withViewToggle('analysesTab')
 )(({
-  apps, name: wsName, namespace, workspace, workspace: { accessLevel, canShare, workspace: { googleProject, bucketName } },
-  refreshApps, onRequesterPaysError, runtimes, persistentDisks, refreshRuntimes, appDataDisks
+  name: wsName, namespace, workspace, workspace: { accessLevel, canShare, workspace: { googleProject, bucketName } },
```
cahrens, line 227: I have put all the analyses-related items into an object to make it clearer who depends on those (as they are currently Google-specific).

These other classes use wrapWorkapce but do not depend on analysesData: Dashboard (workspace), Data, JobHistory, Workflows, SubmissionDetails, WorkflowDashboard, WorkflowView.

integration-tests/utils/integration-helpers.js
```diff
@@ -198,22 +200,30 @@ const enableDataCatalog = async (page, testUrl, token) => {
...
-  await overrideConfig(page, { isAnalysisTabVisible: true })
+const clickNavChildAndLoad = async (page, tab) => {
+  await noSpinnersAfter(page, { action: () => click(page, navChild(tab)) })
+}
 
+const viewWorkspaceDashboard = async (page, token, workspaceName) => {
   await click(page, clickable({ textContains: 'View Workspaces' }))
   await signIntoTerra(page, token)
   await dismissNotifications(page)
+  await fillIn(page, input({ placeholder: 'SEARCH WORKSPACES' }), workspaceName)
```
cahrens, line 211: This is needed for the case of lots of workspaces being present (which is true when logging in with Scarlett).

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -75,13 +75,13 @@ const WorkspaceTabs = ({
 
   const tabs = [
     { name: 'dashboard', link: 'workspace-dashboard' },
-    { name: 'data', link: 'workspace-data' },
-    ...(!isAnalysisTabVisible() ? [{ name: 'notebooks', link: 'workspace-notebooks' }] : []),
+    ...(isGoogleWorkspace ? [{ name: 'data', link: 'workspace-data' }] : []),
```
cahrens, line 78: For the initial rollout of Azure workspaces, we are not displaying tabs that depend on Google project IDs/bucket information. As individual teams change their UI to work with Azure workspaces, they can re-enable their tabs.

This is probably the most controversial part of this PR-- it means that all tabs except the Workspace Dashboard won't appear until we have received the Workspace details from the server and know that it is a Google project. This is generally pretty fast (and something that we then cache), and it's not like you can actually do anything useful with the tabs during that time anyway. However, it is a change in behavior so I want to call it out, and I also am going to touch base with Ean in UX to get his opinion.

The other option would be to show all the tabs initially, and then have unsupported ones disappear if we have an Azure workspace. This would not impact existing Google workspaces, but in general I think that it is more jolting for elements to disappear than to appear. And it seems unlikely at this point that the tabs for Azure and Google workspaces will be identical, so there will most likely always be a need for things to appear/disappear.

integration-tests/tests/workspace-dashboard.js
```diff
@@ -0,0 +1,116 @@
+// This test is owned by the Workspaces Team.
+const _ = require('lodash/fp')
```
cahrens, line 2: I have run this test 150 times with test-flakes and got 0 failures.

src/pages/workspaces/workspace/WorkspaceContainer.js
cahrens: I checked with @eepMoody and he agrees it is better for the tabs to appear than disappear. He did suggest another option of disabling the tabs for Azure, but we came to the conclusion that it may introduce confusion (Azure users aren't necessarily existing users, so having tabs disabled just adds confusion for folks who wouldn't necessarily expect that functionality, plus some of the tabs may never be supported on Azure).

integration-tests/utils/integration-helpers.js
```diff
@@ -198,22 +200,30 @@ const enableDataCatalog = async (page, testUrl, token) => {
...
   await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: workspaceName })) })
-  await noSpinnersAfter(page, { action: () => click(page, navChild('analyses')) })
+}
+
+const performAnalysisTabSetup = async (page, token, testUrl, workspaceName) => {
```
cahrens, line 215: FYI @jdcanas some I did some refactoring here to share the functionality of viewing the workspace dashboard.

src/pages/workspaces/workspace/Analyses.js
cahrens: FYI @jdcanas regarding changes to Analyses-related usage of `wrapWorkspace`.

integration-tests/tests/workspace-dashboard.js
cahrens: The test of the Google workspace is against real sever calls, the one for the Azure workspace uses Ajax mocks.

These tests will be extended (for both Google and Azure workspaces) as we continue to evolve the Workspace Dashboard code.

integration-tests/tests/workspace-dashboard.js
```diff
@@ -0,0 +1,116 @@
...
+
+  const azureWorkspaceDetailsResult = {
+    azureContext: {
+      managedResourceGroupId: 'mrg-terra-workspace-20220412104730',
+      subscriptionId: '3efc5bdf-be0e-44e7-b1d7-c08931e3c16c',
+      tenantId: 'fad90753-2022-4456-9b0a-c7e5b934e408'
+    },
+    workspaceSubmissionStats: { runningSubmissionsCount: 0 },
+    accessLevel: 'OWNER',
+    owners: ['cahrens@gmail.com'],
```
aherbst-broad: Nitpick: You may want to use an explicitly fake user + email domain here (i.e, example@example.com) in case anyone scrapes this code.

aherbst-broad (approved):
I pulled this branch and played with it against an azure workspace, it worked well.

integration-tests/tests/workspace-dashboard.js
cahrens: Good point… I will also change the Azure subscription info so that it isn't real. I just dumped the actual content from my workspace to populate the test.

cahrens: Updated and force pushed.

src/components/FooterWrapper.js
```diff
@@ -43,15 +43,15 @@ const FooterWrapper = ({ children, alwaysShow, fixedHeight }) => {
   })
 
   const popoutItem = ({ link, displayName }) => {
-    return [
+    return h(Fragment, [
```
cahrens, line 46: @marctalbott I noticed this console warning that was introduced in https://github.com/DataBiosphere/terra-ui/pull/2965. Sorry that I didn't catch it on the review.

````
react_devtools_backend.js:3973 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `FooterWrapper`. See https://reactjs.org/link/warning-keys for more information.
    at div
    at FooterWrapper (http://localhost:3000/static/js/src_appLoader_js-data_image_svg_xml_3Csvg_xmlns_27http_www_w3_org_2000_svg_27_viewBox_270_0_1-d7b3bc.chunk.js:5496:23)
    at HeroWrapper (http://localhost:3000/static/js/src_appLoader_js-data_image_svg_xml_3Csvg_xmlns_27http_www_w3_org_2000_svg_27_viewBox_270_0_1-d7b3bc.chunk.js:15996:32)
    at SignIn
    at AuthContainer (http://localhost:3000/static/js/src_appLoader_js-data_image_svg_xml_3Csvg_xmlns_27http_www_w3_org_2000_svg_27_viewBox_270_0_1-d7b3bc.chunk.js:759:23)
    at ErrorWrapper (http://localhost:3000/static/js/src_appLoader_js-data_image_svg_xml_3Csvg_xmlns_27http_www_w3_org_2000_svg_27_viewBox_270_0_1-d7b3bc.chunk.js:5046:208)
    at LocationProvider (http://localhost:3000/static/js/src_appLoader_js-data_image_svg_xml_3Csvg_xmlns_27http_www_w3_org_2000_svg_27_viewBox_270_0_1-d7b3bc.chunk.js:40029:24)
    at Main

File: src/pages/workspaces/workspace/WorkspaceContainer.js (hunk 4 above)
Comment (dmohs) [reply src/pages/workspaces/workspace/WorkspaceContainer.js:78]: Thanks for this discussion. I'm on-board, and we can iterate once we're able to get a feel for it.

File: src/pages/workspaces/workspace/WorkspaceContainer.js [hunk 9]
```diff
@@ -212,7 +216,7 @@ const useCloudEnvironmentPolling = googleProject => {
   }
   const load = async maybeStale => {
     try {
-      const [newDisks, newRuntimes] = googleProject ? await Promise.all([
+      const [newDisks, newRuntimes] = !!googleProject ? await Promise.all([
```
dmohs, line 219: In my judgment, considering empty to be falsy is a bug. In this case, a bug in the language spec. My preference would be:
```suggestion
      const [newDisks, newRuntimes] = !_.isEmpty(googleProject) ? await Promise.all([
```

dmohs (approved):

PR #2968, opened 2022-04-20
[IA-3209] fix "Error loading outdated analyses" error in RStudio

gpcarr:
When in RStudio, users were getting notification errors that said `Error loading outdated analyses` - this PR fixes that

I was able to recreate the bug locally and then validate that the code change makes it so that the error goes away and all other functionality around rmd syncing works as expected

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/pages/workspaces/workspace/applications/AppLauncher.js
+++ b/src/pages/workspaces/workspace/applications/AppLauncher.js
@@ -134,8 +134,11 @@ const ApplicationLauncher = _.flow(
 
   const checkForOutdatedAnalyses = async ({ googleProject, bucketName }) => {
     const analyses = await Ajax(signal).Buckets.listAnalyses(googleProject, bucketName)
-    return _.filter(analysis => _.endsWith(`.${tools.RStudio.ext}`, analysis?.name) &&
-      analysis?.metadata[hashedOwnerEmail] === 'outdated', analyses)
+    return _.filter(({ analysis }) => {
+      const metadataMap = analysis?.metadata
+      return _.endsWith(`.${tools.RStudio.ext}`, analysis?.name) && metadataMap &&
+      analysis?.metadata[hashedOwnerEmail] === 'outdated'
+    }, analyses)
   }
 
   useOnMount(() => {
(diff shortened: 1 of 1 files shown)


jdcanas (approved):

Qi77Qi (approved):

petesantos (approved):

Qi77Qi:
merging since tests failed due to known flakiness

cahrens:
@Qi77Qi @gpcarr there needs to be a ticket linked to this PR.

Qi77Qi:
I think it's okay to have `[no ticket]` in the title...or since when that changed?

cahrens:
> I think it's okay to have `[no ticket]` in the title...or since when that changed?

Every change merged needs a ticket, and the title is where we put it.

Qi77Qi:
I think what I'm asking is where does this requirement come from? Compliance? or UI approvers decided that's the case one day and started enforcing it? (if latter, what's the context?)

cahrens:
> I think what I'm asking is where does this requirement come from? Compliance? or UI approvers decided that's the case one day and started enforcing it?

It is my understanding that it is for compliance reasons, and not specific to terra-ui. I think that it may not have been enforced consistently in terra-ui until a few months ago though.

Qi77Qi:
If for compliance, I don't think that's a must-have... `no ticket` is perfectly acceptable (there was discussions with QA team and decided this way a while ago)

I agree in most cases a ticket is good to have (we should have it in title most times), but I wouldn't want to discourage devs from contributing becuz they're expected to create a ticket first. So I think for small maintenance or bug fixes here and there, `no ticket` should be acceptable

cahrens:
> If for compliance, I don't think that's a must-have... `no ticket` is perfectly acceptable (there was discussions with QA team and decided this way a while ago)
> 
> I agree in most cases a ticket is good to have (we should have it in title most times), but I wouldn't want to discourage devs from contributing becuz they're expected to create a ticket first. So I think for small maintenance or bug fixes here and there, `no ticket` should be acceptable

I would like to learn more about the circumstances in which a ticket is not required… this is the first I have heard of it. I will send an email to folks and cc you.

PR #2969, opened 2022-04-20
[IA-3336] fix hail option in image select

jdcanas:
This fixes an issue that caused the UI to stacktrace and crash when the option for hail is selected in the compute modal dropdown.

This isn't as simple as checking if its undefined, we need to provide intelligent default so the compute modal appropriately displays cost when a user is creating a hail runtime, as so (hail runtimes don't have persistent disks. I verified that this is being handled properly for runtime creation as well):
image

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -496,7 +496,7 @@ export const ComputeModalBase = ({
    */
   const getPendingDisk = () => {
     const { persistentDisk: desiredPersistentDisk } = getDesiredEnvironmentConfig()
-    return { size: desiredPersistentDisk.size, status: 'Ready', diskType: desiredPersistentDisk.diskType }
+    return { size: desiredPersistentDisk ? desiredPersistentDisk.size : 0, status: 'Ready', diskType: desiredPersistentDisk ? desiredPersistentDisk.diskType : pdTypes.standard }
   }
 
   /**
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -495,8 +495,8 @@ export const ComputeModalBase = ({
    * is necessary to compute the cost for potential new disk configurations.
    */
   const getPendingDisk = () => {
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):
I'm able ot select hail image on PR site

src/components/ComputeModal.js
```diff
@@ -496,7 +496,7 @@ export const ComputeModalBase = ({
    */
   const getPendingDisk = () => {
     const { persistentDisk: desiredPersistentDisk } = getDesiredEnvironmentConfig()
```
petesantos: How would you feel about using default values in the destructure? This should simplify this and prevent the case where `size` or `diskType` does not exist and gets set to `undefined`
```suggestion
    const { persistentDisk: { size = 0, diskType = pdTypes.standard } = {}  } = getDesiredEnvironmentConfig()
```

src/components/ComputeModal.js
```diff
@@ -496,7 +496,7 @@ export const ComputeModalBase = ({
...
-    return { size: desiredPersistentDisk.size, status: 'Ready', diskType: desiredPersistentDisk.diskType }
+    return { size: desiredPersistentDisk ? desiredPersistentDisk.size : 0, status: 'Ready', diskType: desiredPersistentDisk ? desiredPersistentDisk.diskType : pdTypes.standard }
```
petesantos: With the above suggestion in place this would get much smaller / easier to parse
```suggestion
    return { size, status: 'Ready', diskType}
```

cahrens:
@jdcanas @Qi77Qi make sure to create a ticket and link it.

src/components/ComputeModal.js
jdcanas: hmm, didn't think to use a nested destructure. I like this a lot better. Thanks!

petesantos (approved):

PR #2974, opened 2022-04-22
[AJ-156] Missing snapshot

davidangb:
https://broadworkbench.atlassian.net/browse/AJ-156

To cause the error:
1. Add a TDR snapshot-by-reference to your workspace
2. Bind a workflow (method config) to that snapshot-by-reference
3. In the Data tab, rename the snapshot-by-reference
4. Attempt to view the workflow.

Before this PR:
Screenshot (13)

After this PR:
Screenshot (15)
Screenshot (14)

Snapshot-by-references are identified by their reference name, which is editable in the UI. The error happened because there are two ajax calls that return errors if the snapshot-by-reference's name isn't found. This PR adds error-handling around both of those ajax calls, as well as some state manipulation and UI error messaging to the end user.

### DIFF
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -476,12 +476,13 @@ const WorkflowView = _.flow(
         !isRedacted ? filterConfigIO(inputsOutputs) : _.identity
       )(config)
 
-      let selectedSnapshotEntityMetadata = undefined
-      if (modifiedConfig.dataReferenceName) {
-        try {
-          selectedSnapshotEntityMetadata = await Ajax(signal).Workspaces.workspace(namespace, name).snapshotEntityMetadata(googleProject, modifiedConfig.dataReferenceName)
-        } catch (error) {
-          // noop
+      const selectedSnapshotEntityMetadata = async () => {
+        if (modifiedConfig.dataReferenceName) {
+          try {
+            return await Ajax(signal).Workspaces.workspace(namespace, name).snapshotEntityMetadata(googleProject, modifiedConfig.dataReferenceName)
+          } catch (error) {
+            return undefined
+          }
         }
       }
 
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -785,7 +785,7 @@ const WorkflowView = _.flow(
                   isDisabled: !!Utils.editWorkspaceError(ws) || !!snapshotReferenceError,
                   'aria-label': 'Snapshot table selector',
                   isClearable: false,
-                  value: modifiedConfig.dataReferenceName && !!!snapshotReferenceError ? modifiedConfig.rootEntityType : undefined,
+                  value: modifiedConfig.dataReferenceName && !snapshotReferenceError ? modifiedConfig.rootEntityType : undefined,
                   onChange: ({ value }) => {
                     this.setState(_.set(['modifiedConfig', 'rootEntityType'], value))
                     this.setState(_.unset(['modifiedConfig', 'entityName']))
--- a/src/pages/workspaces/workspace/workflows/WorkflowView.js
+++ b/src/pages/workspaces/workspace/workflows/WorkflowView.js
@@ -782,10 +782,10 @@ const WorkflowView = _.flow(
               ]),
               entitySelectionModel.type === processSnapshotTable ? div({ style: { margin: '2rem 0 0 2rem' } }, [
                 h(Select, {
-                  isDisabled: !!Utils.editWorkspaceError(ws),
+                  isDisabled: !!Utils.editWorkspaceError(ws) || !!snapshotReferenceError,
                   'aria-label': 'Snapshot table selector',
                   isClearable: false,
-                  value: modifiedConfig.dataReferenceName ? modifiedConfig.rootEntityType : undefined,
+                  value: modifiedConfig.dataReferenceName && !!!snapshotReferenceError ? modifiedConfig.rootEntityType : undefined,
                   onChange: ({ value }) => {
                     this.setState(_.set(['modifiedConfig', 'rootEntityType'], value))
                     this.setState(_.unset(['modifiedConfig', 'entityName']))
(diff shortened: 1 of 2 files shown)


src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -423,17 +423,35 @@ const WorkflowView = _.flow(
...
       if (e.status === 404) {
-        return false
+        const errmsg = await e.text()
+        // distinguish between snapshot-reference-not-found and workflow-not-found
+        if (errmsg?.includes('Reference name') && errmsg?.includes('does not exist in workspace')) {
+          this.setState(_.set(['snapshotReferenceError', errmsg]))
+          return true
+        } else {
+          return false
+        }
```
davidangb, line 438: this is kinda gross, but the back end API does not help us here. It returns 404 in cases where the bound snapshot reference isn't found as well as in cases where the workflow (method config) version isn't found. We have to parse the message itself to distinguish.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -152,8 +152,8 @@ const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange,
         cellRenderer: ({ rowIndex }) => {
           const io = sortedData[rowIndex]
           const { name, optional, inputType } = io
-          const value = config[which][name] || ''
-          const error = errors[which][name]
+          const value = _.get([which, name], config) || ''
+          const error = _.get([which, name], errors)
```
davidangb: in cases where the snapshot reference isn't found, the call to validate() fails. Then, the `errors` object doesn't contain an `outputs` key. Without the lodash safety of `_.get`, this causes failures when clicking on the outputs for a workflow.

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -152,8 +152,8 @@ const WorkflowIOTable = ({ which, inputsOutputs: data, config, errors, onChange,
         cellRenderer: ({ rowIndex }) => {
           const io = sortedData[rowIndex]
           const { name, optional, inputType } = io
-          const value = config[which][name] || ''
-          const error = errors[which][name]
+          const value = _.get([which, name], config) || ''
```
kyuksel: The language now supports the following

```suggestion
          const value = config?.[which]?.[name] || ''
```

which may read more easily when you don't need to programmatically build the key path or need currying provided by `_.get`.

If you like this suggestion, please also update the line below accordingly.

Note that [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) also takes care of the error cases like Lodash does.

src/pages/workspaces/workspace/workflows/WorkflowView.js
kyuksel: Should we file a ticket to encapsulate this semantics in an error code returned by the relevant back-end service so we don't have to rely on message string parsing?

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -423,17 +423,35 @@ const WorkflowView = _.flow(
...
+        } else {
+          return false
+        }
       } else {
         throw e
       }
     }
   }
 
+  async safeGetSnapshotEntityMetadata(googleProject, modifiedConfig) {
```
kyuksel: Not sure if `safe` fits well necessarily here. How about

```suggestion
  async maybeGetSnapshotEntityMetadata(googleProject, modifiedConfig) {
```
?

src/pages/workspaces/workspace/workflows/WorkflowView.js
```diff
@@ -757,10 +782,10 @@ const WorkflowView = _.flow(
               ]),
               entitySelectionModel.type === processSnapshotTable ? div({ style: { margin: '2rem 0 0 2rem' } }, [
                 h(Select, {
-                  isDisabled: !!Utils.editWorkspaceError(ws),
+                  isDisabled: !!Utils.editWorkspaceError(ws) || !!snapshotReferenceError,
                   'aria-label': 'Snapshot table selector',
                   isClearable: false,
-                  value: modifiedConfig.dataReferenceName ? modifiedConfig.rootEntityType : undefined,
+                  value: modifiedConfig.dataReferenceName && !!!snapshotReferenceError ? modifiedConfig.rootEntityType : undefined,
```
kyuksel: Did you mean

```suggestion
                  value: modifiedConfig.dataReferenceName && !snapshotReferenceError ? modifiedConfig.rootEntityType : undefined,
```
?

kyuksel (commented):
Looks good overall. Minor comments inline.
Didn't test.

src/pages/workspaces/workspace/workflows/WorkflowView.js
davidangb: huh, I tried that and it complained about syntax. Lemme give that another shot, I agree it's nicer to read …

src/pages/workspaces/workspace/workflows/WorkflowView.js
davidangb: lol! How about `!!!!!`? I will change this :)

src/pages/workspaces/workspace/workflows/WorkflowView.js
davidangb: dunno what I did wrong the first time, optional chaining works as expected, as you suggested

src/pages/workspaces/workspace/workflows/WorkflowView.js
davidangb: filed https://broadworkbench.atlassian.net/browse/AJ-391!

davidangb:
@kyuksel back to you for re-review and thank you!

src/pages/workspaces/workspace/workflows/WorkflowView.js
kyuksel: The more the merrier 😂

src/pages/workspaces/workspace/workflows/WorkflowView.js
kyuksel: Thank you.

src/pages/workspaces/workspace/workflows/WorkflowView.js
kyuksel: Glad it worked. I feel l end up needing to look up the order of `?` and `.` every other time  😊

kyuksel (approved):

jordansafer (approved):
Agree with Kaan's comments, this looks like a good change to me, lgtm!

PR #2976, opened 2022-04-25
[SU-53] Consolidate billing account management options, per UX

MatthewBemis:
UX has provided a few small tweaks that they'd like us to implement for the billing project page. This is the first one, which consolidates the two icons for changing or removing a billing account into a three-dot dropdown menu.

See [SU-53](https://broadworkbench.atlassian.net/browse/SU-53) comments for additional details.

### Before
Screen Shot 2022-04-25 at 8 35 47 AM

### After
Screen Shot 2022-04-25 at 8 30 56 AM

Screen Shot 2022-04-25 at 8 31 01 AM

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -562,7 +562,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
             }, ['Remove Billing Account'])
           ])
         }, [
-          h(Link, { 'aria-label': 'Column menu' }, [
+          h(Link, { 'aria-label': 'Billing account menu' }, [
             icon('cardMenuIcon', { size: 16 })
           ])
         ]),
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -562,7 +562,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
             }, ['Remove Billing Account'])
           ])
         }, [
-          h(Link, { 'aria-label': 'Billing account menu' }, [
+          h(Link, { 'aria-label': 'Billing account menu', style: { display: 'flex', alignItems: 'center' } }, [
             icon('cardMenuIcon', { size: 16 })
           ])
         ]),
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -533,7 +533,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
       div({ style: { color: colors.dark(), fontSize: 14, display: 'flex', alignItems: 'center', marginTop: '0.5rem', marginLeft: '1rem' } }, [
         !!displayName && span({ style: { flexShrink: 0, fontWeight: 600, fontSize: 14, margin: '0 0.75rem 0 0' } }, 'Billing Account:'),
         !!displayName && span({ style: { flexShrink: 0, marginRight: '0.5rem' } }, displayName),
-        h(MenuTrigger, {
+        isOwner && h(MenuTrigger, {
           closeOnClick: true,
           side: 'bottom',
           style: { marginLeft: '0.5rem' },
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -563,7 +563,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
           ])
         }, [
           h(Link, { 'aria-label': 'Billing account menu', style: { display: 'flex', alignItems: 'center' } }, [
-            icon('cardMenuIcon', { size: 16 })
+            icon('cardMenuIcon', { size: 16, 'aria-haspopup': 'menu' })
           ])
         ]),
         showBillingModal && h(Modal, {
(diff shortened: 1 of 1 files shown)


cahrens:
Make sure that there is some accessible description on the menu.

MatthewBemis:
@cahrens for sure! That's on my list of things to test out before undrafting

kyuksel:
From inspecting the boxes around the new triple-dot menu, it looks like the icon is _not_ vertically centered whereas the text to the left of it _is_. Is this intentional?

image

src/pages/billing/Project.js
```diff
@@ -531,34 +532,40 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+              onClick: async () => {
+                if (Auth.hasBillingScope()) {
+                  setShowBillingRemovalModal(true)
+                } else {
+                  await authorizeAndLoadAccounts()
+                  setShowBillingRemovalModal(Auth.hasBillingScope())
+                }
+              },
+              // (CA-1586) For some reason the api sometimes returns string null, and sometimes returns no field, and sometimes returns null. This is just to be complete.
+              disabled: billingProject.billingAccount === 'null' || billingProject.billingAccount === undefined || billingProject.billingAccount === null
```
kyuksel: Could consolidate a bit. It's really unfortunate that the API can return `'null'`.
```suggestion
              disabled: billingProject.billingAccount === 'null' || _.isNil(billingProject.billingAccount)
```

kyuksel (commented):
Code looks good overall with one minor comment inline. 

I'm seeing some behaviour that was unexpected to me. Curious what you think. Note that it says 'No Access' and then I can click around on the triple-dot menu.

https://user-images.githubusercontent.com/5438223/165380151-697778e3-4187-4e8c-8c82-faedc8552707.mov

MatthewBemis:
@kyuksel thanks for catching that. That dropdown shouldn't be visible if you're a User on the billing project (which is the case in your video, so I'll fix that). But if you're an Owner, it should still be visible in the case of `No Access` because changing the billing account doesn't require access to the existing billing account (this is also in line with current behavior)

Fixed in https://github.com/DataBiosphere/terra-ui/pull/2976/commits/f5057655c39101dd09013b8f34e8a639637bae50

MatthewBemis:
@kyuksel Icon has been centered in https://github.com/DataBiosphere/terra-ui/pull/2976/commits/8d8b6c1df23e77e1f4ff64a1e88a0ab21cd217f9


Screen Shot 2022-04-26 at 4 24 43 PM

src/pages/billing/Project.js
```diff
@@ -562,7 +562,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
             }, ['Remove Billing Account'])
           ])
         }, [
-          h(Link, { 'aria-label': 'Billing account menu' }, [
+          h(Link, { 'aria-label': 'Billing account menu', style: { display: 'flex', alignItems: 'center' } }, [
```
kyuksel, line 565: I see multiple `div`s with `alignItems: 'center'` above, which makes me wonder if we could move up `alignItems: 'center'` to a parent node so we don't have to individually `style` children, which helps with maintainability.

src/pages/billing/Project.js
```diff
@@ -531,34 +532,40 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+                  setShowBillingRemovalModal(Auth.hasBillingScope())
+                }
+              },
+              // (CA-1586) For some reason the api sometimes returns string null, and sometimes returns no field, and sometimes returns null. This is just to be complete.
+              disabled: billingProject.billingAccount === 'null' || _.isNil(billingProject.billingAccount)
+            }, ['Remove Billing Account'])
+          ])
+        }, [
+          h(Link, { 'aria-label': 'Billing account menu', style: { display: 'flex', alignItems: 'center' } }, [
+            icon('cardMenuIcon', { size: 16 })
```
cahrens: I think you may want to add `'aria-haspopup': 'menu'`.

src/pages/billing/Project.js
kyuksel: I played around a bit but couldn't accomplish the simplification I'd hoped to do 🤷🏼

kyuksel (approved):
Tested by playing around the menu for a couple of billing accounts I'm on.

src/pages/billing/Project.js
MatthewBemis: Added in a69db71945c9285793a055a48f1cff00711640c3

src/pages/billing/Project.js
MatthewBemis: Yeah I had the same problem when I originally implemented it. Not sure what the issue is :(

src/pages/billing/Project.js
```diff
@@ -531,34 +532,40 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+              onClick: async () => {
+                if (Auth.hasBillingScope()) {
+                  setShowBillingRemovalModal(true)
+                } else {
+                  await authorizeAndLoadAccounts()
+                  setShowBillingRemovalModal(Auth.hasBillingScope())
+                }
+              },
+              // (CA-1586) For some reason the api sometimes returns string null, and sometimes returns no field, and sometimes returns null. This is just to be complete.
+              disabled: billingProject.billingAccount === 'null' || _.isNil(billingProject.billingAccount)
```
cahrens, line 561: _.isNill checks for null as well as undefined, so the first check is unnecessary. https://www.geeksforgeeks.org/lodash-_-isnil-method/

cahrens (approved):

PR #2977, opened 2022-04-25
[IA-3342] Bump 'jake' and 'async'

kyuksel:
- It turns out the [previous PR](https://github.com/DataBiosphere/terra-ui/pull/2972) to address the [dependabot alert for Prototype Pollution in async](https://github.com/DataBiosphere/terra-ui/security/dependabot/51) didn't address it fully because an older version of `async` remained in use due to `jake`'s dependency on that older version. Therefore this PR bumps `jake` as well.
- Neither `async` nor `jake` seems to be a user-facing library so I'm relying on integration testing for regression detection.
- `async` gets a major version bump but [the list of breaking changes](https://github.com/caolan/async/blob/master/CHANGELOG.md#breaking-changes) look benign.


dmohs (approved):
Neither of these two are even in our package.json. It'd be good if you could figure out what is bringing them in to our project as a transient dependency, but no big deal.

kyuksel:
@dmohs `async` and `jake` are brought in by our direct dependencies of `source-map-explorer` and `react-scripts`:


image

MatthewBemis (approved):

PR #2979, opened 2022-04-26
[DC-312] Make the preview code point to data catalog, rather than data repo

s-rubenstein:
<!--
Hello, friend!
Thanks!

--!>


src/pages/library/DataBrowserDetails.js
```diff
@@ -147,8 +147,9 @@ export const SidebarComponent = ({ dataObj, id }) => {
                   Ajax().Metrics.captureEvent(`${Events.catalogRequestAccess}:popUp`, {
                     // These are still using snapshot as a relic to ensure backwards search
                     // capabilities within the data browser.
-                    snapshotId: _.get('dct:identifier', dataObj),
-                    snapshotName: dataObj['dct:title']
+                    id: dataObj.id,
+                    // TODO: Make point at a data catalog title, rather than the underlying storage system
+                    title: dataObj['dct:title']
```
pshapiro4broad: I think this is correct. It's possible that `dct:title` is the same as a snapshot title, but it's not guaranteed to be so. When a data owner creates the catalog entry, they get to choose what `dct:title` contains. So I think `dct:title` is correct and don't need a TODO.

pshapiro4broad (approved):
Looks good, I like the renamings done here as well.

One overall note, I think the comments about dct:title aren't necessary, maybe a chat about the catalog entry metadata would help to clear this up.

src/pages/library/DataBrowserDetails.js
s-rubenstein: What _is_ `dct:title`? My understanding was that it was the underlying data storage systems title, which I would expect us to not want to use, as opposed to something stored in our metadata (which might derive from `dct:title`?

kyuksel:
@s-rubenstein Please include a Jira ticket reference in the PR title for compliance reasons.

src/pages/library/DataBrowserDetails.js
pshapiro4broad: We don't yet have a specification for exactly what's in a catalog entry and what each attribute means, so there's no way you could know what this is. A `dct:title` and similarly named items are part of the TIM representation of a dataset, which is specified [here](https://github.com/DataBiosphere/terra-interoperability-model/blob/master/releases/1.x/terra-core/TerraDCAT-AP.ttl#L167). The attribute `dct:title` isn't directly specified on `TerraDCAT_ap:Dataset` but is inherited from `dct:Resource` (via `dcat:Dataset`), see https://www.w3.org/TR/vocab-dcat-2/#Property:resource_title.

In theory the UI code should only be relying on properties from `TerraDCAT_ap:Dataset`. All terms in that form are part of the catalog's unified representation and are not storage system specific.

petesantos (approved):
Looks good, manually tested the changes

src/pages/library/DataBrowserDetails.js
pshapiro4broad: To be clear: a catalog entry should be a representation of a `TerraDCAT_ap:Dataset`. The catalog service itself doesn't yet do anything to enforce this, yet.

kyuksel:
FYI @s-rubenstein that I took the liberty to update your branch with `dev` to pick up test fixes and now your PR is ready to merge :)

PR #2980, opened 2022-04-26
[IA-3354] fix issue where creating hail would cause size: undefined

jdcanas:
We didn't previously test creating a hail runtime when no runtime exists for previous fix, we were more focused on persistent disk issue.

I tested:
- Create a hail runtime in a new workspace
- Creating a gce runtime in a workspace with a hail runtime and ensuring a new disk is created for it
- Creating a runtime (ensuring a new disk is created), deleting it but keeping the disk, and ensuring the disk gets attached to the new runtime
- Creating a runtime, changing the image, and ensuring the disk gets attached to the new runtime

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -308,15 +308,22 @@ export const ComputeModalBase = ({
         })
       }
       if (shouldCreateRuntime) {
-        const diskConfig = existingPersistentDisk && !shouldDeletePersistentDisk ? {
-          name: currentPersistentDiskDetails.name
-        } : {
-          name: Utils.generatePersistentDiskName(),
-          size: desiredPersistentDisk.size,
-          diskType: desiredPersistentDisk.diskType.label,
-          labels: { saturnWorkspaceNamespace: namespace, saturnWorkspaceName: name }
-        }
-        const createRuntimeConfig = { ...runtimeConfig, persistentDisk: diskConfig }
+        const diskConfig = Utils.cond(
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):

cahrens (approved):

src/components/ComputeModal.js
```diff
@@ -308,15 +308,22 @@ export const ComputeModalBase = ({
...
-        } : {
-          name: Utils.generatePersistentDiskName(),
-          size: desiredPersistentDisk.size,
-          diskType: desiredPersistentDisk.diskType.label,
-          labels: { saturnWorkspaceNamespace: namespace, saturnWorkspaceName: name }
-        }
-        const createRuntimeConfig = { ...runtimeConfig, persistentDisk: diskConfig }
+        const diskConfig = Utils.cond(
+          [desiredRuntime.cloudService === cloudServices.DATAPROC, () => { return {} }],
+          [existingPersistentDisk && !shouldDeletePersistentDisk, () => { return { persistentDisk: { name: currentPersistentDiskDetails.name } } }],
```
breilly2, line 313: For short arrow functions like this, you can avoid the and `return` by also removing the surrounding `{ }`. When what you want to return is an object literal, put `( )` around it to keep it from being treated as a block statement. I.e.:
```suggestion
          [desiredRuntime.cloudService === cloudServices.DATAPROC, () => ({})],
          [existingPersistentDisk && !shouldDeletePersistentDisk, () => ({ persistentDisk: { name: currentPersistentDiskDetails.name } }) ],
```
Same with the `DEFAULT` case below.

PR #2981, opened 2022-04-26
[IA-3224] Tab Navigation in Cloud Env Modal

cstrandw:
There was an extra, non-visible close button (circled in red in the image below) that was being focused on when a user tried to tab through the modal. I fixed this by changing the tabindex value of the close on the TitleBar component depending on whether the passed in hideCloseButton property. If I did not render the element at all, then the spacing of the arrow on the modal would be affected.

Screen Shot 2022-04-26 at 2 43 30 PM

Tested by going through the modal to make sure that the close was not able to be tabbed to if it was not visible and was able to be tabbed to if it was.

I also included a small fix to the Environments page, as I noticed that the total disk price was not being displayed properly on the page.


src/components/TitleBar.js
```diff
@@ -22,6 +22,7 @@ const TitleBar = ({ id, onPrevious, title, onDismiss, titleExtras, style = {}, t
       onDismiss && h(Link, {
         'aria-label': 'Close',
         style: { marginLeft: '2rem' },
+        tabIndex: hideCloseButton ? -1 : 0,
```
cahrens, line 25: It really would be better to not render this component at all. If you are going to render it, you need to mark it `aria-disabled` when `hideCloseButton` is true so that the screen reader does not traverse it in the dom (I verified that I could still navigate to the close button and trigger it with the screen reader).

cahrens:
It would still be nice to figure out the focus when going from one screen to the next in the wizard (I know you already spent some time on it).

cstrandw:
> It would still be nice to figure out the focus when going from one screen to the next in the wizard (I know you already spent some time on it).

Yeah, the IA team and I discussed making another ticket to try and figure that out in the future, but other work has come up that is more pressing

src/components/TitleBar.js
```diff
@@ -22,6 +22,8 @@ const TitleBar = ({ id, onPrevious, title, onDismiss, titleExtras, style = {}, t
...
+        'aria-disabled': !hideCloseButton,
```
cahrens: This is the opposite of what you want. When the closeButton is hidden, `aria-disabled` should be `true`.

cahrens (approved):

Qi77Qi (approved):

PR #2987, opened 2022-04-27
[IA-3260][IA-3265] IA Azure UI kickoff

jdcanas:
This is a first pass at adding azure support for analysis components, with a lot of TODOs and future tickets to come.

So testing for this is a bit all over the place. There are various variables I had to hardcode to trick the pages into thinking the workspace is an azure workspace since rawls support for MC workspaces isn't in dev yet, which I'll leave for reviewer convenience until this is merged. Additionally, I had to manually create the workspace as described below. You also can't use the workspace I did since there's no easy way (I know of) to add people to an MC workspace.

1. `gcloud auth login` on the terminal as the user I use in the UI
2. `gcloud auth print-access-token` to get a token
3. call createWorkspace on workspace manager dev with the token from #2, which can be done here https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/Workspace/createWorkspace. You need to change the ID to something unique, and set `"stage": "MC_WORKSPACE"`. Write the id down somewhere, it will be needed.
4. create an azure cloud context in WSM here https://workspace.dsde-dev.broadinstitute.org/swagger-ui.html#/Workspace/createCloudContext with the payload below
```
{
  "cloudPlatform": "AZURE",
  "jobControl": {
    "id": "uniqueid1"
  },
  "azureContext": {
    "tenantId": "fad90753-2022-4456-9b0a-c7e5b934e408",
    "subscriptionId": "3efc5bdf-be0e-44e7-b1d7-c08931e3c16c",
    "resourceGroupId": "mrg-terra-integration-test-20211118"
  }
}
```
6. Go to line 215 in WorkspaceContainer.js and set workspaceId to the workspace you created.

Also note several behavior is not yet developed for azure. Rather than stubbing a bunch of things, I just leave it in an undefined (and non-erroring) state since this will not be user facing any time soon.

Additionally, runtime creation is currently bugged in the backend due to some unrelated changes, so all runtimes created will result in errored runtimes. However, the create call itself is accepted by the backend and succeeds. It also errors in such a way that the GET call 404s after. As such, I have provided verification below that the individual calls succeed with the proper backend state.

The base analysis page in azure mode:
image

The start button modal in azure mode:
image

The get azure call which is populating the modal when I have an existing runtime succeeding:
image

The delete call succeeding
image

The create call succeeding:
image

The default create payload succeeding
image

Create payload with all fields changed in modal:
image

Interactions/display for azure also works in the context bar. Everything but creating an azure runtime is grayed out when accessing the context bar in a valid workspace. Additionally, the runtime status/error is properly display in the side bar/details modal respectively.

### DIFF
--- a/src/components/ContextBar.js
+++ b/src/components/ContextBar.js
@@ -126,9 +126,7 @@ export const ContextBar = ({
             style: { flexDirection: 'column', justifyContent: 'center', padding: '.75rem', ...contextBarStyles.contextBarButton, borderBottom: '0px' },
             hover: contextBarStyles.hover,
             tooltipSide: 'left',
-            onClick: () => {
-              setCloudEnvOpen(!isCloudEnvOpen)
-            },
+            onClick: () => setCloudEnvOpen(true),
             tooltip: 'Environment Configuration',
             tooltipDelay: 100,
             useTooltipAsLabel: true
--- a/src/components/ContextBar.js
+++ b/src/components/ContextBar.js
@@ -39,7 +39,7 @@ const contextBarStyles = {
 
 export const ContextBar = ({
   runtimes, apps, appDataDisks, refreshRuntimes, location, locationType, refreshApps,
-  workspace, persistentDisks, workspace: { workspace: { namespace, name: workspaceName } }
+  workspace, persistentDisks, workspace: { workspace: { namespace, name: workspaceName, googleProject } }
 }) => {
   const [isCloudEnvOpen, setCloudEnvOpen] = useState(false)
   const [selectedToolIcon, setSelectedToolIcon] = useState(undefined)
@@ -59,7 +59,8 @@ export const ContextBar = ({
     [tools.Jupyter.label, () => img({ src: jupyterLogo, style: { height: 45, width: 45 }, alt: '' })],
     [tools.Galaxy.label, () => img({ src: galaxyLogo, style: { height: 14, width: 45 }, alt: '' })],
     [tools.Cromwell.label, () => img({ src: cromwellImg, style: { width: 45 }, alt: '' })],
-    [tools.RStudio.label, () => img({ src: rstudioSquareLogo, style: { height: 45, width: 45 }, alt: '' })]
+    [tools.RStudio.label, () => img({ src: rstudioSquareLogo, style: { height: 45, width: 45 }, alt: '' })],
+    [tools.Azure.label, () => img({ src: jupyterLogo, style: { height: 45, width: 45 }, alt: '' })]
   )
 
   const getColorForStatus = status => Utils.cond(
@@ -125,7 +126,12 @@ export const ContextBar = ({
             style: { flexDirection: 'column', justifyContent: 'center', padding: '.75rem', ...contextBarStyles.contextBarButton, borderBottom: '0px' },
             hover: contextBarStyles.hover,
             tooltipSide: 'left',
-            onClick: () => setCloudEnvOpen(!isCloudEnvOpen),
+            onClick: () => {
+              setCloudEnvOpen(!isCloudEnvOpen)
+              if (!googleProject) {
+                setSelectedToolIcon(tools.Azure.label)
+              }
+            },
             tooltip: 'Environment Configuration',
             tooltipDelay: 100,
             useTooltipAsLabel: true
--- a/src/components/ContextBar.js
+++ b/src/components/ContextBar.js
@@ -39,7 +39,7 @@ const contextBarStyles = {
 
 export const ContextBar = ({
   runtimes, apps, appDataDisks, refreshRuntimes, location, locationType, refreshApps,
-  workspace, persistentDisks, workspace: { workspace: { namespace, name: workspaceName, googleProject } }
+  workspace, persistentDisks, workspace: { workspace: { namespace, name: workspaceName } }
 }) => {
   const [isCloudEnvOpen, setCloudEnvOpen] = useState(false)
   const [selectedToolIcon, setSelectedToolIcon] = useState(undefined)
@@ -128,9 +128,6 @@ export const ContextBar = ({
             tooltipSide: 'left',
             onClick: () => {
               setCloudEnvOpen(!isCloudEnvOpen)
-              if (!googleProject) {
-                setSelectedToolIcon(tools.Azure.label)
-              }
             },
             tooltip: 'Environment Configuration',
             tooltipDelay: 100,
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -24,7 +24,8 @@ import * as Nav from 'src/libs/nav'
 import { clearNotification, notify } from 'src/libs/notifications'
 import { useOnMount } from 'src/libs/react-utils'
 import {
-  appIsSettingUp, getComputeStatusForDisplay, getConvertedRuntimeStatus, getCurrentApp, getCurrentRuntime, getPersistentDiskCostHourly, runtimeCost,
+  appIsSettingUp, cloudProviders, getComputeStatusForDisplay, getConvertedRuntimeStatus, getCurrentApp, getCurrentRuntime,
+  getPersistentDiskCostHourly, runtimeCost,
   trimRuntimesOldestFirst
 } from 'src/libs/runtime-utils'
 import { errorNotifiedApps, errorNotifiedRuntimes } from 'src/libs/state'
@@ -72,7 +73,9 @@ export const RuntimeErrorModal = ({ runtime, onDismiss }) => {
     withErrorReporting('Error loading cloud environment details'),
     Utils.withBusyState(setLoadingRuntimeDetails)
   )(async () => {
-    const { errors: runtimeErrors } = await Ajax().Runtimes.runtime(runtime.googleProject, runtime.runtimeName).details()
+    console.log('in loadRuntimeError', runtime, _.lowerCase(runtime.cloudContext.cloudProvider))
+    const { errors: runtimeErrors } = _.lowerCase(runtime.cloudContext.cloudProvider) === cloudProviders.azure.label ? await Ajax().Runtimes.runtimeV2(runtime.workspaceId, runtime.runtimeName).details() :
+      await Ajax().Runtimes.runtime(runtime.googleProject, runtime.runtimeName).details()
     if (_.some(({ errorMessage }) => errorMessage.includes('Userscript failed'), runtimeErrors)) {
       setError(
         await Ajax()
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -288,7 +288,7 @@ export default class RuntimeManager extends PureComponent {
     const { namespace, name, runtimes, refreshRuntimes, canCompute, persistentDisks, apps, appDataDisks, refreshApps, location, locationType, workspace } = this.props
     const { busy, createModalDrawerOpen, errorModalOpen, galaxyDrawerOpen } = this.state
     const { computeRegion } = getRegionInfo(location, locationType)
-    if (!runtimes || !apps || !workspace?.workspace.googleProject) {
+    if (!runtimes || !apps) {
       return null
     }
     const currentRuntime = this.getCurrentRuntime()
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -70,7 +70,7 @@ export const RuntimeErrorModal = ({ runtime, onDismiss }) => {
   const [loadingRuntimeDetails, setLoadingRuntimeDetails] = useState(false)
 
   const loadRuntimeError = _.flow(
-    withErrorReporting('Could Not Retrieve Cloud Environment Log Info'),
+    withErrorReporting('Could Not Retrieve Cloud Environment Error Info'),
     Utils.withBusyState(setLoadingRuntimeDetails)
   )(async () => {
     const { errors: runtimeErrors } = _.lowerCase(runtime.cloudContext.cloudProvider) === cloudProviders.azure.label ?
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -70,10 +70,11 @@ export const RuntimeErrorModal = ({ runtime, onDismiss }) => {
   const [loadingRuntimeDetails, setLoadingRuntimeDetails] = useState(false)
 
   const loadRuntimeError = _.flow(
-    withErrorReporting('Error loading cloud environment details'),
+    withErrorReporting('Could Not Retrieve Cloud Environment Log Info'),
     Utils.withBusyState(setLoadingRuntimeDetails)
   )(async () => {
-    const { errors: runtimeErrors } = _.lowerCase(runtime.cloudContext.cloudProvider) === cloudProviders.azure.label ? await Ajax().Runtimes.runtimeV2(runtime.workspaceId, runtime.runtimeName).details() :
+    const { errors: runtimeErrors } = _.lowerCase(runtime.cloudContext.cloudProvider) === cloudProviders.azure.label ?
+      await Ajax().Runtimes.runtimeV2(runtime.workspaceId, runtime.runtimeName).details() :
       await Ajax().Runtimes.runtime(runtime.googleProject, runtime.runtimeName).details()
     if (_.some(({ errorMessage }) => errorMessage.includes('Userscript failed'), runtimeErrors)) {
       setError(
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -73,7 +73,6 @@ export const RuntimeErrorModal = ({ runtime, onDismiss }) => {
     withErrorReporting('Error loading cloud environment details'),
     Utils.withBusyState(setLoadingRuntimeDetails)
   )(async () => {
-    console.log('in loadRuntimeError', runtime, _.lowerCase(runtime.cloudContext.cloudProvider))
     const { errors: runtimeErrors } = _.lowerCase(runtime.cloudContext.cloudProvider) === cloudProviders.azure.label ? await Ajax().Runtimes.runtimeV2(runtime.workspaceId, runtime.runtimeName).details() :
       await Ajax().Runtimes.runtime(runtime.googleProject, runtime.runtimeName).details()
     if (_.some(({ errorMessage }) => errorMessage.includes('Userscript failed'), runtimeErrors)) {
@@ -289,7 +288,7 @@ export default class RuntimeManager extends PureComponent {
     const { namespace, name, runtimes, refreshRuntimes, canCompute, persistentDisks, apps, appDataDisks, refreshApps, location, locationType, workspace } = this.props
     const { busy, createModalDrawerOpen, errorModalOpen, galaxyDrawerOpen } = this.state
     const { computeRegion } = getRegionInfo(location, locationType)
-    if (!runtimes || !apps) {
+    if (!runtimes || !apps || !workspace?.workspace.googleProject) {
       return null
     }
     const currentRuntime = this.getCurrentRuntime()
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -77,7 +77,8 @@ export const tools = {
   jupyterTerminal: { label: 'terminal' },
   spark: { label: 'spark' },
   Galaxy: { label: 'Galaxy', appType: 'GALAXY' },
-  Cromwell: { label: 'Cromwell', appType: 'CROMWELL', isAppHidden: !isCromwellAppVisible(), isPauseUnsupported: true }
+  Cromwell: { label: 'Cromwell', appType: 'CROMWELL', isAppHidden: !isCromwellAppVisible(), isPauseUnsupported: true },
+  Azure: { label: 'Azure' }
 }
 
 // Returns the tools in the order that they should be displayed for Cloud Environment tools
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -84,9 +84,8 @@ export const tools = {
 // Returns the tools in the order that they should be displayed for Cloud Environment tools
 export const getToolsToDisplay = isAzureWorkspace => _.flow(
   _.remove(tool => tool.isAppHidden),
-  isAzureWorkspace ? _.remove(tool => !tool.isAzureCompatible) : _.remove(tool => tool.isAzureCompatible))(
-  [tools.Jupyter, tools.RStudio, tools.Galaxy, tools.Cromwell, tools.Azure]
-)
+  _.filter(tool => tool.isAzureCompatible === isAzureWorkspace)
+)([tools.Jupyter, tools.RStudio, tools.Galaxy, tools.Cromwell, tools.Azure])
 
 const toolToExtensionMap = { [tools.RStudio.label]: tools.RStudio.ext, [tools.Jupyter.label]: tools.Jupyter.ext }
 
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -84,7 +84,7 @@ export const tools = {
 // Returns the tools in the order that they should be displayed for Cloud Environment tools
 export const getToolsToDisplay = isAzureWorkspace => _.flow(
   _.remove(tool => tool.isAppHidden),
-  _.filter(tool => tool.isAzureCompatible === isAzureWorkspace)
+  _.filter(tool => !!tool.isAzureCompatible === !!isAzureWorkspace)
 )([tools.Jupyter, tools.RStudio, tools.Galaxy, tools.Cromwell, tools.Azure])
 
 const toolToExtensionMap = { [tools.RStudio.label]: tools.RStudio.ext, [tools.Jupyter.label]: tools.Jupyter.ext }
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -73,16 +73,20 @@ export const analysisNameInput = ({ inputProps, ...props }) => h(ValidatedInput,
 
 export const tools = {
   RStudio: { label: 'RStudio', ext: 'Rmd', imageIds: ['RStudio'], defaultImageId: 'RStudio' },
-  Jupyter: { label: 'Jupyter', ext: 'ipynb', imageIds: ['terra-jupyter-bioconductor', 'terra-jupyter-bioconductor_legacy', 'terra-jupyter-hail', 'terra-jupyter-python', 'terra-jupyter-gatk', 'Pegasus', 'terra-jupyter-gatk_legacy'], defaultImageId: 'terra-jupyter-gatk' },
+  Jupyter: { label: 'Jupyter', ext: 'ipynb', imageIds: ['terra-jupyter-bioconductor', 'terra-jupyter-bioconductor_legacy', 'terra-jupyter-hail', 'terra-jupyter-python', 'terra-jupyter-gatk', 'Pegasus', 'terra-jupyter-gatk_legacy'], defaultImageId: 'terra-jupyter-gatk', isLaunchUnsupported: false },
   jupyterTerminal: { label: 'terminal' },
   spark: { label: 'spark' },
   Galaxy: { label: 'Galaxy', appType: 'GALAXY' },
   Cromwell: { label: 'Cromwell', appType: 'CROMWELL', isAppHidden: !isCromwellAppVisible(), isPauseUnsupported: true },
-  Azure: { label: 'Azure' }
+  Azure: { label: 'Azure', isAzureCompatible: true, isPauseUnsupported: true, isLaunchUnsupported: true }
 }
 
 // Returns the tools in the order that they should be displayed for Cloud Environment tools
-export const getToolsToDisplay = _.remove(tool => tool.isAppHidden)([tools.Jupyter, tools.RStudio, tools.Galaxy, tools.Cromwell])
+export const getToolsToDisplay = isAzureWorkspace => _.flow(
+  _.remove(tool => tool.isAppHidden),
+  isAzureWorkspace ? _.remove(tool => !tool.isAzureCompatible) : _.remove(tool => tool.isAzureCompatible))(
+  [tools.Jupyter, tools.RStudio, tools.Galaxy, tools.Cromwell, tools.Azure]
+)
 
 const toolToExtensionMap = { [tools.RStudio.label]: tools.RStudio.ext, [tools.Jupyter.label]: tools.Jupyter.ext }
 
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1487,6 +1487,48 @@ const Runtimes = signal => ({
     }
   },
 
+  listV2: async (labels = {}) => {
+    const res = await fetchLeo(`api/v2/runtimes?${qs.stringify({ saturnAutoCreated: true, ...labels })}`,
+      _.mergeAll([authOpts(), appIdentifier, { signal }]))
+    return res.json()
+  },
+
+  listV2WithWorkspace: async (workspaceId, labels = {}) => {
+    const res = await fetchLeo(`api/v2/runtimes/${workspaceId}?${qs.stringify({ saturnAutoCreated: true, ...labels })}`,
+      _.mergeAll([authOpts(), appIdentifier, { signal }]))
+    return res.json()
+  },
+
+  listV2AzureWithWorkspace: async (workspaceId, labels = {}) => {
+    const res = await fetchLeo(`api/v2/runtimes/${workspaceId}/azure?${qs.stringify({ saturnAutoCreated: true, ...labels })}`,
+      _.mergeAll([authOpts(), appIdentifier, { signal }]))
+    return res.json()
+  },
+
+  runtimeV2: (workspaceId, name, cloudProvider = "azure") => {
+    const root = `api/v2/runtimes/${workspaceId}/${cloudProvider}/${name}`
+
+    return {
+
+      details: async () => {
+        const res = await fetchLeo(root, _.mergeAll([authOpts(), { signal }, appIdentifier]))
+        return res.json()
+      },
+
+      create: options => {
+        const body = _.merge(options, {
+          labels: { saturnAutoCreated: 'true', saturnVersion: version },
+        })
+        return fetchLeo(root, _.mergeAll([authOpts(), jsonBody(body), { signal, method: 'POST' }, appIdentifier]))
+      },
+
+      delete: (deleteDisk = true) => {
+        return fetchLeo(`${root}${qs.stringify({ deleteDisk }, { addQueryPrefix: true })}`,
+          _.mergeAll([authOpts(), { signal, method: 'DELETE' }, appIdentifier]))
+      }
+    }
+  },
+
   fileSyncing: (project, name) => {
     const root = `proxy/${project}/${name}`
 
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1505,7 +1505,7 @@ const Runtimes = signal => ({
     return res.json()
   },
 
-  runtimeV2: (workspaceId, name, cloudProvider = "azure") => {
+  runtimeV2: (workspaceId, name, cloudProvider = 'azure') => {
     const root = `api/v2/runtimes/${workspaceId}/${cloudProvider}/${name}`
 
     return {
@@ -1517,7 +1517,7 @@ const Runtimes = signal => ({
 
       create: options => {
         const body = _.merge(options, {
-          labels: { saturnAutoCreated: 'true', saturnVersion: version },
+          labels: { saturnAutoCreated: 'true', saturnVersion: version }
         })
         return fetchLeo(root, _.mergeAll([authOpts(), jsonBody(body), { signal, method: 'POST' }, appIdentifier]))
       },
--- a/src/libs/azure-utils.js
+++ b/src/libs/azure-utils.js
@@ -13,8 +13,8 @@ export const defaultAzureComputeConfig = {
 
 //TODO: this should be fleshed out once azure region isn't limited to workspace region
 export const azureRegions = { eastus: { label: 'East US' } }
-export const getRegionLabel = label => _.has(label, azureRegions) ? azureRegions[label].label : 'Unknown azure region'
+export const getRegionLabel = key => _.has(key, azureRegions) ? azureRegions[key].label : 'Unknown azure region'
 
 export const azureMachineTypes = { Standard_DS1_v2: { cpu: 1, ramInGb: 3.5 }, Standard_DS2_v2: { cpu: 2, ramInGb: 7 }, Standard_DS3_v2: { cpu: 4, ramInGb: 14 }, Standard_DS4_v2: { cpu: 8, ramInGb: 28 }, Standard_DS5_v2: { cpu: 16, ramInGb: 56 } }
-export const getMachineTypeLabel = label => _.has(label, azureMachineTypes) ? `${label}, ${azureMachineTypes[label].cpu} CPU(s), ${azureMachineTypes[label].ramInGb} GBs` : 'Unknown machine type'
+export const getMachineTypeLabel = key => _.has(key, azureMachineTypes) ? `${key}, ${azureMachineTypes[key].cpu} CPU(s), ${azureMachineTypes[key].ramInGb} GBs` : 'Unknown machine type'
 
--- a/src/libs/azure-utils.js
+++ b/src/libs/azure-utils.js
@@ -1,0 +1,11 @@
+import * as Utils from 'src/libs/utils'
+import _ from 'lodash/fp'
+
+export const defaultAzureMachineType = 'Standard_DS1_v2'
+export const defaultAzureDiskSize = 50
+export const defaultAzureRegion = 'eastus'
+//TODO: this should be fleshed out once azure region isn't limitted to workspace region
+export const azureRegions =  { 'eastus': { label: 'East US' } }
+export const azureMachineTypes = { Standard_DS1_v2: { cpu: 1, ramInGb: 7.75 } }
+export const getMachineTypeLabel = label => _.has(label, azureMachineTypes) ? `${label}, ${azureMachineTypes[label].cpu} CPU(s), ${azureMachineTypes[label].ramInGb} GBs` : 'Unknown machine type'
+export const getRegionLabel = label => _.has(label, azureRegions) ? azureRegions[label].label : 'Unknown azure region'
--- a/src/libs/azure-utils.js
+++ b/src/libs/azure-utils.js
@@ -1,12 +1,20 @@
 import _ from 'lodash/fp'
-import * as Utils from 'src/libs/utils'
 
 
 export const defaultAzureMachineType = 'Standard_DS1_v2'
 export const defaultAzureDiskSize = 50
 export const defaultAzureRegion = 'eastus'
-//TODO: this should be fleshed out once azure region isn't limitted to workspace region
+
+export const defaultAzureComputeConfig = {
+  machineType: defaultAzureMachineType,
+  diskSize: defaultAzureDiskSize,
+  region: defaultAzureRegion
+}
+
+//TODO: this should be fleshed out once azure region isn't limited to workspace region
 export const azureRegions = { eastus: { label: 'East US' } }
+export const getRegionLabel = label => _.has(label, azureRegions) ? azureRegions[label].label : 'Unknown azure region'
+
 export const azureMachineTypes = { Standard_DS1_v2: { cpu: 1, ramInGb: 3.5 }, Standard_DS2_v2: { cpu: 2, ramInGb: 7 }, Standard_DS3_v2: { cpu: 4, ramInGb: 14 }, Standard_DS4_v2: { cpu: 8, ramInGb: 28 }, Standard_DS5_v2: { cpu: 16, ramInGb: 56 } }
 export const getMachineTypeLabel = label => _.has(label, azureMachineTypes) ? `${label}, ${azureMachineTypes[label].cpu} CPU(s), ${azureMachineTypes[label].ramInGb} GBs` : 'Unknown machine type'
-export const getRegionLabel = label => _.has(label, azureRegions) ? azureRegions[label].label : 'Unknown azure region'
+
--- a/src/libs/azure-utils.js
+++ b/src/libs/azure-utils.js
@@ -1,11 +1,12 @@
-import * as Utils from 'src/libs/utils'
 import _ from 'lodash/fp'
+import * as Utils from 'src/libs/utils'
+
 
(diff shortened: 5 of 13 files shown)


cahrens:
I have Azure workspaces so can help with testing this out.

src/libs/ajax.js
```diff
@@ -1489,6 +1489,48 @@ const Runtimes = signal => ({
...
+        return fetchLeo(root, _.mergeAll([authOpts(), jsonBody(body), { signal, method: 'POST' }, appIdentifier]))
+      },
+
+      delete: (deleteDisk = true) => {
+        return fetchLeo(`${root}${qs.stringify({ deleteDisk }, { addQueryPrefix: true })}`,
+          _.mergeAll([authOpts(), { signal, method: 'DELETE' }, appIdentifier]))
+      }
+    }
+  },
+
```
dmohs, line 1531: I've long been concerned with the amount of boilerplate in this file, which contributes to its excessive length. Consider something like the following to tighten things up and shed light on the outlying cases:
```suggestion
  v2: (() => {
    const getStdFetchOpts = () => _.mergeAll([authOpts(), appIdentifier, { signal }])
    const fetchLeoV2Json = (p, q) =>
      fetchLeo('api/v2/' + p + q ? ('?'+qs.stringify(q)) : '',
        getStdFetchOpts()).then(res => res.json())
    return {
      list: (labels = {}) =>
        fetchLeoV2Json(`runtimes`, { saturnAutoCreated: true, ...labels }),
      listWithWorkspace: (workspaceId, labels = {}) =>
        fetchLeoV2Json(`runtimes/${workspaceId}`, { saturnAutoCreated: true, ...labels }),
      listAzureWithWorkspace: async (workspaceId, labels = {}) =>
        fetchLeoV2Json(`runtimes/${workspaceId}/azure`, { saturnAutoCreated: true, ...labels }),
      runtime: (workspaceId, name, cloudProvider = 'azure') => {
        const root = `runtimes/${workspaceId}/${cloudProvider}/${name}`
        return {
          details: () => fetchLeoV2Json(root),
          create: options => {
            const body = _.merge(options, {
              labels: { saturnAutoCreated: 'true', saturnVersion: version }
            })
            return fetchLeo('api/v2/' + root,
              _.mergeAll([authOpts(), jsonBody(body), { signal, method: 'POST' }, appIdentifier]))
          },
          delete: (deleteDisk = true) => {
            return fetchLeo(`api/v2/${root}`
              +`?${qs.stringify({ deleteDisk }, { addQueryPrefix: true })}`,
              _.mergeAll([authOpts(), { signal, method: 'DELETE' }, appIdentifier]))
          }
        }
      }
    }
  })()
```
I don't think I got the abstractions quite right there, but even so I do think it is a bit easier to parse.

src/libs/runtime-utils.js
```diff
@@ -65,6 +65,7 @@ export const defaultComputeRegion = 'US-CENTRAL1'
 
 export const defaultAutopauseThreshold = 30
 export const autopauseDisabledValue = 0
+
 // Leonardo considers autopause disabled when the threshold is set to 0
```
dmohs: Since you touched this line, I judge the comment belongs above the `const`. Feel free to revert if unintentional.
```suggestion
// Leonardo considers autopause disabled when the threshold is set to 0
export const autopauseDisabledValue = 0
```

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+  //TODO: revert before merging
+  var workspaceId = '1aa85f64-5717-4562-b3fc-2c963f66afa6'
+
+  const [computeConfig, setComputeConfig] = useState(defaultAzureComputeConfig)
+  const updateComputeConfig = (key, value) => setComputeConfig(_.set(key, value, computeConfig))
+
+  // Lifecycle
+  useOnMount(() => {
+    // Can't pass an async function into useEffect so we define the function in the body and then call it
+    const doUseOnMount = _.flow(
```
dmohs: I don't think pulling this out into a const is buying us much. Maybe lose it?
```suggestion
    _.flow(
```

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+      setCurrentRuntimeDetails(currentRuntime ? await Ajax().Runtimes.runtimeV2(workspaceId, currentRuntime.runtimeName).details() : null)
+
+      setComputeConfig({
+        machineType: currentRuntimeDetails?.runtimeConfig?.machineType || defaultAzureMachineType,
+        diskSize: currentRuntimeDetails?.diskConfig?.size || defaultAzureDiskSize,
+        region: currentRuntime?.runtimeConfig?.region || defaultAzureRegion
+      })
+    })
+
+    doUseOnMount()
```
dmohs: Without the const:
```suggestion
    })()
```

cahrens:
The default size of the "Type" column needs to be increased if we are going to show AZURE_VM. It would really be nice though if the type was more informative (Jupyter on Azure).

image

cahrens:
When I click see log info, I just get another error. It is awkward to get to the details.

image

image

cahrens:
I'm having issues with creating Jupyter in a GCE workspace (I have removed the overrides in Analyses, WorkspaceContainer, and AzureComputeModal, but perhaps I missed something). It doesn't seem to be recognizing when runtimes already exist.

Could you push up a branch where all the overriding is removed since I have Azure workspaces to test with? That will also allow the integration tests to run. Thanks!

image

src/components/ContextBar.js
```diff
@@ -39,7 +39,7 @@ const contextBarStyles = {
 
 export const ContextBar = ({
   runtimes, apps, appDataDisks, refreshRuntimes, location, locationType, refreshApps,
-  workspace, persistentDisks, workspace: { workspace: { namespace, name: workspaceName } }
+  workspace, persistentDisks, workspace: { workspace: { namespace, name: workspaceName, googleProject } }
```
cahrens: There is a choice here-- you could pass azureContext and use that to decide if you are on Azure. To me that seems somewhat more straightforward, and clarifies that you have no dependency on the workspace's `googleProject`.

Note that you do get `googleProject` from the currentRuntime, so I'm not clear what will happen with that code once Jupyter actually starts up.

src/components/ContextBar.js
```diff
@@ -59,7 +59,8 @@ export const ContextBar = ({
     [tools.Jupyter.label, () => img({ src: jupyterLogo, style: { height: 45, width: 45 }, alt: '' })],
     [tools.Galaxy.label, () => img({ src: galaxyLogo, style: { height: 14, width: 45 }, alt: '' })],
     [tools.Cromwell.label, () => img({ src: cromwellImg, style: { width: 45 }, alt: '' })],
-    [tools.RStudio.label, () => img({ src: rstudioSquareLogo, style: { height: 45, width: 45 }, alt: '' })]
+    [tools.RStudio.label, () => img({ src: rstudioSquareLogo, style: { height: 45, width: 45 }, alt: '' })],
+    [tools.Azure.label, () => img({ src: jupyterLogo, style: { height: 45, width: 45 }, alt: '' })]
```
cahrens, line 63: I find `Azure` as the tool name to be confusing. I would recommend changing both the front-end and back-end code to call this Juptyer_on_Azure.

I do wonder though… in the long term, is the vision that Leo will be able to create Juptyer on either GCP or Azure, regardless of which cloud platform Leo itself is running on? If that's not the vision and there will always be the constraint that Leo creates the tool on the cloud platform that Leo itself is running on, then why not make the APIs so that you tell Leo to create Jupyter, and it does it on the correct cloud platform? I suppose the issue with that would be if different options need to be specified by the user for creating Jupyter on Azure vs. GCP… if so, the front end code does need to be aware at some point so it knows which options to present, though I would still think the server API endpoint could be the same, just with different parameters passed.

src/components/ContextBar.js
```diff
@@ -125,7 +126,12 @@ export const ContextBar = ({
             style: { flexDirection: 'column', justifyContent: 'center', padding: '.75rem', ...contextBarStyles.contextBarButton, borderBottom: '0px' },
             hover: contextBarStyles.hover,
             tooltipSide: 'left',
-            onClick: () => setCloudEnvOpen(!isCloudEnvOpen),
+            onClick: () => {
+              setCloudEnvOpen(!isCloudEnvOpen)
+              if (!googleProject) {
```
cahrens: I think this would be better implementing by filtering for the tools that are supported on Azure. For example, I anticipate Cromwell may be shown here. Perhaps piggyback on `getToolsToDisplay` to help centralize the logic?

src/libs/ajax.js
```diff
@@ -1489,6 +1489,48 @@ const Runtimes = signal => ({
...
+  listV2AzureWithWorkspace: async (workspaceId, labels = {}) => {
+    const res = await fetchLeo(`api/v2/runtimes/${workspaceId}/azure?${qs.stringify({ saturnAutoCreated: true, ...labels })}`,
+      _.mergeAll([authOpts(), appIdentifier, { signal }]))
+    return res.json()
+  },
+
+  runtimeV2: (workspaceId, name, cloudProvider = 'azure') => {
+    const root = `api/v2/runtimes/${workspaceId}/${cloudProvider}/${name}`
```
cahrens, line 1509: I know this is well out of scope here, but it would be awesome to build out an API that is completely dependent on `workspaceId` and would allow the V1 API (which depends on googleProjectID) to be removed.

(From Justin in Slack)
The issue with using the same APIs for azure is that the existing v1 APIs actually have googleProject built directly into the route. We are working on building out v2 support where the routes use a workspaceId to namespace the runtimes, but unfortunately we don’t store the workspace a runtime belongs to today in the DB, so it would not be backwards compatible and would require a fallback with WSM able to translate workspaceIds to googleProjects.
The list runtime is non azure/gcp specific entirely, so it would theoretically be possible to use the same route for both, but it actually makes it slightly more convenient to leverage the backend to filter out runtimes that aren’t relevant to the workspace.
tl;dr its really just a matter of prioritizing the work to allow the v2 routes to support both for all CRUD operations.

src/libs/runtime-utils.js
```diff
@@ -127,9 +128,9 @@ const dataprocCost = (machineType, numInstances) => {
 }
 
 const getHourlyCostForMachineType = (machineTypeName, region, isPreemptible) => {
-  const { cpu, memory } = _.find({ name: machineTypeName }, machineTypes)
+  const { cpu, memory } = _.find({ name: machineTypeName }, machineTypes) || {}
```
cahrens, line 131: It would be nice to show the "None".

image

Also, why not continue to call this "Cloud Environment Details" like on GCP?

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -326,14 +330,13 @@ const Analyses = _.flow(
 
   const noAnalysisBanner = div([
     div({ style: { fontSize: 48 } }, ['A place for all your analyses ']),
-    div({ style: { display: 'flex', flexDirection: 'row', justifyContent: 'center', alignItems: 'center' } }, [
+    !!googleProject ? div({ style: { display: 'flex', flexDirection: 'row', justifyContent: 'center', alignItems: 'center' } }, [
```
cahrens: Again, here I think it would be better to filter the apps for those that are supported on the current cloud environment.

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -401,7 +405,8 @@ const Analyses = _.flow(
         div({ style: { color: colors.dark(), fontSize: 24, fontWeight: 600 } }, ['Your Analyses']),
         h(ButtonOutline, {
           style: { marginLeft: '1.5rem' },
-          onClick: () => setCreating(true),
+          //TODO: azure should eventually leverage create modal
+          onClick: () => !!googleProject ? setCreating(true) : setAzureCreating(true),
```
cahrens, line 413: So the plan is to unify the 2 create modals? I was hoping there is code that could be shared between them, and I'm wondering if it's possible to do that sooner rather than later.

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -228,7 +229,8 @@ const Analyses = _.flow(
   analysesData: { apps, refreshApps, runtimes, refreshRuntimes, appDataDisks, persistentDisks },
   onRequesterPaysError
 }, _ref) => {
-  // State
+  //TODO: revert after testing
+  var googleProject = false
```
dmohs: This is jarring because it appears to be a type mismatch. I think undefined or null would be a better choice.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -114,6 +114,7 @@ const WorkspaceContainer = ({
 
   // If googleProject is not undefined (server info not yet loaded)
   // and not the empty string, we know that we have a Google workspace.
+  //TODO: revert before merging testing
```
cahrens: This TODO is incorrect.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -219,7 +222,10 @@ const useCloudEnvironmentPolling = googleProject => {
       const [newDisks, newRuntimes] = !!googleProject ? await Promise.all([
         Ajax(signal).Disks.list({ googleProject, creator: getUser().email, includeLabels: 'saturnApplication,saturnWorkspaceName' }),
         Ajax(signal).Runtimes.list({ googleProject, creator: getUser().email })
-      ]) : [[], []]
+      ]) : await Promise.all([
+        Promise.resolve([]),
+        workspaceId ? Ajax(signal).Runtimes.listV2AzureWithWorkspace(workspaceId, { creator: getUser().email }) : Promise.resolve([])
```
cahrens: We've been trying to more consistently use the `!!` operator for when we check truthiness of a non-boolean.

cahrens:
When is the terminal button supposed to be enabled? I'm finding that in my GCP workspace, it is enabled when Juptyer is still creating or paused. In that situation, it opens the create modal and it seems like I could create a new runtime. I don't know if this behavior is different from what is on dev, but I'm confused about how the terminal button is supposed to behave when Jupyter isn't actually in the running state.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -202,7 +217,7 @@ export const CloudEnvironmentModal = ({
           shape: 'pause',
           toolLabel,
           onClick: () => stopApp(toolLabel),
-          disabled: busy || !canCompute,
+          disabled: busy || !canCompute || toolLabel === tools.Azure.label, //TODO: Azure stop/start,
```
cahrens: You can avoid this by making the tool not support Pause (like Cromwell). `isPauseUnsupported: true`

My goal was to keep tool-specific logic within the tools constant to avoid sprinkling conditionals throughout the rendering code.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -192,7 +207,7 @@ export const CloudEnvironmentModal = ({
           shape: 'play',
           toolLabel,
           onClick: () => startApp(toolLabel),
-          disabled: busy || !canCompute,
+          disabled: busy || !canCompute || toolLabel === tools.Azure.label, //TODO: Azure stop/start
```
cahrens: I don't think you need this because you won't end up in the stopped condition until Pause is supported.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -287,7 +304,7 @@ export const CloudEnvironmentModal = ({
     const isToolBusy = isToolAnApp(toolLabel) ?
       getIsAppBusy(app) || app?.status === 'STOPPED' || app?.status === 'ERROR' :
       currentRuntime?.status === 'Error'
-    const isDisabled = !doesCloudEnvForToolExist || !cookieReady || !canCompute || busy || isToolBusy || toolLabel === tools.Jupyter.label
+    const isDisabled = !doesCloudEnvForToolExist || !cookieReady || !canCompute || busy || isToolBusy || toolLabel === tools.Jupyter.label || toolLabel === tools.Azure.label
```
cahrens: Might be time to add `isLaunchUnsupported` into the tools constant.

cahrens:
I am seeing this in the Console:

​ Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at AzureComputeModalBase (http://localhost:3000/static/js/src_appLoader_js-data_image_svg_xml_3Csvg_xmlns_27http_www_w3_org_2000_svg_27_viewBox_270_0_1-d7b3bc.chunk.js:69175:24)
printWarning @ :3000/static/js/vend…e2d90.chunk.js:6277

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+            ])
+          ])
+        ])
+      ]),
+      div({ style: { marginBottom: '2rem' } }, [
+        h(IdContainer, [
+          id => h(Fragment, [
+            div({ htmlFor: id, style: { marginBottom: '.5rem', ...computeStyles.label } }, ['Disk Size (GB)']),
+            div({ style: { width: 75 } }, [
+              h(NumberInput, {
```
cahrens, line 153: This is missing an accessible label. I think that `for` only works with labels, not divs.

image

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+          h(Link, { href: 'https://azure.microsoft.com/en-us/pricing/details/virtual-machines/series/', ...Utils.newTabLinkProps }, [
+            'Learn more about cloud compute profiles',
+            icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
+          ])
+        ])
+      ]),
+      div({ style: { marginBottom: '2rem' } }, [
+        h(IdContainer, [
+          id => h(Fragment, [
+            div({ htmlFor: id, style: { marginBottom: '.5rem', ...computeStyles.label } }, ['Location']),
```
cahrens: image

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+  }
+
+  const renderComputeProfileSection = () => {
+    return div({ style: { ...computeStyles.whiteBoxContainer, marginTop: '1rem' } }, [
+      div({ style: { marginBottom: '2rem' } }, [
+        h(IdContainer, [
+          id => h(Fragment, [
+            div({ htmlFor: id, style: { marginBottom: '.5rem', ...computeStyles.label } }, ['Cloud compute profile']),
+            div({ style: { width: 400 } }, [
+              h(Select, {
```
cahrens, line 105: image

cahrens:
I think you mean to hide this inner close button.

image

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+      div({ style: { padding: '1.5rem', overflowY: 'auto', flex: 'auto' } }, [
+        renderApplicationConfigurationSection(),
+        renderComputeProfileSection(),
+        renderBottomButtons()
+      ])
+    ])
+  }
+
+  //TODO this does not actually compute cost as is, see IA-3348
+  //It is possible that once we compute the cost, we would like to parameterize this and make it a shared function between the equivalent in ComputeModal
```
cahrens, line 260: Perhaps don't even show this section until you can display cost information?

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+import colors from 'src/libs/colors'
+import { withErrorReporting, withErrorReportingInModal } from 'src/libs/error'
+import { useOnMount } from 'src/libs/react-utils'
+import {
+  computeStyles, getCurrentRuntime, getIsRuntimeBusy
+} from 'src/libs/runtime-utils'
+import * as Utils from 'src/libs/utils'
+
+
+// Change to true to enable a debugging panel (intended for dev mode only)
```
dmohs: Copy pasta?

jdcanas:
> The default size of the "Type" column needs to be increased if we are going to show AZURE_VM. It would really be nice though if the type was more informative (Jupyter on Azure).
> 
> image

Yeah, I plan to change the type in the backend after our discussion. It will most likely need to wait for next sprint though, since there's some extensive refactoring going on with that code currently.

jdcanas:
> When I click see log info, I just get another error. It is awkward to get to the details.
> 
> image
> 
> image

Yeah, so this is because of the bug I mention in the description. Essentially, the get call to get the error details is failing with a 404 due to how creation fails. Once creation is fixed in the downstream system, this won't be an issue.

```
Additionally, runtime creation is currently bugged in the backend due to some unrelated changes, so all runtimes created will result in errored runtimes. However, the create call itself is accepted by the backend and succeeds. It also errors in such a way that the GET call 404s after
```

src/components/ContextBar.js
jdcanas: it only extracts `googleProject` from `currentRuntime` in the stop function, which is not invoked for azure runtimes as it isn't supported yet.  

I went back and forth on each instance whether azureContext/googleProject is better for making the decision. I'll change it here since generally its easier to read without negation.

src/components/ContextBar.js
jdcanas: Yeah, I comment on this elsewhere. We plan to refactor the backend to make the tool more explicit. 

Currently, Leo can create Jupyter on any cloud platform, dependent on the workspace. In other words, for a given workspace, it will only support one cloud platform and there's no restriction based on where leonardo itself is deployed. With that in mind, it helps to have the front-end differentiate for two reasons. One, for compatibility with existing runtimes which are not scoped to a workspace from Leo's perspective. Two, for the different runtime configs you mention,

src/libs/ajax.js
jdcanas: That is the eventual plan :)

src/libs/runtime-utils.js
jdcanas: I didn't worry too much about cost in this PR, as there's a future ticket to ensure it displays properly. https://broadworkbench.atlassian.net/browse/IA-3348

The title wording is a recent change, if you open the modal by selecting the tool in the context bar on the right, it will display the tool name in the title. Equivalent for Jupyter shown below. 

image

src/pages/workspaces/workspace/Analyses.js
jdcanas: Hmm... I took an approach here which I don't love, but cleanly does the job. Let me know if you have other ideas

src/pages/workspaces/workspace/Analyses.js
jdcanas: So right now the issue isn't that there is unshared code being duplicated. Currently, the flow for non-azure environments is

`setCreating` opens the modal to create an analysis file, and after that, it opens the environment create modal.

For azure, we do not yet support analysis files, so we skip the intermediate step entirely. It would be possible to take the same approach cromwell does and have an entry in the `setCreating` modal, which opens the environment create modal directly. However, we plan to support analysis file creation for azure sooner rather than later, so we'd rather bunch up that work and asset creation.

src/pages/workspaces/workspace/WorkspaceContainer.js
jdcanas: I'll remove before merging. I'm mainly keeping these `TODO: revert ...` comments here to help me find the places I manually modified as I'll need all 6 when I demo it this week.

src/pages/workspaces/workspace/WorkspaceContainer.js
jdcanas: > When is the terminal button supposed to be enabled? I'm finding that in my GCP workspace, it is enabled when Juptyer is still creating or paused. In that situation, it opens the create modal and it seems like I could create a new runtime. I don't know if this behavior is different from what is on dev, but I'm confused about how the terminal button is supposed to behave when Jupyter isn't actually in the running state.

Yeah, if you have a Jupyter environment and its not in the running state, it either
1) is enabled if the runtime will eventually be in an intractable state or (starting/creating)
2) if it can be started

There's also, confusingly, a known bug where the ComputeModal (for creating an env) pops up inappropriately on launch we are targetting for the next sprint https://broadworkbench.atlassian.net/browse/IA-2989

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
jdcanas: I like it there for style / spacing purposes personally

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
jdcanas: ah... not going to lie I thought the htmlFor was to prevent this issue https://sentry.io/answers/unique-key-prop/. Fixed now

src/pages/workspaces/workspace/Analyses.js
cahrens: I was thinking that each tool should provide its image for display here, and there should be a way to filter the tools for the ones that are supported per cloud context. In other words, push the logic down into the tool object so you are just iterating over the tools for the cloud context and getting their images.

Note also that these images need alt text.

src/pages/workspaces/workspace/WorkspaceContainer.js
cahrens: What I meant is that there was no modification to the code here, so there is nothing to revert in this location.

Note that you can now create Azure workspaces from within terra UI now on dev, so you can probably just rip out all of this TODO stuff now. You do need to let Drew or me know the email address you use to log in on dev so we can add you to the appropriate groups.

image

src/components/ContextBar.js
```diff
@@ -125,7 +126,9 @@ export const ContextBar = ({
             style: { flexDirection: 'column', justifyContent: 'center', padding: '.75rem', ...contextBarStyles.contextBarButton, borderBottom: '0px' },
             hover: contextBarStyles.hover,
             tooltipSide: 'left',
-            onClick: () => setCloudEnvOpen(!isCloudEnvOpen),
+            onClick: () => {
+              setCloudEnvOpen(!isCloudEnvOpen)
+            },
```
dmohs: This wants an inline review (not code) comment since I'm compelled to ask, why is this better?

src/components/RuntimeManager.js
```diff
@@ -72,7 +73,8 @@ export const RuntimeErrorModal = ({ runtime, onDismiss }) => {
     withErrorReporting('Error loading cloud environment details'),
     Utils.withBusyState(setLoadingRuntimeDetails)
   )(async () => {
-    const { errors: runtimeErrors } = await Ajax().Runtimes.runtime(runtime.googleProject, runtime.runtimeName).details()
+    const { errors: runtimeErrors } = _.lowerCase(runtime.cloudContext.cloudProvider) === cloudProviders.azure.label ? await Ajax().Runtimes.runtimeV2(runtime.workspaceId, runtime.runtimeName).details() :
+      await Ajax().Runtimes.runtime(runtime.googleProject, runtime.runtimeName).details()
```
dmohs: This is how I like to split long tertiaries to call attention to the fact that there is a conditional here:
```suggestion
    const { errors: runtimeErrors } = _.lowerCase(runtime.cloudContext.cloudProvider) === cloudProviders.azure.label
      ? await Ajax().Runtimes.runtimeV2(runtime.workspaceId, runtime.runtimeName).details()
      : await Ajax().Runtimes.runtime(runtime.googleProject, runtime.runtimeName).details()
```

src/components/notebook-utils.js
```diff
@@ -73,15 +73,20 @@ export const analysisNameInput = ({ inputProps, ...props }) => h(ValidatedInput,
...
+  Azure: { label: 'Azure', isAzureCompatible: true, isPauseUnsupported: true, isLaunchUnsupported: true }
 }
 
 // Returns the tools in the order that they should be displayed for Cloud Environment tools
-export const getToolsToDisplay = _.remove(tool => tool.isAppHidden)([tools.Jupyter, tools.RStudio, tools.Galaxy, tools.Cromwell])
+export const getToolsToDisplay = isAzureWorkspace => _.flow(
+  _.remove(tool => tool.isAppHidden),
+  isAzureWorkspace ? _.remove(tool => !tool.isAzureCompatible) : _.remove(tool => tool.isAzureCompatible))(
+  [tools.Jupyter, tools.RStudio, tools.Galaxy, tools.Cromwell, tools.Azure]
+)
```
dmohs: Not on you, but I don't like having to read these in the order: n, 1, 2, ..., n-1. If this is a "flow," then the river has rocks in it. I think the following indentation helps slightly (with additional suggestions):
```suggestion
export const getToolsToDisplay = isAzureWorkspace => _.flow(
  _.remove(_.get('isAppHidden')),
  _.filter(tool => tool.isAzureCompatible === isAzureWorkspace)
)([tools.Jupyter, tools.RStudio, tools.Galaxy, tools.Cromwell, tools.Azure])
```
@DataBiosphere/terra-ui-code-owners I almost wish for a utility function:
```javascript
const flowInto = (x, ...fns) => _.flow(...fns)(x)
```

src/libs/azure-utils.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const defaultAzureComputeConfig = {
+  machineType: defaultAzureMachineType,
+  diskSize: defaultAzureDiskSize,
+  region: defaultAzureRegion
+}
+
+//TODO: this should be fleshed out once azure region isn't limited to workspace region
+export const azureRegions = { eastus: { label: 'East US' } }
+export const getRegionLabel = label => _.has(label, azureRegions) ? azureRegions[label].label : 'Unknown azure region'
```
dmohs: Since you're getting the label, I think the argument is a key:
```suggestion
export const getRegionLabel = key => azureRegions?.[key].label || 'Unknown azure region'
```

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,332 @@
...
+                onlyInteger: true,
+                value: computeConfig.diskSize,
+                onChange: v => updateComputeConfig('diskSize', v)
+              })
+            ])
+          ])
+        ])
+      ])
+    ])
+  }
```
dmohs, line 167: I judge that collapsing these to a single line gives a better signal-to-noise ratio:
```suggestion
  })])])])])])}
```

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -228,7 +229,8 @@ const Analyses = _.flow(
   analysesData: { apps, refreshApps, runtimes, refreshRuntimes, appDataDisks, persistentDisks },
   onRequesterPaysError
 }, _ref) => {
-  // State
+  // TODO: revert after testing
+  // var googleProject = false
```
dmohs: This looks like a type mismatch. Would undefined or null work here?
Also, `let` gives you mutability with the scope-sanity of `const`.
```suggestion
  // let googleProject = undefined
```

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,332 @@
...
+  //   return !_.isEqual(existingConfig, computeConfig)
+  // }
+  //
+  // const adaptRuntimeDetailsToFormConfig = () => {
+  //   return currentRuntimeDetails ? {
+  //     machineType: currentRuntimeDetails.runtimeConfig?.machineType || defaultAzureMachineType,
+  //     diskSize: currentRuntimeDetails.diskConfig?.size || defaultAzureDiskSize,
+  //     region: currentRuntimeDetails.runtimeConfig?.region || defaultAzureRegion
+  //   } : {}
+  // }
```
dmohs, line 182: Were you able to test this? If so, perhaps helpful to leave it in. Otherwise, maybe it can wait until update is indeed supported.

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,332 @@
...
+
+  const doesRuntimeExist = () => !!currentRuntimeDetails
```
dmohs, line 184: I like utilities like this that makes something potentially ambiguous into something explicit.

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,332 @@
...
+
+  const renderActionButton = () => {
+    const commonButtonProps = {
+      tooltipSide: 'left',
+      disabled: Utils.cond([viewMode === 'deleteEnvironment', () => getIsRuntimeBusy(currentRuntimeDetails)],
+        () => doesRuntimeExist()),
+      tooltip: Utils.cond([viewMode === 'deleteEnvironment', () => getIsRuntimeBusy(currentRuntimeDetails) ? 'Cannot delete a runtime while it is busy' : undefined],
```
dmohs: Lines this long make it a bit harder for me to review.

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,332 @@
...
+            saturnWorkspaceNamespace: namespace,
+            saturnWorkspaceName: workspaceName
+          },
+          disk
+        })
+      }]
+    )
+
+    onSuccess()
+  })
```
dmohs, line 243: The section above could use a little care and feeding. I threw some time in your calendar to talk through it and hopefully finish things up.

dmohs:
This took a bit more time to review than I would have liked. The main catch was that it seemed that AzureComputeModal was copied from ComputeModal and then modified. I value this technique, but it carries the risk of unintentially duplicating code or carrying in artifacts. Thus, as a reviewer, I'm obligated to help the author avoid this.

Knowing this going in, you may be able to find ways to aid your reviewer. For example, if your first two commits were (1) copy the file and (2) first-pass delete of code that won't be used, you could draw the reviewer's attention there. Another option would be to just call it out and mention anything interesting. Either way, anything you can do to avoid surprise will help.

One way to reduce the size might be to PR the AJAX calls first. That would remove them as a distraction when you submit the PR that adds the UI.

src/pages/workspaces/workspace/Analyses.js
kyuksel: Curious what the use case requiring a mutable variable for `googleProject` would be

cahrens: This was just a temporary change to mock having an Azure workspace (see the TODO).

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
jdcanas: Hmm.. I was following the comment above. I don't the useEffect is compatible with async functions as the body. I see a lot of other examples of this pattern throughout the codebase.

src/pages/workspaces/workspace/Analyses.js
jdcanas: See above comment, its to enable testing

src/components/ContextBar.js
jdcanas: Not sure what you're asking here? This line is so when this particular button in the contextBar is clicked the panel is open. It's functionally equivalent to setCloudEnvOpen(true), if that's what you're referring to.

src/components/notebook-utils.js
jdcanas: Ah I like this and the filter better. I knew there was a way to use filter over a ternary remove but my Monday brain was not grasping it :)

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
jdcanas: I personally find the indentation invaluable when modifying the inner structure. I would not be able to properly close a newly-added div without it

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
jdcanas: Yeah, this is tested to ensure the proper button disabling when the modal is in an 'update' state, so it is valid code. It's just not supported in the backend yet.

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
jdcanas: Yeah, there's some history involved here and some lessons learned that went into `renderActionButton` and `applyChanges`. Would be happy to talk about the approach, since a lot of future-proofing went into the thought process behind the respective implementations.

dmohs (approved):
Remaining suggestions covered on a call.

Qi77Qi (approved):

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+        }
+      ])
+    ])
+  }
+
+  const renderDeleteEnvironment = () => {
+    return div({ style: { ...computeStyles.drawerContent, ...computeStyles.warningView } }, [
+      h(TitleBar, {
+        id: titleId,
+        shouldHideCloseButton,
```
cahrens: Close button is still visible. I think the name of the variable is wrong.

image

src/components/ContextBar.js
cahrens: You could make tools.Azure.label "Jupyter" though to help reduce the confusion. It is odd to see "Open Azure", and that would also make the title "Jupyter Environment Details".

image

cahrens:
Is this repeated error expected?

image

Note, to get into this state, click "Open Azure" while Jupyter is still creating (before it errors). I doubt the button is actually supposed to be enabled then.

image

And note also that one of the errors present is a bucket location error. In an Azure Workspace, I wouldn't expect that you are trying to fetch the Bucket location.

image

src/pages/workspaces/workspace/Analyses.js
cahrens: Created https://broadworkbench.atlassian.net/browse/IA-3380 for images missing alt text.

cahrens:
I am not able to delete the Azure instances from the Cloud Environments page.

image

src/components/notebook-utils.js
```diff
@@ -73,15 +73,19 @@ export const analysisNameInput = ({ inputProps, ...props }) => h(ValidatedInput,
...
-  Cromwell: { label: 'Cromwell', appType: 'CROMWELL', isAppHidden: !isCromwellAppVisible(), isPauseUnsupported: true }
+  Cromwell: { label: 'Cromwell', appType: 'CROMWELL', isAppHidden: !isCromwellAppVisible(), isPauseUnsupported: true },
+  Azure: { label: 'Azure', isAzureCompatible: true, isPauseUnsupported: true, isLaunchUnsupported: true }
 }
 
 // Returns the tools in the order that they should be displayed for Cloud Environment tools
-export const getToolsToDisplay = _.remove(tool => tool.isAppHidden)([tools.Jupyter, tools.RStudio, tools.Galaxy, tools.Cromwell])
+export const getToolsToDisplay = isAzureWorkspace => _.flow(
+  _.remove(tool => tool.isAppHidden),
+  _.filter(tool => !!tool.isAzureCompatible === !!isAzureWorkspace)
```
cahrens, line 87: Nit: if something is a boolean to start with, there is no need to use "!!".

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -304,17 +306,19 @@ const Analyses = _.flow(
         reportError('Error creating analysis', error)
       }
     }
-  })
+  }) : () => {}
 
   // Lifecycle
   useOnMount(() => {
     const load = async () => {
```
cahrens: I think you need to wrap these Ajax calls with `withErrorReporting`. Otherwise, if an error occurs with one of them, a rejected promise results and you can end up with integration test failures with no error messages. For example, we just saw this type of failure on a run-notebook test, and in the console there was this 500 error:

````
page.http GET	200  https://leonardo.dsde-dev.broadinstitute.org/api/google/v1/runtimes?saturnAutoCreated=true&googleProject=terra-dev-f6ee1cfe&creator=scarlett.flowerpicker%40test.firecloud.org
page.http POST	500  https://sam.dsde-dev.broadinstitute.org/api/google/user/petServiceAccount/terra-dev-f6ee1cfe/token
page.console ConsoleMessage {
  _type: 'error',
  _text: 'Failed to load resource: the server responded with a status of 500 ()',
  _args: [],
  _stackTraceLocations: [
    {
      url: 'https://sam.dsde-dev.broadinstitute.org/api/google/user/petServiceAccount/terra-dev-f6ee1cfe/token',
      lineNumber: undefined
    }
  ]
}
````

I think that all of our Ajax calls should either be wrapped with `withErrorIgnoring` (if the error is not important and we don't want to notify the user, for instance in the case of NPS survey or metrics logging not working), or `withErrorReporting`.

File: src/pages/workspaces/workspace/WorkspaceContainer.js [hunk 36]
```diff
@@ -79,7 +79,7 @@ const WorkspaceTabs = ({
     ...(isGoogleWorkspace && !isAnalysisTabVisible() ? [{ name: 'notebooks', link: 'workspace-notebooks' }] : []),
     // the spread operator results in no array entry if the config value is false
     // we want this feature gated until it is ready for release
-    ...(isGoogleWorkspace && isAnalysisTabVisible() ? [{ name: 'analyses', link: analysisTabName }] : []),
+    ...(isAnalysisTabVisible() ? [{ name: 'analyses', link: analysisTabName }] : []),
```
cahrens, line 82: In an Azure Workspace, you might want to show the Analyses Tab unconditionally since the audience is already limited (also impacts line 150).

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+      diskSize: currentRuntimeDetails?.diskConfig?.size || defaultAzureDiskSize,
+      region: currentRuntime?.runtimeConfig?.region || defaultAzureRegion
+    })
+  }))
+
+  const renderTitleAndTagline = () => {
+    return h(Fragment, [
+      h(TitleBar, {
+        id: titleId,
+        shouldHideCloseButton,
```
cahrens: Close button is still visible. I think the name of the variable is wrong.

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -0,0 +1,324 @@
...
+  }
+
+  const renderApplicationConfigurationSection = () => {
+    return div({ style: computeStyles.whiteBoxContainer }, [
+      h(IdContainer, [
+        id => h(Fragment, [
+          div({ style: { marginBottom: '1rem' } }, [
+            label({ htmlFor: id, style: computeStyles.label }, ['Application configuration']),
+            h(InfoBox, { style: { marginLeft: '0.5rem' } }, [
+              'Currently, the azure vm is pre-configured. '
```
cahrens: Nit: Azure VM (capitalization).

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -65,6 +66,21 @@ export const CloudEnvironmentModal = ({
     }
   })
 
+  const renderAzureModal = () => h(AzureComputeModalBase, {
+    isOpen: viewMode === NEW_AZURE_MODE,
+    shouldHideCloseButton: true,
```
cahrens: Nit: consider making `shouldHideCloseButton` true by default since you will never want to show it.

jdcanas:
> Is this repeated error expected?
> 
> image
> 
> Note, to get into this state, click "Open Azure" while Jupyter is still creating (before it errors). I doubt the button is actually supposed to be enabled then.
> 
> image
> 
> And note also that one of the errors present is a bucket location error. In an Azure Workspace, I wouldn't expect that you are trying to fetch the Bucket location.
> 
> image

 I fixed the launch button being clickable, but its a bit of undefined behavior to be on the launch page when no launchable apps are available in azure yet. I disabled the offending erroring calls for that page anyways though

jdcanas:
> I am not able to delete the Azure instances from the Cloud Environments page.
> 
> image

Cloud environments page is not in scope for this initial kickoff in the effort to make reasonable chunks of work. See https://broadworkbench.atlassian.net/browse/IA-3384

src/pages/workspaces/workspace/WorkspaceContainer.js
jdcanas: Eventually, analysis tab visibility will go away entirely (likely before azure is released), so I'd rather not complicate it here.

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
jdcanas: it needs to be shown when its open from the `+Start` button, but not from the cloud env modal

cahrens: Ah, interesting! I never realized until now that the 2 paths are at all different (I thought the Start button was just a shortcut to getting to the settings page for a tool).

src/pages/workspaces/workspace/Analyses.js
cahrens: To clarify, I was referring to the useOnMount code below. Usually the block of code in useOnMount is wrapped in a `withErrorReporting`, for example, line you did in `AzureComputeModal`.

https://github.com/DataBiosphere/terra-ui/pull/2987/files#diff-3d432e483156cf5b8150a7569d085e37b87f2e72ccc6be06963ec3d8b67319cbR37

jdcanas: Ah ok. Based on this logic it would seem both could use it :)

cahrens: Yes!

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
jdcanas: Yeah, the difference is from the start button its accessible directly, where as in the other flow its embedded as part of a larger user journey. Eventually, when we support azure analysis (file) creation from the `+Start` button, it will be embedded as part of a larger flow there too, and the default can be made `true` and not specified at any call sites.

cahrens (approved):
Note: did not manually test the last couple of commits.

PR #2988, opened 2022-04-28
[WOR-263] Changes for ToS enforcement (WOR-263).

cahrens:
Don't call server APIs that require ToS to be accepted until the user has actually accepted the ToS.

We will likely do a follow-up PR after auditing all usages of "registrationStatus" to ensure that ToS status is also taken into account if necessary. This may involve changing the definition of "registered" in Terra-ui to include verifying ToS acceptance. But we would like to go ahead and merge this initial change to get `register-user` passing again on dev.

I ran the test against both integration and staging, and it passed there as well.

### DIFF
--- a/src/components/NpsSurvey.js
+++ b/src/components/NpsSurvey.js
@@ -27,7 +27,7 @@ const NpsSurvey = () => {
   const [reasonComment, setReasonComment] = useState('')
   const [changeComment, setChangeComment] = useState('')
 
-  const { registrationStatus } = useStore(authStore)
+  const { registrationStatus, acceptedTos } = useStore(authStore)
 
   const loadStatus = async () => {
     const lastResponseTimestamp = (await Ajax().User.lastNpsResponse()).timestamp
@@ -42,12 +42,12 @@ const NpsSurvey = () => {
   }
 
   useEffect(() => {
-    if (registrationStatus === 'registered') {
+    if (registrationStatus === 'registered' && acceptedTos) {
       loadStatus()
     } else {
       setRequestable(false)
     }
-  }, [registrationStatus])
+  }, [registrationStatus, acceptedTos])
 
   const goAway = shouldSubmit => () => {
     setRequestable(false)
--- a/src/components/NpsSurvey.js
+++ b/src/components/NpsSurvey.js
@@ -30,18 +30,18 @@ const NpsSurvey = () => {
 
   const { registrationStatus, acceptedTos } = useStore(authStore)
 
-  const loadStatus = withErrorIgnoring(async () => {
-    const lastResponseTimestamp = (await Ajax().User.lastNpsResponse()).timestamp
-    // Behavior of the following logic: When a user first accesses Terra, wait 7 days to show the NPS survey.
-    // Once user has interacted with the NPS survey, wait 90 days to show the survey.
-    const askTheUser = !!lastResponseTimestamp ?
-      differenceInCalendarDays(parseJSON(lastResponseTimestamp), Date.now()) >= 90 :
-      differenceInCalendarDays(parseJSON((await Ajax().User.firstTimestamp()).timestamp), Date.now()) >= 7
-    setRequestable(askTheUser)
-  })
-
   useEffect(() => {
     if (registrationStatus === 'registered' && acceptedTos) {
+      const loadStatus = withErrorIgnoring(async () => {
+        const lastResponseTimestamp = (await Ajax().User.lastNpsResponse()).timestamp
+        // Behavior of the following logic: When a user first accesses Terra, wait 7 days to show the NPS survey.
+        // Once user has interacted with the NPS survey, wait 90 days to show the survey.
+        const askTheUser = !!lastResponseTimestamp ?
+          differenceInCalendarDays(parseJSON(lastResponseTimestamp), Date.now()) >= 90 :
+          differenceInCalendarDays(parseJSON((await Ajax().User.firstTimestamp()).timestamp), Date.now()) >= 7
+        setRequestable(askTheUser)
+      })
+
       loadStatus()
     } else {
       setRequestable(false)
--- a/src/components/NpsSurvey.js
+++ b/src/components/NpsSurvey.js
@@ -8,6 +8,7 @@ import { TextArea } from 'src/components/input'
 import Interactive from 'src/components/Interactive'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
+import { withErrorIgnoring } from 'src/libs/error'
 import { getAppName } from 'src/libs/logos'
 import { useStore } from 'src/libs/react-utils'
 import { authStore } from 'src/libs/state'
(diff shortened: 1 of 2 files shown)


src/components/NpsSurvey.js
```diff
@@ -42,12 +42,12 @@ const NpsSurvey = () => {
   }
 
   useEffect(() => {
-    if (registrationStatus === 'registered') {
+    if (registrationStatus === 'registered' && acceptedTos) {
```
cahrens, line 34: Without this, before the ToS is accepted we get this error in the console, which cases the test to fail (but with no indication of why):

asyncToGenerator.js:6 Uncaught (in promise) 
Response {type: 'cors', url: 'https://terra-rex-dev.appspot.com/api/npsResponses/lastTimestamp', redirected: false, status: 403, ok: false, …}
body: (...)
bodyUsed: false
headers: Headers {}
ok: false
redirected: false
status: 403
statusText: ""
type: "cors"
url: "https://terra-rex-dev.appspot.com/api/npsResponses/lastTimestamp"
[[Prototype]]: Response

The fact that the error is uncaught (as opposed to caught and displayed to the user like we normally do) seems to be the issue.

src/libs/auth.js
```diff
@@ -37,6 +37,10 @@ export const hasBillingScope = () => {
   return getAuthInstance().currentUser.get().hasGrantedScopes('https://www.googleapis.com/auth/cloud-billing')
 }
 
+const registeredAndAcceptedTos = (oldState, state) => {
+  return (oldState.registrationStatus !== 'registered' && state.registrationStatus === 'registered' && state.acceptedTos) ||
+    (state.registrationStatus === 'registered' && state.acceptedTos && !oldState.acceptedTos)
+}
```
cahrens, line 43: I don't actually know if _all_ the changes below were necessary, but I am assuming that if the user hasn't accepted the ToS they shouldn't be able to do much. I can try to narrow down which ones actually need this check tomorrow.

cahrens:
@marctalbott what were you trying to accomplish with the check for ToS at https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/auth.js#L175? I keep going back and forth about whether we should change the `registrationStatus` so that it reflects both registered and acceptedTos so that we don't have to check both pieces of information everywhere.

marctalbott (commented):
Looks good. I think changing the concept of "registered" to include having accepted the ToS as you said in Slack makes sense. "registered" used to mean the user was enabled and ready to use Terra, but now that they also have to accept the ToS to be enabled and ready to use Terra, the old definition of "registered" seems outdated

marctalbott (approved):

marctalbott:
> @marctalbott what were you trying to accomplish with the check for ToS at https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/auth.js#L175? I keep going back and forth about whether we should change the `registrationStatus` so that it reflects both registered and acceptedTos so that we don't have to check both pieces of information everywhere.

IIRC, without that check the UI wouldn't re-check if the user was enabled after accepting the ToS. The flow looked something like
1. User signs in, but hasn't accepted the new version of the ToS. When the user signs in, the UI fetches their status in Sam and sees that they are not enabled and haven't accepted the ToS.
2. Since the user hasn't accepted the ToS, the UI prompts the user with the ToS. `authStore` still says the user is disabled.
3. The user accepts the ToS and is enabled in Terra. The UI doesn't reload the user's status and `authStore` still says the user is disabled. This is where the UI _should_ reload the user's status which is what I tried to address with the check for ToS that you linked.
4. Without the linked change, the UI sees that the user is still disabled in Terra and displays the page saying the user is disabled and that an admin will need to re-enable them.

src/libs/auth.js
cahrens: I will continue to look at this in the follow-up PR.

cahrens: Another note for context: we were hitting a problem where the 'getFenceStatus' call was happening when the user became registered, before ToS was accepted. Since ToS is now required, this resulted in a 401, whereas it used to result in a 403. The 401 caused the user to be logged out due to [this code](https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/error.js#L7), failing the test since we did not get to the expected Workspaces list.

Then once I fixed the 'getFenceStatus' issue, the test appeared to run correctly, but still failed due to the NpsSurvey unprotected AJAX call.

cahrens:
> IIRC, without that check the UI wouldn't re-check if the user was enabled after accepting the ToS. The flow looked something like
> 
> 1. User signs in, but hasn't accepted the new version of the ToS. When the user signs in, the UI fetches their status in Sam and sees that they are not enabled and haven't accepted the ToS.
> 2. Since the user hasn't accepted the ToS, the UI prompts the user with the ToS. `authStore` still says the user is disabled.
> 3. The user accepts the ToS and is enabled in Terra. The UI doesn't reload the user's status and `authStore` still says the user is disabled. This is where the UI _should_ reload the user's status which is what I tried to address with the check for ToS that you linked.
> 4. Without the linked change, the UI sees that the user is still disabled in Terra and displays the page saying the user is disabled and that an admin will need to re-enable them.

But we have seen that it was possible to get a registration status of `enabled` and yet the ToS hadn't been accepted (or at least hadn't propagated through the system?). Let's touch base in person about this.

```
      return enabled ? 'registered' : 'disabled'

File: src/libs/auth.js (hunk 2 above)
Comment (kyuksel) [reply src/libs/auth.js:43]: Ah, this context was helpful; thank you.

File: src/libs/auth.js [hunk 3]
```diff
@@ -37,6 +37,10 @@ export const hasBillingScope = () => {
   return getAuthInstance().currentUser.get().hasGrantedScopes('https://www.googleapis.com/auth/cloud-billing')
 }
 
+const registeredAndAcceptedTos = (oldState, state) => {
+  return (oldState.registrationStatus !== 'registered' && state.registrationStatus === 'registered' && state.acceptedTos) ||
+    (state.registrationStatus === 'registered' && state.acceptedTos && !oldState.acceptedTos)
```
kyuksel, line 42: I have a difficult time reasoning about the logic above. I'm sure I'm lacking some domain knowledge around registration status and ToS acceptance. Why do we care about `oldState`? That is, why can't we simply check if `state.registrationStatus === 'registered' && state.acceptedTos`?

kyuksel: FWIW: Assuming we need all of the logic, the following reads a bit easier for me: 

```suggestion
  return state.registrationStatus === 'registered' && state.acceptedTos && (oldState.registrationStatus !== 'registered' || !oldState.acceptedTos)
```

kyuksel (approved):
@cahrens Thank you for fixing the test!

+1 on your and @marctalbott's tinkering around opportunities for better API design, especially if `if (registeredAndAcceptedTos(oldState, state))`s turn out to be all necessary.

src/libs/auth.js
cahrens: My understanding of this code (which I am still learning about) is that we only want to make these Ajax calls once per session, either at log in for someone already registered/accepted TOS, and then at the time that a new user transitions from not being registered/not having accepted TOS to "fully registered". I think that is what the "oldState/newState" checks are about. But I am going to dig into this more.

PR #2990, opened 2022-04-28
[WOR-276] Cleanup of registration status, Ajax call error handling (WOR-276).

cahrens:
This is a follow-up PR to https://github.com/DataBiosphere/terra-ui/pull/2990 (which I wanted to merge as soon as possible to get the `register_user` test passing again).

Now that I've had time to better understand the code and talk with @marctalbott, I am modifying my previous solution to be more explicit and to better address the root issue. Before the ToS rollout (currently on dev only), being "registered" was enough to call most of our server methods. Now you must be both "registered" (have created a user profile) and have "accepted the ToS".

### DIFF
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -6,14 +6,16 @@ import { FrameworkServiceLink, ShibbolethLink, UnlinkFenceAccount } from 'src/co
 import { cookiesAcceptedKey } from 'src/components/CookieWarning'
 import { Ajax, fetchOk } from 'src/libs/ajax'
 import { getConfig } from 'src/libs/config'
-import { withErrorReporting } from 'src/libs/error'
+import { withErrorIgnoring, withErrorReporting } from 'src/libs/error'
 import { captureAppcuesEvent } from 'src/libs/events'
 import { getAppName } from 'src/libs/logos'
 import * as Nav from 'src/libs/nav'
 import { clearNotification, notify, sessionTimeoutProps } from 'src/libs/notifications'
 import { getLocalPref, getLocalPrefForUserId, setLocalPref } from 'src/libs/prefs'
 import allProviders from 'src/libs/providers'
-import { asyncImportJobStore, authStore, cookieReadyStore, requesterPaysProjectStore, workspacesStore, workspaceStore } from 'src/libs/state'
+import {
+  asyncImportJobStore, authStore, cookieReadyStore, requesterPaysProjectStore, userStatus, workspacesStore, workspaceStore
+} from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
 
@@ -37,9 +39,8 @@ export const hasBillingScope = () => {
   return getAuthInstance().currentUser.get().hasGrantedScopes('https://www.googleapis.com/auth/cloud-billing')
 }
 
-const registeredAndAcceptedTos = (oldState, state) => {
-  return (oldState.registrationStatus !== 'registered' && state.registrationStatus === 'registered' && state.acceptedTos) ||
-    (state.registrationStatus === 'registered' && state.acceptedTos && !oldState.acceptedTos)
+const becameRegistered = (oldState, state) => {
+  return (oldState.registrationStatus !== userStatus.registeredWithTos && state.registrationStatus === userStatus.registeredWithTos)
 }
 /*
  * Request Google Cloud Billing scope if necessary.
@@ -167,10 +168,14 @@ authStore.subscribe(withErrorReporting('Error checking registration', async (sta
   const getRegistrationStatus = async () => {
     try {
       const { enabled } = await Ajax().User.getStatus()
-      return enabled ? 'registered' : 'disabled'
+      if (enabled) {
+        return state.acceptedTos ? userStatus.registeredWithTos : userStatus.registeredWithoutTos
+      } else {
+        return userStatus.disabled
+      }
     } catch (error) {
       if (error.status === 404) {
-        return 'unregistered'
+        return userStatus.unregistered
       } else {
         throw error
       }
@@ -192,7 +197,7 @@ authStore.subscribe(withErrorReporting('Error checking TOS', async (state, oldSt
   }
 }))
 
-authStore.subscribe(async (state, oldState) => {
+authStore.subscribe(withErrorIgnoring(async (state, oldState) => {
   if (!oldState.acceptedTos && state.acceptedTos) {
     if (window.Appcues) {
       window.Appcues.identify(state.user.id, {
@@ -201,7 +206,7 @@ authStore.subscribe(async (state, oldState) => {
       window.Appcues.on('all', captureAppcuesEvent)
     }
   }
-})
+}))
 
 authStore.subscribe(state => {
   // We can't guarantee that someone reopening the window is the same user,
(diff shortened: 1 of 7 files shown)


src/libs/auth.js
```diff
@@ -167,10 +168,14 @@ authStore.subscribe(withErrorReporting('Error checking registration', async (sta
   const getRegistrationStatus = async () => {
     try {
       const { enabled } = await Ajax().User.getStatus()
-      return enabled ? 'registered' : 'disabled'
+      if (enabled) {
+        return state.acceptedTos ? userStatus.registeredWithTos : userStatus.registeredWithoutTos
```
cahrens, line 174: From discussions with @marctalbott, I think that it is now impossible for us to get "enabled" from this Sam endpoint and yet have `acceptedTos` be false (it would be a bug, at least on dev). However, to be very explicit and because we do store `acceptedTos` separately, I am including `acceptedTos` in the status check.

I am tempted to make this an assertion instead of having a separate state for something we don't expect. Curious to see what reviewers think….

src/libs/auth.js
```diff
@@ -192,7 +197,7 @@ authStore.subscribe(withErrorReporting('Error checking TOS', async (state, oldSt
   }
 }))
 
-authStore.subscribe(async (state, oldState) => {
+authStore.subscribe(withErrorIgnoring(async (state, oldState) => {
```
cahrens, line 202: Taking this opportunity to wrap more Ajax calls with error handling to prevent potential sporadic and difficult to diagnose test failures.

src/pages/Register.js
```diff
@@ -35,7 +35,7 @@ const Register = () => {
         lastName: familyName,
         contactEmail: email
       })
-      authStore.update(state => ({ ...state, registrationStatus: 'registered' }))
+      authStore.update(state => ({ ...state, registrationStatus: userStatus.createdProfile }))
```
cahrens: This code was the cause of the Ajax calls where `registrationStatus` was `registered` and yet `acceptedTos` was `false`. We should not be assuming `registered` here as the full workflow is not yet complete.

cahrens:
@kyuksel here is my follow-up PR, if you want to see if it addresses the questions/concerns you had in the last one.

src/libs/auth.js
kyuksel: From your description, encapsulating it in an assertion makes sense to me.

src/libs/auth.js
```diff
@@ -37,9 +39,8 @@ export const hasBillingScope = () => {
   return getAuthInstance().currentUser.get().hasGrantedScopes('https://www.googleapis.com/auth/cloud-billing')
 }
 
-const registeredAndAcceptedTos = (oldState, state) => {
-  return (oldState.registrationStatus !== 'registered' && state.registrationStatus === 'registered' && state.acceptedTos) ||
-    (state.registrationStatus === 'registered' && state.acceptedTos && !oldState.acceptedTos)
+const becameRegistered = (oldState, state) => {
+  return (oldState.registrationStatus !== userStatus.registeredWithTos && state.registrationStatus === userStatus.registeredWithTos)
```
kyuksel, line 43: Nit: `newState` reads more easily to me for some reason:

```suggestion
const becameRegistered = (oldState, newState) => {
  return (oldState.registrationStatus !== userStatus.registeredWithTos && newState.registrationStatus === userStatus.registeredWithTos)
```

src/libs/auth.js
kyuksel: Asking to understand: Why is ignoring the desirable behaviour?

src/libs/auth.js
```diff
@@ -214,7 +219,7 @@ authStore.subscribe(state => {
 })
 
 authStore.subscribe(withErrorReporting('Error checking groups for timeout status', async (state, oldState) => {
-  if (registeredAndAcceptedTos(oldState, state)) {
+  if (becameRegistered(oldState, state)) {
```
kyuksel, line 224: I love the much better naming `becameRegistered`.

src/libs/state.js
```diff
@@ -15,6 +15,14 @@ export const authStore = Utils.atom({
   cookiesAccepted: undefined
 })
 
+export const userStatus = {
+  unregistered: 'unregistered',
+  createdProfile: 'createdProfile',
+  registeredWithoutTos: 'registeredWithoutTos',
+  registeredWithTos: 'registered',
+  disabled: 'disabled'
+}
```
kyuksel: I was hoping you'd create this enum 🤩

kyuksel (approved):
Great improvements!

kyuksel:
@petesantos I removed you from the list of assigned reviewers since I ended up looking at this PR out of curiosity (since it was a sequel to a PR I had reviewed) and figured your time could be better spent on a PR not yet reviewed but you're welcome to review this one too of course :)

src/libs/auth.js
marctalbott: I'm looking at the Sam code for this endpoint and I don't think it's possible for Sam to return `enabled: true` when the grace period is disabled and the user hasn't accepted the ToS ([Sam source code](https://github.com/broadinstitute/sam/blob/c62dd6b769adce9215399f1909c62a285f93402f/src/main/scala/org/broadinstitute/dsde/workbench/sam/service/UserService.scala#L171)). Whether that translates to the value stored in `acceptedTos` here in the UI is another question. Maybe `acceptedTos` has the wrong value?

src/libs/state.js
```diff
@@ -15,6 +15,14 @@ export const authStore = Utils.atom({
   cookiesAccepted: undefined
 })
 
+export const userStatus = {
+  unregistered: 'unregistered',
+  createdProfile: 'createdProfile',
+  registeredWithoutTos: 'registeredWithoutTos',
```
marctalbott, line 20: What's the difference between `registeredWithoutTos` and `createdProfile`? In my mental model, a user that has registered but hasn't accepted the ToS is a user that has only created a profile in Terra.

marctalbott (approved):
I think `registeredWithoutTos` and `createdProfile` could be combined into a single value, but I haven't fully thought through the implications of that so I leave it up to you if you want to make that change. Otherwise looks good to me 👍

src/libs/auth.js
cahrens: I think that `acceptTos` will always be true when `enabled` is true when the grace period is over. It turns out that the fact that `registrationStatus` could be `registered` when `acceptedTos` was false was because of the code in Registered.js, not because of this method.

I think the assertion is the better way to go, perhaps after the grace period has ended everywhere though. I could add it to the general terra-ui "cleanup Tos code" ticket.

src/libs/state.js
cahrens: I think they could be collapsed into the single `registeredWithoutTos` state. I was suspicious of what was happening in  `Register.js` and therefore wanted to capture that specific state, but now I'm coming around to collapsing them.

src/libs/auth.js
cahrens: I don't think we want to show the error dialog to the user if these ajax calls fail-- they are things we are doing "behind the scenes" that aren't critical to folks using Terra. At least, that's what I'm reading into the fact that there was no error handling around them.

cahrens: @breilly2 do you have any insight here? It looks like you, Isaac, and Brett wrote most of this code.

src/libs/auth.js
marctalbott: I misread your initial comment as "it is now **possible** for us..." 🤦 . At any rate, this all sounds good to me

src/libs/auth.js
cahrens: Since the rest of this file uses `oldState` and `state`, I maintained it here for consistency.

PR #2992, opened 2022-04-29
[AJ-323] add keys to import jobs so the correct polling jobs dismount

jordansafer:
Fixing AJ-323:
https://broadworkbench.atlassian.net/browse/AJ-323

The issue was, the components for multiple imports were all the same type, ImportStatus, so when the first import completed, React would always dismount the last component (LIFO), even though it was the first component that completed (FIFO). So we would keep polling for the first import to complete indefinitely, and stop polling for the second component. See this doc on reconciliation:
https://reactjs.org/docs/reconciliation.html#recursing-on-children

Solution was to add keys to each ImportStatus, so react would dismount the correct one. See here, three import jobs result in exactly 3 completion notifications (not 300 :) )
Screen Shot 2022-04-29 at 12 24 14 PM

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/components/ImportStatus.js
+++ b/src/components/ImportStatus.js
@@ -13,6 +13,7 @@ import * as Utils from 'src/libs/utils'
 const ImportStatus = () => {
   const jobs = useStore(asyncImportJobStore)
   return h(Fragment, _.map(job => h(ImportStatusItem, {
+    key: job.jobId,
     job,
     onDone: () => {
       asyncImportJobStore.update(_.reject({ jobId: job.jobId }))


davidangb (approved):
did not test it myself, but this change looks correct!

kyuksel (approved):
Thank you for the very nicely explanatory ticket and PR descriptions 🙏🏼 
I love that you were able to fix the issue with only +1 diff  🤩 
Did not test.

sagehen03 (approved):
Great job, Jordan!  I remember learning that child components bound to a collection always need a key back in early 2020; hopefully now I won't forget that info again :)

PR #2993, opened 2022-04-29
[SU-53] Allow deletion of Errored billing projects

MatthewBemis:
I recommend using hermione.owner to test this, as that account has amassed a lot of errored projects over the years.

See UX considerations: https://broadworkbench.atlassian.net/browse/SU-53

List view:
Screen Shot 2022-04-29 at 3 56 16 PM

New menu:
Screen Shot 2022-04-29 at 3 56 29 PM

Deletion modal:
Screen Shot 2022-04-29 at 3 56 36 PM

<!--
Hello, friend!
Thanks!

--!>


cahrens:
Why can you only delete failed ones? What if you wanted to delete a "healthy" billing project?

And can we just show a trash can (or some other delete icon) instead of the menu, given that there is only a single item?

cahrens:
It would be good to extend the billing projects test to do a deletion.

MatthewBemis:
@cahrens the UX around deleting an active billing project is a much more nuanced implementation and the backend doesn't support it too well, particularly in the situation where there are workspaces in the project. I am not sure if it will ever be fully supported at this rate. This is merely being done as a courtesy to our longtime users who may have collected a bunch of these errored projects over the years, rendering their project lists unusable. We recently fixed Rawls support for this [here](https://github.com/broadinstitute/rawls/pull/1645) but we don't want to force users to call the API directly.

And sure, the three dots icon was an explicit choice by Joy but I will double check with her about that versus the trash icon.

petesantos (approved):
Tested with hermione account

MatthewBemis:
@cahrens I confirmed with UX that they would like this to remain a 3-dot icon even though there is only one option

I'll draft this PR until I add the requested test coverage

integration-tests/tests/billing-projects.js
```diff
@@ -89,14 +101,18 @@ const setAjaxMockValues = async (testPage, ownedBillingProjectName, notOwnedBill
     })
   }, _.range(0, numExtraWorkspaces))
 
-  const projectListResult = [{
+  const projectListResult = _.compact([{
```
cahrens, line 104: @MatthewBemis I pushed up a couple of test improvements. The compact is needed here else you end up with `false` as the second of 3 entries, which causes a React rendering warning.

integration-tests/tests/billing-projects.js
```diff
@@ -115,7 +131,13 @@ const setAjaxMockValues = async (testPage, ownedBillingProjectName, notOwnedBill
     email: 'testuser2@example.com', role: 'Owner'
   }]
 
-  return await testPage.evaluate((spendReturnResult, projectListResult, ownedProjectMembersListResult, notOwnedProjectMembersListResult) => {
+  return await testPage.evaluate((spendReturnResult, projectListResult, ownedProjectMembersListResult, notOwnedProjectMembersListResult,
+    ownedBillingProjectName, notOwnedBillingProjectName, erroredBillingProjectName) => {
+
+    const ownedMembersUrl = new RegExp(`api/billing/v2/${ownedBillingProjectName}/members`, 'g')
```
cahrens, line 136: Since we are passing in the project names as variables, hard-coding them in the URLs isn't a good idea.

integration-tests/tests/billing-projects.js
```diff
@@ -130,19 +151,24 @@ const setAjaxMockValues = async (testPage, ownedBillingProjectName, notOwnedBill
...
+        filter: { url: ownedMembersUrl },
         fn: () => () => Promise.resolve(new Response(JSON.stringify(ownedProjectMembersListResult), { status: 200 }))
       },
       {
-        filter: { url: /api\/billing\/v2\/NotOwnedBillingProject\/members$/ },
+        filter: { url: notOwnedMembersUrl },
         fn: () => () => Promise.resolve(new Response(JSON.stringify(notOwnedProjectMembersListResult), { status: 200 }))
       },
+      {
+        filter: { url: erroredBillingProjectUrl, method: 'DELETE' },
```
cahrens, line 162: Specified DELETE just to be clearer.

src/pages/billing/List.js
```diff
@@ -42,7 +43,47 @@ const styles = {
   }
 }
 
-const ProjectListItem = ({ project, project: { roles, status }, isActive }) => {
+const BillingProjectActions = ({ project: { projectName }, loadProjects }) => {
+  const signal = useCancellation()
```
cahrens: I disagree with using cancellation here. If the user triggers a delete operation and then leaves the page, I think they would still expect the delete to happen. We typically use cancellation for GET calls where we don't want to finish them if the user leaves the page.

Looking at other DELETE ajax calls, we do not use cancellation for those.

cahrens (approved):

PR #2995, opened 2022-05-02
[SU-86] Support MN908947.3 SARS-CoV2 reference genome in IGV

MatthewBemis:
https://broadworkbench.atlassian.net/browse/SU-86

Testing this requires jumping through some hoops. The way that I've been testing it is to point my local UI at prod and then use a covid19 workspace that our friends at Theiagen gave me access to. If you'd like to test it out yourself, I can try to get you access. Otherwise, I can hop on a quick call and just demonstrate it. Or neither, up to you.

Screen Shot 2022-05-02 at 2 14 19 PM

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/components/IGVBrowser.js
+++ b/src/components/IGVBrowser.js
@@ -12,6 +12,14 @@ import { useCancellation, useOnMount, withDisplayName } from 'src/libs/react-uti
 import { knownBucketRequesterPaysStatuses, requesterPaysProjectStore } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
+
+const customReferences = [
+  {
+    name: 'MN908947.3',
+    referenceDefinition: { reference: { id: 'sarsCov2RefId.3', indexed: false, fastaURL: 'https://storage.googleapis.com/gcp-public-data--broad-references/sars-cov-2/MN908947.3/nCoV-2019.reference.fasta' } }
+  }
+]
+
 // format for selectedFiles prop: [{ filePath, indexFilePath } }]
 const IGVBrowser = _.flow(
   withDisplayName('IGVBrowser'),
@@ -52,9 +60,6 @@ const IGVBrowser = _.flow(
           const { default: igv } = await import('igv')
           igvLibrary.current = igv
 
-          const sarsCov2RefId = 'MN908947.3'
-          const sarsCov2Ref = { reference: { id: 'sarsCov2RefId.3', indexed: false, fastaURL: 'https://storage.cloud.google.com/firecloud-alerts-dev/nCoV-2019.reference.fasta' } }
-
           const options = {
             genome: refGenome,
             tracks: await Promise.all(_.map(async ({ filePath, indexFilePath }) => {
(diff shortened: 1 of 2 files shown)


Qi77Qi (approved):
@davidangb think you did sth similar before....just tagging you for a review as well..otherwise lgtm

davidangb:
@Qi77Qi et al - I added ASM985889v3, but that was just exposing an already-supported reference genome. This is adding a brand new one, I don't have experience with that.

src/components/IGVBrowser.js
```diff
@@ -12,6 +12,14 @@ import { useCancellation, useOnMount, withDisplayName } from 'src/libs/react-uti
 import { knownBucketRequesterPaysStatuses, requesterPaysProjectStore } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
+
+const customReferences = [
+  {
+    name: 'MN908947.3',
+    referenceDefinition: { reference: { id: 'sarsCov2RefId.3', indexed: false, fastaURL: 'https://storage.googleapis.com/gcp-public-data--broad-references/sars-cov-2/MN908947.3/nCoV-2019.reference.fasta' } }
+  }
+]
```
breilly2: Since this is effectively a key/value map, an object may be a better representation, i.e.:
```
// Additional references supported by Terra that are not included in IGV
const customReferences = {
  'MN908947.3': {
    id: 'sarsCov2RefId.3', indexed: false,
    fastaURL: 'https://storage.googleapis.com/gcp-public-data--broad-references/sars-cov-2/MN908947.3/nCoV-2019.reference.fasta'
  }
}
```
This makes finding the reference by name a simple attribute reference instead of a `_.find`.

src/components/IGVBrowser.js
```diff
@@ -66,8 +74,12 @@ const IGVBrowser = _.flow(
             }, selectedFiles))
           }
 
+          const customReference = _.find({ name: refGenome }, customReferences)
+
+          const mergedOptions = customReference ? _.merge(options, customReference.referenceDefinition) : options
```
breilly2: This works. My only thought is that I had to run an experiment (in a console, though it could have been in my mind if I was more awake and had more recent experience with `_.merge`) to see the final shape of `mergedOptions`. An alternative way of expressing this might be (untested, but I think it works; also incorporating my other suggestion above):
```
// Additional references supported by Terra that are not included in IGV
const customReferences = {
  'MN908947.3': {
    id: 'sarsCov2RefId.3', indexed: false,
    fastaURL: 'https://storage.googleapis.com/gcp-public-data--broad-references/sars-cov-2/MN908947.3/nCoV-2019.reference.fasta'
  }
}

...

    const customReference = customReferences[refGenome]
    const options = {
      genome: refGenome,
      reference: customReference,
      tracks: { ... }
    }
```
No `_.merge` needed. What do you think?

breilly2 (commented):
Code is fine and I believe you that it works. I have some readability suggestions that I'm curious to get your thoughts about.

src/components/IGVBrowser.js
MatthewBemis: I think this is nicer, thanks for this! I've incorporated it into 45a72a6241aadd63636b436808b6d14a5eb8d86d

MatthewBemis: (and btw I tested the suggestion and it does indeed work)

breilly2 (approved):
👍 I love how little code is needed to make this work!

PR #2996, opened 2022-05-02
[QA-1839] Attempt to fix sporadic timeout failure during login

aweng98:
Deal with sporadic `Loading Terra` spinner timeout failure, fix by increase the wait `timeout` to 60 seconds from 30 seconds. Also, remove boilerplate code around `signIntoTerra()`.

### DIFF
--- a/integration-tests/tests/import-dockstore-workflow.js
+++ b/integration-tests/tests/import-dockstore-workflow.js
@@ -24,7 +24,7 @@ const testImportDockstoreWorkflowFn = _.flow(
   withDockstoreCheck
 )(async ({ page, testUrl, token, workspaceName }) => {
   const url = `${testUrl}/#import-tool/dockstore/${testWorkflowIdentifier}`
-  await signIntoTerra(page, { token, url })
+  await signIntoTerra(page, { token, testUrl: url })
 
   await findText(page, 'workflow TopMedVariantCaller')
   await select(page, 'Select a workspace', workspaceName)
--- a/integration-tests/tests/import-dockstore-workflow.js
+++ b/integration-tests/tests/import-dockstore-workflow.js
@@ -23,8 +23,8 @@ const testImportDockstoreWorkflowFn = _.flow(
   withUserToken,
   withDockstoreCheck
 )(async ({ page, testUrl, token, workspaceName }) => {
-  await page.goto(`${testUrl}/#import-tool/dockstore/${testWorkflowIdentifier}`)
-  await signIntoTerra(page, token)
+  const url = `${testUrl}/#import-tool/dockstore/${testWorkflowIdentifier}`
+  await signIntoTerra(page, { token, url })
   await dismissNotifications(page)
   await findText(page, 'workflow TopMedVariantCaller')
   await select(page, 'Select a workspace', workspaceName)
--- a/integration-tests/tests/import-dockstore-workflow.js
+++ b/integration-tests/tests/import-dockstore-workflow.js
@@ -22,9 +22,9 @@ const testImportDockstoreWorkflowFn = _.flow(
   withWorkspace,
   withUserToken,
   withDockstoreCheck
-)(async ({ page, testUrl, token, workspaceName }) => {
-  const url = `${testUrl}/#import-tool/dockstore/${testWorkflowIdentifier}`
-  await signIntoTerra(page, { token, testUrl: url })
+)(async ({ page, testUrl: testUrlRoot, token, workspaceName }) => {
+  const testUrl = `${testUrlRoot}/#import-tool/dockstore/${testWorkflowIdentifier}`
+  await signIntoTerra(page, { token, testUrl })
 
   await findText(page, 'workflow TopMedVariantCaller')
   await select(page, 'Select a workspace', workspaceName)
--- a/integration-tests/tests/import-dockstore-workflow.js
+++ b/integration-tests/tests/import-dockstore-workflow.js
@@ -1,7 +1,7 @@
 const _ = require('lodash/fp')
 const fetch = require('node-fetch')
 const { withWorkspace } = require('../utils/integration-helpers')
-const { click, clickable, dismissNotifications, findText, select, signIntoTerra } = require('../utils/integration-utils')
+const { click, clickable, findText, select, signIntoTerra } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
 
@@ -25,7 +25,7 @@ const testImportDockstoreWorkflowFn = _.flow(
 )(async ({ page, testUrl, token, workspaceName }) => {
   const url = `${testUrl}/#import-tool/dockstore/${testWorkflowIdentifier}`
   await signIntoTerra(page, { token, url })
-  await dismissNotifications(page)
+
   await findText(page, 'workflow TopMedVariantCaller')
   await select(page, 'Select a workspace', workspaceName)
   await click(page, clickable({ text: 'Import' }))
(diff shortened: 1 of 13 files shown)


integration-tests/tests/import-dockstore-workflow.js
```diff
@@ -23,9 +23,9 @@ const testImportDockstoreWorkflowFn = _.flow(
   withUserToken,
   withDockstoreCheck
 )(async ({ page, testUrl, token, workspaceName }) => {
-  await page.goto(`${testUrl}/#import-tool/dockstore/${testWorkflowIdentifier}`)
-  await signIntoTerra(page, token)
-  await dismissNotifications(page)
+  const url = `${testUrl}/#import-tool/dockstore/${testWorkflowIdentifier}`
+  await signIntoTerra(page, { token, testUrl: url })
```
dmohs: The usage here makes me think that `testUrl` is a misnomer in the parameter list. If you did:
```javascript
)(async ({ page, testUrl: testUrlRoot, token, workspaceName }) => {
```
then you could:
```suggestion
  const testUrl = `${testUrlRoot}/#import-tool/dockstore/${testWorkflowIdentifier}`
  await signIntoTerra(page, { token, testUrl })
```

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -29,10 +30,9 @@ const testRunWorkflowOnSnapshotFn = _.flow(
     return
   }
   // IMPORT SNAPSHOT
-  await page.goto(`${testUrl}/#import-data?url=${dataRepoUrlRoot}&snapshotId=${snapshotId}&snapshotName=${snapshotName}&format=snapshot`)
-  await signIntoTerra(page, token)
-  await dismissNotifications(page)
-  await waitForNoSpinners(page)
+  const url = `${testUrl}/#import-data?url=${dataRepoUrlRoot}&snapshotId=${snapshotId}&snapshotName=${snapshotName}&format=snapshot`
+  await signIntoTerra(page, { token, testUrl: url })
```
dmohs: Similarly here if you think it makes sense.

dmohs (approved):

integration-tests/tests/register-user.js
```diff
@@ -1,13 +1,13 @@
 const { withUser } = require('../utils/integration-helpers')
-const { fillIn, findText, click, clickable, input, signIntoTerra, dismissNotifications } = require('../utils/integration-utils')
+const { fillIn, findText, click, clickable, input, signIntoTerra } = require('../utils/integration-utils')
+const { fillInReplace } = require('../utils/integration-utils')
 
 
 const testRegisterUserFn = withUser(async ({ page, testUrl, token }) => {
-  await page.goto(testUrl)
+  await page.goto(testUrl, { waitUntil: ['load', 'domcontentloaded', 'networkidle0'], timeout: 60 * 1000 })
```
kyuksel: How would you feel about factoring out 

```
{ waitUntil: ['load', 'domcontentloaded', 'networkidle0'], timeout: 60 * 1000 }
```

by adding it into `integration-utils` with something like

```javascript
const waitUntilLoadedOrTimeout = timeout => ({ waitUntil: ['load', 'domcontentloaded', 'networkidle0'], timeout })
```

?

integration-tests/utils/integration-utils.js
```diff
@@ -169,10 +169,15 @@ const dismissNotifications = async page => {
   return !!notificationCloseButtons.length && delay(1000) // delayed for alerts to animate off
 }
 
-const signIntoTerra = async (page, token) => {
-  await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true })
+const signIntoTerra = async (page, { token, testUrl }) => {
+  if (testUrl) {
+    await page.goto(testUrl, { waitUntil: ['load', 'domcontentloaded', 'networkidle0'], timeout: 60 * 1000 })
+  }
```
kyuksel: What's the semantics of the predicate (`if (testUrl)`) here? When is a caller of `signIntoTerra` supposed to pass in a `testUrl` and when not?

integration-tests/utils/integration-utils.js
```diff
@@ -169,10 +169,15 @@ const dismissNotifications = async page => {
...
+  await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout: 60 * 1000 })
+  await waitForNoSpinners(page)
+  await page.evaluate(token => window.forceSignIn(token), token)
```
kyuksel, line 176: Could you explain why the line above was added since it doesn't seem to exist in the current version of `dev`?

kyuksel (commented):
Looks good overall. A few questions and comments inline plus pending Mohs' comments to be addressed.

kyuksel:
@aweng98 I ran `test-flakes` with `find-workflow` with `RUNS=100` against this branch and `dev` branch and it didn't indicate a regression, which is great. It'd give further confidence if you could do the same for at least a couple of other tests. `preview-dataset` may be an especially good test to use since it seems to have flaked out on the [CircleCI run](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9240/workflows/1be1971b-a905-4043-aafd-95c517d77937/jobs/41624).

integration-tests/utils/integration-utils.js
cahrens: It was below at line 175.

integration-tests/tests/register-user.js
cahrens: I agree that it would be nice to have a method for this-- it appears in a few different files.

integration-tests/tests/import-dockstore-workflow.js
aweng98: Excellent suggestion.

integration-tests/tests/register-user.js
aweng98: Excellent suggestion.

integration-tests/tests/run-workflow-on-snapshot.js
aweng98: Done

integration-tests/utils/integration-utils.js
aweng98: Intention is to keep the origin design of function. Tests can still use `signIntoTerra` without passing a URL. Optional `testUrl` parameter is meant to reduce boilerplate code `goto`, it's not meant to be a constraint.

integration-tests/utils/integration-utils.js
aweng98: Do you still have question about it?

kyuksel: I see now. Thanks.

integration-tests/utils/integration-utils.js
```diff
@@ -169,10 +169,15 @@ const dismissNotifications = async page => {
   return !!notificationCloseButtons.length && delay(1000) // delayed for alerts to animate off
 }
 
-const signIntoTerra = async (page, token) => {
-  await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true })
+const signIntoTerra = async (page, { token, testUrl }) => {
+  if (testUrl) {
+    await page.goto(testUrl, waitUntilLoadedOrTimeout(60 * 1000))
+  }
```
kyuksel: We typically one-line simple conditionals, like so:

```suggestion
  !!testUrl && await page.goto(testUrl, waitUntilLoadedOrTimeout(60 * 1000))
```

`!!` is to be explicit about our intention of coercing a non-boolean (`testUrl` in this case) into a truthy/falsy.

kyuksel (commented):
One last comment 🙈

cahrens (approved):

integration-tests/utils/integration-utils.js
aweng98: Thanks.

PR #2998, opened 2022-05-03
[SU-36] Allow for data tables to be renamed

MatthewBemis:
New menu option "Rename table":
Screen Shot 2022-05-04 at 3 59 12 PM

New modal:

Screen Shot 2022-05-04 at 3 59 03 PM

https://broadworkbench.atlassian.net/browse/SU-36

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/components/data/RenameTableModal.js
+++ b/src/components/data/RenameTableModal.js
@@ -1,0 +1,25 @@
+import { useState } from 'react'
+import { h } from 'react-hyperscript-helpers'
+import { ButtonPrimary } from 'src/components/common'
+import Modal from 'src/components/Modal'
+import { Ajax } from 'src/libs/ajax'
+import colors from 'src/libs/colors'
+import { reportError } from 'src/libs/error'
+import * as Utils from 'src/libs/utils'
+
+
+const RenameTableModal = ({ onDismiss, selectedDataType, workspace }) => {
+  // State
+  const [renaming, setRenaming] = useState(false)
+
+  return h(Modal, {
+    onDismiss,
+    title: 'Rename Data Table',
+    okButton: h(ButtonPrimary, {
+      disabled: renaming,
+      onClick: console.log(selectedDataType)
+    }, ['Copy'])
+  }, [])
+}
+
+export default RenameTableModal
--- a/src/components/data/RenameTableModal.js
+++ b/src/components/data/RenameTableModal.js
@@ -34,7 +34,7 @@ const RenameTableModal = ({ onDismiss, onUpdateSuccess, namespace, name, selecte
         withErrorReporting('Error renaming data table.'),
         Utils.withBusyState(setRenaming)
       )(async () => {
-        Ajax().Metrics.captureEvent(Events.workspaceDataRenameTable, { oldName: selectedDataType, newName })
+        await Ajax().Metrics.captureEvent(Events.workspaceDataRenameTable, { oldName: selectedDataType, newName })
         await Ajax().Workspaces.workspace(namespace, name).renameEntityType(selectedDataType, newName)
         onUpdateSuccess()
       })
--- a/src/components/data/RenameTableModal.js
+++ b/src/components/data/RenameTableModal.js
@@ -1,6 +1,6 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
-import { h } from 'react-hyperscript-helpers'
+import { div, h } from 'react-hyperscript-helpers'
 import { ButtonPrimary, IdContainer, spinnerOverlay } from 'src/components/common'
 import { ValidatedInput } from 'src/components/input'
 import Modal from 'src/components/Modal'
@@ -40,6 +40,7 @@ const RenameTableModal = ({ onDismiss, onUpdateSuccess, namespace, name, selecte
       })
     }, ['Rename'])
   }, [h(IdContainer, [id => h(Fragment, [
+    div('Workflow configurations that reference the current table name will need to be updated manually.'),
     h(FormLabel, { htmlFor: id, required: true }, ['New Name']),
     tableNameInput({
       inputProps: {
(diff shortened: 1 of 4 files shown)


src/components/data/RenameTableModal.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const signal = useCancellation()
+
+  return h(Modal, {
+    onDismiss,
+    title: 'Rename Data Table',
+    okButton: h(ButtonPrimary, {
+      disabled: renaming,
+      onClick: async () => {
+        setRenaming(true)
```
cahrens: You should likely wrap this in withErrorReporting instead of doing the try/catch (and that makes sure that the Metrics call is also wrapped).

MatthewBemis: Thank you, that's much cleaner. Addressed in 4b3ab4bace2b9c27e8212700ab395fb3136f3cb3

src/libs/events.js
```diff
@@ -53,6 +53,7 @@ const eventsList = {
   workspaceDataOpenWithNotebook: 'workspace:data:notebook',
   workspaceDataImport: 'workspace:data:import',
   workspaceDataUpload: 'workspace:data:upload',
+  workspaceDataRenameTable: 'workspace:data:renametable',
```
breilly2: Nit: I prefer `renameTable` or `rename-table`. Doesn't matter too much as long as the Mixpanel lexicon is updated.

src/pages/workspaces/workspace/Data.js
```diff
@@ -716,7 +733,8 @@ const WorkspaceData = _.flow(
                   after: isDataTabRedesignEnabled() && h(DataTableActions, {
                     tableName: type,
                     rowCount: typeDetails.count,
-                    workspace
+                    workspace,
+                    onSuccess: () => loadMetadata()
```
jordansafer: nit: Could we use a more specific name than "onSuccess"? Maybe: onUpdateSuccess()? Since this only called by rename metadata success, and not by the other menu items.

jordansafer (approved):

src/components/data/RenameTableModal.js
```diff
@@ -0,0 +1,59 @@
...
+    onDismiss,
+    title: 'Rename Data Table',
+    okButton: h(ButtonPrimary, {
+      disabled: renaming,
+      onClick: _.flow(
+        Utils.withBusyState(setRenaming),
+        withErrorReporting('Error renaming data table.')
+      )(async () => {
+        await Ajax().Metrics.captureEvent(Events.workspaceDataRenameTable, { oldName: selectedDataType, newName })
+        await Ajax(signal).Workspaces.workspace(namespace, name).renameEntityType(selectedDataType, newName)
```
cahrens: I don't think you really want to use cancellation here. I think that anytime you are changing the server state, cancellation is not appropriate. When you are calling GET and don't need the response if the user leaves the page, then use cancellation.

MatthewBemis: You're right- I definitely don't want it here!

MatthewBemis:
fyi I am trying to track down the cause of a react console warning that I must have missed when testing this out (or I introduced it in a staged commit that I haven't pushed yet- tbd)

src/components/data/RenameTableModal.js
MatthewBemis: Addressed in 15d3b06a8a3b60caee2a2eefbec76112d4e6e1e6

src/pages/workspaces/workspace/Data.js
MatthewBemis: Sure, addressed in 4cb07b8d2201e7b85f9d751de33e4a62fdc646e0

src/components/data/RenameTableModal.js
```diff
@@ -0,0 +1,56 @@
...
+  return h(Modal, {
+    onDismiss,
+    title: 'Rename Data Table',
+    okButton: h(ButtonPrimary, {
+      disabled: renaming,
+      onClick: _.flow(
+        withErrorReporting('Error renaming data table.'),
+        Utils.withBusyState(setRenaming)
+      )(async () => {
+        await Ajax().Metrics.captureEvent(Events.workspaceDataRenameTable, { oldName: selectedDataType, newName })
```
breilly2, line 37: This will record an event even if the rename fails. This is fine as long as it's understood that the event represents an _intent_ to rename a data table. My guess is that the most common errors would be an invalid or duplicate table name?

breilly2:
Does the table rename also update bindings in method configs? If not, I think it's worth putting a note in the modal mentioning that existing method configs (and other references) may need to be updated.

breilly2 (approved):
👍  Looks good overall. My only real concern is the UX around what happens with references in existing method configs.

src/components/data/RenameTableModal.js
MatthewBemis: That's a good question. I was thinking it'd be good to capture intent but I will double-check with those who have more of a stake in these metrics than I do

And yes, those would be the typical failure reasons

MatthewBemis:
@breilly2 they are not updated (correct me if I'm wrong @jordansafer @sagehen03 ). I'll add some language there to give the user a heads up

src/libs/events.js
MatthewBemis: Addressed in 6cdf3966a244a3f349271d2824fcef40978c55b6

davidangb:
You are correct - method config bindings are not updated. That's a good catch!

src/components/data/RenameTableModal.js
jordansafer: Chiming in here on behalf of the metrics, we added these to capture intent to see commonly used operations. +1 the current implementation, we could refine/debate some edgecases here, but it's over-optimizing for how we're using the metrics right now (just using them as a rough guide for what gets clicked)

davidangb:
heads up that https://github.com/broadinstitute/rawls/pull/1747 may move this API, making this PR invalid - let's coordinate!

MatthewBemis:
@davidangb thank you for the heads up! I will hold off on merging this one until the dust has settled with broadinstitute/rawls#1747

PR #2999, opened 2022-05-03
[DC-227] Add cookie reject modal

s-rubenstein:
Tested by signing in, rejecting cookies, then hitting okay. Also tested using the same flow while signed out.

### DIFF
--- a/src/components/CookieRejectModal.js
+++ b/src/components/CookieRejectModal.js
@@ -11,6 +11,8 @@ const CookieRejectModal = () => {
     title: 'Cookies are required for Terra',
     showCancel: false,
     showX: false,
+    shouldCloseOnOverlayClick: false,
+    shouldCloseOnEsc: false,
     onDismiss: () => authStore.update(_.set('cookiesAccepted', true))
   }, ['By clicking OK, you agree to use cookies in Terra.'])
 }
--- a/src/components/CookieRejectModal.js
+++ b/src/components/CookieRejectModal.js
@@ -12,7 +12,7 @@ const CookieRejectModal = () => {
     showCancel: false,
     showX: false,
     onDismiss: () => authStore.update(_.set('cookiesAccepted', true))
-  }, ['By clicking \'OK\', you agree to use cookies in Terra.'])
+  }, ['By clicking OK, you agree to use cookies in Terra.'])
 }
 
 export default CookieRejectModal
--- a/src/components/CookieRejectModal.js
+++ b/src/components/CookieRejectModal.js
@@ -1,0 +1,18 @@
+import _ from 'lodash/fp'
+import { h } from 'react-hyperscript-helpers'
+import Modal from 'src/components/Modal'
+import { useStore } from 'src/libs/react-utils'
+import { authStore } from 'src/libs/state'
+
+
+const CookieRejectModal = () => {
+  const { cookiesAccepted } = useStore(authStore)
+  return cookiesAccepted === false && h(Modal, {
+    title: 'Cookies are required for Terra',
+    showCancel: false,
+    showX: false,
+    onDismiss: () => authStore.update(_.set('cookiesAccepted', true))
+  }, ['By clicking \'OK\', you agree to use cookies in Terra.'])
+}
+
+export default CookieRejectModal
--- a/src/components/CookieWarning.js
+++ b/src/components/CookieWarning.js
@@ -4,12 +4,11 @@ import { aside, div, h } from 'react-hyperscript-helpers'
 import { Transition } from 'react-transition-group'
 import { ButtonPrimary, ButtonSecondary, Link } from 'src/components/common'
 import { Ajax } from 'src/libs/ajax'
-import { signOut } from 'src/libs/auth'
 import colors from 'src/libs/colors'
 import { getAppName } from 'src/libs/logos'
 import * as Nav from 'src/libs/nav'
 import { useCancellation, useStore } from 'src/libs/react-utils'
-import { authStore } from 'src/libs/state'
+import { authStore, cookieReadyStore } from 'src/libs/state'
 
 
 export const cookiesAcceptedKey = 'cookiesAccepted'
@@ -55,7 +54,9 @@ const CookieWarning = () => {
       const cookieName = eqPos > -1 ? cookie.substr(0, eqPos) : cookie
       document.cookie = `${cookieName}=;expires=Thu, 01 Jan 1970 00:00:00 GMT`
     }, cookies)
-    signOut()
+
+    cookieReadyStore.reset()
+    sessionStorage.clear()
   }
 
   return h(Transition, {
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -114,7 +114,7 @@ export const initializeAuth = _.memoize(async () => {
         fenceStatus: isSignedIn ? state.fenceStatus : {},
         // Load whether a user has input a cookie acceptance in a previous session on this system,
         // or whether they input cookie acceptance previously in this session
-        cookiesAccepted: isSignedIn ? state.cookiesAccepted || getLocalPrefForUserId(user.getId(), cookiesAcceptedKey) : false,
+        cookiesAccepted: isSignedIn ? state.cookiesAccepted || getLocalPrefForUserId(user.getId(), cookiesAcceptedKey) : undefined,
         isTimeoutEnabled: isSignedIn ? state.isTimeoutEnabled : undefined,
         user: {
           token: authResponse ? authResponse.access_token : undefined,
--- a/src/pages/Main.js
+++ b/src/pages/Main.js
@@ -4,6 +4,7 @@ import { h } from 'react-hyperscript-helpers'
 import ReactNotification from 'react-notifications-component'
 import AuthContainer from 'src/components/AuthContainer'
 import ConfigOverridesWarning from 'src/components/ConfigOverridesWarning'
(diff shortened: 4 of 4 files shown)


src/components/CookieRejectModal.js
```diff
@@ -0,0 +1,18 @@
+import _ from 'lodash/fp'
+import { h } from 'react-hyperscript-helpers'
+import Modal from 'src/components/Modal'
+import { useStore } from 'src/libs/react-utils'
+import { authStore } from 'src/libs/state'
+
+
+const CookieRejectModal = () => {
+  const { cookiesAccepted } = useStore(authStore)
+  return cookiesAccepted === false && h(Modal, {
```
s-rubenstein, line 10: This verbiage being okay/better than we have now has been run by Lou

src/components/CookieWarning.js
```diff
@@ -55,7 +54,9 @@ const CookieWarning = () => {
       const cookieName = eqPos > -1 ? cookie.substr(0, eqPos) : cookie
       document.cookie = `${cookieName}=;expires=Thu, 01 Jan 1970 00:00:00 GMT`
     }, cookies)
-    signOut()
```
s-rubenstein, line 58: We no longer want to sign out, because we want users to be able re-accept cookies, and if they are signed out we don't know which user it is who has accepted cookies.

src/libs/auth.js
```diff
@@ -114,7 +114,7 @@ export const initializeAuth = _.memoize(async () => {
         fenceStatus: isSignedIn ? state.fenceStatus : {},
         // Load whether a user has input a cookie acceptance in a previous session on this system,
         // or whether they input cookie acceptance previously in this session
-        cookiesAccepted: isSignedIn ? state.cookiesAccepted || getLocalPrefForUserId(user.getId(), cookiesAcceptedKey) : false,
+        cookiesAccepted: isSignedIn ? state.cookiesAccepted || getLocalPrefForUserId(user.getId(), cookiesAcceptedKey) : undefined,
```
s-rubenstein, line 117: This now defaults to undefined, rather than false, because `undefined` and `false` have different semantic meaning.
`false` means they have explicitly refused cookies.
`undefined` means that we don't know whether they have accepted or refused cookies.

kyuksel (approved):
Code changes LGTM 👍🏼 
Didn't test.

src/components/CookieRejectModal.js
```diff
@@ -0,0 +1,18 @@
...
+    title: 'Cookies are required for Terra',
+    showCancel: false,
+    showX: false,
+    onDismiss: () => authStore.update(_.set('cookiesAccepted', true))
```
s-rubenstein, line 16: My one problem with this is that clicking off the modal seems to also close it, and I kinda want `OK` to be the only clickable option, but I don't know if our modal is structured for that ATM

src/components/CookieRejectModal.js
```diff
@@ -0,0 +1,18 @@
...
+  }, ['By clicking \'OK\', you agree to use cookies in Terra.'])
```
cahrens: I'm not sure single quotes are the best thing to use here. Perhaps check with user ed (unless they recommended them)?

src/components/CookieRejectModal.js
```diff
@@ -0,0 +1,18 @@
...
+import Modal from 'src/components/Modal'
+import { useStore } from 'src/libs/react-utils'
+import { authStore } from 'src/libs/state'
+
+
+const CookieRejectModal = () => {
+  const { cookiesAccepted } = useStore(authStore)
+  return cookiesAccepted === false && h(Modal, {
+    title: 'Cookies are required for Terra',
+    showCancel: false,
```
cahrens, line 12: Note that you can still dismiss the dialog by pressing Escape (it seems like you are trying to prevent the dialog being dismissed, but it's not clear to me what you intend the experience to be for someone who does not want to accept the cookies).

src/components/CookieRejectModal.js
cahrens: Escape also closes it. I personally feel like there should be an option for someone to say "I really don't want to use Cookies… and I understand that means I can't use Terra".

src/components/CookieWarning.js
cahrens: But if we don't sign them out, doesn't that mean that they can continue using Terra even without accepting the use of cookies?

src/libs/auth.js
cahrens: Where does `cookiesAccepted` get set to `false`?

src/components/CookieRejectModal.js
kyuksel: If it's not something specified in these [UX Guidelines page](https://www.sketch.com/s/cca3a834-985b-4165-bbae-21f615aebd2d/a/PqzMVj) or if any objections/questions about it, I'd recommend posting on `#dsp-ui-ux` so we can all be on the same page.

cahrens: It just looked odd to me when I saw it in the dialog, but if Lou already 'OK'd' this usage, ignore me.

s-rubenstein: I ran this screen by the user design team. Let me double check what we use in other dialogs though.

src/components/CookieRejectModal.js
s-rubenstein: So, the short version is that we have said that you can't use the site without accepting cookies. Really the intended experience for someone who does not want to accept the cookies is, sadly, to just close the site atm. We, as a Terra engineering team, probably need to have a discussion about what parts of the site you _can_ use without cookies, but either way, this is better than we have now.

src/components/CookieRejectModal.js
lu-c:
```suggestion
    shouldCloseOnOverlayClick: false,
    shouldCloseOnEsc: false,
    onDismiss: () => authStore.update(_.set('cookiesAccepted', true))
```
We can prevent the modal from doing this by adding to the properties!

src/libs/auth.js
lu-c: I think just in that line - it'll be false if `state.cookiesAccepted` or `getLocalPrefForUserId(...)` is false

src/components/CookieRejectModal.js
s-rubenstein: Looks like the existing pattern is no quotes. I don't know if I agree with that, personally, but for now I'll go to what we do across the site.

lu-c:
I don't really know how I feel about this solution, because if we have to force users to hit the accept button in order to use the site, maybe we shouldn't have a reject button in the first place. It kind of feels bad to have a "accept/reject" option that then tells you in a modal that you HAVE to choose accept anyway. Does that make sense? What do you think?

src/components/CookieWarning.js
s-rubenstein: If we use the modal and block clicks with the overlay, they either need to jump through some large hoops to keep using Terra (go into the console and delete the modal, which I'd argue is equivalent to hitting ok), or agree to cookies. So it blocks use of the site through a different mechanism

src/libs/auth.js
s-rubenstein: It also gets set to false when a user clicks `reject` on the banner. We only save accept/reject for a user, not on the browser as a whole.

src/components/CookieRejectModal.js
s-rubenstein: @cahrens I agree - this is meant as an iterative improvement over what we have right now. If we want the site to be usable without cookies, that is a _behavior_ change over what has been decided for Terra UI, and would require us to design that feature. (which would be a sizable amount of work). I _definitely_ wouldn't want to block this PR on figuring out how to make Terra usable for non-cookie users.

s-rubenstein:
> I don't really know how I feel about this solution, because if we have to force users to hit the accept button in order to use the site, maybe we shouldn't have a reject button in the first place. It kind of feels bad to have a "accept/reject" option that then tells you in a modal that you HAVE to choose accept anyway. Does that make sense? What do you think?

@lu-c - I agree. I wish we had some sort of structure that let users use the site without cookies. We don't right now. The other option would be to remove the reject button, but then we would need to have some sort of design that talks about cookies being mandatory for the site, which we don't currently. We could do that, but I think that the _smallest_ improvement we could make that addresses the problem of "button has no feedback" is to show the modal. From there, we can get UX, Compliance, Product, and Engineering into a room and actually discuss what the solution should be. But I think that without all the stakeholders in the room to have the conversation, I don't want to guess at what we can do for each of those groups/get into the business of designing the solution in a ticket.

kyuksel:
> From there, we can get UX, Compliance, Product, and Engineering into a room and actually discuss what the solution should be. But I think that without all the stakeholders in the room to have the conversation, I don't want to guess at what we can do for each of those groups/get into the business of designing the solution in a ticket.

Makes sense to me. Could you create a ticket for that?

src/components/CookieRejectModal.js
cahrens: It is better now that Escape and clicking outside do not dismiss the dialog.

src/components/CookieRejectModal.js
cahrens: It is better now that Escape and clicking outside do not dismiss the dialog.

cahrens (approved):

s-rubenstein:
@kyuksel https://broadworkbench.atlassian.net/browse/DC-336

lu-c (approved):

fboulnois (approved):

PR #3004, opened 2022-05-05
[QA-1839] Custom app.yaml for staging and alpha env for automation testing

aweng98:
Terra UI app sets `default_expiration` to 2 min for all env. `app.test.yaml` sets `default_expiration` to 30s.

Change:
Update CircleCI `config.yaml` `deploy-env` job to: 
  - If env === staging || alpha, `app.yaml` is replaced by `app.test.yaml`.

Reason:
If terra-ui integration tests have started to load Terra UI immediately after a new app version is deployed, test can fail because Google appengine likely sends cached static files to the client browser. Cached static files will cause terra UI hanging with `Loading Terra...` spinner.

For details on failure investigation, notes is in this [doc](https://docs.google.com/document/d/1BZdNjvAUKuKjqwm7l5Hsr4O5AsiHcZFucoh4cmMdJUo/edit).


.circleci/config.yml
```diff
@@ -74,7 +74,8 @@ commands:
         type: env_var_name
         default: "SA_KEY_JSON"
       env:
-        type: string
+        type: enum
+        enum: ["dev", "staging", "alpha", "perf"]
```
aweng98, line 78: Enum prevents someone sets an invalid `env` value.

.circleci/config.yml
```diff
@@ -92,6 +93,13 @@ commands:
                   fi
       - attach_workspace:
           at: .
+      - when:
+          condition:
+            or:
+              - equal: [ "alpha", << parameters.env >> ]
+              - equal: [ "staging", << parameters.env >> ]
+          steps:
+            - run: cp app.test.yaml app.yaml
```
aweng98: Replace `app.yaml` if env is alpha or staging.

.circleci/config.yml
```diff
@@ -106,8 +114,10 @@ commands:
                 command: |
                   CIRCLE_PR_NUMBER="${CIRCLE_PR_NUMBER:-${CIRCLE_PULL_REQUEST##*/}}"
                   if [[ ! -z "$CIRCLE_PR_NUMBER" ]]; then
-                    gcloud app deploy --project=bvdp-saturn-<< parameters.env >> \
-                      --version="pr-$(expr "$CIRCLE_PR_NUMBER" % $NUM_PR_SITES)" --no-promote --quiet
+                    gcloud app deploy \
+                      --project=bvdp-saturn-<< parameters.env >> \
+                      --version="pr-$(expr "$CIRCLE_PR_NUMBER" % $NUM_PR_SITES)" \
+                      --no-promote --quiet
```
aweng98, line 122: Formatting: Each flag starts on a newline.

.circleci/config.yml
breilly2: Instead of having a separate copy of `app.yaml` that could get out-of-date, How about scripting a modification of `app.yaml` in-place? This `sed` command should do the trick:
```
sed -e 's/default_expiration: ".*"/default_expiration: "0m"/' -i .orig app.yaml
```

petesantos (commented):
This looks good - once the comment @breilly2 left is addressed I will approve. It looks like we can get away without an additional yaml

.circleci/config.yml
kyuksel: `0m` (or `0s`) feels preferable to me over `30s` in case `30s` is long enough for running into the page not loading issue.

kyuksel:
Thinking out loudly: Do we expect no caching of static files to result in noticeable load time degradation, and if so, would we catch it with integration testing? 🤔

kyuksel:
It'd be good to run `test-flakes` on at least a couple of tests with `RUNS=100` or so to gain more confidence that the change won't introduce flakiness regressions.

cahrens:
> It'd be good to run `test-flakes` on at least a couple of tests with `RUNS=100` or so to gain more confidence that the change won't introduce flakiness regressions.

Am I correct that these changes are just about deployment? If so, I don't think that `test-flakes` (which we run locally against an already deployed version of terra) is relevant.

kyuksel:
> Am I correct that these changes are just about deployment? If so, I don't think that test-flakes (which we run locally against an already deployed version of terra) is relevant.

I thought setting [default_expiration](https://cloud.google.com/appengine/docs/standard/php/config/appref#default_expiration) to `0` would cause the browser to fetch static files every time which could cause performance issues in general and not just right after a new app deployment. Is that not the case?

.circleci/config.yml
aweng98: `sed`is a great suggestion. I'll change.

aweng98 (commented):
Hi all, 
FYI, I used `30s` because that's approximately how long between deployed and tests started.
Regarding to set it to `0s`, I will change to `0s`. Pages may load few seconds slower as result. But we don't care about "user experience" (tinny longer time) because it's tests that are running. And we don't measure page performance at this time. And we only do this to staging and alpha (test env). I have not seen a big increase in CI job's run time during PR testing. 
We're making a tradeoff between load time and test reliability. Personally, I prefer test reliability over UI loading performance. Changing `default_expiration` time will have no affect on test reliability. It won't reduce or increase test flakiness.

breilly2 (approved):

kyuksel (approved):
Excited to see if this will fix our alpha/staging-specific failures 😄 

Integration testing was failing. Updated the PR branch with the latest on `dev`, which also triggered a new run. Let's see if it'll pass this time.

kyuksel:
@aweng98 Note the strange logging on the failure with the previous run of the integration test `workspace-dashboard` on your PR branch. 

image

From the test log:
image

[CircleCI link](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9378/workflows/f468ec35-12f6-4948-bf72-96c8930e7308/jobs/42018/tests)

aweng98:
Yeah, test is flaky. But that failure is not new as I have already seen it before. Issue is in the framework code, no error and screenshot when test fails.

aweng98:
Merge PR now. 🤞 it improves test stability.

cahrens:
FYI, I am not aware of flakiness in `google-workspace`. Unfortunately there is no console error indicating what caused the failure. :(

PR #3006, opened 2022-05-06
[SU-104] Store column settings as json

MatthewBemis:
The motivation for this change is that a power-user has reached a point where the JSON string that we use for `columnSettings` has exceeded the 64kb limit for the `value_string` column in the Rawls database (it is a `TEXT` column). This can happen when there are a lot of columns, a lot of tables, a lot of different views saved, or some combination of all of that. In this PR, we coerce the `columnSettings` into a JSON object rather than a JSON string. Rawls stores JSON workspace attributes in a separate column- one that is `LONGTEXT` and able to support up to 4gb of data. 

The best way to test this is:

1) Save a column view for a workspace on dev
2) View that workspace column view on the PR deployment (this verifies that we are backwards compatible with strings)
3) Save a new column view using the PR deployment
4) Ensure that the column view works correctly on the PR deployment

You can manage column settings by clicking the Gear icon visible on the top right of a table in the Data tab of a workspace.

Also see this PR for some screenshots of what the modal looks like: https://github.com/DataBiosphere/terra-ui/pull/2878

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/components/data/SavedColumnSettings.js
+++ b/src/components/data/SavedColumnSettings.js
@@ -78,7 +78,6 @@ const useSavedColumnSettings = ({ workspaceId, snapshotName, entityMetadata, ent
   const updateSavedColumnSettings = async allColumnSettings => {
     const { namespace, name } = workspaceId
     await Ajax(signal).Workspaces.workspace(namespace, name).shallowMergeNewAttributes({
-      // Store settings as a string since Rawls expects lists of objects to be entity references
       [savedColumnSettingsWorkspaceAttributeName]: allColumnSettings
     })
   }
--- a/src/components/data/SavedColumnSettings.js
+++ b/src/components/data/SavedColumnSettings.js
@@ -52,12 +52,24 @@ const useSavedColumnSettings = ({ workspaceId, snapshotName, entityMetadata, ent
 
   const signal = useCancellation()
 
+  // Column settings are now stored as JSON in the rawls database (max size 16mb)
+  // In an earlier version of this feature, they were stored as Strings in rawls (max size 64kb)
+  // For backwards-compatibility purposes, we need to support both cases. So we will try to parse
+  // the column settings as a String and if that fails then we will parse it as regular JSON
+  const parseAsStringOrJson = attr => {
+    try {
+      return JSON.parse(attr)
+    } catch (e) {
+      return attr
+    }
+  }
+
   const getAllSavedColumnSettings = async () => {
     const { namespace, name } = workspaceId
     const { workspace: { attributes } } = await Ajax(signal).Workspaces.workspace(namespace, name).details(['workspace.attributes'])
     return _.flow(
       _.getOr('{}', savedColumnSettingsWorkspaceAttributeName),
-      JSON.parse,
+      parseAsStringOrJson,
       // Drop any column settings for entity types that no longer exist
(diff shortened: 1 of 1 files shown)


src/components/data/SavedColumnSettings.js
```diff
@@ -67,7 +79,7 @@ const useSavedColumnSettings = ({ workspaceId, snapshotName, entityMetadata, ent
     const { namespace, name } = workspaceId
     await Ajax(signal).Workspaces.workspace(namespace, name).shallowMergeNewAttributes({
       // Store settings as a string since Rawls expects lists of objects to be entity references
```
davidangb: nit: does this comment need updating now?

src/components/data/SavedColumnSettings.js
```diff
@@ -52,12 +52,24 @@ const useSavedColumnSettings = ({ workspaceId, snapshotName, entityMetadata, ent
...
+  // In an earlier version of this feature, they were stored as Strings in rawls (max size 64kb)
+  // For backwards-compatibility purposes, we need to support both cases. So we will try to parse
+  // the column settings as a String and if that fails then we will parse it as regular JSON
+  const parseAsStringOrJson = attr => {
+    try {
+      return JSON.parse(attr)
+    } catch (e) {
+      return attr
+    }
+  }
```
davidangb, line 65: `JSON.parse()` will succeed for inputs of numbers, booleans - does this matter to you?

davidangb (approved):
minor comments inline. I did not test this myself.

src/components/data/SavedColumnSettings.js
MatthewBemis: I don't think it matters here. If a user sets `columnSettings` to a number or boolean, the UI handles it gracefully

src/components/data/SavedColumnSettings.js
MatthewBemis: I am somewhat curious about what this comment means now. I will check with Nick on Monday since he will be back then

nawatts: I think this comment was written at an earlier stage of this feature's development, when attribute values for saved column settings were arrays.

It looks like it was referring to the behavior of `shallowMergeNewAttributes`, which converts attribute values to attribute operations and in doing so, assumes arrays of objects are entity reference lists.
https://github.com/DataBiosphere/terra-ui/blob/c514947dd99a14735323ab007f36ee5d82f91da1/src/libs/ajax.js#L881-L884
https://github.com/DataBiosphere/terra-ui/blob/c514947dd99a14735323ab007f36ee5d82f91da1/src/libs/ajax.js#L622-L636

Since the attribute value now is an object, I don't think this comment is relevant anymore.
https://github.com/DataBiosphere/terra-ui/blob/c9d1b60d2040dac05e192d0ee083ef004696aa42/src/components/data/SavedColumnSettings.js#L30-L49

nawatts (approved):
Agree with David's [comment](https://github.com/DataBiosphere/terra-ui/pull/3006/files#r867050752) about updating a comment.

src/components/data/SavedColumnSettings.js
MatthewBemis: Ah thank you for digging up the history on this one!

PR #3007, opened 2022-05-06
[AJ-345] Add snapshot ID to by-reference snapshot display

calypsomatic:
[Jira Ticket](https://broadworkbench.atlassian.net/browse/AJ-345)
Add dataset uuids to snapshot-by-reference info page with clipboard copy icons.

To see:
Add a snapshot reference to a workspace through api/workspaces/{workspaceNamespace}/{workspaceName}/snapshots/v2.
Go to the data tab of the workspace.
Screen Shot 2022-05-10 at 9 50 18 AM
Click the 'i' next to your snapshot.

### DIFF
--- a/src/components/workspace-utils.js
+++ b/src/components/workspace-utils.js
@@ -187,7 +187,8 @@ export const SnapshotInfo = ({
         div([sourceDescription]),
         h(SnapshotLabeledInfo, {
           title: 'Data Repo Snapshot Id:', text: [h(Link, {
-            href: `${getConfig().dataRepoUrlRoot}/snapshots/${snapshotId}`, target: '_blank'
+            href: `${getConfig().dataRepoUrlRoot}/snapshots/${snapshotId}`, target: '_blank',
+            'aria-label': 'Go to the snapshot in a new tab'
           }, [snapshotId]), h(ClipboardButton, { 'aria-label': 'Copy data repo snapshot id to clipboard', text: snapshotId, style: { marginLeft: '0.25rem' } })]
         }),
         div({ style: Style.dashboard.header }, [`Source Data Repo Dataset${source.length > 1 ? 's' : ''}`]),
@@ -202,7 +203,8 @@ export const SnapshotInfo = ({
             div([datasetDescription]),
             h(SnapshotLabeledInfo, {
               title: 'Data Repo Dataset Id:', text: [h(Link, {
-                href: `${getConfig().dataRepoUrlRoot}/datasets/${id}`, target: '_blank'
+                href: `${getConfig().dataRepoUrlRoot}/datasets/${id}`, target: '_blank',
+                'aria-label': 'Go to the dataset in a new tab'
               }, [id]), h(ClipboardButton, { 'aria-label': 'Copy data repo dataset id to clipboard', text: snapshotId, style: { marginLeft: '0.25rem' } })]
             })
           ])
--- a/src/components/workspace-utils.js
+++ b/src/components/workspace-utils.js
@@ -2,8 +2,7 @@ import debouncePromise from 'debounce-promise'
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
 import { b, div, h, p } from 'react-hyperscript-helpers'
-import {
-  AsyncCreatableSelect, ButtonPrimary, ButtonSecondary, ClipboardButton, IdContainer, Link, Select, spinnerOverlay
+import { AsyncCreatableSelect, ButtonPrimary, ButtonSecondary, ClipboardButton, IdContainer, Link, Select, spinnerOverlay
 } from 'src/components/common'
 import { icon } from 'src/components/icons'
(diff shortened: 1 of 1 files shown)


src/components/workspace-utils.js
```diff
@@ -190,6 +197,7 @@ export const SnapshotInfo = ({
             key: id,
             style: { marginBottom: '1rem' }
           }, [
+            h(SnapshotLabeledInfo, { title: 'Source Data Repo Id:', text: [id, h(ClipboardButton, { 'aria-label': 'Copy source data repo id to clipboard', text: id, style: { marginLeft: '0.25rem' } })] }),
```
davidangb: could this be a link too? it would go to /datasets/${id}, same way the snapshot goes to /snapshots/${snapshotId}.

src/components/workspace-utils.js
```diff
@@ -180,6 +182,11 @@ export const SnapshotInfo = ({
       ]),
       div({ style: { paddingLeft: '1rem' } }, [
         div({ style: Style.dashboard.header }, ['Linked Data Repo Snapshot']),
+        h(SnapshotLabeledInfo, {
+          title: 'Data Repo Id:', text: [h(Link, {
+            href: `${getConfig().dataRepoUrlRoot}/snapshots/${snapshotId}`
+          }, [snapshotId]), h(ClipboardButton, { 'aria-label': 'Copy data repo id to clipboard', text: snapshotId, style: { marginLeft: '0.25rem' } })]
+        }),
```
davidangb: could this open in a new window? You should just be able to add `target: '_blank',` to the `Link` definition

src/components/workspace-utils.js
```diff
@@ -2,13 +2,15 @@ import debouncePromise from 'debounce-promise'
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
 import { b, div, h, p } from 'react-hyperscript-helpers'
-import { AsyncCreatableSelect, ButtonPrimary, ButtonSecondary, IdContainer, Link, Select, spinnerOverlay } from 'src/components/common'
+import { AsyncCreatableSelect, ButtonPrimary, ButtonSecondary, ClipboardButton, IdContainer, Link, Select, spinnerOverlay
+} from 'src/components/common'
```
davidangb: this linebreak causes a linting error, which is preventing tests from passing. https://github.com/DataBiosphere/terra-ui/wiki/Developing#code-style-and-linting has some info on configuring a local linter so you can see problems before PR - you may have to adapt those instructions for whatever editor you're using.

src/components/workspace-utils.js
davidangb: Also, the terminology "Source Data Repo Id" here could be confusing. I'd suggest either "Data Repo Id" to use the same terminology as the previous link, or call the snapshot "Data Repo Snapshot Id" and this one "Data Repo Dataset Id" - but those should be Alessandro's call!

davidangb (commented):
Some comments inline!

src/components/workspace-utils.js
calypsomatic: I have no idea where that space came from but I think I've fixed it now.

src/components/workspace-utils.js
calypsomatic: Done.

src/components/workspace-utils.js
calypsomatic: Alessandro agrees!  Updated.

davidangb (approved):
lgtm, thank you! Works for me locally too.

src/components/workspace-utils.js
```diff
@@ -184,6 +185,11 @@ export const SnapshotInfo = ({
         h(SnapshotLabeledInfo, { title: 'Creation Date:', text: Utils.makeCompleteDate(createdDate) }),
         div({ style: { ...Style.elements.sectionHeader, marginBottom: '0.2rem' } }, ['Description:']),
         div([sourceDescription]),
+        h(SnapshotLabeledInfo, {
+          title: 'Data Repo Snapshot Id:', text: [h(Link, {
```
cahrens, line 189: I recommend you try out navigating to this link with a screen reader ([VoiceOver instructions](https://webaim.org/articles/voiceover/)). Right now all you hear is the snapshotID, which is unlikely to meaningful. Here are suggestions on how to improve this-- https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8.

Also, in the future please add steps on how to get to the modified UI elements (a screenshot is also helpful).

calypsomatic: Thanks!  I've added just a description of where the link goes - is that enough or should it also read out the id?

cahrens: I don't think it's necessary to include the ID-- it seems unlikely that it will be meaningful, and if the user really wants to know what it is, they can use VO keys (VO-Shift-Down) to navigate into the link and get the details.

cahrens (approved):

PR #3008, opened 2022-05-09
[WOR-184] Hide non-applicable fields in Create Workspace dialog if Azure billing project, only clone to same cloud evnironment

cahrens:
Things to test:

1. If you are in the azure alpha group (I can add you to it if you would like), the "Create Workspace" dialog will include an Azure billing project in the billing projects select. When you choose that billing project, bucket location and auth domain fields will disappear. Note that in the longterm we will be changing the "Create Workspace" into more of a wizard dialog, where these fields will be on a second screen (after the billing project is selected) so that they are not shown unless they are applicable.

image

image

2. When you choose to Clone a GCP Workspace, the Azure billing project is not in the list of billing projects (again, you have to be in the azure alpha group for this filter to actually be relevant).

image

3. When you bring up the Clone dialog from an Azure Workpace, only the Azure billing project is in the list. **NOTE**: cloning of an Azure workspace does not work, so don't try it! We will soon be disabling/removing the clone menu item for Azure workspaces until this functionality is built out.

image


src/components/NewWorkspaceModal.js
```diff
@@ -118,12 +118,13 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
...
     Ajax(signal).Billing.listProjects()
       .then(_.filter({ status: 'Ready' }))
+      .then(_.filter(project => isBillingProjectApplicable(project)))
       .then(projects => {
         setBillingProjects(projects)
         setNamespace(_.some({ projectName: namespace }, projects) ? namespace : undefined)
       }),
     Ajax(signal).Groups.list().then(setAllGroups),
-    !!cloneWorkspace && Ajax(signal).Workspaces.workspace(namespace, cloneWorkspace.workspace.name).checkBucketLocation(cloneWorkspace.workspace.googleProject, cloneWorkspace.workspace.bucketName)
+    !!cloneWorkspace && !cloneWorkspace.azureContext && Ajax(signal).Workspaces.workspace(namespace, cloneWorkspace.workspace.name).checkBucketLocation(cloneWorkspace.workspace.googleProject, cloneWorkspace.workspace.bucketName)
```
cahrens, line 127: The checkBucketLocation call will error for Azure workspaces. We are going to disable the Clone option until it is fully supported for Azure workspaces, so I could choose not to add this check… but while testing this PR I was distracted/confused by the error popping up.

src/components/NewWorkspaceModal.js
```diff
@@ -153,7 +171,7 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
   const destLocationType = bucketLocation === defaultLocation ? locationTypes.default : locationTypes.region
 
   return Utils.cond(
-    [loading, () => spinnerOverlay],
+    [loading || billingProjects === undefined, () => spinnerOverlay],
```
cahrens, line 174: This addresses a bug I am seeing in dev. Currently, when you select the "Clone" operation, for some workspaces you briefly see the dialog below pop up before the "Clone Workspace" dialog. This is because the code thinks that there are no billing projects configured until the billing projects are fetched, and it is possible for the "loading" flag to get set to false before the billingProjects state variable has been persisted if an error is thrown while obtaining the cloned workspace's bucket location information (we have a lot of dev workspaces where the bucket location call errors).

Dialog we don't want to see flicker up:
image

aherbst-broad (approved):
Tested locally against both the Azure + GCP use cases, worked great.

marctalbott (approved):
lgtm, tested against dev and everything looked good

nawatts (approved):

PR #3013, opened 2022-05-09
[DC-324] accessing google buckets can fail resulting in error notifications for users

slucasbroad:
When using Terra UI we get a lot of error notifications around google buckets not being accessible.

This can cause [test flakiness](https://broadinstitute.slack.com/archives/C01EHNUM73R/p1644358202261189?thread_ts=1644354756.201859&cid=C01EHNUM73R) and also manifests in production, where users tend to see error notifications.

Google recommends [retry logic ](https://cloud.google.com/storage/docs/retry-strategy#rest-apis)for these requests

Where to Start

Take a look at the [ajax.js](https://github.com/DataBiosphere/terra-ui/blob/c1d969c83e59560d86916ce62b2f166b25e8c7ee/src/libs/ajax.js#L56) file, specifically at some of the wrapper functions. withErrorRejection is a simple example to start with.

Create a wrapper function that will retry a request after failure. A key property of a wrapper function is that it should return a function with the same signature as the function that was passed in (again see withErrorRejection)

This is something that can be done in a standalone fashion, if that makes it easier. You can create a wrapper that retries any function upon receiving an error.

It’s Done when

Our google api calls in ajax.js are wrapped in retry logic.

For testing, you can use the browser dev tools to block access to API’s to simulate a failed request

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -52,6 +52,40 @@ const withCancellation = wrappedFetch => async (...args) => {
   }
 }
 
+// 1. Add Button to catalog - when clicked makes fake Ajax call
+// 2. Add Fake Ajax request
+// 3. Add retry wrapper, wrap our fake Ajax request
+
+// TODO: slucas to Implement
+// On a failure, this wrapper should attempt to retry the fetch call.
+// This should hapen every N milliseconds, with a max timeout.
+// When the max timeout is reached, throw an error
+// For random millis have the wait be between 500ms and 2000ms
+// keep retrying until timeout or success
+// on timeout return the error observed
+
+const withRetryOnError = wrappedFetch => async (...args) => {
+
+  let randomWait = 0
+
+  for(let maxoutTimer = 0; maxoutTimer < 3000; maxoutTimer = maxoutTimer + randomWait){
+
+    randomWait = Math.random() * 1500 + 500
+    console.log('Random Milliseconds', randomWait, 'Timer so Far:', maxoutTimer) // wait between 500ms and 2000ms
+
+    try {
+      console.log('try promise')
+      await Utils.withDelay(wrappedFetch,randomWait)(...args)
+    }
+    catch (error){
+      console.log('was connection error')
+    }
+  }
+  console.log('loop end')
+}
+
+
+
 // Converts non-200 responses to exceptions
 const withErrorRejection = wrappedFetch => async (...args) => {
   const res = await wrappedFetch(...args)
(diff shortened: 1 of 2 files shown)


cahrens:
Do we have reason to believe that retrying is going to help? I know that in our dev environment, we have many workspaces where the bucket information never will load properly. In your testing, are you seeing cases where the retrying does help?

petesantos:
> Do we have reason to believe that retrying is going to help? I know that in our dev environment, we have many workspaces where the bucket information never will load properly. In your testing, are you seeing cases where the retrying does help?

@cahrens This ticket was created a while back when we were seeing sporadic test failures around accessing google buckets. During that time, I was doing some unrelated work (to the testing effort) using Production and noticed I was seeing the same types of error notifications in production, indicating this is a user facing issue.

That said, re-running some tests using the flake checker and looking at circle, I have not seen this issue pop up as often as I had been seeing it in the past.

We can update the change to check for certain error codes and not retry in those cases, and we can also be more selective in which bucket calls we want to retry.

fboulnois (approved):

src/libs/ajax.js
```diff
@@ -52,6 +52,23 @@ const withCancellation = wrappedFetch => async (...args) => {
...
 }
 
+const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
+  const timeout = 5000
+  const somePointInTheFuture = Date.now() + timeout
+  const maxDelayIncrement = 1500
+  const minDelay = 500
+
+  while (Date.now() < somePointInTheFuture) {
+    const until = Math.random() * maxDelayIncrement + minDelay
```
cahrens, line 62: What was the thinking behind the random delay? It seems unnecessarily complex to me.

cahrens (approved):
I would remove the random, but otherwise LGTM.

cahrens:
When this changes goes in, I am thinking that we might be able to remove https://github.com/DataBiosphere/terra-ui/blob/e69798efefbadf368f737876d84fd8358b45f6ad/integration-tests/utils/integration-helpers.js#L72 which was added in an attempt to work around permissions not appearing fast enough.

src/libs/ajax.js
cahrens: @slucasbroad can you provide the background on why you are doing a random delay?

slucasbroad: Using Googles recommend retry algorithm. If a lot of clients are retrying the same service, random prevents them from retrying at the same time, staggering the requests.

slucasbroad:
> When this changes goes in, I am thinking that we might be able to remove
> 
> https://github.com/DataBiosphere/terra-ui/blob/e69798efefbadf368f737876d84fd8358b45f6ad/integration-tests/utils/integration-helpers.js#L72
> 
> which was added in an attempt to work around permissions not appearing fast enough.

We added DC-357 to investigate whether we can remove the code.

PR #3015, opened 2022-05-10
[WOR-244] Workspace menu cleanup

cahrens:
This removes the workspace menu duplication, fixes some bugs, and adds disabling of Clone/Share for Azure workspaces (this functionality won't be built out until Q3). Some test coverage is added of the menu enable/disable state and tooltips, but it is not exhaustive (and Azure workspace is mocked).

Things to test:
1. On Workspaces List, try the options for Workspaces with different access levels/cloud contexts. **NOTE**: Delete is not yet implemented for Azure workspaces, so it is not recommended that you try deleting one. 

On the Workspaces List, the workspace details are not fetched until the menu is triggered, so you will see menu item text change or disable. This is consistent with the existing behavior in production, and it is done to avoid fetching the details for all workspaces upfront (which would be a performance issue for users with access to lots of workspaces).

2. Select various Workspaces from the list and view the menu on the Workspace Container/Dashboard.

### DIFF
--- a/integration-tests/tests/workspace-dashboard.js
+++ b/integration-tests/tests/workspace-dashboard.js
@@ -23,6 +23,10 @@ const workspaceDashboardPage = (testPage, testUrl, token, workspaceName) => {
       await Promise.all(_.map(async item => await findText(testPage, item), expectedTextItems))
     },
 
+    assertReadOnly: async () => {
+      await findText(testPage, 'Workspace is read only')
+    },
+
     assertWorkspaceMenuItems: async expectedMenuItems => {
       await click(testPage, clickable({ text: 'Workspace Action Menu' }))
       await Promise.all(_.map(async ({ label, tooltip }) => {
@@ -106,7 +110,7 @@ const setAjaxMockValues = async (testPage, namespace, name, workspaceDescription
     workspaceVersion: 'v2'
   }
   const azureWorkspacesListResult = [{
-    accessLevel: 'WRITER',
+    accessLevel: 'READER',
     public: false,
     workspace: workspaceInfo,
     workspaceSubmissionStats: { runningSubmissionsCount: 0 }
@@ -119,7 +123,7 @@ const setAjaxMockValues = async (testPage, namespace, name, workspaceDescription
       tenantId: 'dummy-tenant-id'
     },
     workspaceSubmissionStats: { runningSubmissionsCount: 0 },
-    accessLevel: 'WRITER',
+    accessLevel: 'READER',
     owners: ['dummy@email.comm'],
     workspace: workspaceInfo,
     canShare: true,
@@ -177,7 +181,10 @@ const testAzureWorkspace = withUserToken(async ({ page, token, testUrl }) => {
   // Check cloud information
   await dashboard.assertCloudInformation(['Cloud NameMicrosoft Azure', 'Resource Group IDdummy-mrg-id'])
 
-  // Verify workspace tooltips on Workspace menu items (all will be disabled, WRITER permissions only).
+  // READER permissions only
+  await dashboard.assertReadOnly()
+
+  // Verify workspace tooltips on Workspace menu items (all will be disabled due to Azure workspace + READER permissions).
   await dashboard.assertWorkspaceMenuItems([
     { label: 'Clone', tooltip: 'Cloning is not supported on Azure Workspaces' },
     { label: 'Share', tooltip: 'Sharing is not supported on Azure Workspaces' },
--- a/integration-tests/tests/workspace-dashboard.js
+++ b/integration-tests/tests/workspace-dashboard.js
@@ -20,7 +20,7 @@ const workspaceDashboardPage = (testPage, testUrl, token, workspaceName) => {
 
     assertCloudInformation: async expectedTextItems => {
       await click(testPage, clickable({ text: 'Cloud information' }))
-      await Promise.all(_.map(async (item) => await findText(testPage, item), expectedTextItems))
+      await Promise.all(_.map(async item => await findText(testPage, item), expectedTextItems))
     },
 
     assertWorkspaceMenuItems: async expectedMenuItems => {
@@ -52,7 +52,7 @@ const workspaceDashboardPage = (testPage, testUrl, token, workspaceName) => {
     },
 
     assertTabs: async (expectedTabs, enabled) => {
-      await Promise.all(_.map(async (tab) => {
+      await Promise.all(_.map(async tab => {
         await (enabled ? clickNavChildAndLoad(testPage, tab) : assertNavChildNotFound(testPage, tab))
       }, expectedTabs))
     }
--- a/integration-tests/tests/workspace-dashboard.js
+++ b/integration-tests/tests/workspace-dashboard.js
@@ -1,30 +1,89 @@
 // This test is owned by the Workspaces Team.
 const _ = require('lodash/fp')
-const { clickNavChildAndLoad, viewWorkspaceDashboard, withWorkspace } = require('../utils/integration-helpers')
-const { assertNavChildNotFound, click, clickable, findText } = require('../utils/integration-utils')
+const { clickNavChildAndLoad, overrideConfig, viewWorkspaceDashboard, withWorkspace } = require('../utils/integration-helpers')
+const { assertNavChildNotFound, assertTextNotFound, click, clickable, findElement, findText, noSpinnersAfter } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
-const { waitUntilLoadedOrTimeout } = require('../utils/integration-utils')
 
+
+const workspaceDashboardPage = (testPage, testUrl, token, workspaceName) => {
+  return {
+    visit: async (loadUrl = true) => {
+      if (loadUrl) {
+        await testPage.goto(testUrl)
+      }
+      await viewWorkspaceDashboard(testPage, token, workspaceName)
+    },
+
+    assertDescription: async expectedDescription => {
+      await findText(testPage, expectedDescription)
+    },
+
+    assertCloudInformation: async expectedTextItems => {
+      await click(testPage, clickable({ text: 'Cloud information' }))
+      await Promise.all(_.map( async(item) => await findText(testPage, item), expectedTextItems ))
+    },
+
+    assertWorkspaceMenuItems: async expectedMenuItems => {
+      await click(testPage, clickable({ text: 'Workspace Action Menu' }))
+      await Promise.all(_.map( async({ label, tooltip }) => {
+        if (!!tooltip) {
+          await findElement(testPage, clickable({ textContains: label, isEnabled: false }))
+          await findText(testPage, tooltip)
+        }
+        else {
+          await findElement(testPage, clickable({ textContains: label }))
+        }
+      }, expectedMenuItems ))
+    },
+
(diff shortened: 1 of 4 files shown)


integration-tests/tests/workspace-dashboard.js
```diff
@@ -1,30 +1,92 @@
...
 const _ = require('lodash/fp')
-const { clickNavChildAndLoad, viewWorkspaceDashboard, withWorkspace } = require('../utils/integration-helpers')
-const { assertNavChildNotFound, click, clickable, findText } = require('../utils/integration-utils')
+const { clickNavChildAndLoad, overrideConfig, viewWorkspaceDashboard, withWorkspace } = require('../utils/integration-helpers')
+const { assertNavChildNotFound, assertTextNotFound, click, clickable, findElement, findText, noSpinnersAfter } = require('../utils/integration-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
-const { waitUntilLoadedOrTimeout } = require('../utils/integration-utils')
 
 
+const workspaceDashboardPage = (testPage, testUrl, token, workspaceName) => {
```
cahrens, line 8: Created a helper page object to make the test cases easier to read/avoid duplication.

integration-tests/tests/workspace-dashboard.js
```diff
@@ -48,7 +110,7 @@ const setAjaxMockValues = async (testPage, namespace, name, workspaceDescription
     workspaceVersion: 'v2'
   }
   const azureWorkspacesListResult = [{
-    accessLevel: 'OWNER',
+    accessLevel: 'READER',
```
cahrens, line 113: I wanted more coverage of the menu tooltips, plus coverage of the Read-only messaging.

image

integration-tests/tests/workspace-dashboard.js
```diff
@@ -93,6 +155,10 @@ const setAjaxMockValues = async (testPage, namespace, name, workspaceDescription
       {
         filter: { url: /api\/workspaces[^/](.*)/ },
         fn: () => () => Promise.resolve(new Response(JSON.stringify(azureWorkspacesListResult), { status: 200 }))
+      },
+      {
+        filter: { url: /api\/v2\/runtimes(.*)/ }, // Needed to prevent errors from the Runtime (IA) component that will be going away.
```
cahrens, line 160: Without this mocked, the API errors because we don't have a real workspace ID.

integration-tests/tests/workspace-dashboard.js
```diff
@@ -105,21 +171,32 @@ const testAzureWorkspace = withUserToken(async ({ page, token, testUrl }) => {
   // Must load page before setting mock responses.
   await page.goto(testUrl)
   await findText(page, 'View Workspaces')
+  await overrideConfig(page, { isAnalysisTabVisible: true })
```
cahrens, line 174: IA has enabled the Analyses Tab in Azure workspaces (behind feature flag), so adding coverage of that.

src/pages/workspaces/List.js
```diff
@@ -47,40 +47,6 @@ const workspaceSubmissionStatus = ({ workspaceSubmissionStats: { runningSubmissi
...
 
-const WorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDelete, onLock }) => {
-  const { workspace } = useWorkspaceDetails({ namespace, name }, ['accessLevel', 'canShare', 'workspace.isLocked'])
-  const canRead = workspace && Utils.canRead(workspace.accessLevel)
-  const canShare = workspace?.canShare
-  const isOwner = workspace && Utils.isOwner(workspace.accessLevel)
-  const isLocked = workspace?.workspace.isLocked
-  return h(Fragment, [
-    h(MenuButton, {
-      disabled: !canRead,
```
cahrens, line 58: We actually never created this menu if the user does not have read permissions for the workspace, so there was no need to carry the check for `canRead` into the shared code.

image

src/pages/workspaces/List.js
```diff
@@ -47,40 +47,6 @@ const workspaceSubmissionStatus = ({ workspaceSubmissionStats: { runningSubmissi
...
-      onClick: onShare
-    }, [makeMenuIcon('share'), 'Share']),
-    h(MenuButton, {
-      disabled: !isOwner,
-      tooltip: !isOwner && ['You have not been granted permission to ', isLocked ? 'unlock' : 'lock', ' this workspace'],
-      tooltipSide: 'left',
-      onClick: onLock
-    }, isLocked ? [makeMenuIcon('unlock'), 'Unlock'] : [makeMenuIcon('lock'), 'Lock']),
-    h(MenuButton, {
-      disabled: !isOwner,
```
cahrens, line 76: This was missing a check for `isLocked`, which was on the other version of the menu.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -85,19 +84,17 @@ const WorkspaceTabs = ({
   ]
   return h(Fragment, [
     h(TabBar, {
-      'aria-label': 'workspace menu',
+      'aria-label': 'Workspace Navigation Tabs',
```
cahrens, line 87: Both the tabs and the workspace action menu were named "Workspace menu", which I found to be confusing.

src/pages/workspaces/workspace/WorkspaceMenu.js
```diff
@@ -0,0 +1,90 @@
...
+
+
+// In `workspaceInfo`, specify either `name and namespace` to fetch the Workspace details,
+// or `canShare, isAzureWorkspace, isLocked, and isOwner` to use previously fetched details.
+const WorkspaceMenu = ({
+  iconSize, popupLocation,
+  callbacks: { onClone, onShare, onLock, onDelete },
+  workspaceInfo: { name, namespace, canShare, isAzureWorkspace, isLocked, isOwner }
+}) => {
+  const navIconProps = {
```
cahrens, line 17: Note that these properties weren't used on the Workspaces list before, but I think the slightly different hover state is nice and it is good to be consistent.

src/pages/workspaces/workspace/WorkspaceMenu.js
```diff
@@ -0,0 +1,90 @@
...
+    h(DynamicWorkspaceMenuContent, { namespace, name, onShare, onClone, onDelete, onLock }) :
+    h(WorkspaceMenuContent, { canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete })
+
+  return h(MenuTrigger, {
+    side: popupLocation,
+    closeOnClick: true,
+    content: menuContent
+  }, [
+    h(Clickable, {
+      'aria-label': !!name ? `Action Menu for Workspace: ${name}` : 'Workspace Action Menu',
```
cahrens, line 32: I added "Action" to try to clarify what the menu is… but could certainly drop it if folks think it's unnecessary.

src/pages/workspaces/workspace/WorkspaceMenu.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded = true }) => {
+  const shareTooltip = Utils.cond(
+    [isAzureWorkspace, () => 'Sharing is not supported on Azure Workspaces'],
+    [workspaceLoaded && !canShare, () => 'You have not been granted permission to share this workspace'],
+    [Utils.DEFAULT, () => '']
+  )
+  const deleteTooltip = Utils.cond(
+    [workspaceLoaded && isLocked, () => 'You cannot delete a locked workspace'],
```
cahrens, line 57: Added a specific tooltip for the locked case.

cahrens:
@MatthewBemis You worked recently on these menus, so I'm tagging you in case you want to take a look (completely up to you). I added a bit of test coverage as well.

cahrens:
@nawatts could you be the terra-ui reviewer for this PR? It didn't get one auto-assigned.

src/pages/workspaces/workspace/WorkspaceMenu.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+const DynamicWorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDelete, onLock }) => {
+  const { workspace } = useWorkspaceDetails({ namespace, name }, ['accessLevel', 'azureContext', 'canShare', 'workspace.isLocked'])
+  const canShare = workspace?.canShare
+  const isOwner = workspace && Utils.isOwner(workspace.accessLevel)
+  const isLocked = workspace?.workspace.isLocked
+  const isAzureWorkspace = !!workspace?.azureContext
+  return WorkspaceMenuContent({
+    canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded: !!workspace
```
nawatts, line 46: Would it be preferable to wait until the workspace is loaded to render the menu content? The initial disabled states can be confusing if the workspace details call is slow.

nawatts (approved):

src/pages/workspaces/workspace/WorkspaceMenu.js
cahrens: I agree that it is a somewhat confusing user experience, but I think changing it is beyond the scope of this ticket (given that it's already this way in production). I will create a Workspaces ticket to consult with UX and look at options. I think it would be odd to show a page-wide spinner when you open the menu, and really it would be best to have this state upfront so we could display the status of the workspaces in the list table itself (currently you have to open the menu to see if a workspace is locked, for example, and we want to be able to show the cloud context for the workspace in the list). Therefore I think we would want to investigate the feasibility of adding cloud context, locked, and canShare to the workspaces list API (accessLevel is already in that API).

nawatts: Sounds good. Agreed about the page-wide spinner, I was thinking something like a small spinner in the popup. But returning the data from the workspaces list API does sound like the ideal solution.

integration-tests/tests/workspace-dashboard.js
aherbst-broad: 👍 I like these more expressive assertions.

aherbst-broad (approved):
Tested locally, lgtm.

src/pages/workspaces/workspace/WorkspaceMenu.js
cahrens: I created https://broadworkbench.atlassian.net/browse/WOR-316. We already have a ticket in an upcoming sprint to add the cloud context info to workspaces list, so hopefully we can tackle `locked` and `canShare` at the same time.

MatthewBemis: @cahrens `isLocked` is already available from the listWorkspaces API, but the UI excludes it from the API response by only requesting a limited number of workspace fields. The history behind this is that we wanted to reduce the size of the JSON response returned, since it can be pretty large. However, I think this is a good reason for the UI to start requesting this field. It should just be a matter of adding `workspace.isLocked` to the list of fields [here](https://github.com/DataBiosphere/terra-ui/blob/dev/src/components/workspace-utils.js#L32)

cahrens: Great! That means that `canShare` and cloud context are the only things we would need to add to the list API.

PR #3017, opened 2022-05-10
[SU-99] Add column to a data table

nawatts:
This relates to the new data tab design, enabled with `configOverridesStore.set({ isDataTabRedesignEnabled: true })`.

## Before

Currently, it's possible to add a column to selected rows in a data table by selecting some rows, clicking "Edit attribute" from the Edit menu, and entering a column name that doesn't already exist.

https://user-images.githubusercontent.com/1156625/167699515-0fc3d16e-f708-4085-a419-41d5f266175b.mov

## After

Feedback from users was that it's not clear what "Edit attribute" does and, generally, what is an attribute?

This adds a separate menu option to "Add column", which applies to all rows in the table (no need to select rows).

https://user-images.githubusercontent.com/1156625/167699883-298ab254-545b-4662-bd90-bddbe1f11d2d.mov

It also removes the option to add a new attribute when editing, requiring an existing attribute to be selected.

https://user-images.githubusercontent.com/1156625/167699991-4d46e0c8-4e75-469a-98ff-242df12e14a3.mov

It also changes wording in several modals and buttons, mostly replacing "attribute" with "value" or "field" and replacing entity types with a generic "row".

### DIFF
--- a/src/components/data/EntitiesContent.js
+++ b/src/components/data/EntitiesContent.js
@@ -7,7 +7,7 @@ import { Fragment, useRef, useState } from 'react'
 import { div, form, h, input } from 'react-hyperscript-helpers'
 import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
 import { ButtonPrimary, ButtonSecondary, Link } from 'src/components/common'
-import { AddEntityModal, EntityDeleter, ModalToolButton, MultipleEntityEditor, saveScroll } from 'src/components/data/data-utils'
+import { AddColumnModal, AddEntityModal, EntityDeleter, ModalToolButton, MultipleEntityEditor, saveScroll } from 'src/components/data/data-utils'
 import DataTable from 'src/components/data/DataTable'
 import ExportDataModal from 'src/components/data/ExportDataModal'
 import { icon, spinner } from 'src/components/icons'
@@ -219,6 +219,7 @@ const EntitiesContent = ({
   const [deletingEntities, setDeletingEntities] = useState(false)
   const [copyingEntities, setCopyingEntities] = useState(false)
   const [addingEntity, setAddingEntity] = useState(false)
+  const [addingColumn, setAddingColumn] = useState(false)
   const [nowCopying, setNowCopying] = useState(false)
   const [refreshKey, setRefreshKey] = useState(0)
   const [showToolSelector, setShowToolSelector] = useState(false)
@@ -381,6 +382,9 @@ const EntitiesContent = ({
           onClick: () => setAddingEntity(true)
         }, 'Add row'),
         h(MenuButton, {
+          onClick: () => setAddingColumn(true)
+        }, ['Add column']),
+        h(MenuButton, {
           disabled: !entitiesSelected,
           tooltip: !entitiesSelected && 'Select rows to edit in the table',
           onClick: () => setEditingEntities(true)
@@ -495,6 +499,16 @@ const EntitiesContent = ({
         workspaceId: { namespace, name },
         onDismiss: () => setAddingEntity(false),
         onSuccess: () => setRefreshKey(_.add(1))
+      }),
+      addingColumn && h(AddColumnModal, {
+        entityType: entityKey,
+        entityMetadata,
+        workspaceId: { namespace, name },
+        onDismiss: () => setAddingColumn(false),
+        onSuccess: () => {
+          setAddingColumn(false)
+          setRefreshKey(_.add(1))
+        }
       }),
       editingEntities && h(MultipleEntityEditor, {
         entityType: entityKey,
--- a/src/components/data/EntitiesContent.js
+++ b/src/components/data/EntitiesContent.js
@@ -347,9 +347,9 @@ const EntitiesContent = ({
         }, ['Download as TSV']),
         !snapshotName && h(MenuButton, {
           disabled: noEdit,
-          tooltip: noEdit ? 'You don\'t have permission to modify this workspace.' : 'Edit an attribute of the selected rows',
+          tooltip: noEdit ? 'You don\'t have permission to modify this workspace.' : 'Edit a field of the selected rows',
           onClick: () => setEditingEntities(true)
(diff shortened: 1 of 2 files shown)


MatthewBemis (approved):

src/components/data/EntitiesContent.js
```diff
@@ -346,9 +347,9 @@ const EntitiesContent = ({
         }, ['Download as TSV']),
         !snapshotName && h(MenuButton, {
           disabled: noEdit,
-          tooltip: noEdit ? 'You don\'t have permission to modify this workspace.' : 'Edit an attribute of the selected rows',
+          tooltip: noEdit ? 'You don\'t have permission to modify this workspace.' : 'Edit a field of the selected rows',
```
breilly2, line 350: So valuable to get user feedback like this. I wouldn't have thought that this wasn't obvious. 👍

src/components/data/EntitiesContent.js
```diff
@@ -496,6 +500,16 @@ const EntitiesContent = ({
...
         onSuccess: () => setRefreshKey(_.add(1))
       }),
+      addingColumn && h(AddColumnModal, {
+        entityType: entityKey,
+        entityMetadata,
+        workspaceId: { namespace, name },
+        onDismiss: () => setAddingColumn(false),
+        onSuccess: () => {
+          setAddingColumn(false)
+          setRefreshKey(_.add(1))
```
breilly2, line 510: Thinking Out Loud: I'm not familiar with the notion of a refresh key. I recognize that this isn't new in this PR, but it's a technique that is relatively new to the codebase and I'm curious what this is for. I'll have to explore this.

src/components/data/data-utils.js
```diff
@@ -759,19 +759,19 @@ export const SingleEntityEditor = ({ entityType, entityName, attributeName, attr
...
   return h(Modal, {
-    title: 'Modify Attribute',
+    title: 'Edit value',
     onDismiss,
     showButtons: false
   }, [
     consideringDelete ?
       h(Fragment, [
-        'Are you sure you want to delete the attribute ', boldish(attributeName),
+        'Are you sure you want to delete the field ', boldish(attributeName),
```
breilly2: Should this (and below) say "value"? I can see how this is somewhat at the intersection of "column" and "value", but "field" feels closer to "column" to me and I _think_ we're dealing with specific rows here, not the entire column of the data table.

There are a few other places where this might apply. What do you see as the distinction between "field" and "value"?

src/components/data/data-utils.js
```diff
@@ -928,6 +928,104 @@ export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames,
   ])
 }
 
+export const AddColumnModal = ({ entityType, entityMetadata, workspaceId: { namespace, name }, onDismiss, onSuccess }) => {
+  const [columnName, setColumnName] = useState('')
+  const [columnNameTouched, setColumnNameTouched] = useState(false)
```
breilly2, line 933: Nice! This is definitely one of those subtle UX things and I love how clearly this expresses the intent of the code.

src/components/data/data-utils.js
```diff
@@ -928,6 +928,104 @@ export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames,
...
+              marginTop: '0.5rem',
+              color: colors.danger()
+            }
+          }, columnNameError)
+        ])
+      ])
+    ]),
+    p([
+      span({ style: { fontWeight: 'bold' } }, ['Default value']),
+      ' (will be entered for all rows)'
```
breilly2: Watching the videos you attached to the PR (think you for that!), it wasn't obvious to me that this was optional, probably because the placeholder above that said "required" was no longer rendered. I think I would like it to be more obvious that this field is optional, either by including that in the placeholder text here or using a "*" or other persistent indicator on the required fields.

breilly2 (commented):
This is really nice! I do have some minor feedback and questions. See individual comments.

src/components/data/data-utils.js
nawatts: Mostly I was going off the design linked in the [ticket](https://broadworkbench.atlassian.net/browse/SU-99). Though I was thinking of "field" as referring to the cell and "value" as referring to its contents.

That thinking is probably influenced by the underlying data model though, which doesn't exactly match the UI. The UI shows a table, but any given row doesn't necessarily have all the attributes represented by the columns in the table. So here, delete doesn't clear the value of the selected attribute, it deletes the attribute from the row entirely.

src/components/data/EntitiesContent.js
nawatts: It's kind of a hack necessitated by having data loading tied to the lifecycle of a component further down in the tree. These "refresh keys" are passed as keys to some component and changing the key causes that component to be unmounted and replaced by a new instance. This causes lifecycle hooks such as `useEffect` and `useOnMount` in that component to re-run.

https://kentcdodds.com/blog/understanding-reacts-key-prop

src/components/data/EntitiesContent.js
kyuksel: The terms `attribute` and `entity` have always confused me  🤷🏼  Might as well use `thing` in lieu of `entity` 😛

src/components/data/data-utils.js
nawatts: In a way, it's not optional. There's always a default value, which defaults to an empty string.

nawatts: Updated the label in https://github.com/DataBiosphere/terra-ui/pull/3017/commits/28ad6d4e7241f4a790be3916a2f7fbe2130859bc.

src/components/data/data-utils.js
nawatts: Changed to "value" in https://github.com/DataBiosphere/terra-ui/pull/3017/commits/641f4cdf8b5979039adaf17fd08e5d9f4c4a9cc4.

breilly2 (approved):
👍  Thank you for the responses to my questions and feedback.

PR #3018, opened 2022-05-10
[DC-310] update how we access table cells in UI tests

slucasbroad:
Determining the column index via the header information and “scanning” the rows of a table to find the correct data we are looking for in a given column.

Fix matching bug


nawatts (approved):

petesantos (approved):
Thanks for fixing this!

PR #3020, opened 2022-05-10
[IA-3290][IA-2989][SUP-559] Added wait to stop ComputeModal from popping up if environment already exists

cstrandw:
Before, refreshing a notebook while a cloud environment was running caused the ComputeModal to open and allowed users to create a duplicate environment. This was caused by the RuntimeKicker element in the NotebookLauncher component opening the ComputeModal when there is a null runtime ([see here](https://github.com/DataBiosphere/terra-ui/blob/e786d1035ffea48f06529f708e5c47b39d1e3468/src/pages/workspaces/workspace/notebooks/NotebookLauncher.js#L82)). I was able to fix this by adding a wait in the RuntimeKicker if the runtime was initially null.

Testing done:

- Followed the replication steps included in [the ticket](https://broadworkbench.atlassian.net/browse/IA-3290) and verified that the ComputeModal did not open any more
- Followed the steps above, and then opened another window and deleted the environment there. I then went back to the original window and refreshed. Verified that the ComputeModal opened as expected.


src/components/runtime-common.js
```diff
@@ -39,10 +39,14 @@ export const RuntimeKicker = ({ runtime, refreshRuntimes, onNullRuntime }) => {
         await refreshRuntimes()
         setBusy(false)
         return
-      } else if (currentRuntime === undefined || status === 'Stopping') {
+      } else if (currentRuntime === undefined || status === 'Stopping' ) {
         await Utils.delay(500)
       } else if (currentRuntime === null) {
-        onNullRuntime()
+        await Utils.delay(1000)
```
cahrens: Is there a way you can poll for a certain condition to be true? Delays are notorious for causing flakiness.

cstrandw: Not that I'm aware of. Since this condition will be true sometimes if there is no runtime, we can't just wait until a runtime is returned, we need a threshold condition like a wait to stop this from infinitely waiting until a runtime is returned. I'm open to suggestions though!

jdcanas: Actually, I think we can just delete this block entirely. There's no real need for this component to ask the user the create a runtime if one doesn't exist. 

The relevant pages (the app and analysis launcher) display an intelligent message if the user ends up on it without a runtime. 

Furthermore, the AppLauncher even prompts the user to create a runtime if one doesn't exist. The AnalysisLauncher doesn't, but we only allow users to launch an app if a runtime exists for it anyway. 

Finally, even if a user finds themselves in this situation, they can also just create a runtime from the context bar.

cahrens: That would be awesome-- thanks for thinking through the scenarios @jdcanas!

breilly2: Yeah, there's definitely a mix here of treating `null` as both "data has not been returned yet" and "returned data indicates that there is no current runtime". If we do need to keep this block, we need to add a little bit of state to track this.

cahrens: I believe the convention in the codebase is that `undefined` is used for "data has not been returned yet" and `null` is used to indicate that the server state is empty.

cstrandw: @jdcanas would removing the block I added and then passing `onNullRuntime: () => undefined` on [this line](https://github.com/DataBiosphere/terra-ui/blob/e786d1035ffea48f06529f708e5c47b39d1e3468/src/pages/workspaces/workspace/notebooks/NotebookLauncher.js#L82) be a better solution?

jdcanas: I think `onNullRuntime` can also be removed entitrely

jdcanas (approved):
LGTM, though the test failure screenshot looks a bit suspicious...

breilly2:
Sorry for the delay. I've been thinking through this.

My understanding is that removing `onNullRuntime` changes the behavior so that it doesn't automatically pop up the runtime create modal if there isn't a runtime. Is that correct? Is that a behavior change that we want? Can this problem still happen if someone manually opens the modal quickly after a reload?

cstrandw:
> Sorry for the delay. I've been thinking through this.
> 
> My understanding is that removing `onNullRuntime` changes the behavior so that it doesn't automatically pop up the runtime create modal if there isn't a runtime. Is that correct? Is that a behavior change that we want? Can this problem still happen if someone manually opens the modal quickly after a reload?

Yes, removing `onNullRuntime` does change the behavior in that way but there are other warnings/ prompts provided to the user to help them that does not cause the user to be able to create a duplicate runtime (see @jdcanas's [comment](https://github.com/DataBiosphere/terra-ui/pull/3020#discussion_r869711658)). This issue is cropping up in a variety of places in the UI, so removing this function helps a few ongoing bugs. Also, it is technically possible for a user to create this issue manually, but it is pretty hard (when I was trying to recreate I was able to open the ComputeModal to create a new runtime only once out of the 8 times, and that was only because I was trying to be very fast)

breilly2:
Thank you for the response, @cstrandw. I will approve this PR in the interest of reducing the potential impact on users by removing a behavior that is more of a convenience than a necessity.

However, I don't think that this is the ultimate solution. Something in the UI code is causing it to make a decision about whether or not there is a runtime before it has all of the information needed to make that decision. That is still true after this change. If we solve the root cause of that problem, then we wouldn't have to remove this convenience behavior.

breilly2 (approved):
👍  This will definitely reduce the likelihood that users will get into a bad state of having 2 runtimes, which is very valuable. I would like to see the root cause investigated and fixed at some point.

src/pages/workspaces/workspace/notebooks/NotebookLauncher.js
```diff
@@ -244,7 +244,12 @@ const PreviewHeader = ({ queryParams, runtime, readOnlyAccess, onCreateRuntime,
             onClick: () => setFileInUseOpen(true)
           }, [makeMenuIcon('lock'), 'Open (In use)'])],
           () => h(HeaderButton, {
-            onClick: () => chooseMode('edit')
+            onClick: () => {
```
cstrandw, line 247: @jdcanas @breilly2 I ended up making another change to trigger the `ComputeModal` to open when a user clicks the `Open` button when there is no existing runtime. This was causing the `run-notebook` integration test to fail and appeared very confusing as nothing happened when pressing that button in this condition.

jdcanas (approved):

PR #3021, opened 2022-05-10
[QA-1839] reload sign-in page once if loading had failed

aweng98:
Failed [CircleCI](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9416/workflows/e128c57a-4bf5-4763-b18c-e9e83f63bf1a).

Tests still failing when log in immediately after new gcloud deploy. [PR #3004](https://github.com/DataBiosphere/terra-ui/pull/3004) has improved stability but did not fix issue completely.

Attempt again to fix loading issue: Reload page once if opening Terra UI (Home page) had failed.

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -191,29 +191,29 @@ const dismissNotifications = async page => {
   return !!notificationCloseButtons.length && delay(1000) // delayed for alerts to animate off
 }
 
-const signIntoTerra = async (page, { token, testUrl }) => {
-  if (!!testUrl) {
-    const timeout = 30 * 1000
-    const pageLoadedPromise = Promise.race([
+const loadSignInPageAndWaitForReady = async (page, url, timeout = 30 * 1000) => {
+  return Promise.all([
+    page.goto(url, waitUntilLoadedOrTimeout(timeout)),
+    Promise.race([
       page.waitForXPath('//*[@id="signInButton"]', { visible: true, timeout }),
       page.waitForXPath('//a[@href="#workspaces"]', { visible: true, timeout })
     ])
-    const loadPagePromise = Promise.all([
-      page.goto(testUrl, waitUntilLoadedOrTimeout(timeout)),
-      pageLoadedPromise
-    ])
-      .then(() => waitForNoSpinners(page))
+  ])
+    .then(() => waitForNoSpinners(page))
+}
 
+const signIntoTerra = async (page, { token, testUrl }) => {
+  if (!!testUrl) {
     try {
-      await loadPagePromise
+      await loadSignInPageAndWaitForReady(page, testUrl)
     } catch (err) {
       console.error(err)
       console.error(`Error: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
       await page.reload({ waitUntil: 'load' })
-      await loadPagePromise
+      await loadSignInPageAndWaitForReady(page, testUrl)
     }
   }
-
+  
   await page.evaluate(token => window.forceSignIn(token), token)
   await dismissNotifications(page)
   await waitForNoSpinners(page)
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -206,8 +206,9 @@ const signIntoTerra = async (page, { token, testUrl }) => {
 
     try {
       await loadPagePromise
-    } catch (e) {
-      console.log(`Warning: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
+    } catch (err) {
+      console.error(err)
+      console.log(`Error: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
       await page.reload({ waitUntil: 'load' })
       await loadPagePromise
     }
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -191,13 +191,10 @@ const dismissNotifications = async page => {
   return !!notificationCloseButtons.length && delay(1000) // delayed for alerts to animate off
 }
 
-const loadSignInPageAndWaitForReady = (page, url, timeout = 30 * 1000) => {
-  return Promise.all([
-    page.goto(url, waitUntilLoadedOrTimeout(timeout)),
-    Promise.race([
-      page.waitForXPath('//*[@id="signInButton"]', { visible: true, timeout }),
-      page.waitForXPath('//a[@href="#workspaces"]', { visible: true, timeout })
-    ])
+const waitForSignInPage = async (page, timeout = 30 * 1000) => {
+  await Promise.race([
+    page.waitForXPath('//*[@id="signInButton"]', { visible: true, timeout }),
+    page.waitForXPath('//a[@href="#workspaces"]', { visible: true, timeout })
   ])
     .then(() => waitForNoSpinners(page))
 }
@@ -205,13 +202,14 @@ const loadSignInPageAndWaitForReady = (page, url, timeout = 30 * 1000) => {
 const signIntoTerra = async (page, { token, testUrl }) => {
   if (!!testUrl) {
     try {
-      console.log(`Loading page: ${testUrl}`)
-      await loadSignInPageAndWaitForReady(page, testUrl)
+      console.log(`Loading sign-in page: ${testUrl}`)
+      await page.goto(testUrl, waitUntilLoadedOrTimeout())
+      await waitForSignInPage(page)
     } catch (err) {
       console.error(err)
       console.error(`Error: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
       await page.reload({ waitUntil: 'load' })
-      await loadSignInPageAndWaitForReady(page, testUrl)
+      await waitForSignInPage(page)
     }
   }
 
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -208,7 +208,7 @@ const signIntoTerra = async (page, { token, testUrl }) => {
       await loadPagePromise
     } catch (err) {
       console.error(err)
-      console.log(`Error: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
+      console.error(`Error: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
       await page.reload({ waitUntil: 'load' })
       await loadPagePromise
     }
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -194,27 +194,22 @@ const dismissNotifications = async page => {
 const signIntoTerra = async (page, { token, testUrl }) => {
   if (!!testUrl) {
     const timeout = 30 * 1000
-    const welcomePagePromise = Promise.all([
-      page.waitForXPath('//a[@href="#workspaces"]', { visible: true, timeout }),
-      page.waitForXPath('//a[@href="#library/showcase"]', { visible: true, timeout }),
-      page.waitForXPath('//a[@href="#library/datasets"]', { visible: true, timeout })
-    ])
-    const pagePromise = Promise.race([
+    const pageLoadedPromise = Promise.race([
       page.waitForXPath('//*[@id="signInButton"]', { visible: true, timeout }),
-      welcomePagePromise
+      page.waitForXPath('//a[@href="#workspaces"]', { visible: true, timeout })
     ])
-    const gotoPromise = Promise.all([
+    const loadPagePromise = Promise.all([
       page.goto(testUrl, waitUntilLoadedOrTimeout(timeout)),
-      pagePromise
+      pageLoadedPromise
     ])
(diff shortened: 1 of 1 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -190,9 +190,27 @@ const dismissNotifications = async page => {
 }
 
 const signIntoTerra = async (page, { token, testUrl }) => {
-  !!testUrl && await page.goto(testUrl, waitUntilLoadedOrTimeout(60 * 1000))
-  await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout: 60 * 1000 })
-  await waitForNoSpinners(page)
```
aweng98, line 197: An issue with waiting for spinner during page loading is that we can "miss" the spinner when it has finished checking before it renders.

integration-tests/utils/integration-utils.js
```diff
@@ -190,9 +190,27 @@ const dismissNotifications = async page => {
...
+    const loadPagePromise = Promise.all([
+      page.goto(testUrl, waitUntilLoadedOrTimeout(timeout)),
+      pageLoadedPromise
+    ])
+      .then(() => waitForNoSpinners(page))
+
+    try {
+      await loadPagePromise
+    } catch (e) {
+      console.log(`Warning: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
```
cahrens: Is it worth logging `e` in case the cause of failure is something else?

rtitle:
Just saw this, we should make sure it works with https://github.com/DataBiosphere/terra-ui/pull/2949 which is changing the sign in flow. :) Will take a closer look in a few..

integration-tests/utils/integration-utils.js
```diff
@@ -190,9 +190,27 @@ const dismissNotifications = async page => {
 }
 
 const signIntoTerra = async (page, { token, testUrl }) => {
-  !!testUrl && await page.goto(testUrl, waitUntilLoadedOrTimeout(60 * 1000))
-  await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout: 60 * 1000 })
-  await waitForNoSpinners(page)
+  if (!!testUrl) {
+    const timeout = 30 * 1000
+    const pageLoadedPromise = Promise.race([
+      page.waitForXPath('//*[@id="signInButton"]', { visible: true, timeout }),
```
rtitle: The new button has the same id ✅

integration-tests/utils/integration-utils.js
aweng98: Sure, I'll add that. 
I was thinking of same thing. But I didn't do it because if reload has worked, I don't really care to see it. If still fails on the second attempt, (same) error will be thrown.

aweng98 (commented):
@rtitle Hi, should I be changing anything here to make sure it won't break your PR or break tests?

cahrens (approved):

aweng98:
@rtitle Hi Rob, I merged your PR #2949 with my changes to test. Here is the CircleCI [job](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9481/workflows/03fe4434-2272-4016-aec8-ef7d4c7e4df7). I don't find any failure on sign in. There is one test that failed with ajax error which you might want to take a look. It could be caused by your PR.

cahrens:
> @rtitle Hi Rob, I merged your PR #2949 with my changes to test. Here is the CircleCI [job](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9481/workflows/03fe4434-2272-4016-aec8-ef7d4c7e4df7). I don't find any failure on sign in. There is one test that failed with ajax error which you might want to take a look. It could be caused by your PR.

I took a look, and the failure is unrelated. It looks like there is a flakiness around a billing project being selected on the billing page (the UI shows project as selected, but then also says on the right that you need to select one). Unrelated to signin.

breilly2:
The PR title is "reload sign-in page once if loading had failed" but it's not clear to me from reading the code that that's what's actually happening. Is that still an accurate description of this change?

breilly2:
The PR title is "reload sign-in page once if loading had failed" but it's not clear to me from reading the code that that's what's actually happening. Is that still an accurate description of this change?

integration-tests/utils/integration-utils.js
```diff
@@ -190,9 +190,28 @@ const dismissNotifications = async page => {
...
+    ])
+      .then(() => waitForNoSpinners(page))
+
+    try {
+      await loadPagePromise
+    } catch (err) {
+      console.error(err)
+      console.error(`Error: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
+      await page.reload({ waitUntil: 'load' })
+      await loadPagePromise
```
nawatts: Does this not re-throw the error, since the catch block would only run if `loadPagePromise` is rejected?

integration-tests/utils/integration-utils.js
kyuksel: @aweng98 IIUC the issue you describe in your comment above, could wrapping the original code in a [noSpinnersAfter](https://github.com/DataBiosphere/terra-ui/blob/e69798efefbadf368f737876d84fd8358b45f6ad/integration-tests/utils/integration-utils.js#L166) suffice?

aweng98: No, I don't think that helps. The best way (IMO) is by finding a visible UI element (or more) first to verify page has rendered then checks and waits for running spinner (ajax loading).

cahrens: I don't think it will work because page.goto changes the page, and `noSpinnersAfter` checks that a spinner appears and disappear within the same page.

aweng98:
> 

Primary objective is to reload the sign-in page. I refactored the function to avoid duplicate code. At the same time, new code eliminates flakiness related to step logic.
I'm terrible at explaining things, suggestion is welcomed. let me know. Thanks.

integration-tests/utils/integration-utils.js
kyuksel: I see. Thanks.

integration-tests/utils/integration-utils.js
kyuksel: Not necessarily. IIUC the goal is to try reloading if loading failed previously. If the reloading succeeds, the test would proceed with subsequent steps.

kyuksel (commented):
I'm a little concerned about the code complexity we may be introducing as it took me sometime to wrap my head around it but I couldn't come up with a suggestion for improvement. 

I'll let @nawatts formally approve since he had asked a question.

integration-tests/utils/integration-utils.js
aweng98: Sorry, I don't fully understand your question. Error will be thrown if failing on the second attempt (L211),  and that should fails (stops) the test. If both the first attempt (L206) and the second attempt (L211) have failed, it might be due to the same root cause.

nawatts: How is line 211 is making a second attempt at anything though? It's awaiting a promise that's already settled (rejected). If the catch block is run, that means `loadPagePromise` was rejected. And if `loadPagePromise` is not changed inside the catch block, how would awaiting it a second time produce a different result?

aweng98: Why do you think L211 is waiting for a promise that's already done (L206)? I may have misunderstanding about async and promises. Would you help me understand your concern better by explain this over a google meet?

kyuksel: Oh yea @nawatts is right. @aweng98 did you mean to do something like below?

```javascript
const loadHomePageAndGoToTestUrl = async (page, testUrl) => {
  const timeout = 30 * 1000

  return Promise.all([
    page.goto(testUrl, waitUntilLoadedOrTimeout(timeout)),
    Promise.race([
      page.waitForXPath('//*[@id="signInButton"]', { visible: true, timeout }),
      page.waitForXPath('//a[@href="#workspaces"]', { visible: true, timeout })
    ])
  ])
    .then(() => waitForNoSpinners(page))
}

  if (!!testUrl) {
    try {
      await loadHomePageAndGoToTestUrl(page, testUrl)
    } catch (err) {
      console.error(err)
      console.error(`Error: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
      await page.reload({ waitUntil: 'load' })
      await loadHomePageAndGoToTestUrl(page, testUrl)
    }
  }

  await page.evaluate(token => window.forceSignIn(token), token)
  await waitForNoSpinners(page)
}
```

nawatts: A promise represents the future result of an asynchronous operation. Awaiting the promise does not run the operation, it blocks until the operation is finished.

You can see this in the console with something like:
```js
const sleep = (t) => {
    return new Promise(resolve => {
        setTimeout(resolve, t * 1000)
    })
}

(async () => {
    const promise = sleep(3)
    
    console.time()
    await promise
    console.timeEnd()

    console.time()
    await promise
    console.timeEnd()
})()
```

The first `await promise` takes time because it has to wait until `sleep` is finished. When `await promise` is called again, it returns quickly because `sleep` is not re-run.

nawatts: Agreed with @kyuksel, I suspect the intention here may have been:

```js
try {
  await doSomething()
} catch (err) {
  // handle error
  await doSomething()
}
```
vs
```js
const promise = doSomething()
try {
  await promise
} catch (err) {
  // handle error
  await promise
}
```

aweng98: Ooh, crap. I understood finally. `await loadPagePromise` call are missing parenthesis. Good Catch! I wish I saw this earlier.

kyuksel: To be clear, the issue is not that `await loadPagePromise` is missing parentheses. The issue is due to the fact that an action wrapped in a `Promise` starts being executed as soon as it is defined (similarly to `Scala` `Future` if you're familiar with that). For instance, in the case of

```javascript
const someActionPromise = page.waitForXPath('//*[@id="signInButton"]', { visible: true, timeout })
```

it starts waiting for the specified page element as soon as the above line is executed. The way to "lazily" define it such that this doesn't happen would be making it a function, like so:

```javascript
const someActionPromise = async () => page.waitForXPath('//*[@id="signInButton"]', { visible: true, timeout })
```

now the waiting action will start only once we execute the function:

```javascript
someActionPromise()
```

regardless of whether we `await` the result where we call the function or not.

aweng98: Fixed problem and took Kaan's code suggestion. Thank you all.

integration-tests/utils/integration-utils.js
```diff
@@ -191,10 +191,29 @@ const dismissNotifications = async page => {
   return !!notificationCloseButtons.length && delay(1000) // delayed for alerts to animate off
 }
 
+const loadSignInPageAndWaitForReady = (page, url, timeout = 30 * 1000) => {
```
kyuksel: Nit: The following reads a bit more naturally to me.

```suggestion
const loadSignInPageAndWaitUntilReady = (page, url, timeout = 30 * 1000) => {
```

kyuksel (commented):
The last set of code changes look good to me. One nit inline.

Again, will let @nawatts give the formal approval in case I'm missing something again.

integration-tests/utils/integration-utils.js
```diff
@@ -191,10 +191,30 @@ const dismissNotifications = async page => {
...
+  return Promise.all([
+    page.goto(url, waitUntilLoadedOrTimeout(timeout)),
+    Promise.race([
```
nawatts: Should `page.goto` and the two `waitForXPath` in `Promise.race` start at the same time? Or should the `waitForXPath`s be run after `page.goto` is done?

integration-tests/utils/integration-utils.js
```diff
@@ -191,10 +191,30 @@ const dismissNotifications = async page => {
...
-  await waitForNoSpinners(page)
+  if (!!testUrl) {
+    try {
+      console.log(`Loading page: ${testUrl}`)
+      await loadSignInPageAndWaitForReady(page, testUrl)
+    } catch (err) {
+      console.error(err)
+      console.error(`Error: Page loading timed out during sign in. Reloading URL: "${testUrl}"`)
+      await page.reload({ waitUntil: 'load' })
+      await loadSignInPageAndWaitForReady(page, testUrl)
```
nawatts: Does `page.goto` in this function cause the page to be loaded again after `page.reload`?

integration-tests/utils/integration-utils.js
aweng98: 👍🏻 It should be in sequential order, `goto` first and done, then `waitForXpath` in `Promise.race`. It's fixed now.

integration-tests/utils/integration-utils.js
aweng98: 😞 yeah, looks that way. How about the updated code?

nawatts (approved):

aweng98:
hooray Thank you all.

PR #3023, opened 2022-05-11
[DE-348] Added "Preview Mode" toggler for the new catalog service

lu-c:
Added ability to toggle between the new catalog service and the old data page.

- The mode you choose will be stored in localstorage, and when you click "browse data", it will pick the mode you last viewed.
- Added hero banner under the covid one
- The link for feedback does not exist right now, so we will have to patch that through again later

Screen Shot 2022-05-11 at 2 25 13 PM
Screen Shot 2022-05-11 at 2 25 19 PM
Screen Shot 2022-05-11 at 2 25 26 PM


integration-tests/tests/preview-dataset.js
```diff
@@ -7,7 +7,7 @@ const datasetName = 'Readable Catalog Snapshot 1'
 
 const testPreviewDatasetFn = withUserToken(async ({ testUrl, page, token }) => {
   await enableDataCatalog(page, testUrl, token)
-  await click(page, clickable({ textContains: 'browse & explore' }))
+  await click(page, clickable({ textContains: 'datasets' }))
```
lu-c, line 10: Had to make this change because the new browser page is intended to replace the datasets page and the mockups have now removed the "browse and explore" tab

src/pages/library/dataBrowser-utils.js
```diff
@@ -100,6 +103,11 @@ const extractTags = dataset => {
   }
 }
 
+export const getDatasetsPath = ({ user: { id } }) => {
+  return isDataBrowserVisible() && catalogPreviewStore.get()[id] ?
```
nawatts: I enabled the data browser using `configOverridesStore.set({ isDataBrowserVisible: true })` and reloaded. The UI crashed with:
```
Uncaught TypeError: Cannot read properties of null (reading 'undefined')
    at getDatasetsPath (dataBrowser-utils.js:107:1)
    at LandingPage (LandingPage.js:67:1)
    ...
```

It worked after running `localStorage.setItem('catalog-beta-preview', '{}')`.

nawatts: The Notebooks/Analyses tab has a similar "preview a new feature" button. In that case, the state is stored in config overrides / session storage. Is there a preference for one way or the other? Using config overrides instead of this new store would also avoid the complexity of passing the user object around / storing the user ID in local storage.

https://github.com/DataBiosphere/terra-ui/blob/f4b5658dbeaa275a8599fa855e5d696e5a163ed6/src/pages/workspaces/workspace/Notebooks.js#L454-L465

src/components/TopBar.js
```diff
@@ -215,14 +215,14 @@ const TopBar = ({ showMenu = true, title, href, children }) => {
               href: Nav.getLink('library-datasets'),
               onClick: hideNav
             }, ['Data']),
+            isDataBrowserVisible() && h(DropDownSubItem, {
+              href: Nav.getLink('library-browser'),
+              onClick: hideNav
+            }, ['Try the BETA Data Catalog']),
```
nawatts, line 218:
```suggestion
            }, ['Data Catalog (BETA)']),
```

A phrase here seems out of place since the other options are all nouns.

src/pages/library/DataBrowserToggler.js
```diff
@@ -0,0 +1,52 @@
...
+import { useStore } from 'src/libs/react-utils'
+import { authStore } from 'src/libs/state'
+import { catalogPreviewStore } from 'src/pages/library/dataBrowser-utils'
+
+
+export const DataBrowserPreviewToggler = ({ checked }) => {
+  const { user: { id } } = useStore(authStore)
+  catalogPreviewStore.set({ [id]: checked })
+
+  return !isDataBrowserVisible() ? div() : div({
```
nawatts, line 15: Should this component be rendered at all if the data browser is not enabled?

Otherwise, could this be simplified to avoid rendering the empty div?

```suggestion
  return !!isDataBrowserVisible() && div({
```

src/pages/library/DataBrowserToggler.js
```diff
@@ -0,0 +1,52 @@
...
+      padding: '10px 15px',
+      margin: 15,
+      border: '1px solid', borderColor: colors.accent(), borderRadius: 3,
+      display: 'flex', flexDirection: 'row'
+    }
+  }, [
+    div({ style: { display: 'flex', flexDirection: 'column' } }, [
+      strong(['Toggle to preview the new Data Catalog']),
+      div({ style: { display: 'flex', flexDirection: 'row', alignItems: 'center', marginTop: 6 } }, [
+        h(Switch, {
```
nawatts, line 30: Is a Switch the best interface to use for this? I think it would be surprising for changing a checkbox to cause page navigation. Would a link or button be more appropriate?

src/pages/library/DataBrowserToggler.js
```diff
@@ -0,0 +1,52 @@
...
+          onChange: () => {
+            catalogPreviewStore.set({ [id]: !checked })
+            if (checked) {
+              Nav.goToPath('library-datasets')
+            } else {
+              Nav.goToPath('library-browser')
+            }
+          }
+        }),
+        strong({ style: { marginLeft: 10 } }, ['BETA Data Catalog ON'])
```
nawatts: This says "On" regardless of whether or not the beta catalog is enabled or not.

src/pages/LandingPage.js
```diff
@@ -86,6 +90,23 @@ const LandingPage = () => {
...
+        color: 'white', padding: '2rem 1rem',
+        width: 'calc(675px + 2rem)', // 3 card widths + 2 card margins to line up with the content directly above
+        marginTop: 15
+      }
+    }, [
+      h2({ style: { fontSize: 18, fontWeight: 500, lineHeight: '22px', margin: 0 } }, ['Preview the BETA Data Catalog and provide valuable feedback.']),
+      h(ButtonOutline, {
+        style: { marginLeft: '2rem', padding: '1.5rem 1rem', textTransform: 'none' },
+        onClick: () => { Nav.goToPath('library-browser') }
+      }, ['Preview BETA Data Catalog'])
```
nawatts, line 123: Is there a way to reword this so that the phrase "Preview the BETA Data Catalog" isn't repeated?

To be more consistent with the COVID banner above it, maybe something like "BETA Data Catalog" as the heading and below that in normal text, "Preview the Data Catalog and provide valuable feedback." with "Preview the Data Catalog" or the whole sentence linked to the catalog?

nawatts (changes requested):
A few things, but mainly requesting changes to handle the case where `catalog-beta-preview` is not set in local storage. Currently, this causes the UI to crash. To reproduce, run `localStorage.removeItem('catalog-beta-preview')`.

src/pages/library/DataBrowserToggler.js
lu-c: I tried doing that initially, and it complained that this function must return an object!

src/components/TopBar.js
lu-c: I want to stick with the mock ups UX gave us for this wording, and revisit it later. Right now, at least, this feature is not ready to be turned on anyway, so there will be things we need to do before that too

src/pages/LandingPage.js
lu-c: I like this, going to make that change! :)

src/pages/library/DataBrowserToggler.js
lu-c: That's true, I think it does feel kind of unexpected... but in the browser, it's not entirely obvious that it's a new page loaded, does it? It looks like you are just toggling a setting, since the tab doesnt actively look like it's changed

src/pages/library/DataBrowserToggler.js
lu-c: Fixed

src/pages/library/dataBrowser-utils.js
lu-c: Fixed this bug. I'd be okay with making that change to use a new configOverridesStore, just trying to think through what the pros and cons of this would be. I'm using the localstorage because I think that's also tied to user... Do we care if someone logs out, the next person who uses the same browser/computer ends up getting the same state? Since you can toggle it back easily, I dont think it's that important 🤔 

The other thing that I'm also thinking about is that this is only supposed to be around as long as we are maintaining both versions. The catalog is supposed to eventually replace it, which means this localstorage vs configoverrides is eventually all going to have to go away.

Dunno! Looking for other opinions on this too! Thanks :)

src/pages/library/DataBrowserToggler.js
nawatts: It doesn't _look_ like it, but the navigation is noticeable in other ways. Focus gets moved back to the top of the page and if you're using a screen reader, it reads the new page. I leave it up to you which way to go.

However, one thing I noticed while trying this with a screen reader is that the switch doesn't have a label. It currently reads as "off, switch" or "on, switch". There's already some descriptive text here, it just needs to be wrapped in a `label` tag with the Switch component. The usage section on https://www.npmjs.com/package/react-switch has an example.

src/pages/LandingPage.js
```diff
@@ -86,6 +90,26 @@ const LandingPage = () => {
...
+        display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems: 'center',
+        boxShadow: '0 2px 5px 0 rgba(0,0,0,0.35), 0 3px 2px 0 rgba(0,0,0,0.12)',
+        color: 'white', padding: '2rem 1rem',
+        width: 'calc(675px + 2rem)', // 3 card widths + 2 card margins to line up with the content directly above
+        marginTop: 15
+      }
+    }, [
+      div([
+        h2({ style: { fontSize: 18, fontWeight: 500, lineHeight: '22px', margin: 0 } }, ['BETA Data Catalog']),
+        'Preview the Data Catalog and provide valuable feedback'
```
nawatts: Ending this sentence with a period would make it consistent with the COVID section.

nawatts (changes requested):
One last thing... the toggle switch needs an accessible label.

s-rubenstein:
I think to link this to Jira it should be `[DC-348]` rather than `[DE-348]` in the title

[DC-348]: https://broadworkbench.atlassian.net/browse/DC-348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/pages/LandingPage.js
```diff
@@ -86,6 +90,26 @@ const LandingPage = () => {
...
       ' for a summary of available resources.'
     ]),
+    isDataBrowserVisible() && div({
+      style: {
+        backgroundColor: colors.primary(),
+        backgroundImage: `url(${terraHero})`, backgroundSize: 'cover', borderRadius: 5,
+        display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems: 'center',
+        boxShadow: '0 2px 5px 0 rgba(0,0,0,0.35), 0 3px 2px 0 rgba(0,0,0,0.12)',
+        color: 'white', padding: '2rem 1rem',
+        width: 'calc(675px + 2rem)', // 3 card widths + 2 card margins to line up with the content directly above
```
s-rubenstein: Do you think it would be worth extracting the card width into a variable, and then making this more explicit/removing the comment and making it:
`width: 'calc(cardWidth * 3 + 'px + ' cardMargin * 2 + 'rem')'`?

That would have the benefit of automatically changing this to match if the cards are changed.

src/pages/library/DataBrowserToggler.js
```diff
@@ -0,0 +1,52 @@
...
+            }
+          }
+        }),
+        strong({ style: { marginLeft: 10 } }, [`BETA Data Catalog ${checked ? 'ON' : 'OFF'}`])
+      ])
+    ]),
+    checked && div({ style: { marginLeft: 80 } }, [
+      'After previewing the Terra Data Catalog, please fill out this quick survey to provide the team with valuable feedback.',
+      h(Link, {
+        href: '', // TODO: Add when this is added.
```
fboulnois:
```suggestion
        href: '#', // TODO: Add when this is added.
```

Should this be something other than empty?

fboulnois (commented):
I like Sky's suggestion. One other question:

src/pages/library/DataBrowserToggler.js
lu-c: O, good call!

src/pages/LandingPage.js
lu-c: Updated!

src/pages/library/DataBrowserToggler.js
lu-c: Ok, updated the label so it is now wrapping the content - Let me know if this is okay!

nawatts (approved):

fboulnois (approved):

fboulnois:
I think it needs to be DC-348 to link to our Jira board, but otherwise lgtm.

src/pages/LandingPage.js
```diff
@@ -75,11 +89,10 @@ const LandingPage = () => {
...
     isTerra() && div({
       style: {
+        ...styles.callToActionBanner,
         backgroundColor: '#191a1c', // This fallback color was extracted from the left edge of the background image
-        backgroundImage: `url(${covidHero})`, backgroundSize: 'cover', borderRadius: 5,
-        boxShadow: '0 2px 5px 0 rgba(0,0,0,0.35), 0 3px 2px 0 rgba(0,0,0,0.12)',
-        color: 'white', padding: '2rem 1rem',
-        width: 'calc(675px + 2rem)' // 3 card widths + 2 card margins to line up with the content directly above
+        backgroundImage: `url(${covidHero})`,
+        width: `calc(${styles.card.width * 3}px + ${styles.card.marginRight} * 2)`
```
s-rubenstein, line 95: I think you need to add `rem` to the end of the margin in the calc string?

src/pages/LandingPage.js
```diff
@@ -92,12 +105,11 @@ const LandingPage = () => {
...
       style: {
+        ...styles.callToActionBanner,
         backgroundColor: colors.primary(),
-        backgroundImage: `url(${terraHero})`, backgroundSize: 'cover', borderRadius: 5,
+        backgroundImage: `url(${terraHero})`,
         display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems: 'center',
-        boxShadow: '0 2px 5px 0 rgba(0,0,0,0.35), 0 3px 2px 0 rgba(0,0,0,0.12)',
-        color: 'white', padding: '2rem 1rem',
-        width: 'calc(675px + 2rem)', // 3 card widths + 2 card margins to line up with the content directly above
+        width: `calc(${styles.card.width * 3}px + ${styles.card.marginRight} * 2)`,
```
s-rubenstein, line 112: Add rem to end of 2 in the calc string.

s-rubenstein (approved):
LGTM. One change, but the merge doesn't need to block on another round of reviews from me.

src/pages/LandingPage.js
nawatts: This looks correct as is. `styles.card.marginRight` is a string with `rem` units. Thus, the other operand for multiplication must be a number (without units) vs a length.

https://developer.mozilla.org/en-US/docs/Web/CSS/calc#syntax

lu-c: If you run this locally tho, it works - I wouldn't be pushing up code without testing it!

PR #3024, opened 2022-05-11
[QA-1864] Terra-UI integration log handler jest-reporter.js fails to capture all logs

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1864

### DIFF
--- a/integration-tests/jest-reporter.js
+++ b/integration-tests/jest-reporter.js
@@ -32,8 +32,7 @@ module.exports = class JestReporter {
     const writableStream = createWriteStream(logFileName)
     writableStream.on('error', ({ message }) => console.error(`Error occurred while writing Console logs to ${logFileName}.\n${message}`))
 
-    testResult.console.forEach(({ message }) =>
-      writableStream.write(`${message}\n`)
+    testResult.console.forEach(({ message }) => writableStream.write(`${message}\n`)
     )
 
     // Write pass/failure summaries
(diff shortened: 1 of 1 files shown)


integration-tests/jest-reporter.js
```diff
@@ -32,7 +32,9 @@ module.exports = class JestReporter {
     const writableStream = createWriteStream(logFileName)
     writableStream.on('error', ({ message }) => console.error(`Error occurred while writing Console logs to ${logFileName}.\n${message}`))
 
-    _.forEach(({ message }) => writableStream.write(`${message}\n`), testResult.console)
```
aweng98, line 35: `_.forEach` is the problem. It didn't get every message in `testResult.console` array.

integration-tests/jest-reporter.js
```diff
@@ -32,7 +32,8 @@ module.exports = class JestReporter {
...
+    testResult.console.forEach(({ message }) => writableStream.write(`${message}\n`)
```
petesantos, line 35: What is the structure of `testResult.console`? I am curious as to why `_.forEach` missed items

petesantos (approved):

integration-tests/jest-reporter.js
aweng98: It's an array. Each element is a Json object. I suspect `_.forEach` fails to "see" all elements because array is very large. I don't have proof for this theory.

Screen Shot 2022-05-13 at 11 29 54 AM

kyuksel: > I suspect _.forEach fails to "see" all elements because array is very large

I can't imagine how that'd be and wonder if something else was happening at the same time when the issue occurred. I don't want to block the merge though.

kyuksel (approved):
Pretty strange about `_.forEach` but  🤷🏼

PR #3025, opened 2022-05-11
[QA-1865] fetch Lyle request err handling

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1865

`fetchLyle()` can fail sometimes. We want to understand why by seeing response body if it fails next time.

CircleCI [job](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9451/workflows/ebef13e3-c33c-4560-9e5a-a495d4a49d4f/jobs/42285)

Error signature:
```
FetchError: invalid json response body at https://terra-lyle.appspot.com/api/delete reason: Unexpected token E in JSON at position 0

       98 |     await test({ ...args, email, token })
       99 |   } finally {
    > 100 |     await fetchLyle('delete', email)
          |     ^
      101 |   }
      102 | }
      103 |

      at ../.yarn/__virtual__/node-fetch-virtual-10f4bc17d9/0/cache/node-fetch-npm-2.6.7-777aa2a6df-8d816ffd1e.zip/node_modules/node-fetch/lib/index.js:273:32
      at utils/integration-helpers.js:100:5
      at utils/integration-utils.js:289:12
 ```

### DIFF
--- a/integration-tests/utils/lyle-utils.js
+++ b/integration-tests/utils/lyle-utils.js
@@ -25,7 +25,7 @@ const fetchLyle = async (path, email) => {
       headers: { 'Content-Type': 'application/json', ...(await authClient.getRequestHeaders(url)) },
       body: JSON.stringify({ email })
     })
-    console.log(`fetchLyle: ${res.status}, ${url}`)
+    console.log(`fetchLyle: POST ${res.status} ${url}`)
     if (res.ok) {
       // response.status >= 200 && response.status < 300
       return res.json()
--- a/integration-tests/utils/lyle-utils.js
+++ b/integration-tests/utils/lyle-utils.js
@@ -24,12 +24,14 @@ const fetchLyle = async (path, email) => {
     headers: { 'Content-Type': 'application/json', ...(await authClient.getRequestHeaders(url)) },
     body: JSON.stringify({ email })
   })
-    .then(resp => {
+    .then(async resp => {
       if (resp.ok) {
         // response.status >= 200 && response.status < 300
         return resp.json()
       }
-      throw Error(`fetch Lyle response: ${resp.text()}`)
+      const respText = await resp.text()
+      console.error(`fetch Lyle response: ${respText}`)
+      return respText
     })
     .catch(async err => {
       console.error(err)
--- a/integration-tests/utils/lyle-utils.js
+++ b/integration-tests/utils/lyle-utils.js
@@ -27,13 +27,10 @@ const fetchLyle = async (path, email) => {
     })
     console.log(`fetchLyle: POST ${res.status} ${url}`)
     if (res.ok) {
-      // response.status >= 200 && response.status < 300
       return res.json()
     }
     // delegate non-2xx response to enclosing try/catch
-    const error = new Error()
-    Object.assign(error, { response: res })
-    throw error
+    throw _.set('response', res, new Error())
   } catch (err) {
     console.error(err)
     const errorBody = await err.response.text()
--- a/integration-tests/utils/lyle-utils.js
+++ b/integration-tests/utils/lyle-utils.js
@@ -19,13 +19,24 @@ const fetchLyle = async (path, email) => {
   const url = `${lyleUrl}/api/${path}`
   const authClient = await makeAuthClient()
 
-  const res = await fetch(url, {
+  return await fetch(url, {
     method: 'POST',
     headers: { 'Content-Type': 'application/json', ...(await authClient.getRequestHeaders(url)) },
     body: JSON.stringify({ email })
   })
-
-  return res.json()
+    .then(resp => {
+      if (resp.ok) {
+        // response.status >= 200 && response.status < 300
+        return resp.json()
+      }
+      throw Error(`fetch Lyle response: ${resp.text()}`)
+    })
+    .catch(async err => {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: integration-tests/utils/lyle-utils.js [hunk 1]
```diff
@@ -19,13 +19,24 @@ const fetchLyle = async (path, email) => {
...
+        return resp.json()
+      }
+      throw Error(`fetch Lyle response: ${resp.text()}`)
+    })
+    .catch(async err => {
+      console.error(err)
+      const errorBody = await err.response.text()
+      console.error(`fetch Lyle request response body: ${errorBody}`)
+      throw err
+    })
```
kyuksel, line 27: Trying to understand the new logic: Do we need to `throw`, `catch` and then `throw` again? Would something like the following accomplish the same?

```javascript
const fetchLyle = async (path, email) => {
  const url = `${lyleUrl}/api/${path}`
  const authClient = await makeAuthClient()

  const res = await fetch(url, {
    method: 'POST',
    headers: { 'Content-Type': 'application/json', ...(await authClient.getRequestHeaders(url)) },
    body: JSON.stringify({ email })
  })

  res.then(resp => {
    if (resp.ok) {
      // response.status >= 200 && response.status < 300
      return resp.json()
    } else {
      const msg = `fetch Lyle response: ${resp.text()}`
      console.error(msg)
      throw Error(msg)
    }
  })
}
```

aweng98: Oops! I didn't realize throw in `then` block will be caught by `catch` block. I fixed problem.

integration-tests/utils/lyle-utils.js
```diff
@@ -19,13 +19,26 @@ const fetchLyle = async (path, email) => {
...
+      const respText = await resp.text()
+      console.error(`fetch Lyle response: ${respText}`)
+      return respText
+    })
+    .catch(async err => {
+      console.error(err)
+      const errorBody = await err.response.text()
+      console.error(`fetch Lyle request response body: ${errorBody}`)
+      throw err
+    })
```
breilly2: This can also be written with try/catch which is more consistent with this codebase and is generally easier to read and reason about.
```
  try {
    const res = await fetch(url, {
      method: 'POST',
      headers: { 'Content-Type': 'application/json', ...(await authClient.getRequestHeaders(url)) },
      body: JSON.stringify({ email })
    })
    if (res.ok) {
      // response.status >= 200 && response.status < 300
      return res.json()
    } else {
      // delegate non-2xx response to enclosing try/catch
      throw { response: res, status: res.status }
    }
  } catch (err) {
    console.error(err)
    const errorBody = await err.response.text()
    console.error(`fetch Lyle request response body: ${errorBody}`)
    throw err
  }
```
Also, delegating to the outer try/catch on non-2xx responses eliminates some code duplication.

(note, I didn't actually test the code above, but something like that should work)

aweng98: This is nice. Code consistency is important. I'll update to use `try/catch`. 
I have one concern. I don't think we should `throw { response: res, status: res.status }` in `else` block. I prefer just to log response text and returns it. Then it can be used for negative testing where non-json response body is expected in fetch. Albeit I don't see any usage like that now. :)
What do you think if I keep the `else` code unchanged?

breilly2: The expected response from this function is JSON, so I'm against returning plain text. Also, this function is purely in support of these tests and I don't see a lot of value in performing negative tests of this function in isolation.

aweng98: Agreed. Updated function. Didn't really need the `else` string.

breilly2 (approved):

integration-tests/utils/lyle-utils.js
```diff
@@ -19,13 +19,27 @@ const fetchLyle = async (path, email) => {
...
+    console.log(`fetchLyle: POST ${res.status} ${url}`)
+    if (res.ok) {
+      // response.status >= 200 && response.status < 300
+      return res.json()
+    }
+    // delegate non-2xx response to enclosing try/catch
+    const error = new Error()
+    Object.assign(error, { response: res })
+    throw error
```
kyuksel: How would you feel about the following?

```suggestion
    if (res.ok) {
      return res.json()
    } else {
      // delegate non-2xx response to enclosing try/catch
      throw _.set('response', res, new Error())
    }
```

It would be a bit more concise; wouldn't mutate an object, and uses Lodash/FP for consistency. The first comment also seemed unnecessary.

aweng98: Like this?
```
if (res.ok) {
  return res.json()
}
// delegate non-2xx response to enclosing try/catch
throw _.set('response', res, new Error())
```	
Code looks clean. Just to confirm, do you know for sure if lodash/fp will work? We saw an issue with `_.forEach` in `jest-reporter.js` so I don't know if fp is a better choice.

kyuksel: Yes, we've been happily using Lodash/FP throughout the codebase. I suspect the root cause for the issue you mentioned was something else.

aweng98: Updated.

kyuksel (approved):

PR #3026, opened 2022-05-11
[SU-57] Refactor upsertEntities Ajax method

nawatts:
Several UI features for editing values in data tables (editing one or more cells in a table and adding/clearing a column) use the `upsertEntities` method to send changes to Rawls' [batch upsert entities](https://rawls.dsde-prod.broadinstitute.org/#/entities/batch_upsert_entities) endpoint.

Entities in data tables can have attributes of multiple types: strings, numbers, booleans, lists, and JSON (arrays or objects). Note that lists are different from JSON arrays.

Currently, setting JSON values cannot be done through the UI.

One obstacle to that is how the `upsertEntities` method accepts parameters. It accepts a list of objects `{ entityType, name, attributes: { attributeName: attributeValue, ... }` and converts the `attributes` object into a list of operations for the Rawls endpoint.

https://github.com/DataBiosphere/terra-ui/blob/f4b5658dbeaa275a8599fa855e5d696e5a163ed6/src/libs/ajax.js#L691-L697

https://github.com/DataBiosphere/terra-ui/blob/f4b5658dbeaa275a8599fa855e5d696e5a163ed6/src/libs/ajax.js#L622-L636

Since `attributesUpdateOps` converts array values into create list / add list item operations, `upsertEntities` cannot currently be used to set an a JSON array value for an attribute.

However, since #2895, we no longer need the behavior provided by `attributesUpdateOps`. We now set list attribute values with something like:
```
operations: [{
  op: "AddUpdateValue",
  attributeName: "attributeName",
  addUpdateAttribute: {
    itemsType: "AttributeValue",
    items: ["foo", "bar", "baz"]
  }
}]
```
Rawls handles this and and stores the value as a list.
https://broadinstitute.slack.com/archives/C02T8G2FJCQ/p1652220052756079?thread_ts=1650480934.910629&cid=C02T8G2FJCQ

This removes `attributesUpdateOps` from `upsertEntities`, which now takes lists of operations. This allows for finer-grained control over the attribute values set and will allow setting JSON array values through `upsertEntities`.

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -688,14 +688,6 @@ const Workspaces = signal => ({
     const root = `workspaces/${namespace}/${name}`
     const mcPath = `${root}/methodconfigs`
 
-    const upsertEntities = entities => {
-      const body = _.map(({ name, entityType, attributes }) => {
-        return { name, entityType, operations: attributesUpdateOps(attributes) }
-      }, entities)
-
-      return fetchRawls(`${root}/entities/batchUpsert`, _.mergeAll([authOpts(), jsonBody(body), { signal, method: 'POST' }]))
-    }
-
     return {
       checkBucketReadAccess: () => {
         return fetchRawls(`${root}/checkBucketReadAccess`, _.merge(authOpts(), { signal }))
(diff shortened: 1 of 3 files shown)


src/libs/ajax.js
```diff
@@ -964,7 +958,11 @@ const Workspaces = signal => ({
         const res = await fetchOk(url)
         const payload = await res.json()
 
-        return upsertEntities(payload)
+        const body = _.map(({ name, entityType, attributes }) => {
+          return { name, entityType, operations: attributesUpdateOps(attributes) }
+        }, payload)
+
+        return fetchRawls(`${root}/entities/batchUpsert`, _.mergeAll([authOpts(), jsonBody(body), { signal, method: 'POST' }]))
```
nawatts, line 965: `importJSON` can be tested with a JSON file such as:
```json
[
  { "entityType": "thing", "name": "thing_1", "attributes": { "value": "foo" } },
  { "entityType": "thing", "name": "thing_2", "attributes": { "value": "bar" } },
  { "entityType": "thing", "name": "thing_3", "attributes": { "value": "baz" } }
]
```

Serve the file with something like `npx http-server . --cors` and navigate to `http://localhost:3000/#import-data?format=entitiesJson&url=http://localhost:8080/things.json`.

MatthewBemis (approved):
lgtm

petesantos (commented):
When I test this code I am not seeing my workspaces on dev. I will look through the PR a bit more to see if I can identify why this is happening

UPDATE - this looks like an issue with my firefox installation, not an issue with the PR

petesantos (approved):

PR #3027, opened 2022-05-12
[WOR-153] Remove calls to cloud function ToS methods

cahrens:
We have been calling both the old cloud function ToS methods and the new Sam ones. Remove the cloud function methods now that the Sam API is working. Terra-ui has always enforced accepting the ToS, although the server API calls have not (and still are not because we are in an extended "grace period").

### DIFF
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -192,9 +192,7 @@ authStore.subscribe(withErrorReporting('Error checking registration', async (sta
 
 authStore.subscribe(withErrorReporting('Error checking TOS', async (state, oldState) => {
   if (!oldState.isSignedIn && state.isSignedIn) {
-    const acceptedCloudFunctionTos = await Ajax().User.getTosAccepted()
-    const acceptedSamTos = await Ajax().User.getSamTosAccepted()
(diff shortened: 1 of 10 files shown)


src/libs/auth.js
```diff
@@ -169,8 +169,8 @@ authStore.subscribe(withErrorReporting('Error checking registration', async (sta
     try {
       const { enabled } = await Ajax().User.getStatus()
       if (enabled) {
-        // Once the ToS grace period has ended in production, if Sam returns `enabled` above, `acceptedTos` should also be true.
-        // At that point, we can add an assertion here.
+        // While initial state is first loading, state.acceptedTos will be undefined (it will then be `true` on the
```
cahrens, line 172: Updating the comment because we can't add the assertion-- this code is executed a couple of times while initial state is getting sorted out, and the first time `acceptedTos` is undefined.

marctalbott (approved):
Compared it to https://github.com/DataBiosphere/terra-ui/pull/2805/files as well and everything looks good to me

petesantos (approved):

PR #3029, opened 2022-05-12
[SU-45] Move data table column settings button to header

nawatts:
Currently, the button for changing data table column settings (which columns are displayed and in what order) floats in the top right of the data table.

Screen Shot 2022-05-12 at 10 22 34 AM

Besides being inconsistent with other table controls, which appear in a row above the table, this also blocks access to the last column's controls when the table is wider than the window.

This moves the button to the heading above the table.

Dat table:
Screen Shot 2022-05-12 at 10 23 25 AM

Redesigned data table (enable with `configOverridesStore.set({ isDataTabRedesignEnabled: true })`):
Screen Shot 2022-05-12 at 10 23 16 AM

Selecting data for a workflow:
Screen Shot 2022-05-12 at 10 24 22 AM

While column state is managed by the DataTable component, the header is controlled by its parent component (EntitiesContent or DataStepContent) through DataTable's `childrenBefore` prop. DataTable exposes some of its state by passing arguments to `childrenBefore`.

### DIFF
--- a/src/components/data/EntitiesContent.js
+++ b/src/components/data/EntitiesContent.js
@@ -313,6 +313,7 @@ const EntitiesContent = ({
   const renderCopyButton = (entities, columnSettings) => {
     return h(Fragment, [
       h(ButtonPrimary, {
+        style: { marginRight: '1rem' },
         tooltip: `Copy only the ${entityKey}s visible on the current page to the clipboard in .tsv format`,
         onClick: _.flow(
           withErrorReporting('Error copying to clipboard'),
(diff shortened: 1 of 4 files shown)


src/components/table.js
```diff
@@ -785,7 +785,7 @@ export const ColumnSettings = ({ columnSettings, onChange }) => {
  * @param {Object} style - style override for the button
  * @param {function(Object[])} onSave - called with modified settings when user saves
  */
-export const ColumnSelector = ({ onSave, columnSettings, columnSettingsComponent = ColumnSettings, modalWidth = 600, style, ...otherProps }) => {
+export const ColumnSelector = ({ onSave, columnSettings, modalWidth = 600, style }) => {
```
nawatts, line 788: Since DataTable no longer uses this component, it doesn't need to support different column settings components.

MatthewBemis (approved):

src/components/data/EntitiesContent.js
```diff
@@ -468,10 +469,14 @@ const EntitiesContent = ({
           selected: selectedEntities,
           setSelected: setSelectedEntities
         },
-        childrenBefore: ({ entities, columnSettings }) => div({ style: { display: 'flex', alignItems: 'center', flex: 'none' } },
+        childrenBefore: ({ entities, columnSettings, showColumnSettingsModal }) => div({ style: { display: 'flex', alignItems: 'center', flex: 'none' } },
```
cahrens, line 472: Does `showColumnSettingsModal` reference the const in `DataTable`? I'm having trouble figuring out how it works.

nawatts: Yes. `EntitiesContent` passes `DataTable` a `childrenBefore` render prop that renders the menu above the table. `DataTable` passes `showColumnSettings` and a few other things to that render prop so that the menu can use them.

cahrens: Thanks!

cahrens (approved):

PR #3030, opened 2022-05-12
[SU-50] Load initial data table column settings from saved column settings

nawatts:
Currently, when you view a data table, the initial column settings (which columns are visible and in what order) are loaded from local storage so that changes persist across sessions. If you're viewing the table for the first time, all columns are shown in alphabetical order.

There's a use case that's important to some groups where one person (Editor) uploads data and prepares it before handing it off to another person (Viewer), who looks at the table only after Editor is done with it. Viewer wants to see a specific set of columns in a specific order.

Currently, this can be done by setting a `workspace-column-defaults` workspace attribute with a JSON encoded object:
```
{
  "tableName": {
    "shown": ["columnA", "columnB"],
    "hidden": ["columnC"]
  },
  ...
}
```

However, it's not convenient to set/edit that attribute and workspaces with many tables or tables with many columns may exceed the maximum size of a string attribute.

We have the ability to save column settings in the UI (the gear button at the top right of the data table). So Editor could save column settings which Viewer could then load.

To streamline this and make it so that Viewer does not have to load column settings, this looks for saved column settings named "Default" and uses that for initial column settings if it exists.

To replicate viewing a data table for the first time, clear column settings from local storage using:
```
userId = localStorage.getItem('apc_user_id')
localStorage.removeItem(`dynamic-storage/${userId}/${workspaceNamespace}/${workspaceName}/${entityType`)
```

### DIFF
--- a/src/pages/workspaces/workspace/workflows/DataStepContent.js
+++ b/src/pages/workspaces/workspace/workflows/DataStepContent.js
@@ -20,7 +20,7 @@ const DataStepContent = ({
   entityMetadata, rootEntityType,
   workspace
 }) => {
-  const { workspace: { namespace, name, googleProject, attributes: { 'workspace-column-defaults': columnDefaults } } } = workspace
+  const { workspace: { namespace, name, googleProject } } = workspace
   // State
   const [type, setType] = useState(entitySelectionModel.type)
   const [selectedEntities, setSelectedEntities] = useState(entitySelectionModel.selectedEntities)
@@ -124,7 +124,7 @@ const DataStepContent = ({
             [chooseRootType, () => rootEntityType],
             [chooseSetType, () => entitySetType]
           ),
-          entityMetadata, setEntityMetadata: () => {}, googleProject, workspaceId: { namespace, name }, workspace, columnDefaults,
+          entityMetadata, setEntityMetadata: () => {}, googleProject, workspaceId: { namespace, name }, workspace,
           selectionModel: {
             selected: selectedEntities, setSelected: setSelectedEntities
           }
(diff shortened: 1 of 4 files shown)


cahrens:
@nawatts integration test failures look suspicious.

nawatts:
Just flaky tests I think. They passed after re-running.

src/pages/workspaces/workspace/workflows/DataStepContent.js
```diff
@@ -119,7 +119,7 @@ const DataStepContent = ({
             [chooseRootType, () => rootEntityType],
             [chooseSetType, () => entitySetType]
           ),
-          entityMetadata, setEntityMetadata: () => {}, googleProject, workspaceId: { namespace, name }, workspace, columnDefaults,
+          entityMetadata, setEntityMetadata: () => {}, googleProject, workspaceId: { namespace, name }, workspace,
```
cahrens, line 127: Was the passing of `columnDefaults` just a mistake? I don't see that it was removed as a parameter to `DataTable`.

nawatts: It was removed here: https://github.com/DataBiosphere/terra-ui/pull/3030/files#diff-967b453c8b03379166f3e719046b3a65574bddf955e95e14d115bea19185ae3bL87

DataTable destructured its props in multiple places.

cahrens:
I don't know very much about this functionality at all, so despite your excellent PR description I don't feel very competent testing this. Would it be worth tagging someone on the team that owns this functionality so that they are aware of the change in behavior and can provide another set of eyes?

nawatts:
I'll ask.

My test plan for this was:
- Clear table settings from local storage.
- View/reload data table. Columns should be sorted in alphabetical order.
- Edit column settings and save some settings named "Default". Cancel the modal so settings are not applied.
- Clear table settings from local storage.
- View/reload data table. Columns should be sorted in the order specified by the "Default" settings.

MatthewBemis (approved):

nawatts:
Rebased on request (no changes).

nawatts:
> Would it be worth tagging someone on the team that owns this functionality so that they are aware of the change in behavior and can provide another set of eyes?

I asked a few people if they wanted to try this out, but doesn't seem like there has been any takers.

cahrens (approved):

PR #3031, opened 2022-05-13
[WOR-324] Wait for page to fully load before clicking.

cahrens:
The recent auth PR that merge seems to have changed some timing related to this page loading.

With this change, the test passed 150/150 times with test-flakes.

### DIFF
--- a/integration-tests/tests/billing-projects.js
+++ b/integration-tests/tests/billing-projects.js
@@ -10,6 +10,7 @@ const billingProjectsPage = (testPage, testUrl) => {
       // Note: not using noSpinnersAfter because this action changes the page, and
       // noSpinners after checks that a spinner appears and disappear within the same page.
       await testPage.goto(`${testUrl}/#billing`)
+      await findText(testPage, 'Select a Billing Project')
       await waitForNoSpinners(testPage)
     },
 


integration-tests/tests/billing-projects.js
```diff
@@ -10,6 +10,7 @@ const billingProjectsPage = (testPage, testUrl) => {
       // Note: not using noSpinnersAfter because this action changes the page, and
       // noSpinners after checks that a spinner appears and disappear within the same page.
       await testPage.goto(`${testUrl}/#billing`)
+      await findText(testPage, 'Select a Billing Project')
       await waitForNoSpinners(testPage)
```
cahrens, line 14: `waitForNoSpinners` was added in the auth PR, but sporadic failures were showing that the billing project had the "hover" effect but was not selected after the click operation. That typically means that a spinner is preventing the click from succeeding. Most likely the `waitForNoSpinners` call completed before the spinner actually was displayed (and as the comment in the code explains, `noSpinnersAfter` doesn't work in this case).

By waiting for the page to render `Select a Billing Project`, we can be assured that we have started loading the page content, and then the `waitForNoSpinners` should work.

rtitle (approved):
Tricky.. thanks for fixing.

nawatts (approved):

PR #3033, opened 2022-05-13
[SU-100] Restyle table panels in data tab

nawatts:
Update data tab panels based on new designs from UX. This changes tables to span the full width of the panel and gives the header at the top a darker gray background. 

To enable the data tab redesign, use `configOverridesStore.set({ isDataTabRedesignEnabled: true })`.

Without the feature flag, the only effect is on the separator between the sidebar and main content of the data tab. It replaces the box shadow with a solid line and unfortunately makes the hover target for resizing the sidebar smaller.

## Before

Data table
Screen Shot 2022-05-13 at 2 42 16 PM

Reference data
Screen Shot 2022-05-13 at 2 42 19 PM

Workspace data
Screen Shot 2022-05-13 at 2 42 21 PM

## After

Data table
Screen Shot 2022-05-13 at 2 42 35 PM

Reference data
Screen Shot 2022-05-13 at 2 42 37 PM

Workspace data
Screen Shot 2022-05-13 at 2 42 40 PM

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -58,7 +58,6 @@ const styles = {
   },
   sidebarSeparator: {
     width: '2px',
-    background: colors.light(),
     height: '100%',
     cursor: 'ew-resize'
   },
@@ -130,7 +129,14 @@ const ReferenceDataContent = ({ workspace: { workspace: { googleProject, attribu
   const { initialY } = firstRender ? StateHistory.get() : {}
 
   return h(Fragment, [
-    div({ style: { display: 'flex', justifyContent: 'flex-end', padding: '1rem' } }, [
+    div({
+      style: {
+        display: 'flex', justifyContent: 'flex-end',
+        padding: '1rem',
+        background: isDataTabRedesignEnabled() ? colors.light() : undefined,
+        borderBottom: isDataTabRedesignEnabled() ? `1px solid ${colors.grey(0.4)}` : undefined
+      }
+    }, [
       h(DelayedSearchInput, {
         'aria-label': 'Search',
         style: { width: 300 },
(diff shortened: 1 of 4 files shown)


MatthewBemis (approved):

src/pages/workspaces/workspace/Data.js
```diff
@@ -50,23 +50,21 @@ const styles = {
...
-    boxShadow: '0 2px 5px 0 rgba(0,0,0,0.25)',
     transition: 'width 100ms'
   },
   dataTypeSelectionPanel: {
     flex: 'none',
     backgroundColor: 'white'
   },
   sidebarSeparator: {
-    width: '0.75rem',
+    width: '2px',
```
kyuksel, line 60: What was the reason for using `px` over `rem`?

kyuksel (approved):
Code changes LGTM.

> Without the feature flag, the only effect is on the separator between the sidebar and main content of the data tab. It replaces the box shadow with a solid line and unfortunately makes the hover target for resizing the sidebar smaller.

I personally would prefer to have the hover target wider but I assume the newer version is what UX preferred?

src/pages/workspaces/workspace/Data.js
nawatts: It's the width of vertical line, so it didn't make sense to scale it with font size. Previously, 0.75rem was the width of the hover target and used `rem` to match padding on the other sides of the panel's content (the actual line width was set in pixels with `background: linear-gradient`).

nawatts:
>> Without the feature flag, the only effect is on the separator between the sidebar and main content of the data tab. It replaces the box shadow with a solid line and unfortunately makes the hover target for resizing the sidebar smaller.

> I personally would prefer to have the hover target wider but I assume the newer version is what UX preferred?

The reduced hover target width is a consequence of now having the table fill the entire panel width. Previously, there was padding around the table that could be used for the hover target. We could probably get a larger hover target by absolutely positioning an element to overlap the panel content. However, depending on the content, that might block interaction with some elements on the left side of the content.

PR #3038, opened 2022-05-16
[IA-3385] Update Galaxy and Jupyter logos in Analyses tab

n8landolt:
## Description of ticket:

The Galaxy and Jupyter logos have been showing up blurry for the users. Let's make sure the logos look clearer.

## Description of solution:

Resize the height of logos to 30px. Adjusted image size restrictions. Added Galaxy SVG image. Added galaxy square image.
Note: I decided against using the Jupyter notebook SVG image because the ratio of logo to text size was unfavorable - it ended up looking worse for our use case.

## Testing:
Viewed each affected component in the app, and compared to the live application to make sure the logos in fact did look better. Screenshots of my testing are below

Old formatting on left, new on right:

Screen Shot 2022-05-16 at 10 56 02 AM
Screen Shot 2022-05-16 at 10 55 51 AM
Screen Shot 2022-05-16 at 10 55 31 AM
Screen Shot 2022-05-16 at 2 13 40 PM


src/components/AnalysisModal.js
```diff
@@ -143,7 +143,7 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
         backgroundColor: 'white', borderRadius: 5, padding: '1rem', display: 'inline-block', verticalAlign: 'middle', marginBottom: '1rem',
         textAlign: 'center', width: '100%', height: 60
       },
-      image: { verticalAlign: 'middle', height: '100%', width: '40%' },
+      image: { verticalAlign: 'middle', height: '30px' },
```
kyuksel: Curious why you preferred an absolute unit (`px`) over relative (`%`)?

kyuksel:
Welcome to Terra UI @n8landolt and thank you for the great PR description!

Since this is your first Terra UI PR, I wanted to mention that we optimize 

1. `svg`s by running `yarn run optimize-image-svgs` from the repo root
2. `png`s via a site like [tinypng.com](https://tinypng.com/)

src/components/AnalysisModal.js
n8landolt: I'd love your input on this. I played around a little bit with the styles and some of the bounds were unnecessary. Height 100% didn't do anything when width was 40% (if I remember correctly)
I forgot to take notes.

I chose to do absolute units because in the other locations the images were in absolute units.

kyuksel: I had to do some reading. What I learnt are:

> Generally speaking, you usually don’t want to control the exact width of an image. Every visitor who comes to your website has a potentially different size screen. If you specify the width, it may be much too small for some users and much too big for others. Most of the time, the best option is to make sure that your image is inside of a responsive (percent-based) container and then let it fill the container.

https://html.com/attributes/img-width

It's good practice to set both `height` and `width` because

> If height and width are set, the space required for an image is reserved when the page is loaded. However, without these attributes, the browser does not know the size of the image, and cannot reserve the appropriate space to it. The effect will be that the page layout will change during loading (while the images load).

https://www.w3schools.com/tags/att_img_height.asp

Those being said, when I applied them to the case above and played around, I didn't see great results so I don't have a better suggestion. 

Nit: could write it as

`height: 30` instead of `height: '30px'`

n8landolt: Thank you for sharing! Of course there's partially conflicting views on images... as if they aren't difficult enough to deal with already :). Both views do make sense, though.

I changed from '30px' to 30 in both places I made an edit.

cahrens (approved):

kyuksel:
@n8landolt Thanks for addressing the PR comments. FYI that each Terra UI PR is expected to have a team reviewer (in your case from the interactive analysis team) in addition to a UI code owner approver. The team reviewers are especially expected to do testing on a local UI they deploy or on the PR site that's auto-deployed and linked to from PR pages, which looks like 

image

Therefore I'll let an IA person give the second thumb 😊

cstrandw (approved):
New images and sizing are definitely less blurry and easier to read, looks good to me!

Qi77Qi (approved):

gpcarr (approved):

PR #3039, opened 2022-05-16
[TOAZ-129] A few more fixes prior to B2C enablement in dev

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-129

1. Fix billing scope elevation for the B2C case. No change in behavior for the Google case.

2. For now Azure logins through B2C don't have a profile picture, so don't render a broken link. There's a ticket to address this better.

3. Make Terra UI source the `oidcClientId` from the backend (Orch), and move it out of UI config. We were already using the backend to redirect to the appropriate authority endpoint (see context [here](https://docs.google.com/document/d/1Z2AlIsg2InYJhSC4jssIXTWnh1mmpUw6qMKxx6vPnH8/edit)). This put us in an awkward situation where it would have required a coupled UI+backend release to update the client+authority. Now, we can control B2C vs Google in backend conf.

4. Fix the `/redirect-from-oauth` to not render the full app in the pop-up and not make any backend calls.

### DIFF
--- a/src/appLoader.js
+++ b/src/appLoader.js
@@ -4,7 +4,7 @@ import _ from 'lodash/fp'
 import ReactDOM from 'react-dom'
 import { h } from 'react-hyperscript-helpers'
 import RModal from 'react-modal'
-import { initializeAuth } from 'src/libs/auth'
+import { initializeAuth, initializeClientId } from 'src/libs/auth'
 import { initializeTCell } from 'src/libs/tcell'
 import Main from 'src/pages/Main'
 
@@ -17,6 +17,8 @@ window.SATURN_VERSION = process.env.REACT_APP_VERSION
 
 window._ = _
 
-ReactDOM.render(h(Main), appRoot)
-initializeAuth()
-initializeTCell()
+initializeClientId().then(() => {
+  ReactDOM.render(h(Main), appRoot)
+  initializeAuth()
+  initializeTCell()
+})
--- a/src/components/icons.js
+++ b/src/components/icons.js
@@ -60,14 +60,20 @@ export const centeredSpinner = ({ size = 48, ...props } = {}) => spinner(_.merge
   }
 }, props))
 
-export const profilePic = ({ size, style, ...props } = {}) => img({
-  alt: 'Google profile image',
-  src: getUser().imageUrl,
-  height: size, width: size,
-  style: { borderRadius: '100%', ...style },
-  referrerPolicy: 'no-referrer',
-  ...props
-})
+export const profilePic = ({ size, style, ...props } = {}) => {
+  // Note Azure logins don't currently have an imageUrl.
+  // For now, don't render anything. In the future, investigate auto-generating a
+  // picture in the UI or adding this capability to B2C.
+  const imageUrl = getUser().imageUrl
+  return imageUrl && img({
+    alt: 'Google profile image',
+    src: imageUrl,
+    height: size, width: size,
+    style: { borderRadius: '100%', ...style },
+    referrerPolicy: 'no-referrer',
+    ...props
+  })
+}
 
 export const wdlIcon = ({ style = {}, ...props } = {}) => div({
   style: {
--- a/src/components/icons.js
+++ b/src/components/icons.js
@@ -61,9 +61,8 @@ export const centeredSpinner = ({ size = 48, ...props } = {}) => spinner(_.merge
 }, props))
 
 export const profilePic = ({ size, style, ...props } = {}) => {
-  // Note Azure logins don't currently have an imageUrl.
-  // For now, don't render anything. In the future, investigate auto-generating a
-  // picture in the UI or adding this capability to B2C.
+  // Note Azure logins don't currently have an imageUrl, so don't render anything.
+  // See TOAZ-147 to support this in the future.
   const imageUrl = getUser().imageUrl
   return imageUrl && img({
     alt: 'Google profile image',
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -23,7 +23,7 @@ import * as Utils from 'src/libs/utils'
 export const getOidcConfig = () => {
   return {
     authority: `${getConfig().orchestrationUrlRoot}/oauth2/authorize`,
-    client_id: getConfig().oidcClientId,
+    client_id: authStore.get().oidcConfig.clientId,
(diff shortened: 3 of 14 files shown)


src/appLoader.js
```diff
@@ -17,6 +17,8 @@ window.SATURN_VERSION = process.env.REACT_APP_VERSION
 
 window._ = _
 
-ReactDOM.render(h(Main), appRoot)
-initializeAuth()
-initializeTCell()
+initializeClientId().then(() => {
```
rtitle, line 20: The client id needs to be known before the react components are rendered. Let me know if there are concerns / other ideas here.

src/components/icons.js
```diff
@@ -60,14 +60,19 @@ export const centeredSpinner = ({ size = 48, ...props } = {}) => spinner(_.merge
...
-  alt: 'Google profile image',
-  src: getUser().imageUrl,
-  height: size, width: size,
-  style: { borderRadius: '100%', ...style },
-  referrerPolicy: 'no-referrer',
-  ...props
-})
+export const profilePic = ({ size, style, ...props } = {}) => {
+  // Note Azure logins don't currently have an imageUrl, so don't render anything.
+  // See TOAZ-147 to support this in the future.
```
rtitle, line 65: This now looks like this for the B2C-Azure case (note no picture):

image

Google logins still have a picture. Potential future options for Azure:
1. Autogenerate a pic if one isn't present. TDR UI [auto-generates](https://github.com/DataBiosphere/jade-data-repo-ui/blob/develop/src/components/common/TerraAvatar.tsx) one using [material-ui/react-avatar](https://mui.com/material-ui/react-avatar/) (cc @nmalfroy). I was hesitant to add this dependency to Terra UI however.

2. Render a static image like 
   image

4. Or do something on the B2C side.

src/libs/auth.js
```diff
@@ -73,7 +77,11 @@ export const reloadAuthToken = (includeBillingScope = false) => {
 
 export const hasBillingScope = () => {
   const scope = getUser().scope
-  return scope && scope.includes('https://www.googleapis.com/auth/cloud-billing')
+
+  // For now B2C always requests the cloud-billing scope from the get-go.
+  // TOAZ-146 is open to create a separate B2C policy with the sensitive scope
```
rtitle, line 96: cc @nmalfroy I made a ticket to setup a separate policy for Terra UI with the sensitive scope.

breilly2:
Still reviewing, but the first thing that I noticed is a behavior that looks like it predates this PR that might have been introduced with the B2C support. In the window that opens to sign in to Google, after signing in (or selecting a signed-in account), I see Terra UI start to load in that window before it closes! I'm not 100% sure that this _didn't_ happen before, but I don't remember it. The consequence is unnecessary network traffic and, depending on timing, even recording usage events in Mixpanel.

I think we should find a way to avoid this, especially if we can confirm that it didn't happen pre-B2C.

rtitle:
@breilly2 yeah Christina and I noticed that earlier. The signin pop-up redirects to `${window.origin}/#redirect-from-oauth`, which renders a spinner and calls the callback to close the popup (see it in action [here](https://bvdp-saturn-dev.appspot.com/#redirect-from-oauth)). However `Loading Terra..` flickers before the spinner is rendered, I think from [here](https://github.com/DataBiosphere/terra-ui/blob/dev/public/index.html#L22).

I'm not certain how to avoid this -- the popup redirect needs to render to complete the auth flow. Is it possible to have a Terra UI route which doesn't display `Loading Terra..`?

nmalfroy:
FWIW, in TDR I made a route just for that redirect that just shows a spinner (and doesn't trigger fetches)

rtitle:
Yeah, now thinking I could have some logic in `index.js` to avoid fetches for that route. Will try it out..

src/pages/RedirectFromOAuth.js
```diff
@@ -1,16 +1,17 @@
...
+const RedirectFromOAuth = (silent = false) => {
+  const userManager = new UserManager({
+    popup_redirect_uri: `${window.origin}/redirect-from-oauth`,
+    silent_redirect_uri: `${window.origin}/redirect-from-oauth-silent`
+  })
   const url = window.location.href.replace('#', '')
   useOnMount(() => silent === true ? userManager.signinSilentCallback(url) : userManager.signinPopupCallback(url))
   return div({ role: 'main' }, [
-    h(centeredSpinner, { style: { position: 'fixed' } })
+    'Signing in to Terra...'
```
rtitle, line 14: The `centeredSpinner` icon requires the `config.json` to be loaded (to get the colors for the branded site…). By just having text I can avoid loading `config.json` in the auth pop-up. I figure this is fine because it's visible for a fraction of a second..

rtitle:
@breilly2 I tried a [thing](https://github.com/DataBiosphere/terra-ui/pull/3039/commits/471efd82033e6336f71385b524a4951088bac4d3). Now the `/redirect-from-oauth` route doesn't render the full app or make any backend calls -- it just renders the single component it needs to complete the oauth callback. I still see `Loading Terra…` flicker (while some assets are loaded) but it's much less than before. Let me know what you think :)

src/components/icons.js
```diff
@@ -60,14 +60,19 @@ export const centeredSpinner = ({ size = 48, ...props } = {}) => spinner(_.merge
...
+  const imageUrl = getUser().imageUrl
+  return imageUrl && img({
```
breilly2, line 67:
```suggestion
  return !!imageUrl && img({
```
Since this is returning something that is going to be treated as a component to render, the check should be coerced to a boolean. If `imageUrl` is `undefined`, this will do what you expect. However, if it is `null` or 0, it will still be falsy, but I think React will end up rendering the string "null" or "0". Coercing it to a boolean will make it decisively `false` instead of relying on falsy.

breilly2 (approved):
👍  Thank you for separating the OAuth redirect so it doesn't start loading the entire app in the sign-in popup window.

I do feel like this could be done inside of `Main` instead of adding more logic to the app bootstrapping. The only problem with doing it this way is that a reader might not expect to find this logic where it is. I can't quickly and easily visualize what a different implementation would look like, so I'm going ahead and approving this as it is. We can always refactor later.

rtitle:
I'm actually going to close this because the thing it depends on was reverted. Will re-open when ready.

PR #3040, opened 2022-05-16
[SU-94] Support searching all data tables

MatthewBemis:
Run `configOverridesStore.set({ isSearchAwesomeNow: true })` in the console to enable this. Note that this will also enable the new data tab design.

After enabling, you should be able to enter in a search term, hit enter and then see something along the lines of the following:

Screen Shot 2022-05-20 at 12 11 45 PM

Clicking into a table with an active cross-table filter term will show the results filtered with that term. Clearing the term will restore the table to showing all results.

Users with column views set when performing the cross-table search will have the column view removed until they clear the filter term.

When there are no tables, the search bar will not appear.

general-dev-billing-account/anonymized-test is a good workspace to stress-test this. Let me know if you need access. This was also tested against the giant CDPH workspace in production and it performed well.

<!--
Hello, friend!
Thanks!

--!>


src/pages/workspaces/workspace/Data.js
```diff
@@ -574,7 +588,10 @@ const WorkspaceData = _.flow(
   const [uploadingFile, setUploadingFile] = useState(false)
   const [entityMetadataError, setEntityMetadataError] = useState()
   const [snapshotMetadataError, setSnapshotMetadataError] = useState()
-  const [sidebarWidth, setSidebarWidth] = useState(280)
+  const [sidebarWidth, setSidebarWidth] = useState(340)
```
MatthewBemis: I need to check with Ean to make sure this is okay, but the addition of the green "pills" when in search-mode made it a bit cramped with the previous default width

src/pages/workspaces/workspace/Data.js
```diff
@@ -666,6 +683,16 @@ const WorkspaceData = _.flow(
     setEntityMetadata(updatedMetadata)
   }
 
+  const searchAcrossTables = async (types, activeCrossTableTextFilter) => {
```
nawatts: This function could probably use some error handling.

src/pages/workspaces/workspace/Data.js
```diff
@@ -69,11 +69,23 @@ const styles = {
   }
 }
 
-const DataTypeButton = ({ selected, entityName, children, entityCount, iconName = 'listAlt', iconSize = 14, buttonStyle, after, ...props }) => {
+const SearchResultsPill = ({ count, searching }) => {
```
nawatts: Should the pill have its own tooltip? Currently, hovering over it shows the tooltip for the table, with a row count that may not match the results count.

Screen Shot 2022-05-17 at 2 17 18 PM

src/pages/workspaces/workspace/Data.js
```diff
@@ -666,6 +683,16 @@ const WorkspaceData = _.flow(
     setEntityMetadata(updatedMetadata)
   }
 
+  const searchAcrossTables = async (types, activeCrossTableTextFilter) => {
+    setCrossTableSearchInProgress(true)
+    const results = await Promise.all(_.map(async ([type]) => {
```
nawatts: I was confused about why the async function's argument was a destructured array. Should `searchAcrossTables` take a list of type names instead of the current list of `[typeName, typeInfo]`?

nawatts (commented):
Nice! I thought this was going to be require way more code.

nawatts:
I'm seeing a pill in rows in other sections of the sidebar.

Screen Shot 2022-05-17 at 2 22 03 PM

src/pages/workspaces/workspace/Data.js
MatthewBemis: That would make more sense!

MatthewBemis:
@nawatts oh yuck, fixed the "bonus" pills in 795f6e9fc4883443d0c63ea18ab1ea55be979985

src/pages/workspaces/workspace/Data.js
MatthewBemis: Added some error reporting/handling in ff3e71d9b1975670734b359d3049afe7d034ce30

src/pages/workspaces/workspace/Data.js
MatthewBemis: Hmm, yeah maybe if there is an active cross-table search we can display something like "thing (10 total rows, 1 filtered row)" What do you think?

src/pages/workspaces/workspace/Data.js
MatthewBemis: Addressed in d068abed0668ed29db995c885beff1f431cce470

src/pages/workspaces/workspace/Data.js
```diff
@@ -69,11 +69,21 @@ const styles = {
...
+  },
+  searching ? [icon('loadingSpinner', { size: 13, color: 'white' })] : `${filteredCount}`)
+}
+
+const DataTypeButton = ({ selected, entityName, children, entityCount, iconName = 'listAlt', iconSize = 14, buttonStyle, filteredCount, crossTableSearchInProgress, activeCrossTableTextFilter, after, ...props }) => {
   const isEntity = entityName !== undefined
 
   return h(Interactive, {
-    style: { ...Style.navList.itemContainer(selected) },
+    style: { ...Style.navList.itemContainer(selected), backgroundColor: selected ? colors.dark(0.1) : 'white' },
```
MatthewBemis, line 86: This style change isn't directly related to this PR but it was requested by UX

src/pages/workspaces/workspace/Data.js
```diff
@@ -740,12 +769,27 @@ const WorkspaceData = _.flow(
...
+                  'aria-label': 'Search all tables',
+                  placeholder: 'Search all tables',
+                  onChange: activeCrossTableTextFilter => {
+                    setActiveCrossTableTextFilter(activeCrossTableTextFilter)
+                    searchAcrossTables(_.keys(entityMetadata), activeCrossTableTextFilter)
+                  },
+                  defaultValue: activeCrossTableTextFilter
+                })
+              ]),
+              activeCrossTableTextFilter !== '' && div({ style: { margin: '0rem 1rem 1rem 1rem' } }, crossTableSearchInProgress ? ['Loading...', [icon('loadingSpinner', { size: 13, color: colors.primary() })]] : [`${_.sum(_.map(c => c.filteredCount, crossTableResultCounts))} results`]),
```
nawatts: The icon here is wrapped in an extra array. This results in a warning from React: `Warning: Each child in a list should have a unique "key" prop.`

```suggestion
              activeCrossTableTextFilter !== '' && div({ style: { margin: '0rem 1rem 1rem 1rem' } }, crossTableSearchInProgress ? ['Loading...', icon('loadingSpinner', { size: 13, color: colors.primary() })] : [`${_.sum(_.map(c => c.filteredCount, crossTableResultCounts))} results`]),
```

src/components/data/DataTable.js
```diff
@@ -237,7 +238,8 @@ const DataTable = props => {
 
 
   // Render
-  const columnSettings = applyColumnSettings(columnState || [], entityMetadata[entityType].attributeNames)
+  // If there is an active cross table search, temporarily reset the column settings to show all columns
+  const columnSettings = activeCrossTableTextFilter ? applyColumnSettings([], entityMetadata[entityType].attributeNames) : applyColumnSettings(columnState || [], entityMetadata[entityType].attributeNames)
```
nawatts, line 242: Since this overrides a user's table column settings, should we disable the settings button while a cross table search is active?

nawatts (approved):

src/pages/workspaces/workspace/Data.js
MatthewBemis: Added in 6473444e567975c124f4fe01998adf9a6b716bae

src/pages/workspaces/workspace/Data.js
MatthewBemis: I reverted this in 6473444e567975c124f4fe01998adf9a6b716bae

We decided to go with truncating long pill numbers, so showing (100141) as (100k), etc

MatthewBemis:
I brought in a new dependency in 6473444e567975c124f4fe01998adf9a6b716bae , but I may back that out because we're only utilizing it for one very small thing

src/libs/utils.js
```diff
@@ -377,3 +377,12 @@ export const formatBytes = bytes => {
   ].find(([_p, d]) => bytes >= d)
   return `${(bytes / divisor).toPrecision(3)} ${prefix}iB`
 }
+
+//Truncates an integer to the thousands, i.e. 10363 -> 10k
+export const truncateInteger = integer => {
+  if (integer < 1000) {
```
jordansafer: Could we increase this to 10,000? I think for smaller numbers, for example if we have 1999, than displaying 1k would be a big shift. Do 4 digits fit?

src/pages/workspaces/workspace/Data.js
```diff
@@ -740,12 +769,27 @@ const WorkspaceData = _.flow(
...
+                  'aria-label': 'Search all tables',
+                  placeholder: 'Search all tables',
+                  onChange: activeCrossTableTextFilter => {
+                    setActiveCrossTableTextFilter(activeCrossTableTextFilter)
+                    searchAcrossTables(_.keys(entityMetadata), activeCrossTableTextFilter)
+                  },
+                  defaultValue: activeCrossTableTextFilter
+                })
+              ]),
+              activeCrossTableTextFilter !== '' && div({ style: { margin: '0rem 1rem 1rem 1rem' } }, crossTableSearchInProgress ? ['Loading...', icon('loadingSpinner', { size: 13, color: colors.primary() })] : [`${_.sum(_.map(c => c.filteredCount, crossTableResultCounts))} results`]),
```
jordansafer: nit that this line is quite long, maybe we could break out the div?

jordansafer (approved):

src/components/data/DataTable.js
MatthewBemis: Addressed here https://github.com/DataBiosphere/terra-ui/pull/3040/commits/33e50019c0700abc1464e3c77ac613bec4323b68

src/libs/utils.js
MatthewBemis: 4 digits do fit! Addressed in c8e7208f19a8c8f05b7a020612c63304ffe32c5f

src/pages/workspaces/workspace/Data.js
MatthewBemis: Addressed in 9addfaafd782e94610393e4c04e3e9b1f87dfe84

PR #3041, opened 2022-05-16
[QA-1871] Terra-UI google-workspace integration test error: Node is either not clickable or not an HTMLElement

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1871

`google-workspace` test failed in this CircleCi [job](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9569/workflows/4a31f6b5-672e-441e-ab54-f666a4c3a0d4/jobs/42706).

Fixes:

- Set option `visible: true` as default for `waitForXPath` in `click` function. Plainly, an element should be visible before click.
- `clickNavChildAndLoad` function is flaky. Update the function to wait for navigation after click tab.

### DIFF
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -198,7 +198,11 @@ const enableDataCatalog = async (page, testUrl, token) => {
 }
 
 const clickNavChildAndLoad = async (page, tab) => {
-  await noSpinnersAfter(page, { action: () => click(page, navChild(tab)) })
+  // click triggers a page navigation event
+  await Promise.all([
+    page.waitForNavigation(waitUntilLoadedOrTimeout()),
+    noSpinnersAfter(page, { action: () => click(page, navChild(tab)) })
+  ])
 }
 
 const viewWorkspaceDashboard = async (page, token, workspaceName) => {
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -103,7 +103,7 @@ const clickTableCell = async (page, { tableName, columnHeader, textContains, isD
   return (await page.waitForXPath(xpath, options)).click()
 }
 
-const click = async (page, xpath, options) => {
+const click = async (page, xpath, options = { visible: true }) => {
   return (await page.waitForXPath(xpath, options)).click()
 }
 


integration-tests/utils/integration-helpers.js
```diff
@@ -198,7 +198,11 @@ const enableDataCatalog = async (page, testUrl, token) => {
 }
 
 const clickNavChildAndLoad = async (page, tab) => {
-  await noSpinnersAfter(page, { action: () => click(page, navChild(tab)) })
+  // click triggers a page navigation event
+  await Promise.all([
+    page.waitForNavigation(waitUntilLoadedOrTimeout()),
+    noSpinnersAfter(page, { action: () => click(page, navChild(tab)) })
+  ])
```
aweng98, line 205: I don't find function handles navigation but the function name clearly indicates a navigation event.

integration-tests/utils/integration-utils.js
```diff
@@ -103,7 +103,7 @@ const clickTableCell = async (page, { tableName, columnHeader, textContains, isD
   return (await page.waitForXPath(xpath, options)).click()
 }
 
-const click = async (page, xpath, options) => {
+const click = async (page, xpath, options = { visible: true }) => {
```
cahrens, line 106: I had no idea that the default value of visible is `false` in `waitForXPath`! That seems like a poor choice on their part.

https://puppeteer.github.io/puppeteer/docs/9.1.1/puppeteer.page.waitforxpath/

That being said, it's probably worth auditing the rest of the calls to `waitForXPath`

integration-tests/utils/integration-helpers.js
cahrens: This is used for internal navigation-- clicking on the tabs within a workspace (Analyses, Workflows, etc.). I'm not sure `waitForNavigation` will do anything useful in this case, but as long as the tests continue to pass I assume there is no harm in calling it.

cahrens (approved):

integration-tests/utils/integration-utils.js
aweng98: Agreed. I'll fix other `waitForXpath` soon.

integration-tests/utils/integration-helpers.js
aweng98: `waitForNavigation` is useful for synchronization when loading triggers new network requests.

integration-tests/utils/integration-utils.js
kyuksel: I'm very surprised we hadn't already been setting `visible` to `true`, which makes me want to check with @petesantos and/or @breilly2 in case there was a particular reason for it.

kyuksel (approved):
LGTM 👍🏼 

One question inline to make sure we're not missing some historical context around `{ visible: true }`.

PR #3043, opened 2022-05-17
[TOAZ-15 follow up] Prevent displaying error messages for sign out requests

rtitle:
Before when the access token expired we would see errors:
image

Now the errors don't popup when this happens. Also make silent refresh a little more robust hopefully.

Testing by logging in waiting:
- If computer is on my access token gets refreshed silently 5m before expiration
- If computer goes to sleep, my access token can expire.
   * It's hard to reproduce though -- e.g. silent refresh still worked for me with a closed laptop
   * In that case I see "session timed out" with no additional errors:
   image
   * Looking into how we can handle going to sleep better (https://github.com/authts/oidc-client-ts/issues/251). Any ideas from reviewers?

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -33,7 +33,8 @@ export const getOidcConfig = () => {
       token_endpoint: `${getConfig().orchestrationUrlRoot}/oauth2/token`
     },
     userStore: new WebStorageStateStore({ store: window.localStorage }),
-    accessTokenExpiringNotificationTimeInSeconds: 120,
+    accessTokenExpiringNotificationTimeInSeconds: 300,
+    automaticSilentRenew: true,
     includeIdTokenInSilentRenew: true
   }
 }
@@ -69,6 +70,18 @@ export const signIn = (includeBillingScope = false) => {
 export const reloadAuthToken = (includeBillingScope = false) => {
   const args = getSigninArgs(includeBillingScope)
   return getAuthInstance().signinSilent(args).catch(() => false)
+}
+
+export const handleSilentRenewError = error => {
+  if (error.name !== 'ErrorResponse') {
+    setTimeout(() => {
+      const auth = getAuthInstance()
+      if (auth.user && !auth.user.expired) {
+        auth.stopSilentRenew()
+        auth.startSilentRenew()
+      }
+    }, 5 * 1000)
+  }
 }
 
 export const hasBillingScope = () => {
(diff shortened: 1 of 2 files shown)


src/libs/auth.js
```diff
@@ -33,7 +33,8 @@ export const getOidcConfig = () => {
       token_endpoint: `${getConfig().orchestrationUrlRoot}/oauth2/token`
     },
     userStore: new WebStorageStateStore({ store: window.localStorage }),
-    accessTokenExpiringNotificationTimeInSeconds: 120,
+    accessTokenExpiringNotificationTimeInSeconds: 300,
+    automaticSilentRenew: true,
```
rtitle, line 37: Using `automaticSilentRenew` instead of a manual callback which is hopefully a bit more robust because it retries errors. Also bumped the time to 5m.

See settings here: https://authts.github.io/oidc-client-ts/interfaces/UserManagerSettings.html

And implementation of silent renew: https://github.com/authts/oidc-client-ts/blob/ff8627d003aef90adbb0a88e8a10ad51c6e60964/src/SilentRenewService.ts

src/libs/auth.js
```diff
@@ -71,6 +72,18 @@ export const reloadAuthToken = (includeBillingScope = false) => {
   return getAuthInstance().signinSilent(args).catch(() => false)
 }
 
+export const handleSilentRenewError = error => {
```
rtitle, line 75: I got this from https://github.com/authts/oidc-client-ts/issues/251, not sure if it helps honestly. Silent renew is working for me, but I'm still observing that it sometimes fails to happen when my computer goes to sleep.

src/libs/auth.js
```diff
@@ -132,14 +145,14 @@ export const bucketBrowserUrl = id => {
...
-export const processUser = user => {
+export const processUser = (user, isSignInEvent) => {
   return authStore.update(state => {
     const isSignedIn = !_.isNil(user)
     const profile = user?.profile
     const userId = profile?.sub
 
     // The following few lines of code are to handle sign-in failures due to privacy tools.
-    if (state.isSignedIn === false && isSignedIn === false) {
+    if (isSignInEvent === true && state.isSignedIn === false && isSignedIn === false) {
```
rtitle, line 155: We only want to display this notification if signIn failed. This callback is raised too for signOut events.

nawatts (approved):

nawatts:
> If computer is on my access token gets refreshed silently 5m before expiration
> * If computer goes to sleep, my access token can expire.
> * It's hard to reproduce though -- e.g. silent refresh still worked for me with a closed laptop
In that case I see "session timed out" with no additional errors:
> * Looking into how we can handle going to sleep better

I don't understand. If the token expires while the machine is asleep, what could we do about that?

rtitle:
> I don't understand. If the token expires while the machine is asleep, what could we do about that?

Yeah I mean nothing really. :) If we had a refresh token we could refresh it on demand but we're not doing that currently for Google or B2C (and it's a potential security issue to have a long-lived refresh token in the UI).

The old gapi library had some magic to reload the access token, but I'm not even certain that worked if it was expired: https://developers.google.com/identity/sign-in/web/reference#googleuserreloadauthresponse

nmalfroy (approved):
that seems reasonable to me!  Someone with more terra UI should probably look though

PR #3045, opened 2022-05-17
[SUP-655] Handle requester pays errors from getting bucket location workspace dashboard

nawatts:
The workspace dashboard makes a request to the GCS API to get the location of the workspace's bucket. For workspaces with a requester pays bucket where the user only has reader access to the workspace, this requires selecting a workspace to bill the request to. This is most likely to occur with public/featured workspaces.

Currently, when viewing the dashboard for such a workspace, a modal pops up prompting for a workspace to bill to. This presents a barrier to browsing information about public/featured workspaces. The user may not care about the bucket location, but clicking cancel in the modal navigates back, preventing them from viewing any information about the workspace. In some cases, a user may not even have a workspace they could select to bill requests to. 

https://user-images.githubusercontent.com/1156625/168826639-d2e5dafe-17b0-42e6-a6d1-e2003d8e5c0c.mov

With this change, if the request for bucket location fails because of a requester pays error, the bucket location will show as "Unknown" and a button will be shown that prompts the user to choose a workspace to bill to and load the bucket location. Dismissing the modal will stay on the workspace dashboard page and leave the bucket location as "Unknown".

https://user-images.githubusercontent.com/1156625/168828064-9ed164de-71a0-4dad-99d9-9329e1a30c9e.mov

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -62,6 +62,14 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
     try {
       return await Utils.withDelay(until, wrappedFetch)(...args)
     } catch (error) {
+      // requesterPaysError may be set on responses from requests to the GCS API that are wrapped in withRequesterPays.
+      // requesterPaysError is true if the request requires a user project for billing the request to. Such errors
+      // are not transient and the request should not be retried.
+      const shouldRetry = !error.requesterPaysError
+
+      if (!shouldRetry) {
+        throw error
+      }
       // ignore error will retry
     }
   }
--- a/src/pages/workspaces/workspace/Dashboard.js
+++ b/src/pages/workspaces/workspace/Dashboard.js
@@ -1,14 +1,15 @@
 import _ from 'lodash/fp'
-import { Fragment, useEffect, useImperativeHandle, useState } from 'react'
+import { Fragment, useCallback, useEffect, useImperativeHandle, useState } from 'react'
 import { dd, div, dl, dt, h, h3, i, span } from 'react-hyperscript-helpers'
 import * as breadcrumbs from 'src/components/breadcrumbs'
-import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
+import { requesterPaysWrapper } from 'src/components/bucket-utils'
 import Collapse from 'src/components/Collapse'
 import { ButtonPrimary, ButtonSecondary, ClipboardButton, Link, spinnerOverlay } from 'src/components/common'
 import { centeredSpinner, icon, spinner } from 'src/components/icons'
 import { MarkdownEditor, MarkdownViewer } from 'src/components/markdown'
 import { InfoBox } from 'src/components/PopupTrigger'
 import { getRegionInfo } from 'src/components/region-common'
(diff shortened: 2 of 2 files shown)


src/libs/ajax.js
```diff
@@ -62,6 +62,14 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
     try {
       return await Utils.withDelay(until, wrappedFetch)(...args)
     } catch (error) {
+      // requesterPaysError may be set on responses from requests to the GCS API that are wrapped in withRequesterPays.
+      // requesterPaysError is true if the request requires a user project for billing the request to. Such errors
+      // are not transient and the request should not be retried.
+      const shouldRetry = !error.requesterPaysError
```
nawatts, line 68: This could probably benefit from some more logic around which requests to retry, based on response status code, etc. Punting on that as out of scope for this issue.

breilly2 (approved):
👍  Love it! Nice improvement!

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -122,9 +123,71 @@ const RightBoxSection = ({ title, info, initialOpenState, onClick, children }) =
   ])
 }
 
+const BucketLocation = requesterPaysWrapper({ onDismiss: _.noop })(({ workspace }) => {
```
cahrens, line 126: None of this is executed for an Azure workspace, so I don't think you need to add the conditional logic around `isGoogleWorkspace` (though I could see adding an assert that `isGoogleWorkspace` is true.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -122,9 +123,71 @@ const RightBoxSection = ({ title, info, initialOpenState, onClick, children }) =
...
+
+  if (loading) {
+    return 'Loading'
+  }
+
+  if (!location) {
+    return h(Fragment, [
+      'Unknown',
+      needsRequesterPaysProject && h(ButtonSecondary, {
+        tooltip: 'This workspace\'s bucket is requester pays. Click to choose a workspace to bill requests to and get the bucket\'s location.',
```
cahrens, line 166: I would suggest adding an `aria-label` with a more concise description of the button. The tooltip text will still be read out at the end, but it makes it easier to understand what the control is. Currently the screen reader announces: **"This workspace's bucket is request pays. Click to choose….." button**.

src/pages/workspaces/workspace/Dashboard.js
nawatts: Good idea. `assert` seems like a more appropriate way of accomplishing what I was going for here.

nawatts: Done in https://github.com/DataBiosphere/terra-ui/pull/3045/commits/54ecc9f0c1d754fc4b2348ccd2f4b42ca4d27dd8

src/pages/workspaces/workspace/Dashboard.js
nawatts: Done in https://github.com/DataBiosphere/terra-ui/pull/3045/commits/3b7e5abb0584badeb601259230455db7856cb433

cahrens (approved):

MatthewBemis (approved):

PR #3047, opened 2022-05-17
[DC-342] Add writer and add to-lower to role definitions

s-rubenstein:
This is in preparation for role changes being done to the catalog. toLower is included in order for us to match both the old definitions and the new ones, and `writer` is added to our list in order to give a more granular approach to role permissioning.

### DIFF
--- a/src/pages/library/dataBrowser-utils.js
+++ b/src/pages/library/dataBrowser-utils.js
@@ -80,7 +80,7 @@ const normalizeDataset = dataset => {
     dataReleasePolicy,
     contacts, curators, contributorNames,
     dataType, dataModality,
-    access: _.intersection(dataset.roles, ['reader', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
+    access: _.intersection(_.map(role => _.toLower(role), dataset.roles), ['reader', 'writer', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
   }
 }
 
--- a/src/pages/library/dataBrowser-utils.js
+++ b/src/pages/library/dataBrowser-utils.js
@@ -80,7 +80,7 @@ const normalizeDataset = dataset => {
     dataReleasePolicy,
     contacts, curators, contributorNames,
     dataType, dataModality,
-    access: _.intersection(_.map(role => _.toLower(role), dataset.roles), ['reader', 'writer', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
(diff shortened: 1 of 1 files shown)


src/pages/library/dataBrowser-utils.js
```diff
@@ -80,7 +80,7 @@ const normalizeDataset = dataset => {
     dataReleasePolicy,
     contacts, curators, contributorNames,
     dataType, dataModality,
-    access: _.intersection(dataset.roles, ['reader', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
+    access: _.intersection(_.map(role => _.toLower(role), dataset.roles), ['reader', 'writer', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
```
nawatts: ESLint is complaining about the unnecessary wrapper function.

```suggestion
    access: _.intersection(_.map(_.toLower, dataset.roles), ['reader', 'writer', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
```

nawatts (approved):

fboulnois (approved):

src/pages/library/dataBrowser-utils.js
```diff
@@ -80,7 +80,7 @@ const normalizeDataset = dataset => {
     dataReleasePolicy,
     contacts, curators, contributorNames,
     dataType, dataModality,
-    access: _.intersection(dataset.roles, ['reader', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
+    access: _.intersection(_.map(_.toLower, dataset.roles), ['reader', 'writer', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
```
pshapiro4broad: I don't think we should use `toLower` here; the UI and the service are both things we control so we should be able to have both ends agree on an API where the client knows what format the server is returning.

Regarding `writer`, I've covered this in the other PR but I don't think we should add it until we have a use case for it in our API/UI.

pshapiro4broad (changes requested):

src/pages/library/dataBrowser-utils.js
s-rubenstein: For `toLower`, I agree, but wanted to have it be backwards compatible. Currently it returns lowercase. Rawls returns uppercase, and TDR returns lowercase. Maybe the solution is to just define our enum lowercase, and not have the toLower here? That would be simpler, but I don't know what the conventions are around enum casing in all of Terra. My understanding was uppercase was the approach, since I have seen uppercase values in both Sam and Rawls, and was surprised we were using lowercase in TDR.

Because the UI doesn't roll out at the same time as the API, I want to make sure it handles both cases well, so it seamlessly transitions when the API rolls out, without need for any coordinated deployments.

PR #3049, opened 2022-05-17
[SU-57] Allowing editing JSON attributes

nawatts:
In addition to the string, number, boolean, and list types that the UI currently supports, data table cells can contain arbitrary JSON values.

For example, an entity with JSON attributes can be created using Rawl's [create entity endpoint](https://rawls.dsde-dev.broadinstitute.org/#/entities/create_entity).
```
{
  "name": "json-attributes",
  "entityType": "test_entity",
  "attributes": {
    "obj": { "key": "value" },
    "objarray": [{ "key1": "value1" }, { "key2": "value2" }, { "key3": "value3" } ],
    "array": ["one", "two", "three"]
  }
}
```

Currently, there is no way to edit such an attribute in the UI.

https://user-images.githubusercontent.com/1156625/168911260-29733ef6-b31e-41b4-9aa3-ac3c68686631.mov

This adds support for editing JSON attributes using the [react-json-view](https://www.npmjs.com/package/react-json-view) library, which is already used in a few places in Terra. It also adds support for converting attributes to/from JSON when selecting one of the type radio buttons.

https://user-images.githubusercontent.com/1156625/168911404-dcbdc7ca-06a9-4bf5-89b4-e468061945b3.mov

### DIFF
--- a/src/components/data/_tests/data-utils.test.js
+++ b/src/components/data/_tests/data-utils.test.js
@@ -82,6 +82,10 @@ describe('getAttributeType', () => {
       ],
       itemsType: 'EntityReference'
     })).toEqual({ type: 'reference', isList: true })
+
+    expect(getAttributeType({ key: 'value' })).toEqual({ type: 'json', isList: false })
+    expect(getAttributeType(['a', 'b', 'c'])).toEqual({ type: 'json', isList: false })
+    expect(getAttributeType([{ idx: 0 }, { idx: 1 }, { idx: 2 }])).toEqual({ type: 'json', isList: false })
   })
 
   it('returns string for null values', () => {
@@ -97,18 +101,28 @@ describe('convertAttributeValue', () => {
     expect(convertAttributeValue('a_string', 'number')).toEqual(0)
     expect(convertAttributeValue('a_string', 'boolean')).toEqual(true)
     expect(convertAttributeValue('a_string', 'reference', 'thing')).toEqual({ entityType: 'thing', entityName: 'a_string' })
+    expect(convertAttributeValue('a_string', 'json')).toEqual({ value: 'a_string' })
 
     expect(convertAttributeValue(7, 'string')).toEqual('7')
     expect(convertAttributeValue(7, 'boolean')).toEqual(true)
     expect(convertAttributeValue(7, 'reference', 'thing')).toEqual({ entityType: 'thing', entityName: '7' })
+    expect(convertAttributeValue(7, 'json')).toEqual({ value: 7 })
 
     expect(convertAttributeValue(true, 'string')).toEqual('true')
     expect(convertAttributeValue(true, 'number')).toEqual(1)
     expect(convertAttributeValue(false, 'reference', 'thing')).toEqual({ entityType: 'thing', entityName: 'false' })
+    expect(convertAttributeValue(true, 'json')).toEqual({ value: true })
 
     expect(convertAttributeValue({ entityType: 'thing', entityName: 'thing_one' }, 'string')).toEqual('thing_one')
     expect(convertAttributeValue({ entityType: 'thing', entityName: 'thing_one' }, 'number')).toEqual(0)
     expect(convertAttributeValue({ entityType: 'thing', entityName: 'thing_one' }, 'boolean')).toEqual(true)
+
+    expect(convertAttributeValue({ key: 'value' }, 'string')).toEqual('{"key":"value"}')
+    expect(convertAttributeValue({ key: 'value' }, 'number')).toEqual(0)
+    expect(convertAttributeValue({ key: 'value' }, 'boolean')).toEqual(true)
+    expect(convertAttributeValue({ key: 'value' }, 'json')).toEqual({ key: 'value' })
+
+    expect(() => convertAttributeValue('abc', 'notatype')).toThrow('Invalid attribute type "notatype"')
   })
 
   it('throws an error if attempting to convert to a reference without an entity type', () => {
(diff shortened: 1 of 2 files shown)


src/components/data/data-utils.js
```diff
@@ -73,7 +74,7 @@ export const renderDataCell = (data, googleProject) => {
 
   const renderArray = items => {
     return _.map(([i, v]) => h(Fragment, { key: i }, [
-      renderCell(v.toString()), i < (items.length - 1) && div({ style: { marginRight: '0.5rem', color: colors.dark(0.85) } }, ',')
+      renderCell(v?.toString()), i < (items.length - 1) && div({ style: { marginRight: '0.5rem', color: colors.dark(0.85) } }, ',')
```
nawatts, line 77: Discovered that arrays with null values would crash the UI.

src/components/data/_tests/data-utils.test.js
```diff
@@ -82,6 +82,10 @@ describe('getAttributeType', () => {
       ],
       itemsType: 'EntityReference'
     })).toEqual({ type: 'reference', isList: true })
+
+    expect(getAttributeType({ key: 'value' })).toEqual({ type: 'json', isList: false })
+    expect(getAttributeType(['a', 'b', 'c'])).toEqual({ type: 'json', isList: false })
+    expect(getAttributeType([{ idx: 0 }, { idx: 1 }, { idx: 2 }])).toEqual({ type: 'json', isList: false })
```
breilly2, line 88: Thinking Out Loud: I'm a little surprised that we support JSON typed attributes since we've effectively reserved some shapes of JSON objects to represent lists and entity references. In fact, I can trick it into switching types when editing JSON by adding an `items` property. 😬 

I don't remember us supporting JSON values in the past. Is this a relatively new addition? I'm not sure there's much we can do to fix it at this point, and I'm guessing that adding JSON editing to the UI is based on a commitment to support JSON values. 🤷‍♂️

breilly2:
I found some strange behavior when toggling between different types, specifically string and JSON.

https://user-images.githubusercontent.com/19228696/169111216-f64e042f-54a0-44e9-825f-cddb10db4cae.mov

On one hand, I think it makes sense to keep a stable version of the original value that can be used as the starting point for any conversion. On the other hand, that would cause any edits to the original value to be lost when switching types. Maybe we shouldn't allow switching types if the value has been edited? Or maybe we should warn that switching types could lose unsaved changes? What do you think?

nawatts:
> I found some strange behavior when toggling between different types, specifically string and JSON.

Ooh. Yeah, that's not great. Maybe we shouldn't attempt to convert JSON values to string and should instead reset to an empty string? Or maybe we shouldn't try to convert values at all when switching between types?

> On one hand, I think it makes sense to keep a stable version of the original value that can be used as the starting point for any conversion.

I'd claim that we do this already. Unsaved changes can be discarded by clicking "Cancel" and then editing the value again.

> On the other hand, that would cause any edits to the original value to be lost when switching types. Maybe we shouldn't allow switching types if the value has been edited? Or maybe we should warn that switching types could lose unsaved changes?

I think I'd claim there's enough precedent in this component for it to be likely understood that converting between types is lossy. For example, converting from a string to a number and back to string doesn't yield the original string, converting from a list to not a list drops all values after the first, etc.

breilly2:
For me, looking at an interface like this, I expect to be able to play "what-if" with the radio buttons and then land back where I started without other things changing. Maybe it would work to not change the value at all, indicate that the current value is invalid, and offer a function to try to auto-convert the value. However, that would be a more significant change that is probably not warranted here.

I don't know that any of the easier options would meet my expectations as a user, so I'm okay with either leaving it as-is or go with one of your other suggestions if you think that's better.

src/components/data/_tests/data-utils.test.js
```diff
@@ -82,6 +82,10 @@ describe('getAttributeType', () => {
       ],
       itemsType: 'EntityReference'
     })).toEqual({ type: 'reference', isList: true })
+
```
breilly2, line 85: I like that we have unit tests for this type of thing now!

breilly2 (approved):
👍  This is definitely an improvement for most use cases. I have concerns about supporting JSON values in general, but that's outside the scope of this PR.

src/components/data/_tests/data-utils.test.js
nawatts: > we've effectively reserved some shapes of JSON objects to represent lists and entity references. In fact, I can trick it into switching types when editing JSON by adding an items property.

Yes. JSON values matching that format actually get translated into lists by Rawls.
https://broadinstitute.slack.com/archives/C02T8G2FJCQ/p1652220052756079?thread_ts=1650480934.910629&cid=C02T8G2FJCQ

> I don't remember us supporting JSON values in the past. Is this a relatively new addition?

From what I can tell, it looks like Rawls has supported JSON values for about 5 years (https://broadinstitute.atlassian.net/browse/GAWB-1117). Handling them in the UI is a recent addition.

nawatts:
> For me, looking at an interface like this, I expect to be able to play "what-if" with the radio buttons and then land back where I started without other things changing.

🤔 I could see that. Or maybe use the current value as the basis for converting type until you change something in the input, and then switch to using the input value? Either way, I think that could be left as a potential future improvement?

src/components/data/_tests/data-utils.test.js
nawatts: One thing we could do is hide the JSON type option unless the attribute already contains JSON. So you could edit existing JSON attributes through the UI but not create new ones. Thoughts?

breilly2:
> maybe use the current value as the basis for converting type until you change something in the input, and then switch to using the input value

I like that improvement, and also agree that that can be a future improvement. I appreciate your help brainstorming ideas, so I'm okay with whatever you decide at this point.

nawatts:
>> maybe use the current value as the basis for converting type until you change something in the input, and then switch to using the input value
>
> I like that improvement, and also agree that that can be a future improvement.

Filed [SU-117](https://broadworkbench.atlassian.net/browse/SU-117) for this.

MatthewBemis (approved):

nawatts:
> I have concerns about supporting JSON values in general, but that's outside the scope of this PR.

https://github.com/DataBiosphere/terra-ui/pull/3049/commits/87e4f7e9b595d090d0af48ba68e9ecf622b015a1 updates this to only show the JSON type option if the attribute already contains JSON. So users could view and edit existing JSON values through the UI, but not add new JSON attributes.

PR #3050, opened 2022-05-18
[SU-89] Allow creating a set from the data table UI

nawatts:
Running a workflow on rows of a data table creates a set containing those rows. Users can also create a set by [uploading a membership TSV](https://support.terra.bio/hc/en-us/articles/360059242671-Adding-data-to-a-workspace-with-a-template#h_01F39ED3V80RE659R0B9934CKA).

Since creating and uploading a membership TSV is a cumbersome process, users will sometimes work around it by starting a workflow and immediately cancelling it just to create a set.

This adds the ability to create a set by selecting rows in the data table UI.

This is part of the new data tab design (enabled with `configOverridesStore.set({ isDataTabRedesignEnabled: true })`).

https://user-images.githubusercontent.com/1156625/169078271-3fb26d1b-3345-471f-93b1-10d8eebe946d.mov

### DIFF
--- a/src/components/data/EntitiesContent.js
+++ b/src/components/data/EntitiesContent.js
@@ -7,7 +7,7 @@ import { Fragment, useRef, useState } from 'react'
 import { div, form, h, input } from 'react-hyperscript-helpers'
 import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
 import { ButtonPrimary, ButtonSecondary, Link } from 'src/components/common'
-import { AddColumnModal, AddEntityModal, EntityDeleter, ModalToolButton, MultipleEntityEditor, saveScroll } from 'src/components/data/data-utils'
+import { AddColumnModal, AddEntityModal, CreateEntitySetModal, EntityDeleter, ModalToolButton, MultipleEntityEditor, saveScroll } from 'src/components/data/data-utils'
 import DataTable from 'src/components/data/DataTable'
 import ExportDataModal from 'src/components/data/ExportDataModal'
 import { icon, spinner } from 'src/components/icons'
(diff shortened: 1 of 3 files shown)


src/components/data/EntitiesContent.js
```diff
@@ -385,11 +386,17 @@ const EntitiesContent = ({
         h(MenuButton, {
           onClick: () => setAddingColumn(true)
         }, ['Add column']),
+        h(MenuDivider),
```
nawatts, line 389: UX requested a divider to separate actions that operate on the table as a whole from actions that operate on selected rows.

kyuksel: Posted [my thoughts](https://broadinstitute.slack.com/archives/C34GAKQ5R/p1652899589845669) in case they lead to further improvements in the future 😊

src/components/data/data-utils.js
```diff
@@ -937,6 +937,90 @@ export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames,
...
+    [!name, () => 'A name for the set is required.'],
+    [!/^[A-Za-z0-9_-]+$/.test(name), () => 'Set name may only contain alphanumeric characters, underscores, dashes, and periods.']
+  )
+
+  const [isBusy, setIsBusy] = useState()
+
+  const createSet = async () => {
+    try {
+      setIsBusy(true)
+
```
kyuksel: Nit: Would it be more clear what we think may error with the following?

```suggestion
    setIsBusy(true)
    try {
```

kyuksel (approved):
Excellent PR description! One nit inline.

src/components/data/data-utils.js
nawatts: That makes sense.

MatthewBemis (approved):

PR #3051, opened 2022-05-18
DC-350: Delete old App Engine versions

fboulnois:
A script to delete old Google App Engine deployments of Terra UI in an environment. You must have GNU date installed.

Usage: `./delete-old-app-engine-versions.sh ENV`

delete-old-app-engine-versions

### DIFF
--- a/scripts/delete-old-app-engine-versions.sh
+++ b/scripts/delete-old-app-engine-versions.sh
@@ -45,11 +45,6 @@ abort() {
     exit 1
 }
 
-# always reset gcloud project regardless of error or exit
-reset() {
-    gcloud config set project "${OLD_PROJECT}" 2>/dev/null
-}
-
 # ensure that date from GNU coreutils is installed
 check_gnu_date_installed() {
     if ! date --version 1>/dev/null 2>&1; then
@@ -57,15 +52,9 @@ check_gnu_date_installed() {
     fi
 }
 
-# switch gcloud project to appropriate terra ui environment
-switch_gcloud_project() {
-    gcloud config set project "${NEW_PROJECT}" 2>/dev/null
-    printf "${INFO} Switched the project to ${GRN}%s${RST}\n" "${NEW_PROJECT}"
-}
-
 # ensure that user has appropriate permissions for app engine
 check_user_permissions() {
-    if ! gcloud app services list 1>/dev/null 2>&1; then
+    if ! gcloud app services list --project="${NEW_PROJECT}" 1>/dev/null 2>&1; then
         GCLOUD_USER=$(gcloud config get-value account)
         abort "User ${GCLOUD_USER} does not have permissions in ${NEW_PROJECT}"
     fi
@@ -73,7 +62,7 @@ check_user_permissions() {
 
 # ensure that the deletion date is always older than the oldest pr date
 set_dev_deletion_date() {
-    OLDEST_PR=$(gcloud app versions list --filter="pr-" --sort-by="LAST_DEPLOYED" | tail -n +2 | head -n 1 | sed 's/ \+/ /g')
+    OLDEST_PR=$(gcloud app versions list --filter="pr-" --sort-by="LAST_DEPLOYED" --project="${NEW_PROJECT}" | tail -n +2 | head -n 1 | sed 's/ \+/ /g')
     OLDEST_PR_NAME=$(echo "${OLDEST_PR}" | cut -d' ' -f2)
     OLDEST_PR_DATE=$(echo "${OLDEST_PR}" | cut -d' ' -f4)
     OLDEST_PR_DATE=$(date --date="${OLDEST_PR_DATE}" +%F)
@@ -90,7 +79,7 @@ set_dev_deletion_date() {
 
 # return versions of app engine that match filter
 filter_app_engine_versions() {
-    gcloud app versions list --filter="$1" 2>/dev/null | tail -n +2 | sed 's/ \+/ /g' | cut -d' ' -f2
+    gcloud app versions list --filter="$1" --project="${NEW_PROJECT}" 2>/dev/null | tail -n +2 | sed 's/ \+/ /g' | cut -d' ' -f2
 }
 
 # ensure that deletions leave a certain number of deployments
@@ -129,7 +118,7 @@ deletion_preflight_summary() {
 
 # actually execute the deletion process
 execute_delete() {
-    printf "${BLD}gcloud app versions delete %s${RST}\n" "${DELETE_LIST_ITEMS[*]}"
+    printf "\n${INFO} Now run this command:${RST}\n${BLD}gcloud app versions delete %s --project=%s${RST}\n" "${DELETE_LIST_ITEMS[*]}" "${NEW_PROJECT}"
 }
 
 # ensure that a deletion phrase must be entered correctly before continuing
@@ -160,20 +149,18 @@ case $1 in
     * ) error "ENV must be one of dev, alpha, staging, or perf";;
 esac
 
-OLD_PROJECT=$(gcloud config get-value project)
 NEW_PROJECT="bvdp-saturn-$1"
 ENV_TO_EXEC="$1"
 
-trap reset EXIT
+printf "${INFO} Selected project ${GRN}%s${RST}\n" "${NEW_PROJECT}"
 
-switch_gcloud_project
 check_user_permissions
 
 DELETION_DATE=$(date --date="-7 days" +%F)
 if [ "$1" == "dev" ]; then
     set_dev_deletion_date
 fi
-printf "${INFO} Setting the deletion date to ${RED}%s and earlier${RST}\n" "${DELETION_DATE}"
+printf "${INFO} Set the deletion date to ${RED}%s and earlier${RST}\n" "${DELETION_DATE}"
 
 deletion_preflight_checks
 deletion_preflight_summary
--- a/scripts/delete-old-app-engine-versions.sh
+++ b/scripts/delete-old-app-engine-versions.sh
@@ -88,9 +88,14 @@ set_dev_deletion_date() {
     fi
 }
 
+# return versions of app engine that match filter
+filter_app_engine_versions() {
+    gcloud app versions list --filter="$1" 2>/dev/null | tail -n +2 | sed 's/ \+/ /g' | cut -d' ' -f2
+}
+
 # ensure that deletions leave a certain number of deployments
 check_remaining_items() {
-    REMAIN_LIST_ITEMS=($(gcloud app versions list --filter="version.createTime.date('%Y-%m-%d', Z)>'${DELETION_DATE}'" 2>/dev/null | tail -n +2 | sed 's/ \+/ /g' | cut -d' ' -f2))
+    REMAIN_LIST_ITEMS=($(filter_app_engine_versions "version.createTime.date('%Y-%m-%d', Z)>'${DELETION_DATE}'"))
     REMAIN_LIST_COUNT="${#REMAIN_LIST_ITEMS[@]}"
     if [ "${REMAIN_LIST_COUNT}" -lt 1 ]; then
         abort "all deployments would be deleted"
@@ -101,7 +106,7 @@ check_remaining_items() {
 
 # ensure that deletions erase a certain number of deployments
 check_deletion_items() {
-    DELETE_LIST_ITEMS=($(gcloud app versions list --filter="version.createTime.date('%Y-%m-%d', Z)<='${DELETION_DATE}'" 2>/dev/null | tail -n +2 | sed 's/ \+/ /g' | cut -d' ' -f2))
+    DELETE_LIST_ITEMS=($(filter_app_engine_versions "version.createTime.date('%Y-%m-%d', Z)<='${DELETION_DATE}'"))
     DELETE_LIST_COUNT="${#DELETE_LIST_ITEMS[@]}"
     if [ "${DELETE_LIST_COUNT}" -lt 1 ]; then
         abort "no deployments to delete"
--- a/scripts/delete-old-app-engine-versions.sh
+++ b/scripts/delete-old-app-engine-versions.sh
@@ -155,10 +155,7 @@ fi
 
 case $1 in
     --help ) usage;;
-    dev ) ;;
-    alpha ) ;;
-    staging ) ;;
-    perf ) ;;
+    dev|alpha|staging|perf ) ;;
     prod ) error "This script cannot be run against prod.";;
     * ) error "ENV must be one of dev, alpha, staging, or perf";;
 esac
--- a/scripts/delete-old-app-engine-versions.sh
+++ b/scripts/delete-old-app-engine-versions.sh
@@ -57,6 +57,12 @@ check_gnu_date_installed() {
     fi
 }
 
+# switch gcloud project to appropriate terra ui environment
+switch_gcloud_project() {
+    gcloud config set project "${NEW_PROJECT}" 2>/dev/null
+    printf "${INFO} Switched the project to ${GRN}%s${RST}\n" "${NEW_PROJECT}"
+}
+
 # ensure that user has appropriate permissions for app engine
 check_user_permissions() {
     if ! gcloud app services list 1>/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,184 @@
...
+
+# print out a summary of the deployments that will be affected
+deletion_preflight_summary() {
+    printf "${INFO} This script will ${GRN}keep %d${RST} deployments and ${RED}delete %d${RST} deployments\n" "${REMAIN_LIST_COUNT}" "${DELETE_LIST_COUNT}"
+    printf "${INFO} Deployments to keep (check this list):\n[ %s ]\n\n" "${REMAIN_LIST_ITEMS[*]}"
+}
+
+# actually execute the deletion process
+execute_delete() {
+    printf "${BLD}gcloud app versions delete %s${RST}\n" "${DELETE_LIST_ITEMS[*]}"
```
fboulnois: Currently this (purposely) only prints what the script would execute.

scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,184 @@
...
+if [ -z "${1+:}" ]; then
+    usage
+fi
+
+case $1 in
+    --help ) usage;;
+    dev ) ;;
+    alpha ) ;;
+    staging ) ;;
+    perf ) ;;
```
pshapiro4broad: You can combine cases with the same body
```suggestion
    dev|alpha|staging|perf) ;;
```

scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,184 @@
...
+    prod ) error "This script cannot be run against prod.";;
+    * ) error "ENV must be one of dev, alpha, staging, or perf";;
```
pshapiro4broad, line 147: This could also/or run `usage`

pshapiro4broad (approved):

scripts/delete-old-app-engine-versions.sh
fboulnois: I did this for consistency with most POSIX tools, which error and show a `try xyz --help ...` when arguments are mistyped. One minor reason this is useful is because it forces the user to be explicit about usage instead of immediately trying another option.

scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,181 @@
...
+# ensure that date from GNU coreutils is installed
+check_gnu_date_installed() {
+    if ! date --version 1>/dev/null 2>&1; then
+        abort "date from GNU coreutils is required"
+    fi
+}
+
+# switch gcloud project to appropriate terra ui environment
+switch_gcloud_project() {
+    gcloud config set project "${NEW_PROJECT}" 2>/dev/null
```
nawatts: Instead of modifying `gcloud` configuration, would it be preferable to add a `--project` flag to all invocations of `gcloud`?

scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,181 @@
...
+    printf "${INFO} Switched the project to ${GRN}%s${RST}\n" "${NEW_PROJECT}"
+}
+
+# ensure that user has appropriate permissions for app engine
+check_user_permissions() {
+    if ! gcloud app services list 1>/dev/null 2>&1; then
```
nawatts: Does this actually check necessary permissions for this script? A user could have viewer access but not be able to delete versions.

scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,181 @@
...
+        GCLOUD_USER=$(gcloud config get-value account)
+        abort "User ${GCLOUD_USER} does not have permissions in ${NEW_PROJECT}"
+    fi
+}
+
+# ensure that the deletion date is always older than the oldest pr date
```
nawatts, line 63: Is it possible that some versions needing to be cleaned up are newer than some PR sites?

scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,181 @@
...
+    UNIX_EPOCH_OLDEST_PR_DATE="$(date --date="${OLDEST_PR_DATE}" +%s)"
+
+    if [ "${UNIX_EPOCH_DELETION_DATE}" -gt "${UNIX_EPOCH_OLDEST_PR_DATE}" ]; then
+        DELETION_DATE=$(date --date="${OLDEST_PR_DATE} -1 day" +%F)
+    fi
+}
+
+# return versions of app engine that match filter
+filter_app_engine_versions() {
+    gcloud app versions list --filter="$1" 2>/dev/null | tail -n +2 | sed 's/ \+/ /g' | cut -d' ' -f2
```
nawatts: `--format` might be an easier way to get the desired information vs processing with tail, sed, and cut.

https://cloud.google.com/sdk/gcloud/reference#--format
https://cloud.google.com/sdk/gcloud/reference/topic/formats

scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,181 @@
...
+execute_delete() {
+    printf "${BLD}gcloud app versions delete %s${RST}\n" "${DELETE_LIST_ITEMS[*]}"
+}
+
+# ensure that a deletion phrase must be entered correctly before continuing
+enter_deletion_phrase() {
+    DELETION_PHRASE="yes delete ${DELETE_LIST_COUNT} deployments in ${ENV_TO_EXEC}"
+    printf "${RED}THIS OPERATION WILL IRREVERSIBLY DELETE ${DELETE_LIST_COUNT} DEPLOYMENTS FROM %s AND EARLIER IN THE ${NEW_PROJECT} PROJECT.${RST}\n" "${DELETION_DATE}"
+    printf "${INFO} To continue with the deletion process, type: ${BLD}%s${RST}\n" "${DELETION_PHRASE}"
+    read -r ACTUAL_PHRASE
```
nawatts, line 129: I think the end goal is to run this script automatically in CircleCI. Is there a way to skip this confirmation prompt in a non-interactive environment?

scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,181 @@
...
+    gcloud app versions list --filter="$1" 2>/dev/null | tail -n +2 | sed 's/ \+/ /g' | cut -d' ' -f2
+}
+
+# ensure that deletions leave a certain number of deployments
+check_remaining_items() {
+    REMAIN_LIST_ITEMS=($(filter_app_engine_versions "version.createTime.date('%Y-%m-%d', Z)>'${DELETION_DATE}'"))
+    REMAIN_LIST_COUNT="${#REMAIN_LIST_ITEMS[@]}"
+    if [ "${REMAIN_LIST_COUNT}" -lt 1 ]; then
+        abort "all deployments would be deleted"
+    elif [ "$REMAIN_LIST_COUNT" -lt 15 ]; then
```
nawatts, line 91: Why do we need to check how many deployments would remain? Can we not be confident that only the desired deployments are selected for deletion?

scripts/delete-old-app-engine-versions.sh
```diff
@@ -0,0 +1,181 @@
...
+}
+
+# ensure that deletions erase a certain number of deployments
+check_deletion_items() {
+    DELETE_LIST_ITEMS=($(filter_app_engine_versions "version.createTime.date('%Y-%m-%d', Z)<='${DELETION_DATE}'"))
+    DELETE_LIST_COUNT="${#DELETE_LIST_ITEMS[@]}"
+    if [ "${DELETE_LIST_COUNT}" -lt 1 ]; then
+        abort "no deployments to delete"
+    elif [ "${DELETE_LIST_COUNT}" -lt 30 ]; then
+        abort "less than 30 deployments to delete"
```
nawatts, line 103: Why abort if less than 30 deployments would be deleted? Should those deployments not be deleted?

kyuksel (commented):
Fancy scripting  🎉 

Should we add somewhere on the script or on the [wiki page this script should be referenced from](url), info on what Google account to have logged in with while running the script?

scripts/delete-old-app-engine-versions.sh
kyuksel: I was thinking the same.

scripts/delete-old-app-engine-versions.sh
fboulnois: I don't think we should clean up anything newer than some PR sites -- it will make handling the logic a lot more difficult.

scripts/delete-old-app-engine-versions.sh
fboulnois: I initially tried with `--format=json` but unfortunately it doesn't keep `sort-by` order and means that additional tools (jq) are required. I'll check if some of the other formats are supported and make processing easier.

scripts/delete-old-app-engine-versions.sh
fboulnois: I think it's both important to check _how many things get deleted_ and _how many things remain_. Both of these are guardrails to ensure that we are left with a working set of deployments. We could remove these guardrails, but in my experience developers assume conditions can never happen and when they do it ends up in disaster (see the recent issue Jira had with their deletion script). Even if the logic was correct on our side, could we guarantee that Google would never introduce an app engine bug in the `gcloud` command? It's defensive coding to ensure our inputs and outputs are what we expect.

scripts/delete-old-app-engine-versions.sh
fboulnois: The goal of the script is to be run when we're nearing the ~200 limit for deployments, so having <30 deployments means that someone recently ran the script / there's a potential issue with the deployments. It's another signal for the user to check what's going on.

scripts/delete-old-app-engine-versions.sh
fboulnois: I think we should introduce that in a future version, but I would strongly discourage us from being over-eager and committing this functionality now before we have been able to run the script manually a few times and ensure it is doing the right thing. I want to be ultra-conservative with the deletion script so we don't accidentally wipe any environments. When we are confident in the capabilities of the script, my thought is to require a specific scary env var to be defined beforehand (something like `YES_I_AM_SURE_I_WANT_TO_AUTOMATICALLY_DELETE_APP_ENGINE_DEPLOYMENTS`) and we can skip the confirmation prompt.

scripts/delete-old-app-engine-versions.sh
fboulnois: I'll give that a shot 👍

scripts/delete-old-app-engine-versions.sh
nawatts: > unfortunately it doesn't keep sort-by order and means that additional tools are required

There's no `--sort-by` in the command currently. Why does that make a difference?

scripts/delete-old-app-engine-versions.sh
fboulnois: That's true. Do you have a suggestion for what we should use instead? I don't see a good way of doing this without parsing the error from the `gcloud app versions delete` command.

scripts/delete-old-app-engine-versions.sh
nawatts: > I think it's both important to check how many things get deleted and how many things remain

But is _how many_ relevant remain? What we actually care about is _which specific versions_ remain. Wouldn't checking how many remain still pass if the script deleted all PR sites and left 15 unused versions?

> Even if the logic was correct on our side, could we guarantee that Google would never introduce an app engine bug in the gcloud command?

Since these checks are also based on the output of `gcloud`, would they actually help in that case?

scripts/delete-old-app-engine-versions.sh
nawatts: I thought the goal was to automate this to run on a regular basis? In that case, it seems like this would produce spurious failures at times where there happened to only be a few deployments.

scripts/delete-old-app-engine-versions.sh
nawatts: The approach I would take is not checking for permissions in advance and letting the `gcloud app versions delete` command fail if the user doesn't have adequate permissions. The error message from gcloud is likely informative enough for the audience using this script.

https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/ (the article is about Python, but the principle applies)

scripts/delete-old-app-engine-versions.sh
fboulnois: I noticed this when I was trying to find the oldest PR (line 76), which does have a `--sort-by`. At that point I decided that `jq` should not be necessary for the rest of the script especially since it doesn't guarantee sort order. For consistency I'd probably want to use one format everywhere.

scripts/delete-old-app-engine-versions.sh
fboulnois: My goal was to avoid having to manually select and delete a bunch of deployments in each environment since normalizing doing that is error prone and very scary. Currently since the script is new I still want to keep a human in the loop to make the ultimate decision of deleting these older deployments. In this situation, the script _should_ be maximally conservative and abort when there are too few deployments as it's a sign for the user to check what's going on.

scripts/delete-old-app-engine-versions.sh
fboulnois: It can't delete all PR sites, since it ensures that the deletion date is at least 1 week or older than the oldest PR, whichever is older (unless my logic here is wrong). I think having at least 15 deployments (maximum 10 PRs) is a good middle ground to be able to roll-back to a working version without manually having to check that each of these 5+ versions work.

I agree that it couldn't help in _all possible cases_ where something wrong might happen with the `gcloud` or the deployments, but it's about minimizing the risks with a minimum amount of checks. I think having these guardrails maximizes the possibility we will be left with working deployments with a minimal amount of code.

Are there checks you think should be included instead?

scripts/delete-old-app-engine-versions.sh
fboulnois: I think it's frustrating for a user to go through all of the steps required in the script and wait while checks complete including entering the deletion phrase only to be told that they don't have permissions to delete the objects anyways.

(Assuming I check the right permission) It's technically possible that the users permissions could change between that first step and the last step when they enter the deletion phrase, but it's about getting maximum value out of the script for the user with minimal frustration, and I think a permission check early on enables that goal.

nawatts: > I think it's frustrating for a user to go through all of the steps required in the script and wait while checks complete including entering the deletion phrase only to be told that they don't have permissions to delete the objects anyways.

But does this actually prevent that? A user with list permissions could still go through all the checks only to find they don't have permission to delete.

nawatts (changes requested):
Didn't test (I don't have access to the GCP projects).

I feel like this is more complicated and conservative than it needs to be. And that things like the confirmation prompt and refusing to delete less than 30 versions will be barriers to automating.

Nevertheless, it is an improvement over manually selecting versions in the Cloud Console. The only thing I want to request changes for is to have the `execute_delete` function run the delete command instead of printing it.

scripts/delete-old-app-engine-versions.sh
fboulnois: I am still trying to identify the best way to check that the user has the right permissions to delete without having to parse the error.

fboulnois:
> I feel like this is more complicated and conservative than it needs to be. And that things like the confirmation prompt and refusing to delete less than 30 versions will be barriers to automating.
> 
> Nevertheless, it is an improvement over manually selecting versions in the Cloud Console. The only thing I want to request changes for is to have the `execute_delete` function run the delete command instead of printing it.

I'll modify the `execute_delete` functionality.

How big do you think the barriers are to modifying this script to supporting automation? If we want to fully automate this process (I think that's a larger Terra UI conversation and outside the scope of this PR), I think there's a very low barrier / small decision to:

1. remove the preflight checks if we find out they are too conservative
2. remove the confirmation prompt or skip it if an environment variable is defined

nawatts:
I meant only that I think this script would require changes in order to be automated (run in CircleCI). I don't think those changes would be particularly difficult to make.

nawatts (approved):

kyuksel:
@fboulnois

- Thank you very much for your taking the initiative to automate the task of deleting old app engine deployments, and being diligent about getting it right. I also loved your back-and-forth's with @nawatts 😊 

- Would you mind updating the [related piece of the wiki](https://github.com/DataBiosphere/terra-ui/wiki/Security-and-Maintenance#cleaning-up-old-non-prod-app-engine-versions) as well?

- FYI that I created [TUIM-2](https://broadworkbench.atlassian.net/browse/TUIM-2) for adding the running of the script to CI. Please feel free to add to the ticket any details I may have missed.

petesantos (approved):

PR #3055, opened 2022-05-19
BT-655 Add Mixpanel event for opening Job Manager

breilly2:
This is so we have something to compare usage against the existing Workflow Dashboard (alpha) page view. Example Mixpanel report: https://mixpanel.com/s/1lGEVl
Screen Shot 2022-05-19 at 8 15 00 AM
Note: I added a "from" attribute (demonstrated in the screenshot above, which is why there are 3 colors) to differentiate between the two places where we link out to Job Manager. This may not be necessary because it would probably be inferred using Mixpanel flows, but I figured it didn't hurt to capture this data. I also included the workflow ID in case it's helpful for retrospectively studying what types of workflows people tend to use the different views for.

### DIFF
--- a/src/libs/events.js
+++ b/src/libs/events.js
@@ -32,6 +32,8 @@ const eventsList = {
   datasetLibraryBrowseData: 'library:browseData',
   dataTableSaveColumnSettings: 'dataTable:saveColumnSettings',
   dataTableLoadColumnSettings: 'dataTable:loadColumnSettings',
+  // Note: "external" refers to the common Job Manager deployment, not a Job Manager bundled in CromwellApp
+  jobManagerOpenExternal: 'job-manager:open-external',
   notebookLaunch: 'notebook:launch',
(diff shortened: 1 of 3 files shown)


cahrens:
@breilly2 when I added a mixpanel event, I found out much later that I was supposed to do [this](https://broadinstitute.slack.com/archives/C015QCR0SKY/p1648758406381279). Just passing the info in case you aren't aware.

cahrens (approved):

jgainerdewar (approved):
LGTM! Thanks for giving me a good excuse to get terra-ui running locally.

breilly2:
Thank you for the reminder, @cahrens. I had added a description in dev (just added event name as well). Looks like I can't add it to prod unless there's already data there? I tried to do an export from dev so I could import it into prod and, while it seemed to imply that it would export only selected rows, it exported all of them. I think I'm going to have to go back in and update the lexicon in prod after this has gone out.

PR #3057, opened 2022-05-19
[AJ-451] display contents of lists in data tables

davidangb:
If a data table cell contains an attribute list, display the contents of that list (at least the beginning of it) instead of the opaque "N items" that requires an additional click to see any values. This PR keeps the click-to-see-values in place; it just adds the ability to see contents before preview.

**_before_**
Screenshot (1)

**_after_**
Screenshot (11)

**_after, entity references_**
Screenshot (10)


nawatts (approved):
Could these be formatted a little more nicely?

Maybe `.join(', ')` instead of `JSON.stringify` for attribute value lists? And maybe formatting entity reference lists to avoid displaying the `entityName` and `entityType` keys? Also, maybe a separator between the number of items and the value?

davidangb:
@nawatts this has changed significantly since your review, so I re-requested review from you … thanks!

sagehen03 (approved):
looks good

nawatts (approved):
I like the new display much better, it's definitely more useful.

One issue I noticed is that when there's enough items to fill the cell, the edit button causes the items link to move when the cell is hovered. This can result in mis-clicking on the edit button when you're going for the items link.

https://user-images.githubusercontent.com/1156625/170109752-38c0f742-e41d-4e8f-bc3e-f13338ad7fed.mov

davidangb:
@nawatts yeah, I saw the same thing with the link jumping around and I don't have great ideas on fixing it, without changing the pre-existing behavior of the edit icon appearing/disappearing on hover. Do you have suggestions?

nawatts:
We could possibly re-order the elements (either `[dataCell, divider, editLink, itemsLink]` or `[itemsLink, dataCell, divider, editLink]`) or reserve space for the edit link (which could be done by wrapping it with a fixed width, `flexShrink: 0`, div).

PR #3059, opened 2022-05-19
[SU-100] Increase font weight of table headers

nawatts:
Requested by UX for new data tab design (enable with `configOverridesStore.set({ isDataTabRedesignEnabled: true })`). I made the change in the base table component so that all tables would be consistent.

## Before
Screen Shot 2022-05-19 at 4 18 59 PM

## After
Screen Shot 2022-05-19 at 4 18 45 PM

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -358,8 +358,7 @@ const DataTable = props => {
                         ]
                       }, [
                         h(HeaderCell, [
-                          !!columnNamespace && span({ style: { fontStyle: 'italic', color: colors.dark(0.75), paddingRight: '0.2rem' } },
-                            columnNamespace),
+                          !!columnNamespace && span({ style: { fontStyle: 'italic', color: colors.dark(0.75), paddingRight: '0.2rem' } }, [columnNamespace]),
                           columnName
                         ])
                       ])
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -359,9 +359,9 @@ const DataTable = props => {
                       }, [
                         h(HeaderCell, [
                           !!columnNamespace && span({ style: { fontStyle: 'italic', color: colors.dark(0.75), paddingRight: '0.2rem' } },
-                            columnNamespace)
-                        ]),
-                        [columnName]
+                            columnNamespace),
+                          columnName
+                        ])
                       ])
                     ]),
                     cellRenderer: ({ rowIndex }) => {
--- a/src/components/table.js
+++ b/src/components/table.js
@@ -13,6 +13,7 @@ import Interactive from 'src/components/Interactive'
 import Modal from 'src/components/Modal'
 import TooltipTrigger from 'src/components/TooltipTrigger'
 import colors from 'src/libs/colors'
+import { isDataTabRedesignEnabled } from 'src/libs/config'
 import { forwardRefWithName, useLabelAssert, useOnMount } from 'src/libs/react-utils'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
@@ -626,7 +627,7 @@ export const TooltipCell = ({ children, tooltip, ...props }) => h(TooltipTrigger
 }, [h(TextCell, props, [children])])
 
 export const HeaderCell = props => {
-  return h(TextCell, _.merge({ style: { fontWeight: 500 } }, props))
+  return h(TextCell, _.merge({ style: { fontWeight: isDataTabRedesignEnabled() ? 600 : 500 } }, props))
 }
 
 export const Sortable = ({ sort, field, onSort, children }) => {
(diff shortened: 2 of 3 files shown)


src/components/data/DataTable.js
```diff
@@ -359,9 +359,9 @@ const DataTable = props => {
                       }, [
                         h(HeaderCell, [
                           !!columnNamespace && span({ style: { fontStyle: 'italic', color: colors.dark(0.75), paddingRight: '0.2rem' } },
-                            columnNamespace)
-                        ]),
-                        [columnName]
+                            columnNamespace),
+                          columnName
+                        ])
```
nawatts, line 363: `columnName` was moved outside of the HeaderCell in #2314 / [AS-503](https://broadworkbench.atlassian.net/browse/AS-503), but is not necessary to accomplish the goal of that change (which is done by wrapping `columnNamespace` in a styled span (#2310).

This puts columnName back inside the HeaderCell so that it gets the expected header styling.

src/components/data/DataTable.js
```diff
@@ -359,9 +359,9 @@ const DataTable = props => {
                       }, [
                         h(HeaderCell, [
                           !!columnNamespace && span({ style: { fontStyle: 'italic', color: colors.dark(0.75), paddingRight: '0.2rem' } },
-                            columnNamespace)
-                        ]),
-                        [columnName]
+                            columnNamespace),
```
kyuksel: I know it's not your change but while touching it, can we enclose `columnNamespace` in an array like we do for other `children` even when it's a single element? FWIW, it took me additional time for my eyes to parse it because of that.

```suggestion
                            [columnNamespace]),
```

breilly2:
This is affecting other tables, in somewhat unexpected ways. It could be a good thing in the end, but I'm not sure. I think changing this universally would require auditing every place where this is used... which could be tricky given some places have very careful width settings to prevent ellipses or line-wrapping. Maybe target this more specifically at the new data table?

breilly2:
Oops... forgot to attach an example screenshot.
Screen Shot 2022-05-19 at 4 34 18 PM

Strangely, even without the config override, "Attribute" is still bolder, yet with the config override, "Variable" is still not bolder.

src/components/data/DataTable.js
```diff
@@ -359,9 +359,9 @@ const DataTable = props => {
                       }, [
                         h(HeaderCell, [
                           !!columnNamespace && span({ style: { fontStyle: 'italic', color: colors.dark(0.75), paddingRight: '0.2rem' } },
```
kyuksel: The exact style (i.e. `{ style: { fontStyle: 'italic', color: colors.dark(0.75), paddingRight: '0.2rem' } }`) is used in `UploadPreviewTable` too. I wonder if it's worth factoring it out so it's a bit more difficult to be inconsistent.

nawatts:
I'd claim that, rather than introducing an issue in the workflow inputs/outputs table, it made an existing issue more visible. Those tables currently have a mix of 3 different font weights for column heading: 500 for task name and type, 400 for variable, and 700 for attribute. I'll make those consistent.

kyuksel (approved):
👍🏼 
Minor comments inline.
Didn't test myself but the screenshot looks good.

src/components/data/DataTable.js
nawatts: Changed in https://github.com/DataBiosphere/terra-ui/pull/3059/commits/aa344dafdbcb3834811979146f87dee702f40c30.

nawatts:
> I think changing this universally would require auditing every place where this is used

I went through every file found by `git grep HeaderCell` and checked that if any column header in a table used HeaderCell, then every column in the table also used it.

breilly2:
`WorkflowView` was really just an example of a case that I found that was impacted outside the scope of the data table redesign that this PR is targeting. I was thinking that a visual inspection of other tables might be needed.

One that I know I was careful about setting column widths for is JobHistory. However, that didn't see to get the extra font weight with this change. I'm not sure why, but if it had, it definitely would have affected the layout and probably required more tweaking of the column widths.

All that said, I did try to find as many of these as I could, and `WorkflowView` was the only one I found that was affected. I probably didn't test all of them, but probably enough of them, and we can always fix others later.

breilly2 (approved):

nawatts:
Going down the rabbit hole, JobHistory doesn't use the HeaderCell component. It uses a different shared component for column headers: HeaderRenderer (I don't know why there are multiple components for column headers).

https://github.com/DataBiosphere/terra-ui/blob/661e8f817013cedb1b83685cd8b9564855035941/src/components/common.js#L573-L575

HeaderRenderer uses a font weight of 600. So this PR will bring HeaderCell and HeaderRenderer in line. JobHistory overrides that font weight though:
https://github.com/DataBiosphere/terra-ui/blob/96d986bb25a73a1541cf18b57744b5a4e8ce8681/src/pages/workspaces/workspace/JobHistory.js#L165-L174

PR #3063, opened 2022-05-20
[QA-1882] rework run-workflow ui test

aweng98:
Rework flaky function `launchWorkflowAndWaitForSuccess()`:
-  Handles unexpected scenarios.
- `pRetry` code fails to handle `Failed` status.

### DIFF
--- a/integration-tests/tests/run-workflow.js
+++ b/integration-tests/tests/run-workflow.js
@@ -49,5 +49,5 @@ const testRunWorkflowFn = _.flow(
 registerTest({
   name: 'run-workflow',
   fn: testRunWorkflowFn,
-  timeout: 15 * 60 * 1000
+  timeout: 20 * 60 * 1000
 })
--- a/integration-tests/utils/workflow-utils.js
+++ b/integration-tests/utils/workflow-utils.js
@@ -33,31 +33,28 @@ const launchWorkflowAndWaitForSuccess = async page => {
   }
 
   // Wait until status is Succeeded or Failed
-  await pRetry(async () => {
-      try {
-        await Promise.race([
-          findInGrid(page, 'Succeeded', { timeout: 60 * 1000 }),
-          findInGrid(page, 'Failed', { timeout: 60 * 1000 })
-        ])
-      } catch (e) {
-        console.info(`Workflow is running, elapsed time (minutes): ${((Date.now() - start) / (1000 * 60)).toFixed(2)}`)
-        throw new Error(e)
-      }
-    },
-    {
-      onFailedAttempt: error => {
-        console.log(`There are ${error.retriesLeft} retries left.`)
-      }, retries: 15, factor: 1
+  const workflowHasSucceeded = await pRetry(async () => {
+    try {
+      return await Promise.race([
+        findInGrid(page, 'Succeeded', { timeout: 60 * 1000 }).then(() => true),
+        findInGrid(page, 'Failed', { timeout: 60 * 1000 }).then(() => false)
+      ])
+    } catch (e) {
+      console.info(`Workflow is running, elapsed time (minutes): ${((Date.now() - start) / (1000 * 60)).toFixed(2)}`)
+      throw new Error(e)
     }
+  },
+  {
+    onFailedAttempt: error => {
+      console.log(`There are ${error.retriesLeft} retries left.`)
+    }, retries: 15, factor: 1
+  }
   )
 
   // pRetry will complete successfully when either Failed or Succeeded status is found.
   // We need to check status here to see if workflow has succeeded or not.
-  try {
-    await findInGrid(page, 'Failed', { timeout: 1000 })
+  if (!workflowHasSucceeded) {
     throw new Error('Workflow has failed')
-  } catch (err) {
-    // Succeeded
   }
 }
 
--- a/integration-tests/utils/workflow-utils.js
+++ b/integration-tests/utils/workflow-utils.js
@@ -51,7 +51,7 @@ const launchWorkflowAndWaitForSuccess = async page => {
     }
   )
 
-  // pRetry will complete successfully if either Failed or Succeeded status if found.
+  // pRetry will complete successfully when either Failed or Succeeded status is found.
   // We need to check status here to see if workflow has succeeded or not.
   try {
     await findInGrid(page, 'Failed', { timeout: 1000 })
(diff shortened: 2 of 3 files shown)


integration-tests/tests/run-workflow.js
```diff
@@ -57,33 +59,52 @@ const launchWorkflowAndWaitForSuccess = async page => {
...
+  // Long enough for the submission details to be Running or Failed.
+  // Workflows might sit in "Submitted" for a long time if there is a large backlog of workflows in the environment
+  await Promise.race([
+    findInGrid(page, 'Failed', { timeout: 5 * 60 * 1000 }),
+    findInGrid(page, 'Running', { timeout: 5 * 60 * 1000 })
+  ])
+
+  // If status is Failed, fails test now
+  try {
+    await findInGrid(page, 'Failed', { timeout: 1000 })
```
nawatts: I think this can be combined with the previous Promise.race with something like:

```js
const workflowStatus = await Promise.race([
  findInGrid(page, 'Failed', { timeout: 5 * 60 * 1000 }).then(() => 'Failed'),
  findInGrid(page, 'Running', { timeout: 5 * 60 * 1000 }).then(() => 'Running')
])
```

aweng98: like this?
Screen Shot 2022-05-20 at 1 47 26 PM

nawatts: Yes, I think that should work.

integration-tests/tests/run-workflow.js
```diff
@@ -57,33 +59,49 @@ const launchWorkflowAndWaitForSuccess = async page => {
...
+    }, retries: 15, factor: 1
+  }
   )
 }
 
 const testRunWorkflow = {
   name: 'run-workflow',
   fn: testRunWorkflowFn,
-  timeout: 15 * 60 * 1000
+  timeout: 20 * 60 * 1000
```
cahrens, line 52: Should also update `run-workflow-on-snapshot`, which uses the helper method above.

aweng98: Done.

integration-tests/tests/run-workflow.js
```diff
@@ -57,33 +59,49 @@ const launchWorkflowAndWaitForSuccess = async page => {
...
+    throw new Error('Workflow has failed')
+  }
+
+  // Wait until status is Succeeded or Failed
   await pRetry(async () => {
     try {
-      await findInGrid(page, 'Succeeded', { timeout: 65 * 1000 }) // long enough for the submission details to refresh
+      await Promise.race([
+        findInGrid(page, 'Succeeded', { timeout: 60 * 1000 }),
+        findInGrid(page, 'Failed', { timeout: 60 * 1000 })
```
nawatts: It looks like this changes the behavior of this test. Previously, the test expected the workflow to succeed. Now it looks like the test will pass if the workflow succeeds or fails. Is that intentional?

aweng98: no, you caught a bug. I added code to perform a final check after pRetry.
```
// pRetry will complete successfully if either Failed or Succeeded status if found.
  // We need to check status here to see if workflow has succeeded or not.
  try {
    await findInGrid(page, 'Failed', { timeout: 1000 })
    throw new Error('Workflow has failed')
  } catch (err) {
    // Succeeded
  }
 ```

**Review by breilly2 — APPROVED** [review]
👍  Looks good. Thank you Christina and Nick for your careful review!

**Comment by nawatts** [comment]
Conflicts are due to #3065 moving the `launchWorkflowAndWaitForSuccess` function to `integration-tests/utils/workflow-utils.js`.

**Comment by aweng98** [comment]
> Conflicts are due to #3065 moving the `launchWorkflowAndWaitForSuccess` function to `integration-tests/utils/workflow-utils.js`.

@nawatts Hi Nick, rebased just now. PTAL.

File: integration-tests/utils/workflow-utils.js [hunk 4]
```diff
@@ -13,27 +13,52 @@ const launchWorkflowAndWaitForSuccess = async page => {
...
+  // Long enough for the submission details to be Running or Failed.
+  // Workflows might sit in "Submitted" for a long time if there is a large backlog of workflows in the environment
+  const isRunning = await Promise.race([
+    findInGrid(page, 'Failed', { timeout: 5 * 60 * 1000 }).then(() => false),
+    findInGrid(page, 'Running', { timeout: 5 * 60 * 1000 }).then(() => true)
+  ])
+
+  // If status is not Running, fails test now
+  if (!isRunning) {
+    throw new Error('Workflow has failed')
```
nawatts, line 32: It looks like this error might also be thrown if the workflow has not started within 5 minutes. Should that be included in the error message?

integration-tests/utils/workflow-utils.js
```diff
@@ -13,27 +13,52 @@ const launchWorkflowAndWaitForSuccess = async page => {
...
   )
+
+  // pRetry will complete successfully when either Failed or Succeeded status is found.
+  // We need to check status here to see if workflow has succeeded or not.
+  try {
+    await findInGrid(page, 'Failed', { timeout: 1000 })
+    throw new Error('Workflow has failed')
+  } catch (err) {
+    // Succeeded
+  }
```
nawatts: Checking for success might read a little clearer.

```suggestion
  try {
    await findInGrid(page, 'Succeeded', { timeout: 1000 })
  } catch (err) {
    throw new Error('Workflow has failed')
  }
```

nawatts: Alternatively, maybe the workflow status from this line in the loop could be saved instead of looking for it in the table again.

```js
await Promise.race([
  findInGrid(page, 'Succeeded', { timeout: 60 * 1000 }),
  findInGrid(page, 'Failed', { timeout: 60 * 1000 })
])
```

nawatts (approved):

integration-tests/utils/workflow-utils.js
aweng98: If `Promise.race` call fails because it can't find `Failed` or `Running` (e.g. if workflow has not started within 5 min), error should be thrown and that should end the test.

integration-tests/utils/workflow-utils.js
aweng98: 👍🏻 Able to use this approach. Tested locally.

PR #3065, opened 2022-05-20
[TUIM-4] Delete Bueller

nawatts:
We do not use the Bueller mechanism for running our integration tests, and maintaining it adds complexity.


nawatts:
The only place I saw a reference to [integration-tests/scripts/makeBearerToken.js](https://github.com/DataBiosphere/terra-ui/blob/c957fd4e7627ecc34f99f58a23273863bf968a1e/integration-tests/scripts/makeBearerToken.js) was in Bueller.md, but I'm not sure if it's used for anything else, so I left it alone.

breilly2:
There's probably more that can be done here. Bueller needed to not have any dependency on jest because jest has some global state that conflicted with a persistent service. That's why we have the `jest` and `tests` folders, as well as the `registerTest` function. Without Bueller in play, we should be able to simplify this again (though maybe we'll still need to keep something like this around to support test-flakes?).

cahrens:
> The only place I saw a reference to [integration-tests/scripts/makeBearerToken.js](https://github.com/DataBiosphere/terra-ui/blob/c957fd4e7627ecc34f99f58a23273863bf968a1e/integration-tests/scripts/makeBearerToken.js) was in Bueller.md, but I'm not sure if it's used for anything else, so I left it alone.

I'm fairly confident this can be deleted, but also OK with progressive improvements/deletions.

kyuksel:
> The only place I saw a reference to [integration-tests/scripts/makeBearerToken.js](https://github.com/DataBiosphere/terra-ui/blob/c957fd4e7627ecc34f99f58a23273863bf968a1e/integration-tests/scripts/makeBearerToken.js) was in Bueller.md, but I'm not sure if it's used for anything else, so I left it alone.

I couldn't find any usages of it either. I'd say let's get rid of it. Worst case something fails and we put it back. At least we'd find out what it's for 😊

kyuksel (approved):
What a joy to see all the reds!

It'd be nice to do [what Brian suggested](https://github.com/DataBiosphere/terra-ui/pull/3065#issuecomment-1133180558) as well but a follow-up ticket for that also sounds good.

cahrens (approved):

breilly2:
> It'd be nice to do [what Brian suggested](https://github.com/DataBiosphere/terra-ui/pull/3065#issuecomment-1133180558) as well but a follow-up ticket for that also sounds good.

I was about to say that I think the simplification is more valuable than just deleting some unused code, but deleting that code also eliminates some dependencies that we no longer have to worry about keeping up-to-date, so yes, this is valuable on its own.

kyuksel:
`test-flakes` seems to be running fine as well.

integration-tests/test-sequencer.js
```diff
@@ -0,0 +1,25 @@
...
+const Sequencer = require('@jest/test-sequencer').default
+
+
+// Start these tests first since they are long-running.
+// This is most relevant on CI, where the number of simultaneous tests is capped.
+const testsToRunFirst = [
+  'create-interactive-analysis',
+  'run-workflow',
+  'run-workflow-on-snapshot'
+]
```
nawatts: `jest/1-run-workflow.integration-test.js` had a comment:

```
/*
 * File named with "1-" to make Jest start it first, since it's a long-running test.
 * This is most relevant on CI, where the number of simultaneous tests is capped.
 */
```

Not sure if that actually worked, since it didn't look like Jest was applying any order to tests by default. But these tests will now be run first.
https://jestjs.io/docs/configuration#testsequencer-string

integration-tests/tests/run-workflow.js
```diff
@@ -46,44 +46,8 @@ const testRunWorkflowFn = _.flow(
   await findInDataTableRow(page, testEntity.name, testEntity.attributes.input)
 })
 
-const launchWorkflowAndWaitForSuccess = async page => {
```
nawatts, line 49: Moved this out into a separate file (`utils/workflow-utils.js`). Previously, `run-workflow-on-snapshot` was requiring it from here. Now, requiring this module has a side effect since it calls `registerTest`.

This will conflict with #3063.

integration-tests/jest.config.js
```diff
@@ -2,7 +2,8 @@ module.exports = {
   verbose: false,
   preset: 'jest-puppeteer',
   testEnvironment: '<rootDir>/jest-circus-environment.js',
-  testRegex: '\\.integration-test\\.js$',
+  testMatch: ['<rootDir>/tests/*.js'],
```
nawatts, line 5: Glob seemed a little easier to read than regex.

https://jestjs.io/docs/configuration#testmatch-arraystring

integration-tests/test-sequencer.js
nawatts: Oh, which tests to run was controlled by CircleCI, not Jest.

https://github.com/DataBiosphere/terra-ui/blob/a53c5252cbacbe99ee885bec05f46444e5c0ac39/.circleci/config.yml#L173-L174

nawatts: Nice, CircleCI was helpful in figuring out what should be passed to `circleci tests split`.

https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9710/workflows/e9c4821a-d1e4-4203-ac1b-53c4cb4f5000/jobs/43255

```
Error autodetecting timing type, falling back to weighting by name. Autodetect no matching filename or classname.  If file names are used, double check paths for absolute vs relative.
Example input file: "/mnt/ramdisk/integration-tests/tests/analysis-context-bar.js"
Example file from timings: "jest/1-create-interactive-analysis.integration-test.js"
```

nawatts: Actually, I guess absolute vs relative path doesn't make a difference here. This failing on the first run is just a consequence of renaming the test files.

nawatts: If CircleCI is splitting tests by timing, I don't think we need to try to control the order of tests. Removed the custom sequencer.

nawatts:
Added calls to registerTest in the files in the tests directory and removed the jest directory.

Changed CircleCI configuration to get the list of tests from Jest rather than listing files. This way, Jest configuration for finding tests should be consistently applied when running tests locally and in CircleCI.

integration-tests/tests/analysis-context-bar.js
```diff
@@ -46,10 +47,9 @@ const testAnalysisContextBarFn = _.flow(
...
-const testAnalysisContextBar = {
+registerTest({
   name: 'analysis-context-bar',
   fn: testAnalysisContextBarFn,
-  timeout: 15 * 60 * 1000
-}
-
-module.exports = { testAnalysisContextBar }
+  timeout: 15 * 60 * 1000,
+  targetEnvironments: [] // temporarily disabled IA-3436
```
cahrens, line 54: Nice! FYI @xuscrbw

integration-tests/tests/run-workflow.js
cahrens: Sounds good. FYI @aweng98

cahrens (approved):
Thanks so much, @nawatts!

integration-tests/tests/analysis-context-bar.js
kyuksel: oh so we'll be able to disable tests more easily this way - cool  🎉

kyuksel (approved):
> Thanks so much, @nawatts!

+1

integration-tests/tests/analysis-context-bar.js
cahrens: And it's a good reminder that we can selectively disable per environment (ie., keep running on dev, disable on perf and staging).

breilly2 (approved):
👍  Nice. As an original author of Bueller, I'm sad that the effort didn't amount to value, but happy to see the decision to eliminate the unnecessary complexity going forward.

PR #3066, opened 2022-05-20
[DC-210] Added ability to manually add entry to data catalog

lu-c:
This was made to allow users to check the output of their scripts / preview what adding their data would look like, but otherwise does not affect the code.

- General gist of what we do is pass the json of the workspace in as a query parameter, and then manually push it into the request result

### DIFF
--- a/src/pages/library/DataBrowser.js
+++ b/src/pages/library/DataBrowser.js
@@ -216,11 +216,11 @@ const makeDataBrowserTableComponent = ({ sort, setSort, selectedData, toggleSele
   return DataBrowserTable
 }
 
-const Browser = () => {
+const Browser = (props) => {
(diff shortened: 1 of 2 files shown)


pshapiro4broad (commented):
Now that I see this in practice, I'm not sure if I like it. It's test code that we'll want to remove at some point, correct? Or are you thinking that we'll leave it in even after we're done debugging our code?

lu-c:
I was thinking it might be nice to keep around for a while, as we work on migrating datasets in - After making this feature, I found some other bugs with how we were transforming the data! In the short term, it'll be useful! In the long term... well, it's still kind of a useful "developer's inside information" tool. It's technically available for any user, so it's only security through obscurity

PR #3068, opened 2022-05-23
[SU-106] Add option to silence expiring NIH access notification

nawatts:
To access some controlled datasets/workspaces, you can link an NIH account to Terra.

This link expires after 30 days. Starting 24 hours before the link expires, we show a notification that the link is about to expire or has expired every time Terra loads (every new tab, every refresh, etc). Feedback from users has been that the frequency of the notification is annoying, especially considering that we don't currently provide an option to un-link an NIH account to stop the notifications after the link has expired.

This adds a "Do not remind me again" button to these notifications which prevents showing the notification again (unless an NIH account is re-linked and the expiration time changes).

Screen Shot 2022-05-23 at 4 13 16 PM

To mock an expiring link, edit these lines in auth.js:

https://github.com/DataBiosphere/terra-ui/blob/610c415f43526eb5dd36645abcdde11946d36605/src/libs/auth.js#L241-L242

```js
authStore.update(state => ({
  ...state,
  nihStatus: {
    linkExpireTime: 1653337361,
    linkedNihUsername: 'me@example.com'
  }
}))
```

The `linkExpireTime` is an epoch timestamp. For example, to mock an NIH link expiring in 12 hours, `linkExpireTime` would be set to the result of `Math.floor(Date.now() / 1000) + 12 * 3600`.

### DIFF
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -270,9 +270,10 @@ authStore.subscribe((state, oldState) => {
       // notification.
       const notificationId = hasExpired ? 'nih-link-expired' : 'nih-link-expires-soon'
 
-      // If/when the notification is muted, the current expiration time of the NIH link is stored in local preferences.
+      // If/when the notification is muted, the expiration time of the current NIH link is stored in local preferences.
       // This lets us apply the mute preference only to the current NIH link. If the user re-links their NIH account
-      // after muting notifications, we want to show these notifications the next time the link will expire.
+      // after muting notifications, we want to show these notifications when the new link will expire. In that case,
+      // the new link will have an expiration time greater than the time stored in the mute preference.
       const muteNotificationPreferenceKey = `mute-nih-notification/${notificationId}`
       const muteNotificationUntil = getLocalPref(muteNotificationPreferenceKey)
       if (muteNotificationUntil && muteNotificationUntil >= expireTime) {
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -259,25 +259,37 @@ authStore.subscribe(withErrorIgnoring(async (state, oldState) => {
 
 authStore.subscribe((state, oldState) => {
   if (state.nihStatus !== oldState.nihStatus) {
-    const notificationId = 'nih-link-warning'
     const now = Date.now()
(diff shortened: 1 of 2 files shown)


src/libs/auth.js
```diff
@@ -259,25 +259,37 @@ authStore.subscribe(withErrorIgnoring(async (state, oldState) => {
...
-    const expireStatus = Utils.cond(
-      [!expireTime, () => null],
-      [now >= expireTime, () => 'has expired'],
-      [now > expireTime - (1000 * 60 * 60 * 24), () => 'will expire soon']
-    )
-    if (expireStatus) {
+    const shouldNotify = expireTime && now > expireTime - (1000 * 60 * 60 * 24)
+    if (shouldNotify) {
+      const hasExpired = now >= expireTime
+      const notificationId = hasExpired ? 'nih-link-expired' : 'nih-link-expires-soon'
```
nawatts, line 271: Split this into two notification IDs so that the expired notification will still be shown if the expires soon notification is muted.

src/libs/auth.js
```diff
@@ -259,25 +259,37 @@ authStore.subscribe(withErrorIgnoring(async (state, oldState) => {
...
+      const expireStatus = hasExpired ? 'has expired' : 'will expire soon'
+
+      const muteNotificationPreferenceKey = `mute-nih-notification/${notificationId}`
+      const muteNotificationUntil = getLocalPref(muteNotificationPreferenceKey)
```
cahrens, line 271: I'm a bit confused about this variable. We don't start showing the notification after `muteNotificationUntil`. Isn't this the time when they pressed "Do not remind me again"?

cahrens, line 271: I'm a bit confused about this variable. We don't start showing the notification after `muteNotificationUntil`. Isn't this the time when they pressed "Do not remind me again"?

cahrens: If my understanding is correct, perhaps `muteNotificationStart` or `muteNotificationSelected`?

cahrens (approved):

MatthewBemis (approved):

nawatts:
Rebased to pull in fixes for data catalog tests in #3071.

PR #3071, opened 2022-05-24
[DC-359] Fix roles in the UI

s-rubenstein:
<!--
Hello, friend!
Thanks!

--!>
Integration tests seem to already be limited to dev and local, so these changes should work universally.

### DIFF
--- a/src/pages/library/dataBrowser-utils.js
+++ b/src/pages/library/dataBrowser-utils.js
@@ -83,7 +83,7 @@ const normalizeDataset = dataset => {
     dataReleasePolicy,
     contacts, curators, contributorNames,
     dataType, dataModality,
-    access: _.intersection(dataset.roles, ['reader', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
+    access: dataset.accessLevel === 'reader' || dataset.accessLevel === 'owner' ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
   }
 }
 


src/pages/library/dataBrowser-utils.js
```diff
@@ -80,7 +80,7 @@ const normalizeDataset = dataset => {
     dataReleasePolicy,
     contacts, curators, contributorNames,
     dataType, dataModality,
-    access: _.intersection(dataset.roles, ['reader', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
+    access: dataset.accessLevel === 'reader' || dataset.accessLevel === 'owner' ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
```
fboulnois, line 86:
```suggestion
    access: (dataset.accessLevel === 'reader' || dataset.accessLevel === 'owner') ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
```

Is there any reason we don't have the same code as before but using `dataset.accessLevel` instead? Is there any issue with operator precedence using the `w || x ? y : z` construction?

fboulnois (commented):
One question:

src/pages/library/dataBrowser-utils.js
s-rubenstein: Nope, parens are fine to use here. There shouldn't be issue with operator precedence, but it does seem easier to read this way.

fboulnois (approved):

nawatts (approved):

src/pages/library/dataBrowser-utils.js
```diff
@@ -80,7 +80,7 @@ const normalizeDataset = dataset => {
     dataReleasePolicy,
     contacts, curators, contributorNames,
     dataType, dataModality,
-    access: _.intersection(dataset.roles, ['reader', 'owner']).length > 0 ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
+    access: (dataset.accessLevel === 'reader' || dataset.accessLevel === 'owner') ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
```
kyuksel: FYI on another way of expressing the predicate in case it's easier to read:
```suggestion
    access: _.includes(dataset.accessLevel, ['reader', 'owner']) ? datasetAccessTypes.GRANTED : datasetAccessTypes.CONTROLLED
```

kyuksel: I'd also recommend creating a `datasetAccessLevels` similarly to `datasetAccessTypes` to avoid using string literals any more than necessary.

kyuksel (commented):
@s-rubenstein Added a couple of recommendations for your consideration but they're not blockers for the PR.

src/pages/library/dataBrowser-utils.js
s-rubenstein: Thank you! I'd been trying to remember the syntax for this, this is good.

I agree on `datasetAccessLevels` and avoiding raw strings, but I might file that as tech debt since its already an issue, and my goal is to fix the UI first.

PR #3072, opened 2022-05-24
[DC-362] add metrics to the retry logic in terra UI

slucasbroad:
added a retry handler for AJAX calls to return metrics to mixpanel.


src/libs/ajax.js
```diff
@@ -66,8 +74,8 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
       // requesterPaysError is true if the request requires a user project for billing the request to. Such errors
       // are not transient and the request should not be retried.
       const shouldRetry = !error.requesterPaysError
-
-      if (!shouldRetry) {
+      if (shouldRetry) {
```
nawatts: This reverses the retry logic and would only retry requester pays errors.

nawatts (changes requested):

src/libs/ajax.js
slucasbroad: Confused was by the use of double `!`.  Unsure why use a double not-not?

nawatts: > Confused was by the use of double `!`. Unsure why use a double not-not?

I suppose this could have been written as:
```js
const shouldNotRetry = Boolean(error.requesterPaysError)

if (shouldNotRetry) {
  throw error
}
```

However, negative variables are generally difficult to work with, since it's very easy to end up in a situation with a double negative that is (as you point out) difficult to understand (for example, `if (!shouldNotRetry) { ...`.

src/libs/ajax.js
```diff
@@ -65,9 +83,10 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
       // requesterPaysError may be set on responses from requests to the GCS API that are wrapped in withRequesterPays.
       // requesterPaysError is true if the request requires a user project for billing the request to. Such errors
       // are not transient and the request should not be retried.
-      const shouldRetry = !error.requesterPaysError
 
+      const shouldRetry = !error.requesterPaysError
       if (!shouldRetry) {
+        captureFailure()
```
nawatts: Should this be outside the `if (!shouldRetry)` block? As is, this would only be called for errors that are not retried (currently, only requester pays errors from the GCS API).

src/libs/events.js
```diff
@@ -16,6 +16,7 @@ const eventsList = {
   analysisDisableBeta: 'analysis:disable',
   analysisLaunch: 'analysis:launch',
   analysisCreate: 'analysis:create',
+  analysisRetry: 'analysis:retry',
```
nawatts: It looks like this is unused.

src/libs/ajax.js
```diff
@@ -51,11 +52,28 @@ const withCancellation = wrappedFetch => async (...args) => {
...
 const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
   const timeout = 5000
   const somePointInTheFuture = Date.now() + timeout
   const maxDelayIncrement = 1500
   const minDelay = 500
+  const captureFailure = _.once(() => {
+    const errorAddress = _.flow(
+      _.values,
+      _.find(v => _.includes(v, args[0]) && v)
+    )([..._.values(getConfig()), 'https://storage.googleapis.com/'])
```
nawatts: It looks like this is getting the API root for the URL of the request that failed? Why not capture the full URL to identify which endpoint failed?

src/libs/ajax.js
```diff
@@ -51,11 +52,28 @@ const withCancellation = wrappedFetch => async (...args) => {
...
+    Ajax().Metrics.captureEvent(Events.requestFailed, { test: errorAddress })
```
nawatts: Is there a more descriptive name for this property than `test`?

slucasbroad: Not too creative myself with names, have any suggestion I could use?

nawatts: Depending on the answer to this question: https://github.com/DataBiosphere/terra-ui/pull/3072#discussion_r880956884, maybe something like `apiUrl` or `requestUrl`?

src/libs/ajax.js
```diff
@@ -51,11 +52,20 @@ const withCancellation = wrappedFetch => async (...args) => {
   }
 }
 
+const captureRetryFailure = (...args) => {
```
nawatts: The name of this function suggests that it is used to capture failures of retries. In reality, it's used to capture the first failure of a request. Failures in retries are not captured (since this function is wrapped in `_.once`).

```suggestion
const captureRequestFailure = (...args) => {
```

src/libs/ajax.js
```diff
@@ -51,11 +52,20 @@ const withCancellation = wrappedFetch => async (...args) => {
...
+  const errorAddress = _.flow(
+    _.values,
+    _.find(v => _.includes(v, args[0]) && v)
+  )([..._.values(getConfig()), 'https://storage.googleapis.com/'])
+  Ajax().Metrics.captureEvent(Events.requestFailed, { test: errorAddress })
```
nawatts: A more descriptive name for this value would be useful when looking at data in Mixpanel.

```suggestion
  Ajax().Metrics.captureEvent(Events.requestFailed, { requestRoot: errorAddress })
```

src/libs/ajax.js
```diff
@@ -51,11 +52,20 @@ const withCancellation = wrappedFetch => async (...args) => {
   }
 }
 
+const captureRetryFailure = (...args) => {
+  const errorAddress = _.flow(
```
nawatts: A comment explaining what this does would be nice. I don't think it's immediately obvious from the code.

src/libs/events.js
slucasbroad: removed

src/libs/ajax.js
```diff
@@ -56,6 +57,7 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
   const somePointInTheFuture = Date.now() + timeout
   const maxDelayIncrement = 1500
   const minDelay = 500
+  //const captureFailure = _.once(() => captureRetryFailure(...args))
```
s-rubenstein: Do we still need the commented out method (and call further down?)

s-rubenstein (approved):
LGTM with the one comment maybe needing to be pulled out.

src/libs/ajax.js
```diff
@@ -76,12 +79,24 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
   return wrappedFetch(...args)
 })
 
+
+// Captures given course in error message, compares to root addresses in dav.json plus https://storage.googleapis.com/,
```
nawatts:
```suggestion
// Captures given course in error message, compares to root addresses in app configuration plus https://storage.googleapis.com/,
```

Spelling: "dav" vs "dev". Also, this isn't necessarily dev.json, but whatever configuration is specified by the environment.

src/libs/ajax.js
```diff
@@ -76,12 +79,24 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
...
+// and returns the root that matches the error.
+const captureRetryFailure = (...args) => {
```
nawatts:
```suggestion
const captureRequestFailure = (...args) => {
```

The current name is a bit misleading. It doesn't have anything to do with retries.

nawatts (commented):
Just a few more minor things. Also, commented out code should be removed.

src/libs/ajax.js
slucasbroad: It was just there for personal convenience if I needed to reinstate it back into withRetryOnError(), took it out now.

src/libs/ajax.js
slucasbroad: changed resolved

nawatts (approved):
When adding new Mixpanel events, please update the [Lexicon](https://help.mixpanel.com/hc/en-us/articles/360001307806-Lexicon-Overview).

src/libs/ajax.js
slucasbroad: the agreed upon Idea was a need to make the metrics easer to group to gather, so focus on a shared root rather than each one having a unique url.

nawatts:
Thinking about this some more, it seems like this will capture many request failures that we don’t actually care about because they’re part of the normal operation of Terra. For example, every time I load Terra, there are several 404 responses from Bard because I don’t have NIH, Anvil, etc accounts linked. This would report events for all of those. There are also some expected error responses in data tables. For example, if you try to delete a row referenced by another row, the API returns 429 to let you know that you have to delete the reference first. Is this going to capture so many false positives that any potential signal is lost in the noise?

src/libs/ajax.js
```diff
@@ -76,12 +77,24 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
   return wrappedFetch(...args)
 })
 
+
+// Captures given course in error message, compares to root addresses in app configuration plus https://storage.googleapis.com/,
+// and returns the root that matches the error.
+const captureRequestFailure = (...args) => {
+  const errorAddress = _.find(v => {
+    return _.includes(v, args[0])
+  },
```
petesantos: nit: this can be a one liner:
```suggestion
  const errorAddress = _.find(v => _.includes(v, args[0]),
```

petesantos:
> Thinking about this some more, it seems like this will capture many request failures that we don’t actually care about because they’re part of the normal operation of Terra. For example, every time I load Terra, there are several 404 responses from Bard because I don’t have NIH, Anvil, etc accounts linked. This would report events for all of those. There are also some expected error responses in data tables. For example, if you try to delete a row referenced by another row, the API returns 429 to let you know that you have to delete the reference first. Is this going to capture so many false positives that any potential signal is lost in the noise?

I agree, we can filter out the bond errors.

src/libs/ajax.js
```diff
@@ -76,12 +77,27 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
   return wrappedFetch(...args)
 })
 
+// Captures given course in error message, compares to root addresses in app configuration plus https://storage.googleapis.com/,
+// and returns the root that matches the error.
+const captureRequestFailure = (...args) => {
+  const errorAddress = _.flow(
+    _.filter(v => v !== 'https://broad-bond-dev.appspot.com'),
```
nawatts: This only filters out the dev version of this service. Also, should _all_ Bond errors be filtered out? Or only those expected in normal operation?

nawatts (changes requested):
Filtering out Bond errors by service URL should use the URL for the current environment (https://github.com/DataBiosphere/terra-ui/pull/3072#discussion_r887054667).

slucasbroad:
* add the helper function, `captureRequestFailure()`, in `ajax.js` to return a failed request field (that is not a bond error) to Mipanal. Is currently applied to `withErrorRejection()`.
* Refactor `shouldRetry` to `shouldNotRetry` to remove confusion involving double negatives.

Co-authored-by: Nick Watts <1156625+nawatts@users.noreply.github.com>
Co-authored-by: Pete Santos <50371603+petesantos@users.noreply.github.com>

slucasbroad:
> Filtering out Bond errors by service URL should use the URL for the current environment ([#3072 (comment)](https://github.com/DataBiosphere/terra-ui/pull/3072#discussion_r887054667)).

Could you clarify? Unsure about instead of using a direct root of `https://broad-bond-dev.appspot.com` and replacing it with the args[0] value of `https://broad-bond-dev.appspot.com/api/link/v1/fence`? Is it redundent?

nawatts:
> Could you clarify? Unsure about instead of using a direct root of https://broad-bond-dev.appspot.com and replacing it with the args[0] value of https://broad-bond-dev.appspot.com/api/link/v1/fence? Is it redundent?

In development, Terra UI makes Bond requests to `broad-bond-dev.appspot.com`. In production, it makes those requests to `broad-bond-prod.appspot.com`. Thus, filtering out requests to `broad-bond-dev.appspot.com` would not affect requests in production.

These service URLs are configured for each environment (dev, alpha, staging, prod, etc.) Instead of hardcoding a URL like `broad-bond-dev.appspot.com`, this should use the configured service URL (`getConfig().bondUrlRoot`).
https://github.com/DataBiosphere/terra-ui/blob/74a981df7a6a5a27a96bd13e5bae60644cbf8298/src/libs/ajax.js#L156-L169

slucasbroad:
> > Filtering out Bond errors by service URL should use the URL for the current environment ([#3072 (comment)](https://github.com/DataBiosphere/terra-ui/pull/3072#discussion_r887054667)).
> 
> Could you clarify? Unsure about instead of using a direct root of `https://broad-bond-dev.appspot.com` and replacing it with the args[0] value of `https://broad-bond-dev.appspot.com/api/link/v1/fence`? Is it redundent?

or you meant replace it with `getConfig().bondUrlRoot`

slucasbroad:
> > > Filtering out Bond errors by service URL should use the URL for the current environment ([#3072 (comment)](https://github.com/DataBiosphere/terra-ui/pull/3072#discussion_r887054667)).
> > 
> > 
> > Could you clarify? Unsure about instead of using a direct root of `https://broad-bond-dev.appspot.com` and replacing it with the args[0] value of `https://broad-bond-dev.appspot.com/api/link/v1/fence`? Is it redundent?
> 
> or you meant replace it with `getConfig().bondUrlRoot`

missed that thanks

nawatts (approved):
Question about filtering out Bond errors remains... is filtering out _all_ errors desired? Or only expected 404s?

slucasbroad:
> Filtering out Bond errors by service URL should use the URL for the current environment

I am personally not sure, more discussion with the team is required.

PR #3076, opened 2022-05-26
[SU-96] Support filterOperator in data table search

MatthewBemis:
UI support for the new filter operators introduced in https://github.com/broadinstitute/rawls/pull/1795

Hidden behind a feature flag. Enable with `configOverridesStore.set({ isSearchAwesomeNow: true })`

Screen Shot 2022-05-26 at 10 40 35 AM

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -2,7 +2,7 @@ import _ from 'lodash/fp'
 import { Fragment, useEffect, useRef, useState } from 'react'
 import { b, div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
-import { ButtonPrimary, Checkbox, Clickable, DeleteConfirmationModal, fixedSpinnerOverlay, Link } from 'src/components/common'
+import { ButtonPrimary, ButtonSecondary, Checkbox, Clickable, DeleteConfirmationModal, fixedSpinnerOverlay, Link, RadioButton } from 'src/components/common'
 import { concatenateAttributeNames, EditDataLink, EntityRenamer, HeaderOptions, renderDataCell, SingleEntityEditor } from 'src/components/data/data-utils'
 import { allSavedColumnSettingsEntityTypeKey, allSavedColumnSettingsInWorkspace, ColumnSettingsWithSavedColumnSettings, decodeColumnSettings } from 'src/components/data/SavedColumnSettings'
 import { icon } from 'src/components/icons'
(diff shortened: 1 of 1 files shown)


nawatts:
It looks like the search results don't update when changing the operator. And pressing enter / clicking the search button also doesn't update the results until the query changes. I suspect `filterOperator` needs to be added to the dependencies list for this `useEffect` hook.

```js
useEffect(() => {
  loadData()
  if (persist) {
    StateHistory.update({ itemsPerPage, pageNumber, sort, activeTextFilter })
  }
}, [itemsPerPage, pageNumber, sort, activeTextFilter, activeCrossTableTextFilter, refreshKey])
```

src/components/data/DataTable.js
```diff
@@ -255,6 +257,38 @@ const DataTable = props => {
...
         childrenBefore && childrenBefore({ entities, columnSettings, showColumnSettingsModal }),
         div({ style: { flexGrow: 1 } }),
+        isSearchAwesomeNow && h(MenuTrigger, {
+          side: 'bottom',
+          closeOnClick: false,
+          popupProps: { style: { width: 250 } },
+          content: h(Fragment, [
+            div({ style: { padding: '1rem' } }, [
+              div({ style: { fontWeight: 600 } }, ['Search logic']),
+              div({ role: 'radiogroup', 'aria-label': 'choose an operator to use for advanced search' }, [
```
nawatts, line 267: Alternatively, since there's already a visible label ("Search logic") for this group, `aria-labelledby` could be used to to associate it with this group vs adding a separate label with `aria-label`.

MatthewBemis:
This was also escaping the feature flag, fixed in bf9286e561ab195f9d71a2912da098478b8df9b7

nawatts (approved):

src/components/data/DataTable.js
```diff
@@ -255,6 +257,38 @@ const DataTable = props => {
...
+                    checked: filterOperator === 'AND',
+                    onChange: () => setFilterOperator('AND'),
+                    labelStyle: { padding: '0.5rem', fontWeight: 'normal' }
+                  })
+                ]),
+                div({ style: { paddingTop: '0.5rem' } }, [
+                  h(RadioButton, {
+                    text: 'OR (rows with any term)',
+                    name: 'advanced-search-operator',
+                    checked: filterOperator !== 'AND',
```
jordansafer: nit: Change to filterOperator === 'OR'. Since we default to 'AND' already, I don't see a case where filterOperator dooesn't equal AND or OR

jordansafer (approved):

src/components/data/DataTable.js
MatthewBemis: Done in 5d9a1af728d9103a18bf14038ab7653ac1d94aac

PR #3078, opened 2022-05-26
Possible improvements to delete app engine script

pshapiro4broad:
`sed` and `cut` didn't work well on my setup, so I replaced them with `awk`.

On my machine GNU date is `gdate`, not `date`, so I updated the script to handle that case.

### DIFF
--- a/scripts/delete-old-app-engine-versions.sh
+++ b/scripts/delete-old-app-engine-versions.sh
@@ -48,7 +48,13 @@ abort() {
 # ensure that date from GNU coreutils is installed
 check_gnu_date_installed() {
     if ! date --version 1>/dev/null 2>&1; then
-        abort "date from GNU coreutils is required"
+        if ! gdate --version 1>/dev/null 2>&1; then
+            abort "date from GNU coreutils is required"
+        fi
+        # GNU date may be installed as 'gdate' instead of 'date'.
+        DATE_CMD='gdate'
+    else
+      DATE_CMD='date'
     fi
 }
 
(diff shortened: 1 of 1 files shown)


scripts/delete-old-app-engine-versions.sh
```diff
@@ -48,7 +48,13 @@ abort() {
...
     if ! date --version 1>/dev/null 2>&1; then
-        abort "date from GNU coreutils is required"
+        if ! gdate --version 1>/dev/null 2>&1; then
+            abort "date from GNU coreutils is required"
+        fi
+        # GNU date may be installed as 'gdate' instead of 'date'.
+        DATE_CMD='gdate'
+    else
+      DATE_CMD='date'
     fi
```
fboulnois, line 58: I think this could more clearly be written as:

```sh
if date --version ... ; then
  DATE_CMD="date"
elif gdate --version ... ; then
  DATE_CMD="gdate"
else
  abort ...
fi
```

Is it possible to use the default MacOS version to run the date commands?

scripts/delete-old-app-engine-versions.sh
```diff
@@ -62,24 +68,24 @@ check_user_permissions() {
...
 # ensure that the deletion date is always older than the oldest pr date
 set_dev_deletion_date() {
-    OLDEST_PR=$(gcloud app versions list --filter="pr-" --sort-by="LAST_DEPLOYED" --project="${NEW_PROJECT}" | tail -n +2 | head -n 1 | sed 's/ \+/ /g')
-    OLDEST_PR_NAME=$(echo "${OLDEST_PR}" | cut -d' ' -f2)
-    OLDEST_PR_DATE=$(echo "${OLDEST_PR}" | cut -d' ' -f4)
-    OLDEST_PR_DATE=$(date --date="${OLDEST_PR_DATE}" +%F)
+    OLDEST_PR=$(gcloud app versions list --filter="pr-" --sort-by="LAST_DEPLOYED" --project="${NEW_PROJECT}" | tail -n +2 | head -n 1)
+    OLDEST_PR_NAME=$(echo "${OLDEST_PR}" | awk '{print $2}')
+    OLDEST_PR_DATE=$(echo "${OLDEST_PR}" | awk '{print $4}')
+    OLDEST_PR_DATE=$($DATE_CMD --date="${OLDEST_PR_DATE}" +%F)
```
fboulnois, line 74: Since we're already doing a lot using `head`, `tail`, `sed`, `cut`, `awk`, I wonder if we should just use `jq` for everything instead, perhaps it will be easier to manipulate the text streams.

fboulnois (commented):
Early suggestions:

pshapiro4broad:
Closing; after discussion with @fboulnois he will continue working on this script to improve portability in a future PR.

PR #3079, opened 2022-05-26
[SUP-666] Added wordBreak: 'break-word' to the table cells.

n8landolt:
## Description

Workspace names wouldn't overflow when underscores were used.

## Solution

Add word break specifically to the workspace cells.

## Testing

See screenshot

Added very long workspace name using underscores in both tables.

Screen Shot 2022-05-26 at 2 56 10 PM

### DIFF
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -215,7 +215,7 @@ const Environments = () => {
   const getWorkspaceCell = (namespace, name, appType, shouldWarn) => {
     return !!name ?
       h(Fragment, [
-        h(Link, { href: Nav.getLink('workspace-dashboard', { namespace, name }) }, [name]),
+        h(Link, { href: Nav.getLink('workspace-dashboard', { namespace, name }), style: { wordBreak: 'break-word'} }, [name]),
         shouldWarn && h(TooltipTrigger, {
           content: `This workspace has multiple active cloud environments${forAppText(appType)}. Only the latest one will be used.`
         }, [icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.warning() } })])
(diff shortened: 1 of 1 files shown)


nawatts (approved):
Alternatively, we use `text-overflow: ellipsis` in many other tables.

https://github.com/DataBiosphere/terra-ui/blob/16598338e69fc6a68b7ed34d00d1f2de3a242b7a/src/components/table.js#L621-L623

https://github.com/DataBiosphere/terra-ui/blob/16598338e69fc6a68b7ed34d00d1f2de3a242b7a/src/libs/style.js#L14

n8landolt:
@nawatts From what I gathered from the ticket and discussion is that we want to see the whole workspace name. Part of the bug is that hyphens will show the full name, but underscores wouldn't count as a break, and wouldn't wrap. Adding the wordBreak forced the wrap.

nawatts:
It's up to you which way to go.

For what it's worth though, we truncate workspace names in the main workspaces list. A tooltip could be added to show the full name.

Screen Shot 2022-05-31 at 10 26 36 AM

petesantos (approved):
I don't want to hold this PR up as it is a big improvement - but could you follow up with UX to see what the preferred styling is in this case - wrapping or ellipses?

Qi77Qi (approved):

Qi77Qi:
for this page, I think wrapping makes a bit more sense to me...

here's what Tiffany said
```[Tiffany Chan]
I personally prefer wrapping the text but let me check with the team to see if we have a preferred pattern for Terra / AoU (e
```

tiffanytuffany:
Finally joined github so I can add my comment: Design team approved that its more beneficial for users to see the full name of the workspace so we'd suggest wrapping the text

PR #3080, opened 2022-05-31
[SU-121] Fix alignment of menu buttons in data tab sidebar

nawatts:
This applies to the new data tab design (enable with `window.configOverridesStore.set({ isDataTabRedesignEnabled: true })`). Currently, long table names push the table menu buttons out of alignment.

After some failed attempts to fix this, I discovered `min-width: 0` is the key to truncating the text as desired here. See https://css-tricks.com/flexbox-truncated-text/ for more information.

This also tidies up some unnecessary styles and ensures there's some spacing between elements.

## Before
Screen Shot 2022-05-26 at 5 38 33 PM

## After
Screen Shot 2022-05-31 at 9 14 40 AM

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -89,7 +89,7 @@ const DataTypeButton = ({ selected, entityName, children, entityCount, iconName 
     role: 'listitem'
   }, [
     h(Clickable, {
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):

dmohs (approved):
Very nice.

PR #3088, opened 2022-05-31
[QA-1895] Integration test: speculative fix for navigation timeout error

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1895

Seeing sporadic UI test failures caused by Navigation timeout. 

This was recently added
```
const waitUntilLoadedOrTimeout = timeout => ({ waitUntil: ['load', 'domcontentloaded', 'networkidle0'], timeout })
```

I think it has to do with `waitUntil: load, domcontentloaded` options so I'm going to remove these two.

### DIFF
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -33,6 +33,8 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
       try {
         return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {} })
       } catch (err) {
+        console.error(err)
+        console.error(typeof err)
         const message = await err.text()
         throw message
       }
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -1,11 +1,10 @@
 const _ = require('lodash/fp')
 const uuid = require('uuid')
 const {
-  click, clickable, dismissNotifications, fillIn, findText, input, signIntoTerra, waitForNoSpinners, navChild, noSpinnersAfter
+  click, clickable, dismissNotifications, fillIn, findText, gotoPage, input, signIntoTerra, waitForNoSpinners, navChild, noSpinnersAfter, navOptionNetworkIdle
 } = require('./integration-utils')
 const { fetchLyle } = require('./lyle-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
-const { waitUntilLoadedOrTimeout } = require('../utils/integration-utils')
 
 
 const defaultTimeout = 5 * 60 * 1000
@@ -196,7 +195,7 @@ const overrideConfig = async (page, configToPassIn) => {
 }
 
 const enableDataCatalog = async (page, testUrl, token) => {
-  await page.goto(testUrl, waitUntilLoadedOrTimeout(60 * 1000))
+  await gotoPage(page, testUrl)
   await waitForNoSpinners(page)
 
   await findText(page, 'Browse Data')
@@ -209,7 +208,7 @@ const enableDataCatalog = async (page, testUrl, token) => {
 const clickNavChildAndLoad = async (page, tab) => {
   // click triggers a page navigation event
   await Promise.all([
-    page.waitForNavigation(waitUntilLoadedOrTimeout()),
+    page.waitForNavigation(navOptionNetworkIdle()),
     noSpinnersAfter(page, { action: () => click(page, navChild(tab)) })
   ])
 }
@@ -223,7 +222,7 @@ const viewWorkspaceDashboard = async (page, token, workspaceName) => {
 }
 
 const performAnalysisTabSetup = async (page, token, testUrl, workspaceName) => {
-  await page.goto(testUrl, waitUntilLoadedOrTimeout(60 * 1000))
+  await gotoPage(page, testUrl)
   await findText(page, 'View Workspaces')
   await overrideConfig(page, { isAnalysisTabVisible: true })
   await viewWorkspaceDashboard(page, token, workspaceName)
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -159,8 +159,8 @@ const select = async (page, labelContains, text) => {
   return click(page, `//div[starts-with(@id, "react-select-") and contains(normalize-space(.),"${text}")]`)
 }
 
-const waitForNoSpinners = async page => {
-  await page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
+const waitForNoSpinners = page => {
+  return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
 }
 
 // Puppeteer works by internally using MutationObserver. We are setting up the listener before
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -347,7 +347,7 @@ const withPageLogging = fn => async options => {
   return await fn(options)
 }
 
-const navOptionNetworkIdle = (timeout = 60 * 1000) => ({ waitUntil: ['domcontentloaded'], timeout })
+const navOptionNetworkIdle = (timeout = 60 * 1000) => ({ waitUntil: ['networkidle0'], timeout })
 
 const gotoPage = (page, url) => {
   return page.goto(url, navOptionNetworkIdle())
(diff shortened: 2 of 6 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -346,7 +353,7 @@ const withPageLogging = fn => async options => {
   return await fn(options)
 }
 
-const waitUntilLoadedOrTimeout = timeout => ({ waitUntil: ['load', 'domcontentloaded', 'networkidle0'], timeout })
+const waitUntilNetworkIdleOrTimeout = timeout => ({ waitUntil: ['networkidle0'], timeout })
```
aweng98, line 356: I'm renaming function accordingly since removing `load` and `domcontentloaded` options,

integration-tests/utils/integration-utils.js
```diff
@@ -159,8 +159,13 @@ const select = async (page, labelContains, text) => {
   return click(page, `//div[starts-with(@id, "react-select-") and contains(normalize-space(.),"${text}")]`)
 }
 
-const waitForNoSpinners = page => {
-  return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
+const waitUntilDomContentLoaded = page => {
+  return page.waitForSelector('div:not(empty):not([id="root"])') // Used to prevent checking loadingSpinner while page is blank
```
aweng98: This CSS selector should be a good start for most pages but probably not sufficient for edge cases.

integration-tests/utils/integration-utils.js
```diff
@@ -159,8 +159,14 @@ const select = async (page, labelContains, text) => {
   return click(page, `//div[starts-with(@id, "react-select-") and contains(normalize-space(.),"${text}")]`)
 }
 
-const waitForNoSpinners = page => {
-  return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
+// Used to prevent checking loadingSpinner while page is blank
+const waitUntilDomContentLoaded = page => {
+  return page.waitForSelector('div:not(empty):not([id="root"])')
```
aweng98: This CSS selector should be a good start for most pages but probably not sufficient for edge cases.

nawatts: If I understand this correctly, the purpose of this is to wait for the initial "Loading Terra..." message to go away? If so, the name seems misleading. Based on the name, I would have assumed this had something to do with the [DOMContentLoaded event](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event).

Also, this selector looks like it matches the loading message itself, so wouldn't this function return while the loading message is still visible?
https://github.com/DataBiosphere/terra-ui/blob/2c58e0f146245c119c5cdf4d1732591bad950ab1/public/index.html#L17-L30

Finally, was the selector meant to be `:not(:empty)` vs `:not(empty)`?

integration-tests/utils/integration-utils.js
```diff
@@ -192,6 +198,7 @@ const dismissNotifications = async page => {
 }
 
 const signIntoTerra = async (page, { token, testUrl }) => {
+  console.log('signIntoTerra ...')
```
nawatts, line 195: Is this intentional or left over from development?

integration-tests/utils/integration-utils.js
nawatts: I realize the old function was also called `waitUntil...`, but this name seems misleading. Based on the name, I would have assumed that this waits for the network to be idle. But it actually just returns options to be passed to another function which does the waiting.

nawatts:
Not necessarily for this PR, but based on the number of `page.goto` calls that had to be changed here, maybe a `goToPage` helper function would nice. So that timeout options don't have to be specified every time.

integration-tests/utils/integration-utils.js
aweng98: Purpose of `waitUntilDomContentLoaded` function is wait until page is not blank otherwise time out, to prevent checking for `Loading Terra ...` spinner too early.
It doesn't matter if the selector is finding the loading message so yes, this function returns while loading message is still visible. Waiting for spinner to stop is done by `await page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })`. 
See `waitForNoSpinners` function (below).

integration-tests/utils/integration-utils.js
aweng98: Intentional. It makes easier (hopefully for everyone) to follow test events when reading failed test log.

integration-tests/utils/integration-utils.js
aweng98: Good point. What do you think about `navigationOptionNetworkIdle` as new name?

aweng98 (commented):
👍🏻 `:not(:empty)` looks to me is a better selector. On home page `https://bvdp-saturn-staging.appspot.com/`, `:not(empty)` found 70 web elements. `:not(:empty)` found 51 web elements. 70 web elements included empty div element like `<div></div>` which is "bad".

aweng98:
Can you think a better way to hold off checking loading spinner until page is "ready" besides checking for blank page?

integration-tests/utils/integration-utils.js
```diff
@@ -159,8 +159,17 @@ const select = async (page, labelContains, text) => {
   return click(page, `//div[starts-with(@id, "react-select-") and contains(normalize-space(.),"${text}")]`)
 }
 
-const waitForNoSpinners = page => {
-  return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
+// Used to prevent checking loadingSpinner while page is blank
+const waitUntilDomContentLoaded = page => {
+  return Promise.all([
+    page.waitForFunction(() => document),
+    page.waitForSelector('div:not(:empty):not([id="root"])')
```
nawatts: I still don't understand this function. This selector matches the initial "Loading Terra..." message as well as anything rendered after that message goes away. So why is this complex selector needed? It seems like we could just wait for the root div with `#root"`.

integration-tests/utils/integration-utils.js
```diff
@@ -159,8 +159,17 @@ const select = async (page, labelContains, text) => {
...
+  ])
+}
+
+const waitForNoSpinners = async page => {
+  await waitUntilDomContentLoaded(page)
```
nawatts: It seems like `waitUntilDomContentLoaded` would only be needed on the initial page load. Since `waitForNoSpinners` is frequently used, is it worth calling `waitUntilDomContentLoaded` every time? Would putting it in `goToPage` work?

integration-tests/utils/integration-utils.js
aweng98: I have removed `waitUntilDomContentLoaded`.

integration-tests/utils/integration-helpers.js
```diff
@@ -34,6 +33,8 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
       try {
         return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {} })
       } catch (err) {
+        console.error(err)
+        console.error(typeof err)
```
aweng98, line 37: I'd like to use this log to determine why `await err.text()` fails sometimes.

nawatts: Isn't this already logged below? With:

```js
console.error(`Failed to create workspace: ${workspaceName} with billing project: ${billingProject}`)
console.error(e)
```

integration-tests/utils/integration-utils.js
```diff
@@ -346,7 +347,11 @@ const withPageLogging = fn => async options => {
   return await fn(options)
 }
 
-const waitUntilLoadedOrTimeout = timeout => ({ waitUntil: ['load', 'domcontentloaded', 'networkidle0'], timeout })
+const navOptionNetworkIdle = (timeout = 60 * 1000) => ({ waitUntil: ['domcontentloaded'], timeout })
```
nawatts: The PR description says:
> Seeing sporadic UI test failures caused by Navigation timeout.
> ...
> I think it has to do with waitUntil: load, domcontentloaded options so I'm going to remove these two.

But this removes `load` and `networkidle0`, not `domcontentloaded`. What changed?

Also, the function name is misleading since `navOptionNetworkIdle` does not wait for `networkidle`.

aweng98: Oops! I think I just forgot to change it back. it should be `networkidle0`.

integration-tests/utils/integration-helpers.js
aweng98: No. because that catches error caused by `await err.text()`. I want to see what the original `err` is and what the `typeof` is.

nawatts (approved):

cahrens (approved):

PR #3091, opened 2022-06-01
[SU-124] Defer loading submission statuses on workspaces list

nawatts:
Some users, have reported that the workspaces list page is slow to the point of being unusable (10s of seconds to load). This is particularly true for the Field Engineering team, who have hundreds to thousands of workspaces.

@MatthewBemis investigated and found that removing the `workspaceSubmissionStats` field from the list workspaces request reduced the time required for that request by as much as 90%.

We can't drop `workspaceSubmissionStats` entirely, since it is used to display the status of the last submitted workflow for each workspace and the list can also be filtered on submission status.

This changes the workspaces list page to make an initial request for the workspaces list without the `workspaceSubmissionStats` and then make a second request for `workspaceSubmissionStats` in the background. This lets us render the workspace list without waiting for the slow request. If it takes more than a second to load `workspaceSubmissionStats`, spinners are shown in place of the status icons. And if a user attempts to filter by submission status before submission statuses are loaded, "Loading submission statuses" is shown in place of a no results message.

The screen recording here uses an artificial delay for loading submission statuses since my account doesn't have enough workspaces/workflows to noticeably slow down this request.

https://user-images.githubusercontent.com/1156625/171487551-b92502ed-802f-4885-88a6-cf58aef89ada.mov

As a bonus, many places throughout the UI are using the `useWorkspaces` hook to populate a menu for selecting a workspace. For example, importing data, exporting data to a workspace, etc. prompt for a workspace. Removing the `workspaceSubmissionStats` field from the `useWorkspaces` hook should speed up all of those as well.

### DIFF
--- a/src/pages/workspaces/List.js
+++ b/src/pages/workspaces/List.js
@@ -4,7 +4,7 @@ import * as qs from 'qs'
 import { useEffect, useMemo, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
-import { HeaderRenderer, Link, Select, topSpinnerOverlay, transparentSpinnerOverlay } from 'src/components/common'
+import { DelayedRender, HeaderRenderer, Link, Select, topSpinnerOverlay, transparentSpinnerOverlay } from 'src/components/common'
 import FooterWrapper from 'src/components/FooterWrapper'
 import { icon, spinner } from 'src/components/icons'
 import { DelayedSearchInput } from 'src/components/input'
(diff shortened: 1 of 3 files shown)


src/pages/workspaces/List.js
```diff
@@ -244,6 +277,12 @@ export const WorkspaceList = () => {
                 })
               ]),
               div({ style: styles.tableCellContent }, [
+                loadingSubmissionStats && h(DelayedRender, [
```
nawatts, line 280: Showing lots of spinners might be controversial, but I thought it may be warranted given that users could previously find information here. To make them less obtrusive, `DelayedRender` is used so that they only show up if submission statuses take more than a second to load.

MatthewBemis (approved):
LGTM. I really like the UX of this!

kyuksel (approved):
The code changes and the video look good. 

I was running into the lag with my prod account. Looking forward to checking it after the changes are deployed there :)

cahrens (approved):
This looks great-- thanks @nawatts and @MatthewBemis !

PR #3094, opened 2022-06-02
[IA-3042] adding environment key into config files and update dropdown logic

xuscrbw:
This PR is to address an underlying issue that we have been running into when we update the Leo Cache images.

At the moment, we only have one config file with the list of images for all environments. This means when we update Leo code on dev through a PR merge, we will cause issues with the dev UI environment because the dropdown will be referencing images that are no longer cached. While this had not been an issue in the past, a recent Google behavior change has caused non-cached images to fail, which exposed this problem.

The code here is relatively straightforward. You can test this locally by adjusting the environmentKey variable in the config.json file, which will change the versions of the images in the dropdown when creating a custom environment. 

Screen Shot 2022-06-02 at 9 27 13 AM

We have created 4 files in the terra-docker-image-documentation bucket: 

terra-docker-versions-dev.json, terra-docker-versions-alpha.json, terra-docker-versions-staging.json, terra-docker-versions-prod.json

I will be making a companion PR in terra-docker to update the readme so we are aware of this new workflow. These will be replacing the current terra-docker-versions-new.json and terra-docker-versions.json files we have now.

The way I anticipate this working would be: User merges new leo cached images to dev on Leonardo > User updates -dev file, alpha gets promoted overnight > We update the -alpha file when we can to avoid too many failures.

During release, we update -staging when staging is promoted. When Prod is released, we update -prod.

The only worry I have is the alpha overnight promotion timing, since that will mean we will have failures on the UI tests. Not sure if there's a clean way to have it done in the same job(?)

For sake of testing, I manually altered the json files to have different versions of the Default GATK image for testing purposes. Before this is merged, I will change these files to be the same as the most updated version.

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/config/perf.json
+++ b/config/perf.json
@@ -26,5 +26,5 @@
   },
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
   "alphaRegionalityGroup": "Alpha_Regionality_Users",
-  "environmentKey": "dev"
+  "terraDockerVersionsFile": "terra-docker-versions-dev.json"
 }
--- a/config/perf.json
+++ b/config/perf.json
@@ -25,5 +25,6 @@
     "apiKey": "AQEBBAEkx4iE2KxNyI7Wx08EwU1ycTM7E4FMSmaibbMUQxNU6uQvuAJt7fyABAtFYSYfgEE"
   },
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
-  "alphaRegionalityGroup": "Alpha_Regionality_Users"
+  "alphaRegionalityGroup": "Alpha_Regionality_Users",
+  "environmentKey": "dev"
 }
--- a/config/perf.json
+++ b/config/perf.json
@@ -26,5 +26,5 @@
   },
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
   "alphaRegionalityGroup": "Alpha_Regionality_Users",
-  "terraDockerVersionsFile": "terra-docker-versions-dev.json"
+  "terraDockerVersionsFile": "terra-docker-versions-perf.json"
 }
--- a/config/perf.json
+++ b/config/perf.json
@@ -26,5 +26,6 @@
   },
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
   "alphaRegionalityGroup": "Alpha_Regionality_Users",
-  "terraDockerVersionsFile": "terra-docker-versions-perf.json"
+  "terraDockerVersionsFile": "terra-docker-versions-perf.json",
+  "terraDockerImageBucket": "terra-docker-image-documentation-perf"
 }
--- a/config/prod.json
+++ b/config/prod.json
@@ -24,5 +24,6 @@
   },
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
   "alphaRegionalityGroup": "Alpha_Regionality_Users",
-  "terraDockerVersionsFile": "terra-docker-versions-prod.json"
+  "terraDockerVersionsFile": "terra-docker-versions-prod.json",
+  "terraDockerImageBucket": "terra-docker-image-documentation-prod"
 }
--- a/config/prod.json
+++ b/config/prod.json
@@ -23,5 +23,6 @@
     "apiKey": "AQEBBAEGP2gTM2pIdJAQOIeNrm8dcTM7E4FMSmaibbMUQxNU6qy6nLPOBK8QfSvPFSsX8PQ"
   },
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
-  "alphaRegionalityGroup": "Alpha_Regionality_Users"
+  "alphaRegionalityGroup": "Alpha_Regionality_Users",
+  "environmentKey": "prod"
 }
--- a/config/prod.json
+++ b/config/prod.json
@@ -24,5 +24,5 @@
   },
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
   "alphaRegionalityGroup": "Alpha_Regionality_Users",
-  "environmentKey": "prod"
+  "terraDockerVersionsFile": "terra-docker-versions-prod.json"
 }
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -17,6 +17,7 @@ import TooltipTrigger from 'src/components/TooltipTrigger'
 import { cloudServices, machineTypes } from 'src/data/machines'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
+import { getConfig } from 'src/libs/config'
 import { withErrorReporting, withErrorReportingInModal } from 'src/libs/error'
 import Events, { extractWorkspaceDetails } from 'src/libs/events'
 import { betaVersionTag } from 'src/libs/logos'
@@ -689,6 +690,7 @@ export const ComputeModalBase = ({
       const { googleProject } = getWorkspaceObject()
       const currentRuntime = getCurrentRuntime(runtimes)
       const currentPersistentDisk = getCurrentPersistentDisk(runtimes, persistentDisks)
+      const envKey = getConfig().environmentKey
 
       Ajax().Metrics.captureEvent(Events.cloudEnvironmentConfigOpen, {
         existingConfig: !!currentRuntime, ...extractWorkspaceDetails(getWorkspaceObject())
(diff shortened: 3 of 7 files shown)


kyuksel:
> The only worry I have is the alpha overnight promotion timing, since that will mean we will have failures on the UI tests. Not sure if there's a clean way to have it done in the same job(?)

Integration test failures are expensive (in terms of the effort they require looking into) so it'd be really great if we don't cause known failures. I wonder if we could have Leonardo expose the bucket path through an API such that Terra UI (and any other UI that need to fetch image info) wouldn't have to keep track of what environment it is. It wouldn't even need to know the bucket path. There may be a UX implication of making one more API call but I feel we can mitigate that by caching the bucket path for some time since it won't be changing frequently.

xuscrbw:
> I wonder if we could have Leonardo expose the bucket path through an API such that Terra UI (and any other UI that need to fetch image info) wouldn't have to keep track of what environment it is.

Hm, I don't think I'm completely understanding what the API call would do in this case. Would the API pull the different version.json based on the environment without having to pass in a variable? I feel like the update json process would still need to happen unless the API could determine when the promotion process was occurring.

kyuksel:
> > I wonder if we could have Leonardo expose the bucket path through an API such that Terra UI (and any other UI that need to fetch image info) wouldn't have to keep track of what environment it is.
> 
> Hm, I don't think I'm completely understanding what the API call would do in this case. Would the API pull the different version.json based on the environment without having to pass in a variable? I feel like the update json process would still need to happen unless the API could determine when the promotion process was occurring.

For visibility: Talked to @xuscrbw offline about a scheme where a separate `gs://....json` file is created for each custom image we care about and it's kept track of in Leo [reference.conf](https://github.com/DataBiosphere/leonardo/blob/5f958804e92bb9f292be62049b269bd0d8277669/http/src/main/resources/reference.conf#L103) like custom image hashes are kept track of today. When a Leo PR is made to update a custom image hash, the json file pointer is also updated. The new Leo endpoint exposes the value of that pointer, which Terra UI queries to get. The json file pointer follows the same environment promotion cycle that custom image hash does so there shouldn't be any out-of-sync situation causing Terra UI tests to fail.

kyuksel (approved):

kyuksel:
> > > I wonder if we could have Leonardo expose the bucket path through an API such that Terra UI (and any other UI that need to fetch image info) wouldn't have to keep track of what environment it is.
> > 
> > 
> > Hm, I don't think I'm completely understanding what the API call would do in this case. Would the API pull the different version.json based on the environment without having to pass in a variable? I feel like the update json process would still need to happen unless the API could determine when the promotion process was occurring.
> 
> For visibility: Talked to @xuscrbw offline about a scheme where a separate `gs://....json` file is created for each custom image we care about and it's kept track of in Leo [reference.conf](https://github.com/DataBiosphere/leonardo/blob/5f958804e92bb9f292be62049b269bd0d8277669/http/src/main/resources/reference.conf#L103) like custom image hashes are kept track of today. When a Leo PR is made to update a custom image hash, the json file pointer is also updated. The new Leo endpoint exposes the value of that pointer, which Terra UI queries to get. The json file pointer follows the same environment promotion cycle that custom image hash does so there shouldn't be any out-of-sync situation causing Terra UI tests to fail.

For posterity: Upon further offline talk, we'll prevent the out-of-sync issue by appending a `gsutil cp` to the job that promotes Leo (and accordingly the custom image hash). For instance, as Leo `dev` is promoted to Leo `alpha`, the `gsutil` command will copy the `dev` json file to the `alpha` json file in GCS so that Terra UI will always display the cached images on the image pull-down menu.

config/prod.json
```diff
@@ -23,5 +23,6 @@
     "apiKey": "AQEBBAEGP2gTM2pIdJAQOIeNrm8dcTM7E4FMSmaibbMUQxNU6qy6nLPOBK8QfSvPFSsX8PQ"
   },
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
-  "alphaRegionalityGroup": "Alpha_Regionality_Users"
+  "alphaRegionalityGroup": "Alpha_Regionality_Users",
+  "environmentKey": "prod"
```
jack-r-warren:
```suggestion
  "terraDockerVersionsFile": "terra-docker-versions-prod.json"
```

xuscrbw: Addressed

src/components/ComputeModal.js
```diff
@@ -697,7 +698,7 @@ export const ComputeModalBase = ({
         currentRuntime ? Ajax().Runtimes.runtime(currentRuntime.googleProject, currentRuntime.runtimeName).details() : null,
         Ajax()
           .Buckets
-          .getObjectPreview(googleProject, 'terra-docker-image-documentation', 'terra-docker-versions.json', true)
+          .getObjectPreview(googleProject, 'terra-docker-image-documentation', getConfig().terraDockerVersionsFile, true)
```
jack-r-warren: One new question--will this work for paths inside the bucket? If `getConfig().terraDockerVersionsFile` was `myfolder/terra-docker-versions-temp-123.json`, would the getObjectPreview function know to do the small bit of directory traversal to grab the file?

jack-r-warren:
> 



> > > > I wonder if we could have Leonardo expose the bucket path through an API such that Terra UI (and any other UI that need to fetch image info) wouldn't have to keep track of what environment it is.
> > > 
> > > 
> > > Hm, I don't think I'm completely understanding what the API call would do in this case. Would the API pull the different version.json based on the environment without having to pass in a variable? I feel like the update json process would still need to happen unless the API could determine when the promotion process was occurring.
> > 
> > 
> > For visibility: Talked to @xuscrbw offline about a scheme where a separate `gs://....json` file is created for each custom image we care about and it's kept track of in Leo [reference.conf](https://github.com/DataBiosphere/leonardo/blob/5f958804e92bb9f292be62049b269bd0d8277669/http/src/main/resources/reference.conf#L103) like custom image hashes are kept track of today. When a Leo PR is made to update a custom image hash, the json file pointer is also updated. The new Leo endpoint exposes the value of that pointer, which Terra UI queries to get. The json file pointer follows the same environment promotion cycle that custom image hash does so there shouldn't be any out-of-sync situation causing Terra UI tests to fail.
> 
> For posterity: Upon further offline talk, we'll prevent the out-of-sync issue by appending a `gsutil cp` to the job that promotes Leo (and accordingly the custom image hash). For instance, as Leo `dev` is promoted to Leo `alpha`, the `gsutil` command will copy the `dev` json file to the `alpha` json file in GCS so that Terra UI will always display the cached images on the image pull-down menu.

Adding my own note here for posterity--after @xuscrbw and @Qi77Qi brought this up to devops, we decided that the promotion job itself isn't a reliable place to run this line of code (because Jenkins isn't the actual workhorse for deployments, isn't strictly necessary for deployments today, and won't be involved at all in the future). Instead, Leonardo's Dockerfile will curl the `-new.json` file, so that each Leonardo image knows what the latest images were _at the time the image was built_.

When Leonardo is started up, we'll have an initialization step encoded in Leonardo's Helm manifests that will take whatever `-new.json` file _it has stored_ (as in, whatever was the latest when that Leonardo was built) and will `gsutil cp` the file up to that environment's specific file--whether it be `-dev.json`, `-alpha.json`, whatever (Leonardo's Helm chart knows what environment it is deployed to). The benefit here is that the updating mechanism is entirely encapsulated with Leonardo's deployment: rollbacks and deployments, however they happen, will always keep the environment's images file exactly in sync with whatever was the latest when that environment's leonardo version was created.

config/perf.json
```diff
@@ -25,5 +25,6 @@
     "apiKey": "AQEBBAEkx4iE2KxNyI7Wx08EwU1ycTM7E4FMSmaibbMUQxNU6uQvuAJt7fyABAtFYSYfgEE"
   },
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
-  "alphaRegionalityGroup": "Alpha_Regionality_Users"
+  "alphaRegionalityGroup": "Alpha_Regionality_Users",
+  "terraDockerVersionsFile": "terra-docker-versions-dev.json"
```
jack-r-warren:
```suggestion
  "terraDockerVersionsFile": "terra-docker-versions-perf.json"
```

jack-r-warren (commented):
All infrastructure prerequisites have been merged, so we're waiting on release cycle for this particular change to be safe to merge. Once [this link](https://storage.googleapis.com/terra-docker-image-documentation-prod/terra-docker-versions-prod.json) returns successfully, we're set.

**However**, as we talked about offline, we can't use the `terra-docker-image-documentation` bucket at all in Terra UI--it is a multipurpose bucket not suitable for production use. There's a new bucket for each environment, you can see the full paths for each [at the bottom here](https://github.com/broadinstitute/terra-helmfile/pull/2880/files). Note that devops needs both the bucket name and the file path within the bucket to be individually configurable in Terra UI's config.json file. (Them being one value like `terra-docker-image-documentation-alpha/terra-docker-versions-alpha.json` would be fine, what won't work is a value like `alpha` and then using interpolation after the fact)

kyuksel (approved):

jack-r-warren (approved):
LGTM

PR #3095, opened 2022-06-02
[SU-29] Add delete data table button

nawatts:
Currently, deleting a data table requires selecting all rows and deleting them. Because of the Rawls API, this requires fetching the IDs of all entities in the table and passing them back to the delete entities endpoint. This adds a button to use a more efficient endpoint that requires only the entity type.

https://user-images.githubusercontent.com/1156625/171656214-3d79659a-0317-42cc-b84d-7e7f6c1b0f56.mov

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -8,7 +8,7 @@ import { AutoSizer } from 'react-virtualized'
 import * as breadcrumbs from 'src/components/breadcrumbs'
 import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
 import Collapse from 'src/components/Collapse'
-import { ButtonOutline, Clickable, Link, spinnerOverlay } from 'src/components/common'
+import { ButtonOutline, Clickable, DeleteConfirmationModal, Link, spinnerOverlay } from 'src/components/common'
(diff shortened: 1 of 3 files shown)


MatthewBemis (approved):

src/pages/workspaces/workspace/Data.js
```diff
@@ -544,9 +550,28 @@ const DataTableActions = ({ workspace, tableName, rowCount, onUpdateSuccess }) =
...
+      onUpdateSuccess: onRenameTable,
       namespace, name,
       selectedDataType: tableName
+    }),
+    deleting && h(DeleteConfirmationModal, {
+      objectType: 'table',
+      objectName: tableName,
+      onDismiss: () => setDeleting(false),
+      onConfirm: async () => {
+        setDeleting(false)
```
cahrens: Is there a reason to not use `withErrorReporting` and `setBusyState`?

nawatts: I was thinking that the `setBusyState(false)` after successfully deleting a table would cause a state update on an unmounted component because deleting the table removes the table and this menu from the sidebar. But I tried it with `withErrorReporting` and `withBusyState` and there was no warning.

cahrens (approved):

PR #3097, opened 2022-06-02
[SU-126] Remove feature flag for cross table and advanced search

nawatts:
Enable these features for all users.

Screen Shot 2022-06-02 at 12 07 47 PM

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -12,7 +12,6 @@ import { MenuButton, MenuTrigger } from 'src/components/PopupTrigger'
 import { GridTable, HeaderCell, paginator, Resizable } from 'src/components/table'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
-import { isSearchAwesomeNow } from 'src/libs/config'
 import { withErrorReporting } from 'src/libs/error'
 import Events, { extractWorkspaceDetails } from 'src/libs/events'
 import { getLocalPref, setLocalPref } from 'src/libs/prefs'
@@ -257,7 +256,7 @@ const DataTable = props => {
       }, [
         childrenBefore && childrenBefore({ entities, columnSettings, showColumnSettingsModal }),
         div({ style: { flexGrow: 1 } }),
-        isSearchAwesomeNow() && h(MenuTrigger, {
+        h(MenuTrigger, {
           side: 'bottom',
           closeOnClick: false,
           popupProps: { style: { width: 250 } },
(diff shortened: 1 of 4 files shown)


MatthewBemis (approved):

breilly2:
Really cool feature! I found a couple of oddities in my testing.

First, the single table search in the upper right is still enabled but seems to have no effect when there's a search term in the "all tables" filter. I would expect that the advanced search settings would apply to that as well, especially since that's right next to the single table search input.

I went to a table, selected all 5 rows, then performed a filter. The text above the table says "5 rows selected" and that's what gets used for "open with..." and "export". However, the select column header menu says "Filtered (3)".
Screen Shot 2022-06-03 at 8 14 24 AM

I'm not blocking this PR over either of these, but we should probably follow on with fixes/clarifications.

On the flip side, filtering, selecting all, and clearing the filter correctly retains the selected rows, so that's cool. 👍 

Out of curiosity, why is "settings" disabled when there's an "all tables" filter? Is that intentional?

breilly2 (approved):

nawatts:
Thanks @breilly2. Nice finds. I'll go ahead and merge this as-is and file tickets for follow up. We're kind of in a bind since this feature was already publicly announced in https://terra.bio/new-in-terra-uploading-and-organizing-data-just-got-easier/.

> First, the single table search in the upper right is still enabled but seems to have no effect when there's a search term in the "all tables" filter. I would expect that the advanced search settings would apply to that as well, especially since that's right next to the single table search input.

I'm not sure I agree about expecting the advanced search settings to apply to cross table search since that settings button is in a different section than the cross table search input. Definitely agreed that we need to clarify how the two search inputs interact though. Maybe clearing and disabling the single table search input when there's a cross table search query. Filed [SU-130](https://broadworkbench.atlassian.net/browse/SU-130) for this.

> I went to a table, selected all 5 rows, then performed a filter. The text above the table says "5 rows selected" and that's what gets used for "open with..." and "export". However, the select column header menu says "Filtered (3)".
> ...
> On the flip side, filtering, selecting all, and clearing the filter correctly retains the selected rows, so that's cool. 👍

🤔 Off the top of my head, I'm not sure how to fix the first without losing the second. We can't know if any of the selected rows are not in the filtered table without loading all pages of the table. Filed [SU-131](https://broadworkbench.atlassian.net/browse/SU-131).

> Out of curiosity, why is "settings" disabled when there's an "all tables" filter? Is that intentional?

It is. Searching across tables overrides column settings to show all columns.
https://github.com/DataBiosphere/terra-ui/blob/9eb0d2ee0487a3db636dd0aee16acf959bd95619/src/components/data/DataTable.js#L243-L244

We should add a tooltip explaining that though. That's easy enough to add to this PR (https://github.com/DataBiosphere/terra-ui/pull/3097/commits/4bf393e18d7899ff10fe5768ad6fc47e0dca8083).

PR #3098, opened 2022-06-02
[SU-115] Add Mixpanel event for cross table search

nawatts:
To enable cross-table search, use `window.configOverridesStore.set({ isSearchAwesomeNow: true })`. Then enter a query in the search box in the sidebar.

Screen Shot 2022-06-02 at 1 14 00 PM


MatthewBemis (approved):
I appreciate that you're also capturing the number of tables, I think that will be very insightful!

petesantos (approved):

PR #3099, opened 2022-06-02
[IA-3405] Fixes a notification bug

gpcarr:
Users have been reporting they are receiving "Error loading outdated analyses" when leaving their RStudio runtime idle for a while and then returning to it they get a bunch of notifications.

This PR adds logic to only check for outdated analyses if the runtime is `Running`

I tested this by creating an rstudio runtime with a really short autopause threshold (2min), leaving the runtime idle and then coming back to it

### DIFF
--- a/src/pages/workspaces/workspace/applications/AppLauncher.js
+++ b/src/pages/workspaces/workspace/applications/AppLauncher.js
@@ -187,16 +187,18 @@ const ApplicationLauncher = _.flow(
       setShouldSetupWelder(true)
     }
 
-    const findOutdatedAnalyses = withErrorReporting('Error loading outdated analyses', async () => {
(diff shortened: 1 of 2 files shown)


nawatts (approved):

Qi77Qi (approved):

gpcarr:
@nawatts looks like the `run-workflow` test failed which is unrelated to my changes - am I okay to merge?

nawatts:
I think generally the preference would be to [re-run the CircleCI workflow](https://support.circleci.com/hc/en-us/articles/360050303671-How-to-rerun-a-workflow) unless there's a test that is known to be consistently failing.

gpcarr:
Just kicked off a re-run!

PR #3100, opened 2022-06-02
[SU-62] Support for column renaming

MatthewBemis:
Screen Shot 2022-06-02 at 3 40 39 PM

Screen Shot 2022-06-02 at 3 40 44 PM


src/components/data/DataTable.js
```diff
@@ -497,6 +501,17 @@ const DataTable = props => {
       },
       onDismiss: () => setUpdatingEntity(undefined)
     }),
+    !!renamingColumn && h(RenameColumnModal, {
+      namespace, name,
+      entityType: _.head(entities).entityType,
```
nawatts:
```suggestion
      entityType,
```

I notice that this is consistent with other code, but alternatively, the `entityType` prop passed to DataTable can be used here vs getting it from the entities list.

src/components/data/RenameColumnModal.js
```diff
@@ -0,0 +1,55 @@
...
+        withErrorReporting('Error renaming column name.'),
+        Utils.withBusyState(setRenaming)
+      )(async () => {
+        await Ajax().Workspaces.workspace(namespace, name).renameEntityColumn(entityType, oldAttributeName, newAttributeName)
+        onSuccess()
+      })
+    }, ['Rename'])
+  }, [h(IdContainer, [id => h(Fragment, [
+    div('Workflow configurations that reference the current column name will need to be updated manually.'),
+    h(FormLabel, { htmlFor: id, required: true }, ['New Name']),
```
nawatts, line 36: Today I learned that this component exists.

src/components/data/RenameColumnModal.js
```diff
@@ -0,0 +1,55 @@
...
+import { ButtonPrimary, IdContainer, spinnerOverlay } from 'src/components/common'
+import { ValidatedInput } from 'src/components/input'
+import Modal from 'src/components/Modal'
+import { Ajax } from 'src/libs/ajax'
+import { withErrorReporting } from 'src/libs/error'
+import { FormLabel } from 'src/libs/forms'
+import * as Utils from 'src/libs/utils'
+
+
+export const tableNameInput = ({ inputProps, ...props }) => h(ValidatedInput, {
```
nawatts: What's the motivation for extracting this into a separate function? Given that this only sets a few props and is only called once, I think I'd vote to inline it.

src/components/data/RenameColumnModal.js
```diff
@@ -0,0 +1,55 @@
...
+  const [renaming, setRenaming] = useState(false)
+
+  return h(Modal, {
+    onDismiss,
+    title: 'Rename Column',
+    okButton: h(ButtonPrimary, {
+      disabled: renaming,
+      onClick: _.flow(
+        withErrorReporting('Error renaming column name.'),
+        Utils.withBusyState(setRenaming)
```
nawatts: Using the `Utils.withBusyState` wrapper, if the rename request is successful, `setRenaming(false)` is called after the `onSuccess` callback. Since `onSuccess` dismisses the modal, this results in `setRenaming(false)` causing a React warning about a state update on an unmounted component.

The way I've chosen to avoid this in some other components (SingleEntityEditor, AddColumnModal, etc) is to call `setBusy` and `reportError` directly. Alternatively, delaying `onSuccess` by wrapping it in `setTimeout(..., 0)` would probably work as well.
https://github.com/DataBiosphere/terra-ui/blob/c2f2ef23bffaf3f0c215638a3fb25b50ecbae5c2/src/components/data/data-utils.js#L1075-L1106

src/components/data/RenameColumnModal.js
```diff
@@ -0,0 +1,55 @@
...
+  const [newAttributeName, setNewAttributeName] = useState('')
+  const [renaming, setRenaming] = useState(false)
+
+  return h(Modal, {
+    onDismiss,
+    title: 'Rename Column',
+    okButton: h(ButtonPrimary, {
+      disabled: renaming,
+      onClick: _.flow(
+        withErrorReporting('Error renaming column name.'),
```
nawatts:
```suggestion
        withErrorReporting('Error renaming column.'),
```



Also FYI, there's a `withErrorReportingInModal` function that takes the `onDismiss` function and dismisses the modal before showing the error notification.

src/components/data/DataTable.js
```diff
@@ -497,6 +501,17 @@ const DataTable = props => {
...
+      oldAttributeName: renamingColumn,
+      onSuccess: () => {
+        setRenamingColumn(undefined)
+        Ajax().Metrics.captureEvent(Events.workspaceDataRenameColumn, extractWorkspaceDetails(workspace.workspace))
+        loadMetadata()
```
nawatts, line 510: Not necessary to change, but FYI the already loaded entityMetadata could be updated using `setEntityMetadata` to avoid this request.

src/components/data/RenameColumnModal.js
MatthewBemis: I'm guilty of copying this from another component. Inlined in c18c281c65133c0b6633cea508bb334896898066

jordansafer (approved):
Looks good!

src/components/data/RenameColumnModal.js
MatthewBemis: Addressed in 3612f06503e935b2c9f198fae6defb15f278b55b

src/components/data/RenameColumnModal.js
nawatts: Looks like it was inlined in RenameTableModal. Did you mean to inline it here as well?

src/components/data/DataTable.js
```diff
@@ -386,6 +389,7 @@ const DataTable = props => {
                         extraActions: [
```
nawatts, line 389: Not specific to this PR, but we should probably hide these actions to edit the column when selecting data for a workflow. Filed [SU-140](https://broadworkbench.atlassian.net/browse/SU-140) for this.

nawatts (approved):

src/components/data/RenameColumnModal.js
MatthewBemis: Oh whoops- yes I did

MatthewBemis: Inlined in the 2nd spot: d4c8621320b1efac31922b86ba8b07ec10709da4

PR #3101, opened 2022-06-03
[IA-3473] Fix bug where user incorrectly gets prompted to update their runtime

gpcarr:
This PR fixes a bug where:
- A user has an RStudio runtime that is `Stopped`
- that user navigates to the analyses tab, clicks on an Rmd file and clicks on the "Open" button that opens RStudio and the cloud environment modal opens and prompts the user to update their runtime when clicking this button should instead just start the user's runtime since it's currently paused.
Screen Shot 2022-06-03 at 12 19 53 AM
Screen Shot 2022-06-03 at 12 20 28 AM

This PR makes it so that when a user has a paused RStudio environment and want to launch an Rmd file, their runtime is started (same behavior as jupyter)
Screen Shot 2022-06-03 at 10 26 47 AM

We will be creating a follow up ticket to address RStudio modes in the UI - this PR fixes when a runtime is Stopped and a user wants to start it but we still need to address other corner cases

### DIFF
--- a/src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
+++ b/src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
@@ -90,7 +90,7 @@ const AnalysisLauncher = _.flow(
               { key: runtimeName, workspace, runtime, analysisName, mode, toolLabel, styles: iframeStyles }) :
             h(Fragment, [
               h(PreviewHeader, {
-                styles: iframeStyles, queryParams, runtime, analysisName, toolLabel, workspace, setCreateOpen,
+                styles: iframeStyles, queryParams, runtime, analysisName, toolLabel, workspace, setCreateOpen, refreshRuntimes,
                 readOnlyAccess: !(Utils.canWrite(accessLevel) && canCompute), onCreateRuntime: () => setCreateOpen(true)
               }),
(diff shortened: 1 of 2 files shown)


src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -292,7 +306,12 @@ const PreviewHeader = ({
                 Ajax().Metrics.captureEvent(Events.analysisLaunch,
                   { origin: 'analysisLauncher', source: tools.RStudio.label, application: tools.RStudio.label, workspaceName: name, namespace })
                 Nav.goToPath(appLauncherTabName, { namespace, name, application: 'RStudio' })
-              } else {
+                } else if (runtimeStatus === 'Stopped' && currentRuntimeTool === toolLabel) {
+               // we make it here because mode is undefined. we don't have modes for rstudio (that i'm aware of)
+                chooseMode('edit')
```
Qi77Qi: why is `chooseMode` needed here? just to force a reload or sth?

n8landolt: It's for line 284, so mode is defined and line 284 will be false, so we can continue to checking logic on the other headers (see line 339 on)

dmohs (approved):

Qi77Qi (approved):

PR #3102, opened 2022-06-03
QA-1825 Modify fiddly browser/test synchronization for signing in to FireCloud

breilly2:
Back before Terra, I was involved in the development of the FireCloud UI tests. At that time, I did a bunch of debugging around this exact same problem. The result is [here](https://github.com/broadinstitute/firecloud-ui/blob/ab678610367c8388d70de768db0ad84c6df63997/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/user/SignInPage.scala#L31). That's the origin of the comment explaining the fixed-duration delay that also made its way into the Terra UI tests (which I was also involved in the development of).

Based on my memory of that and some further experimentation, I've made a couple of changes here, including re-introducing that delay. I expanded on the code comment to try to make it more clear, but I'll also go into further detail here.

FireCloud UI's main `App` component has a funky initialization process where the initial render will update the component's state which will then cause a re-render. There are at least 3 separate render operations in this chain. This might not be a problem except that some UI elements get rendered before the page is fully ready. If you watch closely, you might even be able to see an extra flicker when loading the page. What's happening in the test is that `forceSignedIn` is being called before the final render, but the state change is still getting queued. When that final state change happens, the result is re-rendering the sign-in page (which is what gets captured in the screenshot). Since there's really no observable way to detect that the final render has happened, the only thing I can come up with to do is to add a fixed-duration sleep.

I changed the code to look for the sign in button instead of using `page.waitForResponse`. While I like the idea of using `waitForResponse`, it wasn't working reliably for me. In my manual testing, if I went directly to `https://firecloud.dsde-dev.broadinstitute.org/?return=terra#methods/gatk/echo_to_file/`, I also would not see `iframerpc`. However, if I navigated forward from Terra (like the test does) I _would_ see a request for `iframerpc`. Also, when running the test non-headless, there was never a request for `https://accounts.google.com/o/oauth2/iframerpc` even though the test seemed to work when running headless. Ultimately, it just seems too unreliable to use. The old FireCloud UI tests used the sign-in button as an initial check so I think that'll be fine here too.

In the end, this passed several `CONCURRENCY=10 RUNS=100 yarn test-flakes find-workflow` runs on my laptop so I think this will be a good improvement.

### DIFF
--- a/integration-tests/utils/firecloud-utils.js
+++ b/integration-tests/utils/firecloud-utils.js
@@ -10,7 +10,6 @@ const signIntoFirecloud = async (page, token) => {
   await page.waitForXPath('//title[text()="FireCloud | Broad Institute"]')
   await findText(page, 'content you are looking for is currently only accessible')
 
-  // console.log('waiting for google auth library to load')
   await page.waitForXPath('//*[@id="sign-in-button"]')
 
   /*
--- a/integration-tests/utils/firecloud-utils.js
+++ b/integration-tests/utils/firecloud-utils.js
@@ -1,4 +1,4 @@
-const { click, findText, waitForNoSpinners } = require('./integration-utils')
+const { click, delay, findText, waitForNoSpinners } = require('./integration-utils')
 
 
 const selectWorkspace = async (page, billingAccount, workspace) => {
@@ -7,6 +7,12 @@ const selectWorkspace = async (page, billingAccount, workspace) => {
 }
 
 const signIntoFirecloud = async (page, token) => {
+  await page.waitForXPath('//title[text()="FireCloud | Broad Institute"]')
+  await findText(page, 'content you are looking for is currently only accessible')
+
+  // console.log('waiting for google auth library to load')
+  await page.waitForXPath('//*[@id="sign-in-button"]')
+
   /*
    * The FireCloud not-signed-in page renders the sign-in button while it is still doing some
    * initialization. If you log the status of the App components state for user-status and auth2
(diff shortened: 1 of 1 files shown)


integration-tests/utils/firecloud-utils.js
```diff
@@ -7,28 +7,40 @@ const selectWorkspace = async (page, billingAccount, workspace) => {
 }
 
 const signIntoFirecloud = async (page, token) => {
+  await page.waitForXPath('//title[text()="FireCloud | Broad Institute"]')
+  await findText(page, 'content you are looking for is currently only accessible')
```
breilly2, line 11: This is not new, just relocating what used to be on lines 29-30.

integration-tests/utils/firecloud-utils.js
```diff
@@ -7,28 +7,39 @@ const selectWorkspace = async (page, billingAccount, workspace) => {
...
+
+  await page.waitForXPath('//*[@id="sign-in-button"]')
```
breilly2, line 13: This replaces the `page.waitForResponse` call from lines 22-27.

integration-tests/utils/firecloud-utils.js
```diff
@@ -7,28 +7,39 @@ const selectWorkspace = async (page, billingAccount, workspace) => {
...
-  await page.waitForResponse(response => {
-    return response.url().startsWith('https://accounts.google.com/o/oauth2/iframerpc') &&
-        response.request().method() === 'GET' &&
-        response.status() === 200
-  }, { timeout: 30 * 1000 }
-  )
-
-  await page.waitForXPath('//title[text()="FireCloud | Broad Institute"]')
-  await findText(page, 'content you are looking for is currently only accessible')
-  await waitForNoSpinners(page)
```
breilly2, line 31: I removed this because `waitForNoSpinners` is looking specifically for spinners as Terra UI renders them. Since we're in FireCloud UI (which renders spinners differently), this had no effect.

integration-tests/utils/firecloud-utils.js
```diff
@@ -7,28 +7,39 @@ const selectWorkspace = async (page, billingAccount, workspace) => {
...
+   *
    * If the page is used before this is complete (for example window.forceSignedIn adding
    * :signed-in to user-status), bad things happen (for example :signed-in being dropped from
-   * user-status). Instead of reworking the sign-in logic for a case that (for the most part) only
-   * a computer will operate fast enough to encounter, we'll just slow the computer down a little.
+   * user-status). Put another way, if forceSignedIn is called between the 2nd and 3rd render, the
+   * sign-in doesn't work. The visible effect (failure screenshot) is that the test gets stuck on
+   * the FireCloud sign-in page.
+   *
+   * There is no observable evidence that the component has settled into its final state. I (Brian)
```
nawatts, line 29: Is settling into its final state based on network requests? Could [`page.waitForNetworkIdle`](https://pptr.dev/#?product=Puppeteer&version=v13.1.3&show=api-pagewaitfornetworkidleoptions) be useful here?

But if this test is passing several hundred runs in a row using a fixed delay, then I'm happy with that approach.

nawatts (approved):

cahrens (approved):
Thanks for the history and clear comments!

integration-tests/utils/firecloud-utils.js
breilly2: Interesting idea. I'm not sure. Since the last thing that's needed for it to settle is a re-render, which may be the result of a network request, I suspect there's still a small gap. I wish this component was better about not rendering until it's actually ready to interact with. But, it's FireCloud UI which we hope to retire someday so I don't want to fuss with it any more than absolutely necessary.

salonishah11 (approved):
LGTM! 👍

integration-tests/utils/firecloud-utils.js
```diff
@@ -7,28 +7,39 @@ const selectWorkspace = async (page, billingAccount, workspace) => {
...
+   * that (for the most part) only a computer will operate fast enough to encounter, we'll just slow
+   * the computer down a little with a fixed-length sleep.
+   *
+   * Note that the amount of sleep needed is dependent on CPU load where the browser is running,
+   * which can be affected by CPU speed, memory, and other processes running at the same time
+   * (notably tests running in parallel). The duration below may be overkill in many conditions, but
+   * we want it to always be enough, whether running 10 parallel executions on a developer laptop or
+   * running in a CI server.
    */
-  await page.waitForResponse(response => {
```
breilly2, line 22: @aweng98, I really did like this approach, but as I described in the PR description, the difference between headless and non-headless was a deal-breaker for me. Also, neither this nor searching for the sign-in button are alone sufficient to guarantee that the page is ready. Since we can't completely eliminate the `delay`, I chose the sign-in button approach because I found it to be more reliable.

I'm looking forward to hearing your thoughts about this PR. Thanks.

aweng98: Wait for the `sign-in-button` is a good call.
There was a 1 sec sleep before, but I removed it in this [PR](https://github.com/DataBiosphere/terra-ui/pull/3036) because it didn't eliminate flaky failure. 
Maybe, waiting for `sign-in-button` and `delay(1500)` is going to remove the flakiness for good.
Since you have determined this is not a product or UI code issue, I think adding a [retry](https://github.com/DataBiosphere/terra-ui/commit/cab27f927dfa8a6dd58f4afb71a3466c104a01d5) will guarantee no more flaky failure. Or maybe we add it in future if flakiness continues.

aweng98 (approved):

PR #3105, opened 2022-06-03
[IA-3451] Persistent disk monthly cost showing unknown on env delete.

n8landolt:
## Description

When a user goes to delete an environment, the persistent disk monthly cost shows unknown.

## Solution

The issue is the diskType is being returned as a string, e.g. 'pd-standard' when it's really expecting the object listed in pdTypes. 
Use pdTypes.fromString() in the disk.details() method to return the proper object.


Qi77Qi (approved):

Qi77Qi:
Tested on PR site and works as expected
:shipit:

src/libs/ajax.js
```diff
@@ -1667,7 +1668,11 @@ const Disks = signal => ({
       details: async () => {
         const res = await fetchLeo(`api/google/v1/disks/${project}/${name}`,
           _.mergeAll([authOpts(), appIdentifier, { signal, method: 'GET' }]))
-        return res.json()
+        const resJs = res.json().then(val => {
+          val.diskType = pdTypes.fromString(val.diskType)
+          return val
+        })
+        return resJs
```
kyuksel: Would the following be more concise and consistent with how we typically mutate objects (using `_.set()`)?

```suggestion
        return res.json().then(_.set('diskType', pdTypes.fromString(val.diskType)))
```

Note that `_.set('diskType', pdTypes.fromString(val.diskType))` is equivalent to `val => _.set('diskType', pdTypes.fromString(val.diskType), val)` since [Lodash-FP functions are auto-curried](https://blog.codeminer42.com/functional-programming-with-lodash-fp-8fe0619b3024/#:~:text=Auto%20Curried%20Methods,calls%20provide%20the%20same%20result.).

n8landolt: Thank you for this! This looks way better

src/libs/ajax.js
```diff
@@ -1667,7 +1668,7 @@ const Disks = signal => ({
       details: async () => {
         const res = await fetchLeo(`api/google/v1/disks/${project}/${name}`,
           _.mergeAll([authOpts(), appIdentifier, { signal, method: 'GET' }]))
-        return res.json()
+        return res.json().then(val => _.set('diskType', pdTypes.fromString(val.diskType)))
```
kyuksel: `_.set('diskType', pdTypes.fromString(val.diskType))` would return a function that takes an object (e.g. `val`) and replace the value of that object's `diskType` property with the result of `pdTypes.fromString(val.diskType)`. Since `then()` expects such a function, we can have

```suggestion
        return res.json().then(_.set('diskType', pdTypes.fromString(val.diskType)))
```

n8landolt: The only issue with your suggested change is that val would then be undefined

kyuksel: It wouldn't. It would implicitly be what's wrapped inside `res.json()`. See below for a simpler example:

image

What's happening here is that we have a function (`_.set()` in this case but it doesn't matter) that takes three inputs. Let's say it is

```
const foo = (x, y, z) => x * y * z
```

We could also define the function in a curried fashion:

```
const foo = (x, y) => z => x * y * z
```

in which case the following would be `true`

```
foo(1, 2, 3) === foo(1, 2)(3)
```

The only thing that's special about Lodash FP is that most (all?) of its functions support both definitions of a `foo` and determine which one to use based on how many of the inputs are specified at a time.

Does that make sense?

n8landolt: Maybe I'm missing something, but in your examples you are defining all of the variables, but in your suggested code change val doesn't exist until it needs to be accessed? I put in your suggested change and I get 'val' is not defined as an error.

kyuksel: Ahhh so sorry. I overlooked the fact that `pdTypes.fromString()` needs a `diskType` and not the enclosing object, which necessitates explicitly specifying the input (i.e. `val`) to the function passed into to `then()` as you had.

kyuksel: Slight correction: Since we want `then()` to have a function that outputs the modified object, we'll want the following:
```suggestion
        return res.json().then(val => _.set('diskType', pdTypes.fromString(val.diskType), val))
```

Without that third argument into `_.set()`, it'll return a function.

n8landolt: Ahhh I'm finally understanding. On the 'val' object, we're setting the 'diskType' field to be equal to pdTypes.fromString(...), and still returning that val object.

kyuksel: Yes!

kyuksel (approved):

PR #3106, opened 2022-06-03
[QA-1898] Terra UI test: update waitForXPath

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1898

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -5,7 +5,7 @@ const { Storage } = require('@google-cloud/storage')
 const { screenshotBucket, screenshotDirPath } = require('../utils/integration-config')
 
 
-const waitForXpathWithVisibleTrue = opts => _.defaults({ visible: true }, opts)
+const defaultVisibleTrue = _.defaults({ visible: true })
 
 const waitForFn = async ({ fn, interval = 2000, timeout = 10000 }) => {
   const readyState = new Promise(resolve => {
@@ -32,7 +32,7 @@ const findIframe = async page => {
 }
 
 const findInGrid = (page, textContains, options) => {
-  return page.waitForXPath(`//*[@role="table"][contains(normalize-space(.),"${textContains}")]`, waitForXpathWithVisibleTrue(options))
+  return page.waitForXPath(`//*[@role="table"][contains(normalize-space(.),"${textContains}")]`, defaultVisibleTrue(options))
 }
 
 const getClickablePath = (path, text, textContains, isDescendant = false) => {
@@ -106,11 +106,11 @@ const clickTableCell = async (page, { tableName, columnHeader, text, isDescendan
 }
 
 const click = async (page, xpath, options) => {
-  return (await page.waitForXPath(xpath, waitForXpathWithVisibleTrue(options))).click()
+  return (await page.waitForXPath(xpath, defaultVisibleTrue(options))).click()
 }
 
 const findText = (page, textContains, options) => {
-  return page.waitForXPath(`//*[contains(normalize-space(.),"${textContains}")]`, waitForXpathWithVisibleTrue(options))
+  return page.waitForXPath(`//*[contains(normalize-space(.),"${textContains}")]`, defaultVisibleTrue(options))
 }
 
 const assertTextNotFound = async (page, text) => {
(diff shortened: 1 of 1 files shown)


petesantos:
Is this fixing a specific problem? What is the reason for this change?

aweng98:
UI elements may exists in DOM but not yet visible. This will help to reduce flakiness.

petesantos (approved):

integration-tests/utils/integration-utils.js
```diff
@@ -29,7 +29,7 @@ const findIframe = async page => {
   return hasFrame() || await waitForFn({ fn: hasFrame })
 }
 
-const findInGrid = (page, textContains, options) => {
+const findInGrid = (page, textContains, options = { visible: true }) => {
```
nawatts: This default value will only be used if no options argument is passed to the function. For example, if `{ timeout: 1000 }` is passed as options, `visible: true` will not be applied. Would something like [`_.defaults`](https://lodash.com/docs/4.17.15#defaults) be preferable here?

kyuksel: Good catch! Were you thinking something like

```javascript
const findInGrid = (page, textContains, options) => {
  const optionsWithVisibleTrueAsDefault = _.defaults({ visible: true }, options)
  return page.waitForXPath(`//*[@role="table"][contains(normalize-space(.),"${textContains}")]`, optionsWithVisibleTrueAsDefault)
}
```

?

aweng98: 👍🏻👍🏻

integration-tests/utils/integration-utils.js
```diff
@@ -5,6 +5,8 @@ const { Storage } = require('@google-cloud/storage')
 const { screenshotBucket, screenshotDirPath } = require('../utils/integration-config')
 
 
+const waitForXpathWithOptionVisibleTrue = opts => _.defaults({ visible: true }, opts)
```
kyuksel: 1. Because Lodash-FP supports auto-currying, `opts => _.defaults({ visible: true }, opts)` can just be `_.defaults({ visible: true })` without needing to change the call site.
2. It'd be good to make it clear in the function name that we're setting `visible` to `true` only if a caller doesn't specify a value for `visible`.

```suggestion
const waitForXpathWithVisibleTrueIfUnspecified = _.defaults({ visible: true })
```

integration-tests/utils/integration-utils.js
```diff
@@ -30,7 +32,7 @@ const findIframe = async page => {
 }
 
 const findInGrid = (page, textContains, options) => {
-  return page.waitForXPath(`//*[@role="table"][contains(normalize-space(.),"${textContains}")]`, options)
+  return page.waitForXPath(`//*[@role="table"][contains(normalize-space(.),"${textContains}")]`, waitForXpathWithOptionVisibleTrue(options))
```
kyuksel: As an alternative, this could be
```suggestion
  return page.waitForXPathWithVisibleTrueIfUnspecified(`//*[@role="table"][contains(normalize-space(.),"${textContains}")]`, options)
```

if we refactored the helper to be

```javascript
const waitForXPathWithVisibleTrueIfUnspecified = (page, xpath, options) => page.waitForXPath(xpath, _.defaults({ visible: true }, options)
```

Up to you to decide which one reads better 😊

aweng98: I feel suggested code is confusing, at least true for me. I'd prefer to leave it as-is if you don't feel strongly for change.

kyuksel: I don't feel strongly.

integration-tests/utils/integration-utils.js
```diff
@@ -5,6 +5,8 @@ const { Storage } = require('@google-cloud/storage')
 const { screenshotBucket, screenshotDirPath } = require('../utils/integration-config')
 
 
+const waitForXpathWithVisibleTrue = opts => _.defaults({ visible: true }, opts)
```
kyuksel: Can we rename the function to something that more closely matches what it does, e.g. `defaultVisibleToTrue`?

Note also that explicitly specifying input (`opts`) is unnecessary.

```suggestion
const defaultVisibleToTrue = _.defaults({ visible: true })
```

aweng98: Thanks, updated code.

kyuksel (approved):
Nit: Add a `To` to `defaultVisibleTrue` so it reads `defaultToVisibleTrue`

aweng98:
Test failures in CircleCI jobs are not related to this PR.

PR #3108, opened 2022-06-06
[QA-1899] Test failure tool: Save page's content to file when test fails

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1899

Save page's content to a .html file along with screenshot to help with test failure troubleshooting.

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -1,5 +1,5 @@
 const _ = require('lodash/fp')
-const { mkdirSync } = require('fs')
+const { mkdirSync, writeFileSync } = require('fs')
 const { resolve } = require('path')
 const { Storage } = require('@google-cloud/storage')
 const { screenshotBucket, screenshotDirPath } = require('../utils/integration-config')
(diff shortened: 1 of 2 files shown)


nawatts:
`maybeSaveScreenshot` has a section:
```js
if (screenshotBucket) {
  const storage = new Storage()
  await storage.bucket(screenshotBucket).upload(path)
  if (errorsPresent) {
    await storage.bucket(screenshotBucket).upload(failureNotificationDetailsPath)
  }
}
```

Does `savePageContent` also need something similar?

aweng98:
I think it's fine without it. Is anything (like dashboard) currently using the artifacts saved to the bucket?

nawatts:
> I think it's fine without it. Is anything (like dashboard) currently using the artifacts saved to the bucket?

👍  After some investigation, it looks like the screenshot bucket is a lingering remnant of the Bueller code, which was previously removed. So I don't think that anything is saving screenshots to a bucket.

nawatts (approved):

integration-tests/utils/integration-utils.js
```diff
@@ -277,11 +277,16 @@ const openError = async page => {
...
     screenshotDirPath :
     process.env.SCREENSHOT_DIR || process.env.LOG_DIR || resolve(__dirname, '../test-results/screenshots')
+  mkdirSync(dir, { recursive: true })
+  return dir
+}
 
+const maybeSaveScreenshot = async (page, testName) => {
+  const dir = getScreenshotDir()
   try {
     mkdirSync(dir, { recursive: true })
```
cahrens: Shouldn't need this because `getScreenshotDir` makes the directory.

cahrens (approved):
One minor comment, otherwise LGTM.

integration-tests/utils/integration-utils.js
aweng98: Yeah forgot to remove it. Thanks for catching that.

PR #3109, opened 2022-06-07
[IA-3474] Dismissing analyses modal blocks user input for too long.

n8landolt:
## Description

Clicking the + Start button on the analyses tab and then dismissing the modal caused a long pause in the UI while an asynchronous call was made.

## Solution

Simplify onDismiss to purely dismiss the modal window.
Add a new method, onError, which will capture the error cases.

## Testing

Opened and closed modals with success and error states.


jdcanas:
So I actually looked into this and addressed it  the same way (via moving `setCreating` to the top) since it was causing some test instability in this currently open PR 

https://github.com/DataBiosphere/terra-ui/pull/3037/files#diff-e19c52f5a9d3432ff7537aebcbb3d2a439a5fefc7346c1b0ac0017e2f9e8652bR462

Worth noting the same bug affects the onSuccess and onDismiss. I believe `refreshAnalyses`, `refreshApps`, and `refreshRuntimes` are all needed, as the modal can create an analysis, app, or runtime. However, it is an async function, so those portions won't cause anything to hang. 

editted to add: the behavior if we don't call the refresh functions here in the onDismiss/onSuccess is that an analysis could be added to the bucket but not show up on the page until it is manually refreshed. The same not-showing-up-until-page-refresh applies to runtimes/apps not showing up in the context bar for the respective refresh function

n8landolt:
My issue with calling **any** of the refresh methods on dismiss is that on dismiss happens when the user makes no changes. onSuccess will be able to take care of the refreshing. 

Also the UI hangs due to the method I removed setting busy to true, which blocks all user input until the method returns, even though it's asynchronous. I think it should still be called on success, but onDismiss will cause no changes to the user's workspace or analyses.

jdcanas:
> My issue with calling **any** of the refresh methods on dismiss is that on dismiss happens when the user makes no changes. onSuccess will be able to take care of the refreshing.
> 
> Also the UI hangs due to the method I removed setting busy to true, which blocks all user input until the method returns, even though it's asynchronous. I think it should still be called on success, but onDismiss will cause no changes to the user's workspace or analyses.

This issue is that onDismiss is passed to and called by all modals that are opened within the `AnalysisModal` in the corresponding onDismiss. For example, within the `ComputeModal` onDismiss is called after various calls to Leonardo that can fail. If a call to Leonardo fails, there may indeed be a change to the runtime's a user has (i.e. a runtime can now be in error status), which necessitates calling the refresh functions in onDismiss as well. We need to be very careful when modifying these onSuccess/onDismiss functions as child components often depend on them.

For an example of what I mean, see the `applyChanges` function within the `ComputeModal`

n8landolt:
I'll do some more digging around, I wonder if we want an "onFail" method in addition to the "onDismiss" and "onSuccess."

n8landolt:
I've implemented what I believe to be the correct solution for this ticket. I still need to do testing.

The reason I want an onError instead of using onDismiss as a catch all - onDismiss is supposed to be like a "reserved" function. It's intended to have the simplest implementation - close the modal.
onError will still dismiss the modal, but it will additionally call the proper refresh methods to update the state of the application.

Additionally, I think I want to wait until https://github.com/DataBiosphere/terra-ui/pull/3037 is merged - there may be merge conflicts. I may merge that branch into mine to make things smoother.

n8landolt:
I'm uncertain how to properly test - is there any easy way to force an error in the modals?

src/components/AnalysisModal.js
```diff
@@ -117,6 +117,10 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
         resetView()
         onDismiss()
       },
+      onError: () => {
```
jdcanas: may be worth defining onDismiss above in this case as to not duplicate the code

jdcanas (approved):
Like the approach... I'd just take a scan to ensure all trickle-down usages of onDismiss have no API calls associated with them

src/components/AnalysisModal.js
n8landolt: How does the modalDismiss function look to you?
Should I name it onModalDismiss ?

n8landolt:
Need to fix one more thing before I send this to a "real" PR

src/components/AnalysisModal.js
```diff
@@ -129,12 +137,15 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
...
+        onDismissModal()
+      },
+      onError: () => {
+        onError()
+        resetView()
       },
       onSuccess: () => {
-        setViewMode(undefined)
         onSuccess()
+        resetView()
```
petesantos: nit: Since the order matters in each of these calls, would it make sense to codify these as functions?

src/components/AnalysisModal.js
```diff
@@ -129,12 +137,15 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
       apps,
       appDataDisks,
       onDismiss: () => {
-        setViewMode(undefined)
-        onDismiss()
+        onDismissModal()
```
petesantos: nit: Could this be:

```onDismiss: onDismissModal,```

File: src/components/AnalysisModal.js [hunk 4]
```diff
@@ -312,8 +323,7 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
         titleStyles: { margin: '1.5rem 0 0 1.5rem', display: !!viewMode ? 'none' : undefined },
         width,
         onDismiss: () => {
-          resetView()
-          onDismiss()
+          onDismissModal()
```
petesantos: nit: same comment as Line 140

src/components/AnalysisModal.js
```diff
@@ -324,8 +334,7 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
     const modalProps = {
       isOpen, width, 'aria-labelledby': titleId,
       onDismiss: () => {
-        resetView()
-        onDismiss()
+        onDismissModal()
```
petesantos: nit: same comment as line 140

petesantos:
> I'm uncertain how to properly test - is there any easy way to force an error in the modals?

Are the errors you want to force a result of an API call? If so, you can use the `ajaxOverridesStore` util.
Example:
```javascript
window.ajaxOverridesStore.set([{
filter: { url: /rawls/ }, fn: window.ajaxOverrideUtils.makeError({ status: 404, frequency: 1 }) }])
}])
```

Will intercept all calls to rawls and return a 404. You can run that code in the browser console.

petesantos (commented):
A few nits.
Looking at the other comments - do you need help on testing?

src/components/AnalysisModal.js
n8landolt: Yes! Good catch. 

Does nit mean "nit picky?"

src/components/AnalysisModal.js
n8landolt: I added onSuccessModal and onErrorModal methods.
Not sure if that's what we want the naming convention to be, but I copied onDismissModal (even though the dismiss is referencing the modal)

Maybe onSuccessEvent and onErrorEvent instead?

src/components/AnalysisModal.js
kyuksel: Yes. I personally use it to also imply that my approval isn't contingent on addressing (by which I mean responding to a comment, whether it involves code changes or it's just a reply pushing back) it. I've added a topic to the UIWG agenda to get on the same page in terms of PR guidelines including nomenclature :)

petesantos (approved):
Looks good - thanks!
Please do take a look at Kaan's naming suggestion. My apologies, I know it was at odds with my naming suggestion.

src/components/AnalysisModal.js
```diff
@@ -49,10 +49,30 @@ export const AnalysisModal = withDisplayName('AnalysisModal')(
...
-      setNotebookKernel('python3')
+      //Set timeout is used here to match the drawer closing animation speed.
+      //This prevents the drawer from suddenly changing to default on close.
+      setTimeout(() => {
+        setViewMode(undefined)
+        setAnalysisName('')
+        setCurrentTool(undefined)
+        setNotebookKernel('python3')
+      },
+      200)
```
petesantos: I missed this earlier - Is there any way we could prevent coupling this to the animation speed of the drawer?

The modal drawer is based on react-modal. Is there a handler for when the modal has closed that we could use?

n8landolt: That's a good idea - I will look into it. If I can't get it done quick, should I remove this change for the time being?

n8landolt: I think I'm going to revert this change, and maybe another story can handle this specific issue.

My idea is to have all of the dismiss functions used by the RModal allow a timeout argument, so the passed in dismiss function would look like:

```
const onDismissCallback = (timeout = 0) => {
    setTimeout(() => {
      <Dismiss logic>
    }, timeout    
}
```

And the onRequestClose in ModalDrawer.js would be defined as such:

```
onRequestClose: () => {
  onDismiss(200)
},
```

PR #3111, opened 2022-06-07
[QA-1902] Fix launchWorkflowAndWaitForSuccess test helper

nawatts:
Currently, the `launchWorkflowAndWaitForSuccess` waits for a workflow to enter the "Running" state. However, the job history table polls for workflow status updates every minute, so the test is not guaranteed to see a "Running" state if the workflow starts and completes in between polls. The test handles the case where the workflow fails quickly, but it does not handle the case where the workflow succeeds quickly.

This test is currently frequently failing on dev. Looking at the screenshots shows that the workflow actually succeeded. Examples:
- https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9966/workflows/dbafece7-8b97-4954-add3-a58c0dbb0363/jobs/44314/tests
- https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9964/workflows/62f0d879-6e0c-43e6-b851-17f362a27376/jobs/44306/tests
- https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/9963/workflows/545a30fb-ef0e-4831-9470-192b46ef0d53/jobs/44304/tests

### DIFF
--- a/integration-tests/utils/workflow-utils.js
+++ b/integration-tests/utils/workflow-utils.js
@@ -22,13 +22,18 @@ const launchWorkflowAndWaitForSuccess = async page => {
 
   // Long enough for the submission details to be Running or Failed.
   // Workflows might sit in "Submitted" for a long time if there is a large backlog of workflows in the environment
-  const isRunning = await Promise.race([
-    findInGrid(page, 'Failed', { timeout: 5 * 60 * 1000 }).then(() => false),
-    findInGrid(page, 'Running', { timeout: 5 * 60 * 1000 }).then(() => true)
+  const workflowStatus = await Promise.race([
+    findInGrid(page, 'Succeeded', { timeout: 5 * 60 * 1000 }).then(() => 'Succeeded'),
+    findInGrid(page, 'Failed', { timeout: 5 * 60 * 1000 }).then(() => 'Failed'),
+    findInGrid(page, 'Running', { timeout: 5 * 60 * 1000 }).then(() => 'Running')
   ])
 
+  if (workflowStatus === 'Succeeded') {
+    return
+  }
+
   // If status is not Running, fails test now
-  if (!isRunning) {
+  if (workflowStatus === 'Failed') {
     throw new Error('Workflow has failed')
   }
 


aweng98 (approved):
Thank you.

integration-tests/utils/workflow-utils.js
```diff
@@ -22,13 +22,18 @@ const launchWorkflowAndWaitForSuccess = async page => {
 
   // Long enough for the submission details to be Running or Failed.
   // Workflows might sit in "Submitted" for a long time if there is a large backlog of workflows in the environment
-  const isRunning = await Promise.race([
-    findInGrid(page, 'Failed', { timeout: 5 * 60 * 1000 }).then(() => false),
-    findInGrid(page, 'Running', { timeout: 5 * 60 * 1000 }).then(() => true)
+  const workflowStatus = await Promise.race([
+    findInGrid(page, 'Succeeded', { timeout: 5 * 60 * 1000 }).then(() => 'Succeeded'),
+    findInGrid(page, 'Failed', { timeout: 5 * 60 * 1000 }).then(() => 'Failed'),
+    findInGrid(page, 'Running', { timeout: 5 * 60 * 1000 }).then(() => 'Running')
```
aednichols, line 28: I admittedly do not fully understand how this test works, but it seems like we could completely ignore `Running`?

Testing the outcome of `Succeeded` transitively verifies the implementation detail that it was `Running` at some point in the past, and tests are best when they are about outcomes.

```suggestion
  const workflowStatus = await Promise.race([
    findInGrid(page, 'Succeeded', { timeout: 5 * 60 * 1000 }).then(() => 'Succeeded'),
    findInGrid(page, 'Failed', { timeout: 5 * 60 * 1000 }).then(() => 'Failed'),
```

aednichols (approved):

integration-tests/utils/workflow-utils.js
nawatts: The way the test works now is that it waits up to 5 minutes for the workflow to start running (or fail).
https://github.com/DataBiosphere/terra-ui/blob/3ccf9fb37c47754b7bff174c71018950b57dd8fb/integration-tests/utils/workflow-utils.js#L23-L28

After the workflow starts running, the test waits up to 15 minutes for the workflow to succeed or fail.
https://github.com/DataBiosphere/terra-ui/blob/3ccf9fb37c47754b7bff174c71018950b57dd8fb/integration-tests/utils/workflow-utils.js#L35-L52

I don't know if that part of the test that asserts the workflow starts within 5 minutes is desirable or not.

aednichols: I guess what I'm saying is that if  the test has succeeded at the 15 minute mark, we don't care whether it was running at the 5 minute mark, because everything worked. But this is more of a thinking out loud exercise than an actual change request. Just trying to make things simpler and more resilient to variation.

aednichols (approved):

cahrens (approved):

integration-tests/utils/workflow-utils.js
nawatts: I'm going to go ahead and merge this as-is to attempt to resolve the current flakiness in tests. I agree with what you're saying though. Does anyone in the Batch or QA teams want to take on further updates to this test?

PR #3113, opened 2022-06-07
[QA-1903] Fix UI analysis-context-bar test

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1903

Clicking `'Jupyter Environment ( Creating )'` icon randomly stops working. After watching test playback in headful mode, I believe adding code to check page readiness should resolve this flaky issue for good.

### DIFF
--- a/integration-tests/tests/analysis-context-bar.js
+++ b/integration-tests/tests/analysis-context-bar.js
@@ -26,7 +26,9 @@ const testAnalysisContextBarFn = _.flow(
   await findElement(page, getAnimatedDrawer('Jupyter Cloud Environment'), { timeout: 40000 })
   await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Create' })) })
 
-  // For page synchronization, find the Start button
+  // We need a way to determine UI has finished loading after click Create. The START button is located in the main page.
+  // Loading of page could take few seconds or less so we find the START button and wait for visible.
+  // Click 'Jupyter Environment ( Creating )' icon could fail if page has not finished loading.
   await findElement(page, clickable({ textContains: 'Start' }), { visible: true })
 
   // Ensure UI displays the runtime is creating and the Terminal icon is present + enabled
--- a/integration-tests/tests/analysis-context-bar.js
+++ b/integration-tests/tests/analysis-context-bar.js
@@ -2,7 +2,7 @@
 const _ = require('lodash/fp')
 const { withRegisteredUser, withBilling, withWorkspace, performAnalysisTabSetup } = require('../utils/integration-helpers')
 const {
-  click, clickable, getAnimatedDrawer, findElement, noSpinnersAfter
+  click, clickable, getAnimatedDrawer, findElement, noSpinnersAfter, findTooltipText
 } = require('../utils/integration-utils')
 const { registerTest } = require('../utils/jest-utils')
 
@@ -15,6 +15,10 @@ const testAnalysisContextBarFn = _.flow(
   // Navigate to appropriate part of UI (the analysis tab)
   await performAnalysisTabSetup(page, token, testUrl, workspaceName)
 
+  // Ensure UI displays the runtime Terminal icon is present + disabled
+  let iconTooltipText = 'Terminal'
+  await findElement(page, clickable({ textContains: iconTooltipText, isEnabled: false }))
+
   // Create a runtime
   await click(page, clickable({ textContains: 'Environment Configuration' }))
   await findElement(page, getAnimatedDrawer('Cloud Environment Details'))
@@ -22,9 +26,16 @@ const testAnalysisContextBarFn = _.flow(
   await findElement(page, getAnimatedDrawer('Jupyter Cloud Environment'), { timeout: 40000 })
   await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Create' })) })
 
-  // Ensure UI displays the runtime is creating and the terminal icon is present + disabled
-  await findElement(page, clickable({ textContains: 'Terminal', isEnabled: false }))
-  await click(page, clickable({ textContains: 'Jupyter Environment ( Creating )' }), { timeout: 40000 })
+  // Ensure UI displays the runtime is creating and the Terminal icon is present + enabled
+  const terminalIcon = await findElement(page, clickable({ textContains: iconTooltipText, isEnabled: true }))
+  await terminalIcon.hover()
+  await findTooltipText(page, iconTooltipText)
+
+  iconTooltipText = 'Jupyter Environment ( Creating )'
+  const jupyterEnvIcon = await findElement(page, clickable({ textContains: iconTooltipText, isEnabled: true }))
+  await jupyterEnvIcon.hover()
+  await findTooltipText(page, iconTooltipText)
+  await click(page, clickable({ textContains: iconTooltipText }), { timeout: 40000 })
 
   // Updating/modifying the environment should be disabled when the env is creating
   await findElement(page, getAnimatedDrawer('Jupyter Environment Details'), { timeout: 40000 })
--- a/integration-tests/tests/analysis-context-bar.js
+++ b/integration-tests/tests/analysis-context-bar.js
@@ -2,7 +2,7 @@
 const _ = require('lodash/fp')
 const { withRegisteredUser, withBilling, withWorkspace, performAnalysisTabSetup } = require('../utils/integration-helpers')
 const {
-  click, clickable, getAnimatedDrawer, findElement, noSpinnersAfter, findTooltipText
+  click, clickable, getAnimatedDrawer, findElement, noSpinnersAfter
 } = require('../utils/integration-utils')
 const { registerTest } = require('../utils/jest-utils')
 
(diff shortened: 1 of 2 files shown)


xuscrbw (approved):

integration-tests/utils/integration-utils.js
```diff
@@ -111,6 +111,10 @@ const findText = (page, textContains, options) => {
   return page.waitForXPath(`//*[contains(normalize-space(.),"${textContains}")]`, options)
 }
 
+const findTooltipText = (page, textContains, options = { visible: true }) => {
+  return page.waitForXPath(`//*[@id="modal-root"]//*[@role="tooltip" and contains(normalize-space(.),"${textContains}")]`, options)
+}
```
kyuksel: See https://github.com/DataBiosphere/terra-ui/pull/3106#discussion_r891356438

integration-tests/tests/analysis-context-bar.js
```diff
@@ -15,16 +15,27 @@ const testAnalysisContextBarFn = _.flow(
   // Navigate to appropriate part of UI (the analysis tab)
   await performAnalysisTabSetup(page, token, testUrl, workspaceName)
 
+  // Ensure UI displays the runtime Terminal icon is present + disabled
+  let iconTooltipText = 'Terminal'
+  await findElement(page, clickable({ textContains: iconTooltipText, isEnabled: false }))
```
kyuksel: We try to avoid mutations to make it easier to reason about code (meaning, one wouldn't have to look around the code to see if a variable at hand might be changed somewhere else). In this case, the values are simple string literals, which could be inlined. That'd also save readers of the code having to scroll up/down to see values.

```suggestion
  // Ensure UI displays the runtime Terminal icon is present + disabled
  await findElement(page, clickable({ textContains: 'Terminal', isEnabled: false }))
```

Similarly for below.

If you feel that'd cause too much repetition of the string literals, you could define two `consts`, something like:

```javascript
const tooltipTextTerminal = 'Terminal'
```
```javascript
const tooltipTextEnvCreating = 'Jupyter Environment ( Creating )'
```

integration-tests/utils/integration-utils.js
aweng98: Yeah, I plan to rebase then update this code after that PR is merged.

integration-tests/tests/analysis-context-bar.js
aweng98: yeah, I want to avoid text duplication. I'll change to two `const`.

integration-tests/tests/analysis-context-bar.js
```diff
@@ -15,16 +15,26 @@ const testAnalysisContextBarFn = _.flow(
   // Navigate to appropriate part of UI (the analysis tab)
   await performAnalysisTabSetup(page, token, testUrl, workspaceName)
 
+  // Ensure UI displays the runtime Terminal icon is present + disabled
+  const tooltipTextTerminal = 'Terminal'
+  await findElement(page, clickable({ textContains: tooltipTextTerminal, isEnabled: false }))
```
kyuksel, line 20: Why is it necessary and how is it related to the [error on the QA ticket](https://broadworkbench.atlassian.net/browse/QA-1903) to "Ensure UI displays the runtime Terminal icon is present + disabled"?

integration-tests/tests/analysis-context-bar.js
```diff
@@ -15,16 +15,26 @@ const testAnalysisContextBarFn = _.flow(
...
+
   // Create a runtime
   await click(page, clickable({ textContains: 'Environment Configuration' }))
   await findElement(page, getAnimatedDrawer('Cloud Environment Details'))
   await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Settings' })) })
```
kyuksel, line 25: I know it's not your PR but would it be good to make the search expression more specific via

```suggestion
  await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Environment Settings' })) })
```

?

image

integration-tests/tests/analysis-context-bar.js
```diff
@@ -15,16 +15,26 @@ const testAnalysisContextBarFn = _.flow(
...
   await findElement(page, getAnimatedDrawer('Jupyter Cloud Environment'), { timeout: 40000 })
   await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Create' })) })
 
-  // Ensure UI displays the runtime is creating and the terminal icon is present + disabled
-  await findElement(page, clickable({ textContains: 'Terminal', isEnabled: false }))
-  await click(page, clickable({ textContains: 'Jupyter Environment ( Creating )' }), { timeout: 40000 })
+  // For page synchronization, find the Start button
+  await findElement(page, clickable({ textContains: 'Start' }), { visible: true })
```
kyuksel: What Start button is this? Could you send a screenshot of it?

integration-tests/utils/integration-utils.js
kyuksel: I see. 
It also doesn't seem to be used anywhere?

integration-tests/tests/analysis-context-bar.js
aweng98: Adding this to ensure UI correctness before Create runtime. Not at all related to the ticket but it improves the test. I can remove it if you disagree.

integration-tests/tests/analysis-context-bar.js
aweng98: Tried but it doesn't work. Because it's not one span element. 
Screen Shot 2022-06-07 at 6 13 25 PM

So now it raises a new concern. There are 3 buttons all have `Settings` text in the label. `page.waitForXpath` always returns the first element. What if three elements have shifted in display order? I'm going to leave this as-is for now but it is a real issue with the test.

integration-tests/utils/integration-utils.js
aweng98: I edited test so it no longer does hover then check tooltip text because hover seems to be unreliable (good thing that I caught this before merge). I left this func alone because I think it can be useful in future to others. Do you want it removed?

kyuksel: I'd suggest removing as a lot of what I had thought would be needed didn't end up being needed :)

integration-tests/tests/analysis-context-bar.js
kyuksel: Good point! Thinking how we could ensure the `Jupyter` button ends up being clicked on  🤔

integration-tests/tests/analysis-context-bar.js
kyuksel: By `UI correctness`, do you mean `UI readiness`? My assumption reading the change was that the change tries to ensure that

```javascript
```

doesn't happen too soon, by first checking for another element (`Terminal`) that shows up on the same page as `Environment Configuration` does. If my assumption is correct, I'd like to understand why checking for `Environment Configuration` isn't sufficient. 

Or was it that you just wanted to add another check to make sure `Terminal` is also there?

integration-tests/tests/analysis-context-bar.js
kyuksel: @breilly2 had the idea to use the `aria-label` which would [currently read 'Jupyter Environment'](https://github.com/DataBiosphere/terra-ui/blob/7355929aea1bd2401c7631d15413237df3d19520/src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js#L387). He also thought the source code should be changed to `Jupyter Environment Settings`, which I agree with. 

Cc: @cahrens @jdcanas @nawatts

integration-tests/tests/analysis-context-bar.js
kyuksel: Thank you.

Are you saying that without the line

```javascript
await findElement(page, clickable({ textContains: 'Start' }), { visible: true })
```
the line 
```javascript
await findElement(page, clickable({ textContains: tooltipTextTerminal, isEnabled: true }), { visible: true })
```

executes too soon causing failures?

integration-tests/tests/analysis-context-bar.js
aweng98: Test checks the Terminal is enabled at L33: `await findElement(page, clickable({ textContains: tooltipTextTerminal, isEnabled: true }), { visible: true })` after create runtime. What if the Terminal was already enabled before Create Runtime? Is that a bug or still correct in UI?

integration-tests/tests/analysis-context-bar.js
aweng98: Yes, it can. Watching the playback in headful mode, I noticed page (grey area) was blank while the sidebar icons already visible.

integration-tests/tests/analysis-context-bar.js
kyuksel: Gotcha  👍🏼

integration-tests/tests/analysis-context-bar.js
kyuksel: > page (grey area) was blank while the sidebar icons already visible

Why would that be a problem?

aweng98: because page load is not 100% done, click while loading in progress is a common cause for flaky.

kyuksel: Clarified offline. No action necessary.

integration-tests/tests/analysis-context-bar.js
kyuksel: @aweng98 If you could make the improvement mentioned above in this PR, great, if not, it'd be good to create an IA ticket and ping them on `#dsp-interactive-analysis`.

kyuksel (approved):
@aweng98 Thank you for patiently answering my questions. If you could make the [improvement mentioned on a thread](https://github.com/DataBiosphere/terra-ui/pull/3113/files#r892559242), great, if not, it'd be good to create an IA ticket and ping them on `#dsp-interactive-analysis`.

integration-tests/tests/analysis-context-bar.js
aweng98: I will make a follow-up PR for this change right after merge.

PR #3115, opened 2022-06-07
[IA-3483] Fix for location not being set correctly in WorkspaceContainer

wnojopra:
There is currently a bug where we're Unable to create cloud environments outside of US even when the workspace bucket is outside of US.

To reproduce:
1) Create/Load a workspace that has a bucket outside of the US (Montreal for example).
2) Attempt to create a cloud environment in this workspace. Click customize
3)  Notice that the dropdown to choose a location is stuck on US.
Expectation: Location dropdown should be set to Montreal, with the option to change to US. 

This happens because the `isGoogleWorkspace` prop doesn't get set properly when we first try to call `loadBucketLocation`. However, the googleProject is, so we can use that to determine that we are not an Azure workspace and can safely attempt to load the bucket location.

### DIFF
--- a/src/pages/workspaces/workspace/WorkspaceContainer.js
+++ b/src/pages/workspaces/workspace/WorkspaceContainer.js
@@ -310,7 +310,7 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
     }
 
     const loadBucketLocation = async (googleProject, bucketName) => {
-      if (isGoogleWorkspace) {
+      if (!!googleProject) {
         const bucketLocation = await Ajax(signal).Workspaces.workspace(namespace, name).checkBucketLocation(googleProject, bucketName)
         setBucketLocation(bucketLocation)
       }


nawatts (approved):

nawatts:
Nice, this resolves [this comment](https://github.com/DataBiosphere/terra-ui/pull/3037/files#r874812211) on #3037.

cahrens (approved):
Thanks @wnojopra -- I think I introduced this bug.

At some point we should think about what automated test coverage we could add to cover some of these workflows.

wnojopra:
Thanks @cahrens and @nawatts !

When this feature was first being developed, I suggested that we add an integration test similar to the analysis integration test. It would create a workspace, launch a cloud environment, and run a notebook, but just all in a non-US region. It would be straightforward to implement. Unfortunately at the time, tests were very flaky and we didn't want to add another flaky test. It was decided that manual testing would be sufficient. 

But this was a bug that could have been caught easily with that integration test. Is now a good time to revisit that decision?

nawatts:
I think we're generally more confident in the stability of the terra-ui test suite now, though there are a few that are still known to be flaky or have other issues (as you're probably seeing with google-workspace, azure-workspace, and import-cohort-data tests on this PR). If you're motivated to write a test for this, I think we would welcome it.

cahrens:
Yes, I agree that we are now more open to adding new tests, as long as you are willing to keep on top of it and help address issues that arise. Note that you can mock out specific AJAX calls if it makes your test easier to write/maintain.

PR #3116, opened 2022-06-08
[QA-1905] Change button text in runtime manager error notifications

nawatts:
When a cloud environment fails to start in integration tests, we would like to get more information about the error. [Slack thread for context](https://broadinstitute.slack.com/archives/C01EHNUM73R/p1654625184893469).

When a terra-ui integration test fails, it will take a screenshot of the page.

https://github.com/DataBiosphere/terra-ui/blob/a8969e58d20dff8652e955d96c6d27d1e8b9babf/integration-tests/jest-circus-environment.js#L16-L18

If there are any error notifications present, it will also expand those notifications and take a screenshot of the details modal.

https://github.com/DataBiosphere/terra-ui/blob/a8969e58d20dff8652e955d96c6d27d1e8b9babf/integration-tests/utils/integration-utils.js#L300-L304

To expand the notification, the tests look for a button labeled "Details".

https://github.com/DataBiosphere/terra-ui/blob/a8969e58d20dff8652e955d96c6d27d1e8b9babf/integration-tests/utils/integration-utils.js#L272-L281

This is based on the notifications shown by the shared `reportError` helper.

https://github.com/DataBiosphere/terra-ui/blob/a8969e58d20dff8652e955d96c6d27d1e8b9babf/src/libs/error.js#L11

https://github.com/DataBiosphere/terra-ui/blob/a8969e58d20dff8652e955d96c6d27d1e8b9babf/src/libs/notifications.js#L91-L94

However, the runtime manager's error notifications do not use this helper. They show a notification that behaves similarly: it has a button that opens a modal with details about the error. But the button is labeled "SEE LOG INFO". The tests don't recognize this button, so we don't get a screenshot of the error details.

This changes the button label to "Details" to make these error notifications work with the integration tests. Though this is a test issue, I think it's reasonable to make this change in the UI since it makes these error notifications consistent with others in the app.

### DIFF
--- a/src/components/RuntimeManager.js
+++ b/src/components/RuntimeManager.js
@@ -112,7 +112,7 @@ const RuntimeErrorNotification = ({ runtime }) => {
         textDecoration: 'underline',
         fontWeight: 'bold'
       }
-    }, ['SEE LOG INFO']),
+    }, ['Details']),
     modalOpen && h(RuntimeErrorModal, {
       runtime,
       onDismiss: () => setModalOpen(false)
@@ -157,7 +157,7 @@ const AppErrorNotification = ({ app }) => {
         textDecoration: 'underline',
         fontWeight: 'bold'
       }
-    }, ['SEE LOG INFO']),
+    }, ['Details']),
     modalOpen && h(AppErrorModal, {
       app,
       onDismiss: () => setModalOpen(false)


Qi77Qi (approved):

Qi77Qi (approved):
thanks @nawatts !

aweng98 (approved):

src/components/RuntimeManager.js
```diff
@@ -112,7 +112,7 @@ const RuntimeErrorNotification = ({ runtime }) => {
         textDecoration: 'underline',
         fontWeight: 'bold'
       }
-    }, ['SEE LOG INFO']),
+    }, ['Details']),
```
kyuksel, line 115: I wonder if we should define a `const` for the string literal `Details` and reference it from this file, the helper and the test so if the string changes in the future, we can easily make the code update.

kyuksel (approved):
Thanks for the fix! One thought inline. Up to you to take an action or not. Feel free to merge without another approval.

src/components/RuntimeManager.js
nawatts: I'm not convinced of the benefit... either way, for these to work with the integration tests' screenshots, a dev has to realize that there's a specific label to use. I'm skeptical that having the label in a constant makes it any easier to discover that fact. Also, non-standard error notifications that have a details button like this one are rare (so far as I know, these are the only ones and they're intended to be removed), so I don't think we need to go out of our way to accommodate them.

Also, I'm not sure if we can share code between UI and tests because of the different module systems (ES modules in the UI, CommonJS in the tests).

kyuksel: All good reasons to not make the change 👍🏼 Thank you.

PR #3117, opened 2022-06-08
[TOAZ-15] [TUIM-1] Integrate new OAuth library to Terra UI to support B2C

rtitle:
### JIRAs:
* https://broadworkbench.atlassian.net/browse/TOAZ-15
* https://broadworkbench.atlassian.net/browse/TUIM-1

### Description
This is an un-revert of [the original PR](https://github.com/DataBiosphere/terra-ui/pull/2949) to swap out Terra UI's oauth library to support Google and B2C. The original change was reverted because it was causing session timeouts to happen too frequently. The reason this started happening is the `gapi` client library is able to keep access tokens fresh indefinitely through some magic -- they render an iframe on a Google server and are able to store/reload previous auth responses to renew expired access tokens. The new library wasn't doing that, so access tokens would expire if the browser wasn't able to renew them in time (e.g. if the computer was asleep). 

The OAuth solution for this is refresh tokens. We now request an access token _and_ a refresh token from Google (or B2C). The refresh token can be used to renew an expired Google access token. Terra UI `signOut` clears the access token and refresh token from the browser, so we still adhere to the session timeout requirement for sensitive users. TDR has implemented a similar change here: https://github.com/DataBiosphere/jade-data-repo-ui/pull/1132. @dinvlad is in the loop on this from an AppSec/compliance perspective (Denis feel free to correct anything here).

### Commits:
* The [first commit](https://github.com/DataBiosphere/terra-ui/pull/3117/commits/5649d1615034a7cc65c633f814141b9ffb441120) is a pure un-revert of the original change.
* The [second commit](https://github.com/DataBiosphere/terra-ui/pull/3117/commits/3dbd5c92b26b6d0126d1e9974b621a806433597f) adds functionality to request and use a refresh token

### Testing notes:
1. Test on PR site if possible. While developing I observed different behavior on the PR site vs localhost w.r.t. refresh tokens.
2. Verify user login and access to all Terra UI screens.
3. Verify the billing page in particular -- it should automatically elevate scopes and display billing accounts if the user has previously granted `cloud-billing` to the application; otherwise it should display the Google prompt to elevate scopes (which is actually blocked on non-prod environments).
4. Verify session timeout. This is kind of annoying to test. I've tried leaving Terra UI open all day while I was working; and also all night while my laptop was asleep. In both cases I was still signed in when I came back, which is the same behavior as Terra prod today.

Thank you!


dinvlad:
Does logout just remove the refresh token from browser, or also invalidates it? I think it'd be ideal to trigger the latter also, if possible.

src/libs/auth.js
```diff
@@ -21,26 +21,29 @@ import * as Utils from 'src/libs/utils'
...
+    metadata,
+    prompt: 'consent',
     scope: 'openid email profile',
-    metadata: {
-      authorization_endpoint: `${getConfig().orchestrationUrlRoot}/oauth2/authorize`,
-      token_endpoint: `${getConfig().orchestrationUrlRoot}/oauth2/token`
-    },
+    loadUserInfo: isGoogleAuthority(),
     userStore: new WebStorageStateStore({ store: window.localStorage }),
-    accessTokenExpiringNotificationTimeInSeconds: 300,
```
cahrens: @rtitle can you say more about the removal of `accessTokenExpiringNotificationTimeInSeconds`? I'm also having trouble understanding where the refresh token is requested.

rtitle:
> Does logout just remove the refresh token from browser, or also invalidates it? I think it'd be ideal to trigger the latter also, if possible.

@dinvlad that's a good catch. I think currently it just removes it from the browser. I just did a test:
1. Logged in, copied the refresh token from browser storage
2. I was able to exchange the refresh token for an access token via curl to `https://www.googleapis.com/oauth2/v4/token` (note I did need to pass the `client_secret` to Google, which we do not store in the browser)
3. Logged out of Terra UI
4. I could still exchange the refresh token for an access token.

I can look into if we can explicitly revoke the refresh token upon logout.

EDIT: seems like it's possible to revoke a refresh token in Google/B2C, so I'll add that as a logout hook 🚀

src/libs/auth.js
rtitle: Yes. We actually use 2 callbacks:
1. `accessTokenExpiring` -- this fires shortly _before_ the access token expires. In this case we silently renew the access token without needing the refresh token. This is not always reliable (e.g. it can be missed if the browser is asleep).
2. `accessTokenExpired` -- this fires _after_ the access token expires. In this case the refresh_token is used to get a new access_token.

I could hear an argument that (1) isn't necessary if we're just using a refresh token anyway.

In any case this parameter only affects case (1). I had previously increased the time-before-expire to 5 minutes in the hope that it would be more reliable. Now it's using the default of 2 minutes I believe.

rtitle: Also the refresh token is requested via:
```
extraQueryParams: { access_type: 'offline' }
```

cahrens: Thanks, but I don't actually see `accessTokenExpiring` anyplace in the code. Can you point me to where it is?

src/libs/auth.js
```diff
@@ -19,29 +20,87 @@ import {
...
+    token_endpoint: `${getConfig().orchestrationUrlRoot}/oauth2/token`,
+    ...(isGoogleAuthority() && { userinfo_endpoint: 'https://openidconnect.googleapis.com/v1/userinfo' })
+  }
+  return {
+    authority: `${getConfig().orchestrationUrlRoot}/oauth2/authorize`,
+    client_id: authStore.get().oidcConfig.clientId,
+    popup_redirect_uri: `${window.origin}/redirect-from-oauth`,
+    silent_redirect_uri: `${window.origin}/redirect-from-oauth-silent`,
+    metadata,
+    prompt: 'consent',
```
cahrens: What is the difference between 'consent' and 'login'?

src/libs/auth.js
rtitle: It comes from `automaticSilentRenew: true`: https://github.com/DataBiosphere/terra-ui/pull/3117/files#diff-a4ba38b20453374fed28d8c26c28f5b20b486db965bbfd9c47032a9cb99c122cR39. The library sets the accessTokenExpiring callback if that param is enabled [[source code](https://github.com/authts/oidc-client-ts/blob/main/src/SilentRenewService.ts#L23)].

Previously I experimented with setting my own `accessTokenExpiring` callback vs letting the library do it, but the library version seems to work fairly well (it has some retries in case of errors, etc).

cahrens: Got it, thank you!

src/libs/auth.js
rtitle: `prompt: 'login'` forces the Google login popup (e.g. it doesn't use the currently signed in user) but it wasn't giving me a refresh token on Google. `consent` seems to do both.

Google docs: https://developers.google.com/identity/protocols/oauth2/openid-connect#authenticationuriparameters

B2C docs:
https://docs.microsoft.com/en-us/azure/active-directory-b2c/authorization-code-flow

Now that I'm reading this I wonder if it should be `consent` for Google and `login` for B2C. I'll do some more testing with B2C enabled in general.

(The subtle differences between Google and B2C have been really annoying.)

rtitle: cc @nmalfroy as well if you have any insights

dinvlad: Indeed, seems like for B2C
> Currently, the only valid value is `login`

nmalfroy: I left it as consent in B2C and it seems to behave ok.  I'm actually not sure that it makes much of a difference for B2C since we don't ask for sensitive scopes in B2C

src/libs/auth.js
```diff
@@ -19,29 +20,93 @@ import {
...
 }
 
 export const signOut = () => {
   // Note: Ideally, we'd actually clear the Leo proxy cookie here, but there's not currently an endpoint to do that.
   // When IA-2236 is done, add that call here.
   cookieReadyStore.reset()
   sessionStorage.clear()
-  getAuthInstance().signOut()
+  const auth = getAuthInstance()
+  auth.revokeTokens()
```
rtitle: I confirmed the refresh and access token are successfully revoked when I sign out.

dinvlad: Perfect!

rtitle:
@cahrens (or @breilly2) just a friendly nudge for any more feedback here. I did some more testing against a BEE with Google and B2C enabled, and things are looking good I think. Thank you

src/libs/auth.js
```diff
@@ -19,29 +20,94 @@ import {
...
+    .catch(() => false)
+    .then(() => auth.removeUser())
+    .then(() => auth.clearStaleState())
```
kyuksel, line 66: Would it be desirable to call `clearStaleState()` if `removeUser()` fails?

```suggestion
    .finally(() => auth.clearStaleState())
```

kyuksel: Would it be desirable to call the subsequent functions when a previous fails, like so?

```suggestion
    .finally(() => auth.removeUser())
    .finally(() => auth.clearStaleState())
```

rtitle: `finally` does seem nicer, and works as intended. Will change.

kyuksel: I also wonder if we need to do any reporting/alerting in case one or more of those three calls fail, for security purposes.

Cc: @dinvlad

rtitle: `revokeTokens` calls the token revoke endpoint and _can_ fail if the token is already revoked. So I think I'm going to leave the `catch` there to recover errors. `removeUser` and `clearStaleState` just modifies browser state, so makes sense to do them in finally (we want them to always occur).

kyuksel: I see. 

> revokeTokens calls the token revoke endpoint and can fail if the token is already revoked.

Would it make sense / be possible to recover from that specific error and not the others in case we'd not want to swallow other error modes? I don't know how fussy we want to be about this security-wise 😊

nmalfroy: I've noticed in the wild that it's actually expected that it will fail (there's a race condition in there).  In the case of google it revokes 2 tokens: the refresh_token and the access_token.  If you revoke one, it seems to revoke the other but there's some time t where that hasn't happened yet (< a second it seems)

rtitle: Makes sense, just pushed a change to only recover from `invalid_token` errors. :)

rtitle:
Hm integration tests started failing where they weren't before. Not sure if it's flakiness or a real issue. Will investigate…

EDIT: I think it was Sam issues in dev… other PRs were failing too, they just passed on re-run.

src/components/common.js
```diff
@@ -523,18 +523,25 @@ export const HeroWrapper = ({ showMenu = true, bigSubhead = false, children }) =
...
-        `${getAppName(true)} is a ${Utils.cond(
+        `${getAppName(!isTerra())} is a ${Utils.cond(
           [isTerra(), () => 'cloud-native platform'],
           [isFirecloud(), () => 'NCI Cloud Resource project powered by Terra'],
           [isProjectSingular(), () => 'project funded by Additional Ventures and powered by Terra'],
           () => 'project powered by Terra'
         )} for biomedical researchers to `,
         heavyWrapper('access data'), ', ', heavyWrapper('run analysis tools'), ', ',
-        span({ style: { whiteSpace: 'nowrap' } }, ['and', strong([' collaborate'])]),
-        isProjectSingular() && ' to advance research around single ventricle heart disease',
```
cahrens, line 536: It looks like an incorrect merge resolution here-- this removes some code that @kyuksel recently added.

rtitle: Oops - thanks, good catch. Will fix.

rtitle: Done

src/components/common.js
```diff
@@ -534,7 +537,13 @@ export const HeroWrapper = ({ showMenu = true, bigSubhead = false, children }) =
         heavyWrapper('access data'), ', ', heavyWrapper('run analysis tools'), ', ',
         span({ style: { whiteSpace: 'nowrap' } }, ['and', strong([' collaborate'])]),
```
cahrens: @kyuksel the usage of `strong` here instead of `heavyWrapper` is causing this styling issue:

image

cahrens (approved):
I've looked through the commits after the initial revert commit, and then skimmed through the entire PR for merge conflict issues. I did some minimal testing with the billing page but did not test the expiration/refresh tokens (it seems like a couple of folks have tested that pretty extensively).

src/pages/workspaces/workspace/DashboardPublic.js
```diff
@@ -57,7 +57,7 @@ const DashboardPublic = ({ namespace, name }) => {
       ]),
       div({ style: Style.dashboard.rightBox }, [
         div({ style: signInStyle }, [
-          div(['Sign in to view full workspace']),
+          div(['Log in to view full workspace']),
```
breilly2, line 60: Why the change in terminology? I think we use "Sign in" everywhere, and I'd expect to have to change this is many more places to keep it all consistent.

src/components/common.js
kyuksel: Thanks for the catch!

@rtitle would you be able to revert the change in your PR or shall I make a separate PR?

src/pages/workspaces/workspace/DashboardPublic.js
cahrens: I think the new standardized terminology per Lou is "Log in". And it was changed on the button text to be consistent.

src/components/common.js
cahrens: @kyuksel Rob didn't introduce this issue. You can see it in dev. The issue is that in one code path (the one we were testing apparently) `heavyWrapper` makes things bold/strong, but in the other it does not. I would say we should simplify our lives and just always go with bold. Or you need to consistently call `heavyWrapper` (I think you were trying to avoid wrapping CSS).

https://bvdp-saturn-dev.appspot.com/#workspaces

image

rtitle: I can change it back to `heavyWrapper`. (I thought **collaborate** was supposed to be emphasized 😂 )

src/pages/workspaces/workspace/DashboardPublic.js
rtitle: Yeah `Log in` came from Lou. (Trying to find the wireframes.. I can't access invision anymore for some reason.) I updated the button on the dashboard and in the hamburger menu.

src/components/common.js
kyuksel: I didn't mean to imply Rob introduced the issue. Reverting my change would be a small one and this PR seems close to being merged so I thought it could piggy-back on this PR but I can make a separate one if it'll complicate things.

> I would say we should simplify our lives and just always go with bold.

I don't feel comfortable changing it without knowing why it was `bold`ed in one case and not the other in the first place.

> I think you were trying to avoid wrapping CSS

Wrapping would be okay. I mainly thought it was unnecessary but missed the `bigSubHead = false` case.

rtitle: Done

kyuksel:
```suggestion
        span({ style: { whiteSpace: 'nowrap' } }, ['and', heavyWrapper(' collaborate')]),
```

kyuksel: Thank you!

rtitle: Done here: https://github.com/DataBiosphere/terra-ui/pull/3117/commits/4a91462ec9e19709399768abbf617161567d7907

src/components/TopBar.js
```diff
@@ -187,16 +186,14 @@ const TopBar = ({ showMenu = true, title, href, children }) => {
...
-                }, ['SIGN IN']) :
-                h(SignInButton)
+              h(Clickable, {
+                ...(isDatastage() || isBioDataCatalyst()) ? { href: Nav.getLink('workspaces') } : { onClick: signIn },
+                style: {
+                  backgroundColor: 'white', fontSize: 18, fontWeight: 500, color: colors.accent(),
+                  borderRadius: 5, boxShadow: '0 2px 4px 0 rgba(0,0,0,.25)',
+                  width: 250, height: 56, display: 'flex', alignItems: 'center', justifyContent: 'center'
+                }
+              }, ['LOG IN'])
```
nmalfroy, line 196: Ah wait, are we using LOG IN?  I just changed it to SIGN IN in TDR 🤦 I'll change it back

src/components/common.js
```diff
@@ -523,18 +523,27 @@ export const HeroWrapper = ({ showMenu = true, bigSubhead = false, children }) =
...
       }
     }, [
-      h1({ style: { fontSize: 54 } }, `Welcome to ${getAppName()}`),
+      h1({ style: { fontSize: 54 } }, [
+        'Welcome to ',
+        span({ style: { display: isTerra() ? 'block' : 'inline-block' } }, getAppName(isTerra()))
+      ]),
       div({ style: { margin: '1rem 0', width: 575, ...(bigSubhead ? { fontSize: 20, lineHeight: '28px' } : { fontSize: 16, lineHeight: 1.5 }) } }, [
-        `${getAppName(true)} is a ${Utils.cond(
+        `${getAppName(!isTerra())} is a ${Utils.cond(
```
nmalfroy, line 531: Not totally sure where this gets used but is it expected that it's `!isTerra()` here but `isTerra()` on line 528?

src/libs/auth.js
```diff
@@ -19,29 +20,110 @@ import {
...
+    silent_redirect_uri: `${window.origin}/redirect-from-oauth-silent`,
+    metadata,
+    prompt: 'consent login',
+    scope: 'openid email profile',
+    loadUserInfo: isGoogleAuthority(),
+    userStore: new WebStorageStateStore({ store: window.localStorage }),
+    automaticSilentRenew: true,
+    includeIdTokenInSilentRenew: true,
+    extraQueryParams: { access_type: 'offline' },
+    revokeTokenTypes: ['access_token', 'refresh_token']
```
nmalfroy: FYI: you don't need this line since those are the defaults

nmalfroy (approved):
Looks good!  I had a quick question about boolean logic and there's an extraneous option

src/components/TopBar.js
rtitle: Using 'Log in' per Lou 🤷

src/libs/auth.js
rtitle: Ah cool 👍

src/components/common.js
rtitle: It's confusing :/ but expected. Basically there's a longName and a shortName for the site. For Terra we want:
```
Welcome to Terra Community Workbench! [long name]

Terra [short name] is a…
```

For AnVIL (and other sites) it's reverse:
```
Welcome to AnVIL! [short name]

The NHGRI AnVIL (Genomic Data Science Analysis, Visualization, and Informatics Lab-space) [long name] is a …
```

src/libs/auth.js
rtitle: Confirmed revoke works w/o this line, removed.

src/components/TopBar.js
cahrens: And NOT "LOGIN", which is a noun.

PR #3119, opened 2022-06-08
[QA-1909] UI test analysis-context-bar improvement

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1909

Follow-up PR to address another issue in analysis-context-bar test.

From PR [DataBiosphere/terra-ui/pull/3113](https://github.com/DataBiosphere/terra-ui/pull/3113) [comments](https://github.com/DataBiosphere/terra-ui/pull/3113/files#r891747388).

### DIFF
--- a/integration-tests/utils/analysis-utils.js
+++ b/integration-tests/utils/analysis-utils.js
@@ -1,0 +1,7 @@
+const findButtonInCloudEnvDetailsDialog = (page, ariaLabelButtonText) => {
+  return page.waitForXPath(`//*[@role="dialog" and @aria-hidden="false"]//*[@role="button" and contains(@aria-label,"${ariaLabelButtonText}")]`,
+    { visible: true }
+  )
+}
+
+module.exports = { findButtonInCloudEnvDetailsDialog }
--- a/integration-tests/utils/analysis-utils.js
+++ b/integration-tests/utils/analysis-utils.js
@@ -1,7 +1,0 @@
-const findButtonInCloudEnvDetailsDialog = (page, ariaLabelButtonText) => {
-  return page.waitForXPath(`//*[@role="dialog" and @aria-hidden="false"]//*[@role="button" and contains(@aria-label,"${ariaLabelButtonText}")]`,
-    { visible: true }
-  )
-}
-
-module.exports = { findButtonInCloudEnvDetailsDialog }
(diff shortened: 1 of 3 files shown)


cahrens (approved):

integration-tests/utils/analysis-utils.js
```diff
@@ -0,0 +1,7 @@
+const findButtonInCloudEnvDetailsDialog = (page, ariaLabelButtonText) => {
+  return page.waitForXPath(`//*[@role="dialog" and @aria-hidden="false"]//*[@role="button" and contains(@aria-label,"${ariaLabelButtonText}")]`,
+    { visible: true }
+  )
+}
```
kyuksel: 1. Does this function have anything specific to CloudEnvironmentModal? If not, I'd suggest moving it to `integration-utils.js`?
2. I also see that `getAnimatedDrawer` in `integration-utils.js` does a similar search and there doesn't seem to be anything specific in that function about an animated drawer so I wonder if we could refactor `getAnimatedDrawer` to work for its current use cases as well as the new case you're adding  🤔

aweng98: 1) Yes, function is for Cloud Env Details modal only. I thought about putting it in `integration-utils` before, but I think it didn't make any sense.
2) I'd like to keep PR scope small if possible. Someone else from IA team or me in future can make additional improvement such as refactoring `getAnimatedDrawer`. What do you think?

kyuksel: 1. What about it is specific to `CloudEnvironmentModal`?
2. Okay.

aweng98: Hmm, it could be useful for other tests but I don't know any right now. 
Can you tell me:
1) What other tests beside analysis tests can use this finder function?
2) How many other UI elements (in other UI pages) where you should find elements by `aria-label` (not by visible text) in a dialog?

I'm happy to move it to `integration-utils` file if it will be needed by non-analysis tests. But we shouldn't do it if we thought it would be needed but didn't end up being needed at all. 🙂

kyuksel: 1. I think I hadn't asked my question clearly. Can we rename the helper as something like `findButtonByAriaLabel` so it's clear that it could be used in _any_ modal as long as the purpose is matching some aria-label text?
2. I agree with you about not prematurely factoring code out. For the same reason, the helper could be placed in the only file it's used from (`analysis-context-bar.js`) but I don't feel strongly.

aweng98: Rename function to `findButtonInDialogByAriaLabel` and moved it to `integration-utils.js`.

kyuksel (approved):

aweng98:
Thank you. Sorry for my misunderstanding.

PR #3120, opened 2022-06-09
[QA-1912] Update app.yaml to re-group cached files

aweng98:
Set appropriate `expiration` for handlers.

Slack conversation about problem with [Loading Terra... spinner](https://broadinstitute.slack.com/archives/C01EHNUM73R/p1654538923441459)

https://broadworkbench.atlassian.net/browse/QA-1912

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -199,13 +199,14 @@ jobs:
           paths:
             - .yarn/unplugged
             - node_modules/.cache
-      - run: tar -czf build.tgz .gcloudignore build config
+      - run: tar -czf build.tgz .gcloudignore app.yaml build config
       - store_artifacts:
           path: build.tgz
       - persist_to_workspace:
           root: .
           paths:
             - .gcloudignore
+            - app.yaml
             - build
             - config
   deploy-pr:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,15 +93,6 @@ commands:
                   fi
       - attach_workspace:
           at: .
-      - when:
-          condition:
-            or:
-              - equal: [ "alpha", << parameters.env >> ]
-              - equal: [ "staging", << parameters.env >> ]
-          steps:
-            - run:
-                command: |
-                  sed -e 's/default_expiration: ".*"/default_expiration: "0s"/' -i.orig app.yaml
       - run: cp config/<< parameters.env >>.json build/config.json
       - run: echo "$<< parameters.sa_key_var >>" > /tmp/sa.json
       - run: gcloud auth activate-service-account --key-file=/tmp/sa.json
@@ -208,14 +199,13 @@ jobs:
           paths:
             - .yarn/unplugged
             - node_modules/.cache
-      - run: tar -czf build.tgz .gcloudignore app.yaml build config
+      - run: tar -czf build.tgz .gcloudignore build config
       - store_artifacts:
           path: build.tgz
       - persist_to_workspace:
           root: .
           paths:
             - .gcloudignore
-            - app.yaml
             - build
             - config
   deploy-pr:
--- a/app.yaml
+++ b/app.yaml
@@ -2,16 +2,100 @@ runtime: nodejs16
 default_expiration: "2m"
 
 handlers:
-- url: /(.*\.(css|eot|gz|html|ico|jpg|jpeg|js|json|map|png|svg|otf|ttf|woff|woff2))
+
+# font format
+- url: /(.*\.(woff|woff2))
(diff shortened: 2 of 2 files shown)


app.yaml
```diff
@@ -2,15 +2,109 @@ runtime: nodejs16
...
-- url: /(.*\.(css|eot|gz|html|ico|jpg|jpeg|js|json|map|png|svg|otf|ttf|woff|woff2))
+
+# font format
+- url: /(.*\.(woff|woff2))
+  mime_type: application/x-font-woff
+  static_files: build/\1
+  upload: build/(.*\.(woff|woff2))
+  secure: always
+  http_headers:
+    X-Content-Type-Options: "nosniff"
```
aweng98: Had to copy/paste two keys (`secure` and `http_headers`) for all `url`. I wish there is an easy way where just set it once.

nawatts: I don't think I'd recommend using them here, but YAML does support reusing blocks of code with [anchors and aliases](https://spec.yaml.io/main/spec/1.3.0/#anchors-and-aliases).

app.yaml
```diff
@@ -2,15 +2,109 @@ runtime: nodejs16
 default_expiration: "2m"
 
 handlers:
-- url: /(.*\.(css|eot|gz|html|ico|jpg|jpeg|js|json|map|png|svg|otf|ttf|woff|woff2))
+
+# font format
+- url: /(.*\.(woff|woff2))
+  mime_type: application/x-font-woff
```
nawatts: Why do we need to manually set MIME types now? We weren't previously.

app.yaml
```diff
@@ -2,15 +2,109 @@ runtime: nodejs16
...
+
+- url: /(.*\.js)
+  mime_type: text/javascript
+  static_files: build/\1
+  upload: build/(.*\.js)
+  secure: always
+  http_headers:
+    X-Content-Type-Options: "nosniff"
+
+- url: /(.*\.(json|map))
```
nawatts: These two extensions probably should not be treated the same. Sourcemaps can be cached like their associated JS files, while we probably want a shorter cache time for JSON files such as app configuration.

nawatts (changes requested):
I'm not sure that it makes sense to group files by extension. I suspect most of the files in the `static` directory (JS, CSS, images, fonts, etc) can be treated the same.

And there's at least one JS file (`jupyter-iframe-extension.js`) where we probably want different cache settings than JS outputs from webpack.

.circleci/config.yml
```diff
@@ -208,14 +199,13 @@ jobs:
           paths:
             - .yarn/unplugged
             - node_modules/.cache
-      - run: tar -czf build.tgz .gcloudignore app.yaml build config
```
nawatts: Removing `app.yaml` here and in the `persist_to_workspace` step below is causing the `deploy-pr` job to fail: https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/10080/workflows/b5877506-90bf-4017-ba0e-2ef3f1d6d212/jobs/44680

`app.yaml` here and in the `persist_to_workspace` step below were not part of #3004. `app.yaml` has been in these lists for years (181bdb27dbc50ddb951859feb9682a3d559e8322).

nawatts (changes requested):

.circleci/config.yml
aweng98: oh, that's not good. fixed now.

nawatts (approved):

app.yaml
```diff
@@ -1,16 +1,19 @@
 runtime: nodejs16
-default_expiration: "2m"
+default_expiration: 2m
 
 handlers:
-- url: /(.*\.(css|eot|gz|html|ico|jpg|jpeg|js|json|map|png|svg|otf|ttf|woff|woff2))
```
nawatts, line 5: It turns out that we do have other HTML files (`google*.html` in the `public` directory). Removing `html` here would prevent those files from being served.

https://github.com/DataBiosphere/terra-ui/tree/48dc8b3e695407c96b352c8c56a9a65d7729df95/public

I think the best way to handle this would be to add a separate handler for HTML files.

nawatts (changes requested):

app.yaml
aweng98: Nice catch! Added `.html` hander now.

dmohs (approved):

nawatts (approved):

nawatts:
@aweng98 this needs a JIRA ticket.

aweng98:
Thank you for the reminder. I will create one Jira ticket today.

PR #3122, opened 2022-06-09
[WOR-192] Change default GCP storage location to us-central1

aherbst-broad:
## Context
[Relevant ticket](https://broadworkbench.atlassian.net/browse/WOR-192)
Google is changing it's pricing policy; movement between multiregion and "regular" regions will no longer be free as of 2022-10-01. 

## This PR
* Changes the default location to `us-central1` for new workspaces
* Adds a warning to the create workspace modal that will notify users of the impending pricing change if they choose the `US` multiregion
* Fixes an issue with the notebooks ComputeModal where `unknown` would display for the location

### DIFF
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -6,7 +6,6 @@ const {
 } = require('./integration-utils')
 const { fetchLyle } = require('./lyle-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
-const { defaultLocation } = require('../../src/libs/runtime-utils')
 
 
 const defaultTimeout = 5 * 60 * 1000
@@ -34,7 +33,7 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
     const response = await page.evaluate(async (name, billingProject) => {
       try {
         // explicitly setting us-central1 in this call until rawls defaults to us-central1 as well
-        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: defaultLocation })
+        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: 'us-central1' })
       } catch (err) {
         console.error(err)
         console.error(typeof err)
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -6,7 +6,6 @@ const {
 } = require('./integration-utils')
 const { fetchLyle } = require('./lyle-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
-const { defaultLocation } = require('src/libs/runtime-utils')
 
 
 const defaultTimeout = 5 * 60 * 1000
@@ -33,7 +32,7 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
   try {
     const response = await page.evaluate(async (name, billingProject) => {
       try {
-        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: defaultLocation })
+        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: 'us-central1' })
       } catch (err) {
         console.error(err)
         console.error(typeof err)
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -32,7 +32,7 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
   try {
     const response = await page.evaluate(async (name, billingProject) => {
       try {
-        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: 'us-central1' })
+        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {} })
       } catch (err) {
         console.error(err)
         console.error(typeof err)
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -32,6 +32,7 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
   try {
     const response = await page.evaluate(async (name, billingProject) => {
       try {
+        // explicitly setting us-central1 in this call until rawls defaults to us-central1 as well
         return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: 'us-central1' })
       } catch (err) {
         console.error(err)
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -32,7 +32,7 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
   try {
     const response = await page.evaluate(async (name, billingProject) => {
       try {
-        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {} })
+        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: 'us-central1' })
       } catch (err) {
         console.error(err)
         console.error(typeof err)
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -6,6 +6,7 @@ const {
 } = require('./integration-utils')
 const { fetchLyle } = require('./lyle-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
+const { defaultLocation } = require('../../src/libs/runtime-utils')
 
 
 const defaultTimeout = 5 * 60 * 1000
@@ -33,7 +34,7 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
     const response = await page.evaluate(async (name, billingProject) => {
       try {
         // explicitly setting us-central1 in this call until rawls defaults to us-central1 as well
-        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: 'us-central1' })
+        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: defaultLocation })
       } catch (err) {
         console.error(err)
         console.error(typeof err)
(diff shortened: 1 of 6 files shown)


src/components/NewWorkspaceModal.js
```diff
@@ -249,6 +249,14 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
           options: _.sortBy('label', isAlphaRegionalityUser ? allRegions : availableBucketRegions)
         })
       ])]),
+      isLocationMultiRegion(bucketLocation) && div({ style: { ...warningStyle } }, [
+        icon('warning-standard', { size: 24, style: { color: colors.warning(), flex: 'none', marginRight: '0.5rem' } }),
+        div({ style: { flex: 1 } }, [
+          `Effective October 1, 2022, Google Cloud will charge egress fees on data stored in multi-region storage buckets.`,
+          p(`Choosing a multi-region bucket location may result in additional storage costs for your workspace.`),
+          p(`Unless you require geo-redundancy for maximum availabity for your data, you should choose a single region bucket location. For more information see <LINK TO BLOG POST>.`)
```
aherbst-broad: TODO: I need to get a real link in this copy

src/components/ComputeModal.js
```diff
@@ -737,7 +737,7 @@ export const ComputeModalBase = ({
       setCustomEnvImage(!foundImage ? imageUrl : '')
       setJupyterUserScriptUri(currentRuntimeDetails?.jupyterUserScriptUri || '')
 
-      const locationType = location === defaultLocation ? locationTypes.default : locationTypes.region
+      const locationType = getLocationType(location)
```
aherbst-broad, line 740: The previous logic wasn't compatible with the change from `US` to `US-CENTRAL1`; specifically since the `location` variable was == to the `defaultLocation` (since the `defaultComputeLocation` is already `US-CENTRAL1`), we were setting the location type to `default` which is multi-region, which is incorrect.

I created a helper function in `region-common.js` that explicitly looks at the `allRegions` list and pulls the location type out that way rather than inferring it here. 

cc: @wnojopra as I believe you've been working on notebooks regionality

src/components/NewWorkspaceModal.js
wnojopra: Maybe https://support.terra.bio/hc/en-us/articles/360060777272-US-regional-versus-Multi-regional-US-buckets-trade-offs#h_01F5GJ8B50DDV28N92P2BPN12P ?

integration-tests/utils/integration-helpers.js
```diff
@@ -31,7 +32,7 @@ const makeWorkspace = withSignedInPage(async ({ page, billingProject }) => {
   try {
     const response = await page.evaluate(async (name, billingProject) => {
       try {
-        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {} })
+        return await window.Ajax().Workspaces.create({ namespace: billingProject, name, attributes: {}, bucketLocation: 'us-central1' })
```
wnojopra, line 36: My understanding is that rawls is being updated so that workspaces will be created with us-central1 buckets by default. Is that correct? So I'm guessing that this code is added because that rawls change isn't live yet. I'd suggest adding a comment here explaining why we need to specify the bucketLocation at this point.

wnojopra (changes requested):
Code overall looks pretty good, thanks for working on this. 

When cloning a workspace, I see this:
image
US-CENTRAL1 is capitalized, and its incorrectly labelled as multi-region. I think you can fix the multi-region labeling by adjusting the default locationTypes in region-common.js line 62. I think that will also fix the casing which is happening in region-common.js line 18.

integration-tests/utils/integration-helpers.js
cahrens: This is a good point-- can we just rely on the Rawls default instead of duplicating it in both repos? I would suggest you just wait until the Rawls PR is merged (though it will need to make its way to alpha/staging for the integration tests to run against it in those environments… can you merge it tomorrow?).

aherbst-broad: True, I'll wait until that is merged and remove this change.

src/components/NewWorkspaceModal.js
```diff
@@ -133,7 +133,7 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
   ]))
 
   const shouldShowDifferentRegionWarning = () => {
-    return !!cloneWorkspace && bucketLocation !== sourceWorkspaceLocation
+    return !!cloneWorkspace && bucketLocation !== sourceWorkspaceLocation && isLocationMultiRegion(bucketLocation) !== true
```
cahrens: I'm confused about this. Couldn't bucketLocation be multi-region but still not include sourceWorkspaceLocation (for example, bucketLocation is "US Multiregion", but sourceWorkspaceLocation is "Montreal")?

src/components/NewWorkspaceModal.js
cahrens: https://support.terra.bio/hc/en-us/articles/360060777272-US-regional-versus-Multi-regional-US-buckets-trade-offs Does seem like the right article, but someone should follow up with Allie to get it corrected to match this change. For instance, it currently says "If you don't need your data to be stored in multiple US regions, you can save storage costs in Terra by changing from the default (multi-regional) to regional cloud storage buckets.", and it doesn't mention Montreal.

xuscrbw:
Think the test code for analysis-context-bar and run-notebook is broken here because this change added another screen in the modal which we are not expecting. You will need to add the changes. I think it's just clicking 'Next' and then 'Create' on this view: 
Screen Shot 2022-06-10 at 11 20 52 AM

Let me know if you need more help with finding where to make changes.

src/components/NewWorkspaceModal.js
aherbst-broad: Ignore, this was a bad change on my part which has been backed out.

integration-tests/utils/integration-helpers.js
cahrens: FWIW, we are putting this back in to be consistent with the fact that for non-test usage we do pass the bucketLocation instead of relying on the Rawls default.

But @aherbst-broad, can you reference the default location constant instead of explicitly putting `us-central1`?

aherbst-broad:
@xuscrbw thanks, the breakage was due to the IT helper not placing the workspace in the new default storage region of us-central1. I've adjusted the IT helper to default to set us-central1 explicitly. When [this change ](https://github.com/broadinstitute/rawls/pull/1851) to rawls ships, we can back out the IT helper change. The linked rawls PR is blocked on a downstream API consumer making a needed change so they're resilient to the us-central1.

integration-tests/utils/integration-helpers.js
aherbst-broad: Following up: we are blocked on shipping the rawls PR due to a downstream consumer needing to make a change. Until that ships I'm going to explicitly set the region on the IT helper. I've added a comment to that effect here as well.

src/components/NewWorkspaceModal.js
aherbst-broad: I have added the needed link to a forthcoming new article. I'll make sure that it's live for this change.

integration-tests/utils/integration-helpers.js
cahrens: To be consistent with the fact that we explicitly pass it for non-test usage, I think it is fine to always pass the default value. But I think you should use the constant.

wnojopra (approved):
Testing and code LGTM. However there is one more small thing you can fix. On line 170 and 171 in NewWorkspaceModal.js, these variables are incorrectly set to 'region' if the location is 'US'. It should be set to 'multi-region'. This is because the logic assumed the defaultLocation is multi-region. This is obviously no longer true. The current consequence of this bug is that it causes the word 'us' to be lowercase instead of uppercase in the clone workspace dialog:
image

But it would still be good to fix so that future developers can use these variables correctly.

integration-tests/utils/integration-helpers.js
```diff
@@ -1,10 +1,12 @@
 const _ = require('lodash/fp')
 const uuid = require('uuid')
 const {
-  click, clickable, dismissNotifications, fillIn, findText, gotoPage, input, signIntoTerra, waitForNoSpinners, navChild, noSpinnersAfter, navOptionNetworkIdle
+  click, clickable, dismissNotifications, fillIn, findText, gotoPage, input, signIntoTerra, waitForNoSpinners, navChild, noSpinnersAfter,
+  navOptionNetworkIdle
 } = require('./integration-utils')
 const { fetchLyle } = require('./lyle-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
+const { defaultLocation } = require('../../src/libs/runtime-utils')
```
cahrens: Ugh, I guess it isn't feasible to import from src in the integration tests. Feel free to go back to hard-coding us-central1.

src/components/NewWorkspaceModal.js
```diff
@@ -246,9 +246,27 @@ const NewWorkspaceModal = withDisplayName('NewWorkspaceModal', ({
...
       ])]),
+      isLocationMultiRegion(bucketLocation) && div({ style: { ...warningStyle } }, [
+        icon('warning-standard', { size: 24, style: { color: colors.warning(), flex: 'none', marginRight: '0.5rem' } }),
+        div({ style: { flex: 1 } }, [
+          `Effective October 1, 2022, Google Cloud will charge egress fees on data stored in multi-region storage buckets.`,
+          p(`Choosing a multi-region bucket location may result in additional storage costs for your workspace.`),
+          p(
+            [
+              `Unless you require geo-redundancy for maximum availabity for your data, you should choose a single region bucket location.`,
+              h(Link, { href: 'https://terra.bio/moving-away-from-multi-regional-storage-buckets', ...Utils.newTabLinkProps },
```
cahrens, line 260: This link isn't working for me-- I expect it just hasn't been made public yet?

aherbst-broad: Correct, will double check re: timing w/adam.

cahrens:
@aherbst-broad I'm still seeing the "us" capitalization issue/incorrect values in lines 170-171 that @wnojopra mentioned above.

integration-tests/utils/integration-helpers.js
aherbst-broad: As discussed below, we can't reference the desired constant from this module, I've restored the hardcoded string here.

aherbst-broad:
@cahrens fixed in the latest push
Screen Shot 2022-06-13 at 8 57 28 AM

cahrens (approved):

PR #3127, opened 2022-06-09
[SUP-660] Support Dockstore workflows on the featured workflows page

nawatts:
The featured workflows page (https://app.terra.bio/#library/code) fetches a list of featured workflows (identified by namespace and name) from a JSON file stored in the `firecloud-alerts` bucket. It also fetches information about those workflows from the Methods Repository and uses that to display a description of the workflow. This adds support for displaying featured workflows from Dockstore.

Dockstore workflows from the [gatk-workflows organization](https://dockstore.org/search?organization=gatk-workflows&entryType=workflows&searchMode=files) can be added to the featured workflows list by adding an entry containing the workflow's TRS ID.

For example:
```
{ "id": "#workflow/github.com/gatk-workflows/seq-format-conversion/BAM-to-Unmapped-BAM" }
```

The TRS ID can be found on the workflow page on Dockstore.

https://dockstore.org/workflows/github.com/gatk-workflows/seq-format-conversion/BAM-to-Unmapped-BAM:3.0.0?tab=info

Screen Shot 2022-06-09 at 5 46 19 PM

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1686,10 +1686,8 @@ const Dockstore = signal => ({
     return res.json()
   },
 
-  listTools: async (params = {}) => {
-    const response = await fetchDockstore(`api/ga4gh/v1/tools?${qs.stringify(params)}`)
-    return response.json()
-  }
+  listTools: (params = {}) =>
+    fetchDockstore(`api/ga4gh/v1/tools?${qs.stringify(params)}`).then(r => r.json())
 })
 
 
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1684,6 +1684,11 @@ const Dockstore = signal => ({
   getVersions: async ({ path, isTool }) => {
     const res = await fetchDockstore(dockstoreMethodPath({ path, isTool }), { signal })
     return res.json()
+  },
+
+  listTools: async (params = {}) => {
+    const response = await fetchDockstore(`api/ga4gh/v1/tools?${qs.stringify(params)}`)
+    return response.json()
   }
 })
 
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1686,8 +1686,7 @@ const Dockstore = signal => ({
     return res.json()
   },
 
-  listTools: (params = {}) =>
-    fetchDockstore(`api/ga4gh/v1/tools?${qs.stringify(params)}`).then(r => r.json())
+  listTools: (params = {}) => fetchDockstore(`api/ga4gh/v1/tools?${qs.stringify(params)}`).then(r => r.json())
 })
 
 
--- a/src/pages/library/Code.js
+++ b/src/pages/library/Code.js
@@ -97,7 +97,7 @@ const getFeaturedMethods = async signal => {
   ])
 
   const methodsRepoMethodDetails = _.flow(
-    _.groupBy(method => `${method.namespace}/${method.name}`),
+    _.groupBy(({ namespace, name }) => `${namespace}/${name}`),
     _.mapValues(_.maxBy('snapshotId'))
   )(methodsRepoMethodsList)
 
@@ -135,7 +135,7 @@ const Code = () => {
           div({ style: { display: 'flex', flexWrap: 'wrap' } }, [
             _.map(method => {
               const { namespace, name, id } = method
-              const isMethodsRepoMethod = namespace && name
+              const isMethodsRepoMethod = !!(namespace && name)
               const href = isMethodsRepoMethod ?
                 `${getConfig().firecloudUrlRoot}/?return=${returnParam()}#methods/${namespace}/${name}/` :
                 `${getConfig().dockstoreUrlRoot}/workflows/${id.replace(/^#workflow\//, '')}`
--- a/src/pages/library/Code.js
+++ b/src/pages/library/Code.js
@@ -1,5 +1,5 @@
 import _ from 'lodash/fp'
-import { useEffect, useState } from 'react'
+import { useState } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
 import { Clickable, Link } from 'src/components/common'
 import FooterWrapper from 'src/components/FooterWrapper'
@@ -14,9 +14,8 @@ import { withErrorReporting } from 'src/libs/error'
 import { getAppName, returnParam } from 'src/libs/logos'
 import { useCancellation, useOnMount, useStore } from 'src/libs/react-utils'
 import { authStore } from 'src/libs/state'
-import * as StateHistory from 'src/libs/state-history'
 import * as Style from 'src/libs/style'
-import { withBusyState } from 'src/libs/utils'
+import * as Utils from 'src/libs/utils'
 
 
 const styles = {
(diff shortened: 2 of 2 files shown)


src/pages/library/Code.js
```diff
@@ -26,9 +25,26 @@ const styles = {
...
+  let description
+
+  if (isMethodsRepoMethod) {
+    name = method.name
+    description = method.synopsis
+    href = `${getConfig().firecloudUrlRoot}/?return=${returnParam()}#methods/${method.namespace}/${method.name}/`
+  } else {
+    name = method.toolname
+    description = method.description
+    href = `${getConfig().dockstoreUrlRoot}/workflows/${method.id.replace(/^#workflow\//, '')}`
```
nawatts, line 42: I'm kind of guessing at the URL format for Dockstore. This seems to work.

src/pages/library/Code.js
```diff
@@ -26,9 +25,26 @@ const styles = {
...
+  let href
+  let description
+
+  if (isMethodsRepoMethod) {
+    name = method.name
+    description = method.synopsis
+    href = `${getConfig().firecloudUrlRoot}/?return=${returnParam()}#methods/${method.namespace}/${method.name}/`
+  } else {
+    name = method.toolname
+    description = method.description
```
nawatts, line 41: Dockstore workflow descriptions contain Markdown, so it's less useful to display them here but the Featured Workspaces page also has unrendered Markdown descriptions.

src/pages/library/Code.js
```diff
@@ -26,7 +25,12 @@ const styles = {
   }
 }
 
-export const MethodCard = ({ method: { name, synopsis }, ...props }) => {
+export const MethodCard = ({ method, ...props }) => {
+  const isMethodsRepoMethod = method.namespace && method.name
+
+  const name = isMethodsRepoMethod ? method.name : method.toolname
+  const description = isMethodsRepoMethod ? method.synopsis : method.description
```
nawatts, line 32: Dockstore workflow descriptions contain Markdown, so it doesn't look great here. The Featured Workspaces page also has unrendered Markdown descriptions though.

src/pages/library/Code.js
```diff
@@ -122,10 +134,15 @@ const Code = () => {
           div({ style: styles.header }, 'GATK4 Best Practices workflows'),
           div({ style: { display: 'flex', flexWrap: 'wrap' } }, [
             _.map(method => {
-              const { namespace, name } = method
+              const { namespace, name, id } = method
+              const isMethodsRepoMethod = namespace && name
+              const href = isMethodsRepoMethod ?
+                `${getConfig().firecloudUrlRoot}/?return=${returnParam()}#methods/${namespace}/${name}/` :
+                `${getConfig().dockstoreUrlRoot}/workflows/${id.replace(/^#workflow\//, '')}`
```
nawatts, line 141: Kind of guessing on the Dockstore URL format. This seems to work.

src/libs/ajax.js
```diff
@@ -1684,6 +1684,11 @@ const Dockstore = signal => ({
   getVersions: async ({ path, isTool }) => {
     const res = await fetchDockstore(dockstoreMethodPath({ path, isTool }), { signal })
     return res.json()
+  },
+
+  listTools: async (params = {}) => {
+    const response = await fetchDockstore(`api/ga4gh/v1/tools?${qs.stringify(params)}`)
+    return response.json()
   }
```
dmohs: I know this follows the existing pattern, but it is unnecessarily verbose, in my opinion. I'm not sure anyone has the bandwidth to put in a refactoring PR to address this universally, and in any case I'd like to start small, so I suggest:
```suggestion
  listTools: (params = {}) =>
    fetchDockstore(`api/ga4gh/v1/tools?${qs.stringify(params)}`).then(r => r.json())
```

dmohs (approved):

src/pages/library/Code.js
kyuksel: Did UX have any thoughts on it?

src/pages/library/Code.js
```diff
@@ -85,34 +89,42 @@ export const MethodRepoTile = () => {
...
 
+const getFeaturedMethods = async signal => {
+  const [featuredMethods, methodsRepoMethodsList, dockstoreMethodsList] = await Promise.all([
+    Ajax(signal).FirecloudBucket.getFeaturedMethods(),
+    Ajax(signal).Methods.list({ namespace: 'gatk' }),
+    Ajax(signal).Dockstore.listTools({ organization: 'gatk-workflows' })
+  ])
+
+  const methodsRepoMethodDetails = _.flow(
+    _.groupBy(method => `${method.namespace}/${method.name}`),
```
kyuksel: Nit: For consistency with below `_.map`:
```suggestion
    _.groupBy(({ namespace, name }) => `${namespace}/${name}`),
```

src/pages/library/Code.js
```diff
@@ -122,10 +134,15 @@ const Code = () => {
           div({ style: styles.header }, 'GATK4 Best Practices workflows'),
           div({ style: { display: 'flex', flexWrap: 'wrap' } }, [
             _.map(method => {
-              const { namespace, name } = method
+              const { namespace, name, id } = method
+              const isMethodsRepoMethod = namespace && name
```
kyuksel: For explicitness and consistency with the codebase (at least the more recent portion):
```suggestion
              const isMethodsRepoMethod = !!namespace && !!name
```

src/pages/library/Code.js
kyuksel: If Dockstore doesn't guarantee a contract regarding the URL format, I wonder if we should have a test that checks it so we can detect any breakages in the future  🤔

src/pages/library/Code.js
nawatts: Done in [71356c2](https://github.com/DataBiosphere/terra-ui/pull/3127/commits/71356c27c257695e216f095763cb789b6ce51e62).

src/pages/library/Code.js
nawatts: Done in [71356c2](https://github.com/DataBiosphere/terra-ui/pull/3127/commits/71356c27c257695e216f095763cb789b6ce51e62).

If we care about this style, let's turn on https://eslint.org/docs/rules/prefer-destructuring.

src/pages/library/Code.js
nawatts: While I didn't find see anything documented for URL formats, I'm not worried about it changing.

On further inspection, this looks consistent with how we handle Dockstore URLs elsewhere.
https://github.com/DataBiosphere/terra-ui/blob/f4329d39eaa8f7f8e6f8513a358a038851008268/src/libs/ajax.js#L176
https://github.com/DataBiosphere/terra-ui/blob/f4329d39eaa8f7f8e6f8513a358a038851008268/src/components/common.js#L368-L373

src/pages/library/Code.js
nawatts: I didn't ask. The Featured Workspaces is the same way, so I took that as adequate precedent.

kyuksel (approved):

PR #2709, opened 2021-12-03
[PF-692] Phase 3 regionality release (support Montreal buckets)

wnojopra:
Follow up from #2680. Phase 3 allows the UI to create workspace buckets in northamerica-northeast1 (Montreal). 

To test:
1) Create a new workspace. Verify that the location dropdown now includes Montreal. Create a workspace in Montreal.
2) Attempt to clone the new workspace. Verify the location dropdown includes US multi-regional, Montreal, and Iowa. Verify the egress warning appears properly when choosing a different location. Create a cloned workspace.
3) Repeat step 2 but for a workspace in the US. 

image

PR #3128, opened 2022-06-09
[IA-3483] Add integration test for running a notebook in a non-US region

wnojopra:
Followup from #3115.
Generally tests the regionality features, which was added in #2680 and #2709

### DIFF
--- a/integration-tests/tests/run-regional-notebook.js
+++ b/integration-tests/tests/run-regional-notebook.js
@@ -1,6 +1,4 @@
-// The notebook run-regional-notebook.js is intended to be the same as this,
-// but with additional regional requirements. Generally any changes to this
-// file should also be made in run-regional-notebook.js
+// This test is intended to be the same as run-notebook.js, but in a non-US region.
 const _ = require('lodash/fp')
 const { withRegisteredUser, withBilling, withWorkspace } = require('../utils/integration-helpers')
 const {
@@ -10,12 +8,13 @@ const { registerTest } = require('../utils/jest-utils')
 
 
 const notebookName = 'TestNotebook'
+const bucketLocation = 'northamerica-northeast1'
 
-const testRunNotebookFn = _.flow(
+const testRunRegionalNotebookFn = _.flow(
   withWorkspace,
   withBilling,
   withRegisteredUser
-)(async ({ workspaceName, page, testUrl, token }) => {
+)(async ({ workspaceName, bucketLocation, page, testUrl, token }) => {
   await signIntoTerra(page, { token, testUrl })
 
   await click(page, clickable({ textContains: 'View Workspaces' }))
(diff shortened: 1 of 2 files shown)


kyuksel:
@wnojopra `run-notebook` test will be deprecated in favour of [run-analysis](https://github.com/DataBiosphere/terra-ui/pull/3037) test soonish™ so it may be better to base the new test you're adding off of the `run-analysis`.

integration-tests/tests/run-notebook.js
```diff
@@ -1,3 +1,6 @@
+// The notebook run-regional-notebook.js is intended to be the same as this,
+// but with additional regional requirements. Generally any changes to this
+// file should also be made in run-regional-notebook.js
```
kyuksel, line 3: Is it possible to factor out the common code?

wnojopra:
> @wnojopra `run-notebook` test will be deprecated in favour of [run-analysis](https://github.com/DataBiosphere/terra-ui/pull/3037) test soonish™ so it may be better to base the new test you're adding off of the `run-analysis`.

I like the idea, but that test won't catch a bug like [IA-3483]. This test needs to exist to confirm that the current way to get a regional cloud environment is working. 

[IA-3483]: https://broadworkbench.atlassian.net/browse/IA-3483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

integration-tests/tests/run-notebook.js
cahrens: Yes, please do not duplicate the code. And is it necessary to do the whole test with the different location, or do you just want to verify a subset of the functionality?

kyuksel:
> > @wnojopra `run-notebook` test will be deprecated in favour of [run-analysis](https://github.com/DataBiosphere/terra-ui/pull/3037) test soonish™ so it may be better to base the new test you're adding off of the `run-analysis`.
> 
> I like the idea, but that test won't catch a bug like [IA-3483](https://broadworkbench.atlassian.net/browse/IA-3483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ). This test needs to exist to confirm that the current way to get a regional cloud environment is working.

You mean, `run-analysis` as is won't catch the bug because it doesn't `Customize` the compute? If so, that's fine. I didn't mean we not add a test. I was trying to suggest the new test use the `Analyses` tab functionality rather than `Notebooks` tab functionality as the latter will be removed (and so will `run-notebook` be).

> I didn't mean we not add a test.

That being said, I do wonder if we could just modify `run-analysis` to use a custom location. I feel it'd still test all the other functionality and we'd accomplish two goals with one test?

cahrens:
@wnojopra any objection to closing this PR given that it's been open for so long?

PR #3129, opened 2022-06-10
[IA-3343] Create a disk reattachment integration test

n8landolt:
## Description

Write an integration test that tests creating and destroying an environment, and confirm the persistent disk reattaches.

## Testing

- [ ] Run flaky with 10 retries


kyuksel:
@n8landolt have you considered appending the testing of disk re-attachment functionality to the existing test, or mocking the first runtime/notebook creation piece (probably difficult in this case but worth considering)? These end-to-end (aka critical user journey -- CUJ) tests are quite resource intensive and difficult to get right (especially in terms of avoiding flakiness) so adding a test whose steps overlap with an existing test should be the last resort in my opinion.

integration-tests/tests/disk-reattachment.js
```diff
@@ -23,37 +23,35 @@ const testDiskReatachmentFn = _.flow(
...
-
   // Create a runtime
   await click(page, clickable({ textContains: 'Environment Configuration' }))
   await findElement(page, getAnimatedDrawer('Cloud Environment Details'))
   await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Settings' })) })
-  await findElement(page, getAnimatedDrawer('Jupyter Cloud Environment'), { timeout: 40000 })
+  await findElement(page, getAnimatedDrawer('Jupyter Cloud Environment'), { timeout: 45000 })
   await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Create' })) })
-
+  await delay(10000)
```
cahrens: Is there nothing that you can poll for on the page to indicate that the thing you are waiting for has completed?

Delays like this should only be introduced as an absolute last resort, and they should be well commented. Delays have the following problems--

1) They can lead to flakiness because if the operation is taking longer than expected, the delay may not be long enough.
2) They bloat the length of the test because they usually have to be much longer than the expected length of the operation to prevent the previous issue from occurring. If instead you poll for the state you are waiting for, the test will only take as long as actually necessary.

cahrens:
@n8landolt what are the results when you run this with flakes (100 runs, concurrency of 10)?

kyuksel:
> Run flaky with 10 retries

What does this mean?

Also, could you please report results from [running test-flakes](https://github.com/DataBiosphere/terra-ui/tree/dev/integration-tests#detecting-flakiness)?

n8landolt:
> > Run flaky with 10 retries
> 
> What does this mean?
> 
> Also, could you please report results from [running test-flakes](https://github.com/DataBiosphere/terra-ui/tree/dev/integration-tests#detecting-flakiness)?

Will run this over lunch time! I want to make sure to check for flakiness before we merge. Locally I've had 100% success rate with a few iterations, but circle ci has not been happy...

kyuksel:
> > > Run flaky with 10 retries
> > 
> > 
> > What does this mean?
> > Also, could you please report results from [running test-flakes](https://github.com/DataBiosphere/terra-ui/tree/dev/integration-tests#detecting-flakiness)?
> 
> Will run this over lunch time! I want to make sure to check for flakiness before we merge. Locally I've had 100% success rate with a few iterations, but circle ci has not been happy...

Thank you! `test-flakes` runs with `RUNS=100` (or higher) and `CONCURRENCY=10` tend to be very good in surfacing test issues that a few individual runs don't. Would recommend having that run before even un-drafting a PR  😊

PR #3130, opened 2022-06-10
[DO NOT MERGE] Wire bucketLocation through workspace creation

kyuksel:
**Note: Opening for illustrative purposes only.** 

Currently, `withWorkspace` wrapper creates a workspace in a default region. These changes aim for it to be optionally customizable by a test.

Didn't test end-to-end.

One thing I don't like is that the changes aren't backwards-compatible and require existing tests calling `withWorkspace` to be modified (even though it's just a two-char change, i.e. `withWorkspace()` instead of `withWorkspace`).

### DIFF
--- a/integration-tests/tests/find-workflow.js
+++ b/integration-tests/tests/find-workflow.js
@@ -7,7 +7,7 @@ const { withUserToken } = require('../utils/terra-sa-utils')
 
 
 const testFindWorkflowFn = _.flow(
-  withWorkspace(),
+  withWorkspace,
   withUserToken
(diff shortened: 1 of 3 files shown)


integration-tests/tests/find-workflow.js
```diff
@@ -7,7 +7,7 @@ const { withUserToken } = require('../utils/terra-sa-utils')
 
 
 const testFindWorkflowFn = _.flow(
-  withWorkspace,
+  withWorkspace(),
```
kyuksel: This is to illustrate how tests would need to be modified if they don't care about customizing `bucketLocation`.

PR #3132, opened 2022-06-10
[SU-134] Improve error handling for deleting a data table

nawatts:
Currently, if the request to delete a data table fails, an error notification is shown but the modal is not dismissed. Furthermore, one possible reason that deleting a data table may fail is if rows in the table are referenced in other tables. It would be nice to distinguish this case (requires some user action) from a system error.

With this change, the confirmation modal will be dismissed if an error occurs. And if the request failed because the table is referenced in another table, it will show a warning notification explaining the situation instead of the standard error notification.

This can be further improved by having Rawls return the list of references so that we can offer to delete them as well, similar to how deleting selected rows from a table works.

## Before
https://user-images.githubusercontent.com/1156625/173149726-fe8de16f-23b0-4626-9796-363a5dfeb231.mov

## After
https://user-images.githubusercontent.com/1156625/173149752-016ba0c1-2847-44c7-a1b4-5f57fa6ed2b7.mov

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -31,6 +31,7 @@ import { getConfig } from 'src/libs/config'
 import { reportError, withErrorReporting } from 'src/libs/error'
 import Events, { extractWorkspaceDetails } from 'src/libs/events'
 import * as Nav from 'src/libs/nav'
+import { notify } from 'src/libs/notifications'
 import { forwardRefWithName, useCancellation, useOnMount, useStore, withDisplayName } from 'src/libs/react-utils'
 import { asyncImportJobStore } from 'src/libs/state'
 import * as StateHistory from 'src/libs/state-history'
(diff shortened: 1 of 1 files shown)


kyuksel (approved):
The code changes look great! Didn't test but the screen recording also looks great.

src/pages/workspaces/workspace/Data.js
```diff
@@ -558,15 +559,23 @@ const DataTableActions = ({ workspace, tableName, rowCount, onRenameTable, onDel
...
+      onConfirm: Utils.withBusyState(setLoading)(async () => {
+        try {
+          await Ajax().Workspaces.workspace(namespace, name).deleteEntitiesOfType(tableName)
+          Ajax().Metrics.captureEvent(Events.workspaceDataDeleteTable, {
+            ...extractWorkspaceDetails(workspace.workspace)
+          })
+          onDeleteTable(tableName)
+        } catch (error) {
+          setDeleting(false)
+          if (error.status === 409) {
```
kyuksel, line 571: I assume we get `409` only when a table can't be deleted due to would-be-dangling references, which seems to be the case based on Rawls swagger.

image

cahrens (approved):

src/pages/workspaces/workspace/Data.js
nawatts: Yes. Deleting selected rows behaves the same way.
https://rawls.dsde-prod.broadinstitute.org/#/entities/delete_entities

Issue #527, opened 2018-08-15
SPIKE: Notebooks version management


bradtaylor:
Currently, we localize notebooks to GCS object storage. If two users are working concurrently, they will overwrite each others changes.

We would like to be able to provide some measure of version control. For example, checkpoint versioning. This might include the ability to:
Save a checkpoint version of the notebook, with a comment.
Access a list of checkpoint versions, included who/when saved and the comment
View a meaningful diff of each checkpoint against the current version of the notebook
Restore a previous version as the current version
Delete a previous version

This describes the behavior in Overleaf, a latex authoring tool. The chain of saving is linear, and we don't have to worry necessarily about branches

How practical would achieving something like the above be?

bradtaylor:
This is being handled elsewhere?

PR #3134, opened 2022-06-13
[PE-4] Landing page text update for Project Singular

kyuksel:
The PR changes
- involve no functional changes;
- add the text highlighted on the screenshot below;
- can be enabled via `isProjectSingular: true` flag while running locally.

image

**EDIT:** I had somehow missed half of the requested changes :( The latest looks like the following:

image

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -54,39 +54,40 @@ const styles = {
   }
 }
 
-export const Clickable = forwardRefWithName('Clickable', ({ href, as = (!!href ? 'a' : 'div'), disabled, tooltip, tooltipSide, tooltipDelay, useTooltipAsLabel, onClick, children, ...props }, ref) => {
-  const child = h(Interactive, {
-    'aria-disabled': !!disabled,
-    as, disabled, ref,
-    onClick: (...args) => onClick && !disabled && onClick(...args),
-    href: !disabled ? href : undefined,
-    tabIndex: disabled ? '-1' : '0',
-    ...props
-  }, [children])
+export const Clickable = forwardRefWithName('Clickable',
+  ({ href, as = (!!href ? 'a' : 'div'), disabled, tooltip, tooltipSide, tooltipDelay, useTooltipAsLabel, onClick, children, ...props }, ref) => {
+    const child = h(Interactive, {
+      'aria-disabled': !!disabled,
+      as, disabled, ref,
+      onClick: (...args) => onClick && !disabled && onClick(...args),
+      href: !disabled ? href : undefined,
+      tabIndex: disabled ? '-1' : '0',
+      ...props
+    }, [children])
 
-  // To support accessibility, every link must have a label or contain text or a labeled child.
-  // If an unlabeled link contains just a single unlabeled icon, then we should use the tooltip as the label,
-  // rather than as the description as we otherwise would.
-  //
-  // If the auto-detection can't make the proper determination, for example, because the icon is wrapped in other elements,
-  // you can explicitly pass in a boolean as `useTooltipAsLabel` to force the correct behavior.
-  //
-  // Note that TooltipTrigger does this same check with its own children, but since we'll be passing it an
-  // Interactive element, we need to do the check here instead.
-  const useAsLabel = _.isNil(useTooltipAsLabel) ? containsUnlabelledIcon({ children, ...props }) : useTooltipAsLabel
+    // To support accessibility, every link must have a label or contain text or a labeled child.
+    // If an unlabeled link contains just a single unlabeled icon, then we should use the tooltip as the label,
+    // rather than as the description as we otherwise would.
+    //
+    // If the auto-detection can't make the proper determination, for example, because the icon is wrapped in other elements,
+    // you can explicitly pass in a boolean as `useTooltipAsLabel` to force the correct behavior.
+    //
+    // Note that TooltipTrigger does this same check with its own children, but since we'll be passing it an
+    // Interactive element, we need to do the check here instead.
+    const useAsLabel = _.isNil(useTooltipAsLabel) ? containsUnlabelledIcon({ children, ...props }) : useTooltipAsLabel
 
-  // If we determined that we need to use the tooltip as a label, assert that we have a tooltip.
-  // Do the check here and pass empty properties, to bypass the check logic in useLabelAssert() which doesn't take into account the icon's properties.
-  if (useAsLabel && !tooltip) {
-    useLabelAssert('Clickable', { allowTooltip: true, allowContent: true })
-  }
+    // If we determined that we need to use the tooltip as a label, assert that we have a tooltip.
+    // Do the check here and pass empty properties, to bypass the check logic in useLabelAssert() which doesn't take into account the icon's properties.
+    if (useAsLabel && !tooltip) {
+      useLabelAssert('Clickable', { allowTooltip: true, allowContent: true })
+    }
 
-  if (tooltip) {
-    return h(TooltipTrigger, { content: tooltip, side: tooltipSide, delay: tooltipDelay, useTooltipAsLabel: useAsLabel }, [child])
-  } else {
-    return child
-  }
-})
+    if (tooltip) {
+      return h(TooltipTrigger, { content: tooltip, side: tooltipSide, delay: tooltipDelay, useTooltipAsLabel: useAsLabel }, [child])
+    } else {
+      return child
+    }
+  })
 
 export const Link = forwardRefWithName('Link', ({ disabled, variant, children, baseColor = colors.accent, ...props }, ref) => {
   return h(Clickable, _.merge({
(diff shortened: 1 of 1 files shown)


cahrens (approved):

nawatts (approved):

kyuksel:
@cahrens @nawatts Thank you for the review! I'm afraid I'll have to request re-reviews as I'd somehow missed half of the requested changes :( It initially looked too straightforward so I didn't open the PR as draft until getting the ✅ from the PO. So sorry about that.

src/components/common.js
```diff
@@ -54,39 +54,40 @@ const styles = {
   }
 }
```
kyuksel, line 56: --- Beginning of unrelated reformatter changes ---

src/components/common.js
```diff
@@ -423,8 +424,7 @@ export const UnlinkFenceAccount = ({ linkText, provider }) => {
             message: `Successfully unlinked your account from ${provider.name}`,
             timeout: 30000
           })
-        }
-        )
+        })
```
kyuksel, line 427: --- End of unrelated reformatter changes ---

src/components/common.js
```diff
@@ -510,6 +510,11 @@ export const Switch = forwardRefWithName('Switch', ({ onChange, onLabel = 'True'
 export const HeroWrapper = ({ showMenu = true, bigSubhead = false, children }) => {
   const heavyWrapper = text => bigSubhead ? b({ style: { whiteSpace: 'nowrap' } }, [text]) : text
 
+  const endText = Utils.cond(
+    [isProjectSingular(), () => div(['and ', heavyWrapper('collaborate'), ' to advance research around single ventricle heart disease.'])],
```
cahrens: @kyuksel can you see about changing `heavyWrapper` so that it uses the `strong` tag instead of the `b` tag? With the HTML as-is, the screen reader experience is very poor. It stops reading before the first bold word, then you have to individually select each bit of text afterwards (and it doesn't state that it is emphasized). I haven't tried changing it to `strong`, but I'm hoping it will fix the issues.

https://help.siteimprove.com/support/solutions/articles/80000448460-accessibility-bold-vs-strong-and-italic-vs-emphasis

cahrens: Also, it feels odd to put the last part of the sentence in a different div. If the screen reader can read the whole thing seamlessly I guess it is OK, but semantically it feels wrong.

src/components/common.js
nawatts: What was the motivation for re-formatting this?

src/components/common.js
kyuksel: @cahrens Thanks for the catch! Addressed in [e1ce4a7](https://github.com/DataBiosphere/terra-ui/pull/3134/commits/e1ce4a7f3e3a57205a552839e3761913e205c83e).

nawatts: Agreed that `div` seems like the wrong element for wrapping a bit of text. I think `span` would be preferable. Does this need a DOM element at all?

src/components/common.js
```diff
@@ -508,7 +508,12 @@ export const Switch = forwardRefWithName('Switch', ({ onChange, onLabel = 'True'
 })
 
 export const HeroWrapper = ({ showMenu = true, bigSubhead = false, children }) => {
-  const heavyWrapper = text => bigSubhead ? b({ style: { whiteSpace: 'nowrap' } }, [text]) : text
+  const heavyWrapper = text => bigSubhead ? strong({ style: { whiteSpace: 'nowrap' } }, [text]) : text
```
kyuksel, line 511: For improving accessibility as [described by @cahrens](https://github.com/DataBiosphere/terra-ui/pull/3134#discussion_r895971066)

src/components/common.js
nawatts: Also, does the `collaborate` styling need to be repeated in both branches of the condition? Could only the `to advance research around single ventricle heart disease` phrase be conditionally appended?

kyuksel: > Agreed that div seems like the wrong element for wrapping a bit of text. I think span would be preferable. Does this need a DOM element at all?

No. Addressed in [fede1ef](https://github.com/DataBiosphere/terra-ui/pull/3134/commits/fede1efd03d410169d0c7c0d61d26ffaf56680fd).

kyuksel: > does the collaborate styling need to be repeated in both branches of the condition? Could only the to advance research around single ventricle heart disease phrase be conditionally appended?

Yes. Addressed in [894efab](https://github.com/DataBiosphere/terra-ui/pull/3134/commits/894efab69c93a594b2227af43333fdf65e06c2c6).

src/components/common.js
```diff
@@ -528,10 +528,12 @@ export const HeroWrapper = ({ showMenu = true, bigSubhead = false, children }) =
         `${getAppName(true)} is a ${Utils.cond(
           [isTerra(), () => 'cloud-native platform'],
           [isFirecloud(), () => 'NCI Cloud Resource project powered by Terra'],
+          [isProjectSingular(), () => 'project funded by Additional Ventures and powered by Terra'],
           () => 'project powered by Terra'
         )} for biomedical researchers to `,
         heavyWrapper('access data'), ', ', heavyWrapper('run analysis tools'), ', ',
-        span({ style: { whiteSpace: 'nowrap' } }, ['and', heavyWrapper(' collaborate'), '.'])
+        h(Fragment, [span({ style: { whiteSpace: 'nowrap' } }, ['and', strong([' collaborate'])]),
+          isProjectSingular() && ' to advance research around single ventricle heart disease', '.'])
```
nawatts:
```suggestion
        span({ style: { whiteSpace: 'nowrap' } }, ['and', strong([' collaborate'])]),
        isProjectSingular() && ' to advance research around single ventricle heart disease', '.'
```

Now that this is inlined, I don't think the Fragment is necessary.

kyuksel: Thank you! Addressed in [d3bf3a7](https://github.com/DataBiosphere/terra-ui/pull/3134/commits/d3bf3a7fe23dc8c5c801f2a394f4424a52764570).

nawatts (approved):

src/components/common.js
kyuksel: I have IntelliJ do the reformatting when I'm unsure about how to format code that I'm changing. I don't have a way to selectively apply that reformatting on a piece of a file. I hate to increase the PR diff but not reformatting now means I'd have to manually unapply reformatting every time I need to run the reformatter.

src/components/common.js
```diff
@@ -528,10 +528,13 @@ export const HeroWrapper = ({ showMenu = true, bigSubhead = false, children }) =
         `${getAppName(true)} is a ${Utils.cond(
           [isTerra(), () => 'cloud-native platform'],
           [isFirecloud(), () => 'NCI Cloud Resource project powered by Terra'],
+          [isProjectSingular(), () => 'project funded by Additional Ventures and powered by Terra'],
           () => 'project powered by Terra'
         )} for biomedical researchers to `,
         heavyWrapper('access data'), ', ', heavyWrapper('run analysis tools'), ', ',
-        span({ style: { whiteSpace: 'nowrap' } }, ['and', heavyWrapper(' collaborate'), '.'])
+        span({ style: { whiteSpace: 'nowrap' } }, ['and', strong([' collaborate'])]),
```
cahrens, line 535: FYI, when I use VoiceOver it is breaking the text here at this span. It would be nice if it wasn't necessary to force `nowrap` around "and **collaborate**". (But don't feel like you have to keep working on this)

nawatts: Using a non-breaking space character between them instead of wrapping in a span might work.

cahrens: I played around a little with this in the browser, and it seems like the screen reader is trying to be smart about where to break the text (if you make the region wide enough so it can all render on one line, it doesn't split the content when reading).

src/components/common.js
cahrens: @kyuksel you can select just a portion of the text and choose "Reformat Code".

image

src/components/common.js
kyuksel: > Using a non-breaking space character between them instead of wrapping in a span might work.

Good to know. Were you thinking 
```javascript
`and&nbsp;${strong([' collaborate'])}`
```
?

kyuksel: > if you make the region wide enough so it can all render on one line, it doesn't split the content when reading

Do you mean by increasing `width: 575` on [line #527](https://github.com/DataBiosphere/terra-ui/pull/3134/files#diff-06c322acffe4a65c324584f6184927c7d4fca7e49530365c92671c0ea1fd30e1R527)?

src/components/common.js
kyuksel: Oh wow -- so many pearls, you two 🤩

src/components/common.js
nawatts: > Were you thinking
>
> `` `and&nbsp;${strong([' collaborate'])}` ``

I think React will escape `&nbsp;`, so you may have to use the Unicode character: `\u00a0`.

```js
'and\u00a0', strong(['collaborate']),
```

cahrens: To be clear, I'm not suggesting you increase the width. I was just trying it out to understand the effect of the width on how the screen reader was behaving.

PR #3032, opened 2022-05-13
[DC-325] Enable UI integration tests on alpha / staging

slucasbroad:
Remove `targetEnvironments` and update test for Alpha environment compatibility.

PR #3135, opened 2022-06-13
[QA-1864] Integration logs handler jest-reporter fails to write out all logs

aweng98:
Repeat fix made in https://github.com/DataBiosphere/terra-ui/pull/3024 because it was broken by https://github.com/DataBiosphere/terra-ui/pull/3032.

### DIFF
--- a/integration-tests/jest-reporter.js
+++ b/integration-tests/jest-reporter.js
@@ -31,8 +31,8 @@ module.exports = class JestReporter {
 
     const writableStream = createWriteStream(logFileName)
     writableStream.on('error', ({ message }) => console.error(`Error occurred while writing Console logs to ${logFileName}.\n${message}`))
-
-    testResult.console.forEach(({ message }) => writableStream.write(`${message}\n`))
+    
+    _.forEach(({ message }) => { writableStream.write(`${message}\n`) }, testResult?.console)
 
     // Write pass/failure summaries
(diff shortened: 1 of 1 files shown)


nawatts (approved):
Since there was some conversation on #3024 about `testResult.console.forEach(...)` vs `_.forEach(..., testResult.console)`, I did some investigating. One way that [`_.forEach`](https://lodash.com/docs/4.17.15#forEach) differs from `Array.prototype.forEach` is that `_.forEach` allows breaking out of the loop:
> Iteratee functions may exit iteration early by explicitly returning false.

And [`writeableStream.write`](https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback) may return false.
> Returns: [<boolean>](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type) false if the stream wishes for the calling code to wait for the 'drain' event to be emitted before continuing to write additional data; otherwise true.

I suspect `_.forEach` would work fine if the iteratee returned undefined.
```js
_.forEach(({ message }) => { writableStream.write(`${message}\n`) }, testResult?.console)
```

aweng98:
Thank you. Worth a try. Updated code to wrap `writableStream.write(`${message}\n`)` inside curly bracket.

aednichols (approved):

PR #3137, opened 2022-06-13
[SU-88][SU-118] Sync workspaces list query parameters and filter state

nawatts:
The workspaces list initializes the selected tab and filter state from URL query parameters. However, it only updates the query parameters when the "Search workspaces" filter text is changed.

Thus, it's easy to get into a state where the page does not match the query parameters. For example, if you go to a tab, enter something in the search box, and go to a different tab. If you then refresh the page, the selected tab will change.

The root of the issue is duplicated state. The tab and filter state is stored in both component state and query parameters, and the component attempts to sync the two. This removes the duplication so that the query parameters are the source of truth for the selected tab and filter state. This way, the page content always matches the query parameters.

### DIFF
--- a/src/libs/nav.js
+++ b/src/libs/nav.js
@@ -106,7 +106,7 @@ export const Router = () => {
 }
 
 export const updateSearch = params => {
-  const newSearch = qs.stringify(params, { addQueryPrefix: true })
+  const newSearch = qs.stringify(params, { addQueryPrefix: true, arrayFormat: 'brackets' })
 
   if (newSearch !== history.location.search) {
     history.replace({ search: newSearch })
--- a/src/libs/nav.js
+++ b/src/libs/nav.js
@@ -105,10 +105,8 @@ export const Router = () => {
   ])
 }
 
-export const updateSearch = (query, params) => {
-  const newSearch = qs.stringify({
-    ...query, ...params
-  }, { addQueryPrefix: true })
+export const updateSearch = params => {
+  const newSearch = qs.stringify(params, { addQueryPrefix: true })
 
   if (newSearch !== history.location.search) {
     history.replace({ search: newSearch })
--- a/src/pages/workspaces/List.js
+++ b/src/pages/workspaces/List.js
@@ -1,6 +1,5 @@
 import { isAfter, parseJSON } from 'date-fns/fp'
 import _ from 'lodash/fp'
-import * as qs from 'qs'
 import { useEffect, useMemo, useState } from 'react'
 import { div, h, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
@@ -65,7 +64,9 @@ const useWorkspacesWithSubmissionStats = () => {
     }
   }, [workspaces, submissionStats, signal])
 
-  const workspacesWithSubmissionStats = _.map(ws => _.set('workspaceSubmissionStats', _.get(ws.workspace.workspaceId, submissionStats), ws), workspaces)
+  const workspacesWithSubmissionStats = useMemo(() => {
+    return _.map(ws => _.set('workspaceSubmissionStats', _.get(ws.workspace.workspaceId, submissionStats), ws), workspaces)
+  }, [workspaces, submissionStats])
 
   return { workspaces: workspacesWithSubmissionStats, refresh, loadingWorkspaces, loadingSubmissionStats }
 }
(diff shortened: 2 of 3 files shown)


src/libs/nav.js
```diff
@@ -105,10 +105,8 @@ export const Router = () => {
   ])
 }
 
-export const updateSearch = (query, params) => {
-  const newSearch = qs.stringify({
-    ...query, ...params
-  }, { addQueryPrefix: true })
+export const updateSearch = params => {
+  const newSearch = qs.stringify(params, { addQueryPrefix: true })
```
nawatts: There was no need for this function to take two arguments. If there are two objects, the calling function can merge them.

src/pages/workspaces/List.js
```diff
@@ -65,7 +64,9 @@ const useWorkspacesWithSubmissionStats = () => {
     }
   }, [workspaces, submissionStats, signal])
 
-  const workspacesWithSubmissionStats = _.map(ws => _.set('workspaceSubmissionStats', _.get(ws.workspace.workspaceId, submissionStats), ws), workspaces)
+  const workspacesWithSubmissionStats = useMemo(() => {
+    return _.map(ws => _.set('workspaceSubmissionStats', _.get(ws.workspace.workspaceId, submissionStats), ws), workspaces)
+  }, [workspaces, submissionStats])
```
nawatts, line 69: #3091 broke the memoization of the filtered workspaces list. This restores it.

MatthewBemis (approved):
Tested it out and it looks good to me!

src/pages/workspaces/List.js
```diff
@@ -343,7 +336,7 @@ export const WorkspaceList = () => {
             placeholder: 'Access levels',
             'aria-label': 'Filter by access levels',
             value: accessLevelsFilter,
-            onChange: data => setAccessLevelsFilter(_.map('value', data)),
+            onChange: data => Nav.updateSearch({ ...query, accessLevelsFilter: _.map('value', data) }),
```
breilly2, line 339: I don't love how these array values (this an access level) end up serializing:
Screen Shot 2022-06-15 at 2 48 51 PM
Is there a way we can do away with the array indexes and instead use multi-valued attributes, inferring ordering by order in the query string? Do you think it's not worth the effort to try?

Also, fresh off the security training, this is one of those serialization implementation details that it mentions watching out for (though I don't think there's really a problem here).

nawatts: > Is there a way we can do away with the array indexes and instead use multi-valued attributes, inferring ordering by order in the query string? Do you think it's not worth the effort to try?

So instead of `list[0]=foo&list[1]=bar`, `list[]=foo&list[]=bar`? Or `list=foo&list=bar`? The library we use for handling query strings does allow controlling how arrays are serialized.
https://www.npmjs.com/package/qs#user-content-stringifying

breilly2: I like `list=foo&list=bar` and I think that's the most natural representation of the data in a query string.

nawatts: After trying `list=foo&list=bar`, I found that it would complicate the parsing logic because there's nothing hinting that `list=foo` should be parsed into an array with one item vs a string. It can be handled here, but if we're trying to make this a standard format for URL query parameters across Terra, then I think it's likely to create some bugs (or at least some surprises for devs) down the road.

breilly2: Ah, interesting point. Oh well.

src/pages/workspaces/List.js
```diff
@@ -79,17 +80,21 @@ const workspaceSubmissionStatus = workspace => {
...
 
   const { query } = Nav.useRoute()
   const filter = query.filter || ''
-  const [accessLevelsFilter, setAccessLevelsFilter] = useState(query.accessLevelsFilter || [])
-  const [projectsFilter, setProjectsFilter] = useState(query.projectsFilter || undefined)
-  const [submissionsFilter, setSubmissionsFilter] = useState(query.submissionsFilter || [])
-  const [tab, setTab] = useState(query.tab || 'myWorkspaces')
-  const [tagsFilter, setTagsFilter] = useState(query.tagsFilter || [])
+  // Using the EMPTY_LIST constant as a default value instead of creating a new empty array on
+  // each render avoids unnecessarily recomputing the memoized filteredWorkspaces value.
```
breilly2, line 92: Nice catch!

breilly2 (approved):
👍 Thanks for finishing the work started in #2553!

src/pages/workspaces/List.js
nawatts: Can still get part of the way there. Changed from `list[0]=foo&list[1]=bar` to `list[]=foo&list[]=bar` in [3ad7337](https://github.com/DataBiosphere/terra-ui/pull/3137/commits/3ad7337a0368817c0b2545550642f8deddb2e5fa).

src/pages/workspaces/List.js
nawatts: ESLint caught it for me 😁.

PR #3138, opened 2022-06-13
DC-396: App Engine cleanup script improvements

fboulnois:
v2 of the App Engine deletion script:

- All operations which would have been done by `tail`, `head`, `awk`, `sed`, `cut`, `tr`, and `date` now directly use `jq` instead
- The deletion phrase has been removed, as gcloud asks you to confirm deletion anyways

### DIFF
--- a/scripts/delete-old-app-engine-versions.sh
+++ b/scripts/delete-old-app-engine-versions.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 #
 # Deletes old Google App Engine deployments of Terra UI in an environment. You
-# MUST have GNU date installed to be able to use this script.
+# MUST have jq installed to be able to use this script.
 #
 # USAGE: ./delete-old-app-engine-versions.sh ENV
 #   ENV must be one of dev, alpha, staging, or perf
--- a/scripts/delete-old-app-engine-versions.sh
+++ b/scripts/delete-old-app-engine-versions.sh
@@ -101,8 +101,8 @@ check_deletion_items() {
     DELETE_LIST_COUNT="${#DELETE_LIST_ITEMS[@]}"
     if [ "${DELETE_LIST_COUNT}" -lt 1 ]; then
         abort "no deployments to delete"
-    elif [ "${DELETE_LIST_COUNT}" -lt 30 ]; then
-        abort "less than 30 deployments to delete"
+    elif [ "${DELETE_LIST_COUNT}" -lt 10 ]; then
+        abort "less than 10 deployments to delete"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


scripts/delete-old-app-engine-versions.sh
```diff
@@ -163,6 +155,4 @@ printf "${INFO} Set the deletion date to ${RED}%s and earlier${RST}\n" "${DELETI
 deletion_preflight_checks
 deletion_preflight_summary
 
-enter_deletion_phrase
```
s-rubenstein, line 166: Do we no longer want a confirmation? Is the goal to automate here? If so, that sounds good.

s-rubenstein (approved):
One question. If the answer is yes, this looks good. otherwise I would probably want to talk about that before continuing.

scripts/delete-old-app-engine-versions.sh
fboulnois: `gcloud` asks you to confirm deletion anyways, so the deletion phrase ends up adding additional friction and is redundant.

scripts/delete-old-app-engine-versions.sh
```diff
@@ -45,10 +45,10 @@ abort() {
...
 }
 
-# ensure that date from GNU coreutils is installed
-check_gnu_date_installed() {
-    if ! date --version 1>/dev/null 2>&1; then
-        abort "date from GNU coreutils is required"
+# ensure that jq is installed
+check_jq_installed() {
+    if ! jq --version 1>/dev/null 2>&1; then
+        echo "jq v1.6 or above is required"
```
pshapiro4broad: Maybe add `brew install jq` to the error message?

scripts/delete-old-app-engine-versions.sh
```diff
@@ -148,13 +140,13 @@ case $1 in
...
 
 NEW_PROJECT="bvdp-saturn-$1"
-ENV_TO_EXEC="$1"
 
 printf "${INFO} Selected project ${GRN}%s${RST}\n" "${NEW_PROJECT}"
 
 check_user_permissions
 
-DELETION_DATE=$(date --date="-7 days" +%F)
+DELETION_TIME=$(jq -n 'now - 604800') # 604800 seconds = 7 days
```
pshapiro4broad: Might be clearer to write out the math
```suggestion
DELETION_TIME=$(jq -n 'now - (7 * 24 * 60 * 60)') # 7 days
```

pshapiro4broad (approved):
It's a shame `date` is so different everywhere. But at least the `jq` is pretty easy to follow.

scripts/delete-old-app-engine-versions.sh
fboulnois: Nice, `jq` really does everything.

scripts/delete-old-app-engine-versions.sh
fboulnois: Can we guarantee that everyone who runs this is using MacOS? I'm not sure.

petesantos: I personally like error messages with the install step as well. Given that it is a shell script, I think adding a `brew install jq` message is ok. You can run install and run brew on Linux as well.

petesantos (approved):

scripts/delete-old-app-engine-versions.sh
fboulnois: I added this to the error message.

PR #3139, opened 2022-06-13
[QA-1913] Terra UI tests: additional logging for troubleshooting test failures

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1913

Capture all page errors and write summary of all test results to `tests-summary.json` file. Because sometimes a UI test marked as failed but there was no failure messages in test log. A summary of test results may be helpful to determine what has happened.

Example:
`import-dockstore-workflow` failed but failure messages field is blank in test log.
[CircleCI](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/10117/workflows/b71b8ab1-c50b-40e0-bc47-7c234f4d8422)

One `tests-summary.json` in every CirlcleCI VM:

Screen Shot 2022-06-13 at 8 46 25 PM

### DIFF
--- a/integration-tests/jest-reporter.js
+++ b/integration-tests/jest-reporter.js
@@ -4,21 +4,19 @@ const { createWriteStream, mkdirSync, writeFileSync } = require('fs')
 
 
 module.exports = class JestReporter {
-  name
-
   constructor(_globalConfig, _options) {
     this.logRootDir = process.env.LOG_DIR || '/tmp/test-results'
   }
 
   onTestStart(test) {
     const { path } = test
-    this.name = parse(path).name
-    console.info(`**  Running ${this.name} at ${this.timeNow()}`)
+    console.info(`**  Running ${parse(path).name} at ${this.timeNow()}`)
   }
 
   onTestResult(_testRunConfig, testResult, _runResults) {
-    const { testResults } = testResult
-    const logFileName = `${this.getLogPath(testResults)}/${this.name}-${Date.now()}.log`
+    const { testFilePath, testResults } = testResult
+    const { name } = parse(testFilePath)
+    const logFileName = `${this.getLogPath(testResults)}/${name}-${Date.now()}.log`
 
     const writableStream = createWriteStream(logFileName)
     writableStream.on('error', ({ message }) => console.error(`Error occurred while writing Console logs to ${logFileName}.\n${message}`))
@@ -43,16 +41,16 @@ module.exports = class JestReporter {
     }, testResults)
 
     writableStream.end()
-    writableStream.on('finish', () => console.log(`**  Saved ${this.name} results: ${logFileName}`))
+    writableStream.on('finish', () => console.log(`**  Saved ${name} results: ${logFileName}`))
   }
 
   // Called after all tests have completed
   onRunComplete(_test, runResults) {
     // Save run summary to a file.
     const logDir = this.getLogPath(runResults)
-    const summaryFile = `${this.name}-summary.json`
+    const summaryFile = `tests-summary.json`
     writeFileSync(`${logDir}/${summaryFile}`, JSON.stringify(runResults, null, 2))
-    console.info(`**  Saved ${this.name} summary: ${logDir}/${summaryFile}`)
+    console.info(`**  Saved all tests summary: ${logDir}/${summaryFile}`)
     return runResults
   }
 
--- a/integration-tests/jest-reporter.js
+++ b/integration-tests/jest-reporter.js
@@ -10,7 +10,7 @@ module.exports = class JestReporter {
 
   onTestStart(test) {
     const { path } = test
-    console.info(`**  Running ${parse(path).name} at ${this.timeNow()}`)
+    console.log(`**  Running ${parse(path).name} at ${this.timeNow()}`)
   }
 
   onTestResult(_testRunConfig, testResult, _runResults) {
@@ -30,7 +30,6 @@ module.exports = class JestReporter {
       writableStream.write('----------------------------------------------\n')
       writableStream.write(`Test: ${title}\n`)
       writableStream.write(`Status: ${status}\n`)
-      console.log(`failureMessages: ${failureMessages}`)
       if (!_.isEmpty(failureMessages)) {
         writableStream.write(`Failure messages: ${failureMessages}\n`)
       }
@@ -50,7 +49,7 @@ module.exports = class JestReporter {
     const logDir = this.getLogPath(runResults)
     const summaryFile = `tests-summary.json`
     writeFileSync(`${logDir}/${summaryFile}`, JSON.stringify(runResults, null, 2))
-    console.info(`**  Saved all tests summary: ${logDir}/${summaryFile}`)
+    console.log(`**  Saved all tests summary: ${logDir}/${summaryFile}`)
     return runResults
   }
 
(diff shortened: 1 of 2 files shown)


integration-tests/jest-reporter.js
```diff
@@ -37,16 +28,49 @@ module.exports = class JestReporter {
...
+      const { title, status, failureMessages, failureDetails } = result
       writableStream.write('----------------------------------------------\n')
-      writableStream.write(`Test: ${result.title}\n`)
-      writableStream.write(`Status: ${result.status}\n`)
-      const { failureMessages = [] } = result
-      !_.isEmpty(failureMessages) && writableStream.write(`Failure message: ${failureMessages}\n`)
+      writableStream.write(`Test: ${title}\n`)
+      writableStream.write(`Status: ${status}\n`)
+      console.log(`failureMessages: ${failureMessages}`)
+      if (!_.isEmpty(failureMessages)) {
```
aweng98, line 44: Replacing usage of `&&` with `if`.

integration-tests/jest-reporter.js
```diff
@@ -37,16 +28,49 @@ module.exports = class JestReporter {
...
+  onRunComplete(_test, runResults) {
+    // Save run summary to a file.
+    const logDir = this.getLogPath(runResults)
+    const summaryFile = `${this.name}-summary.json`
+    writeFileSync(`${logDir}/${summaryFile}`, JSON.stringify(runResults, null, 2))
+    console.info(`**  Saved ${this.name} summary: ${logDir}/${summaryFile}`)
+    return runResults
+  }
+
+  getLogPath(result) {
```
aweng98: Extracted code from `onTestResult` function so that `onRunComplete` can use it as well.

integration-tests/jest-reporter.js
```diff
@@ -37,16 +28,49 @@ module.exports = class JestReporter {
...
-    writableStream.on('finish', () => console.log(`Finished writing logs to ${logFileName}`))
+    writableStream.on('finish', () => console.log(`**  Saved ${this.name} results: ${logFileName}`))
+  }
+
+  // Called after all tests have completed
+  onRunComplete(_test, runResults) {
+    // Save run summary to a file.
+    const logDir = this.getLogPath(runResults)
+    const summaryFile = `${this.name}-summary.json`
+    writeFileSync(`${logDir}/${summaryFile}`, JSON.stringify(runResults, null, 2))
```
aweng98: `runResults` is a JSON object.

cahrens:
@aweng98 When there is no failure message in the logs, I've usually been able to trace the cause back to a 500 error (which is printed now in the logs). Do you have an example failure I can look at?

aweng98:
I have one CircleCI link in PR description. Test was `import-dockstore-workflow` in `integration-tests-alpha` job.

aweng98:
Test failed due to 403 failed request. However, I think we should see failure stacktrace in test log but none exists for whatever reason unknown to me.

cahrens:
@aweng98 I'm not sure the 403 caused it to fail… I'm pretty sure we get 403s in passing tests. Sadly I don't see anything that is obviously the cause of the failure.

integration-tests/jest-reporter.js
```diff
@@ -10,24 +10,13 @@ module.exports = class JestReporter {
...
-    // Since Node 10, mkdirSync knows to not do anything if the dir exists (https://stackoverflow.com/a/71767385/2851999) so not checking explicitly
-    // We may want to specify what to do if an error occurs. Not sure if throwing is what we'd want.
-    try {
-      mkdirSync(logDir, { recursive: true })
-    } catch (err) {
-      console.error(err)
-      throw err
-    }
+    const { name } = parse(testFilePath)
+    const logFileName = `${this.getLogPath(testResults)}/${name}-${Date.now()}.log`
```
nawatts: Here this function is called with an array of `AssertionResult` objects.

https://github.com/facebook/jest/blob/main/packages/jest-reporters/src/types.ts
https://github.com/facebook/jest/blob/main/packages/jest-test-result/src/types.ts

integration-tests/jest-reporter.js
```diff
@@ -37,16 +26,48 @@ module.exports = class JestReporter {
...
 
     writableStream.end()
-    writableStream.on('finish', () => console.log(`Finished writing logs to ${logFileName}`))
+    writableStream.on('finish', () => console.log(`**  Saved ${name} results: ${logFileName}`))
+  }
+
+  // Called after all tests have completed
+  onRunComplete(_test, runResults) {
+    // Save run summary to a file.
+    const logDir = this.getLogPath(runResults)
```
nawatts: Here, this function is called with an `AggregatedResult` object.

https://github.com/facebook/jest/blob/main/packages/jest-reporters/src/types.ts
https://github.com/facebook/jest/blob/main/packages/jest-test-result/src/types.ts

It's written assuming it will be passed an array of AssertionResult objects. The Lodash functions used in `getLogPath` will handle this without throwing an error, but the logic is invalid when called with the wrong argument type.

nawatts: Even if the function handled this type of argument correctly, `getLogPath` returns a path to either a `successes` or `failures` directory. Does that make sense for a summary file which may contain information from both successful and failed tests?

integration-tests/utils/integration-utils.js
```diff
@@ -319,6 +319,20 @@ const withScreenshot = _.curry((testName, fn) => async options => {
...
+// Emitted when the page crashes
+const logError = page => {
+  const handle = msg => console.error('page.error', msg)
+  page.on('error', handle)
+  return () => page.off('error', handle)
+}
+
+// Emitted when an uncaught exception happens within the page
+const logPageError = page => {
+  const handle = msg => console.error('page.error', msg)
```
nawatts: Should the messages for logging `error` and `page.error` events be different? Currently, both are `page.error`.

integration-tests/jest-reporter.js
```diff
@@ -37,16 +26,48 @@ module.exports = class JestReporter {
...
+    const summaryFile = `tests-summary.json`
+    writeFileSync(`${logDir}/${summaryFile}`, JSON.stringify(runResults, null, 2))
+    console.log(`**  Saved all tests summary: ${logDir}/${summaryFile}`)
+    return runResults
```
nawatts: It looks like `onRunComplete` isn't intended to return anything.

https://github.com/facebook/jest/blob/main/packages/jest-reporters/src/types.ts

nawatts (changes requested):
`getLogPath` looks like it is misused (called with different/unexpected argument types) in both `onTestResult` and `onRunComplete`.

integration-tests/jest-reporter.js
```diff
@@ -37,16 +26,48 @@ module.exports = class JestReporter {
...
+  }
+
+  getLogPath(result) {
+    const { testResults } = result
```
nawatts: In `onTestResult`, this function is passed an array of AssertionResults. That array would not have a `testResults` property, so this function would always return the `successes` directory for logs.

integration-tests/utils/integration-utils.js
aweng98: 👍🏻 Updated for clarity.

integration-tests/jest-reporter.js
aweng98: 👍🏻 Fixed.

aweng98 (commented):
Removed `getLogPath` function. Save summary file to `logRootDir`.

nawatts (approved):

cahrens (approved):

PR #3141, opened 2022-06-14
[IA-3387] Azure Jupyter Lab support completed

jdcanas:
This completes the iframe support for accessing Jupyter lab hosted in an azure instance

image

To test this, you need to have an azure workspace. This leverages the existing create azure runtime flow, and then adds cookie support. The iframe proxyUrl is computed in `appLauncher` in a previous PR, and leveraged here.

One thing to note is the `setCookie` call itself takes ~30seconds, and then the jupyter lab UI takes another ~30secs to render.... not sure entirely why, though I'm fairly confident that it is a backend issue. This is being investigated, but I don't think it should block merging this work since this **will not be** user facing as of this PR.

### DIFF
--- a/src/components/runtime-common.js
+++ b/src/components/runtime-common.js
@@ -110,7 +110,6 @@ export const PeriodicCookieSetter = () => {
 }
 
 export const PeriodicAzureCookieSetter = ({ proxyUrl }) => {
-  console.log('using periodic cookie setter')
   const signal = useCancellation()
   usePollingEffect(
     withErrorIgnoring(async () => {
--- a/src/components/runtime-common.js
+++ b/src/components/runtime-common.js
@@ -11,7 +11,7 @@ import Events from 'src/libs/events'
 import { getLocalPref } from 'src/libs/prefs'
 import { useCancellation, useGetter, useOnMount, usePollingEffect, usePrevious, useStore } from 'src/libs/react-utils'
 import { getConvertedRuntimeStatus, usableStatuses } from 'src/libs/runtime-utils'
-import { authStore, cookieReadyStore, userStatus } from 'src/libs/state'
+import { authStore, azureCookieReadyStore, cookieReadyStore, userStatus } from 'src/libs/state'
 import * as Utils from 'src/libs/utils'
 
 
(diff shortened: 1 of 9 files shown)


src/components/runtime-common.js
```diff
@@ -108,6 +108,18 @@ export const PeriodicCookieSetter = () => {
   return null
 }
 
+export const PeriodicAzureCookieSetter = ({ proxyUrl }) => {
+  const signal = useCancellation()
+  usePollingEffect(
+    withErrorIgnoring(async () => {
+      await Ajax(signal).Runtimes.setAzureCookie(proxyUrl)
```
Qi77Qi, line 115: will this happen for all Azure runtimes the user has across workspaces?

Qi77Qi:
can we also update `open azure` text to just `open`

src/components/runtime-common.js
```diff
@@ -108,6 +108,18 @@ export const PeriodicCookieSetter = () => {
   return null
 }
 
+export const PeriodicAzureCookieSetter = ({ proxyUrl }) => {
```
breilly2, line 111: I think it's worth extracting the commonality between `PeriodicCookieSetter` and `PeriodicAzureCookieSetter`. Something like:
```
export const CookieSetterWrapper = f => () => {
  usePollingEffect(withErrorIgnoring(f(signal)), { ms: 5 * 60 * 1000, leading: true })
  return null
}

export const PeriodicCookieSetter = CookieSetterWrapper(async signal => {
  await Ajax(signal).Runtimes.setCookie()
  cookieReadyStore.set(true)
})
```
Which could then also be used for `PeriodicAzureCookieSetter`.

breilly2 (approved):
👍 One suggestion, but code looks good to me. I don't know how to set up an Azure workspace so I was not able to test. I'm going to rely on others doing that.

src/components/runtime-common.js
jdcanas: the point where this is invoked is when you are launching an azure app, so a specific (running) runtime is present when this component is attached, since we need the proxyUrl associated with a runtime to call setCookie. If you have more than one runtime, there is logic to select a specific one [here](https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/runtime-utils.js#L282), which the cookie setter will use. 

See here for where the cookie setter is attached https://github.com/DataBiosphere/terra-ui/pull/3141/files#diff-44105412b7d7ce49a5ee9ac6b7202ea01d275dbd1054339f21abebdf163349e5R213

src/components/runtime-common.js
jdcanas: Yeah I thought of doing that... I chose not to though since its a small enough number of lines with enough params that the abstraction would be too complex to be worth it (i.e. they both use different cookie stores, call different APIs, and one takes a param and the other doesn't).

Qi77Qi (approved):
one more comment but otherwise lgtm 
https://github.com/DataBiosphere/terra-ui/pull/3141#issuecomment-1159142080

PR #3144, opened 2022-06-14
[DC-377] Sr null safe date

s-rubenstein:
<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/pages/library/DataBrowserDetails.js
+++ b/src/pages/library/DataBrowserDetails.js
@@ -48,7 +48,7 @@ const MetadataDetailsComponent = ({ dataObj, name }) => {
       ]),
       div({ style: styles.attributesColumn }, [
         h3({ style: styles.headers }, ['Last Updated']),
-        Utils.makeStandardDate(dataObj['dct:modified'])
+        dataObj['dct:modified'] && Utils.makeStandardDate(dataObj['dct:modified'])
       ]),
       div({ style: styles.attributesColumn }, [
         h3({ style: styles.headers }, ['Version']),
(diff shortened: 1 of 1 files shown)


src/pages/library/DataBrowserDetails.js
```diff
@@ -48,7 +48,7 @@ const MetadataDetailsComponent = ({ dataObj, name }) => {
       ]),
       div({ style: styles.attributesColumn }, [
         h3({ style: styles.headers }, ['Last Updated']),
-        Utils.makeStandardDate(dataObj['dct:modified'])
+        dataObj.lastUpdated
```
pshapiro4broad: For context, this property is initialized in https://github.com/DataBiosphere/terra-ui/blob/dev/src/pages/library/dataBrowser-utils.js#L82 
```
    lastUpdated: !!dataset['dct:modified'] && new Date(dataset['dct:modified']),
```

pshapiro4broad (approved):
Is there a way to keep track of cases like this so we can regression test for them?

Or maybe this can wait until we have a schema, and can use it to generate random test data with all the possible missing values.

src/pages/library/DataBrowserDetails.js
s-rubenstein: Yeah, although interestingly it didn't seem to actually work here, so I am moving to `dataObj['dct:modified'] && Utils.makeStandardDate(dataObj['dct:modified'])`

I think maybe we wait until we have a schema, then go through the UI and make it manage fields based off the schema. Because it's not like we have definitive information right now.

s-rubenstein: (It worked for the null case, but didn't work for the non-null case, which I hadn't checked yet.)

s-rubenstein: @pshapiro4broad I am going to file a ticket to audit the UI for cases like this to do after we are done with the schema, because it would be good to drive the UI off of only schema defined values.

fboulnois (approved):
lgtm

nawatts (approved):

PR #3147, opened 2022-06-15
DC-396: More accurately check App Engine permissions

fboulnois:
Before the script checked permissions using `list`, but now they are correctly checked using `delete` instead.

See previous discussion here: https://github.com/DataBiosphere/terra-ui/pull/3051#discussion_r876353448

### DIFF
--- a/scripts/delete-old-app-engine-versions.sh
+++ b/scripts/delete-old-app-engine-versions.sh
@@ -54,7 +54,8 @@ check_jq_installed() {
 
 # ensure that user has appropriate permissions for app engine
 check_user_permissions() {
-    if ! gcloud app services list --project="${NEW_PROJECT}" 1>/dev/null 2>&1; then
+    FAKE_VERSION="example-invalid-version-for-permission-testing"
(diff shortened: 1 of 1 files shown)


nawatts (approved):

pshapiro4broad (approved):
LGTM

scripts/delete-old-app-engine-versions.sh
```diff
@@ -54,7 +54,8 @@ check_jq_installed() {
 
 # ensure that user has appropriate permissions for app engine
 check_user_permissions() {
-    if ! gcloud app services list --project="${NEW_PROJECT}" 1>/dev/null 2>&1; then
+    FAKE_VERSION="example-invalid-version-for-permission-testing"
+    if ! gcloud app versions delete "${FAKE_VERSION}" --project="${NEW_PROJECT}" 1>/dev/null 2>&1; then
```
nawatts, line 58: Does this also fail if the user has delete permissions but `$FAKE_VERSION` does not exist?

fboulnois: No, it gives a warning instead, which is recorded as command success.

PR #3152, opened 2022-06-17
[SU-136] Fix position of data table ID column

nawatts:
The first column (besides the checkbox) in a data table contains the name of the entity displayed in that row. When horizontally scrolling through a table with many columns, that column scrolls out of view. This makes it easy to lose track of which row you are currently looking at. This fixes the checkbox and name columns in place so that they are always visible when horizontally scrolling the table. This is similar to how the header row is fixed in place when vertically scrolling the table.

## Before
https://user-images.githubusercontent.com/1156625/174343432-7203064f-1c3f-401f-8cfe-02e7e9be4ef7.mov

## After
https://user-images.githubusercontent.com/1156625/174343449-e43895d6-37ef-4e0f-935e-426bd5e42221.mov

A challenge here was that, for accessibility, the cell elements must be rendered inside the row elements. This requires rendering them inside RVGrid's inner scroll container (where the row elements are rendered). However, they cannot be rendered like most cells in the RVGrid, because RVGrid does not render cells it thinks have been scrolled out of view. Thus, RVGrid is only told about the other (unfixed) columns and its [`cellRangeRenderer`](https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md#cellrangerenderer), which is already customized to render row elements, is further customized to always render the fixed columns in each visible row.

To position the cells in fixed columns correctly and avoid having them scroll out of view like other cells in RVGrid's inner scroll container, the cells are styled with `position: fixed` GridTable's root element with `contain: strict`. This makes the root element of GridTable act as a [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block) for `position: fixed` descendants.

However, RVGrid sets `will-change: transform` on its root element, which would make it the containing block for the fixed columns' cells. To prevent this, RVGrid's default style is overridden with `willChange: 'auto'`.

I'm unsure of the performance implications of removing `will-change: transform`. There was a comment in react-virtualized's code explaining it:
> Without this property, Chrome repaints the entire Grid any time a new row or column is added.
> Firefox only repaints the new row or column (regardless of this property).
> Safari and IE don't support the property at all.

I didn't notice any degradation in scrolling with this change. Even if there is, I'll claim that this may change may be worth it if it avoids the need to scroll back and forth to see row IDs. `contain: strict` on GridTable's root element should at least prevent any impact on the rest of the page. In both current dev and this branch, scrolling the data table in the workflow data selection seem smoother than scrolling the data table in the workspace data tab.

### DIFF
--- a/src/components/table.js
+++ b/src/components/table.js
@@ -422,6 +422,14 @@ export const GridTable = forwardRefWithName('GridTable', ({
     }
   }))
 
+  const fixedColumns = _.slice(0, numFixedColumns, columns)
+  const totalFixedColumnsWidth = _.sum(_.map('width', fixedColumns))
+  // The value at index i in this array is a sum of the widths of columns to the left of the column at index i.
+  const fixedColumnOffsets = _.transform((offsets, column) => { offsets.push(column.width) }, [0])(columns)
+
+  const unfixedColumns = _.slice(numFixedColumns, _.size(columns), columns)
+
+  // columnIndex in this function is an index in the original columns array.
   const renderHeaderCell = ({ key, columnIndex, rowIndex, style }) => {
     const field = columns[columnIndex].field
     return div({
@@ -442,6 +450,7 @@ export const GridTable = forwardRefWithName('GridTable', ({
     ])
   }
 
+  // columnIndex in this function is an index in the original columns array.
   const renderCell = ({ key, columnIndex, rowIndex, style }) => {
     return div({
       key,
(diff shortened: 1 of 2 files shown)


src/components/table.js
```diff
@@ -432,40 +480,76 @@ export const GridTable = forwardRefWithName('GridTable', ({
...
+                  columnIndex,
+                  rowIndex: 0,
+                  style: {
+                    height: rowDatum.size,
+                    left: fixedColumnOffsets[columnIndex] + horizontalOffsetAdjustment,
+                    position: 'fixed',
+                    top: rowDatum.offset + verticalOffsetAdjustment,
+                    width: fixedColumns[columnIndex].width,
+                    // Show header cell above body cells in fixed columns when vertically scrolling the grid.
+                    zIndex: 2
```
cahrens, line 535: Nit: below at line 616 use used 1 for zIndex. Can you make both values the same for consistency?

cahrens (approved):
Thank you for the very thorough comments, both in the PR and in the code. I really appreciate your attention to detail!

I tried out VoiceOver with the table, and there are definitely challenges to using the table. Some of what is announced doesn't make a lot of sense, and as far as I can tell it's pretty much impossible to trigger the edit icons. However, those are pre-existing issues, so definitely out of scope for this PR.

src/components/table.js
nawatts: The `zIndex` is higher on the fixed columns in the header row so that the header row is shown above the `zIndex: 1` cells in the table body when scrolling vertically. If they are the same, the body cell is shown since it's later in the DOM order.

Screen Shot 2022-06-21 at 2 20 36 PM

cahrens: Got it, thanks!

nawatts:
Thanks.

> as far as I can tell it's pretty much impossible to trigger the edit icons

Yes, they're only reachable via the keyboard in cells that contain another interactive element, such as a cell with a list value.

https://github.com/DataBiosphere/terra-ui/blob/95d4c3c07e3260e7b22992eb0c35f24e523d8a95/src/style.css#L83-L85

Created a few tickets for a11y issues in the data tables:
* [SU-142](https://broadworkbench.atlassian.net/browse/SU-142)
* [SU-143](https://broadworkbench.atlassian.net/browse/SU-143)

MatthewBemis (approved):

PR #3153, opened 2022-06-17
DC-399: Implement UI for recording user feedback on the new catalog view

pshapiro4broad:
Implement UI for recording user feedback on the new catalog view. The dialog performs two validations; it checks for feedback text between 1 and 2000 characters long, and a valid email, if present. If validation fails, the Submit button is disabled and the error text is shown as a tooltip on the submit button.

The UI mocks for the dialogs are here: https://broadworkbench.atlassian.net/browse/DC-372

To test this out, enable the new UI in the javascript console:
```
configOverridesStore.set({ isDataBrowserVisible: true })
```
And then open the Library / Data page, and enable the new view using the toggle button. The dialog is shown when the Feedback link is clicked.

The responses are recorded using google forms, although no google forms UI is used.

Form responses are collected here: https://docs.google.com/forms/d/1fE86o9KxCQxY_v73EO_8H_0eEOFYGDG8qKoXxxnt-YM/edit#responses


fboulnois (approved):
I checked the PR site, tested the functionality, and this lgtm

src/components/group-common.js
```diff
@@ -163,8 +163,7 @@ export const NewUserModal = ({
   const errors = validate({ userEmail }, { userEmail: { email: true } })
   const isAdmin = _.includes(adminLabel, roles)
 
-  const userEmailInvalid = !!validate({ userEmail }, { userEmail: { email: true } })
-  const canAdd = value => value !== userEmail || !userEmailInvalid
+  const canAdd = value => value !== userEmail || !errors
```
s-rubenstein, line 166: Just to note, this was an unrelated change, but in pairing we found some pretty unnecessary complexity and validation that could get removed and fixed on the way.

s-rubenstein (approved):
LGTM

src/pages/library/DataBrowserToggler.js
```diff
@@ -1,15 +1,83 @@
...
+  const submit = Utils.withBusyState(setSubmitting, async () => {
+    await Ajax().Surveys.submitForm('1FAIpQLSevEVLKiLNACAsti8k2U8EVKGHmQ4pJ8_643MfdY2lZEIusyw',
+      { 'entry.477992521': feedback, 'entry.82175827': contactEmail })
+    onSuccess()
+  })
+
+  const errors = validate({ feedback, contactEmail },
+    { feedback: { length: { minimum: 1, maximum: 2000 } }, contactEmail: contactEmail ? { email: true } : {} })
+
+  return h(Modal, {
```
breilly2, line 35: I suggest adding
```
    shouldCloseOnEsc: false,
```
I've encountered feedback forms that are too quick to dismiss from a seemingly non-destructive or accidental action, at which point any feedback that I've carefully crafted is lost. This makes me then either have two pieces of feedback or, more likely, not take the time to re-type my original feedback.

src/pages/library/DataBrowserToggler.js
```diff
@@ -1,15 +1,83 @@
...
+import validate from 'validate.js'
+
+
+const DataBrowserFeedbackModal = ({ onSuccess, onDismiss }) => {
+  const [contactEmail, setContactEmail] = useState('')
+  const [feedback, setFeedback] = useState('')
+  const [submitting, setSubmitting] = useState(false)
+
+  const submit = Utils.withBusyState(setSubmitting, async () => {
+    await Ajax().Surveys.submitForm('1FAIpQLSevEVLKiLNACAsti8k2U8EVKGHmQ4pJ8_643MfdY2lZEIusyw',
```
breilly2, line 27: Do we want to capture feedback from different environments in different sheets? Or maybe include the current env in the payload?

src/pages/library/DataBrowserToggler.js
```diff
@@ -1,15 +1,83 @@
...
 import * as Nav from 'src/libs/nav'
 import { useStore } from 'src/libs/react-utils'
 import { authStore } from 'src/libs/state'
+import * as Utils from 'src/libs/utils'
 import { catalogPreviewStore } from 'src/pages/library/dataBrowser-utils'
+import validate from 'validate.js'
+
+
+const DataBrowserFeedbackModal = ({ onSuccess, onDismiss }) => {
+  const [contactEmail, setContactEmail] = useState('')
```
breilly2, line 22: Could (probably should) pre-populate the user's contact email from their profile.

pshapiro4broad: We talked about how we didn't want to auto-submit the user's contact info to let people send feedback anonymously, which is why it's a field vs being sent along with the feedback. Also, the email being present is an extra step a user must take to indicate that it's OK to contact them for more information. Auto-filling the field would make it too easy for a user to leave their email in place IMO.

But I could be convinced otherwise, and it's easy enough to change.

src/pages/library/DataBrowserToggler.js
pshapiro4broad: That's good idea, although I don't think we need to worry about too much feedback from different places. It seems unlikely that we'll get feedback from dev that wouldn't be applicable for prod as well.

We did think about making the form elements configurable, and that way could be different in different environments, but it doesn't seem likely that this would be needed.

src/pages/library/DataBrowserToggler.js
pshapiro4broad: We did think about this and wanted to allow a focus loss click to dismiss the dialog, since it can be annoying when dialogs are hard to close. But if it's more typical in the terra-ui to not allow this, I can add these flags. Personally I don't like UIs where I can't use ESC to close a dialog.

breilly2: Yeah, I generally agree with that. In this case, I think the greater risk is not collecting the feedback and wasting a person's time than someone getting slightly annoyed that they have to activate the Cancel button to close the modal. Though now that I type that, I wonder if a11y has anything to say on this topic?

The other option would be to add a confirmation if someone tries to close the modal without submitting if they've started typing feedback. That might be the better solution. I made the suggestion I did because it's trivial to add.

src/pages/library/DataBrowserToggler.js
breilly2: Ah, I was assuming that it was a required field, which is why I wished it was pre-populated. I guess I didn't read the "Can we contact you with further questions?" prompt. 😅

src/pages/library/DataBrowserToggler.js
breilly2: To be clear, I speak from personal experience here. I once sent some feedback about the Gmail webapp. I had fully composed it and then explored some feature about taking a screenshot, I think, and dismissing that ended up closing the entire feedback form and throwing away what I'd typed. (You'd think that I'd have learned to _always_ compose outside of web forms by now. 🙄) I did end up sending two pieces of feedback that day, but I was _very_ unhappy about it.

breilly2 (approved):
👍 Based on responses to my comments, I don't have any reason to hold back approval. I'd be happy if the modal closing was tweaked a little, but I'm going to leave it up to you to decide what to do here. You could also ask the UX team what they think.

PR #1980, opened 2020-01-03
Remove remove-markdown to avoid regex DoS

breilly2:
<!--
Hello, friend!
Remember to mention what sort of testing/verification you performed in the course of building this PR, i.e. checking functionality in the browser locally.
Thanks!
--!>
Details here:
https://github.com/stiang/remove-markdown/issues/47
https://snyk.io/vuln/SNYK-JS-REMOVEMARKDOWN-73635

PR #670, opened 2018-09-18
Strip markdown from workspace cards

MattPutnam:
Fixes #656 

I couldn't find a single library that did both markdown stripping and rendering. There was a pair from the same developer that shared some dependencies, but they were harder to use (very imperative, would have had to write our own wrapper to make a Markdown component). This one's pretty small and has no dependencies of its own.

PR #3155, opened 2022-06-21
[SU-139] Render first paragraph of featured workspace descriptions

nawatts:
Currently, the [featured workspaces page](https://app.terra.bio/#library/showcase) displays the unrendered Markdown source for workspaces' description.

Screen Shot 2022-06-21 at 7 44 16 AM

Simply rendering the description isn't a great alternative though. The first line of many workspaces' description is a header repeating the workspace name.

Screen Shot 2022-06-21 at 7 45 22 AM

A decent compromise seems to be rendering only the first paragraph of the description.

Screen Shot 2022-06-21 at 7 45 44 AM

### DIFF
--- a/src/components/markdown.js
+++ b/src/components/markdown.js
@@ -41,6 +41,35 @@ export const MarkdownViewer = ({ children, renderers, ...props }) => {
   })
 }
 
+export const FirstParagraphMarkdownViewer = ({ children, renderers, ...props }) => {
+  let renderedFirstParagraph = false
+  return h(MarkdownViewer, {
+    ...props,
+    renderers: {
+      // See https://marked.js.org/using_pro#renderer for list of renderer methods.
+      blockquote: () => '',
+      checkbox: () => '',
+      code: () => '',
+      heading: () => '',
+      hr: () => '',
+      image: () => '',
+      list: () => '',
+      listitem: () => '',
+      paragraph: text => {
+        if (!renderedFirstParagraph) {
+          renderedFirstParagraph = true
+          return `<p>${text}</p>`
+        }
+        return ''
+      },
+      table: () => '',
+      tablerow: () => '',
+      tablecell: () => '',
+      ...renderers
+    }
+  }, [children])
+}
+
 export const newWindowLinkRenderer = (href, title, text) => {
   return `<a href="${href}" ${(title ? `title=${title}` : '')} target="_blank">${text}</a>`
 }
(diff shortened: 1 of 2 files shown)


src/components/markdown.js
```diff
@@ -41,6 +41,35 @@ export const MarkdownViewer = ({ children, renderers, ...props }) => {
   })
 }
 
+export const FirstParagraphMarkdownViewer = ({ children, renderers, ...props }) => {
+  let renderedFirstParagraph = false
+  return h(MarkdownViewer, {
```
breilly2, line 46: We should be a little careful here. Several years ago, I _think_ we used to use `marked` for these description previews (on the workspace list page), but we had to stop because `marked` would fail for some content in a way that made the page unusable. It was tricky to track down because it was content-dependent, and I don't know that we ever tracked down exactly what content was causing the problem.

nawatts: We're using `marked` to render workspace descriptions on the workspace dashboard.
https://github.com/DataBiosphere/terra-ui/blob/d20d3f1c48a3d5906e8179dc3a20d9338a93a432/src/pages/workspaces/workspace/Dashboard.js#L386

nawatts: Ah, looks like [AS-643](https://broadworkbench.atlassian.net/browse/AS-643) / #2732 could be the past issue? In which case, this should call `toString` on the description value.

nawatts: Going back farther, I also found #1980, which sounds similar to the issue described. But so far as I can tell, the workspaces list never rendered the Markdown. It looks like descriptions were added in #292 and #670 later attempted to remove the Markdown markup from them.

nawatts: Added a `toString` in [04fb237](https://github.com/DataBiosphere/terra-ui/pull/3155/commits/04fb237987a478ba9cfc675c3714fc320ad99e76).

MatthewBemis (approved):

src/components/markdown.js
petesantos: I believe it was #1980. At one point the workspaces list was a card view rather than a table view and each card would display the description.
IIRC there were certain markdown patterns that, when run through the `marked`, would result in running a regex in one of the dependencies or transitive dep. This would end up in a tight loop crashing the app.

edit: looking at the PR it looks like `remove-markdown` was causing the issue

nawatts: So far as I can tell, these descriptions have never been rendered in the workspaces list. #670 added [`remove-markdown`](https://www.npmjs.com/package/remove-markdown) to remove Markdown formatting (not render the Markdown). That was later removed in #1980 because of a [regular expression DoS issue](https://security.snyk.io/vuln/SNYK-JS-REMOVEMARKDOWN-73635).

`marked` is currently used to render these Markdown workspace descriptions on the workspace dashboard page.
https://github.com/DataBiosphere/terra-ui/blob/9892004cf19e735788a25c6fb8f6b35111639778/src/pages/workspaces/workspace/Dashboard.js#L386

petesantos (approved):

PR #3161, opened 2022-06-22
[SU-130][SU-132] Change behavior of cross data table search

nawatts:
Currently, when performing a cross data table search, the cross table search query takes precedence over the single table search query. The table search input is still enabled, but has no effect.

Also, when performing a cross table search, the user's column settings for the table are overridden and all columns are shown. This is not done when performing a single table search though.

This changes the single table search to take priority over the cross table search. The results shown in the table are always based on the single table search query. When a table is opened from cross table search results, the single table search query is initialized with the cross table search query. After that, it can be changed independently.

Since cross table searches are no longer handled differently by DataTable, this also removes the override on column settings. In the future, we can improve searches to show a notice if rows match the search query because of values in hidden columns.

Screen Shot 2022-06-22 at 8 26 21 AM

## Before
https://user-images.githubusercontent.com/1156625/175028405-702eba21-8219-4ec6-99e2-332c406b7eab.mov

## After
https://user-images.githubusercontent.com/1156625/175028437-ee43e6d2-f994-4940-83b1-0b122a1963e6.mov

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -79,7 +79,7 @@ const DataTable = props => {
   const [itemsPerPage, setItemsPerPage] = useState(stateHistory.itemsPerPage || 100)
   const [pageNumber, setPageNumber] = useState(stateHistory.pageNumber || 1)
   const [sort, setSort] = useState(stateHistory.sort || { field: 'name', direction: 'asc' })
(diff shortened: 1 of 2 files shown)


MatthewBemis (approved):

breilly2 (approved):
👍 I think I still wish that the two searches were additive, but this is an improvement.

PR #3163, opened 2022-06-22
[IA-3507] wait for workspaceName before listing disks and runtimes in leo

gpcarr:
This PR ensures that `workspaceName` is defined before making API calls to leo for listDisks and listRuntimes. This was causing a bug where listDisks was returning every disk a user had since workspaceName was not defined. It is only not defined on initial load. So now we wait for `workspaceName` to be defined before listing disks and runtimes

Users may have seen the error icon in the Cloud Environment widget as a result of this bug and had to navigate away and then come back for the error icon to go away. This makes it so the icon is not incorrectly displayed

### DIFF
--- a/src/pages/workspaces/workspace/WorkspaceContainer.js
+++ b/src/pages/workspaces/workspace/WorkspaceContainer.js
@@ -200,12 +200,15 @@ const WorkspaceAccessError = () => {
   ])
 }
 
-const useCloudEnvironmentPolling = (googleProject, workspaceNamespace, workspaceName) => {
+const useCloudEnvironmentPolling = (googleProject, workspace) => {
   const signal = useCancellation()
   const timeout = useRef()
   const [runtimes, setRuntimes] = useState()
   const [persistentDisks, setPersistentDisks] = useState()
   const [appDataDisks, setAppDataDisks] = useState()
+
+  const workspaceNamespace = workspace?.workspace.namespace
+  const workspaceName = workspace?.workspace.name
 
   const reschedule = ms => {
     clearTimeout(timeout.current)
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):
the bug is that `workspace` might've been undefined when `listDisks` and `listRuntimes` are called...the intended behavior is we only call these 2 APIs when `workspace` is defined...

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -200,23 +200,26 @@ const WorkspaceAccessError = () => {
...
-const useCloudEnvironmentPolling = (googleProject, workspaceNamespace, workspaceName) => {
+const useCloudEnvironmentPolling = (googleProject, workspace) => {
   const signal = useCancellation()
   const timeout = useRef()
   const [runtimes, setRuntimes] = useState()
   const [persistentDisks, setPersistentDisks] = useState()
   const [appDataDisks, setAppDataDisks] = useState()
 
+  const workspaceNamespace = workspace?.workspace.namespace
+  const workspaceName = workspace?.workspace.name
```
aweng98: Hi, I'm curious. Is this simpler? 
```
const saturnWorkspaceNamespace = workspace?.workspace.namespace
const saturnWorkspaceName = workspace?.workspace.name
```

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -200,23 +200,26 @@ const WorkspaceAccessError = () => {
...
     timeout.current = setTimeout(refreshRuntimesSilently, ms)
   }
   const load = async maybeStale => {
     try {
-      const [newDisks, newRuntimes] = await Promise.all([
+      const [newDisks, newRuntimes] = !!workspace ? await Promise.all([
         Ajax(signal).Disks.list({ googleProject, creator: getUser().email, includeLabels: 'saturnApplication,saturnWorkspaceName', saturnWorkspaceName: workspaceName }),
-        !!workspaceNamespace ? Ajax(signal).Runtimes.listV2({ creator: getUser().email, saturnWorkspaceNamespace: workspaceNamespace, saturnWorkspaceName: workspaceName }) : []
-      ])
+        Ajax(signal).Runtimes.listV2({ creator: getUser().email, saturnWorkspaceNamespace: workspaceNamespace, saturnWorkspaceName: workspaceName })
```
dmohs: Alex's suggestion above would allow this to be condensed:
```suggestion
        Ajax(signal).Runtimes.listV2({ creator: getUser().email, saturnWorkspaceNamespace, saturnWorkspaceName })
```

dmohs (approved):

src/pages/workspaces/workspace/WorkspaceContainer.js
gpcarr: I made that change and just pushed it! cc @aweng98 thanks for the suggestion

aweng98 (approved):

PR #3170, opened 2022-06-27
[IA-3420][IA-3526] fix bug and add workspaceId to details

jdcanas:
Fixes a bug where the page would not load due to cloudService not being defined for a runtime, and adds workspaceId to the details pop-up where relevant.

Unfortunately, leo only has this information for runtimes through runtimes created with the V2 API. For backwards compatibility, we must ask rawls for the workspaceId using runtime labels.

### DIFF
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -269,6 +269,7 @@ const Environments = () => {
   const renderDetailsApp = (app, disks) => {
     const { appName, diskName, googleProject, auditInfo: { creator } } = app
     const disk = _.find({ name: diskName }, disks)
+    // There's no current plans for workspaceId to be defined for apps
     return getDetailsPopup(appName, googleProject, disk, creator, undefined)
   }
 
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -175,8 +175,8 @@ const Environments = () => {
 
   const getCloudProvider = cloudEnvironment => Utils.cond(
     [isApp(cloudEnvironment), () => 'Kubernetes'],
-    [cloudEnvironment.runtimeConfig.cloudService === 'DATAPROC', () => 'Dataproc'],
-    [Utils.DEFAULT, () => cloudEnvironment.runtimeConfig.cloudService])
+    [cloudEnvironment?.runtimeConfig?.cloudService === 'DATAPROC', () => 'Dataproc'],
+    [Utils.DEFAULT, () => cloudEnvironment?.runtimeConfig?.cloudService])
 
   const getCloudEnvTool = cloudEnvironment => isApp(cloudEnvironment) ?
     _.capitalize(cloudEnvironment.appType) :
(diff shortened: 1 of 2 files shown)


src/pages/Environments.js
```diff
@@ -254,27 +254,29 @@ const Environments = () => {
...
     }, [h(Link, ['view'])])
   }
 
   const renderDetailsApp = (app, disks) => {
-    const { appName, diskName, googleProject } = app
+    const { appName, diskName, googleProject, auditInfo: { creator } } = app
     const disk = _.find({ name: diskName }, disks)
-    return getDetailsPopup(appName, googleProject, disk)
+    // There's no current plans for workspaceId to be defined for apps
+    return getDetailsPopup(appName, googleProject, disk, creator, undefined)
```
Qi77Qi: `workspaceId` is specific to workspace, so it should apply for apps as well, no?

jdcanas: Hmm... so currently we rely on leo to attach the workspaceId to the runtime, which is only done for azure VMs since we get this info through WSM. It should be possible to use the labels on the app/runtime (workspaceNamespace and workspaceName) to retrieve this ID from rawls UI-side though. Let me investigate...

Qi77Qi: I'm referring to the UUID for workspace, is that what you're thinking as well?

jdcanas: Yes. This information is optional from the leo response (and only populated for azure runtime currently). I now get this info from rawls directly instead of omitting it if leo hasn't returned it. Check it out now. It has a bit longer of a load time, but it seems worth it. 

https://github.com/DataBiosphere/terra-ui/pull/3170/files#diff-db07f65dd0ba5edbd535b2a231709c1f5901f4e1c43f39c333165d97c1abacc4R144

Qi77Qi (approved):
tried locally..lgtm

cahrens:
I am not able to load the Cloud Environments page on the PR site due to this error:

image

On Dev the page loads, and I can see the error by clicking on the error icon.

image

image

jdcanas:
> I am not able to load the Cloud Environments page on the PR site due to this error:
> 
> image
> 
> On Dev the page loads, and I can see the error by clicking on the error icon.
> 
> image
> 
> image

Hmm... I think this is a bug left over from early azure VMs. This is actually an error with getting the VM from leo. The [list](https://leonardo.dsde-dev.broadinstitute.org/#/runtimes/listRuntimesV2) call returns this VM in error state, but there was a bug when registering it in WSM. Leo can find the record via `list`, but when [get](https://leonardo.dsde-dev.broadinstitute.org/#/runtimes/getAzureRuntime) is called with additional filters and a WSM lookup is done, a 404 is thrown.

You can confirm if this is a  case if you also get a 404 when you try to delete it. 

Can you open the network tab and show me the URL from the request that is erroring? I'll need to clean up this entry in the dev db to recover your workspace from this error, then it should work fine.

cahrens (approved):

PR #3171, opened 2022-06-27
[AJ-481] update fixed columns logic to allow for empty tables in the UI

jordansafer:
In Terra UI, when uploading an empty table, the table view page crashes due to an error in the DataGrid. The problem is that the first 2 columns are set to "fixed", so the DataGrid works with 0 unfixed columns. The component can't handle this case, and crashes with an IndexOutOfBoundsException when trying to compute the width of the unfixed part of the table. This fix addresses that by only fixing 2 columns if we have more than 2.
<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -318,7 +318,7 @@ const DataTable = props => {
               initialX,
               initialY,
               sort,
-              numFixedColumns: visibleColumns.length > 2 ? 2 : 0,
+              numFixedColumns: visibleColumns.length > 0 ? 2 : 0,
               columns: [
                 {
                   width: 70,
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -306,6 +306,8 @@ const DataTable = props => {
       div({ style: { flex: 1 } }, [
         h(AutoSizer, [
           ({ width, height }) => {
+            const visibleColumns = _.filter('visible', columnSettings)
+
             return h(GridTable, {
               ref: table,
               'aria-label': `${entityType} data table, page ${pageNumber} of ${Math.ceil(totalRowCount / itemsPerPage)}`,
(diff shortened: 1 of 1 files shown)


MatthewBemis (approved):
Thanks for fixing this!

cahrens (approved):

src/components/data/DataTable.js
```diff
@@ -306,6 +306,8 @@ const DataTable = props => {
       div({ style: { flex: 1 } }, [
         h(AutoSizer, [
           ({ width, height }) => {
+            const visibleColumns = _.filter('visible', columnSettings)
```
cahrens, line 309: FYI @nawatts. This LGTM, but tagging you as the author of https://github.com/DataBiosphere/terra-ui/pull/3152/files.

src/components/data/DataTable.js
```diff
@@ -316,7 +318,7 @@ const DataTable = props => {
               initialX,
               initialY,
               sort,
-              numFixedColumns: 2,
+              numFixedColumns: visibleColumns.length > 2 ? 2 : 0,
```
nawatts:
```suggestion
              numFixedColumns: visibleColumns.length > 0 ? 2 : 0,
```

`visibleColumns` is based on `columnSettings` and does not include the first two columns (checkbox and row ID). As is, this would only fix the first two columns when there are more than 4 columns in the table.

nawatts (approved):
Ideally, I think GridTable should be resilient to this instead of relying on consumers to pass a combination of props that doesn't trigger the bug. This fixes the immediate issue though, so let's get this merged.

jordansafer:
> Ideally, I think GridTable should be resilient to this instead of relying on consumers to pass a combination of props that doesn't trigger the bug. This fixes the immediate issue though, so let's get this merged.

Agreed, could be worth a PR to GridTable

jordansafer:
Tests passed in CircleCI after retry- (syncing issue behind circleCI and github shows that tests were canceled)

https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/10375/workflows/7807d871-cbe0-46a6-9f57-3882ff27ae8c

PR #3173, opened 2022-06-27
[IA-3506] add rstudio test

jdcanas:
This test creates an rstudio runtime and runs a sample code snippet as a smoke test.

This test would be a good candidate for not running in PR tests, but running in the nightly regression suite. It is more of a feature test, and most of the CUJ is covered in `run-analysis`.

I ran test-flakes with 100 twice, and didn't receive any xPath related failures (though neither run had 100% pass rate due to noise around other workspace-related calls in the shared set-up code)

### DIFF
--- a/integration-tests/tests/run-rstudio.js
+++ b/integration-tests/tests/run-rstudio.js
@@ -19,7 +19,7 @@ const testRunRStudioFn = _.flow(
   // Create analysis file
   await click(page, clickable({ textContains: 'Start' }))
   await findElement(page, getAnimatedDrawer('Select an application'))
-  await click(page, image({ text: 'Create new R markdown file' }))
+  await click(page, image({ text: 'Create new R file' }))
   await fillIn(page, input({ placeholder: 'Enter a name' }), rFileName)
   await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Create Analysis' })) })
 
--- a/integration-tests/tests/run-rstudio.js
+++ b/integration-tests/tests/run-rstudio.js
@@ -48,8 +48,7 @@ const testRunRStudioFn = _.flow(
   await dismissNotifications(page)
   await click(page, clickable({ textContains: 'Open' }))
 
-  // This is code is duplicated, but will be deleted from the run-notebook test shortly.
-  // Find the iframe, wait until the Jupyter kernel is ready, and execute some code
+  // Find the iframe, wait until the RStudio iframe is loaded, and execute some code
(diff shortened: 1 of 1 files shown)


integration-tests/tests/run-rstudio.js
```diff
@@ -0,0 +1,66 @@
...
+  await click(page, clickable({ textContains: rFileName }))
+  await dismissNotifications(page)
+
+  await noSpinnersAfter(page, {
+    action: () => click(page, clickable({ textContains: 'Open' }))
+  })
+
+  //Create a cloud env from analysis launcher
+  await noSpinnersAfter(page, { action: () => click(page, clickable({ text: 'Create' })) })
+  await findElement(page, clickable({ textContains: 'RStudio Environment ( Creating )' }), { timeout: 40000 })
```
dmohs, line 44: One thing I've started doing on AoU tests is to create constants for these timeouts, e.g.,
```javascript
const RStudioCreationWaitMs = 40000
```
Then, my whole test timeout becomes something like:
```javascript
timeout: RStudioCreationWaitMs + otherWaitMs + [some buffer]
```

dmohs (approved):
Love the comments throughout the test to walk one through the steps.

integration-tests/tests/run-rstudio.js
jdcanas: Hmm... I'll keep this in mind for other tests that use the same/similar timeouts a lot, but there are only 3 custom timeouts and the purposes are fairly different, so I think keeping the numbers in the test helps readability

integration-tests/tests/run-rstudio.js
```diff
@@ -0,0 +1,66 @@
...
+  withBilling,
+  withRegisteredUser
+)(async ({ workspaceName, page, testUrl, token }) => {
+  await performAnalysisTabSetup(page, token, testUrl, workspaceName)
+
+  // Create analysis file
+  await click(page, clickable({ textContains: 'Start' }))
+  await findElement(page, getAnimatedDrawer('Select an application'))
+  await click(page, image({ text: 'Create new R markdown file' }))
+  await fillIn(page, input({ placeholder: 'Enter a name' }), rFileName)
```
gpcarr, line 23: Do we not have to select a `File Type`? 

Screen Shot 2022-06-29 at 4 55 50 PM

jdcanas: No, I just use the default value here. This isn't valid for file name since there is no default

gpcarr (approved):

PR #3174, opened 2022-06-27
[IA-3524] Fix filtering of older runtimes

gpcarr:
https://broadworkbench.atlassian.net/browse/IA-3524

older runtimes might not have `saturnWorkspaceNamespace` or `saturnWorkspacename` labels which is what we use in the UI to filter them per workspace. This PR adds logic to check if `googleProject` is not undefined and if not use it to filter runtimes for a workspace

Tested workspaces both pre and post PPW. everything is working as expected but there currently are no runtimes in dev that don't have either of the aforementioned labels - currently trying fudge dev data to create this scenario that exists in prod 

**This is currently impacting 974 runtimes in prod that don't have the `saturnWorkspaceNamespace` label and 636 runtimes in prod that don't have the `saturnWorkspacename` label**


src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -218,7 +218,7 @@ const useCloudEnvironmentPolling = (googleProject, workspace) => {
     try {
       const [newDisks, newRuntimes] = !!workspace ? await Promise.all([
         Ajax(signal).Disks.list({ googleProject, creator: getUser().email, includeLabels: 'saturnApplication,saturnWorkspaceName', saturnWorkspaceName }),
-        Ajax(signal).Runtimes.listV2({ creator: getUser().email, saturnWorkspaceNamespace, saturnWorkspaceName })
+        !!googleProject ? Ajax(signal).Runtimes.listV2({ googleProject, creator: getUser().email, saturnWorkspaceNamespace, saturnWorkspaceName }) : Ajax(signal).Runtimes.listV2({ creator: getUser().email, saturnWorkspaceNamespace, saturnWorkspaceName })
```
Qi77Qi: do we still need to pass `saturnWorkspaceNamespace, saturnWorkspaceName` in first case?


Also, can we do the same for listDisks as well

jdcanas: I dont really like how this reads... it uses a side effect to determine which information to filter on when we really want to include those filters if they are defined.

could we instead construct an object that is destructured into the list call with undefined values filtered?

Something like 
```
const labelFilters = _.omit(_.isUndefined,{ saturnWorkspaceNamespace, saturnWorkspaceName, googleProject })
Ajax(signal).Runtimes.listV2({creator: getUser().email, ...labelFilters })
```

(note, _.omit code syntax may not be right, but the idea is there. Took it from this stackoverflow thread, but it doesn't use fp lodash https://stackoverflow.com/questions/30812765/how-to-remove-undefined-and-null-values-from-an-object-using-lodash)

kyuksel:
@gpcarr I'd encourage you to keep this PR as a draft PR until the IA team signs off on it. This would help code-owner reviewers better judge when to look at it.

src/pages/workspaces/workspace/WorkspaceContainer.js
gpcarr: it's already in place for disks since `googleProject` is still be used [here](https://github.com/DataBiosphere/terra-ui/blob/dev/src/pages/workspaces/workspace/WorkspaceContainer.js#L220) (guess the UI work to support PDs in azure has not been done which makes sense since we haven't added support for PDs yet for azure)

gpcarr: > do we still need to pass saturnWorkspaceNamespace, saturnWorkspaceName in first case?

we might not need to pass these anymore, good point. i can do some testing

gpcarr: i really like that too, will update!

gpcarr: @jdcanas updated to compute `runtimeFilters` before making the ajax call. @Qi77Qi brings up a good point - if `googleProject` is present then we don't need workspaceName or workspaceNamespace. I think there's room for refactoring here to support Azure that we can follow up on once this bug is fixed. I do have one concern - runtimes exist without the `workspaceName` and `workspaceNamespace` labels and we can use `googleProject` however, these runtimes were probably created pre PPW which means that there could be multiple workspaces that share the same `googleProject`

jdcanas: not sure exactly what you mean... 

I think the approach here covers all cases, we simply include all filters that we can find. For azure runtimes, google project isn't defined, so this is handled. For pre-PPW workspaces that share the same google project, we use workspaceName+workspaceNamespace, which represent a unique identifier for a workspace together, so that case is covered as well.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -210,17 +210,19 @@ const useCloudEnvironmentPolling = (googleProject, workspace) => {
...
 
+  console.log(workspace)
+
   const reschedule = ms => {
     clearTimeout(timeout.current)
     timeout.current = setTimeout(refreshRuntimesSilently, ms)
   }
   const load = async maybeStale => {
     try {
+      const runtimeFilters = !!googleProject ? { googleProject } : {saturnWorkspaceNamespace, saturnWorkspaceName}
```
jdcanas: I'd like to use lodash to filter all undefined values if possible. Imagine a case where for some runtime a label isn't defined for some reason. Passing `x: undefined` as a filter to leo could cause the filter it to match no runtimes, as opposed to runtimes without those labels present.

This code should work, tested in the console
```
const runtimeFilters = _.pickBy(x => x !== undefined, 
    { googleProject, saturnWorkspaceName, saturnWorkspace, creator: getUser().email }
)

const [newDisks, newRuntimes] = !!workspace ? await Promise.all([
        Ajax(signal).Disks.list({ googleProject, creator: getUser().email, includeLabels: 'saturnApplication,saturnWorkspaceName', saturnWorkspaceName }),
        Ajax(signal).Runtimes.listV2(runtimeFilters)
 ]) : [[], []]
```

gpcarr: The one edge case I'm concerned about is if there are multiple runtimes (pre-PPW) that share a `googleProject` and don't have `saturnWorkspaceName` and `saturnWorkspaceNamespace` labels

Qi77Qi: there're about 296 google projects that have more than 1 runtime, and the runtimes don't have `saturnWorkspaceName`

Query
```
select count(*), cloudContext from CLUSTER LEFT JOIN `LABEL` as L on CLUSTER.id = L.`resourceId` where CLUSTER.runtimeName LIKE 'saturn%' and CLUSTER.status ="Stopped" and L.`resourceType`="runtime" and NOT EXISTS (select * from LABEL AS L2 where L2.key="saturnWorkspaceName" and L2.`resourceId`=L.resourceId and L2.`resourceType`="runtime") group by CLUSTER.cloudContext having count(*)>1 ;
```

Qi77Qi (approved):

Qi77Qi:
:shipit:

src/pages/workspaces/workspace/WorkspaceContainer.js
gpcarr: We talked about this offline and decided that 
```
const cloudEnvFilters = _.pickBy(l => !_.isUndefined(l),
        { googleProject, saturnWorkspaceName, saturnWorkspaceNamespace, creator: getUser().email }
      )
```
will account for both pre PPW workspaces and post PPW. This logic will make it so that whatever is defined in `{ googleProject, saturnWorkspaceName, saturnWorkspaceNamespace, creator: getUser().email }` will be used to filter runtimes and disks

gpcarr:
> @gpcarr I'd encourage you to keep this PR as a draft PR until the IA team signs off on it. This would help code-owner reviewers better judge when to look at it.

@kyuksel IA has signed off so I moved this back to Ready for Review!

jdcanas (approved):

breilly2 (approved):
👍 I didn't test, but the change seems reasonable to me.

PR #3175, opened 2022-06-27
AJ-433 filter ids to prevent duplication

calypsomatic:
[AJ-433](https://broadworkbench.atlassian.net/browse/AJ-433)
If a column of a snapshot-by-reference is also the id, then it will display twice in the table - once as the id, once as an attribute.  This filters out ids when loading the metadata.  Should only affect snapshots whose id is also a column.

### DIFF
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -478,7 +478,7 @@ const WorkspaceData = _.flow(
       setSnapshotDetails(_.set([snapshotName, 'error'], false))
       const entities = await Ajax(signal).Workspaces.workspace(namespace, name).snapshotEntityMetadata(googleProject, snapshotName)
       //Prevent duplicate id columns
-      const entitiesWithoutIds = _.mapValues(entity => _.update(['attributeNames'], _.without([entity.idName]), value), entities)
+      const entitiesWithoutIds = _.mapValues(entity => _.update(['attributeNames'], _.without([entity.idName]), entity), entities)
       setSnapshotDetails(_.set([snapshotName, 'entityMetadata'], entitiesWithoutIds))
     } catch (error) {
       reportError(`Error loading entities in snapshot ${snapshotName}`, error)
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -477,6 +477,10 @@ const WorkspaceData = _.flow(
     try {
       setSnapshotDetails(_.set([snapshotName, 'error'], false))
       const entities = await Ajax(signal).Workspaces.workspace(namespace, name).snapshotEntityMetadata(googleProject, snapshotName)
+      //Prevent duplicate id columns
+      Object.keys(entities).forEach( key => {
+        entities[key].attributeNames = entities[key].attributeNames.filter(attr => attr != entities[key].idName)
+      })
       setSnapshotDetails(_.set([snapshotName, 'entityMetadata'], entities))
     } catch (error) {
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -477,6 +477,10 @@ const WorkspaceData = _.flow(
     try {
       setSnapshotDetails(_.set([snapshotName, 'error'], false))
       const entities = await Ajax(signal).Workspaces.workspace(namespace, name).snapshotEntityMetadata(googleProject, snapshotName)
+      //Prevent duplicate id columns
+      Object.keys(entities).forEach(key => {
+        entities[key].attributeNames = entities[key].attributeNames.filter(attr => attr !== entities[key].idName)
+      })
```
kyuksel: Could this be

```suggestion
      const entitiesWithoutIds = _.pickBy(value => value.attributeNames !== value.idName, entities)
```

which could also be written as 

```suggestion
      const entitiesWithoutIds = _.pickBy(({attributeNames, idName }) => attributeNames !== idName, entities)
```

?

Note that the reason I suggest [pickBy](https://lodash.com/docs/4.17.15#pickBy) instead of [omitBy](https://lodash.com/docs/4.17.15#omitBy) is that `omit` is noted to be "considerably slower" than `pick` but this may be a bad extrapolation and/or premature optimization as I haven't tested either.

In general, we prefer Lodash FP for the following reasons:

- Lodash handles the case where you try to run the function on an incorrect type. e.g.: null, undefined, Object  (in the case of slice)
- Lodash FP functions support currying. This is especially useful for code changes later. If someone wants to add functionality, it makes it easy to do something like add the function to a flow.
- Lodash FP will not mutate the array. In the case of native slice, it also returns a new array. But there are other functions that can mutate (e.g. push/pop)
- For consistency within the codebase, consistency with how the code operates, and to prevent mutation we tend to prefer using Lodash functions.
- Mixing native and Lodash functions can lead to confusion about behavior.

calypsomatic: Absolutely!  I'm still a bit unfamiliar with Lodash, and here we're trying to filter an array that is a property of an object for every property in an object but I think I got a version that works.

src/pages/workspaces/workspace/Data.js
```diff
@@ -477,7 +477,9 @@ const WorkspaceData = _.flow(
     try {
       setSnapshotDetails(_.set([snapshotName, 'error'], false))
       const entities = await Ajax(signal).Workspaces.workspace(namespace, name).snapshotEntityMetadata(googleProject, snapshotName)
-      setSnapshotDetails(_.set([snapshotName, 'entityMetadata'], entities))
+      //Prevent duplicate id columns
+      const entitiesWithoutIds = _.mapValues(value => _.update(['attributeNames'], _.without([value.idName]), value), entities)
```
jordansafer:
```suggestion
      const entitiesWithoutIds = _.mapValues(entity => _.update(['attributeNames'], _.without([entity.idName]), value), entities)
```

small readability nit! could we name the "value" as "entity"?

jordansafer (approved):
LGTM :)

src/pages/workspaces/workspace/Data.js
kyuksel: Ah I see…we want to filter each `entity`'s `attributeNames` and not `entity`s themselves as a whole. Sorry for my misleading suggestion.

kyuksel (approved):

src/pages/workspaces/workspace/Data.js
calypsomatic: No problem!  I learned some good Lodash.

PR #3177, opened 2022-06-28
[QA-1932] Terra-UI testing: Handle NPS survey popup if found

aweng98:
[QA-1932](https://broadworkbench.atlassian.net/browse/QA-1932)

Unexpected NPS survey popup cause UI test to fail.

Screen Shot 2022-06-27 at 7 20 26 PM

### DIFF
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -200,6 +200,7 @@ const clickNavChildAndLoad = async (page, tab) => {
 }
 
 const viewWorkspaceDashboard = async (page, token, workspaceName) => {
+  // Sign in to handle unexpected NPS survey popup and Loading Terra... spinner
   await signIntoTerra(page, { token })
   await click(page, clickable({ textContains: 'View Workspaces' }))
   await dismissNotifications(page)
--- a/integration-tests/utils/integration-helpers.js
+++ b/integration-tests/utils/integration-helpers.js
@@ -200,8 +200,8 @@ const clickNavChildAndLoad = async (page, tab) => {
 }
 
 const viewWorkspaceDashboard = async (page, token, workspaceName) => {
-  await click(page, clickable({ textContains: 'View Workspaces' }))
   await signIntoTerra(page, { token })
+  await click(page, clickable({ textContains: 'View Workspaces' }))
   await dismissNotifications(page)
   await fillIn(page, input({ placeholder: 'SEARCH WORKSPACES' }), workspaceName)
   await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: workspaceName })) })
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -160,10 +160,8 @@ const select = async (page, labelContains, text) => {
   return click(page, `//div[starts-with(@id, "react-select-") and contains(normalize-space(.),"${text}")]`)
 }
 
-const waitForNoSpinners = async page => {
-  // Wait for id="root" element first to prevent checking Loading Terra... spinner too soon
-  await page.waitForXPath('//*[@id="root"]')
-  await page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
+const waitForNoSpinners = page => {
+  return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
 }
 
 // Puppeteer works by internally using MutationObserver. We are setting up the listener before
@@ -183,16 +181,14 @@ const delay = ms => {
 }
 
 const dismissNotifications = async page => {
-  const notificationCloseButtonXpath = '(//a | //*[@role="button"] | //button)[contains(@aria-label,"Dismiss") and not(contains(@aria-label,"error"))]'
-  // Don't sleep for 3 seconds if a notification was found already
-  await Promise.race([
-    delay(3000), // delayed for any alerts to show
-    page.waitForXPath(notificationCloseButtonXpath, { visible: true })
-  ])
-  const notificationCloseButtons = await page.$x(notificationCloseButtonXpath)
+  await delay(3000) // delayed for any alerts to show
+  const notificationCloseButtons = await page.$x(
+    '(//a | //*[@role="button"] | //button)[contains(@aria-label,"Dismiss") and not(contains(@aria-label,"error"))]')
+
   await Promise.all(
-    notificationCloseButtons.map(button => button.evaluateHandle(btn => btn.click(), button))
+    notificationCloseButtons.map(handle => handle.click())
   )
+
   return !!notificationCloseButtons.length && delay(1000) // delayed for alerts to animate off
 }
 
@@ -225,7 +221,7 @@ const signIntoTerra = async (page, { token, testUrl }) => {
   await page.waitForXPath('//*[contains(normalize-space(text()),"Loading Terra")][./*[contains(@src, "/loading-spinner.svg")]]', { hidden: true })
   await waitForNoSpinners(page)
 
-  await page.waitForFunction('window["forceSignIn"]', { polling: 100 })
+  await page.waitForFunction('!!window["forceSignIn"]')
   await page.evaluate(token => window.forceSignIn(token), token)
 
   await dismissNotifications(page)
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -218,12 +218,13 @@ const dismissNPSSurvey = async page => {
 const signIntoTerra = async (page, { token, testUrl }) => {
   if (!!testUrl) {
     await page.goto(testUrl, navOptionNetworkIdle())
-    // Wait for id="root" element first to prevent checking Loading Terra... spinner too soon
-    await page.waitForXPath('//*[@id="root"]')
-    await page.waitForXPath('//*[contains(normalize-space(text()),"Loading Terra")][./*[contains(@src, "/loading-spinner.svg")]]', { hidden: true })
   }
 
+  // Wait for id="root" element first to prevent checking Loading Terra... spinner too soon
+  await page.waitForXPath('//*[@id="root"]')
+  await page.waitForXPath('//*[contains(normalize-space(text()),"Loading Terra")][./*[contains(@src, "/loading-spinner.svg")]]', { hidden: true })
   await waitForNoSpinners(page)
+
   await page.waitForFunction('window["forceSignIn"]', { polling: 100 })
   await page.evaluate(token => window.forceSignIn(token), token)
 
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -160,8 +160,10 @@ const select = async (page, labelContains, text) => {
   return click(page, `//div[starts-with(@id, "react-select-") and contains(normalize-space(.),"${text}")]`)
 }
 
-const waitForNoSpinners = page => {
-  return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
+const waitForNoSpinners = async page => {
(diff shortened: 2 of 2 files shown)


integration-tests/utils/integration-helpers.js
```diff
@@ -200,8 +200,8 @@ const clickNavChildAndLoad = async (page, tab) => {
 }
 
 const viewWorkspaceDashboard = async (page, token, workspaceName) => {
-  await click(page, clickable({ textContains: 'View Workspaces' }))
   await signIntoTerra(page, { token })
+  await click(page, clickable({ textContains: 'View Workspaces' }))
```
aweng98, line 205: Call `signIntoTerra()` first to handle unexpected NPS survey popup.

integration-tests/utils/integration-utils.js
```diff
@@ -181,24 +183,52 @@ const delay = ms => {
...
+  let element
+  try {
+    element = await page.waitForXPath('//iframe[@aria-label="NPS Survey"]', { timeout: 1000 })
+  } catch (e) {
+    return // NPS survey was not found
+  }
+  try {
+    console.log('dismissing NPS survey')
+    const iframe = await element.contentFrame()
+    const [closeButton] = await iframe.$x('.//*[@id="ask-me-later"]') // Find "Ask Me Later" button to dismiss survey popup
```
aweng98: Tried to use text like this but unable to make it work. 
`await iframe.$x(clickable({text: "Ask Me Later"}))`.

integration-tests/utils/integration-helpers.js
cahrens, line 205: Was there a reason to switch this order?

integration-tests/utils/integration-utils.js
cahrens: Can you use the user-visible text instead of an ID?

integration-tests/utils/integration-helpers.js
aweng98: See my comment for this change.

aweng98: You're too fast! 🙂 I was still typing comments while you were reviewing it.

cahrens: Ah right, you are doing the dismissal within signIntoTerra. Can you add a comment in the code? I think the original idea was that "View Workspaces" was sort of an entryway into logging into Terra, so people might be tempted to swap the order without a comment.

integration-tests/utils/integration-utils.js
```diff
@@ -181,24 +183,53 @@ const delay = ms => {
...
+  const notificationCloseButtonXpath = '(//a | //*[@role="button"] | //button)[contains(@aria-label,"Dismiss") and not(contains(@aria-label,"error"))]'
+  // Don't sleep for 3 seconds if a notification was found already
+  await Promise.race([
+    delay(3000), // delayed for any alerts to show
+    page.waitForXPath(notificationCloseButtonXpath, { visible: true })
+  ])
+  const notificationCloseButtons = await page.$x(notificationCloseButtonXpath)
   await Promise.all(
-    notificationCloseButtons.map(handle => handle.click())
+    notificationCloseButtons.map(button => button.evaluateHandle(btn => btn.click(), button))
```
aweng98: Changed how click works here because `elementHandle.click()` uses element's (x, y) coordinates. If there are more than one notifications, after one notification is closed, position of next notification moves. So click by coordinates is bad idea here as closing additional notifications can unexpectedly fail.

nawatts:
Why do we even show the NPS survey on non-prod environments? Can it be turned off instead of adding code to handle it?

Also, it looks like there are a lot of changes in this PR unrelated to the NPS survey. What motivated those changes?

integration-tests/utils/integration-helpers.js
aweng98: Good idea. Added comment: `// Sign in to handle unexpected NPS survey popup and Loading Terra... spinner`

aweng98:
> Also, it looks like there are a lot of changes in this PR unrelated to the NPS survey.

"Unrelated" changes are `dismissNotifications()` and `waitForNoSpinners()`, right? I discovered them as flaky while I was testing new code for NPS survey. If you feel strongly about they don't belong here, I can remove them. I don't have a ticket for them alone. I thought that was going to be brought up as it is kind of "controversial".

aweng98:
> Why do we even show the NPS survey on non-prod environments? Can it be turned off instead of adding code to handle it?

That's a good question for PM and team that owns the functionality.

integration-tests/utils/integration-utils.js
nawatts: It seems like this could be simplified to
```js
notificationCloseButtons.map(button => button.evaluate(btn => btn.click()))
```
`evaluate` and `evaluateHandle` automatically pass the handle as the first argument to the function, so passing `button` again is unnecessary.

https://pptr.dev/#?product=Puppeteer&version=v15.1.1&show=api-elementhandleevaluatepagefunction-args

integration-tests/utils/integration-utils.js
```diff
@@ -181,24 +183,53 @@ const delay = ms => {
...
+    return // NPS survey was not found
+  }
+  try {
+    console.log('dismissing NPS survey')
+    const iframe = await element.contentFrame()
+    const [closeButton] = await iframe.$x('.//*[@id="ask-me-later"]') // Find "Ask Me Later" button to dismiss survey popup
+    await closeButton.evaluateHandle(button => button.click(), closeButton)
+    await delay(500) // delayed for survey to animate off
+  } catch (e) {
+    console.error(e)
```
nawatts, line 209: Why does this error need to be caught and logged here? Would it not be be logged by the `page.pageerror` handler attached in `enablePageLogging`?

integration-tests/utils/integration-utils.js
```diff
@@ -181,24 +183,53 @@ const delay = ms => {
...
+  try {
+    element = await page.waitForXPath('//iframe[@aria-label="NPS Survey"]', { timeout: 1000 })
+  } catch (e) {
+    return // NPS survey was not found
+  }
+  try {
+    console.log('dismissing NPS survey')
+    const iframe = await element.contentFrame()
+    const [closeButton] = await iframe.$x('.//*[@id="ask-me-later"]') // Find "Ask Me Later" button to dismiss survey popup
+    await closeButton.evaluateHandle(button => button.click(), closeButton)
```
nawatts: It looks like this could be simplified to
```js
await closeButton.evaluate(button => button.click())
```

`evaluate` and `evaluateHandle` automatically pass the handle as the first argument to the function. So there's no need to pass `closeButton` again.
https://pptr.dev/#?product=Puppeteer&version=v15.1.1&show=api-elementhandleevaluatepagefunction-args

integration-tests/utils/integration-utils.js
```diff
@@ -160,8 +160,10 @@ const select = async (page, labelContains, text) => {
   return click(page, `//div[starts-with(@id, "react-select-") and contains(normalize-space(.),"${text}")]`)
 }
 
-const waitForNoSpinners = page => {
-  return page.waitForXPath('//*[@data-icon="loadingSpinner"]', { hidden: true })
+const waitForNoSpinners = async page => {
+  // Wait for id="root" element first to prevent checking Loading Terra... spinner too soon
+  await page.waitForXPath('//*[@id="root"]')
```
nawatts: Why do we need to wait for the root element? The root element is part of the initial HTML response, so if the page is loaded, it should be present.

aweng98:
To avoid back-and-forth comments, I've reverted `dismissNotifications()` and `waitForNoSpinners()` changes.

integration-tests/utils/integration-utils.js
```diff
@@ -192,13 +192,40 @@ const dismissNotifications = async page => {
...
+
 const signIntoTerra = async (page, { token, testUrl }) => {
-  !!testUrl && await page.goto(testUrl, navOptionNetworkIdle())
-  await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout: 60 * 1000 })
+  if (!!testUrl) {
+    await page.goto(testUrl, navOptionNetworkIdle())
+  }
+
+  // Wait for id="root" element first to prevent checking Loading Terra... spinner too soon
+  await page.waitForXPath('//*[@id="root"]')
```
nawatts: Why do we need to wait the root element to be present? It's part of the HTML response, so if the page is loaded, it should be present.

integration-tests/utils/integration-utils.js
```diff
@@ -192,13 +192,40 @@ const dismissNotifications = async page => {
...
+  await page.waitForXPath('//*[contains(normalize-space(text()),"Loading Terra")][./*[contains(@src, "/loading-spinner.svg")]]', { hidden: true })
```
nawatts: Why change the selector here?

nawatts:
> "Unrelated" changes are dismissNotifications() and waitForNoSpinners(), right? I discovered them as flaky while I was testing new code for NPS survey. If you feel strongly about they don't belong here, I can remove them. I don't have a ticket for them alone. I thought that was going to be brought up as it is kind of "controversial".

I think it's fine include a few small semi-related changes, but they should be explained in the PR description or comments. Also, grouping multiple changes in one PR blocks merging any of those changes until all of them are approved. So if some change is likely to generate discussion, it's likely worth separating that into a smaller, focused PR.

integration-tests/utils/integration-utils.js
cahrens: That might be because the button is an iframe. Did you try `iframe.$x('.//*[normalize-space(.)="Ask Me Later"]')? In other words, build up the xPath using the text.

integration-tests/utils/integration-utils.js
aweng98: Nice! Done. I didn't know this before.

cahrens:
> > Why do we even show the NPS survey on non-prod environments? Can it be turned off instead of adding code to handle it?
> 
> That's a good question for PM and team that owns the functionality.

I posted in the UX channel to ask about this-- https://broadinstitute.slack.com/archives/C34GAKQ5R/p1656435309063549

integration-tests/utils/integration-utils.js
aweng98: Log here because it would be easy to find them in (middle of somewhere) test log file instead finding them at the end of log file in test Summary.
For the second question, I don't know if error would be caught by the `page.pageerror`.

aweng98 (commented):
Hi, Nick. I have reverted unrelated improvement in this PR.

integration-tests/utils/integration-utils.js
nawatts: Doesn't the `page.pageerror` handler log errors when they occur? Why would this message be in a different place in the logs than the "dismissing NPS survey" message?

aweng98: I think you're specifically talking about any error from clicking, right? I still don't know if `page.pageerror` will catch click error.

This catch block also catches error if the button in iframe cannot not be found. Other type of errors are not going to be captured by `page.pageerror`.

nawatts: Oh yeah, I'm getting mixed up. `page.pageerror` wouldn't catch this. Though I'm still not convinced there's a need to log this separately vs letting the usual error handling machinery log it.

nawatts (approved):

integration-tests/utils/integration-utils.js
aweng98: Sorry for the delayed response. Yes, that xpath worked! It took me some time to run tests repeatly to make sure proposed changes works.

cahrens (approved):
It seems that Joy might be turning off the NPS Survey, but I don't see harm in checking in the code to handle it when it comes back.

aweng98:
Thank you all.

PR #3178, opened 2022-06-28
[SU-136] Fix bug with GridTable fixed column positions

nawatts:
Noticed this bug while looking at #3171. Positions for fixed columns are calculated incorrectly.

This only manifests when there are more than 2 fixed columns in a GridTable, which is not currently the case anywhere in Terra.

Screenshots below show 4 fixed columns.

## Before
Screen Shot 2022-06-28 at 1 51 30 PM

## After
Screen Shot 2022-06-28 at 1 51 41 PM


MatthewBemis (approved):

dmohs (approved):
I love so much that you take the time to fix stuff like this.

PR #3180, opened 2022-06-28
[QA-1933] Terra UI test: Fix flaky workflow tests

aweng98:
[QA-1933](https://broadworkbench.atlassian.net/browse/QA-1933)

Failures caused by bad page synchronization.

### DIFF
--- a/integration-tests/tests/find-workflow.js
+++ b/integration-tests/tests/find-workflow.js
@@ -46,6 +46,7 @@ const testFindWorkflowFn = _.flow(
     backToTerra()
   ])
 
+  await page.waitForXPath('//*[@id="signInButton"]', { visible: true })
   await signIntoTerra(page, { token })
   await findText(page, `${workflowName}-configured`)
   await findText(page, 'inputs')
--- a/integration-tests/tests/find-workflow.js
+++ b/integration-tests/tests/find-workflow.js
@@ -39,6 +39,7 @@ const testFindWorkflowFn = _.flow(
     } else {
       await click(page, clickable({ textContains: 'Yes' }))
     }
+    await page.waitForXPath('//*[@id="signInButton"]', { visible: true })
   }
 
   await Promise.all([
@@ -46,7 +47,6 @@ const testFindWorkflowFn = _.flow(
     backToTerra()
   ])
 
-  await page.waitForXPath('//*[@id="signInButton"]', { visible: true })
   await signIntoTerra(page, { token })
   await findText(page, `${workflowName}-configured`)
   await findText(page, 'inputs')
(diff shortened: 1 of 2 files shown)


integration-tests/tests/find-workflow.js
```diff
@@ -46,6 +46,7 @@ const testFindWorkflowFn = _.flow(
     backToTerra()
   ])
 
+  await page.waitForXPath('//*[@id="signInButton"]', { visible: true })
```
aweng98: To confirm page is ready before calling `signIntoTerra()` function.

Encountered pageerror one time while running UI `find-workflow.js` test repeatly on localhost.

Screen Shot 2022-06-27 at 5 32 19 PM

integration-tests/tests/run-workflow-on-snapshot.js
```diff
@@ -66,6 +66,9 @@ const testRunWorkflowOnSnapshotFn = _.flow(
   await launchWorkflowAndWaitForSuccess(page)
 
   await clickNavChildAndLoad(page, 'data')
+  // Before click "Workspace Data" link, we need to make sure page has rendered all UI elements.
+  // "testsnapshot" link can take few seconds more to render on page.
+  await findElement(page, clickable({ textContains: 'testsnapshot' }))
```
aweng98, line 73: Test failed because click "Workspace Data" did not work. UI has not finished with loading the final element "testsnapshot". When `testsnapshot" element is finally displayed on page, elements which are below it have all shifted downward.
Screen Shot 2022-06-28 at 3 17 53 PM

davidangb (approved):
changes look fine to me, tests appear to be passing

jordansafer (approved):
Thank you for doing this!

aweng98:
Thank you for the quick review!

integration-tests/tests/run-workflow-on-snapshot.js
nawatts: It doesn't seem immediately obvious why we need `testsnapshot` to appear before clicking the "Workspace Data" button. Maybe worth a comment explaining that this is so that the "Workspace Data" button doesn't move when trying to click on it?

integration-tests/tests/find-workflow.js
nawatts: Why do we need to wait for the sign in button? `signIntoTerra` doesn't use it.

https://github.com/DataBiosphere/terra-ui/blob/261ed6e51c5ec71df3f9120c7f8cf64af2ae0b21/integration-tests/utils/integration-utils.js#L214-L225

aweng98: No, it doesn't. But the new code is not for `signIntoTerra` function. I use it to determine whether page (opened by `backToTerra` function) has finished rendering.

nawatts: It might be clearer to put it in the `backToTerra` function then.

nawatts (approved):

integration-tests/tests/run-workflow-on-snapshot.js
aweng98: Okay. I added more to my orginal comment.

integration-tests/tests/find-workflow.js
aweng98: Makes sense! Moved it to `backToTerra` func.

PR #3181, opened 2022-06-28
[QA-1935] Terra-UI test log: Capture page loading requests from ui server

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1935

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -361,7 +361,7 @@ const logPageConsoleMessages = page => {
   return () => page.off('console', handle)
 }
 
-const logPageAjaxResponses = page => {
+const logPageResponses = page => {
   const terraRequests = [
     'broad',
     'terra',
@@ -379,7 +379,7 @@ const logPageAjaxResponses = page => {
 }
 
 const enablePageLogging = page => {
-  logPageAjaxResponses(page)
+  logPageResponses(page)
   logPageConsoleMessages(page)
   logPageError(page)
   logError(page)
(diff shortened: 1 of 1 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -365,7 +365,8 @@ const logPageAjaxResponses = page => {
   const terraRequests = [
     'broad',
     'terra',
-    'googleapis'
+    'googleapis',
+    'bvdp'
```
aweng98, line 369: Missed it before, but we should also capture `bvdp-*` UI server requests and write them to test log.

kyuksel (approved):

integration-tests/utils/integration-utils.js
nawatts: What requests are made to `bvdp-*`? So far as I can tell, it's only the request to load `config.json`?

aweng98: Static files. Example screenshot for project: `bvdp-saturn-staging`.

Screen Shot 2022-06-30 at 10 51 51 AM

nawatts: Ah, I see. I was misled by the name of this function (`logPageAjaxResponses`). Requests for static files are not AJAX requests. But it looks like the `response` event is not specific to AJAX requests.
https://pptr.dev/#?product=Puppeteer&version=v13.5.1&show=api-event-response

nawatts (approved):
Maybe rename `logPageAjaxResponses` since this will now also log responses for non-AJAX requests for static files.

aweng98:
Good idea. Renamed function to `logPageResponses`.

PR #3182, opened 2022-06-29
DC-401: Get catalog UI ready for production release

tcjiang9:


package.json
```diff
@@ -28,7 +28,7 @@
     "jszip": "^3.7.1",
     "lodash": "^4.17.21",
     "marked": "^4.0.10",
-    "oidc-client-ts": "^2.0.4",
+    "oidc-client-ts": "^2.0.5",
```
nawatts: This looks unrelated to the ticket and other changes. What was the motivation for upgrading this library?

src/pages/library/Datasets.js
```diff
@@ -427,11 +429,24 @@ const topMed = () => h(Participant, {
   }, ['Browse Data'])
 ])
 
+const DatasetsAuthWrapper = () => {
```
nawatts: What is the purpose of this wrapper?

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -32,7 +32,7 @@ export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
       strong(['Pending']), `.`
     ]),
     div({ style: { marginTop: 30, display: 'flex', flexDirection: 'row', justifyContent: 'space-between', alignItems: 'center' } }, [
-      `Let's continue testing the BETA Data Catalog!`,
+      `Let's continue testing the New Data Catalog!`,
```
nawatts: Nit: should "New" be lowercase?

nawatts (changes requested):
Changes related to the ticket look mostly fine, but there are a lot of ESLint warnings that need to be addressed (run `yarn run lint` to see them). Upgrading the `oidc-client-ts` library seems unrelated to this ticket, so I think it should be removed from this PR.

src/pages/library/Datasets.js
tcjiang9: it's a bit of a hack to ensure the Datasets component rerenders when a new user login completes

package.json
tcjiang9: this was done automatically when I ran yarn install, I can undo it

src/pages/library/Datasets.js
nawatts: Ah, because `getLocalPref('catalog-toggle')` depends on the current user ID and needs to be rerun when the user changes?

I think this would be worth a comment since it's not obvious.

Also, I suspect `useStore(authStore)` could be placed in the `Datasets` component instead of a wrapper component. And the return value doesn't need to be stored in a variable.

nawatts: Then again, since Datasets is a child of Router, which is a child of AuthContainer and AuthContainer subscribes to changes in the auth store, I would expect this to re-render when the user changes even without subscribing to the auth store itself.

src/pages/library/Datasets.js
```diff
@@ -445,8 +461,8 @@ export const navPaths = [
   {
     name: 'library-datasets',
     path: '/library/datasets',
-    component: Datasets,
-    public: true,
+    component: DatasetsAuthWrapper,
+    public: false,
```
nawatts, line 476: If this page is going to require login, should it still appear in the sidebar when the user is not logged in?

Screen Shot 2022-06-29 at 2 48 48 PM

src/pages/library/Datasets.js
s-rubenstein: Oh! When we were pairing I missed that the Router will trigger a re-render when the user changes. That's great in this case!

We can't put `useStore(authStore)` in `Datasets`. I think we tried that first and it didn't work, which is why we moved to this approach. We wanted a complete re-instantiation of the component, rather than a partial re-rendering.

integration-tests/tests/preview-dataset.js
```diff
@@ -9,7 +9,7 @@ const datasetName = 'Readable Catalog Snapshot 1'
 const testPreviewDatasetFn = withUserToken(async ({ testUrl, page, token }) => {
   await enableDataCatalog(page, testUrl, token)
   await click(page, clickable({ textContains: 'datasets' }))
-  await click(page, clickable({ textContains: 'BETA Data Catalog OFF' }))
+  await click(page, clickable({ textContains: 'New Data Catalog OFF' }))
```
s-rubenstein: Is there a ticket in the backlog to remove the word `New` in all these places? I feel like we're gunna wanna make that change at some point.

integration-tests/utils/integration-helpers.js
```diff
@@ -185,7 +185,6 @@ const enableDataCatalog = async (page, testUrl, token) => {
   await waitForNoSpinners(page)
 
   await findText(page, 'Browse Data')
-  await overrideConfig(page, { isDataBrowserVisible: true })
```
s-rubenstein, line 188: I think these tests might still need a change to flip the switch to turn on the data catalog?

src/pages/library/DataBrowserToggler.js
```diff
@@ -97,22 +92,15 @@ export const DataBrowserPreviewToggler = ({ checked }) => {
...
           width: 55, height: 25,
-          onChange: () => {
-            catalogPreviewStore.set({ [id]: !checked })
-            if (checked) {
-              Nav.goToPath('library-datasets')
-            } else {
-              Nav.goToPath('library-browser')
-            }
-          }
+          onChange: value => onChange(value)
```
s-rubenstein: I know we changed to this here, but I think we might just be able to have this be `onChange`, and it should pick up value?

I tested in the console with the following set of commands:
```
const foo = (val) => console.log(val)
undefined
foo
(val) => console.log(val)
const obj = { foo }
undefined
obj.foo("a")
a
```

src/pages/library/Datasets.js
```diff
@@ -427,16 +429,30 @@ const topMed = () => h(Participant, {
...
 
+const DatasetsAuthWrapper = () => {
+  // unused variable to re-render Datasets
+  const authState = useStore(authStore)
+  return h(Datasets)
+}
 
 const Datasets = () => {
+  const [catalogShowing, setCatalogShowing] = useState(!!getLocalPref('catalog-toggle'))
+  console.log(`DATASETS CAT SHOWING ${catalogShowing}`)
```
s-rubenstein: We can remove the console log at this point.

package.json
tcjiang9: @nawatts 
looks like if I undo this, I get a circleci build failure with
➤ YN0056: │ oidc-client-ts@npm:2.0.5: Cache entry required but missing for oidc-client-ts@npm:2.0.5

src/pages/library/Datasets.js
tcjiang9: @nawatts looks like you're right here, it turned out to not be necessary, thanks!

src/pages/library/Datasets.js
tcjiang9: @nawatts, following this logic, should I remove the card for "Browse Data" from the front page also?

tcjiang9: the 'View Workspaces' card also requires login

src/pages/LandingPage.js
```diff
@@ -103,7 +101,7 @@ const LandingPage = () => {
       }, ['See this article']),
       ' for a summary of available resources.'
     ]),
-    isDataBrowserVisible() && div({
+    div({
```
pshapiro4broad: I don't know if we want to show this `div()` yet. It's a change to the terra landing page, and it's a banner that will direct users to the catalog view. I think we want to wait on this until we're ready to more publicly promote the new catalog view. If so, we'll need to keep a feature flag in place to hide this banner.

src/pages/library/DataBrowserToggler.js
```diff
@@ -97,22 +90,15 @@ export const DataBrowserPreviewToggler = ({ checked }) => {
...
         }),
-        div({ style: { marginLeft: 10 } }, [`BETA Data Catalog ${checked ? 'ON' : 'OFF'}`])
+        div({ style: { marginLeft: 10 } }, [`New Data Catalog ${catalogShowing ? 'ON' : 'OFF'}`])
```
pshapiro4broad: It looks like doesn't reflect the latest invision mocks. In the one I see, this text says `New Catalog OFF` or `New Catalog ON`. https://projects.invisionapp.com/d/main?origin=v7#/console/22047689/467561712/preview

src/pages/library/RequestDatasetAccessModal.js
```diff
@@ -21,7 +21,7 @@ export const RequestDatasetAccessModal = ({ onDismiss, datasets }) => {
   const [showWipModal, setShowWipModal] = useState(false)
 
   return showWipModal ? h(Modal, {
-    title: `Keep in mind we're still in BETA`,
+    title: `Keep in mind we're still in development`,
```
pshapiro4broad, line 24: This is fine although we don't have any immediate plans to add controlled access data to the catalog, and this dialog won't be seen until we do. At that point we'll probably revisit this UI as well.

integration-tests/utils/integration-helpers.js
tcjiang9: looks like  the test to flip the catalog is already performed in other tests
```
const testPreviewDatasetFn = withUserToken(async ({ testUrl, page, token }) => {
  await click(page, clickable({ textContains: 'datasets' }))
  await click(page, clickable({ textContains: 'New Data Catalog OFF' }))
  await waitForNoSpinners(page)

```

I think this test was just to test the config override and so can probably be deleted

src/pages/library/DataBrowserToggler.js
```diff
@@ -97,22 +90,15 @@ export const DataBrowserPreviewToggler = ({ checked }) => {
...
       ])
     ]),
-    checked && div({ style: { display: 'flex' } }, [
+    catalogShowing && div({ style: { display: 'flex' } }, [
       icon('talk-bubble', { size: 45, style: { marginLeft: '1.5rem', margin: '0 0.5rem' } }),
       div({ style: { display: 'flex', flexDirection: 'column' } }, [
         strong('Provide feedback'),
```
pshapiro4broad: The feedback text here has also been updated in the latest UI mocks https://projects.invisionapp.com/d/main?origin=v7#/console/22047689/467561713/preview

pshapiro4broad (commented):
The code looks OK, there are a few more UI changes that need to be added though.

src/pages/library/Datasets.js
nawatts: >  We can't put useStore(authStore) in Datasets. I think we tried that first and it didn't work, which is why we moved to this approach. We wanted a complete re-instantiation of the component, rather than a partial re-rendering.

Both updating component state and updating state in the wrapper component would re-render the component. If you want to create a new component instance, you can change the component's `key`. 
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key

src/pages/library/Datasets.js
nawatts: Yeah, on second thought, keeping the links in the sidebar and on the front page and then prompting to login when a user clicks the link makes sense.

src/pages/library/Datasets.js
s-rubenstein: My understanding was that state updates triggered a smart re-render, basically that react made a best effort guess on what should or should not re-render when state was updated, where as props triggered a re-render of the whole component. However, yeah, it wouldn't like… completely shut down and recreate the component from scratch.

Either way, this is fine as it is now.

nawatts: A change to either props or state causes the component to render. React then reconciles the DOM with the new element tree returned by that render.

https://reactjs.org/docs/faq-state.html#what-is-the-difference-between-state-and-props
https://reactjs.org/docs/reconciliation.html

src/libs/config.js
```diff
@@ -11,8 +11,7 @@ export const getConfig = () => {
 export const isAnvil = () => (window.location.hostname === 'anvil.terra.bio') || getConfig().isAnvil
 export const isAnalysisTabVisible = () => getConfig().isAnalysisTabVisible
 export const isCromwellAppVisible = () => getConfig().isCromwellAppVisible
-// configOverridesStore.set({ isDataBrowserVisible: true }) in browser console to enable
-export const isDataBrowserVisible = () => getConfig().isDataBrowserVisible
+export const isDataBrowserFrontPage = () => false
```
pshapiro4broad, line 14: If there isn't one already, can you make a ticket for the work to show the front page banner, or to remove it if we decide not to include it after all?

pshapiro4broad (approved):
LGTM!

src/pages/library/DataBrowserToggler.js
```diff
@@ -97,30 +89,23 @@ export const DataBrowserPreviewToggler = ({ checked }) => {
...
       icon('talk-bubble', { size: 45, style: { marginLeft: '1.5rem', margin: '0 0.5rem' } }),
       div({ style: { display: 'flex', flexDirection: 'column' } }, [
-        strong('Provide feedback'),
+        'Provide feedback or looking for a dataset not listed?',
         h(Link, {
           onClick: () => setFeedbackShowing(true),
           style: { display: 'block', marginTop: 10 }
         },
-        ['What do you think about the new Data Catalog?'])
+        ['Let us know!'])
```
nawatts, line 43: It would be nice to add an `aria-label` to this link since "Let us know!" isn't descriptive without the context of "Provide feedback or looking for a dataset not listed?".

src/pages/library/DataBrowserToggler.js
```diff
@@ -142,6 +127,10 @@ export const DataBrowserPreviewToggler = ({ checked }) => {
     [div({ style: { fontWeight: 600, fontSize: 18 } },
       'Thank you for helping us improve the Data Catalog experience!')]
     )
-  ])
+  ]), !catalogShowing ? div({
+    style: {
+      padding: '15px 15px', wordWrap: 'break-word', width: 300, lineHeight: 1.5
```
nawatts, line 56: Why limit the width of this and make the text wrap?

Screen Shot 2022-07-05 at 3 41 32 PM

nawatts (approved):

src/pages/library/DataBrowserToggler.js
tcjiang9: just to match our UI mock
https://projects.invisionapp.com/d/main?origin=v7#/console/22047689/467561712/preview

fboulnois (approved):
lgtm

PR #3185, opened 2022-06-29
[SUP-687] Rewrite column_grabber.svg to remove IDs

nawatts:
Some SVG files used for icons contain ID attributes. When one of those icons appears multiple times on the same page, axe warns about non-unique IDs: https://dequeuniversity.com/rules/axe/3.5/duplicate-id.

One instance of this occurs in the data table with the column grabber icon. column_grabber.svg contains two path elements that are referenced by ID by [use](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use) elements. Since each path is referenced only once, they can be inlined instead.

This also removes an unnecessary group element and `fill-rule` attributes.

Note: the `d` attribute of the two paths is duplicated. I suspect the intention here was to have one path element that was referenced twice.


breilly2 (approved):
👍 I'm not too familiar with SVGs, but I believe you that this works.

A few thinking-out-loud thoughts:
* We have many SVGs that have IDs: `find src -name '*.svg' |xargs grep -l 'id='`. I guess we haven't run into this problem before because those don't get used multiple times on a page.
* We have a `optimize-image-svgs` script in `package.json` that uses svgo to optimize SVG files. It only operates on `src/images`, not `src/icons`. We should probably update that script.
* It looks like `optimize-image-svgs` hasn't been run in a while.
* `optimize-image-svgs` does remove the `fill-rule` and some other things, but keeps the IDs. I wonder if it can be configured to avoid IDs.
* Looking at the output of `optimize-image-svgs` is part of what convinced me that the change in the PR is good.

nawatts:
> We have many SVGs that have IDs: find src -name '*.svg' |xargs grep -l 'id='. I guess we haven't run into this problem before because those don't get used multiple times on a page.

Also, I don't think we've really looked for this problem much in the past. This was identified by [axe DevTools](https://www.deque.com/axe/). I removed IDs from a few other icon SVGs in #3186.

I haven't looked at `optimize-image-svgs`, but one thing to note about the images vs icons directories is the difference in the way they're included in the page. I think SVGs in images tend to be used as the src for `img` tags, while SVGs in icons are converted to React components by some Babel transform or webpack loader. Since they do go through webpack/Babel, I'm curious if `optimize-image-svgs` is necessary at all.

kyuksel:
I wonder if running `optimize-image-svgs` on _images_ (and _icons_ if applicable) could be part of the CI flow. It's difficult to remember during PR authoring and reviewing.

cahrens (approved):

PR #3187, opened 2022-06-29
[SU-150] Fix rendering lists of URLs in data tables

nawatts:
Currently, lists of GCS URLs are not rendered correctly in data tables. The data table cell shows the filename of the last URL in the list and the link URL is the comma delimited concatenation of all URLs in the list. The root of the issue is that DataTable passes the result of `Utils.entityAttributeText` to `renderDataCell`, so `renderDataCell` (which looks like it was written to accept an attribute value) may receive a stringified version of the attribute value.

https://github.com/DataBiosphere/terra-ui/blob/d4533d8e90a2bc1f947da99c9bf2339222b32191/src/components/data/DataTable.js#L407

https://github.com/DataBiosphere/terra-ui/blob/d4533d8e90a2bc1f947da99c9bf2339222b32191/src/components/data/data-utils.js#L65-L87

This fixes rendering of lists of URLs. Since there are many different attribute types that need to be rendered by the data table, this also adds some unit tests for `renderDataCell` and adds [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) to support that.

Note for reviewers: unit tests were added in e7b01d54cc948c2d802cc27705d49524fc6edacf before making changes to `renderDataCell`. After those changes, `renderDataCell` doesn't output exactly the same thing in all cases, but the changes are minor and can be seen in 608d898036f0630209c679c972a1ede086a6a9e6.

## Before
Screen Shot 2022-06-29 at 5 02 41 PM
Screen Shot 2022-06-29 at 5 02 53 PM

## After
Screen Shot 2022-06-29 at 5 02 19 PM
Screen Shot 2022-06-29 at 5 02 24 PM


src/components/data/_tests/data-utils.test.js
```diff
@@ -95,6 +99,93 @@ describe('getAttributeType', () => {
   })
 })
 
+describe('renderDataCell', () => {
```
cahrens, line 102: WooHoo, first unit tests with rendering!! Awesome @nawatts!

MatthewBemis (approved):

nawatts:
Rebased to resolve merge conflicts in yarn.lock.

petesantos (approved):
I had some questions around how this would behave with various data types - the unit tests put my mind at ease - thanks!

PR #3188, opened 2022-06-29
[DDO-2155] GHA to push a Docker image to GAR

em-may:


Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Bundle app source
+COPY . .
+
+# Install deps
+RUN yarn install --immutable-cache
+RUN npm install --location=global serve # to serve the app
+
+# Run the linter
+RUN yarn eslint --max-warnings=0 .
```
nawatts: Do we need to run ESLint in the Docker build? This seems like something that should be checked before the image is built, similar to tests.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Build the app
+RUN DISABLE_ESLINT_PLUGIN=true yarn build # already linted above
+
+# Copy app config to root dir
+COPY config/emay.json build/config.json
+
+EXPOSE 8080
+
+CMD npx serve build -l 8080
```
nawatts: Since terra-ui uses client side routing, index.html needs to served for all URLs that match client side routes. The App Engine deployment does this by responding with index.html for all URLs that don't match a static file.

https://github.com/DataBiosphere/terra-ui/blob/d4533d8e90a2bc1f947da99c9bf2339222b32191/app.yaml#L21-L22

Does this handle such URLs? It looks like we may need to configure some rewrites for `serve`.
https://github.com/vercel/serve-handler#options

.github/workflows/build.yaml
```diff
@@ -0,0 +1,105 @@
...
+# message.
+#
+# When run on a PR, it simulates bumping the tag and appends a hash to the pushed image.
+#
+# The workflow relies on github secrets:
+# - BROADBOT_TOKEN - the broadbot token, so we can avoid two reviewer rule on GHA operations
+on:
+  pull_request:
+    paths-ignore:
+      - 'README.md'
```
nawatts: Should changes in the `integration-tests` folder also be ignored?

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN npm install --location=global serve # to serve the app
+
+# Run the linter
+RUN yarn eslint --max-warnings=0 .
+
+# Build the app
+RUN DISABLE_ESLINT_PLUGIN=true yarn build # already linted above
+
+# Copy app config to root dir
+COPY config/emay.json build/config.json
```
kyuksel: Should `emay` be replaced with something non-personal?

config/emay.json
```diff
@@ -0,0 +1,25 @@
...
+  "marthaUrlRoot": "https://martha.fiab-emay-moving-deer.bee.envs-terra.bio",
+  "orchestrationUrlRoot": "https://firecloudorch.fiab-emay-moving-deer.bee.envs-terra.bio",
+  "rawlsUrlRoot": "https://rawls.fiab-emay-moving-deer.bee.envs-terra.bio",
+  "rexUrlRoot": "https://terra-rex-dev.appspot.com",
+  "samUrlRoot": "https://sam.fiab-emay-moving-deer.bee.envs-terra.bio",
+  "shibbolethUrlRoot": "https://broad-shibboleth-prod.appspot.com/dev",
+  "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
+  "alphaRegionalityGroup": "Alpha_Regionality_Users",
+  "terraDockerVersionsFile": "terra-docker-versions-dev.json",
+  "terraDockerImageBucket": "terra-docker-image-documentation-dev"
```
kyuksel: Should `emay` references in `UrlRoot`s be replaced?

kyuksel (commented):
A couple of comments inline + echoing Nick's comments

Dockerfile
em-may: There's not technical requirement to do it (it still happens in CircleCI), I was just including it to be as close to the CircleCI as possible. Will remove.

.github/workflows/build.yaml
em-may: I can definitely add that, as well as any other resources that y'all want ignored.

Dockerfile
em-may: Definitely, it's there for testing and will be removed soon. It will be replaced by a k8s ConfigMap in the helm chart that is populated with the correct vars for each env (see: BEE).

config/emay.json
em-may: See above comment.

Dockerfile
em-may: This might explain the 404 I get after passing OAuth. Will try this, thanks!

Dockerfile
kyuksel: Gotcha, thanks. FYI that we typically open Terra UI PRs in draft status and keep them that way until testing is done. This helps with the automated PR review triaging (as Terra UI is not owned by a particular team) 😊

em-may: Done! Will reopen when ready.

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN yarn install --immutable-cache
+
+# Build the app.
+RUN yarn build
+
+# App port forwarding.
+EXPOSE 8080
+
+# Use yarn to host the app.
+CMD yarn start
```
nawatts: I'm not sure where this image will be used, but `yarn start` serves the app using [webpack-dev-server](https://github.com/webpack/webpack-dev-server), which is recommended to only be used for development.

It also serves the app with React in development mode instead of using the production build created by `yarn build`.

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM node:16
+
+RUN set -x \
+  && git clone https://github.com/DataBiosphere/terra-ui.git \
+  && cd terra-ui \
+  && git checkout dev \
```
nawatts: Should this image be built from the local copy of terra-ui instead of downloading from GitHub?

Would it not defeat the purpose of a BEE if the terra-ui service is always built from the tip of the `dev` branch instead of a specified revision?

.github/workflows/tag-publish.yaml
```diff
@@ -0,0 +1,114 @@
...
+env:
+  GOOGLE_PROJECT: dsp-artifact-registry
+  # Name of the app-specific Docker repository configured in GOOGLE_PROJECT
+  REPOSITORY_NAME: ${{ github.event.repository.name }}
+  # Region-specific Google Docker repository where GOOGLE_PROJECT/REPOSITORY_NAME can be found
+  GOOGLE_DOCKER_REPOSITORY: us-central1-docker.pkg.dev
+
+jobs:
+  tag-push-job:
+    if: startsWith(github.ref, 'refs/tags/')
```
jack-r-warren: I don't think we want this, we want to run this action upon every commit to dev and every PR branch commit

.github/workflows/tag-publish.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Add latest tag to Docker image
+        if: github.event_name != 'pull_request'
+        shell: bash
+        run: |
+          gcloud artifacts docker tags add \
+            "${{ steps.image-name.outputs.tagged }}" \
+            "${{ steps.image-name.outputs.name }}:latest"
+
+      - name: Deploy to Terra Dev environment
+        uses: broadinstitute/repository-dispatch@master
```
jack-r-warren, line 108:
```suggestion
        uses: broadinstitute/repository-dispatch@master
        if: github.event_name != 'pull_request'
```
We don't want to update dev.yaml upon PR builds

.github/workflows/tag-publish.yaml
```diff
@@ -0,0 +1,114 @@
+name: Bump, Tag, and Publish
```
jack-r-warren: I'd actually change this to say "Generate BEE Build" or something, because it doesn't do any real publishing (and the UI isn't actually really versioned by the bumper tags, either)

Dockerfile
em-may: Good catch, thanks for this. Fixed!

knapii-developments (approved):

nawatts:
The currently builds an image that does not contain a `config.json` file. I think that because [`public/config.json`](https://github.com/DataBiosphere/terra-ui/blob/dev/public/config.json) in git is symlinked to `config/dev.json`, it does not get copied between the stages of the Docker build. 

`config.json` is the first thing that Terra UI loads and it tells Terra UI where to send all API requests. When running in a BEE, I assume that we would mount a configuration file specific to the BEE. Is it worth baking a default configuration into the image?

nawatts (approved):

em-may:
@nawatts You are correct that a BEE-specific config is created and mounted. The config.json file is generated [here](https://github.com/broadinstitute/terra-helmfile/blob/24c84b6f5eaef18603da7b74486a0e40a5e85c5d/charts/terraui/templates/configmap.yaml#L8) using values from [here](https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/terraui/bee.yaml.gotmpl). Since this file is specific to each BEE and we don't have a unified environment like the live envs, we cannot create a default config, that is, unless for some reason we want to point everything at dev by default. However, since that could give false positives and potentially create confusion, I think it's best to rely on per-BEE config files generated as per above.

PR #3189, opened 2022-06-30
[TOAZ-146] [B2C] Use separate policy for sensitive scopes from Terra UI

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-146

Handle billing scope elevation for Google logins through B2C. Note there should be no change in behavior for direct Google logins as a result of this PR.

For B2C we have a similar flow: initially only request minimal Google scopes; if the user navigates to the billing page we request the `cloud-billing` scope via a custom policy.

I tested against a fiab with both B2C enabled and disabled, and it seems to work. For reviewers: please test signin/signout flow and the billing page works as expected.

Also attempted to fix an issue @nawatts reported where we get spurious error notifications if the token expires, even when the refresh token handshake works.

### DIFF
--- a/config/alpha.json
+++ b/config/alpha.json
@@ -26,5 +26,6 @@
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
   "alphaRegionalityGroup": "Alpha_Regionality_Users",
   "terraDockerVersionsFile": "terra-docker-versions-alpha.json",
-  "terraDockerImageBucket": "terra-docker-image-documentation-alpha"
+  "terraDockerImageBucket": "terra-docker-image-documentation-alpha",
+  "b2cBillingPolicy": "b2c_1a_signup_signin_billing_alpha"
 }
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -82,19 +82,24 @@ const revokeTokens = async () => {
 }
 
 const getSigninArgs = includeBillingScope => {
-  return includeBillingScope === true ? {
-    scope: `
-    openid
-    email
-    profile
-    https://www.googleapis.com/auth/cloud-billing
-  `.trim().split(/\s+/).join(' ')
-  } : {}
+  return Utils.cond(
+    [includeBillingScope === false, () => {}],
+    // If the login authority is Google, we can just append the scope to the signin args.
+    [isGoogleAuthority(), () => ({ scope: 'openid email profile https://www.googleapis.com/auth/cloud-billing' })],
+    // If the login authority is B2C, switch to a dedicated policy endpoint configured for
+    // the GCP cloud-billing scope.
+    () => ({
+      extraQueryParams: { access_type: 'offline', p: getConfig().b2cBillingPolicy },
+      extraTokenParams: { p: getConfig().b2cBillingPolicy }
+    })
+  )
 }
 
 export const signIn = (includeBillingScope = false) => {
   const args = getSigninArgs(includeBillingScope)
-  return getAuthInstance().signinPopup(args).catch(() => false)
+  return getAuthInstance().signinPopup(args)
+    .then(user => processUser(user, true, includeBillingScope))
+    .catch(() => false)
 }
 
 export const reloadAuthToken = (includeBillingScope = false) => {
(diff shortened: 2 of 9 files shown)


src/libs/auth.js
```diff
@@ -82,19 +82,30 @@ const revokeTokens = async () => {
...
-    profile
-    https://www.googleapis.com/auth/cloud-billing
-  `.trim().split(/\s+/).join(' ')
-  } : {}
+  return Utils.cond(
+    [includeBillingScope === false, () => {}],
+    // For Google just append the scope to the signin args.
+    [isGoogleAuthority(), () => ({ scope: 'openid email profile https://www.googleapis.com/auth/cloud-billing' })],
+    // For B2C switch to a dedicated policy endpoint configured for the GCP cloud-billing scope.
+    () => ({
```
rtitle, line 90: @nmalfroy not sure if TDR is doing this yet -- but fyi I was able to dynamically switch the B2C policy by setting these params on the login request.

config/alpha.json
```diff
@@ -26,5 +26,6 @@
   "alphaSpendReportGroup": "Alpha_Spend_Report_Users",
   "alphaRegionalityGroup": "Alpha_Regionality_Users",
   "terraDockerVersionsFile": "terra-docker-versions-alpha.json",
-  "terraDockerImageBucket": "terra-docker-image-documentation-alpha"
+  "terraDockerImageBucket": "terra-docker-image-documentation-alpha",
+  "b2cBillingPolicy": "b2c_1a_signup_signin_billing_alpha"
```
rtitle, line 30: These policies exist in all non-prod envs as a result of https://github.com/broadinstitute/terraform-ap-deployments/pull/707

src/libs/auth.js
```diff
@@ -82,19 +82,30 @@ const revokeTokens = async () => {
...
-export const signIn = (includeBillingScope = false) => {
+export const signIn = async (includeBillingScope = false) => {
   const args = getSigninArgs(includeBillingScope)
-  return getAuthInstance().signinPopup(args).catch(() => false)
+  const user = await getAuthInstance().signinPopup(args)
+
+  // For B2C record in the auth store whether we requested the GCP cloud-billing scope since there
+  // is no way to determine it after the fact.
+  // For Google we don't need to do this since we can inspect the scope directly in the user object.
+  if (!isGoogleAuthority()) {
```
nawatts, line 104: Would it be worth storing a flag here for Google as well as B2C? Then `hasBillingScope` would not need to be special cased.

src/libs/auth.js
```diff
@@ -82,19 +82,30 @@ const revokeTokens = async () => {
...
-  return includeBillingScope === true ? {
-    scope: `
-    openid
-    email
-    profile
-    https://www.googleapis.com/auth/cloud-billing
-  `.trim().split(/\s+/).join(' ')
-  } : {}
+  return Utils.cond(
+    [includeBillingScope === false, () => {}],
```
nawatts: Is this intended to return an empty object? As is, it looks like it returns undefined (the function body is empty).

rtitle: Oh I meant to return empty object (though I guess undefined works..). Will fix

src/libs/auth.js
rtitle: I actually thought about this and tried it. I decided against it because Google does some magic -- sometimes even if I don't initially sign in with the billing scope I have it _anyway_ (maybe from a prior session). So I thought it was better to test the actual value from Google.

For B2C there is no magic because they are distinct signin endpoints. So the state value is pretty reliable. 

(Btw it's probably possible to make B2C pass the Google scopes as a claim so the client could inspect them directly instead of using our own state -- but this works well enough I think.)

src/libs/auth.js
nmalfroy: Cool.  We don't do any scope switching but this might be useful to do in the near future

nmalfroy: that's interesting...is the `p` parameter a documented one or did you have to figure that one out

nmalfroy (approved):
lgtm!

nawatts:
Not sure if it's something specific to my account, but when I try to log in (by clicking the log in button in the sidebar) with this branch, I get "This app is blocked. This app tried to access sensitive info in your Google Account. To keep your account safe, Google blocked this access."

src/libs/auth.js
rtitle: `p` is the B2C policy (e.g. https://terradevb2c.b2clogin.com/terradevb2c.onmicrosoft.com/v2.0/.well-known/openid-configuration?p=b2c_1a_signup_signin_billing)

These options just append extra querystring params to the authorize/token endpoints. (I also updated the backend routes to handle incoming requests with `p` correctly.)

nawatts:
Ah, looks like my issue is specific to which log in button is clicked. I think the one in the sidebar probably needs to be updated to call `signIn(false)`.

https://github.com/DataBiosphere/terra-ui/blob/bf29a934aa97a70c58246085cedbc7b12149beea/src/components/TopBar.js#L190

rtitle:
Ohh! Good catch, will fix.
edit: done

nawatts (approved):

PR #3190, opened 2022-06-30
[TUIM-14][TUIM-17] Update node-fetch and protobufjs

breilly2:
https://broadworkbench.atlassian.net/browse/TUIM-14
https://broadworkbench.atlassian.net/browse/TUIM-17

Followed instructions at https://github.com/DataBiosphere/terra-ui/wiki/Security-and-Maintenance#dependabot-security-updates


nawatts:
For what it's worth, we're already using the patched versions of `node-fetch` and `protobufjs` referenced in those alerts. Those alerts are based on `integration-tests/yarn.lock`, which no longer exists.

yarn.lock
```diff
@@ -9689,7 +9674,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"node-fetch@npm:2.6.7, node-fetch@npm:^2.6.1, node-fetch@npm:^2.6.6, node-fetch@npm:^2.6.7":
+"node-fetch@npm:2.6.7, node-fetch@npm:^2.6.1, node-fetch@npm:^2.6.7":
```
kyuksel, line 9677: Do we need `node-fetch@npm:^2.6.1` to go away as well?

kyuksel:
> For what it's worth, we're already using the patched versions of `node-fetch` and `protobufjs` referenced in those alerts. Those alerts are based on `integration-tests/yarn.lock`, which no longer exists.

Too bad [it doesn't appear to be possible to exclude modules from being scanned for Dependabot alerts](https://github.com/DataBiosphere/terra-ui/pull/3190#issuecomment-1171376178).

EDIT: Reading the quoted comment again, we shouldn't even need to exclude anything since we removed `integration-tests/yarn.lock` a while ago.

breilly2:
> For what it's worth, we're already using the patched versions of node-fetch and protobufjs referenced in those alerts. Those alerts are based on integration-tests/yarn.lock, which no longer exists.

I can't say anything definite about protobufjs, but `integration-tests/package.json` explicitly referenced node-fetch 2.6.6. Maybe it somehow still resolved to 2.6.7 in practice, but I couldn't prove that to myself.

My trust in dependabot is pretty shaken now, so I'm not sure how to proceed. It seems like there's some hidden state, such as some cached `yarn.lock` files ~and some memory of alerts that have been dismissed~. EDIT: I found the record of alerts, which allows you to filter on dismissal reason: https://github.com/DataBiosphere/terra-ui/security/dependabot?q=is%3Aclosed.

yarn.lock
breilly2: I _think_ this line is specifying the versions as they're requested by various project dependencies (direct or indirect) that will be resolved to this specific version of the dependency. In theory, I think this would allow a block like:
```
"node-fetch@npm:^3.0.0":
  version: 3.0.1
  resolution: "node-fetch@npm:3.0.1"
  ...
```
to coexist in the same project, and each library could get the major version of node-fetch that it wants.

nawatts:
> I can't say anything definite about protobufjs, but integration-tests/package.json explicitly referenced node-fetch 2.6.6. Maybe it somehow still resolved to 2.6.7 in practice, but I couldn't prove that to myself.

It specifies `^2.6.6`, which is equivalent to `>=2.6.6 <3.0.0`.

https://github.com/DataBiosphere/terra-ui/blob/a8bd12430d0f16725622100e8f228858e348c562/integration-tests/package.json#L15

https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004

In yarn.lock, that is resolved to `2.6.7`.

https://github.com/DataBiosphere/terra-ui/blob/a8bd12430d0f16725622100e8f228858e348c562/yarn.lock#L9692-L9704

breilly2:
Ah, of course. That makes sense based on my response to Kaan's question.

Well, I guess I still want to merge this, even if just as an exercise in tidying up.

kyuksel:
> Well, I guess I still want to merge this, even if just as an exercise in tidying up.

Please do.

kyuksel (approved):

nawatts (approved):

nawatts:
> Well, I guess I still want to merge this, even if just as an exercise in tidying up.

👍  And since `node-fetch` is a direct dependency, updating the version selector in `integration-tests/package.json` to require the patched version seems like a good practice.

PR #3192, opened 2022-06-30
[QA-1941] Add lint-staged in husky pre-commit

aweng98:
[QA-1941](https://broadworkbench.atlassian.net/browse/QA-1941)

Git's pre-commit hook (with `husky` and `lints-staged`) to check whether modified code passes eslint checks.


aweng98:
I had no idea .yarn/cache files are saved to repo. So many new cache files added.

package.json
```diff
@@ -112,5 +113,10 @@
   "packageManager": "yarn@3.2.0",
   "workspaces": [
     "integration-tests"
-  ]
+  ],
+  "lint-staged": {
+    "**/*.{js,jsx}": [
```
aweng98: I did not find any `.jsx` file. But this at least gives us an assurance for additions in future.

aweng98 (commented):
New checks blocked "bad" code from commit.
Screen Shot 2022-06-30 at 1 22 12 PM

package.json
kyuksel: We don't use JSX (and I hope we won't use for the foreseeable future 😊).  I'd suggest removing it to avoid confusion and/or premature optimization.

aweng98: I think we should keep it. But it's not a big deal for me. I need approvals so I'm removing it.

kyuksel:
I'm surprised by the massive amount of changes to `yarn` `cache` and `lock` 😮  Any thoughts @nawatts?

aweng98:
Again, I was surprised to see cache files stored in repo. IMO, we shouldn't do that.
Here is the article on what should be ignored from yarn documentation. https://next.yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

nawatts:
> Again, I was surprised to see cache files stored in repo. IMO, we shouldn't do that.
> Here is the article on what should be ignored from yarn documentation. https://next.yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

Storing cache files in git is intentional (added in #2906) since we are using zero-installs.

As for the additional dependencies, I assume they're dependencies of `lint-staged`. I don't think we _need_ `lint-staged`, since `yarn lint` is probably fast enough to just run on the entire code base in the pre-commit hook. But `lint-staged` is a nice to have improvement to hook speed if it runs ESLint only on changed files.

kyuksel (approved):
Thank you for this improvement that will save time for PR authors and reviewers! 🤩

package.json
```diff
@@ -112,5 +113,10 @@
   "packageManager": "yarn@3.2.0",
   "workspaces": [
     "integration-tests"
-  ]
+  ],
+  "lint-staged": {
+    "**/*.js": [
+      "yarn eslint"
```
nawatts: Since many ESLint rules are configured as warnings instead of errors and CI runs ESLint with `--max-warnings=0`, this will let through many things that would fail on CI. This should include the `--max-warnings=0` flag the same as in CI.

https://github.com/DataBiosphere/terra-ui/blob/fac6806358f52c6df2cc6b3a3becab2b64a3d6ec/.circleci/config.yml#L194

nawatts (changes requested):

package.json
aweng98: Done. Good catch! I'm wondering if we should remove that step in CircleCI since we're doing the same check on commit. It doesn't hurt to leave it as is.

nawatts: Adding it to a pre-commit hook does not remove the need to run it in CI. There is no guarantee that a contributor runs the pre-commit hook. It's like server side vs client side validation: client side is a nice to have enhancement for quicker feedback, but only server side can be relied on.

nawatts (approved):

PR #3193, opened 2022-06-30
[SUP-517] Add non-human reference genomes

MatthewBemis:
New references in selection dropdown:
Screen Shot 2022-06-30 at 2 33 45 PM 1

New reference attributes after adding:
Screen Shot 2022-06-30 at 2 33 57 PM

<!--
Hello, friend!
Thanks!

--!>


src/data/reference-data.js
```diff
@@ -68,6 +68,48 @@ export default {
...
+  'ROS-Cfam-1-0': {
+    ref_fasta: 'gs://gcp-public-data--broad-references/C.lupus_familiaris/ROS_Cfam_1.0/GCF_014441545.1_ROS_Cfam_1.0_genomic.fna'
+  },
+  'UU-Cfam-GSD-1-0': {
+    ref_fasta: 'gs://gcp-public-data--broad-references/C.lupus_familiaris/UU_Cfam_GSD_1.0/GCA_011100685.1_UU_Cfam_GSD_1.0_genomic.fna'
+  },
+  'Sscrofa11-1': {
+    ref_fasta: 'gs://gcp-public-data--broad-references/S.scrofa/Sscrofa11.1/GCF_000003025.6_Sscrofa11.1_genomic.fna'
+  },
+  'ARS-UI-Ramb-v2-0': {
```
MatthewBemis, line 111: Unfortunately the true assembly names for these include dots, which are not allow in attribute names. I wonder if it may be worth also including a displayName value, which is what we'd show in the dropdown...

nawatts: I like the idea, and I think that would probably be a better UX. I think that it would also require some additional complexity in the workflow inputs menu since workflow inputs would need to store the underlying attribute name.

nawatts: Then again, I guess adding that wouldn't change anything here, so it could be done later.

nawatts (approved):

calypsomatic (approved):

PR #3195, opened 2022-07-05
DC-430: Ensure that the oldest PR cannot be deleted

fboulnois:
Fixes a bug where if the deletion date is the same date as the oldest PR the oldest PR deployment could accidentally be deleted. Set the deletion PR time to 24h before the oldest PR and do that comparison instead.

terminal

### DIFF
--- a/scripts/delete-old-app-engine-versions.sh
+++ b/scripts/delete-old-app-engine-versions.sh
@@ -75,8 +75,9 @@ set_dev_deletion_date() {
 
     printf "${INFO} ${GRN}%s${RST} is the oldest PR and was deployed on ${GRN}%s${RST}\n" "${OLDEST_PR_NAME}" "${OLDEST_PR_DATE}"
 
-    if [ "${DELETION_TIME}" -gt "${OLDEST_PR_TIME}" ]; then
-        DELETION_DATE=$(echo "${OLDEST_PR_TIME}" | jq -r '. - (24 * 60 * 60) | strftime("%Y-%m-%d")') # 1 day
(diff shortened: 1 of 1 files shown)


pshapiro4broad (approved):
Looks good. I ran the script locally to verify, with the `gcloud app versions delete` line commented out.

pshapiro4broad (commented):
FWIW I ran `shellcheck` and it generated two warnings:
```
$ shellcheck delete-old-app-engine-versions.sh 

In delete-old-app-engine-versions.sh line 91:
    REMAIN_LIST_ITEMS=($(filter_app_engine_versions "version.createTime.date('%Y-%m-%d', Z)>'${DELETION_DATE}'"))
                       ^-- SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In delete-old-app-engine-versions.sh line 102:
    DELETE_LIST_ITEMS=($(filter_app_engine_versions "version.createTime.date('%Y-%m-%d', Z)<='${DELETION_DATE}'"))
                       ^-- SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).

For more information:
  https://www.shellcheck.net/wiki/SC2207 -- Prefer mapfile or read -a to spli...
$
```

fboulnois:
> FWIW I ran `shellcheck` and it generated two warnings:

Yes. Unfortunately there isn't really a portable way to get around this due to the version of Bash which is installed by default on MacOS.

cahrens (approved):

PR #3196, opened 2022-07-05
[DC-428] Add feedback context and to details page

s-rubenstein:
<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/pages/library/DataBrowserDetails.js
+++ b/src/pages/library/DataBrowserDetails.js
@@ -217,7 +217,6 @@ export const SidebarComponent = ({ dataObj, id }) => {
       ])
     ]),
     feedbackShowing && h(DataBrowserFeedbackModal, {
-      title: '',
       onDismiss: () => setFeedbackShowing(false),
       onSuccess: () => setFeedbackShowing(false),
       primaryQuestion: 'Is there anything missing or that you would like to see in this dataset view?',
--- a/src/pages/library/DataBrowserDetails.js
+++ b/src/pages/library/DataBrowserDetails.js
@@ -212,9 +212,7 @@ export const SidebarComponent = ({ dataObj, id }) => {
         div({ style: { marginLeft: 10, lineHeight: '1.3rem' } }, [
           h(Link, {
             onClick: () => setFeedbackShowing(true)
-          },
-          ['Provide feedback']),
-          ' on this dataset view'
+          }, ['Provide feedback on this dataset view'])
         ])
       ])
     ]),
--- a/src/pages/library/DataBrowserDetails.js
+++ b/src/pages/library/DataBrowserDetails.js
@@ -5,7 +5,6 @@ import { ButtonOutline, ButtonPrimary, ButtonSecondary, Link } from 'src/compone
 import FooterWrapper from 'src/components/FooterWrapper'
 import { centeredSpinner, icon } from 'src/components/icons'
 import { libraryTopMatter } from 'src/components/library-common'
-import Modal from 'src/components/Modal'
 import { ReactComponent as AzureLogo } from 'src/images/azure.svg'
 import { ReactComponent as GcpLogo } from 'src/images/gcp.svg'
 import { Ajax } from 'src/libs/ajax'
@@ -110,9 +109,8 @@ const MainContent = ({ dataObj }) => {
 
 export const SidebarComponent = ({ dataObj, id }) => {
   const { access } = dataObj
-  const [showRequestAccessModal, setShowRequestAccessModal] = useState()
-  const [feedbackShowing, setFeedbackShowing] = useState()
-  const [thanksShowing, setThanksShowing] = useState()
+  const [showRequestAccessModal, setShowRequestAccessModal] = useState(false)
+  const [feedbackShowing, setFeedbackShowing] = useState(false)
   const sidebarButtonWidth = 230
 
   return h(Fragment, [
@@ -223,24 +221,10 @@ export const SidebarComponent = ({ dataObj, id }) => {
     feedbackShowing && h(DataBrowserFeedbackModal, {
       title: '',
       onDismiss: () => setFeedbackShowing(false),
-      onSuccess: () => {
-        setFeedbackShowing(false)
-        setThanksShowing(true)
-      },
+      onSuccess: () => setFeedbackShowing(false),
       primaryQuestion: 'Is there anything missing or that you would like to see in this dataset view?',
       sourcePage: 'Catalog Details'
     }),
-    thanksShowing && h(Modal, {
-      onDismiss: () => setThanksShowing(false),
-      showCancel: false,
-      okButton: h(ButtonPrimary, {
-        onClick: () => setThanksShowing(false)
-      },
-      ['OK'])
-    },
-    [div({ style: { fontWeight: 600, fontSize: 18 } },
-      'Thank you for helping us improve the Data Catalog experience!')]
-    ),
     showRequestAccessModal && h(RequestDatasetAccessModal, {
       datasets: [dataObj],
       onDismiss: () => setShowRequestAccessModal(false)
--- a/src/pages/library/DataBrowserDetails.js
+++ b/src/pages/library/DataBrowserDetails.js
@@ -113,7 +113,7 @@ export const SidebarComponent = ({ dataObj, id }) => {
   const [showRequestAccessModal, setShowRequestAccessModal] = useState()
   const [feedbackShowing, setFeedbackShowing] = useState()
   const [thanksShowing, setThanksShowing] = useState()
-  const sidebarWidth = 230
+  const sidebarButtonWidth = 230
 
   return h(Fragment, [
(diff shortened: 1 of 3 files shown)


fboulnois (approved):
I checked the PR site out and I appreciate the responses in our internal channel so this lgtm

src/pages/library/DataBrowserDetails.js
```diff
@@ -109,6 +111,9 @@ const MainContent = ({ dataObj }) => {
 export const SidebarComponent = ({ dataObj, id }) => {
   const { access } = dataObj
   const [showRequestAccessModal, setShowRequestAccessModal] = useState()
+  const [feedbackShowing, setFeedbackShowing] = useState()
+  const [thanksShowing, setThanksShowing] = useState()
```
nawatts: Nit: it would be nice to initialize these state variables to false instead of undefined.

src/pages/library/DataBrowserDetails.js
```diff
@@ -195,16 +200,47 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
+      title: '',
+      onDismiss: () => setFeedbackShowing(false),
+      onSuccess: () => {
+        setFeedbackShowing(false)
+        setThanksShowing(true)
+      },
+      primaryQuestion: 'Is there anything missing or that you would like to see in this dataset view?',
+      sourcePage: 'catalog-details'
+    }),
+    thanksShowing && h(Modal, {
```
nawatts: Since there's now two "thanks for the feedback" modals (one here and one in DataBrowserToggler.js, is it worth creating a component for this? Or better yet, making it part of DataBrowserFeedbackModal?

pshapiro4broad (approved):
The code looks fine to me. Regarding the google form, I would change it so that the page field is required. You could also make it a multiple choice rather than free text field, so the input values for the field are restricted in google forms.

src/pages/library/DataBrowserDetails.js
```diff
@@ -195,16 +200,47 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
         onClick: () => {
           Ajax().Metrics.captureEvent(`${Events.catalogWorkspaceLink}:detailsView`, {
             id,
             title: dataObj['dct:title']
           })
           importDataToWorkspace([dataObj])
         }
-      }, ['Link to a workspace'])
+      }, ['Link to a workspace']),
+      div({ style: { display: 'flex', width: sidebarButtonWidth, marginTop: 20 } }, [
```
nawatts, line 210: It may be preferable to specify vertical margins in `em` or `rem` units instead of pixels.
https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/#vertical-margins

src/pages/library/DataBrowserDetails.js
```diff
@@ -195,16 +200,47 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
+        icon('talk-bubble', { size: 60, style: { width: 60, height: 45 } }),
+        div({ style: { marginLeft: 10, lineHeight: '1.3rem' } }, [
+          h(Link, {
+            onClick: () => setFeedbackShowing(true)
+          },
+          ['Provide feedback']),
+          ' on this dataset view'
```
nawatts: Is there a reason not to link this entire text? This effects what gets read by a screen reader when the link is focused and the full text is more descriptive.

src/pages/library/DataBrowserFeedbackModal.js
```diff
@@ -0,0 +1,68 @@
...
+
+  const errors = validate({ feedback, contactEmail },
+    { feedback: { length: { minimum: 1, maximum: 2000 } }, contactEmail: contactEmail ? { email: true } : {} })
+
+  return h(Modal, {
+    onDismiss,
+    shouldCloseOnEsc: false,
+    width: 500,
+    title: 'Give feedback',
+    okButton: h(TooltipTrigger, { content: !!errors && _.map(error => div({ key: error }, [error]), errors) },
```
nawatts: It would be nice if errors were shown in a more visible location than the submit button's tooltip. There is a ValidatedInput component in src/components/input.js to support this.

Also, ButtonPrimary accepts a `tooltip` prop so there's no need to wrap it with a TooltipTrigger component.

src/pages/library/DataBrowserFeedbackModal.js
```diff
@@ -0,0 +1,68 @@
...
+      [h(ButtonPrimary, {
+        disabled: errors,
+        onClick: submit
+      }, ['Submit'])]
+    )
+  }, [
+    h(IdContainer, [id => h(Fragment, [
+      h(FormLabel, { required: true, htmlFor: id, style: { fontSize: 14, fontWeight: 300 } },
```
nawatts: Does this label need the extremely light font weight? I don't think anything else in Terra is below 400. Also, could this label and the "Can we contact you with further questions?" label be styled consistently?

src/pages/library/DataBrowserDetails.js
s-rubenstein: I noticed that setRequestAccessModal was undefined beforehand, so I followed the pattern. (I actually wrote false first)

Now that I know preferred pattern is false default, I'll follow that moving forward.

src/pages/library/DataBrowserDetails.js
s-rubenstein: According to Lou's mocks, only the text there is linked. I can follow up with Lou

src/pages/library/DataBrowserDetails.js
s-rubenstein: Sounds good, I will move it to that modal.

src/pages/library/DataBrowserDetails.js
s-rubenstein: Here I actually will push back. In general, I do personally prefer `em` or `rem`. However, in this case the page is built around px margins. (look at the buttons above for examples)

Consistency within the page is more important, in my opinion. If we want to move consistently to `em` or `rem` for margins/padding, that would be a pretty comprehensive site refactor, and I would want to make sure we talked about it at a UI group meeting/full thread.

s-rubenstein: I, in fact, disagree with the point that the author is making about vertical margins specifically. Here, I actually think that the vertical spacing is for aesthetic purposes. (see this quote here:

```
This space has a “functional” purpose when it comes to text. We aren't using it aesthetically.

For these reasons, I think it does make sense to scale these margins with the user's chosen root font size.
```
)

s-rubenstein: UX question @lstrano-broad

src/pages/library/DataBrowserDetails.js
s-rubenstein: UX question @lstrano-broad

src/pages/library/DataBrowserFeedbackModal.js
s-rubenstein: UX question @lstrano-broad

src/pages/library/DataBrowserFeedbackModal.js
s-rubenstein: UX question @lstrano-broad

lstrano-broad: @nawatts You are correct, mock fixed to reflect same bold style for both..."Please tell us..." & "Can we contact you..."

src/pages/library/DataBrowserDetails.js
lstrano-broad: @nawatts Ok to link the entire text for the reader. "Provide feedback on this dataset view"

src/pages/library/DataBrowserDetails.js
nawatts: > However, in this case the page is built around px margins. (look at the buttons above for examples)

> I, in fact, disagree with the point that the author is making about vertical margins specifically. Here, I actually think that the vertical spacing is for aesthetic purposes

Fair points. Anecdotally, I think styles throughout terra-ui more often use `rem` than pixel units. But inline styling makes it very difficult to have much consistency.

lstrano-broad: You both have great points and I see the logic in em/rem vs. px. Reason UX team is using px is that our software Sketch does not support em/rem which is crazy.

src/pages/library/DataBrowserDetails.js
```diff
@@ -195,16 +198,31 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
+        icon('talk-bubble', { size: 60, style: { width: 60, height: 45 } }),
+        div({ style: { marginLeft: 10, lineHeight: '1.3rem' } }, [
+          h(Link, {
+            onClick: () => setFeedbackShowing(true)
+          }, ['Provide feedback on this dataset view'])
+        ])
+      ])
     ]),
+    feedbackShowing && h(DataBrowserFeedbackModal, {
+      title: '',
```
nawatts: Nit: Is this necessary? DataBrowserFeedbackModal doesn't take a `title` prop.

nawatts (approved):

PR #3197, opened 2022-07-05
[WOR-84] Show location and storage container name for Azure workspaces

cahrens:
To test this, you need to be able to create an Azure workspace; let me know if you need to be added to the Azure beta testing group so you have access to the Azure billing project.

image

This PR adds 2 new bits of data to the Workspace Dashboard-- location and storage container name.

image


src/libs/ajax.js
```diff
@@ -1102,6 +1105,33 @@ const DataRepo = signal => ({
   }
 })
 
+const AzureStorage = signal => ({
+  details: async (workspaceId = {}) => {
+    const res = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources?stewardship=CONTROLLED`,
```
cahrens, line 1110: There is an optional parameter to request just a resources of a specific type, but it does not allow multiple resource types to be specified. Therefore I would have to make 2 different calls if I wanted to filter by resource type, and that does not seem worth the complexity.

src/libs/ajax.js
```diff
@@ -1102,6 +1105,33 @@ const DataRepo = signal => ({
...
+      _.merge(authOpts(), { signal })
+    )
+    const data = await res.json()
+    const storageAccount = _.find({ metadata: { resourceType: 'AZURE_STORAGE_ACCOUNT' } }, data.resources)
```
cahrens, line 1114: There should be only 1 storage account.

src/libs/azure-utils.js
```diff
@@ -12,8 +12,9 @@ export const defaultAzureComputeConfig = {
 }
 
 //TODO: this should be fleshed out once azure region isn't limited to workspace region
-export const azureRegions = { eastus: { label: 'East US' } }
+export const azureRegions = { eastus: { flag: '🇺🇸', label: 'East US' } }
 export const getRegionLabel = key => _.has(key, azureRegions) ? azureRegions[key].label : 'Unknown azure region'
+export const getRegionFlag = key => _.has(key, azureRegions) ? azureRegions[key].flag : '❓'
```
cahrens, line 17: The question mark is what we use for unknown in the Google code as well (`region-common`). I would reference that variable, but `region-common` is under `components`, and it feels odd for something in `libs` to import from `components`.

integration-tests/tests/workspace-dashboard.js
```diff
@@ -173,6 +192,10 @@ const setAzureAjaxMockValues = async (testPage, namespace, name, workspaceDescri
         filter: { url: detailsUrl },
         fn: () => () => Promise.resolve(new Response(JSON.stringify(azureWorkspaceDetailsResult), { status: 200 }))
       },
+      {
+        filter: { url: workspaceResourcesUrl },
+        fn: () => () => Promise.resolve(new Response(JSON.stringify(azureWorkspaceResourcesResult), { status: 200 }))
+      },
```
dmohs, line 198: I like that we're trying out this mocking strategy. I've always worried about drift between the mock and real APIs, and this is a chance to see whether those worries have merit. I kind of doubt we'll see them change very often.

dmohs (approved):

aherbst-broad (approved):

PR #3202, opened 2022-07-06
[WOR-377] Add clipboard copy link for Azure container SAS URL

aherbst-broad:
## Context
[Relevant ticket](https://broadworkbench.atlassian.net/browse/WOR-377)
We want to surface Azure container SAS URls for usage in other tooling like Azure storage explorer.

## This PR
* Threads in the SAS URL from the WSM `getSasToken` response ([related PR](https://github.com/DataBiosphere/terra-workspace-manager/pull/711)) and places it in the Azure storage info box, as a clipboard copyable item.
   * I've added a `setInterval` call that will refresh the token every 59 minutes to help guard against stale tokens
* Updates the workspace dashboard integration test

Screen Shot 2022-07-07 at 2 54 22 PM

<!--

Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1122,7 +1122,8 @@ const AzureStorage = signal => ({
     if (storageAccount === undefined) { // Internal users may have early workspaces with no storage account.
       return {
         location: 'Unknown',
-        storageContainerName: 'None'
+        storageContainerName: 'None',
+        sasUrl: 'None'
       }
     } else {
       const container = _.find(
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -154,6 +154,7 @@ const withRequesterPays = wrappedFetch => (url, ...args) => {
 
 export const fetchOk = _.flow(withInstrumentation, withCancellation, withErrorRejection)(fetch)
 
+const fetchWsm = withUrlPrefix(`${getConfig().workspaceManagerUrlRoot}/api/`, fetchOk)
 const fetchSam = _.flow(withUrlPrefix(`${getConfig().samUrlRoot}/`), withAppIdentifier)(fetchOk)
 const fetchBuckets = _.flow(withRequesterPays, withRetryOnError, withUrlPrefix('https://storage.googleapis.com/'))(fetchOk)
 const fetchRawls = _.flow(withUrlPrefix(`${getConfig().rawlsUrlRoot}/api/`), withAppIdentifier)(fetchOk)
@@ -669,6 +670,18 @@ const CromIAM = signal => ({
 
 
 const Workspaces = signal => ({
+
+  resources: async workspaceId => {
+    const res = await fetchWsm(`workspaces/v1/${workspaceId}/resources`, _.merge(authOpts(), { signal }))
+    return res.json()
+  },
+
+  sasToken: async (workspaceId, containerId) => {
+    const res = await fetchWsm(`workspaces/v1/${workspaceId}/resources/controlled/azure/storageContainer/${containerId}/getSasToken`,
+      _.merge(authOpts(), { signal, method: 'POST' }))
+    return res.json()
+  },
+
   list: async fields => {
     const res = await fetchRawls(`workspaces?${qs.stringify({ fields }, { arrayFormat: 'comma' })}`, _.merge(authOpts(), { signal }))
     return res.json()
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -156,7 +156,6 @@ const withRequesterPays = wrappedFetch => (url, ...args) => {
 
 export const fetchOk = _.flow(withInstrumentation, withCancellation, withErrorRejection)(fetch)
 
-const fetchWsm = withUrlPrefix(`${getConfig().workspaceManagerUrlRoot}/api/`, fetchOk)
 const fetchSam = _.flow(withUrlPrefix(`${getConfig().samUrlRoot}/`), withAppIdentifier)(fetchOk)
 const fetchBuckets = _.flow(withRequesterPays, withRetryOnError, withUrlPrefix('https://storage.googleapis.com/'))(fetchOk)
 const fetchRawls = _.flow(withUrlPrefix(`${getConfig().rawlsUrlRoot}/api/`), withAppIdentifier)(fetchOk)
@@ -1134,17 +1133,17 @@ const AzureStorage = signal => ({
   },
 
   resources: async workspaceId => {
-    const res = await fetchWsm(`workspaces/v1/${workspaceId}/resources`, _.merge(authOpts(), { signal }))
+    const res = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources`, _.merge(authOpts(), { signal }))
     return res.json()
   },
 
   sasToken: async workspaceId => {
-    const resourcesResponse = await fetchWsm(`workspaces/v1/${workspaceId}/resources`, _.merge(authOpts(), { signal }))
-    const resourcesData = await resourcesResponse.json()
-    const container = _.find(resource => resource.metadata.resourceType === 'AZURE_STORAGE_CONTAINER', resourcesData.resources)
+    const resourcesResponse = await AzureStorage(signal).resources(workspaceId)
+    const container = _.find(resource => resource.metadata.resourceType === 'AZURE_STORAGE_CONTAINER', resourcesResponse.resources)
 
-    const tokenResponse = await fetchWsm(`workspaces/v1/${workspaceId}/resources/controlled/azure/storageContainer/${container.metadata.resourceId}/getSasToken`,
+    const tokenResponse = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources/controlled/azure/storageContainer/${container.metadata.resourceId}/getSasToken`,
       _.merge(authOpts(), { signal, method: 'POST' }))
+
     return tokenResponse.json()
   }
 })
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1106,6 +1106,13 @@ const DataRepo = signal => ({
 })
 
 const AzureStorage = signal => ({
+  sasToken: async (workspaceId, containerId) => {
+    const tokenResponse = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources/controlled/azure/storageContainer/${containerId}/getSasToken`,
+      _.merge(authOpts(), { signal, method: 'POST' }))
+
+    return tokenResponse.json()
+  },
+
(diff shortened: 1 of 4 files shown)


src/libs/ajax.js
```diff
@@ -154,6 +154,7 @@ const withRequesterPays = wrappedFetch => (url, ...args) => {
 
 export const fetchOk = _.flow(withInstrumentation, withCancellation, withErrorRejection)(fetch)
 
+const fetchWsm = withUrlPrefix(`${getConfig().workspaceManagerUrlRoot}/api/`, fetchOk)
```
cahrens: You will want to resolve conflicts with my PR, which I just merged.

src/libs/ajax.js
```diff
@@ -669,6 +670,18 @@ const CromIAM = signal => ({
 
 
 const Workspaces = signal => ({
+
+  resources: async workspaceId => {
```
cahrens: I decided to create a separate "AzureStorage" const, but we could move my details method into Workspaces (I was somewhat on the fence). We will definitely want to share `resources`.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -210,6 +210,8 @@ const WorkspaceDashboard = _.flow(
   const [busy, setBusy] = useState(false)
   const [consentStatus, setConsentStatus] = useState(undefined)
   const [tagsList, setTagsList] = useState(undefined)
+  const [sasToken, setSasToken] = useState(undefined)
+  const [isFetchingSasToken, setIsFetchingSasToken] = useState(true)
```
cahrens: I don't think you need this state-- you can go off of whether `sasToken` is undefined for enabling the Clipboard button.

And `Utils.withBusyState` will show a spinner, which I don't think we want to do.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -296,6 +300,16 @@ const WorkspaceDashboard = _.flow(
     setTagsList(await Ajax().Workspaces.workspace(namespace, name).deleteTag(tag))
   })
 
+  const loadSasToken = _.flow(
+    withErrorReporting('Error loading SAS token'),
+    Utils.withBusyState(setIsFetchingSasToken)
+  )(async workspaceId => {
+    const response = await Ajax().Workspaces.resources(workspaceId)
+    const container = _.find(resource => resource.metadata.resourceType === 'AZURE_STORAGE_CONTAINER', response.resources)
+    const sasToken = await Ajax().Workspaces.sasToken(workspaceId, container.metadata.resourceId)
```
cahrens: I would consider putting all the logic for getting the sasToken into the `sasToken` method (fetching the resources, finding the AZURE_STORAGE_CONTAINER). At least, that's what I decided to do for `AzureStorage.details`.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -343,6 +357,9 @@ const WorkspaceDashboard = _.flow(
         h(InfoRow, { title: 'Resource Group ID' }, [
           h(TooltipCell, [azureContext.managedResourceGroupId]),
           h(ClipboardButton, { 'aria-label': 'Copy resource group id to clipboard', text: azureContext.managedResourceGroupId, style: { marginLeft: '0.25rem' } })
+        ]),
+        h(InfoRow, { title: 'Azure Storage' }, [
```
cahrens: I don't think we need to repeat Azure here… perhaps just call it it `Storage SAS URL`? Definitely get Adam's opinion.

cahrens:
@aherbst-broad please also add coverage in the (mocked) dashboard "integration" test. You will need to add a new mocked AJAX response for the SAS token, and then add to the list of items we verify in the [Cloud Environment portion](https://github.com/DataBiosphere/terra-ui/pull/3197/files#diff-96fda9961f180503fd4a6879a877067e3d0e75ac6eba72dac2631633eb97b3c8R226) of the UI. If you have questions, let me know.

src/libs/ajax.js
aherbst-broad: I'll move to the `AzureStorage` const 👍

src/libs/ajax.js
```diff
@@ -1130,6 +1130,21 @@ const AzureStorage = signal => ({
         storageContainerName: container.resourceAttributes.azureStorageContainer.storageContainerName
       }
     }
+  },
+
+  resources: async workspaceId => {
```
cahrens: Can you also call the new resources function up at line 1110?

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -226,6 +227,10 @@ const WorkspaceDashboard = _.flow(
       loadBucketSize()
     } else {
       loadAzureStorage()
+      loadSasToken(workspace.workspace.workspaceId)
+
+      // sas tokens expire after 1 hour
+      setInterval(loadSasToken, Utils.durationToMillis({ minutes: 59 }), workspace.workspace.workspaceId)
```
cahrens: Nit: perhaps make this more like 50 minutes so they would conceivably have time to do something with the token (if they are near the timeout)?

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -362,6 +372,13 @@ const WorkspaceDashboard = _.flow(
             'aria-label': 'Copy storage container name to clipboard',
             text: storageContainerName, style: { marginLeft: '0.25rem' }
           })
+        ]),
+        h(InfoRow, { title: 'SAS URL' }, [
```
cahrens: I would suggest "Storage SAS URL" for clarity.

src/pages/workspaces/workspace/Dashboard.js
nawatts: Does this interval ever get cleared? As is, it looks like it would keep going after leaving the page. Also, it looks like multiple calls to refresh would result in multiple intervals running at the same time.

src/libs/ajax.js
```diff
@@ -1130,6 +1130,21 @@ const AzureStorage = signal => ({
...
+    const res = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources`, _.merge(authOpts(), { signal }))
+    return res.json()
+  },
+
+  sasToken: async workspaceId => {
+    const resourcesResponse = await AzureStorage(signal).resources(workspaceId)
+    const container = _.find(resource => resource.metadata.resourceType === 'AZURE_STORAGE_CONTAINER', resourcesResponse.resources)
```
cahrens: Note that you can also use this syntax (I don't have a strong preference):

```
const storageAccount = _.find({ metadata: { resourceType: 'AZURE_STORAGE_ACCOUNT' } }, data.resources)

File: src/libs/ajax.js (hunk 6 above)
Comment (cahrens) [reply src/libs/ajax.js]: I think we might as well limit to CONTROLLED resources.

`workspaces/v1/${workspaceId}/resources?stewardship=CONTROLLED`

File: src/libs/ajax.js [hunk 10]
```diff
@@ -1130,6 +1130,21 @@ const AzureStorage = signal => ({
...
+
+    const tokenResponse = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources/controlled/azure/storageContainer/${container.metadata.resourceId}/getSasToken`,
```
cahrens: Do you want to guard against the storage container not existing (which is the case for old workspaces)? I chose to do it in my code (line 1115), but also recognize that the additional complexity may not be worth carrying forward.

Note also that I am looking up the storage account first (assuming there will only be 1), and then finding the storage container that matches the storage account. You are assuming there is only one storage container… which is probably fine. But I am starting to think that perhaps just fold the SasToken into my `details` method so that we aren't doing this storage container lookup twice (and assured it is in fact the same storage container). Then the details method would return {location, storageContainerName, sasToken}. What do you think?

src/pages/workspaces/workspace/Dashboard.js
cahrens: Good point--we should be passing `signal` in at line 314. @nawatts, would that be sufficient, or does the interval also need to be explicitly cleared in a cleanup method?

nawatts: Passing `signal` to `Ajax` would cancel an in-progress request when the component is unmounted. But we'd also need to cancel the interval triggering the requests.

cahrens: @aherbst-broad you can see an example of how to do this in `AppLauncher.js` (search for usages of `interval`).

aherbst-broad: Ah good call, checkout https://github.com/DataBiosphere/terra-ui/pull/3202/commits/c7e23b55c98872c7f01f21f9d2c1101e21511b3e

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -239,6 +246,11 @@ const WorkspaceDashboard = _.flow(
 
   useEffect(() => {
     setLocalPref(persistenceId, { workspaceInfoPanelOpen, cloudInfoPanelOpen, ownersPanelOpen, authDomainPanelOpen, tagsPanelOpen })
+
+    return () => {
+      clearInterval(sasTokenRefreshInterval.current)
```
nawatts, line 251: I think this cleanup should be in a separate `useEffect` hook. Since `useEffect` hooks are cleaned up and re-run whenever their dependencies change, having it in this hook would stop the SAS token refresh interval any time a user toggled one of the panels in the sidebar.
https://reactjs.org/docs/hooks-effect.html#explanation-why-effects-run-on-each-update

src/libs/ajax.js
aherbst-broad: Yeah I think that makes sense so both code paths are in agreement about the container we're looking at.

src/libs/ajax.js
aherbst-broad: I've refactored to centralize the calls into your details fn, take a look.

src/libs/ajax.js
```diff
@@ -1125,9 +1132,11 @@ const AzureStorage = signal => ({
         },
         data.resources
       )
+      const sas = await AzureStorage(signal).sasToken(workspaceId, container.metadata.resourceId)
       return {
         location: storageAccount.resourceAttributes.azureStorage.region,
-        storageContainerName: container.resourceAttributes.azureStorageContainer.storageContainerName
+        storageContainerName: container.resourceAttributes.azureStorageContainer.storageContainerName,
+        sasUrl: sas.url
```
cahrens, line 1140: I would advice also adding sasUrl up at line 1126, value of 'None'.

cahrens (approved):

nawatts (approved):

PR #3203, opened 2022-07-06
[IA-3496] Fix creating new environment and changing size of persistent disk

n8landolt:
# Description

Currently, the scenario where you have an existing persistent disk but no environment and a user goes to create an environment is not working correctly.

# Solution
When the user creates an environment with an existing persistent disk, and they increase the size, the disk will expand in size, but will maintain the data and all other properties.

# Testing
Manual testing:

Recreated above mentioned scenario, checked logs to ensure the PD ID stayed the same.

Additionally, made sure to re-test the scenario where the user reattaches the same disk at the same size.


src/components/ComputeModal.js
```diff
@@ -308,9 +309,14 @@ export const ComputeModalBase = ({
           autopauseThreshold: computeConfig.autopauseThreshold
         })
       }
+
       if (shouldCreateRuntime) {
         const diskConfig = Utils.cond(
           [desiredRuntime.cloudService === cloudServices.DATAPROC, () => ({})],
+          [shouldUpdatePersistentDisk, () => {
+            Ajax().Disks.disk(googleProject, currentPersistentDiskDetails.name).update(desiredPersistentDisk.size)
```
petesantos: Do we need to do anything if there is an error on this `update` call?

jdcanas: I dont like this diskConfig  resolution `Utils.cond` calling the API. Can we call this down on line 332 before the runtime create call so all the API calls happen in the same place? This entire function is already wrapped in a `withErrorReportingInModal('Error modifying cloud environment', onError)`, so an error in the call here is handled.

n8landolt: Made an adjustment, thoughts?

jdcanas (approved):

petesantos (approved):

PR #3204, opened 2022-07-06
[NEED TICKET] UI context-analysis-bar test minor improvement

aweng98:


integration-tests/tests/analysis-context-bar.js
```diff
@@ -21,7 +21,13 @@ const testAnalysisContextBarFn = _.flow(
 
   // Create a runtime
   await click(page, clickable({ textContains: 'Environment Configuration' }))
+
   await findElement(page, getAnimatedDrawer('Cloud Environment Details'))
+  // Verify Jupyter Environment Settings button is visible and enabled. Pause Environments and Open buttons are visible and disabled because no runtime.
+  await findButtonInDialogByAriaLabel(page, 'Jupyter Environment')
+  await findButtonInDialogByAriaLabel(page, 'Jupyter Status', true)
+  await findButtonInDialogByAriaLabel(page, 'Launch Jupyter', true)
```
aweng98: Checks buttons displayed correctly as intented.

Screen Shot 2022-07-06 at 4 34 09 PM

integration-tests/utils/integration-utils.js
```diff
@@ -273,8 +273,9 @@ const findInDataTableRow = (page, entityName, text) => {
   return findElement(page, elementInDataTableRow(entityName, text))
 }
 
-const findButtonInDialogByAriaLabel = (page, ariaLabelText) => {
-  return page.waitForXPath(`//*[@role="dialog" and @aria-hidden="false"]//*[@role="button" and contains(@aria-label,"${ariaLabelText}")]`,
+const findButtonInDialogByAriaLabel = (page, ariaLabelText, isDisabled = false) => {
+  return page.waitForXPath(
+    `//*[@role="dialog" and @aria-hidden="false"]//*[self::div[@role="button"] | self::a][@aria-disabled="${isDisabled}" and contains(@aria-label,"${ariaLabelText}")]`,
```
aweng98, line 278: Update Xpath to find different types of buttons: `//a` and `//div[@role="button"]`

integration-tests/utils/integration-utils.js
```diff
@@ -273,8 +273,9 @@ const findInDataTableRow = (page, entityName, text) => {
   return findElement(page, elementInDataTableRow(entityName, text))
 }
 
-const findButtonInDialogByAriaLabel = (page, ariaLabelText) => {
-  return page.waitForXPath(`//*[@role="dialog" and @aria-hidden="false"]//*[@role="button" and contains(@aria-label,"${ariaLabelText}")]`,
+const findButtonInDialogByAriaLabel = (page, ariaLabelText, isDisabled = false) => {
```
breilly2: I don't love the lone boolean, especially because callers passing `true` means that it should be disabled, which feels backwards. What would you think about an options object?
```suggestion
const findButtonInDialogByAriaLabel = (page, ariaLabelText, { isDisabled = false } = {}) => {
```
Then calls would look like `findButtonInDailogByAriaLabel(page, 'Click Me', { isDisabled: true })`.

aweng98: I like this 👍🏻

breilly2 (approved):
👍 once there's a ticket for this.

I'm curious... is this covering a failure mode you've seen? What part of this expectation was broken? Were these buttons enabled even though they shouldn't have been?

PR #3206, opened 2022-07-07
[SUP-558] Add filtering and sorting to call table in workflow dashboard

MatthewBemis:
This adds sorting on columns and filtering via call status to the Calls collapsable section in the Workflow Dashboard.
Screen Shot 2022-07-07 at 3 49 04 PM

This is a solid workflow to test it on: https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/DRS%20Data%20Access%20Scale%20Testing%20-%20Dev/job_history/aaa8dd84-38b1-4981-be73-c84ff5149b02/66102117-8ad6-4079-b42f-3ac487f0b787

<!--
Hello, friend!
Thanks!

--!>


src/components/job-common.js
```diff
@@ -115,7 +115,7 @@ export const makeCromwellStatusLine = (executionStatus, backendStatus) => {
 export const makeSection = (label, children) => div({
   style: {
     flex: '0 0 33%', padding: '0 0.5rem 0.5rem', marginTop: '1rem',
-    whiteSpace: 'pre', textOverflow: 'ellipsis', overflow: 'hidden'
+    whiteSpace: 'pre', textOverflow: 'ellipsis'
```
MatthewBemis: This was causing the status dropdown to become partially cut off when there were only a few rows. I don't believe that it has any adverse consequences in any of the places that it is used

src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -90,7 +121,7 @@ const CallTable = ({ namespace, name, submissionId, workflowId, callName, callOb
             size: { basis: 200, grow: 2 },
             headerRenderer: () => 'Links',
```
MatthewBemis, line 122: I didn't think it was necessary to add a sort on the Links column

src/components/job-common.js
nawatts: This change affects the layout of the top section of the submission details page for some window sizes.

## Before
Screen Shot 2022-07-07 at 3 48 09 PM

## After
Screen Shot 2022-07-07 at 3 48 25 PM

nawatts: Based on the 33% flex-basis in `makeSection`, I'm guessing it was designed for the top section of the workflow dashboard page ("Workflow Status", "Workflow Timing", and "Links"). I think we can replace it where it's used for the "Total Call Status Counts" and "Call Lists" sections.

nawatts (changes requested):
The bulk of the changes look good. Just one issue with the style change in `makeSection`.

src/components/job-common.js
MatthewBemis: Fixed in 8c6988f726aadfd166fd8b901d8098f741dca636

It didn't seem possible to have the child override the `overflow` property of the parent, so I allowed overriding the style in `makeSection`

nawatts (approved):

cjllanwarne (approved):

PR #3207, opened 2022-07-07
[SU-155] Preserve column settings when renaming a table

nawatts:
Currently, when a table is renamed, any saved column settings for that table are lost. This changes adds a step to the rename process to update the workspace attribute containing saved column settings.

## Before
https://user-images.githubusercontent.com/1156625/177814289-e1fe32b0-dc7a-422b-9c73-0093e1f2f539.mov

## After
https://user-images.githubusercontent.com/1156625/177814244-f94bafcb-63b3-4425-bd4d-41b16fc71b4a.mov

### DIFF
--- a/src/components/data/SavedColumnSettings.js
+++ b/src/components/data/SavedColumnSettings.js
@@ -52,7 +52,7 @@ export const allSavedColumnSettingsEntityTypeKey = ({ snapshotName, entityType }
   return entityTypeKey
 }
 
-const useSavedColumnSettings = ({ workspaceId, snapshotName, entityMetadata, entityType }) => {
+export const useSavedColumnSettings = ({ workspaceId, snapshotName, entityMetadata, entityType }) => {
   // Saved column settings for all tables are stored in the same attribute in the format:
   // {
   //   snapshots: {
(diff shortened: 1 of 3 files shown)


src/components/data/SavedColumnSettings.js
```diff
@@ -88,9 +88,9 @@ const useSavedColumnSettings = ({ workspaceId, snapshotName, entityMetadata, ent
     )(workspace)
   }
 
-  const updateSavedColumnSettings = async allColumnSettings => {
+  const updateAllSavedColumnSettings = async allColumnSettings => {
     const { namespace, name } = workspaceId
-    await Ajax(signal).Workspaces.workspace(namespace, name).shallowMergeNewAttributes({
```
nawatts, line 93: This request writes data, so it shouldn't be cancelled if the component is unmounted.

src/components/data/SavedColumnSettings.js
```diff
@@ -127,17 +127,19 @@ const useSavedColumnSettings = ({ workspaceId, snapshotName, entityMetadata, ent
...
     // Re-fetch column settings to reduce the chances of overwriting changes made by other users since we first loaded settings.
     const allColumnSettings = await getAllSavedColumnSettings()
     const newColumnSettings = _.omit(`${entityTypeKey}.${columnSettingsName}`, allColumnSettings)
-    await updateSavedColumnSettings(newColumnSettings)
+    await updateAllSavedColumnSettings(newColumnSettings)
   }
 
   return {
+    getAllSavedColumnSettings,
+    updateAllSavedColumnSettings,
```
nawatts, line 142: Expose functions to get/set saved column settings for all tables in the workspace. Previously, for convenience, `useSavedColumnSettings` only exposed functions to get/set saved column settings for the table specified by the `entityType` parameter.

MatthewBemis (approved):
Thank you for taking care of this one! 🙏

dmohs (approved):

PR #3208, opened 2022-07-07
[QA-1948] Integration test SignIn failure

aweng98:
**Sign In** still fails after new deploy. I want to add a retry if fails to see it helps. I'll make a PR to revert in future if this workaround does not help to resolve the issue.

A number of PRs have been merged to address the issue with Sign In after new gcloud deploy (e.g. to Staging env), but nothing seems to make any difference.

For example, failed CircleCI `integration-tests-staging` [job](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/10488/workflows/a35846e2-dd85-4743-a3e7-2598d7af93a1).

`billing-projects` test is one of two started first. Test failed with following stacktrace.

```
Error reading historical timing data: file does not exist
Requested weighting by historical based timing, but they are not present. Falling back to weighting by name.
**  Running billing-projects at 7/5/2022, 11:18:41
**  Running workspace-dashboard at 7/5/2022, 11:18:41
Captured screenshot: /tmp/staging/test-results/failure-screenshots/failure-1657034388288-billing-projects.png
Saved screenshot page content: /tmp/staging/test-results/failure-screenshots/failure-1657034391417-billing-projects.html
```

```
page.http GET 200 https://bvdp-saturn-staging.appspot.com/
page.http GET 200 https://bvdp-saturn-staging.appspot.com/logo-grey.svg
page.http GET 200 https://bvdp-saturn-staging.appspot.com/static/css/main.34135f2a.css
page.http GET 200 https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js
page.http GET 200 https://bvdp-saturn-staging.appspot.com/loading-spinner.svg
page.http GET 200 https://bvdp-saturn-staging.appspot.com/config.json
page.http GET 404 https://bvdp-saturn-staging.appspot.com/static/js/688.97cab371.chunk.js
page.console Failed to load resource: the server responded with a status of 404 ()
page.pageerror [Error: ChunkLoadError: Loading chunk 688 failed.
(error: https://bvdp-saturn-staging.appspot.com/static/js/688.97cab371.chunk.js)
    at Object.r.f.j (https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js:1:42175)
    at https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js:1:39349
    at Array.reduce (<anonymous>)
    at Function.r.e (https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js:1:39314)
    at https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js:1:43692
    at c (https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js:1:4985)
    at Generator._invoke (https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js:1:4773)
    at Generator.next (https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js:1:5414)
    at o (https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js:1:25838)
    at s (https://bvdp-saturn-staging.appspot.com/static/js/main.ae1969ee.js:1:26041)]
```

Another example,

```
withWorkspace ...
page.http GET 200 https://bvdp-saturn-staging.appspot.com/
page.http GET 200 https://bvdp-saturn-staging.appspot.com/loading-spinner.svg
page.http GET 200 https://bvdp-saturn-staging.appspot.com/logo-grey.svg
page.http GET 404 https://bvdp-saturn-staging.appspot.com/static/js/main.09c38c86.js
page.console Failed to load resource: the server responded with a status of 404 ()
page.http GET 200 https://bvdp-saturn-staging.appspot.com/static/css/main.34135f2a.css
Captured screenshot: /tmp/staging/test-results/failure-screenshots/failure-1657547372003-google-workspace.png
```

[QA-1948](https://broadworkbench.atlassian.net/browse/QA-1948)

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -220,7 +220,7 @@ const signIntoTerra = async (page, { token, testUrl }) => {
     minTimeout: 1000, // This is min wait time between retries
     onFailedAttempt: error => {
       console.error(
-        `Sign in to Terra attempt ${error.attemptNumber} failed. There are ${error.retriesLeft} retries left.`
+        `Sign in to Terra attempt ${error.attemptNumber} failed. ${error.retriesLeft} retries left.`
       )
     },
     retries: 2
@@ -238,6 +238,8 @@ const signIntoTerra = async (page, { token, testUrl }) => {
       await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout: 60 * 1000 })
     } catch (e) {
       console.error(e)
+      // Stop page loading, as if you hit "X" in the browser. ignore exception.
+      await page._client.send('Page.stopLoading').catch(err => void err)
       throw new Error(e)
     }
   }
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -231,15 +231,11 @@ const signIntoTerra = async (page, { token, testUrl }) => {
   const run = async url => {
     try {
       const httpResponse = await page.goto(url, { waitUntil: ['networkidle0', 'domcontentloaded'], timeout })
-      if (httpResponse.status() >= 400) {
+      if (!(httpResponse.ok() || httpResponse.status() === 304)) {
         throw new Error(`Error loading URL: ${url}. Http response status: ${httpResponse.statusText()}`)
       }
-      await page.waitForFunction(url => {
-        return window.location.href.includes(url)
-      }, {}, url)
       await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout })
     } catch (e) {
-      console.error(e)
       // Stop page loading, as if you hit "X" in the browser. ignore exception.
       await page._client.send('Page.stopLoading').catch(err => void err)
       throw new Error(e)
@@ -249,8 +245,10 @@ const signIntoTerra = async (page, { token, testUrl }) => {
   if (!!testUrl) {
     console.log(`Loading URL: ${testUrl}`)
     await pRetry(() => run(testUrl), retryOptions)
+  } else {
+    await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout })
   }
-  await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout })
+
   await waitForNoSpinners(page)
 
   await page.waitForFunction('!!window["forceSignIn"]')
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -212,7 +212,7 @@ const dismissNPSSurvey = async page => {
   }
 }
 
-// Test workaround with one retry: Loading of Terra UI does not work sometimes after new deploy to Staging/Alpha.
+// Test workaround: Retry loading of Terra UI if fails first time. This issue often happens after new deploy to Staging/Alpha.
 const signIntoTerra = async (page, { token, testUrl }) => {
   console.log('signIntoTerra ...')
 
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -230,7 +230,7 @@ const signIntoTerra = async (page, { token, testUrl }) => {
   const timeout = 60 * 1000
   const run = async url => {
     try {
-      const httpResponse = await page.goto(url, { waitUntil: ['networkidle0', 'domcontentloaded'], timeout })
+      const httpResponse = await gotoPage(page, url)
       if (!(httpResponse.ok() || httpResponse.status() === 304)) {
         throw new Error(`Error loading URL: ${url}. Http response status: ${httpResponse.statusText()}`)
       }
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -212,6 +212,7 @@ const dismissNPSSurvey = async page => {
   }
 }
 
+// Test workaround with one retry: Loading of Terra UI does not work sometimes after new deploy to Staging/Alpha.
 const signIntoTerra = async (page, { token, testUrl }) => {
   console.log('signIntoTerra ...')
 
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -216,7 +216,15 @@ const signIntoTerra = async (page, { token, testUrl }) => {
 
   const loadUrl = async url => {
     console.log(`Loading URL: ${url}`)
-    await page.goto(url, { waitUntil: ['networkidle0', 'domcontentloaded'] })
+    let httpResponse
+    try {
+      httpResponse = await page.goto(url, { waitUntil: ['networkidle0', 'domcontentloaded'] })
+    } catch (e) {
+      console.error(e) // Known issue with goto() throws Navigation Timeout error. Log error then continue
+    }
+    if (httpResponse.status() >= 400) {
+      throw new Error(`Error loading URL: ${url}. Http response status: ${httpResponse.status()}`)
+    }
     await page.waitForFunction(url => {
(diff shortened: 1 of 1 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -211,9 +212,45 @@ const dismissNPSSurvey = async page => {
...
+      await page._client.send('Page.stopLoading').catch(err => void err)
+      throw new Error(e)
+    }
+  }
+
+  if (!!testUrl) {
+    console.log(`Loading URL: ${testUrl}`)
+    await pRetry(() => run(testUrl), retryOptions)
+  }
+  await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout })
```
cahrens: I don't think you want this since it is also at line 240 (unless you expect that `testUrl` might not be defined, in which case put this call in an else).

integration-tests/utils/integration-utils.js
```diff
@@ -211,9 +212,45 @@ const dismissNPSSurvey = async page => {
...
+      )
+    },
+    retries: 2
+  }
+
+  const timeout = 60 * 1000
+  const run = async url => {
+    try {
+      const httpResponse = await page.goto(url, { waitUntil: ['networkidle0', 'domcontentloaded'], timeout })
+      if (httpResponse.status() >= 400) {
```
breilly2: Would anything but 200 be a success?

integration-tests/utils/integration-utils.js
```diff
@@ -211,9 +212,45 @@ const dismissNPSSurvey = async page => {
...
+        throw new Error(`Error loading URL: ${url}. Http response status: ${httpResponse.statusText()}`)
+      }
+      await page.waitForFunction(url => {
+        return window.location.href.includes(url)
```
nawatts: Why is this check necessary? What would cause `page.goto(url)` to succeed but `window.location.href` to not contain `url`?

integration-tests/utils/integration-utils.js
```diff
@@ -211,9 +212,45 @@ const dismissNPSSurvey = async page => {
...
+      }, {}, url)
```
breilly2: What is the purpose of this?

integration-tests/utils/integration-utils.js
```diff
@@ -211,9 +212,45 @@ const dismissNPSSurvey = async page => {
...
+        `Sign in to Terra attempt ${error.attemptNumber} failed. There are ${error.retriesLeft} retries left.`
+      )
+    },
+    retries: 2
+  }
+
+  const timeout = 60 * 1000
+  const run = async url => {
+    try {
+      const httpResponse = await page.goto(url, { waitUntil: ['networkidle0', 'domcontentloaded'], timeout })
```
nawatts: Why switch away from `navOptionNetworkIdle` here? We use `navOptionNetworkIdle` for other page navigations.
https://github.com/DataBiosphere/terra-ui/blob/8f48397a8bcef3ebffa75b143a249aa5ffd78ae2/integration-tests/utils/integration-utils.js#L396-L398

integration-tests/utils/integration-utils.js
```diff
@@ -211,9 +212,45 @@ const dismissNPSSurvey = async page => {
...
+      if (httpResponse.status() >= 400) {
+        throw new Error(`Error loading URL: ${url}. Http response status: ${httpResponse.statusText()}`)
+      }
+      await page.waitForFunction(url => {
+        return window.location.href.includes(url)
+      }, {}, url)
+      await page.waitForXPath('//*[contains(normalize-space(.),"Loading Terra")]', { hidden: true, timeout })
+    } catch (e) {
+      console.error(e)
```
nawatts: Will this error be logged twice? Once here and once in the `onFailedAttempt` function in `retryOptions`?

breilly2 (commented):
There's more code than I expect to see for a retry. I wonder if it would be simpler to wrap the existing body of `signIntoTera` with a retry that checks whether the `forceSignIn` had the desired effect. That might be an easier condition to detect.

integration-tests/utils/integration-utils.js
aweng98: There are few tests that do not pass in `testUrl`. It's in `else` now.

integration-tests/utils/integration-utils.js
aweng98: Serves as a verification. Assertion matcher `expect` functions are normally called in test file only.

integration-tests/utils/integration-utils.js
aweng98: I have gotten 304 (Not Modified) response code. Personally, I'd consider 304 to be a success. I tried with Puppeteer Response `ok()` before but this function does not consider 304 be a success.
```
   ok() {
        // TODO: document === 0 case?
        return this._status === 0 || (this._status >= 200 && this._status <= 299);
    }
 ```

breilly2: Okay. I'd rather we be more explicit about that by checking for both `ok()` and `304` instead of treating anything less than 400 as good.

integration-tests/utils/integration-utils.js
breilly2: The question I had was better expressed by Nick:
> What would cause `page.goto(url)` to succeed but `window.location.href` to not contain `url`?

integration-tests/utils/integration-utils.js
aweng98: I think it may be a _mistake_ to use `networkidle0` in page navigation/goto. It considers navigation to be finished when there are no more than 0 `fetch` network requests for 500 ms. Since there are other types of network requests triggered automatically when loading `index.html` page, I think it self alone may not be sufficient.

integration-tests/utils/integration-utils.js
aweng98: I'm not sure because Sign In does not fail when I run tests locally and in CircleCI without deploy first. I removed `console.error(e)` for now.

integration-tests/utils/integration-utils.js
aweng98: It's just an assertion. Removed for now since it seems to be redudant.

integration-tests/utils/integration-utils.js
aweng98: Sounds good. Updated to `if (!(httpResponse.ok() || httpResponse.status() === 304))`.

integration-tests/utils/integration-utils.js
nawatts: Can we use the `gotoPage` function from integration-utils.js here? If `gotoPage` needs to be changed, then let's do that.

However...

> It considers navigation to be finished when there are no more than 0 fetch network requests for 500 ms. Since there are other types of network requests triggered automatically when loading index.html page, I think it self alone may not be sufficient.

Where do you see that `networkidle0` is specific to XHR requests? The [`page.goto` docs](https://github.com/puppeteer/puppeteer/blob/v13.1.3/docs/api.md#pagegotourl-options) do not mention that. Also, if we're concerned about additional network requests, then the [`DOMContentLoaded` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event) "fires ... without waiting for stylesheets, images, and subframes to finish loading."

aweng98: I'll change it back. Re `fetch`, I found it [here](https://github.com/puppeteer/puppeteer/issues/1552#issuecomment-350954419). He was one of original Puppeteer developers.

nawatts (approved):

breilly2 (approved):
Still kinda feel like this is more code (complexity) than needed for this, and I'd still prefer that we run alpha/staging tests against an isolated UI build to avoid this problem, but this is a fine stop-gap. 👍

aweng98:
Thanks all. I'll continue to monitore for this issue after PR merge.

PR #3209, opened 2022-07-07
[QA-1946] integration-test: Update noSpinnerAfter function

aweng98:
[QA-1946](https://broadworkbench.atlassian.net/browse/QA-1946)

noSpinnerAfter function assumes that a spinner will always be shown. But if UI does not show a spinner, the test function fails.

I had a meeting with Pete S. this afternoon to talk in details about `noSpinnerAfter` function. This PR summaries changes we agreed to.

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -171,9 +171,9 @@ const noSpinnersAfter = async (page, { action, debugMessage }) => {
   if (debugMessage) {
     console.log(`About to perform an action and wait for spinners. \n\tDebug message: ${debugMessage}`)
   }
-  const foundSpinner = page.waitForXPath('//*[@data-icon="loadingSpinner"]')
+  const foundSpinner = page.waitForXPath('//*[@data-icon="loadingSpinner"]', { timeout: 5000 })
   await Promise.all([
-    // loadingSpinner is not always visible. Ignore but log error if loadingSpinner is not found
+    // loadingSpinner is not always visible. Log but ignore error if loadingSpinner is not found
     foundSpinner.catch(e => { console.error(e) }),
     action()
   ])
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -172,7 +172,11 @@ const noSpinnersAfter = async (page, { action, debugMessage }) => {
     console.log(`About to perform an action and wait for spinners. \n\tDebug message: ${debugMessage}`)
   }
   const foundSpinner = page.waitForXPath('//*[@data-icon="loadingSpinner"]')
-  await Promise.all([foundSpinner, action()])
+  await Promise.all([
+    // loadingSpinner is not always visible. Ignore but log error if loadingSpinner is not found
+    foundSpinner.catch(e => { console.error(e) }),
+    action()
+  ])
   return waitForNoSpinners(page)
 }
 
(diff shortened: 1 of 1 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -171,8 +171,12 @@ const noSpinnersAfter = async (page, { action, debugMessage }) => {
   if (debugMessage) {
     console.log(`About to perform an action and wait for spinners. \n\tDebug message: ${debugMessage}`)
   }
-  const foundSpinner = page.waitForXPath('//*[@data-icon="loadingSpinner"]')
-  await Promise.all([foundSpinner, action()])
+  const foundSpinner = page.waitForXPath('//*[@data-icon="loadingSpinner"]', { timeout: 5000 })
```
aweng98, line 174: Display of loading spinner should not take 30 seconds. 5 seconds is enough time to wait for the loading spinner.

integration-tests/utils/integration-utils.js
```diff
@@ -171,8 +171,12 @@ const noSpinnersAfter = async (page, { action, debugMessage }) => {
...
+  await Promise.all([
+    // loadingSpinner is not always visible. Log but ignore error if loadingSpinner is not found
+    foundSpinner.catch(e => { console.error(e) }),
```
kyuksel: Was this meant to be

```suggestion
    foundSpinner.catch(e => console.error(e)),
```

?

integration-tests/utils/integration-utils.js
```diff
@@ -171,8 +171,12 @@ const noSpinnersAfter = async (page, { action, debugMessage }) => {
   if (debugMessage) {
     console.log(`About to perform an action and wait for spinners. \n\tDebug message: ${debugMessage}`)
   }
-  const foundSpinner = page.waitForXPath('//*[@data-icon="loadingSpinner"]')
-  await Promise.all([foundSpinner, action()])
+  const foundSpinner = page.waitForXPath('//*[@data-icon="loadingSpinner"]', { timeout: 5000 })
+  await Promise.all([
+    // loadingSpinner is not always visible. Log but ignore error if loadingSpinner is not found
```
breilly2, line 176: This goes against the guarantee that the function is still asserting. Do we think that the guarantee isn't important anymore or is this just going to lead to another type of flaky failure (such as what led us to add this function in the first place)?

integration-tests/utils/integration-utils.js
aweng98: Yes, Thanks.

integration-tests/utils/integration-utils.js
aweng98: Guarantee still is true for the 99.999% of time. @petesantos, can you help to explain this change? It's difficult for me to explain UI behavior with React.

aweng98: In case we don't reach an agreement, I'd revert back just this change and keep the other change in this PR.

petesantos: @aweng98 When we chatted about this, my hypothesis was as follows:
The browser attempts to render at 60FPS, approximately every 16ms.
When a request is made, a spinner is rendered. If the response is received within the 16 ms frame, react may run through the render function multiple times. This could result in no spinner ever getting rendered. This would mean there is no mutation event associated with the spinner.

To visualize, if each dash or below represents 1ms and the word blocks representing 1 ms it would look something like:
`- Request Render(spinner) - - - - - - - - Response - - - Render(no spinner) -(browser render) `

However, after we spoke, I experimented with this a little and could not reproduce this scenario. We may need some more investigation.

breilly2: Oh, that's really interesting.

My first thought was: I wonder if there's something we can do to guarantee that spinners will be rendered for at least one frame. However, instead of effectively artificially slowing down the app, can we detect the effect of the action in the problematic case and not care whether or not a spinner was rendered? In other words, can we somehow avoid the call to `noSpinnersAfter`?

PR #3210, opened 2022-07-07
[SU-162] Fix TSV download for JSON attributes

nawatts:
Currently, JSON attribute values in data tables are rendered in downloaded TSVs as `[object Object]`.

To reproduce this, use Rawls' [create entity endpoint](https://rawls.dsde-dev.broadinstitute.org/#/entities/create_entity) to create an entity with a JSON object attribute:
```
{
  "name": "test-1",
  "entityType": "test_entity",
  "attributes": {
    "object": {
       "key": "value"
    }
  }
}
```

When downloading selected rows as a TSV from the Export menu, the TSV is generated in the UI using the `Utils.entityAttributeText` function.
https://github.com/DataBiosphere/terra-ui/blob/9045a42b031d4c3426f6fcd9ef0f6bb4f776fee9/src/components/data/EntitiesContent.js#L238-L243

However, `entityAttributeText` does not return a string for some attribute types. In those cases, the text rendered in the TSV is the result of calling `toString` on the value returned from `entityAttributeText`. For attributes containing JSON objects, this results in the TSV containing `[object Object]`.

Added tests for the existing behavior of `entityTextAttribute` in 63cd32c535c42001c93df058f0551292b8c74f1c. Changes in behavior can be seen in b004e43323a814853df04c6cd4c7a1f7d786153a.

### DIFF
--- a/src/components/data/EntitiesContent.js
+++ b/src/components/data/EntitiesContent.js
@@ -7,7 +7,7 @@ import { Fragment, useState } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
 import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
 import { ButtonSecondary } from 'src/components/common'
-import { AddColumnModal, AddEntityModal, CreateEntitySetModal, EntityDeleter, ModalToolButton, MultipleEntityEditor, saveScroll } from 'src/components/data/data-utils'
+import { AddColumnModal, AddEntityModal, CreateEntitySetModal, entityAttributeText, EntityDeleter, ModalToolButton, MultipleEntityEditor, saveScroll } from 'src/components/data/data-utils'
 import DataTable from 'src/components/data/DataTable'
 import ExportDataModal from 'src/components/data/ExportDataModal'
 import { icon } from 'src/components/icons'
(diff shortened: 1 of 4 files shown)


breilly2:
One concern I have is that this accentuates the asymmetry between export/import for objects. If we start exporting JSON, I think there will be an expectation that we also import JSON into objects. Currently, importing results in a string value containing the JSON.

nawatts:
For better or worse, Terra does support JSON attributes in data tables, so we need to handle them as best we can. I think exporting the JSON encoded value is clearly better than exporting `[object Object]`.

JSON object attributes are exported correctly through the Orchestration API's [browserDownloadAttributes](https://api.firecloud.org/#/Workspaces/browserDownloadAttributes) endpoint (click "Download TSV" from the table menu in the Data tab sidebar to download from that API). After https://github.com/DataBiosphere/terra-ui/pull/3210/commits/dd211939ba70c8004b18194686c95480d54b8477, this makes partial TSVs exported through the UI consistent with TSVs exported through the API.

nawatts:
There are also already inconsistencies in export/import besides importing a JSON encoded object as a string. Data tables can contain attributes with JSON arrays. Importing a TSV with a JSON encoded array of objects results in an error ("Only arrays of strings, numbers, booleans, or entity references are supported."). Importing a TSV with a JSON encoded array of strings/numbers/booleans/entity references results in a list, not a JSON array.

MatthewBemis (approved):
Tested it out. Code looks good to me. 👍  The inconsistencies re: TSVs are unfortunate but I think that goes beyond this PR and I agree that this is a good improvement over the current broken behavior

breilly2 (approved):
👍 Nice tests!

PR #3212, opened 2022-07-08
[PE-5] RARE-X Landing Page Updates

kyuksel:
# Ticket
https://broadworkbench.atlassian.net/browse/PE-5

# Screenshots
## Terra
Styling changes only:

### Before
image

### After
image

### Before
image

### After
image

- The text and hyperlink changes are highlighted on the [ticket](https://broadworkbench.atlassian.net/browse/PE-5) description.
- Ran the changes by UX and PO on [this Slack thread](https://broadinstitute.slack.com/archives/C34GAKQ5R/p1657305260988549).
- Visually tested with and without the following config overrides:
  - `isRareX: true`
  - `isProjectSingular: true`
  - `isElwazi: true`
  - `isAnvil: true`
- VoiceOver announcements didn't indicate a regression as far as I could tell.
- **Note for code reviewers**: I'd recommend skipping white spaces via [?w=1](https://github.com/DataBiosphere/terra-ui/pull/3212/files?w=1).


src/libs/logos.js
```diff
@@ -23,15 +23,15 @@ import { isAnvil, isBaseline, isBioDataCatalyst, isDatastage, isElwazi, isFirecl
 import * as Utils from 'src/libs/utils'
 
 
-export const getAppName = (longName = false) => Utils.cond(
+export const getAppName = ({ longName = false, capitalInitial = false } = { longName: false, capitalInitial: false }) => Utils.cond(
```
nawatts: Is this additional complexity _really_ necessary? Can we just always capitalize the "The" like with AnVIL, Baseline, and elwazi? It's surprising that this option is inconsistently applied.

src/pages/LandingPage.js
```diff
@@ -67,20 +62,29 @@ const makeCard = (link, title, body) => h(Clickable, {
...
-const LandingPage = () => {
-  return h(HeroWrapper, { bigSubhead: true }, [
+const getSiteSpecificHyperlinks = () => Utils.cond(
+  [isRareX(), () => [
+    makeDocLink('https://rare-x.org/DataAnalysisPlatform-Documentation',
+      'Find RARE-X Data Analysis Platform documentation, tutorials and Jupyter notebook examples'),
+    makeDocLink('https://rare-x.org/researchers/', 'Learn more about the RARE-X Data Analysis Platform'),
+    makeDocLink('https://support.terra.bio/hc/en-us', 'Find Terra how-to\'s, documentation, video tutorials, and discussion forums')
+  ]],
+  () => [
```
nawatts: Should the links conditional on `isTerra` and `isFirecloud` be separate cases for this `Utils.cond` instead of grouped in the default option?

src/components/common.js
```diff
@@ -505,29 +505,30 @@ export const HeroWrapper = ({ showMenu = true, bigSubhead = false, showDocLink =
...
-        '. ',
-        showDocLink ?
-          h(Link, {
-            style: { textDecoration: 'underline' },
-            href: `https://support.terra.bio/hc/en-us`,
-            ...Utils.newTabLinkProps
-          }, ['Learn more about Terra.']) : null
-      ]),
-      children
+      div({ style: { maxWidth: '50%' } }, [
```
kyuksel, line 508: This was to limit the cases where text and doc links on the landing page from bleed into the hexagons on the hero image.

src/components/common.js
```diff
@@ -505,29 +505,30 @@ export const HeroWrapper = ({ showMenu = true, bigSubhead = false, showDocLink =
...
+        h1({ style: { fontSize: 54 } }, [`Welcome to ${getAppName({ longName: isTerra() })}`]),
```
kyuksel, line 509: @rtitle The change above was so the Welcome text isn't wrapped in an undesirable way (split into three lines), which would particularly affect `RARE-X` branding since that has a long name (i.e. The RARE-X Data Analysis Platform). The change was [approved by UX and PO](https://broadinstitute.slack.com/archives/C34GAKQ5R/p1657305260988549). I realized it essentially reverts your [change on PR #3117](https://github.com/DataBiosphere/terra-ui/commit/aff77d126b81cf22a60d6fff102bb3ee8222d186#diff-06c322acffe4a65c324584f6184927c7d4fca7e49530365c92671c0ea1fd30e1R526) so I wanted to check if I'm missing something.

src/libs/logos.js
kyuksel: The extensive defaulting is so a good number of call sites using the default values could remain concise.

nawatts: This could be simplified to `({ longName = false, capitalInitial = false } = {}) => ...`.

src/libs/nav.js
```diff
@@ -87,7 +87,7 @@ export const TitleManager = () => {
   const newTitle = Utils.cond(
     [_.isFunction(title), () => title({ ...params, queryParams: query })],
     [title, () => title],
-    getAppName
+    () => getAppName({ capitalInitial: true })
```
kyuksel: This is to capitalize `t` in `the` in the `RARE-X` site name (i.e. The RARE-X Data Analysis Platform) displayed on the site's browser tab.

src/pages/LandingPage.js
```diff
@@ -30,17 +30,12 @@ const styles = {
 }
 
 const makeDocLink = (href, title) => {
-  return div({
-    style: { marginBottom: '1rem', fontSize: 18, width: 600 }
```
kyuksel, line 34: `width` specification isn't needed / shouldn't be here. It's handled by the enclosing `div`'s `style`ing ([link](https://github.com/DataBiosphere/terra-ui/pull/3212/files?w=1#diff-42f0ee6fafb8d0d7920cee5508dee7b74b4607b26f0091ea6b33dbc3c880a32dR86)).

src/libs/logos.js
kyuksel: I was told by Product that RARE-X wanted their welcome text to be specifically `Welcome to the RARE-X Data Analysis Platform`, and not `Welcome to The RARE-X Data Analysis Platform` or `Welcome to RARE-X`.

src/pages/LandingPage.js
kyuksel: I was on the fence about it. Changed in [5084c9d](https://github.com/DataBiosphere/terra-ui/pull/3212/commits/5084c9df38630c09bbedaf932ac0fbedd764cc04).

src/libs/logos.js
kyuksel: I thought I had run into an issue with that but don't seem to be able to reproduce it. 
Changed in [665699a](https://github.com/DataBiosphere/terra-ui/commit/665699a31b5b949bc14817ea23100b2c9ca9b01c).

src/libs/logos.js
```diff
@@ -23,15 +23,15 @@ import { isAnvil, isBaseline, isBioDataCatalyst, isDatastage, isElwazi, isFirecl
...
+export const getAppName = ({ longName = false, capitalInitial = false } = {}) => Utils.cond(
   [isFirecloud(), () => 'FireCloud'],
   [isDatastage(), () => 'DataStage'],
   [isAnvil(), () => longName ? 'The NHGRI AnVIL (Genomic Data Science Analysis, Visualization, and Informatics Lab-space)' : 'AnVIL'],
   [isBioDataCatalyst(), () => 'NHLBI BioData Catalyst'],
   [isBaseline(), () => longName ? 'The Baseline Health Study Data Portal' : 'Project Baseline'],
   [isElwazi(), () => longName ? 'The eLwazi Open Data Science Platform' : 'eLwazi'],
   [isProjectSingular(), () => 'Project Singular'],
-  [isRareX(), () => longName ? 'The Rare Disease Database Platform' : 'RARE-X'],
+  [isRareX(), () => `${capitalInitial ? 'The' : 'the'} RARE-X Data Analysis Platform`],
```
nawatts: Replacing the hyphen here with a non-breaking hyphen (`\u2011`) would prevent this:

Screen Shot 2022-07-12 at 11 33 10 AM

src/pages/LandingPage.js
```diff
@@ -67,20 +64,29 @@ const makeCard = (link, title, body) => h(Clickable, {
...
+    makeDocLink('https://support.terra.bio/hc/en-us/articles/360033416912',
+      'Learn more about the Cancer Research Data Commons and other NCI Cloud Resources')
+  ]],
+  [isRareX(), () => [
+    makeDocLink('https://rare-x.org/DataAnalysisPlatform-Documentation',
+      'Find RARE-X Data Analysis Platform documentation, tutorials and Jupyter notebook examples'),
+    makeDocLink('https://rare-x.org/researchers/', 'Learn more about the RARE-X Data Analysis Platform'),
+    makeDocLink('https://support.terra.bio/hc/en-us', 'Find Terra how-to\'s, documentation, video tutorials, and discussion forums')
+  ]]
+)
```
nawatts, line 85: This is missing a default case. Currently, all themes show this link:
```js
makeDocLink('https://support.terra.bio/hc/en-us', 'Find how-to\'s, documentation, video tutorials, and discussion forums')
```

src/libs/logos.js
```diff
@@ -23,15 +23,15 @@ import { isAnvil, isBaseline, isBioDataCatalyst, isDatastage, isElwazi, isFirecl
 import * as Utils from 'src/libs/utils'
 
 
-export const getAppName = (longName = false) => Utils.cond(
+export const getAppName = ({ longName = false, capitalInitial = false } = {}) => Utils.cond(
```
nawatts: I think it would be preferable to rename this argument to clarify that it only applies to RARE-X.

src/libs/logos.js
kyuksel: Good catch! Done in [1b88e00](https://github.com/DataBiosphere/terra-ui/pull/3212/commits/1b88e00ddeaea844736149eb8b423bca06acecda).

src/libs/logos.js
kyuksel: An editor like IntelliJ makes it easy to find out the currently only use case in a snap, and I'd rather maintain the argument name reflect its functionality, which doesn't need to be specific to RARE-X. It would also cause 5 call sites to be further verbose. I'll change it if you feel strongly though.

src/pages/LandingPage.js
kyuksel: Ugh - thanks for catching this! Fixed in [e45589e](https://github.com/DataBiosphere/terra-ui/pull/3212/commits/e45589e68730eb53fb6507e826980dfab38d4c1d).

src/libs/logos.js
cahrens: I find `capitalInitial` to be confused. I expected it to be whether the first word of the application (FireCloud, eLwazi, RARE-X) was capitalized, not "the". And I would expect the default to be `true` instead of `false`.

Where is this desired with `false`? I'm wondering if something creative can be done to avoid the need to introduce this variable.

cahrens: @kyuksel can you possibly use https://www.geeksforgeeks.org/lodash-_-lowerfirst-method/ for the special place(s) where you want the first letter to be lowercase?

nawatts: With this change, the function no longer works as advertised. If I call `getAppName({ capitalInitial: false })` in AnVIL, I would expect to get "the NGHRI AnVIL ...". But that's not the case. This parameter's only purpose is to special case the RARE-X label.

kyuksel: (Discussed offline with @cahrens @nawatts)

Short-term changes are implemented in [283d720](https://github.com/DataBiosphere/terra-ui/pull/3212/commits/283d720aac2aa07c40c04775ec735f1393947512).

Will create a ticket to do some refactoring around the branding code for the longer-term.

cahrens:
It looks like the size of "New User?…" is too big.

image

kyuksel:
> It looks like the size of "New User?…" is too big.

They seem to be `36px` on both the PR'ed version

image

and the prod version

image

no?

cahrens (approved):

nawatts (approved):

PR #3213, opened 2022-07-10
[TUIM-20] Disable running analysis tests on dev.

cahrens:
All 4 of these tests failed at one run or the other on the various commits in this PR. Oddly, it doesn't seem like they all run all the time (or else perhaps the logic to split up the logs into test-based log files isn't working consistently).

### DIFF
--- a/integration-tests/tests/analysis-context-bar.js
+++ b/integration-tests/tests/analysis-context-bar.js
@@ -62,5 +62,6 @@ const testAnalysisContextBarFn = _.flow(
 registerTest({
   name: 'analysis-context-bar',
   fn: testAnalysisContextBarFn,
-  timeout: 15 * 60 * 1000
+  timeout: 15 * 60 * 1000,
+  targetEnvironments: ['local', 'alpha', 'perf', 'staging']
 })
--- a/integration-tests/tests/analysis-context-bar.js
+++ b/integration-tests/tests/analysis-context-bar.js
@@ -63,5 +63,5 @@ registerTest({
   name: 'analysis-context-bar',
   fn: testAnalysisContextBarFn,
   timeout: 15 * 60 * 1000,
-  targetEnvironments: ['local', 'alpha', 'perf', 'staging']
+  targetEnvironments: []
 })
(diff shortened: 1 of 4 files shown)


integration-tests/tests/analysis-context-bar.js
```diff
@@ -62,5 +62,6 @@ const testAnalysisContextBarFn = _.flow(
 registerTest({
   name: 'analysis-context-bar',
   fn: testAnalysisContextBarFn,
-  timeout: 15 * 60 * 1000
+  timeout: 15 * 60 * 1000,
+  targetEnvironments: ['alpha', 'perf', 'staging']
```
cahrens, line 66: The tests still run on the PR if 'local' is in the array, which isn't what I would have expected. Omitting both 'local' and 'dev'.

nawatts (approved):

integration-tests/tests/analysis-context-bar.js
nawatts: Yeah, since a local UI can be pointed at the APIs in any environment, having "local" as one of the environment options is not ideal.

We never actually run the tests with `environment` set to dev.

The `integration-tests-against-dev` job sets the `env` parameter to `local`.
https://github.com/DataBiosphere/terra-ui/blob/9045a42b031d4c3426f6fcd9ef0f6bb4f776fee9/.circleci/config.yml#L243-L249

Which gets passed to integration test code via the `ENVIRONMENT` environment variable.
https://github.com/DataBiosphere/terra-ui/blob/9045a42b031d4c3426f6fcd9ef0f6bb4f776fee9/.circleci/config.yml#L154-L165

And that's used to determine whether or not to run the test.
https://github.com/DataBiosphere/terra-ui/blob/9045a42b031d4c3426f6fcd9ef0f6bb4f776fee9/integration-tests/utils/jest-utils.js#L12
https://github.com/DataBiosphere/terra-ui/blob/9045a42b031d4c3426f6fcd9ef0f6bb4f776fee9/integration-tests/utils/jest-utils.js#L29

n8landolt (approved):

PR #3217, opened 2022-07-11
[SU-92] Improve screen reader output for collapse component

nawatts:
Currently, when tabbing to one of the headings in the workspace dashboard sidebar, VoiceOver reads out the entire content of that section. This is because the button has an [`aria-owns`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-owns) attribute telling VoiceOver that the main content of the Collapse component is a child of the button in the heading. After removing that attribute, VoiceOver only reads out the button text.

Removing `aria-owns` and leaving `aria-expanded` and `aria-controls` aligns with the ARIA Authoring Practices Guide's recommendations for a disclosure: https://w3c.github.io/aria-practices/#disclosure.

Screen Shot 2022-07-11 at 2 29 24 PM

## Before
Screen Shot 2022-07-11 at 2 30 15 PM

## After
Screen Shot 2022-07-11 at 2 30 39 PM

### DIFF
--- a/src/components/Collapse.js
+++ b/src/components/Collapse.js
@@ -25,7 +25,6 @@ const Collapse = ({ title, buttonStyle, buttonProps = {}, summaryStyle, detailsS
       h(Link, {
         'aria-expanded': isOpened,
         'aria-controls': isOpened ? id : undefined,
-        'aria-owns': isOpened ? id : undefined,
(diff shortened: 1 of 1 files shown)


cahrens (approved):
This works well, thanks!

MatthewBemis (approved):

PR #3219, opened 2022-07-11
[QA-1949] Terra-UI test: Update dismissNotification function to avoid sleep if not needed

aweng98:
[QA-1949](https://broadworkbench.atlassian.net/browse/QA-1949)

Little refactoring, simplify the code.

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -182,14 +182,16 @@ const delay = ms => {
 }
 
 const dismissNotifications = async page => {
-  await delay(3000) // delayed for any alerts to show
-  const notificationCloseButtons = await page.$x(
-    '(//a | //*[@role="button"] | //button)[contains(@aria-label,"Dismiss") and not(contains(@aria-label,"error"))]')
+  const notificationCloseButtonXpath = '(//a | //*[@role="button"] | //button)[contains(@aria-label,"Dismiss") and not(contains(@aria-label,"error"))]'
+  await Promise.race([
+    delay(3000), // delayed for any alerts to show
+    page.waitForXPath(notificationCloseButtonXpath, { visible: true })
+  ])
 
+  const notificationCloseButtons = await page.$x(notificationCloseButtonXpath)
   await Promise.all(
-    notificationCloseButtons.map(handle => handle.click())
+    notificationCloseButtons.map(button => button.evaluateHandle(btn => btn.click(), button))
   )
-
   return !!notificationCloseButtons.length && delay(1000) // delayed for alerts to animate off
 }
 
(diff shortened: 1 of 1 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -181,14 +181,16 @@ const delay = ms => {
...
+  const notificationCloseButtonXpath = '(//a | //*[@role="button"] | //button)[contains(@aria-label,"Dismiss") and not(contains(@aria-label,"error"))]'
+  await Promise.race([
+    delay(3000), // delayed for any alerts to show
+    page.waitForXPath(notificationCloseButtonXpath, { visible: true })
+  ])
 
+  const notificationCloseButtons = await page.$x(notificationCloseButtonXpath)
   await Promise.all(
-    notificationCloseButtons.map(handle => handle.click())
+    notificationCloseButtons.map(button => button.evaluateHandle(btn => btn.click(), button))
```
aweng98, line 194: `elementHandle.click()` is not reliable because it depends on element's (x, y) coordinates. If there are more than one notifications, after one notification is closed, position of next notification shifts (moves up or down). Click by coordinates when closing additional notifications can unexpectedly break.

nawatts:
This seems like it would work ok if there were only one notification. However, what if multiple notifications appear, both within the 3 second time frame but 1 second apart? It looks like this would see the first notification, close it, and then continue, leaving the second notification blocking parts of the UI.

aweng98:
I didn't expect that. Doesn't UI gurantees all notifications will show around the same time on landing page during log in?

kyuksel (commented):
> Doesn't UI gurantees all notifications will show around the same time on landing page during log in?

I didn't think so. Have you encountered code that indicates it does?

nawatts:
> Doesn't UI gurantees all notifications will show around the same time on landing page during log in?

Error notifications are often shown after an API request fails and there's no guarantee that requests all respond at exactly the same time.

aweng98:
Removed `Promise.race`. Reworked code. PTAL.

integration-tests/utils/integration-utils.js
```diff
@@ -182,15 +182,18 @@ const delay = ms => {
 }
 
 const dismissNotifications = async page => {
-  await delay(3000) // delayed for any alerts to show
-  const notificationCloseButtons = await page.$x(
-    '(//a | //*[@role="button"] | //button)[contains(@aria-label,"Dismiss") and not(contains(@aria-label,"error"))]')
-
+  const closeButtonXpath = '(//a | //*[@role="button"] | //button)[contains(@aria-label,"Dismiss") and not(contains(@aria-label,"error"))]'
+  try {
+    await page.waitForXPath(closeButtonXpath, { timeout: 3000 }) // 3 seconds wait for any alerts to show
```
nawatts, line 187: Again, this seems like it would work ok if there were only one notification. However, if multiple notifications appear with a slight delay between them, then it looks like this would see the first notification, close it, and then continue, leaving the second notification blocking parts of the UI.

aweng98: Sorry, missed that point new code. I just don't like the sleep. Closing PR.

PR #3222, opened 2022-07-11
[IA-3550] Fix for regionality conflict warning for workspaces with US multiregion buckets

wnojopra:
This warning was erroneously showing for workspaces with a US multi-regional bucket.
37d7d947-b103-436c-9d9f-d6b36366b75c
This is because the default workspace bucket location was changed, but this logic was not changed with it.

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -10,7 +10,7 @@ import { NumberInput, TextInput, ValidatedInput } from 'src/components/input'
 import { withModalDrawer } from 'src/components/ModalDrawer'
 import { getToolForImage, tools } from 'src/components/notebook-utils'
 import { InfoBox } from 'src/components/PopupTrigger'
-import { getAvailableComputeRegions, getLocationType, getRegionInfo, isUSLocation } from 'src/components/region-common'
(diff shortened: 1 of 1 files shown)


src/components/ComputeModal.js
```diff
@@ -661,13 +661,12 @@ export const ComputeModalBase = ({
...
 
   const isDifferentLocation = () => {
-    // If the bucket is regional, we can easily compare the bucketLocation with the compute region.
-    // bucketLocation === 'US' means the bucket is US multi-regional.
-    // For a US multi-regional bucket, the computeRegion needs to be US-CENTRAL1 in order to be considered "in the same location".
-    // Currently, US is the only multi-region supported in Terra
-    if (location === defaultLocation) {
+    if (isLocationMultiRegion(location)) {
+      // For a US multi-regional bucket, the computeRegion needs to be US-CENTRAL1 in order to be considered "in the same location".
+      // Currently, US is the only multi-region supported in Terra
```
kyuksel, line 666: I wonder if it may be too easy to forget to update the logic here if we were to start supporting another multi-regional bucket but I'm guessing that such a scenario is unlikely to be supported anytime soon to begin with.

kyuksel (approved):
The code change LGTM.
Didn't test.

breilly2 (approved):
👍 Code seems fine. I did not test because I don't have much knowledge/experience in the domain of Terra support for different regions.

PR #3223, opened 2022-07-12
[IA-3493] fix rstudio launch

jdcanas:
I highly recommend looking at the file in its current state instead of the github diff view. This code was extremely confusing for me, so I did a fairly large refactor. I split out the app conditionals (and other logic) into separate top-level `Util.conds`, instead of mixing them inside a few very large conditionals.

The issue was that the user would be inappropriately prompted to create a new RStudio when the runtime was not in a launchable status.

I did a lot of testing here, with both Jupyter and RStudio runtimes. Some examples of testing:

- Clicking an RStudio file with a Jupyter runtime and ensuring Open prompts runtime creation
    - Same applies to clicking a Jupyter file with an RStudio Runtime
- Clicking on RStudio/Jupyter files and then Open when the runtime is in Stopped should start the runtime
-  Clicking on RStudio/Jupyter files and then Open when the runtime is in running should launch the tool (means different things for both tools)
- Clicking on a file with no runtime should prompt runtime creation for the appropriate tool
- Workspace-level options should work (the `...` menu)
- Playground mode for Jupyter analyses should work

### DIFF
--- a/src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
+++ b/src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
@@ -227,6 +227,8 @@ const PreviewHeader = ({
   const { mode } = queryParams
   const analysisLink = Nav.getLink(analysisLauncherTabName, { namespace, name, analysisName })
 
+  const currentRuntimeTool = getToolFromRuntime(runtime)
+
   const checkIfLocked = withErrorReporting('Error checking analysis lock status', async () => {
     const { metadata: { lastLockedBy, lockExpiresAt } = {} } = await Ajax(signal)
       .Buckets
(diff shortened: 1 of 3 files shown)


src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -253,69 +254,76 @@ const PreviewHeader = ({
...
-                await clipboard.writeText(`${window.location.host}/${analysisLink}`)
-                notify('success', 'Successfully copied URL to clipboard', { timeout: 3000 })
-              })
-            }, ['Copy URL to clipboard'])
-          ]),
-          side: 'bottom'
+          }
+        },
+        openMenuIcon)],
+      // Jupyter is slightly different since it interact with editMode and playground mode flags as well
```
kyuksel: Nit: typo

```suggestion
      // Jupyter is slightly different since it interacts with editMode and playground mode flags as well
```

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
```diff
@@ -253,69 +254,76 @@ const PreviewHeader = ({
...
   useOnMount(() => { checkIfLocked() })
 
+  const openMenuIcon = [makeMenuIcon('rocket'), 'Open']
+
   return h(ApplicationHeader, {
     label: 'PREVIEW (READ-ONLY)',
     labelBgColor: colors.dark(0.2)
   }, [
+    // App-specific controls
     Utils.cond(
```
kyuksel, line 264: I had a hard time wrapping my head around the branching logic below (a lot of which predates this PR). It makes me wonder if they could be refactored such that all possible states could be part of an enum object. This may have the following benefits:

1. The states could be reviewed more easily by stakeholders like Product.
2. The names could be self-documenting.
3. We could unit-test them more easily.
4. It may be conducive to unnesting conditionals, which may expose potential issues (e.g. if we're branching on status being Stopped, why are we then checking if `runtime` is `truthy` in an inner conditional?)

```javascript
const states = {
  isReadOnlyAccess: 'isReadOnlyAccess',
  noModeOrStatusNullOrStopped: 'noModeOrStatusNullOrStopped', // Why can status be null? Can it be encoded better?,
  …….
}

// Helper that can be unit tested
const getState = () => Utils.cond(
  [readOnlyAccess, () => states.isReadOnlyAccess],
  [!mode || _.includes(runtimeStatus, [null, 'Stopped']), () => states.noModeOrStatusNullOrStopped,
  ……
)

// Rendering code
Utils.switchCase(getState(),
  [isReadOnlyAccess, () => h(HeaderButton, ……)],
  [noModeOrStatusNullOrStopped, () => h(Fragment, …….) ],
  ……
)
```

jdcanas: I'm having trouble understanding your suggestion... I think this incremental change is much more readable than the previous version, and it un-nests all the Utils.cond to the same level. I'm wary to make large sweeping changes as opposed to incremental improvement to this file, as this change is a fix to an issue users are currently seeing in production that as an added bonus makes it more readable

kyuksel: I agree and that all sounds good. To clarify, my comment was a thinking exercise around how we could further make the code easier to reason about, inspect and test as we spoke about refactoring.

kyuksel (approved):
Code changes LGTM. I suggest an IA person thoroughly test the changes manually.

src/pages/workspaces/workspace/notebooks/AnalysisLauncher.js
jdcanas: Yeah, I'd agree. Definitely something we should look for opportunities to refactor as it gets more complicated.

n8landolt (approved):

PR #3225, opened 2022-07-13
[IA-3516] Add runtime and storage cost to Analyses tab

n8landolt:
## Description

Add runtime and storage cost to the CloudEnvironmentModal, Galaxy modal, and as tooltips for the context bar.

## Testing

Tested an environment with only a jupyter runtime and persistent disk

Tested an environment with Juptyter and Galaxy runtimes

Tested an environment with Galaxy runtime

Note - the current dev instability makes things a little more difficult. Once the requests succeed, all information appears to be correct.

Screenshots:

Screen Shot 2022-07-19 at 12 25 24 PM
Screen Shot 2022-07-19 at 12 25 35 PM
Screen Shot 2022-07-19 at 12 25 42 PM
Screen Shot 2022-07-19 at 12 26 23 PM
Screen Shot 2022-07-19 at 12 26 29 PM


tiffanytuffany:
@n8landolt UX review: The "Rate" text looks fairly large compared to the cost text below it. Can we have them be the same size or have the cost "$0.00/hr" bigger? Since this is the main focus for the user. Also not sure if its possible to keep the bold just for the numbers and the "/hr" in regular font. 

For the hover text, can we bring the last sentence about workflows onto the next line? That way theres some separation of content Screen Shot 2022-07-15 at 4 28 47 PM

kyuksel:
@tiffanytuffany @n8landolt Is it communicated to users somewhere that the cost rates are _estimates_ and therefore should not be taken as bills?

tiffanytuffany:
@kyuksel I didn't realize those costs were estimates since I was basing it off of the existing cloud environment cost display. Have we just been displaying estimates as concrete numbers?
Screen Shot 2022-07-18 at 1 20 55 PM

n8landolt:
Might I suggest adding estimate in the tooltip text like so:
This rate reflects the **estimated** aggregate hourly cost..."

I'm planning to add that in (Just the word, no bold)

tiffanytuffany:
UX approved

src/components/ContextBar.js
```diff
@@ -43,6 +44,7 @@ export const ContextBar = ({
 }) => {
   const [isCloudEnvOpen, setCloudEnvOpen] = useState(false)
   const [selectedToolIcon, setSelectedToolIcon] = useState(undefined)
+  const [computeRegion] = useState(getRegionInfo(location, locationType).computeRegion)
```
jdcanas: do we need to use `useState` for this if it is never updated via a setter, or can it be a regular `const` variable?

src/components/ContextBar.js
```diff
@@ -117,11 +137,34 @@ export const ContextBar = ({
...
+            hover: contextBarStyles.hover,
+            tooltipSide: 'left',
+            tooltip: [
+              div('This rate reflects the estimated aggregate hourly cost for running and paused applications, as well as associated persistent disks. For more details, click on the Cloud icon.'),
+              br(),
+              div('Workflow and workspace storage costs are not included.')
+            ],
+            tooltipDelay: 100
+          }, [
+            div({ style: { textAlign: 'center', color: '#333F52', fontSize: '0.8em' } }, 'Rate'),
```
jdcanas: is it possible to use a color from the standard `color.js` palette for this? We generally like to avoid including colors directly into components, as terra has a few different skins, and using a keyword from colors.js ensures the colors make sense complimentary in other skins.

jdcanas: Seems like colors.dark is what you are looking for, same hex code

src/components/TooltipTrigger.js
```diff
@@ -16,14 +16,17 @@ const baseToolTip = {
 
 const styles = {
   tooltip: {
-    background: 'black', color: 'white',
-    padding: '0.5rem',
+    background: '#FFFFFF',
+    color: '#333F52',
+    borderColor: '#6D6E70',
```
jdcanas: could colors.grey be used here?

src/libs/runtime-utils.js
```diff
@@ -37,14 +37,18 @@ export const pdTypes = {
...
     regionToPricesName: 'monthlySSDDiskPrice'
   },
-  fromString: str => Utils.switchCase(str,
-    [pdTypes.standard.label, () => pdTypes.standard],
-    [pdTypes.balanced.label, () => pdTypes.balanced],
-    [pdTypes.ssd.label, () => pdTypes.ssd],
-    [Utils.DEFAULT, () => console.error(`Invalid disk type: Should not be calling pdTypes.fromString for ${str}`)]
-  )
+  fromString: str => {
+    return Utils.switchCase(str,
```
jdcanas: you shouldn't need the brackets and return, assuming its leftover from some prints that were put here?

src/libs/runtime-utils.js
```diff
@@ -37,14 +37,18 @@ export const pdTypes = {
...
+      [pdTypes.balanced.label, () => pdTypes.balanced],
+      [pdTypes.ssd.label, () => pdTypes.ssd],
+      [Utils.DEFAULT, () => console.error(`Invalid disk type: Should not be calling pdTypes.fromString for ${str}`)]
+    )
+  }
+}
+export const updatePdType = disk => {
+  return disk && _.update('diskType', pdTypes.fromString, disk)
 }
-export const updatePdType = disk => disk && _.update('diskType', pdTypes.fromString, disk)
```
jdcanas: the original form should be fine here too

src/libs/runtime-utils.js
```diff
@@ -209,7 +213,7 @@ const ephemeralExternalIpAddressCost = ({ numStandardVms, numPreemptibleVms }) =
   return numStandardVms * ephemeralExternalIpAddressPrice.standard + numPreemptibleVms * ephemeralExternalIpAddressPrice.preemptible
 }
 
-export const runtimeCost = ({ runtimeConfig, status }) => {
+export const getRuntimeCost = ({ runtimeConfig, status }) => {
   switch (status) {
```
jdcanas: I like the name change here. If you're feeling ambitious you could also update this code to use `Utils.switchcase` 🙏

src/libs/runtime-utils.js
```diff
@@ -339,6 +343,16 @@ export const getCurrentPersistentDisk = (appType, apps, appDataDisks, workspaceN
     )(appDataDisks))
 }
 
+export const getCurrentPersistentDisk = (runtimes, persistentDisks) => {
```
jdcanas, line 361: thanks for moving this to a better file

src/libs/runtime-utils.js
```diff
@@ -347,9 +361,62 @@ export const isCurrentGalaxyDiskDetaching = apps => {
 export const getGalaxyCostTextChildren = (app, appDataDisks) => {
   const dataDisk = getCurrentAttachedDataDisk(app, appDataDisks)
   return app ?
-    [getComputeStatusForDisplay(app.status), dataDisk ? ` (${Utils.formatUSD(getGalaxyCost(app, dataDisk))} / hr)` : ``] : ['None']
+    [getComputeStatusForDisplay(app.status), dataDisk ? ` ${Utils.formatUSD(getGalaxyCost(app, dataDisk))}/hr` : ``] : ['']
+}
+
+// TODO: multiple runtime: this is a good example of how the code should look when multiple runtimes are allowed, over a tool-centric approach
+export const getCostDisplayForTool = (app, appDataDisks, currentRuntime, currentRuntimeTool, toolLabel) => {
```
jdcanas: you shouldn't need the brackets and return here as well

src/components/ContextBar.js
```diff
@@ -117,11 +137,34 @@ export const ContextBar = ({
         await refreshRuntimes(true)
         await refreshApps()
       },
-      runtimes, apps, appDataDisks, refreshRuntimes, refreshApps, workspace, canCompute, persistentDisks, location, locationType
+      runtimes, apps, appDataDisks, refreshRuntimes, refreshApps, workspace, canCompute, persistentDisks, location, computeRegion
     }),
     div({ style: { ...Style.elements.contextBarContainer, width: 70 } }, [
       div({ style: contextBarStyles.contextBarContainer }, [
         h(Fragment, [
+          h(Clickable, {
```
jdcanas: I don't know if clickable is the right element to use here. This results in it having a hover-over color, and the cursor changing into a `click here` affordance. Was a bit surprised nothing happened when I clicked based on this. A div or the like may be more applicable

image

jdcanas:
few nits, looks good overall

n8landolt:
General conversation - do we know that by using the colors object that we'll be compliant for different UIs?
Screen Shot 2022-07-19 at 1 34 09 PM

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -378,7 +359,12 @@ export const CloudEnvironmentModal = ({
             style: { height: 30 },
             alt: `${toolLabel}`
           }),
-          getCostForTool(toolLabel)
+          div(
```
jdcanas: seems a bit extra to have the open array on a newline (and the `])` close), surprised the linter is happy with it

cahrens:
> General conversation - do we know that by using the colors object that we'll be compliant for different UIs?

It won't be guaranteed, but your odds will be better. When I've been doing the colors for a white label site, I do check color contrast and adjust as necessary. That being said, I don't know all the places every color is used, so it's possible there are usages that are problematic.

cahrens:
@n8landolt can you let me know when you have addressed @jdcanas's comments so I know when to review? If there are specific things you want me to look at earlier, just tag me on them.

src/components/TooltipTrigger.js
```diff
@@ -16,14 +16,17 @@ const baseToolTip = {
 
 const styles = {
   tooltip: {
-    background: 'black', color: 'white',
-    padding: '0.5rem',
+    background: '#FFFFFF',
```
cahrens: I am thrilled to see the tooltip color changing… I hated the black. However, something needs to be done to make the arrow pointer render better.

image

image

src/components/ContextBar.js
```diff
@@ -44,6 +48,7 @@ export const ContextBar = ({
   const [isCloudEnvOpen, setCloudEnvOpen] = useState(false)
   const [selectedToolIcon, setSelectedToolIcon] = useState(undefined)
 
+  const computeRegion = getRegionInfo(location, locationType).computeRegion
```
cahrens, line 51: This works for Azure workspaces too?

src/components/ContextBar.js
```diff
@@ -101,6 +113,17 @@ export const ContextBar = ({
     ])
   }
 
+  //This excludes cromwellapp in the calculation.
```
cahrens, line 116: Are you excluding Cromwell because cost calculations are not available for it? 

Ideally it would be nice to just iterate though each of the tools and ask them for their cost (and make it up to the tools to figure out their cost… Cromwell could just return 0 for now).

src/components/ContextBar.js
```diff
@@ -101,6 +113,17 @@ export const ContextBar = ({
...
+  const getTotalToolAndDiskCostDisplay = () => {
+    const galaxyApp = getCurrentApp(tools.Galaxy.appType)(apps)
+    const galaxyDisk = getCurrentAppDataDisk(tools.Galaxy.appType, apps, appDataDisks, workspaceName)
+    const galaxyCost = galaxyApp && galaxyDisk ? getGalaxyCost(galaxyApp, galaxyDisk) : 0
+    const runtimeCost = currentRuntime ? getRuntimeCost(currentRuntime) : 0
+    const curPd = getCurrentPersistentDisk(runtimes, persistentDisks)
+    const diskCost = curPd ? getPersistentDiskCostHourly(curPd, computeRegion) : 0
+    return `${Utils.formatUSD(galaxyCost + runtimeCost + diskCost)}`
```
cahrens: Why is this in US dollars? If someone is using Montreal, will the correct costs be displayed?

src/libs/runtime-utils.js
```diff
@@ -347,9 +355,60 @@ export const isCurrentGalaxyDiskDetaching = apps => {
 export const getGalaxyCostTextChildren = (app, appDataDisks) => {
   const dataDisk = getCurrentAttachedDataDisk(app, appDataDisks)
   return app ?
-    [getComputeStatusForDisplay(app.status), dataDisk ? ` (${Utils.formatUSD(getGalaxyCost(app, dataDisk))} / hr)` : ``] : ['None']
+    [getComputeStatusForDisplay(app.status), dataDisk ? ` ${Utils.formatUSD(getGalaxyCost(app, dataDisk))}/hr` : ``] : ['']
```
cahrens, line 379: Again, I worry about USD.

src/libs/runtime-utils.js
```diff
@@ -347,9 +355,60 @@ export const isCurrentGalaxyDiskDetaching = apps => {
...
+}
+
+// TODO: multiple runtime: this is a good example of how the code should look when multiple runtimes are allowed, over a tool-centric approach
+export const getCostDisplayForTool = (app, appDataDisks, currentRuntime, currentRuntimeTool, toolLabel) => Utils.cond(
```
cahrens: Or the tools themselves have a method for their cost.

src/libs/runtime-utils.js
```diff
@@ -347,9 +355,60 @@ export const isCurrentGalaxyDiskDetaching = apps => {
...
+  [toolLabel === tools.Galaxy.label, () => getGalaxyCostTextChildren(app, appDataDisks)],
+  [toolLabel === tools.Cromwell.label, () => ''], // We will determine what to put here later
+  [toolLabel === tools.Azure.labels, () => ''], //TODO: Azure cost calculation
+  [getRuntimeForTool(toolLabel, currentRuntime, currentRuntimeTool), () => {
+    const runtime = getRuntimeForTool(toolLabel, currentRuntime, currentRuntimeTool)
+    const totalCost = getRuntimeCost(runtime)
+    return `${getComputeStatusForDisplay(runtime.status)} ${Utils.formatUSD(totalCost)}/hr`
```
cahrens: Another place where USD is assumed.

src/libs/runtime-utils.js
```diff
@@ -347,9 +355,60 @@ export const isCurrentGalaxyDiskDetaching = apps => {
 export const getGalaxyCostTextChildren = (app, appDataDisks) => {
   const dataDisk = getCurrentAttachedDataDisk(app, appDataDisks)
   return app ?
-    [getComputeStatusForDisplay(app.status), dataDisk ? ` (${Utils.formatUSD(getGalaxyCost(app, dataDisk))} / hr)` : ``] : ['None']
+    [getComputeStatusForDisplay(app.status), dataDisk ? ` ${Utils.formatUSD(getGalaxyCost(app, dataDisk))}/hr` : ``] : ['']
+}
+
+// TODO: multiple runtime: this is a good example of how the code should look when multiple runtimes are allowed, over a tool-centric approach
+export const getCostDisplayForTool = (app, appDataDisks, currentRuntime, currentRuntimeTool, toolLabel) => Utils.cond(
+  [toolLabel === tools.Galaxy.label, () => getGalaxyCostTextChildren(app, appDataDisks)],
```
cahrens: Can't this method call `getCostForTool` and just format the result?

src/libs/runtime-utils.js
```diff
@@ -347,9 +355,60 @@ export const isCurrentGalaxyDiskDetaching = apps => {
...
+    return getRuntimeCost(runtime)
+  }],
+  [Utils.DEFAULT, () => {
+    return ''
+  }]
+)
+
+export const getCostDisplayForDisk = (app, appDataDisks, computeRegion, currentRuntimeTool, persistentDisks, runtimes, toolLabel) => {
+  const diskCost = getCostForDisk(app, appDataDisks, computeRegion, currentRuntimeTool, persistentDisks, runtimes, toolLabel)
+  return diskCost ? `Disk ${Utils.formatUSD(getCostForDisk(app, appDataDisks, computeRegion, currentRuntimeTool, persistentDisks, runtimes, toolLabel))}/hr` : ''
```
cahrens: Why do you call `getCostForDisk` again instead of using the variable you just defined?

src/components/TooltipTrigger.js
cahrens: @n8landolt I will take on fixing this issue in https://broadworkbench.atlassian.net/browse/TUIM-23. Please back out the change for now.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -448,3 +571,119 @@ describe('getDisplayList', () => {
...
+    const toolLabel = tools.Galaxy.label
+
+    //Act
+    const result = getCostDisplayForTool(app, currentRuntime, currentRuntimeTool, toolLabel)
+
+    //Assert
+    expect(result).toBe(expectedResult)
+  })
+  it('will get compute cost and compute status for Jupyter runtime', () => {
+    //Arrange
```
msilva-broad: I favor styling these with all caps... and comments should always have a space after the // or /* delimiter

// ARRANGE

// ACT

// ASSERT

this makes secondary comments within each section look secondary, and these will still stand out.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -253,6 +272,106 @@ const cromwellProvisioningDiskUpdatedPd = {
   zone: 'us-central1-a'
 }
 
+const jupyterStopped = {
```
msilva-broad: if these mock data consts are only used in one test, I would move them to be in the // ARRANGE section of that test.  If one or more of them are used in more then one test, I would move them to the top of the inside of the appropriate describe arrow function block.  Module scope should be the last resort.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -1,9 +1,28 @@
...
+        n1HourlyGBRamPrice: 0.004237, n1HourlyCpuPrice: 0.031611, preemptibleN1HourlyGBRamPrice: 0.000892, preemptibleN1HourlyCpuPrice: 0.006655,
+        t4HourlyPrice: 0.35, p4HourlyPrice: 0.6, k80HourlyPrice: 0.45, v100HourlyPrice: 2.48, p100HourlyPrice: 1.46,
+        preemptibleT4HourlyPrice: 0.11, preemptibleP4HourlyPrice: 0.216, preemptibleK80HourlyPrice: 0.0375,
+        preemptibleV100HourlyPrice: 0.74, preemptibleP100HourlyPrice: 0.43
+      }
+    ]
+  }
+})
+
+jest.clearAllMocks(true)
```
msilva-broad: I am curious why this is needed.... could it be that any weirdness you were seeing could be fixed by changing regionToPrices export to an arrow function that returns that array of info instead of the array directly?

src/libs/_tests/runtime-utils.test.js
kyuksel: @msilva-broad It'd be good to add any styling guidelines to https://docs.google.com/document/d/1cwQ2IckdSBPqVYUGLjk-Z1zq6qwunml4wkavPoDH48s/edit#heading=h.9np93p8k8ga8 so we can review them more easily and enforce them for all :)

cahrens: I personally don't want to have to us all caps for comments. Definitely something to discuss before adding to the style guidelines.

src/libs/runtime-utils.js
```diff
@@ -325,7 +323,7 @@ export const appIsSettingUp = app => {
   return app && (app.status === 'PROVISIONING' || app.status === 'PRECREATING')
 }
 
-export const getCurrentPersistentDisk = (appType, apps, appDataDisks, workspaceName) => {
+export const getCurrentAppDataDisk = (appType, apps, appDataDisks, workspaceName) => {
```
msilva-broad, line 336: I think /** jsdoc */ style function comment blocks would go a long way to documenting these methods' intended use... me coming in cold, I have no idea why this one was renamed getCurrentAppDataDisk, and a different implementation is being put in for getCurrentPersistentDisk below.

src/components/ContextBar.js
cahrens: For clarity, the IA team will follow up about localization/currency in a future ticket.

src/libs/runtime-utils.js
```diff
@@ -351,9 +359,45 @@ export const isCurrentGalaxyDiskDetaching = apps => {
 export const getGalaxyCostTextChildren = (app, appDataDisks) => {
   const dataDisk = getCurrentAttachedDataDisk(app, appDataDisks)
   return app ?
-    [getComputeStatusForDisplay(app.status), dataDisk ? ` (${Utils.formatUSD(getGalaxyCost(app, dataDisk))} / hr)` : ``] : ['None']
+    [getComputeStatusForDisplay(app.status), dataDisk ? ` ${Utils.formatUSD(getGalaxyCost(app, dataDisk))}/hr` : ``] : ['']
 }
 
+// TODO: multiple runtime: this is a good example of how the code should look when multiple runtimes are allowed, over a tool-centric approach
+export const getCostDisplayForTool = (app, currentRuntime, currentRuntimeTool, toolLabel) => {
```
msilva-broad, line 383: as was discussed in slack.  If these methods are only used by one visual component (especially since they mention "display" in their name), lets move them over to live next to the visual component as whatever-component-utils.js.  And if this lets you move over your unit tests and mocks, great!

If you want to be conservative, you can move the unit tests over to a _tests folder inside the appropriate folder... src/components I think.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -1,9 +1,28 @@
 import { tools } from 'src/components/notebook-utils'
 import {
-  getAnalysesDisplayList, getCurrentApp, getCurrentAppDataDisk, getCurrentAppIncludingDeleting, getDiskAppType, workspaceHasMultipleApps, workspaceHasMultipleDisks
+  getAnalysesDisplayList, getCostDisplayForTool, getCurrentApp, getCurrentAppDataDisk, getCurrentAppIncludingDeleting, getDiskAppType, workspaceHasMultipleApps, workspaceHasMultipleDisks
 } from 'src/libs/runtime-utils'
 
 
+jest.mock('src/data/machines', () => {
```
cahrens, line 7: It's very cool to see how this mocking works!

Can you say more about why you decided to mock the return values for 'US-CENTRAL1'? I get that if these values change, the unit test would have to be updated, but I would expect to make that type of change.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -378,6 +397,10 @@ const mockBucketAnalyses = [
       }
   }
 ]
+//Expecting (.toBeCalled()) only one call needs to clear between tests
```
cahrens: I don't understand what this comment means.

src/libs/_tests/runtime-utils.test.js
```diff
@@ -474,11 +574,114 @@ describe('getDisplayList', () => {
...
+    const toolLabel = tools.Jupyter.label
+    const expectedResult = 'Disk < $0.01/hr'
+
+    //Act
+    const result = getCostDisplayForDisk(app, appDataDisks, computeRegion, currentRuntimeTool, persistentDisks, runtimes, toolLabel)
+
+    //Assert
+    expect(result).toBe(expectedResult)
+  })
+  it('will return empty string', () => {
```
cahrens: Can you add why an empty string is expected (what this case covers)?

src/libs/_tests/runtime-utils.test.js
```diff
@@ -272,7 +272,7 @@ const cromwellProvisioningDiskUpdatedPd = {
   zone: 'us-central1-a'
 }
 
-const jupyterStopped = {
+const jupyter1 = {
```
cahrens, line 273: If the fact that it is running is important, I would make that part of the variable name (`jupyterRunning`).

cahrens: Or consider always overriding the status where it is used for clarity (I see below that you override this for Stopped).

cahrens:
I seem to get into some inconsistent states between the cost summary and what the tools/apps individually say, especially when switching between tabs within the workspace and reloading Terra.

Sometimes the running Galaxy app doesn't appear at all (especially on the Analyses tab). Sometimes it does, but total cost displays as $0. Sometimes it shows and the total cost is correct. Each option seems to happen with pretty equal frequency.

image
image

Note that when "Azure environment" is showing, I'm actually in a GCP workspace, not an Azure one.
image

src/libs/_tests/runtime-utils.test.js
msilva-broad: great stuff!  This effort was meant to drive discussion.  The footprint is small.  I suggest we ask Nathan to have a follow-up PR to make further tweaks that come out of us discussing this topic at the UIWG.  I wouldn't want to hold up his PR.

msilva-broad: an yes, @kyuksel once we have agreement in UIWG, we can update the styling guidelines.

src/libs/_tests/runtime-utils.test.js
msilva-broad: Nathan and I discussed this very question as he was getting started on this work.  My thinking was that over time (if not already), this config object being exported could be used in a number of areas and unit/functional tests.  It would be a bit onerous to have to update all of them for every little config change if the intrinsic shape/meaning of the object has no substantive impact on the consuming code's logic.  

So as long as the shape (contract) of the dependency isn't changing, the unit tests should not need updates.   It also becomes crisper to understand the minimal stub data required to drive the code being tested, and give the test file the ability to manufacture key variations of the dependency that may not be guaranteed to exist in the real object at a given time.  Also, this gets easier to manage with Typescript, which is why it's a topic I plan to raise for discussion at this week's UIWG.

 But the goal is to drive discussion and feedback :)

src/libs/_tests/runtime-utils.test.js
kyuksel: To clarify, that doc is just a medium I've used to jot down styling guidelines that've come up throughout the past year in PR reviews, mobings, etc., in hopes of solidifying them as part of UIWG so don't worry about adding to it. All of the existing bullets will need buy-ins as well.

cahrens:
With the latest commit, the information shown in the total cost more accurately represents what is shown in the tooltip for the individual app. Note that it is possible for the disk cost to not return, or the app cost to not return (in which case the app isn't shown at all).

image
image
image

Jupyter creation isn't working at all well, so it's hard to say what is going on in that case. All of these things will hopefully be improved/fixed in future iterations or dev environment fixes.

cahrens (approved):

Qi77Qi (approved):
PR site looks good

PR #3227, opened 2022-07-13
[PE-5] Adjust landing page layout

nawatts:
For some window widths, the text on the cards on the landing page overflows.

The cause of this is a `max-width: 50%` style added in https://github.com/DataBiosphere/terra-ui/pull/3212/files#r918374677 to prevent content from overlapping the background image. This change uses `width: 'calc(100% - <number>px')` individually on the heading, paragraph, and doc link list elements to accomplish that.

## Before

before_terra_2

before_terra

before_rarex

## After

after_terra_2

after_terra

after-rarex

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -505,30 +505,28 @@ export const HeroWrapper = ({ showMenu = true, bigSubhead = false, showDocLink =
         backgroundRepeat: 'no-repeat', backgroundSize: '750px', backgroundPosition: 'right 0 top 0'
       }
     }, [
-      div({ style: { maxWidth: '50%' } }, [
-        h1({ style: { fontSize: 54 } }, [`Welcome to ${getAppName({ longName: isTerra() })}`]),
-        div({ style: { margin: '1rem 0', width: 700, ...(bigSubhead ? { fontSize: 20, lineHeight: '28px' } : { fontSize: 16, lineHeight: 1.5 }) } }, [
-          `${getAppName({ longName: !isTerra(), capitalizeThe: true })} is a ${Utils.cond(
-            [isTerra(), () => 'cloud-native platform'],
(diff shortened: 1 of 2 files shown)


breilly2 (approved):
👍 Thanks for fixing this!

src/components/common.js
```diff
@@ -505,30 +505,28 @@ export const HeroWrapper = ({ showMenu = true, bigSubhead = false, showDocLink =
...
-            h(Link, {
-              style: { textDecoration: 'underline' },
-              href: `https://support.terra.bio/hc/en-us`,
-              ...Utils.newTabLinkProps
-            }, ['Learn more about Terra.']) : null
-        ]),
-        children
-      ])
+      h1({ style: { fontSize: 54, width: 'calc(100% - 460px)' } }, [`Welcome to ${getAppName({ longName: isTerra() })}`]),
+      div({ style: { margin: '1rem 0', width: 'calc(100% - 460px)', maxWidth: 700, ...(bigSubhead ? { fontSize: 20, lineHeight: '28px' } : { fontSize: 16, lineHeight: 1.5 }) } }, [
```
kyuksel, line 510: Would it be good to have one source for the value `460` and add a comment about why that value was chosen? I feel it may be too easy to modify it in one place and not the other, and not notice the consequences.

kyuksel (approved):
Thank you for the fix!
One comment inline.
Tested locally.

src/components/common.js
nawatts: Added a comment in https://github.com/DataBiosphere/terra-ui/pull/3227/commits/4b808727f5e72865d58677f47b57feaf0b2794e0. I don't think we really need to optimize for changes to the background image. Such a change is probably rare. And if it does change, the content of the background image isn't necessarily rectangular, so widths of the different elements wouldn't necessarily all need to be the same (currently, one could argue for allowing the h1 element to be wider).

PR #3229, opened 2022-07-13
[SU-146] Ability to star workspaces in workspace list

MatthewBemis:
This PR allows a user to star a workspace in their list view. Starred workspaces are always floated to the top, even when sorting the list.

Due to a column size limit in Thurloe's database, we can only support 50 starred workspaces per user. This should be more than enough, even for power users. If a user wants to star more than 50 workspaces, they are either abusing the star feature or that may be a sign that we need to implement additional organization features to the workspace list page (perhaps folders).

For the related Orch PR, see https://github.com/broadinstitute/firecloud-orchestration/pull/966 which made it possible to add a `starredWorkspaces` key

Screen Shot 2022-07-14 at 10 22 02 AM

<!--
Hello, friend!
Thanks!

--!>


src/components/workspace-utils.js
```diff
@@ -340,6 +342,46 @@ export const WorkspaceTagSelect = props => {
...
+  const { workspace: { workspaceId } } = workspace
+  const isStarred = _.includes(workspaceId, starredWorkspaceIds)
+
+  const toggleStar = _.flow(
+    Utils.withBusyState(setUpdating),
+    withErrorReporting(`Unable to ${isStarred ? 'unstar' : 'star'} workspace`)
+  )(async star => {
+    if (star) {
+      if (_.size(starredWorkspaceIds) < 50) {
+        const updatedWorkspaceIds = _.concat(starredWorkspaceIds, [workspaceId])
```
nawatts: Since one attribute contains all the starred workspaces, it would be good to reload the list immediately before updating. Currently, if you open two tabs and star workspaces in each one, the first one gets overwritten.

src/components/workspace-utils.js
```diff
@@ -340,6 +342,46 @@ export const WorkspaceTagSelect = props => {
...
+      }
+    } else {
+      const updatedWorkspaceIds = _.without([workspaceId], starredWorkspaceIds)
+      await Ajax().User.profile.setPreferences({ starredWorkspaces: _.join(',', updatedWorkspaceIds) })
+      authStore.update(_.set('profile.starredWorkspaces', updatedWorkspaceIds))
+    }
+    onUpdate(star)
+  })
+
+  return h(Interactive, {
```
nawatts: Should this button be disabled and show a tooltip if there are 50 workspaces starred?

src/pages/workspaces/List.js
```diff
@@ -143,9 +144,14 @@ export const WorkspaceList = () => {
     }),
     initialFiltered), [accessLevelsFilter, filter, initialFiltered, projectsFilter, submissionsFilter, tagsFilter])
 
+  const { profile: { starredWorkspaces } } = useStore(authStore)
+  const starredWorkspaceIds = _.isEmpty(starredWorkspaces) ? [] : _.split(',', starredWorkspaces)
+
+  //Starred workspaces are always floated to the top
```
nawatts, line 155: It's unfortunate that this makes workspaces change place in the list when they are starred/unstarred. But I also think this sort order makes sense and I don't see a good way around that.

nawatts (commented):
Has the associated Orchestration change already gone out to production?

MatthewBemis:
@nawatts it's scheduled to go out to prod on Monday. I'm planning to hold this PR back as needed

src/components/workspace-utils.js
```diff
@@ -340,6 +342,46 @@ export const WorkspaceTagSelect = props => {
...
+
+  const { workspace: { workspaceId } } = workspace
+  const isStarred = _.includes(workspaceId, starredWorkspaceIds)
+
+  const toggleStar = _.flow(
+    Utils.withBusyState(setUpdating),
+    withErrorReporting(`Unable to ${isStarred ? 'unstar' : 'star'} workspace`)
+  )(async star => {
+    if (star) {
+      if (_.size(starredWorkspaceIds) < 50) {
```
cahrens: I would suggest a code comment explaining the 50 star limitation.

src/pages/workspaces/List.js
cahrens: I think it's pretty confusing/unexpected that it disappears when you toggle the star. I feel like what I've seen in other UIs is that the item you have toggled remains where it is until you refresh the page or select a sort option in the table.

cahrens: And I think that would be easy to do by removing the `onUpdate` call to refresh the workspaces.

Could you touch base with UX about the experience?

src/components/workspace-utils.js
```diff
@@ -340,6 +342,46 @@ export const WorkspaceTagSelect = props => {
...
+      authStore.update(_.set('profile.starredWorkspaces', updatedWorkspaceIds))
+    }
+    onUpdate(star)
+  })
+
+  return h(Interactive, {
+    as: 'span',
+    role: 'checkbox',
+    'aria-checked': isStarred,
+    tooltip: isStarred ? 'Unstar this workspace' : 'Star this workspace',
```
cahrens: I am not seeing a tooltip, nor am I hearing anything with the screenreader that explains what this control is (I just get "checkbox unchecked" or something to that effect.

I think we should consider adding in a tooltip the impact of starting a workspace ("Starred workspaces are listed first" or something like that).

cahrens: image

src/pages/workspaces/List.js
MatthewBemis: It is unfortunately a bit more convoluted than just removing the onUpdate call. A bunch of state and sorting code needs to be updated as well. I am checking with UX about this now

cahrens: @MatthewBemis any update?

MatthewBemis: Not yet, but I think we will have this resolved with Joy soon. I should have put this back in draft in the meantime- I'll do that now.

MatthewBemis: @cahrens the conclusion with Joy is that we will keep the behavior as-is, but we will of course be listening carefully for user feedback and adjust the behavior if needed

src/components/workspace-utils.js
```diff
@@ -378,7 +378,7 @@ export const WorkspaceStarControl = ({ workspace, starredWorkspaceIds, onUpdate,
     as: 'span',
     role: 'checkbox',
     'aria-checked': isStarred,
-    tooltip: isStarred ? 'Unstar this workspace' : 'Star this workspace',
+    tooltip: isStarred ? 'Unstar this workspace' : 'Star this workspace. Starred workspace will appear at the top of your workspace list',
```
cahrens: Can you add a period at the end of the sentence?

src/components/workspace-utils.js
MatthewBemis: I switched this from an Interactive to a Clickable, which seems to improve the accessibility. I also added an aria label to announce when a workspace is starred, rather than just "checkbox checked". I am not sure if I should also announce if a workspace is unstarred- I chose not to because that is the default state for a workspace and I didn't want it to be too chatty. The tooltip also works as expected now as a Clickable

Improved in acb0d64fbd2c7f03e7ad3cb83a5547cc60d17968

Also added a better description of what starring a workspace does in https://github.com/DataBiosphere/terra-ui/pull/3229/commits/aa33ae1d07f13e12701536423b2cce5dac68b702

src/components/workspace-utils.js
MatthewBemis: Done in 740f866cc7b1bbea5687bd00973264508f053d44

src/components/workspace-utils.js
MatthewBemis: Done in 67a0acd5ceb98455fdd74f399aabdbf6ff917aaf

src/components/workspace-utils.js
MatthewBemis: Handled in 7deb3043e3c8496d8ab0a2363ad08706d21109b8

Note that I did make a small tweak to that code in https://github.com/DataBiosphere/terra-ui/pull/3229/commits/2a80a5abd17abe0b43cb61425e0ae30881a6219d

cahrens:
@MatthewBemis when I toggle a star, the keyboard focus does not go to a logical place. I would expect keyboard focus to remain on the star, but it seems to go to a random star in the list (when I press Tab after the refresh).

Also, the experience with a screen reader is pretty awful. When you toggle a star, what is read out after the refresh is "My Workspaces (100)", which is the name of the tab. Hopefully ensuring that focus returns to the correct element will improve this.

src/components/workspace-utils.js
```diff
@@ -340,6 +341,57 @@ export const WorkspaceTagSelect = props => {
...
+    onUpdate(star)
+  })
+
+  return h(Clickable, {
+    as: 'span',
+    role: 'checkbox',
+    'aria-checked': isStarred,
+    tooltip: Utils.cond(
+      [isStarred, () => 'Unstar this workspace'],
+      [!isStarred && !maxStarredWorkspacesReached, () => 'Star this workspace. Starred workspace will appear at the top of your workspace list.'],
```
cahrens: Should be "Starred workspace**s** will appear…"

MatthewBemis: corrected in https://github.com/DataBiosphere/terra-ui/pull/3229/commits/66ea3d50cc4aec80721fb96ee70b6a6d0bf7b671

nawatts (commented):
> when I toggle a star, the keyboard focus does not go to a logical place. I would expect keyboard focus to remain on the star, but it seems to go to a random star in the list.

Seconding this. This could possibly be resolved by programmatically focusing the star control after updating, but there may be similar focus issues if a user tabs to another control while the star status is updating. This may be another point in favor of preserving the sort order when a workspace is starred/unstarred.

MatthewBemis:
@cahrens @nawatts apologies for the initial resistance- I'm on board with trying to preserve the order. There are multiple very good reasons at this point...

MatthewBemis:
@cahrens @nawatts I've updated it so the workspace remains in-place if the star is toggled. This also seems to have helped with the screen reader readout and focus issues.

Currently the behavior is that the workspaces will remain in-place until the page is refreshed. It's proving a bit tricky to have that happen in an elegant way when a sort option is changed. I tried using an effect to detect if the sort was changed and then trigger a refresh of the list of IDs that are starred, but was unable to get it to work.

cahrens:
@MatthewBemis I'm still seeing odd behavior on the PR site. When I toggle the star via the keyboard, the list scrolls so that what I just toggled is not visible. The keyboard focus does not appear to remain on the star, as when I press Tab it goes to an incorrect location (and if I scroll up to find what I toggled, the star does not have the focus indication).

MatthewBemis:
@cahrens which browser are you using? this is the behavior that I see:

https://user-images.githubusercontent.com/7257391/181565017-b8a2fe15-ba53-4efd-bedf-6fd655baef5d.mov

cahrens:
> @cahrens which browser are you using? this is the behavior that I see:

That looks lovely. I'm using Chrome. When I use the mouse it behaves nicely, but when I use the keyboard to toggle the star it doesn't. I'm using space though to toggle… I can try different things once I can log in again (getting all sorts of strange TOS/log in registration check errors).

nawatts:
@MatthewBemis it looks like in your [video](https://github.com/DataBiosphere/terra-ui/pull/3229#issuecomment-1198242421), there aren't enough workspaces for the list to scroll.

If there are, "clicking" the star button with the space bar causes the list to scroll. I was able to make that stop by adding a keydown event handler to the Clickable in WorkspaceStarControl. It's mostly the same as [Interactive's default keydown handler](https://github.com/DataBiosphere/terra-ui/blob/3747142e9506a4c5b3df28b9c826d0bb52316993/src/components/Interactive.js#L49-L54), but adds `e.preventDefault()`.

```js
onKeyDown: e => {
  if (e.key === 'Enter' || e.key === ' ') {
    e.preventDefault()
    e.stopPropagation()
    e.target.click()
  }
}
```

I think the scrolling is also what's causing focus to sometimes move after "clicking" a star button.

MatthewBemis:
@nawatts ah I was using Enter to click the star, which doesn't have the same side effect even with a long list of workspaces.

Thank you for the tip about the event handler- I will add that in!

MatthewBemis:
Added the event handler in `c92fe17a6bb3ae557b6a5faf1333761b241c378e` to stop the scrolling!

nawatts (approved):

src/components/workspace-utils.js
```diff
@@ -340,6 +341,63 @@ export const WorkspaceTagSelect = props => {
...
+      [!isStarred && maxStarredWorkspacesReached, () => ['A maximum of ',
+        MAX_STARRED_WORKSPACES, ' workspaces can be starred. Please un-star another workspace before starring this workspace.']]
+    ),
+    'aria-label': isStarred ? 'This workspace is starred' : '',
+    className: 'fa-layers fa-fw',
+    disabled: maxStarredWorkspacesReached && !isStarred,
+    style: { verticalAlign: 'middle', ...style },
+    onKeyDown: e => {
+      if (e.key === 'Enter' || e.key === ' ') {
+        e.preventDefault()
```
cahrens, line 396: I do wonder if `e.preventDefault` should be pushed down to `Interactive`, but it's difficult for me to reason through if that would break desired behavior in any cases.

src/components/workspace-utils.js
```diff
@@ -340,6 +341,63 @@ export const WorkspaceTagSelect = props => {
...
+
+  //Thurloe has a limit of 2048 bytes for its VALUE column. That means we can store a max of 55
+  //workspaceIds in list format. We'll use 50 because it's a nice round number and should be plenty
+  //for the intended use case. If we find that 50 is not enough, consider introducing more powerful
+  //workspace organization functionality like folders
+  const MAX_STARRED_WORKSPACES = 50
+  const maxStarredWorkspacesReached = _.size(stars) >= MAX_STARRED_WORKSPACES
+
+  const refreshStarredWorkspacesList = async () => {
+    const { starredWorkspaces } = Utils.kvArrayToObject((await Ajax().User.profile.get()).keyValuePairs)
```
cahrens, line 364: It is odd that the get and set of preferences is not symmetrical (`register/profile` vs. `api/profile/preferences`).

src/components/workspace-utils.js
```diff
@@ -340,6 +341,63 @@ export const WorkspaceTagSelect = props => {
...
+    return _.isEmpty(starredWorkspaces) ? [] : _.split(',', starredWorkspaces)
+  }
+
+  const toggleStar = _.flow(
+    Utils.withBusyState(setUpdating),
+    withErrorReporting(`Unable to ${isStarred ? 'unstar' : 'star'} workspace`)
+  )(async star => {
+    const refreshedStarredWorkspaceList = await refreshStarredWorkspacesList()
```
cahrens, line 372: When you toggle a star, why do you first fetch the list of stars instead of using what is already in state?

MatthewBemis: This was added due to the discussion in https://github.com/DataBiosphere/terra-ui/pull/3229#discussion_r921239405

We're somewhat limited in what we can store in Thurloe and how we can store it- so that's why we store them as a single list. We also wanted these stars to be permanent so local storage wasn't suitable.

src/components/workspace-utils.js
MatthewBemis: Agreed- these APIs are pretty ancient and there tends to be a lot of asymmetry and non-RESTful quirks about them

src/components/workspace-utils.js
nawatts: My thoughts exactly.

src/pages/workspaces/List.js
```diff
@@ -86,6 +87,10 @@ export const WorkspaceList = () => {
   const { workspaces, refresh: refreshWorkspaces, loadingWorkspaces, loadingSubmissionStats } = useWorkspacesWithSubmissionStats()
   const [featuredList, setFeaturedList] = useState()
 
+  const { profile: { starredWorkspaces } } = useStore(authStore)
```
cahrens, line 91: It's unfortunate that the state management can't live entirely within the `WorkspaceStarControl` component.

cahrens (approved):
Thanks for iterating on the design, @MatthewBemis !

cahrens:
@MatthewBemis I doubt the test failures are at all related, but please run them again before you merge.

MatthewBemis:
@cahrens will do. And just a heads up that I'll be holding off on merging this PR for a bit longer. We have another workspaces-related feature that I'll be PRing soon to display recently viewed workspaces. We'd like them both to go out on the same day (I think there will be a blog post about it, not sure)

PR #3233, opened 2022-07-14
[DC-354] Add mocks to UI test to run in all envs

tcjiang9:


breilly2 (approved):
👍 Makes sense to me.

fboulnois (approved):
nice work.

PR #3234, opened 2022-07-14
[IA-3530] Add cloud env page loading metrics 

jdcanas:
This is a small PR to time how long API calls take. Usually this is done server-side, but this particular timing involves multiple calls, and the interesting part is the user-specific summation of the return of these calls.

I.e., it is possible for a particular user to have a lot of resources and this call could take a while.

### DIFF
--- a/src/pages/Environments.js
+++ b/src/pages/Environments.js
@@ -16,6 +16,7 @@ import { Ajax } from 'src/libs/ajax'
 import { getUser } from 'src/libs/auth'
 import colors from 'src/libs/colors'
 import { withErrorIgnoring, withErrorReporting } from 'src/libs/error'
+import Events from 'src/libs/events'
 import * as Nav from 'src/libs/nav'
 import { useCancellation, useGetter, useOnMount, usePollingEffect } from 'src/libs/react-utils'
 import {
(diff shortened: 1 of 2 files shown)


xuscrbw (approved):

src/pages/Environments.js
```diff
@@ -135,24 +136,33 @@ const Environments = () => {
 
   const refreshData = Utils.withBusyState(setLoading, async () => {
     const creator = getUser().email
+
+    const startTimeForLeoCallsEpochMs = new Date().getTime()
```
nawatts: FYI, [`Date.now`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now) is a more concise way to get a timestamp.

nawatts (approved):

src/pages/Environments.js
```diff
@@ -135,24 +136,33 @@ const Environments = () => {
...
+    const [decoratedRuntimes, decoratedDisks, decoratedApps] = await Promise.all([
+      Promise.all(_.map(decorateLabeledCloudObjWithWorkspaceId, newRuntimes)),
+      Promise.all(_.map(decorateLabeledCloudObjWithWorkspaceId, newDisks)),
+      Promise.all(_.map(decorateLabeledCloudObjWithWorkspaceId, newApps))
+    ])
+    const endTimeForRawlsCallsEpochMs = new Date().getTime()
+
+    const leoCallTimeTotalMs = endTimeForLeoCallsEpochMs - startTimeForLeoCallsEpochMs
+    const rawlsCallTimeTotalMs = endTimeForRawlsCallsEpochMs - startTimeForRawlsCallsEpochMs
+    Ajax().Metrics.captureEvent(Events.cloudEnvironmentDetailsLoad, { leoCallTimeMs: leoCallTimeTotalMs, rawlsCallTimeMs: rawlsCallTimeTotalMs, totalCallTimeMs: leoCallTimeTotalMs + rawlsCallTimeTotalMs })
```
kyuksel, line 165: I wonder if it may be worth having a helper for stopwatching, perhaps a wrapper like

```javascript
// Helper wrapper
const withStopwatch = fn => {
  const startTime =  Date.now()
  const res = fn()
  const endTime = Date.now()
  const timeElapsed = endTime - startTime

  return { res, elapsedTime }
}

// Main
const { res: [decoratedRuntimes, decoratedDisks, decoratedApps], leoCallTimeMs } = withStopwatch( () => 
  await Promise.all([
      Promise.all(_.map(decorateLabeledCloudObjWithWorkspaceId, newRuntimes)),
      Promise.all(_.map(decorateLabeledCloudObjWithWorkspaceId, newDisks)),
      Promise.all(_.map(decorateLabeledCloudObjWithWorkspaceId, newApps))
    ])
)

Ajax().Metrics.captureEvent(Events.cloudEnvironmentDetailsLoad, { leoCallTimeMs })
```

jdcanas: Not opposed to it if it becomes more widespread, but seems like an over-eager generalization that adds complication

kyuksel: My main motivation was due to start/stop's being sprinkled around looking harder to me to discern the business logic, rather than factoring out duplicate code but I don't feel strongly.

n8landolt: I like Kaan's idea, it might look a bit cleaner.

n8landolt (approved):

PR #3237, opened 2022-07-14
[IA-3567] Fix 'choppy' modal dismiss animation

n8landolt:
# Description

When closing the CloudEnvironmentModal or AnalysisModal after navigating within the modal, the view would reset to the default screen before finishing the closing animation.

# Solution

Fixed CloudEnvironmentModal and AnalysisModal so that we delay the transition back to the default screen for the same duration as the closing animation.

The implementation takes advantage of the onExited callback within the Transition component. The resetView method will be passed in to the onExited callback, so the higher order components don't have to worry about animation timing.

# Testing

Ensured that current implementation of onDismiss doesn't break with the new method argument

Ensured the transition is smooth when closing a modal.

# Video

The first is current live app - the second is with my fix.

https://user-images.githubusercontent.com/11773357/179776157-e364fbcb-fd8c-40d0-a242-bf24b1bd055c.mov

### DIFF
--- a/src/components/ModalDrawer.js
+++ b/src/components/ModalDrawer.js
@@ -6,6 +6,8 @@ import colors from 'src/libs/colors'
 import { useLabelAssert } from 'src/libs/react-utils'
 
 
+const TRANSITION_DURATION = 200
+
 const drawer = {
   overlay: transitionState => transitionState === 'entering' || transitionState === 'entered' ? {
     backgroundColor: 'rgba(0, 0, 0, 0.5)', padding: '2rem 1rem',
@@ -29,7 +31,7 @@ const ModalDrawer = ({ isOpen, onDismiss, width = 450, children, ...props }) => 
 
   return h(Transition, {
     in: isOpen,
-    timeout: { exit: 200 },
+    timeout: { exit: TRANSITION_DURATION },
     appear: true,
     mountOnEnter: true,
     unmountOnExit: true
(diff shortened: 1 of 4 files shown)


petesantos:
I am not sure what the original issue is. @n8landolt Could you provide reproduction steps and possibly some before/after videos?

n8landolt:
@petesantos That's my bad. The current code doesn't actually fix the issue - I thought it did. I have a different solution incoming, plus video.

Edit: Fixed

gpcarr (approved):
lgtm

kyuksel:
@n8landolt Before going on vacation, Pete had this [Slack comment](https://broadinstitute.slack.com/archives/C01EHNUM73R/p1658265855688779). Did you have any response, particularly to the following piece?

> My main concerns are around adding a timeout to deal with these types of issues…This seems like it may be better solved in the drawer, possibly using `requestAnimationFrame`.

src/components/ModalDrawer.js
```diff
@@ -29,7 +30,7 @@ const ModalDrawer = ({ isOpen, onDismiss, width = 450, children, ...props }) =>
 
   return h(Transition, {
     in: isOpen,
-    timeout: { exit: 200 },
+    timeout: { exit: DEFAULT_TRANSITION_DURATION },
```
nawatts: Using `DEFAULT_TRANSITION_DURATION` doesn't make it clear that the values in the different modals are related and need to be synchronized with `ModalDrawer`. Those could be read as unrelated things that all happen to use the same default value. Instead of a common `DEFAULT_TRANSITION_DURATION` constant, maybe export a constant from this module (`modalDrawerExitTimeout` or something like that) and use that in the consuming components?

nawatts: Or, maybe the need for synchronized timeouts could be removed by updating `ModalDrawer` to allow passing callback props to `Transition` and using [`onExited`](https://reactcommunity.org/react-transition-group/transition#Transition-prop-onExited) to reset views?

gpcarr (approved):
lgtm

nawatts (approved):

kyuksel:
It'd be good to update the PR title since the time-out solution is no longer relevant.

PR #3238, opened 2022-07-15
[SU-156] Add warning for shared billing project with one owner

nawatts:
If a Terra workspace or billing project with one owner and that person loses access to their account (leaves their organization, etc), the associated resources can no longer be fully managed through Terra. Due to security policy, Terra Support is generally unable to add owners to workspaces or billing projects. Thus, we want to encourage having multiple owners for workspaces and billing projects. This adds a warning to the billing project page if the billing project has multiple users but only one owner.

What the billing project owner sees:
Screen Shot 2022-07-15 at 10 25 23 AM

What other users of the billing project see:
Screen Shot 2022-07-15 at 10 27 14 AM

### DIFF
--- a/integration-tests/tests/billing-projects.js
+++ b/integration-tests/tests/billing-projects.js
@@ -1,6 +1,7 @@
 // This test is owned by the Workspaces Team.
 const _ = require('lodash/fp')
 const { assertTextNotFound, click, clickable, findText, gotoPage, noSpinnersAfter, select, signIntoTerra, waitForNoSpinners } = require('../utils/integration-utils')
+const { userEmail } = require('../utils/integration-config')
 const { registerTest } = require('../utils/jest-utils')
 const { withUserToken } = require('../utils/terra-sa-utils')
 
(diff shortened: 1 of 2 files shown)


MatthewBemis (approved):

cahrens:
@nawatts how would you feel about adding some coverage for whether or not this warning appears to `billing-projects.js`?

nawatts:
Tests added in https://github.com/DataBiosphere/terra-ui/pull/3238/commits/6ed30cc59be12dc0efb7636dee0ae08cee7e7b2b.

breilly2 (approved):
👍 Looks good. Great improvement!

Thinking out loud: At first, the test felt to me like a lot of code for a simple behavior. Then I understood that it's actually 5 different test cases. I'd love to see us work toward refactoring the code to inject data instead of mocking AJAX responses. I think that would make the tests easier to write, run more quickly, be less brittle, and be isolated when appropriate (though I'm admittedly on the fence about whether or not this should be 5 separate tests).

nawatts:
Agreed. Theoretically, tests like this would probably be better done in a unit test. However, many of our components (such as ProjectDetail) are so large and include so much functionality (including network requests) in one component that we'd probably still end up mocking AJAX responses in unit tests vs rendering with specific props.

PR #3239, opened 2022-07-15
[WOR-418] Hardcode a large limit when fetching workspace resources

aherbst-broad:
https://broadworkbench.atlassian.net/browse/WOR-418

By default WSM limits the number of workspaces fetched to 10, which can cause problems when we are rendering azure storage info and the needed container information isn't in the first 10 resources.

This PR adds a limit clause of 1000 as a short term fix to hopefully get all resources.

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1120,7 +1120,7 @@ const AzureStorage = signal => ({
   },
 
   details: async (workspaceId = {}) => {
-    const res = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources?stewardship=CONTROLLED`,
+    const res = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources?stewardship=CONTROLLED&limit=1000`,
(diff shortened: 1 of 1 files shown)


src/libs/ajax.js
```diff
@@ -1120,7 +1120,7 @@ const AzureStorage = signal => ({
   },
 
   details: async (workspaceId = {}) => {
-    const res = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources?stewardship=CONTROLLED`,
+    const res = await fetchWorkspaceManager(`workspaces/v1/${workspaceId}/resources?stewardship=CONTROLLED&limit=1000`,
```
cahrens, line 1123: Why 1000?

aherbst-broad: No good reason other than it seemed a sufficiently large number.

cahrens (approved):

nawatts (approved):

PR #3240, opened 2022-07-18
[SU-156] Restructure Collapse component

nawatts:
I want to add a warning icon with a tooltip to a collapsible section in the workspace dashboard. Something like this:

goal

However, with the current structure of the Collapse component, this would require adding it to the Collapse's title and would result in an accessibility issue (nested interactive elements) because the tooltip trigger would be inside the Collapse's button element.

Currently, the button element covers entire "summary" section of the Collapse.

structure

Currently the only option to add interactive elements is to place them after the button, as in the Data tab sidebar.

alternative

This restructures Collapse's summary section so that the button no longer covers the entire width. It uses an absolutely positioned child element to make it so that clicking anywhere in the summary section still toggles the Collapse. This is based on the "Accessible Card" in https://kittygiraudel.com/2022/04/02/accessible-cards/.

It also moves the `afterToggle` prop from after the icon / at the far right of the summary section to after the button.

new_structure

This does not add the icon/tooltip to the Owners section in the workspace dashboard.

Collapse is used several places in the app and with several different styles. Many of those had to be adapted for the new structure.

## Workflow dashboard
Before
workspace dashboard before
After
workspace dashboard after

## Data tab sidebar
Before
data sidebar before
After
data sidebar after

## Snapshots
Reversed the order of info button and arrow icon.

Before
snapshot before
After
snapshot after

## URL viewer
Before
uri viewer before
After
add row after

## Add row modal
Before
add row before
After
uri viewer after

## Workflow dashboard
Before
workflow dashboard before
After
workflow dashboard after

## Profile
Before
profile before
After
profile after

## Billing projects
Before
billing projects before
After
billing projects after

## Featured workspaces
Before
feature workspaces before
After
featured workspaces after

### DIFF
--- a/src/components/Collapse.js
+++ b/src/components/Collapse.js
@@ -56,7 +56,10 @@ const Collapse = ({ title, hover, tooltip, tooltipDelay, summaryStyle, detailsSt
         title
       ]),
       // zIndex: 2 lifts afterToggle controls above the absolutely positioned div in Link so that they can be clicked.
-      afterToggle && div({ style: { display: 'flex', flex: 1, margin: '0 1ch', zIndex: 2 } }, [afterToggle]),
+      // display: flex and flex: 1 causes this div to fill available space. This makes it easy to position afterToggle
+      // controls just after the summary text or at the right edge of the summary section. height: 0 prevents the unused
+      // space in this div from blocking clicks on the summary section.
+      afterToggle && div({ style: { display: 'flex', flex: 1, alignItems: 'center', height: 0, margin: '0 1ch', zIndex: 2 } }, [afterToggle]),
       titleFirst && angleIcon
     ]),
     isOpened && div({ id, style: detailsStyle }, [children])
--- a/src/components/Collapse.js
+++ b/src/components/Collapse.js
@@ -50,10 +50,12 @@ const Collapse = ({ title, hover, tooltip, tooltipDelay, summaryStyle, detailsSt
       }, [
         div({
           'aria-hidden': true,
+          // zIndex: 1 lifts this element above angleIcon, so that clicking on the icon will toggle the Collapse.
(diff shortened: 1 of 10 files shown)


MatthewBemis (approved):

src/components/Collapse.js
```diff
@@ -21,19 +28,34 @@ const Collapse = ({ title, buttonStyle, buttonProps = {}, summaryStyle, detailsS
...
+        hover,
+        tooltip,
+        tooltipDelay
       }, [
-        titleFirst && div({ style: { flexGrow: 1, ...(noTitleWrap ? Style.noWrapEllipsis : {}) } }, [title]),
-        angleIcon,
-        !titleFirst && title
+        div({
+          'aria-hidden': true,
+          style: { position: 'absolute', top: 0, left: 0, bottom: 0, right: 0, zIndex: 1 }
```
dmohs, line 54: `zIndex` usage almost always wants a comment. Why wasn't absolute positioning enough here?

src/components/Collapse.js
```diff
@@ -21,19 +28,34 @@ const Collapse = ({ title, buttonStyle, buttonProps = {}, summaryStyle, detailsS
...
+        }),
+        title
       ]),
-      afterToggle
+      afterToggle && div({ style: { display: 'flex', flex: 1, margin: '0 1ch', zIndex: 2 } }, [afterToggle]),
```
dmohs: Same.

dmohs (approved):

src/components/Collapse.js
nawatts: Added comments in https://github.com/DataBiosphere/terra-ui/pull/3240/commits/7ee084e2707f1222cb814c68200867588e327526. `zIndex: 1` lifts this element above sibling elements (such as `angleIcon`), so that clicking anywhere on the summary section will toggle the Collapse.

PR #3242, opened 2022-07-19
PROD-695 fix regex for Rmd and R files

Qi77Qi:
https://broadworkbench.atlassian.net/browse/PROD-695

Tested by running local UI against dev

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -91,7 +91,7 @@ export const toolExtensionDisplay = {
 }
 
 export const getPatternFromTool = toolLabel => Utils.switchCase(toolLabel,
-  [tools.RStudio.label, () => '.+(.R|.Rmd)$'],
+  [tools.RStudio.label, () => '.+(\\.R|\\.Rmd)$'],
   [tools.Jupyter.label, () => '.*\\.ipynb']
 )
 
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -91,7 +91,7 @@ export const toolExtensionDisplay = {
 }
 
 export const getPatternFromTool = toolLabel => Utils.switchCase(toolLabel,
-  [tools.RStudio.label, () => '.*(.R|.Rmd)$'],
+  [tools.RStudio.label, () => '.+(.R|.Rmd)$'],
   [tools.Jupyter.label, () => '.*\\.ipynb']
 )
 
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -91,7 +91,7 @@ export const toolExtensionDisplay = {
 }
 
 export const getPatternFromTool = toolLabel => Utils.switchCase(toolLabel,
-  [tools.RStudio.label, () => '.*\\.R(md)'],
+  [tools.RStudio.label, () => '.*(.R)$|(.Rmd)$'],
   [tools.Jupyter.label, () => '.*\\.ipynb']
 )
 
--- a/src/components/notebook-utils.js
+++ b/src/components/notebook-utils.js
@@ -91,7 +91,7 @@ export const toolExtensionDisplay = {
 }
 
 export const getPatternFromTool = toolLabel => Utils.switchCase(toolLabel,
-  [tools.RStudio.label, () => '.*(.R)$|(.Rmd)$'],
+  [tools.RStudio.label, () => '.*(.R|.Rmd)$'],
   [tools.Jupyter.label, () => '.*\\.ipynb']
 )
 


xuscrbw (approved):

cahrens (approved):

src/components/notebook-utils.js
```diff
@@ -91,7 +91,7 @@ export const toolExtensionDisplay = {
 }
 
 export const getPatternFromTool = toolLabel => Utils.switchCase(toolLabel,
-  [tools.RStudio.label, () => '.*\\.R(md)'],
+  [tools.RStudio.label, () => '.*(.R)$|(.Rmd)$'],
```
breilly2: Does whatever is consuming this care about the groupings? I'm also not 100% sure of how much of the expression the `|` considers. In fact, I'm wondering if what you want is `.*(.R|.Rmd)$`.

Qi77Qi: I think it's essentially equivalent to what you have there...I did test both `.Rmd` and `.R` files

kyuksel: 1. It seems to match `.Rmd` as well, which I assume is not something we'd want.
2. Would we want to restrict the length and the types of characters we want to allow in file names?
3. It looks easy enough to get it wrong, which to me warrants adding a unit test.

Qi77Qi: we do want to sync both Rmd and R files...it's a bug that currently we're not syncing either (not just this bug in UI)

Regarding 2, we don't do that for notebooks either...if this actually becomes an issue, then we can add that

Regarding 3, we do have tests in welder repo, not sure what test will look like just for the regex here

src/components/notebook-utils.js
```diff
@@ -91,7 +91,7 @@ export const toolExtensionDisplay = {
 }
 
 export const getPatternFromTool = toolLabel => Utils.switchCase(toolLabel,
-  [tools.RStudio.label, () => '.*\\.R(md)'],
+  [tools.RStudio.label, () => '.*(.R|.Rmd)$'],
```
kyuksel: Should the notebook name contain at least one character?

```suggestion
  [tools.RStudio.label, () => '.+(.R|.Rmd)$'],
```

aweng98: `+` is not better than `*`. Consider an case of one letter notebook name. Regex with `+` won't match `t.R`. I've created notebook name with 1 letter before in AoU. 
You may want to consider an edge case with file extension occurs more than once. Such as `hello.R.R` or `helo.Rmd.Rmd` or `hello.Rmd.R`.
Consider using this regex or come up with a better version: `.+[^(.R|.Rmd)](.R|.Rmd)$`

breilly2 (approved):
👍 `.+(.R|.Rmd)$` looks correct to me.

src/components/notebook-utils.js
```diff
@@ -91,7 +91,7 @@ export const toolExtensionDisplay = {
 }
 
 export const getPatternFromTool = toolLabel => Utils.switchCase(toolLabel,
-  [tools.RStudio.label, () => '.*\\.R(md)'],
+  [tools.RStudio.label, () => '.+(.R|.Rmd)$'],
```
kyuksel: `.`s inside the parentheses need escaping or else the expression would match things like `abR`.

```suggestion
  [tools.RStudio.label, () => '.+(\.R|\.Rmd)$'],
```

src/components/notebook-utils.js
Qi77Qi: @aweng98 it will match things like `t.R`...regarding the repeating `.R.R`, I'm leaning towards keeping things simple for now until we have a case to make further decisions. Users can create arbitrary file names in terminal, and we can't really prevent them from doing so. This regex here is just determining what files will be synced to workspace bucket.

src/components/notebook-utils.js
```diff
@@ -91,7 +91,7 @@ export const toolExtensionDisplay = {
 }
 
 export const getPatternFromTool = toolLabel => Utils.switchCase(toolLabel,
-  [tools.RStudio.label, () => '.*\\.R(md)'],
+  [tools.RStudio.label, () => '.+(\\.R|\\.Rmd)$'],
```
kyuksel, line 94: Does the above work for you? I don't think it would since what are escaped are the slashes (vs. periods).

```suggestion
  [tools.RStudio.label, () => '.+(\.R|\.Rmd)$'],
```

src/components/notebook-utils.js
breilly2: Oh, interesting. I wasn't sure if they still needed escaping when inside `()`s.

Qi77Qi:
are ppl ok with merging without the register user test in circle ci passing? https://github.com/DataBiosphere/welder/pull/86 is the PR for updating test around regex

src/components/notebook-utils.js
Qi77Qi: you need double slash...it's not happy with one slash in VS code...we have double slash for .ipynb as well

kyuksel:
> are ppl ok with merging without the register user test in circle ci passing? [DataBiosphere/welder#86](https://github.com/DataBiosphere/welder/pull/86) is the PR for updating test around regex

I think it's fine since the change is very contained and the test seems unrelated.

kyuksel (approved):

src/components/notebook-utils.js
aweng98: Totally. Thanks.

PR #3243, opened 2022-07-19
WOR-264 De-emphasize Other Costs

msilva-broad:
- Billing report now shows "other costs" in a footnote, and removes the card, leaving only 3 cost cards in summary area.
- updated end2end integration test.
- visual testing done locally.
---

aria-live region fixes
- screen reader announcement on report render is now consistent with content being announced in order.
- changed over to a single larger aria-live region for the entire report content, which is marked with aria-busy when loading to cut down on screen-reader noise.
- promoted a couple of sub-controls to be full React.FC components instead of just rendering sub-routines.  This looks to enable React rendering optimizations better, and help a great deal with removing some screen reader repetition of the entire live region content for a single report refresh.
- KNOWN ISSUE: still seeing initial announcement of the cost card headers place-holder content on initial load and hard tab switches (like switching projects.  I suspect this may be related to aria-controls prop not being implemented in our Tabs control/usage.  I will file a seperate ticket for that fix.
---
image

### DIFF
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -304,7 +304,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
       div(
         {
           style: { flex: 'none', padding: '0.625rem 1.25rem' },
-          'aria-live': projectCost !== null ? 'polite' : 'off',
+          'aria-live': isProjectCostReady ? 'polite' : 'off',
           'aria-atomic': true
         },
         [
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -375,13 +375,13 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
         )
       ])
     ]),
-    [spendReportKey]: div({ style: { display: 'grid', rowGap: '1.66rem' } }, [
+    [spendReportKey]: div({ style: { display: 'grid', rowGap: '0.5rem' } }, [
       div(
         {
           style: {
             display: 'grid',
             gridTemplateColumns: 'repeat(3, minmax(max-content, 1fr))',
-            rowGap: '1.25rem',
+            rowGap: '1.66rem',
             columnGap: '1.25rem'
           }
         }, [
@@ -415,7 +415,7 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
       h(OtherMessaging, { cost: isProjectCostReady ? projectCost['other'] : null }),
       costPerWorkspace.numWorkspaces > 0 && div(
         {
-          style: { minWidth: 500 }
+          style: { minWidth: 500, marginTop: '1rem' }
         }, [ // Set minWidth so chart will shrink on resize
           h(Suspense, { fallback: null }, [h(LazyChart, { options: spendChartOptions })])
         ]
--- a/src/pages/billing/Project.js
+++ b/src/pages/billing/Project.js
@@ -195,6 +195,7 @@ const groupByBillingAccountStatus = (billingProject, workspaces) => {
 const LazyChart = lazy(() => import('src/components/Chart'))
 const maxWorkspacesInChart = 10
 const spendReportKey = 'spend report'
+const otherCostsDomId = 'billing-other-costs'
 const otherMessaging = cost => `Total spend includes ${cost} in other infrastructure or query costs related to the general operations of Terra.`
 
 const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProject, isAlphaSpendReportUser, isOwner, reloadBillingProject }) => {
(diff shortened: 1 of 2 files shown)


cahrens:
When I try this out with VoiceOver, the aria-live region for the "other cost" explanation is getting read before the cards for total, compute, and storage. It would be great if that could be reversed… not sure what is causing it.

cahrens:
Another possible solution would be to have the explanation text linked to the total cost using `aria-describedby` (and not marked as `aria-live`). That might work better, and it more logically represents the connection.

msilva-broad:
> 



> Another possible solution would be to have the explanation text linked to the total cost using `aria-describedby` (and not marked as `aria-live`). That might work better, and it more logically represents the connection.

implemented. try out out plz.

src/pages/billing/Project.js
```diff
@@ -195,6 +195,7 @@ const groupByBillingAccountStatus = (billingProject, workspaces) => {
 const LazyChart = lazy(() => import('src/components/Chart'))
 const maxWorkspacesInChart = 10
 const spendReportKey = 'spend report'
+const otherCostsDomId = 'billing-other-costs'
```
cahrens: This approach will work, and we should never have multiple of these sections on the page. However, the general pattern in the repo is to use `IdContainer` to generate the ID and obtain the reference to it. For example, on line 602.

aherbst-broad (approved):

src/pages/billing/Project.js
```diff
@@ -195,6 +195,8 @@ const groupByBillingAccountStatus = (billingProject, workspaces) => {
...
+const otherMessaging = cost => `Total spend includes ${cost} in other infrastructure or query costs related to the general operations of Terra.`
```
cahrens, line 243: The message looks pretty odd when the cost is "…" during the loading time. I get that we probably want to have some message to avoid text elements shifting once the cost has been loaded, but can we possibly have something else that reads better? Perhaps just "Total spend includes infrastructure or query costs related to the general operations of Terra"?

msilva-broad: fixed

src/pages/billing/Project.js
msilva-broad: removed from newer implementation

src/pages/billing/Project.js
```diff
@@ -393,41 +401,21 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
-                showAsterisk: name === 'spend',
-                key: name
-              }),
-              ['spend', 'compute', 'storage'])
-              )
-            ]
-          ),
-          h(OtherMessaging, { cost: isProjectCostReady ? projectCost['other'] : null })
+          ]),
+          ...(_.map(name => CostCard({
```
cahrens: @msilva-broad I thought there were good reasons to keep CostCard as a component. Does it cause issues as a component?

msilva-broad: I thought so too, but as I reverted to the original setup, I was still seeing some wierdness and was tired of fussing with it all... reverting fixed things.  so, meh.

src/pages/billing/Project.js
```diff
@@ -278,21 +288,39 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
       }
     }, [
-      div({ style: { flex: 'none', padding: '0.625rem 1.25rem' }, 'aria-live': projectCost !== null ? 'polite' : 'off', 'aria-atomic': true }, [
-        h3({ style: { fontSize: 16, color: colors.accent(), margin: '0.25rem 0.0rem', fontWeight: 'normal' } }, title),
-        div({ style: { fontSize: 32, height: 40, fontWeight: 'bold', gridRowStart: '2' } }, [amount])
-      ])
+      div(
+        {
+          style: { flex: 'none', padding: '0.625rem 1.25rem' },
+          'aria-live': projectCost !== null ? 'polite' : 'off',
```
cahrens: Nit: for consistency you could replace `projectCost !== null` with `isProjectCostReady`.

src/pages/billing/Project.js
```diff
@@ -278,21 +288,39 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
...
+          'aria-atomic': true
+        },
+        [
+          div({ style: { fontSize: 16, color: colors.accent(), margin: '0.25rem 0.0rem', fontWeight: 'normal' } }, [title]),
+          div({ style: { fontSize: 32, height: 40, fontWeight: 'bold', gridRowStart: '2' } }, [
+            amount,
+            (!!showAsterisk && isProjectCostReady) ? span(
```
cahrens: Nit: `showAsterisk` is a boolean, so I think having the `!!` in front just makes it harder to read.

src/pages/billing/Project.js
```diff
@@ -361,19 +396,32 @@ const ProjectDetail = ({ authorizeAndLoadAccounts, billingAccounts, billingProje
                   value: days
                 }), [7, 30, 90]),
                 onChange: ({ value: selectedDays }) => {
-                  setSpendReportLengthInDays(selectedDays)
-                  setProjectCost(null)
+                  if (selectedDays !== spendReportLengthInDays) {
```
cahrens, line 399: @msilva-broad This is a pre-existing bug that was bothering me so I pushed up a commit to fix it. Selecting an already-selected date range caused the cost cards to all clear and never reload.

src/pages/billing/Project.js
cahrens: I pushed this up.

cahrens (approved):

PR #3244, opened 2022-07-19
[SU-164] Improve IGV file selection

nawatts:
When opening data table rows in IGV, Terra looks through all attribute values in the selected rows to find files viewable in IGV. It then allows the user to select which of those files to view. Currently, the way it does identifies files is based on file extension (it looks for a `.bam`, `.cram`, `.vcf` or `.bed` file extension). For some file extensions, it also looks for a related index file (a matching `.bai` for a `.bam` file, etc).

https://github.com/DataBiosphere/terra-ui/blob/fee26014a657de6a9bef68856584fd90711d292c/src/components/IGVFileSelector.js#L30-L47

This method also selects values that are not GCS URLs. For example, if an attribute value contains a file name or path instead of a URL (`test.bed` instead of `gs://workspace-bucket/test.bed`), Terra will include that file name in the list of viewable files.

The IGVBrowser component that those selected files are passed to expects GCS URLs and it parses selected files as such.

https://github.com/DataBiosphere/terra-ui/blob/752e98114d4b5c9c9143fed0f69672ab1b6b5c45/src/components/IGVBrowser.js#L28

Thus, selecting a value that is not a URL to view in IGV causes IGV to fail to load and Terra to display an infinite spinner.

With this change, IGVFileSelector will only show files from URL values that IGVBrowser can handle.

## Before
https://user-images.githubusercontent.com/1156625/179816650-1d1c1b12-1fd6-4e13-978b-cad6f5568e45.mov

## After
https://user-images.githubusercontent.com/1156625/179816681-827b0cb3-9d2a-4817-9fd5-e7fb1222f702.mov

---

This also fixes another bug found while testing this change. Currently, Terra cannot handle list values when opening data in IGV. Attempting to open a row containing a list in IGV crashes Terra.

## Before
https://user-images.githubusercontent.com/1156625/179816978-59eb07f9-d920-488f-976d-12cbbefac2c9.mov

## After
https://user-images.githubusercontent.com/1156625/179817007-de17bd3e-6755-4d0d-94a1-738957574b87.mov

### DIFF
--- a/src/components/IGVFileSelector.js
+++ b/src/components/IGVFileSelector.js
@@ -4,6 +4,7 @@ import { div, h } from 'react-hyperscript-helpers'
 import { AutoSizer, List } from 'react-virtualized'
 import ButtonBar from 'src/components/ButtonBar'
 import { ButtonPrimary, LabeledCheckbox, Link } from 'src/components/common'
+import { parseGsUri } from 'src/components/data/data-utils'
 import IGVReferenceSelector, { addIgvRecentlyUsedReference, defaultIgvReference } from 'src/components/IGVReferenceSelector'
 import * as Style from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
@@ -21,7 +22,7 @@ export const getValidIgvFiles = values => {
   return _.flatMap(value => {
     const possibleFile = /(.+)\.([^.]+)$/.exec(value)
 
-    if (!possibleFile) {
+    if (_.isEmpty(parseGsUri(value)) || !possibleFile) {
       return []
     }
 
--- a/src/components/IGVFileSelector.js
+++ b/src/components/IGVFileSelector.js
@@ -12,9 +12,36 @@ import * as Utils from 'src/libs/utils'
 const getStrings = v => {
   return Utils.cond(
     [_.isString(v), () => [v]],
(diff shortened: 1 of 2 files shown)


MatthewBemis (approved):

breilly2:
Does IGV only support `gs://` or does it also support `http[s]://`?

nawatts:
IGV.js supports HTTP(S) URLs. IGV in Terra currently only supports GCS URLs because of its code for handling requester-pays.
https://github.com/DataBiosphere/terra-ui/blob/752e98114d4b5c9c9143fed0f69672ab1b6b5c45/src/components/IGVBrowser.js#L26-L28

This change brings the IGV file selector in line with what's currently supported by the IGVBrowser component. I'm certainly open to adding support for HTTP(S) URLs, but I think that could be a separate change.

Also, Terra itself doesn't have great support for HTTP(S) URLs. While GCS and DRS URLs are rendered as links in data tables, HTTP(S) URLs are shown as plain text.

Screen Shot 2022-07-21 at 12 54 35 PM

breilly2 (approved):
👍 Looks good to me. I like that you extracted functions for unit testing.

nawatts:
Filed [SU-173](https://broadworkbench.atlassian.net/browse/SU-173) for supporting viewing HTTP(S) URLs in IGV.

PR #3245, opened 2022-07-19
[SU-169] Limit number of list items rendered in data table cells

nawatts:
Data table cells containing lists render elements for each list item.

https://github.com/DataBiosphere/terra-ui/blob/ca57072ff8692c147c2d9511a5ffd0d751a4508f/src/components/data/data-utils.js#L88-L92

When a data table has many long lists, this makes the data table slow to render.

This became more of an issue after #3187 (specifically https://github.com/DataBiosphere/terra-ui/pull/3187/files#diff-967b453c8b03379166f3e719046b3a65574bddf955e95e14d115bea19185ae3bL407-R407). Before that, we converted lists to strings before rendering data table cells.

This change limits data table cells to render at most 100 items from lists. If a list has more items, an "and [N] more" message is added. Users can still view all items in the list by clicking the "[N] entities/items" button. 100 seemed low enough to improve performance and high enough that there would be no visible change to many tables.

## Before
20 sets with 10,000 items each. Note the loading spinner stops and the browser freezes.

https://user-images.githubusercontent.com/1156625/179835227-8568f011-c99f-4cd9-9dfb-b8642be29fd6.mov

## After
https://user-images.githubusercontent.com/1156625/179835291-723218b8-9483-43af-9992-f7ff59ed411f.mov

### DIFF
--- a/src/components/data/_tests/data-utils.test.js
+++ b/src/components/data/_tests/data-utils.test.js
@@ -3,6 +3,7 @@ import '@testing-library/jest-dom'
 import { render } from '@testing-library/react'
 import _ from 'lodash/fp'
 import { concatenateAttributeNames, convertAttributeValue, entityAttributeText, getAttributeType, prepareAttributeForUpload, renderDataCell } from 'src/components/data/data-utils'
+import * as Utils from 'src/libs/utils'
 
 
 describe('concatenateAttributeNames', () => {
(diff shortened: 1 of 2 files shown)


davidangb (approved):
nice, thanks for tackling this! :+1: for the changes in here. For the cells that get truncated and show " and N more," could/should you also update the tooltip? I believe currently the tooltip shows all values.

MatthewBemis (approved):

nawatts:
> For the cells that get truncated and show " and N more," could/should you also update the tooltip?

Done in https://github.com/DataBiosphere/terra-ui/pull/3245/commits/47ff56a5d0a71af95b54ae21a78f81e6c69e4ea7.

davidangb (approved):
thank you!

dmohs (approved):
Nice.

PR #3246, opened 2022-07-20
[SUP-698] Create placeholder objects for folders in GCS

nawatts:
Multiple users have said that it's surprising that folders created in the Files tab do not persist unless a file is uploaded to the folder. This happens because GCS doesn't actually have folders and the folder hierarchy is constructed from object names.

With this change, Terra will create a placeholder object `gs://my-bucket/path/to/folder/` so that the folder appears even if no files are uploaded into it. This mimics the Cloud Console's behavior. See https://cloud.google.com/storage/docs/folders for more information.

This required a few other changes... the placeholder object would show up as a blank row in the files table. This change filters it out. This also improves Terra's handling of folders created through the Cloud Console. This also adds a button to delete empty folders, which will delete the placeholder object.

## Before
https://user-images.githubusercontent.com/1156625/180019272-13fcfc81-b8a2-46e6-a3cb-2b291eaa4cfc.mov

## After
https://user-images.githubusercontent.com/1156625/180019211-508344ee-9c9c-467e-945f-aa174f033273.mov

### DIFF
--- a/src/components/data/FileBrowser.js
+++ b/src/components/data/FileBrowser.js
@@ -409,7 +409,8 @@ const BucketBrowser = (({
 
                 // Since prefixes have a trailing slash, the last item in the split array will be an empty string.
                 // Dropping the last two items returns the parent "folder".
-                setPrefix(_.flow(_.split('/'), _.dropRight(2), _.join('/'))(prefix))
+                const parentPrefix = _.flow(_.split('/'), _.dropRight(2), _.join('/'))(prefix)
+                setPrefix(parentPrefix === '' ? '' : `${parentPrefix}/`)
               })
             }, ['Delete this folder'])
           ])],
--- a/src/components/data/FileBrowser.js
+++ b/src/components/data/FileBrowser.js
@@ -51,7 +51,9 @@ const useBucketContents = ({ googleProject, bucketName, prefix, pageSize = 1000 
         nextPageToken: pageNextPageToken
       } = await Ajax(signal).Buckets.list(googleProject, bucketName, prefix, requestOptions)
 
-      setAllObjects(prevObjects => _.concat(prevObjects, _.defaultTo([], pageObjects)))
+      // _.remove({ name: prefix }) filters out folder placeholder objects.
+      // See https://cloud.google.com/storage/docs/folders for more information.
+      setAllObjects(prevObjects => _.flow(_.defaultTo([]), _.remove({ name: prefix }), _.concat(prevObjects))(pageObjects))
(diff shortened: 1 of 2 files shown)


src/components/data/FileBrowser.js
```diff
@@ -424,6 +454,11 @@ const BucketBrowser = (({
         onSuccess: ({ name }) => {
           setCreatingNewFolder(false)
           setPrefix(`${prefix}${name}/`)
+
+          // Create a placeholder object for the new folder.
+          // See https://cloud.google.com/storage/docs/folders for more information.
+          const placeholderObject = new File([''], `${name}/`, { type: 'text/plain' })
+          Ajax().Buckets.upload(googleProject, bucketName, prefix, placeholderObject)
```
cahrens: What if this Ajax call were to throw an error?

src/components/data/FileBrowser.js
```diff
@@ -387,6 +389,34 @@ const BucketBrowser = (({
...
+                  }
+                }
+
+                setBusy(false)
+
+                // Since prefixes have a trailing slash, the last item in the split array will be an empty string.
+                // Dropping the last two items returns the parent "folder".
+                setPrefix(_.flow(_.split('/'), _.dropRight(2), _.join('/'))(prefix))
+              }
+            }, ['Delete this folder'])
```
cahrens, line 415: What does the Delete button up near New Folder do?

nawatts: That button deletes objects selected in the table. We don't currently allow selecting folders in the table because we don't support recursively deleting all objects in a folder.

src/components/data/FileBrowser.js
nawatts: Added error reporting in https://github.com/DataBiosphere/terra-ui/pull/3246/commits/392c79c057aa7e03fbcbdca80723c8da2feff872.

cahrens: Why does withErrorReporting not work? Is it because you need to set the busy flag without using the spinner? What does the busy flag control?

nawatts: I guess it does. I kind of got in the habit of not using the wrappers after running into several "state change on unmounted component" warnings when using `Utils.withBusyState` in modals (where using `withBusyState` would cause the busy state to be set after the modal was closed). Switched this to use wrappers in https://github.com/DataBiosphere/terra-ui/pull/3246/commits/c433d9d31ab25e2571dcfa2494a2af5100e548fb.

The busy flag controls the spinner.

cahrens (approved):

MatthewBemis (approved):
lgtm

PR #3247, opened 2022-07-20
[ID-119] Port Terra UI to use DrsHub instead of Martha in Dev

Shakespeared:
This is part of the process to getting DrsHub to Prod. This PR is only to transition Dev Terra to use DrsHub instead of Martha, later PRs will transition the rest of the environments as well. To make that easy, I have added a DrsHub config to all envs which currently points to Martha everywhere except dev, and will be updated in each config as we get DrsHub running in other envs. 

There is one usage of Martha in `UriViewer` to get signed URLs that is not being ported yet - that will be a later enhancement for DrsHub. 

Tested locally with this DRS URI: `drs://dg.4DFC:011a6a54-1bfe-4df9-ae24-990b12a812d3`

### DIFF
--- a/config/prod.json
+++ b/config/prod.json
@@ -6,6 +6,7 @@
   "catalogUrlRoot": "https://catalog.dsde-prod.broadinstitute.org",
   "dataRepoUrlRoot": "https://data.terra.bio",
   "dockstoreUrlRoot": "https://dockstore.org",
+  "drsHubUrlRoot": "https://us-central1-broad-dsde-prod.cloudfunctions.net",
   "firecloudBucketRoot": "https://storage.googleapis.com/firecloud-alerts",
   "firecloudUrlRoot": "https://portal.firecloud.org",
   "isProd": true,
--- a/config/prod.json
+++ b/config/prod.json
@@ -18,6 +18,7 @@
   "rexUrlRoot": "https://terra-rex-prod.appspot.com",
   "samUrlRoot": "https://sam.dsde-prod.broadinstitute.org",
   "shibbolethUrlRoot": "https://broad-shibboleth-prod.appspot.com",
+  "shouldUseDrsHub": false,
   "workspaceManagerUrlRoot": "https://workspace.dsde-prod.broadinstitute.org",
   "tCell": {
     "appId": "saturnprod-VuHKy",
--- a/config/prod.json
+++ b/config/prod.json
@@ -6,7 +6,7 @@
   "catalogUrlRoot": "https://catalog.dsde-prod.broadinstitute.org",
   "dataRepoUrlRoot": "https://data.terra.bio",
   "dockstoreUrlRoot": "https://dockstore.org",
-  "drsHubUrlRoot": "https://drshub.dsde-dev.broadinstitute.org",
+  "drsHubUrlRoot": "https://drshub.dsde-prod.broadinstitute.org",
   "firecloudBucketRoot": "https://storage.googleapis.com/firecloud-alerts",
   "firecloudUrlRoot": "https://portal.firecloud.org",
   "isProd": true,
--- a/config/prod.json
+++ b/config/prod.json
@@ -6,7 +6,7 @@
   "catalogUrlRoot": "https://catalog.dsde-prod.broadinstitute.org",
   "dataRepoUrlRoot": "https://data.terra.bio",
   "dockstoreUrlRoot": "https://dockstore.org",
-  "drsHubUrlRoot": "https://us-central1-broad-dsde-prod.cloudfunctions.net",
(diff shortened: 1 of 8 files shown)


src/libs/ajax.js
```diff
@@ -1754,12 +1755,23 @@ const Martha = signal => ({
...
+      'getSignedUrlV1',
       _.mergeAll([jsonBody({ bucket, object, dataObjectUri }), authOpts(), appIdentifier, { signal, method: 'POST' }]))
     return res.json()
   }
 })
 
+const DrsHub = signal => ({
+  getDataObjectMetadata: async (url, fields) => {
+    const res = await fetchDrsHub(
+      '/api/v4/drs/resolve',
```
breilly2: Doesn't this need to still be `'martha_v3'` for non-dev environments?

nawatts (commented):
It looks like this changes the endpoint used for `getDataObjectMetadata` in production from `https://us-central1-broad-dsde-prod.cloudfunctions.net/martha_v3` to `https://us-central1-broad-dsde-prod.cloudfunctions.net/api/v4/drs/resolve`. Is that intentional?

src/libs/ajax.js
Shakespeared: Ah, I was so excited to get this tested that I forgot to add a flag for dev

Shakespeared:
> It looks like this changes the endpoint used for `getDataObjectMetadata` in production from `https://us-central1-broad-dsde-prod.cloudfunctions.net/martha_v3` to `https://us-central1-broad-dsde-prod.cloudfunctions.net/api/v4/drs/resolve`. Is that intentional?

Updated with feature flag that I meant to implement the first time!

src/libs/ajax.js
```diff
@@ -1744,35 +1745,34 @@ const Dockstore = signal => ({
   listTools: (params = {}) => fetchDockstore(`api/ga4gh/v1/tools?${qs.stringify(params)}`).then(r => r.json())
 })
 
+const shouldUseDrsHub = !!fetchShouldUseDrsHub()
```
nawatts:
```suggestion
const shouldUseDrsHub = !!getConfig().shouldUseDrsHub
```


As is, calling `fetchShouldUseDrsHub` will make a request to `<Terra URL>/true/undefined` or `<Terra URL>/false/undefined` depending on the value of `getConfig().shouldUseDrsHub`. This request will succeed because Terra responds with the app HTML for any unknown URL. Since `fetchShouldUseDrsHub()` returns a Promise, this will always be truthy.

nawatts (changes requested):

src/libs/ajax.js
Shakespeared: Thank you!

src/libs/ajax.js
Shakespeared: updated!

config/prod.json
```diff
@@ -6,6 +6,7 @@
   "catalogUrlRoot": "https://catalog.dsde-prod.broadinstitute.org",
   "dataRepoUrlRoot": "https://data.terra.bio",
   "dockstoreUrlRoot": "https://dockstore.org",
+  "drsHubUrlRoot": "https://us-central1-broad-dsde-prod.cloudfunctions.net",
```
nawatts: After the change in approach in https://github.com/DataBiosphere/terra-ui/pull/3247/commits/9778ef7f4a207cdaff083048d50db845c3d16bb3, should non-dev configurations have DRS Hub URLs for `drsHubUrlRoot` instead of Martha URLs?

Shakespeared: Yep great point. Thanks for the thorough review on this - sorry about the back and forth, I'm a bit rusty on the UI.

nawatts: No cause to apologize! Back and forth is just the review process working.

tlangs (approved):

src/components/UriViewer.js
```diff
@@ -178,12 +179,13 @@ const UriViewer = _.flow(
         const metadata = await loadObject(googleProject, bucket, name)
         setMetadata(metadata)
       } else {
+        // TODO: change below comment after switch to DRSHub is complete
```
andy7i: it might be a good idea for us to add the ticket number here and make a note in the ticket to update this. I tend to forget these things but also, having it in the ticket means that whoever picks up the ticket will know to update this.

andy7i (approved):
Looks good! Added a suggestion. Also, don't forget the add the ticket number (ID-119) to the title!

src/components/UriViewer.js
Shakespeared: Ah great idea to make sure this doesn't get forgotten. I also totally forgot to add a ticket number the PR name, I will update that as well!

nawatts (approved):

breilly2 (approved):

PR #3249, opened 2022-07-20
[IA-3584] Bump default/min dataproc master disk size

gpcarr:
https://broadworkbench.atlassian.net/browse/IA-3584

And fixes a bug where changing the application configuration option from the dropdown menu to hail does not change the masterDiskSize to 150GBs


Qi77Qi (approved):
lgtm

xuscrbw:
So seeing something interesting. On the PR deployment environment, the default seems to stay at 120, but if we use the up and down buttons increase the disk size, it flips to 150. Going to test this locally

Edit: Can see this locally also. It will also try to create a 120 GB master disk but will fail until you tick it up to 150 GB

gpcarr:
> So seeing something interesting. On the PR deployment environment, the default seems to stay at 120, but if we use the up and down buttons increase the disk size, it flips to 150. Going to test this locally
> 
> Edit: Can see this locally also. It will also try to create a 120 GB master disk but will fail until you tick it up to 150 GB

I was able to recreate this locally, good catch! I'm close to a fix and will update once I'm done

src/components/ComputeModal.js
```diff
@@ -1527,6 +1527,9 @@ export const ComputeModalBase = ({
         updateComputeConfig('componentGatewayEnabled', isDataproc(newRuntimeType))
         const machineType = getDefaultMachineType(isDataproc(newRuntimeType), desiredTool)
         updateComputeConfig('masterMachineType', machineType)
+        if (isDataproc(newRuntimeType) && computeConfig.masterDiskSize < defaultDataprocMasterDiskSize) {
+          updateComputeConfig('masterDiskSize', defaultDataprocMasterDiskSize)
```
n8landolt, line 1531: In my opinion, this type of logic should really be in a backend migration. Eventually this 'if' statement will never be true.

gpcarr: when will it not be true? what would the migration do?

gpcarr:
> So seeing something interesting. On the PR deployment environment, the default seems to stay at 120, but if we use the up and down buttons increase the disk size, it flips to 150. Going to test this locally
> 
> Edit: Can see this locally also. It will also try to create a 120 GB master disk but will fail until you tick it up to 150 GB

@xuscrbw I fixed this and pushed the changes if you want to take a look!

src/components/ComputeModal.js
sjacobs146: I can see @n8landolt 's point - once everything is migrated (i.e. all disk sizes are increased ) this is essentially dead code. But we will likely increase the default disk size again some time and this code would handle that case. I'm also not a fan of bulk database updates.

cahrens (approved):

PR #3250, opened 2022-07-21
[SU-156] Add notice for shared workspaces with one owner

nawatts:
If a Terra workspace or billing project with one owner and that person loses access to their account (leaves their organization, etc), the associated resources can no longer be fully managed through Terra. Due to security policy, Terra Support is generally unable to add owners to workspaces or billing projects. Thus, we want to encourage having multiple owners for workspaces and billing projects. This adds a notice to the workspace dashboard has multiple users but only one owner.

Related to #3238

Screen Shot 2022-07-21 at 9 58 48 AM
Screen Shot 2022-07-21 at 9 59 15 AM

### DIFF
--- a/src/pages/workspaces/workspace/Dashboard.js
+++ b/src/pages/workspaces/workspace/Dashboard.js
@@ -350,7 +350,7 @@ const WorkspaceDashboard = _.flow(
 
   const oneOwnerNotice = Utils.cond(
     // No warning if there are multiple owners.
-    [_.size(owners) > 1, () => null],
+    [_.size(owners) !== 1, () => null],
     // If the current user does not own the workspace, then then workspace must be shared.
     [!Utils.isOwner(accessLevel), () => h(Fragment, [
       'This shared workspace has only one owner. Consider requesting ',
--- a/src/pages/workspaces/workspace/Dashboard.js
+++ b/src/pages/workspaces/workspace/Dashboard.js
@@ -111,7 +111,7 @@ const DashboardAuthContainer = props => {
   )
 }
 
-const RightBoxSection = ({ title, info, initialOpenState, onClick, children }) => {
+const RightBoxSection = ({ title, info, initialOpenState, afterToggle, onClick, children }) => {
   return div({ style: { paddingTop: '1rem' } }, [
     div({ style: Style.dashboard.rightBoxContainer }, [
       h(Collapse, {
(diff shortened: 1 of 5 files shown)


MatthewBemis (approved):

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -111,14 +111,15 @@ const DashboardAuthContainer = props => {
   )
 }
 
-const RightBoxSection = ({ title, info, initialOpenState, onClick, children }) => {
+const RightBoxSection = ({ title, info, initialOpenState, afterToggle, onClick, children }) => {
```
breilly2: Thinking out loud: I wish `afterToggle` had a different name... something like `afterTitle`. I was confused at first because I read "toggle" as a verb instead of a noun. Probably just an artifact of how the `Collapse` component evolved (I know you're just using the same name that `Collapse` uses).

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -335,6 +348,19 @@ const WorkspaceDashboard = _.flow(
     refresh()
   })
 
+  const oneOwnerNotice = Utils.cond(
+    // No warning if there are multiple owners.
+    [_.size(owners) > 1, () => null],
```
breilly2: Hopefully we'll never encounter a workspace with 0 owners. 😅

breilly2 (approved):

src/pages/workspaces/workspace/Dashboard.js
nawatts: Changed this condition in https://github.com/DataBiosphere/terra-ui/pull/3250/commits/9bf1218fd7d8d10ef616990bf4ce430b9c75000a just in case.

src/pages/workspaces/workspace/Dashboard.js
nawatts: I was also regretting not changing the prop name in #3240 where I changed the placement of this element. I'll change `afterToggle` to `afterTitle`.

nawatts: Changed in https://github.com/DataBiosphere/terra-ui/pull/3250/commits/c67e2ccae491e8dac1033296f48c33af78753f01.

PR #3257, opened 2022-07-22
[PE-8] Refactor Terra branding code

kyuksel:
### Ticket
https://broadworkbench.atlassian.net/browse/PE-8

### Goals
- Take a step towards making it easier to add/edit Terra co-branding for non-software engineers and/or software engineers not familiar with Terra UI by splitting out some of the branding config from the rest of the code.
- Make the branding code easier to follow by eliminating several conditionals.

### Details
- The vast majority of the changes are for refactoring that should be invisible to users unless otherwise noted.
- Most of the refactoring is around `getAppName()` to "denormalize" a lot of the brand config (e.g. brand name, description, etc.) that was interleaved with the business logic. This causes some duplication in brands' config. However, we're optimizing for the first goal above. 
- [Worked with Product and Marketing](https://broadinstitute.slack.com/archives/C34GAKQ5R/p1657899906579349) to further simplify the code by removing `bold`ing on the landing page text and new user text, as shown below, respectively.

image

image (1)

### Testing
Compared local UI pages to those in production.

### Documentation
Will update [the confluence page](https://broadworkbench.atlassian.net/wiki/spaces/WOR/pages/2369388553/Cobranding+and+White+Label+Sites) once the PR is merged.

### Future Work
- Eliminate the need to make code changes (e.g. `isMyNewBrand = () => ...` in order to add new brands.
- Make it easier for non-software engineers to enable a brand on the browser console.


src/libs/logos.js
```diff
@@ -56,15 +39,11 @@ const pickBrandLogo = (color = false) => Utils.cond(
...
-  alt: getAppName(), role: 'img',
+  alt: getEnabledBrand().name, role: 'img',
   src: pickBrandLogo(color),
   style: { height: size, marginRight: '1.5rem' }
 })
 
-export const signInLogo = () => isTerra() ?
-  terraLogoMaker(terraLogo, { height: 150 }) :
-  brandLogoMaker(70, true)
-
```
kyuksel, line 67: Unrelated to the ticket. Removed unused function opportunistically.

src/libs/brand-utils.js
```diff
@@ -0,0 +1,28 @@
+import { brands } from 'src/libs/brands'
+import { getConfig } from 'src/libs/config'
+import * as Utils from 'src/libs/utils'
+
+
+// TODO: Deprecate Datastage (https://broadworkbench.atlassian.net/browse/SATURN-1414)
```
cahrens: Can you move the comment down to the line of code it relates to?

src/libs/brand-utils.js
```diff
@@ -0,0 +1,28 @@
...
+export const isAnvil = () => (window.location.hostname === brands.anvil.hostName) || getConfig().isAnvil
+export const isBaseline = () => (window.location.hostname === brands.baseline.hostName) || getConfig().isBaseline
+export const isBioDataCatalyst = () => (window.location.hostname === brands.bioDataCatalyst.hostName) || getConfig().isBioDataCatalyst
+export const isDatastage = () => (window.location.hostname === brands.datastage.hostName) || getConfig().isDatastage
+export const isElwazi = () => (window.location.hostname === brands.elwazi.hostName) || getConfig().isElwazi
+export const isFirecloud = () => (window.location.hostname === brands.firecloud.hostName) || getConfig().isFirecloud
+export const isProjectSingular = () => (window.location.hostname === brands.projectSingular.hostName) || getConfig().isProjectSingular
+export const isRareX = () => (window.location.hostname === brands.rareX.hostName) || getConfig().isRareX
+export const isTerra = () => (window.location.hostname === brands.terra.hostName) || getConfig().isTerra || (!isFirecloud() && !isDatastage() && !isAnvil() && !isBioDataCatalyst() && !isBaseline() && !isElwazi() && !isProjectSingular() && !isRareX())
```
cahrens: Nit: I wonder if it would be cleaner to have these `isX` methods call a common method with `Utils.case` so that they can have a default case for `isTerra`.

src/libs/brands.js
```diff
@@ -0,0 +1,146 @@
...
+  },
+  rareX: {
+    name: `The RARE${nonBreakingHyphen}X Data Analysis Platform`,
+    signInName: `the RARE${nonBreakingHyphen}X Data Analysis Platform`,
+    welcomeHeader: `Welcome to the RARE${nonBreakingHyphen}X Data Analysis Platform`,
+    description: `The RARE${nonBreakingHyphen}X Data Analysis Platform is a federated data repository of rare disease patient health data, including patient reported outcomes, clinical and molecular information. The platform is powered by Terra for biomedical researchers to access data, run analysis tools, and collaborate.`,
+    hostName: 'rare-x.terra.bio',
+    docLinks: [
+      {
+        link: 'https://rare-x.org/DataAnalysisPlatform-Documentation',
```
cahrens, line 173: This link returns a 404. Is that expected?

src/libs/colors.js
```diff
@@ -1,6 +1,6 @@
 import Color from 'color'
 import _ from 'lodash/fp'
-import { isAnvil, isBaseline, isBioDataCatalyst, isDatastage, isElwazi, isFirecloud, isProjectSingular, isRareX, isTerra } from 'src/libs/config'
+import { isAnvil, isBaseline, isBioDataCatalyst, isDatastage, isElwazi, isFirecloud, isProjectSingular, isRareX, isTerra } from 'src/libs/brand-utils'
```
cahrens, line 3: What do you think about moving the color palette into the brands data structure?

I'm thinking that in general whenever "isX" is called, it means something is missing from the brands data structure.

src/libs/logos.js
```diff
@@ -18,30 +18,13 @@ import rareXLogoWhite from 'src/images/brands/rareX/rarex-logo-white.svg'
...
-    [isElwazi(), () => longName ? `${maybeCapitalizedThe} eLwazi Open Data Science Platform` : 'eLwazi'],
-    [isProjectSingular(), () => 'Project Singular'],
-    [isRareX(), () => `${maybeCapitalizedThe} RARE${nonBreakingHyphen}X Data Analysis Platform`],
-    () => longName ? 'Terra Community Workbench' : 'Terra'
-  )
-}
-
-export const returnParam = () => getAppName().toLowerCase()
-
 const pickBrandLogo = (color = false) => Utils.cond(
```
cahrens: It seems like the brand logo should go into the brand structure.

src/pages/SignIn.js
```diff
@@ -20,12 +19,14 @@ const styles = {
...
     div({ style: { maxWidth: 600 } }, [
       div({ style: { fontSize: 36, color: colors.dark(0.6) } }, ['New User?']),
-      div({ style: { fontSize: 36, marginBottom: '2rem' } }, [`${getAppName({ capitalizeThe: true })} requires a Google Account.`]),
+      div({ style: { fontSize: 36, marginBottom: '2rem' } }, [`${brand.name} requires a Google Account.`]),
       div({ style: { fontSize: 16, lineHeight: 1.5, marginBottom: '2rem' } }, [
-        `${getAppName({ capitalizeThe: true })} uses your Google account. Once you have signed in and completed the user profile registration step, you can start using ${getAppName()}.`
+        `Once you have signed in and completed the user profile registration step, you can start using ${brand.signInName}.`
       ]),
       h(SignInButton),
       !isAnvil() && !isElwazi() && div({ style: styles.warningNoticeContainer }, [
```
cahrens, line 32: It seems like this warning text could also go into the brands structure.

src/libs/brand-utils.js
cahrens: Or even better if you can eliminate the need for `isX` methods entirely by putting everything into brands.

cahrens (approved):
I think this is a great first step, and I'm fine with you merging it as-is. But I think it would be good to continue moving more things into the brands structure with the goal of eliminating the need for the `isX` methods.

src/libs/brand-utils.js
kyuksel: I was thinking it'd be harder to miss up top but I don't feel strongly about it. Will move it.

cahrens: Thanks, I was just confused because it had no relationship to the following line.

src/libs/brand-utils.js
kyuksel: > Nit: I wonder if it would be cleaner to have these `isX` methods call a common method with `Utils.case` so that they can have a default case for `isTerra`.

`() => brands.terra` _is_ the case for defaulting to Terra. Perhaps I'm misunderstanding the ask?

> Or even better if you can eliminate the need for isX methods entirely by putting everything into brands.

That's what I'm hoping to do too but I don't want that work to be part of this PR.

src/libs/brands.js
kyuksel: Yes. I'll bring it to the PM's attention though, to make sure they're aware.

src/libs/colors.js
kyuksel: I agree with you and was envisioning it as part of the follow-up work.

src/libs/brand-utils.js
cahrens: I mean that it would be good to avoid the need for this part of the conditional:

```
(!isFirecloud() && !isDatastage() && !isAnvil() && !isBioDataCatalyst() && !isBaseline() && !isElwazi() && !isProjectSingular() && !isRareX())
```

I think a Utils.cond block that deals with all the different types would be able to accomplish this by having `isTerra` be the default condition.

src/libs/logos.js
kyuksel: My overall thinking on where things should go is as follows:

- If it's a _config_-like thing (stuff currently in `brands.js` that doesn't involve Javascript functions, etc.) that can be changed/added to by a non-software engineer, we put it in `brands.js`.
- If it's branding-specific but involves knowing some Javascript, it goes into `brand-utils.js`.

I was punting on moving the logo stuff into `brand-utils.js` because I wanted to think more holistically about what else can be made _config_-like as far as images go but I can go ahead and move the two functions `pickBrandLogo` and `brandLogoMaker` now.

src/pages/SignIn.js
kyuksel: I thought about that myself but I don't have a great idea about how to structure the stuff in `SignIn.js` in a way to could enable a non-software engineer persona to modify it so decided to punt on that as well.

kyuksel:
> I think this is a great first step, and I'm fine with you merging it as-is. But I think it would be good to continue moving more things into the brands structure with the goal of eliminating the need for the isX methods.

💯

src/libs/brand-utils.js
kyuksel: Good to know and a reason enough to move it for me 😊

src/libs/brand-utils.js
kyuksel: Yea I think that could be nicely handled by the further refactoring work I have in mind.

src/libs/logos.js
kyuksel: Changed in [c8e4697](https://github.com/DataBiosphere/terra-ui/pull/3257/commits/c8e4697cea9233df7c2e2fb58fdf2dcdd8472094).

src/libs/brand-utils.js
kyuksel: Changed in [9d11924](https://github.com/DataBiosphere/terra-ui/pull/3257/commits/9d11924c69e01df4057dd2b51b770eb6f3e1c02e).

src/libs/brand-utils.js
```diff
@@ -26,3 +43,14 @@ export const getEnabledBrand = () => Utils.cond(
   [isTerra(), () => brands.terra],
   () => brands.terra
 )
+
+export const pickBrandLogo = (color = false) => Utils.cond(
+  [isFirecloud(), () => color ? fcLogo : fcLogoWhite],
```
cahrens: What I meant is that each branch should return its own logo. I think you should be able to define everything about a brand in a single place.

cahrens: I think that would mean that each brand would need to return 2 different logos (color and white).

kyuksel: Ah I see. Changed in [6ffd0e3](https://github.com/DataBiosphere/terra-ui/pull/3257/commits/6ffd0e33c6fe793ccda024345afc45e42bbbeb9c).

For the time being, I've put the references to logo files inside `brands`. Depending on how the `non-software engineer` persona shapes up, it may make sense to move the references to another place as they could be considered code changes rather than "config" changes.

src/libs/brands.js
```diff
@@ -17,7 +38,11 @@ export const brands = {
         link: 'https://support.terra.bio/hc/en-us',
         text: 'Find how-to\'s, documentation, video tutorials, and discussion forums'
       }
-    ]
+    ],
+    logos: {
```
cahrens, line 43: Love it!

cahrens (approved):

src/libs/brand-utils.js
```diff
@@ -0,0 +1,34 @@
...
+// TODO: Deprecate Datastage (https://broadworkbench.atlassian.net/browse/SATURN-1414)
+export const isDatastage = () => (window.location.hostname === brands.datastage.hostName) || getConfig().isDatastage
+export const isElwazi = () => (window.location.hostname === brands.elwazi.hostName) || getConfig().isElwazi
+export const isFirecloud = () => (window.location.hostname === brands.firecloud.hostName) || getConfig().isFirecloud
+export const isProjectSingular = () => (window.location.hostname === brands.projectSingular.hostName) || getConfig().isProjectSingular
+export const isRareX = () => (window.location.hostname === brands.rareX.hostName) || getConfig().isRareX
+export const isTerra = () => (window.location.hostname === brands.terra.hostName) || getConfig().isTerra ||
+  (!isFirecloud() && !isDatastage() && !isAnvil() && !isBioDataCatalyst() && !isBaseline() && !isElwazi() && !isProjectSingular() && !isRareX())
+
+export const getEnabledBrand = () => Utils.cond(
```
nawatts, line 18: To reduce some repetitive code, I think I would propose swapping the `isBrand` and `getEnabledBrand` functions. Maybe something like:

```js
export const getEnabledBrand = () => _.find(brand => brand.hostname ==== window.location.hostname || getConfig().brand === brand.name, brands) || brands.terra

export const isAnvil = () => getEnabledBrand() === brands.anvil
```

src/pages/LandingPage.js
```diff
@@ -62,30 +51,27 @@ const makeCard = (link, title, body) => h(Clickable, {
...
+      style: { fontSize: 18 }
+    }, [
+      title,
+      icon('pop-out', { size: 18, style: { marginLeft: '0.5rem' } })
+    ])
+  ])
+}
+
+const makeDocLinks = linkTextPairs => _.map(
+  ({ link, text }) => makeDocLink(link, text), linkTextPairs
```
nawatts: Nit: If `makeDocLink` took one of the link configuration objects as a single parameter vs each value as separate parameters, this could be simplified a bit.

```js
const makeDocLink = ({ link, text }) => ...

const makeDocLinks = _.map(makeDocLink)
```

src/pages/SignIn.js
```diff
@@ -20,12 +19,14 @@ const styles = {
...
+  const brand = getEnabledBrand()
+
   return h(HeroWrapper, { showMenu: false, showDocLink: true }, [
     div({ style: { maxWidth: 600 } }, [
       div({ style: { fontSize: 36, color: colors.dark(0.6) } }, ['New User?']),
-      div({ style: { fontSize: 36, marginBottom: '2rem' } }, [`${getAppName({ capitalizeThe: true })} requires a Google Account.`]),
+      div({ style: { fontSize: 36, marginBottom: '2rem' } }, [`${brand.name} requires a Google Account.`]),
       div({ style: { fontSize: 16, lineHeight: 1.5, marginBottom: '2rem' } }, [
-        `${getAppName({ capitalizeThe: true })} uses your Google account. Once you have signed in and completed the user profile registration step, you can start using ${getAppName()}.`
+        `Once you have signed in and completed the user profile registration step, you can start using ${brand.signInName}.`
```
nawatts, line 29: `brand.signInName` is used both for a user-facing label here and also to construct return URLs for Firecloud. Should those two things use separate configuration values? Are they guaranteed to always use the same values?

nawatts (approved):
I agree with all of the feedback @cahrens already gave.

> I think this is a great first step, and I'm fine with you merging it as-is. But I think it would be good to continue moving more things into the brands structure with the goal of eliminating the need for the isX methods.

> What do you think about moving the color palette into the brands data structure?

> I'm thinking that in general whenever "isX" is called, it means something is missing from the brands data structure.

src/libs/brand-utils.js
kyuksel: I like the idea. However, `getConfig()` doesn't have a defined `brand` property.

My goal is to further refactor the code around this and I think your comment will be taken care of then.

nawatts: > However, getConfig() doesn't have a defined brand property.

Yes, this proposal would change the way that the current brand is manually overridden from `configOverridesStore.set({ isAnvil: true })` to `configOverridesStore.set({ brand: 'AnVIL' })`. The current approach allows getting into an invalid state by activating multiple brands at once (`configOverridesStore.set({ isAnvil: true, isBaseline: true })`).

src/pages/SignIn.js
kyuksel: I've thought about the same thing and I agree that it'd be ideal to have distinct properties for them. I initially felt that the likelihood of a divergence is small enough to not warrant increasing the size of `brands` noticeably but I don't feel confident anymore plus code readability matters so I'll implement your suggestion.

src/pages/LandingPage.js
kyuksel: Changed in [0f65cf6](https://github.com/DataBiosphere/terra-ui/pull/3257/commits/0f65cf6b943fcd45ebc65d53e4ba004743ea2ea6).

src/libs/brand-utils.js
kyuksel: You're right. What I have in mind for the follow-up refactoring also involves providing a helper (e.g. `enableBrand('baseline')` or perhaps `enableBrand(brands.baseline)`) and hiding the implementation detail (i.e. `configOverridesStore.set({ isBrand: true })`), which should also take care of the issue you raised. 

Since the functional issue you raised is a development feature and is unlikely to happen to a non-developer, I'd like to defer the fix to the follow-up refactoring. Are you okay with that?

nawatts: Sounds good.

src/pages/SignIn.js
kyuksel: Changed in [a8e8598](https://github.com/DataBiosphere/terra-ui/pull/3257/commits/a8e8598f0c8066356145bf35c6acf3ddcd4dbae6).

PR #3261, opened 2022-07-26
DC-452: remove selection UI and export to workspace UI from the catalog list page

pshapiro4broad:
This change removes the selection checkboxes, selection feedback text, and the `Link to a workspace` UI from the catalog list page. The link to a workspace feature is not yet working, so an error is generated if a user tries to use this feature.

Rather than commenting out the code, this change removes the code completely since we may not support this operation in the list view after all, depending on the difficulty of implementation and the user need for this feature.

Note that there are still disabled tests that reference the UI in this page. I didn't remove these tests because the Link feature will be implemented soon on the Catalog Details page (https://broadworkbench.atlassian.net/browse/DC-328) and the same test can be used for that page instead of the list page.

Before
image

After
image


s-rubenstein (approved):
LGTM

fboulnois (approved):

breilly2 (approved):

PR #3262, opened 2022-07-26
[IA-3599] Add unit test to ContextBar.js

n8landolt:
## Description
Add tests for ContextBar.js

### DIFF
--- a/src/components/ContextBar.test.js
+++ b/src/components/ContextBar.test.js
@@ -5,10 +5,29 @@ import { div, h } from 'react-hyperscript-helpers'
 import { tools } from 'src/components/notebook-utils'
 import { MenuTrigger } from 'src/components/PopupTrigger'
 import { Ajax } from 'src/libs/ajax'
+import { getGalaxyComputeCost, getGalaxyDiskCost, getPersistentDiskCostHourly, getRuntimeCost, runtimeConfigCost } from 'src/libs/runtime-utils'
+import * as Utils from 'src/libs/utils'
+import { ContextBar } from 'src/pages/workspaces/workspace/analysis/ContextBar'
 import { CloudEnvironmentModal } from 'src/pages/workspaces/workspace/analysis/modals/CloudEnvironmentModal'
 
-import { ContextBar } from '../ContextBar'
 
+const GALAXY_COMPUTE_COST = 10
+const GALAXY_DISK_COST = 1
+const RUNTIME_COST = 0.1
+const PERSISTENT_DISK_COST = 0.01
+
+jest.mock('src/libs/runtime-utils', () => {
+  const originalModule = jest.requireActual('src/libs/runtime-utils')
+
+  return {
+    ...originalModule,
+    getGalaxyComputeCost: jest.fn(),
+    getGalaxyDiskCost: jest.fn(),
+    getPersistentDiskCostHourly: jest.fn(),
+    getRuntimeCost: jest.fn(),
+    runtimeConfigCost: jest.fn()
+  }
+})
 
 // Mocking for terminalLaunchLink using Nav.getLink
 jest.mock('src/libs/nav', () => {
@@ -75,6 +94,22 @@ beforeEach(() => {
         runtime: mockRuntime
       }
     }
+  })
+
+  getGalaxyComputeCost.mockImplementation(() => {
+    return GALAXY_COMPUTE_COST
+  })
+  getGalaxyDiskCost.mockImplementation(() => {
+    return GALAXY_DISK_COST
+  })
+  getRuntimeCost.mockImplementation(() => {
+    return RUNTIME_COST
+  })
+  getPersistentDiskCostHourly.mockImplementation(() => {
+    return PERSISTENT_DISK_COST
+  })
+  runtimeConfigCost.mockImplementation(() => {
+    return RUNTIME_COST + PERSISTENT_DISK_COST
   })
 })
 
@@ -143,7 +178,7 @@ const galaxyDisk = {
   zone: 'us-central1-a'
 }
 
-const jupyter1 = {
+const jupyter = {
   id: 75239,
   workspaceId: null,
   runtimeName: 'saturn-eae9168f-9b99-4910-945e-dbab66e04d91',
@@ -160,7 +195,7 @@ const jupyter1 = {
   },
   runtimeConfig: {
     machineType: 'n1-standard-1',
-    persistentDiskId: 14692,
+    persistentDiskId: 15778,
     cloudService: 'GCE',
     bootDiskSize: 120,
     zone: 'us-central1-a',
@@ -224,8 +259,8 @@ const azureRunning = {
   }
 }
 
-const jupyter1Disk = {
-  id: 14692,
+const runtimeDisk = {
+  id: 15778,
   googleProject: 'terra-dev-cf677740',
   cloudContext: {
     cloudProvider: 'GCP',
@@ -270,14 +305,14 @@ const rstudioRuntime = {
   },
   runtimeConfig: {
     machineType: 'n1-standard-4',
-    persistentDiskId: 15774,
+    persistentDiskId: 15778,
     cloudService: 'GCE',
     bootDiskSize: 120,
     zone: 'us-central1-a',
     gpuConfig: null
   },
   proxyUrl: 'https://leonardo.dsde-dev.broadinstitute.org/proxy/terra-dev-98897219/saturn-48afb74a-15b1-4aad-8b23-d039cf8253fb/rstudio',
-  status: 'Creating',
+  status: 'Running',
   labels: {
     saturnWorkspaceNamespace: 'general-dev-billing-account',
     'saturn-iframe-extension': 'https://bvdp-saturn-dev.appspot.com/jupyter-iframe-extension.js',
@@ -326,8 +361,8 @@ describe('ContextBar - buttons', () => {
     // Arrange
     const jupyterContextBarProps = {
       ...contextBarProps,
-      runtimes: [jupyter1],
-      persistentDisks: [jupyter1Disk]
+      runtimes: [jupyter],
+      persistentDisks: [runtimeDisk]
     }
 
     // Act
@@ -338,16 +373,39 @@ describe('ContextBar - buttons', () => {
     expect(screen.getByLabelText('Environment Configuration'))
     expect(screen.getByLabelText(new RegExp(/Jupyter Environment/i)))
     expect(screen.getByLabelText('Terminal button')).toBeEnabled()
+    expect(screen.getByText(Utils.formatUSD(RUNTIME_COST + PERSISTENT_DISK_COST)))
+    expect(screen.getByText(/Running \$.*\/hr/))
+  })
+
+  it('will render Jupyter in Creating status', () => {
+    // Arrange
+    const jupyterContextBarProps = {
+      ...contextBarProps,
+      runtimes: [{ ...jupyter, status: 'Creating' }],
+      persistentDisks: [runtimeDisk]
+    }
+
+    // Act
+    render(h(ContextBar, jupyterContextBarProps))
+
+    //Assert
+    expect(screen.getByText('Rate:'))
+    expect(screen.getByLabelText('Environment Configuration'))
+    expect(screen.getByLabelText(new RegExp(/Jupyter Environment/i)))
+    expect(screen.getByLabelText('Terminal button')).toBeEnabled()
+    expect(screen.getByText(Utils.formatUSD(RUNTIME_COST + PERSISTENT_DISK_COST)))
+    expect(screen.getByText(/Creating \$.*\/hr/))
+    expect(screen.getByText(/Disk \$.*\/hr/))
   })
 
   it('will render Galaxy and RStudio buttons with a disabled Terminal Button', () => {
     // Arrange
     const rstudioGalaxyContextBarProps = {
       ...contextBarProps,
-      runtimes: [rstudioRuntime],
+      runtimes: [{ ...rstudioRuntime, status: 'Creating' }],
       apps: [galaxyRunning],
       appDataDisks: [galaxyDisk],
-      persistentDisks: [jupyter1Disk]
+      persistentDisks: [runtimeDisk]
     }
 
     // Act
@@ -355,10 +413,14 @@ describe('ContextBar - buttons', () => {
 
     //Assert
     expect(screen.getByText('Rate:'))
+    expect(screen.getByText(Utils.formatUSD(RUNTIME_COST + GALAXY_COMPUTE_COST + GALAXY_DISK_COST + PERSISTENT_DISK_COST)))
     expect(screen.getByLabelText('Environment Configuration'))
     expect(screen.getByLabelText(new RegExp(/RStudio Environment/i)))
     expect(screen.getByLabelText(new RegExp(/Galaxy Environment/i)))
     expect(screen.getByLabelText('Terminal button')).toHaveAttribute('disabled')
+    expect(screen.getByText(/Running \$.*\/hr/))
+    expect(screen.getByText(/Creating \$.*\/hr/))
+    expect(screen.getByText(/Disk \$.*\/hr/))
   })
 
   it('will render a Cromwell button with a disabled Terminal Button', () => {
@@ -374,6 +436,7 @@ describe('ContextBar - buttons', () => {
 
     //Assert
     expect(screen.getByText('Rate:'))
+    expect(screen.getByText('$0.00'))
     expect(screen.getByLabelText('Environment Configuration'))
     expect(screen.getByLabelText('Terminal button')).toHaveAttribute('disabled')
     expect(screen.getByLabelText(new RegExp(/Cromwell Environment/i)))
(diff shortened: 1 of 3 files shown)


src/components/_tests/ContextBar.test.js
```diff
@@ -0,0 +1,152 @@
...
+      runtimes={runtimes}
+      apps={apps}
+      appDataDisks={appDataDisks}
+      refreshRuntimes={refreshRuntimes}
+      location={location}
+      locationType={locationType}
+      refreshApps={refreshApps}
+      workspace={workspace}
+      persistentDisks={persistentDisks}
+    />)
```
kyuksel: Could this be

```javascript
    render(h(ContextBar, {
      runtimes,
      apps,
      appDataDisks,
      refreshRuntimes,
      location,
      locationType,
      refreshApps,
      workspace,
      persistentDisks
    }))
```
?

src/components/_tests/ContextBar.test.js
```diff
@@ -0,0 +1,392 @@
...
+            start: jest.fn()
+          }
+        }
+      }
+    }
+  })
+})
+
+//Note - These constants are copied from src/libs/runtime-utils.test.js
+const galaxyRunning = {
```
jdcanas: can we make a common test data source and override the necessary fields?

n8landolt: I discussed that with Miguel, and I believe our agreement was that for the time being we'll keep the data in this file.

src/components/_tests/ContextBar.test.js
n8landolt: Taking a different approach - this is no longer relevant!

src/components/_tests/ContextBar.test.js
```diff
@@ -0,0 +1,561 @@
...
+    render(h(ContextBar, jupyterContextBarProps))
+
+    //Assert
+    expect(screen.getByText('Rate:'))
+    expect(screen.getByLabelText('Environment Configuration'))
+    expect(screen.getByLabelText(new RegExp(/Azure Environment/i)))
+    expect(screen.getByLabelText('Terminal button')).toHaveAttribute('disabled')
+  })
+
+  it('will render Azure Environment button', () => {
```
n8landolt: Fixing title and adding an appropriate assertion.

src/components/_tests/ContextBar.test.js
jdcanas: I discussed this with Miguel. His argument is that this creates dependencies in tests, but since a lot of tests will need a generic 'running galaxy' runtime (for example), fields can be overrode as needed. This is a pattern we use everywhere for leo tests

I am a strong believer we should form up a common data file for runtimes/disks/etc type data.

Especially since this data is duplicated almost verbatim, and there will be thousands of lines of duplicated data otherwise. 

Very important to have a strong set of 'base' data to extends from here, so if you test needs something specific (i.e. a stopped runtime over a running one), I'd encourage defining a base model and having your test override that specific field in conjunction with the spread operator over the base model.

jdcanas:
This is great! Tons of test coverage. A few things things that come to mind in terms of coverage (aka while the lines in question may be hit according to the coverage tool, that does not mean we are necessarily verifying that the branch had the correct visual result):

- perhaps we want to test that the tooltip for a runtime has the proper info (namely cost and status)
     - the test could simply verify that the tooltip contains the text from the cost function, not verify the cost calculation itself  
- Testing that the status icon color is correct for a runtime against an array of all possible statuses (see `CreateAzureBillingProjectModalTest` for how to test using an array of values as input)

src/components/_tests/ContextBar.test.js
jdcanas: If you'd like, you can keep it as-is, and I can refactor.

src/components/_tests/ContextBar.test.js
```diff
@@ -0,0 +1,561 @@
+import '@testing-library/jest-dom'
+
+import { fireEvent, render, screen } from '@testing-library/react'
+import { div, h } from 'react-hyperscript-helpers'
+import { tools } from 'src/components/notebook-utils'
+import { MenuTrigger } from 'src/components/PopupTrigger'
+import { Ajax } from 'src/libs/ajax'
+import { CloudEnvironmentModal } from 'src/pages/workspaces/workspace/analysis/modals/CloudEnvironmentModal'
+
+import { ContextBar } from '../ContextBar'
```
jdcanas: Probably shouldn't use a relative import

src/components/_tests/ContextBar.test.js
```diff
@@ -0,0 +1,561 @@
...
+    const rstudioGalaxyContextBarProps = {
+      ...contextBarProps,
+      runtimes: [rstudioRuntime],
+      apps: [galaxyRunning],
+      appDataDisks: [galaxyDisk],
+      persistentDisks: [jupyter1Disk]
+    }
+
+    // Act
+    render(h(ContextBar, rstudioGalaxyContextBarProps))
```
jdcanas: do you not need to do `await act(async () => render(...))` here? I was getting a lot of warnings before I did...

jdcanas (changes requested):
Thinks its very necessary to have single source of truth for test data. Imagine if a runtime has a new field that is used in a lot of places all of a sudden, but the tests don't really care what value it is. Would require changing all of the runtimes in our test code...

src/components/_tests/ContextBar.test.js
n8landolt: I haven't had a problem with it in it's current state

src/components/_tests/ContextBar.test.js
msilva-broad: My thoughts on this topic are certainly colored by working in Typescript codebases.

With types on these data objects, I do prefer a bias towards locally specified mock objects for all but auto-generated large datasets... for a few reasons:
1) when the shape of an object changes, its great to get the signal of the impact by seeing and fixing the errors across the various code and test files.
2) the signal in (1) is a good code smell to maybe address DRY or SOLID in the code structure to cut down on overly broad propagation/repetition of business logic, etc.
3) with Typescript, individual test files can use Partial<T> to reduce the verbosity of certain mock objects to just the essential properties/sub-objects that drive the given code-under-test.
4) point (3) can also be a code smell that smaller interfaces should be defined to reduce the object-shape dependency surfaces of the code-under-test, which improves code reusability.
5) avoids cross-talk and dependency tentacles between test files.

Having said all that, this is not a hill I would die on.  What I would recommend though, is to refrain from any test data unification until AFTER the components folder is better organized to not be so flat, allowing team/functional areas to have a more natural place to put their specific test data.  Also, having Typescript in place first would help with cases where teams decide to use base mock objects with ...spread overrides, to assist in type correctness and discoverability.

src/components/_tests/ContextBar.test.js
msilva-broad: I would instead push to move this test file to be co-located next to the code-under-test... I think this was discussed a bit in the UIWG quite a few weeks ago... but my memory is fuzzy.  We may not have gotten a hard consensus on it.

n8landolt: You're correct - and Justin's comment reminded me to move the test!

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,626 @@
...
+  }
+}
+
+describe('ContextBar - buttons', () => {
+  it('will render default icons', () => {
+    // Act
+    render(h(ContextBar, contextBarProps))
+
+    // Assert
+    expect(screen.getByText('Rate:'))
```
nawatts: Instead of querying all of the document via `screen`, can these tests use the queries returned by `render`?

https://testing-library.com/docs/react-testing-library/api#queries-1

n8landolt: Is there a particular advantage or difference to this? I didn't realize it was an option!

nawatts: I'll claim it's more idiomatic.

It may also be useful if there's a case like a component opens a modal and you want to query the component or the modal specifically.

n8landolt: Ahh I see now. Let me make the changes!

nawatts: Also good for test isolation. If another test leaves some elements in the DOM, querying with screen could pick those up.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+})
+
+// Mocking for terminalLaunchLink using Nav.getLink
+jest.mock('src/libs/nav', () => {
+  const originalModule = jest.requireActual('src/libs/nav')
+
+  return {
+    ...originalModule,
+    getPath: jest.fn(() => '/test/'),
+    getLink: jest.fn(() => '/')
```
nawatts: [CRA's default configuration](https://create-react-app.dev/docs/running-tests/#configuration) [resets mocks](https://jestjs.io/docs/configuration#resetmocks-boolean) between tests. Do these implementations ever get used?

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+
+  return {
+    ...originalModule,
+    getGalaxyComputeCost: jest.fn(),
+    getGalaxyDiskCost: jest.fn(),
+    getPersistentDiskCostHourly: jest.fn(),
+    getRuntimeCost: jest.fn(),
+    runtimeConfigCost: jest.fn()
+  }
+})
```
nawatts: A minor stylistic thing, but these can be written a little more concisely as:

```suggestion
jest.mock('src/pages/workspaces/workspace/analysis/runtime-utils', () => ({
  ...jest.requireActual('src/pages/workspaces/workspace/analysis/runtime-utils'),
  getGalaxyComputeCost: jest.fn(),
  getGalaxyDiskCost: jest.fn(),
  getPersistentDiskCostHourly: jest.fn(),
  getRuntimeCost: jest.fn(),
  runtimeConfigCost: jest.fn()
}))
```

Might be nice when there are many modules being mocked.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+  const originalModule = jest.requireActual('src/libs/config')
+  return {
+    ...originalModule,
+    isCromwellAppVisible: () => {
+      return true
+    }
+  }
+})
+
+// Necessary to mock the AJAX module.
```
nawatts: This comment seems unnecessary. The next line of code is self explanatory.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+jest.mock('src/libs/ajax')
+const mockRuntimesStartFn = jest.fn()
+const mockRuntime = jest.fn()
```
nawatts, line 56: I don't see any `expect`s using this function. Does it need to be mocked?

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+      },
+      Runtimes: {
+        runtime: mockRuntime
+      }
+    }
+  })
+
+  getGalaxyComputeCost.mockImplementation(() => {
+    return GALAXY_COMPUTE_COST
+  })
```
nawatts, line 83: These could all use `mockReturnValue` instead of `mockImplementation`.

```suggestion
  getGalaxyComputeCost.mockReturnValue(GALAXY_COMPUTE_COST)
```

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+    return PERSISTENT_DISK_COST
+  })
+  runtimeConfigCost.mockImplementation(() => {
+    return RUNTIME_COST + PERSISTENT_DISK_COST
+  })
+})
+
+afterEach(() => {
+  jest.clearAllMocks()
+})
```
nawatts, line 100: CRA's default configuration will do this automatically via `restoreMocks`.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+
+//Note - These constants are copied from src/libs/runtime-utils.test.js
```
nawatts, line 102: If these are used in multiple tests, it would be nice to move them to a separate module instead of duplicating them.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+  }
+}
+
+describe('ContextBar - buttons', () => {
+  it('will render default icons', () => {
+    // Act
+    const { getByText, getByLabelText } = render(h(ContextBar, contextBarProps))
+
+    // Assert
+    expect(getByText('Rate:'))
```
nawatts, line 337: `expect` without a matcher is kind of unusual.

It looks like this test checks for the presence of an element by relying on the `getBy...` query to throw if it doesn't find a matching element.

This could either omit the `expect` or use `expect(getByText('Rate:')).toBeTruthy()`.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+    expect(getByLabelText('Environment Configuration'))
+    expect(getByLabelText('Terminal button')).toHaveAttribute('disabled')
```
nawatts, line 339: Would jest-dom's [`toBeDisabled` matcher](https://github.com/testing-library/jest-dom#tobedisabled) work here?.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+    }
+
+    // Act
+    const { getByText, getByLabelText } = render(h(ContextBar, jupyterContextBarProps))
+
+    //Assert
+    expect(getByText('Rate:'))
+    expect(getByLabelText('Environment Configuration'))
+    expect(getByLabelText(new RegExp(/Jupyter Environment/i)))
+    expect(getByLabelText('Terminal button')).toBeEnabled()
```
nawatts, line 357: It looks like this is actually testing more than it claims to ("will render Jupyter button with an enabled Terminal Button"). Might be worth splitting some of this out into separate tests or relabeling the test.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+    expect(getByText(Utils.formatUSD(RUNTIME_COST + PERSISTENT_DISK_COST)))
+    expect(getByText(/Creating \$.*\/hr/))
+  })
+
+  it('will render Galaxy and RStudio buttons with a disabled Terminal Button', () => {
```
nawatts, line 382: Why test Jupyter on its own but Galaxy and RStudio together?

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+  it('clicking environment configuration opens CloudEnvironmentModal', () => {
+    // Act
+    const { getByText, getByLabelText } = render(h(ContextBar, contextBarProps))
+
+    const envConf = getByLabelText('Environment Configuration')
+    fireEvent.click(envConf)
+
+    // Assert
+    expect(getByText('Cloud Environment Details'))
+  })
```
nawatts, line 478: Nit: formatting. Most other tests have an empty line in between.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+      persistentDisks: [runtimeDisk]
+    }
+
+    // Act
+    const { getByText, getByLabelText } = render(h(ContextBar, jupyterContextBarProps))
+    fireEvent.click(getByLabelText(new RegExp(/Jupyter Environment/i)))
+
+    // Assert
+    getByText('Cloud Environment Details')
+    getByText(tools.Jupyter.label)
```
nawatts, line 494: It may also be worth testing for the absence of other tools here.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+    // Arrange
+    global.window = Object.create(window)
+    const url = 'http://dummy.com'
+    Object.defineProperty(window, 'location', {
+      value: {
+        href: url
+      },
+      writable: true,
+      hash: '/'
+    })
```
nawatts, line 543: It might be worth adding a function to Nav that wraps access to `window.location` to make mocking easier.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,625 @@
...
+      ],
+      persistentDisks: [runtimeDisk]
+    }
+
+    // Act
+    const { getByLabelText } = render(h(ContextBar, jupyterContextBarProps))
+    fireEvent.click(getByLabelText('Terminal button'))
+
+    // Assert
+    expect(mockRuntimesStartFn).toBeCalledTimes(0)
```
nawatts: Either way is fine, but I think `expect(mockRuntimesStartFn).not.toHaveBeenCalled()` reads a little more clearly for things like this.

nawatts (approved):

jdcanas (approved):
looks good, lets just remember to try to use the `testData` file added in the future after that is mered.

src/components/ContextBar.test.js
```diff
@@ -0,0 +1,610 @@
+import '@testing-library/jest-dom'
```
msilva-broad, line 1: the location of ContextBar.js has moved to a more dedicated area owned by IA team:

src/pages/workspaces/workspace/analysis/ContextBar.js

Also, because the UIWG has approved the convention of co-locating test files to be right next to the file being tested, this .test.js file should move over to that new location as well.

cahrens: @jdcanas has moved this file in his PR: https://github.com/DataBiosphere/terra-ui/pull/3422

PR #3263, opened 2022-07-26
[QA-1972] Clean up integration-tests circleci config.yaml

aweng98:
[QA-1972](https://broadworkbench.atlassian.net/browse/QA-1972)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,19 @@ executors:
     docker:
       - image: google/cloud-sdk:alpine
 
+anchors:
+  # Run jobs on PR branches only
+  filter-pr-branch: &filter-pr-branch
+    filters:
+      branches:
+        ignore: dev
+
+  # Run jobs on Dev branch only
+  filter-dev-branch: &filter-dev-branch
+    filters:
+      branches:
+        only: dev
+
 commands:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -17,6 +17,19 @@ executors:
     docker:
       - image: google/cloud-sdk:alpine
 
+anchors:
+  # Run jobs on PR branches only
```
dmohs: This comment isn't always true since you can place an anchor in multiple locations. I'd remove both of these, since the implementation speaks for itself.

.circleci/config.yml
```diff
@@ -128,8 +141,9 @@ commands:
         type: boolean
         default: false
       env:
-        type: string
-        default: local
+        type: enum
+        enum: [ "local", "staging", "alpha" ]
+        default: "local"
```
dmohs: I've come to avoid quotes unless necessary, since they add a little noise.

dmohs (approved):
This is fantastic. Substantial de-duplication and a great example of applying knowledge across teams (I've seen this in AoU's Circle config).

.circleci/config.yml
aweng98: yeah, quotes are removed.

nawatts (approved):

PR #3264, opened 2022-07-26
DC-488: remove footer from Browser (catalog) component

pshapiro4broad:
Now that Browser is a component embedded in another page, it no longer needs to include the footer element. This change was missed when the catalog code was refactored to share the same route as the existing library data page.

image

### DIFF
--- a/src/pages/library/DataBrowser.js
+++ b/src/pages/library/DataBrowser.js
@@ -2,7 +2,6 @@ import _ from 'lodash/fp'
 import { useState } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
 import { ButtonOutline, ButtonPrimary, ButtonSecondary, Checkbox, Link, spinnerOverlay } from 'src/components/common'
-import FooterWrapper from 'src/components/FooterWrapper'
 import { icon } from 'src/components/icons'
 import { MiniSortable, SimpleTable } from 'src/components/table'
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
oops 👍

src/pages/library/DataBrowser.js
```diff
@@ -217,7 +216,7 @@ export const Browser = () => {
 
   const toggleSelectedData = data => setSelectedData(_.xor([data]))
 
-  return h(FooterWrapper, { alwaysShow: true }, [
+  return [
```
nawatts: Currently, this results in a React warning: `Warning: Each child in a list should have a unique "key" prop.`.

This can be resolved by returning a [Fragment](https://reactjs.org/docs/fragments.html) element here instead of an array.

```js
import { Fragment } from 'react'

...

return h(Fragment, [
  h(SearchAndFilterComponent, {
  ...
])

nawatts (changes requested):

s-rubenstein (approved):
LGTM (pending @nawatts change)

src/pages/library/DataBrowser.js
s-rubenstein: Hi @nawatts! I took this over since Phil is OOO for the rest of the week. I think things should be resolved now, would you mind taking a look?

nawatts (approved):

PR #3269, opened 2022-07-26
[SU-147] Read submission and workflow GCS links from Rawls/Cromwell rather than assuming

MatthewBemis:
This is in preparation for [SU-166](https://broadworkbench.atlassian.net/browse/SU-166). Previously, Terra UI would make assumptions about where the submission and workflow files lived because they only ever lived in one particular predictable path. With SU-166, that is no longer the case- so we need to pull that information from a source of truth (either rawls, or cromwell via rawls)

Submission ID link in the submission table:
Screen Shot 2022-07-26 at 6 47 33 PM

Submission ID link in submission details:
Screen Shot 2022-07-26 at 6 47 43 PM

Execution directory in workflow links:
Screen Shot 2022-07-26 at 6 47 47 PM

Execution Directory link in the Workflow dashboard:
Screen Shot 2022-07-26 at 6 47 56 PM

<!--
Hello, friend!
Thanks!

--!>


src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
```diff
@@ -195,7 +195,7 @@ const SubmissionWorkflowsTable = ({ workspace, submission }) => {
                 inputName && h(Link, {
                   key: 'directory',
                   ...Utils.newTabLinkProps,
-                  href: bucketBrowserUrl(`${bucketName}/${submissionId}/${inputName.split('.')[0]}/${workflowId}`),
+                  href: bucketBrowserUrl(`${submissionRoot.replace('gs://', '')}/${inputName.split('.')[0]}/${workflowId}`),
```
nawatts, line 198: Based on changes to WorkflowDashboard.js, should this also use `workflowRoot` instead of constructing a URL from the input and workflow ID?

MatthewBemis: The `workflowRoot` is only available if we load an individual workflow from Rawls, which can pull additional optional fields from Cromwell. 

In this submission details context, we only load the individual submission which only gives very basic details about the workflow. There is no way to grab the `workflowRoot` for all of the workflows in this list view, so sadly we still have to do _some_ url constructing

nawatts (approved):

src/pages/workspaces/workspace/jobHistory/SubmissionDetails.js
nawatts: Ok. And since we're only changing `submissionRoot`, `workflowRoot` will still always follow this pattern?

MatthewBemis: As far as I know it will. Unfortunately we're at the whim of convention here because there is no other scalable way to retrieve that info

cjllanwarne (approved):

PR #3271, opened 2022-07-27
[SU-168] Reset Appcues on sign out

nawatts:
Currently, Appcues' user ID remains after signing out of Terra. This can result in Appcues showing content like the NPS survey when there is no user logged in.

To see the effect of this change, run `window.Appcues.debug()` and then sign out of Terra (note the "User Identified" section of the "Appcues Status" popup).

See https://docs.appcues.com/article/161-javascript-api for more info on the Appcues API.


MatthewBemis (approved):
lgtm, tested it out too 👍

breilly2 (approved):

PR #3273, opened 2022-07-27
[QA-1980] Add Slack function in Terra-UI integration test project

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1980

Part of work to implement an automated test failures Slack notification to teams' Slack channels. 

Using `node-fetch` to keep it simple and without adding additional dependency libraries. `post-message-test.js` is a temporary test script to show `post-message.js` is working as designed.

Screen Shot 2022-07-27 at 4 12 08 PM

Screen Shot 2022-07-27 at 4 11 57 PM


integration-tests/slack/post-message-test.js
```diff
@@ -0,0 +1,23 @@
+const { postMessage } = require('./post-message')
+
+
+// Post a message to Slack channels
+const postMessageTest = async () => {
```
cahrens: I am not clear what this is testing. Just that `postMessage` doesn't error?

cahrens: Is this for debugging only (not executed in integration test suite)? Some comments would be helpful.

aweng98: `post-message-test.js` is to show `post-message.js` is working. I forgot to say I will remove it in PR before merge.

cahrens (approved):

integration-tests/slack/post-message.js
```diff
@@ -0,0 +1,48 @@
...
+    responseJson = await fetchResponse.json()
+  } catch (e) {
+    console.error(`**  ERROR: Encountered unexpected error when posting message to Slack channel: ${channel}.`, e)
+    throw e
+  }
+  if (responseJson.ok) {
+    console.log(`Successfully posted message to Slack channel: ${channel}. Message: ${JSON.stringify(data)}`)
+  } else {
+    console.error(`**  ERROR: ${new Date().toTimeString()}: Failed to post message to Slack channel: ${channel}.`, responseJson.error)
+    throw new responseJson.error()
```
nawatts: It looks like `responseJson.error` is a string containing an error code from the Slack API, not a function.
https://api.slack.com/methods/chat.postMessage#examples

```suggestion
    throw new Error(responseJson.error)
```

integration-tests/slack/post-message.js
```diff
@@ -0,0 +1,48 @@
...
+
+
+const defaultToken = process.env.SLACK_BOT_TOKEN
+const defaultChannel = process.env.SLACK_CHANNEL_ID
+const apiUrl = 'https://slack.com/api'
+
+const postMessage = async ({ channel = defaultChannel, token = defaultToken, blocks }) => {
+  if (!channel) {
+    console.error(`**  ERROR: Missing Slack channel. Failed to post message to Slack.`)
+    return
```
nawatts: Should these cases throw errors instead of returning?

nawatts (changes requested):
It looks like there is a type error in some of the error handling code.

integration-tests/slack/post-message.js
aweng98: I thought about it before. If throws error here and all tests passed, job would be marked as fail. I didn't want to see that. What do you think?

integration-tests/slack/post-message.js
aweng98: Good catch!

integration-tests/slack/post-message.js
kyuksel: It feels like having a false alarm (due to a test failure because of a Slack notification issue) is better than risking not getting a notification when there's an actual test failure (because of a Slack notification issue that's swallowed), no?

nawatts: There's currently some mixed messaging in this function. If no channel or token is configured, it logs "ERROR: ..." but then returns. I think either of these options would be preferable:
- Missing channel / token configuration is treated as misconfiguration and errors.
- Missing channel / token configuration is treated as disabling Slack notifications and does not log an error (maybe an info or warning).

> If throws error here and all tests passed, job would be marked as fail. I didn't want to see that.

On this subject, this function currently throws an error if the Slack API request fails. Should that fail the test?

nawatts (approved):

integration-tests/slack/post-message.js
aweng98: If all tests passed but the Slack request failed. it would cause CircleCI job to fail. 
I agree with your reasoning, I'm changing this to throw error instead return.
Thank you all.

PR #3275, opened 2022-07-28
[TUIM-6] Add CSP header

nawatts:
Currently, we use a webpack plugin to add a `meta` element containing a [content security policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).

https://github.com/DataBiosphere/terra-ui/blob/5de7c1eaf024bbc9981ef1842c5ffd35ba51a3bb/config-overrides.js#L17-L32

This adds a [`Content-Security-Policy` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) as well. This policy was copied from `view-source:https://app.terra.bio/`.

One thing that the webpack plugin added that is not included in this static header is a few [nonce sources](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources#sources). Currently, the CSP generated by the webpack plugin is:
```
base-uri 'self'; object-src 'none'; script-src 'self' https://fast.appcues.com https://us.jsagent.tcell.insight.rapid7.com https://cdnjs.cloudflare.com 'nonce-2q2A4++7GPOTeXsiBSy9IA==' 'nonce-Av7aa2wd7e2J1bC0etU3Jg=='; style-src * 'unsafe-inline'
```
Those two nonce sources correspond to `https://fast.appcues.com/49767.js` and the main JS bundle generated by webpack. Those files should be covered by `https://fast.appcues.com` and `'self'` in the `script-src` directive, so I don't think the nonce sources were necessary.

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -24,6 +24,7 @@ handlers:
   expiration: 0s
   secure: always
   http_headers:
+    Content-Security-Policy: "base-uri 'self'; object-src 'none'; script-src 'self' https://fast.appcues.com https://us.jsagent.tcell.insight.rapid7.com https://cdnjs.cloudflare.com; style-src * 'unsafe-inline'"
     X-Frame-Options: "SAMEORIGIN"
     Strict-Transport-Security: "max-age=31536000; includeSubdomains; preload"
     X-Content-Type-Options: "nosniff"


sarahgibs (approved):
Looks good. Thanks.

app.yaml
```diff
@@ -24,6 +24,7 @@ handlers:
   expiration: 0s
   secure: always
   http_headers:
+    Content-Security-Policy: "base-uri 'self'; object-src 'none'; script-src 'self' https://fast.appcues.com https://us.jsagent.tcell.insight.rapid7.com https://cdnjs.cloudflare.com; style-src * 'unsafe-inline'"
```
fboulnois, line 27: Do we need to include any of the other services in here, like TDR or Rawls?

sarahgibs: Only if JS or CSS is loaded from them.

rtitle:
I didn't test, but make sure that Leo/notebooks functionality works correctly with the CSP change. The UI sets a cookie from the Leo domain (notebooks.firecloud.org) and renders the content in an iframe.

nawatts:
In the PR deployment, that endpoint is returning 401. I can't see the response because Leo doesn't set a CORS header in that case. This happens in other PR deployments as well.
```
Access to fetch at 'https://leonardo.dsde-dev.broadinstitute.org/proxy/setCookie' from origin 'https://pr-8-dot-bvdp-saturn-dev.appspot.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

That endpoint is loaded using `fetch`, so I think it should be fine.
https://github.com/DataBiosphere/terra-ui/blob/5de7c1eaf024bbc9981ef1842c5ffd35ba51a3bb/src/libs/ajax.js#L1518-L1520

rtitle:
> In the PR deployment, that endpoint is returning 401. I can't see the response because Leo doesn't set a CORS header in that case. This happens in other PR deployments as well.

This should be fixed in dev now as of https://github.com/broadinstitute/firecloud-develop/pull/3042. Leo `setCookie` worked for me on a PR site. 😁

nawatts:
I checked that I was able to open a Jupyter notebook and run a cell in the PR deployment.

breilly2 (approved):

PR #3277, opened 2022-07-29
[IA-3607] Update styling for new cost display in context bar.

n8landolt:
https://broadworkbench.atlassian.net/browse/IA-3607

## Description
There has been some feedback on the new cost display feature. Make some visual tweaks according to the UI team.

## Screenshots

Tool tip Line Height

Old (No line height specified)
Screen Shot 2022-08-05 at 9 22 03 AM
Screen Shot 2022-08-05 at 9 30 19 AM

New 1.5 line height
Screen Shot 2022-08-05 at 9 29 33 AM
Screen Shot 2022-08-05 at 9 30 26 AM


Qi77Qi (approved):

src/components/ContextBar.js
```diff
@@ -145,15 +145,20 @@ export const ContextBar = ({
...
         h(Fragment, [
           h(TooltipTrigger,
             {
               side: 'left',
               delay: 100,
               content: [
-                div({ key: 'p1' }, 'This rate reflects the estimated aggregate hourly cost for running and paused applications, as well as associated persistent disks. For more details, click on the Cloud icon.'),
+                div({ key: 'p1' },
+                  ['Estimated hourly rate for all applications in a running or paused state, and associated persistent disks. For details, click',
+                    img({ src: cloudIcon, style: { height: 20, paddingLeft: 5, paddingRight: 5, verticalAlign: 'text-bottom' }, alt: 'Environment Configuration' }),
```
dmohs: Concision is beneficial here, I think:
```suggestion
                    img({ src: cloudIcon, style: { height: 20, padding: '0 5px', verticalAlign: 'text-bottom' }, alt: 'Environment Configuration' }),
```
While you're here, I think "Environment Configuration Icon" would be better alt text, since this isn't the actual control but an example of it.

src/components/ContextBar.js
```diff
@@ -145,15 +145,20 @@ export const ContextBar = ({
     }),
     div({ style: { ...Style.elements.contextBarContainer, width: 70 } }, [
       div({ style: contextBarStyles.contextBarContainer }, [
+
```
dmohs: Intentional? Benefit is not clear to me.
```suggestion
```

src/components/ContextBar.js
```diff
@@ -163,17 +168,18 @@ export const ContextBar = ({
                 hover: { ...contextBarStyles.hover }
               },
               [
-                div({ style: { textAlign: 'center', color: colors.dark(), fontSize: '0.8em' } }, 'Rate'),
```
dmohs, line 166: `0.x` is the convention in the codebase. Changing the convention requires justification, via PR comment or otherwise.

src/components/TooltipTrigger.js
```diff
@@ -21,6 +21,7 @@ const styles = {
     borderColor: colors.secondary(),
     padding: '1rem',
     boxShadow: '0 1px 3px 2px rgba(0,0,0,0.3)',
+    lineHeight: '1.5em',
```
dmohs: I tried and failed to see the impact of this on the PR site. It might be worth an extra screenshot or two since this component is used so widely.

dmohs (approved):

src/components/ContextBar.js
n8landolt: I see your point - made adjustments.

src/components/TooltipTrigger.js
n8landolt: Added screenshots - we COULD apply the line height just to the tooltip I added (We need the space for the cloud env image in the new tooltip)
However - 1.5 line height is a bit more readable, and consistency is nice in my opinion.

Should we have a conversation with someone about it?

cahrens: I would suggesting pinging UX to see if they are OK with changing this in all of our tooltips.

PR #3278, opened 2022-07-29
[DC-363] Move to passing errors to ignore as param

s-rubenstein:
<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -55,7 +55,7 @@ const withCancellation = wrappedFetch => async (...args) => {
   }
 }
 
-const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
+const withRetryOnError = _.curry((errorsToIgnore, wrappedFetch) => async (...args) => {
   const timeout = 5000
   const somePointInTheFuture = Date.now() + timeout
   const maxDelayIncrement = 1500
@@ -66,11 +66,7 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
     try {
       return await Utils.withDelay(until, wrappedFetch)(...args)
     } catch (error) {
-      // requesterPaysError may be set on responses from requests to the GCS API that are wrapped in withRequesterPays.
-      // requesterPaysError is true if the request requires a user project for billing the request to. Such errors
-      // are not transient and the request should not be retried.
-
-      const shouldNotRetry = Boolean(error.requesterPaysError)
+      const shouldNotRetry = _.intersection(_.keys(error), errorsToIgnore).length > 0
       if (shouldNotRetry) {
         throw error
       }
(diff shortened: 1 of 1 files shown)


src/libs/ajax.js
```diff
@@ -66,11 +66,7 @@ const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
     try {
       return await Utils.withDelay(until, wrappedFetch)(...args)
     } catch (error) {
-      // requesterPaysError may be set on responses from requests to the GCS API that are wrapped in withRequesterPays.
-      // requesterPaysError is true if the request requires a user project for billing the request to. Such errors
-      // are not transient and the request should not be retried.
-
-      const shouldNotRetry = Boolean(error.requesterPaysError)
+      const shouldNotRetry = _.intersection(_.keys(error), errorsToIgnore).length > 0
```
pshapiro4broad: It's a minor point since the collections are so small, but is there a lodash way to do a filter find first? Since you don't need a full intersection, you only need to know if at least one element of X is in Y.

pshapiro4broad (approved):

src/libs/ajax.js
```diff
@@ -55,7 +55,7 @@ const withCancellation = wrappedFetch => async (...args) => {
   }
 }
 
-const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
+const withRetryOnError = _.curry((errorsToIgnore, wrappedFetch) => async (...args) => {
```
nawatts: A more flexible way to do this might be to have `withRetryOnError` accept a function instead of a list of properties.

```js
const withRetryOnError = _.curry((shouldRetry, wrappedFetch) ...
...
if (!shouldRetry(error)) {
  throw error
}
```

src/libs/ajax.js
s-rubenstein: I think this is no longer needed with @nawatts' approach.

nawatts (approved):

cahrens (approved):

src/libs/ajax.js
kyuksel: > It's a minor point since the collections are so small, but is there a lodash way to do a filter find first? Since you don't need a full intersection, you only need to know if at least one element of X is in Y.

Good thinking. I know it's not needed anymore but for future reference Lodash's [findIndex](https://lodash.com/docs/4.17.15#findIndex) or [find](https://lodash.com/docs/4.17.15#find) could be used for this purpose.

petesantos (approved):

fboulnois (approved):

src/libs/ajax.js
```diff
@@ -55,7 +55,7 @@ const withCancellation = wrappedFetch => async (...args) => {
   }
 }
 
-const withRetryOnError = _.curry(wrappedFetch => async (...args) => {
+const withRetryOnError = _.curry((shouldNotRetryFn, wrappedFetch) => async (...args) => {
```
pshapiro4broad, line 58: Another minor comment, in general I prefer positive over negative names, since if you ever have to negate them in an expression it makes it harder (for me) to figure out what's going on. So maybe `shouldThrowFn` or `shouldRetryFn` and negate the sense in the caller.

PR #3279, opened 2022-07-29
Ignore

n8landolt:
Please ignore

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -58,7 +58,7 @@ const DataTable = props => {
     childrenBefore,
     editable,
     activeCrossTableTextFilter,
-    persist, refreshKey, firstRender,
+    persist, refreshKey,
     snapshotName,
     deleteColumnUpdateMetadata,
     controlPanelStyle,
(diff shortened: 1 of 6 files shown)


Qi77Qi:
does this mean whatever the first location in avalalble location list will populate the field?

src/components/ComputeModal.js
```diff
@@ -1163,7 +1163,7 @@ export const ComputeModalBase = ({
                 isSearchable: false,
                 value: computeConfig.computeRegion,
                 onChange: ({ value, locationType }) => updateComputeLocation(value, locationType),
-                options: _.flow(_.filter(l => l.value !== defaultLocation), _.sortBy('label'))(getAvailableComputeRegions(location))
+                options: _.sortBy('label')(getAvailableComputeRegions(location))
```
kyuksel: We don't need to use the curried nature of `sortBy` here since both inputs (`'label'` and `getAvailableComputeRegions(location)`) are available at the same time (as opposed to that they weren't in the previous case). It doesn't matter that much but FWIW my brain starts wondering if it's missing something when it sees the curried call 😊 

```suggestion
                options: _.sortBy('label', getAvailableComputeRegions(location))
```

kyuksel (approved):
- The code change LGTM with a minor comment inline. 
- Didn't test. The team reviewer should thoroughly test given it's easy to have a lot of corner cases when regionality is involved.

src/components/ComputeModal.js
n8landolt: > does this mean whatever the first location in avalalble location list will populate the field?

What do we want the outcome to be? 
It seems like the _.sortBy under the options doesn't really do anything, and if anything could be moved to region-common getAvailableComputeRegions.

kyuksel: @wnojopra in case he has comments on the business logic

wnojopra: First comment: This change actually fixes a bug introduced by the change of defaultLocation from multi-regional to us-central1. When you're in a workspace who's bucket is in Montreal, your options for Cloud Environment location should be Montreal AND us-central. It seems on prod currently, us-central1 is missing. I never noticed this bug until today. So thanks for doing this. 

Second comment: The filter and sort stuff was here because getAvailableComputeRegions didn't use to exist. Instead we used to just pass it all compute regions. 

Ultimately, we want the list of options to be the available compute regions, sorted alphabetically. When the modal is first loaded, we want the default selection to be the same region as their workspace bucket (which is set correctly on line 1164 `value: computeConfig.computeRegion,`). 

Overall this change should be good.

PR #3280, opened 2022-07-29
[SU-151] Display recently viewed workspaces above workspaces list

MatthewBemis:
This adds a row above the workspaces list that displays the 4 most recently viewed workspaces. If there are no recently viewed workspaces or the user does not have any workspaces, this section is hidden.

Screen Shot 2022-08-02 at 4 57 39 PM

See [SU-151](https://broadworkbench.atlassian.net/browse/SU-151) for the mockup from Ean. There is one notable stylistic difference here. The mocks display a gray background color and then a flat white card for each of the recently viewed workspaces. We are not ready to switch to the gray background yet, so this uses the standard card style seen in other places in Terra UI to improve contrast

<!--
Hello, friend!
Thanks!

--!>


MatthewBemis:
Seems like one of the tests is failing. Can't repro it locally yet, but I think this is ready for a review. Seems to just be a mishandling of an undefined value somewhere

src/components/workspace-utils.js
```diff
@@ -340,6 +346,55 @@ export const WorkspaceTagSelect = props => {
   })
 }
 
+export const WorkspaceSubmissionStatusIcon = ({ status, loadingSubmissionStats, size = 20 }) => {
+  return div({}, [
```
nawatts: Nit: Since the spinner and status icon should be mutually exclusive, what do you think about something like:

```js
return Utils.cond(
  [loadingSubmissionStats, () => ...],
  [status, () => ...],
  () => null
)
```

That would get rid of the unnecessary div and clarify the expected states.

src/pages/workspaces/List.js
```diff
@@ -87,6 +87,13 @@ export const WorkspaceList = () => {
   const { workspaces, refresh: refreshWorkspaces, loadingWorkspaces, loadingSubmissionStats } = useWorkspacesWithSubmissionStats()
   const [featuredList, setFeaturedList] = useState()
 
+  const persistenceId = 'workspaces/recentlyViewed'
+  const [recentlyViewed] = useState(() => getLocalPref(persistenceId)?.recentlyViewed || [])
```
nawatts: Since the component doesn't need to react to changes in this variable (because it is never updated), it doesn't need to be stored in component state. Storing it in state does avoid re-reading from local storage on every render though. Alternatively, that could be done with [`useMemo`](https://reactjs.org/docs/hooks-reference.html#usememo).

```js
const recentlyViewed = useMemo(() => getLocalPref(persistenceId)?.recentlyViewed || [], [])
```

src/pages/workspaces/List.js
```diff
@@ -87,6 +87,13 @@ export const WorkspaceList = () => {
...
+
+  useEffect(() => {
+    setLocalPref(persistenceId, { recentlyViewed })
+  }, [recentlyViewed]) // eslint-disable-line react-hooks/exhaustive-deps
```
nawatts: Since `recentlyViewed` doesn't change, this `useEffect` hooks is unnecessary.

src/pages/workspaces/List.js
nawatts: Also, should `recentlyViewed` be filtered based on the `workspaces` list? As is, if you view a workspace, then lose access to it, would it still show up in the recently viewed list?

src/pages/workspaces/List.js
```diff
@@ -380,6 +375,18 @@ export const WorkspaceList = () => {
           })
         ])
       ]),
+      !_.isEmpty(workspaces) && !_.isEmpty(recentlyViewed) && div({}, [
```
nawatts: FYI, if the props object is empty, it can be omitted (`div([...])` vs `div({}, [...])`).

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -263,6 +266,10 @@ const WorkspaceDashboard = _.flow(
     }
   }, [sasTokenRefreshInterval])
 
+  useEffect(() => {
+    setLocalPref(recentlyViewedPersistenceId, { recentlyViewed })
+  }, [recentlyViewed]) // eslint-disable-line react-hooks/exhaustive-deps
```
nawatts: For this, I don't think it's necessary have a state variable and an effect hook based on that state variable. `getLocalPref` and `setLocalPref` could be called directly by `updateRecentlyViewedWorkspaces`.

src/pages/workspaces/workspace/Dashboard.js
```diff
@@ -420,6 +427,18 @@ const WorkspaceDashboard = _.flow(
     ]
   }
 
+  const updateRecentlyViewedWorkspaces = () => {
```
nawatts: I think it would be better to put this in `WorkspaceContainer` or `wrapWorkspace` in WorkspaceContainer.js. That way, recently viewed workspaces would be updated if the user navigates directly to some other workspace page like Data or Notebooks instead of the dashboard.

nawatts (changes requested):
Two main things:
- The recently viewed workspaces list should be filtered to available workspaces.
- Recently viewed workspaces should be updated when viewing any workspace page, not just the dashboard.

cahrens:
FYI, I'm unable to view the Workspaces list on the PR site because I encounter this error:

image

(I'm not trying to review… just wanted to try it out and see what the "cards" at the top are)

src/components/workspace-utils.js
```diff
@@ -340,6 +346,55 @@ export const WorkspaceTagSelect = props => {
...
+  return h(Clickable, {
+    style: { ...Style.elements.card.container, margin: '0 0.25rem 0 0.25rem', lineHeight: '22px', flex: '0 1 calc(25% - 10px)' },
+    href: Nav.getLink('workspace-dashboard', { namespace, name })
+  }, [
+    div({ style: { flex: 'none' } }, [
+      div({ style: { color: colors.accent(), ...Style.noWrapEllipsis, fontSize: 16, marginBottom: 7 } }, name),
+      div({ style: { display: 'flex', justifyContent: 'space-between' } }, [
+        div({ style: { ...Style.noWrapEllipsis, whiteSpace: 'pre-wrap', fontStyle: 'italic' } }, `Viewed ${dateViewed}`),
+        div({ style: { display: 'flex', alignItems: 'center' } }, [
+          h(WorkspaceSubmissionStatusIcon, {
```
cahrens, line 400: This WorkflowSubmissionStatusIcon is not accessible-- there seems to be no way to get the screen reader to announce anything about it.

cahrens: I also called this out in one of our tickets (https://broadworkbench.atlassian.net/browse/WOR-149?focusedCommentId=64541), so if you didn't actually change `WorkflowSubmissionStatusIcon`, that fix could be considered out of scope for this PR.

cahrens:
> FYI, I'm unable to view the Workspaces list on the PR site because I encounter this error:

I'm actually seeing this error on dev too, so not related to your PR. Trying to figure out what happened…

src/pages/workspaces/List.js
MatthewBemis: Done in 29ee1789c580e63d03be97caf276efcdcd79da28

src/components/workspace-utils.js
MatthewBemis: Done in ce8677729ce72aee7cb9281982da90ad8892f7a7

src/pages/workspaces/List.js
MatthewBemis: Ah totally- leftover from an earlier implementation. Removed in 86cf6a6b5410dd575a2d88c9cf2259ee2cb7ec42

cahrens:
@MatthewBemis What is the status of this PR? We are going to be doing some work on this page soon (including moving the submission status icon), and it would be good to get this PR merged first if possible.

The integration test failure looks like something you probably should look into.

MatthewBemis:
@cahrens yes the test is on my list to look into. Hoping to get this back to reviewers today, been bogged down by some other work. I'll tag you directly when it's ready for a look.

src/components/workspace-utils.js
MatthewBemis: This was just extraction, so I will leave it be for this PR

src/pages/workspaces/List.js
MatthewBemis: Using useMemo in e149f68df9f4fbe0452a282d3b0aea2ce2e02356

MatthewBemis: Also handled filtering based on `workspaces` in 27d0204e7507fcd427586025d2cff6623636a0bb, which fixed the broken test too

MatthewBemis:
@nawatts @cahrens this should be ready for review now. Please take another look- thanks!

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -325,6 +329,18 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
       }
     }
 
+    const updateRecentlyViewedWorkspaces = workspaceId => {
```
nawatts: The method for updating the recently viewed list seems kind of round about. It looks like this function could directly call get/setLocalPref and that would remove the need for a state variable and effect hook.

src/pages/workspaces/List.js
```diff
@@ -380,6 +371,18 @@ export const WorkspaceList = () => {
...
+      !_.isEmpty(workspaces) && !_.isEmpty(recentlyViewed) && div([
+        p({ style: { textTransform: 'uppercase' } }, 'Recently viewed'),
+        div({ style: { display: 'flex', flexWrap: 'wrap', paddingBottom: '1rem' } },
+          _.map(({ workspaceId, timestamp }) => {
+            const workspace = getWorkspace(workspaceId)
+            return h(RecentlyViewedWorkspaceCard, {
+              workspace, loadingSubmissionStats, timestamp,
+              submissionStatus: workspaceSubmissionStatus(workspace)
+            })
+          }, _.filter(w => _.find({ workspace: { workspaceId: w.workspaceId } }, workspaces), recentlyViewed))
```
nawatts: Should this filter be done before checking `.isEmpty(...)`? It looks like if you've lost access to all recently viewed workspaces, this would display "Recently viewed" and an empty list.

src/components/workspace-utils.js
```diff
@@ -340,6 +346,55 @@ export const WorkspaceTagSelect = props => {
...
+    () => null)
+}
+
+export const RecentlyViewedWorkspaceCard = ({ workspace, submissionStatus, loadingSubmissionStats, timestamp }) => {
+  const { workspace: { namespace, name, googleProject } } = workspace
+
+  const dateViewed = Utils.makeCompleteDate(new Date(parseInt(timestamp)).toString())
+
+  return h(Clickable, {
+    style: { ...Style.elements.card.container, margin: '0 0.25rem 0 0.25rem', lineHeight: '22px', flex: '0 1 calc(25% - 10px)' },
```
nawatts: Does not need to be changed, but FYI many CSS properties have shorthand forms for when multiple values are the same. Margin here could also be set as `margin: '0 0.25rem'`.
https://developer.mozilla.org/en-US/docs/Web/CSS/margin#syntax

nawatts: I think line height should be specified in `rem` so that it scales with text size.

More info on absolute vs relative units in CSS here: https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#numbers_lengths_and_percentages

nawatts (changes requested):
Main requesting changes to the update flow in WorkspaceContainer.

src/components/workspace-utils.js
MatthewBemis: margin css shorthanded in d63db7d21ea173ce492eeb165e8cb5eefcb368ec, switched lineHeight to 1.5rem in https://github.com/DataBiosphere/terra-ui/pull/3280/commits/ac62229ded3443c6ea61a12e32011d9cca11ab15

src/pages/workspaces/workspace/WorkspaceContainer.js
MatthewBemis: Simplified in 51762ff1f058a223a8413898cf6ec22239f9f950

src/pages/workspaces/List.js
MatthewBemis: For sure- 70e611a3d607d9dae30fd40b52f22033840273f3

cahrens:
It feels like the recent workspace cards should go above the workspace list filter options because the filter options do not impact the recent workspaces. For example, I had 2 recent Azure workspaces and 2 recent Google ones. When I search for "Azure", the Google ones disappear from the workspace list, but the 2 Google workspaces remain in the recent workspaces cards. That seemed unexpected because the filter controls are above the recent workspace cards.

src/components/workspace-utils.js
```diff
@@ -340,6 +346,55 @@ export const WorkspaceTagSelect = props => {
...
+      div({ style: { display: 'flex', justifyContent: 'space-between' } }, [
+        div({ style: { ...Style.noWrapEllipsis, whiteSpace: 'pre-wrap', fontStyle: 'italic' } }, `Viewed ${dateViewed}`),
+        div({ style: { display: 'flex', alignItems: 'center' } }, [
+          h(WorkspaceSubmissionStatusIcon, {
+            status: submissionStatus,
+            loadingSubmissionStats,
+            size: 16
+          }),
+          !!googleProject ?
+            h(GcpLogo, { title: 'Google Cloud', role: 'img', style: { marginLeft: 5, height: 16 } }) :
```
cahrens, line 406: I think we may want more compact icons-- @yunip20 is also looking at this for the ticket she's working on.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -325,6 +326,20 @@ export const wrapWorkspace = ({ breadcrumbs, activeTab, title, topBarContent, sh
...
+      const recentlyViewedPersistenceId = 'workspaces/recentlyViewed'
```
cahrens: This ID is also defined in `workspace-utils`. It would be good to avoid that duplication, and I think this code might make more sense in that file anyway.

MatthewBemis:
@cahrens that's a good point re: the placement of the cards. This is how Ean had it in his mockup (see the ticket linked in the description), but I'll check again with UX. 

I went ahead and made the change now because I have a feeling we'll roll with it: bf4ba35948390e521f93ba14a1af112696939a78

cahrens:
> @cahrens that's a good point re: the placement of the cards. This is how Ean had it in his mockup (see the ticket linked in the description), but I'll check again with UX.
> 
> I went ahead and made the change now because I have a feeling we'll roll with it: [bf4ba35](https://github.com/DataBiosphere/terra-ui/commit/bf4ba35948390e521f93ba14a1af112696939a78)

Thanks, I like that better-- less confusing.

nawatts (approved):
+1 to @cahrens's comments about icons and workspace-utils.

src/components/workspace-utils.js
cahrens: @yunip20 has some smaller icons in https://github.com/DataBiosphere/terra-ui/pull/3303/files

MatthewBemis: @cahrens @yunip20 Nice! I'll wait for that PR to merge and then rebase this branch to grab those new icons

src/pages/workspaces/workspace/WorkspaceContainer.js
MatthewBemis: Moved everything to workspace-utils in ba5c1d9244b8dce9af6b7c57889cad888a153d51

cahrens (approved):

src/components/workspace-utils.js
cahrens: @MatthewBemis There are other things in Yunyi's PR that are going to create merge issues (specifically, the workflows submission icon). I think it would be easier if you go ahead and merge this one, and we will handle the updates/merge conflicts.

MatthewBemis: Works for me- thanks!

PR #3282, opened 2022-08-01
[SU-161] Allow adding tracks to IGV

nawatts:
Currently, to view files using IGV in Terra, the file URLs must be in a data table and only files from a single data table can be viewed together. This adds an option to add arbitrary GCS files as tracks in the IGV browser. This can be useful for things like reference data.

This required refactoring how IGVBrowser handles requester pays files. Currently, IGVBrowser assumes that all files will be specified when the component is created. It uses [`requesterPaysWrapper` from components/bucket-utils](https://github.com/DataBiosphere/terra-ui/blob/6d8a86c1b583eb5672aa2121d31a41064c610d08/src/components/bucket-utils.js#L23-L41) which, if a requester pays error is encountered, will unmount IGVBrowser, prompt for a workspace to bill to, and re-mount IGVBrowser. This method is not viable if we want to preserve the state of tracks in IGV while prompting for a workspace to bill requester pays requests to. Thus, [3bd984e](https://github.com/DataBiosphere/terra-ui/pull/3282/commits/3bd984e331a30c66b228e743e5262e86f8e80742) adds an `addTracks` function to IGVBrowser that can handle requester pays files and makes IGVBrowser handle showing the RequesterPaysModal.

If testing in a read only workspace with a requester pays bucket, ToolDrawer will prompt for a billing workspace before IGVBrowser is shown. To get to IGVBrowser's requester pays error handling, comment out [`loadNotebookNames()` in ToolDrawer](https://github.com/DataBiosphere/terra-ui/blob/6d8a86c1b583eb5672aa2121d31a41064c610d08/src/components/data/EntitiesContent.js#L79).

Also, currently [IGVBrowser calls `getUserProjectForWorkspace` once for each file viewed](https://github.com/DataBiosphere/terra-ui/blob/6d8a86c1b583eb5672aa2121d31a41064c610d08/src/components/IGVBrowser.js#L61). When viewing files in a requester pays bucket in a read-only workspace, this results in a call to the list billing projects API for each file. This changes IGVBrowser to call `getUserProjectForWorkspace` only once per set of tracks added.

https://user-images.githubusercontent.com/1156625/182165689-29dbf5b6-a41e-44e2-8100-b754a27c441d.mov

### DIFF
--- a/src/components/IGVAddTrackModal.js
+++ b/src/components/IGVAddTrackModal.js
@@ -23,7 +23,7 @@ const IGVAddTrackModal = ({ onDismiss, onSubmitTrack }) => {
     () => null
   )
 
-  const isTrackValid = !urlError
+  const isTrackValid = _.isNull(urlError)
 
   const submit = () => {
     const track = {
--- a/src/components/IGVAddTrackModal.js
+++ b/src/components/IGVAddTrackModal.js
@@ -1,0 +1,104 @@
+import _ from 'lodash/fp'
+import { Fragment, useState } from 'react'
+import { form, h, label } from 'react-hyperscript-helpers'
+import { ButtonPrimary, IdContainer } from 'src/components/common'
+import { parseGsUri } from 'src/components/data/data-utils'
+import { TextInput, ValidatedInput } from 'src/components/input'
+import Modal from 'src/components/Modal'
+import * as Utils from 'src/libs/utils'
+
+
+const IGVAddTrackModal = ({ onDismiss, onSubmitTrack }) => {
+  const [value, setValue] = useState({
+    name: '',
+    url: '',
+    indexUrl: ''
+  })
+
+  const [urlInputTouched, setUrlInputTouched] = useState(false)
+
+  const urlError = Utils.cond(
+    [!value.url, () => 'A URL is required'],
+    [_.isEmpty(parseGsUri(value.url)), () => 'A gs:// URL is required'],
+    () => null
+  )
+
+  const isTrackValid = !urlError
+
+  const submit = () => {
+    const track = {
+      name: value.name || value.url,
+      url: value.url
+    }
+    if (value.indexUrl) {
+      track.indexURL = value.indexUrl
+    } else {
+      track.indexed = false
+    }
+
+    onSubmitTrack(track)
+  }
+
+  return h(Modal, {
+    title: 'Add track',
+    okButton: h(ButtonPrimary, {
+      disabled: !isTrackValid,
+      onClick: submit
+    }, ['Add track']),
+    onDismiss
+  }, [
+    form({
+      onSubmit: e => {
+        e.preventDefault()
+        submit()
+      }
+    }, [
+      h(IdContainer, [id => h(Fragment, [
+        label({
+          htmlFor: id,
+          style: { display: 'block', margin: '0.5rem 0 0.25rem' }
+        }, ['Name (optional)']),
+        h(TextInput, {
+          id,
+          placeholder: 'My track',
+          value: value.name,
+          onChange: name => setValue(_.set('name', name))
+        })
+      ])]),
+
+      h(IdContainer, [id => h(Fragment, [
+        label({
+          htmlFor: id,
+          style: { display: 'block', margin: '0.5rem 0 0.25rem' }
+        }, ['URL (required)']),
+        h(ValidatedInput, {
+          inputProps: {
+            id,
+            placeholder: 'gs://my-bucket/file.bam',
+            value: value.url,
+            onChange: url => {
+              setValue(_.set('url', url))
+              setUrlInputTouched(true)
+            }
+          },
+          error: urlInputTouched && urlError
+        })
+      ])]),
+
+      h(IdContainer, [id => h(Fragment, [
+        label({
+          htmlFor: id,
+          style: { display: 'block', margin: '0.5rem 0 0.25rem' }
+        }, ['Index URL (optional)']),
+        h(TextInput, {
+          id,
+          placeholder: 'gs://my-bucket/file.bai',
+          value: value.indexUrl,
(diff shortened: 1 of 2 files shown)


src/components/IGVAddTrackModal.js
```diff
@@ -0,0 +1,104 @@
...
+
+  const [urlInputTouched, setUrlInputTouched] = useState(false)
+
+  const urlError = Utils.cond(
+    [!value.url, () => 'A URL is required'],
+    [_.isEmpty(parseGsUri(value.url)), () => 'A gs:// URL is required'],
+    () => null
+  )
+
+  const isTrackValid = !urlError
```
kyuksel: [Potentially a nit] Would

```suggestion
  const isTrackValid = !_.isNull(urlError)
```

be clearer/safer? Probably not a big deal in this case but when I see JS coercion of a non-boolean value into boolean with assumptions (e.g. it's `null` when there's an error, in this case), I wonder if that value could be modified to be something else (e.g. `0`) down the road, falsefying the assumption without noticing its implications.

src/components/IGVBrowser.js
```diff
@@ -1,80 +1,115 @@
...
+    )(tracks)
+
+    // Learn the requester pays status of each bucket.
+    // Requesting a file will store its requester pays status in knownBucketRequesterPaysStatuses.
+    const isRequesterPays = await Promise.all(_.map(async url => {
+      const [bucket, file] = parseGsUri(url)
+
+      if (knownBucketRequesterPaysStatuses.get()[bucket] === undefined) {
+        try {
+          await Ajax(signal).Buckets.getObject(workspace.workspace.googleProject, bucket, file, { fields: 'kind' })
```
kyuksel, line 44: Trying to understand: Looks like we make this API call, only to see if it'll fail?

MatthewBemis (approved):

src/components/IGVBrowser.js
```diff
@@ -1,80 +1,115 @@
...
-  requesterPaysWrapper({ onDismiss: ({ onDismiss }) => onDismiss() })
-)(({ selectedFiles, refGenome: { genome, reference }, workspace, onDismiss, onRequesterPaysError }) => {
+const IGVBrowser = ({ selectedFiles, refGenome: { genome, reference }, workspace, onDismiss }) => {
   const containerRef = useRef()
   const signal = useCancellation()
   const [loadingIgv, setLoadingIgv] = useState(true)
   const igvLibrary = useRef()
+  const igvBrowser = useRef()
+  const [requesterPaysModal, setRequesterPaysModal] = useState(null)
+  const [showAddTrackModal, setShowAddTrackModal] = useState(false)
```
kyuksel, line 21: Our convention for clustering component-level `const`s has been to put `useState`s together on the top, which I find helpful because often I glance at state variables first when trying to understand a component.

src/components/IGVBrowser.js
```diff
@@ -1,80 +1,115 @@
...
-                indexURL: !!indexFilePath ? Utils.mergeQueryParams(userProjectParam, indexFilePath) : undefined
+    // If any bucket is requester pays, files in that bucket will need to have a user project included in the request.
+    let userProject
+    if (_.some(_.identity, isRequesterPays)) {
+      // Check if the user can bill to the current workspace.
+      userProject = await getUserProjectForWorkspace(workspace)
+
+      // If not, prompt to select a workspace to bill to.
+      if (!userProject) {
+        userProject = await new Promise((resolve, reject) => {
```
kyuksel, line 62: I couldn't tell why `userProject` had to be a non-`const` since it doesn't seem to be used besides in `if (!userProject)`. Could it be

```suggestion
    if (_.some(_.identity, isRequesterPays)) {
      // Check if the user can bill to the current workspace.
      const userProject = await getUserProjectForWorkspace(workspace)

      // If not, prompt to select a workspace to bill to.
      if (!userProject) {
        await new Promise((resolve, reject) => {
```
?

src/components/IGVBrowser.js
nawatts: Sorted hooks in https://github.com/DataBiosphere/terra-ui/pull/3282/commits/1ca56d704aaf9adf87844dd9da5649eda2494e62. I think there's also an argument to be made for grouping based on concern vs type of hook.

src/components/IGVBrowser.js
nawatts: Correct. We need to find out if any of the buckets are requester pays. The only way to do that is to attempt a request and see if it fails with a specific error.

src/components/IGVBrowser.js
nawatts: `userProject` is used to add a query parameter to requests for files in requester pays buckets.
```js
 _.forEach(({ name, url, indexURL }) => {
  const [bucket] = parseGsUri(url)
  const userProjectParam = { userProject: knownBucketRequesterPaysStatuses.get()[bucket] ? userProject : undefined }
  ...
```

If we can't automatically determine a user project using `getUserProjectForWorkspace`, we have to prompt the user to select one.

src/components/IGVAddTrackModal.js
nawatts: Changed in https://github.com/DataBiosphere/terra-ui/pull/3282/commits/c2b065d1e873c55682e8d821768028e960a6e4f3.

src/components/IGVBrowser.js
kyuksel: Sure. Perhaps a good topic to discuss as part of coding guidelines.

src/components/IGVBrowser.js
kyuksel: That's unfortunate. Thanks.

src/components/IGVBrowser.js
kyuksel: Ah okay. I'd missed that somehow.

kyuksel (approved):

PR #3283, opened 2022-08-01
[SUP-646] Add screen reader announcement to loading spinner

nawatts:
From [SUP-646](https://broadworkbench.atlassian.net/browse/SUP-646):
> Across Terra, we use a loading spinner when we are making a server call to let the user know that an action is in progress. The spinner does not have any accessible text attached to it, so a screen reader does not have anything to announce.
>
> It would be helpful if the spinner announced that it was loading content. The proposal would be to add an aria role of alert with text that is not visible to sighted users. For an example, see https://codeburst.io/how-to-create-a-simple-css-loading-spinner-make-it-accessible-e5c83c2e464c 
>
> It might be nice if the text is not added unless the operation has been in process for a certain length of time, as often the loads are quite fast.

This adds a visually hidden [alert](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role) to the spinner icon. DelayedRender is used to make an announcement only if the spinner is shown for more than 150ms.

To use DelayedRender here, I had to move it to its own module to avoid a circular import issue (DelayedRender was in components/common, components/common imports spinner from icons and creates a few spinner elements that are exported from the module).


MatthewBemis (approved):

src/components/icons.js
```diff
@@ -47,7 +48,10 @@ export const icon = (shape, { size = 16, ...props } = {}) => {
   return _.invokeArgs(shape, [{ size, 'data-icon': shape, ...props }], iconDict)
 }
 
-export const spinner = props => icon('loadingSpinner', _.merge({ size: 24, style: { color: colors.primary() } }, props))
+export const spinner = ({ message = 'Loading', ...props }) => h(Fragment, [
+  icon('loadingSpinner', _.merge({ size: 24, style: { color: colors.primary() } }, props)),
+  h(DelayedRender, { delay: 150 }, [div({ className: 'sr-only', role: 'alert' }, [message])])
```
cahrens, line 53: @msilva-broad This type of approach might work better on the spend report instead of the aria-live regions.

cahrens:
This behaves very nicely-- thanks @nawatts !

cahrens (approved):

PR #3286, opened 2022-08-02
[SU-185] Allow muting notifications for expired/expiring fence links

nawatts:
In order to gain access to some datasets, users can link other types of accounts (NHLBI BioData Catalyst Framework Services, NCI CRDC Framework Services, NHGRI AnVIL Data Commons Framework Services, Kids First DRC Framework Services) to their Terra account using the Bond service. Those links expire after some time (usually 30 days). When they are expiring soon (within the next 5 days) or have expired, Terra shows a notification.

Users have complained about the same notification popping up every time they open Terra and have asked for a way to permanently dismiss it. This adds a "Do not remind me again" button to these notifications which prevents showing the notification again (until the account is re-linked).

Screen Shot 2022-08-02 at 2 23 21 PM

This is similar to what #3068 did for linked NIH accounts. While #3068's approach was specific to NIH link notifications, this adds a more generic option to mute notifications. `notify` now checks if a local preference `mute-notification/${notificationId}` is set. If that preference contains a timestamp later than the current time or a -1 (for mute forever), the notification will not be shown. This also updates the NIH link notifications to use this new method.

This also fixes a bug. Currently, the notification code assumes that all fence links expire after 30 days. However, NCI links expire after only 15 days. Expiration times for different account types are specified in [libs/providers](https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/providers.js).

## Testing

NIH and fence link statuses are fetched when a user signs in. To manually test this, set AJAX overrides before signing into Terra.

AnVIL
```js
window.ajaxOverridesStore.set([
  {
    filter: { url: /\/api\/link\/v1\/anvil$/ },
    fn: window.ajaxOverrideUtils.makeSuccess({
      issued_at: (new Date(Date.now() - 90 * 24 * 3600 * 1000)).toISOString(),
      username: 'user@example.com'
    })
  }
])
```

NIH
```js
window.ajaxOverridesStore.set([
  {
    filter: { url: /\/api\/nih\/status$/ },
    fn: window.ajaxOverrideUtils.makeSuccess({
      linkedNihUsername: 'user@example.com',
      datasetPermissions: [],
      linkExpireTime: Math.floor(Date.now() / 1000) - 24 * 3600
    })
  }
])
```

### DIFF
--- a/src/libs/auth.js
+++ b/src/libs/auth.js
@@ -11,7 +11,7 @@ import { getConfig } from 'src/libs/config'
 import { withErrorIgnoring, withErrorReporting } from 'src/libs/error'
 import { captureAppcuesEvent } from 'src/libs/events'
 import * as Nav from 'src/libs/nav'
-import { clearNotification, notify, sessionTimeoutProps } from 'src/libs/notifications'
+import { clearMatchingNotifications, clearNotification, notify, sessionTimeoutProps } from 'src/libs/notifications'
 import { getLocalPref, getLocalPrefForUserId, setLocalPref } from 'src/libs/prefs'
 import allProviders from 'src/libs/providers'
 import {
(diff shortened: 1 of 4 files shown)


MatthewBemis (approved):
Nice!

src/libs/auth.js
```diff
@@ -358,47 +358,55 @@ authStore.subscribe(withErrorIgnoring(async (state, oldState) => {
...
+  const dateOfExpiration = status && new Date(status.linkExpireTime * 1000)
+  const shouldNotify = Boolean(dateOfExpiration) && now >= addDays(-1, dateOfExpiration)
+  if (shouldNotify) {
+    const hasExpired = now >= dateOfExpiration
+
+    // Previously, muting notifications for an expiring NIH account link would store the expiration time of
+    // the link for which notifications were muted. Now, these notifications are muted using muteNotification
+    // from libs/notifications. This is kept to avoid showing notifications that were muted using the old method.
+    // Once enough time has passed for NIH account links to expire and users to re-link them, this can be removed
+    // with less disruption.
```
dmohs, line 376: Great comment.

src/libs/auth.js
```diff
@@ -411,32 +419,57 @@ authStore.subscribe(withErrorReporting('Error loading Framework Services account
...
+
+  if (shouldNotify) {
+    const hasExpired = now >= dateOfExpiration
+
+    // Separate notifications for expired and expiring. If a user mutes the notification that a link will
+    // expire soon, then we still want to notify them once the link has actually expired.
+    // The link expiration time is included in the notification ID so that a muting notifications only
+    // applies to the current instance of the link. If an account is re-linked, notifications should
+    // be shown again when that link is expiring.
+    const notificationId = `fence-link-expiration/${key}/${dateOfExpiration.getTime()}/${hasExpired ? 'expired' : 'expiring'}`
```
dmohs, line 441: Consider extracting this line into a function, which will provide a single location for the comment.

dmohs (approved):

src/libs/auth.js
nawatts: I tried this and found it kind of awkward. The notification IDs generated here need to match the prefix passed to `clearMatchingNotifications`, so it felt like those prefixes should also be extracted. And I'd either have to pass both `dateOfExpiration` and `hasExpired` (which is derived from `dateOfExpiration`) to the function or recalculate `hasExpired`.

dmohs: Hmm... I think you could manage the key issue, but yeah about the others:
```javascript
const getNotificationIdPath = (dateOfExpiration, hasExpired) =>
  // comment about expired vs. expiring
  `${dateOfExpiration.getTime()}/${hasExpired ? 'expired' : 'expiring'}`
```
Usage:
```javascript
const notificationId = `fence-link-expiration/${key}/${getNotificationIdPath(dateOfExpiration, hasExpired)}`
```
I agree. Not an obvious win.

PR #3287, opened 2022-08-02
[QA-1984] UI Test results Slack notification from CircleCI

aweng98:
[QA-1984](https://broadworkbench.atlassian.net/browse/QA-1984)

CircleCI workflow example: [terra-ui/11131/workflows/2cad15b5-9051-4064-8ab0-0c93928649fc](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/11131/workflows/2cad15b5-9051-4064-8ab0-0c93928649fc)

Google [doc](https://docs.google.com/document/d/1w3akIJJjOX2BUbYsa0upo9Z1gDX3sf745kW05P9VQNY/edit?usp=sharing).

An [example](https://broadinstitute.slack.com/archives/C01AG7GEJ3D/p1659624574078209) of failed test notification to `cromwell_jenkins_ci_errors" channel.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,7 +273,7 @@ jobs:
       - integration-tests:
           env: staging
   notify-test-results:
-    executor: puppeteer
+    executor: node
     parameters:
       job_name:
         type: enum
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -262,72 +262,18 @@ jobs:
           env: local
   integration-tests-alpha:
     executor: puppeteer
-    parallelism: 4
+    parallelism: 1 # No way to prevent multiple Slack messages otherwise
     steps:
       - integration-tests:
           env: alpha
+      - notify-qa
   integration-tests-staging:
     executor: puppeteer
-    parallelism: 4
+    parallelism: 1 # No way to prevent multiple Slack messages otherwise
     steps:
       - integration-tests:
           env: staging
-  notify-test-results:
-    executor: node
-    parameters:
-      job_name:
-        type: enum
-        enum: [ integration-tests-alpha, integration-tests-staging, integration-tests-branch-against-dev ]
-        default: integration-tests-alpha
-    steps:
-      - checkout
-      - when:
-          condition: << pipeline.parameters.start_notify_job >>
-          steps:
-            - attach_workspace:
-                at: .
-            - restore_cache:
-                keys:
-                  - deps-{{ .Branch }}-{{ checksum ".pnp.cjs" }}
-            - run:
-                name: Wait up to 15 minutes for << parameters.job_name >> to finish
-                command: |
-                  x=0
-                  while [[ $x -le 900 ]]
-                    do
-                      status=$(curl --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job" --header "Circle-Token: $CIRCLECI_TOKEN" | jq -r '.items[]|select(.name == "<< parameters.job_name >>")|.status')
-                      echo "Job status: ${status}"
-                      if [[ $status == "success" ]] || [[ $status == "failed" ]]; then
-                        break
-                      elif [[ $status == "canceled" ]]; then
-                        circleci-agent step halt
-                      fi
-
-                      sleep 10
-                      x=$(( $x + 10 ))
-                    done
-
-                  echo "Waited $x seconds"
-                  date
-
-                  # Is job stuck?
-                  if [[ x -ge 900 ]]; then
-                    status=$(curl --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job" --header "Circle-Token: $CIRCLECI_TOKEN" | jq -r '.items[]|select(.name == "<< parameters.job_name >>")|.status')
-                    echo "Final job status: ${status}"
-                    exit 1
-                  fi
-            - run:
-                name: Set JOB_BUILD_NUM from << parameters.job_name >>
-                command: |
-                  build_num=$(curl --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job" --header "Circle-Token: $CIRCLECI_TOKEN" | jq -r '.items[]|select(.name == "<< parameters.job_name >>")|.job_number')
-                  echo "export JOB_BUILD_NUM=$build_num" >> "$BASH_ENV"
-                  echo "export JOB_NAME=<< parameters.job_name >>" >> "$BASH_ENV"
-                  source $BASH_ENV
-            - run:
-                name: Slack test results
-                working_directory: integration-tests
-                command: |
-                  node --require ../.pnp.cjs ./slack/slack-circleci-results.js
+      - notify-qa
 
 workflows:
   version: 2
@@ -366,19 +312,6 @@ workflows:
           <<: *filter-dev-branch
           requires:
             - deploy-staging
-      - notify-test-results:
-          name: notify-test-results-alpha
-          <<: *filter-dev-branch
-          requires:
-            - deploy-alpha
-          job_name: integration-tests-alpha
-      - notify-test-results:
-          name: notify-test-results-staging
-          <<: *filter-dev-branch
-          requires:
-            - deploy-staging
-          job_name: integration-tests-staging
-
   nightly-integration-tests:
     triggers:
       - schedule:
@@ -389,9 +322,3 @@ workflows:
     jobs:
       - integration-tests-alpha
       - integration-tests-staging
-      - notify-test-results:
-          name: notify-test-results-alpha
-          job_name: integration-tests-alpha
-      - notify-test-results:
-          name: notify-test-results-staging
-          job_name: integration-tests-staging
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -262,18 +262,69 @@ jobs:
           env: local
   integration-tests-alpha:
     executor: puppeteer
-    parallelism: 1 # No way to prevent multiple Slack messages otherwise
+    parallelism: 4
     steps:
       - integration-tests:
           env: alpha
-      - notify-qa
   integration-tests-staging:
     executor: puppeteer
-    parallelism: 1 # No way to prevent multiple Slack messages otherwise
+    parallelism: 4
     steps:
       - integration-tests:
           env: staging
-      - notify-qa
+  notify-test-results:
+    executor: puppeteer
+    parameters:
+      job_name:
+        type: enum
+        enum: [ integration-tests-alpha, integration-tests-staging ]
+        default: integration-tests-alpha
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - restore_cache:
+          keys:
+            - deps-{{ .Branch }}-{{ checksum ".pnp.cjs" }}
+      - run:
+          name: Wait up to 15 minutes for << parameters.job_name >> to finish
+          command: |
+            x=0
+            while [[ $x -le 900 ]]
+              do
+                status=$(curl --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job" --header "Circle-Token: $CIRCLECI_TOKEN" | jq -r '.items[]|select(.name == "<< parameters.job_name >>")|.status')
+                echo "Job status: ${status}"
+                if [[ $status == "success" ]] || [[ $status == "failed" ]]; then
+                  break
+                elif [[ $status == "canceled" ]]; then
+                  circleci-agent step halt
+                fi
+
+                sleep 10
+                x=$(( $x + 10 ))
+              done
+
+            echo "Waited $x seconds"
+            date
+
+            # Is job stuck?
+            if [[ x -ge 900 ]]; then
+              status=$(curl --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job" --header "Circle-Token: $CIRCLECI_TOKEN" | jq -r '.items[]|select(.name == "<< parameters.job_name >>")|.status')
+              echo "Final job status: ${status}"
+              exit 1
+            fi
+      - run:
+          name: Set JOB_BUILD_NUM from << parameters.job_name >>
+          command: |
+            build_num=$(curl --request GET "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job" --header "Circle-Token: $CIRCLECI_TOKEN" | jq -r '.items[]|select(.name == "<< parameters.job_name >>")|.job_number')
+            echo "export JOB_BUILD_NUM=$build_num" >> "$BASH_ENV"
+            echo "export JOB_NAME=<< parameters.job_name >>" >> "$BASH_ENV"
+            source $BASH_ENV
+      - run:
+          name: Slack test results
+          working_directory: integration-tests
+          command: |
+            node --require ../.pnp.cjs ./slack/slack-circleci-results.js
 
 workflows:
   version: 2
@@ -312,6 +363,19 @@ workflows:
           <<: *filter-dev-branch
           requires:
             - deploy-staging
+      - notify-test-results:
+          name: notify-test-results-alpha
+          <<: *filter-dev-branch
+          requires:
+            - deploy-alpha
+          job_name: integration-tests-alpha
+      - notify-test-results:
+          name: notify-test-results-staging
+          <<: *filter-dev-branch
+          requires:
+            - deploy-staging
+          job_name: integration-tests-staging
+
   nightly-integration-tests:
     triggers:
       - schedule:
@@ -322,3 +386,9 @@ workflows:
     jobs:
       - integration-tests-alpha
       - integration-tests-staging
+      - notify-test-results:
+          name: notify-test-results-alpha
+          job_name: integration-tests-alpha
+      - notify-test-results:
+          name: notify-test-results-staging
+          job_name: integration-tests-staging
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,11 +28,6 @@ anchors:
       branches:
         only: dev
 
-parameters:
-  start_notify_job:
-    type: boolean
-    default: false
-
 commands:
   notify-github: # borrows from https://circleci.com/developer/orbs/orb/mmnfst/github-deployments#orb-source
     parameters:
@@ -186,17 +181,6 @@ commands:
       - store_artifacts:
           path: << parameters.log_dir >>/<< parameters.env >>/test-results
           destination: results
-      - run:
-          name: Trigger notify-test-results job
-          command: |
-            if [[ $CIRCLE_NODE_INDEX -eq 0 ]]; then
-              curl -u ${CIRCLECI_TOKEN}: -X POST \
-                --header 'content-type: application/json' \
-                --url https://circleci.com/api/v2/project/github/DataBiosphere/terra-ui/pipeline \
-                --data '{"parameters":{"start_notify_job":true}}'
-            fi
-          when: always
-
   notify-qa:
     parameters:
       channel:
@@ -354,12 +338,6 @@ workflows:
           <<: *filter-pr-branch
           requires:
             - build
-      - notify-test-results:
-          name: notify-test-results-pr
-          <<: *filter-pr-branch
-          requires:
-            - build
-          job_name: integration-tests-branch-against-dev
       - deploy-pr:
           <<: *filter-pr-branch
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -187,12 +187,13 @@ commands:
           path: << parameters.log_dir >>/<< parameters.env >>/test-results
           destination: results
       - run:
+          name: Trigger notify-test-results job
           command: |
             if [[ $CIRCLE_NODE_INDEX -eq 0 ]]; then
-              curl --request POST \
-                --url https://circleci.com/api/v2/project/gh/DataBiosphere/terra-ui/pipeline \
+              curl -u ${CIRCLECI_TOKEN}: -X POST \
                 --header 'content-type: application/json' \
-                --data '{"branch":"dev","parameters":{"start_notify_job":true}}'
+                --url https://circleci.com/api/v2/project/github/DataBiosphere/terra-ui/pipeline \
+                --data '{"parameters":{"start_notify_job":true}}'
             fi
           when: always
 
@@ -295,15 +296,15 @@ jobs:
         enum: [ integration-tests-alpha, integration-tests-staging, integration-tests-branch-against-dev ]
         default: integration-tests-alpha
     steps:
-      - checkout
-      - attach_workspace:
-          at: .
-      - restore_cache:
-          keys:
-            - deps-{{ .Branch }}-{{ checksum ".pnp.cjs" }}
       - when:
           condition: << pipeline.parameters.start_notify_job >>
           steps:
+            - checkout
+            - attach_workspace:
+                at: .
+            - restore_cache:
+                keys:
+                  - deps-{{ .Branch }}-{{ checksum ".pnp.cjs" }}
             - run:
                 name: Wait up to 15 minutes for << parameters.job_name >> to finish
                 command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,11 @@ anchors:
       branches:
         only: dev
 
+parameters:
+  start_notify_job:
+    type: boolean
+    default: false
+
 commands:
   notify-github: # borrows from https://circleci.com/developer/orbs/orb/mmnfst/github-deployments#orb-source
     parameters:
@@ -181,6 +186,16 @@ commands:
       - store_artifacts:
           path: << parameters.log_dir >>/<< parameters.env >>/test-results
           destination: results
+      - run:
+          command: |
+            if [[ $CIRCLE_NODE_INDEX -eq 0 ]]; then
+              curl --request POST \
+                --url https://circleci.com/api/v2/project/gh/DataBiosphere/terra-ui/pipeline \
+                --header 'content-type: application/json' \
+                --data '{"branch":"dev","parameters":{"start_notify_job":true}}'
+            fi
+          when: always
+
   notify-qa:
     parameters:
       channel:
@@ -277,7 +292,7 @@ jobs:
     parameters:
       job_name:
         type: enum
-        enum: [ integration-tests-alpha, integration-tests-staging ]
+        enum: [ integration-tests-alpha, integration-tests-staging, integration-tests-branch-against-dev ]
         default: integration-tests-alpha
     steps:
       - checkout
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -262,18 +262,69 @@ jobs:
...
-      - notify-qa
   integration-tests-staging:
     executor: puppeteer
-    parallelism: 1 # No way to prevent multiple Slack messages otherwise
+    parallelism: 4
     steps:
       - integration-tests:
           env: staging
-      - notify-qa
+  notify-test-results:
```
aweng98: `notify-test-results`: Waits for test job (e.g. integration-tests-staging) to finish by polling job status. Finally, calls `./slack/slack-circleci-results.js` to collect all tests summaries, find any failed tests, then send Slack notifications.

integration-tests/slack/slack-notify-channels.json
```diff
@@ -29,8 +29,8 @@
       ]
     },
     {
-      "name": "dsp-batch",
-      "id": "C1EH66VCM",
+      "name": "cromwell_jenkins_ci_errors",
+      "id": "C01AG7GEJ3D",
```
aweng98: Unrelated: This change was requested by Adam Nicholes from batch team.

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,139 @@
...
+
+  return idsMap
+}
+
+/**
+ * Notify job failed. Message contains failed test names.
+ * @param failedTests
+ * @returns {Promise<void>}
+ */
+// eslint-disable-next-line
```
aweng98: eslint warning. I don't know how to resolve it.
```
104:41  warning  Async arrow function has no 'await' expression  require-await
```

integration-tests/jest-reporter.js
```diff
@@ -57,7 +57,7 @@ module.exports = class JestReporter {
   // Called after all tests have completed
   onRunComplete(_test, runResults) {
     // Save run summary to a file.
-    const summaryFile = `tests-summary.json`
+    const summaryFile = `tests-summary${process.env.CIRCLE_NODE_INDEX ? `-${process.env.CIRCLE_NODE_INDEX}` : ''}.json`
```
aweng98: Make test summary file name unique on CircleCI.

.circleci/config.yml
```diff
@@ -312,6 +363,19 @@ workflows:
           <<: *filter-dev-branch
           requires:
             - deploy-staging
+      - notify-test-results:
+          name: notify-test-results-alpha
+          <<: *filter-dev-branch
+          requires:
+            - deploy-alpha
+          job_name: integration-tests-alpha
```
aweng98: `job_name` in workflow:
The name key can be used to invoke reusable jobs across any number of workflows. Using the name key ensures numbers are not appended to your job name (i.e. sayhello-1 , sayhello-2, etc.). The name you assign to the name key needs to be unique, otherwise the numbers will still be appended to the job name.

.circleci/config.yml
```diff
@@ -262,18 +262,69 @@ jobs:
...
   integration-tests-staging:
     executor: puppeteer
-    parallelism: 1 # No way to prevent multiple Slack messages otherwise
+    parallelism: 4
     steps:
       - integration-tests:
           env: staging
-      - notify-qa
+  notify-test-results:
+    executor: puppeteer
```
nawatts: Nit: does this job need Puppeteer? Can it use the node executor?

integration-tests/slack/message-templates.js
```diff
@@ -0,0 +1,73 @@
+const {
+  JOB_NAME: CIRCLE_JOB,
+  JOB_BUILD_NUM: CIRCLE_BUILD_NUM,
+  CIRCLE_PROJECT_REPONAME,
+  CIRCLE_SHA1,
+  CIRCLE_USERNAME
+} = process.env
+
+const printTimestamp = () => {
```
nawatts: To me, the `print...` names for these function suggest that they are outputting with console.log or something similar. Maybe consider something like `getTimestampSlackBlock`? Same comment for other functions in this file.

integration-tests/slack/message-templates.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+}
+
+const printJobDetail = () => {
+  return {
+    type: 'section',
+    fields: [
+      {
+        type: 'mrkdwn',
+        text: `*Github Project*: <https://github.com/DataBiosphere/terra-ui/commit/${CIRCLE_SHA1} | ${CIRCLE_PROJECT_REPONAME}>`
```
nawatts: I find some of these labels misleading. As is, in Slack, it looks like this shows up as "GitHub Project: terra-ui".
1. That makes me think it links to https://github.com/DataBiosphere/terra-ui.
2. It seems like redundant information since the title already includes "Terra-UI Test".

Maybe label this "Revision" or "Commit" and show the short hash as the link text?

integration-tests/slack/message-templates.js
```diff
@@ -0,0 +1,73 @@
...
+      },
+      {
+        type: 'mrkdwn',
+        text: `*User*: ${CIRCLE_USERNAME}`
```
nawatts: Is this useful? The idea of including the username makes sense. But in many cases, GitHub usernames are unrelated to real names or Slack usernames, which I think limits the value of including this.

integration-tests/slack/message-templates.js
nawatts: Also, is including this timestamp block in the message useful? How often would this be different from the message timestamp in Slack.

Screen Shot 2022-08-04 at 4 53 18 PM

integration-tests/slack/message-templates.js
```diff
@@ -0,0 +1,73 @@
...
+      },
+      {
+        type: 'mrkdwn',
+        text: `*Job*: <https://circleci.com/gh/DataBiosphere/terra-ui/${CIRCLE_BUILD_NUM} | ${CIRCLE_JOB}>`
+      },
+      {
+        type: 'mrkdwn',
+        text: `*Job Build Number*: ${CIRCLE_BUILD_NUM}`
```
nawatts: To trim down the amount of text in notifications, maybe this could be part of the "Job" block / link text?

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+  JOB_BUILD_NUM: circleJobBuildNum
+} = process.env
+
+/**
+ * Fetch CircleCI artifact links to tests summary JSON files (were created in onRunComplete() in jest-reporter.js)
+ * @param { string } token
+ * @param { string } buildNum
+ * @returns { Promise<Array[string]> } URL to tests-summary-[0-9].json
+ */
+const fetchCircleJobArtifacts = async ({ token = defaultToken, buildNum = circleJobBuildNum } = {}) => {
```
nawatts: `token` is used for the CircleCI API, but `defaultToken` is taken from an environment variable called `SLACK_BOT_TOKEN`, which sounds like it's for the Slack API.

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+  if (!buildNum) {
+    throw new Error(`**  ERROR: Missing CircleCI build number. Failed to fetch CircleCI job artifacts.`)
+  } else if (!token) {
+    throw new Error(`**  ERROR: Missing CircleCI API token. Failed to fetch CircleCI job artifacts.`)
+  }
+
+  // Find more arguments and details of the response: https://circleci.com/docs/api/v2/index.html#operation/getJobArtifacts
+  const apiUrl = 'https://circleci.com/api/v2/project/github/DataBiosphere/terra-ui'
+  const headers = { Accept: 'application/json', Authorization: `Bearer ${token}` }
```
nawatts: Does this API call need to be authenticated? [find-build.js in saturn-ui-prod-deploy](https://github.com/DataBiosphere/saturn-ui-prod-deploy/blob/dev/src/find-build.js) does a similar API call without authentication.

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+
+  try {
+    const response = await fetch(`${apiUrl}/${buildNum}/artifacts`, { method: 'GET', headers })
+    const json = await response.json()
+    const { items } = json
+    const filtered = _.filter(item => item['path'].toString().includes('tests-summary'), items)
+    return filtered.map(item => item.url)
```
nawatts: Maybe consider something like:
```suggestion
    const { items } = await response.json()
    const testSummaryArtifacts = _.filter(_.flow(_.get('path'), _.includes('tests-summary')), items)
    return _.map(_.get('url'), testSummaryArtifacts)
```

- Doesn't mix Lodash filter and native map (https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide#coding-style)
- More descriptive variable name than "filtered"
- `item.path` syntax is more concise than `item['path']`
- `item.path` is a string and doesn't need to be converted (https://circleci.com/docs/api/v2/index.html#operation/getJobArtifacts)

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+}
+
+const getFailedTestFileName = aggregatedResult => {
+  const failedTests = []
+  _.forEach(testResult => {
+    if (testResult.numFailingTests > 0) {
+      failedTests.push(parse(testResult.testFilePath).name)
+    }
+  }, aggregatedResult.testResults)
+  return failedTests
```
nawatts: This can be written with `filter` and `map` instead of appending to an array.

```suggestion
  return _.flow(
    _.filter(testResult => testResult.numFailingTests > 0)
    _.map(testResult => parse(testResult.testFilePath).name)
  )(aggregatedResult.testResults)
```

https://github.com/DataBiosphere/terra-ui/wiki/Contributor-Guide#coding-style

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+    const { items } = json
+    const filtered = _.filter(item => item['path'].toString().includes('tests-summary'), items)
+    return filtered.map(item => item.url)
+  } catch (e) {
+    console.error(`**  ERROR: Encountered unexpected error when getting CircleCI build_number: ${buildNum} artifacts:`, e)
+    throw e
+  }
+}
+
+const getFailedTestFileName = aggregatedResult => {
```
nawatts: The name of this function suggests it returns one file name, but it actually returns an array.

```suggestion
const getFailedTestFileNames = aggregatedResult => {
```

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+    if (testResult.numFailingTests > 0) {
+      failedTests.push(parse(testResult.testFilePath).name)
+    }
+  }, aggregatedResult.testResults)
+  return failedTests
+}
+
+/**
+ *
+ * @returns {Promise<Array>}
```
nawatts: This annotation looks incomplete. What is the type of array elements?

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+  } else if (!token) {
+    throw new Error(`**  ERROR: Missing CircleCI API token. Failed to fetch CircleCI job artifacts.`)
+  }
+
+  // Find more arguments and details of the response: https://circleci.com/docs/api/v2/index.html#operation/getJobArtifacts
+  const apiUrl = 'https://circleci.com/api/v2/project/github/DataBiosphere/terra-ui'
+  const headers = { Accept: 'application/json', Authorization: `Bearer ${token}` }
+
+  try {
+    const response = await fetch(`${apiUrl}/${buildNum}/artifacts`, { method: 'GET', headers })
```
nawatts: To make this more concise, `method: 'GET'` could be omitted since that's the default. Based on [find-build.js in saturn-ui-prod-deploy](https://github.com/DataBiosphere/saturn-ui-prod-deploy/blob/dev/src/find-build.js), the `Accept: 'application/json` header could also omitted.

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+  const headers = { Accept: 'application/json' }
+  const urls = await fetchCircleJobArtifacts()
+  const tests = []
+  for (const url of urls) {
+    try {
+      // Parse all tests-summary JSON for failed tests
+      const response = await fetch(url, { method: 'GET', headers })
+      const json = await response.json()
+      const failedTest = getFailedTestFileName(json)
+      tests.push(...failedTest)
```
nawatts: Nit: Naming makes this confusing. Because of the singular variable name, it wasn't obvious `failedTest` is an array. When I first read this, I assumed `failedTest` is an object and thus spreading it when calling `tests.push` didn't make sense.

```suggestion
      const failedTests = getFailedTestFileName(json)
      tests.push(...failedTests)
```

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+  const { fail: jsonItems } = data
+
+  _.forEach(item => {
+    const id = _.get(['id'], item)
+    const { tests } = item
+    const names = _.map(item => item.name, tests)
+    channels.set(id, names)
+  }, jsonItems)
+
+  // Construct a map object: key: channel_id, value: test_names array
```
nawatts: This comment would be more helpful in the `returns` section of the function's JSDoc.

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+  const channels = new Map()
+  const data = JSON.parse(fs.readFileSync('./slack/slack-notify-channels.json', 'utf8'))
+  const { fail: jsonItems } = data
+
+  _.forEach(item => {
+    const id = _.get(['id'], item)
+    const { tests } = item
+    const names = _.map(item => item.name, tests)
+    channels.set(id, names)
+  }, jsonItems)
```
nawatts: This is very hard to read. The variable names could be more descriptive (`data`, `jsonItems`, `item`, `names` are all very generic) and style could be more consistent (in three lines, there are three different ways of reading an object property: `_.get`, destructuring, and dot notation).

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+
+  // Construct a map object: key: channel_id, value: test_names array
+  const idsMap = new Map()
+  _.forEach(test => {
+    const filtered = Array.from(channels.keys())
+      .map(key => channels.get(key).includes(test) || channels.get(key).length === 0 ? key : undefined)
+      .filter(key => key)
```
nawatts:
```suggestion
    const filtered = Array.from(channels.keys())
      .filter(key => channels.get(key).includes(test) || channels.get(key).length === 0)
```

nawatts: Also, similar comment as above on variable names. `filtered` and `key` are very generic and don't help understand what this is doing.

integration-tests/slack/slack-utils.js
nawatts: To resolve this, remove the `async` keyword.

```js
const notifyFailure = failedTests => {
```

https://eslint.org/docs/latest/rules/require-await

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+ * Slack to notify job failed. Message contains failed test names.
+ * @param { Array[string] } failedTests
+ * @returns { Promise<void> }
+ */
+// eslint-disable-next-line
+const notifyFailure = async failedTests => {
+  const data = getChannelsNotifyFailed(failedTests)
+  // Slack issue: No way to post the same message to multiple channels at once. https://github.com/slackapi/bolt-js/issues/696
+  _.forEach(async ([k, v]) => {
+    // k is channel_id. v is test_name array.
```
nawatts: Better than a comment would be to name those variables `channelId` and `testNames` instead of `k` and `v`.

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+    const blocks = printMessageTemplate(1)
```
nawatts: Is `1` left over from development? Is this function supposed to be passed `failedTests.length`?

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+    blocks.push({
+      type: 'section',
+      fields:
+        [
+          {
+            type: 'mrkdwn',
+            text: `:errrrr:  Tests:\n\`\`\`*  ${v.join('\n*  ')}\`\`\``
```
nawatts: To make the notification less noisy, maybe consider omitting the emoji and "Tests" header and using a Markdown list instead of a code block?

integration-tests/slack/slack-utils.js
```diff
@@ -0,0 +1,148 @@
...
+  }, _.toPairs(data))
+}
+
+// Slack to notify job succeeded
+const notifySuccess = () => {
+  const blocks = printMessageTemplate()
+  const data = JSON.parse(fs.readFileSync('./slack/slack-notify-channels.json', 'utf8'))
+  const { pass: passNotifyChannels } = data
+  const channels = []
+  _.forEach(key => channels.push(_.get(['id'], key)), passNotifyChannels)
```
nawatts:
```suggestion
  const channels = _.map(_.get('id'), passNotifyChannels)
```

integration-tests/slack/slack-utils.js
nawatts: Oh, based on [post-message.js](https://github.com/DataBiosphere/terra-ui/blob/781fcd5259a80341ae8bf16f89d164f13dd89b2d/integration-tests/slack/post-message.js), `SLACK_BOT_TOKEN` _is_ for the Slack API. This token should probably be rotated since it looks like it's been sent to CircleCI.

integration-tests/slack/slack-utils.js
nawatts: If authentication is needed, CircleCI uses a `Circle-Token` header or HTTP basic auth, not an `Authorization: 'Bearer ...'` header.

https://circleci.com/docs/api/v2/index.html

nawatts (changes requested):
As I understand it, this starts a job (notify-test-results) that runs in parallel with integration-tests-staging/alpha. notify-test-results occasionally checks the CircleCI API to see if integration-tests-staging/alpha has finished.

Since build minutes are how CircleCI bills, I'm not keen on doubling our usage for integration-tests-staging/alpha jobs. I think there are a few alternatives that would not involve a job that polls for another job to finish.

1. When each parallel job of integration-tests-staging/alpha finishes, it could check if the overall job has finished and, if so, send the Slack notification.
2. Make the integration-tests-staging/alpha job always pass and use [requires](https://circleci.com/docs/configuration-reference#requires) to run another job afterward. That job checks tests results, sends the Slack notification, and fails if any tests failed.

.circleci/config.yml
aweng98: No, doesn't need Puppeteer. I'll change to node executor.

integration-tests/slack/message-templates.js
aweng98: Removed. Personally, I think dd/mm/yy to is helpful. But I don't feel strongly.

integration-tests/slack/message-templates.js
aweng98: Done. Renamed to `Commit` and slice `CIRCLE_SHA1` to show first 7 char.

integration-tests/slack/message-templates.js
aweng98: Agreed. It's blank for the nightly-test runs anyway. I'll remove.

integration-tests/slack/message-templates.js
aweng98: Done.

integration-tests/slack/slack-utils.js
aweng98: Done. I didn't know.

integration-tests/slack/slack-utils.js
aweng98: It seems CircleCI API request did not need a token to work. I wonder if it's because we're on the Free Plan.

integration-tests/slack/slack-utils.js
aweng98: Removed token.

integration-tests/slack/slack-utils.js
aweng98: Awesome! 👍🏻

integration-tests/slack/slack-utils.js
aweng98: Done.

integration-tests/slack/slack-utils.js
aweng98: Added `[string]`. pretty much same for all array/map.

integration-tests/slack/slack-utils.js
aweng98: 👍🏻

integration-tests/slack/slack-utils.js
aweng98: But `await postMessage` is called inside the `_.forEach`. I'll try remove it to see if still works.

integration-tests/slack/slack-utils.js
aweng98: 👍🏻

integration-tests/slack/slack-utils.js
aweng98: `failedTests.length` is right.

integration-tests/slack/slack-utils.js
aweng98: 👍🏻
There is no Markdown list in Slack. What I have is what Slack recommends to do to make a list. I'll remove header and emoij.

integration-tests/slack/slack-utils.js
aweng98: 👍🏻

integration-tests/slack/slack-utils.js
aweng98: Updated JSDoc.

aweng98 (commented):
So many good feedback, Thank you Nick. Sorry about the size of the PR this time. I don't normally make such big PR.

aweng98:
> As I understand it, this starts a job (notify-test-results) that runs in parallel with integration-tests-staging/alpha. notify-test-results occasionally checks the CircleCI API to see if integration-tests-staging/alpha has finished.
> 
> Since build minutes are how CircleCI bills, I'm not keen on doubling our usage for integration-tests-staging/alpha jobs. I think there are a few alternatives that would not involve a job that polls for another job to finish.
> 
> 1. When each parallel job of integration-tests-staging/alpha finishes, it could check if the overall job has finished and, if so, send the Slack notification.
> 2. Make the integration-tests-staging/alpha job always pass and use [requires](https://circleci.com/docs/configuration-reference#requires) to run another job afterward. That job checks tests results, sends the Slack notification, and fails if any tests failed.

1 and 2 are good suggestions. Either requires some good amount of time for code refactoring and testing. I'll start looking into this. In the mean time, I've addressed the majority of feedback.

integration-tests/slack/slack-utils.js
aweng98: Thanks.
I tried to wrtie FP functions as best as I can but I'm not an advanced JS programmer. I focused on making it work, I have not iterated towards better code styles. I'll try do better next time.
Just my own opinion, such guide is not really useful to me because it doesn't actually show how to guide someone like me to write better FP code. Real code examples will be helpful if they exist.

integration-tests/slack/slack-utils.js
aweng98: Done. Oh boy, I really thought I had this FP right by myself.
Renamed `filtered` to `channelIdsForFailedTest` and `key` to `channelId`.

integration-tests/slack/slack-utils.js
nawatts: Yes, but this function is not awaiting the functions called by `_.forEach`.

integration-tests/slack/slack-utils.js
nawatts: I think it's because terra-ui is a public repository on GitHub.
https://circleci.com/docs/oss#security

aweng98:
@nawatts Hi, I have a question. In order to make the change that you suggested (avoid running monitoring job in parallel), I would need more time for changes and testing (more commits). I don't know how long it would take and I don't know if I can make it work. So just to check, is that change absolutely required? If yes, I'd like to put PR back to draft. If not, maybe I can do the change in a follow-up PR after merge this PR first.

integration-tests/slack/slack-utils.js
nawatts: For added context, having this in the JSDoc instead of the function body allows a reader to see the function's interface from the JSDoc without having to get into the implementation. Some editors also surface information from JSDoc.

integration-tests/slack/slack-utils.js
nawatts: It wasn't _wrong_, but it was unnecessarily complex. Since this is getting a subset of item from the original collection, it can be done with `filter`. If it was transforming collection items, then `map` would be needed.

integration-tests/slack/slack-utils.js
nawatts: There should be a way to render this as something like:
Screen Shot 2022-08-05 at 1 16 41 PM

Instead of:
Screen Shot 2022-08-05 at 1 17 09 PM

I think removing the backticks for the code block formatting would do it.
https://api.slack.com/reference/surfaces/formatting#block-formatting

integration-tests/slack/message-templates.js
```diff
@@ -0,0 +1,50 @@
...
+} = process.env
+
+const getHeaderBlock = numFailedTests => {
+  if (numFailedTests === 0) {
+    return {
+      type: 'header',
+      text: {
+        type: 'plain_text',
+        text: `:circleci-pass:  Terra-UI Test has Passed on CircleCI`,
+        emoji: true
```
nawatts: This block for successful tests has `emoji: true`. The block for failed tests does not. In Slack, it looks like emoji were rendered for both of them. So is this necessary?

integration-tests/slack/slack-utils.js
nawatts: I linked to that mainly for the "Use Lodash/fp for data transformation" guideline.

Leaving out `_.flow` and Lodash entirely, this could also have been written as:
```js
return aggregatedResult.testResults
  .filter(testResult => testResult.numFailingTests > 0)
  .map(testResult => parse(testResult.testFilePath).name)
```

Generally, I think the functional style and immutable data helps reason about what code is doing by more clearly showing how data is being transformed. As opposed to mutable state (such as the `failedTests` array in the previous version of this code) where you have to step through how it's changing over time to understand how the output relates to the input.

integration-tests/slack/message-templates.js
aweng98: Just tested it. It's optional. I'll remove it.

integration-tests/slack/slack-utils.js
aweng98: Done. Originally, I wanted to make more contrast.

aweng98 (commented):
Hi @nawatts,
I thought one PR is the only way to do this because changes are related. But now I think I should break up this PR into few smaller PRs, hoping to make reviews easier. Let me know if you agree or disagree.

- `slack-notify-channels.json` (unrelated) change is one PR. 
- `jest-reporter.js`: one PR.
- `message-templates.js`: alone in one PR. I can show message look-like in screenshots.
- `post-message.js`, `slack-circleci-results.js` and `slack-utils.js`: together in one PR.
- `config.yml`: in final PR.

PR #3288, opened 2022-08-02
[IA-3279] Release Analysis Tab - Remove legacy code and beta toggle

jdcanas:
This deletes a lot of code!

A lot less maintenance for the IA team, as we release the analysis tab feature 🎉


cahrens (approved):
Very exciting-- deleting code is the best! I wasn't able to test the PR because state related to runtimes/apps/disks updates so unreliably on dev. But I read through the code changes.

integration-tests/tests/workspace-dashboard.js
```diff
@@ -1,6 +1,6 @@
 // This test is owned by the Workspaces Team.
 const _ = require('lodash/fp')
-const { overrideConfig, viewWorkspaceDashboard, withWorkspace } = require('../utils/integration-helpers')
```
cahrens, line 3: @jdcanas please also delete `overrideConfig`, as both usages are removed. The method as-is has a problem because it does not wait for the spinner (@aweng98 was actually in the process of investigating this). I don't want us to leave an unused method sitting around that could potentially cause a problem for someone else in the future.

cahrens: Ping again on ^

jdcanas: Yeah, this is on hold until to official release date and bug bash, so focusing on that currently :). Fixing this now though...

Qi77Qi (approved):

PR #3291, opened 2022-08-04
[SU-181] Persist service alert dismissal when x button is clicked

nawatts:
Currently, clicking the "Dismiss" button in a service alert notification hides the notification for 24 hours. Clicking the X button, which is also labeled as "Dismiss notification", does not. If a service alert notification is dismissed using the X button, it reappears when the page is reloaded or Terra is opened in a new tab. Some users have complained about the inconsistency. This change makes the X button behave the same as the "Dismiss" button.

## Before
https://user-images.githubusercontent.com/1156625/182876569-30a3c306-b380-48e4-8311-07afdb7db177.mov

## After
https://user-images.githubusercontent.com/1156625/182876600-a45352f3-97f9-49d7-ad22-d290fe5ae41d.mov

## Testing

Since service alerts are fetched when Terra is first loaded, ajaxOverrides can't be used to mock the response (the request is made before there's a chance to set the ajaxOverrides). This can be tested locally by changing the `FirecloudBucket.getServiceAlerts` function in libs/ajax.js.

```js
getServiceAlerts: async () => {
  return Promise.resolve([{
    title: 'Service alert',
    message: 'The systems are down!',
    link: 'https://support.terra.bio'
  }])
}
```

### DIFF
--- a/src/components/ServiceAlerts.js
+++ b/src/components/ServiceAlerts.js
@@ -32,6 +32,10 @@ export const ServiceAlerts = () => {
       return !dismissal || differenceInDays(parseJSON(dismissal.date), now) >= 1
     }, _.zip(newAlerts, newHashes))
     _.forEach(([{ link: readMoreLink, message, title, 'link-title': linkTitle, severity }, hash]) => {
+      const notificationId = `alert-${hash}`
+      const onDismiss = () => {
+        setDynamic(localStorage, `dismiss-alerts/${hash}`, { date: now })
+      }
       notify(
         severity === 'info' ? 'info' : 'warn',
         h(Fragment, [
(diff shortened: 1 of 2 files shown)


MatthewBemis (approved):
👍 👍 👍

breilly2 (approved):
👍 As a user, I prefer having both options... make this go away right now and make it go away permanently (or for 24 hours in this case). Not having the former option, I really wish that we had a way to view this notification again if I want to (instead of having to wait 24 hours for it to appear again). If it were me, I would not make this change without also adding some way to view current notifications, but that's a requirements issue that isn't really relevant to my review of this PR.

nawatts:
From a product perspective, I think hiding the notification for 24 hours is fine since that's already the primary action and it's reasonable to assume that both "Dismiss" buttons behave the same way. 

Also, given the number of tickets we've had asking for the ability to persistently dismiss notifications that appear on load ([CA-1053](https://broadworkbench.atlassian.net/browse/CA-1053), [SU-129](https://broadworkbench.atlassian.net/browse/SU-129), [SU-185](https://broadworkbench.atlassian.net/browse/SU-185)), I think many users are generally frustrated with those types of notifications and just want to get them out of the way.

But I agree that it would be good to have an indication that there's an ongoing service alert and a way to view the alert again.  It looks like service alerts are posted in the Support forum (https://support.terra.bio/hc/en-us/sections/4415104213787), but it's not obvious from the app UI that that page exists. We had some general ideas for a notification center ([SU-107](https://broadworkbench.atlassian.net/browse/SU-107)) and I filed [SU-191](https://broadworkbench.atlassian.net/browse/SU-191) specifically for service alerts.

PR #3292, opened 2022-08-04
[IA-3602] Fix inappropriate destructure

jdcanas:
We were de-structuring an object that could be undefined. Provided appropriate default values in this case.

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -402,7 +402,8 @@ export const getCostForDisk = (app, appDataDisks, computeRegion, currentRuntimeT
   let diskCost = ''
   if (currentRuntimeTool === toolLabel && persistentDisks && persistentDisks.length > 0) {
     const curPd = getCurrentPersistentDisk(runtimes, persistentDisks)
-    diskCost = getPersistentDiskCostHourly(curPd, computeRegion)
+    const { size, status, diskType } = curPd || { size: 0, status: 'Running', diskType: pdTypes.standard }
+    diskCost = getPersistentDiskCostHourly({ size, status, diskType }, computeRegion)
   } else if (app && appDataDisks && (toolLabel === 'Galaxy')) {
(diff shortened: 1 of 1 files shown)


src/libs/runtime-utils.js
```diff
@@ -402,7 +402,8 @@ export const getCostForDisk = (app, appDataDisks, computeRegion, currentRuntimeT
   let diskCost = ''
   if (currentRuntimeTool === toolLabel && persistentDisks && persistentDisks.length > 0) {
     const curPd = getCurrentPersistentDisk(runtimes, persistentDisks)
-    diskCost = getPersistentDiskCostHourly(curPd, computeRegion)
+    const { size = 0, status = 'Running', diskType = pdTypes.standard } = curPd || {}
```
kyuksel, line 405: Not directly related to what this PR is trying to do but since it includes the piece `status = 'Running'`, I wanted to encourage the team to consider opportunistically refactoring `status`es such that they follow the pattern [tools](https://github.com/DataBiosphere/terra-ui/blob/6f14a95c41a9fee80000bcba21b5f25ceff663e6/src/components/notebook-utils.js#L75) does, with the benefits of less risk to mess up strings and ease of reading due to things like the ability to collect `status` related properties into a single place (e.g. [usableStatuses](https://github.com/DataBiosphere/terra-ui/blob/c31a3f631426b5649309c1daafd306dc5a007686/src/libs/runtime-utils.js#L80) could be a property).

kyuksel (approved):

src/libs/runtime-utils.js
n8landolt: I've wondered why we don't have a status enum

n8landolt (approved):

src/libs/runtime-utils.js
jdcanas: Yeah...definitely worth doing. It would require a good bit of refactoring though. Probably best to do piece meal as we go

PR #3296, opened 2022-08-08
[IA-3565] Add hover effects and save user sort preferences.

n8landolt:
https://broadworkbench.atlassian.net/browse/IA-3565

## Description
Users are having a hard time seeing that sorting files in the analyses tab is an active feature. Add a hover effect to show the sort arrow to better communicate this functionality.
Users would like their sort preferences to be persistent between pages.

## Testing
Manual testing.

Mouseover each column, ensure the correct arrow is showing even if sorted and hovering.

Selected a sort order that isn't the default, navigated away from the page, came back and verified the sort order.
Additionally tested on logout and verified.


Qi77Qi (approved):
tried on PR site..seems to work

n8landolt:
Looking for feedback on any "nits"
Such as the names of the constants I introduced, and the if statement logic to have the arrow icon on hover vs when it's sorted.

src/components/table.js
```diff
@@ -757,20 +757,33 @@ export const Sortable = ({ sort, field, onSort, children }) => {
   ])])
 }
 
-export const MiniSortable = ({ sort, field, onSort, children }) => {
+export const MiniSortable = ({ sort, field, onSort, children, sortHover }) => {
```
nawatts: In order to function correctly and show the icon on hover, it looks like this relies on the parent component to attach onMouseEnter/Leave event handlers and set `sortHover` to the same value as `field`.

Is there no way for MiniSortable (and Sortable) to handle this themselves? One of the goals for [IA-3548](https://broadworkbench.atlassian.net/browse/IA-3548) is that this should apply to all of Terra. That will be much easier if this can be implemented within these components instead of requiring changes to all tables/lists that use them.

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -39,27 +40,34 @@ import ExportAnalysisModal from 'src/pages/workspaces/workspace/notebooks/Export
 import { wrapWorkspace } from 'src/pages/workspaces/workspace/WorkspaceContainer'
 
 
+const FIELD_APPLICATION = 'application'
```
nawatts: Though it's not documented in the wiki, generally Terra UI code uses camel case instead of upper case for constants like these.

nawatts (changes requested):

src/components/table.js
n8landolt: I'll see if I can come up with a solution - when I initially tried it in MiniSortable the area for hover is smaller than the clickable area. This is good feedback, thank you.

n8landolt: Oh, it's super easy. I think i was getting thrown off because the IDE wanted onMouseEnter to be onmouseenter, so I must not have gone back to test.

src/components/common.js
```diff
@@ -553,7 +553,7 @@ export const ClipboardButton = ({ text, onClick, children, ...props }) => {
   }, [children, icon(copied ? 'check' : 'copy-to-clipboard', !!children && { style: { marginLeft: '0.5rem' } })])
 }
 
-export const HeaderRenderer = ({ name, label, sort, onSort, style, ...props }) => h(MiniSortable, { sort, field: name, onSort }, [
+export const HeaderRenderer = ({ name, label, sort, onSort, style, sortHover, ...props }) => h(MiniSortable, { sort, field: name, onSort, sortHover }, [
```
nawatts: After e2b208dd34d2c8d6b227479626c7c2519d2bc215, MiniSortable no longer takes a sortHover prop.

src/pages/workspaces/workspace/Analyses.js
```diff
@@ -232,7 +239,7 @@ const Analyses = _.flow(
   const [copyingAnalysisName, setCopyingAnalysisName] = useState(undefined)
   const [deletingAnalysisName, setDeletingAnalysisName] = useState(undefined)
   const [exportingAnalysisName, setExportingAnalysisName] = useState(undefined)
-  const [sortOrder, setSortOrder] = useState(() => StateHistory.get().sortOrder || defaultSort.value)
+  const [sortOrder, setSortOrder] = useState(() => getLocalPref(KEY_ANALYSES_SORT_ORDER) || defaultSort.value)
```
nawatts, line 244: Not strictly related to this PR, but it looks like `defaultSort.label` is unused and `defaultSort` could be simplified.

src/components/table.js
```diff
@@ -741,36 +741,56 @@ export const HeaderCell = props => {
   return h(TextCell, _.merge({ style: { fontWeight: 600 } }, props))
 }
 
+const getSortIcon = (sort, field, sortHover, style) => {
```
nawatts: The current implementation seems overly complicated. There's a lot of duplication in the two if conditions, when the only thing that varies is the icon name.

src/components/table.js
```diff
@@ -741,36 +741,56 @@ export const HeaderCell = props => {
...
+      style
+    }, [
+      icon('long-arrow-alt-down')
+    ])
+  }
+  return sortIcon
+}
+
 export const Sortable = ({ sort, field, onSort, children }) => {
+  const [sortHover, setSortHover] = useState()
```
nawatts: Would the purpose of this state be clearer with a boolean like `hovered` vs a string?

nawatts (approved):
There's a few things that I think could be cleaned up some more, but generally looks good.

nawatts:
Also, should this show the sort icon when the column header receives keyboard focus?

src/components/data/data-utils.js
```diff
@@ -1397,10 +1397,10 @@ export const HeaderOptions = ({ sort, field, onSort, extraActions, children }) =
   ])
 
   return h(Sortable, {
-    sort, field, onSort
+    sort, field, onSort, columnMenu
```
nawatts, line 1400: Not necessary to explore here, but I wonder if this could have also been solved by moving `columnMenu` outside of Sortable. Maybe by returning a Fragment.

nawatts (approved):

PR #3300, opened 2022-08-08
[QA-1988] UI Integration Test: Slack message templates

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1988

Avoid clutter code, make Slack message format and contents easier to read and update.
Let me know if there are other CircleCI env variables which may be useful to have in notification.

Fields:
- Title
- `Job [BUILD_NUM`: Link to CircleCI job
- `Commit`: Link to Github commit hash that triggered CircleCI job

screenshots (from mock CircleCI env variables):

- Test job succeded

   Screen Shot 2022-08-08 at 3 11 01 PM

- Test job failed (two failed tests in list)

   Screen Shot 2022-08-08 at 3 09 57 PM

### DIFF
--- a/integration-tests/slack/message-templates.js
+++ b/integration-tests/slack/message-templates.js
@@ -53,9 +53,18 @@ const getTestsListBlock = testNames => {
 }
 
 // any non-zero number to indicate test(s) has failed in circleci job
-const getMessageBlockTemplate = (numFailedTests = 0) => [
-  getHeaderBlock(numFailedTests),
-  getJobDetailBlock()
-]
+const getMessageBlockTemplate = testNames => {
+  const size = testNames.length
 
-module.exports = { getMessageBlockTemplate, getTestsListBlock }
+  const blocksArray = [
+    getHeaderBlock(size),
+    getJobDetailBlock()
+  ]
+
+  if (size === 0) {
+    return blocksArray
+  } else {
+    return blocksArray.concat(getTestsListBlock(testNames))
+  }
+}
+module.exports = { getMessageBlockTemplate }
--- a/integration-tests/slack/message-templates.js
+++ b/integration-tests/slack/message-templates.js
@@ -40,10 +40,22 @@ const getJobDetailBlock = () => {
   }
 }
 
+const getTestsListBlock = testNames => {
+  return {
+    type: 'section',
+    fields: [
+      {
+        type: 'mrkdwn',
+        text: `*  ${testNames.join('\n*  ')}`
+      }
+    ]
+  }
+}
+
 // any non-zero number to indicate test(s) has failed in circleci job
 const getMessageBlockTemplate = (numFailedTests = 0) => [
   getHeaderBlock(numFailedTests),
   getJobDetailBlock()
 ]
 
-module.exports = { getMessageBlockTemplate }
+module.exports = { getMessageBlockTemplate, getTestsListBlock }
--- a/integration-tests/slack/message-templates.js
+++ b/integration-tests/slack/message-templates.js
@@ -24,8 +24,8 @@ const getHeaderBlock = numFailedTests => {
   }
 }
 
-const getJobDetailBlock = () => {
-  return {
+const getJobDetailBlock = () => (
+  {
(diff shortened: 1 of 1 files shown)


cahrens (approved):

integration-tests/slack/message-templates.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+// any non-zero number to indicate test(s) has failed in circleci job
+const getMessageBlockTemplate = (numFailedTests = 0) => [
+  getHeaderBlock(numFailedTests),
+  getJobDetailBlock()
+]
+
+module.exports = { getMessageBlockTemplate, getTestsListBlock }
```
nawatts: Do these need to be separated? Could `getMessageBlockTemplate` accept the list of failed tests instead of the number of failed tests and add the tests list block?

aweng98: No, doesn't need to be separated. For me, it's clean and clear. 
Below is what I came up with for `getMessageBlockTemplate(testNames)`. It works and one less func to export. But it's more code. Do you recommend to change to this?
```
const getMessageBlockTemplate = (testNames) => {
  const size = testNames.length

  const blocksArray = [
    getHeaderBlock(size),
    getJobDetailBlock()
  ]

  if (size === 0) {
    return blocksArray
  } else {
    return blocksArray.concat(getTestsListBlock(testNames))
  }
}
```

nawatts: I think I would be in favor of this change. The code for conditionally appending the tests list block has to go somewhere, and this way the logic for building the message is encapsulated in this module.

aweng98: That's a good point.  Updated and thank you.

integration-tests/slack/message-templates.js
```diff
@@ -0,0 +1,69 @@
...
+    fields: [
+      {
+        type: 'mrkdwn',
+        text: `*  ${testNames.join('\n*  ')}`
+      }
+    ]
+  }
+}
+
+const getMessageBlockTemplate = testNames => {
```
nawatts: `testNames` is a generic name. It would be nice to clarify that this function takes a list of _failed_ tests.

nawatts (approved):

integration-tests/slack/message-templates.js
```diff
@@ -0,0 +1,69 @@
...
+  return {
+    type: 'section',
+    fields: [
+      {
+        type: 'mrkdwn',
+        text: `*  ${testNames.join('\n*  ')}`
+      }
+    ]
+  }
+}
```
kyuksel: FYI that it could be written more concisely as follows:

```suggestion
const getTestsListBlock = testNames => (
  {
    type: 'section',
    fields: [
      {
        type: 'mrkdwn',
        text: `*  ${testNames.join('\n*  ')}`
      }
    ]
  }
)
```

or

```javascript
const getTestsListBlock = testNames => ({
  type: 'section',
  fields: [
    {
      type: 'mrkdwn',
      text: `*  ${testNames.join('\n*  ')}`
    }
  ]
})
```

integration-tests/slack/message-templates.js
```diff
@@ -0,0 +1,69 @@
...
+  const blocksArray = [
+    getHeaderBlock(size),
+    getJobDetailBlock()
+  ]
+
+  if (size === 0) {
+    return blocksArray
+  } else {
+    return blocksArray.concat(getTestsListBlock(testNames))
+  }
```
kyuksel: Nit: Straightforward conditionals like these are worth writing more concisely IMO

```suggestion
  size === 0 ? blocksArray : blocksArray.concat(getTestsListBlock(testNames))
```

kyuksel (approved):
The Slack messages look great - thank you!
A couple of nits are inline.

integration-tests/slack/message-templates.js
aweng98: Cool. Updated `getTestsListBlock()` and `getJobDetailBlock()`. Thank you.

PR #3301, opened 2022-08-08
[PPWM-76] Mark persistent disks as 'Offline' post workspace migration

ehigham:
RR: https://broadworkbench.atlassian.net/browse/PPWM-75
When a v1 workspace is migrated to a v2 workspace, users need to "migrate" their old persistent disks from the original google project to the workspace's newly-created google project. Since this is a per-user operation and we have no way of knowing which workspace a disk should belong in, we require the user to do this.
In the "Cloud Environments" page:
- Show peristent disks that require migration as "Offline"
- Include a to a new modal that allows users to select which workspaces they'd like to copy the disk to or delete the disk.
- Show a banner when user has "Offline" persistent disks.

"Offline" disk:
image

Migrate disk modal:
image

Delete disk option:
image

Delete disk confirm:
image

When user cannot copy to any workspace:
image

Updated tooltip in create environment
image

<!--
Hello, friend!
Thanks!

--!>


src/libs/ajax.js
```diff
@@ -1739,6 +1739,12 @@ const Disks = signal => ({
 
   disk: (project, name) => {
     return {
+      create: ({ ...props }) => {
+        const body = jsonBody({ ...props })
+        return fetchLeo(`api/google/v1/disks/${project}/${name}`,
+          _.mergeAll([authOpts(), appIdentifier, { signal, method: 'POST' }, body])
+        )
```
nawatts: There's no need for the spread operator here. It's needlessly creating shallow copies of the object passed in.

```suggestion
      create: props => {
        const body = jsonBody(props)
        return fetchLeo(`api/google/v1/disks/${project}/${name}`,
          _.mergeAll([authOpts(), appIdentifier, { signal, method: 'POST' }, body])
        )
```
This could also be shortened a bit:
```suggestion
      create: props => fetchLeo(`api/google/v1/disks/${project}/${name}`,
        _.mergeAll([authOpts(), appIdentifier, { signal, method: 'POST' }, jsonBody(props)]))
```

src/pages/Environments.js
```diff
@@ -116,8 +120,165 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
   ])
 }
 
+const MigratePersistentDisksBanner = ({ count }) => {
+  const deadline = new Date('01 January 2023 00:00 UTC')
```
nawatts: I think ISO 8601 is the only format of string that is guaranteed to work here.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters

Alternatively, the  [date-fns](https://date-fns.org/docs/Getting-Started/) library can use different formats.

src/pages/Environments.js
```diff
@@ -116,8 +120,165 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+        strong([`You have ${Math.floor((deadline - Date.now()) / 86400000)} days to migrate ${count} shared persistent ${count > 1 ?
+          'disks' :
+          'disk'}. `]),
+        `Un-migrated disks will be DELETED after ${Utils.makeCompleteDate(deadline)}.`
+      ])
+    ])
+  ])
+}
+
+const MigratePersistentDiskCell = ({ disk, setMigrateDisk }) => div({
```
nawatts: Nit: For a cleaner boundary between components, maybe calls this prop `onClick` instead of `setMigrateDisk`.

src/pages/Environments.js
```diff
@@ -116,8 +120,165 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+    div(['Offline', icon('warning-standard', { style: { marginLeft: '0.25rem', color: colors.danger() } })])
+  ]),
+  h(Link, { onClick: () => setMigrateDisk(disk), style: { wordBreak: 'break-word' } },
+    ['Migrate']
+  )
+])
+
+const MigratePersistentDiskModal = ({ disk, workspaces, onSuccess, onDismiss, contactSupport, deleteDiskId }) => {
+  const [migrating, setMigrating] = useState()
+  const [isSelected, setIsSelected] = useState({})
```
nawatts: A comment explaining the keys/values of this object would be helpful. Also, the `is...` name makes me think this is a boolean rather than an object.

src/pages/Environments.js
```diff
@@ -116,8 +120,165 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+    ({ googleProject, namespace: saturnWorkspaceNamespace, name: saturnWorkspaceName }) => {
+      return reportErrorAndRethrow(`Error copying persistent disk to workspace '${saturnWorkspaceName}'`,
+        () => Ajax().Disks.disk(googleProject, disk.name).create({
+          ..._.pick(['size', 'blockSize', 'zone'], disk),
+          diskType: disk.diskType.label,
+          labels: { saturnWorkspaceNamespace, saturnWorkspaceName },
+          sourceDisk: _.pick(['googleProject', 'name'], disk)
+        })
+      )()
+    }
```
nawatts: Instead of immediately calling the function returned by `reportErrorAndRethrow`, it can be used like this:

```suggestion
  const copyDiskToWorkspace = reportErrorAndRethrow(
    `Error copying persistent disk to workspace '${saturnWorkspaceName}'`
    ({ googleProject, namespace: saturnWorkspaceNamespace, name: saturnWorkspaceName }) => {
      return Ajax().Disks.disk(googleProject, disk.name).create({
          ..._.pick(['size', 'blockSize', 'zone'], disk),
          diskType: disk.diskType.label,
          labels: { saturnWorkspaceNamespace, saturnWorkspaceName },
          sourceDisk: _.pick(['googleProject', 'name'], disk)
      })
    }
  )
```

src/pages/Environments.js
```diff
@@ -116,8 +120,165 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+
+  const migrateDisk = _.flow(onError(onDismiss), Utils.withBusyState(setMigrating))(async () => {
```
nawatts: The combination of `reportErrorAndRethrow` and `onError(onDismiss)` looks similar to `withErrorReportingInModal`.

https://github.com/DataBiosphere/terra-ui/blob/4c71dba6556eeccb520dc2038202529f4fa0fae2/src/libs/error.js#L43-L57

src/pages/Environments.js
```diff
@@ -116,8 +120,165 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+    const destinations = _.filter(({ workspaceId }) => isSelected[workspaceId], workspaces)
+    await Promise.all(_.map(copyDiskToWorkspace, destinations))
+    await onSuccess()
+  })
+
+  const DeleteDiskSelection = ({ ...props }) => {
```
nawatts: Similar to a previous comment, this can be:

```suggestion
  const DeleteDiskSelection = props => {
```

src/pages/Environments.js
```diff
@@ -116,8 +120,165 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+    okButton: deleteDisk || _.isEmpty(workspaces) ?
+      h(ButtonPrimary, { disabled: !deleteDisk, onClick: _.flow(onDismiss, _.constant(disk.id), deleteDiskId) }, 'Delete') :
+      h(ButtonPrimary, { disabled: numberOfCopies === 0, onClick: migrateDisk }, 'Migrate'),
+    onDismiss
+  }, [
+    div({ style: { display: 'flex', flexDirection: 'column' } }, Array.prototype.concat(
+      [span(space, 'Due to data security policies, persistent disks can no longer be shared between workspaces.')],
+      _.isEmpty(workspaces) ? [
+        strong(space, ['You own this disk but do not have access to any workspaces where it can be shared.']),
+        h(DeleteDiskSelection, space),
```
nawatts: Because `DeleteDiskSelection` is defined inside `MigratePersistentDiskModal`, it will be a different function in each render of `MigratePersistentDiskModal`. Thus, when it's rendered as a component like this, React will see it as a different component and unmount and destroy the previous instance and create and mount a new instance. This is inefficient and can lead to unexpected behavior if any hooks were used in `DeleteDiskSelection`

Instead of using `DeleteDiskSelection` as a component, you can use it as a function (you may want to rename it in this case).

```suggestion
        DeleteDiskSelection(space),
```

src/pages/Environments.js
```diff
@@ -195,7 +361,7 @@ const Environments = () => {
     await loadData()
   })
 
-  useOnMount(() => { loadData() })
+  useOnMount(loadData)
```
nawatts: I think this was probably intentional. Effect functions return a cleanup function which React will call when a dependency changes or the component is unmounted. An async effect function will return a Promise, which will cause an error when React attempts to call it as a function.

https://reactjs.org/docs/hooks-effect.html

src/pages/Environments.js
```diff
@@ -133,36 +294,41 @@ const Environments = () => {
...
   const [shouldFilterRuntimesByCreator, setShouldFilterRuntimesByCreator] = useState(true)
+  const [migrateDisk, setMigrateDisk] = useState()
 
   const refreshData = Utils.withBusyState(setLoading, async () => {
+    await refreshWorkspaces()
     const creator = getUser().email
 
+    const getWorkspace = (ws => (namespace, name) => _.get(`${namespace}.${name}`, ws))(
+      getWorkspaces()
+    )
```
nawatts: This took me a minute to figure out what it was doing. Is there a need to capture the return value of `getWorkspaces` like this. It's returning the value of a ref, so it wouldn't be expensive to call.

```suggestion
    const getWorkspace = (namespace, name) => _.get(`${namespace}.${name}`, getWorkspaces())
```

src/pages/Environments.js
```diff
@@ -116,8 +120,165 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
 
+const MigratePersistentDisksBanner = ({ count }) => {
+  const deadline = new Date('01 January 2023 00:00 UTC')
+  return count > 0 && div({
+    style: {
+      position: 'absolute', top: topBarHeight, left: '50%', transform: 'translate(-50%, -50%)',
+      backgroundColor: colors.warning(0.15),
+      border: '2px solid', borderRadius: '12px', borderColor: colors.warning(),
+      zIndex: 8000 // Must be under 9000 to not draw over notifications
+    }
```
nawatts: Does the banner really need to overlap with the top bar? We could get rid of all of this position, transform, and zIndex stuff if it was across the top of the main content area.

src/pages/Environments.js
ehigham: I guess not, I was just eliminating an extra function call each time `getWorkspace` is called. I still prefer to not code in inefficiencies like this.

src/pages/Environments.js
ehigham: According to the UX mock-ups in the ticket, the banner does. I'm just the messenger haha

src/pages/Environments.js
```diff
@@ -647,8 +814,29 @@ const Environments = () => {
...
+        onDismiss: () => setMigrateDisk(undefined),
+        onSuccess: () => {
+          setMigrateDisk(undefined)
+          loadData()
+        },
+        contactSupport: () => contactUsActive.set(true),
+        deleteDiskId: setDeleteDiskId
+      })
+    ]),
+    h(MigratePersistentDisksBanner, { count: _.countBy('requiresMigration', disks).true }, []),
```
nawatts: Instead of having `MigratePersistentDisksBanner` render nothing if `count` is 0, it might be clearer to conditionally render this component.

```suggestion
    _.countBy('requiresMigration', disks).true > 0 && h(MigratePersistentDisksBanner, { count: _.countBy('requiresMigration', disks).true }, []),
```

src/pages/Environments.js
nawatts: On narrower windows (Terra UI supports down to 1200px), the banner overlaps the title in the top bar.

Screen Shot 2022-09-08 at 11 21 38 AM

src/pages/Environments.js
ehigham: not sure how this works as the binding for `saturnWorkspaceName` comes from the ~lambda~ arrow?

src/pages/Environments.js
ehigham: Almost, though i want to show every failure but call `onDismiss` once

src/pages/Environments.js
ehigham: I'd like to understand this more, in particular when to `h` stuff and when to not (you can tell I'm *clearly* very knowledgeable...). Maybe a topic to discuss in our meeting

src/pages/Environments.js
nawatts: Oops, didn't notice that. Ignore this then.

src/pages/Environments.js
nawatts: Sure, but since `getWorkspaces` is only reading a value from a ref object, I'd claim that the optimization isn't worth the loss in readability.
https://github.com/DataBiosphere/terra-ui/blob/56d4157468bab113ae025880c4d1dcce513d30f7/src/libs/react-utils.js#L30-L34

src/pages/Environments.js
nawatts: Ooh, I missed that detail as well.

src/pages/Environments.js
ehigham: Yeah - that's what the mockup does

ehigham: See attachments in https://broadworkbench.atlassian.net/browse/PPWM-76

src/pages/Environments.js
```diff
@@ -116,8 +120,176 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
 }
 
+const MigratePersistentDisksBanner = ({ count }) => {
+  const deadline = new Date('2023-01-01T00:00:00.000Z')
+  return div({
+    style: {
+      position: 'absolute', top: topBarHeight, left: '50%', transform: 'translate(-50%, -50%)',
+      backgroundColor: colors.warning(0.15),
+      border: '2px solid', borderRadius: '12px', borderColor: colors.warning(),
+      zIndex: 2 // Must be under 9000 to not draw over notifications
```
ehigham:
```suggestion
      zIndex: 2 // Draw over top bar but behind contact support dialog
```

src/pages/Environments.js
```diff
@@ -647,8 +824,32 @@ const Environments = () => {
...
+        disk: migrateDisk,
+        workspaces: _.flow(
+          _.get(migrateDisk.googleProject),
+          _.values,
+          _.filter(({ accessLevel, workspace: { googleProject } }) => {
+            return Utils.canWrite(accessLevel) && _.isEmpty(disksByProject[googleProject])
+          }),
+          _.map('workspace'),
+          _.sortBy(({ name }) => _.lowerCase(name)),
+          _.constant([])
```
ehigham:
```suggestion
          _.sortBy(({ name }) => _.lowerCase(name))
```

src/pages/Environments.js
nawatts: Instead of showing one notification for each failure, you could collect all failed disks and show a notification with the list.

src/libs/error.js
```diff
@@ -63,3 +63,11 @@ export const withErrorReportingInModal = _.curry((title, onDismiss, fn) => async
 export const withErrorReporting = _.curry((title, fn) => {
   return withErrorIgnoring(reportErrorAndRethrow(title)(fn))
 })
+
+
+/**
+ * Invoke the `callback` with any error thrown when evaluating the async `fn` with `...args`.
+ */
+export const onError = _.curry((callback, fn) => async (...args) => {
```
nawatts: Many (most?) functions in Terra UI that wrap other functions follow the naming pattern `with...`. Maybe something like `withErrorHandling`?

Also, it looks like most of the other `with...` functions in this module could be implemented using this.

src/pages/Environments.js
```diff
@@ -116,8 +120,176 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+      position: 'absolute', top: topBarHeight, left: '50%', transform: 'translate(-50%, -50%)',
+      backgroundColor: colors.warning(0.15),
+      border: '2px solid', borderRadius: '12px', borderColor: colors.warning(),
+      zIndex: 2 // Draw over top bar but behind contact support dialog
+    }
+  }, [
+    div({ style: { display: 'flex', alignItems: 'center', margin: '0.75rem 1.5rem 0.75rem 1.5rem' } }, [
+      icon('warning-standard', { size: 32, style: { color: colors.warning(), marginRight: '0.25rem' } }),
+      div([
+        strong([`You have ${Math.floor((deadline - Date.now()) / 86400000)} days to migrate ${count} shared persistent ${count > 1 ?
```
nawatts: FYI, date-fns has functions for things like this: https://date-fns.org/v2.29.2/docs/differenceInDays

src/pages/Environments.js
```diff
@@ -116,8 +120,176 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+        `Un-migrated disks will be DELETED after ${Utils.makeCompleteDate(deadline)}.`
+      ])
+    ])
+  ])
+}
+
+const MigratePersistentDiskCell = ({ onClick }) => div({
+  style: {
+    display: 'flex', flex: 1, flexDirection: 'column', textAlign: 'center',
+    height: '100%', margin: '-1rem', padding: '0.5rem 0 0 0',
```
nawatts, line 150: Is there a way to avoid the negative margin? Perhaps with FlexTable's `styleCell` prop?

src/pages/Environments.js
```diff
@@ -116,8 +120,176 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+      _.isEmpty(workspaces) ? [
+        strong(space, ['You own this disk but do not have access to any workspaces where it can be shared.']),
+        renderDeleteDiskSelection(space),
+        span(space, 'OR'),
+        div([
+          h(Link, { onClick: _.flow(onDismiss, onContactSupport) }, 'Contact Terra Support'),
+          ' to have it transferred to another user.'
+        ])
+      ] : [
+        strong(space, ['Select workspaces where you what to use a copy of this disk.']),
```
nawatts: Is this supposed to say "where you **want** to use a copy"?

n8landolt (approved):

src/pages/Environments.js
```diff
@@ -116,8 +120,176 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+    }
+
+  // dismiss the UI at most once on error
+  const migrateDisk = _.flow(onError(onDismiss), Utils.withBusyState(setMigrating))(async () => {
+    const destinations = _.filter(({ workspaceId }) => isWorkspaceSelected[workspaceId], workspaces)
+    await Promise.all(_.map(copyDiskToWorkspace, destinations))
+    await onSuccess()
+  })
+
+  const renderDeleteDiskSelection = props => {
```
nawatts: It looks like this is always called with the same input (`space`). Since that input contains only a constant style, does this need to take an argument?

src/pages/Environments.js
```diff
@@ -116,8 +120,176 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+                  onChange: selected => {
+                    setIsWorkspaceSelected(_.set(workspaceId, selected, isWorkspaceSelected))
+                    setDeleteDisk(false)
+                  }
+                })
+              }
+            },
+            {
+              field: 'workspace',
+              headerRenderer: () => strong(['Workspace name']),
```
nawatts: Maybe use the `HeaderCell` component from the table module instead of strong here. [strong](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong) doesn't seem really appropriate here and that would make the style consistent with other tables.

src/pages/Environments.js
```diff
@@ -116,8 +120,176 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+              cellRenderer: ({ rowIndex }) => {
+                const { authorizationDomain, name } = workspaces[rowIndex]
+                const authDomains = _.map('membersGroupName', authorizationDomain)?.join(', ')
+                const text = Utils.cond(
+                  [authorizationDomain.length === 1, () => `${name} (Authorization Domain: ${authDomains})`],
+                  [authorizationDomain.length > 1, () => `${name} (Authorization Domains: ${authDomains})`],
+                  [Utils.DEFAULT, () => name]
+                )
+                return div({ title: text, style: { ...Style.noWrapEllipsis } }, [text])
```
nawatts: There's also a `TextCell` component that adds the `noWrapEllipsis` style.

src/pages/Environments.js
```diff
@@ -647,8 +824,31 @@ const Environments = () => {
...
+          setMigrateDisk(undefined)
+          loadData()
+        },
+        onContactSupport: () => contactUsActive.set(true),
+        onDeleteDisk: () => setDeleteDiskId(migrateDisk.id)
+      })
+    ]),
+    (count => count > 0 && h(MigratePersistentDisksBanner, { count }, []))(
+      _.countBy('requiresMigration', disks).true
+    ),
```
nawatts: You explained your reasons, but I still think things like this are less readable/more confusing than storing `count` before the `return` expression.

nawatts (approved):

src/pages/Environments.js
```diff
@@ -1,45 +1,49 @@
...
   const [deleting, setDeleting] = useState()
   const deleteRuntime = _.flow(
     Utils.withBusyState(setDeleting),
     withErrorReporting('Error deleting cloud environment')
   )(async () => {
-    cloudProvider === cloudProviders.gcp.label ?
+    isGcpContext(cloudContext) ?
       await Ajax().Runtimes.runtime(googleProject, runtimeName).delete(deleteDisk) :
-      await Ajax().Runtimes.runtimeV2(workspaceId, runtimeName).delete(deleteDisk)
+      await Ajax().Runtimes.runtimeV2(workspace.workspace.workspaceId, runtimeName).delete(deleteDisk)
```
jdcanas: why aren't we destructuring this in the props anymore?

src/pages/Environments.js
```diff
@@ -116,8 +120,176 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+  const [migrating, setMigrating] = useState()
+  // workspaceId -> boolean indicating if we should copy the disk to the workspace
+  const [isWorkspaceSelected, setIsWorkspaceSelected] = useState({})
+  // users can choose to delete their disk instead of coping via a checkbox. Mutually exclusive with `isWorkspaceSelected`
+  const [deleteDisk, setDeleteDisk] = useState(false)
+
+  const copyDiskToWorkspace =
+    ({ googleProject, namespace: saturnWorkspaceNamespace, name: saturnWorkspaceName }) => {
+      // show an error for each failed copy operation
+      return reportErrorAndRethrow(`Error copying persistent disk to workspace '${saturnWorkspaceName}'`,
```
jdcanas: Can we use `withErrorReportingInModal` here? Not too sure what the difference between that and this function is, but the one I mention handles some annoying things about error handling in modals, such as a modal obscuring the error based on positioning, and onDismiss/cleanup not properly being done after an error

src/pages/Environments.js
```diff
@@ -116,8 +120,176 @@ const DeleteAppModal = ({ app: { googleProject, appName, diskName, appType }, on
...
+          ..._.pick(['size', 'blockSize', 'zone'], disk),
+          diskType: disk.diskType.label,
+          labels: { saturnWorkspaceNamespace, saturnWorkspaceName },
+          sourceDisk: _.pick(['googleProject', 'name'], disk)
+        })
+      )()
+    }
+
+  // dismiss the UI at most once on error
+  const migrateDisk = _.flow(onError(onDismiss), Utils.withBusyState(setMigrating))(async () => {
```
jdcanas: consider `withErrorReportingInModal`

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -226,12 +226,18 @@ const useCloudEnvironmentPolling = (googleProject, workspace) => {
       // TODO: after PPW migration - we should only need saturnWorkspaceName and saturnWorkspaceNamespace labels
       const cloudEnvFilters = _.pickBy(l => !_.isUndefined(l), saturnWorkspaceVersion === 'v1' ? { creator: getUser().email, googleProject } : { creator: getUser().email, saturnWorkspaceName, saturnWorkspaceNamespace })
 
+      // v2 workspaces may have been migrated from v1 workspaces, in which case the googleProject
+      // associated with runtimes and disks will not match the workspace googleProject. These
+      // should be hidden from the user.
+      const isV1Artefact = ({ googleProject = undefined }) => googleProject && googleProject !== workspace.workspace.googleProject
```
jdcanas: `art**i**fact` 🇬🇧

src/pages/Environments.js
ehigham: This used to be the workspaceId - now it's the whole workspace. I've since removed one level of nesting but to destructure futher doesn't add clarity

src/pages/workspaces/workspace/WorkspaceContainer.js
ehigham: blast.

src/pages/Environments.js
ehigham: Do we have code to do this already? Otherwise I'll make a follow up change.

src/pages/Environments.js
ehigham: See https://github.com/DataBiosphere/terra-ui/pull/3301#discussion_r966031338

src/pages/Environments.js
ehigham: As above

src/pages/Environments.js
nawatts: ImportData has something similar to what I had in mind. Use [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) to wait for all requests, then throw a custom error if any failed. And the usual `withErrorReporting` wrapper could be used with that.

https://github.com/DataBiosphere/terra-ui/blob/ff9c7a2b69eb4cc5f0be9d4b3d2956c4ab8ff076/src/pages/ImportData.js#L152-L171

ehigham: Would the error ballon/thing layout the error correctly as it's designed for displaying one error rather than many?

ehigham: I'd rather not have to add a new component to layout these errors and I'm not expecting users to select many workspaces to copy to

nawatts: You can throw a single error if any of the requests fail. `withErrorReporting` would handle that error.

Similar to this in ImportData:
https://github.com/DataBiosphere/terra-ui/blob/ff9c7a2b69eb4cc5f0be9d4b3d2956c4ab8ff076/src/pages/ImportData.js#L171

ehigham: Right, but it'll be one error containing multiple messages - ie
Failed to copy disk to workspace A because ...
Failed to copy disk to workspace D because ...

src/pages/Environments.js
ehigham: I've looked into it a bit and changing the `SimpleFlexTable` to be a `FlexTable` is a bit beyond my competency. This code is temporary and should be deleted within 6 months so I'm not sure if it's worth it.

nawatts: Oh, I missed that this was using SimpleFlexTable instead of FlexTable.

src/libs/error.js
ehigham: Would you mind taking a look at this again and making sure you're happy with the new implementations?

ehigham: ok they're wrong. I'll fix them. `_.curry` does not behave how I expected.

ehigham: Fixed - can you take a look?

src/libs/error.js
```diff
@@ -15,29 +15,28 @@ export const reportError = async (title, obj) => {
...
  * evaluation fails.
  */
-export const withErrorIgnoring = fn => async (...args) => {
-  try {
-    return await fn(...args)
-  } catch (error) {
-    // ignore error
-  }
-}
+export const withErrorIgnoring = withErrorHandling(() => {})
```
nawatts: Could also use [`_.noop`](https://lodash.com/docs/4.17.15#noop) here.

src/libs/error.js
```diff
@@ -15,29 +15,28 @@ export const reportError = async (title, obj) => {
...
 /**
  * Return a Promise to the result of evaluating the async `fn` with `...args`. If evaluation fails,
  * report the error to the user with `title` as a side effect.
  */
-export const reportErrorAndRethrow = _.curry((title, fn) => async (...args) => {
-  try {
-    return await fn(...args)
-  } catch (error) {
+export const reportErrorAndRethrow = _.curry((title, fn) => {
+  return _.flip(withErrorHandling)(fn, error => {
```
nawatts, line 36: Is `_.flip` needed here?

ehigham: Needed? No. I think it's nicer to look at than: 
```javascript
return withErrorHandling(error => {
  reportError(title, error)
  throw error
}, fn)
```

PR #3302, opened 2022-08-09
[QA-1990] UI integration test slack notification from circleci

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1990

(For context, this is split from https://github.com/DataBiosphere/terra-ui/pull/3287)

Changes:
- Retreive CircleCI job artifacts
- Process Jest test summary JSON files to retrieve tests information
- Compose and send Slack message when test job is finished
- Provide a way to run JS code in Terminal window so it can be launched in a CircleCI step (not included in this PR)

To test this:

- Set `CIRCLE_BUILD_NUM` and `SLACK_BOT_TOKEN` env variables
- From `integration-tests` dir, cmd to invoke script: `node --require ../.pnp.cjs ./slack/notify-circleci-test-results.js` 

Example: this notification was posted to `dsp-workspaces-test-alerts` today.

Screen Shot 2022-08-09 at 3 14 25 PM

### DIFF
--- a/integration-tests/slack/circleci-utils.js
+++ b/integration-tests/slack/circleci-utils.js
@@ -4,10 +4,10 @@ const { parse } = require('path')
 
 
 /**
- * Fetch CircleCI artifact links to tests summary JSON files (were created in onRunComplete() in jest-reporter.js)
- * @param { string } token
- * @param { string } buildNum
- * @returns { Promise<Array[string]> } URL to tests-summary-[0-9].json
+ * Fetch CircleCI artifact links to tests summary JSON files (created in onRunComplete() in jest-reporter.js)
+ * @param {string} token
+ * @param {string} buildNum
+ * @returns {Promise<Array[string]>} URL to tests-summary-[0-9].json
  */
 const fetchJobArtifacts = async ({ buildNum = process.env.CIRCLE_BUILD_NUM } = {}) => {
   if (!buildNum) {
@@ -30,8 +30,8 @@ const fetchJobArtifacts = async ({ buildNum = process.env.CIRCLE_BUILD_NUM } = {
 
 /**
  *
- * @param { string } aggregatedResults A JSON object
- * @returns { Array[string] }
+ * @param {@link https://github.com/facebook/jest/blob/240587bde5dae1467ced0fdeee2e668e01caf896/packages/jest-test-result/src/types.ts#L77 AggregatedResult} Results from the test run.
+ * @returns {Array[string]}
  */
 const getFailedTestNames = aggregatedResults => {
   return _.flow(
@@ -41,12 +41,11 @@ const getFailedTestNames = aggregatedResults => {
 }
 
 /**
- *
- * @returns { Promise<Array[string]> }
+ * Parse all tests-summary JSON to look for failed test names
+ * @returns {Promise<Array[string]>}
  */
 const getFailedTestNamesFromArtifacts = async () => {
   const tests = []
-  // Parse all tests-summary JSON to look for failed test names
   const urls = await fetchJobArtifacts()
   await Promise.all(_.map(async url => {
     try {
--- a/integration-tests/slack/circleci-utils.js
+++ b/integration-tests/slack/circleci-utils.js
@@ -21,7 +21,7 @@ const fetchJobArtifacts = async ({ buildNum = process.env.CIRCLE_BUILD_NUM } = {
     const response = await fetch(`${apiUrlRoot}/${buildNum}/artifacts`)
     const { items } = await response.json()
     const testSummaryArtifacts = _.filter(_.flow(_.get('path'), _.includes('tests-summary')), items)
-    return _.map(_.get('url'), testSummaryArtifacts)
+    return _.map('url', testSummaryArtifacts)
   } catch (e) {
     console.error(`**  ERROR: Encountered error when getting CircleCI JOB_BUILD_NUM: ${buildNum} artifacts.`, e)
     throw e
@@ -45,20 +45,16 @@ const getFailedTestNames = aggregatedResults => {
  * @returns {Promise<Array[string]>}
  */
 const getFailedTestNamesFromArtifacts = async () => {
-  const tests = []
   const urls = await fetchJobArtifacts()
-  await Promise.all(_.map(async url => {
+  return _.flatten(await Promise.all(_.map(async url => {
     try {
       const response = await fetch(url)
-      const failedTests = getFailedTestNames(await response.json())
-      tests.push(...failedTests)
+      return getFailedTestNames(await response.json())
     } catch (e) {
       console.error(`**  ERROR: Encountered error when getting CircleCI artifacts tests-summary.json: ${url}.`, e)
       throw e
     }
-  }, urls))
-
-  return tests
+  }, urls)))
 }
 
 module.exports = {
--- a/integration-tests/slack/circleci-utils.js
+++ b/integration-tests/slack/circleci-utils.js
@@ -1,0 +1,67 @@
+const _ = require('lodash/fp')
+const fetch = require('node-fetch')
+const { parse } = require('path')
+
+
+/**
+ * Fetch CircleCI artifact links to tests summary JSON files (were created in onRunComplete() in jest-reporter.js)
+ * @param { string } token
+ * @param { string } buildNum
+ * @returns { Promise<Array[string]> } URL to tests-summary-[0-9].json
+ */
+const fetchJobArtifacts = async ({ buildNum = process.env.CIRCLE_BUILD_NUM } = {}) => {
+  if (!buildNum) {
+    throw new Error(`**  ERROR: Missing CircleCI build number. Failed to fetch CircleCI job artifacts.`)
+  }
+
+  // For more arguments and details of the response, see: https://circleci.com/docs/api/v2/index.html#operation/getJobArtifacts
+  const apiUrlRoot = 'https://circleci.com/api/v2/project/github/DataBiosphere/terra-ui'
+  try {
+    // Because terra-ui is a public repository on GitHub, API token is not required. See: https://circleci.com/docs/oss#security
+    const response = await fetch(`${apiUrlRoot}/${buildNum}/artifacts`)
+    const { items } = await response.json()
+    const testSummaryArtifacts = _.filter(_.flow(_.get('path'), _.includes('tests-summary')), items)
+    return _.map(_.get('url'), testSummaryArtifacts)
+  } catch (e) {
+    console.error(`**  ERROR: Encountered error when getting CircleCI JOB_BUILD_NUM: ${buildNum} artifacts.`, e)
+    throw e
+  }
+}
+
+/**
+ *
+ * @param { string } aggregatedResults A JSON object
+ * @returns { Array[string] }
+ */
+const getFailedTestNames = aggregatedResults => {
+  return _.flow(
+    _.filter(testResult => testResult.numFailingTests > 0),
+    _.map(testResult => parse(testResult.testFilePath).name)
+  )(aggregatedResults.testResults)
+}
+
+/**
+ *
+ * @returns { Promise<Array[string]> }
+ */
+const getFailedTestNamesFromArtifacts = async () => {
+  const tests = []
+  // Parse all tests-summary JSON to look for failed test names
+  const urls = await fetchJobArtifacts()
+  await Promise.all(_.map(async url => {
+    try {
+      const response = await fetch(url)
+      const failedTests = getFailedTestNames(await response.json())
+      tests.push(...failedTests)
+    } catch (e) {
+      console.error(`**  ERROR: Encountered error when getting CircleCI artifacts tests-summary.json: ${url}.`, e)
+      throw e
+    }
+  }, urls))
+
+  return tests
+}
+
+module.exports = {
+  getFailedTestNamesFromArtifacts
+}
--- a/integration-tests/slack/notify-circleci-test-results.js
+++ b/integration-tests/slack/notify-circleci-test-results.js
@@ -21,7 +21,6 @@ const getAllSlackChannelsForFailedJob = () => {
   const allChannelsAndTests = new Map()
   _.forEach(item => {
     const channelId = _.get('id', item)
-    // const { tests } = item
     const testNames = _.map(test => test.name, item.tests)
     allChannelsAndTests.set(channelId, testNames)
   }, testsInfo.fail)
@@ -61,7 +60,6 @@ const notifyCircleCITestResults = async () => {
   if (failedTestNames.length === 0) {
     // Slack notification: CircleCI job succeeded
     const channelIDs = getAllSlackChannelsForPassedJob()
-    console.log('channelIDs:', channelIDs)
     const messageBlocks = getMessageBlockTemplate(failedTestNames)
     // Using _.forEach to post same message to each Slack channels separately because a Slack issue: No way to post the same message to multiple channels at once.
     // See: https://github.com/slackapi/bolt-js/issues/696
--- a/integration-tests/slack/notify-circleci-test-results.js
+++ b/integration-tests/slack/notify-circleci-test-results.js
@@ -1,0 +1,71 @@
+const fs = require('fs')
+const _ = require('lodash/fp')
+const { getFailedTestNamesFromArtifacts } = require('./circleci-utils')
+const { getMessageBlockTemplate } = require('./message-templates')
+const { postMessage } = require('./post-message')
+
+
+/**
+ * Get a map object of Slack channel IDs and array of test names for failed tests notification
+ * @param { Array[string] } failedTests
+ * @returns { Map<string, Array[string]> } A map object where key is channel_id, value is test_names array
+ */
+const getFailedTestsAndChannelIDs = failedTests => {
+  const { fail: jsonBlock } = JSON.parse(fs.readFileSync('./slack/slack-notify-channels.json', 'utf8'))
+  const idsAndNames = new Map() // Map<string, Array[string]>
+  _.forEach(item => {
+    const channelId = _.get('id', item)
+    const { tests } = item
+    const testNames = _.map(test => test.name, tests)
+    idsAndNames.set(channelId, testNames)
+  }, jsonBlock)
+
+  const filteredIDsAndNames = new Map() // Map<string, Array[string]>
+  _.forEach(test => {
+    const channelIdsForFailedTest = Array.from(idsAndNames.keys())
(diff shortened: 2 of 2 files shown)


integration-tests/slack/circleci-utils.js
```diff
@@ -0,0 +1,67 @@
...
+    return _.map(_.get('url'), testSummaryArtifacts)
+  } catch (e) {
+    console.error(`**  ERROR: Encountered error when getting CircleCI JOB_BUILD_NUM: ${buildNum} artifacts.`, e)
+    throw e
+  }
+}
+
+/**
+ *
+ * @param { string } aggregatedResults A JSON object
```
nawatts: The documented type ("string") and the description ("A JSON object") conflict. Also, "A JSON object" isn't really helpful. It doesn't give a reader any more information than a correct type annotation would.

integration-tests/slack/circleci-utils.js
```diff
@@ -0,0 +1,67 @@
...
+  )(aggregatedResults.testResults)
+}
+
+/**
+ *
+ * @returns { Promise<Array[string]> }
+ */
+const getFailedTestNamesFromArtifacts = async () => {
+  const tests = []
+  // Parse all tests-summary JSON to look for failed test names
```
nawatts: Since this comment describes the purpose of the function, it would be more useful in the JSDoc comment where it's more easily findable than in the body of the function.

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,71 @@
...
+const { postMessage } = require('./post-message')
+
+
+/**
+ * Get a map object of Slack channel IDs and array of test names for failed tests notification
+ * @param { Array[string] } failedTests
+ * @returns { Map<string, Array[string]> } A map object where key is channel_id, value is test_names array
+ */
+const getFailedTestsAndChannelIDs = failedTests => {
+  const { fail: jsonBlock } = JSON.parse(fs.readFileSync('./slack/slack-notify-channels.json', 'utf8'))
```
nawatts: FYI, JSON files can also be loaded using [require](https://nodejs.org/docs/latest-v16.x/api/modules.html#all-together). For a configuration file like this, it might be appropriate to require it once at the top of the file instead of reading it in multiple places.

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,71 @@
...
+    const channelIDs = getSlackChannelIDsForPassedTests()
+    const messageBlocks = getMessageBlockTemplate(failedTestNames)
+    // Post same message blocks for all Slack channels
+    _.forEach(async channelId => await postMessage({ channel: channelId, blocks: messageBlocks }), channelIDs)
+    return
+  }
+
+  // Slack notification: CircleCI job failed. Message contains list of failed test names.
+  const channelIDsAndNames = getFailedTestsAndChannelIDs(failedTestNames)
+  // Slack issue: No way to post the same message to multiple channels at once. https://github.com/slackapi/bolt-js/issues/696
```
nawatts: Why is this an issue? It seems like this is just the API that Slack provides. We're not trying to post the same message to multiple channels: the list of tests varies for each channel.

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,71 @@
...
+
+  return filteredIDsAndNames
+}
+
+/**
+ * Get array of Slack channel IDs for succeeded job notification
+ * @returns { Array[string] }
+ */
+const getSlackChannelIDsForPassedTests = () => {
+  const { pass: jsonBlock } = JSON.parse(fs.readFileSync('./slack/slack-notify-channels.json', 'utf8'))
```
nawatts: Is there a more descriptive name for this variable?

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,71 @@
...
+const { getMessageBlockTemplate } = require('./message-templates')
+const { postMessage } = require('./post-message')
+
+
+/**
+ * Get a map object of Slack channel IDs and array of test names for failed tests notification
+ * @param { Array[string] } failedTests
+ * @returns { Map<string, Array[string]> } A map object where key is channel_id, value is test_names array
+ */
+const getFailedTestsAndChannelIDs = failedTests => {
```
nawatts: Variable names in this function could use some improvement. `jsonBlock`, `idsAndNames`, `filteredIdsAndNames` are all very generic. Are there more descriptive names?

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,71 @@
...
+    const channelId = _.get('id', item)
+    const { tests } = item
+    const testNames = _.map(test => test.name, tests)
+    idsAndNames.set(channelId, testNames)
+  }, jsonBlock)
+
+  const filteredIDsAndNames = new Map() // Map<string, Array[string]>
+  _.forEach(test => {
+    const channelIdsForFailedTest = Array.from(idsAndNames.keys())
+      .filter(key => idsAndNames.get(key).includes(test) || idsAndNames.get(key).length === 0)
```
nawatts: Probably worth a comment here explaining the `idsAndNames.get(key).length === 0` case. Something about how no list of tests means notify that channel for any failing test.

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,71 @@
...
+    if (channelIdsForFailedTest.length === 0) {
+      throw new Error(`Test: ${test} was not found in slack-notify-channels.json`)
+    }
+    _.forEach(channelId => filteredIDsAndNames.has(channelId) ? filteredIDsAndNames.get(channelId).push(test) : filteredIDsAndNames.set(channelId, new Array(test)),
+      channelIdsForFailedTest
+    )
```
nawatts: This is opinion, but I think the arrow function and ternary suggest that the return value here is important, when really we care about the side effects (updating the map and array with set and push). I'll claim this would be clearer as something like:

```suggestion
    _.forEach(channelId => {
      if (!filteredIDsAndNames.has(channelId)) {
        filteredIDsAndNames.set(channelId, [])
      }
      filteredIDsAndNames.get(channelId).push(test)
    }, channelIdsForFailedTest)
```

nawatts (changes requested):

integration-tests/slack/circleci-utils.js
aweng98: Updated jsDoc with a link to the type definition. Hope it helps.
```
@param {@link https://github.com/facebook/jest/blob/240587bde5dae1467ced0fdeee2e668e01caf896/packages/jest-test-result/src/types.ts#L77 AggregatedResult} aggregatedResults Results from the test run.
```

Screen Shot 2022-08-09 at 4 25 37 PM

integration-tests/slack/circleci-utils.js
aweng98: Yes.

integration-tests/slack/notify-circleci-test-results.js
aweng98: Yes

integration-tests/slack/notify-circleci-test-results.js
aweng98: Yes.

integration-tests/slack/notify-circleci-test-results.js
aweng98: Right. This comment explains why `._forEach` is used to send a message to every channel separately. 
It's valid but I had it at the wrong place (copy/past mistake). It shoud be in `if (failedTestNames.length === 0)` block.

integration-tests/slack/notify-circleci-test-results.js
aweng98: Made an attempt. Please let me know what you think about new changes.

integration-tests/slack/circleci-utils.js
nawatts: Nice. I would have settled for `{Object}`.

integration-tests/slack/notify-circleci-test-results.js
nawatts: Ah, that makes more sense in the success notifications.

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Get all Slack channel IDs and test names for failed job notification
+ * @returns {Map<string, Array[string]>}
+ */
+const getAllSlackChannelsForFailedJob = () => {
+  const allChannelsAndTests = new Map()
+  _.forEach(item => {
+    const channelId = _.get('id', item)
+    // const { tests } = item
```
nawatts: This can be removed.

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,82 @@
...
+const getFailedTestsAndChannelIDs = failedTests => {
+  const allChannelsAndTests = getAllSlackChannelsForFailedJob()
+
+  const filteredIncludeOnlyFailedTests = new Map() // Map<string, Array[string]>
+  _.forEach(test => {
+    const channelIdsForFailedTest = Array.from(allChannelsAndTests.keys())
+      .filter(key => allChannelsAndTests.get(key).includes(test) || allChannelsAndTests.get(key).length === 0) // A channel without a list of tests means notify that channel for any failing test
+    if (channelIdsForFailedTest.length === 0) {
+      throw new Error(`Test: ${test} was not found in slack-notify-channels.json`)
+    }
```
nawatts: Since the current configuration always notifies the dsde-qa channel for any failing test, will this condition ever be true?

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// Post CircleCI UI test report to Slack channels
+const notifyCircleCITestResults = async () => {
+  const failedTestNames = await getFailedTestNamesFromArtifacts()
+
+  if (failedTestNames.length === 0) {
+    // Slack notification: CircleCI job succeeded
+    const channelIDs = getAllSlackChannelsForPassedJob()
+    console.log('channelIDs:', channelIDs)
```
nawatts: Is this needed? It looks like it's maybe left over from development.

nawatts (approved):

integration-tests/slack/notify-circleci-test-results.js
aweng98: Done.

integration-tests/slack/notify-circleci-test-results.js
aweng98: Ah, I think we still need it. This check is good for ensuring JSON should also be updated when someone rename or add new test.
For example, Workspace team added a new test. No update to `slack-notify-channels.json`. When this test fails in future, `#dsde-qa` is the only channel that will receive Slack notification. Team that owns this test will never know until manually be told or update JSON in new PR.

integration-tests/slack/notify-circleci-test-results.js
aweng98: No.

integration-tests/slack/notify-circleci-test-results.js
aweng98: Oops, I'll check on this. You may be right about condition is never true.

aweng98: I just had an idea. I will remove this check now and create a new PR to add a func that checks whether failed test is in json or not.

nawatts: Sounds good. If we want to require all tests be configured to notify a specific Slack channel, it would be nice if that check happened before integration tests ran.

aweng98:
@breilly2 @cahrens Hi, can you take a look? I need two approvals. thank you.

cahrens:
I'm deferring to @breilly2 as the assigned reviewer (I'm in the midst of other reviews).

integration-tests/slack/circleci-utils.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  // For more arguments and details of the response, see: https://circleci.com/docs/api/v2/index.html#operation/getJobArtifacts
+  const apiUrlRoot = 'https://circleci.com/api/v2/project/github/DataBiosphere/terra-ui'
+  try {
+    // Because terra-ui is a public repository on GitHub, API token is not required. See: https://circleci.com/docs/oss#security
+    const response = await fetch(`${apiUrlRoot}/${buildNum}/artifacts`)
+    const { items } = await response.json()
+    const testSummaryArtifacts = _.filter(_.flow(_.get('path'), _.includes('tests-summary')), items)
+    return _.map(_.get('url'), testSummaryArtifacts)
```
breilly2: Nit: FYI, [`_.map`](https://lodash.com/docs/4.17.15#map) supports a `_.property` shorthand, so you can simply do:
```suggestion
    return _.map('url', testSummaryArtifacts)
```
(Note: this is from memory and reading the docs; I didn't actually test this suggestion)

integration-tests/slack/circleci-utils.js
```diff
@@ -0,0 +1,66 @@
...
+  await Promise.all(_.map(async url => {
+    try {
+      const response = await fetch(url)
+      const failedTests = getFailedTestNames(await response.json())
+      tests.push(...failedTests)
+    } catch (e) {
+      console.error(`**  ERROR: Encountered error when getting CircleCI artifacts tests-summary.json: ${url}.`, e)
+      throw e
+    }
+  }, urls))
```
breilly2: Nit:
```suggestion
  const urls = await fetchJobArtifacts()
  const tests = _.flatten(await Promise.all(_.map(async url => {
    try {
      const response = await fetch(url)
      return getFailedTestNames(await response.json())
    } catch (e) {
      console.error(`**  ERROR: Encountered error when getting CircleCI artifacts tests-summary.json: ${url}.`, e)
      throw e
    }
  }, urls)))
```
Again, I didn't test this, but something like this should work if you want to do this in a functional way that avoids mutable data structures.

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,77 @@
...
+const { postMessage } = require('./post-message')
+const testsInfo = require('./slack-notify-channels.json')
+
+
+/**
+ * Get array of Slack channel IDs for succeeded job notification
+ * @returns {Array[string]}
+ */
+const getAllSlackChannelsForPassedJob = () => {
+  return _.map(_.get('id'), testsInfo.pass)
```
breilly2: Nit: Again, I think this shorthand will work.
```suggestion
  return _.map('id', testsInfo.pass)
```

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,77 @@
...
+ * @returns {Map<string, Array[string]>}
+ */
+const getAllSlackChannelsForFailedJob = () => {
+  const allChannelsAndTests = new Map()
+  _.forEach(item => {
+    const channelId = _.get('id', item)
+    const testNames = _.map(test => test.name, item.tests)
+    allChannelsAndTests.set(channelId, testNames)
+  }, testsInfo.fail)
+  return allChannelsAndTests
```
breilly2, line 27: Nit: this could probably also be written something like:
```suggestion
  return _.fromPairs(_.map(item => [item.id, _.map('name', item.tests)], testsInfo.fail))
```

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Get a map object of Slack channel IDs and array of test names for failed tests notification
+ * @param {Array[string]} failedTests
+ * @returns {Map<string, Array[string]>} A map object where key is channel_id, value is test_names array
+ */
+const getFailedTestsAndChannelIDs = failedTests => {
+  const allChannelsAndTests = getAllSlackChannelsForFailedJob()
+
+  const filteredIncludeOnlyFailedTests = new Map() // Map<string, Array[string]>
```
breilly2, line 38: Nit: I leave this one as an exercise to the reader. :)
In general, `_.forEach` is fine if you want some side effect. If instead you have a list of things that you want to transform into a list of other things, `_.map` is a more self-documenting choice.

integration-tests/slack/notify-circleci-test-results.js
```diff
@@ -0,0 +1,77 @@
...
+const notifyCircleCITestResults = async () => {
+  const failedTestNames = await getFailedTestNamesFromArtifacts()
+
+  if (failedTestNames.length === 0) {
+    // Slack notification: CircleCI job succeeded
+    const channelIDs = getAllSlackChannelsForPassedJob()
+    const messageBlocks = getMessageBlockTemplate(failedTestNames)
+    // Using _.forEach to post same message to each Slack channels separately because a Slack issue: No way to post the same message to multiple channels at once.
+    // See: https://github.com/slackapi/bolt-js/issues/696
+    _.forEach(async channelId => await postMessage({ channel: channelId, blocks: messageBlocks }), channelIDs)
```
breilly2: Here (and below) is a perfect example to support my other comment. Here `_.forEach` is appropriate[*] because you want a set of side-effects for each item in the list.

[*] One could argue that `await Promise.all(_.map(async channelId => ...))` would be better because the order of execution of the side-effects doesn't matter.

breilly2 (approved):
👍 This all looks fine to me. I made various suggestions about making the code style more functional than imperative, but:

1. I didn't test any of my code suggestions, and therefore it's possible that there are specific reasons why my suggestions aren't appropriate because of some detail that I missed.
2. None of this is critical to my approval of the code.

Also, please accept my apology for the review delay. I've had more distractions than I expected to the past few days.

integration-tests/slack/circleci-utils.js
aweng98: I had no idea. Suggested code works!

integration-tests/slack/notify-circleci-test-results.js
aweng98: Done.

integration-tests/slack/circleci-utils.js
aweng98: Good stuff!

integration-tests/slack/notify-circleci-test-results.js
aweng98: Unchanged after tested this because it returns a Dictionary instead of a Map object.

aweng98 (commented):
thank you.

PR #3303, opened 2022-08-09
[WOR-149] Add cloud context icon and last workflow status

ghost:
https://broadworkbench.atlassian.net/browse/WOR-149 
Cloud Context Icon: 
Screen Shot 2022-08-09 at 3 27 42 PM
Cloud Context Icon with Last Workflow Submitted Status:
Screen Shot 2022-08-09 at 3 34 37 PM

<!--
Hello, friend!
Remember to mention what sort of testing/verification you performed in the course of building this PR, i.e. checking functionality in the browser locally.
Also, so if a screen recording [1] and/or screenshots would be a helpful way to communicate context, please consider including some in your PR description.
Thanks!

[1] https://support.apple.com/guide/quicktime-player/record-your-screen-qtp97b08e666/mac
--!>

PR #3305, opened 2022-08-10
[WOR-149] Add cloud context icon

ghost:
https://broadworkbench.atlassian.net/browse/WOR-149
Screen Shot 2022-08-10 at 2 14 47 PM
Screen Shot 2022-08-10 at 2 15 09 PM

<!--
Hello, friend!
Thanks!

--!>


src/pages/workspaces/List.js
```diff
@@ -251,19 +253,47 @@ export const WorkspaceList = () => {
             ])
           },
           size: { basis: 120, grow: 1, shrink: 0 }
+        }, {
+          headerRenderer: () => div({ className: 'sr-only' }, ['Last Workflow Submitted Status']),
+          cellRenderer: ({ rowIndex }) => {
+            const { ...workspace } = sortedWorkspaces[rowIndex]
```
nawatts: For clarity:
```suggestion
            const workspace = sortedWorkspaces[rowIndex]
```

src/pages/workspaces/List.js
```diff
@@ -251,19 +253,47 @@ export const WorkspaceList = () => {
...
         }, {
           headerRenderer: () => div({ className: 'sr-only' }, ['Actions']),
           cellRenderer: ({ rowIndex }) => {
-            const { accessLevel, workspace: { workspaceId, namespace, name }, ...workspace } = sortedWorkspaces[rowIndex]
+            const { accessLevel, workspace: { workspaceId, namespace, name } } = sortedWorkspaces[rowIndex]
             if (!Utils.canRead(accessLevel)) {
-              // No menu shown if user does not have read acccess.
-              return null
+              // No menu shown if user does not have read access.
+              return div({ className: 'sr-only' }, ['Action menu is not available due to access level.'])
```
nawatts: Do we need this message? For users that don't use screen readers, we don't have any equivalent message and we simply don't render any content here.

cahrens: If you don't have it, you get "Blank". Plus it renders in an unattractive way. See https://github.com/DataBiosphere/terra-ui/pull/3303#discussion_r942526064

cahrens:
@yunip20 can you also change the "recently viewed workspace card" to use the more concise icons?

src/pages/workspaces/List.js
nawatts: Ah, ok.

This could possibly be reworded to be more similar to other permission messages.

```suggestion
              return div({ className: 'sr-only' }, ['You do not have permission to perform actions on this workspace.'])
```

https://github.com/DataBiosphere/terra-ui/blob/9e6b5554b1425c8f1b1dcbdd0e4f1466eb440a61/src/libs/utils.js#L189-L202

cahrens:
Nit, do we want icon sizes to be consistent between the cards and the list view? FYI @MatthewBemis 

image

src/components/workspace-utils.js
```diff
@@ -417,12 +417,12 @@ export const WorkspaceSubmissionStatusIcon = ({ status, loadingSubmissionStats,
       side: 'left'
     }, [
       Utils.switchCase(status,
-        ['success', () => icon('success-standard', { size, style: { color: colors.success() } })],
-        ['failure', () => icon('error-standard', { size, style: { color: colors.danger(0.85) } })],
-        ['running', () => icon('sync', { size, style: { color: colors.success() } })]
+        ['success', () => icon('success-standard', { size, style: { color: colors.success() }, 'aria-label': 'Workflow Status Success' })],
+        ['failure', () => icon('error-standard', { size, style: { color: colors.danger(0.85) }, 'aria-label': 'Workflow Status Failure' })],
+        ['running', () => icon('sync', { size, style: { color: colors.success() }, 'aria-label': 'Workflow Status Running' })]
```
nawatts, line 422: I would have thought that TooltipTrigger's useAsLabel should provide a label here. With this aria-label, it seems like VoiceOver also picks up the aria-describedby added by TooltipTrigger, and reads:
"Last Workflow Submitted Status" (column header)
"Workflow Status Success" (aria label)
"Last submitted workflow status: Success" (aria describedby)

https://github.com/DataBiosphere/terra-ui/blob/9e6b5554b1425c8f1b1dcbdd0e4f1466eb440a61/src/components/TooltipTrigger.js#L109

nawatts (approved):

src/components/workspace-utils.js
cahrens: Yes, it's reading all 3. Slightly wordy, but better than nothing being read out.

cahrens (approved):

src/components/workspace-utils.js
nawatts: I was curious why it's read with aria-label but not aria-labelledby. Maybe something to do with aria-hidden on the icon.

ghost: If I change `aria-label` to `aria-labelledby`, VoiceOver only reads the column header ("Last Workflow Submitted Status") and the tooltip ("Last submitted workflow status: Success"), but when I check with React Devtools, `aria-labelledby` is set to undefined. Should I use `aria-labelledby` instead of `aria-label`? 
Screen Shot 2022-08-10 at 3 44 50 PM

```
Utils.switchCase(status,
        ['success', () => icon('success-standard', { size, style: { color: colors.success() }, 'aria-labelledby': 'Workflow Status Success' })],
        ['failure', () => icon('error-standard', { size, style: { color: colors.danger(0.85) }, 'aria-labelledby': 'Workflow Status Failure' })],
        ['running', () => icon('sync', { size, style: { color: colors.success() }, 'aria-labelledby': 'Workflow Status Running' })]
      )
```

nawatts: An `aria-labelledby` attribute should have the ID of another element that contains label text, not the label text itself. 

I think this is fine as is. Up to you whether you want to dig into it or not.

cahrens: `aria-labelledby` would have to reference the ID of another component. For instance, if there was a visible label beside the icon, you could set the icon to be labelled by that other element. It is showing as undefined because the text you specified is not the ID of an element in the dom.

ghost: Thanks, that makes sense!

PR #3306, opened 2022-08-10
[PROD-718] Changed occurrences (for apps) of googleProject to cloudContext.cloudResource

n8landolt:
## Description

Starting and stopping apps is currently broken in production.

The reason being is the app property "googleProject" is now an object, when it should be a string.

### DIFF
--- a/src/components/GalaxyModal.js
+++ b/src/components/GalaxyModal.js
@@ -64,7 +64,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
       Utils.withBusyState(setLoading),
       withErrorReportingInModal('Error deleting galaxy instance', onError)
     )(async () => {
-      await Ajax().Apps.app(app.googleProject, app.appName).delete(attachedDataDisk ? shouldDeleteDisk : false)
+      await Ajax().Apps.app(app.cloudContext?.cloudResource, app.appName).delete(attachedDataDisk ? shouldDeleteDisk : false)
       Ajax().Metrics.captureEvent(Events.applicationDelete, { app: 'Galaxy', ...extractWorkspaceDetails(workspace) })
       return onSuccess()
     })
@@ -73,7 +73,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
       Utils.withBusyState(setLoading),
       withErrorReportingInModal('Error stopping galaxy instance', onError)
     )(async () => {
-      await Ajax().Apps.app(app.googleProject, app.appName).pause()
+      await Ajax().Apps.app(app.cloudContext?.cloudResource, app.appName).pause()
       Ajax().Metrics.captureEvent(Events.applicationPause, { app: 'Galaxy', ...extractWorkspaceDetails(workspace) })
       return onSuccess()
     })
@@ -82,7 +82,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
       Utils.withBusyState(setLoading),
       withErrorReportingInModal('Error starting galaxy instance', onError)
     )(async () => {
-      await Ajax().Apps.app(app.googleProject, app.appName).resume()
+      await Ajax().Apps.app(app.cloudContext?.cloudResource, app.appName).resume()
       Ajax().Metrics.captureEvent(Events.applicationResume, { app: 'Galaxy', ...extractWorkspaceDetails(workspace) })
       return onSuccess()
     })
--- a/src/components/GalaxyModal.js
+++ b/src/components/GalaxyModal.js
@@ -64,7 +64,10 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
       Utils.withBusyState(setLoading),
       withErrorReportingInModal('Error deleting galaxy instance', onError)
     )(async () => {
-      await Ajax().Apps.app(app.cloudContext?.cloudResource, app.appName).delete(attachedDataDisk ? shouldDeleteDisk : false)
+      if (!app.cloudContext) {
+        throw new Error('No galaxy app found.')
+      }
+      await Ajax().Apps.app(app.cloudContext.cloudResource, app.appName).delete(attachedDataDisk ? shouldDeleteDisk : false)
       Ajax().Metrics.captureEvent(Events.applicationDelete, { app: 'Galaxy', ...extractWorkspaceDetails(workspace) })
       return onSuccess()
     })
@@ -73,7 +76,10 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
       Utils.withBusyState(setLoading),
       withErrorReportingInModal('Error stopping galaxy instance', onError)
     )(async () => {
-      await Ajax().Apps.app(app.cloudContext?.cloudResource, app.appName).pause()
+      if (!app.cloudContext) {
+        throw new Error('No galaxy app found.')
+      }
+      await Ajax().Apps.app(app.cloudContext.cloudResource, app.appName).pause()
       Ajax().Metrics.captureEvent(Events.applicationPause, { app: 'Galaxy', ...extractWorkspaceDetails(workspace) })
       return onSuccess()
     })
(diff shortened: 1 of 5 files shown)


gpcarr (approved):
:shipit:

src/components/GalaxyModal.js
```diff
@@ -64,7 +64,7 @@ export const GalaxyModalBase = withDisplayName('GalaxyModal')(
       Utils.withBusyState(setLoading),
       withErrorReportingInModal('Error deleting galaxy instance', onError)
     )(async () => {
-      await Ajax().Apps.app(app.googleProject, app.appName).delete(attachedDataDisk ? shouldDeleteDisk : false)
+      await Ajax().Apps.app(app.cloudContext?.cloudResource, app.appName).delete(attachedDataDisk ? shouldDeleteDisk : false)
```
cahrens: What happens when you call this method with `undefined` as the first argument?

src/components/RuntimeManager.js
```diff
@@ -128,7 +128,7 @@ export const AppErrorModal = ({ app, onDismiss }) => {
     withErrorReporting('Error loading app details'),
     Utils.withBusyState(setLoadingAppDetails)
   )(async () => {
-    const { errors: appErrors } = await Ajax().Apps.app(app.googleProject, app.appName).details()
+    const { errors: appErrors } = await Ajax().Apps.app(app.cloudContext.cloudResource, app.appName).details()
```
cahrens, line 131: It looks like in some places you are confident that `app.cloudContext` will be defined, and in other places you are not confident. What is the difference?

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
```diff
@@ -147,18 +148,19 @@ export const CloudEnvironmentModal = ({
 
   // We assume here that button disabling is working properly, so the only thing to check is whether it's an app or the current (assumed to be existing) runtime
   const startApp = toolLabel => Utils.cond([isToolAnApp(toolLabel), () => {
-    const { googleProject, appName } = currentApp(toolLabel)
+    const { appName, cloudContext } = currentApp(toolLabel)
     executeAndRefresh(toolLabel,
-      Ajax().Apps.app(googleProject, appName).resume())
+      //TODO: Potential issue in leonardo needs to be fixed.
```
cahrens: What is the TODO here referencing?

src/components/RuntimeManager.js
n8landolt: The reason we need it is the field comes from an API request and we aren't always going to get 200s

src/pages/workspaces/workspace/notebooks/modals/CloudEnvironmentModal.js
n8landolt: That there may be an issue in leo that was the root of it - however, I'm planning to remove the TODO because we want to move to using cloudContext over googleProject regardless.

src/components/GalaxyModal.js
n8landolt: This will take some time to test... not sure how long galaxy envs take to start after being paused.

n8landolt: Screen Shot 2022-08-10 at 4 45 19 PM

cahrens: If you are going to allow for this to be `undefined`, you need to ensure that the code does something logical in that case. I don't see that `undefined` is a reasonable option for this argument given that it is used as `project` in `api/google/v1/apps/${project}/${name}`.

n8landolt: I'm not sure what would be an appropriate argument, I figure getting that error is better than the whole UI stopping due to a missing property?

cahrens: I think you need to make sure you don't attempt to call `Ajax().Apps.app(app.cloudContext?.cloudResource, app.appName).delete` when `app.cloudContext` is undefined.

kyuksel: > This will take some time to test... not sure how long galaxy envs take to start after being paused.

May want to mock the Ajax response to be able to iterate faster since Galaxy-related operations tend to take a long time in the backend.

kyuksel: > I'm not sure what would be an appropriate argument, I figure getting that error is better than the whole UI stopping due to a missing property?

Perhaps but sometimes not (e.g. the UI could do some undesirable stuff, which can go unnoticed because there aren't obvious errors right away). Anytime using the optional chaining operator (`?`), it's worth thinking if we really need to use it or if it can be a code smell (for the reason I mentioned; i.e., wrong things can be performed without noticing soon). Ideally, it should only be used for things outside the UI's control, like API responses. For those cases, if we could do the validation as close to the source (e.g. API response fetching), we could prevent proliferation of `?` downstream.

cahrens: To be clear, `app.cloudContext` can legitimately be undefined. And therefore the code needs to take that into account and prevent the user from calling this Ajax method while it is undefined. That may involve something like disabling the delete button… I haven't looked at the surrounding code to know what would be reasonable. But at the point where the Ajax call is made, you should be confident that `app.cloudContext` is not undefined because you have already checked for that condition.

n8landolt: From what I understand, if we have an app (or runtime) cloudContext should never be undefined, and we already have a null check on app for these methods I'm updating.

The main reason I decided to add the optional operator is because there have been inconsistencies on the backend and I wasn't sure that I could "trust" that "cloudResource" would be the correct name.

I don't believe app.cloudContext could ever be undefined - but this change stems from a backend issue where a response was changed.

I'm just hesitant to add extra bloat to the code with a new if statement check when it's already essentially handled.

src/components/RuntimeManager.js
n8landolt: I actually think my above comment is incorrect. If we get an app, we are supposed to have a cloudContext and a cloudResource. The reason for the optional chaining was because this whole issue stemmed from a backend bug.

src/components/GalaxyModal.js
cahrens: @n8landolt I'm confused… from your screenshot, it seemed to indicate that you did find a condition in which it is undefined. Can you explain more about the screenshot?

Also, `app.cloudContext?.cloudResource` doesn't help you at all if `cloudResource` is not the correct name. What is says is that if `app.cloudContext` is falsey (undefined/null), then return undefined from `app.cloudContext?.cloudResource` instead of blowing up. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
```diff
@@ -83,7 +83,7 @@ const DeleteWorkspaceModal = ({ workspace: { workspace: { namespace, name, bucke
       await Ajax().Workspaces.workspace(namespace, name).delete()
       if (isGoogleWorkspace) {
         await Promise.all(
-          _.map(async app => await Ajax().Apps.app(app.googleProject, app.appName).delete(), deletableApps)
+          _.map(async app => await Ajax().Apps.app(app.cloudContext?.cloudResource, app.appName).delete(), deletableApps)
```
cahrens: This still has the optional chaining.

src/components/GalaxyModal.js
n8landolt: The screenshot was from me forcing the property to be undefined

n8landolt: As far as I know, the property is supposed to be cloudContext.cloudResource returned from the API, as it stands it is the correct property

src/pages/workspaces/workspace/DeleteWorkspaceModal.js
n8landolt: Whoops, didn't CMD+S them

cahrens (approved):
Code changes look good-- I have not tested them though. I am counting on IA folks to do the testing.

PR #3307, opened 2022-08-10
[SU-184] Separate attribute type input

nawatts:
We allow editing an attribute for multiple rows in a data table at once. However, currently the only option is to set the attribute to the same value in every row. We would like to add the option to change the type of the attribute, converting each row's value to the new type individually.

Currently, there is a component (AttributeInput) for editing an attribute value, including its type. This separates the type inputs out so they can be used without the value inputs.

Screen Shot 2022-08-10 at 4 29 13 PM

### DIFF
--- a/src/components/data/_tests/data-utils.test.js
+++ b/src/components/data/_tests/data-utils.test.js
@@ -3,7 +3,7 @@ import '@testing-library/jest-dom'
 import { fireEvent, render } from '@testing-library/react'
 import _ from 'lodash/fp'
 import { h } from 'react-hyperscript-helpers'
-import { AttributeInput, concatenateAttributeNames, convertAttributeValue, entityAttributeText, getAttributeType, prepareAttributeForUpload, renderDataCell } from 'src/components/data/data-utils'
+import { AttributeInput, AttributeTypeInput, concatenateAttributeNames, convertAttributeValue, entityAttributeText, getAttributeType, prepareAttributeForUpload, renderDataCell } from 'src/components/data/data-utils'
 import * as Utils from 'src/libs/utils'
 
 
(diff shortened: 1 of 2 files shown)


src/components/data/_tests/data-utils.test.js
```diff
@@ -428,3 +429,304 @@ describe('prepareAttributeForUpload', () => {
     })
   })
 })
+
+describe('AttributeTypeInput', () => {
+  it('renders radio buttons for available types', () => {
```
dmohs, line 434: This looks suspiciously like it is testing React's ability to render, which may be a valid thing to test, but I wouldn't do so for each component. I would add a comment here that answers the question, "What value am I getting for the cost of having to update the code in two places if I want to add another option?" Example: "Some options are conditional, so this ensures only the default options are present."

src/components/data/_tests/data-utils.test.js
```diff
@@ -428,3 +429,304 @@ describe('prepareAttributeForUpload', () => {
...
+  it('calls onChange callback when a type is selected', () => {
+    const onChange = jest.fn()
+    const { getByLabelText } = render(h(AttributeTypeInput, {
+      value: { type: 'string' },
+      onChange
+    }))
+
+    fireEvent.click(getByLabelText('Number'))
+
+    expect(onChange).toHaveBeenCalledWith({ type: 'number' })
```
dmohs, line 467: In my opinion, this is a decent alternative to TypeScript for improving the visibility of a contract. I hope to find opportunities to steal this technique.

src/components/data/_tests/data-utils.test.js
```diff
@@ -428,3 +429,304 @@ describe('prepareAttributeForUpload', () => {
...
+  describe('references', () => {
+    it('renders an entity types menu for reference values', () => {
+      const { getByLabelText } = render(h(AttributeTypeInput, {
+        value: { type: 'reference', entityType: 'foo' },
+        entityTypes: ['foo', 'bar', 'baz'],
+        onChange: jest.fn()
+      }))
+
+      const entityTypeInput = getByLabelText('Referenced entity type:')
+      expect(entityTypeInput.getAttribute('role')).toBe('combobox')
```
dmohs, line 479: This looks like a way for us to enforce good practice regarding accessibility.

src/components/data/_tests/data-utils.test.js
```diff
@@ -428,3 +429,304 @@ describe('prepareAttributeForUpload', () => {
...
+    })
+
+    it('selecting reference type uses the default reference entity type is one is provided', () => {
```
dmohs:
```suggestion
    it('selecting reference type uses the default reference entity type if one is provided', () => {
```

src/components/data/_tests/data-utils.test.js
```diff
@@ -428,3 +429,304 @@ describe('prepareAttributeForUpload', () => {
...
+      const onChange = jest.fn()
+      const { getByLabelText } = render(h(AttributeInput, {
+        value: { itemsType: 'AttributeValue', items: ['foo', 'bar', 'baz'] },
+        onChange
+      }))
+
+      const listCheckbox = getByLabelText('Value is a list')
+      fireEvent.click(listCheckbox)
+
+      expect(onChange).toHaveBeenCalledWith('foo')
```
dmohs, line 687: Love the clarity of this, but I wonder if the behavior could be surprising.

src/components/data/data-utils.js
```diff
@@ -711,64 +770,22 @@ const AttributeInput = ({ autoFocus = false, value: attributeValue, initialValue
...
-                _.set('entityType', value, attributeValue)
-              onChange(newAttributeValue)
-            }
-          })
-        ])])
-      ])
-    ]),
+        onChange(newAttributeValue)
+      }
+    }),
```
dmohs, line 788: The length of this file compels me to wonder if this is worth a follow-up PR that pulls some code out into a new file.

dmohs (approved):
In my experience, unit tests have a high risk of not paying for themselves. These do not bring up that fear for me. Well done.

cahrens:
Some awesome examples of unit tests here @nawatts!

src/components/data/data-utils.js
```diff
@@ -585,6 +585,66 @@ export const getAttributeType = attributeValue => {
   return { type, isList }
 }
 
+export const AttributeTypeInput = ({ label: labelText = 'Type', value, onChange, entityTypes = [], defaultReferenceEntityType = null, showJsonTypeOption = false }) => {
+  const { type, entityType: referenceEntityType } = value
+
+  const typeOptions = [
+    { type: 'string' },
+    { type: 'reference', tooltip: 'A link to another entity' },
```
MatthewBemis: TOL: have we settled on what we want to actually call entities for user-facing text? (out of scope for this PR)

MatthewBemis (approved):

src/components/data/_tests/data-utils.test.js
nawatts: 🤔 Tests like this might be better accomplished with Jest's [snapshot testing](https://jestjs.io/docs/snapshot-testing), which would reduce the cost of having to update the test code. I think tests like these function better as documentation than snapshot tests. But, as with comments, I suppose there's an argument to be made here that the intent is obvious enough that it doesn't need explanation.

src/components/data/_tests/data-utils.test.js
nawatts: Testing Library highlights accessibility in their recommendation for which queries to use in tests: https://testing-library.com/docs/queries/about#priority.

src/components/data/_tests/data-utils.test.js
nawatts: What would the alternative be? I suppose string lists could be concatenated to convert them to a single value, but for other types I can't think of anything that would be clearer than truncation. And for what it's worth, the impact of surprising behavior is somewhat limited since the user can see the new value before saving it to the data table.

src/components/data/data-utils.js
nawatts: Definitely. I'm all for breaking up the `*-utils` modules into more cohesive pieces.

src/components/data/data-utils.js
nawatts: Yeah, I think the design attached to [SU-99](https://broadworkbench.atlassian.net/browse/SU-99) updated the user facing labels. "Row" would probably be most consistent with the rest of the data table UI.

nawatts: Updated in https://github.com/DataBiosphere/terra-ui/pull/3307/commits/7c047d4d3c0851a00c2f40abc38a0ac0c0ecb036.

src/components/data/_tests/data-utils.test.js
nawatts: I think tests like this are good for ensuring that a component continues to adhere to a contract. But I'll claim that if/when a contract changes, TypeScript makes it much easier to check that all consumers have been updated to the new contract.

src/components/data/_tests/data-utils.test.js
dmohs: This comment was unhelpful, sorry. I should have suggested an alternative. If I had tried, I would have discovered that I didn't have any good ideas.

PR #3309, opened 2022-08-11
[DC-487] Add catalog export case to import-data

s-rubenstein:
<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1097,7 +1097,7 @@ const Catalog = signal => ({
     return res.json()
   },
   exportDataset: async ({ id, workspaceId }) => {
-    return await fetchCatalog(`v1/datasets/${id}/export/${workspaceId}`, _.merge(authOpts(), { signal }))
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
Well, that's convenient! Do we need unit test mocks for this functionality?

pshapiro4broad (approved):
Looks good. I assume export for snapshots will be a bit more messy...

pshapiro4broad:
So the code that uses this new export case will be a separate PR? Is there a way to test this in the UI at this point? Or were we already calling this code with `catalog`?

breilly2 (approved):
👍 I'll confess that I didn't test this because I don't have enough familiarity with the Data Catalog domain, but the code looks fine to me.

s-rubenstein:
@pshapiro4broad yeah, exporting snapshots will be much more tricky! This is not currently called with the catalog, although I could test this by hand-rolling a URL with the right query params. I may try to do that.

@fboulnois its unclear to me what we do for unit tests rn in the UI. I would say we may want to add a test similar to what we have for other tests for this, but we need to wait for that until the button works.

s-rubenstein:
I have manually tested now with both existing and new workspace, and the new code works

PR #3310, opened 2022-08-11
[IA-3634][IA-3636][IA-3652] misc bug fixes

jdcanas:
All attached tickets are subtasks of this ticket https://broadworkbench.atlassian.net/browse/IA-3279

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -736,7 +736,7 @@ export const ComputeModalBase = ({
 
       const locationType = getLocationType(location)
       const { computeZone, computeRegion } = getRegionInfo(location || defaultLocation, locationType)
-      const runtimeConfig = currentRuntimeDetails?.runtimeConfig
+      const runtimeConfig = currentRuntimeDetails?.runtimeConfig || computeConfig
       const gpuConfig = runtimeConfig?.gpuConfig
       const autopauseThresholdCalculated = !!currentRuntimeDetails ? currentRuntimeDetails.autopauseThreshold : defaultAutopauseThreshold
       const newRuntimeType = Utils.switchCase(runtimeConfig?.cloudService,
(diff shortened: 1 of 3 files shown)


src/components/ComputeModal.js
```diff
@@ -1468,18 +1468,18 @@ export const ComputeModalBase = ({
...
               'To reduce the size of the PD, the existing PD will be deleted and a new one will be created and attached to your virtual machine instance. This will ',
               span({ style: { fontWeight: 600 } }, ['delete all files on the disk.'])
             ]),
-            existingRuntime.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
+            existingRuntime?.tool === 'RStudio' ? h(SaveFilesHelpRStudio) : h(SaveFilesHelp)
           ])],
           [willRequireDowntime(), () => h(Fragment, [
-            existingRuntime.tool !== desiredTool ?
-              p(['By continuing, you will be changing the application of your cloud environment from ', strong([existingRuntime.tool]), ' to ',
+            existingRuntime && existingRuntime.tool !== desiredTool ?
```
nawatts, line 1481: Could this use optional chaining like the other `existingRuntime.tool` conditions?

src/components/ComputeModal.js
```diff
@@ -1468,18 +1468,18 @@ export const ComputeModalBase = ({
...
+              p(['By continuing, you will be changing the application of your cloud environment from ', strong([existingRuntime?.tool]), ' to ',
```
nawatts, line 1482: This is in a branch that's conditional on `existingRuntime` being truthy, so optional chaining is not necessary.

nawatts (approved):
Did not test. Could not test because neither the PR description nor the tickets describes what the bugs are or how to reproduce them.

In Terra UI, since at least one PR reviewer is unlikely to be from the same team, please consider including some context for the change in the PR description.

Qi77Qi (approved):

src/components/ComputeModal.js
jdcanas: I thought about it, but i like that the others have a sensible defaults if there is no existing runtime

src/components/ComputeModal.js
jdcanas: Yes, but it seems code written not-defensively eventually gets refactored to a place where it bites us... I'd prefer to be overly cautious here.

nawatts: I would argue that this is not helpful for that purpose. Maybe it prevents a JS error (reading a property of undefined), but if `existingRuntime` is undefined here, then the output _is still incorrect_ because the tool name would be missing. It's an error in a different form. Given the choice between those two types of errors, I'll claim that it's preferable to fail loudly since it reduces the chances that the problem will go undetected during development or testing.

PR #3311, opened 2022-08-11
[QA-1984] Integration test: Add slack notification step in CircleCI config

aweng98:
https://broadworkbench.atlassian.net/browse/QA-1984
This is the final piece of the puzzle. Slack notification for failed UI tests will be complete after merge. 

Big thank you to Nick Watts for giving me this excellent idea.
> When each parallel job of integration-tests-staging/alpha finishes, it could check if the overall job has finished and, if so, send the Slack notification.

This PR does exactly that. I was lucky to find the right CircleCI API ([/getJobDetails](https://circleci.com/docs/api/v2/index.html#operation/getJobDetails)). API response tells status of individal nodes in a parallized run.

Created new `wait-for-job-finish.sh` to keep shell scripts out of `config.yml`.

This change enables `integration-tests-alpha` and `integration-tests-staging` to run parallelized and then send one Slack notification (currently `parallelism: 1` due to CircleCI Slack/orb limitation). Let me know if we want to increase their parallelism in this PR.


.circleci/config.yml
```diff
@@ -181,6 +181,16 @@ commands:
       - store_artifacts:
           path: << parameters.log_dir >>/<< parameters.env >>/test-results
           destination: results
+      - run:
+          working_directory: integration-tests
+          name: Test results Slack notification on Dev branch
+          command: |
+            if [ "$CIRCLE_NODE_INDEX" -eq 0 ] && [ "$CIRCLE_BRANCH" = "dev" ]; then
```
aweng98, line 188: Run this additional step only if `CIRCLE_NODE_INDEX` is 0 and on Dev branch because we can't have all nodes excute the same script.
Chose `CIRCLE_NODE_INDEX=0` because this node always exists: works for parallelism >= 1.

.circleci/config.yml
```diff
@@ -268,14 +278,12 @@ jobs:
     steps:
       - integration-tests:
           env: alpha
-      - notify-qa
```
aweng98, line 271: No longer need CircleCI Slack/orb for `integration-tests-alpha` and `integration-tests-staging` jobs.

.circleci/config.yml
aweng98: Screen Shot 2022-08-12 at 11 52 17 AM

.circleci/wait-for-job-finish.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# Exit script if you try to use an uninitialized variable.
+set -o nounset
+
+# Use the error status of the first failure, rather than that of the last item in a pipeline.
+set -o pipefail
+
+x=0
```
nawatts: This variable could use a more descriptive name.

.circleci/wait-for-job-finish.sh
```diff
@@ -0,0 +1,34 @@
...
+EXCLUDE_NODE_INDEX=0
```
nawatts: Could this variable be replaced with `CIRCLE_NODE_INDEX`? I think that might make it clearer that it's waiting for all other nodes to finish.

.circleci/wait-for-job-finish.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Wait up to 5 minutes for job to finish. A job can run on multiple nodes: parallelism > 1
+while [ $x -le 300 ]; do
+  # Find number of nodes in running
+  job_detail=$(curl -s "https://circleci.com/api/v2/project/github/DataBiosphere/terra-ui/job/$CIRCLE_BUILD_NUM" --header 'Circle-Token: "'$CIRCLECI_USER_TOKEN'"')
+  job_running_nodes_count=$(echo $job_detail | jq -r '.parallel_runs[] | select(.status == "running") | select(.index != '$EXCLUDE_NODE_INDEX')' | grep "running" | wc -l)
+  echo "job_running_nodes_count: $job_running_nodes_count"
```
nawatts: Do we need this output every 10 seconds?

.circleci/wait-for-job-finish.sh
```diff
@@ -0,0 +1,34 @@
...
+
+  if [ $job_running_nodes_count -eq 0 ]; then
+    exit 0
+  fi
+
+  echo "sleep 10 seconds"
```
nawatts: Is this output valuable?

.circleci/config.yml
nawatts: Would putting this in a `when` step instead of checking a condition in the script make it clearer what's happening in the CircleCI UI?

https://circleci.com/docs/configuration-reference#the-when-step-requires-version-21

.circleci/wait-for-job-finish.sh
```diff
@@ -0,0 +1,34 @@
...
+  sleep 10
+  x=$(($x + 10))
+done
+
+echo "Waited total $x seconds"
+date
+
+# Something is wrong. Log response for error troubleshooting
+curl -s "https://circleci.com/api/v2/project/github/DataBiosphere/terra-ui/job/$CIRCLE_BUILD_NUM" --header 'Circle-Token: "'$CIRCLECI_USER_TOKEN'"' | jq -r '.'
+echo "ERROR: Exceeded maximum waiting time 5 minutes is exceeded."
```
aweng98: I'll fix log in next commit.

nawatts (changes requested):
This isn't exactly what I had in mind. I was thinking that each parallel job could check if all others had finished and either send a notification or do nothing. That way the last job to finish would send the notification and no job would wait.

But this way, having one designated parallel job run its tests and then wait for others to finish, seems like it's probably more reliable. My idea is likely vulnerable to race conditions caused by the time between a job checking if other jobs had finished and the job itself finishing. And since we [split tests into parallel jobs based on timing data](https://circleci.com/docs/parallelism-faster-jobs#splitting-by-timing-data) and thus expect all parallel jobs to take a similar amount of time to run, I think spinning and waiting with this approach is acceptable.
https://github.com/DataBiosphere/terra-ui/blob/0c99f2032050f47320577ecd07653f98ddd9c84a/.circleci/config.yml#L177

A few comments, but I'll request changes only over the output: I don't think we need to log the same information every 10 seconds. Also, consider running shell scripts through [ShellCheck](https://www.shellcheck.net/).

.circleci/wait-for-job-finish.sh
aweng98: Yeah, updated `jq` selector to use it: `select(.index != '$CIRCLE_NODE_INDEX')'`

.circleci/wait-for-job-finish.sh
aweng98: I used them during development. Removed.

.circleci/wait-for-job-finish.sh
aweng98: Removed

.circleci/wait-for-job-finish.sh
aweng98: Changed to `counter`.

.circleci/config.yml
aweng98: With `when:` (took few tries):

Screen Shot 2022-08-12 at 12 53 40 PM

nawatts (approved):

aweng98:
Running tests few times to confirm new changes are still working.

.circleci/config.yml
aweng98: Couldn't get `when` to work in step. I found an [explanation](https://discuss.circleci.com/t/using-environment-variables-inside-a-conditional/39152/4) from CircleCI help.

> Because _**when**_ statements are evaluated when the configuration is compiled prior to running. Environment variables are expanded at runtime.

.circleci/wait-for-job-finish.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Exit script if you try to use an uninitialized variable.
+set -o nounset
+
+# Use the error status of the first failure, rather than that of the last item in a pipeline.
+set -o pipefail
+
+counter=0
+
+# Wait up to 5 minutes for job to finish. A job can run on multiple nodes: parallelism > 1
```
cahrens, line 11: We don't know if the individual nodes will finish within 5 minutes of each; I've seen some pretty unbalanced times. I think we may need to increase this timeout, but we can monitor how it behaves in practice.

cahrens (approved):

PR #3312, opened 2022-08-12
fix value copied for data repo dataset id

calypsomatic:
[AJ-550](https://broadworkbench.atlassian.net/browse/AJ-550) Copying source dataset repo id instead copies snapshot id 
#3007 (from [AJ-345](https://broadworkbench.atlassian.net/browse/AJ-345)) added the linked data repo snapshot id and source data repo dataset id to the snapshot info page for snapshot-by-reference, including the ability to copy these ids with the clipboard icon.  Both clipboard icons, however, copy the same value.  This apparently went unnoticed for several months; fixing it now.


davidangb (approved):
looks good, did not test myself

cahrens (approved):

PR #3313, opened 2022-08-12
[SU-184] Allow changing attribute type for multiple rows

nawatts:
Currently, Terra UI allows editing a column in multiple rows of a data table at once. However, the only option when doing so is to set the field to the same value in all selected rows. Some users have requested the ability to, instead of overwriting the field in the selected rows, convert the field's value to a different type. This change adds that capability. With this change, when editing multiple rows, users can choose between "Set value" and "Convert type" operations.

## Before
Screen Shot 2022-08-12 at 6 10 17 PM

## After

https://user-images.githubusercontent.com/1156625/184451255-a8c428ce-86c4-4689-991e-628e2a4e159b.mov

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -949,7 +949,7 @@ export const SingleEntityEditor = ({ entityType, entityName, attributeName, attr
   ])
 }
 
-export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames, entityTypes, workspaceId: { namespace, name }, onDismiss, onSuccess }) => {
+export const MultipleEntityEditor = ({ entityType, entities, attributeNames, entityTypes, workspaceId: { namespace, name }, onDismiss, onSuccess }) => {
   const [attributeToEdit, setAttributeToEdit] = useState('')
   const [attributeToEditTouched, setAttributeToEditTouched] = useState(false)
   const attributeToEditError = attributeToEditTouched && Utils.cond(
@@ -967,15 +967,15 @@ export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames, 
     try {
       setIsBusy(true)
 
-      const entityUpdates = _.map(entityName => ({
+      const entityUpdates = _.map(entity => ({
         entityType,
-        name: entityName,
+        name: entity.name,
         operations: [{
           op: 'AddUpdateAttribute',
           attributeName: attributeToEdit,
           addUpdateAttribute: prepareAttributeForUpload(newValue)
         }]
-      }), entityNames)
+      }), entities)
 
       await Ajax(signal)
         .Workspaces
@@ -991,7 +991,7 @@ export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames, 
   const doDelete = async () => {
     try {
       setIsBusy(true)
-      await Ajax(signal).Workspaces.workspace(namespace, name).deleteAttributeFromEntities(entityType, attributeToEdit, entityNames)
+      await Ajax(signal).Workspaces.workspace(namespace, name).deleteAttributeFromEntities(entityType, attributeToEdit, _.map('name', entities))
       onSuccess()
     } catch (e) {
       onDismiss()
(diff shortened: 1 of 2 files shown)


src/components/data/data-utils.js
```diff
@@ -949,72 +948,91 @@ export const SingleEntityEditor = ({ entityType, entityName, attributeName, attr
...
+      attributeName: attributeToEdit,
+      addUpdateAttribute: convertAttributeValue(entity.attributes[attributeToEdit], type, referenceEntityType)
+    }]
+  }), entities)
+
+  return Ajax()
+    .Workspaces
+    .workspace(namespace, name)
+    .upsertEntities(entityUpdates)
+}
```
kyuksel: There appears to be quite a bit of code duplication between `editEntitiesSetValue` and `editEntitiesConvertType` and they are used only in `saveAttributeEdits`. Have you considered inlining them in `saveAttributeEdits`?

src/components/data/data-utils.js
```diff
@@ -949,72 +948,91 @@ export const SingleEntityEditor = ({ entityType, entityName, attributeName, attr
...
   const [attributeToEditTouched, setAttributeToEditTouched] = useState(false)
   const attributeToEditError = attributeToEditTouched && Utils.cond(
     [!attributeToEdit, () => 'An attribute name is required.'],
     [!_.includes(attributeToEdit, attributeNames), () => 'The selected attribute does not exist.']
   )
 
+  const [operation, setOperation] = useState('setValue')
+
   const [newValue, setNewValue] = useState('')
+  const [newType, setNewType] = useState({ type: 'string' })
```
kyuksel: Have you considered enumerating `operation` and `newType` in JSON objects to make valid values more explicit, like so:

```javascript
const newTypes = {
  string: 'string',
  reference: 'reference',
  etc.
}

?

File: src/components/data/data-utils.js [hunk 3]
```diff
@@ -1057,12 +1075,46 @@ export const MultipleEntityEditor = ({ entityType, entityNames, attributeNames,
...
+                ],
+                value: operation,
+                onChange: ({ value: newOperation }) => {
+                  setOperation(newOperation)
+                  if (newOperation === 'setValue') {
+                    setNewValue('')
+                  }
+                  if (newOperation === 'convertType') {
+                    setNewType({ type: 'string' })
+                  }
```
kyuksel: Nit: Is the vertical space above warranted vs. ?

```suggestion
                  newOperation === 'setValue' && setNewValue('')
                  newOperation === 'convertType' && setNewType({ type: 'string' })
```

kyuksel (approved):
Looks like good usability improvement! 

To me, displaying the `operation` values as radio buttons rather than pull-down menu items would help with discoverability without taking up a lot of space since there are only two possible values but I assume this had been considered by UX.

Also a few nitty questions are inline but they're not blockers.

src/components/data/data-utils.js
nawatts: There's lots of code in data-utils that uses stringly typed attribute types. I'm open to changing those to use an enum like object, but I think it would make more sense to do in a separate change.

src/components/data/data-utils.js
nawatts: Personally, I think this is the clearest way to write code like this. I think the alternatives below obscure the idea that the purpose of the expression is the side effect of `setState`.

- ```js
  newOperation === 'setValue' && setNewValue('')
  ```

- ```js
  newOperation === 'setValue' ? setNewValue('') : undefined
  ```
  ESLint complains about this because of the [no-unused-expressions](https://eslint.org/docs/latest/rules/no-unused-expressions) rule.

src/components/data/data-utils.js
nawatts: Done in [7435e64](https://github.com/DataBiosphere/terra-ui/pull/3313/commits/7435e6447778474f733b7f1ae3cb8bf9d15c1f5b).

src/components/data/data-utils.js
nawatts: Added an enum like object for operation in [12798ce](https://github.com/DataBiosphere/terra-ui/pull/3313/commits/12798ce2db5d1a466b9db425ddd52beef5bd71e2).

MatthewBemis (approved):

src/components/data/data-utils.js
```diff
@@ -956,7 +956,11 @@ export const MultipleEntityEditor = ({ entityType, entities, attributeNames, ent
     [!_.includes(attributeToEdit, attributeNames), () => 'The selected attribute does not exist.']
   )
 
-  const [operation, setOperation] = useState('setValue')
+  const Operation = {
```
kyuksel, line 959: Thanks for introducing this.

For consistency with the rest of the codebase, I'd lower-case the initial in `Operation`, and I think we typically pluralize them.

src/components/data/data-utils.js
kyuksel: Ah, sorry for the typo (`?` instead of `&&`) in the original suggestion.

The arguments you made for not making the change are very reasonable 👍🏼

PR #3316, opened 2022-08-16
[SU-196] Refactor WorkspaceData component state

nawatts:
The WorkspaceData component uses a selectedDataType state variable to determine what is shown in the main panel. However, this state variable is overloaded. It may contain an ID for the type of data being shown (“localVariables” or “\_\_bucket_objects__”), an array containing a snapshot name and snapshot table name, a reference dataset ID, or a data table name.

https://github.com/DataBiosphere/terra-ui/blob/24e98a2d8e494093fb78c1f5e0a1a76bcfa5ed25/src/pages/workspaces/workspace/Data.js#L422-L432

This makes it difficult to extend WorkspaceData to support additional data types and I want to add support for a new type of data: data table versions ([SU-180](https://broadworkbench.atlassian.net/browse/SU-180)).

This change replaces selectedDataType with selectedData, which is an object containing a type key and additional identifier keys depending on the type. For example: `{ type: workspaceDataTypes.entities, entityType: 'sample' }`.

### DIFF
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -368,3 +368,5 @@ export const truncateInteger = integer => {
 
   return `${Math.floor(integer / 1000)}k`
 }
+
+export const enumify = keys => _.flow(_.map(key => [key, key]), _.fromPairs)(keys)
--- a/src/libs/utils.js
+++ b/src/libs/utils.js
@@ -369,4 +369,9 @@ export const truncateInteger = integer => {
   return `${Math.floor(integer / 1000)}k`
 }
 
-export const enumify = keys => _.flow(_.map(key => [key, key]), _.fromPairs)(keys)
+/**
+ * Convert a list of keys into an enum object.
+ * @param {string[]} keys
+ * @returns {Object}
+ */
+export const enumify = _.flow(_.map(key => [key, key]), _.fromPairs)
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -387,13 +387,7 @@ const DataTableActions = ({ workspace, tableName, rowCount, entityMetadata, onRe
   ])
 }
 
-const workspaceDataTypes = {
-  entities: 'entities',
-  snapshot: 'snapshot',
-  referenceData: 'referenceData',
-  localVariables: 'localVariables',
-  bucketObjects: 'bucketObjects'
-}
+const workspaceDataTypes = Utils.enumify(['entities', 'snapshot', 'referenceData', 'localVariables', 'bucketObjects'])
 
 const WorkspaceData = _.flow(
   forwardRefWithName('WorkspaceData'),
--- a/src/pages/workspaces/workspace/Data.js
+++ b/src/pages/workspaces/workspace/Data.js
@@ -35,9 +35,6 @@ import * as Utils from 'src/libs/utils'
 import { wrapWorkspace } from 'src/pages/workspaces/workspace/WorkspaceContainer'
 
 
-const localVariables = 'localVariables'
-const bucketObjects = '__bucket_objects__'
-
 const styles = {
   tableContainer: {
     display: 'flex', flex: 1, flexBasis: '15rem',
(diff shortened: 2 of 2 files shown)


src/pages/workspaces/workspace/Data.js
```diff
@@ -390,6 +387,14 @@ const DataTableActions = ({ workspace, tableName, rowCount, entityMetadata, onRe
   ])
 }
 
+const workspaceDataTypes = {
+  entities: 'entities',
+  snapshot: 'snapshot',
+  referenceData: 'referenceData',
+  localVariables: 'localVariables',
+  bucketObjects: 'bucketObjects'
+}
```
dmohs: It took me a bit to recognize that this is really just helping to avoid misspelling the data types. Given that level of importance, I might compress the space used a bit:
```suggestion
const workspaceDataTypes =
  'entities snapshot referenceData localVariables bucketObjects'.split(/\s+/).reduce((acc, k) => ({...acc, [k]: k}), {})
```
This also has the advantage of ensuring the keys and values are indeed identical at the expense of taking more time to understand. I don't have a strong preference, just highlighting the tradeoffs for you to make the decision.

src/pages/workspaces/workspace/Data.js
```diff
@@ -417,32 +422,20 @@ const WorkspaceData = _.flow(
...
-
   const loadEntityMetadata = async () => {
     try {
       setEntityMetadata(undefined)
       setEntityMetadataError(false)
       const entityMetadata = await Ajax(signal).Workspaces.workspace(namespace, name).entityMetadata()
 
-      setSelectedDataType(getSelectionType() === 'entities' && !entityMetadata[selectedDataType] ? undefined : selectedDataType)
+      if (selectedData?.type === workspaceDataTypes.entities && !entityMetadata[selectedData.entityType]) {
+        setSelectedData(null)
```
dmohs: I have no preference here, but curious why you chose `null` over `undefined`. Maybe pop-in a PR comment right here?

src/pages/workspaces/workspace/Data.js
```diff
@@ -417,32 +422,20 @@ const WorkspaceData = _.flow(
...
-    return Utils.cond(
-      [!selectedDataType, () => 'none'],
-      [selectedDataType === localVariables, () => 'localVariables'],
-      [selectedDataType === bucketObjects, () => 'bucketObjects'],
-      [_.isArray(selectedDataType), () => 'snapshots'],
-      [_.includes(selectedDataType, _.keys(referenceData)), () => 'referenceData'],
-      () => 'entities'
-    )
-  }
-
```
dmohs, line 433: So basically you obviated the need for this entire conditional by choosing a more appropriate data structure to hold the state. That's pretty great.

dmohs (approved):
This is the way.

src/pages/workspaces/workspace/Data.js
nawatts: We have a couple of these enum like objects in various places. Perhaps it's worth adding a utils function to generate them?

src/pages/workspaces/workspace/Data.js
nawatts: Mostly habit I suppose. But looking back, `undefined` would be more consistent with other state variables in this component.

nawatts: Switched to undefined in [f3ba647](https://github.com/DataBiosphere/terra-ui/pull/3316/commits/f3ba647675a2348dbe049a1dd205d6df502a5196).

src/pages/workspaces/workspace/Data.js
nawatts: Added a `Utils.enumify` function in [ec3187e](https://github.com/DataBiosphere/terra-ui/pull/3316/commits/ec3187ec4c7f6b2b2ae8073dccfe8757f2618919).

src/libs/utils.js
```diff
@@ -368,3 +368,5 @@ export const truncateInteger = integer => {
 
   return `${Math.floor(integer / 1000)}k`
 }
+
+export const enumify = keys => _.flow(_.map(key => [key, key]), _.fromPairs)(keys)
```
kyuksel: Could it be

```suggestion
export const enumify = _.flow(_.map(key => [key, key]), _.fromPairs)
```

?

Also, it may be good to add a comment with the input type/shape that this function is meant to take.

nawatts: I thought the named `keys` parameter gave more of a hint about the expected input. Replaced that with JSDoc in [5c7968e](https://github.com/DataBiosphere/terra-ui/pull/3316/commits/5c7968e3875664c7c5b3e24497c9817b8374b3cb).

src/libs/utils.js
```diff
@@ -369,4 +369,9 @@ export const truncateInteger = integer => {
   return `${Math.floor(integer / 1000)}k`
 }
 
-export const enumify = keys => _.flow(_.map(key => [key, key]), _.fromPairs)(keys)
+/**
+ * Convert a list of keys into an enum object.
```
cahrens, line 373: Documentation! Awesome.

MatthewBemis (approved):

PR #3317, opened 2022-08-16
[DC-510] Reenable preview dataset

s-rubenstein:
Do not merge until DC-347 is in prod


fboulnois (approved):
Look good. Will the mixpanel events be tracked in a separate ticket?

s-rubenstein:
@fboulnois we actually already seem to track a mixpanel event, so that's a noop!

cahrens (approved):

tcjiang9 (approved):

PR #3319, opened 2022-08-16
[DC-488] link to workspace

s-rubenstein:
<!--
Hello, friend!
Thanks!

--!>


src/libs/utils.js
```diff
@@ -368,3 +368,13 @@ export const truncateInteger = integer => {
 
   return `${Math.floor(integer / 1000)}k`
 }
+
+// can this be done with usePollingEffect?
```
s-rubenstein: @kyuksel I'd specifically like eyes on this, if possible! I feel like there should probably be an existing way to do this in the codebase, but couldn't figure it out. `usePollingEffect` seemed to be mostly used for infinite polling, or something of that nature.

nawatts: It makes sense to avoid `usePollingEffect` for this so that it can be used outside of a React component.

src/pages/library/dataBrowser-utils.js
```diff
@@ -123,14 +132,34 @@ export const useDataCatalog = () => {
...
-  Nav.history.push({
-    pathname: Nav.getPath('import-data'),
-    search: qs.stringify({
-      url: getConfig().dataRepoUrlRoot, format, referrer: 'data-catalog',
-      snapshotIds: _.map('dct:identifier', datasets)
-    })
-  })
+export const importDataToWorkspace = async (dataset, asyncHandler) => {
+  const routeOptions = await Utils.cond(
+    [isWorkspace(dataset), () => {
```
pshapiro4broad: JS doesn't mind that one is `async` and one isn't?

src/pages/library/dataBrowser-utils.js
```diff
@@ -123,14 +132,34 @@ export const useDataCatalog = () => {
...
+          snapshotName: dataset['dct:title'],
+          catalogDatasetId: dataset.id
+        })
+      }
+    }],
+    [
+      isDatarepoSnapshot(dataset), async () => {
+        asyncHandler()
+        const jobInfo = await Ajax().DataRepo.snapshot(dataset['dct:identifier']).exportSnapshot()
+        await poll(async () => await Ajax().DataRepo.job(jobInfo.id).details(), 1000, jobStatus => jobStatus['job_status'] !== 'running')
```
pshapiro4broad: This may take a while to complete. Do we need to think about providing user feedback while this is running (as the TDR UI does)? If the user navigates away while this is running, what happens to the response of the async steps?

Also, do we poll for TDR job completion elsewhere in the UI? Even if we don't, it's something that may be useful elsewhere too.

src/pages/library/dataBrowser-utils.js
```diff
@@ -123,14 +132,34 @@ export const useDataCatalog = () => {
...
+        const jobResult = await Ajax().DataRepo.job(jobInfo.id).result()
+        const jobResultManifest = jobResult && jobResult.format && jobResult.format.parquet && jobResult.format.parquet.manifest
+        return {
+          pathname: Nav.getPath('import-data'),
+          search: qs.stringify({
+            url: getConfig().dataRepoUrlRoot, format: 'tdrexport', referrer: 'data-catalog',
+            snapshotId: dataset['dct:identifier'], snapshotName: dataset['dct:title'], tdrmanifest: jobResultManifest
```
pshapiro4broad: If we arrive here and `jobResultManifest` doesn't exist, what will happen?

src/pages/library/dataBrowser-utils.js
```diff
@@ -123,14 +132,34 @@ export const useDataCatalog = () => {
...
+        })
+      }
+    }],
+    [
+      isDatarepoSnapshot(dataset), async () => {
+        asyncHandler()
+        const jobInfo = await Ajax().DataRepo.snapshot(dataset['dct:identifier']).exportSnapshot()
+        await poll(async () => await Ajax().DataRepo.job(jobInfo.id).details(), 1000, jobStatus => jobStatus['job_status'] !== 'running')
+        const jobResult = await Ajax().DataRepo.job(jobInfo.id).result()
+        const jobResultManifest = jobResult && jobResult.format && jobResult.format.parquet && jobResult.format.parquet.manifest
```
pshapiro4broad: can you instead use `?.`?
```suggestion
        const jobResultManifest = jobResult?.format?.parquet?.manifest
```

src/libs/utils.js
s-rubenstein: We could, however, probably use this in `usePollingEffect` to simplify that code. I can make that refactor. (Not quite out of the box, it looks like - if I am understanding correctly - `usePollingEffect` also has a variable that says whether the function calls should start after the delay, or before the delay)

src/pages/library/dataBrowser-utils.js
```diff
@@ -123,14 +132,34 @@ export const useDataCatalog = () => {
...
+        return {
+          pathname: Nav.getPath('import-data'),
+          search: qs.stringify({
+            url: getConfig().dataRepoUrlRoot, format: 'tdrexport', referrer: 'data-catalog',
+            snapshotId: dataset['dct:identifier'], snapshotName: dataset['dct:title'], tdrmanifest: jobResultManifest
+          })
+        }
+      }
+    ]
```
pshapiro4broad: What do we want to happen if both `isX()` methods return `false`?

src/pages/library/DataBrowserDetails.js
```diff
@@ -194,17 +195,21 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
             title: dataObj['dct:title']
           })
-          importDataToWorkspace([dataObj])
+          importDataToWorkspace(dataObj, () => setPreparingExport(true))
         }
-      }, ['Link to a workspace']),
+      }, [preparingExport ?
+        div({ style: { fontStyle: 'italic', display: 'flex', alignItems: 'center', justifyContent: 'space-around' } }, [
+          spinner({}),
+          div({ style: { marginLeft: '1rem' } }, ['Preparing data'])
```
pshapiro4broad, line 253: So this is shown until the `async` operation is complete? That would work.

src/libs/utils.js
```diff
@@ -368,3 +368,13 @@ export const truncateInteger = integer => {
...
+export const poll = async (pollFn, pollTime, pollUntil) => {
+  let result = await pollFn()
+  while (!pollUntil(result)) {
```
pshapiro4broad: Maybe a `do` `while` would let you avoid duplicating `await pollFn()`?

pshapiro4broad (approved):
I had some questions but it looks OK, as far as I can tell.

src/pages/library/dataBrowser-utils.js
nawatts: `await` can be used on any value.

> If the value of the expression following the await operator is not a Promise, it's converted to a resolved Promise.
>
> -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#description

src/pages/library/dataBrowser-utils.js
s-rubenstein: So, I was mostly following the mocks that Lou has laid out. I agree that it is a while to wait with such minimal feedback. If the user navigates away, it will actually pull them back to the `import-data` page when finished, regardless of what they're doing. This is probably bad UX, and it should probably cancel when the user leaves the page. I'll see what I can do for that.

We may want to change it, but I am inclined to get user feedback first, and go from there. Starting with the minimal feedback, and if we need more than that, we can follow up with something more intrusive.

src/pages/library/dataBrowser-utils.js
s-rubenstein: So, I believe that shouldn't happen as long as the job completes, but it is likely an error will pop up (in the top right, which is where our errors go for the user), and it will redirect with undefined or null in the URL.

src/pages/library/dataBrowser-utils.js
s-rubenstein: That's a good question. I was thinking about some sort of modal saying "We do not yet support export for this dataset." with maybe a comment to alert support? But open to UX thoughts here.

src/pages/library/DataBrowserDetails.js
s-rubenstein: Yes, correct

src/libs/utils.js
s-rubenstein: Discussed offline - it depends on where the logic is put for whether it polls immediately. (either here or in the delay)

This follows existing code style based on `usePollingEffect` so opted to stick with present styling

src/libs/ajax.js
```diff
@@ -1108,6 +1108,22 @@ const DataRepo = signal => ({
...
+      details: async () => {
+        const res = await fetchDataRepo(`repository/v1/jobs/${jobId}`, _.merge(authOpts(), { signal }))
+        return res.json()
+      },
+      result: async () => {
+        const res = await fetchDataRepo(`repository/v1/jobs/${jobId}/result`, _.merge(authOpts(), { signal }))
+        return res.json()
       }
     }
   }
```
kyuksel: FYI that the above could be written as below for concision:

```javascript
  job: jobId => ({
    details: async () => {
      const res = await fetchDataRepo(`repository/v1/jobs/${jobId}`, _.merge(authOpts(), { signal }))
      return res.json()
    },
    result: async () => {
      const res = await fetchDataRepo(`repository/v1/jobs/${jobId}/result`, _.merge(authOpts(), { signal }))
      return res.json()
    }
  })
```

fboulnois (approved):
I like @pshapiro4broad questions and suggestions. Otherwise, this seems to accurately reflect the functionality and user journey we're expecting, so lgtm.

src/libs/utils.js
kyuksel: I agree that refactoring `poll()` so it can be used from within `usePollingEffect()` sounds ideal.

> Not quite out of the box, it looks like - if I am understanding correctly - usePollingEffect also has a variable that says whether the function calls should start after the delay, or before the delay

[Thinking out loudly] I wonder if something like

```javascript
const poll = async (pollFn, pollTime, pollUntil, leading, signal) => {
  const initialResult = leading ? await pollFn() : undefined
  
  if (!_.isUndefined(initialResult) && pollUntil(initialResult)) {
    return initialResult
  }
  
  let result = initialResult
  while (!signal.aborted && !pollUntil(result)) {
    await delay(pollTime)
    result = !signal.aborted && await pollFn()
  }
  
  return result
}
```

could be used by both `usePollingEffect()` and from within `dataBrowser-utils.js`. Note that I haven't tested it.

Either way, the function could probably use [JSDoc](https://en.wikipedia.org/wiki/JSDoc)'ing.

src/pages/library/DataBrowserDetails.js
```diff
@@ -194,17 +195,21 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
         onClick: () => {
           Ajax().Metrics.captureEvent(`${Events.catalogWorkspaceLink}:detailsView`, {
             id,
             title: dataObj['dct:title']
           })
-          importDataToWorkspace([dataObj])
+          importDataToWorkspace(dataObj, () => setPreparingExport(true))
         }
-      }, ['Link to a workspace']),
+      }, [preparingExport ?
```
kyuksel, line 250: I wonder if this may stop working when we upgrade to React 18 (currently using `17.0.2`) due to [automatic batching](https://github.com/reactwg/react-18/discussions/21). Probably not based on the examples on that page but I figured I'd bring it up.

src/pages/library/dataBrowser-utils.js
```diff
@@ -39,6 +40,14 @@ export const datasetReleasePolicies = {
   releasepolicy_other: { policy: 'SnapshotReleasePolicy_Other', label: 'Other', desc: 'Misc release policies' }
 }
 
+export const isWorkspace = dataset => {
+  return _.toLower(dataset['dcat:accessURL']).includes('/#workspaces/')
+}
```
kyuksel, line 41: FYI that the above could be written as below for concision and a bit better error handling (thanks to Lodash):

```javascript
export const isWorkspace = _.flow( 
  _.toLower,
  _.includes('/#workspaces/')
)
```

Similarly for the function below.

src/pages/library/dataBrowser-utils.js
```diff
@@ -123,14 +132,34 @@ export const useDataCatalog = () => {
...
-    pathname: Nav.getPath('import-data'),
-    search: qs.stringify({
-      url: getConfig().dataRepoUrlRoot, format, referrer: 'data-catalog',
-      snapshotIds: _.map('dct:identifier', datasets)
-    })
-  })
+export const importDataToWorkspace = async (dataset, asyncHandler) => {
+  const routeOptions = await Utils.cond(
+    [isWorkspace(dataset), () => {
+      return {
```
kyuksel: Nit: This `return` wrapping could be removed for concision as explained in an earlier comment.

src/pages/library/DataBrowserDetails.js
```diff
@@ -194,17 +195,21 @@ export const SidebarComponent = ({ dataObj, id }) => {
         ])
       ]),
       h(ButtonPrimary, {
-        disabled: true,
-        tooltip: 'We are currently working on the link to workspace feature which will be available soon.',
+        disabled: dataObj.access !== datasetAccessTypes.GRANTED,
+        tooltip: dataObj.access === datasetAccessTypes.GRANTED ? '' : uiMessaging.controlledFeature_tooltip,
```
kyuksel: I know it wasn't introduced by this PR but can we rename `controlledFeature_tooltip` as `controlledFeatureTooltip` to be consistent with the `camelCase` convention we follow?

src/pages/library/dataBrowser-utils.js
kyuksel: [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) for the win here 😊 
 
```suggestion
        const jobResultManifest = jobResult?.format?.parquet?.manifest
```

src/pages/library/dataBrowser-utils.js
kyuksel: May want to test and handle that error case?

src/pages/library/dataBrowser-utils.js
kyuksel: Either way, should we define a default case within `Utils.cond()`?

src/pages/library/dataBrowser-utils.js
s-rubenstein: Okay, here's the update from talking with Lou:
Before navigating, the user is going to need to click again (so there is a secondary click when it is ready)
We're also going to open a toast, that will say "your data is being prepared for analysis", in case the user clicks away. They can also click a link in the toast when the export is finished to go to the import page. This toast is going to be non-closable.

src/pages/library/dataBrowser-utils.js
s-rubenstein: I'll add that as an error state (we want to handle error states differently)

integration-tests/tests/run-catalog-workflow.js
```diff
@@ -21,6 +21,5 @@ const testCatalogFlowFn = _.flow(
 registerTest({
   name: 'run-catalog',
   fn: testCatalogFlowFn,
-  timeout: 2 * 60 * 1000,
-  targetEnvironments: []
+  timeout: 2 * 60 * 1000
```
kyuksel, line 24: Since we're re-enabling these two tests, have we run them with `test-flakes`

src/libs/utils.js
```diff
@@ -369,12 +369,30 @@ export const truncateInteger = integer => {
   return `${Math.floor(integer / 1000)}k`
 }
 
-// can this be done with usePollingEffect?
-export const poll = async (pollFn, pollTime, pollUntil) => {
-  let result = await pollFn()
-  while (!pollUntil(result)) {
```
kyuksel: Thank you for adding this helpful JSDoc below. It just misses `*` on lines between `/**` and `*/`. [Examples here](https://jsdoc.app/about-getting-started.html) in case it helps.

src/libs/utils.js
```diff
@@ -369,12 +369,30 @@ export const truncateInteger = integer => {
...
+/**
+ Polls using a given function until the pollUntil function returns true.
+
+ @author srubenstein
```
kyuksel: This is probably unnecessary since we have tools like `git blame` to determine authors as needed. It also could get out of date easily.
```suggestion
```

src/libs/utils.js
```diff
@@ -369,12 +369,30 @@ export const truncateInteger = integer => {
...
+ @param {() => A} pollFn The function to poll using
```
kyuksel: [It looks like](https://jsdoc.app/about-getting-started.html) we want a `-`, like so

```suggestion
 @param {() => A} pollFn - The function to poll using
```

Similarly for below.

src/libs/utils.js
```diff
@@ -369,12 +369,30 @@ export const truncateInteger = integer => {
...
+  let result = initialResult
+
+  while (!signal.aborted && !pollUntil(result)) {
     await delay(pollTime)
-    result = !pollUntil(result) && await pollFn()
+    result = !signal.aborted && await pollFn()
   }
+
   return result
 }
```
kyuksel, line 393: The function above may be a good unit test candidate.

src/pages/library/DataBrowserDetails.js
```diff
@@ -112,6 +117,36 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
+      [isWorkspace(dataset), () => Nav.history.push({
+        pathname: Nav.getPath('import-data'),
+        search: qs.stringify({
+          format: 'catalog',
+          snapshotName: dataset['dct:title'],
+          catalogDatasetId: dataset.id
+        })
+      })],
+      [
+        isDatarepoSnapshot(dataset), async () => {
```
kyuksel: Nit: We typically would start `isDatarepoSnapshot(dataset)…` right after the left bracket on the line above, similarly to how it is on line 122.

src/pages/library/dataBrowser-utils.js
s-rubenstein: So, this doesn't work out of the box. Because it is actually using the dcat:accessURL property, specifically. I tried adding a map `_.map(dataset => dataset['dcat:accessURL']`, but that also didn't work, so I am going to revert to what it was

kyuksel: Reverting sounds good. Sorry for my misleading suggestion.

integration-tests/tests/run-catalog-workflow.js
s-rubenstein: I have not. Can do. I don't think they've been flaky though?

kyuksel: We wouldn't know given the tests haven't being run?

s-rubenstein: They were running for a good long while. But yeah, the flow is different now. I can absolutely run a test.

PR #3320, opened 2022-08-16
[IA-3632] fix error display

jdcanas:
onError was omitted for some modal-based error reporting

### DIFF
--- a/src/pages/workspaces/workspace/analysis/AzureComputeModal.js
+++ b/src/pages/workspaces/workspace/analysis/AzureComputeModal.js
@@ -13,7 +13,7 @@ import {
   getRegionLabel
 } from 'src/libs/azure-utils'
 import colors from 'src/libs/colors'
-import { withErrorReporting, withErrorReportingInModal } from 'src/libs/error'
+import { withErrorReportingInModal } from 'src/libs/error'
 import { useOnMount } from 'src/libs/react-utils'
 import {
   computeStyles, getCurrentRuntime, getIsRuntimeBusy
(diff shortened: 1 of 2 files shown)


Qi77Qi (approved):

src/pages/workspaces/workspace/analysis/AzureComputeModal.js
```diff
@@ -24,7 +24,7 @@ import * as Utils from 'src/libs/utils'
 const titleId = 'azure-compute-modal-title'
 
 export const AzureComputeModalBase = ({
-  onDismiss, onSuccess, workspace: { workspace: { namespace, name: workspaceName, workspaceId } }, runtimes, hideCloseButton = false
+  onDismiss, onSuccess, onError = onDismiss, workspace: { workspace: { namespace, name: workspaceName, workspaceId } }, runtimes, hideCloseButton = false
```
breilly2, line 27: Why not just pass `onDismiss` to `withErrorReportingInModal`? `onDismiss` is both the default value _and_ the value that seems to be passed in explicitly in all cases (in the other file in this PR).

jdcanas: It does diverge in a few cases. I'd like the caller to have to think about it, but protect against them forgetting to pass it in.

breilly2: Ahh, okay. I didn’t notice the higher-order component from `withModalDrawer` so I didn’t trace the props back far enough. The varied `onError` comes from the use of `AzureComputeModal` in `Analyses.js`.

breilly2 (approved):

PR #3321, opened 2022-08-17
[SU-197] Remove broken state history from data tab

nawatts:
Several components on the data tab read information from state history (scroll position, current page, etc). However, no data they read this way is ever visible to a user.

Some children of WorkspaceData read from state history on their first render `firstRender`:
https://github.com/DataBiosphere/terra-ui/blob/86327f299540fe87d0570c7e9e2dcff597492f28/src/components/data/EntitiesContent.js#L369

`firstRender` is a state variable in WorkspaceData and passed down to children:
https://github.com/DataBiosphere/terra-ui/blob/86327f299540fe87d0570c7e9e2dcff597492f28/src/pages/workspaces/workspace/Data.js#L393-L401

On mount, WorkspaceData loads metadata:
https://github.com/DataBiosphere/terra-ui/blob/86327f299540fe87d0570c7e9e2dcff597492f28/src/pages/workspaces/workspace/Data.js#L528-L531

Which clears entityMetadata state:
https://github.com/DataBiosphere/terra-ui/blob/86327f299540fe87d0570c7e9e2dcff597492f28/src/pages/workspaces/workspace/Data.js#L434-L436

While entity metadata is loading, a spinner is the only child rendered. Other children are unmounted:
https://github.com/DataBiosphere/terra-ui/blob/86327f299540fe87d0570c7e9e2dcff597492f28/src/pages/workspaces/workspace/Data.js#L553-L555

Thus, by the time children such as EntitiesContent are mounted, firstRender will be false and they will not use anything from state history.

I could attempt to make this work correctly. However, the WorkspaceData component is messy enough that I'd rather clear out broken functionality and reduce complexity before attempting a refactor.

Also, currently, using state history in the data tab is complicated by having multiple pages (data tables, snapshots, reference data, files, etc) all share the same URL (`<namespace>/<name>/data`). Thus, they also share the same state history entry, which is what necessitated reading state history based on `firstRender`. This functionality will be much easier to restore in a possible future where all of those pages have their own URLs and their own entries in state history.

This also removes state history from UploadPreviewTable. There, any restored state is useless without the table data itself, which is ephemeral.

### DIFF
--- a/src/components/data/DataTable.js
+++ b/src/components/data/DataTable.js
@@ -58,7 +58,7 @@ const DataTable = props => {
     childrenBefore,
     editable,
     activeCrossTableTextFilter,
-    persist, refreshKey, firstRender,
+    persist, refreshKey,
     snapshotName,
     deleteColumnUpdateMetadata,
     controlPanelStyle,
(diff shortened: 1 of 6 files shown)


cahrens (approved):
Thanks for this cleanup @nawatts! I agree that it is much better to start over if this functionality is ever truly desired vs. trying to fix the various issues with this implementation.

MatthewBemis (approved):
:shipit:

nawatts:

PR #3322, opened 2022-08-17
[IA-3663] Updated defaultDataprocWorkerDiskSize from 120 to 150.

n8landolt:
## Description

Currently there is a UI bug where 120 is shown as the default and lowest value for worker dataproc disk sizes. However, when a user goes to change that value to anything under 150, it will "jump" to 150.

Additionally, the runtimeDefault in Leonardo is set to 150. This will fix the "jump" issue and simultaneously synchronize default values with Leonardo.

## Implementation

Update the default dataproc disk size to 150.

TODO: Fix the "jump" issue if dataProc diskSize is set to less than 150 (future proofing)

## Testing

Go to create a cluster for a runtime, confirm that 150 is the minimum.

### DIFF
--- a/src/libs/runtime-utils.js
+++ b/src/libs/runtime-utils.js
@@ -49,7 +49,7 @@ export const mapToPdTypes = _.map(updatePdType)
 
 // Dataproc clusters don't have persistent disks.
 export const defaultDataprocMasterDiskSize = 150
-export const defaultDataprocWorkerDiskSize = 120
+export const defaultDataprocWorkerDiskSize = 150
 // Since Leonardo started supporting persistent disks (PDs) for GCE VMs, boot disk size for a GCE VM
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):
don't think this is needed unless automation tests proves otherwise

Qi77Qi (changes requested):
the bug is that in UI, default worker size is 120, but it flips to 150 if user attempts to update its value

A better fix will be not increasing worker size, but actually fixing the flip in UI

Qi77Qi (approved):
this is fine since Leo's worker disk size was bumped to 150 at some point (for whatever reason) https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L64

cahrens (approved):

breilly2 (approved):

PR #3325, opened 2022-08-19
[SU-179][SU-182] Add ability to save and view data table versions

nawatts:
This allows users to save "versions" of a data table. Currently, saving a version downloads the data table as a TSV and saves it in the workspace bucket. Version TSVs are saved under a specific prefix and have some metadata attached. This is not necessarily the final way that versions will be stored and it does have its limitations (users can alter the version files in the workspace bucket), but we think it's ok for a proof of concept that will let us try out the UI and workflow with users.

Saved versions can be viewed by clicking the "Show version history" button in the table menu in the data tab sidebar. Selecting a version in that list shows options to download or delete the version.

This is behind a feature flag. To enable it, run `window.configOverridesStore.set({ isDataTableVersioningEnabled: true })` in the console and reload Terra.

https://user-images.githubusercontent.com/1156625/185674007-8e945b62-89c8-4f49-9147-48fe990c2be2.mov

More information about entity versioning is available in the doc linked from the tickets ([SU-179](https://broadworkbench.atlassian.net/browse/SU-179) and [SU-182](https://broadworkbench.atlassian.net/browse/SU-182)).

### DIFF
--- a/src/components/data/data-table-versions.js
+++ b/src/components/data/data-table-versions.js
@@ -1,0 +1,134 @@
+import _ from 'lodash/fp'
+import { Fragment, useEffect, useState } from 'react'
+import { div, h, h1, li, ol, p } from 'react-hyperscript-helpers'
+import { ButtonPrimary, ButtonSecondary, DeleteConfirmationModal, IdContainer, spinnerOverlay } from 'src/components/common'
+import { parseGsUri } from 'src/components/data/data-utils'
+import { icon, spinner } from 'src/components/icons'
+import { TextInput } from 'src/components/input'
+import Modal from 'src/components/Modal'
+import { Ajax } from 'src/libs/ajax'
+import colors from 'src/libs/colors'
+import { FormLabel } from 'src/libs/forms'
+import { useCancellation } from 'src/libs/react-utils'
+import * as Style from 'src/libs/style'
+import * as Utils from 'src/libs/utils'
+
+
+const DownloadVersionButton = ({ url }) => {
+  const signal = useCancellation()
+  const [downloadUrl, setDownloadUrl] = useState()
+
+  useEffect(() => {
+    const loadUrl = async () => {
+      try {
+        setDownloadUrl(null)
+        const [bucket, object] = parseGsUri(url)
+        const { url: signedUrl } = await Ajax(signal).DrsUriResolver.getSignedUrl({ bucket, object })
+        setDownloadUrl(signedUrl)
+      } catch (error) {
+        setDownloadUrl(null)
+      }
+    }
+    loadUrl()
+  }, [signal, url])
+
+  return h(ButtonPrimary, {
+    disabled: !downloadUrl,
+    href: downloadUrl
+  }, [
+    !downloadUrl && icon('loadingSpinner', { size: 12, style: { marginRight: '1ch' } }),
+    'Download TSV'
+  ])
+}
+
+export const DataTableVersion = ({ version, onDelete }) => {
+  const { entityType, timestamp, description } = version
+
+  const [showDeleteConfirmation, setShowDeleteConfirmation] = useState(false)
+  const [deleting, setDeleting] = useState(false)
+
+  return div({ style: { padding: '1rem' } }, [
+    h1({
+      style: {
+        ...Style.elements.sectionHeader, fontSize: 20,
+        paddingBottom: '0.5rem', borderBottom: Style.standardLine, marginBottom: '1rem'
+      }
+    }, [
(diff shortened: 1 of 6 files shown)


MatthewBemis (approved):
I had a few thoughts as I played around with it. Since this is a POC I have no expectation that these should be addressed in this PR or at all, but wanted to capture them while they were on my mind in case they’re useful for future implementations:

1) There is no indicator for when saving a version is in progress. I versioned a table with 20k rows and was left sitting for a bit wondering if it went through. The success notification is a nice touch!

2) I created a bunch of versions of a table and am starting to realize that the version description might be more useful to display in the list than the timestamp (or maybe both are useful, but I wouldn't want that list to get too crowded since it's already a pretty tight area...). I found myself clicking into each version to view the description to have a sense of which version it was. I imagine users will run into the same thing over a long period of time once their memory of a particular version has faded a bit.

3) I wonder if we can also add `versionedBy:some-user@gmail.com` to the file metadata and display that in the info panel as well.

This is a sweet POC and an exciting new feature to be able to play around with!

nawatts:
All good ideas.

Re: 2, if we move the description to the sidebar as well, I'm starting to wonder if it even makes sense to have anything in the main panel. The download and delete buttons could be under a popup menu in the sidebar similar to the data tables themselves.

src/components/data/data-table-versions.js
```diff
@@ -0,0 +1,139 @@
...
+  const [downloadUrl, setDownloadUrl] = useState()
+
+  useEffect(() => {
+    const loadUrl = async () => {
+      try {
+        setDownloadUrl(null)
+        const [bucket, object] = parseGsUri(url)
+        const { url: signedUrl } = await Ajax(signal).DrsUriResolver.getSignedUrl({ bucket, object })
+        setDownloadUrl(signedUrl)
+      } catch (error) {
```
dmohs, line 31: I'm going to take your word for it that swallowing the error is the right move.

src/libs/data-table-versions.js
```diff
@@ -0,0 +1,109 @@
...
+import { useState } from 'react'
+import { parseGsUri } from 'src/components/data/data-utils'
+import { Ajax } from 'src/libs/ajax'
+import { getUser } from 'src/libs/auth'
+import { notify } from 'src/libs/notifications'
+import { useCancellation } from 'src/libs/react-utils'
+import * as Utils from 'src/libs/utils'
+
+
+export const dataTableVersionsRoot = '.data-table-versions'
```
dmohs: This is already a long name, so use your own judgment, but I might call this `dataTableVersionsPathRoot` so as not to confuse it with a css selector.

src/libs/data-table-versions.js
```diff
@@ -0,0 +1,109 @@
...
+
+  const objectName = `${dataTableVersionsRoot}/${entityType}/${versionName}`
+  const createdBy = getUser().email
+  await Ajax().Buckets.patch(googleProject, bucketName, objectName, {
+    metadata: {
+      createdBy,
+      entityType,
+      timestamp,
+      description
+    }
```
dmohs: I think this is just a smidge clearer in compact form:
```suggestion
    metadata: { createdBy, entityType, timestamp, description }
```
Your call.

src/libs/data-table-versions.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export const listDataTableVersions = async (workspace, entityType, { signal } = {}) => {
+  const { workspace: { googleProject, bucketName } } = workspace
+
+  const prefix = `${dataTableVersionsRoot}/${entityType}/`
+  const { items } = await Ajax(signal).Buckets.listAll(googleProject, bucketName, { prefix })
+
+  return _.flow(
+    _.filter(item => item.metadata?.entityType && item.metadata?.timestamp),
```
dmohs, line 46: Tried this out as an alternative, but didn't end up liking it.
```javascript
    _.filter(item => ['entityType', 'timestamp'].every(k => item.metadata?[k]),
```

dmohs (approved):

src/libs/data-table-versions.js
nawatts: Renamed in https://github.com/DataBiosphere/terra-ui/pull/3325/commits/b6c0e80be3f92a53372f3f4e776e4e965f95ce28.

src/components/data/data-table-versions.js
nawatts: Added error handling in https://github.com/DataBiosphere/terra-ui/pull/3325/commits/39d4ef77e945799c7ab455b94c01c335e82d2bbf.

PR #3326, opened 2022-08-19
[IA-3587] Display default location in location select

n8landolt:
## Description

Users would like to be able to explicitly see the location in the select component instead of the current "Select..."

## Testing

Tested on the 3 available regions to me, US multiregional, US, and Montreal.

Screenshot from creating Jupyter notebook in default US space and US multiregional.

Screen Shot 2022-07-29 at 3 22 50 PM

Screenshot from creating a Jupyter Notebook in a Montreal located workspace. 
Is this the expected outcome?

Screen Shot 2022-07-29 at 3 21 14 PM)

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -1162,7 +1162,7 @@ export const ComputeModalBase = ({
                 isSearchable: false,
                 value: computeConfig.computeRegion,
                 onChange: ({ value, locationType }) => updateComputeLocation(value, locationType),
-                options: _.sortBy('label', getAvailableComputeRegions(location))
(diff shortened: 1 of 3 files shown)


xuscrbw (approved):
LGTM, tested this on the PR site by creating workspaces and seeing the default location was set  and not blank.

Also verified that when selecting a montreal based workspace, you can switch between two locations.

I did notice that sometimes after creating a workspace in Montreal with Hermione, it will spit an error about having problms finding the bucket. I am not sure if this is just an issue with Hermione or an issue overall... Don't think that would be caused by this PR..

dmohs (approved):
Nice.

PR #3329, opened 2022-08-22
[SU-186] Preparation for new profile page

MatthewBemis:
Once https://github.com/broadinstitute/firecloud-orchestration/pull/992 goes live, we will no longer bother to collect the PI, Non-Profit status, and institutional program values. This won't be a critical breaking change but it would be better to avoid the temporary state where Terra UI accepts values for these fields and throws them away.

[SU-160](https://broadworkbench.atlassian.net/browse/SU-160) will redesign the profile page entirely but it is easier to not try to time the merge of that with the Orch PR mentioned above.

Before

Screen Shot 2022-08-22 at 11 42 38 AM

After

Screen Shot 2022-08-22 at 11 42 21 AM

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/pages/Profile.js
+++ b/src/pages/Profile.js
@@ -467,7 +467,7 @@ const Profile = ({ queryParams }) => {
             ]),
             div({ style: { margin: '1rem' } }, [proxyGroup]),
 
-            sectionTitle('Program Info'),
+            sectionTitle('Location'),
 
             line([
               textField('programLocationCity', 'City'),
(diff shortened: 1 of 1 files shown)


src/pages/Profile.js
```diff
@@ -476,18 +469,6 @@ const Profile = ({ queryParams }) => {
 
             sectionTitle('Program Info'),
```
nawatts: With the non-profit status and PI/program lead gone, would it make sense to go ahead and change this section's label to "Location"?

nawatts (approved):

tlangs (approved):
LGTM

PR #3330, opened 2022-08-22
DC-513 re-enable landing page banner for catalog

tcjiang9:


src/libs/config.js
```diff
@@ -12,4 +12,4 @@ export const getConfig = () => {
  * Flags for hidden features
  */
 export const isCromwellAppVisible = () => getConfig().isCromwellAppVisible
-export const isDataBrowserFrontPage = () => false
+export const isDataBrowserFrontPage = () => true
```
nawatts: Why do we need the `isDataBrowserFrontPage` function if it always returns a constant? Can this function and the `isDataBrowserFrontPage() &&` condition in LandingPage be removed?

nawatts (approved):

petesantos (approved):

fboulnois (approved):
lgtm

s-rubenstein (approved):

src/pages/LandingPage.js
```diff
@@ -109,6 +94,7 @@ const LandingPage = () => {
       h(ButtonOutline, {
         style: { marginLeft: '2rem', padding: '1.5rem 1rem', textTransform: 'none' },
         onClick: () => {
+          Ajax().Metrics.captureEvent(`${Events.catalogView}:landingPageBanner`)
```
nawatts: By convention, this string should be added to the list in events.js instead of constructing an event name here.

src/libs/config.js
```diff
@@ -12,4 +12,5 @@ export const getConfig = () => {
  * Flags for hidden features
  */
 export const isCromwellAppVisible = () => getConfig().isCromwellAppVisible
-export const isDataBrowserFrontPage = () => false
+export const isDataTableProvenanceEnabled = () => getConfig().isDataTableProvenanceEnabled
+export const isDataTableVersioningEnabled = () => getConfig().isDataTableVersioningEnabled
```
nawatts: These functions have been removed in the dev branch and should not be re-added here.

PR #3331, opened 2022-08-23
[TUIM-31] Upgrade yarn to v3.2.2

nawatts:
The Dockerfile used to build Terra UI for BEEs does not pin the version of Node.js
https://github.com/DataBiosphere/terra-ui/blob/14ad4e0a3e49b84d934523e87fd69bbc1213ef76/Dockerfile#L1

Node v16.17.0 includes some changes to the [Experimental ESM Loader Hooks API](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V16.md#experimental-esm-loader-hooks-api) that are not compatible with the version of Yarn that we currently use (v3.2.0).

This caused the latest Docker build to fail: https://github.com/DataBiosphere/terra-ui/runs/7974523092?check_suite_focus=true.

This updates Yarn to v3.2.2, which works with Node v16.17.0.

To upgrade Yarn, I used:
```
yarn set version 3.x
yarn
```


breilly2 (approved):
👍 Thanks for figuring this out. I didn't test locally, but I recently did something similar with Job Manager so the diff makes sense to me.

MatthewBemis (approved):

PR #3332, opened 2022-08-23
[SU-172] Add provenance information for data table columns

nawatts:
This is a proof of concept for adding provenance information for data table columns. It shows which submissions used the selected column as an output. This is intended as a prototype to collect feedback from some users.

Currently, it is limited to checking the last 25 submissions that used the data table as the root entity type. This is done to limit the number of submission configurations that are individually fetched from the API. If we decide to make something like this generally available, we would likely add a new API endpoint or change existing ones to make fetching the provenance data more efficient.

This is behind a feature flag. To enable it, run `window.configOverridesStore.set({ isDataTableProvenanceEnabled: true })` in the console.

Screen Shot 2022-08-23 at 11 19 06 AM

Screen Shot 2022-08-23 at 11 19 08 AM

### DIFF
--- a/src/libs/data-table-provenance.js
+++ b/src/libs/data-table-provenance.js
@@ -1,6 +1,7 @@
 import _ from 'lodash/fp'
 import { useEffect, useState } from 'react'
 import { Ajax } from 'src/libs/ajax'
+import { isDataTableProvenanceEnabled } from 'src/libs/config'
 import { reportError } from 'src/libs/error'
 import { useCancellation } from 'src/libs/react-utils'
 import * as Utils from 'src/libs/utils'
@@ -82,7 +83,9 @@ export const useColumnProvenance = (workspace, entityType) => {
       }
     }
 
-    loadColumnProvenance()
+    if (isDataTableProvenanceEnabled()) {
+      loadColumnProvenance()
+    }
   }, [workspace, entityType, signal])
 
   return { columnProvenance, loading, error }
--- a/src/libs/data-table-provenance.js
+++ b/src/libs/data-table-provenance.js
@@ -1,7 +1,6 @@
 import _ from 'lodash/fp'
-import { useEffect, useState } from 'react'
+import { useCallback, useState } from 'react'
 import { Ajax } from 'src/libs/ajax'
-import { isDataTableProvenanceEnabled } from 'src/libs/config'
 import { reportError } from 'src/libs/error'
 import { useCancellation } from 'src/libs/react-utils'
 import * as Utils from 'src/libs/utils'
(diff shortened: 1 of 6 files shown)


src/libs/data-table-provenance.js
```diff
@@ -0,0 +1,90 @@
...
+      submissionsMaybeUsingEntityType
+    )
+  )
+
+  return _.flow(
+    _.map(([submission, configuration]) => ({ submission, configuration })),
+    _.filter(({ configuration }) => configuration.rootEntityType === entityType),
+    _.orderBy(({ submission }) => new Date(submission.submissionDate), 'desc')
+  )(_.zip(submissionsMaybeUsingEntityType, submissionConfigurations))
+}
```
dmohs, line 44: Nice FP through this function.

src/libs/data-table-provenance.js
```diff
@@ -0,0 +1,90 @@
...
+
+const getColumnFromOutputExpression = expr => {
+  const match = /^this\.([^.]+)$/.exec(expr)
+  return match ? match[1] : null
```
dmohs, line 48: Cases like this are when I really wish we had an either. If we did...
```javascript
return Either.of(/^this\.([^.]+)$/.exec(expr)).map(m => m[1])
```

src/libs/data-table-provenance.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+const getColumnProvenance = async (workspace, entityType, { signal } = {}) => {
+  const submissions = await getSubmissionsWithRootEntityType(workspace, entityType, { signal })
+
+  return _.flow(
+    _.flatMap(({ submission, configuration }) => _.map(([output, expression]) => ({ submission, configuration, output, expression }), _.toPairs(configuration.outputs))),
+    _.map(({ submission, configuration, output, expression }) => ({ submission, configuration, output, column: getColumnFromOutputExpression(expression) })),
```
dmohs: Then here, you could handle the `Left` case in the context of where it is used.
```javascript
column: getColumnFromOutputExpression(expression).getOr(null)
```
I like this because it is here where I know best what to do with the failure case.

Now, as for the line, what I think is happening here is that you are essentially converting `expression` to `column`, but the line is a bit noisy for that to be obvious. I might write it this way to give the transformation more signal:
```javascript
_.map(x => ({ ..._.omit(['expression'], x), column: getColumnFromOutputExpression(expression) }))
```

src/libs/data-table-provenance.js
```diff
@@ -0,0 +1,90 @@
...
+    _.filter(({ column }) => !!column),
```
dmohs: Consider whether this might be a clearer way of saying, "remove any items where this key has no value:"
```suggestion
    _.filter(x => !!x.column),
```

dmohs (approved):

src/libs/data-table-provenance.js
nawatts: Changed the latter in 959268e1cfa7295c1d3848b5e4f86a37413fdb13. I think part of the reason I had the arguments destructured was because I was unsure what to name a single argument. In retrospect though, appropriate names seem fairly obvious.

src/libs/data-table-provenance.js
nawatts: Changed in https://github.com/DataBiosphere/terra-ui/commit/959268e1cfa7295c1d3848b5e4f86a37413fdb13.

src/libs/data-table-provenance.js
dmohs: I hear that.

Not to belabor this, but I was reading the destructuring docs yesterday and it made me think you could do patterns like this without the `omit` (which struck me because it's a common pattern):
```javascript
_.map(({expression ...rest}) => ({ ...rest, column: getColumnFromOutputExpression(expression) }))
```

nawatts: I like it. Seems easier to read than `..._.omit`. Changed in https://github.com/DataBiosphere/terra-ui/pull/3332/commits/5ff7d5d67d3e5e323c636677799b025c74c2432c.

MatthewBemis (approved):

PR #3333, opened 2022-08-23
[SUP-688] Fix nested interactive controls in data table column headers

nawatts:
Currently, axe dev tools warn about nested interactive controls when viewing a data table. This is because, in the column heading, button to show the column actions menu is nested inside a button to sort the table by that column. Moving the menu button outside of the sort button fixes the issue.

### DIFF
--- a/src/components/data/data-utils.js
+++ b/src/components/data/data-utils.js
@@ -1433,27 +1433,28 @@ export const ModalToolButton = ({ icon, text, disabled, ...props }) => {
 }
 
 export const HeaderOptions = ({ sort, field, onSort, extraActions, children }) => {
-  const columnMenu = span({ onClick: e => e.stopPropagation() }, [
-    h(MenuTrigger, {
-      closeOnClick: true,
-      side: 'bottom',
-      content: h(Fragment, [
-        h(MenuButton, { onClick: () => onSort({ field, direction: 'asc' }) }, ['Sort Ascending']),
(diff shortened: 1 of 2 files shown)


src/components/data/data-utils.js
```diff
@@ -1433,27 +1433,28 @@ export const ModalToolButton = ({ icon, text, disabled, ...props }) => {
 }
 
 export const HeaderOptions = ({ sort, field, onSort, extraActions, children }) => {
-  const columnMenu = span({ onClick: e => e.stopPropagation() }, [
```
nawatts, line 1436: This wrapper was to fix a bug caused by the nested buttons (see #2959).

MatthewBemis (approved):

breilly2:
I found a rendering bug where the actions menu is rendered on top of the column width adjuster when the column isn't wide enough to fit everything. The column width adjuster seems to receive all clicks so I can't actually activate the menu unless I expand the column width to avoid the overlap.

Screen Shot 2022-08-24 at 2 08 16 PM

nawatts:
> I found a rendering bug where the actions menu is rendered on top of the column width adjuster when the column isn't wide enough to fit everything.

Nice catch. Should be fixed in https://github.com/DataBiosphere/terra-ui/pull/3333/commits/fed782caeef2d867d9a94a50b01edc6d8b62a1cc.

breilly2 (approved):

PR #3334, opened 2022-08-23
[dc-488] link to workspace blocking

s-rubenstein:
Tested manually both aborting and not aborting.

Ran integration tests.

Ran test-flakes against link-to-new-workspace.

We are starting with a blocking approach, despite the length it will take to export. If we see large dropoffs in mixpanel after release, we will re-evaluate then


src/pages/library/dataBrowser-utils.js
```diff
@@ -122,15 +127,3 @@ export const useDataCatalog = () => {
   })
   return { dataCatalog, refresh, loading }
 }
-
-export const importDataToWorkspace = datasets => {
```
s-rubenstein, line 126: I moved this into the using components and changed it to a single dataset on the principle of keeping it as close to the working code as possible - it isn't used elsewhere anymore.

src/libs/ajax.js
```diff
@@ -1110,9 +1110,23 @@ const DataRepo = signal => ({
       details: async () => {
         const res = await fetchDataRepo(`repository/v1/snapshots/${snapshotId}`, _.merge(authOpts(), { signal }))
         return res.json()
+      },
+      exportSnapshot: async () => {
+        const res = await fetchDataRepo(`repository/v1/snapshots/${snapshotId}/export`, _.merge(authOpts(), { signal }))
+        return res.json()
```
nmalfroy, line 1116: For these: what happens if the call fails?  Do we need to have error handling if the export kick off fails?

src/libs/ajax.js
```diff
@@ -1110,9 +1110,23 @@ const DataRepo = signal => ({
...
     }
-  }
+  },
+  job: jobId => ({
+    details: async () => {
+      const res = await fetchDataRepo(`repository/v1/jobs/${jobId}`, _.merge(authOpts(), { signal }))
+      return res.json()
+    },
+    result: async () => {
+      const res = await fetchDataRepo(`repository/v1/jobs/${jobId}/result`, _.merge(authOpts(), { signal }))
```
nmalfroy, line 1126: Same question here: probably need some error handling?

src/libs/ajax.js
nmalfroy: I also just ask in case the response isn't JSON (e.g. if a token has expired that can happen). That may be a fundamental change that is out of scope for the ticket...just wanted to check

src/libs/utils.js
```diff
@@ -369,6 +369,29 @@ export const truncateInteger = integer => {
...
+/**
+ * Polls using a given function until the pollUntil function returns true.
+ * @param {() => { result: A, shouldContinue: boolean }} pollFn - The function to poll using
+ * @param {number} pollTime - How much time there should be in ms between calls of the pollFn
+ * @param {boolean} leading - Whether the function should wait {pollTime} ms before running for the first time
+
+ * @returns {A} - The response of pollFn
+ */
+export const poll = async (pollFn, pollTime, leading = true) => {
+  do {
```
nmalfroy, line 381: I'm surprised...do we not have a polling function already?  Like, for the import service, does it not use one already?

src/libs/ajax.js
```diff
@@ -1110,9 +1110,23 @@ const DataRepo = signal => ({
       details: async () => {
         const res = await fetchDataRepo(`repository/v1/snapshots/${snapshotId}`, _.merge(authOpts(), { signal }))
         return res.json()
+      },
+      exportSnapshot: async () => {
+        const res = await fetchDataRepo(`repository/v1/snapshots/${snapshotId}/export`, _.merge(authOpts(), { signal }))
```
nmalfroy: Worth noting: there are other URL parameters that can be passed in to export:
- `exportGsPaths`: can be ignored for now and defaults to false
- `validatePrimaryKeyUniqueness`: this is set to true for now but if we're working with azure snapshots, that needs to be set to false (for now) or the export job will fail.  The cloud provider for the snapshot is returned as part of the snapshot now (as of yesterday)
Does it make sense to make that change now or as a followon?

nmalfroy (commented):
looks good!  I have a couple questions around error handling (mostly cause I'm not super familiar with the codebase) and one about support for Azure snapshots that would be good to handle

integration-tests/utils/catalog-utils.js
```diff
@@ -17,7 +17,9 @@ const linkDataToWorkspace = async (page, testUrl, token) => {
   await click(page, checkbox({ text: 'Granted', isDescendant: true }))
   // TODO: add test data with granted access DC-321
   await clickTableCell(page, { tableName: 'dataset list', columnHeader: 'Dataset Name', text: 'Readable Catalog Snapshot 1', isDescendant: true })
-  await noSpinnersAfter(page, { action: () => click(page, clickable({ textContains: 'Link to a workspace' })) })
+  await click(page, clickable({ textContains: 'Prepare for analysis' }))
+  await delay(20000)
```
petesantos: Why use a delay instead of `noSpinnersAfter`?

src/libs/events.js
```diff
@@ -31,6 +31,7 @@ const eventsList = {
   catalogToggle: 'catalog:toggle',
   catalogView: 'catalog:view',
   catalogWorkspaceLink: 'catalog:workspaceLink',
+  catalogWorkspaceLinkExportFinished: 'catalog:workspaceLink:completed',
```
petesantos, line 34: Could you make sure this item gets added to our [mixpanel lexicon](https://mixpanel.com/project/2120588/view/19059/app/lexicon#transformations)?

petesantos (commented):
A few minor comments - I still need to take a look at `DataBrowserDetails`

src/libs/ajax.js
s-rubenstein: So, terra-ui has a generic error handler. I believe it won't finish the await, and will likely either cancel the execution of the method, or poll, that will error, and it will pop up the unsupported modal. I can test though.

src/libs/utils.js
s-rubenstein: We have a `usePollingEffect`, but that is component driven, not webpage wide. (Unless the component is top level)

I created this as a need to have a webpage wide poll while we were doing the async approach, and think it provides enough value to leave in the PR.

src/libs/ajax.js
s-rubenstein: I think followon, if at all. I think there's conversations we need to have internally to the catalog on whether we want to support what here, or just use sensible defaults and obfuscate it from the non-power users.

integration-tests/utils/catalog-utils.js
s-rubenstein: I was finding with the polling it wasn't picking up the spinners with `noSpinnersAfter`. If it should be, then I can look into it a bit more.

src/libs/events.js
s-rubenstein: Yes, I will add it once this goes to prod, because I believe it will make the base event in the lexicon, and then I just need to edit the event.

src/libs/ajax.js
nmalfroy: Right.  I do think we'll want the ability to set it in the call (and not expose it to user...but just to make sure it works from Catalog). By setting that, we'll be able to support azure snapshots starting today (which is super cool)

src/libs/utils.js
nmalfroy: ah cool.  Yeah, that makes sense

petesantos (commented):
I am seeing an error when trying to link and not seeing an error notification.

src/libs/ajax.js
s-rubenstein: Added error handling for this and below!

src/libs/ajax.js
s-rubenstein: Added! (We do it at call site, not in the ajax library itself, usually)

src/libs/ajax.js
s-rubenstein: Lets do it as a follow on, because we haven't yet trial ran azure snapshots in the catalog, so we may need to make some metadata changes to be able to detect it. @tcjiang9 this may relate to schema definitions.

src/pages/library/DataBrowserDetails.js
```diff
@@ -194,17 +238,13 @@ export const SidebarComponent = ({ dataObj, id }) => {
         ])
       ]),
       h(ButtonPrimary, {
-        disabled: true,
-        tooltip: 'We are currently working on the link to workspace feature which will be available soon.',
+        disabled: dataObj.access !== datasetAccessTypes.GRANTED || tdrSnapshotPreparePolling,
```
pshapiro4broad: Shouldn't this also include `&& (isWorkspace() || isDatarepoSnapshot())`? If it's not one of these, then we don't know how to import it. Or maybe it's assumed that in those cases `dataObj.access` is never `GRANTED`? If so maybe a comment about that assumption would be good enough.

I see that the disabled cases is handled by `setDatasetNotSupportedForExport()`. Is it preferable to enable a button that doesn't work (and provide feedback to that effect once it's pressed) vs disabling it with a tooltip?

src/pages/library/DataBrowserDetails.js
```diff
@@ -223,10 +263,81 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
+    }, ['This dataset is not hosted in a storage system that currently has the ability to export to a research environment.']),
+    !!snapshotExportJobId && h(SnapshotExportModal, {
+      jobId: snapshotExportJobId,
+      dataset: dataObj,
+      onDismiss: () => {
+        setSnapshotExportJobId()
+        setTdrSnapshotPreparePolling(false)
+      },
+      onFailure: () => {
+        setDatasetNotSupportedForExport(true)
```
pshapiro4broad, line 277: Do we know what conditions will cause the operation fail here? Or it this a backstop handler for a case that we don't expect to see in practice?

src/libs/ajax.js
pshapiro4broad: I think we should add `validatePrimaryKeyUniqueness=false` for now, to support azure, and open a ticket in the backlog to disable this in the future once azure snapshots support this operation. In the case of a snapshot that's been created for sharing, I'd hope that the data owner has already verified that the data is correct and there's no primary key conflict in the data.

Although if primary key validation doesn't work at all for azure, I'm not sure why an option is needed. Is there a case where a caller would prefer the operation to fail instead of skipping validation for azure?

src/libs/ajax.js
pshapiro4broad: The main danger I see here is if the operation never completes in a way that causes the UI to return to normal. I suppose a user can always reload the page or close the tab. If we had some idea of an upper bound we could add a timeout and dismiss the modal state after the timer expires.

src/libs/ajax.js
nmalfroy: ok sounds good!

src/pages/library/DataBrowserDetails.js
```diff
@@ -146,7 +146,9 @@ export const SidebarComponent = ({ dataObj, id }) => {
           type: 'tdrSnapshot'
         })
         setTdrSnapshotPreparePolling(true)
-        const jobInfo = await Ajax().DataRepo.snapshot(dataset['dct:identifier']).exportSnapshot()
+        // const jobInfo = await Ajax().DataRepo.snapshot(dataset['dct:identifier']).exportSnapshot()
```
nmalfroy: I think these were left as stragglers

nmalfroy (approved):
lgtm!

src/pages/library/DataBrowserDetails.js
```diff
@@ -223,10 +261,81 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
+    datasetNotSupportedForExport === true && h(Modal, {
+      title: 'Cannot Export Dataset',
+      showCancel: false,
+      onDismiss: () => setDatasetNotSupportedForExport(false)
+    }, ['This dataset is not hosted in a storage system that currently has the ability to export to a research environment.']),
+    !!snapshotExportJobId && h(SnapshotExportModal, {
+      jobId: snapshotExportJobId,
+      dataset: dataObj,
+      onDismiss: () => {
+        setSnapshotExportJobId()
```
breilly2: Nit: IntelliJ is saying that an argument is expected. We should probably pass `undefined` here and on line 279 (since that's the default from `useState`).

src/libs/utils.js
```diff
@@ -369,6 +369,29 @@ export const truncateInteger = integer => {
...
+    !leading && await delay(pollTime)
+    const r = await pollFn()
+    if (!r.shouldContinue) { return r.result }
+    leading && await delay(pollTime)
+  } while (true)
```
pshapiro4broad: Possible improvement
```suggestion
    leading = false
  } while (r.shouldContinue)
  return r.result
```
Modifying `leading` avoids the need for the early return and avoids the repeated fragment `await delay(pollTime)`. This is at the cost of making `leading` mutable; it seems worth it to me though.

src/pages/library/DataBrowserDetails.js
```diff
@@ -223,10 +261,81 @@ export const SidebarComponent = ({ dataObj, id }) => {
     showRequestAccessModal && h(RequestDatasetAccessModal, {
       datasets: [dataObj],
       onDismiss: () => setShowRequestAccessModal(false)
+    }),
+    datasetNotSupportedForExport === true && h(Modal, {
```
breilly2: Nit: It looks to me like `datasetNotSupportedForExport` is always `true` or `false`, so I don't think the `=== true` is necessary. Did you find something different in your testing?

src/libs/utils.js
```diff
@@ -369,6 +369,29 @@ export const truncateInteger = integer => {
...
+ * Polls using a given function until the pollUntil function returns true.
+ * @param {() => { result: A, shouldContinue: boolean }} pollFn - The function to poll using
+ * @param {number} pollTime - How much time there should be in ms between calls of the pollFn
+ * @param {boolean} leading - Whether the function should wait {pollTime} ms before running for the first time
+
+ * @returns {A} - The response of pollFn
+ */
+export const poll = async (pollFn, pollTime, leading = true) => {
+  do {
+    !leading && await delay(pollTime)
```
pshapiro4broad: Is this correct? Doesn't this call `await` when `leading` is `false`? If so then it waits before the first call, which is the opposite of how `leading` is described above.

(Also, isn't `leading ||` equivalent to `!leading &&`? Avoiding the `!` would make it easier to read IMO.)

breilly2 (commented):
I'm seeing the same thing as Pete. It looks like it's due to getting a 401 from the TDR export call.

src/pages/library/DataBrowserDetails.js
```diff
@@ -109,8 +116,43 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
   const sidebarButtonWidth = 230
 
+
+  const importDataToWorkspace = dataset => {
+    Utils.cond(
+      [isWorkspace(dataset), () => {
+        Ajax().Metrics.captureEvent(`${Events.catalogWorkspaceLink}:detailsView`, {
+          id,
+          title: dataObj['dct:title'],
+          type: 'workspace'
```
pshapiro4broad: I'd rather not get in the habit of including type in our event logs. We can always derive type from the catalog ID, and we don't want the UI looking at type at all if we can avoid it.

src/pages/library/DataBrowserDetails.js
```diff
@@ -109,8 +116,43 @@ export const SidebarComponent = ({ dataObj, id }) => {
...
+  const [tdrSnapshotPreparePolling, setTdrSnapshotPreparePolling] = useState(false)
   const sidebarButtonWidth = 230
 
+
+  const importDataToWorkspace = dataset => {
+    Utils.cond(
+      [isWorkspace(dataset), () => {
+        Ajax().Metrics.captureEvent(`${Events.catalogWorkspaceLink}:detailsView`, {
+          id,
+          title: dataObj['dct:title'],
```
pshapiro4broad: Is `title` something we usually record in the event logs?

pshapiro4broad (approved):
Looks OK although I have a few comments

src/libs/utils.js
s-rubenstein: That's correct! Will fix!

src/libs/utils.js
s-rubenstein: I like the leading change. I am not going to change the while though, because it involves pulling the `r` out of the loop, and I like the immutability of setting it within the loop.

src/pages/library/DataBrowserDetails.js
s-rubenstein: I don't think so, but I think in this case we want it, because product would like to know which datasets people are most interested in. (Although id could give us that, its not readable for product, so they would need to ask us for it.)

src/pages/library/DataBrowserDetails.js
s-rubenstein: I disagree that we don't want to include things that are derivable with the id - I think that if we want this to be readable by product, we want to ensure they have the information they commonly will need without needing to ask us to go into the DB for them.

src/pages/library/DataBrowserDetails.js
s-rubenstein: Interesting. I did not get that message from intellij. I wonder if it is a configurable setting. Either way, maybe the current team on the terra-ui management rotation can take a ticket to add that to the linter?

src/pages/library/DataBrowserDetails.js
s-rubenstein: Backstop handler for a case we don't expect to see in practice.

s-rubenstein (commented):
@breilly2 the 401 error is from the TDR. They have a bug right now where you need owner permissions to export to a workspace, where it should only require READER+. They are fixing that, but I don't know if it is fixed yet.

s-rubenstein:
The reason there is no error message now, actually @breilly2 is that the TDR returns a 401. Technically they should return a 403 (@nmalfroy )

The 401 is getting detected as an expired token, and attempting to reload the auth token, which succeeds. This is because 401 is most often used as a proxy issue, and not as insufficient permissions.

Rather than fixing in the UI, I would say that we should fix in the API. However, when the TDR bug is fixed that allows READERs to export to workspace, this is likely to not be a problem, because the granted check will disable the button in someone doesn't have access.

src/pages/library/DataBrowserDetails.js
pshapiro4broad: At some future point I hope we can remove the conditional code and handle both workspace and snapshot cases using a single code path, like we do for all other catalog API calls. At that point we'd have to decide if we want to keep the `isWorkspace` logic around for the purposes of event logging. This could be a case of planning for a future that never arrives, but I'd like to avoid that future decision by not including it now.

Maybe as a proxy for this value you could include the access URL itself? That would be OK I think.

nmalfroy:
> 

Yeah, that's a known thing. We need to go through all of our 401 (or most of them at least) and convert them to 403's

src/pages/library/DataBrowserDetails.js
pshapiro4broad: Is the mix panel UI extensible? The title for an ID could be extracted programatically at display time by using the catalog `datasets` endpoint.

s-rubenstein: I don't believe it is - and if it is, it would mean giving admin credentials to mixpanel? Otherwise we couldn't guarantee access?

pshapiro4broad: Right now all catalog entries are publicly discoverable, so any user can get this metadata. Conversely, if a dataset _isn't_ publicly discoverable, maybe its metadata shouldn't be stored in mixpanel?

Overall I have a preference here, to only send `id`, but isn't a strong preference and I'm OK with sending more data. I want to explore the impact / possibility of not sending other data as well, but I don't want to hold up the PR on this point.

petesantos (approved):

PR #3336, opened 2022-08-24
[WX-838] Don't render execution directory link if there's no workflowRoot

breilly2:
I found a case where a [failed workflow](https://app.terra.bio/#workspaces/broad-firecloud-dsde/complex-featured-workflow/job_history/55258820-bcec-4db9-9c19-afc458b16126) didn't have a `workflowRoot`. Is this expected or is there some other problem? Not sure this is the right solution or if we should have some sort of fallback.


MatthewBemis (approved):

cjllanwarne (approved):

cahrens:
@breilly2 should this PR be closed?

PR #3337, opened 2022-08-24
[AJ-536] handle deleted TDR snapshots under references 

davidangb:
https://broadworkbench.atlassian.net/browse/AJ-536

Handles the case where:
1. a user has added a snapshot **reference** to their workspace
2. the TDR snapshot underneath the reference has been deleted

Previously, when viewing the details of the reference, the UI would throw an infinite-spinner error. Since the "Delete Snapshot" link was inside the details component, this prevented the user from removing the now-invalid reference.

As of this PR, we display the details component, but omit the offending fields, and allow the user to delete the reference.

In the deleted-snapshot case, we still log some console errors but they seem benign and do not impact functionality for the end user. Since this is a corner case, I am choosing to not spend the time to resolve those console errors.

Deleted snapshot:
Screenshot 8-25-2022 at 12 33 PM

Non-deleted snapshot:
Screenshot 8-24-2022 at 03 06 PM

### DIFF
--- a/src/components/workspace-utils.js
+++ b/src/components/workspace-utils.js
@@ -106,6 +106,7 @@ export const SnapshotInfo = ({
 }) => {
   // State
   const [snapshotInfo, setSelectedSnapshotInfo] = useState()
+  const [snapshotLoadError, setSnapshotLoadError] = useState()
   const [newName, setNewName] = useState(snapshotName)
   const [editingName, setEditingName] = useState(false)
   const [newDescription, setNewDescription] = useState(undefined)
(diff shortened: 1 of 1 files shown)


src/components/workspace-utils.js
```diff
@@ -150,6 +155,43 @@ export const SnapshotInfo = ({
     length: { maximum: 63, tooLong: 'Name can\'t be more than 63 characters.' }
   })
 
+  const tdrErrorDisplay = () => div({ style: { paddingLeft: '1rem' } }, [
+    div({ style: Style.dashboard.header }, ['Error']),
+    `Details of this snapshot could not be loaded due to a problem calling Terra Data Repo: ${JSON.stringify(snapshotLoadError)}`
+  ])
```
davidangb, line 165: in the case of deleted snapshots, `snapshotLoadError` is an empty object and so this error message isn't super helpful. Again, since this is a corner case I'm not bothering to make it a lot better, and leaving this in place for other cases such as transient network errors.

src/components/workspace-utils.js
```diff
@@ -131,14 +132,18 @@ export const SnapshotInfo = ({
...
   useOnMount(() => {
     const loadSnapshotInfo = async () => {
-      const snapshotInfo = await Ajax(signal).DataRepo.snapshot(snapshotId).details()
-      setSelectedSnapshotInfo(snapshotInfo)
+      try {
+        const snapshotInfo = await Ajax(signal).DataRepo.snapshot(snapshotId).details()
+        setSelectedSnapshotInfo(snapshotInfo)
+        setSnapshotLoadError(undefined)
+      } catch (e) {
+        setSelectedSnapshotInfo({ createdDate: Date.now() }) // fake data prevents problems with Utils.makeCompleteDate below
```
nawatts: The only place I see `createdDate` used is in a block that's conditional on `!snapshotLoadError`. If these two lines were switched, so that `setSnapshotLoadError(e)` is called before `setSelectedSnapshotInfo`, is the fake data still necessary?

nawatts (approved):

src/components/workspace-utils.js
davidangb: ahh, good call! 🤯 changed, and that works much better. Thank you.

nawatts: I think such problems will go away with [React 18's automatic batching](https://reactjs.org/blog/2022/03/29/react-v18.html#new-feature-automatic-batching).

jordansafer (approved):
LGTM!

PR #3339, opened 2022-08-26
[SUP-723] Fix "not supported on Azure workspaces" error 

ghost:
https://broadworkbench.atlassian.net/browse/SUP-723

### DIFF
--- a/src/pages/workspaces/workspace/WorkspaceContainer.js
+++ b/src/pages/workspaces/workspace/WorkspaceContainer.js
@@ -93,7 +93,7 @@ const WorkspaceTabs = ({
       h(WorkspaceMenu, {
         iconSize: 27, popupLocation: 'bottom',
         callbacks: { onClone, onShare, onLock, onDelete },
-        workspaceInfo: { canShare, isAzureWorkspace, isLocked, isOwner }
+        workspaceInfo: { canShare, isAzureWorkspace, isLocked, isOwner, workspaceLoaded: !!workspace }
       })
     ])
   ])
--- a/src/pages/workspaces/workspace/WorkspaceMenu.js
+++ b/src/pages/workspaces/workspace/WorkspaceMenu.js
@@ -12,7 +12,7 @@ import * as Utils from 'src/libs/utils'
 const WorkspaceMenu = ({
   iconSize, popupLocation,
   callbacks: { onClone, onShare, onLock, onDelete },
-  workspaceInfo: { name, namespace, canShare, isAzureWorkspace, isLocked, isOwner }
+  workspaceInfo: { name, namespace, canShare, isAzureWorkspace, isLocked, isOwner, workspaceLoaded }
 }) => {
   const navIconProps = {
     style: { opacity: 0.65, marginRight: '1rem', height: iconSize },
@@ -21,7 +21,7 @@ const WorkspaceMenu = ({
 
   const menuContent = !!namespace ?
     h(DynamicWorkspaceMenuContent, { namespace, name, onShare, onClone, onDelete, onLock }) :
-    h(WorkspaceMenuContent, { canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete })
+    h(WorkspaceMenuContent, { canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded })
 
   return h(MenuTrigger, {
     side: popupLocation,
@@ -47,9 +47,9 @@ const DynamicWorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDele
   })
 }
 
-const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded = true }) => {
+const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded = false }) => {
   const shareTooltip = Utils.cond(
-    [isAzureWorkspace, () => 'Sharing is not currently supported on Azure Workspaces'],
+    [workspaceLoaded && isAzureWorkspace, () => 'Sharing is not currently supported on Azure Workspaces'],
     [workspaceLoaded && !canShare, () => 'You have not been granted permission to share this workspace'],
     [Utils.DEFAULT, () => '']
   )
@@ -61,25 +61,25 @@ const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, o
 
   return h(Fragment, [
     h(MenuButton, {
-      disabled: isAzureWorkspace,
+      disabled: !workspaceLoaded || isAzureWorkspace,
       tooltip: workspaceLoaded && isAzureWorkspace && 'Cloning is not currently supported on Azure Workspaces',
       tooltipSide: 'left',
       onClick: onClone
     }, [makeMenuIcon('copy'), 'Clone']),
     h(MenuButton, {
-      disabled: !canShare || isAzureWorkspace,
+      disabled: !workspaceLoaded || !canShare || isAzureWorkspace,
       tooltip: shareTooltip,
       tooltipSide: 'left',
       onClick: onShare
     }, [makeMenuIcon('share'), 'Share']),
     h(MenuButton, {
-      disabled: !isOwner,
+      disabled: !workspaceLoaded || !isOwner,
       tooltip: workspaceLoaded && !isOwner && ['You have not been granted permission to ', isLocked ? 'unlock' : 'lock', ' this workspace'],
       tooltipSide: 'left',
       onClick: onLock
     }, isLocked ? [makeMenuIcon('unlock'), 'Unlock'] : [makeMenuIcon('lock'), 'Lock']),
     h(MenuButton, {
-      disabled: !isOwner || isLocked,
+      disabled: !workspaceLoaded || !isOwner || isLocked,
       tooltip: deleteTooltip,
       tooltipSide: 'left',
       onClick: onDelete
--- a/src/pages/workspaces/workspace/WorkspaceMenu.js
+++ b/src/pages/workspaces/workspace/WorkspaceMenu.js
@@ -47,7 +47,7 @@ const DynamicWorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDele
   })
 }
 
-const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded = false }) => {
+const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded }) => {
   const shareTooltip = Utils.cond(
     [workspaceLoaded && isAzureWorkspace, () => 'Sharing is not currently supported on Azure Workspaces'],
     [workspaceLoaded && !canShare, () => 'You have not been granted permission to share this workspace'],
--- a/src/pages/workspaces/workspace/WorkspaceMenu.js
+++ b/src/pages/workspaces/workspace/WorkspaceMenu.js
@@ -47,27 +47,32 @@ const DynamicWorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDele
   })
 }
 
-export const TooltipText = {
-  AzureCloneTooltip: 'Cloning is not currently supported on Azure Workspaces',
-  AzureShareTooltip: 'Sharing is not currently supported on Azure Workspaces'
+export const tooltipText = {
+  cloneAzureUnsupported: 'Cloning is not currently supported on Azure Workspaces',
+  shareAzureUnsupported: 'Sharing is not currently supported on Azure Workspaces',
+  shareNoPermission: 'You have not been granted permission to share this workspace',
+  deleteLocked: 'You cannot delete a locked workspace',
+  deleteNoPermission: 'You must be an owner of this workspace or the underlying billing project',
+  lockNoPermission: 'You have not been granted permission to lock this workspace',
+  unlockNoPermission: 'You have not been granted permission to unlock this workspace'
 }
 
 const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded }) => {
   const shareTooltip = Utils.cond(
-    [workspaceLoaded && isAzureWorkspace, () => TooltipText.AzureShareTooltip],
-    [workspaceLoaded && !canShare, () => 'You have not been granted permission to share this workspace'],
+    [workspaceLoaded && isAzureWorkspace, () => tooltipText.shareAzureUnsupported],
+    [workspaceLoaded && !canShare, () => tooltipText.shareNoPermission],
     [Utils.DEFAULT, () => '']
   )
   const deleteTooltip = Utils.cond(
-    [workspaceLoaded && isLocked, () => 'You cannot delete a locked workspace'],
-    [workspaceLoaded && !isOwner, () => 'You must be an owner of this workspace or the underlying billing project'],
+    [workspaceLoaded && isLocked, () => tooltipText.deleteLocked],
+    [workspaceLoaded && !isOwner, () => tooltipText.deleteNoPermission],
     [Utils.DEFAULT, () => '']
   )
 
   return h(Fragment, [
     h(MenuButton, {
       disabled: !workspaceLoaded || isAzureWorkspace,
-      tooltip: workspaceLoaded && isAzureWorkspace && TooltipText.AzureCloneTooltip,
+      tooltip: workspaceLoaded && isAzureWorkspace && tooltipText.cloneAzureUnsupported,
       tooltipSide: 'left',
       onClick: onClone
     }, [makeMenuIcon('copy'), 'Clone']),
@@ -79,7 +84,7 @@ const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, o
     }, [makeMenuIcon('share'), 'Share']),
     h(MenuButton, {
       disabled: !workspaceLoaded || !isOwner,
-      tooltip: workspaceLoaded && !isOwner && ['You have not been granted permission to ', isLocked ? 'unlock' : 'lock', ' this workspace'],
+      tooltip: workspaceLoaded && !isOwner && [isLocked ? tooltipText.unlockNoPermission : tooltipText.lockNoPermission],
       tooltipSide: 'left',
       onClick: onLock
     }, isLocked ? [makeMenuIcon('unlock'), 'Unlock'] : [makeMenuIcon('lock'), 'Lock']),
--- a/src/pages/workspaces/workspace/WorkspaceMenu.js
+++ b/src/pages/workspaces/workspace/WorkspaceMenu.js
@@ -47,9 +47,14 @@ const DynamicWorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDele
   })
 }
 
+export const TooltipText = {
+  AzureCloneTooltip: 'Cloning is not currently supported on Azure Workspaces',
+  AzureShareTooltip: 'Sharing is not currently supported on Azure Workspaces'
+}
+
 const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded }) => {
   const shareTooltip = Utils.cond(
-    [workspaceLoaded && isAzureWorkspace, () => 'Sharing is not currently supported on Azure Workspaces'],
+    [workspaceLoaded && isAzureWorkspace, () => TooltipText.AzureShareTooltip],
     [workspaceLoaded && !canShare, () => 'You have not been granted permission to share this workspace'],
     [Utils.DEFAULT, () => '']
   )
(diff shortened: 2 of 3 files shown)


src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+})
+
+jest.mock('src/components/TooltipTrigger', () => ({
+  ...jest.requireActual('src/components/TooltipTrigger'),
+  __esModule: true,
+  default: jest.fn()
+}))
+
+
+describe('test workspace undefined', () => {
```
msilva-broad: describe labels should be nouns (what is being tested).  Here "WorkspaceMenu - undefined workspace" would work.  Or "WorkspaceMenu with undefined workspace".

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+            }
+          },
+          [children]
+        ),
+        open && !!content && div([content])
+      ]))
+    })
+  })
+
+  it('tests disabled state for clone menu item when workspace is undefined', () => {
```
msilva-broad: it() labels refer to the thing being tested (workspace menu with undefined workspace here). So "renders Clone menu item as disabled" would work here.

src/pages/workspaces/workspace/WorkspaceMenu.js
```diff
@@ -47,9 +47,9 @@ const DynamicWorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDele
   })
 }
 
-const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded = true }) => {
+const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded = false }) => {
```
cahrens: This default value of false can be removed (since the default value is to be undefined, which is falsey). In other words, just don't specify a default value.

src/pages/workspaces/workspace/WorkspaceMenu.js
```diff
@@ -61,25 +61,25 @@ const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, o
...
     h(MenuButton, {
-      disabled: isAzureWorkspace,
+      disabled: !workspaceLoaded || isAzureWorkspace,
       tooltip: workspaceLoaded && isAzureWorkspace && 'Cloning is not currently supported on Azure Workspaces',
       tooltipSide: 'left',
       onClick: onClone
     }, [makeMenuIcon('copy'), 'Clone']),
     h(MenuButton, {
-      disabled: !canShare || isAzureWorkspace,
+      disabled: !workspaceLoaded || !canShare || isAzureWorkspace,
```
cahrens, line 80: For the benefit of other reviewers: I added checks for `!workspaceLoaded` on all of these for clarity. But because the default value of `canShare` and `isOwner` is falsey before the initial Ajax response is received, these menus were already correctly disabled during the pre-loaded period.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+        ),
+        open && !!content && div([content])
+      ]))
+    })
+  })
+
+  it('renders Clone menu item as disabled', () => {
+    const onClone = jest.fn()
+    render(h(WorkspaceMenu, {
+      iconSize: 20, popupLocation: 'left',
```
cahrens: Do you actually need to specify `iconSize`, `popupLocation`, and `callbacks`? We aren't actually triggering the callback function during the tests, as we are only checking the enable and tooltip states.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+
+  it('renders Clone tooltip', () => {
+    const onClone = jest.fn()
+    render(h(WorkspaceMenu, {
+      iconSize: 20, popupLocation: 'left',
+      callbacks: { onClone },
+      workspaceInfo: { name: 'example1', namespace: 'example-billing-project' }
+    }))
+    fireEvent.mouseOver(screen.getByText('Clone'))
+    expect(screen.queryByText('Cloning is not currently supported on Azure Workspaces')).toBeNull()
```
cahrens: I don't love this test-- even if a tooltip about Azure was incorrectly showing, the test would pass if the tooltip text was changed in the source code file. To make this test useful, I think a constant needs to be introduced for the text so it can be referenced in both places.

What I would prefer to see is that we are testing that the correct tooltip text **is** visible, but we don't show a tooltip during this initial loading period. Perhaps we should say something like "Workspace information is loading", but I think we would need to run that by UX.

cahrens: I also would be tempted fold the verification of the enable/disable state and the tooltip text into the same test case to share the setup code and make the relationship between them clearer. But I imagine some folks would prefer test cases to be as small and targeted as possible. @msilva-broad, what do you think?

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+
+  it('renders Share tooltip', () => {
+    const onShare = jest.fn()
+    render(h(WorkspaceMenu, {
+      iconSize: 20, popupLocation: 'left',
+      callbacks: { onShare },
+      workspaceInfo: { name: 'example1', namespace: 'example-billing-project' }
+    }))
+    fireEvent.mouseOver(screen.getByText('Share'))
+    expect(screen.queryByText('Sharing is not currently supported on Azure Workspaces')).toBeNull()
```
cahrens: Same comments as above.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+  })
+})
```
cahrens, line 96: I think we should go ahead and add coverage for Lock and Delete so all 4 menu items are tested (here and for GCP and Azure).

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+
+
+describe('WorkspaceMenu - GCP workspace', () => {
```
cahrens, line 98: I am surprised that `workspaceLoaded` is not getting passed in for any of these tests. I don't understand how they are working because the menu items should always be in the disabled state then.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+  beforeEach(() => {
+    useWorkspaceDetails.mockReturnValue({ workspace: { canShare: false, workspace: { isLocked: false } } })
+    MenuTrigger.mockImplementation(({ content }) => { return div([content]) })
+    TooltipTrigger.mockImplementation(({ content, children }) => {
```
cahrens: I wonder if you could define this TooltipTrigger mock implementation someplace where it could be reused by all the tests (since it is a bit complicated). Perhaps a base test object would work, or at least a const.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+
+  it('renders Clone menu item as disabled', () => {
+    const onClone = jest.fn()
+    render(h(WorkspaceMenu, {
+      iconSize: 20, popupLocation: 'left',
+      callbacks: { onClone },
+      workspaceInfo: { name: 'example1', namespace: 'example-billing-project' }
+    }))
+    const menuText = screen.getByText('Clone')
+    expect(menuText).toHaveAttribute('disabled')
```
cahrens, line 66: Why are you checking that `menuText` isn't disabled? And is it really text that is being returned, or the menu element?

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+
+  it('renders Clone menu item as disabled', () => {
+    const onClone = jest.fn()
+    render(h(WorkspaceMenu, {
+      iconSize: 20, popupLocation: 'left',
+      callbacks: { onClone },
+      workspaceInfo: { name: 'example1', namespace: 'example-billing-project' }
+    }))
+    const menuText = screen.getByText('Clone')
+    expect(menuText).not.toHaveAttribute('disabled')
```
cahrens: Clone should be enabled… if you pass `workspaceLoaded`. I don't think these assertions are doing what you intend

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+  it('renders Share menu item as enabled', () => {
+    useWorkspaceDetails.mockReturnValue({ workspace: { canShare: true, workspace: { isLocked: false } } })
+    const onShare = jest.fn()
+    render(h(WorkspaceMenu, {
+      iconSize: 20, popupLocation: 'left',
+      callbacks: { onShare },
+      workspaceInfo: { name: 'example1', namespace: 'example-billing-project' }
+    }))
+    const menuText = screen.getByText('Share')
+    expect(menuText).not.toHaveAttribute('disabled')
```
cahrens: Note that this assertion is identical to the one at line 157… again, I don't think the assertion is correct.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+  it('renders Share tooltip', () => {
+    const onShare = jest.fn()
+    render(h(WorkspaceMenu, {
+      iconSize: 20, popupLocation: 'left',
+      callbacks: { onShare },
+      workspaceInfo: { name: 'example1', namespace: 'example-billing-project' }
+    }))
+    fireEvent.mouseOver(screen.getByText('Share'))
+    screen.getByText('You have not been granted permission to share this workspace')
+    expect(screen.queryByText('Sharing is not currently supported on Azure Workspaces')).toBeNull()
```
cahrens: Let's not test the tooltip text that shouldn't be present given that you are testing what actually is present.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+
+  it('renders Share tooltip', () => {
+    const onShare = jest.fn()
+    render(h(WorkspaceMenu, {
+      iconSize: 20, popupLocation: 'left',
+      callbacks: { onShare },
+      workspaceInfo: { name: 'example1', namespace: 'example-billing-project' }
+    }))
+    fireEvent.mouseOver(screen.getByText('Share'))
+    screen.getByText('You have not been granted permission to share this workspace')
```
cahrens: Do you need to add an assertion that this isn't null?

ghost: I didn't add the assertion that it isn't null because I thought getByText by default throws an error if it can't find a match (if it's null).

cahrens:
@yunip20 I made a first pass-- let me know when you are ready for me to take a look again. If you have questions, feel free to ping me in Slack.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
cahrens: Oh, I didn't notice `getByText` which is different form `queryByText`. I would find it easier to read/understand the code if you always used `queryByText` and verified that the result either is or is not null.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
ghost: I wrote the label wrong for this test--this test renders Clone menu item as enabled (not disabled). I think the test should pass because I negated the `toHaveAttribute('disabled')` in `expect(menuText).not.toHaveAttribute('disabled')` , which checks that Clone is enabled.

cahrens: How is it rendering as enabled when `workspaceLoaded` is not specified?

ghost: I thought mocking useWorkspaceDetails `useWorkspaceDetails.mockReturnValue({ workspace: { canShare: false, workspace: { isLocked: false } } })` at line 106 would return a gcp workspace and set workspaceLoaded to truthy

cahrens: Hmmm, let me check out the branch and test it. Since you are passing `workspaceInfo` to `WorkspaceMenu` at line 132, I would have expected that `workspaceLoaded` needed to be included there. But perhaps I'm missing something!

cahrens: Oh, I get it! You are using the "dynamic menu content" path, which gets the information from the workspace details. That's fine (the other path remains uncovered, but it is less complex).

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+            }
+          },
+          [children]
+        ),
+        open && !!content && div([content])
+      ]))
+    })
+  })
+
+  it('renders Clone menu item as disabled', () => {
```
cahrens: Change this to enabled.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
cahrens: This is because the dynamic path is being executed (call to workspace details).

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+    const onClone = jest.fn()
+    render(h(WorkspaceMenu, {
+      iconSize: 20, popupLocation: 'left',
+      callbacks: { onClone },
+      workspaceInfo: { name: 'example1', namespace: 'example-billing-project' }
+    }))
+    const menuText = screen.getByText('Clone')
```
cahrens: The element returned is actually a `HTMLDivElement` (a React element), but the name suggests a string. I would suggest calling it `cloneMenuItem` or `menuItem`.

src/pages/workspaces/workspace/WorkspaceContainer.js
```diff
@@ -93,7 +93,7 @@ const WorkspaceTabs = ({
       h(WorkspaceMenu, {
```
breilly2, line 93: Would it make sense to just not render the menu until the workspace is loaded here?

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
cahrens: Because `iconSize` and `popupLocation` do not have default values and are referenced, they do need to be specified. If possible, leave callbacks as empty though so it doesn't make it appear that the callback is actually tested.

src/pages/workspaces/workspace/WorkspaceMenu.js
```diff
@@ -21,7 +21,7 @@ const WorkspaceMenu = ({
 
   const menuContent = !!namespace ?
     h(DynamicWorkspaceMenuContent, { namespace, name, onShare, onClone, onDelete, onLock }) :
-    h(WorkspaceMenuContent, { canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete })
+    h(WorkspaceMenuContent, { canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded })
 
   return h(MenuTrigger, {
```
breilly2, line 26: Another idea would be to render a disabled menu icon here if the workspace isn't loaded.

breilly2 (commented):
Mid-way comments; I still need to review the test code.

I'm also seeing some behavior I can't figure out. On the Workspace list where the permissions are loaded dynamically when the menu is opened, "Clone" seems to always be enabled while the workspace details are loading, including when it's an Azure workspace.

src/pages/workspaces/workspace/WorkspaceMenu.js
cahrens: @breilly2 the thing is that the workspace details aren't fetched until you actually click on the menu icon… we don't want to get the details for all the workspaces since there may be a very large number.

I don't see the behavior with clone that you are describing (initially enabled) on the PR site.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
msilva-broad: i agree, exposing a constant for the text value/s is a good thing.... I often expose stuff like this as an object with multiple keys:

export const TEXT = {
   blah: "the blah blah"
   foo: "the foo bar"
}

since we don't do internationalization, something like this should be fine.

msilva-broad: I think having seperate tests for the menu item checks and the tooltips checks are ok.  That gives better signal on what exactly broke if one of the tests fails.  It's ok however to group related assertions into each of those tests... like assert that all the menu items are rendered as expected.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
msilva-broad: yes, this was according to my guidance... getByText is a handy assert or throw where the failure will output super-handy DOM debug to the console.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+jest.mock('src/components/workspace-utils', () => {
+  const originalModule = jest.requireActual('src/components/workspace-utils')
+  return {
+    ...originalModule,
+    useWorkspaceDetails: jest.fn()
+  }
+})
+
+jest.mock('src/components/PopupTrigger', () => {
+  const originalModule = jest.requireActual('src/components/PopupTrigger')
```
msilva-broad, line 24: lets put a comment on this mock and the TooltipTrigger mock  below to state that we are mocking them to avoid test environment issues with React Portal's requirement to use DOM measure services which are not available in jest env.

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,258 @@
...
+          },
+          [children]
+        ),
+        open && !!content && div([content])
+      ]))
+    })
+  })
+
+  it('renders Clone menu item as disabled', () => {
+    const onClone = jest.fn()
```
msilva-broad: Let's add section headers inside each test function to follow the AAA pattern:

// Arrange

// Act

// Assert

(Sorry I didn't mention this earlier... will be part of a UI Unit Testing guide to be authored soon"

src/pages/workspaces/workspace/WorkspaceMenu.js
```diff
@@ -47,39 +47,49 @@ const DynamicWorkspaceMenuContent = ({ namespace, name, onClone, onShare, onDele
   })
 }
 
-const WorkspaceMenuContent = ({ canShare, isAzureWorkspace, isLocked, isOwner, onClone, onShare, onLock, onDelete, workspaceLoaded = true }) => {
+export const tooltipText = {
```
cahrens, line 50: @yunip20 I believe our convention is to use lowercase initial letters. I moved all the tooltip text options into this object so we could fully test everything.

cahrens:
@yunip20 I decided to take this unit test leveraging "each" to the full extent possible… hopefully you don't hate it!

cahrens (approved):

ghost:
I think that works, it definitely makes the tests more readable and understandable!

src/pages/workspaces/workspace/_tests/WorkspaceMenu.test.js
```diff
@@ -0,0 +1,320 @@
...
+  })
+
+  it.each([
+    { accessLevel: 'READER', locked: true },
+    { accessLevel: 'OWNER', locked: true },
+    { accessLevel: 'READER', locked: false },
+    { accessLevel: 'OWNER', locked: false }
+  ])('renders Delete tooltip for access level $accessLevel and locked status $locked', ({ accessLevel, locked }) => {
+    // Arrange
+    useWorkspaceDetails.mockReturnValue({ workspace: { workspace: { isLocked: locked }, accessLevel } })
```
ghost: I had a quick question: I see that `azureContext` is missing when mocking useWorkspaceDetails for some tests in azure case, should I include it?

cahrens: Good catch! I forgot that when I copied over the cases from GCP. I will push up a fix momentarily.

cahrens: @yunip20 I pushed up the changes-- please review when you get a chance. Thanks!

src/pages/workspaces/workspace/WorkspaceContainer.js
cahrens: I think we would want to render it as disabled instead of not rendering at all, but given that we already do the dynamic loading/enabling for the workspaces list, I don't think it is worth making the behavior different here.

msilva-broad (approved):
really great work!  The unit tests look fabulous!!!

PR #3340, opened 2022-08-26
DC-526 fix run-catalog-filter test in staging

tcjiang9:


integration-tests/tests/run-catalog-filter.js
```diff
@@ -15,8 +15,16 @@ const testCatalogFilterFn = withUserToken(async ({ testUrl, page, token }) => {
...
-  await enableDataCatalog(page, testUrl, token)
+  await gotoPage(page, testUrl)
+  await waitForNoSpinners(page)
+
   await setDatasetsMockValues(page)
+
+  await findText(page, 'Browse Data')
+  await click(page, clickable({ textContains: 'Browse Data' }))
+  await signIntoTerra(page, { token })
+  await enableDataCatalog(page)
```
pshapiro4broad, line 26: Could this call `navigateToDataCatalog()` instead of having this code here? It looks like it's doing the same thing.

pshapiro4broad (approved):
Looks good, just one comment.

integration-tests/tests/run-catalog-filter.js
tcjiang9: no because I need to call await setDatasetsMockValues(page) in the middle of the navigate logic

fboulnois: Could you create a `goToPageDataCatalogAndWait` with the first two function calls, and a `findAndSignInAndEnableCatalog` with the other 4 function calls? Then `navigateToDataCatalog` could call these two functions, and this code could use both of these two new functions, too.

fboulnois (approved):
Overall lgtm, one suggestion to address @pshapiro4broad 's comment.

integration-tests/tests/run-catalog-filter.js
dmohs: I usually find that when I am inclined to put "and" in a function name, I prefer to divide it and call each one individually.

dmohs (approved):

PR #3341, opened 2022-08-29
Update AMP PD dataset to point to Featured Workspace

emmarogge:
Remove both existing "Browse Data" buttons on Terra datasets page and
add new "Explore AMP PD" button which points to the new AMP PD
Featured Workspace. This increases discoverability of the new "home" for AMP PD resources in Terra, since all viewers with Terra accounts (not just those with AMP PD access) can view the AMP PD Featured Workspace. [Note: images shows tooltip when mouse hovering.]

image

### DIFF
--- a/src/pages/library/Datasets.js
+++ b/src/pages/library/Datasets.js
@@ -188,16 +188,11 @@ const amppd = () => h(Participant, {
   sizeText: 'Participants: 10,772'
 }, [
   h(ButtonPrimary, {
-    'aria-label': 'Browse AMP-PD Tier 1 data',
-    style: { marginBottom: '1rem' },
-    href: Nav.getLink('data-explorer-private', { dataset: 'AMP PD Clinical - 2020_v2release_1218' }),
-    onClick: () => captureBrowseDataEvent('AMP PD Clinical - 2020_v2release_1218')
-  }, ['Browse Tier 1 Data']),
-  h(ButtonPrimary, {
-    'aria-label': 'Browse AMP-PD Tier 2 data',
-    href: Nav.getLink('data-explorer-private', { dataset: 'AMP PD - 2020_v2release_1218' }),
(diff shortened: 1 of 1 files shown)


mbookman:
From an AMP PD program perspective, this looks great.

Really like having the new Featured Workspace, which provides a lot more detail to both existing AMP PD researchers as well as Terra users interested in Parkinson's research. Being able to point people to the Featured Workspace (which also has links to the data browsers) is a better user experience.

Thanks Emma!

cahrens (approved):

src/pages/library/Datasets.js
```diff
@@ -188,16 +188,11 @@ const amppd = () => h(Participant, {
...
-    href: Nav.getLink('data-explorer-private', { dataset: 'AMP PD Clinical - 2020_v2release_1218' }),
-    onClick: () => captureBrowseDataEvent('AMP PD Clinical - 2020_v2release_1218')
-  }, ['Browse Tier 1 Data']),
-  h(ButtonPrimary, {
-    'aria-label': 'Browse AMP-PD Tier 2 data',
-    href: Nav.getLink('data-explorer-private', { dataset: 'AMP PD - 2020_v2release_1218' }),
-    onClick: () => captureBrowseDataEvent('AMP PD - 2020_v2release_1218')
-  }, ['Browse Tier 2 Data'])
+    'aria-label': 'Learn about AMP PD data',
+    tooltip: 'Visit the featured workspace',
```
mbookman: nit: "Featured Workspace" is a named thing in Terra (at least per the docs):

https://support.terra.bio/hc/en-us/articles/360033599791-Feature-Your-Workspace

so let's upper-case it here. Otherwise LGTM.

mbookman (changes requested):

src/pages/library/Datasets.js
emmarogge: Fixed!

mbookman (approved):
LGTM

PR #3342, opened 2022-08-29
[QA-2000] UI integration test failed with Error: Node is detached from document

aweng98:
[QA-2000](https://broadworkbench.atlassian.net/browse/QA-2000)

Flaky test fix: In `Click()` function, add one retry in `try/catch` block. This PR handles `Node is detached from document` only.

CircleCI `integration-tests-staging` [job](https://app.circleci.com/pipelines/github/DataBiosphere/terra-ui/11562/workflows/57e79d0e-8e4a-48bd-a5f5-cfc5e7e53fbf/jobs/49486).
```

Tests Summary
----------------------------------------------
Test: import-cohort-data
Status: failed
Failure messages: Error: Node is detached from document
    at ElementHandle._scrollIntoViewIfNeeded (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:475:22)
    at ElementHandle.click (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:617:5)
    at /mnt/ramdisk/integration-tests/tests/import-cohort-data.js:21:3
    at /mnt/ramdisk/integration-tests/utils/integration-helpers.js:64:5
    at /mnt/ramdisk/integration-tests/utils/integration-utils.js:344:12
    at /mnt/ramdisk/integration-tests/utils/integration-utils.js:398:10
Failure details: [
  {}
]
```

`Error: Node is detached from document` happens when the referenced web element is no longer attached to the DOM when clicking.

### DIFF
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -109,11 +109,7 @@ const click = async (page, xpath, options) => {
   try {
     return (await page.waitForXPath(xpath, defaultToVisibleTrue(options))).click()
   } catch (e) {
-    // Handle Error: Node is detached from document
-    //   at ElementHandle._scrollIntoViewIfNeeded (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:475:22)
-    //   at processTicksAndRejections (node:internal/process/task_queues:96:5)
-    //   at ElementHandle.click (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:617:5)
-    if (e instanceof Error && e.message?.includes('Node is detached from document')) {
+    if (e.message.includes('Node is detached from document')) {
       return (await page.waitForXPath(xpath, defaultToVisibleTrue(options))).click()
     }
     throw e
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -106,7 +106,19 @@ const clickTableCell = async (page, { tableName, columnHeader, text, isDescendan
 }
 
 const click = async (page, xpath, options) => {
-  return (await page.waitForXPath(xpath, defaultToVisibleTrue(options))).click()
+  const element = await page.waitForXPath(xpath, defaultToVisibleTrue(options))
+  try {
+    return element.click()
+  } catch (e) {
+    // Handle Error: Node is detached from document
+    //   at ElementHandle._scrollIntoViewIfNeeded (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:475:22)
+    //   at processTicksAndRejections (node:internal/process/task_queues:96:5)
+    //   at ElementHandle.click (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:617:5)
+    if (e instanceof Error && e.message?.includes('Node is detached from document')) {
+      return (await page.waitForXPath(xpath, defaultToVisibleTrue(options))).click()
+    }
+    throw e
+  }
 }
 
 const findText = (page, textContains, options) => {
--- a/integration-tests/utils/integration-utils.js
+++ b/integration-tests/utils/integration-utils.js
@@ -106,9 +106,8 @@ const clickTableCell = async (page, { tableName, columnHeader, text, isDescendan
 }
 
 const click = async (page, xpath, options) => {
-  const element = await page.waitForXPath(xpath, defaultToVisibleTrue(options))
   try {
-    return element.click()
+    return (await page.waitForXPath(xpath, defaultToVisibleTrue(options))).click()
   } catch (e) {
     // Handle Error: Node is detached from document
     //   at ElementHandle._scrollIntoViewIfNeeded (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:475:22)
--- a/integration-tests/tests/import-cohort-data.js
+++ b/integration-tests/tests/import-cohort-data.js
@@ -1,8 +1,7 @@
 const _ = require('lodash/fp')
 const { withWorkspace } = require('../utils/integration-helpers')
 const {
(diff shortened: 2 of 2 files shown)


integration-tests/utils/integration-utils.js
```diff
@@ -106,7 +106,18 @@ const clickTableCell = async (page, { tableName, columnHeader, text, isDescendan
...
 const click = async (page, xpath, options) => {
-  return (await page.waitForXPath(xpath, defaultToVisibleTrue(options))).click()
+  try {
+    return (await page.waitForXPath(xpath, defaultToVisibleTrue(options))).click()
+  } catch (e) {
+    // Handle Error: Node is detached from document
+    //   at ElementHandle._scrollIntoViewIfNeeded (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:475:22)
+    //   at processTicksAndRejections (node:internal/process/task_queues:96:5)
+    //   at ElementHandle.click (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:617:5)
+    if (e instanceof Error && e.message?.includes('Node is detached from document')) {
```
aweng98: Is this the right way to check error message?

integration-tests/utils/integration-utils.js
```diff
@@ -106,7 +106,18 @@ const clickTableCell = async (page, { tableName, columnHeader, text, isDescendan
...
 
 const click = async (page, xpath, options) => {
-  return (await page.waitForXPath(xpath, defaultToVisibleTrue(options))).click()
+  try {
+    return (await page.waitForXPath(xpath, defaultToVisibleTrue(options))).click()
+  } catch (e) {
+    // Handle Error: Node is detached from document
+    //   at ElementHandle._scrollIntoViewIfNeeded (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:475:22)
+    //   at processTicksAndRejections (node:internal/process/task_queues:96:5)
+    //   at ElementHandle.click (/mnt/ramdisk/.yarn/unplugged/puppeteer-npm-13.5.1-231e586c7a/node_modules/puppeteer/src/common/JSHandle.ts:617:5)
```
nawatts: This comment isn't really providing any useful information. The error message is in the `if` statement and the sample stack trace refers to Puppeteer internals. An explanation of what causes this error and why it is expected to occasionally occur would be more useful.

integration-tests/utils/integration-utils.js
nawatts: It's probably safe to assume that `e` is an instance of Error here. Ideally, it's better to check errors based on an error code or something like that instead of a user facing message. That way, you avoid the risk of code breaking if the message is slightly reworded. Not all libraries provide that option though. It may be worth checking if Puppeteer uses a specific error class for this type of error ([puppeteer.errors](https://pptr.dev/api/puppeteer.errors)).

nawatts (commented):
Do we know why this error occurs?

cahrens:
In my past life, this meant that the handle the Selenium page object was out-of0date (ie., the page content had refreshed, and it was necessary to "visit" the page again to get a new page object reference). In that scenario, trying again with the stale reference would do no good.

I'm not sure how that maps to Puppeteer and our "page" objects, which are different. But I am skeptical that trying again will help.

breilly2:
@cahrens This code retries `page.waitForXPath` so it should find a new node.

I haven't had time to look into the stack trace, but I'm curious what is being clicked on here. If we're rendering something that someone could click on and then taking it away before it's actually clicked, that feels like a bug to me.

aweng98:
> but I'm curios what is being clicked on here.

The web element to be clicked is a text label for a plot/graph. This page has a bounch of plots to paint.
Screen Shot 2022-09-01 at 10 39 18 AM

integration-tests/utils/integration-utils.js
aweng98: Removed `instanceof` check.
I can't find the unique error code for this error. Based on the source [code](https://github.com/puppeteer/puppeteer/blob/b5e6474374ae6a88fc73cdb1a9906764c2ac5d70/src/common/JSHandle.ts#L376) tells me, only text string is returned.

integration-tests/utils/integration-utils.js
aweng98: Removed comment. I think the code is obvious already.

breilly2:
Since that's a UI that we don't control and therefore can't fix any rendering issues, I'd rather the retry be localized to that specific interaction. Having the retry in our general `click` utility could mask rendering issues that we _can_ fix.

aweng98:
> Do we know why this error occurs?

Not sure. My guess is page has not rendered all graphs while targeted text web element is already visible. A page with graphic web elements is not easy to work with because we don't know how to tell if graph has rendered.

integration-tests/utils/integration-utils.js
nawatts: Fair enough. Yeah, there's no guarantee that a library will always provide an error code or similar. Sometimes the only option is to look at the message.

nawatts (approved):
There are some places where we restore state from local storage, then immediately start loading fresh data from the backend and render a spinner. I don't know for sure, but I could imagine something like this could happen in those cases, like if the test finds a button rendered from the restored state and that button is removed by the time it clicks on it.

In any case, I like the move toward retrying smaller tasks within tests vs our current approach of retrying the entire test suite.

aweng98:
Do you feel strongly about making localized change in test? I wanted to handle this type of error in `click()` function because it can happen to any test when test script fails to sync properly with page rendering.

breilly2:
I'm going to call out my own perfectionism here. My gut preference is strong, but I'm not going to impose that on everyone in this case if others are okay with having the resilience built into `click`.

cahrens:
> I'm going to call out my own perfectionism here. My gut preference is strong, but I'm not going to impose that on everyone in this case if others are okay with having the resilience built into `click`.

I actually saw this failure in a different test very recently. So I am OK having it built-in to `click`… that should give us data faster about whether or not this actually fixes the problem.

breilly2:
@aweng98 I removed myself as a reviewer. I still don't love the approach, but I may be in the minority here. I would like someone other than me to approve to show that this change has more support.

cahrens (approved):
I'm still not convinced that this will fix the issue, but it will be easier for us to determine that if we go ahead and get it in.

PR #3343, opened 2022-08-30
[IA-3650] Fixed PD value not showing issue. Updated spark cluster UI formatting.

n8landolt:
https://broadworkbench.atlassian.net/browse/IA-3650

## Description

The input box for Persistent disk is not the correct size in a few spots in Terra-UI.
There were also other inputs improperly sized for the values that are able to be selected. 
Additionally, added menuPlacement: "auto" for the CPU and memory options for spark clusters to prevent the modal from stretching down further.

Old on left, new on right (same values selected)
Screen Shot 2022-08-30 at 11 03 17 AM

### DIFF
--- a/src/components/ComputeModal.js
+++ b/src/components/ComputeModal.js
@@ -94,6 +94,7 @@ const WorkerSelector = ({ value, machineTypeOptions, onChange }) => {
         div([
           h(Select, {
             id,
+            menuPlacement: 'auto',
             isSearchable: false,
             value: currentCpu,
             onChange: option => onChange(_.find({ cpu: option.value }, machineTypeOptions)?.name || value),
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):
looks good...tried on PR site

src/components/ComputeModal.js
```diff
@@ -94,6 +94,7 @@ const WorkerSelector = ({ value, machineTypeOptions, onChange }) => {
         div([
           h(Select, {
             id,
+            menuPlacement: 'auto',
```
jdcanas, line 97: what does this do? Not sure exactly from what you put in desc, just curious :)

breilly2 (approved):

PR #3345, opened 2022-08-30
[IA-2848] turn RuntimeManager into a functional component

jdcanas:
https://broadworkbench.atlassian.net/browse/IA-2848

Non-functional refactoring. Main thing that this component does is compare previous runtime/app status to it's current status and handle various transitions as they occur in the background. For example, it will display errors if a runtime transitions from a happy status to error, and also display the pop-up to launch an app when in transitions from provisioning/starting to running. 

Converting it to a functional component found a bunch of unused params that I got to clean up because the linter plays better with functional components :) 

I thought about changing the logic to be a Utils.cond instead of a massive if/else, but I'd prefer to refactor this incrementally until we have some tests in place for it...


src/components/RuntimeManager.js
```diff
@@ -199,10 +184,16 @@ export default class RuntimeManager extends PureComponent {
       })
       errorNotifiedApps.update(Utils.append(galaxyApp.appName))
     }
-  }
+  }, [runtimes, apps]) // eslint-disable-line react-hooks/exhaustive-deps
+
+  return null
```
jdcanas, line 189: Not sure if there is a better way of defining a pure component in functional react. It errors out if I remove the return and tells me to put the null back. Weird that it doesn't just figure out I don't want anything rendered if I omit the return...

cahrens: Yes, I believe it is required.

src/components/RuntimeManager.js
```diff
@@ -199,10 +184,16 @@ export default class RuntimeManager extends PureComponent {
       })
       errorNotifiedApps.update(Utils.append(galaxyApp.appName))
     }
-  }
+  }, [runtimes, apps]) // eslint-disable-line react-hooks/exhaustive-deps
```
cahrens: What are the other things the effect depends on? Can they just all be listed?

src/components/RuntimeManager.js
```diff
@@ -122,35 +122,20 @@ const AppErrorNotification = ({ app }) => {
...
+  const prevApps = usePrevious(apps)
+
+  useEffect(() => {
+    const runtime = getCurrentRuntime(runtimes) || {}
+    const prevRuntime = _.last(_.sortBy('auditInfo.createdDate', _.remove({ status: 'Deleting' }, prevRuntimes))) || {}
     const twoMonthsAgo = _.tap(d => d.setMonth(d.getMonth() - 2), new Date())
     const welderCutOff = new Date('2019-08-01')
-    const createdDate = new Date(runtime.createdDate)
-    const dateNotified = getDynamic(sessionStorage, `notifiedOutdatedRuntime${runtime.id}`) || {}
+    const createdDate = new Date(runtime?.createdDate)
```
cahrens, line 134: What has changed that now there may not be a current runtime?

src/components/RuntimeManager.js
jdcanas: It depends on `prevRuntimes`, `prevApps` which are computed off the included dependencies, and react is not able to figure it out. It also depends on `namespace, name` technically. Currently, the runtime is already scoped by these variables so its trivial to include them as well here since the runtime will refresh if they do. Can't hurt include these though in case that changes at some point.

src/components/RuntimeManager.js
jdcanas: There could always have been no current runtime. The definition does use a default [here](https://github.com/DataBiosphere/terra-ui/pull/3345/files#diff-aabe65a220cb1e02fd8b0e1a38c532eb92df64bcc9b0e244801141b6330d1ca5R130), but I'd rather be defensive here since the data still comes from an API response and we have no way of enforcing contract adherence currently.

cahrens: I can't tell what the effect would be if the `runtime` ends up being undefined-- the variable access is defensive and wouldn't throw an error, but would the usage of those variables be robust to undefined? It might be worth thinking through that a bit.

cahrens (approved):
Caveats-- I haven't tested it, and I'm by no means a React expert. But this looks generally straightforward.

src/components/RuntimeManager.js
jdcanas: Essentially, if there is no runtime, there's no meaningful work this component has to do since there are no errors/notifications/status transitions that the user needs to be notified about

nawatts: In that case, it seems like it might be clearer to do something like:
```js
if (!runtime) {
  return
}
```

vs have to deal with possibly undefined values all through the effect function.

jdcanas: Ah, the same thing applies to apps though. So no meaningful work in terms of processing around that runtime's state if its undefined, but we can't return due to the possibility we need to notify on apps. Honestly, this whole component is a massive if/else tree that should be simplified into a Utils.cond. I really think something like that is necessary soon or we're talking about dressing up something gross here.

n8landolt (approved):

PR #3347, opened 2022-08-31
Add the ability to create an Azure-backed billing project (WOR-386).

cahrens:
Adds the ability on dev **only** for a user in the Azure alpha group to create an Azure-backed billing project. Note that the user must already have a Terra Managed Application deployment in a subscription that they have owner permissions on.

If the user is in the alpha group, they will see a selector to choose either GCP or Azure:

image

Choosing Azure:

image

Azure Portal showing this managed app in the subscription:

image


config/dev.json
```diff
@@ -1,6 +1,7 @@
 {
   "agoraUrlRoot": "https://agora.dsde-dev.broadinstitute.org",
   "bardRoot": "https://terra-bard-dev.appspot.com",
+  "billingProjectManagerUrlRoot": "https://bpm.dsde-dev.broadinstitute.org",
```
cahrens: Currently dev is the only one that exists, but everything then prod should be available shortly. And Azure support is not enabled in prod, so we won't attempt to hit the prod BPM service.

integration-tests/tests/workspace-dashboard.js
```diff
@@ -172,10 +172,6 @@ const setAzureAjaxMockValues = async (testPage, namespace, name, workspaceDescri
     const workspaceSasTokenUrl = new RegExp(`api/workspaces/v1/${workspaceId}/resources/controlled/azure/storageContainer/(.*)/getSasToken`)
 
     window.ajaxOverridesStore.set([
-      {
```
cahrens, line 175: This is unrelated, but @aweng98 noticed it when working on switching tests to running against staging. This override was actually never used, as the namespace is incorrect (correct override is below at line 185).

src/pages/billing/List.js
```diff
@@ -133,11 +166,8 @@ const billingProjectNameValidator = existing => ({
   }
 })
 
-const noBillingMessage = onClick => div({ style: { fontSize: 20, margin: '2rem' } }, [
-  div([
-    'To get started, ',
-    h(Link, { onClick }, ['click here to create a Billing Project'])
-  ]),
+const noBillingMessage = div({ style: { fontSize: 20, margin: '4rem' } }, [
+  div(['To get started, use the Create button to create a Terra Billing Project']),
```
cahrens, line 170: Removed the link, as it was Google-specific.

src/pages/billing/List.js
```diff
@@ -183,14 +213,14 @@ const NewBillingProjectModal = ({ onSuccess, onDismiss, billingAccounts, loadAcc
   return h(Modal, {
     onDismiss,
     shouldCloseOnOverlayClick: false,
-    title: 'Create Billing Project',
+    title: 'Create Terra Billing Project',
```
cahrens, line 216: Minor changes to be consistent with Azure version.

There is enough different between the Azure and GCP versions that I don't think it makes sense to try to share the code between them. 

It would be nice to move the GCP project creation dialog out into its own file (and add tests), but I didn't want to do the refactor as part of this PR because I wanted it to be clear what actually changed in this dialog.

src/pages/billing/List.js
```diff
@@ -316,12 +347,14 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
 
   const authorizeAndLoadAccounts = () => authorizeAccounts().then(loadAccounts)
 
-  const showCreateProjectModal = async () => {
-    if (Auth.hasBillingScope()) {
-      setCreatingBillingProject(true)
+  const showCreateProjectModal = async type => {
+    if (type === billingProjectTypes.azure) {
+      setCreatingBillingProject(type)
```
cahrens, line 357: Do not need to elevate auth scope in the Azure case.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
```diff
@@ -0,0 +1,236 @@
...
+jest.mock('src/libs/ajax')
+
+describe('CreateAzureBillingProjectModal', () => {
+  beforeEach(() => {
+    // Arrange
+    Ajax.mockImplementation(() => {})
+    render(h(CreateAzureBillingProjectModal, { onSuccess: jest.fn(), onDismiss: jest.fn(), billingProjectNameValidator }))
+  })
+
+  const tooShortError = 'Billing project name is too short (minimum is 6 characters)'
```
cahrens, line 34: I am not attempting to reference the message from the source files because most all of these messages are partially constructed by the validator library, and thus we do not have references to the complete messages.

cahrens:
@dmohs unless you have an Azure deployed managed app, you won't be able to fully test this. However, you can log in as Hermione, who is in the Azure alpha group, to at least see the dialog. You can then try out some of the error handling, such as putting in a subscription ID that is invalid or one that you don't have access to (62b22893-6bc1-46d9-8a90-806bb3cce3c9). And of course you can check that the GCP path (which is all that is exposed in prod) still works.

src/libs/ajax.js
```diff
@@ -158,6 +158,7 @@ const fetchSam = _.flow(withUrlPrefix(`${getConfig().samUrlRoot}/`), withAppIden
 // are not transient and the request should not be retried.
 const fetchBuckets = _.flow(withRequesterPays, withRetryOnError(error => Boolean(error.requesterPaysError)), withUrlPrefix('https://storage.googleapis.com/'))(fetchOk)
 const fetchRawls = _.flow(withUrlPrefix(`${getConfig().rawlsUrlRoot}/api/`), withAppIdentifier)(fetchOk)
+const fetchBillingProjectManager = _.flow(withUrlPrefix(`${getConfig().billingProjectManagerUrlRoot}/api/`), withAppIdentifier)(fetchOk)
```
aherbst-broad: Nitpitck: It's the Billing _Profile_ Manager

src/pages/billing/List.js
```diff
@@ -43,6 +44,39 @@ const styles = {
   }
 }
 
+const billingProjectTypes = {
+  azure: 'azure',
+  gcp: 'gcp'
+}
+
+const CreateBillingProjectControl = ({ isAlphaSpendReportUser, showCreateProjectModal }) => {
```
aherbst-broad: Is this the name for this flag? We want the alpha azure group right?

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
```diff
@@ -0,0 +1,236 @@
...
+      _.forEach(error => expect(screen.queryByText(error)).not.toBeNull(), errors)
+      _.forEach(unexpectedError => expect(screen.queryByText(unexpectedError)).toBeNull(), _.difference(allErrors, errors))
+    }
+    // Create always disabled because all fields must have valid inputs.
+    verifyDisabled(getCreateButton())
+  })
+
+  it('validates the subscription ID', () => {
+    // Arrange
+    expect(screen.queryByText(invalidUuidError)).toBeNull() // Should not initially be visible
```
msilva-broad: to move this assertion down to the // Assert section I recommend you save the result of screen.queryByText(...) to a const initialXxxxError here and assert it .toBeNull() in the later section.

cahrens: @msilva-broad to be honest, I don't really care too much for the AAA pattern. As the tests become more complicated, it is natural to "assert" at multiple points and "act" at multiple points; "arranging" and even "acting" may also happen partially or completely in test setup. I think it is easier to understand the test if you can see "this is what I expect the state to be at this point of the test", vs. storing variables and doing all the assertions at the end.

src/pages/billing/CreateAzureBillingProjectModal.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+  })
+
+  // Field validation
+  const billingProjectNameErrors = validate({ billingProjectName }, { billingProjectName: billingProjectNameValidator(existing) })
+  const subscriptionIdErrors = validate({ subscriptionId }, { subscriptionId: { type: 'uuid' } })
+  const isValidSubscriptionId = !!subscriptionId && !subscriptionIdErrors
+
+  useEffect(() => {
+    if (isValidSubscriptionId) {
```
aherbst-broad, line 55: Nitpick: Consider an early return here

src/pages/billing/CreateAzureBillingProjectModal.js
```diff
@@ -0,0 +1,151 @@
...
+            const error = await (obj instanceof Response ? obj.text() : obj)
+            console.error(error)
+          }
+        }
+      )
+      fetchManagedApps()
+    }
+  }, [isValidSubscriptionId, subscriptionId, signal])
+
+  const subscriptionIdError = Utils.cond(
```
aherbst-broad, line 76: 👍 I like the pattern of _not_ exposing backend error messages to the presentation layer.

aherbst-broad (commented):
Very cool, had a question about the flag being checked.

src/libs/ajax.js
cahrens: Thanks-- I keep getting them confused. :( Will fix.

src/pages/billing/List.js
cahrens: Correct, I will fix it.

integration-tests/tests/workspace-dashboard.js
cahrens, line 175: This is unrelated, but @aweng98 noticed it when working on switching tests to running against staging. This override was actually never used, as the namespace is incorrect (correct override is below at line 185).

src/pages/billing/List.js
cahrens: I pushed up the change and tested with a user who is in one group but not the other. Thanks for catching this!

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
msilva-broad: a compromise I have seen work well is to have additional // Assert and // Act sections as needed for these types of scenarios.

// Assert
expect(...)

// Act
...

// Assert
expect(...)

A goal of the AAA pattern is to have reasonably crisp, small tests that give good signal on what broke if they fail.  I hope to introduce additional design patterns that should make it easier to avoid large multi-step test scenarios since the code under test will be more modularized and testable as smaller units.   Having to do AAAAAA in large tests is usually a code-smell that the code under test could use some refactoring to be easier to test.

But I am not one to let perfect be the enemy of good.  AAAA or whatever is a good compromise since it keeps things well labeled and readable for possible refactoring later.

cahrens: @msilva-broad this component is new and small-- it's just 3 fields. However, the fields depend on each other for state, and the interaction between them is what I am most interested in testing. This results in multi-step test scenarios, but I still think they can be very readable without requiring AAA labels. I would rather see comments or self-documenting code that describe the setup (which is what I think of as `Arrange`, where setup includes validating the expected initial state), the actions that are being triggered, and the results being validated.

I am going to push up a commit where I remove the AAA labels and instead increase the comments. I can revert this or go to a mixture of AAA labels and comments, but this will let you see more concretely how I would write the tests if I wasn't trying to follow the pattern of including AAA labels.

aherbst-broad (approved):
lgtm; tested locally against dev rawls.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
msilva-broad: to summarize our video chat on this topic:
- renaming this section of the test as:
     // Assert - Initial State
   should work.
- the AAA or AAAAA labels are very helpful for onboarding developers who are less experienced in writing good unit/functional tests, and also helpful for folks reviewing the tests, which will have some variations (helper methods, beforeEach, it.each, etc) in large test suites, and help reviewers who are less familiar with the code area in general.

package.json
```diff
@@ -72,8 +72,10 @@
     "postinstall": "husky install"
   },
   "devDependencies": {
+    "@testing-library/dom": "^8.17.1",
```
nawatts, line 75: Do we need a direct dependency on this? I didn't see any direct imports from it. `@testing-library/react` depends on it, and I think these queries can be imported. from `@testing-library/react`.

cahrens: I got a build error without this.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
cahrens: @msilva-broad I pushed up the changes we talked about.

package.json
cahrens: Just tried again and confirmed that I get an error:

```
Cannot find module '@testing-library/dom/dist/helpers.js' from '/Users/roberts/Documents/source_new/terra-ui/.yarn/__virtual__/@testing-library-user-event-virtual-3c4cd93763/0/cache/@testing-library-user-event-npm-14.4.3-2d1a75355f-852c48ea6d.zip/node_modules/@testing-library/user-event/dist/cjs/utils/misc'

Comment (nawatts) [reply package.json:75]: 👍 I guess v8.14.0 that got included with `@testing-library/react` is missing that file.

https://github.com/DataBiosphere/terra-ui/blob/0d45cf4b022531ced0940ebb6646dc31fc5ec3bb/yarn.lock#L2512-L2542

File: src/libs/ajax.js [hunk 13]
```diff
@@ -628,6 +638,12 @@ const Billing = signal => ({
       await billing.addProjectUser(projectName, _.difference(newRoles, oldRoles), email)
       return billing.removeProjectUser(projectName, _.difference(oldRoles, newRoles), email)
     }
+  },
+
+  listAzureManagedApplications: async subscriptionId => {
+    const response = await fetchBillingProfileManager(`azure/v1/managedApps?azureSubscriptionId=${subscriptionId}`,
+      _.merge(authOpts(), { signal }))
+    return response.json()
   }
```
dmohs, line 647: I believe this more concise equivalent improves readability. Your call:
```suggestion
  listAzureManagedApplications: subscriptionId =>
    fetchBillingProfileManager(`azure/v1/managedApps?azureSubscriptionId=${subscriptionId}`,
      _.merge(authOpts(), { signal })).then(r => r.json())
```

src/pages/billing/CreateAzureBillingProjectModal.js
```diff
@@ -0,0 +1,151 @@
...
+        async () => {
+          try {
+            const managedApps = await Ajax(signal).Billing.listAzureManagedApplications(subscriptionId)
+            setManagedApps(managedApps.managedApps)
+            setErrorFetchingManagedApps(false)
+          } catch (obj) {
+            setManagedApps([])
+            setChosenManagedApp('')
+            setErrorFetchingManagedApps(true)
+            // We can't rely on the formatting of the error, so show a generic message but include the error in the console for debugging purposes.
```
dmohs, line 66: Good call.

src/pages/billing/CreateAzureBillingProjectModal.js
```diff
@@ -0,0 +1,151 @@
...
+      h(IdContainer, [id => h(Fragment, [
+        h(FormLabel, { htmlFor: id, required: true }, ['Managed application']),
+        div({ style: { fontSize: 14 } }, [
+          h(Select, {
+            id,
+            isMulti: false,
+            placeholder: 'Select a managed application',
+            isDisabled: errorFetchingManagedApps !== false || !!subscriptionIdError,
+            value: chosenManagedApp,
+            onChange: ({ value }) => setChosenManagedApp(value),
```
dmohs, line 136: Not a suggestion, just something I tried to see how it would look and sharing the result.
```javascript
            onChange: _.flow(_.get('value'), setChosenManagedApp),
```

src/pages/billing/List.js
```diff
@@ -283,6 +315,10 @@ export const BillingList = ({ queryParams: { selectedName } }) => {
     setIsAlphaSpendReportUser(await Ajax(signal).Groups.group(getConfig().alphaSpendReportGroup).isMember())
   })
 
+  const loadAlphaAzureMember = reportErrorAndRethrow('Error loading azure alpha group membership')(async () => {
+    setIsAlphaAzureUser(await Ajax(signal).Groups.group(getConfig().alphaAzureGroup).isMember())
+  })
```
dmohs, line 320: I sometimes think the promise methods can make the flow more clear. Your call:
```suggestion
  const loadAlphaAzureMember = reportErrorAndRethrow('Error loading azure alpha group membership')(() =>
    Ajax(signal).Groups.group(getConfig().alphaAzureGroup).isMember()
    .then(setIsAlphaAzureUser)
  )
```

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
dmohs: Great PR comment.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
```diff
@@ -0,0 +1,239 @@
...
+    verifyDisabled(getCreateButton())
+
+    // Act - Select one of the managed apps
+    await userEvent.click(getManagedAppInput())
+    const selectOption = await screen.findByText(appName)
+    await userEvent.click(selectOption)
+    // Assert
+    verifyEnabled(getCreateButton())
+
+    // Act - Click Create
```
dmohs, line 188: I don't see how this adds value to the code, which already says this.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
```diff
@@ -0,0 +1,239 @@
...
+
+    // Assert
+    verifyEnabled(getCreateButton())
+
+    // Act - Click Create
+    await act(async () => {
+      await userEvent.click(getCreateButton())
+    })
+
+    // Assert - Verify expected error message and Create button disabled.
```
dmohs, line 234: I suspect you are getting conflicting advice, but my judgment is that the readability of this entire section could be improved simply by removing all instances of, "Arrange," "Act," and "Assert," leaving behind just the useful comments by themselves.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
```diff
@@ -0,0 +1,239 @@
...
+      await userEvent.type(getBillingProjectInput(), projectName)
+      await userEvent.type(getSubscriptionInput(), uuid())
+    })
+    await waitFor(() => verifyEnabled(getManagedAppInput()))
+    // Select one of the managed apps
+    await userEvent.click(getManagedAppInput())
+    const selectOption = await screen.findByText(appName)
+    await userEvent.click(selectOption)
+
+    // Assert
```
dmohs, line 226: I fail to see how comments like this one add enough value to pay for the vertical space used.

dmohs (approved):

src/libs/ajax.js
cahrens: I agree this is better, thanks!

src/pages/billing/CreateAzureBillingProjectModal.js
cahrens: To me this is less clear/more difficult to read.

src/pages/billing/List.js
cahrens: I tend to forget this way of writing the code, but I do think it's somewhat cleaner. I'll switch it.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
msilva-broad: The Arrange Act Assert comments are to follow the AAA pattern for unit testing, a common way to organize test methods for easier glanceability and promotion of other good practices and assist in providing consistency and learnability for developer onboarding.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
cahrens: I tend to agree, but @msilva-broad and I talked and there is a strong desire to make the tests follow a standardized pattern that will make it easier for others to skim them and understand what is going on. So for the purposes of standardization/education, I'm OK with it being slightly more verbose.

src/libs/ajax.js
cahrens: @dmohs I do still need to have the `return`, and then I get a warning about `async` not having an `await`. Do I just remove the `async`?

cahrens: Actually, I think I'm going to keep the code as-is because it follows the pattern of the other code in Ajax, and I am getting confused then about how some of the error handling would work.

src/pages/billing/List.js
cahrens: This lead to some lint warnings which I'm sure I can fix, but I think I'm going to stick with the async/await syntax.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
```diff
@@ -0,0 +1,238 @@
...
+    const listAzureManagedApplications = jest.fn(() => Promise.resolve({ managedApps: [] }))
+    Ajax.mockImplementation(() => { return { Billing: { listAzureManagedApplications } } })
+
+    // Act - Supply valid UUID
+    await act(async () => {
+      await userEvent.type(getSubscriptionInput(), subscriptionId)
+    })
+
+    // Assert
+    expect(listAzureManagedApplications).toHaveBeenCalledWith(subscriptionId)
```
msilva-broad, line 117: I recommend pairing args assertions like this with a call count == 1 check.

cahrens: I'll keep that in mind…. I actually had the call count check first, but decided it was overkill. Note that if the mock isn't called, the error `toHaveBeenCalledWith` is very clear, so the only thing you are losing is checking that it isn't called more than once.

src/libs/ajax.js
dmohs: When one removes the braces, `return` is implied. For example, the identity function can be defined as:
```javascript
x => x
```
Labeling a function with `async` causes it to return a promise, regardless of whether you `await` within the body. Since the example I gave returns a promise, there is no need for any other decoration.

src/pages/billing/_tests/CreateAzureBillingProjectModal.test.js
dmohs: I understand the technique, but the value is not obvious to me. I don't see much difference between these comments and any others, meaning that if you have to label the code "Act" vs. "Assert," I'd prefer you to see if you can rewrite it to make what you're doing obvious and eliminate the need for the comment. I don't treat test code any differently than other code in this regard.

PR #3349, opened 2022-08-31
[SU-160] Profile page redesign

MatthewBemis:
A redesign of the profile page, see the mocks in SU-160 https://projects.invisionapp.com/share/EM12WVICD9Q5#/screens/468966709

I dropped the "Other" field for the areas of research because it was actually a huge PITA to get working properly given how these fields are stored in Thurloe. I'll add that later on if we want but I want to get this moving and not hold it up over something silly.

Proxy group also wasn't in the redesign but I didn't feel like it was a good idea to drop that since it's useful to users.

Personal info tab:
Screen Shot 2022-09-01 at 9 00 16 AM

External identities tab:
Screen Shot 2022-08-31 at 4 41 28 PM

Notification settings tab:
Screen Shot 2022-08-31 at 4 41 38 PM

<!--
Hello, friend!
Thanks!

--!>


src/pages/Profile.js
```diff
@@ -355,22 +356,131 @@ const PassportLinker = ({ queryParams: { state, code } = {}, provider, prettyNam
...
+  return h(PageBox, { role: 'main', style: { flexGrow: 1 }, variant: PageBoxVariants.LIGHT }, [
+    h(NihLink, { nihToken: queryParams?.['nih-username-token'] }),
+    _.map(provider => h(FenceLink, { key: provider.key, provider }), allProviders),
+    !!getConfig().externalCredsUrlRoot && h(PassportLinker, { queryParams, provider: 'ras', prettyName: 'RAS' })
+  ])
+}
+
+const NotificationCheckbox = ({ notificationKeys, setSaving, prefsData }) => {
+  const notificationKeysWithValue = ({ notificationKeys, value }) => {
+    return Object.assign(...notificationKeys.map(notificationKey => ({ [notificationKey]: `${JSON.stringify(value)}` })))
```
nawatts: Lodash's [fromPairs](https://lodash.com/docs/4.17.15#fromPairs) might be useful here. Something like:
```suggestion
    return _.fromPairs(_.map(notificationKey => [notificationKey, JSON.stringify(value)], notificationKeys))
```

src/pages/Profile.js
```diff
@@ -355,22 +356,131 @@ const PassportLinker = ({ queryParams: { state, code } = {}, provider, prettyNam
...
+        prefsData,
+        label: 'Workspace Access Removed',
+        notificationKeys: ['notifications/WorkspaceRemovedNotification']
+      })
+    ]),
+    div({ style: Style.cardList.toolbarContainer }, [
+      h2({ style: { ...Style.elements.sectionHeader, marginTop: '2rem', textTransform: 'uppercase' } }, [
+        'Submission Notifications',
+        h(InfoBox, { style: { marginLeft: '0.5rem' } },
+          'You may receive email notifications regarding terminal submissions in Terra (i.e. if a submission has succeeded, failed, or been aborted). You may opt in or out of receiving these emails for individual workspaces.'
```
nawatts: "terminal submissions" seems kind of jargony.

```suggestion
          'You may receive email notifications when a submission has succeeded, failed, or been aborted. You may opt in or out of receiving these emails for individual workspaces.'
```

src/pages/Profile.js
```diff
@@ -355,22 +356,131 @@ const PassportLinker = ({ queryParams: { state, code } = {}, provider, prettyNam
...
+    !!getConfig().externalCredsUrlRoot && h(PassportLinker, { queryParams, provider: 'ras', prettyName: 'RAS' })
+  ])
+}
+
+const NotificationCheckbox = ({ notificationKeys, setSaving, prefsData }) => {
+  const notificationKeysWithValue = ({ notificationKeys, value }) => {
+    return Object.assign(...notificationKeys.map(notificationKey => ({ [notificationKey]: `${JSON.stringify(value)}` })))
+  }
+
+  return h(LabeledCheckbox, {
```
nawatts: These checkboxes are missing label text. It might be better to use `Checkbox` instead of `LabeledCheckbox` and add `aria-label`s. Alternatively, you could add visually hidden / screen reader only text for the label.

src/pages/Profile.js
```diff
@@ -355,22 +356,131 @@ const PassportLinker = ({ queryParams: { state, code } = {}, provider, prettyNam
...
+    )
+  ])
+}
+
+const PersonalInfoTab = ({ setSaving }) => {
   const [profileInfo, setProfileInfo] = useState(() => _.mapValues(v => v === 'N/A' ? '' : v, authStore.get().profile))
   const [proxyGroup, setProxyGroup] = useState()
-  const [saving, setSaving] = useState()
+  const { researchArea } = profileInfo
+  const [areasOfResearch, setAreasOfResearch] = useState(_.isEmpty(researchArea) ? [] : _.split(',', researchArea))
```
nawatts: Since `researchArea` is part of `profileInfo` and `profileInfo` is stored in state, is there a need for a separate state variable here? With two state variables, they may get out of sync.

src/pages/Profile.js
```diff
@@ -465,48 +573,82 @@ const Profile = ({ queryParams }) => {
...
+        ]),
 
-            line([
-              textField('programLocationCity', 'City'),
-              textField('programLocationState', 'State')
-            ]),
-            line([
-              textField('programLocationCountry', 'Country')
-            ]),
+        sectionTitle('What is your area of research? (multi-select)'),
```
nawatts: Instead of "multi-select", maybe add a paragraph below this that says something like "Check all that apply."?

nawatts (changes requested):
Requesting changes over checkbox labels. Currently, all you get in a screen reader is "checked, checkbox".

src/pages/Profile.js
MatthewBemis: Updated in 5103bd555bfcbae02e9ae3e56748bfeb13d26155

src/pages/Profile.js
MatthewBemis: Updated to pass the `label` text through to `aria-label` in 3d2fff105e421b1562be4fc820481a7ce8540cf7

src/pages/Profile.js
MatthewBemis: Good call- removed the extra state in b6c918f788607a70867e28791012ced4e4d48857

nawatts:
It looks like `text-overflow: ellipsis` isn't working for workspace names.

Screen Shot 2022-09-01 at 2 09 54 PM

Would it make sense to put the checkbox first and leave the rest of the card for the workspace name instead of blocking off half the card for the checkbox?

nawatts (approved):

MatthewBemis:
@nawatts I'll see if I can fix the text overflow issue. This might be premature on my part, but I think if we want to keep workspace-level notifications on this page, then we'll ultimately end up filling out the space to the right with more checkboxes (and change the header to not be submission-specific), which I think would make more sense with the workspace name on the left. There's already 1 existing workspace-level notification in our backlog that we can add pretty easily.

src/pages/Profile.js
```diff
@@ -465,48 +573,84 @@ const Profile = ({ queryParams }) => {
...
-                await Promise.all([
-                  Ajax().User.profile.set(_.pickBy(_.identity, profileData)),
-                  Ajax().User.profile.setPreferences(prefsData)
-                ])
-                await refreshTerraProfile()
-              }),
-              disabled: !!errors,
-              tooltip: !!errors && 'Please fill out all required fields'
-            }, ['Save Profile'])
+        line([
```
andy7i: minor/nit - maybe remove the top margin? There's a lot of space between `Check all that apply` and the checkboxes
Screen Shot 2022-09-07 at 2 12 28 PM

src/pages/Profile.js
```diff
@@ -465,48 +573,84 @@ const Profile = ({ queryParams }) => {
                 }, ['user guide.'])
               ])
             ]),
-            div({ style: { margin: '1rem' } }, [proxyGroup]),
+            div({ style: { margin: '1rem' } }, proxyGroup ? proxyGroup : 'Loading...')
+          ])
+        ]),
 
-            sectionTitle('Location'),
+        sectionTitle('What is your area of research?'),
```
andy7i, line 586: I couldn't find any info about this section. Is there a mock for this? / could you point me to more info about this section?

src/pages/Profile.js
```diff
@@ -355,22 +356,131 @@ const PassportLinker = ({ queryParams: { state, code } = {}, provider, prettyNam
...
-  // State
+const ExternalIdentitiesTab = ({ queryParams }) => {
+  return h(PageBox, { role: 'main', style: { flexGrow: 1 }, variant: PageBoxVariants.LIGHT }, [
+    h(NihLink, { nihToken: queryParams?.['nih-username-token'] }),
+    _.map(provider => h(FenceLink, { key: provider.key, provider }), allProviders),
+    !!getConfig().externalCredsUrlRoot && h(PassportLinker, { queryParams, provider: 'ras', prettyName: 'RAS' })
+  ])
+}
+
+const NotificationCheckbox = ({ notificationKeys, label, setSaving, prefsData }) => {
```
andy7i, line 368: The mocks show a couple additional features: a checkbox for `Opt In All`, and splitting out the `Opt in` checkboxes into 2 columns for `Banners` and `Emails`. Is the plan to implement those in future iterations or has the design changed?

andy7i (approved):
https://broadworkbench.atlassian.net/browse/SU-160

tested locally (but did not test notifications)

overall, looks nice! I can see these changes being very useful for users.

I had a couple of questions/suggestions but nothing blocking. 👍

src/pages/Profile.js
MatthewBemis: The `Opt In All` is something that will be coming in a future iteration, likely as part of [SUP-519](https://broadworkbench.atlassian.net/browse/SUP-519).

The `Banners` and `Emails` separation is a bit outdated- we don't actually have any real plans to support banners for these types of notifications

src/pages/Profile.js
MatthewBemis: This came directly as a request from Eric Karofsky as part of user research, but it was after the mocks were designed so it's not reflected in there

src/pages/Profile.js
MatthewBemis: Reduced margin by half in 95398d07aa7457ffcacaa1593683672a79301a6b

PR #3350, opened 2022-09-01
[SU-199] Make recently viewed workspace panel collapsible

MatthewBemis:
From user feedback.

https://user-images.githubusercontent.com/7257391/187914918-6d3de5e5-3dbd-4c45-a62f-b6868c6b51d3.mov

<!--
Hello, friend!
Thanks!

--!>

### DIFF
--- a/src/pages/workspaces/List.js
+++ b/src/pages/workspaces/List.js
@@ -3,6 +3,7 @@ import _ from 'lodash/fp'
 import { useEffect, useMemo, useState } from 'react'
 import { div, h, p, span } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
+import Collapse from 'src/components/Collapse'
 import { HeaderRenderer, Link, Select, topSpinnerOverlay, transparentSpinnerOverlay } from 'src/components/common'
 import FooterWrapper from 'src/components/FooterWrapper'
 import { icon } from 'src/components/icons'
@@ -23,7 +24,7 @@ import colors from 'src/libs/colors'
 import { withErrorReporting } from 'src/libs/error'
 import Events from 'src/libs/events'
 import * as Nav from 'src/libs/nav'
-import { getLocalPref } from 'src/libs/prefs'
+import { getLocalPref, setLocalPref } from 'src/libs/prefs'
 import { useCancellation, useOnMount, useStore } from 'src/libs/react-utils'
 import { authStore } from 'src/libs/state'
 import * as Style from 'src/libs/style'
@@ -101,6 +102,9 @@ export const WorkspaceList = () => {
   //A user may have lost access to a workspace after viewing it, so we'll filter those out just in case
   const recentlyViewed = useMemo(() => _.filter(w => _.find({ workspace: { workspaceId: w.workspaceId } }, workspaces), getLocalPref(recentlyViewedPersistenceId)?.recentlyViewed || []), [workspaces])
 
+  const persistenceId = `workspaces/list`
+  const [recentlyViewedOpen, setRecentlyViewedOpen] = useState(() => getLocalPref(persistenceId)?.recentlyViewedOpen)
+
   const { query } = Nav.useRoute()
   const filter = query.filter || ''
   // Using the EMPTY_LIST constant as a default value instead of creating a new empty array on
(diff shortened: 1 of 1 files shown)


src/pages/workspaces/List.js
```diff
@@ -128,6 +132,10 @@ export const WorkspaceList = () => {
     loadFeatured()
   })
 
+  useEffect(() => {
+    setLocalPref(persistenceId, { recentlyViewedOpen })
+  }, [recentlyViewedOpen]) // eslint-disable-line react-hooks/exhaustive-deps
```
nawatts: Unfortunately, `react-hooks/exhaustive-deps` doesn't allow ignoring specific dependencies. That being the case, in my opinion, it's preferable to leave this rule enabled and add the unnecessary `persistenceId` dependency rather than chance missing a dependency if the effect function is changed later.

src/pages/workspaces/List.js
```diff
@@ -101,6 +102,9 @@ export const WorkspaceList = () => {
   //A user may have lost access to a workspace after viewing it, so we'll filter those out just in case
   const recentlyViewed = useMemo(() => _.filter(w => _.find({ workspace: { workspaceId: w.workspaceId } }, workspaces), getLocalPref(recentlyViewedPersistenceId)?.recentlyViewed || []), [workspaces])
 
+  const persistenceId = `workspaces/list`
+  const [recentlyViewedOpen, setRecentlyViewedOpen] = useState(() => getLocalPref(persistenceId)?.recentlyViewedOpen)
```
nawatts: Currently, there are two places that handle the case where `recentlyViewedOpen` is undefined: the Collapse component's `initialOpenState` and `onClick`. Alternatively, that could be handled once here.
```suggestion
  const [recentlyViewedOpen, setRecentlyViewedOpen] = useState(() => _.defaultTo(true, getLocalPref(persistenceId)?.recentlyViewedOpen))
```

src/pages/workspaces/List.js
```diff
@@ -356,8 +364,13 @@ export const WorkspaceList = () => {
           href: 'https://support.terra.bio/hc/en-us/articles/360024743371-Working-with-workspaces'
         }, ['Learn more about workspaces.'])
       ]),
-      !_.isEmpty(workspaces) && !_.isEmpty(recentlyViewed) && div([
-        p({ style: { textTransform: 'uppercase' } }, 'Recently viewed'),
+      !_.isEmpty(workspaces) && !_.isEmpty(recentlyViewed) && h(Collapse, {
+        title: 'Recently Viewed',
+        initialOpenState: recentlyViewedOpen !== undefined ? recentlyViewedOpen : true,
```
nawatts: If an undefined local preference is handled in the state initialization, then this becomes
```suggestion
        initialOpenState: recentlyViewedOpen,
```

src/pages/workspaces/List.js
```diff
@@ -356,8 +364,13 @@ export const WorkspaceList = () => {
...
+        noTitleWrap: true,
+        onClick: () => setRecentlyViewedOpen(recentlyViewedOpen === undefined ? false : !recentlyViewedOpen),
```
nawatts: This could use the [function form of setState](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous). If an undefined local preference is handled in the state initialization, it could be:
```suggestion
        onClick: () => setRecentlyViewedOpen(v => !v),
```

nawatts (approved):

src/pages/workspaces/List.js
MatthewBemis: Much cleaner, thank you!

src/pages/workspaces/List.js
MatthewBemis: Updated in 0c10d40fb89481394ebfac66175db863a634bef1

cahrens:
If I press space on a workspace card in the recently viewed section, it both takes me to the workspace and collapses the section (which the history remembers). Both Enter and Space is correct for both an anchor tag and a button, but I think you need to make sure that when a workspace card is focused, the event is consumed so it doesn't bubble up to the collapse button. To toggle the collapse, it will be necessary to shift-Tab back to the title control.

MatthewBemis:
@cahrens nice catch! Fixed in 659949162999a48d43e21634ccbd783479877d71

cahrens (approved):

PR #3351, opened 2022-09-01
DC-531: Revert preview dataset and link to workspace

fboulnois:
Reverts DataBiosphere/terra-ui#3334
Reverts DataBiosphere/terra-ui#3317 

We've seen an increasing amount of errors in `link-to-workspace`, `run-catalog-workflow`, and the `preview workspace` feature is also partially broken. This also is blocking Terra release.

- For preview workspace, the data we are getting back from certain workspaces in prod is very different from the ones we have in dev, and this is causing react to crash
- For export to workspace, tests are failing in alpha and staging but not in dev

### DIFF
--- a/integration-tests/tests/link-to-new-workspace.js
+++ b/integration-tests/tests/link-to-new-workspace.js
@@ -46,5 +46,6 @@ const testLinkToNewWorkspaceFn = withUserToken(async ({ billingProject, page, te
 registerTest({
   name: 'link-to-new-workspace',
   fn: testLinkToNewWorkspaceFn,
-  timeout: 2 * 60 * 1000
+  timeout: 2 * 60 * 1000,
+  targetEnvironments: []
 })
--- a/integration-tests/tests/preview-dataset.js
+++ b/integration-tests/tests/preview-dataset.js
@@ -26,5 +26,6 @@ const testPreviewDatasetFn = withUserToken(async ({ testUrl, page, token }) => {
 registerTest({
   name: 'preview-dataset',
   fn: testPreviewDatasetFn,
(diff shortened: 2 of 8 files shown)


pshapiro4broad (approved):

petesantos (commented):
Does this need to be a full revert, or can we disable the buttons and tests?
The rationale being: if we have the current code in place, the follow up PR to correct the bug will be smaller.

aweng98 (approved):
Thank you for disabling failing tests.

nawatts:
> For export to workspace, tests are failing in alpha and staging but not in dev

Looking forward to having PR tests run against staging to avoid issues like this.

cahrens:
@fboulnois I am taking a look, but if the UI reviewers who originally reviewed these changes want to jump in and take over, that's great with me. This PR is more complex than a simple commit revert.

pshapiro4broad:
@cahrens The PR may look complicated but I believe the commits are a direct revert of the earlier commits, not a manual change to back out the earlier changes.

cahrens:
@pshapiro4broad what about the last commit?

cahrens:
@fboulnois it looks like the last commit is actually no longer needed-- those changes have already merged, correct? Would you mind removing the commit from this PR so that this PR is just simple reverts of previous commits?

fboulnois:
@cahrens the last commit is needed because two PRs were reverted but the polling functionality (which was introduced in one of those PRs) needs to stay in place (in other words one is a partial revert). But I can squash the commits instead, if you'd rather see that (I think they get squashed when merged though?)

cahrens (approved):

PR #3352, opened 2022-09-01
[PROD-732] Renaming analyses deletes the file instead of renaming.

n8landolt:
https://broadworkbench.atlassian.net/browse/PROD-732

## Issue:

When a user went to rename a file on the analyses tab, the file would instead get deleted.

## Root cause:
The async functions that handle copying and deleting weren't running sequentially - the delete function would be fired immediately after the copyTo (rename) funciton, when it SHOULD have been after copyTo completed.

## Solution
Removed function that incorrectly attempts retries. 
Fixed async/await would function properly.

Re-enabled rename button.

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -1301,7 +1301,7 @@ const Buckets = signal => ({
 
       rename: async newName => {
         await copy(newName, bucket, false)
-        await doDelete()
+        return doDelete()
       }
     }
   },
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -55,6 +55,26 @@ const withCancellation = wrappedFetch => async (...args) => {
   }
 }
 
+const withRetryOnError = _.curry((shouldNotRetryFn, wrappedFetch) => async (...args) => {
+  const timeout = 5000
+  const somePointInTheFuture = Date.now() + timeout
+  const maxDelayIncrement = 1500
+  const minDelay = 500
+
+  while (Date.now() < somePointInTheFuture) {
+    const until = Math.random() * maxDelayIncrement + minDelay
+    try {
+      return await Utils.withDelay(until, wrappedFetch)(...args)
+    } catch (error) {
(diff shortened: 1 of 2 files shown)


src/libs/ajax.js
```diff
@@ -1338,7 +1318,7 @@ const Buckets = signal => ({
     const encodeFileName = name => encodeAnalysisName(getFileName(name))
 
     const doCopy = async (newName, newBucket, body) => {
-      fetchBuckets(
+      return fetchBuckets(
```
nawatts, line 1341: I'm skeptical that `withRetryOnError` needs to be removed. As I understand it, the issue seems to be that the UI is not recognizing when the copy request fails and then proceeding with deleting the original file. It looks like this `return` will fix that issue.

@petesantos any comment on `withRetryOnError` (added for [DC-324](https://broadworkbench.atlassian.net/browse/DC-324))?

n8landolt: I'll test it again with withRetryOnError, however, I still don't believe that function is working as it should be. Let me see...

n8landolt: Seems like things worked when adding back in withRetryOnError.

I think what was happening that caused me to remove it - the delete happened sneakily after the first request was triggered, then the console flooded with retries. It's likely working as intended - just the lack of this return made the outcome strange.

n8landolt: Loading spinners last significantly longer with this function being used, FWIW.

nawatts: > I think what was happening that caused me to remove it - the delete happened sneakily after the first request was triggered, then the console flooded with retries. It's likely working as intended - just the lack of this return made the outcome strange.

> Loading spinners last significantly longer with this function being used, FWIW.

Makes sense. It sounds like the delete request was being sent right after the copy request, so there was possibly a race. Vs now they'd be run sequentially.

nawatts (approved):

Qi77Qi (approved):

PR #3353, opened 2022-09-06
[SUP-546] Add provenance for files written by workflows

nawatts:
This adds a "Where did this file come from?" section to the modal shown when viewing a file from a data table or the workspace files browser.

For files that are workflow outputs, it shows a link to the workflow and submission.

Screen Shot 2022-09-06 at 10 54 10 AM

This is behind a feature flag. To enable it, run `window.configOverridesStore.set({ isDataTableProvenanceEnabled: true })` in the console and reload.

### DIFF
--- a/src/components/UriViewer.js
+++ b/src/components/UriViewer.js
@@ -5,6 +5,7 @@ import { div, h, img, input } from 'react-hyperscript-helpers'
 import { requesterPaysWrapper, withRequesterPaysHandler } from 'src/components/bucket-utils'
 import Collapse from 'src/components/Collapse'
 import { ButtonPrimary, ClipboardButton, Link } from 'src/components/common'
+import { FileProvenance } from 'src/components/data/data-table-provenance'
 import { getDownloadCommand, getUserProjectForWorkspace, parseGsUri } from 'src/components/data/data-utils'
 import { spinner } from 'src/components/icons'
 import Modal from 'src/components/Modal'
@@ -12,6 +13,7 @@ import DownloadPrices from 'src/data/download-prices'
 import { Ajax } from 'src/libs/ajax'
 import { bucketBrowserUrl } from 'src/libs/auth'
 import colors from 'src/libs/colors'
+import { isDataTableProvenanceEnabled } from 'src/libs/config'
 import Events, { extractWorkspaceDetails } from 'src/libs/events'
 import { useCancellation, useOnMount, withDisplayName } from 'src/libs/react-utils'
 import { knownBucketRequesterPaysStatuses, workspaceStore } from 'src/libs/state'
@@ -267,7 +269,11 @@ const UriViewer = _.flow(
             els.data(new Date(updated).toLocaleString())
           ])
         ]),
-        div({ style: { fontSize: 10 } }, ['* Estimated. Download cost may be higher in China or Australia.'])
+        div({ style: { fontSize: 10 } }, ['* Estimated. Download cost may be higher in China or Australia.']),
+        isDataTableProvenanceEnabled() && h(Fragment, [
+          div({ style: { margin: '2rem 0 0.5rem', fontWeight: 500 } }, ['Where did this file come from?']),
+          h(FileProvenance, { workspace, fileUrl: uri })
+        ])
       ])],
       () => h(Fragment, [
         isGs(uri) ? 'Loading metadata...' : 'Resolving DOS object...',
--- a/src/components/UriViewer.js
+++ b/src/components/UriViewer.js
@@ -164,7 +164,9 @@ const DownloadButton = ({ uri, metadata: { bucket, name, fileName, size }, acces
 const UriViewer = _.flow(
   withDisplayName('UriViewer'),
   requesterPaysWrapper({ onDismiss: ({ onDismiss }) => onDismiss() })
-)(({ googleProject, uri, onDismiss, onRequesterPaysError }) => {
+)(({ workspace, uri, onDismiss, onRequesterPaysError }) => {
+  const { workspace: { googleProject } } = workspace
+
   const signal = useCancellation()
   const [metadata, setMetadata] = useState()
   const [loadingError, setLoadingError] = useState()
@@ -275,7 +277,7 @@ const UriViewer = _.flow(
   ])
 })
 
-export const UriViewerLink = ({ uri, googleProject }) => {
+export const UriViewerLink = ({ uri, workspace }) => {
   const [modalOpen, setModalOpen] = useState(false)
   return h(Fragment, [
     h(Link, {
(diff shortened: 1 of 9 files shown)


MatthewBemis (approved):

cahrens:
@nawatts What would you think about moving this information into "More information" (which would mean changing where the estimated asterisk is displayed also)? My thoughts are:

1) It's odd to have "More information" be collapsible above an element that is always present.
2) The provenance information seems like "More information" about the file.
3) The dialog is getting quite tall, which I assume is the reason that "More information" is collapsible.
4) This dialog is pretty crazy and the information seems to have little structure. The more structure we can give it and the less information we show by default, the better.

image

nawatts:
> What would you think about moving this information into "More information"

Those reasons all make sense to me. Done in https://github.com/DataBiosphere/terra-ui/pull/3353/commits/42cd00de852c52bd1898305ec8a499faf2e84aa0.

Screen Shot 2022-09-07 at 10 06 13 AM

src/components/data/data-table-provenance.js
```diff
@@ -38,3 +38,63 @@ export const DataTableColumnProvenance = ({ workspace, column, provenance }) =>
     ])
   ])
 }
+
+export const FileProvenance = ({ workspace, fileUrl }) => {
```
cahrens, line 42: @nawatts are you thinking of adding a unit test for this?

nawatts: Added tests for this component in https://github.com/DataBiosphere/terra-ui/pull/3353/commits/d0d175c0e04588d7afa7d31546bef5868b13477c, which also revealed a bug (both the workflow and submission links were labeled with the submission ID).

nawatts:
Rebased to run tests against staging instead of dev.

src/components/data/data-table-provenance.test.js
```diff
@@ -0,0 +1,58 @@
...
+      { type: fileProvenanceTypes.workflowOutput, submissionId: '842371e3-2cea-4929-94f4-dda074e6fd71', workflowId: 'e8e3447c-10c7-4265-b812-e6a5183e99a5' },
+      'This file is an output of workflow e8e3447c-10c7-4265-b812-e6a5183e99a5 (part of submission 842371e3-2cea-4929-94f4-dda074e6fd71).'
+    ],
+    [
+      { type: fileProvenanceTypes.workflowLog, submissionId: '842371e3-2cea-4929-94f4-dda074e6fd71', workflowId: 'e8e3447c-10c7-4265-b812-e6a5183e99a5' },
+      'This file is a log from workflow e8e3447c-10c7-4265-b812-e6a5183e99a5 (part of submission 842371e3-2cea-4929-94f4-dda074e6fd71).'
+    ]
+  ]
+
+  afterEach(() => {
```
cahrens: Is restoring the mocks necessary? I ask only because I didn't do it in my own tests.

src/components/UriViewer.js
```diff
@@ -5,13 +5,15 @@ import { div, h, img, input } from 'react-hyperscript-helpers'
...
 import { ButtonPrimary, ClipboardButton, Link } from 'src/components/common'
+import { FileProvenance } from 'src/components/data/data-table-provenance'
 import { getDownloadCommand, getUserProjectForWorkspace, parseGsUri } from 'src/components/data/data-utils'
 import { spinner } from 'src/components/icons'
 import Modal from 'src/components/Modal'
 import DownloadPrices from 'src/data/download-prices'
 import { Ajax } from 'src/libs/ajax'
 import { bucketBrowserUrl } from 'src/libs/auth'
 import colors from 'src/libs/colors'
+import { isDataTableProvenanceEnabled } from 'src/libs/config'
```
cahrens, line 16: Out of curiosity, why put this behind a feature flag?

src/libs/data-table-provenance.js
```diff
@@ -88,3 +90,98 @@ export const useColumnProvenance = (workspace, entityType) => {
...
+
+const isLog = (url, task) => _.some(log => _.includes(url, [log.backendLogs.log, log.stdout, log.stderr]), task.logs)
+
+const isOutput = (url, task) => _.some(
+  // outputs can contain a value, a collection of values, or a collection of collections of values
+  output => output === url || _.includes(url, output) || _.some(_.includes(url), output),
+  task.outputs
+)
+
+const getFileProvenance = async (workspace, fileUrl, { signal } = {}) => {
```
cahrens: You could consider unit testing this as well.

src/components/data/data-table-provenance.js
```diff
@@ -38,3 +38,63 @@ export const DataTableColumnProvenance = ({ workspace, column, provenance }) =>
...
+        h(Link, {
+          href: Nav.getLink('workspace-submission-details', {
+            namespace, name,
+            submissionId: fileProvenance.submissionId
+          })
+        }, [fileProvenance.submissionId]),
+        ', but it was not found in workflow outputs.'
+      ])],
+      [fileProvenanceTypes.workflowOutput, () => span([
+        'This file is an output of workflow ',
```
cahrens, line 64: These links are a pretty bad user experience for screenreaders. The link text itself is a UUID that is meaningless, and the context around it isn't read out if you are only navigating links/focusable components.

I would suggest adding an aria-label to the links that describe what they are, and I don't think there is any need to actually read out the ID (which will be the behavior-- the label will be read instead of the ID). So for instance, here it would be "parent workflow", above it would be "possible parent submission", etc.

src/components/data/data-table-provenance.test.js
nawatts: Create React App configures Jest to restore mocks by default, but there's been talk of changing that setting for Terra UI. I wanted to ensure isolation between tests.
https://create-react-app.dev/docs/running-tests/#configuration

src/components/UriViewer.js
nawatts: This feature is kind of experimental, so we wanted to collect some feedback and maybe adjust before making it generally available.
https://broadworkbench.atlassian.net/browse/SU-180

src/components/data/data-table-provenance.test.js
nawatts: Restoring the mocks is not necessary here.

cahrens: I would like us to keep the default isolation between tests because otherwise the risk of accidental drool seems high.

nawatts: 👍 I think this can still be removed here because this test doesn't use `jest.spyOn`.
https://jestjs.io/docs/jest-object#jestrestoreallmocks

Removed in https://github.com/DataBiosphere/terra-ui/pull/3353/commits/05c7a643ee7fecaabd2740d6873188119bd991fc.

src/libs/data-table-provenance.js
nawatts: Added tests in https://github.com/DataBiosphere/terra-ui/pull/3353/commits/51d53a2fa19e9e46f8471c67cb7a99a4b5ab8f4e.

src/components/data/data-table-provenance.js
nawatts: Added labels in https://github.com/DataBiosphere/terra-ui/pull/3353/commits/f21e9e6119eb4eb3434f4c751cd078e2c3a31c56.

cahrens (approved):

PR #3354, opened 2022-09-06
[BT-740] Add documentation about adding marking new events as expected in Mixpanel

breilly2:
Just adding some helpful documentation.

### DIFF
--- a/src/libs/events.js
+++ b/src/libs/events.js
@@ -4,11 +4,9 @@ import { Ajax } from 'src/libs/ajax'
 import { useRoute } from 'src/libs/nav'
 
 /*
- * NOTE: In order to show up in reports, new events MUST be marked as expected in the
- * Mixpanel lexicon:
- *   https://mixpanel.com/project/2085496/view/19055/app/lexicon#transformations
- * See our Mixpanel Documentation Guide for more details:
- *   https://docs.google.com/document/d/11FTFwcSby_Y3syQCvjLz124xgv5LGd0HBhEjup59O7Q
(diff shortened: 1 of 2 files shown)


src/libs/events.js
```diff
@@ -3,7 +3,13 @@ import { useEffect } from 'react'
 import { Ajax } from 'src/libs/ajax'
 import { useRoute } from 'src/libs/nav'
 
-
+/*
+ * NOTE: In order to show up in reports, new events MUST be marked as expected in the
+ * Mixpanel lexicon:
+ *   https://mixpanel.com/project/2085496/view/19055/app/lexicon#transformations
```
nawatts: This links to the dev Lexicon, but the production Lexicon also needs to be updated and is more important.

src/libs/events.js
```diff
@@ -3,7 +3,13 @@ import { useEffect } from 'react'
...
+ * See our Mixpanel Documentation Guide for more details:
+ *   https://docs.google.com/document/d/11FTFwcSby_Y3syQCvjLz124xgv5LGd0HBhEjup59O7Q
```
nawatts: Could this go in the Terra UI wiki?

nawatts (approved):

src/libs/events.js
breilly2: I'll update with links to both.

src/libs/events.js
breilly2: I like that idea. I'll check with the authors of the Google doc to see if that makes sense to them.

jgainerdewar (approved):

PR #3356, opened 2022-09-06
[DC-533][DC-534] Revert "DC-531: Revert preview dataset and link to workspace (#3351)"

s-rubenstein:
This reverts commit 004592f8a7af983cf741828653255ede616e3c3d.

I added permissions to scarlett.flowerpicker on the readable snapshot in both staging and alpha to fix the link-to-workspace test failures, and tested locally by updating my local config to point at alpha and running the `link-to-workspace` test against that local. The long term fix is captured in a TDR ticket (already filed and I don't have the ticket number handy)

I fixed the preview bug by wrapping our default case in `_.toString`. This is not ideal, but it gets a partial preview out the door and can be further improved to handle objects properly in a follow up.


pshapiro4broad (approved):
This looks fine but I don't see the `toString` fix, is that fixed in a different PR?

cahrens:
@s-rubenstein can you answer @pshapiro4broad question about the `toString` fix?

s-rubenstein:
@pshapiro4broad I missed committing it. :D

src/pages/library/DataBrowserPreview.js
```diff
@@ -83,7 +83,7 @@ const DataBrowserPreview = ({ id }) => {
           style: { fontSize: 16, textTransform: 'none' },
           onClick: () => setViewJSON({ title: `${table}, Row ${rowIndex} - ${cellKey}`, cellData: maybeJSON })
         }, ['View JSON'])],
-        [Utils.DEFAULT, () => cellContent]
+        [Utils.DEFAULT, () => cellContent ? cellContent.toString() : '']
```
s-rubenstein: I added the ternary as well, because our linter prefers `.toString` over `_.toString`. I wanted it to be nullsafe, so this seemed like a good approach.

pshapiro4broad: Can `?.` be used here?
```suggestion
        [Utils.DEFAULT, () => cellContent?.toString()]
```

cahrens: _.toString is null-safe, so I think it's the better option.

```
The _.toString() method is used to convert the given value to a string. An empty string is returned for null and undefined values. The sign of the value -0 is preserved.

Comment (s-rubenstein) [reply src/pages/library/DataBrowserPreview.js]: @cahrens I agree, but the linter actually wouldn't let me use it:
```
ESLint found too many warnings (maximum: 0).

/Users/srubenstein/Desktop/code/terra-ui/src/pages/library/DataBrowserPreview.js
  86:31  warning  Use `function toString() { [native code] }` instead of `toString`  lodash-fp/preferred-alias

```

**Review by fboulnois — APPROVED** [review]
Thank you for this follow-up. Lgtm once this change is tested in alpha and staging.

File: src/pages/library/DataBrowserPreview.js [hunk 2]
```diff
@@ -83,7 +83,8 @@ const DataBrowserPreview = ({ id }) => {
           style: { fontSize: 16, textTransform: 'none' },
           onClick: () => setViewJSON({ title: `${table}, Row ${rowIndex} - ${cellKey}`, cellData: maybeJSON })
         }, ['View JSON'])],
-        [Utils.DEFAULT, () => cellContent]
+        [typeof cellContent === 'object', () => JSON.stringify(cellContent)],
```
cahrens: Any reason not to use `_.isObject`? It seems much friendlier.

s-rubenstein: Nope, that seems better to me!

cahrens (approved):

PR #3357, opened 2022-09-07
[WOR-481] Filter storage containers to avoid application private container

cahrens:
Fixes a bug where we showed the Leonardo PRIVATE_ACCESS container on the Workspace Dashboard, and the SAS token was for that container instead of the workspace public container.

If we end up having multiple SHARED_ACCESS containers at some point, requiring us find a way to mark "the" workspace container. However, we aren't confident yet about the best way to do that, so we are going to wait to see if it truly becomes necessary.

To test:
1) Create an Azure workspace (must be in Azure alpha group).
2) Create a Jupyter notebook in that workspace
3) Note the storage container name (ls-saturn is for "leonardo", sc is "the" workspace bucket).
4) You can also check that the SAS tokens are different (the container name is used in the request to get the SAS token).

Before:
image

After:
image

### DIFF
--- a/integration-tests/tests/workspace-dashboard.js
+++ b/integration-tests/tests/workspace-dashboard.js
@@ -157,7 +157,15 @@ const setAzureAjaxMockValues = async (testPage, namespace, name, workspaceDescri
       },
       {
         metadata: {
-          resourceType: 'AZURE_STORAGE_CONTAINER'
+          resourceType: 'AZURE_STORAGE_CONTAINER',
+          controlledResourceMetadata: { accessScope: 'PRIVATE_ACCESS' }
+        },
+        resourceAttributes: { azureStorageContainer: { storageAccountId: 'dummy-sa-resource-id', storageContainerName: 'private-sc-name' } }
+      },
+      {
+        metadata: {
+          resourceType: 'AZURE_STORAGE_CONTAINER',
+          controlledResourceMetadata: { accessScope: 'SHARED_ACCESS' }
         },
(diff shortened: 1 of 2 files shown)


src/libs/ajax.js
```diff
@@ -1138,7 +1138,7 @@ const AzureStorage = signal => ({
     } else {
       const container = _.find(
         {
-          metadata: { resourceType: 'AZURE_STORAGE_CONTAINER' },
+          metadata: { resourceType: 'AZURE_STORAGE_CONTAINER', controlledResourceMetadata: { accessScope: 'SHARED_ACCESS' } },
```
cahrens, line 1141: Another difference between the Leo container and the workspace one is that the Leo container has `managedBy: "APPLICATION"` vs. `managedBy: "USER"`. My thought was that checking the `accessScope` is sufficient, but we could check both if we wish. @aherbst-broad do you have an opinion?

integration-tests/tests/workspace-dashboard.js
```diff
@@ -157,7 +157,15 @@ const setAzureAjaxMockValues = async (testPage, namespace, name, workspaceDescri
       },
       {
         metadata: {
-          resourceType: 'AZURE_STORAGE_CONTAINER'
+          resourceType: 'AZURE_STORAGE_CONTAINER',
```
cahrens, line 160: Without the fix, this causes the test to fail because we don't find the storage container we expect on the workspace dashboard page.

nawatts (approved):

src/libs/ajax.js
aherbst-broad: I think accessScope is ok for now.

aherbst-broad (approved):

